PR #4709, opened 2021-07-14
[unused-private-member] add logic for checking nested functions

### LINKED ISSUE #4673: False Positive on `Unused-private-member` for nested function
Looks like the `unused-private-member` warning is getting a false positive for nested functions.
E.g: `testthing.py`
``` python
""" Testing for false positive of the unused-private-member pylint test. """

class TestNest:
    """ The testing class """

    def __init__(self, in_thing):
        self.thing = False
        self.do_thing(in_thing)

    def do_thing(self, in_thing):
        """ Checks the false-positive condition, sets a property. """
        def __false_positive(in_thing):
            print(in_thing)
        __false_positive(in_thing)
        self.thing = True

    def undo_thing(self):
        """ Unsets a property. """
        self.thing = False
```
And turns out that even for test-cases I can't bring myself to get a bad pylint score.
Oh well.

``` sh
************* Module testthing
testthing.py:12:8: W0238: Unused private member `TestNest.__false_positive(in_thing)` (unused-private-member)

Your code has been rated at 9.09/10 (previous run: 9.09/10, +0.00)
```

Currently running:
``` sh
# pylint --version
pylint 2.9.1
astroid 2.6.2
Python 3.7.5 (default, Nov  7 2019, 10:50:52)
[GCC 8.3.0]
```

yushao2:
## Steps

## Description
Add logic for checking (private) nested functions in a class which was causing a false-negative as reported in 
#4673 

also, improve error message for nested functions

## Type of Changes

## Related Issue
Closes #4673

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -173,7 +173,6 @@ Release date: 2021-07-20
 
   Closes #4673
 
-
 What's New in Pylint 2.9.3?
 ===========================
 Release date: 2021-07-01
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,7 +14,7 @@ Release date: TBA
 
 * pyreverse: Show class has-a relationships inferred from the type-hint
 
-Closes #4744
+  Closes #4744
 
 * Added ``ignored-parents`` option to the design checker to ignore specific
   classes from the ``too-many-ancestors`` check (R0901).
@@ -55,6 +55,10 @@ Closes #4744
 * Added ``forgotten-debug-statement``: Emitted when ``breakpoint``, ``pdb.set_trace`` or ``sys.breakpointhook`` calls are found
 
   Closes #3692
+
+* Fix false-positive of ``unused-private-member`` when using nested functions in a class
+
+  Closes #4673
 
 
 What's New in Pylint 2.9.6?
@@ -168,10 +172,6 @@ Release date: 2021-07-20
   a faulty function in a context manager.
 
   Closes #4723
-
-* Fix false-positive of ``unused-private-member`` when using nested functions in a class
-
-  Closes #4673
 
 
 What's New in Pylint 2.9.3?
--- a/ChangeLog
+++ b/ChangeLog
@@ -173,6 +173,7 @@ Release date: 2021-07-20
 
   Closes #4673
 
+
 What's New in Pylint 2.9.3?
 ===========================
 Release date: 2021-07-01
--- a/ChangeLog
+++ b/ChangeLog
@@ -169,6 +169,10 @@ Release date: 2021-07-20
 
   Closes #4723
 
+* Fix false-positive of ``unused-private-member`` when using nested functions in a class
+
+  Closes #4673
+
 
 What's New in Pylint 2.9.3?
 ===========================
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -946,16 +946,11 @@ a metaclass class method.",
                     curr = curr.parent.scope()
 
                 outer_level_names = f"{'.'.join(reversed(name_stack))}"
-
+                function_repr = f"{outer_level_names}.{function_def.name}({function_def.args.as_string()})"
                 self.add_message(
                     "unused-private-member",
                     node=function_def,
-                    args=(
-                        node.name,
-                        f"{outer_level_names}.{function_def.name}({function_def.args.as_string()})".lstrip(
-                            "."
-                        ),
-                    ),
+                    args=(node.name, function_repr.lstrip(".")),
                 )
 
     def _check_unused_private_variables(self, node: astroid.ClassDef) -> None:
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -939,23 +939,22 @@ a metaclass class method.",
                         break
             else:
                 name_stack = []
-                outer_level_names = ""
                 curr = parent_scope
                 # Generate proper names for nested functions
                 while curr != node:
                     name_stack.append(curr.name)
                     curr = curr.parent.scope()
 
-                if name_stack:
-                    outer_level_names = f"{'.'.join(reversed(name_stack))}."
+                outer_level_names = f"{'.'.join(reversed(name_stack))}"
 
-                function_repr = f"{outer_level_names}{function_def.name}({function_def.args.as_string()})"
                 self.add_message(
                     "unused-private-member",
                     node=function_def,
                     args=(
                         node.name,
-                        function_repr,
+                        f"{outer_level_names}.{function_def.name}({function_def.args.as_string()})".lstrip(
+                            "."
+                        ),
                     ),
                 )
 
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -912,9 +912,8 @@ a metaclass class method.",
             parent_scope = function_def.parent.scope()
             if isinstance(parent_scope, astroid.FunctionDef):
                 # Handle nested functions
-                outer_fn = parent_scope
                 if function_def.name in (
-                    n.name for n in outer_fn.nodes_of_class(astroid.Name)
+                    n.name for n in parent_scope.nodes_of_class(astroid.Name)
                 ):
                     continue
             for attribute in node.nodes_of_class(astroid.Attribute):
@@ -940,19 +939,23 @@ a metaclass class method.",
                         break
             else:
                 name_stack = []
+                outer_level_names = ""
                 curr = parent_scope
                 # Generate proper names for nested functions
                 while curr != node:
                     name_stack.append(curr.name)
                     curr = curr.parent.scope()
 
-                function_repr = f"{function_def.name}({function_def.args.as_string()})"
+                if name_stack:
+                    outer_level_names = f"{'.'.join(reversed(name_stack))}."
+
+                function_repr = f"{outer_level_names}{function_def.name}({function_def.args.as_string()})"
                 self.add_message(
                     "unused-private-member",
                     node=function_def,
                     args=(
                         node.name,
-                        f"{'.'.join(reversed(name_stack))}{'.' if name_stack else ''}{function_repr}",
+                        function_repr,
                     ),
                 )
 
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -909,12 +909,13 @@ a metaclass class method.",
             function_def = cast(astroid.FunctionDef, function_def)
             if not is_attr_private(function_def.name):
                 continue
-            if isinstance(function_def.parent.scope(), astroid.FunctionDef):
+            parent_scope = function_def.parent.scope()
+            if isinstance(parent_scope, astroid.FunctionDef):
                 # Handle nested functions
-                outer_fn = function_def.parent.scope()
-                if function_def.name in [
+                outer_fn = parent_scope
+                if function_def.name in (
                     n.name for n in outer_fn.nodes_of_class(astroid.Name)
-                ]:
+                ):
                     continue
             for attribute in node.nodes_of_class(astroid.Attribute):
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas (approved):
👍

pylint/checkers/classes.py
```diff
@@ -908,6 +908,13 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
             function_def = cast(astroid.FunctionDef, function_def)
             if not is_attr_private(function_def.name):
                 continue
+            if isinstance(function_def.parent.scope(), astroid.FunctionDef):
+                # Handle nested functions
+                outer_fn = function_def.parent.scope()
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Nvm, I think mypy can infer it from the condition just above.

pylint/checkers/classes.py
```diff
@@ -908,6 +908,13 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
             function_def = cast(astroid.FunctionDef, function_def)
             if not is_attr_private(function_def.name):
                 continue
+            if isinstance(function_def.parent.scope(), astroid.FunctionDef):
```
cdce8p, line 911: If you move `outer_fn`, you can reuse it here. Would `parent_scope` be a more fitting name?

pylint/checkers/classes.py
```diff
@@ -908,6 +908,13 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
...
+                # Handle nested functions
+                outer_fn = function_def.parent.scope()
+                if function_def.name in [
+                    n.name for n in outer_fn.nodes_of_class(astroid.Name)
+                ]:
```
cdce8p, line 916:
```suggestion
                if function_def.name in (
                    n.name for n in outer_fn.nodes_of_class(astroid.Name)
                ):
```
You should be able to use a generator here

pylint/checkers/classes.py
```diff
@@ -930,11 +937,18 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
                     ):
                         break
             else:
+                name_stack = ""
+                curr = function_def.parent.scope()
```
cdce8p, line 941: Can you reuse `outer_fn` / `parent_scope` here?

pylint/checkers/classes.py
```diff
@@ -930,11 +937,18 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
...
+                # Generate proper names for nested functions
+                while curr != node:
+                    name_stack = f"{curr.name}." + name_stack
```
cdce8p, line 944: I don't like the string concatenation here. It would be better to add each name to a list and then do something like `".".join(reversed(name_stack))`

tests/functional/u/unused/unused_private_member.py
```diff
@@ -135,3 +135,44 @@ def __init__(self):
     def attr_c(self):
         """Get c."""
         return cls.__attr_c  # [undefined-variable]
+
+# Test cases for false-positive reported in #4673
```
cdce8p, line 139: Can you add an additional empty line here?

cdce8p (changes requested):

pylint/checkers/classes.py
yushao2: pardon for my lack of experience here, but the suggestion is preferred because it's immutable right?

@cdce8p

cdce8p: Not exactly. Here it's the difference between a list comprehension and a generator expression. The later can be slightly faster in some cases as it doesn't need to create a new list before iterating. Especially since `nodes_of_class` also returns a generator.

Immutability would be the case if it were normal list / tuple. Then a tuple would be preferred. (There is a pylint check for it in the CodeStyle extension).

tests/functional/u/unused/unused_private_member.py
yushao2: This is weird, I've added one but it doesn't seem to show up in the ChangeLog

yushao2:
```suggestion

# Test cases for false-positive reported in #4673
```

yushao2: Ah okay, seems okay now if I look at the files changed view

cdce8p: Sometimes empty lines get added to the end of a section in diff view

pylint/checkers/classes.py
```diff
@@ -908,6 +908,14 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
             function_def = cast(astroid.FunctionDef, function_def)
             if not is_attr_private(function_def.name):
                 continue
+            parent_scope = function_def.parent.scope()
+            if isinstance(parent_scope, astroid.FunctionDef):
+                # Handle nested functions
+                outer_fn = parent_scope
```
cdce8p, line 914: I'm wondering if the reassignment is necessary

ChangeLog
```diff
@@ -67,6 +71,7 @@ Release date: TBA
   Closes #4668
 
 
+
```
cdce8p, line 74:
```suggestion
```
Empty line :)

pylint/checkers/classes.py
```diff
@@ -930,11 +938,21 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
...
+                    curr = curr.parent.scope()
+
                 function_repr = f"{function_def.name}({function_def.args.as_string()})"
                 self.add_message(
                     "unused-private-member",
                     node=function_def,
-                    args=(node.name, function_repr),
+                    args=(
+                        node.name,
+                        f"{'.'.join(reversed(name_stack))}{'.' if name_stack else ''}{function_repr}",
```
cdce8p, line 954: This is quite difficult to read. You already define `function_repr` earlier, why not do the same for the first part?

cdce8p (changes requested):
Some last comments

yushao2:
@cdce8p busy week for me, so sorry for the lack of response. Have updated my PR based on comments :)

ChangeLog
```diff
@@ -120,6 +120,10 @@ Release date: 2021-07-20
 
   Closes #4723
 
+* Fix false-positive of ``unused-private-member`` when using nested functions in a class
```
Pierre-Sassoulas, line 123: Could you move this to 2.9.6, please :) ?

pylint/checkers/classes.py
```diff
@@ -931,11 +938,25 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
...
+                    outer_level_names = f"{'.'.join(reversed(name_stack))}."
+
+                function_repr = f"{outer_level_names}{function_def.name}({function_def.args.as_string()})"
                 self.add_message(
                     "unused-private-member",
                     node=function_def,
-                    args=(node.name, function_repr),
+                    args=(
+                        node.name,
+                        function_repr,
```
Pierre-Sassoulas, line 958: You can remove the comma at the end so black keep the code like it was before.
```suggestion
                        function_repr
```

Pierre-Sassoulas (changes requested):

pylint/checkers/classes.py
```diff
@@ -931,11 +938,25 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
...
+                name_stack = []
+                outer_level_names = ""
+                curr = parent_scope
+                # Generate proper names for nested functions
+                while curr != node:
+                    name_stack.append(curr.name)
+                    curr = curr.parent.scope()
+
+                if name_stack:
+                    outer_level_names = f"{'.'.join(reversed(name_stack))}."
```
cdce8p, line 950: You don't need to guard this with an `IF`. Join with an empty list will just return an empty string.
```py
'.'.join(reversed([]))  # -> ""
```

pylint/checkers/classes.py
```diff
@@ -931,11 +938,25 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
                     ):
                         break
             else:
-                function_repr = f"{function_def.name}({function_def.args.as_string()})"
+                name_stack = []
+                outer_level_names = ""
```
cdce8p, line 942: With that, you don't need this line anymore.

cdce8p (commented):
@yushao2 Don't worry about it! Just let us know if we can help.

Left more more comment, but after that it will be good for merging.

pylint/checkers/classes.py
```diff
@@ -939,23 +939,22 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
...
                 # Generate proper names for nested functions
                 while curr != node:
                     name_stack.append(curr.name)
                     curr = curr.parent.scope()
 
-                if name_stack:
-                    outer_level_names = f"{'.'.join(reversed(name_stack))}."
+                outer_level_names = f"{'.'.join(reversed(name_stack))}"
 
-                function_repr = f"{outer_level_names}{function_def.name}({function_def.args.as_string()})"
```
cdce8p: I really liked this line. It helped make the code more readable. In contrast to having a long f-string as an argument.

pylint/checkers/classes.py
```diff
@@ -939,23 +939,22 @@ def _check_unused_private_functions(self, node: astroid.ClassDef) -> None:
...
                 self.add_message(
                     "unused-private-member",
                     node=function_def,
                     args=(
                         node.name,
-                        function_repr,
+                        f"{outer_level_names}.{function_def.name}({function_def.args.as_string()})".lstrip(
+                            "."
+                        ),
```
cdce8p, line 957: Why did you add the `lstring(".")`? `'.'.join` will never yield a single `.` if that was your intention.

```py
>>> ".".join([])
''
```

cdce8p (commented):
Almost there

cdce8p:
@yushao2 Did you see my latest comments?

pylint/checkers/classes.py
yushao2: > Why did you add the `lstring(".")`? `'.'.join` will never yield a single `.` if that was your intention.
> 
> ```python
> >>> ".".join([])
> ''
> ```

This is necessary due to the dot between `outer_level_names` and `function_def.name`

it's true that the join would yield an empty string, but this would result in error messages showing something like

`.Class.my_var` -- therefore i added the lstrip here

yushao2:
> @yushao2 Did you see my latest comments?



> @yushao2 Did you see my latest comments?

oops, did i miss out anything? I only saw the latest 2 comments on the lstrip, and to undo the removal of the temp variable that's holding the function's representation

pylint/checkers/classes.py
cdce8p: Makes sense. Thanks for the explanation!

cdce8p (approved):
LGTM

cdce8p:
> oops, did i miss out anything? I only saw the latest 2 comments on the lstrip, and to undo the removal of the temp variable that's holding the function's representation

No, is fine now! After you forced-pushed some changes, I wasn't sure if you had seen my comment. But apparently you have 😉 Anyway, thank you for working on it!

yushao2:
> LGTM

thanks! opened #4782 and #4783 earlier too related to the same checker for your review :)

yushao2:
> > oops, did i miss out anything? I only saw the latest 2 comments on the lstrip, and to undo the removal of the temp variable that's holding the function's representation
> 
> No, is fine now! After you forced-pushed some changes, I wasn't sure if you had seen my comment. But apparently you have wink Anyway, thank you for working on it!

yeah my personal dev machine is a little messy due to the massive changes in the codebase so have been rebasing my work and been typing `-f` more than i should have hahaha

Issue #4296, opened 2021-04-05
pylint's return code is still 0 when analysed code contains useless-suppressions


MarkCBell:
### Steps to reproduce

Given a file `a.py`:
```
def foo(x):  # pylint: disable=unused-argument
    return x+1
```

### Current behavior

Result of `pylint a.py --disable=all --enable=useless-suppression`:
```
a.py:1:0: I0021: Useless suppression of 'unused-argument' (useless-suppression)
```
however the return code of this is command `0`. This makes it much more difficult create CI/CD pipelines that check whether there are unneeded suppressions, for example.

This appears to be because of the fact that even though a message is added to the reporter, the [MSG_TYPES_STATUS](https://github.com/PyCQA/pylint/blob/master/pylint/constants.py#L35) of info messages is 0. Hence when this message is parsed, msg_status is or'ed with 0 and so the return code of the command is still 0.

### Expected behavior

The return code should be non-zero since a check which was enabled has failed.

There are a number of different ways to achieve this including:
1. Change `MSG_TYPES_STATUS['I']` to a non-zero number so that raising any `info` messages results in a non-zero exit code
2. Change the `useless-suppression` from type `info` to be type `warning`

### pylint --version output

Result of `pylint --version` output:
```
pylint 2.7.4
astroid 2.5.2
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0]
```

dbaty:
In the meantime, you may be interested in the [pylint-strict-informational plugin](https://github.com/Polyconseil/pylint-strict-informational/) which, once installed, transforms informational messages as warnings.

TylerYep:
Though 1) sounds fine, I really hope 2) is eventually implemented, even though it's a breaking change. Enabling `useless-suppression` by default helps clean up repositories considerably as Pylint improves, and matches the behavior of tools like mypy or flake8+yesqa.

papb:
@dbaty That plugin did not work for me unfortunately, should it still work with latest pylint?

papb:
@TylerYep @dbaty I managed to make another workaround, see https://github.com/Polyconseil/pylint-strict-informational/issues/2

Issue #3363, opened 2020-01-24
Add --strict option to fail on informational messages


xavfernandez:
### Is your feature request related to a problem? Please describe
A bunch of useful informational messages exist in pylint but enforcing them in CI is tricky since their presence doesn't affect the exit code.

### Describe the solution you'd like

The activation of this `--strict` (or `--fail-on-informational`) option would simply change
`MSG_TYPES_STATUS['I']` value to a none-zero value to be determined (I'd go with 1)

https://github.com/PyCQA/pylint/blob/f2f4e6f42416644471ab003d4df7ecf052c3c411/pylint/constants.py#L33

to make pylint fail when an informational message is produced.

### Additional context
This could be an alternative to #3251

xavfernandez:
In the meantime, a fake plugin [pylint-strict-informational](https://pypi.org/project/pylint-strict-informational) that does exactly that has been published.

To use, `pip install pylint-strict-informational` then use `pylint --load-plugins=pylint_strict_informational` (or set it in your configuration file).
(From an original idea of @dbaty).

PCManticore:
A strict mode could make sense for other checks as well. It's definitely something we should consider.

papb:
@xavfernandez Unfortunately `pylint-strict-informational` did not work for me. I ended up with the following workaround (executed before calling `pylint`):

```bash
find .venv/lib -path '*/pylint/constants.py' -exec sed -r -i -e 's@(^MSG_TYPES_STATUS = .*):\s*0,(.*)@\1: 4,\2  # monkey-patched, waiting for pylint#3363@g' {} ';'
```

MarkCBell:
Note, since PyCQA/pylint@ab094ae, it is not just enough to change the `MSG_TYPES_STATUS` of the informational type messages. This is because this commit means that a shortcut `exit(0)` is applied whenever a run completes with a score >= `config.fail_under` (which is 10.0 by default). Since informational messages do not affect the score, if only informational messages are raised during a run then the run completes with a perfect score and so this shortcut is applied and the job exits with return code `0` regardless of the value of `self.linter.msg_status`.

papb:
Hello @Pierre-Sassoulas and everyone, sorry to take long, but: why was this issue closed? PR #4713 does not seem to add a flag to allow failing on any informational message.

At least it seems I can now achieve it by manually adding `--fail-on` for all existing informational messages, but I don't like that. Thoughts?

papb:
[...]

PR #4713, opened 2021-07-15
Ensure --fail-on always returns a non-zero exit code when issue is matched

MarkCBell:
## Steps

- [x] If it's a new feature or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.

## Description

This fixes a bug in which pylint can still return a zero exit code when using --fail-on even if the specified issue occurs when the issues matched are all of info type.

Without this patch, running: `pylint --fail-on=useless-suppression a.py` when `a.py` contains:
```
"""Test file to trigger an useless-suppression error msg"""
# pylint: disable=broad-except
pass
```
returns exit code 0 despite matching the useless-suppression (info) issue.

## Type of Changes

## Related Issue

This is related to Issue #4296 and #3363. However although it will provides some other options for these issues, I don't think it will fully resolve them.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -19,6 +19,9 @@ Release date: TBA
 
 * Fix bug in which --fail-on can return a zero exit code even when the specified issue is present
 
+  Closes #4296 
+  Closes #3363 
+
 * Fix hard failure when handling missing attribute in a class with duplicated bases
 
   Closes #4687
--- a/ChangeLog
+++ b/ChangeLog
@@ -19,8 +19,8 @@ Release date: TBA
 
 * Fix bug in which --fail-on can return a zero exit code even when the specified issue is present
 
-  Closes #4296 
-  Closes #3363 
+  Closes #4296
+  Closes #3363
 
 * Fix hard failure when handling missing attribute in a class with duplicated bases
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,8 @@ Release date: TBA
 ..
   Put bug fixes that should not wait for a new minor version here
 
+* Fix bug in which --fail-on can return a zero exit code even when the specified issue is present
+
 * Fix hard failure when handling missing attribute in a class with duplicated bases
 
   Closes #4687
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -512,3 +512,5 @@ contributors:
 * Daniel Dorani (doranid): contributor
 
 * Will Shanks: contributor
+
+* Mark Bell: contributor
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -707,7 +707,9 @@ class PyLinter(
                         self.fail_on_symbols.append(msg.symbol)
 
     def any_fail_on_issues(self):
-        return any(x in self.fail_on_symbols for x in self.stats["by_msg"])
+        return self.stats is not None and any(
+            x in self.fail_on_symbols for x in self.stats["by_msg"]
+        )
 
     def disable_noerror_messages(self):
         for msgcat, msgids in self.msgs_store._msgs_by_category.items():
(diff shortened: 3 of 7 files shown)


Pierre-Sassoulas (changes requested):
Thank you for the changes, thjis look reasonable ! Could you add integration tests for this, please ?  I'm thinking of something like:
```python

@pytest.mark.parametrize("args,expected", [["fail-under", 0], ...])
def test_exit_code(args, expected):
    with pytest.raises(SystemExit as e:
        with patch("sys.argv", ["pylint", "tests/functional/...",] + args):
             ...

**Comment by MarkCBell** [comment]
@Pierre-Sassoulas I've added some test and I think what I've done works (the tests implement the correct behaviour, fail without the functionality patch and pass with the functionality patch) but if you have any suggestions for how to refine the structure of the tests or change where they should be located then please let me know.

File: ChangeLog [hunk 1]
```diff
@@ -17,6 +17,8 @@ Release date: TBA
 ..
   Put bug fixes that should not wait for a new minor version here
 
+* Fix bug in which --fail-on can return a zero exit code even when the specified issue is present
```
Pierre-Sassoulas, line 20:
```suggestion
* Fix bug in which --fail-on can return a zero exit code even when the specified issue is present

  Closes #4296 
  Closes #3363 
```

Pierre-Sassoulas (approved):
Nice jobs, the tests are on point, I don't see any forgotten use cases, but if they exists it will be really easy to add them later on. I think this fix the issues you linked, we're not going to add an option for failing on information messages, it makes more sense has the default behavior. Do yo agree ?

Pierre-Sassoulas:
I'm putting this in 2.10 because it'll feel like a breaking change is a pipeline start to fail because there were informational messages. Could you also add an explaining text in 2.10's what's new, please ?

MarkCBell:
So I understand there is a need for caution around changes that it may change the behaviour of pipeline. However in order to have a pipeline that is currently succeeding become failing because of this patch, I think the following conditions would need to be met:
 - pylint is run without the `--exit-zero` flag
 - pylint generates only informational messages
 - at least one of the informational messages that is generated is provided as an argument to the `--fail-on` flag

So this means that although a developer has explicitly set `--fail-on=i0011` (for example) and that message is being generated, their pipeline is still (I would say incorrectly) passing.

I would suggest that this should just be treated as a bug in the implementation of `fail-on` which despite it's documentation:
```
    --fail-on=<msg ids>
                        Return non-zero exit code if any of these
                        messages/categories are detected, even if score is
                        above --fail-under value. Syntax same as enable.
```
can currently return a zero exit code even when a specified (info) message is detected. In which case I don't think that this would need to wait for 2.10. Does that sound reasonable to you or have I missed something?

Pierre-Sassoulas:
Ho, I did not notice the ``fail-on`` requirement. Shouldn't pylint exit with an error code if there is an enabled informational message even without the fail-on option ? If I understand correctly we could merge this as it is, and make the exit code change for 2.10 later ?

MarkCBell:
I'm sorry but I don't think I understand your last comment. Suppose we were to merge this PR exactly as it is right now into 2.9.4, what exactly is "the exit code change" that you are proposing could be done separately / later for 2.10?

Pierre-Sassoulas:
Sorry for being unclear. Considering that we now have 3 conditions:
- pylint is run without the --exit-zero flag
- pylint generates only informational messages
- at least one of the informational messages that is generated is provided as an argument to the --fail-on flag

I'd want to remove that last condition regarding the ``--fail-on flag`` later on.

MarkCBell:
Ok, I think I understand now. In which case I suggest that the right thing to do is to merge this PR into 2.9.4 as a bug fix to make the implementation of `--fail-on` match how it is described in the documentation. As you suggest, breaking changes can be made as part of 2.10 later and hopefully that should be very straightforward based on the structure and tests provided by the PR.

PR #4720, opened 2021-07-19
Pylint fix for invalid TOML config

### LINKED ISSUE #4580: Pylint Crash on invalid TOML config
### Steps to reproduce

It was not immediately clear to me that the main table should be `[tool.pylint.master]` and not `[tool.pylint]` (I searched in the docs but did not find an example).

Given any python file and a config file called `repro.toml` (I initially encountered this on my `pyproject.toml`, but for the sake of a simple example):

```toml
[tool.pylint]
load-plugins = []
```

### Current behavior

```
pylint --rcfile=repro.toml repro.py       
  File "C:\Program Files\Python310\lib\runpy.py", line 196, in _run_module_as_main
  File "C:\Program Files\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\...\venv\Scripts\pylint.exe\__main__.py", line 7, in <module>
  File "d:\...\venv\lib\site-packages\pylint\__init__.py", line 24, in run_pylint
    PylintRun(sys.argv[1:])
  File "d:\...\venv\lib\site-packages\pylint\lint\run.py", line 316, in __init__
    linter.read_config_file(verbose=self.verbose)
  File "d:\...\venv\lib\site-packages\pylint\config\option_manager_mixin.py", line 281, in read_config_file
    for option, value in values.items():
AttributeError: 'str' object has no attribute 'items'
```

### Expected behavior

It should not crash.

### pylint --version output

I just upgraded to prerelease versions as suggested in this template.

```
pylint 3.0.0a3
astroid 2.5.8
Python 3.10.0b1 (tags/v3.10.0b1:ba42175, May  3 2021, 20:22:30) [MSC v.1928 64 bit (AMD64)]
```

### Additional crashes for invalid data

Additional crashes are possible for "mapping" configs where I thought a TOML table might be used rather than a list of strings with a built in separator as is used in `.pylintrc` (again, I could not find examples).

Given config:

```toml
[tool.pylint.imports]
preferred-modules = { "a"="b" }
```

Or config:

```toml
[tool.pylint.basic]
name-group = { "a"="b" }
```

Results in (same traceback for both configs):

```
pylint --rcfile=repro.toml repro.py
  File "C:\Program Files\Python310\lib\runpy.py", line 196, in _run_module_as_main
  File "C:\Program Files\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\...\venv\Scripts\pylint.exe\__main__.py", line 7, in <module>
  File "d:\...\venv\lib\site-packages\pylint\__init__.py", line 24, in run_pylint
    PylintRun(sys.argv[1:])
  File "d:\...\venv\lib\site-packages\pylint\lint\run.py", line 333, in __init__
    linter.load_config_file()
  File "d:\...\venv\lib\site-packages\pylint\config\option_manager_mixin.py", line 313, in load_config_file
    for option, value in parser.items(section):
  File "C:\Program Files\Python310\lib\configparser.py", line 860, in items
    return [(option, value_getter(option)) for option in orig_keys]
  File "C:\Program Files\Python310\lib\configparser.py", line 860, in <listcomp>
    return [(option, value_getter(option)) for option in orig_keys]
  File "C:\Program Files\Python310\lib\configparser.py", line 856, in <lambda>
    value_getter = lambda option: self._interpolation.before_get(self,
  File "C:\Program Files\Python310\lib\configparser.py", line 395, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "C:\Program Files\Python310\lib\configparser.py", line 412, in _interpolate_some
    p = rest.find("%")
AttributeError: 'DynamicInlineTableDict' object has no attribute 'find'
```

Related issues: #4518 #4204

tanvimoharir:
## Steps

## Description
This is a fix for handling bad or invalid toml configuration in pyproject.toml

## Type of Changes

## Related Issue

Closes #4580

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -153,6 +153,14 @@ Release date: TBA
 
   Closes #5261
 
+* Crashes when a list is encountered in a toml configuration do not happen anymore. A new
+  ``bad-configuration-option-value`` checker was added that will emit for these problems.
+  Some bad configurations might still silently not be taken into account which is tracked in
+  a follow-up issue.
+
+  Closes #4580
+  Follow-up in #5259
+
 * Added new checker ``useless-with-lock`` to find incorrect usage of with statement and threading module locks.
   Emitted when ``with threading.Lock():`` is used instead of ``with lock_instance:``.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -157,7 +157,7 @@ Release date: TBA
 
   Closes #4580
 
-* A new ``bad-configuration-option-value`` checker was added that will emit for misplaced option
+* A new ``bad-configuration-section`` checker was added that will emit for misplaced option
   in pylint's top level namespace for toml configuration. Top-level dictionaries or option defined
   in the wrong section will still silently not be taken into account, which is tracked in a
   follow-up issue.
--- a/ChangeLog
+++ b/ChangeLog
@@ -153,12 +153,15 @@ Release date: TBA
 
   Closes #5261
 
-* Crashes when a list is encountered in a toml configuration do not happen anymore. A new
-  ``bad-configuration-option-value`` checker was added that will emit for these problems.
-  Some bad configurations might still silently not be taken into account which is tracked in
-  a follow-up issue.
+* Crashes when a list is encountered in a toml configuration do not happen anymore.
 
   Closes #4580
+
+* A new ``bad-configuration-option-value`` checker was added that will emit for misplaced option
+  in pylint's top level namespace for toml configuration. Top-level dictionaries or option defined
+  in the wrong section will still silently not be taken into account, which is tracked in a
+  follow-up issue.
+
   Follow-up in #5259
 
 * Added new checker ``useless-with-lock`` to find incorrect usage of with statement and threading module locks.
--- a/doc/whatsnew/2.12.rst
+++ b/doc/whatsnew/2.12.rst
@@ -61,6 +61,8 @@ New checkers
   in the wrong section will still silently not be taken into account, which is tracked in a
   follow-up issue.
 
+  Follow-up in #5259
+
 Removed checkers
 ================
 
--- a/doc/whatsnew/2.12.rst
+++ b/doc/whatsnew/2.12.rst
@@ -56,7 +56,7 @@ New checkers
 
   Closes #5208
 
-* A new ``bad-configuration-option-value`` checker was added that will emit for misplaced option
+* A new ``bad-configuration-section`` checker was added that will emit for misplaced option
   in pylint's top level namespace for toml configuration. Top-level dictionaries or option defined
   in the wrong section will still silently not be taken into account, which is tracked in a
   follow-up issue.
--- a/doc/whatsnew/2.12.rst
+++ b/doc/whatsnew/2.12.rst
@@ -56,6 +56,11 @@ New checkers
 
   Closes #5208
 
+* A new ``bad-configuration-option-value`` checker was added that will emit for misplaced option
+  in pylint's top level namespace for toml configuration. Top-level dictionaries or option defined
+  in the wrong section will still silently not be taken into account, which is tracked in a
+  follow-up issue.
+
 Removed checkers
 ================
 
@@ -156,5 +161,9 @@ Other Changes
   have functioning shared semaphore implementation.
 
   Closes #5216
+
+* Crashes when a list is encountered in a toml configuration do not happen anymore.
+
+  Closes #4580
 
 * Make yn validator case insensitive, to allow for ``True`` and ``False`` in config files.
--- a/pylint/config/option_manager_mixin.py
+++ b/pylint/config/option_manager_mixin.py
@@ -310,7 +310,7 @@ class OptionsManagerMixIn:
             if not isinstance(values, dict):
                 # This class is a mixin: add_message comes from the `PyLinter` class
                 self.add_message(  # type: ignore
-                    "bad-configuration-option-value", line=0, args=(section, values)
+                    "bad-configuration-section", line=0, args=(section, values)
                 )
                 continue
             for option, value in values.items():
--- a/pylint/config/option_manager_mixin.py
+++ b/pylint/config/option_manager_mixin.py
@@ -308,7 +308,7 @@ class OptionsManagerMixIn:
             # TOML has rich types, convert values to
             # strings as ConfigParser expects.
             if not isinstance(values, dict):
-                # This class is a mixin the add_message come from the pylinter
+                # This class is a mixin: add_message comes from the `PyLinter` class
                 self.add_message(  # type: ignore
                     "bad-configuration-option-value", line=0, args=(section, values)
                 )
@@ -332,8 +332,7 @@ class OptionsManagerMixIn:
         options provider)"""
         parser = self.cfgfile_parser
         for section in parser.sections():
-            items = parser.items(section)
-            for option, value in items:
+            for option, value in parser.items(section):
                 try:
                     self.global_set_option(option, value)
                 except (KeyError, optparse.OptionError):
--- a/pylint/config/option_manager_mixin.py
+++ b/pylint/config/option_manager_mixin.py
@@ -293,9 +293,8 @@ class OptionsManagerMixIn:
             msg = "No config file found, using default configuration"
         print(msg, file=sys.stderr)
 
-    @staticmethod
     def _parse_toml(
-        config_file: Union[Path, str], parser: configparser.ConfigParser
+        self, config_file: Union[Path, str], parser: configparser.ConfigParser
     ) -> None:
         """Parse and handle errors of a toml configuration file."""
         with open(config_file, encoding="utf-8") as fp:
@@ -305,23 +304,36 @@ class OptionsManagerMixIn:
         except KeyError:
             return
         for section, values in sections_values.items():
+            section_name = section.upper()
             # TOML has rich types, convert values to
             # strings as ConfigParser expects.
+            if not isinstance(values, dict):
+                # This class is a mixin the add_message come from the pylinter
+                self.add_message(  # type: ignore
+                    "bad-configuration-option-value", line=0, args=(section, values)
+                )
+                continue
             for option, value in values.items():
                 if isinstance(value, bool):
                     values[option] = "yes" if value else "no"
-                elif isinstance(value, (int, float)):
-                    values[option] = str(value)
                 elif isinstance(value, list):
                     values[option] = ",".join(value)
-            parser._sections[section.upper()] = values  # type: ignore
+                else:
+                    values[option] = str(value)
+            for option, value in values.items():
+                try:
+                    parser.set(section_name, option, value=value)
+                except configparser.NoSectionError:
+                    parser.add_section(section_name)
+                    parser.set(section_name, option, value=value)
 
     def load_config_file(self):
         """Dispatch values previously read from a configuration file to each
         options provider)"""
         parser = self.cfgfile_parser
         for section in parser.sections():
-            for option, value in parser.items(section):
+            items = parser.items(section)
+            for option, value in items:
                 try:
                     self.global_set_option(option, value)
                 except (KeyError, optparse.OptionError):
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -160,8 +160,8 @@ MSGS = {
     ),
     "E0014": (
         "Out-of-place setting encountered in top level configuration-section '%s' : '%s'",
-        "bad-configuration-option-value",
-        "Used when we detect a setting in the top level of a toml configuration that shouldn't be there.
+        "bad-configuration-section",
+        "Used when we detect a setting in the top level of a toml configuration that shouldn't be there.",
     ),
 }
 
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -159,10 +159,9 @@ MSGS = {
         "Used when a bad value is used in 'load-plugins'.",
     ),
     "E0014": (
-        "Incorrect setting encountered in top level configuration-section '%s' : '%s'",
+        "Out-of-place setting encountered in top level configuration-section '%s' : '%s'",
         "bad-configuration-option-value",
-        "Used when a top section value can't be parsed because it does not belong to the top section "
-        "or has an incorrect value type.",
+        "Used when we detect a setting in the top level of a toml configuration that shouldn't be there.
     ),
 }
 
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -158,6 +158,12 @@ MSGS = {
         "bad-plugin-value",
         "Used when a bad value is used in 'load-plugins'.",
     ),
+    "E0014": (
+        "Incorrect setting encountered in top level configuration-section '%s' : '%s'",
+        "bad-configuration-option-value",
+        "Used when a top section value can't be parsed because it does not belong to the top section "
+        "or has an incorrect value type.",
+    ),
 }
 
 
--- a/tests/config/functional/toml/issue_4580/invalid_data_for_import.toml
+++ b/tests/config/functional/toml/issue_4580/invalid_data_for_import.toml
@@ -1,0 +1,7 @@
+# Both disable and load-plugins do not belong in the imports section
+[tool.pylint."imports"]
+disable = [
+    "logging-not-lazy",
+    "logging-format-interpolation",
+]
+preferred-modules = { "a"="b" }
--- a/tests/config/functional/toml/issue_4580/invalid_data_for_import.toml
+++ b/tests/config/functional/toml/issue_4580/invalid_data_for_import.toml
@@ -1,4 +1,5 @@
 # Both disable and load-plugins do not belong in the imports section
+# TODO This should be fixed in #5259
 [tool.pylint."imports"]
 disable = [
     "logging-not-lazy",
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -529,6 +529,9 @@ contributors:
 
 * Marcin Kurczewski (rr-): contributor
 
+* Tanvi Moharir: contributor
+  - Fix for invalid toml config
+
 * Eisuke Kawashima (e-kwsm): contributor
 
 * Daniel van Noord (DanielNoord): contributor
--- a/tests/config/functional/toml/issue_4580/correct_basic_name_group.toml
+++ b/tests/config/functional/toml/issue_4580/correct_basic_name_group.toml
@@ -1,0 +1,2 @@
+[tool.pylint.basic]
+name-group = { "a"="b" }
--- a/tests/config/functional/toml/issue_4580/correct_import_preferred_module.toml
+++ b/tests/config/functional/toml/issue_4580/correct_import_preferred_module.toml
@@ -1,0 +1,2 @@
+[tool.pylint.imports]
+preferred-modules = { "a"="b" }
--- a/tests/config/functional/toml/issue_4580/empty_list.out
+++ b/tests/config/functional/toml/issue_4580/empty_list.out
@@ -1,2 +1,2 @@
 ************* Module {abspath}
-{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'load-plugins' : '[]' (bad-configuration-option-value)
+{relpath}:1:0: E0014: Out-of-place setting encountered in top level configuration-section 'load-plugins' : '[]' (bad-configuration-section)
--- a/tests/config/functional/toml/issue_4580/empty_list.out
+++ b/tests/config/functional/toml/issue_4580/empty_list.out
@@ -1,0 +1,2 @@
+************* Module {abspath}
+{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'load-plugins' : '[]' (bad-configuration-option-value)
--- a/tests/config/functional/toml/issue_4580/empty_list.toml
+++ b/tests/config/functional/toml/issue_4580/empty_list.toml
@@ -1,0 +1,3 @@
+[tool.pylint]
+# load-plugins does not belong in top level section
+load-plugins = []
--- a/tests/config/functional/toml/issue_4580/invalid_data_for_basic.out
+++ b/tests/config/functional/toml/issue_4580/invalid_data_for_basic.out
@@ -1,3 +1,3 @@
 ************* Module {abspath}
-{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'load-plugins' : '[]' (bad-configuration-option-value)
-{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'disable' : 'logging-not-lazy,logging-format-interpolation' (bad-configuration-option-value)
+{relpath}:1:0: E0014: Out-of-place setting encountered in top level configuration-section 'load-plugins' : '[]' (bad-configuration-section)
+{relpath}:1:0: E0014: Out-of-place setting encountered in top level configuration-section 'disable' : 'logging-not-lazy,logging-format-interpolation' (bad-configuration-section)
--- a/tests/config/functional/toml/issue_4580/invalid_data_for_basic.out
+++ b/tests/config/functional/toml/issue_4580/invalid_data_for_basic.out
@@ -1,0 +1,3 @@
+************* Module {abspath}
+{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'load-plugins' : '[]' (bad-configuration-option-value)
+{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'disable' : 'logging-not-lazy,logging-format-interpolation' (bad-configuration-option-value)
--- a/tests/config/functional/toml/issue_4580/invalid_data_for_basic.toml
+++ b/tests/config/functional/toml/issue_4580/invalid_data_for_basic.toml
@@ -1,0 +1,4 @@
+[tool.pylint]
+# Both disable and load-plugins do not belong in the top level section
+load-plugins = []
+disable = "logging-not-lazy,logging-format-interpolation"
--- a/tests/config/functional/toml/issue_4580/invalid_data_for_import.result.json
+++ b/tests/config/functional/toml/issue_4580/invalid_data_for_import.result.json
@@ -1,0 +1,5 @@
+{
+  "functional_append": {
+    "disable": [["logging-not-lazy"], ["logging-format-interpolation"]]
+  }
+}
--- a/tests/config/functional/toml/issue_4580/rich_types.result.json
+++ b/tests/config/functional/toml/issue_4580/rich_types.result.json
@@ -1,0 +1,7 @@
+{
+  "functional_append": {
+    "disable": [["logging-not-lazy"], ["logging-format-interpolation"]]
+  },
+  "jobs": 10,
+  "reports": true
+}
--- a/tests/config/functional/toml/issue_4580/rich_types.toml
+++ b/tests/config/functional/toml/issue_4580/rich_types.toml
@@ -1,0 +1,7 @@
+[tool.pylint."messages control"]
+disable = [
+    "logging-not-lazy",
+    "logging-format-interpolation",
+]
+jobs = 10
+reports = true
--- a/tests/config/functional/toml/issue_4580/top_level_disable.out
+++ b/tests/config/functional/toml/issue_4580/top_level_disable.out
@@ -1,0 +1,2 @@
+************* Module {abspath}
+{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'disable' : 'logging-not-lazy,logging-format-interpolation' (bad-configuration-option-value)
--- a/tests/config/functional/toml/issue_4580/top_level_disable.out
+++ b/tests/config/functional/toml/issue_4580/top_level_disable.out
@@ -1,2 +1,2 @@
 ************* Module {abspath}
-{relpath}:1:0: E0014: Incorrect setting encountered in top level configuration-section 'disable' : 'logging-not-lazy,logging-format-interpolation' (bad-configuration-option-value)
+{relpath}:1:0: E0014: Out-of-place setting encountered in top level configuration-section 'disable' : 'logging-not-lazy,logging-format-interpolation' (bad-configuration-section)
--- a/tests/config/functional/toml/issue_4580/top_level_disable.toml
+++ b/tests/config/functional/toml/issue_4580/top_level_disable.toml
@@ -1,0 +1,3 @@
+[tool.pylint]
+# disable does not belong in top level section
+disable = "logging-not-lazy,logging-format-interpolation"
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -2,6 +2,12 @@ import os
 from pathlib import Path
 from typing import Optional, Set
 
+<<<<<<< HEAD
+=======
+import pytest
+
+import pylint.lint
+>>>>>>> Adding check for dict
 from pylint.lint.run import Run
 from pylint.testutils.configuration_test import run_using_a_configuration_file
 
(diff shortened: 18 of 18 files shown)


Pierre-Sassoulas (commented):
Could you add a test with the failing toml, please ? :)

tanvimoharir:
> Could you add a test with the failing toml, please ? :)

Yes I will do that, I also need to accommodate the other failed case which was `AttributeError: 'DynamicInlineTableDict' object has no attribute 'find' `

tanvimoharir:
@Pierre-Sassoulas I added three tests, one for empty list and two more for the other two cases mentioned in #4580 
However for the case for `AttributeError: 'DynamicInlineTableDict' object has no attribute 'find'` its still crashing.

tests/test_config.py
```diff
@@ -91,3 +93,45 @@ def test_can_read_toml_rich_types(tmp_path):
...
+
+def test_toml_with_invalid_data_for_basic(tmp_path):
+    # This would test config with invalid data for basic section
+    # refer #4580
+    config_file = tmp_path / "pyproject.toml"
+    config_file.write_text(
+        """
+[tool.pylint."basic"]
+name-group = { "a"="b" }
+"""
```
tanvimoharir, line 134: Per my other comment these two cases are still crashing.

Pierre-Sassoulas (commented):
The tests looks nice, do you need help to fix the use case that is failing ?

tanvimoharir:
> The tests looks nice, do you need help to fix the use case that is failing ?

I have attempted a fix.

tanvimoharir:
> The tests looks nice, do you need help to fix the use case that is failing ?

Yes, it would be great if you help me with this. I have tried few things and they dont seem to be working although the tests pass when I executed locally.

Pierre-Sassoulas:
I think we need to define better what is expected in the ``pyproject.toml`` for pylint. The current crash might be due to ``disable = "logging-not-lazy,logging-format-interpolation"`` not being inside "MESSAGES CONTROL". But we'd need to have proper error message when something is misplaced, because right now it's not very convenient and it's just crashing when the configuration is not correct. 

I guess the expected configuration for the failing test should be:
```toml
[tool.pylint]
    [tool.pylint."messages control"]
        disable = "logging-not-lazy,logging-format-interpolation"
    [tool.pylint."master"]
        load-plugins = []
```
(Correct me if I'm wrong, I'm no toml expert. But when using the current conf instead of crashing we should have an error message like:
```
Disable is not a valid section for pylint, did you mean '[tool.pylint."messages control"]  disable = ...' ?
```

tanvimoharir:
> I think we need to define better what is expected in the `pyproject.toml` for pylint. The current crash might be due to `disable = "logging-not-lazy,logging-format-interpolation"` not being inside "MESSAGES CONTROL". But we'd need to have proper error message when something is misplaced, because right now it's not very convenient and it's just crashing when the configuration is not correct.
> 
> I guess the expected configuration for the failing test should be:
> 
> ```toml
> [tool.pylint]
>     [tool.pylint."messages control"]
>         disable = "logging-not-lazy,logging-format-interpolation"
>     [tool.pylint."master"]
>         load-plugins = []
> ```
> 
> (Correct me if I'm wrong, I'm no toml expert. But when using the current conf instead of crashing we should have an error message like:
> 
> ```
> Disable is not a valid section for pylint, did you mean '[tool.pylint."messages control"]  disable = ...' ?
> ```

Thanks @Pierre-Sassoulas I have updated with the configuration you have specified here. I will look more into TOML types since I'm not that familiar with it.

tanvimoharir:
@Pierre-Sassoulas  Could you help me  proceed with this ? I'm not able to understand it clearly.

tanvimoharir:
> Hey sorry for the time it took to answer this. It look like the configuration from the toml is not taken into account like before as some integration tests are failing. If I were you I would add unit test on the TOML configuration in order to understand what is going wrong in the integrations tests first. Do not hesitate to ask if there is something specific that is blocking you.

The current failures which i see through via pipeline details are because of the tests which I added.
Do you think I should remove those?
Let me know if I'm missing something.

DanielNoord (changes requested):
Small comments here and there. I don't know what's wrong with `ModuleStats()` here, but I think that shouldn't be fixed in this PR.

I think we should also add a changelog entry for this.

tests/config/test_config_pyproject_toml.py
```diff
@@ -0,0 +1,35 @@
...
+        with patch(
+            "sys.argv",
+            ["pylint", "--rcfile", str(toml_path), str(HERE / "file_to_lint.py")],
+        ):
+            try:
+                run_pylint()
+            except SystemExit as ex:
+                out, err = capsys.readouterr()
+                msg = f"Wrong result with configuration {toml_path}"
+                assert expected_result.get("out", "¤") in out or not out, msg
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Sorry I changed a lot of thing in the latest commit, it's refactored now.

tests/config/issue_4746/pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.poe.tasks]
+docs = {cmd = "sphinx-build docs build", help = "Build documentation"}
+
+[tool.pylint.MASTER]
+load-plugins = 'pylint_websockets'
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done

pylint/config/option_manager_mixin.py
```diff
@@ -273,25 +274,7 @@ def read_config_file(self, config_file=None, verbose=None):
             parser = self.cfgfile_parser
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done

pylint/config/option_manager_mixin.py
```diff
@@ -311,6 +294,35 @@ def read_config_file(self, config_file=None, verbose=None):
             msg = "No config file found, using default configuration"
         print(msg, file=sys.stderr)
 
+    @staticmethod
+    def _parse_toml(
+        config_file: Union[Path, str], parser: configparser.ConfigParser
+    ) -> None:
+        """Upgrade the parser from a toml configuration file."""
```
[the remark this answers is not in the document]
Pierre-Sassoulas: I guess it was "Upgrade [the value inside] the parser". I like your suggestion better.

pylint/config/option_manager_mixin.py
```diff
@@ -311,6 +294,35 @@ def read_config_file(self, config_file=None, verbose=None):
...
+        with open(config_file, encoding="utf-8") as fp:
+            content = toml.load(fp)
+        try:
+            sections_values = content["tool"]["pylint"]
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done

pylint/config/option_manager_mixin.py
```diff
@@ -311,6 +294,35 @@ def read_config_file(self, config_file=None, verbose=None):
...
+                    for option, value in values.items():
+                        if isinstance(value, bool):
+                            values[option] = "yes" if value else "no"
+                        elif isinstance(value, (int, float)):
+                            values[option] = str(value)
+                        elif isinstance(value, list):
+                            values[option] = ",".join(value)
+                        else:
+                            values[option] = str(value)
+                parser._sections[section.upper()] = values  # type: ignore
```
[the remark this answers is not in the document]
Pierre-Sassoulas: ``pylint/config/option_manager_mixin.py:321: error: "ConfigParser" has no attribute "_sections"``

pylint/config/option_manager_mixin.py
```diff
@@ -311,6 +294,35 @@ def read_config_file(self, config_file=None, verbose=None):
             msg = "No config file found, using default configuration"
         print(msg, file=sys.stderr)
 
+    @staticmethod
+    def _parse_toml(
+        config_file: Union[Path, str], parser: configparser.ConfigParser
```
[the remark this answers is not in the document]
Pierre-Sassoulas: I'm trying to use the largest possible type here because ``open`` can take both ``str`` or ``Path`` so using this typing is true and permit to not cast to string for no reason when we use a Path before calling.

ChangeLog
```diff
@@ -140,6 +140,11 @@ Release date: TBA
 
   Closes #5216
 
+* Crashes when a list exists in a toml configuration does not happen anymore. A new ``bad-configuration-option-value``
+  was added that will emit for some problem, but a bad configuration might still silentely not be taken into account.
+
+  Closes #4580
+
```
DanielNoord, line 147:
```suggestion
* Crashes when a list is encountered in a toml configuration do not happen anymore. A new ``bad-configuration-option-value``
  checker was added that will emit for these problems. 
  Some bad configurations might still silently not be taken into account which is tracked in a follow-up issue.

  Closes #4580
  Follow-up in #5259
```

DanielNoord: Shouldn't this also be in `2.12`?

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
DanielNoord: It is just from experience that I know that allowing both `Path` and `str` can create some problems later on. 
I think it was the typing of `file_location` or something in which initially both `Path` and `str` were allowed, which required a lot of `isinstance()`'s only to later find out that it couldn't ever be `Path`. I think in this case it should be fine, but I have noticed that we spare ourselves some headaches if we normalise all `Path`'s to be `str` when we are handling and passing paths around internally.

pylint/config/option_manager_mixin.py
```diff
@@ -311,12 +292,44 @@ def read_config_file(self, config_file=None, verbose=None):
...
     def load_config_file(self):
         """Dispatch values previously read from a configuration file to each
         options provider)"""
         parser = self.cfgfile_parser
         for section in parser.sections():
-            for option, value in parser.items(section):
+            try:
+                items = parser.items(section)
+            except ValueError:
+                self.add_message("bad-configuration-option-value", line=0, args=section)
```
DanielNoord, line 331: I think we need to return here? Otherwise we will handle half of the config if we find a bad value somewhere in the middle?

DanielNoord (changes requested):

tests/config/test_config_pyproject_toml.py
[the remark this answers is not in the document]
DanielNoord: Much better!

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
DanielNoord: I think we should use `parser.options(section.upper())` here. `mypy` is correctly complaining about accessing internal/private attributes. `options()` seems to be the correct way to do this looking at the `ConfigParser` class.

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: Good catch, I think we need to continue to parse the rest of the configuration file.

DanielNoord: Do we though? It seems weird to continue parsing half of the file if we know we can't parse it fully? What if there are actual correct sections after the section that raises the `ValueError`? As a user I think I would expect a program to stop parsing an incorrect configuration instead of parsing half of it.

Pierre-Sassoulas: But right now we're still running pylint after that, so if the program still runs I would expect the program to run with all the valid configurations it can parse. 

I agree that it would also make sense to raise a parsing error (``argparse.ArgumentParser.error(...)`` (we're not using argparse yet though) or cleanly raise and handle another error and print a nice message for it as it's not really a problem related to parsing anymore.

But we're not detecting a lot of parsing error for stuff option that do not exists but are in an existing section. We don't know what we expect in this configuration file. Isn't stopping the process for some errors and not the other would make pylint inconsistent ?  It's not like not taking things into account is particularly good either...

I'm half inclined to just fix the crash, release 2.12, and do that properly in 2.13.

DanielNoord: There is no easy way to keep continuing parsing the sections after the faulty section right? If that's not possible I think it is indeed perhaps best to fix this like this for 2.12 and improve handling in 2.13.

Pierre-Sassoulas: I realized that for top level option we can sometime be certain that it's a bad option. I'm going to upgrade the MR to reflect that. This does not permit to detect bad value inside a valid top level section but is still better.

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
Pierre-Sassoulas: > required a lot of isinstance()'s

I guess it's to stop mypy from complaining ? Path(str) or str(path) should work to go from one to the other (if the path can't be None) so I would cast directly instead of using isinstance ? But I agree that if we have some function that require one or the other casting to string/path is indeed a lot of work.

In the current function ``Union[Path, str]`` is not a "lie".  But If we want to be able to use Path some day I think using the real typing we can handle when we can is important. It prevents having to cast to str to satisfy mypy.

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
Pierre-Sassoulas: Nice catch, I should have read the doc :) !

pylint/config/option_manager_mixin.py
```diff
@@ -311,12 +292,44 @@ def read_config_file(self, config_file=None, verbose=None):
...
+            elif isinstance(values, dict):
+                for option, value in values.items():
+                    if isinstance(value, bool):
+                        values[option] = "yes" if value else "no"
+                    elif isinstance(value, (int, float)):
+                        values[option] = str(value)
+                    elif isinstance(value, list):
+                        values[option] = ",".join(value)
+                    else:
+                        values[option] = str(value)
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done in 3e0f65a9 (not yet pushed to save some github integration run)

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
DanielNoord: I wonder if there is every any reason to use `Path`. It seems the only difference between `Path` and `str` is that `Path` can be a `bytes` and can be easily converted between `str` and `bytes` in both ways.
See:
https://docs.python.org/3/glossary.html#term-path-like-object

I might be wrong here, but I don't immediately see why we would ever need it to be `Path`. 
Besides, I think being able to know that pathnames within `pylint` are always `str` makes it a lot easier to handle them and write code that accepts it. That level of consistency makes the codebase more maintainable imo.

pylint/utils/linterstats.py
```diff
@@ -285,9 +285,18 @@ def increase_single_module_message_count(
         increase: int,
     ) -> None:
         """Increase the message type count of an individual message type of a module"""
-        self.by_module[modname][type_name] = (
-            self.by_module[modname][type_name] + increase
-        )
+        if self.by_module.get(modname, None) is None:
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done in 1abc4b963fb92e13db615dccf23a7e0892173f04, this permits to have the following output:

```
************* Module tests/config/issue_4580/invalid_data_for_basic.toml
tests/config/issue_4580/invalid_data_for_basic.toml:1:0: E0014: Bad top level configuration section name encountered 'load-plugins' : '[]' (bad-configuration-option-value)
tests/config/issue_4580/invalid_data_for_basic.toml:1:0: E0014: Bad top level configuration section name encountered 'disable' : 'logging-not-lazy,logging-format-interpolation' (bad-configuration-option-value)
```

Pierre-Sassoulas:
I changed this to draft because I'd like to merge https://github.com/PyCQA/pylint/pull/5271 first then rebase on it. But this is almost ready now.

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
Pierre-Sassoulas: There's 17 files using ``pathlib.Path`` already (I'm probably mostly the one who added them). It's personal preference but I rather do ``my_path.resolve() / "directory" / "file.py"`` than ``os.path.abspath(os.path.join(my_path, "directory", "file.py))`` or ``Path(__file__).resolve().parent.parent`` than ``os.path.dirname(os.path.dirname(os.path.abspath(__file__)))``. Also ``mypath.glob('*.csv')`` does not require an additional import to work. Last thing, Path is more semantic, a string can be something else than a path so using Path indicate intent.

I'm not on a crusade to change everything to Path everywhere, but I think being able to use Path would be convenient. Also all python functions handle both, so if we can't, it's probably an internal pylint problem. Putting the real typing in new function is the way to slowly fix this problem, so I think we should do it. Putting only str will raise mypy error that are not really errors (which might be the reason you had so much trouble with Path vs str in the first place ?).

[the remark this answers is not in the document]
DanielNoord: Yeah you convinced me, we should allow both here.

I looked at what the initial problem was in https://github.com/PyCQA/pylint/pull/4980/files. Seems like this was because of a similar discussion: should we use `Union[str, Path]` or should we type like we are currently using the functions (`str`)?. In that case we decided against allowing `Path` because it would mean a lot of additional typing and some refactoring. Here it might be fine to do allow it as we are also not returning the `parameter` that is typed as `Path`.

Pierre-Sassoulas (commented):
This is ready for review following #5276 and #5277.

tests/config/issue_4580/invalid_data_for_import.toml
```diff
@@ -0,0 +1,7 @@
+# Both disable and load-plugins are not top level section
```
DanielNoord, line 1:
```suggestion
# Both disable and load-plugins do not belong in the imports section
```

tests/config/issue_4580/empty_list.toml
```diff
@@ -0,0 +1,2 @@
+[tool.pylint]
+load-plugins = []
```
DanielNoord, line 2:
```suggestion
# load-plugins does not belong in top level section
load-plugins = []
```

tests/config/issue_4580/invalid_data_for_basic.toml
```diff
@@ -0,0 +1,4 @@
+[tool.pylint]
+# Both disable and load-plugins are not top level section
```
DanielNoord, line 2:
```suggestion
# Both disable and load-plugins do not belong in the top level section
```

tests/config/issue_4580/top_level_disable.toml
```diff
@@ -0,0 +1,2 @@
+[tool.pylint]
+disable = "logging-not-lazy,logging-format-interpolation"
```
DanielNoord, line 2:
```suggestion
# disable does not belong in top level section
disable = "logging-not-lazy,logging-format-interpolation"
```

tests/config/test_config_pyproject_toml.py
```diff
@@ -0,0 +1,70 @@
...
+    for toml_path in pyproject_toml_paths:
+        expected_result = expected_results[toml_path.name]
+        file_to_lint = str(HERE / "file_to_lint.py")
+        with patch(
+            "sys.argv",
+            ["pylint", "--rcfile", str(toml_path), file_to_lint, "-r", "n"],
+        ):
+            try:
+                run_pylint()
+            except SystemExit as ex:
```
DanielNoord, line 59: Not parametrising this test might create a lot of duplicate lines but these asserts are really hard to follow. 
Wouldn't it be better to just separate these into different tests and assert on the value of the exit code instead of parametrising?

pylint/lint/pylinter.py
```diff
@@ -158,6 +158,11 @@ def _load_reporter_by_class(reporter_class: str) -> type:
         "bad-plugin-value",
         "Used when a bad value is used in 'load-plugins'.",
     ),
+    "E0014": (
+        "Bad top level configuration section name encountered '%s' : '%s'",
```
DanielNoord, line 162:
```suggestion
        "Incorrect setting encountered in top level configuration-section '%s' : '%s'",
```

tests/config/test_config_pyproject_toml.py
```diff
@@ -0,0 +1,70 @@
...
+                out, err = capsys.readouterr()
+                msg = f"Wrong result with configuration {toml_path}"
+                if "out" not in expected_result:
+                    assert not out, msg
+                else:
+                    assert expected_result["out"] in out, msg
```
DanielNoord, line 65: Also, I think we should check for the full message.
Checking for `"encountered 'disable' : 'logging-not-lazy,",` does not really tell me what the expected output of the message is.

pylint/config/option_manager_mixin.py
```diff
@@ -292,36 +300,53 @@ def read_config_file(self, config_file=None, verbose=None):
...
     def load_config_file(self):
         """Dispatch values previously read from a configuration file to each
         options provider)"""
         parser = self.cfgfile_parser
         for section in parser.sections():
-            for option, value in parser.items(section):
+            try:
+                items = parser.items(section)
+            except ValueError:
+                self.add_message(
```
DanielNoord, line 345: Might want to add an explanation for why we get here.

Can we get the incorrect item from the ValueError message? If so we might be able to pass it as an `args`.

pylint/config/option_manager_mixin.py
```diff
@@ -271,6 +271,14 @@ def read_config_file(self, config_file=None, verbose=None):
...
         use_config_file = config_file and os.path.exists(config_file)
         if use_config_file:
+            mod_name = config_file
+            self.current_name = mod_name
+            try:
+                self.set_current_module(mod_name)
+            except AttributeError:
+                # 'OptionsManagerMixIn' object has no attribute 'set_current_module'
+                # Ie: the class is not used as a mixin
+                pass
```
DanielNoord, line 281: Do we actually use this without initialising `PyLinter`?

DanielNoord (commented):
Change (like previously) looks good! I like how we are now handling most errors while still continuing.

Mostly some comments about the tests.

tests/config/test_config_pyproject_toml.py
Pierre-Sassoulas: I'm half inclined to create a file with expected result like for the functional tests.

tests/config/test_config_pyproject_toml.py
Pierre-Sassoulas: Resolving in favor of this discussion : https://github.com/PyCQA/pylint/pull/4720/files#r744555630

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: It seems to be only in ``test_can_read_toml_env_variable`` now (I remembered 4/5 tests being affected), we might want to change the test itself.

DanielNoord: I think that might be better indeed.

tests/config/test_config_pyproject_toml.py
DanielNoord: Wouldn't we need to create a new system for it? 
With the functional tests you can still see the expected behaviour (because of the comments) in the tests, while the full data is in another file. That's quite useful.

Pierre-Sassoulas: This is going to require another merge request I guess. Do you think we should do this before or after we handle this one ? I'm going to push what I currently have (without annotation in the toml for expected error).

DanielNoord: I'm not sure how easy it is to implement such a system. Might be a bit overkill?

Separating the tests might be easier? It is not as if we're going for the most perfomance-optimized test suite anyway. It might take a similar amount of time as creating a new system to allow external expected results.

pylint/lint/pylinter.py
```diff
@@ -158,6 +158,11 @@ def _load_reporter_by_class(reporter_class: str) -> type:
         "bad-plugin-value",
         "Used when a bad value is used in 'load-plugins'.",
     ),
+    "E0014": (
+        "Incorrect setting encountered in top level configuration-section '%s' : '%s'",
+        "bad-configuration-option-value",
+        "Used when a top section value can't be parsed probably because it does not exists.",
```
DanielNoord, line 164:
```suggestion
        "Used when a top section value can't be parsed because it does not belong to the top section "
        "or has an incorrect value type.",
```

tests/config/test_functional_toml_parsing.py
```diff
@@ -0,0 +1,51 @@
...
+    with patch(
+        "sys.argv", ["pylint", "--rcfile", pyproject_toml_path, FILE_TO_LINT, "-r", "n"]
+    ):
+        with pytest.raises(SystemExit) as exc_info:
+            run_pylint()
+        assert exc_info.value.code == expected_code, msg
+        out, err = capsys.readouterr()
+        # rstrip() applied so we can have a final newline in the expected test file
+        assert expected_output.rstrip() == out.rstrip(), msg
+        assert not err, msg
```
DanielNoord, line 51: I like this! At some point we might want to make this accessible for other tests as well, but for this PR I think this works fine!

Pierre-Sassoulas: Yes, I made that for toml file, only but it could be any configuration file... We'll just have to select file not ending in ".expected".

tests/config/functional/issue_4580/invalid_data_for_import.toml
```diff
@@ -0,0 +1,7 @@
+# Both disable and load-plugins do not belong in the imports section
```
Pierre-Sassoulas, line 1: Notice how we do not detect anything here, but we should.

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: This will be fixed in #5284

Pierre-Sassoulas (commented):
Following #5287 this became a more manageable change, it's ready for review @DanielNoord  :)

tests/config/functional/toml/issue_4580/invalid_data_for_import.toml
```diff
@@ -0,0 +1,7 @@
+# Both disable and load-plugins do not belong in the imports section
```
Pierre-Sassoulas, line 1:
```suggestion
# Both disable and load-plugins do not belong in the imports section
# TODO This should be fixed in #5259
```

ChangeLog
```diff
@@ -153,6 +153,14 @@ Release date: TBA
 
   Closes #5261
 
+* Crashes when a list is encountered in a toml configuration do not happen anymore. A new
```
DanielNoord, line 156: This should also be in `2.12.rst`

pylint/config/option_manager_mixin.py
```diff
@@ -305,23 +304,36 @@ def _parse_toml(
         except KeyError:
             return
         for section, values in sections_values.items():
+            section_name = section.upper()
             # TOML has rich types, convert values to
             # strings as ConfigParser expects.
+            if not isinstance(values, dict):
+                # This class is a mixin the add_message come from the pylinter
```
DanielNoord, line 311:
```suggestion
                # This class is a mixin: add_message comes from the `PyLinter` class
```

pylint/config/option_manager_mixin.py
```diff
@@ -305,23 +304,36 @@ def _parse_toml(
...
+                    parser.set(section_name, option, value=value)
 
     def load_config_file(self):
         """Dispatch values previously read from a configuration file to each
         options provider)"""
         parser = self.cfgfile_parser
         for section in parser.sections():
-            for option, value in parser.items(section):
+            items = parser.items(section)
+            for option, value in items:
```
DanielNoord, line 336:
```suggestion
            for option, value in parser.items(section):
```

This is not too unreadable, right? Or does this create issues?

pylint/config/option_manager_mixin.py
```diff
@@ -305,23 +304,36 @@ def _parse_toml(
...
                 elif isinstance(value, list):
                     values[option] = ",".join(value)
-            parser._sections[section.upper()] = values  # type: ignore
+                else:
+                    values[option] = str(value)
+            for option, value in values.items():
+                try:
+                    parser.set(section_name, option, value=value)
+                except configparser.NoSectionError:
+                    parser.add_section(section_name)
```
DanielNoord, line 327: 👍

pylint/config/option_manager_mixin.py
```diff
@@ -305,23 +304,36 @@ def _parse_toml(
         except KeyError:
             return
         for section, values in sections_values.items():
+            section_name = section.upper()
             # TOML has rich types, convert values to
             # strings as ConfigParser expects.
+            if not isinstance(values, dict):
+                # This class is a mixin the add_message come from the pylinter
+                self.add_message(  # type: ignore
+                    "bad-configuration-option-value", line=0, args=(section, values)
```
DanielNoord, line 313: I'm probably overlooking something really obvious, but how does this check whether `load-plugins = []` should be in the `tools.pylint` section (as one of the tests does)?

pylint/lint/pylinter.py
```diff
@@ -158,6 +158,12 @@ def _load_reporter_by_class(reporter_class: str) -> type:
         "bad-plugin-value",
         "Used when a bad value is used in 'load-plugins'.",
     ),
+    "E0014": (
+        "Incorrect setting encountered in top level configuration-section '%s' : '%s'",
+        "bad-configuration-option-value",
+        "Used when a top section value can't be parsed because it does not belong to the top section "
```
DanielNoord, line 164: See other comment, but based on the `isinstance()` test I feel like this message is only emitted when the type of the value is incorrect?

DanielNoord (commented):
I don't really understand how the new checker works, or at least: how its description fits the single case where it can be emitted in `option_manager_mixin`.

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: Nice catch, I probably just needed to do that for debugging.

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: There is nothing directly in the ``tools.pylint`` except header (master, message controls, etc), so at top level, if it's not a dict it can't be right. We know nothing about what to expect in the configuration except that.

pylint/lint/pylinter.py
Pierre-Sassoulas: At the top level of the configuration if we do not have a dict it's not a proper top section header. We actually want to change that so we don't have to guess what is the checker that contain this option, but it's a breaking change.

ChangeLog
Pierre-Sassoulas: 04c0603615d263679b7178a3e9f4aa368d07c525

doc/whatsnew/2.12.rst
```diff
@@ -56,6 +56,11 @@ New checkers
 
   Closes #5208
 
+* A new ``bad-configuration-option-value`` checker was added that will emit for misplaced option
+  in pylint's top level namespace for toml configuration. Top-level dictionaries or option defined
+  in the wrong section will still silently not be taken into account, which is tracked in a
+  follow-up issue.
```
DanielNoord, line 62: Dont't we want to link to the follow up issue here?

pylint/config/option_manager_mixin.py
DanielNoord: Perhaps `bad-configuration-section` would be better then? For me `option-value` implies there is something wrong with the type of the option value (but that might be just me).

I think such a check (option value types) would be helpful anyway, if we don't already check it, so reserving `bad-configuration-option-value` for that checker would be better. 
Putting stuff in the wrong section and giving stuff the incorrect type are different mistakes and require different messages. For the the latter we could (for example) also include the expected type in the message.

pylint/lint/pylinter.py
```diff
@@ -158,6 +158,12 @@ def _load_reporter_by_class(reporter_class: str) -> type:
...
+        "or has an incorrect value type.",
```
DanielNoord, line 165:
```suggestion
        "Used when we detect a setting in the top level of a toml configuration that shouldn't be there.
```

pylint/lint/pylinter.py
```diff
@@ -158,6 +158,12 @@ def _load_reporter_by_class(reporter_class: str) -> type:
         "bad-plugin-value",
         "Used when a bad value is used in 'load-plugins'.",
     ),
+    "E0014": (
+        "Incorrect setting encountered in top level configuration-section '%s' : '%s'",
```
DanielNoord, line 162:
```suggestion
        "Out-of-place setting encountered in top level configuration-section '%s' : '%s'",
```

DanielNoord (commented):
I understand what the message is checking now. I think renaming and rewording the description might help future confusion. Rest looks good to go!

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: I agree, I'm going to change the name and upgrade the tests.

Pierre-Sassoulas: 22ed9232

doc/whatsnew/2.12.rst
Pierre-Sassoulas: 0fbaadcd

PR #4615, opened 2021-06-27
Pyreverse: PlantUML support + coloring (from PR #4521)

DudeNr33:
## Steps

- [x] Add a ChangeLog entry describing what your PR does.
- [x] If it's a new feature or an important bug fix, add a What's New entry in
      `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Description

This is a rather large PR, so I try my best to summarize the changes.
It solves both #4488 and #4498. 
The requested changes from PR #4521 are already included, and PR #4521 can be deleted (the introduction of ``NamedTuple`` for better typing was easier to do with the changes already made for the PlantUML support).

So here is an overview what this PR includes:

1. New Features:
    * Optional parameters ``--colorized`` and ``--max-color-depth`` to automatically color classes/modules of the same package (merged from PR #4521)
    * Implement a ``PlantUmlPrinter`` to support output to .puml files.  
    Note that automatic conversion to .png or other image formats is not supported yet; specifying "png" as output format will still rely on the Dot backend.
    * Positive side effect of refactoring: The various import-graphs provided by the ``ImportsChecker`` can now be produced in .vcg and .puml format, not only .dot.

2. Refactoring:
    * Defined a common ``Printer`` interface in ``pylint.pyreverse.printer`` for the possible backends (Dot, VCG, PlantUML).  
    Moved ``DotBackend`` and ``VCGPrinter`` to ``printer`` module.  
    * YAGNI: remove optional parameters from various methods in the ``Printer`` subclasses that were never used by the callers in the codebase, to find a minimal superset of parameters for each method signature.
    * Introduce factory function for picking the corresponding ``Printer`` or ``DiagramWriter`` subclass for the desired output format.
    * Remove module ``pylint.graph`` as it is no longer needed (``DotBackend`` is replaced by ``pylint.pyreverse.printer.DotPrinter``). Same for ``pylint.pyreverse.vcgutils``.
    Move orphaned ``get_cycles()`` function to ``pylint.checkers.imports``, which is (in my opinion) the better place for it anyway.
    * Refactor tests:
        * move to own subdirectory
        * split into functional tests and unit tests
        * extract for ``pyreverse.utils`` into own test module
    * Improve typing by adding type hints and use NamedTuples etc. where appropriate.
    * Use a consistent style for defining the dictionaries of the configuration options
[...]

PR #4731, opened 2021-07-20
Create common ``Printer`` base class for ``pyreverse`` and improve typing.

DudeNr33:
## Steps

## Description
This is the first MR which shall split MR #4615 into smaller and more easily reviewable parts.
It involves the necessary refactoring which is needed to prepare for the new features (PlantUML support and colorization).
It also adds type annotations.

Summary of changes:

* Rename ``vcgutils.py`` to ``printer.py`` and introduce a ``Printer`` base class which defines the common interface for the different backends.
* Adapt ``VCGPrinter`` to adhere to the new printer base class
* Add a ``DotPrinter`` class which essentially implements the features of ``pylint.graph.DotBackend``. The latter will be removed in a future MR, as it is currently also used by the ``ImportsChecker`` and I wanted to keep this MR smaller and focused (it is already not very small...).
* Add type annotations for the ``Printer`` class and its children. This also uncovered some problems where ``mypy`` correctly complained, so I had to add specialised ``PackageEntity`` and ``ClassEntity`` child classes in ``diagrams.py``.
* Added tests for VCG output, which was missing before
* Updated the test data as the new ``Printer`` backends now outputs some more information. The final image produced by Graphviz is still the same.

## Type of Changes


pylint/pyreverse/utils.py
```diff
@@ -275,3 +275,15 @@ def infer_node(node: Union[astroid.AssignAttr, astroid.AssignName]) -> set:
         return set(node.infer())
     except astroid.InferenceError:
         return set()
+
+
+def get_file_extension(filename_or_extension: str) -> str:
```
Pierre-Sassoulas, line 280: I think it's replaceable by ``pathlib.Path(filename_or_extension).suffix``

Pierre-Sassoulas (commented):
Just a small comment, and I did not look in detail but the new typing is 🔥 , thanks! I'll review when this is no longer a draft :)

pylint/pyreverse/printer.py
```diff
@@ -0,0 +1,489 @@
...
+
+
+class NodeProperties(NamedTuple):
+    label: str
+    color: Optional[str] = None
+    fontcolor: Optional[str] = None
+    body: Optional[str] = None
+
+
+class Printer:
```
Pierre-Sassoulas, line 57:
```suggestion
class Printer(abc.ABC):
```
It's more explicit than raising ``NotImplementError``s :)

pylint/pyreverse/printer.py
```diff
@@ -0,0 +1,489 @@
...
+        self._close_graph()
+        with open(outputfile, "w", encoding="utf-8") as outfile:
+            outfile.writelines(self.lines)
+
+    def _close_graph(self) -> None:
+        """Emit the lines needed to properly close the graph."""
+        raise NotImplementedError
+
+
+class VCGPrinter(Printer):
```
Pierre-Sassoulas, line 111: We can create a new file for VCGPrinter and another one for DotPrinter :) (It would also help with the diff and possible future conflict if vcgutils.py is recognized as the old file for it).

pylint/pyreverse/writer.py
```diff
@@ -44,94 +57,104 @@ def write(self, diadefs):
...
+    def get_class_properties(self, obj: ClassEntity) -> NodeProperties:
+        """get label and shape for classes."""
         raise NotImplementedError
 
+    def save(self) -> None:
+        """write to disk"""
+        self.printer.generate(self.file_name)
+
 
 class DotWriter(DiagramWriter):
```
Pierre-Sassoulas, line 133: Maybe we could start with a first merge request that create the following architecture and move relevant part of the code without modifying anything:
```
pyreverse:
    writter:
        diagram_writter.py
        dot_writter.py
        vcg_writter.py
    printer:
        dot_printer.py
        vcg_printer.py
```
What do you think ?

Pierre-Sassoulas (changes requested):
Hey, thank you for this work on this and sorry for the delay before reviewing. Could we reorganize and move the code around first then do the modification on it ? I'd be more confident if there was a proper diff for the change in the printer.py file, right now everything is considered new by git and it's hard to tell what changed exactly :)

DudeNr33:
Thanks for the feedback. I will implement it when #4745 is merged and I did a rebase.

pylint/pyreverse/diagrams.py
```diff
@@ -45,6 +45,19 @@ def __init__(self, title="No name", node=None):
...
+
+
+class ClassEntity(DiagramEntity):
+    """A diagram object representing a class"""
+
+    def __init__(self, title, node):
+        super().__init__(title=title, node=node)
+        self.attrs = None
+        self.methods = None
+
```
DudeNr33, line 59: Adding the type hints revealed that it was necessary to distinguish between a ``PackageEntity`` and a ``ClassEntity``, because the ``ClassEntity`` has additional attributes that were dynamically added in the previous code, which confused ``mypy``.

pylint/pyreverse/writer.py
```diff
@@ -44,94 +52,104 @@ def write(self, diadefs):
...
-    def get_values(self, obj):
+    def get_package_properties(self, obj: PackageEntity) -> NodeProperties:
+        """get label and shape for packages."""
+        return NodeProperties(
+            label=self.get_title(obj),
+            color="black",
+        )
+
+    def get_class_properties(self, obj: ClassEntity) -> NodeProperties:
         """get label and shape for classes.
```
DudeNr33, line 148: ``get_values`` is now ``get_class_properties``.

``get_package_properties`` is currently trivial and the color is hardcoded, but this will change when the new feature for automatic coloring are merged.

pylint/pyreverse/writer.py
```diff
@@ -159,53 +179,33 @@ def get_values(self, obj):
...
-            dict(
-                arrowstyle="solid",
-                backarrowstyle="none",
-                linestyle="dotted",
-                backarrowsize=10,
-            ),
-            dict(arrowstyle="solid", backarrowstyle="none", textcolor="green"),
-        ]
-        DiagramWriter.__init__(self, config, styles)
-
```
DudeNr33, line 187: The ``__init__`` is now obsolete. It formerly defined the styles to apply to the different types of edges for displaying relationships.
These are now defined in the ``ARROWS`` dictionary in the module of the corresponding printer class.

DudeNr33:
I implemented most of the feedback.
I reintroduced the module ``vcgutils.py``, which should make it now easier to diff.
As soon as the review is finished, I would suggest to make a last commit where this file is simply renamed to ``vcg_printer.py``.

Regarding the last suggestion (doing a first MR to implement the structure 
```
pyreverse:
    writter:
        diagram_writter.py
        dot_writter.py
        vcg_writter.py
    printer:
        dot_printer.py
        vcg_printer.py
```
):

If the current cleanup is not enough to properly review the changes, let me know and I can do that.

pylint/pyreverse/vcgutils.py
```diff
@@ -152,73 +154,129 @@
...
                 raise Exception(
-                    f"""value {value} isn't correct for attribute {key}
-correct values are {type}"""
+                    f"value {value} isn't correct for attribute {key} correct values are {type}"
                 )
 
     def _inc_indent(self):
         """increment indentation"""
-        self._indent = "  %s" % self._indent
+        self._indent += "  "
```
Pierre-Sassoulas, line 279: 👍

pylint/pyreverse/writer.py
```diff
@@ -140,12 +158,14 @@ def get_values(self, obj):
                 )
 
                 if func.args.args:
-                    args = [arg for arg in func.args.args if arg.name != "self"]
+                    argument_list = [
```
Pierre-Sassoulas, line 161:
```suggestion
                    arguments: List[str] = [
```
Personal preference but when naming an iterable I like a plural word so I can do ``for argument in arguments`` later on.

pylint/pyreverse/dot_printer.py
```diff
@@ -0,0 +1,141 @@
...
+    NodeType.INTERFACE: "record",
+    NodeType.CLASS: "record",
+}
+ARROWS: Dict[EdgeType, Dict] = {
+    EdgeType.INHERITS: dict(arrowtail="none", arrowhead="empty"),
+    EdgeType.IMPLEMENTS: dict(arrowtail="node", arrowhead="empty", style="dashed"),
+    EdgeType.ASSOCIATION: dict(
+        fontcolor="green", arrowtail="none", arrowhead="diamond", style="solid"
+    ),
+    EdgeType.USES: dict(arrowtail="none", arrowhead="open"),
```
Pierre-Sassoulas, line 31: Maybe we could create enums for fontcolor, arrowtail, style and arrowhead ?

pylint/pyreverse/dot_printer.py
```diff
@@ -0,0 +1,141 @@
...
+                dot_sourcepath = outputfile
+        with open(dot_sourcepath, "w", encoding="utf8") as outfile:
+            outfile.writelines(self.lines)
+        if target not in graphviz_extensions:
+            if shutil.which("dot") is None:
+                raise RuntimeError(
+                    f"Cannot generate `{outputfile}` because 'dot' "
+                    "executable not found. Install graphviz, or specify a `.gv` "
+                    "outputfile to produce the DOT source code."
+                )
```
Pierre-Sassoulas, line 130: We could create a function for this I think.

pylint/pyreverse/dot_printer.py
```diff
@@ -0,0 +1,141 @@
...
+ARROWS: Dict[EdgeType, Dict] = {
+    EdgeType.INHERITS: dict(arrowtail="none", arrowhead="empty"),
+    EdgeType.IMPLEMENTS: dict(arrowtail="node", arrowhead="empty", style="dashed"),
+    EdgeType.ASSOCIATION: dict(
+        fontcolor="green", arrowtail="none", arrowhead="diamond", style="solid"
+    ),
+    EdgeType.USES: dict(arrowtail="none", arrowhead="open"),
+}
+RANKDIR: Dict[Layout, str] = {
+    Layout.LEFT_TO_RIGHT: "LR",
```
Pierre-Sassoulas, line 34: I think we can use ``Layout.LEFT_TO_RIGHT.name`` / ``Layout.LEFT_TO_RIGHT.value`` and remove this dict.

pylint/pyreverse/vcgutils.py
```diff
@@ -152,73 +154,129 @@
...
+        arrowstyle="solid",
+        backarrowstyle="none",
+        linestyle="dotted",
+        backarrowsize=10,
+    ),
+    EdgeType.ASSOCIATION: dict(
+        arrowstyle="solid", backarrowstyle="none", textcolor="green"
+    ),
+}
+ORIENTATION: Dict[Layout, str] = {
```
Pierre-Sassoulas, line 177: Maybe a function that return the proper value in the Layout enum (Orientation or rankdir) would reduce the number of thing to keep in mind when using the code.

pylint/pyreverse/writer.py
```diff
@@ -44,94 +52,104 @@ def write(self, diadefs):
...
+            self.save()
 
-    def write_packages(self, diagram):
+    def write_packages(self, diagram: PackageDiagram) -> None:
         """write a package diagram"""
         # sorted to get predictable (hence testable) results
-        for i, obj in enumerate(sorted(diagram.modules(), key=lambda x: x.title)):
-            self.printer.emit_node(i, label=self.get_title(obj), shape="box")
-            obj.fig_id = i
+        for obj in sorted(diagram.modules(), key=lambda x: x.title):
```
Pierre-Sassoulas, line 60:
```suggestion
        for module in sorted(diagram.modules(), key=lambda x: x.title):
```

pylint/pyreverse/writer.py
Pierre-Sassoulas: Thank you for reorganizing the code the diff is manageable now :)

Pierre-Sassoulas (changes requested):
Thank you for your work on this, very impressive ! Ad thank you for making the diff manageable. I left some minor comment. We can still do the architecture change later if you feel like it (probably after merging the other MR as the rebase will be hard to do if we do it before).

pylint/pyreverse/writer.py
DudeNr33: I agree, changed it.

pylint/pyreverse/dot_printer.py
DudeNr33: Yes, good catch. ``main.py`` already does a similar check, they just differ in the consequence if ``dot`` is not found - one uses ``sys.exit(32)``, the other raises ``RuntimeError``. Settled for ``sys.exit(32)``.

pylint/pyreverse/dot_printer.py
DudeNr33: Yes, for the ``DotPrinter`` this is possible. Changed it.

pylint/pyreverse/vcgutils.py
DudeNr33: Hm, I don't know - we would just copy the content of the dictionaries into the ``Layout`` enum. Plus, the enum would then need to know the details of each backend (how the parameters have to look for ``dot``, ``vcg``, and ``puml`` - which is even more tricky as it only supports top to bottom and left to right, but not the other two). So every time a new backend is added, we would not only need to add the new printer class but also modify the ``Layout``.
This would be against the Open-Closed-Principle in my opinion.

pylint/pyreverse/writer.py
DudeNr33: Yes, that's better. Changed it.

pylint/pyreverse/vcgutils.py
Pierre-Sassoulas: You're right I did not realize that :) !

DudeNr33:
Thank you very much for your review! 
The only part I missed in the last push is the request for creating Enums for ``fontcolor``, ``arrowhead`` etc. 

I also need to rename the ``vcgutils.py`` into ``vcg_printer.py``, so please don't merge yet, will do that in the next few days!

Pierre-Sassoulas:
No problem, just request a new review when the code is ready :)

DudeNr33:
I thought about implementing enums for arrowstyle, arrowhead and so on a little longer.
In my opinion there is not much benefit from it in this case. We would have to create those enums for each printer separately, as the names and values for the arguments are different. And then they would only be used in the ``ARROWS`` dictionary, of which the items are not really individually processed:

```python
# in vcgutils.py
attributes = ARROWS[type_]
if label:
    attributes["label"] = label
self._write_attributes(
    EDGE_ATTRS,
    **attributes,
)

# in dot_printer.py
arrowstyle = ARROWS[type_]
attrs = [f'{prop}="{value}"' for prop, value in arrowstyle.items()]
```

I think we would just clutter the modules with enum definitions, without gaining much in terms of readability or reusability.

Pierre-Sassoulas (approved):
Thank you for making a first step easier to review :) !

PR #4737, opened 2021-07-22
[#3791] propose a fix

### LINKED ISSUE #3791: False positive undefined-variable when variable name in decoration matches function argument
### Steps to reproduce

```py
def decorate(seq):
    return lambda f: f

@decorate(x*x for x in range(3))
def f(x):
    pass

@decorate(x*x for x in range(3))
def g(y):
    pass
```

### Current behavior

pylint reports:
```
4:12: E0602: Undefined variable 'x' (undefined-variable)
```
Line 4 being the `@decorate` preceding `f()`.

Note that no error is reported for `g()`: the problem only occurs if the variable name inside the comprehension matches one of the decorated function's arguments. Which is strange, since they should be in separate scopes.

### Expected behavior

No error is reported.

### pylint --version output

```
pylint 2.6.0
astroid 2.4.2
Python 3.8.4 (default, Jul 14 2020, 09:08:56) [GCC]
```

This problem did not occur until I upgraded from pylint 2.5.3 to 2.6.0.

rr-:
## Steps

- [X] Add yourself to CONTRIBUTORS if you are a new contributor.
- [X] Write a good description on what the PR does.

## Description

This is a proposed fix for an issue where defining a function decorated with a list comprehension that shares argument name, and the argument is used twice in the list comprehension, it shows an undefined-variable error.

To put it in code:

```python3
def decorator(seq):
    return lambda f: f

@decorator(x * x for x in "123")
def func(x):
    pass
```

This wrongfully shows an `undefined-variable` error since version at least 2.7.

```python3
def decorator(seq):
    return lambda f: f

@decorator(x for x in "123")
def func(x):
   pass
```

This is rightfully OK on the current `main`.

This probably will need changes, but I am not familiar enough with the codebase to go better about it.

## Type of Changes

## Related Issue

Closes #3791

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,4 +1,4 @@
-Contributors
+﻿Contributors
 ------------
 
 Current team:
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,4 +1,4 @@
-﻿Contributors
+Contributors
 ------------
 
 Current team:
@@ -519,3 +519,5 @@ contributors:
   - Documented Jupyter integration
 
 * Yilei Yang: contributor
+
+* Marcin Kurczewski (rr-): contributor
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -945,6 +945,8 @@ class VariablesChecker(BaseChecker):
 
     def visit_name(self, node):
         """Check that a name is defined in the current scope"""
+        # refactoring this block is a pain.
+        # pylint: disable=too-many-nested-blocks,too-many-branches
         stmt = node.statement()
         if stmt.fromlineno is None:
             # name node from an astroid built from live code, skip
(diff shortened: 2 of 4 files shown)


pylint/checkers/variables.py
```diff
@@ -1015,6 +1016,12 @@ def visit_name(self, node):
                 self._loopvar_name(node, name)
                 break
 
+            # the name has already been consumed, skip decorators
+            if name in current_consumer.consumed and utils.is_func_decorator(
+                current_consumer.node
```
Pierre-Sassoulas, line 1023: I think it should be an ``or`` line 1012 instead :)

CONTRIBUTORS.txt
```diff
@@ -1,4 +1,4 @@
-﻿Contributors
+Contributors
```
Pierre-Sassoulas, line 1: There's a modification here, I don't know why ;)

Pierre-Sassoulas (changes requested):
Look good thank you for the analysis, I have a small comment but this looks mergeable already.

CONTRIBUTORS.txt
rr-: This is just a guess but I think my editor ate the byte order mark?

pylint/checkers/variables.py
rr-: Thanks, moved

CONTRIBUTORS.txt
rr-: That was it. Restored :)

Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
Thank you for the fix and congratulation on becoming a pylint contributor :) !

PR #4749, opened 2021-07-26
Improve Issue forms

cdce8p:
## Description
Follow up to #4742

* Some improvements to the current issue forms
* Allow black issues. Especially useful in cases where the default templates don't match.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
@@ -19,78 +19,73 @@ body:
       description:
         What is the bug about? Please provide the code that is causing the issue, and
         configurations used if required
-      value: |
-        Given a file `a.py`:
-
+      placeholder: |
         ```python
         # Please disable message unrelated to the bug
         # pylint: disable=missing-docstring,
         <a> = b + 1
         ```
+      render: python
     validations:
       required: true
   - type: textarea
     id: configuration
     attributes:
-      label: (Configuration)
+      label: Configuration
       description:
         Please provide the part of the configuration that is causing the bug if required
         (Leave this part blank if the configuration is not relevant)
-      value: |
-        Using the following configuration:
-        ```ini
-        ```
-    validations:
-      required: false
+      render: ini
   - type: textarea
     id: current-behavior
     attributes:
-      label: Command used
-      description: What is the command used and its actual output ?
-      value: |
-        Result of `pylint a.py`:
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,4 +1,4 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
```
Pierre-Sassoulas, line 1: It's a very small option in the interface, maybe we should keep that so a very specific issue with a lot of work (for performance for example) do not get shoehorned into our bug template.

Pierre-Sassoulas (commented):
Left some comment to discuss, you(re right about the parenthesis around optional part, this look horrible when the issue is actually created.

.github/ISSUE_TEMPLATE/BUG-REPORT.yml
```diff
@@ -19,78 +19,73 @@ body:
...
-        ```ini
-        ```
-    validations:
-      required: false
+      render: ini
   - type: textarea
     id: current-behavior
     attributes:
-      label: Command used
-      description: What is the command used and its actual output ?
```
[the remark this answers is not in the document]
cdce8p: Added a new `textarea` for the command

.github/ISSUE_TEMPLATE/config.yml
cdce8p: Tbh I wanted to set this to `enabled` but didn't realize this this was already the case 😅
Thanks for pointing it out.

.github/ISSUE_TEMPLATE/BUG-REPORT.yml
```diff
@@ -35,17 +33,37 @@ body:
       description:
         Please provide the part of the configuration that is causing the bug if required
         (Leave this part blank if the configuration is not relevant)
+      placeholder: |
+        # Leave this black if the configuration is not relevant!
```
Pierre-Sassoulas, line 37:
```suggestion
        # Leave this blank if the configuration is not relevant!
```

.github/ISSUE_TEMPLATE/BUG-REPORT.yml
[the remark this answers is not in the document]
Pierre-Sassoulas: 👍

Pierre-Sassoulas (approved):

PR #4753, opened 2021-07-26
Add unspecified-encoding checker #3826

### LINKED ISSUE #3826: PEP 597, require encoding kwarg in open call (and other calls that delegate to io.open)
there's a proposal to deprecate not passing an encoding to `open(...)`
https://www.python.org/dev/peps/pep-0597/#motivation

> Developers using macOS or Linux may forget that the default encoding is not always UTF-8.
> 
> For example, long_description = open("README.md").read() in setup.py is a common mistake. Many Windows users can not install the package if there is at least one non-ASCII character (e.g. emoji) in the README.md file which is encoded in UTF-8.
> 
> For example, 489 packages of the 4000 most downloaded packages from PyPI used non-ASCII characters in README. And 82 packages of them can not be installed from source package when locale encoding is ASCII. [1] They used the default encoding to read README or TOML file.
> 
> Another example is logging.basicConfig(filename="log.txt"). Some users expect UTF-8 is used by default, but locale encoding is used actually. [2]
> 
> Even Python experts assume that default encoding is UTF-8. It creates bugs that happen only on Windows. See [3] and [4].
> 
> Raising a warning when the encoding option is omitted will help to find such mistakes.

raise a warning similar to subprocess-run-check

### Additional context
Add any other context about the feature request here.

DanielNoord:

## Type of Changes

## Description

This adds an unspecified-encoding checker that adds a warning whenever open() is called without an explicit encoding argument.
This closes #3826.

I stumbled upon this issue today while I experienced the importance of adding `encoding` to `open()`. This is my first contribution to pylint so please let me know if anything needs to be changed or improved!


ChangeLog
```diff
@@ -6,8 +6,9 @@ What's New in Pylint 2.10.0?
 ============================
 Release date: TBA
 
-..
-  Put new features here and also in 'doc/whatsnew/2.10.rst'
+* Added ``unspecified-encoding``: Emitted when open() is called without specifying an encoding
```
Pierre-Sassoulas, line 9:
```suggestion
..
  Put new features here and also in 'doc/whatsnew/2.10.rst'
  
* Added ``unspecified-encoding``: Emitted when open() is called without specifying an encoding
```

pylint/checkers/stdlib.py
```diff
@@ -425,6 +426,13 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             "deprecated-decorator",
             "The decorator is marked as deprecated and will be removed in the future.",
         ),
+        "W1514": (
+            "Using open without explicitly specifying an encoding",
+            "unspecified-encoding",
+            "It is better to specify an encoding when opening documents. "
+            "Failing to do so can create problems when opening documents on other systems. "
```
Pierre-Sassoulas, line 433:
```suggestion
            "Using the system default implicitly can create problems on other operating systems. "
```

pylint/checkers/stdlib.py
```diff
@@ -573,6 +584,17 @@ def _check_open_mode(self, node):
             ):
                 self.add_message("bad-open-mode", node=node, args=mode_arg.value)
 
+    def _check_open_encoded(self, node):
```
Pierre-Sassoulas, line 587:
```suggestion
    def _check_open_encoded(self, node: astroid.Call) -> None:
```

Pierre-Sassoulas (changes requested):
Thank you for handling this issue. This looks great. I made some small comments but this is definitely going in ``2.10.0`` :).

tests/functional/u/unspecified-encoding_py3.py
```diff
@@ -0,0 +1,7 @@
+"""Warnings for using open() without specifying an encoding"""
+# pylint: disable=consider-using-with
+
+NAME = "foo.bar"
+open(NAME, "wb", encoding="utf-8")
+open(NAME, "w", encoding=None)  # [unspecified-encoding]
+open(NAME, "rb")  # [unspecified-encoding]
```
Pierre-Sassoulas, line 7: Let's add the example given in the issue :)
```suggestion
open(NAME, "rb")  # [unspecified-encoding]

with open(filename) as f:  # [unspecified-encoding]
    ...

with open(filename, encoding=None) as f:  # [unspecified-encoding]
    ...


with open(filename, encoding="utf8", errors="surrogateescape") as f:
    ...

locale_encoding = getattr(io, "LOCALE_ENCODING", None)
with open(filename, encoding=locale_encoding) as f:
    ...
```
Also we should probably check more use case to handle other open method (method named open, other method from io, etc.)

DanielNoord:
@Pierre-Sassoulas I merged all your suggestions + added checks for `b` mode, which does not need an encoding.

I do not really have any experience with other `open` methods. Perhaps you can add some additional test cases your self, or give me a link to some documentation on `open` methods you would like to see added.

Pierre-Sassoulas:
I think we should handle [``io.open``](https://docs.python.org/3/library/io.html#io.open), it's just an alias, but also [``io.open_code`` which is implicitly "rb"](https://docs.python.org/3/library/io.html#io.open_code) and make sure that we don't have false positive for it.

DanielNoord:
Added the check for `io.open` and `io.open_code`. 

However, the test suite is failing on some files which have not been edited by these commits while it is up to date with the main branch (besides the latest commit to issue forms).
Not sure what is going wrong here. Should I fix that in this PR? I don't immediately know where these errors are coming from.

Pierre-Sassoulas:
> Not sure what is going wrong here. 

Well what happened is you discovered style problem inside pylint with your new checker 😄 The new problem in pylint should be fixed in this merge request, we can discuss them here. Maybe it'll fix strange bug that happened only on some particular OS 🎉 

Or it can be problem in existing functional tests that are now detected. You can disable those new messages in the existing tests so there is no change in the expected output, or update the output if you think this make more sense.

Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
Always happy to help :) Also don't stress out, this won't be merged before 2.10, so not before 2.9.6 is out 🙂 

Regarding the latest change to fix the tests, it looks most of the changes are in functional tests. Please don't modify the functional test too much, you can disable the new checker with ``pylint: disable=unspecified-encoding`` at the top of the file (avoid creating a newline as the expected output will change completely if you add a line). The formatting changes seems like they are automated by your IDE and need to be reverted as having strange formattign is also part of what we're testing (``j+1;lambda ...`` for example). Sorry for not being clearer in my earlier comment.

DanielNoord:
Ah shit, I just changed some of the tests to include `encodings` and fixed line-numbering in the test data. Do you want me to remove those and add the override? Or is adding `encodings` to fix the errors also fine?
For me (Mac) the tests are also throwing a number of `c-extension-no-member`, `no-member` and `no-name-in-module` errors. Are those also related to the changes introduced by this PR? I don't think so, right?

Pierre-Sassoulas:
A mix of disabling, changing the code and checking that the warning is raised is fine, it should not be one sided though, more like a third of each. (To be faire disable or checking that it's raised is an easier change but you already did the work of changing the code so... 😄)

I don't think the test suite run in the CI for mac right now as we have linux and windows only.  So, its possible that we have some existing problem on mac.  We supposed mac/linux were close enough and not worth the added pipeline time. Apparently it's not. You can check on the master branch for the test that are already failing while we fix this.

DanielNoord:
Hmm, after fixing my formatter the tests are now also raising a number of issues with the code of pylint itself.
I guess `open()` is used often, who would have thought... :)

I would be very glad to fully fix all errors, but this won't be done tonight. Seeing as pylint has such a good test coverage already, I would feel bad if my contribution would lead to an increase in skipped checks. I will try and group some of the changes in similar commits for easier review. We can always squash those before the final merge.

pylint/checkers/stdlib.py
```diff
@@ -44,7 +45,7 @@
 from pylint.checkers import BaseChecker, DeprecatedMixin, utils
 from pylint.interfaces import IAstroidChecker
 
-OPEN_FILES = {"open", "file"}
+OPEN_FILES_MODE = {"open", "file"}
```
cdce8p, line 48:
```suggestion
OPEN_FILES_MODE = ("open", "file")
```
Why not use a tuple here? A set is a bit over the top for just two entries.

pylint/checkers/stdlib.py
```diff
@@ -494,8 +503,12 @@ def visit_call(self, node):
                 if inferred is astroid.Uninferable:
                     continue
                 if inferred.root().name == OPEN_MODULE:
-                    if getattr(node.func, "name", None) in OPEN_FILES:
+                    if getattr(node.func, "name", None) in OPEN_FILES_MODE:
                         self._check_open_mode(node)
+                    if getattr(node.func, "name", None) in {"open"} or getattr(
+                        node.func, "attrname", None
+                    ) in {"open"}:
```
cdce8p, line 510:
```suggestion
                    if getattr(node.func, "name", None) in ("open",) or getattr(
                        node.func, "attrname", None
                    ) in ("open",):
```
Same here. Tuples will work just fine.

cdce8p, line 510:
```suggestion
                    if (
                        isinstance(node.func, astroid.Name) and node.func.name in ("open",)
                        or isinstance(node.func, astroid.Attribute) and node.func.attrname in ("open",)
                    ):
```
Furthermore, it might be worth considering to replace the `getattr` calls with `isinstance` once.

doc/whatsnew/2.10.rst
```diff
@@ -12,7 +12,9 @@ Summary -- Release highlights
 New checkers
 ============
 
+* Added ``unspecified-encoding``: Emitted when open() is called without specifying an encoding
 
+  Closes #3826
```
cdce8p, line 17:
```suggestion
  Closes #3826

```
Just a small suggestion: Can you add an empty line her? Usually there should be two empty ones before a new section.

cdce8p (commented):
What a first PR 🔥 I like it a lot! Did a test run and it already worked beautifully.
As this will go in `2.10`, there is still some time. So you don't need to hurry.

pylint/checkers/stdlib.py
DanielNoord: I mainly did this because is was already as set. For example, on line 53 `ENV_GETTERS` is even a set with only one entry.
Want me to change both to a tuple? Or change this to a tuple and line 53 to a string?

pylint/checkers/stdlib.py
DanielNoord: Wouldn't a comparison with the string be fine as well? Or is comparison to a tuple better performance wise?

pylint/checkers/stdlib.py
cdce8p: Let's stick with a tuple for `ENV_GETTERS`. That way the `if` statement doesn't need to change if we ever need to add additional entries.

pylint/checkers/stdlib.py
cdce8p: Same reason as above. Although a string comparison might be possible, it would require us to change the `if` statement if we ever add additional entries. It might even be a good idea to create a new constant for `("open",)`.

DanielNoord:
Almost made it work. Again, sorry for taking so much of your time...

Action at [personal repo](https://github.com/DanielNoord/pylint/pull/1) shows that the PR succeeds for CPython 3.8 =<, but fails for 3.7, 3.6 and pypy. I thought appending `_py38` would make it automatically skip for 3.6 and 3.7 but I guess it is not that simple.
Furthermore, it is only one line that is failing (line 34) as `io.open_code()` was only added in 3.8. Perhaps there is a way to only make it skip that line for 3.6 and 3.7?

cdce8p:
You would need to create a file like this one: [generic_alias_typing.rc](https://github.com/PyCQA/pylint/blob/c04f92ef68e5ea779a60bfddb91dc677c5470fd0/tests/functional/g/generic_alias/generic_alias_typing.rc). With the same name as the test file, only `rc` as extension.

One more thing, I would appreciate if you wouldn't force push all your new changes. That makes it really difficult to review them. We can squash merge the PR in the end so it doesn't make a difference if there are 10+ commits. Usually a force push should only be used when doing a rebase to catch up with the latest changes on `main`.

DanielNoord:
I'm really sorry! I thought you might want a commit history that represents a possible final history.
Will refrain from force pushing in the future!

pylint/checkers/stdlib.py
```diff
@@ -494,8 +504,19 @@ def visit_call(self, node):
...
+                    if (
+                        isinstance(node.func, astroid.Name)
+                        and node.func.name in OPEN_FILES_MODE
+                    ):
                         self._check_open_mode(node)
+                    if (
+                        isinstance(node.func, astroid.Name)
+                        and node.func.name in OPEN_FILES_ENCODING
+                    ) or (
+                        isinstance(node.func, astroid.Attribute)
```
cdce8p, line 516:
```suggestion
                        or isinstance(node.func, astroid.Attribute)
```
The parentheses aren't really necessary as `and` binds tighter than `or`.

pylint/lint/pylinter.py
```diff
@@ -617,8 +619,9 @@ def set_option(self, optname, value, action=None, optdict=None):
                 except KeyError:
                     meth = self._bw_options_methods[optname]
                     warnings.warn(
-                        "%s is deprecated, replace it by %s"
-                        % (optname, optname.split("-")[0]),
+                        "{} is deprecated, replace it by {}".format(
+                            optname, optname.split("-")[0]
+                        ),
```
cdce8p, line 624: As this isn't related, it would be best to exclude the change here. Even if technically it's an improvement.

cdce8p (approved):
Looks good 🚀
Left some last comments

pylint/lint/pylinter.py
DanielNoord: Want me to create another commit/PR to do this? Or just let it be?

cdce8p: If you really like to, you can create another one. But if you ask me, I would let it be. It doesn't harm anybody and is still valid code.

Pierre-Sassoulas:
Haaa. I made a mistake I should have waited for 2.9.6 to be released before merging.

cdce8p:
> Haaa. I made a mistake I should have waited for 2.9.6 to be released before merging.

In that case, what do you think about releasing `2.10` next? As far as I've seen, there aren't any major bugfixes that need to go into `2.9.6`. All can move to `2.10`

DanielNoord:
Also note that I didn't have time to merge @cdce8p's last comment

Pierre-Sassoulas:
> Also note that I didn't have time to merge @cdce8p's last comment

Don't worry, I applied the suggestion :)

I'm going to release 2.9.6 with #4765  and we'll start to merge the 2.10 features for the next release after that. I moved the remaining MR in 2.9.6 to 2.10.1.

PR #4758, opened 2021-07-27
Add ignored-parents option to design checker

### LINKED ISSUE #3057: Count ancestors only in user code
Sometimes I need to inherit from classes defined in third-party libs. Sometimes these classes have long inheritance chain. I can't do anything about this, but still get too-many-ancestors.

I want there was an option to count ancestors only in "local" code, so pylint will complain only about the part of problem I can fix.

### Additional context
Here's example where it would be useful: https://stackoverflow.com/questions/22834392/understanding-too-many-ancestors-from-pylint/43090406

9999years:
## Type of Changes

## Description

Allow users to specify classes to ignore while counting parent classes.

When one of the ignored classes is detected in the ancestor tree, it _and its parents_ are ignored.  Suppose we have the following inheritance diagram:

```
       F
      /
  D  E
   \/
    B  C
     \/
      A      # class A(B, C): ...
```

If we add `E` to `ignored-parents`, then the classes counted for the `too-many-ancestors` check will be `B`, `D`, and `C`.

If we have another class `G(F)`, then `F` *will* be counted -- ancestors of ignored parents are pruned from the tree for counting, but we don't go the other way and exclude all parents of ignored parents.

Partially closes #3057 by offering a configuration mechanism to ignore non-user classes from counting.

## Rationale

`typing.Protocol` is ignored from counting, but I'm using `typing_extensions`, so `typing_extensions.Protocol` is increasing the count and triggering the `too-many-ancestors` message.


ChangeLog
```diff
@@ -14,6 +14,11 @@ What's New in Pylint 2.9.6?
 ===========================
 Release date: TBA
 
+* Added ``ignored-parents`` option to the design checker to ignore classes
```
Pierre-Sassoulas, line 17: I think this should be in 2.10 as it's a new feature.

pylint/checkers/design_analysis.py
```diff
@@ -237,6 +238,35 @@ def _count_methods_in_class(node):
     return all_methods
 
 
+def _get_parents(
+    node: nodes.ClassDef, ignored_parents: FrozenSet[str]
```
Pierre-Sassoulas, line 242: Good job adding the typing 👍

pylint/checkers/design_analysis.py
```diff
@@ -379,11 +418,11 @@ def _ignored_argument_names(self):
...
     def visit_classdef(self, node: nodes.ClassDef):
         """check size of inheritance hierarchy and number of instance attributes"""
-        nb_parents = sum(
-            1
-            for ancestor in node.ancestors()
-            if ancestor.qname() not in STDLIB_CLASSES_IGNORE_ANCESTOR
+        parents = _get_parents(
+            node, STDLIB_CLASSES_IGNORE_ANCESTOR.union(self.config.ignored_parents)
         )
+        print(parents)
```
Pierre-Sassoulas, line 424: Debug print forgotten here :)

tests/checkers/unittest_design.py
```diff
@@ -0,0 +1,44 @@
...
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+from pathlib import Path
+
+import astroid
+
+from pylint.checkers import design_analysis
+from pylint.testutils import CheckerTestCase, set_config
+
+REGR_DATA_DIR = str(Path(__file__).parent / ".." / "regrtest_data")
```
Pierre-Sassoulas, line 13:
```suggestion

```

tests/checkers/unittest_design.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2021 Rebecca Turner <rturner@starry.com>
```
Pierre-Sassoulas, line 1: This is automated, don't worry about it next time :)

Pierre-Sassoulas (changes requested):
Thank you for the merge request and the added option, this looks nice. I left some comments. Could you also add a new functional test in tests/functional/.../too_many_ancestors.py with a configuration file, please ?

9999years:
Added a functional test -- should I keep the `tests/checkers/unittest_design.py` file I added or is just the functional test OK?

Pierre-Sassoulas (approved):
Thank you for implementing the functional tests ! It's true that the unittest could be a duplicate now. I don't know for sure if it's covering more line than simply the functional test.

9999years:
I'll leave it -- it won't hurt and it'll make it easier for the next person who wants to unit test the design checker.

Pierre-Sassoulas:
There will be a slight delay before merging as we want to release 2.9.6 first and this is going in 2.10 🙂

Pierre-Sassoulas:
Thank you again for your work on this, this will be available in 2.10.0 !

Issue #4693, opened 2021-07-08
False positive no-member


fluffy-critter:
### Steps to reproduce

In https://github.com/PlaidWeb/Authl/blob/dc53e065e7cfe5e21bcfb169aea4588293531da8/tests/handlers/test_fediverse.py#L89:

```
    result = handler.check_callback(result.url, parse_args(result.url), {})
    assert isinstance(result, disposition.Verified)
    assert result.identity == 'https://mastodon.example/@moo'
    assert result.redir == 'qwerpoiu'
    assert result.profile == { 
        'name': 'moo friend',
        'bio': 'a cow',
        'avatar': 'https://placekitten.com/1280/1024',
        'homepage': 'https://moo.example',
        'pronouns': 'moo/moo'
    }

```

### Current behavior

```
tests/handlers/test_fediverse.py:93:11: E1101: Instance of 'Error' has no 'profile' member (no-member)
```

### Expected behavior

In this particular test, `result` can never be `disposition.Error` and there's even a previous assertion to ensure that it's [`disposition.Verified`](https://github.com/PlaidWeb/Authl/blob/dc53e065e7cfe5e21bcfb169aea4588293531da8/authl/disposition.py#L35) which does have the `'profile'` member. I assume that there's some erroneous caching of `result`'s type from another test.

### pylint --version output

Result of `pylint --version` output:

```
pylint 3.0.0-a4
astroid 2.6.2
Python 3.8.6 (default, Jan  6 2021, 12:14:27) 
[Clang 12.0.0 (clang-1200.0.32.28)]
```

jolaf:
I've encountered this error too – twice already.

PR #4764, opened 2021-07-28
Omit no-member error if guarded behind if stmt

### LINKED ISSUE #1990: Add `isinstance` in `if` condition check as a special case
### Steps to reproduce
1. Write this simple code and save it:
```py
import subprocess

try:
    subprocess.check_call(['ls', '-']) # Deliberately made error in this line
except Exception as err:
    if isinstance(err, subprocess.CalledProcessError):
        print 'Subprocess error occured. Return code:', err.returncode
        # Error in above line.......................here^^^
    else:
        print 'An error occured:', str(err)
```
2. Run `pylint` on this file.
3. Output of Pylint:
```
$ pylint pylint_test.py 
No config file found, using default configuration
************* Module pylint_test
C:  1, 0: Missing module docstring (missing-docstring)
W:  5, 7: Catching too general exception Exception (broad-except)
E:  7,56: Instance of 'Exception' has no 'returncode' member (no-member)

Your code has been rated at 0.00/10 (previous run: 2.86/10, -2.86)
```

### Current behavior
I is showing Error in line: 7 which should not be an error.

### Expected behavior
It should not show Error in line-7 as I am already checking if `err` is an instance of `subprocess.CalledProcessError` which has member `returncode`.

### pylint --version output
```
No config file found, using default configuration
pylint 1.8.3, 
astroid 1.6.2
Python 2.7.12 (default, Dec  4 2017, 14:50:18) 
[GCC 5.4.0 20160609]
```

### LINKED ISSUE #4168: false positive no-member after isinstance
### Steps to reproduce

Given a file `a.py`:
```python

mixed_tuple = (b"a", b"b", b"c", b"d")
byte_tuple = [arg.encode('utf8') if isinstance(arg, str) else arg for arg in mixed_tuple]

for arg in mixed_tuple:
    if isinstance(arg, str):
        print(arg.encode('utf8'))
    else:
        print(arg)
```

As soon as the mixed_tuple contains at least one string, the no-member warning is no longer shown.

### Current behavior
Result of `pylint a.py`:
```
************* Module pixy.a
src/pixy/a.py:4:14: E1101: Instance of 'bytes' has no 'encode' member (no-member)
src/pixy/a.py:8:14: E1101: Instance of 'bytes' has no 'encode' member (no-member)

Your code has been rated at -6.67/10 (previous run: -15.00/10, +8.33)
```

### Expected behavior
No warning should be shown

### pylint --version output

```
pylint 2.7.2
astroid 2.5.1
Python 3.9.1
```

This behaviour has been introduced with 2.7.0

cdce8p:
## Type of Changes

## Description

Omit `no-member` errors if attribute / method access is guarded behind if statement.

This change doesn't implement `control-flow` logic. Instead I chose to check for an `IF` / `IFExp` node in the node parents and evaluate it. That doesn't solve all issue, especially not those with `assert` #4693, but I would never the less consider it an improvement.

As this is a larger change, I would like to target `2.10` for it.

### Example
```py
from datetime import datetime

value = "Hello World"
value.isoformat()  # [no-member]

if isinstance(value, datetime):
    value.isoformat()  # This is now fixed and doesn't emit 'no-member' anymore
```

--
Ref #1162 (not closed, but issue would be addressed)
Closes #1990
Closes #4168

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,12 @@ Release date: TBA
 
   Close #4120
 
+* Don't emit ``no-member`` error if guarded behind if statement.
+
+  Ref #1162
+  Closes #1990
+  Closes #4168
+
 
 What's New in Pylint 2.9.6?
 ===========================
--- a/doc/whatsnew/2.10.rst
+++ b/doc/whatsnew/2.10.rst
@@ -24,5 +24,12 @@ Other Changes
 * Performance of the Similarity checker has been improved.
 
 * Added ``time.clock`` to deprecated functions/methods for python 3.3
+
 * Added ``ignored-parents`` option to the design checker to ignore specific
   classes from the ``too-many-ancestors`` check (R0901).
+
+* Don't emit ``no-member`` error if guarded behind if statement.
+
+  Ref #1162
+  Closes #1990
+  Closes #4168
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -454,6 +454,7 @@ def _emit_no_member(node, owner, owner_name, ignored_mixins=True, ignored_none=T
         * the owner is a class and the name can be found in its metaclass.
         * The access node is protected by an except handler, which handles
           AttributeError, Exception or bare except.
+        * The node is guarded behind and `IF` or `IFExp` node
     """
     # pylint: disable=too-many-return-statements
     if node_ignores_exception(node, AttributeError):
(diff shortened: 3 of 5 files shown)


Pierre-Sassoulas (commented):
I'm surprised that this does not break more functional tests. I think this is a nice way to remove some false positive using astroid inference.

Pierre-Sassoulas (approved):

PR #4768, opened 2021-07-28
Extended `consider-using-tuple` check to cover `in` comparisons

cdce8p:
## Type of Changes

## Description
Extend `consider-using-tuple` check from the `CodeStyleChecker` extension to recognize lists / sets in `in` comparisons.

**Some examples**
```py
x in [1, 2, 3]  # [consider-using-tuple]

if x in [1, 2, 3]:  # [consider-using-tuple]
    pass

assert x in {1, 2, 3}  # [consider-using-tuple]
```

Previously the check was only used inside `For` looks and `Comprehensions`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -36,6 +36,10 @@ Release date: TBA
 
   Closes #3878
 
+* ``CodeStyleChecker``
+
+  * Extended ``consider-using-tuple`` check to cover ``in`` comparisons.
+
 
 What's New in Pylint 2.9.6?
 ===========================
--- a/doc/whatsnew/2.10.rst
+++ b/doc/whatsnew/2.10.rst
@@ -17,6 +17,14 @@ New checkers
   Closes #3826
 
 
+Extensions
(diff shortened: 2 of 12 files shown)


Pierre-Sassoulas (approved):
Well done!

ChangeLog
```diff
@@ -3,21 +3,20 @@ Pylint's ChangeLog
...
-
 What's New in Pylint 2.10.0?
 ============================
 Release date: TBA
 
-* Added ``ignored-parents`` option to the design checker to ignore specific
-  classes from the ``too-many-ancestors`` check (R0901).
-
-  Partially closes #3057
-
```
Pierre-Sassoulas, line 15: Maybe we could minimise the diff here so there is less conflict in other MR.

cdce8p: I moved the Changelog style changes to #4770

Pierre-Sassoulas (approved):

PR #4771, opened 2021-07-29
Add ``forgotten-debug-statement`` checker

### LINKED ISSUE #3692: New Check: Warning for `breakpoint` calls. no-breakpoint
Since python [3.7](https://docs.python.org/3.7/whatsnew/3.7.html) there is a new builtin [`breakpoint()`](https://docs.python.org/3.7/whatsnew/3.7.html#pep-553-built-in-breakpoint). Although there are environment variables you can set in production to make `breakpoint()` a no-op, the default behavior (unset) is to allow `breakpoint()`. Further, setting this flag makes it circuitous to use a breakpoint when you need one in a shell, or doing some debugging.

How often do you intentionally commit breakpoints into your packages? Projects? Production? Even in test environments its unusual.

A new warning :)

### Additional context
From [PEP 553](https://www.python.org/dev/peps/pep-0553/#environment-variable) re: breakpoints:

> Completely disabling all accidental breakpoint() calls pushed to production. This could be accomplished by setting PYTHONBREAKPOINT=0 in the execution environment. Another suggestion by reviewers of the PEP was to set PYTHONBREAKPOINT=sys.exit in this case.

### Uncertain

Should `sys.breakpointhook()` also be disallowed?

DanielNoord:

## Type of Changes

## Description

This adds a checker for calls to ``breakpoint()`` or ``sys.breakpointhook()``.
This closes #3692


DanielNoord:
After thinking about this some more:
Do we want to check for redefined functions? And if so, do we want to produce the message even if they are redefined, or do we want to mute the message after redefinition?

cdce8p (changes requested):
> Do we want to check for redefined functions? And if so, do we want to produce the message even if they are redefined, or do we want to mute the message after redefinition?

That should be covered with inference. You could add an example for it to the test.

DanielNoord:
> That should be covered with inference. You could add an example for it to the test.

I found that the following code passes without throwing the error:
```
def breakpoint():
    pass
```

However, the following code does throw an error. 
```
def overwrite():
    pass
sys.breakpointhook = overwrite
sys.breakpointhook()
```
I don't think anyone should ever want to overwrite functions like in the second example, but should we consider such code? And if so, much we allow it?

pylint/checkers/stdlib.py
```diff
@@ -55,6 +55,7 @@
 SUBPROCESS_POPEN = "subprocess.Popen"
 SUBPROCESS_RUN = "subprocess.run"
 OPEN_MODULE = "_io"
+BREAKPOINTS = ("builtins.breakpoint", "sys.breakpointhook")
```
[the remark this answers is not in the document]
DanielNoord: Probably, although then we might need a new message code as well? 
`no-breakpoint` seems counterintuitive for checking `pdb.set_trace`.

cdce8p:
> However, the following code does throw an error.
> 
> ```python
> def overwrite():
>     pass
> sys.breakpointhook = overwrite
> sys.breakpointhook()
> ```
> 
> I don't think anyone should ever want to overwrite functions like in the second example, but should we consider such code? And if so, much we allow it?

Tbh I haven't seen anybody reassign stdlib function yet. So I would just ignore that case.

--
I initially though you meant the other way round, i.e. setting an alias. That is common.
```py
b = breakpoint
b()
```
That should through a `no-breakpoint` error and can / should be covered by the tests.

pylint/checkers/stdlib.py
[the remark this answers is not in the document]
Pierre-Sassoulas: ``forgotten-debug-statement`` ?

tests/functional/f/forgotten_debug_statement_py37.py
```diff
@@ -0,0 +1,10 @@
+# pylint: disable=missing-docstring
+
+import sys
+
+breakpoint()  # [forgotten-debug-statement]
+sys.breakpointhook()  # [forgotten-debug-statement]
+b = breakpoint
+b()  # [forgotten-debug-statement]
+bsys = sys.breakpointhook
+bsys()  # [forgotten-debug-statement]
```
cdce8p, line 10: Just one of those alias checks is already fine. They won't yield a different result for different functions.

tests/functional/f/forgotten_debug_statement_py37.py
```diff
@@ -0,0 +1,10 @@
+# pylint: disable=missing-docstring
+
+import sys
+
+breakpoint()  # [forgotten-debug-statement]
+sys.breakpointhook()  # [forgotten-debug-statement]
```
cdce8p, line 6: Can you add `pdb.set_trace()` here as well?

cdce8p (commented):
Just a small last comment

tests/functional/f/forgotten_debug_statement_py37.py
DanielNoord: Shit, totally forgot about this! Doing this now

cdce8p (approved):
LGTM, thanks @DanielNoord!

doc/whatsnew/2.10.rst
```diff
@@ -16,6 +16,10 @@ New checkers
 
   Closes #3826
 
+* Added ``forgotten-debug-statement``: Emitted when breakpoint() or sys.breakpointhook() calls are found
```
Pierre-Sassoulas, line 27:
```suggestion
* Added ``forgotten-debug-statement``: Emitted when ``breakpoint``, ``pdb.set_trace`` or ``sys.breakpointhook()`` calls are found
```

Pierre-Sassoulas (approved):

doc/whatsnew/2.10.rst
```diff
@@ -24,6 +24,10 @@ New checkers
 
   Closes #4365
 
+* Added ``forgotten-debug-statement``: Emitted when ``breakpoint``, ``pdb.set_trace`` or ``sys.breakpointhook()`` calls are found
```
Pierre-Sassoulas, line 27:
```suggestion
* Added ``forgotten-debug-statement``: Emitted when ``breakpoint``, ``pdb.set_trace`` or ``sys.breakpointhook`` calls are found
```

ChangeLog
```diff
@@ -48,6 +48,10 @@ Release date: TBA
 
   * Extended ``consider-using-tuple`` check to cover ``in`` comparisons.
 
+* Added ``forgotten-debug-statement``: Emitted when breakpoint() or sys.breakpointhook() calls are found
```
Pierre-Sassoulas, line 51:
```suggestion
* Added ``forgotten-debug-statement``: Emitted when ``breakpoint``, ``pdb.set_trace`` or ``sys.breakpointhook`` calls are found
```

Issue #3249, opened 2019-11-17
Assignment Expression Reports superfluous-parens False Positive


t-mart:
An `if` statement that tests a negated (`not`) parenthesized [Assignment Expression](https://www.python.org/dev/peps/pep-0572/) _without any other operators (See "Non-issues" section below)_ is reported by Pylint to have superfluous parentheses.

I believe this is a false positive. Removing the parentheses is invalid syntax.

### Steps to reproduce
1. Write an if statement that tests a negated (`not`) [Assignment Expression](https://www.python.org/dev/peps/pep-0572/), such as:

```python
if not (my_var := False):
    print(my_var)
```

2. Run pylint against that code.

### Current behavior
Pylint reports `superfluous-parens:Unnecessary parens after 'not' keyword`.

### Expected behavior
Pylint should not report about this syntax.

Removing the parentheses is a syntax error: `SyntaxError: cannot use named assignment with operator`.

### pylint --version output
```
pylint 2.4.4
astroid 2.3.3
Python 3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:21:23) [MSC v.1916 32 bit (Intel)]
```

### Non-issues
- If the negated assignment-expression `if` statement contains another binary operator, no Pylint report is generated about it. (Or at least, that seems to be the only difference.)

  For example, this slightly-modified snippet from [What’s New In Python 3.8](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions) does not cause Pylint to report:

  ```python
  if not (n := len(list(range(11))) > 10:
      print(f"List is too long ({n} elements, expected <= 10)")
  ```

PCManticore:
Thanks for the report!

t-mart:
Ah, it seems this issue is more general than I first thought. **The negation has nothing to do with it.**

Example that also causes `superfluous-parens`:

```python
if (foo := True):
    print(foo)
```

huwcbjones:
> Ah, it seems this issue is more general than I first thought. **The negation has nothing to do with it.**
> 
> Example that also causes `superfluous-parens`:
> 
> ```python
> if (foo := True):
>     print(foo)
> ```

I would say that example is a legit case where we _do_ want to raise this warning. In the example below, I believe the expected behaviour would be to raise a superfluous warning for 2, a syntax error for 3, and no warning for 1 & 4

```python
# 1: Legal python, not superfluous
if foo := True:
    print(foo)

# 2: Legal python, superfluous
[...]

PR #4784, opened 2021-08-01
Fix false positives for ``superfluous-parens``

### LINKED ISSUE #2818: superfluous-parens in if-expression are not superfluous
### Steps to reproduce
Create file t.py:
```
def f(b):
    return [a for a in (b if b else [b])]
```
Related to #574 but this example has an if-expression instead of short circut.

### Current behavior
```
t.py:2:0: C0325: Unnecessary parens after 'in' keyword (superfluous-parens)
```
Note that following this suggestion actually creates a SyntaxError:
```
def g(b):
    return [a for a in b if b else [b]]
```
```
t.py:2:0: E0001: invalid syntax (<unknown>, line 2) (syntax-error)
```
### Expected behavior
This warning is not emitted in this case.

### pylint --version output
```
pylint 2.3.1
astroid 2.2.5
Python 3.7.2+ (default, Feb 27 2019, 15:41:59) 
[GCC 8.2.0]
```

### LINKED ISSUE #3429: Disable rule for multi-line statement
Having something like:

```python
def func(
    foo
):
    pass
```

Pylint gives:

```
foo.py:1:0: C0102: Black listed name "foo" (blacklisted-name)
...
```

This cannot be disabled like:

```python
def func(  # pylint: disable=blacklisted-name
    foo  # pylint: disable=blacklisted-name
):  # pylint: disable=blacklisted-name
    pass
```

None of these work. A [SO answer](https://stackoverflow.com/a/25779494/165629) suggests to:

```python
# pylint: disable=blacklisted-name
def func(
    foo
):
    # pylint: enable=blacklisted-name
    pass
```

But that's pretty annoying...

I would prefer if this was possible:

```python
def func(
    foo
):  # pylint: disable=blacklisted-name
    pass
```

### LINKED ISSUE #3608: Incorrectly identifying parens as superfluous in comprehensions with ternary operators
Pylint incorrectly identifies parens as superfluous when they are used to isolate a ternary operator that determines the iterable inside of a list comprehension (or set comprehension, dict comprehension etc.).

### Steps to reproduce
1. Create a file called `scratch.py` with contents:
```
""" A test script. """

print([i**2 for i in (range(10) if 7 > 5 else range(3))])
```
2. Evaluate this using the latest version of pylint:
    ``$ pylint scratch.py --enable=all``

### Current behavior

Pylint reports that there is one issue with this code: 
``scratch.py:3:0: C0325: Unnecessary parens after 'in' keyword (superfluous-parens)``

### Expected behavior
Pylint should report no issues with this code since the parenthesis are required. Following Pylints suggestion and changing the file to:
```
""" A test script. """

print([i**2 for i in range(10) if 7 > 5 else range(3)])
```
results in invalid Python code:
```
$ python scratch.py
  File "scratch.py", line 3
    print([i**2 for i in range(10) if 7 > 5 else range(3)])
                                            ^
SyntaxError: invalid syntax
```

### pylint --version output
```
$ pylint --version
pylint 2.5.2
astroid 2.4.1
Python 3.8.1 (default, Mar  7 2020, 12:46:28)
[Clang 10.0.0 (clang-1000.10.44.4)]
```

### LINKED ISSUE #4346: False positive ``superfluous-parens`` with walrus operator
I've came across a `superfluous-parens` false positive when using the walrus operator:

### Steps to reproduce

Given a file `test.py`:
```python
"""Test example."""
if (var := 3) in [1, 2, 3]:
    print('Hello world !')
```

We obtain the following:
### Current behavior

Result of `pylint test.py`:
```
************* Module test
test.py:2:0: C0325: Unnecessary parens after 'if' keyword (superfluous-parens)
```

### Expected behavior

Normally, this should not raise any errors as the use of parentheses here is relevant. Because, otherwise the value of `var` will be the inclusion boolean instead of the value itself. 

### pylint --version output

Result of pylint --version output on my local macOS:

```
pylint 2.7.4
astroid 2.5.2
Python 3.8.8 (v3.8.8:024d8058b0, Feb 19 2021, 08:48:17) 
[Clang 6.0 (clang-600.0.57)]
```

DanielNoord:

## Type of Changes

## Description

This fixes the false positives identified in #2818, #3249, #3608 & #4346
All false positives reported fell under keywords before walrus operator or if-keyword within generators/comprehension.
This closes #2818, closes #3429, closes #3608, closes #4346

#3249 reported a lot of different false positives, numerous of which also included code that is not valid python code.
I think this should fix all false positives within valid code, but before closing that issue somebody might need to run through the issue again to see if all reports have been taken care of.


Pierre-Sassoulas (approved):
Wow, so much issue closed in one MR :) Thank you for doing the project management works too and checking all the issue with the problem !

Pierre-Sassoulas (commented):
Could you add something in the changelog and in 2.10.0's whatsnew ? :)

DanielNoord:
Not sure what is going on:
[This run](https://github.com/DanielNoord/pylint/actions/runs/1087751735) worked and I thought I added lines to the changelog. Next run the CI failed.. Going to (try and) fix!

ChangeLog
```diff
@@ -60,6 +60,13 @@ Release date: TBA
 
   Closes #4673
 
+* Fix false positives for ``superfluous-parens``
```
Pierre-Sassoulas, line 63:
```suggestion
* Fix false positives for ``superfluous-parens`` with walrus operator, ternary operator and inside list comprehension.
```

doc/whatsnew/2.10.rst
```diff
@@ -54,3 +54,10 @@ Other Changes
   Ref #1162
   Closes #1990
   Closes #4168
+
+* Fix false positives for ``superfluous-parens``
```
Pierre-Sassoulas, line 58:
```suggestion
* Fix false positives for ``superfluous-parens`` with walrus operator, ternary operator and inside list comprehension.
```

Pierre-Sassoulas (commented):
This is strange, the tests aren't usually flaky. Maybe the first run was with the current master ?

DanielNoord:
I guess so. However, when I copy the code of ``format.py`` to my local installation the following code works and produces correct output:
```python
if not (my_var := False):
    print(my_var)

print([i ** 2 for i in (range(10) if 7 > 5 else range(3))])


def a_function():
    return (x for x in ((3, 4)))  # Bad at col 25


print([i ** 2 for i in ((3, 4))])  # Bad at col 25

if not ((odd := 1)):  # Bad at col 9
    pass
```

However, the following messages raise errors in the test of `unittest_format.py`:
```python 
( # Line 187
    Message("superfluous-parens", line=1, args="in"),
    "return (x for x in ((3, 4)))",
    0,
)

(Message("superfluous-parens", line=1, args="not"), "if not ((odd := 1)):", 0), # Line 215
 
```

I believe those messages are tested in the test file in the 4th and 5th test. Any idea why it does work when replacing the code in the locally installed `pylint` but not in the test suite?

Pierre-Sassoulas:
I don't see any problem with the unittest but this is arder to udnerstand than functional tests. Could you try to create a functional tests in ``tests/functional/s/superfluous_parenthesis.py`` and see how it goes (=copy paste this code and signal expected message with ``# [message-name]`` ? You might have to create another file for warlus operator that only work in python 3.10 with a corresponding ``.rc`` file.

tests/functional/s/superfluous_parens.py
```diff
@@ -1,19 +1,27 @@
 """Test the superfluous-parens warning."""
 from __future__ import print_function
-# pylint: disable=unneeded-not
+
+# pylint: disable=unneeded-not, missing-function-docstring, unnecessary-comprehension
 i = 3
-if (i == 5):  # [superfluous-parens]
+if i == 5:
```
Pierre-Sassoulas, line 6: This one is valid and should still work after we fix the false positives.

Pierre-Sassoulas (changes requested):
We should not remove valid use cases from the functional tests.

tests/functional/s/superfluous_parens.py
DanielNoord: Shit, this is due to black auto-removing the parens and me wondering why the error should show up for `if i == 5`

tests/functional/s/superfluous_parens.py
```diff
@@ -1,19 +1,27 @@
...
     pass
 if not (i == 5):  # [superfluous-parens]
     pass
 if not (3 or 5):
     pass
-for (x) in (1, 2, 3):  # [superfluous-parens]
+for x in (1, 2, 3):
```
DanielNoord, line 12: Same goes for this one..

pylint/checkers/format.py
```diff
@@ -393,19 +395,25 @@ def _check_keyword_parentheses(
...
                 if tokens[i + 1].string in (":", ")", "]", "}", "in") or tokens[
                     i + 1
                 ].type in (tokenize.NEWLINE, tokenize.ENDMARKER, tokenize.COMMENT):
-                    # The empty tuple () is always accepted.
                     if contains_walrus_operator and walrus_operator_depth - 1 == depth:
-                        # Reset variable for possible following expressions
-                        contains_walrus_operator = False
-                        continue
+                        return
+                    # The empty tuple () is always accepted.
```
DanielNoord, line 416: I moved this comment to 416 because I believe it cover 417-418 and is misplaced in the current code.

DanielNoord:
Force-pushed to clean up the commit history.

First commit covers splitting the functional tests for `superfluous-parens` into a base and `python 3.8` version to test the walrus operator. As mentioned in the OP, #3249 covers many different reports of false positives, but I think the additions to the functional tests cover all of them. I've also added some tests for cases of `(((my_var)))` which were not discussed, but were initially failing.

Second commit covers changes to the code and to the unittest. The codes passes the functional tests but fails the unittests. I don't know why, so will need some guidance to fix this before this PR can be undrafted.

Once again, sorry for creating this PR before it was ready. I thought it was finished but apparently not. I'm going to fix my local development environment before doing any additional work so I have better test results before creating PR's. With 15+ tests always failing I often miss cases where 16 instead of 15 tests fail...

Pierre-Sassoulas:
> Once again, sorry for creating this PR before it was ready. I thought it was finished but apparently not. 

Don't worry, the continuous integration is here for checking on every environment and python-interpreters and it's expected that fixing everything might take some time :)

DanielNoord:
Do you have any idea why the unit tests are failing?
I have not used those before and just copied the pattern used on other places in the file, but my additions seem to fail. Testing similar patterns with the functional testing does seem to work...

Pierre-Sassoulas:
I made some tests locally, I pushed the result. I don't know exactly why but ``self.checker._check_keyword_parentheses`` was not resulting from the same output than for functional tests. I changed to ``process_token`` and I think the test case that is failing is a real issue now.

DanielNoord:
I think the unittest is failing on `( Message("superfluous-parens", line=1, args="in"), "return (x for x in ((3, 4)))", 0)`, right?
However, that case seems to be tested by the functional test on line 21 I think. I don't see what makes pylint respond differently to both cases.

Pierre-Sassoulas:
I rarely use the unittest myself, if it's a pain to do let's convert to functional tests :)

DanielNoord:
All tests are now passing. Please see if you think all cases are covered in the tests, especially those mentioned in #3249.

tests/functional/s/super/superfluous_parens_walrus_py38.py
```diff
@@ -0,0 +1,24 @@
...
+if odd := is_odd(i): # [undefined-variable]
+    pass
+if not (foo := 5):
+    pass
+
+if not ((x := y)):  # [superfluous-parens]
+    pass
+
+if ((x := y)):   # [superfluous-parens]
+    pass
```
Pierre-Sassoulas, line 24:
```suggestion
    pass

assert (ret := str(42))

# False positive (str, tuple)
A = [int(n) for n in (nums := '123')]
B = [int(n) for n in (nums := ('1', '2', '3'))]

# No false positive (list, set, dict)
C = [int(n) for n in (nums := ['1', '2', '3'])]
D = [int(n) for n in (nums := {'1', '2', '3'})]
E = [int(n) for n in (nums := {'1': 1, '2': 2, '3': 3})]

# False positive (brackets only)
F = [int(n) for n in (nums := ['1'])]
G = [int(n) for n in (nums := {'1'})]
H = [int(n) for n in (nums := {'1': 1})]
I = [int(n) for n in (nums := ('1'))]
J = [int(n) for n in (nums := ('1', ))]

# No false positive (brackets plus comma)
K = [int(n) for n in (nums := ['1', ])]
L = [int(n) for n in (nums := {'1', })]
M = [int(n) for n in (nums := {'1': 1, })]

# False positives (comma outside brackets)
N = [int(next(iter(n))) for n in (nums := (['1'], ['2'], ['3']))]
O = [int(next(iter(n))) for n in (nums := ({'1'}, {'2'}, {'3'}))]
P = [int(next(iter(n))) for n in (nums := ({'1': 1}, {'2': 2}, {'3': 3}))]

# False positives (aliasing)
NUMS_LIST = ['1', '2', '3']
NUMS_SET = {'1', '2', '3'}
NUMS_DICT = {'1': 1, '2': 2, '3': 3}
Q = [int(n) for n in (nums := NUMS_LIST)]
R = [int(n) for n in (nums := NUMS_SET)]
S = [int(n) for n in (nums := NUMS_DICT)]
T = [int(n) for n in (nums := [*NUMS_LIST])]
U = [int(n) for n in (nums := NUMS_LIST[:])]

# 2D arrays with numpy (since they allow commas within indexing)
import numpy
ARRAY_2D = numpy.zeroes((3, 3), dtype=bool)
V = not (vals := ARRAY_2D[:, :])  # No false positive
X = not (vals := ARRAY_2D[2, 2])  # No false positive
W = not (vals := ARRAY_2D[::])  # False positive
Y = not (vals := ARRAY_2D[2:2])  # False positive
Z = not (just_want_to_finish_the_alphabet := True)  # False positive

x = ("Version " + __version__) in output 
y = 2 + (5 * 3)
assert "Version " + __version__ in output
assert "" + ("Version " + __version__) in output
assert ("Version " + __version__) in output

class YieldRegression3249:
    @classmethod
    def from_json(cls, json):
        """Yields records from a JSON-ish dict."""
        modules = json.pop('modules', None) or ()
        yield (order := super().from_json(json))
        for module in modules:
            yield OrderedModule(order=order, module=Module(module))
            
            
```

I think those are the relevant use case from #3249, it's not super a problem if we have a few duplicate :)

Pierre-Sassoulas (changes requested):
Thank you for fixing the tests ! I think once we incorporate use case from the pinned issue we can merge :)

tests/functional/s/super/superfluous_parens_walrus_py38.py
```diff
@@ -0,0 +1,86 @@
...
+
+assert (ret := str(42))
+
+# False positive (str, tuple)
+A = [int(n) for n in (nums := '123')]
```
DanielNoord, line 29: A-R are not valid python code (you cannot use the walrus operator there).
I think we can therefore exclude them?

tests/functional/s/super/superfluous_parens_walrus_py38.py
```diff
@@ -0,0 +1,86 @@
...
+# 2D arrays with numpy (since they allow commas within indexing)
+import numpy
+ARRAY_2D = numpy.zeroes((3, 3), dtype=bool)
+V = not (vals := ARRAY_2D[:, :])  # No false positive
+X = not (vals := ARRAY_2D[2, 2])  # No false positive
+W = not (vals := ARRAY_2D[::])  # False positive
+Y = not (vals := ARRAY_2D[2:2])  # False positive
+Z = not (just_want_to_finish_the_alphabet := True)  # False positive
+
+x = ("Version " + __version__) in output
```
DanielNoord, line 73: I don't really understand these tests. What is output supposed to be? A string? A list? And where is __version__ taken from.

tests/functional/s/super/superfluous_parens_walrus_py38.py
```diff
@@ -0,0 +1,86 @@
...
+assert ("Version " + __version__) in output
+
+class YieldRegression3249:
+    @classmethod
+    def from_json(cls, json):
+        """Yields records from a JSON-ish dict."""
+        modules = json.pop('modules', None) or ()
+        yield (order := super().from_json(json))
+        for module in modules:
+            yield OrderedModule(order=order, module=Module(module))
```
DanielNoord, line 86: I think OrderedModule is a non-standard class. Can we replace this by a simple function with a yield statement?

tests/functional/s/super/superfluous_parens_walrus_py38.py
Pierre-Sassoulas: Yeap, we can also remove Z :)

tests/functional/s/super/superfluous_parens_walrus_py38.py
Pierre-Sassoulas: It's one of the example a use case in 3249 (https://github.com/PyCQA/pylint/issues/3249#issuecomment-683424549). `__version__` was probably a string or this would fail. I think the parenthesis is really superfluous here

```suggestion
x = ("Version " + __version__) in output # [superfluous-parenthesis]
```

tests/functional/s/super/superfluous_parens_walrus_py38.py
Pierre-Sassoulas: :+1:  Let's replace OrderedModule by something builtin

DanielNoord:
I have updated the tests, let me know what you think.

I have removed these two tests:
```python
I = ("Version " + G) in F
assert "" + ("Version " + G) in F
```
The current checker focuses on unnecessary parens after keywords. These tests cover unnecessary parens with string combination. Because of 
```python
if tokens[start + 1].string != "(":
            return
``` 
on line `373` in `format.py` and the fact that we only run when Keywords are present we won't be able to catch all of these errors. For example the first test described, as it doesn't include a keyword. I think opening a new issue for "Superfluous parens checker for string combinations" would be the best way forward.

Pierre-Sassoulas:
> I think opening a new issue for "Superfluous parens checker for string combinations" would be the best way forward.

Sure thing, this MR already fix so much issue at once already !  What do you think of letting the test case in the functional test without the `# [superfluous-parentheses]` and with a TODO ? (Like what was done for 3429 before)

DanielNoord:
Sounds good to me, but perhaps first create the issue so we can link to it in the TODO?

Pierre-Sassoulas (approved):
Amazing work, thank you ! :clap:

Issue #4776, opened 2021-07-30
False positive ``consider-using-tuple`` for ``x in set``, because frozen sets are faster


samueltardieu:
### Bug description

The change introduced in #4768 is unwarranted.

Considering the following code:

```python
def f(a):
    if a in {"abc", "def", "ghi", "klm", "nop"}:
        print(a)

def g():
    for a in {"abc", "def", "ghi", "klm", "nop"}:
        print(a)
```

it suggests to use a tuple in both cases:

```bash
$ pylint /tmp/t.py -v
************* Module t
/tmp/t.py:2:12: R6102: Consider using an in-place tuple instead of set (consider-using-tuple)
/tmp/t.py:6:13: R6102: Consider using an in-place tuple instead of set (consider-using-tuple)
```

While the second suggestion is right (in-place sets should not be used for iterating over their content), the first one (added in #4768) is a very bad advice. In the case of a membership lookup a set is faster, compiled as a `frozenset` by Python, and the right tool for the job.

I would suggest that this commit be reverted before the release as it encourages bad practices. A new style check could rather suggest to replace in-place lists and tuples in membership tests by sets when the data type is known to be hashable (as strings or numbers are for example).

### Command used

```shell
pylint /tmp/t.py
```

### Pylint output

```shell
************* Module t
/tmp/t.py:2:12: R6102: Consider using an in-place tuple instead of set (consider-using-tuple)
/tmp/t.py:6:13: R6102: Consider using an in-place tuple instead of set (consider-using-tuple)
```

### Expected behavior

I would expect to see the second suggestion (which is sound), and not the first one (which is a bad advice).

### Pylint version

```shell
pylint 2.10.0-dev0 (5e5f48d05fb6f1322a7420d6ea7586966b08f80a)
```

### OS / Environment

ArchLinux

Pierre-Sassoulas:
Hey thank you for opening the issue. Do you have a source for the performance comparison between the two ?

samueltardieu:
```python
>>> def in_set(a): return a in {"abc", "def", "ghi", "klm", "nop"}
...
>>> def in_tuple(a): return a in ("abc", "def", "ghi", "klm", "nop")
...
>>> from timeit import timeit
>>> timeit(lambda: in_set("foobar"))
0.08363517999532633
>>> timeit(lambda: in_tuple("foobar"))
0.1204142420028802
```

[...]

PR #4787, opened 2021-08-02
Add ``possible-f-string-as-string`` checker

### LINKED ISSUE #2507: Detect when f-string-syntax is used in a string, but not marked as an f-string
### Problem statement

Consider the following codeblock

```python
name = "Katie"

stmt1 = f"Hello {name}"
stmt2 = "Hello {name}"
```

While the block is correct syntax-wise, the content is almost certainly logically invalid, as seen below: 

```python
>>> print(stmt1)
Hello Katie
>>> print(stmt2)
Hello {name}
```

`stmt2` is a valid string, but declares string interpolation formatting that is compatible with f-strings, but the string does not have the `f` prefix to invoke the interpolation. 

PyLint should detect strings that use f-string-like syntax, but have not been declared as f-strings, in Python 3.6+ 

This could either be by:
 * attempting to evaluate the content within `{}`, and if valid Python, return a warning
 * attempting to evaluate the entire string as an f-string, and if the evaluation does not equal the original string, return a warning

**Edge Case One**: Having `{` and `}` characters are perfectly valid, but on the edge case that the content within them is a reference to an in-scope variable, or some other expression, it is likely the string was meant to be an f-string. This would be desirable to catch as a linting error.

**Edge Case Two**: `f'...'` is effectively shorthand for `'...'.format(locals())`; so this:
```
mystring = 'hello {name}'
print(mystring.format(name='Katie'))
```
is entirely legal, and isn't an error. There is a possibly design decision about whether this linting check should be

* enforced by default
* enforceable as an option
* enforceable if the string that triggers the problem doesn't have '.format' invoked on it in the same method.

### Additional context
I understand PyLint already warns against f-strings in logging, but I can't locate the base functionality to detect f-strings in the first place. I'd be willing to help implement this feature, but would require some mentoring.

DanielNoord:

## Type of Changes

## Description

This checks if text in strings in between `{}`'s are variables.
The var the string is assigned to is also checked for a format() call.
If this does not happen, the string should probably be a f-string and a message is emitted.
This closes #2507

## Reason for this PR being a draft

Ideally I would also add a check for the following code:
```python
x = "This is a calculation {1 + 1}" # bad

x = f"This is a calculation {1 + 1}" # good
```
Can we check that  `1 + 1` is a calculation without calling `eval`?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -68,6 +68,10 @@ Release date: TBA
 
   Closes #4681
 
+* Added ``possible-f-string-as-string``: Emitted when variables are used in normal strings in between "{}"
+
+  Closes #2507
+
 
 What's New in Pylint 2.9.6?
 ===========================
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -26,6 +26,7 @@
 # Copyright (c) 2020 Anthony <tanant@users.noreply.github.com>
 # Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
 # Copyright (c) 2021 Peter Kolbus <peter.kolbus@garmin.com>
+# Copyright (c) 2021 Daniel van Noord <13665637+DanielNoord@users.noreply.github.com>
 
 
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
@@ -203,6 +204,12 @@ MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
         "f-string-without-interpolation",
         "Used when we detect an f-string that does not use any interpolation variables, "
         "in which case it can be either a normal string or a bug in the code.",
+    ),
+    "W1310": (
+        "Using an string which should probably be a f-string",
+        "possible-f-string-as-string",
+        "Used when we detect a string that uses '{}' with a local variable inside. "
+        "This string is probably meant to be an f-string.",
     ),
 }
 
@@ -891,6 +898,57 @@ class StringConstantChecker(BaseTokenChecker):
             # another character can always be consumed whole: the second
             # character can never be the start of a new backslash escape.
             index += 2
+
+    @check_messages("possible-f-string-as-string")
+    def visit_const(self, node: astroid.Const):
+        self._detect_possible_f_string(node)
+
+    def _detect_possible_f_string(self, node: astroid.Const):
+        """Check whether strings include local/global variables in '{}'
+        Those should probably be f-strings's
+        """
+
+        def detect_if_used_in_format(node: astroid.Const, assign_name: str) -> bool:
+            """Check if the node is used in a call to format() if so return True"""
+            # the skip_class is to make sure we don't go into inner scopes, but might not be needed per se
+            for attr in node.scope().nodes_of_class(
+                astroid.Attribute, skip_klass=(astroid.FunctionDef,)
+            ):
+                if isinstance(attr.expr, astroid.Name):
+                    if attr.expr.name == assign_name and attr.attrname == "format":
+                        return True
+            return False
+
+        if node.pytype() == "builtins.str" and not isinstance(
+            node.parent, astroid.JoinedStr
+        ):
+            # Find all pairs of '{}' within a string
+            inner_matches = re.findall(r"(?<=\{).*?(?=\})", node.value)
+            if len(inner_matches) != len(set(inner_matches)):
+                return
+            if inner_matches:
+                for match in inner_matches:
+                    # Check if match is a local or global variable
+                    if match := node.scope().locals.get(
+                        match
+                    ) or node.root().locals.get(match):
+                        assign_node = node
+                        while not isinstance(assign_node, astroid.Assign):
+                            assign_node = assign_node.parent
+                        if isinstance(assign_node.value, astroid.Tuple):
+                            node_index = assign_node.value.elts.index(node)
+                            assign_name = assign_node.targets[0].elts[node_index].name
+                        else:
+                            assign_name = assign_node.targets[0].name
+                        if not detect_if_used_in_format(node, assign_name):
+                            self.add_message(
+                                "possible-f-string-as-string",
+                                line=node.lineno,
+                                node=node,
+                            )
+                            return
+                    else:
+                        return
 
 
 def register(linter):
--- a/tests/functional/p/possible_f_string_as_string.py
+++ b/tests/functional/p/possible_f_string_as_string.py
@@ -1,0 +1,12 @@
+# pylint: disable=missing-module-docstring, invalid-name
+var = "string"
+var_two = "extra string"
+
+x = f"This is a {var} which should be a f-string"
+x = "This is a {var} used twice, see {var}"
+x = "This is a {var} which should be a f-string"  # [possible-f-string-as-string]
+x = "This is a {var} and {var_two} which should be a f-string"  # [possible-f-string-as-string]
+x1, x2, x3 = (1, 2, "This is a {var} which should be a f-string")  # [possible-f-string-as-string]
+
+y = "This is a {var} used for formatting later"
+z = y.format(var="string")
--- a/doc/whatsnew/2.10.rst
+++ b/doc/whatsnew/2.10.rst
@@ -28,6 +28,10 @@ New checkers
 
   Closes #3692
 
+* Added ``possible-f-string-as-string``: Emitted when variables are used in normal strings in between "{}"
+
+  Closes #2507
+
 
 Extensions
 ==========
--- a/tests/functional/p/possible_f_string_as_string.txt
+++ b/tests/functional/p/possible_f_string_as_string.txt
@@ -1,0 +1,3 @@
+possible-f-string-as-string:7:4::Using an string which should probably be a f-string:HIGH
+possible-f-string-as-string:8:4::Using an string which should probably be a f-string:HIGH
+possible-f-string-as-string:9:20::Using an string which should probably be a f-string:HIGH


pylint/checkers/strings.py
```diff
@@ -26,6 +26,7 @@
 # Copyright (c) 2020 Anthony <tanant@users.noreply.github.com>
 # Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
 # Copyright (c) 2021 Peter Kolbus <peter.kolbus@garmin.com>
+# Copyright (c) 2021 Daniel van Noord <13665637+DanielNoord@users.noreply.github.com>
```
Pierre-Sassoulas, line 29: This is done automatically, you can skip it next time ;)

pylint/checkers/strings.py
```diff
@@ -204,6 +205,12 @@
         "Used when we detect an f-string that does not use any interpolation variables, "
         "in which case it can be either a normal string or a bug in the code.",
     ),
+    "W1310": (
+        "Using an string which should probably be a f-string",
+        "possible-f-string-as-string",
```
Pierre-Sassoulas, line 210:
```suggestion
        "f-string-syntax-in-standard-string" 
```
```suggestion
        "f-string-syntax-in-string" 
```
```suggestion
        "f-string-syntax-in-str" 
```
```suggestion
        "forgotten-f-string-prefix"
```
```suggestion
        "possible-forgotten-f-prefix"
```
Some suggestion for another name, I think the current one is a little long.

pylint/checkers/strings.py
```diff
@@ -892,6 +899,55 @@ def process_non_raw_string_token(
             # character can never be the start of a new backslash escape.
             index += 2
 
+    @check_messages("possible-f-string-as-string")
+    def visit_const(self, node: astroid.Const):
+        self._detect_possible_f_string(node)
```
Pierre-Sassoulas, line 904:
```suggestion
        if node.pytype() == "builtins.str" and not isinstance(
            node.parent, astroid.JoinedStr
        ):
            self._detect_possible_f_string(node)
```
I would put the conditional here to diminish the indentation in the other function.

pylint/checkers/strings.py
```diff
@@ -892,6 +899,55 @@ def process_non_raw_string_token(
...
+            node.parent, astroid.JoinedStr
+        ):
+            # Find all pairs of '{}' within a string
+            inner_matches = re.findall(r"(?<=\{).*?(?=\})", node.value)
+            if len(inner_matches) != len(set(inner_matches)):
+                return
+            if inner_matches:
+                for match in inner_matches:
+                    # Check if match is a local or global variable
+                    if node.scope().locals.get(match) or node.root().locals.get(match):
```
Pierre-Sassoulas, line 932:
```suggestion
                    if not(node.scope().locals.get(match) or node.root().locals.get(match)):
                        return
                    ... 
```
I would do a premature return to avoid the long if / else.

pylint/checkers/strings.py
```diff
@@ -204,6 +205,12 @@
         "Used when we detect an f-string that does not use any interpolation variables, "
         "in which case it can be either a normal string or a bug in the code.",
     ),
+    "W1310": (
+        "Using an string which should probably be a f-string",
```
Pierre-Sassoulas, line 209:
```suggestion
        "The %s syntax imply an f-string but the leading 'f' is missing",
```
If we could provide the "{local_var}" as an argument it would be great so the message is easier to understand.

Ie final message would be:
``The '{var}' and '{var}' syntax imply an f-string but the leading 'f' is missing``
``The '{var}' and '{var_two}' syntax imply an f-string but the leading 'f' is missing``

tests/functional/p/possible_f_string_as_string.py
```diff
@@ -0,0 +1,12 @@
...
+var_two = "extra string"
+
+x = f"This is a {var} which should be a f-string"
+x = "This is a {var} used twice, see {var}"
+x = "This is a {var} which should be a f-string"  # [possible-f-string-as-string]
+x = "This is a {var} and {var_two} which should be a f-string"  # [possible-f-string-as-string]
+x1, x2, x3 = (1, 2, "This is a {var} which should be a f-string")  # [possible-f-string-as-string]
+
+y = "This is a {var} used for formatting later"
+z = y.format(var="string")
```
Pierre-Sassoulas, line 12: I think we need some more tests here. We usually have a tons of false positive when adding a new checker because there is a lot of use case and some coders do really unexpected things when there is so much users. In the worst case it can be crashes. I guess we could start by adding actual pylint f-string and remove the f here. There can be expression inside f-strings too for example ``"This is an example with a list: {''.join(examples) + 'well...' }" that should not crash.``

Pierre-Sassoulas (commented):
Thank you for working on this checker, much appreciated !

tests/functional/p/possible_f_string_as_string.py
DanielNoord: I wonder if there is a better way to do this. Similar to the problem identified in the OP with calculations not involving any variables. Those should ideally also be added.

Can we use `eval`? And just check and see if the `{}` and its contents change when they are evaluated as part of an f-string?

Pierre-Sassoulas: I'd rather we not use ``eval``, pylint is launched on a lot of continuous integration and this would create an attack vector to execute random code contained in a string with containing ``{}``. Or did I misunderstand ?

DanielNoord: No, you understood and this is what I expected. I hoped there would perhaps be a 'utils' function that allow us to check for this.
Any other good ideas how to do this? Especially the calculations without variables seems difficult to do without some sort of evaluation/comparison between both string types.

Pierre-Sassoulas: Could you elaborate on the "calculations without variables" part ? 😄

DanielNoord: 'f"This example {1 + 1}"' for example...

Pierre-Sassoulas: Well, I suggested hastily something that is probably not very reasonable. We'll always have false positive if we check for more than local variable (think of someone generating python code inside a dict). Maybe the sane thing to do it to limit false positive even if we have a lot of false negative for this check and warn only if a local var is detected. But maybe we could create an optional checker that is really aggressive and use ``ast.parse`` in order to check if something look like valid python code then warn for every string with parseable python code inside ``{}`` ?

DanielNoord: I think that is the best way forward. However, does that aggressive check need to be optional? I don't have as much experience with python code as you do, but does it happen often that valid python code is intentionally put inside `{}` within a normal string?

Pierre-Sassoulas: > does it happen often that valid python code is intentionally put inside {} within a normal string?

You're right, probably not... pylint is notorious for its false positives and I wanted to be cautious 😄 

I'm even struggling to come up with an example. Here's a silly example that work because I think ``first_key: "int"`` is valid python:

```python
generated_python = """
first_key = "version"
DYNAMIC_DICT = {
first_key: "int"
}
"""
```

DanielNoord: There is also an example in the original issue. However the valid python also needs to return or represent something. Your example is not a valid f-string. I'm not sure if astroid allows us to check for this?

Pierre-Sassoulas: You're right it also needs to return something... Maybe there would not be that much false positive after all. astroid can parse python too and infer the value of things, but there is not check for "this string is a valid python code" or "this string is a valid f-string as far as I know".

DanielNoord: I don't know the best way forward here. After googling a bit, f-strings seem to be fairly complicated as well as tricky to parse. 
Adding a checker for local variables in normal strings containing "{}" could be a partial solution, but I feel whatever we end up writing for that checker will be incompatible with a full-fledged `possible-f-string` checker. Therefore, I don't know if that warrants the time-investment.

DanielNoord: I think a problem here is that `ast.parse` is not able to recognise `bar` as a variable from the local and global scope, and I don't think we are able to tell it do so.
I have played around with letting `ast.parse` parse all lines up until the line with the string, changing the string to an f-string and comparing the difference. However, if the string is at the bottom of a 300+ lines file this would have major performance impact (I think). 
Other than that, I had no real ideas of how to supply local and global variables included in the f-string to `ast.parse`)

Pierre-Sassoulas: Ha yes, you're right. We have to parse it in the context of our function. We can probably try to infer ``'\n'.join(bar)`` as if it's code inside the function at the position of the string using astroid (or ast.parse for that matter but when analyzing the core around the string we're supposed to have a parsed context in astroid already, so it should cost less). This is clearly non trivial and we should consider performance implication as you said. How about we focus on the easy case with only local_variable ?

DanielNoord: I think the problem with fixing the easy case now is that it is almost certainly going to be incompatible with a full implementation of a f-string checker. For me, this is a reason to consider whether the time investment needed to make this work (although most of the work is already present in this PR) is warranted.

Pierre-Sassoulas: Clearly. In term of maintainer time as well, we'll get swamped with false negatives report as soon as we add the check in a live version. User will expect it to work properly on all use case (as they should). Might as well do it properly from the beginning and avoid the pain.  I'll add the work in progress label.

DanielNoord:
For future contributors stumbling across this PR.

We would like to add a `possible-f-string-as-string` checker which checks for normal strings that seem to be f-strings.
The following example show some basic test cases:
```python
param = "string"
"This is a string" # good
"This is a {param}" # bad 
f"This is a {param}" # good

"This is a calculation: 1 + 1" # good
"This is a calculation: {1 + 1}" # bad
f"This is a calculation: {1 + 1}" # good

"This is a nice string" # good
"This is a {"nice" + param}" # bad
f"This is a {"nice" + param}" # good
```

The code in this PR is (almost) able to test the first three cases by looking for regex matches in between `{}` and seeing if the matches can be found in the local or global scope.
The second and third batch of tests is more problematic. We need a way to test whether the `{}` contain valid python code. The second batch is (fairly) easy, the third is problematic. We need to evaluate the code within the `{}` within the current global and local scope to see if any of the variables are in there (such as `param` in the example). We currently have no solution to do this without any considerable performance impact.

TLDR: How to evaluate the code in between `{}` within a normal string in the context of that string without too much performance impact?

Crissal1995:
Sorry to pop out from nowhere, but this PR got my interest. 

Have you read the [f-string PEP](https://www.python.org/dev/peps/pep-0498)? There are two interesting sections, [Code equivalence](https://www.python.org/dev/peps/pep-0498/#code-equivalence) and [Expression evaluation](https://www.python.org/dev/peps/pep-0498/#expression-evaluation), that shows how the f-string is treated internally, what's its representation with `format` statements, and how is parsed with `ast`. 

I think that expressions should be evaluated as done in the PEP, i.e. with [`ast.parse('(' + expression + ')', '<fstring>', 'eval')`](https://docs.python.org/3/library/ast.html#ast.parse); because the mode is set to `eval`, [only a single statement](https://docs.python.org/3/library/functions.html#compile) can be executed, so I think that should be ok for security reasons.

If accepted, this should address every problem related with valid and invalid expressions inside curly braces.

DanielNoord:
@Crissal1995 This is indeed interesting, thanks for the tip!

I have not tested this, so please forgive me if I my questions are non-relevant, but:
1) Can't a single statement also produce security issues? I have no real experience with this security and `eval`, but I am thinking of `print(organization_secret)` or using an already imported `requests` module to send sensitive information from an automatically run CI? 
2) How can we pass the context of the string to `ast.parse('(' + expression + ')', '<fstring>', 'eval')`. It seems as if we should still pass all lines up until the line with the string to the ast-parser to infer any local or global variables. For performance reasons, this is something we should (obviously) avoid.

Crissal1995:
> 1. Can't a single statement also produce security issues? I have no real experience with this security and `eval`, but I am thinking of `print(organization_secret)` or using an already imported `requests` module to send sensitive information from an automatically run CI?

Excuse me, but I don't understand this question. If you run PyLint, you are supposed to be in your own execution environment, so you know your secrets and have no interest to `print` them, or am I missing something?

> 2. How can we pass the context of the string to `ast.parse('(' + expression + ')', '<fstring>', 'eval')`. It seems as if we should still pass all lines up until the line with the string to the ast-parser to infer any local or global variables. For performance reasons, this is something we should (obviously) avoid.

This is indeed a good question; however, I think that should be as PyLint already works. It's not needed to execute all the code up until your expression; rather, your variable needs to be in the scope of local or global variables.
To explain me better:
- if I execute `ast.parse("(y +/ 2)", "<fstring>", "eval")`, then I got a `SyntaxError`.
- If I execute `ast.parse("(y + 2)", "<fstring>", "eval")`, however, I got a valid `ast.Expression`; then I need to check every leaf, and see if it's a `ast.Name` or not - like `y` is in this case.
- 
When I got a `Name`, I can check its `id` - that is the string representation of its name - to see if it is found in the current scope. If not, raise an error; otherwise you're good to go, there is no need to execute all code before.

If I try to access an object variable, like `myvar.z`, there is just a need of checking if `myvar` is present, not that is present **and** has the attribute `z`.

DanielNoord:
> Excuse me, but I don't understand this question. If you run PyLint, you are supposed to be in your own execution environment, so you know your secrets and have no interest to `print` them, or am I missing something?

As I said, I'm not quite sure how CI's and secrets behave but I am imagining a situation where an automatically run CI prints or sends an organization's secrets as a result of evaluating f-strings. I'm not sure if the is actually a feasible line of attack, but we should be 100% certain of this before merging such behaviour into the tool.

> * if I execute `ast.parse("(y +/ 2)", "<fstring>", "eval")`, then I got a `SyntaxError`.
> * If I execute `ast.parse("(y + 2)", "<fstring>", "eval")`, however, I got a valid `ast.Expression`; then I need to check every leaf, and see if it's a `ast.Name` or not - like `y` is in this case.

This seems like the right way forward. If I can find the time I might start working on this implementation, but I am more than happy for somebody else to start working on this.

> If I try to access an object variable, like `myvar.z`, there is just a need of checking if `myvar` is present, not that is present **and** has the attribute `z`; this will be done at runtime from Python itself.

I think we should do this check, as Pylint tries to inform users of errors before running their code. Relying on errors on runtime is therefore not truly in the spirit of pylint.

Pierre-Sassoulas:
> As I said, I'm not quite sure how CI's and secrets behave but I am imagining a situation where an automatically run CI prints or sends an organization's secrets as a result of evaluating f-strings. I'm not sure if the is actually a feasible line of attack, but we should be 100% certain of this before merging such behaviour into the tool.

I was thinking the same. Let's say a public open source project use pylint on Github and has a release job. In it's CI in a job, someone change a string to ``"{print(os.environ('PYPI_TOKEN')}", if we eval ``print(os.environ('PYPI_TOKEN')`` without special care,  the attacker can see the value of the secret in the CI log (maybe they would also needs to make a test fail so the output is printed, but it's not that sophisticated). Then they can immediately publish a compromised package. This would be really bad.

Crissal1995:
> As I said, I'm not quite sure how CI's and secrets behave but I am imagining a situation where an automatically run CI prints or sends an organization's secrets as a result of evaluating f-strings. I'm not sure if the is actually a feasible line of attack, but we should be 100% certain of this before merging such behaviour into the tool.
> 
With `ast.parse` you only `compile` valid Python code, parsing its tree, but not execute a single line of code.
Furthermore, in compiling phase, there is also a flag automatically set by `ast` that is [`ast.PyCF_ONLY_AST`](https://docs.python.org/3/library/ast.html#ast.PyCF_ONLY_AST).
> [`ast.PyCF_ONLY_AST`] Generates and returns an abstract syntax tree instead of returning a compiled code object.

In fact, when you `eval` a node generated by `ast.parse`, you get a nice `TypeError`.

```python
>>> import ast
>>> f1 = lambda x: ast.parse("(" + x + ")", "<fstring>", "eval")
>>> f2 = lambda x: compile(x, "<fstring>", "eval")
>>> eval(f1("1"))
TypeError: eval() arg 1 must be a string, bytes or code object
>>> eval(f2("1"))
1
>>> x = "MY SECRET"
>>> y = "print(x)"
>>> eval(f1(y))
TypeError: eval() arg 1 must be a string, bytes or code object
>>> eval(f2(y))
MY SECRET
```

> > If I try to access an object variable, like `myvar.z`, there is just a need of checking if `myvar` is present, not that is present **and** has the attribute `z`; this will be done at runtime from Python itself.
> 
> I think we should do this check, as Pylint tries to inform users of errors before running their code. Relying on errors on runtime is therefore not truly in the spirit of pylint.
> 
But being Python a dynamically typed language, this turns out to be a difficult task, if not impossible.
The only possible scenario to see this happen is to force type hinting for every object passed in the context of a fstring, otherwise how can we know if `myvar` is an int, a `Foo` object or what else - and thus has in fact the `z` attribute?

For example, when you `parse("(foo.x)")`, you get back an `Expression` made out of an [`Attribute`](https://docs.python.org/3/library/ast.html#ast.Attribute). Without executing the code up until that point, you cannot really tell if `foo` really has that attribute; that's because I suggest to let the execution fail, or at least warn the user that this particular string - `"This is my var: {foo.x}"` - is **surely** a missed fstring.

Pierre-Sassoulas:
I tried something with ast.parse, but I fail to find a test case where the returned parsed ast is not an expression. I think it create a syntax error the way it's parsed if it isn't one anyway, so we probably don't need to check that it's an ast.Expression. Other than that checking that a valid python expression contain a a variable from the global or local scope seems complicated in term of performance so we can probably not do it. What we probably need to do is to check that the string is not used with format later on because the number of false positive will be huge if we don't.

DanielNoord:
I was also toying around with `ast.parse` but I failed to find an easy way to implement it with the expected behaviour. Perhaps we need to think about this the other way around: "why would `{...}` be included in a normal string and see if we can test for those cases?

pylint/checkers/strings.py
```diff
@@ -188,7 +189,7 @@
     "W1307": (
         "Using invalid lookup key %r in format specifier %r",
         "invalid-format-index",
-        "Used when a PEP 3101 format string uses a lookup specifier "
+        "Used when a PEP 3101 format string uses a lookup specifier "  # pylint: disable=possible-forgotten-f-prefix
```
DanielNoord, line 192: Shouldn't this be on the next line?

pylint/checkers/strings.py
```diff
@@ -210,6 +211,12 @@
         "Used when we detect a string that does not have any interpolation variables, "
         "in which case it can be either a normal string without formatting or a bug in the code.",
     ),
+    "W1311": (
+        "The %s syntax imply an f-string but the leading 'f' is missing",
```
DanielNoord, line 215: This should be `implies`, my mistake..

pylint/checkers/strings.py
```diff
@@ -210,6 +211,12 @@
...
+        "possible-forgotten-f-prefix",
+        "Used when we detect a string that uses '{}' with a local variable inside. "
```
DanielNoord, line 217: Depending on whether we are able to do this, this might need to be "with a local variable or valid expression inside".

pylint/checkers/strings.py
```diff
@@ -912,12 +919,51 @@ def process_non_raw_string_token(
...
-            self._detect_u_string_prefix(node)
+            self._detect_possible_f_string(node)
+        self._detect_u_string_prefix(node)
+
+    def _detect_possible_f_string(self, node: astroid.Const):
+        """Check whether strings include local/global variables in '{}'
+        Those should probably be f-strings's
+        """
+
+        def detect_if_used_in_format(node: astroid.Const) -> bool:
```
DanielNoord, line 936: Why did you remove this function? I think it worked quite okay in my previous commits?

tests/functional/p/possible_forgotten_f_prefix.py
```diff
@@ -0,0 +1,38 @@
...
+var = "string"
+var_two = "extra string"
+
+x = f"This is a {var} which should be a f-string"
+x = "This is a {var} used twice, see {var}"
+x = "This is a {var} which should be a f-string"  # [possible-forgotten-f-prefix]
+x = "This is a {var} and {var_two} which should be a f-string"  # [possible-forgotten-f-prefix, possible-forgotten-f-prefix]
+x1, x2, x3 = (1, 2, "This is a {var} which should be a f-string")  # [possible-forgotten-f-prefix]
+
+y = "This is a {var} used for formatting later"  # [possible-forgotten-f-prefix]
```
DanielNoord, line 11: This should not raise the message, as it is used later in a `format()` call. However, you changed this, so perhaps I am missing a reason why this should raise?

tests/functional/p/possible_forgotten_f_prefix.py
```diff
@@ -0,0 +1,38 @@
...
+z = y.format(var="string")
+
+g = "This is a {another_var} used for formatting later"  # [possible-forgotten-f-prefix]
```
DanielNoord, line 14: This should not raise the message, as it is used later in a `format()` call.

pylint/checkers/strings.py
Pierre-Sassoulas: Probably a problem with pylint but the warning is signaled on this one.

tests/functional/p/possible_forgotten_f_prefix.py
Pierre-Sassoulas: At first I thought it was acceptable (contrary to another_var where the var do not exist in the local scope). But I changed my mind, and I think you're right.

pylint/checkers/strings.py
Pierre-Sassoulas: I tried to simply detect a format call without checking assign_name but failed. (This is really just a work in progress it can be reverted)

DanielNoord:
I couldn't get the pre-commit to pass without fixing some of the `consider-using-tuple` messages.
This PR is probably not the place to fix those. I am wondering how the current CI is not failing on the main branch when it should raise these messages?

tests/functional/p/possible_forgotten_f_prefix.py
```diff
@@ -0,0 +1,53 @@
+"""Check various forms of strings which could be f-strings without a prefix"""
+# pylint: disable=invalid-name, line-too-long, pointless-string-statement, pointless-statement
+# pylint: disable=missing-function-docstring
+
+# Check for local variable interpolation
+PARAM = "string"
+PARAM_TWO = "extra string"
+
+A = f"This is a {PARAM} which should be a f-string"
+B = "This is a {PARAM} used twice, see {PARAM}"
```
Pierre-Sassoulas, line 10:
```suggestion
B = "This is a {PARAM} used twice, see {PARAM}"  # [possible-forgotten-f-prefix, possible-forgotten-f-prefix]
```

tests/functional/p/possible_forgotten_f_prefix.py
```diff
@@ -0,0 +1,53 @@
...
+M = "This is a calculation: {1 + 1}"  # [possible-forgotten-f-prefix]
+
+# Check invalid Python code
+N = "This is {invalid /// python /// inside}"
+O = "This is {not */ valid python.}"
+P = "This is {def function(): return 42} valid python but not an expression"
+
+
+def function():
+    return 42
```
Pierre-Sassoulas, line 38: We can remove that, it was a check for in-lined function definition in ``P``
```suggestion
```

pylint/checkers/typecheck.py
```diff
@@ -1780,7 +1780,7 @@ def visit_compare(self, node):
             return
 
         op, right = node.ops[0]
-        if op in ["in", "not in"]:
```
Pierre-Sassoulas, line 1783: I don't know why it was not caught in the CI before, but I think this warning is too opinionated, we can disable it in the pylintrc instead. There's a discussion to replace it by ``use-set-for-membership`` as sets are faster for membership, which is what we should use here but clearly not related to this MR. See #4776 / https://github.com/PyCQA/pylint/pull/4841, if you're interested about this issue :)

Pierre-Sassoulas (approved):
Looks very nice, thank you for fixing my draft 👍

pylint/checkers/typecheck.py
DanielNoord: What should I do for now? I can't push any commits without changing these..

tests/functional/p/possible_forgotten_f_prefix.py
DanielNoord: I have intentionally left this as a case without warnings. Whenever users use the same `parameter` twice it is likely that they are going to use `format()` later on.
See lines 950-952:
```python
        # If a variable is used twice it is probably used for formatting later on
        if len(inner_matches) != len(set(inner_matches)):
            return
```

pylint/checkers/typecheck.py
Pierre-Sassoulas: Please disable the check by adding ``consider-using-tuple``after ``disable=`` inside the ``pylintrc`` file (at the root of the repository).

DanielNoord: Wouldn't that disable the check even after the updates of #4841 have been added?

Pierre-Sassoulas: I would not mind if it's disabled it's very opinionated. It will permit to keep working on this and I'll check the root cause meanwhile :)

ChangeLog
```diff
@@ -85,6 +85,10 @@ Release date: TBA
 
   Closes #4681
 
+* Added ``possible-forgotten-f-prefix``: Emitted when variables are used in normal strings in between "{}"
```
DanielNoord, line 88:
```suggestion
* Added ``possible-forgotten-f-prefix``: Emitted when variables or valid expressions are used in normal strings in between "{}"
```

DanielNoord:
This finally seems to be going towards a finished state. The functional tests are now catching a large number of ways to use `format()` and `{...}`.

The only thing we won't catch is:
```python
A = "This is a {PARAM} which should be a f-string"  # [possible-forgotten-f-prefix]
A = "This is a {PARAM} which should be a f-string"
A.format(PARAM)
```
This is because I use quite a hacky way to see if an assignment is later used in a `format()` call. We could do with a function such as `variable_used_later_in_call(variable_name, function_name)`, which is basically what I wrote now, but I think this is dependent on actual control flow.
My hacky solution is to create a set of variable names and subscripts that are later used by `format()`. However, this does not allow us to infer whether a variable gets reassigned before that call. Therefore, the above example will fail to produce a message.

Left to do before undrafting:
- Remove changes of `[]` to `()`
- See if messages should be ignored in the cases the `pre-commit` is referring to

Pierre-Sassoulas (approved):
Nice job ! The false negative on ``"{PARAM}".format(PARAM)`` is acceptable imo. It's a KeyError: the user will notice or another warning will handle (I'm surprised there seem to be none right now).

DanielNoord:
I think I'm going crazy...
Found two additional cases where the message should not be reported:
```python
U = r"a{1}"

M = "This is a {parameter} used for formatting later"

def func_three():
    x = M.format(parameter=PARAM)
    return x
```
The former is especially important considering how often `{...}` is used in regex. Do we know of a good way to infer if a string was originally a raw string? Using `node.kind` is not possible sadly.

The latter can be solved by also checking inner scopes but this would make the following example no longer produce a message:
```python
A = "{PARAM}" # bad

def function_one():
    A = "{PARAM}"
    return A.format(PARAM="string")
```
Although users should refrain from redefining global variables, the `A` one line 1 would be found in the scope of `function_one` even though it should raise an error.

I'm starting to think there might be too many use cases for `{...}` in normal strings and this checker might only create a large amount of false positives..

Pierre-Sassoulas:
>  Using node.kind is not possible sadly.

Would it be possible if we do the check only for python 3.8 or superior ? If the alternative is to not have a check at all the choice is easy.

DanielNoord:
`.kind` only works for the `u` prefix, not for the `r` prefix. The `ast` module doesn't seem to differentiate between strings and raw strings... We could perhaps do something with checking the `coloffset - 1`? Or do you know if this problem came up before?

Pierre-Sassoulas:
Ha.. I thought it worked for raw string too, my bad. If we need to make a false "kind" for string node it would be in astroid that could be useful for other project or checker too.

DanielNoord:
Adding the "false kind" would be helpful anyway I think, as well as back porting `kind` to python 3.6. Checkers which use it can now only use >=3.8 (I believe).

However, for this checker I'm on the fence about the best way forward. 
It turns out there are many use-cases of normal strings with `{...}` and after every iteration of working on this I find 10 more. Perhaps we should limit the scope of the checker to only check for exact matches with variables names in both global and local scope. This is fairly trivial and would likely lead to only a small amount of false positives. I think one of the earliest commits actually already did this.
All other extensions of the checkers (such as functions with local/global variables, or valid expressions without any variables) that I have tried to add have created considerable amounts of false positives. But I am open to suggestions and any other good ideas about how we could include such extensions!
Perhaps postpone this checker to 2.11 to see if we get any more feedback and otherwise merge the minimal checker in to 2.11?

Pierre-Sassoulas:
> Perhaps postpone this checker to 2.11 to see if we get any more feedback and otherwise merge the minimal checker in to 2.11?

Minimal checker for 2.10, then we try to get less false negative for 2.11 sounds good. Especially if by that time astroid can handle kind.

It's true that there is a lot of false positive, because it's hard to check if a string is used later. Maybe we can limit the warning for string that we know won't leave the scope (ie function or method variables). So if the string can be used later with format or % and are hard to check, then we never make any assumption ? Maybe we can raise a warning but with a lower confidence level ? I'm not sure a lot of person are using the confidence level in their configuration.

Pierre-Sassoulas:
I misunderstood what you said, you're right let's delay everything to 2.11.

DanielNoord:
> It's true that there is a lot of false positive, because it's hard to check if a string is used later. Maybe we can limit the warning for string that we know won't leave the scope (ie function or method variables). So if the string can be used later with format or % and are hard to check, then we never make any assumption?

I am not sure if we are already doing this, but having checks work for some code and not for others without clearly showing that the user is not something I am really in favour of. When I run the linter in VS Code I expect all messages I haven't disabled to be checked.

> Maybe we can raise a warning but with a lower confidence level ? I'm not sure a lot of person are using the confidence level in their configuration.

I never did, but this might be an option.

Pierre-Sassoulas:
> I am not sure if we are already doing this, but having checks work for some code and not for others without clearly showing that the user is not something I am really in favour of. When I run the linter in VS Code I expect all messages I haven't disabled to be checked.

I think this is a conversation about would we rather have false positive vs false negative and about probabilities here. Pylint is known to have annoying false positives so we're trying to lower their number (almost 75% of issue are about false positives). But no one complain about a false negative :) So as a maintainer I'd rather have less false positive. Now we can consider that creating a string for format is rather rare and they can be disabled (Like you did in pylint there are maybe 4 occurrence for 20 ksloc...) so false positive are ok in this case.

To go into more detail I think the probability of creating a format string is higher on a non local string (it's a string that is going to be reused) and imo most of the false positive will be on such string. That's why I wanted to exclude the string that are not used locally from the check. Dou you have an opinion @cdce8p ?

cdce8p:
> Dou you have an opinion @cdce8p ?

In general, I agree with @Pierre-Sassoulas here. We should do our best to prevent as many `false-positives` as possible. If they get too frequent, they turn users away just because they are annoying. `false-negatives` in contrast are just areas we can improve in the future.

Regarding the specifics of f-strings, I'm not sure there is a good way to determine if it should be one or not. Thus this will likely lead to more false-positives than helpful warnings. @Pierre-Sassoulas mentioned assignments and I agree: Those should definitely be excluded. It's common to see code like this:
```py
TEMPLATE_STR = "Some string with a placeholder: {}"
print(TEMPLATE_STR.format("Hello World"))
```

In the end, I don't know if we should move forward with it. In case you still would like to, it might be better to create a new extension for it.

cdce8p:
A new extension could also include the `consider-using-f-string` check. See my comment here https://github.com/PyCQA/pylint/pull/4796#issuecomment-908424386

Pierre-Sassoulas:
I think `possible-forgotten-f-prefix` should be a default checker as it's asked by a lot of user : it's a very popular issue, one of the most popular in fact (and it's also a very popular question on stackoverflow that I can't find anymore). The forgotten f happen very often. But I realize now that `consider-using-f-string` is a little opinionated and could be an extension. This is the last issue that we have to put in 2.11. Maybe we can delay  `possible-forgotten-f-prefix` in 2.12 and put `consider-using-f-string`  in the `RecommendationChecker` then release 2.11 ?

DanielNoord:
While I agree `consider-using-f-string` is opinionated I believe the fact that there are real performance benefits might warrant a place in the default checkers. However, obviously this is a decision for the project maintainers so we should move it.

For this PR to move forward the following tasks remain:
- [ ] Implement a `nodes.Const.kind` that also picks up on the `r` prefix rather than only `u` (and possibly backport to Python<3.8). This should probably be done in `astroid`.
 I have created an issue: https://github.com/PyCQA/astroid/issues/1156
- [ ] Rewrite tests to only include 1) non-assignments in global scope, 2) assignments in method or function scopes and 3) only exact matches with variables in local or global scope
- [ ] Rewrite code to match new tests. See commit 4 for a 95% working example.

Pierre-Sassoulas:
The performance benefit convinced me, I'm for keeping `consider-using-f-string` the way it is. Should we suggest to use pyupgrade in the message ? This automate a lot of the boring/easy stuff when migrating to f-string and could reduce a lof of the frustration.

Thank you for opening the issue in astroid and making a checklist. It seems there is still a lot of work on this one so let's not block 2.11 because of it.

cdce8p:
> While I agree `consider-using-f-string` is opinionated I believe the fact that there are real performance benefits might warrant a place in the default checkers. However, obviously this is a decision for the project maintainers so we should move it.

Ok, this would indeed be an argument to keep it enabled by default. I would nevertheless recommend to move it to the `RecommendationChecker`. That would just be a better fit for it. (The one is loaded by default too.)

--
> Should we suggest to use pyupgrade in the message ? This automate a lot of the boring/easy stuff when migrating to f-string and could reduce a lof of the frustration.

The documentation for pyupgrade mentions some good arguments for when it should **not** be used: https://github.com/asottile/pyupgrade#f-strings

```txt
note: pyupgrade is intentionally timid and will not create an f-string
if it would make the expression longer or if the substitution parameters
are anything but simple names or dotted names (as this can decrease readability).
```

I wasn't aware that pyupgrade has already implemented `fstrings`. This is a good argument, for moving it to an optional extension. If `pyupgrade` can do it automatically, we should recommend using that instead!

DanielNoord:
I'm going to close this PR (sadly).

The logic I worked on worked (sort of) but there were just too many false positives an edge cases that I would be comfortable merging it. I think https://github.com/PyCQA/astroid/issues/1156 would need to be fixed for me to ever start working on this again.

Issue #4391, opened 2021-04-24
False negative ``unused-variable`` for catched exception when used in the inner context


Pierre-Sassoulas:
### Steps to reproduce

Given a file `a.py`:
```python
#pylint: disable=missing-docstring

def function():
    unused = 1  # [unused-variable]
    try:
        1 / 0
    except ZeroDivisionError as error:  # [unused-variable]
        try:
            1 / 0
        except ZeroDivisionError as error:
            raise Exception("") from error

```

### Current behavior

```
************* Module a
a.py:4:4: W0612: Unused variable 'unused' (unused-variable)

------------------------------------------------------------------
Your code has been rated at 8.89/10 (previous run: 8.89/10, +0.00)
```

### Expected behavior

```
************* Module a
a.py:4:4: W0612: Unused variable 'unused' (unused-variable)
a.py:8:30: W0612: Unused variable 'error' (unused-variable)
```

### pylint --version output

```
pylint 2.7.4
astroid 2.5.3
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0]
```

ksaketou:
Hello, I think that pylint considers the `error` variable as used due to the `raise ... from error`. I ran pylint  on this code and I got the correct output.
```
#pylint: disable=missing-docstring

def function():
    unused = 1  # [unused-variable]
    try:
        1 / 0
    except ZeroDivisionError as error:  # [unused-variable]
        try:
            1 / 0
        except ZeroDivisionError as error:
            raise Exception("")
```
Output:
```
unused-variable.py:4:4: W0612: Unused variable 'unused' (unused-variable)
unused-variable.py:7:4: W0612: Unused variable 'error' (unused-variable)
```

Pierre-Sassoulas:
Good point, this is certainely part of it, but I think that without the ` raise Exception("") from error` the correct output would become:
```
unused-variable.py:4:4: W0612: Unused variable 'unused' (unused-variable)
unused-variable.py:7:4: W0612: Unused variable 'error' (unused-variable)
unused-variable.py:10:4: W0612: Unused variable 'error' (unused-variable)

```

ksaketou:
Without the `raise Exception("") from error` the output stays the same. However, the output you suggest makes sense. I could give this a try if there is no one working on it

Pierre-Sassoulas:
I assigned you :) Thanks for handling this !

ksaketou:
[...]

PR #4791, opened 2021-08-02
Fix false negative for used-before-assignment (ExceptHandler)

### LINKED ISSUE #626: pylint doesn't understand the scope of except clause variables
Originally reported by: **Tom Dryer (BitBucket: [tdryer](http://bitbucket.org/tdryer), GitHub: @tdryer?)**

---

In the following code, pylint fails to identify that the `e` in the except clause is an unused variable, and that the `e` in `print(e)` is an undefined variable:

```
#!python

def main():
    try:
        raise ValueError
    except ValueError as e:
        pass

    print(e)


main()
```

The issue appears to be that pylint doesn't understand that the scope of the exception variable is limited to the except clause.

---
- Bitbucket: https://bitbucket.org/logilab/pylint/issue/626

david-yz-liu:
## Type of Changes

## Description

This deals with the same logical issue as the third point in https://github.com/PyCQA/astroid/pull/1111: handling the fact that exception variables in `ExceptHandler`s are only in scope inside the handling block. The way I fixed this was to explicitly filter out the `AssignName` nodes that were the exception variable in an `except` clause, in `NamesConsumer.get_next_to_consume`.

⭐ I also modified `NamesConsumer.mark_as_consumed` to allow for a name to be "partially consumed"⭐, so that I could piggyback on the existing check for `used-before-assignment` and still get an `unused-variable` for the except variable, as shown below:

```python
def f():
    try:
        1 / 0
        pass

    print(e)
```

```console
$ python -m pylint test.py
************* Module test
test.py:7:10: E0601: Using variable 'e' before assignment (used-before-assignment)
test.py:4:4: W0612: Unused variable 'e' (unused-variable)
```

But if you don't like that change, I could remove the changes to `mark_as_consumed` too, and either try to write a special case to get the `unused-variable` error, or just leave it without checking for that.

As a comment: the main complication I encountered is that the methods currently only look at the first `AssignName` node, e.g.

https://github.com/PyCQA/pylint/blob/9ae559d202814091896cdd3b95068a12fa1bd94b/pylint/checkers/variables.py#L1027

So I use some new filtering to avoid introducing false positives, like in 

```python

def f():
    try:
        1 / 0
    except ZeroDivisionError as e:  # This AssignName comes first but doesn't matter
        pass

    e = 10
    print(e)  # This is defined
```

That said, pylint's `VariablesChecker` doesn't do much to handle control flow, so I could also revert the filtering and instead add a special case for `ExceptHandler` in `visit_name` directly.

Closes #626


ChangeLog
```diff
@@ -68,6 +68,11 @@ Release date: TBA
 
   Closes #4681
 
+* Fix false negative for ``used-before-assignment`` when the variable is assigned
+  in an exception handler, but used outside of the handler.
+
+  Closes #626
```
Pierre-Sassoulas, line 74: I think it also closes #4391?

Pierre-Sassoulas (changes requested):
Thank you, amazing fix ! I don't know if the change to consume is the way to go but this is certainely the expected result. Regarding control flow, if I understand correctly we would have to refactor rhe variable checker in order to do it properly ? Or is that a limitation of the visitor pattern used by pylint and there's notre much we can do ?

ChangeLog
david-yz-liu: Unfortunately it doesn't---the changes don't handle shadowing:

```python
def function():
    try:
        1 / 0
    except ZeroDivisionError as error:
        try:
            1 / 0
        except ZeroDivisionError as error:
            raise Exception("") from error
```

You can also see that there are tests in `tests/functional/u/unused/unused_variable.py:108-158` that still have TODOs in them.

Both occurrences of the `as error` are part of the same consumer in "`to_consume`", and they're both removed when the `error` `astroid.Name` node in the inner block is visited. I can think of a couple of different ways of fixing this, but none that are particularly easy/elegant. I can point out that we get a similar issue with loop variable shadowing:

```python
def function():
    for i in range(10):
        for i in range(20):
            print(i)
```

Pylint currently reports a `redefined-outer-name` error for the inner `i` loop variable, but does *not* report `unused-variable` for the outer `i` loop variable. The `redefined-outer-name` check is handled in a separate `visit_for` method; I could do something similar in `visit_excepthandler` if you want.

P.S. I'll reply to your general review comment later, need to grab lunch now!

Pierre-Sassoulas: Thank you for your very clear answer !

david-yz-liu:
@Pierre-Sassoulas so in terms of the overall approach: right now each `NamesConsumer` supports *one* lookup per variable. Variable starts in `to_consume`, then after a `Name` is visited, the entry gets moved to `consumed`. Moreover, it's all or nothing; either all `AssignName`s associated with that variable get moved from `to_consume` to `consumed`, or they all stay in `to_consume`.

My changes allow for some `AssignName`s to stay in `to_consume`, but *don't* change the fact that each variable is looked up once per `NamesConsumer`, because of the check here:

https://github.com/PyCQA/pylint/blob/161a27199fce6c12411c61042212b7a669d04ee3/pylint/checkers/variables.py#L1026-L1040

This allows for the exception variable `e` to both be visited and marked as `used-before-assignment`, but leaves the `AssignName` in the consumer so that it gets marked as `unused-variable`.

If we don't want to change `to_consume` in this way, another way to detect `unused-variable` would be to define `visit_excepthandler`/`leave_excepthandler` to look for unused exception variables as a special case, similar to what `visit_for` does. This keeps the main algorithm (`visit_name`) simpler, but might increase the overall running time. (The brute force check would be something like `any(n.name == except_handler.name for n in except_handler.nodes_of_class(astroid.Name))`. Don't know if we could do much better.) Or again, could just not check for `unused-variable` in this case at all for this PR.

<hr>

Regarding control flow, I think the main limitation is the fact that the variable checkers rely on `NamesConsumer`'s `to_consume`/`consumed` states, and these are updated "all-or-nothing" per variable, with additional special-case checks here and there. To give a simpler example, pylint doesn't report any error on:

```python
def function():
    x = 10
    del x
    print(x)

function()
```

I don't know the history of the Variables checker, but it seems to me like pylint will keep running into issues around control flow as long as this basic approach is used. I wouldn't say that this is a limitation of the AST visitor pattern though. I think improvements could be made by doing any of (in order of increasing code complexity and running time, but also correctness IMO):

1. Modifying `NamesConsumer` so that variables can be processed more than once.
2. Using `astroid`'s variable `lookup` functionality, which handles control flow better.
3. Building a control flow graph and then running `VariablesChecker` checks on the graph rather than the AST itself. 

But all of these would be really significant changes so I'm not exactly advocating for them, just thought I'd mention. Personally I'm still very happy using pylint with the current functionality!

Pierre-Sassoulas:
Thank you for your insightful answer. If we want to implement control flow at some point we'll need that kind of insight and I can't attain it while tending to the constant fire of new issues and pull requests, so I appreciate that very much. Maybe I could create a "control flow" issue referencing the possibilities you gave so it become the reference when the issue of control flow comes up ? This is a problem that appear often and I have no solution but hacks, personally.

Pierre-Sassoulas (approved):
👍

david-yz-liu:
Sure, that sounds fine to me. (Honestly I feel like it's your and the other pylint maintainers' call about whether to pursue this or not!!)

PR #4793, opened 2021-08-03
Refactor ``--list-msgs`` & ``--list-msgs-enabled``

### LINKED ISSUE #4778: --list-msgs and --list-msgs-enabled output different sets of messages
### Bug description

```python
`--list-msgs` and `--list-msgs-enabled` output different sets of message ids; there are more in the latter.
```

### Configuration

```ini
Default out of the box.
```

### Command used

```shell
pylint --list-msgs | sed -rne 's/^:([^ ]+).*/\1/p' | sort -u > list-msgs
pylint --list-msgs-enabled | sed -rne 's/^  ([^ ]+).*/\1/p' | sort -u > list-msgs-enabled
diff -U0 list-msgs list-msgs-enabled
```

### Pylint output

```shell
--- list-msgs	2021-07-31 11:18:23.627278871 +0300
+++ list-msgs-enabled	2021-07-31 11:19:05.295546080 +0300
@@ -42,0 +43 @@
+boolean-datetime
@@ -75,0 +77 @@
+continue-in-finally
@@ -133,0 +136 @@
+import-star-module-level
@@ -183,0 +187 @@
+long-suffix
@@ -218,0 +223 @@
+non-ascii-bytes-literal
@@ -239,0 +245,2 @@
+old-ne-operator
+old-octal-literal
@@ -269,0 +277 @@
+return-arg-in-generator
```

### Expected behavior

No difference, same (complete) set of message ids output from both.

### Pylint version

```shell
pylint 2.9.6
astroid 2.6.5
[GCC 9.4.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Both options now show which messages can't be emitted with the current interpreter. This makes function more like their name implies. Code clearly shows in which order/format the messages are emitted.
This closes #4778

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -775,25 +775,23 @@ class PyLinter(
         self._python3_porting_mode = True
 
     def list_messages_enabled(self):
+        emittable, non_emittable = self.msgs_store.find_emittable_messages()
         enabled = []
         disabled = []
-        non_emittable = []
-        for message in self.msgs_store.messages:
-            if not message.may_be_emitted():
-                non_emittable.append(f"  {message.symbol} ({message.msgid})")
-            elif self.is_message_enabled(message.msgid):
+        for message in emittable:
+            if self.is_message_enabled(message.msgid):
                 enabled.append(f"  {message.symbol} ({message.msgid})")
             else:
                 disabled.append(f"  {message.symbol} ({message.msgid})")
         print("Enabled messages:")
-        for msg in sorted(enabled):
+        for msg in enabled:
             print(msg)
         print("\nDisabled messages:")
-        for msg in sorted(disabled):
+        for msg in disabled:
             print(msg)
         print("\nNon-emittable messages with current interpreter:")
-        for msg in sorted(non_emittable):
-            print(msg)
+        for msg in non_emittable:
+            print(f"  {msg.symbol} ({msg.msgid})")
         print("")
 
     # block level option handling #############################################
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -775,21 +775,24 @@ class PyLinter(
         self._python3_porting_mode = True
 
     def list_messages_enabled(self):
-        enabled = [
-            f"  {message.symbol} ({message.msgid})"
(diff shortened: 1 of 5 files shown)


pylint/lint/pylinter.py
```diff
@@ -775,22 +775,25 @@ def python3_porting_mode(self):
...
-        ]
-        disabled = [
-            f"  {message.symbol} ({message.msgid})"
-            for message in self.msgs_store.messages
-            if not self.is_message_enabled(message.msgid)
-        ]
+        enabled = []
+        disabled = []
+        non_emittable = []
+        for message in self.msgs_store.messages:
```
Pierre-Sassoulas, line 781: Nice refactor 👍

Pierre-Sassoulas (changes requested):
Nice job ! Could we refactor a little further and create a common function to recover emitable/non-emittable ? This way ``list_messages_enabled`` would loop on the result of this function and the other one would use them directly.

DanielNoord:
Where should I put this common function? Is there a good place for utils function for files in different directories?

Thinking of:
```python
def find_emittable_messages(
    messages: List[MessageDefinition],
) -> Tuple[List[MessageDefinition], List[ValuesView[MessageDefinition]]]:
    emittable = []
    non_emittable = []
    for message in messages:
        if message.may_be_emitted():
            emittable.append(message)
        else:
            non_emittable.append(message)
    return emittable, non_emittable
```

Pierre-Sassoulas:
``MessageDefinitionStore`` is the logical place so you don't have to provide the messages (they're already a class attribute).

Pierre-Sassoulas (approved):
Thank you !

scop:
FYI, my recent pylint bash completion PR at https://github.com/scop/bash-completion/pull/562 is what triggered to me to file #4778. Looking at the code here it seems my scraper would be still doing the right thing after these changes, cool.

The scraper is _far_ from pretty, but all things considered I don't think we have a better choice at the moment, and if we want to support old pylint versions out there, new additions of e.g. better machine parseable outputs for these in new releases wouldn't help either.

PR #2036, opened 2018-04-24
Breaking down C0111(Missing Docstring) into multiple "Convention" messages

Pierre-Sassoulas:
This is an incomplete pull request in order to see more easily what needs to be changed in order to implement #1164 after a long period of inactivity.

PR #4794, opened 2021-08-03
Add ``format-string-without-interpolation`` checker

### LINKED ISSUE #4042: format-string inconsistencies and bugs
At work we stumbled upon a bug in an old code, that used `%` formatting, but the fmt string was wrong, because it didn't contain any replacement fields. It would be nice if pylint could warn about format strings without any replacement fields.

### Current behavior

```py
def demo(param):
    print("1asfasfdsf dasfdadfsdf" % param)  # no warning (code like this line was used in our code-base)
    print("2asfasfdsf dasfdadfsdf" % ())  # no warning
    print("3asfasfdsf dasfdadfsdf" % [])  # pylint outputs too-many-format-args

    # this line raises when you run the script using cpython
    print("4asfasfdsf dasfdadfsdf" % None)  # pylint outputs too-many-format-args

demo(())
```

```
demo.py:4:10: E1305: Too many arguments for format string (too-many-format-args)
demo.py:7:10: E1305: Too many arguments for format string (too-many-format-args)
```

When using `.format` the behavior is a bit different (both in cpython as well as in pylint)

```py
def demo(param):
    print("1asfasfdsf dasfdadfsdf".format(param))  # pylint outputs too-many-format-args
    print("2asfasfdsf dasfdadfsdf".format(()))  # pylint outputs too-many-format-args
    print("3asfasfdsf dasfdadfsdf".format([]))  # pylint outputs too-many-format-args
    print("4asfasfdsf dasfdadfsdf".format(None))  # pylint outputs too-many-format-args

demo(())
```

Contrary to the first example, this one doesn't raise an exception when you run it using cpython. However, pylint considers all print  function calls as invalid, which is wrong, because cpython doesn't raise.

```py
demo_usint_dot_format.py:2:10: E1305: Too many arguments for format string (too-many-format-args)
demo_usint_dot_format.py:3:10: E1305: Too many arguments for format string (too-many-format-args)
demo_usint_dot_format.py:4:10: E1305: Too many arguments for format string (too-many-format-args)
demo_usint_dot_format.py:7:10: E1305: Too many arguments for format string (too-many-format-args)
```

### Expected behavior

I don't understand why python supports format strings without any replacement fields. Maybe because if the fmt string and the args passed to e.g. `.format` are determined dynamically.

**Anyway, I think pylint should output an error (e.g. `bad-format-string`) if the format string is a string without any replacement fields** (as in the examples above).

Things I want to mention but I don't care about them:
 
There is an inconsistency between lists and tuples if used as format string args:

```py
    print("2asfasfdsf dasfdadfsdf" % ())  # no warning
    print("3asfasfdsf dasfdadfsdf" % [])  # pylint outputs too-many-format-args
```

There are differences between `%` formating and `.format` formatting in pylint.

### pylint --version output

```
pylint 2.6.0
astroid 2.4.2
Python 3.8.6 | packaged by conda-forge | (default, Oct  7 2020, 19:08:05) 
[GCC 7.5.0]
```

DanielNoord:

## Type of Changes

## Description

This adds a checker that checks strings with '%' or format() applied to them.
If no variables to be replaced are found the warning is emitted.
Closes #4042

### DIFF
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -209,7 +209,6 @@ MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
         "format-string-without-interpolation",
         "Used when we detect a string that does not have any interpolation variables, "
         "in which case it can be either a normal string without formatting or a bug in the code.",
-        {"old_names": [("E1305", "too-many-format-args")]},
     ),
 }
 
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -209,6 +209,7 @@ MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
         "format-string-without-interpolation",
         "Used when we detect a string that does not have any interpolation variables, "
         "in which case it can be either a normal string without formatting or a bug in the code.",
+        {"old_names": [("E1305", "old-empty-docstring")]},
     ),
 }
 
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -204,6 +204,12 @@ MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
         "Used when we detect an f-string that does not use any interpolation variables, "
         "in which case it can be either a normal string or a bug in the code.",
     ),
+    "W1310": (
+        "Using formatting for a string that does not have any interpolated variables",
+        "format-string-without-interpolation",
+        "Used when we detect a string that does not have any interpolation variables, "
+        "in which case it can be either a normal string without formatting or a bug in the code.",
+    ),
 }
 
 OTHER_NODES = (
@@ -273,6 +279,7 @@ class StringFormatChecker(BaseChecker):
         "too-many-format-args",
         "too-few-format-args",
         "bad-string-format-type",
+        "format-string-without-interpolation",
     )
     def visit_binop(self, node):
         if node.op != "%":
@@ -300,6 +307,9 @@ class StringFormatChecker(BaseChecker):
             return
         except utils.IncompleteFormatString:
             self.add_message("truncated-format-string", node=node)
+            return
+        if not required_keys and not required_num_args:
+            self.add_message("format-string-without-interpolation", node=node)
             return
         if required_keys and required_num_args:
             # The format string uses both named and unnamed format
@@ -518,6 +528,9 @@ class StringFormatChecker(BaseChecker):
         if check_args:
             # num_args can be 0 if manual_pos is not.
             num_args = num_args or manual_pos
+            if not num_args:
+                self.add_message("format-string-without-interpolation", node=node)
+                return
             if len(positional_arguments) > num_args:
                 self.add_message("too-many-format-args", node=node)
             elif len(positional_arguments) < num_args:
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -209,7 +209,7 @@ MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
         "format-string-without-interpolation",
         "Used when we detect a string that does not have any interpolation variables, "
         "in which case it can be either a normal string without formatting or a bug in the code.",
-        {"old_names": [("E1305", "old-empty-docstring")]},
+        {"old_names": [("E1305", "too-many-format-args")]},
     ),
 }
 
--- a/tests/functional/s/string/string_formatting.py
+++ b/tests/functional/s/string/string_formatting.py
@@ -65,7 +65,7 @@ def print_good():
 
 def pprint_bad():
     """Test string format """
-    "{{}}".format(1) # [too-many-format-args]
+    "{{}}".format(1) # [format-string-without-interpolation]
     "{} {".format() # [bad-format-string]
     "{} }".format() # [bad-format-string]
     "{0} {}".format(1, 2) # [format-combined-specification]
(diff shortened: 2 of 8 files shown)


DanielNoord:
I'm not sure how to handle the failing check in `pylint/config/__init__.py` on `line 129`. 
Seems to me that the checker has identified a problem in the current code, as `globals()` is not interpolated into the string. Do we want to do this? Or should we add the checker to the ignore list of this file?

pylint/checkers/strings.py
```diff
@@ -204,6 +204,12 @@
         "Used when we detect an f-string that does not use any interpolation variables, "
         "in which case it can be either a normal string or a bug in the code.",
     ),
+    "W1310": (
+        "Using formatting for a string that does not have any interpolated variables",
+        "format-string-without-interpolation",
+        "Used when we detect a string that does not have any interpolation variables, "
+        "in which case it can be either a normal string without formatting or a bug in the code.",
```
Pierre-Sassoulas, line 211: We could use old_names to prevent legacy disable of ``too-many-format-args`` in user code to become useless-supression and for a new warning to appear instead. (I'm not sure if it's possible while keeping the old message in some other place, disregard this comment if it's causing problems.)

Pierre-Sassoulas (changes requested):
Look good ! I only have a small comment.

pylint/checkers/strings.py
DanielNoord: But the messages are not really the same, right? Looking at other cases of `old_names` the new message often seems to exactly do the same, just separated in individual messages. Furthermore, looking at #4042 in some cases `too-many-format-args` is not currently raised. I think it might be better to really consider this a new checker, perhaps postponing it till 3.0?

Pierre-Sassoulas: Some old names are not exactly the same for example we split up ``missing-docstrings`` in ``missing-module-docstring``, ``missing-class-docstrings``, ``missing-function-docstring``. The idea is to be able to change a name without forcing the user to change its disable or configuration, not having an exact match. So I think it's ok, to use this for a partial rename to have more precise message.

DanielNoord: Trying to pass test as we speak.
Have you considered the failing test on `pylint/config/__init__.py` and what to do with it?

DanielNoord: Ah, `too-many-format-args` already has an old-name.
Looking at 
```python
"W9012": (
            "Missing return type documentation",
            "missing-return-type-doc",
            "Please document the type returned by this method.",
            # we can't use the same old_name for two different warnings
            # {'old_names': [('W9007', 'missing-returns-doc')]},
        ),
``` 
this is not possible?

Pierre-Sassoulas:
Regarding config init : Very good news we discovered a problem in pylint with this new checker ? 😄 I think we can remove globals() from this string (to keep the same behavior at least). I think this would help during debug but clearly no one used this in a long time or we'd have a bug report at this point.

pylint/checkers/strings.py
Pierre-Sassoulas: Since then we can but it was a lot of work. Actually it's my first merge request to split up missing-docstring into multiple warning. Good time. Took 2 years and 10 MR. Search for C0111, C01112, C0114 or C0115 😄. By the way we can  probably uncomment the line and remove the comment, but maybe it's a bit risqué.

Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/2036

tests/functional/s/string/string_formatting.py
```diff
@@ -65,7 +65,7 @@ def print_good():
 
 def pprint_bad():
     """Test string format """
-    "{{}}".format(1) # [too-many-format-args]
+    "{{}}".format(1) # [format-string-without-interpolation]
```
Pierre-Sassoulas, line 68:
```suggestion
    "{{}}".format(1) # [too-many-format-args]

```
Doing it this way is a functional test for the old_names we added :)

DanielNoord:
This is probably me not understanding how old_names work, but now all tests are failing as all `too-many-format-args` seem to have become `format-string-without-interpolation`. Probably my mistake, but not sure what happened here..

Pierre-Sassoulas:
Ho no sorry my bad. It's not possible to rename only a part of it. We'd have to rename both. It means we'd need to rename ``[too-many-format-args`` too. I don't know about that. this seems hacky. Maybe if we find a better name we can kill two bird with one stone ?

DanielNoord:
Hmm `too-many-format-args` seems pretty good as is. Is there no way to keep it the same?
Perhaps `additional-format-args`? Although I think `too-many-format-args` is still better, especially considering `too-few-format-args` is also a thing..

Pierre-Sassoulas:
Another way would be to refactor the MessageIdStore so it's possible to partially rename a message. I think the consequences is that you'd need to instantly rename all the messages that need to be renamed when partially renaming a message or you'd get the old name (kinda defeating the purpose of it in the first place). Beside if you have a ``too-many-format-args`` this is most likely an error so you won't disable it very often. I think reverting the old_names I asked is the sane decision here. Sorry for wasting your time with this comment.

DanielNoord:
No problem, I can imagine why you wanted to add `old_message` initially!

Pierre-Sassoulas (approved):
Thank you for reverting ! 👍

DanielNoord:
Not that I care too much about it, but it seems like the Copyright lines at the top of a document are not updated automatically. Should I create a PR to do so?

Pierre-Sassoulas:
It's done automatically but during release (check ``doc/release.md``). It's a pretty heavy calculation using ``copyrite``,  also it changes the order of line randomly which is not ideal but not a real problem either. This lib could be optimized or changed but it never bothered me enough to actually do anything. I  you have a better solution on the shelf I'm open to suggestion :)

DanielNoord:
Ah okay! Don't really have a better solution, but will keep my eyes open!

PR #4796, opened 2021-08-03
Add ``Consider-using-f-string`` checker

### LINKED ISSUE #3592: Consider nudging users towards using f-strings
Python 3.5 is end-of-life in September 2020.  Since Python 3.6 introduced f-strings, there is no good reason to interpolate _literal strings_ with the string interpolation operator
```python
'Python %f' % version
```
or the `format` method
```python
'Python {}'.format(version)
```
—except when the arguments involve unpacking:
```python
'Python {} or {}`.format(*versions)
```

Can we consider a pylint warning to nudge users towards using f-strings when possible?  It would be extremely helpful to write the f-string for them.  For example:
```
a.py:1:0: W999: Not using f-string: f"Python {version}" (not-using-fstring)
```
What do you think?

DanielNoord:

## Type of Changes

## Description

Add ``consider-using-f-string checker``
This adds a checker for normal strings which are formatted
with `.format()` or '%'.
The message is a convention to nudge users towards using f-strings.
This closes #3592


DanielNoord:
Many changes in the second commit of this PR. However, I think they are warranted as f-strings are considerably faster, see [here](https://www.scivision.dev/python-f-string-speed/) for example.
Let me know if anything needs to be changed!

tests/functional/l/logging_not_lazy.py
```diff
@@ -14,7 +14,6 @@
 renamed_logging.warn('%s' + ' the rest of a single string') # [logging-not-lazy]
 renamed_logging.log(renamed_logging.INFO, var_name + var) # [logging-not-lazy]
 
-var_name = 'Var:'
```
Pierre-Sassoulas, line 17: Why remove this line ?

tests/functional/r/raise/raising_format_tuple.py
```diff
@@ -12,23 +12,23 @@ def bad_percent(arg):
 
 def good_percent(arg):
     '''Instead of passing multiple arguments, format the message'''
-    raise KeyError('Bad key: %r' % arg)
```
Pierre-Sassoulas, line 15: This file should probably not be touched as the formatting of the string seems to be a test in itself ?

Pierre-Sassoulas (changes requested):
Huge cleanup work alongside a very useful new checker, thanks a lot ! Did you check the change in performance by using the benchmark result locally ? If f-string are faster if could be another good news 😄 I have minor comments, let me know what you think.

tests/functional/l/logging_not_lazy.py
DanielNoord: I didn't think it added anything after line 8. Seems a duplicate

tests/functional/r/raise/raising_format_tuple.py
DanielNoord: True. Going to fix now!

DanielNoord:
Should be good to go now for `2.11`.

I did not do the benchmarks, you might want to do them but I think there should be a performance increase. At least for some of the tests.

Pierre-Sassoulas (approved):
Thanks a lot, great checker and great clean up too !

Pierre-Sassoulas:
Regarding the performance increase, I also think that it should increase because of f-string but we'd switch to f-string even with a sligth decrease in performance because of the readability and modernity :)

cdce8p:
I don't think this should have been part of the `StringChecker`. It's only a recommendation, so at the very least this should go into the `RecommendationChecker`.

IMO this should even move to a new extension. `str.format` is still perfectly valid and in some cases even the more readable option!

PR #4777, opened 2021-07-30
Add regression tests for issue #3711

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

This is a regression tests for https://github.com/PyCQA/astroid/pull/1111

Closes #3711

PR #4797, opened 2021-08-04
Add ``disable-next`` option

### LINKED ISSUE #1682: Add an option to disable a message on the next line so a line does not become too long because of a disable
Hi. I'm trying to ignore a pylint error for a single line and I don't know how it should be done.

Example use case:

```python
        if self._wrapper_class is not None:
            data = self._wrapper_class(data)
            [...]
```

The `data = ...` line triggers a `not-callable` error I would like to mask.

I'm not sure how to do that.

This triggers a `line-too-long` error, unless I also disable `line-too-long` in the line, which makes it even longer (in my real-life use case, data is a longer word and the line is more than 80 characters):

```python
        if self._wrapper_class is not None:
            data = self._wrapper_class(data)  # pylint: disable=not-callable
            [...]
```

This disables the warning for the whole `if` context:

```python
        if self._wrapper_class is not None:
            # pylint: disable=not-callable
            data = self._wrapper_class(data)
            [...]
```

This if functionally correct but a bit too verbose:

```python
        if self._wrapper_class is not None:
            # pylint: disable=not-callable
            data = self._wrapper_class(data)
            # pylint: enable=not-callable
            [...]
```

This was discussed a while ago on [a mailing-list thread](http://python-projects.logilab.narkive.com/P3Cl687J/disabling-warning-for-one-long-line-only-potential-ppylint-bug).

I guess this is bound to happen as long as the only way to disable a warning for a single line is to add a trailing pragma, especially since pylint supports disable by (long) name rather than (short) code.

Wouldn't it be handy to have some sort of ignore-for-next-statement pragma?

```python
        if self._wrapper_class is not None:
            # pylint: disable-next=not-callable
            data = self._wrapper_class(data)
            [...]
```

DanielNoord:

## Type of Changes

## Description

Adding `# pylint: disable-next=msgid` to your file will disable the message for the next line.
Similar to how many linter and checkers do this.
This closes #1682


DanielNoord:
#1682 also discusses another issue about disabling `line-too-long` and whether disable messages should count towards line length. This is not covered by this PR but I think warrants a different issue, as the feature requested in the OP of the issue is implemented here.

DanielNoord:
I think the failing tests are due to changes to the Python 3.10 runner. Probably related to the release of RC1 on Monday. The tests on the `main` branch were passing yesterday, but when I merged `main` into my own repo this morning they failed. See [here](https://github.com/DanielNoord/pylint/tree/main).
Probably need to add an `.rc` file to the failing tests. The CodeQL failures are new to me. Should we do something with them?

Pierre-Sassoulas (approved):
> I think the failing tests are due to changes to the Python 3.10 runner. 

Definitely, we can also see it in #4777 

Regarding code QL, I dismissed them, there's no secret here, it's just a variable name in a functional test.

I think this solve the issue with line too long, we have a way to disable only one line when the line is long. thanks a lot for the implementation :+1:

Pierre-Sassoulas:
> I doesn't fully solve this comment though. 

I understand this comment as `not counting the # disable=line-too-long in the line length defeat the purpose of line-too-long because the ling is even longer than before`. And I agree with this, we should not do it. I think this disable-next is the perfect solution :)

DanielNoord:
I don't fully agree. I think the comment that we shouldn't cut off the `disable-param` when calculating the line-length is valid.
If a user has set their screen width to fit a certain number of columns they should be able to see all code, including any comments and `disable-param`. This PR gives users the ability to move these `disable-param`s to the line above, but I think we should indeed raise an (optional) message whenever `line + comments > max line width`. That does not fit within this PR, but that request is included in #1682.

Pierre-Sassoulas:
Ok it makes sense, what about creating another issue for a checker for code + comment > max char ?

Pierre-Sassoulas (changes requested):
We need a little example in the documentation (FAQ and message control chapter). This is a pretty important feature, and I think it need to be visible :)

DanielNoord:
Not the best documentation writer. Please add whatever you deem necessary!

Pierre-Sassoulas (approved):
No no, it was perfect, thank you 😄

lafrech:
Original poster of #1682, here. Thank you guys for this feature.

Issue #4439, opened 2021-05-04
AttributeError: 'Subscript' object has no attribute 'name'


MrTrick:
Seems distinct from https://github.com/PyCQA/pylint/issues/2250
Maybe a regression, or another edge case?

### Steps to reproduce

Given a file `a.py`:
```python
from attr import attrs, attrib
from typing import Optional

@attrs()
class User:
    name: str = attrib()
    age: int = attrib()
    occupation = Optional[str] = attrib(default=None)
```
Where a typo was made in the last line - first `=` should be `:`.

### Current behavior

Pylint crashes if executed on the file, or on any file that imports it.

```
# pylint a.py
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/pylint/lint/pylinter.py", line 1041, in get_ast
    return MANAGER.ast_from_file(filepath, modname, source=True)
  File "/usr/local/lib/python3.7/site-packages/astroid/manager.py", line 98, in ast_from_file
    return AstroidBuilder(self).file_build(filepath, modname)
  File "/usr/local/lib/python3.7/site-packages/astroid/builder.py", line 142, in file_build
    return self._post_build(module, encoding)
  File "/usr/local/lib/python3.7/site-packages/astroid/builder.py", line 166, in _post_build
    module = self._manager.visit_transforms(module)
  File "/usr/local/lib/python3.7/site-packages/astroid/manager.py", line 75, in visit_transforms
    return self._transform.visit(node)
  File "/usr/local/lib/python3.7/site-packages/astroid/transforms.py", line 90, in visit
    module.body = [self._visit(child) for child in module.body]
  File "/usr/local/lib/python3.7/site-packages/astroid/transforms.py", line 90, in <listcomp>
    module.body = [self._visit(child) for child in module.body]
  File "/usr/local/lib/python3.7/site-packages/astroid/transforms.py", line 58, in _visit
    return self._transform(node)
  File "/usr/local/lib/python3.7/site-packages/astroid/transforms.py", line 41, in _transform
    ret = transform_func(node)
  File "/usr/local/lib/python3.7/site-packages/astroid/brain/brain_attrs.py", line 58, in attr_attributes_transform
    node.locals[target.name] = [rhs_node]
AttributeError: 'Subscript' object has no attribute 'name'
************* Module a
a.py:1:0: F0002: <class 'AttributeError'>: 'Subscript' object has no attribute 'name' (astroid-error)
```

[...]

PR #4801, opened 2021-08-04
Hot-fix master following release of python 3.10-rc's new version

Pierre-Sassoulas:
## Type of Changes

## Description

This is a hot fix, the real fix escape me right now.

### DIFF
--- a/tests/functional/r/regression/regression_4439.rc
+++ b/tests/functional/r/regression/regression_4439.rc
@@ -1,0 +1,2 @@
+[testoptions]
+max_pyver=3.10


DanielNoord:
I wanted to fix this, but not quite sure what the original test is trying to test for.
In 3.10 you can assign types to create your own types, which is why the test is no longer emitting the warning.

Pierre-Sassoulas:
Yes the part that look like a type assignment in 3.10 was invalid before 3.10 and crashed in astroid <=2.6.4, see #4439

DanielNoord:
So no need to change the tests to be able to also run on 3.10?

Pierre-Sassoulas:
Yes I think the change is ok as it is, if there was multiple use case in the functional test we'd have to separate them but there is just this one.

PR #4811, opened 2021-08-07
``pyreverse``: Move creation of label from ``DiagramWriter`` to ``Printer`` 

DudeNr33:
## Description

Currently the logic to create diagrams is split between the ``Printer`` classes and the ``DiagramWriter`` classes.
In my opinion the responsibilities should be:
- ``DiagramWriter`` defines WHAT elements should be included in the diagram
- ``Printer`` handles HOW this needs to be translated into a format the desired tool (``dot``/Graphviz or ``vcg``, later on also PlantUML) understands

With the current implementation the ``DiagramWriter`` subclasses also hold a lot of logic how to construct the ``label`` of individual nodes (i.e. to include attributes and methods of the classes). However, how this label needs to be constructed is an implementation detail of the backend and thus should reside in the ``Printer`` classes.

This MR moves the logic inside the ``Printer`` classes.
Please note that this eliminates the need for specialized ``DiagramWriter`` subclasses - they now only differ in the ``set_printer`` method, and the ``Printer`` class could now be passed directly in the constructor.
To ease the review I did not include this change just yet, so it is easier to see what parts of the current logic were moved.
We can open a new MR for this, or after these first changes are reviewed I just make another push on this MR and request a new review.

## Type of Changes

### DIFF
--- a/pylint/pyreverse/dot_printer.py
+++ b/pylint/pyreverse/dot_printer.py
@@ -41,6 +41,7 @@ class DotPrinter(Printer):
         layout: Optional[Layout] = None,
         use_automatic_namespace: Optional[bool] = None,
     ):
+        layout = layout or Layout.BOTTOM_TO_TOP
         self.charset = "utf-8"
         self.node_style = "solid"
         super().__init__(title, layout, use_automatic_namespace)
--- a/pylint/pyreverse/dot_printer.py
+++ b/pylint/pyreverse/dot_printer.py
@@ -11,10 +11,12 @@ import subprocess
 import sys
 import tempfile
 from pathlib import Path
-from typing import Dict, FrozenSet, Optional
+from typing import Dict, FrozenSet, List, Optional
+
+import astroid
 
 from pylint.pyreverse.printer import EdgeType, Layout, NodeProperties, NodeType, Printer
-from pylint.pyreverse.utils import check_graphviz_availability
+from pylint.pyreverse.utils import check_graphviz_availability, get_annotation_label
 
 ALLOWED_CHARSETS: FrozenSet[str] = frozenset(("utf-8", "iso-8859-1", "latin1"))
 SHAPES: Dict[NodeType, str] = {
@@ -65,7 +67,7 @@ class DotPrinter(Printer):
             properties = NodeProperties(label=name)
         shape = SHAPES[type_]
         color = properties.color if properties.color is not None else "black"
-        label = properties.label
+        label = self._build_label_for_node(properties)
         if label:
             if type_ is NodeType.INTERFACE:
                 label = "<<interface>>\\n" + label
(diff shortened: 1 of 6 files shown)


pylint/pyreverse/writer.py
```diff
@@ -111,10 +107,6 @@ def set_printer(self, file_name: str, basename: str) -> None:
         """set printer"""
         raise NotImplementedError
 
-    def get_title(self, obj: DiagramEntity) -> str:
-        """get project title"""
-        raise NotImplementedError
-
```
DudeNr33, line 117: This method is no longer necessary.

pylint/pyreverse/vcg_printer.py
```diff
@@ -223,15 +223,38 @@ def emit_node(
...
             NODE_ATTRS,
-            label=label,
+            label=self._build_label_for_node(properties),
             shape=SHAPES[type_],
         )
         self.emit("}")
 
+    @staticmethod
+    def _build_label_for_node(properties: NodeProperties) -> str:
+        fontcolor = "\f09" if properties.fontcolor == "red" else ""
```
DudeNr33, line 238: I am not particularly happy with hard-coding the "red" here. I am also not sure if the ``\f09`` really changes the color: the only program I could find that can display ``vcg`` graphs does not output the label with a colored text. The documentation of ``vcg`` indicates that it would theoretically also support a ``fontcolor`` attribute.

To be honest I feel like ``vcg`` is not really used any more. As I have a hard time verifying the result due to the lack of a tool which can properly display the graphs I am reluctant to change the implementation of the vcg related code too much.

Pierre-Sassoulas: Shouldn't the signature be the same with VcgPrinter and DotPrinter (to respect the Liskov substitution principle) ?

pylint/pyreverse/writer.py
Pierre-Sassoulas: 👍

Pierre-Sassoulas (commented):
Looks nice thank you ! I only have a comment regarding general design.

pylint/pyreverse/vcg_printer.py
DudeNr33: Good question - but I would argue no, not in this case:
This method is a "private" method for ``VCGPrinter`` and ``DotPrinter``, but it is intentionally not a part of the ``Printer`` base class. 
For Dot and VCG I need to put the attributes and methods of a class inside the ``label`` attribute for the node, while for the upcoming ``PlantUmlPrinter`` this is not necessary as the syntax is different.
So ``PlantUmlPrinter`` won't have such a ``_build_label_for_node`` method. 
``VCGPrinter`` and ``DotPrinter`` just somewhat "coincidentally" share a very similar method.

Pierre-Sassoulas: Make sense 👍

Pierre-Sassoulas (approved):
Do you think it's ready to merge ? :)

DudeNr33:
I would suggest that I make another push where I remove the ``VCGWriter`` and ``DotWriter`` instances, unless you prefer a separate MR for this? 
The only reason I did not include it directly is that it might have made it harder to follow along in the diffs what logic was moved between the classes. But now that you've seen and reviewed the diffs I could push the last changes and we could do it all in one MR.

Pierre-Sassoulas:
Let's do that then :)

pylint/pyreverse/dot_printer.py
```diff
@@ -41,6 +41,7 @@ def __init__(
         layout: Optional[Layout] = None,
         use_automatic_namespace: Optional[bool] = None,
     ):
+        layout = layout or Layout.BOTTOM_TO_TOP
```
DudeNr33, line 44: Since ``DiagramWriter`` no longer passes the layout explicitly I have to set the default here in ``DotPrinter``. 
Otherwise the ``pyreverse`` output would change for either ``dot`` or ``vcg``, as they both use different layouts.

DudeNr33:
Ok, now this MR is ready to be merged from my side. :)

Pierre-Sassoulas (approved):
👍

PR #4815, opened 2021-08-08
``pyreverse``: refactor tests

DudeNr33:
This should be the last "preparation" MR before the new ``pyreverse`` features (PlantUML output and colored output) can be merged.

In this MR one of the initial review remarks of the original MR is addressed: the existing tests for ``pyreverse`` imported from one another, and some proper fixtures can help clean up the code.
Additionally, I moved the tests to a separate subdirectory so the hierarchy under ``tests`` matches the one under ``pylint``, and renamed the test modules to better conform to the naming conventions of the remaining test modules.

I also included some tests that I forgot to merge when I refactored the ``Printer`` class.

## Type of Changes


.pre-commit-config.yaml
```diff
@@ -85,7 +85,7 @@ repos:
         args: []
         require_serial: true
         additional_dependencies: ["types-pkg_resources==0.1.3", "types-toml==0.1.3"]
-        exclude: tests/functional/|tests/input|tests/extensions/data|tests/regrtest_data/|tests/data/|doc/|bin/
+        exclude: tests/functional/|tests/input|tests(/.*)*/data|tests/regrtest_data/|tests/data/|tests(/.*)+/conftest.py|doc/|bin/
```
DudeNr33, line 88: ``mypy`` is not happy with having two modules with the same name inside a directory tree without ``__init__.py`` files.
Also, all ``data`` folders should probably be ignored, the original exclude pattern listed ``tests/extensions/data`` and ``tests/data`` separately.

pylint/testutils/pyreverse.py
```diff
@@ -0,0 +1,16 @@
...
+from pylint.pyreverse.inspector import Project, project_from_files
+
+
+def get_project(module: str, name: Optional[str] = "No Name") -> Project:
+    """return an astroid project representation"""
+
+    def _astroid_wrapper(func, modname):
+        return func(modname)
+
+    return project_from_files([module], _astroid_wrapper, project_name=name)
```
DudeNr33, line 16: This method was used in several test modules, but is not suitable to be replaced by a fixture. That's why I put it under ``pylint.testutils``.

tests/pyreverse/test_utils.py
```diff
@@ -0,0 +1,119 @@
+# Copyright (c) 2021 Mark Byrne <mbyrnepr2@gmail.com>
+# Copyright (c) 2021 Andreas Finkler <andi.finkler@gmail.com>
+
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/master/LICENSE
+
+"""Tests for pylint.pyreverse.utils"""
+
```
DudeNr33, line 8: Those tests were initially inside ``test_writer.py``, but I think they belong in an own module.

Pierre-Sassoulas (changes requested):
Thank you for making digestible little part for me 😄  I made minor remarks but this looks mergeable already.

pylint/testutils/pyreverse.py
```diff
@@ -0,0 +1,16 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/master/LICENSE
```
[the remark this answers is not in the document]
DudeNr33: Ah, missed that one too. Changed it.

DudeNr33:
Thanks for the review. Good idea to return a callable from a fixture, haven't thought of that before. :)

Pierre-Sassoulas (approved):
👍

PR #4816, opened 2021-08-08
Add regression tests for astroid 2.7.0 fixes

Pierre-Sassoulas:
## Type of Changes

## Description

Tests for https://github.com/PyCQA/astroid/pull/1121 and https://github.com/PyCQA/astroid/pull/1126


tests/functional/t/too/too_many_instance_attributes.py
```diff
@@ -1,4 +1,6 @@
 # pylint: disable=missing-docstring, too-few-public-methods, useless-object-inheritance
+from dataclasses import dataclass, InitVar
```
david-yz-liu, line 2: @Pierre-Sassoulas this was an oversight on my part---`dataclasses` was introduced in Python 3.7, which is why the tests are failing on 3.6. I think the right fix is to move this test into a separate file do do the `min_pyver` configuration?

Pierre-Sassoulas: Yes there are other example in the functional test. (same name with _py37 would work). You made the code we'll need in 4 months :) We need to support python 3.6 until its end of life, but it's close now.

Issue #2121, opened 2018-05-19
undefined-loop-variable in presence of cell-var-from-loop


sk-:
Similar to bug #1921 and related, I found that an extra `undefined-loop-variable` error is shown  in presence of `cell-var-from-loop` when the same variable is used later in another loop.

### Steps to reproduce
The following code:
```py
def call(f, a):
    return f(a)
for x in [1]:
    call(lambda a: a+x, 3)
for x in [2]:
    pass
```

outputs

```
W:  4,21: Cell variable x defined in loop (cell-var-from-loop)
W:  4,21: Using possibly undefined loop variable 'x' (undefined-loop-variable)
```

However, when changing the name of the second `x` or commenting out those two lines, then the output is just 

```
W:  4,21: Cell variable x defined in loop (cell-var-from-loop)
``` 

### pylint --version output
```
No config file found, using default configuration
pylint 1.8.4,
astroid 1.6.3
Python 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 03:03:55)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
```

PCManticore:
Thanks for the bug report!

SamyCookie:
Does https://github.com/PyCQA/pylint/pull/2276 fixed this issue ?

Pierre-Sassoulas:
Closing as it's fixed on main, thank you for the help @SamyCookie  and @david-yz-liu

PR #4827, opened 2021-08-10
Fix bugs in W0640 cell-var-from-loop checker

### LINKED ISSUE #2846: cell-var-from-loop misses case with nested loop
### Steps to reproduce
1. Create a file  `a.py` containing:
```py
TEST_A = [
    (lambda: n)
    for n in range(3)
]

TEST_B = [
    (lambda: [n for _ in range(3)])
    for n in range(3)
]

print(TEST_A[1]())  # User probably intended for this to be 1, actually 2

print(TEST_B[1]())  # User probably intended for this to be [1, 1, 1], actually [2, 2, 2]
```
2. Run `pylint a.py`

### Current behavior

There is a lint failure: `a.py:2:13: W0640: Cell variable n defined in loop (cell-var-from-loop)`
Tere is no such failure for `TEST_B`.

### Expected behavior

Lint failures for both `TEST_A` and `TEST_B`.

### pylint --version output

```
pylint 2.3.1
astroid 2.2.5
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 22:20:52) [MSC v.1916 32 bit (Intel)]
```

### LINKED ISSUE #3107: False positive for cell-var-from-loop
### Steps to reproduce
```
for i in [[2], [3]]:
    next(filter(lambda j, ix=i[0]: j == ix, [1, 3]))
```

### Current behavior
W0640: Cell variable i defined in loop (cell-var-from-loop)

### Expected behavior
I would expect to avoid the warning because the "i" is passed as default argument to the lambda. I get the false positive also if I loop along objects instead of list.
But I don't get it for the code below (looping along ints)
```
for i in [2, 3]:
    next(filter(lambda j, ix=i: j == ix, [1, 3]))
```

### pylint --version output

pylint 2.3.1
astroid 2.2.5
Python 3.7.4 (v3.7.4:e09359112e, Jul  8 2019, 14:54:52) 

thanks

david-yz-liu:
## Type of Changes

## Description

I reviewed some of the issues related to W0640, and since they were all related to the same method (`VariablesChecker._check_late_binding_closure`) I tackled them in one go, reviewing the existing code and adding some additional tests too.

1. https://github.com/PyCQA/pylint/issues/3107. The current code checks for names which are used as default arguments as follows:

    https://github.com/PyCQA/pylint/blob/b8021d33a4f34b410d73fa389621e2c4ae5e22c0/pylint/checkers/variables.py#L1820-L1821

     This only works for names used as simple default arguments, e.g. `lambda x=i : x`, but not ones with a more complex node structure like `lambda x=i[0] : x`. The obvious change is to use `utils.is_default_argument`, but I realized that there are cases where a default argument *is* a cell var defined from a loop, it just requires two levels of function nesting, as in the following (new) test case:

    ```python
    def bad_case10():
        """Detect when a loop variable is the default argument for a nested function"""
        lst = []
        for i in range(10):
            def func():
                def func2(arg=i):  # [cell-var-from-loop]
                    return arg

                return func2

            lst.append(func)
        return lst
    ```

    So I modified the algorithm a little to "jump up" a scope level when encountering a name in a default argument, rather than just returning right away.

2. https://github.com/PyCQA/pylint/issues/2846 The method currently uses the following to decide to skip the cell-var-from-loop check:

    https://github.com/PyCQA/pylint/blob/b8021d33a4f34b410d73fa389621e2c4ae5e22c0/pylint/checkers/variables.py#L1817-L1819

    This results in a false negative when the cell variable appears inside a comprehension (itself inside a function body), e.g.

    ```python
    TEST_B = [
        (lambda: [n for _ in range(3)])
        for n in range(3)
    ]
    ```

    My fix here was simply to use `node.frame()` rather than `node.scope()`, since the former skips comprehension scopes. A bit further below, I also added an explicit `node.name in node_scope.locals` to skip the remaining work if the node is known to not be a cell variable at all.

3.  I modified the algorithm so that instead of taking an `assignment_node` from the caller (which was the first `AssignName` in the "current consumer" in `VariablesChecker.visit_name`), it instead uses astroid's `lookup` method to find assignments associated with the name, taking into account basic control flow. This fixes a false negative like the following (also added as new test case):

    ```python
    def bad_case9():
        """Detect when loop variable shadows an outer assignment."""
        lst = []
        i = 100
        for i in range(10):
            lst.append(lambda: i)  # [cell-var-from-loop]
        return lst
    ```

4. Previously, this method was called in the following location in `visit_name`:

    https://github.com/PyCQA/pylint/blob/b8021d33a4f34b410d73fa389621e2c4ae5e22c0/pylint/checkers/variables.py#L1054-L1057

    This meant that if both `undefined-variable` and `used-before-assignment` were disabled, there would be many false negatives for `cell-var-from-loop` as well. I'm pretty sure this was unintentional. For example:

    ```python
    $ python -m pylint --disable=undefined-variable,used-before-assignment tests/functional/c/cellvar_escaping_loop.py
    ************* Module functional.c.cellvar_escaping_loop
    tests\functional\c\cellvar_escaping_loop.py:70:27: W0640: Cell variable i defined in loop (cell-var-from-loop)

    Your code has been rated at 9.85/10 (previous run: 9.91/10, -0.06)
    ```

    I moved the check out of the if statement. (This is in a separate commit.)

I also did a little bit of code cleanup to use `checker.utils` a bit more, hope that's okay. But if not happy to revert some of those changes.

Closes #2846
Closes #3107

*Comment*: Also, I think #2121 is fixed on main?


pylint/checkers/utils.py
```diff
@@ -455,6 +455,11 @@ def is_ancestor_name(
     return False
 
 
+def is_being_called(node: astroid.node_classes.NodeNG) -> bool:
```
Pierre-Sassoulas, line 458: I like this style of doing things better than having super big conditional without explanation. Even though I know this won't be re-used it's nice to have a function if only for clarity.

pylint/checkers/variables.py
```diff
@@ -1804,27 +1804,41 @@ def _check_unused_arguments(self, name, node, stmt, argnames):
 
         self.add_message("unused-argument", args=name, node=stmt, confidence=confidence)
 
-    def _check_late_binding_closure(self, node, assignment_node):
+    def _check_late_binding_closure(self, node: astroid.Name) -> None:
+        """Check whether node is a cell var that is assigned within a containing loop.
+
+        Special cases where we don't care about the error:
+        1. When the node's function is immediately called, e.g. (lambda: i)()
+        2. When the node's function is returned from within the loop, e.g. return lambda: i
```
Pierre-Sassoulas, line 1812: 👍

tests/functional/c/cellvar_escaping_loop.py
```diff
@@ -123,6 +154,63 @@ def bad_case6():
     return lst
 
 
+def bad_case7():
+    """Multiple variables unpacked in comprehension."""
+    return [
+        lambda: (
+            x  # [cell-var-from-loop]
+            + y)  # [cell-var-from-loop]
```
Pierre-Sassoulas, line 162:
```suggestion
            x+ y)  # [cell-var-from-loop, cell-var-from-loop]
```
I don't know if it's intentional or if you want to differentiate x and y to see that a message is send for each, but you could also do that.

Pierre-Sassoulas (approved):
Thank you for this MR ! Closes multiple issues, very well tested and also super clean, which is on par with that what you usually do 😄

tests/functional/c/cellvar_escaping_loop.py
david-yz-liu: @Pierre-Sassoulas I actually wasn't aware you could test for multiple error occurrences in the same line, good to know! 😅 

Since you've merged in this PR I'm assuming you don't actually need me to make this change, but I'll keep it in mind for the future.

kowalcj0:
Hi,

I'm little confused by this warning which `pylint 2.12.2` found in the following `example.py`:

```python
"""cell-var-from-loop example."""

grouped_messages = {
    "category_a": [
        {"key_a": 3},
        {"key_a": 1},
        {"key_a": 2},
    ],
    "category_b": [
        {"key_b": 5},
        {"key_b": 3},
        {"key_b": 8},
    ],
}
sorting_keys = {
    "category_a": "key_a",
    "category_b": "key_b",
}

sorted_messages = {}

for category, messages in grouped_messages.items():
    sorted_messages[category] = sorted(
        (message for message in messages),
        key=lambda msg: msg[sorting_keys[category]]
    )

print(sorted_messages)
```

```bash
$ pylint example.py 
************* Module example
example.py:25:41: W0640: Cell variable category defined in loop (cell-var-from-loop)

Your code has been rated at 8.33/10 (previous run: 8.33/10, +0.00)
```

I got rid of this warning by replacing `key=lambda msg: msg[sorting_keys[category]]` with `key=operator.itemgetter(sorting_keys[category])`.
TBH I'm not sure why pylint complains about `category` being defined in loop.
Especially, because both versions of the code, give exactly the same result:

```shell
python3 example.py 
{'category_a': [{'key_a': 1}, {'key_a': 2}, {'key_a': 3}], 'category_b': [{'key_b': 3}, {'key_b': 5}, {'key_b': 8}]}
```

DanielNoord:
This looks like a false positive to me. In any case, could you open a new issue for this? Continuing the discussion in closed PRs can lead to the discussion being lost (as they are not easily visible in the issue list).

PR #4829, opened 2021-08-11
Fix false positives for invalid-all-format (#4711)

### LINKED ISSUE #4711: False positive finding for E0605: Invalid format for __all__
### Steps to reproduce
Given a file `__init__.py`:
```python
from foo import bar
__all__ = list(globals().keys())
```

### Current behavior
Result of `pylint __init__.py`:
```
configuration:1:0: E0605: Invalid format for __all__, must be tuple or list (invalid-all-format)
```

### Expected behavior
No finding. The attribute is a list at runtime but I guess the check doesn't see this

### pylint --version output
```
pylint 2.9.3
astroid 2.6.2
Python 3.7.1 (default, Oct 22 2018, 13:16:18) [GCC]
```

kasium:
## Type of Changes

## Description

invalid-all-format only considers lists and tuples and the direct form (e.g `[]`) but ignores the builtin functions

Closes #4711


Pierre-Sassoulas (approved):
You did pretty much everything for this bug :D Created the issue analyzed the problem, small diff for the fix, proper tests, that's a pretty great first contribution :) And it will be in 2.10.0 :)

kasium:
Thanks a lot 😃

PR #4831, opened 2021-08-12
Fix false-positive used-before-assignment with := in Return node

### LINKED ISSUE #4828: `used-before-assignment` false positive with walrus operator
### Bug description

`used-before-assignment` false positive occurs on the following code:
```python
def foo() -> str:
    return bar if (bar := "") else ""
```

### Configuration

_No response_

### Command used

```shell
pylint foo.py
```

### Pylint output

```shell
E0601: Using variable 'bar' before assignment (used-before-assignment)
```

### Expected behavior

no error

### Pylint version

```shell
pylint 2.9.6
astroid 2.6.6
Python 3.10.0b4 (v3.10.0b4:2ba4b20854, Jul  9 2021, 21:56:21) [Clang 12.0.5 (clang-1205.0.22.9)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

cdce8p:
## Type of Changes

## Description

```py
def foo() -> str:
    return bar if (bar := "") else ""
```

Closes #4828

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -114,6 +114,10 @@ Release date: TBA
 
   Closes #3839
 
+* Fix false-positive ``used-before-assignment`` with an assignment expression in a ``Return`` node
+
+  Closes #4828
+
 
 What's New in Pylint 2.9.6?
 ===========================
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
That is clean !

Issue #3512, opened 2020-04-25
Disable certain messages by default


AWhetter:
Following on from the discussion on #746, this is a draft proposal to start disabling certain messages by default. Furthermore, to make it easier to run pylint side by side with flake8, we want to document which messages are duplicated in flake8 and possible make them easier to disable as a group.

The following table contains every pylint message excluding those from plugins and py3k checks. For each message it is noted whether it is a style check or a functional check (by functional I mean it validates some sort of run time behaviour of the code being linted or checks for programmer mistakes) or other (when it's technically a style check but I don't think we would want to disable it when a user opts out of style checks, or if the message doesn't fit in the other categories), whether it duplicates a message in flake8, and **my opinion** on whether I think it should be disabled by default so that we can use it as a discussion point. I normally prefer the linter to be quite strict so it might be worth gathering user feedback on this. Another option would be to give messages levels of strictness and you pick a level to lint at.

Message Name | Message ID | Style or Functional or Other | flake8 ID | Disable by Default | Notes 
--- | --- | --- | --- | --- | --- 
blacklisted-name | C0102 | S |  | N | 
invalid-name | C0103 | S |  | N | 
empty-docstring | C0112 | S |  | Y | 
unneeded-not | C0113 | O | F713 | N | 
missing-module-docstring | C0114 | S | "D100,D104" | Y | 
missing-class-docstring | C0115 | S | "D101,D106" | Y | 
missing-function-docstring | C0116 | S | "D102,D103,D105" | Y | 
singleton-comparison | C0121 | S |  | N | 
misplaced-comparison-constant | C0122 | S |  | N | 
unidiomatic-typecheck | C0123 | F |  | N | 
non-ascii-name | C0144 | F |  | N | 
consider-using-enumerate | C0200 | O |  | N | 
consider-iterating-dictionary | C0201 | O |  | N | 
bad-classmethod-argument | C0202 | O |  | N | 
bad-mcs-method-argument | C0203 | O |  | N | 
bad-mcs-classmethod-argument | C0204 | O |  | N | 
single-string-used-for-slots | C0205 | F |  | N | 
line-too-long | C0301 | S | E501 | N | 
too-many-lines | C0302 | S |  | N | 
trailing-whitespace | C0303 | S | "W291,W293" | N | 
missing-final-newline | C0304 | S | W292 | N | 
trailing-newlines | C0305 | S |  | N | 
multiple-statements | C0321 | O | E701 | N | 
[...]

PR #4832, opened 2021-08-12
Revert "Extended consider-using-tuple check to cover 'in' comparisons (#4768)"

cdce8p:
## Description

Revert changes made in #4768 after discussion #4776.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -48,10 +48,6 @@ Release date: TBA
 
   Closes #4365
 
-* ``CodeStyleChecker``
-
-  * Extended ``consider-using-tuple`` check to cover ``in`` comparisons.
-
 * Added ``forgotten-debug-statement``: Emitted when ``breakpoint``, ``pdb.set_trace`` or ``sys.breakpointhook`` calls are found
 
   Closes #3692
--- a/doc/whatsnew/2.10.rst
+++ b/doc/whatsnew/2.10.rst
@@ -29,14 +29,6 @@ New checkers
   Closes #3692
 
 
-Extensions
-==========
-
-* ``CodeStyleChecker``
-
-  * Extended ``consider-using-tuple`` check to cover ``in`` comparisons.
-
-
 Other Changes
 =============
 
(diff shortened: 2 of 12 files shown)


cdce8p:
@Pierre-Sassoulas I would like to revert the commit before starting to work on a replacement.
After that we might what to consider if the new "generic" `use-sequence-for-iterating` and `use-set-for-membership` checks should be enabled by default as refactoring checks. Technically they do provide a performance benefit.

The alternative would be to keep them in the `code_style` extension.

Pierre-Sassoulas (approved):
Couldn't we replace ``consider-using-tuple`` to ``use-sets-for-membership`` for every example in this MR without reverting first ? Thiw way we would not have to think about it because the initial MR was already well thought. I guess you know better how you want to do it so I'm approving anyway :)

> The alternative would be to keep them in the code_style extension.

We could keep them separated in their own checker so the decision is easier to make.  I think if we want to be able to do #3512 comfortably we'll have to separate more anyway, what do you think ?

cdce8p:
> Couldn't we replace `consider-using-tuple` to `use-sets-for-membership` for every example in this MR without reverting first ? This way we would not have to think about it because the initial MR was already well thought.

I tried that initially, but it became such a mess that I didn't really wanted to work on it. Starting from scratch is the better and cleaner approach IMO. I'll use most of it for the new check though.

> > The alternative would be to keep them in the code_style extension.
> 
> We could keep them separated in their own checker so the decision is easier to make. I think if we want to be able to do #3512 comfortably we'll have to separate more anyway, what do you think ?

I thought the additional effort wouldn't be worth it. We can still separate individual checks from the rest by creating methods for them. What it comes down to IMO is if we want to enable these by default or have them in an extension.
- If enabled by default, the `RecommendationChecker` might be a pretty good place (my favorite)
- Otherwise, I would suggest the `CodeStyle` extension.

If you do have another idea, especially a name for the new (?) checker, I would be open to that too.

Pierre-Sassoulas:
Finding name is hard 😄 ``AppropriateContainerChecker`` ? Using ``RecommendationChecker`` works too. I'm convinced by ``Technically they do provide a performance benefit.`` - albeit a small one, so I'm slightly in favor of making it default without strong opinion about it.

cdce8p:
Performance benefit and good coding practice.
I'll work on a PR with it as part of the `RecommendationChecker` for now. Can still change it later if we decide to.

PR #4835, opened 2021-08-12
Add `use-sequence-for-iteration` checker

cdce8p:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |

## Description

Part of https://github.com/PyCQA/pylint/issues/4776#issuecomment-890496656
It is more efficient to iterate over sequence types than sets.

```py
# old
for x in {1, 2, 3}:  # [use-sequence-for-iteration]
    pass
```

```py
# better
for x in (1, 2, 3):
    pass
```

PR #4838, opened 2021-08-13
Improvements `consider-using-tuple`

cdce8p:
## Type of Changes

## Description
Followup to https://github.com/PyCQA/pylint/pull/4835#issuecomment-898052028

* Add additional typing to `CodeStyleChecker`
* Limit `consider-using-tuple` to in-place defined lists to prevent double warnings.
* Emit `consider-using-tuple` even for starred expressions as they can be replaced as well.
```py
var = (1, 2, 3)

# old
[*var]

# new
(*var,)  # trailing comma is important
```

### DIFF
--- a/pylint/extensions/code_style.py
+++ b/pylint/extensions/code_style.py
@@ -36,7 +36,7 @@ class CodeStyleChecker(BaseChecker):
             "Emitted when dictionary values can be replaced by namedtuples or dataclass instances.",
         ),
         "R6102": (
-            "Consider using an in-place tuple%s",
+            "Consider using an in-place tuple instead of list",
             "consider-using-tuple",
             "Emitted when an in-place defined list or set can be "
             "replaced by a slightly faster tuple.",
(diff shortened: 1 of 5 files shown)


pylint/extensions/code_style.py
```diff
@@ -36,7 +36,7 @@ class CodeStyleChecker(BaseChecker):
             "Emitted when dictionary values can be replaced by namedtuples or dataclass instances.",
         ),
         "R6102": (
-            "Consider using an in-place tuple%s",
+            "Consider using an in-place tuple instead of list",
             "consider-using-tuple",
             "Emitted when an in-place defined list or set can be "
             "replaced by a slightly faster tuple.",
```
Pierre-Sassoulas, line 42: We should probably change that text to say that it's only for consistency as we have use-sequence-for-iteration that handle the performance issue.

cdce8p: Good catch, must have missed that! I've updated the documentation. Let me know what you think

Pierre-Sassoulas: 👍

PR #4844, opened 2021-08-14
Allow true / false in pylintrc

cdce8p:
## Type of Changes

## Description

Pylint does support rich types in `.toml` config files. However when copying a `true` / `false` option back to a `pylintrc` file, it is an error.

This is how the conversion is done for `.toml` files:
https://github.com/PyCQA/pylint/blob/09ffc07fcb9ac2ac99d56b1d2c818ce93e509426/pylint/config/option_manager_mixin.py#L280-L284

### Example
```toml
# toml
reports = true
```

```ini
# pylintrc

# old
reports = yes


# with PR also allowed
reports = true
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -130,6 +130,8 @@ Release date: TBA
 
   Closes #4839
 
+* Allow ``true`` and ``false`` values in ``pylintrc`` for better compatibility with ``toml`` config.
+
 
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
Make sense, thank you !

PR #4850, opened 2021-08-15
``pyreverse``: Add option for colored output

### LINKED ISSUE #4488: Proposal: automatically color modules in package diagram

I really like the possibility to generate package diagrams using ``pyreverse``. However, a complete diagram can become very big and complex. 
The currently just "black and white" output makes it hard to spot modules of the same package.
This is for example the package diagram for pylint itself:
image

By coloring modules of the same package with the same color, it is much easier to spot what parts belong together:
image

### Additional context

I also experimented with grouping modules of the same package together, but the result is not as clear:
image

To discuss is the best way to implement this. The image above was generated by automatically picking colors from a list based on the first two parts of the module name. 
Possible alternatives/extensions would be:
* Let the user define the colors to use in the configuration file, instead of picking colors automatically 
* Let the user define up to what depth he wants to distinguish subpackages

DudeNr33:

## Type of Changes

## Description
This MR adds two command line options to ``pyreverse``:
* ``--colorized``: Use colored output. Classes/modules of the same package get the same color.
* ``--max-color-depth``: Use separate colors up to package depth of <depth>

This feature is usable with ``DotPrinter`` and ``PlantUmlPrinter``. 
``VCGPrinter`` will simply ignore the color passed to it. 

Closes #4488

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,8 @@ Release date: TBA
 
 * pyreverse: add option to produce colored output.
 
+  Closes #4488
+
 * pyreverse: add output in PlantUML format.
 
   Closes #4498
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,7 +12,9 @@ Release date: TBA
 ..
   Put bug fixes that should not wait for a new minor version here
 
-+ pyreverse: add output in PlantUML format.
+* pyreverse: add option to produce colored output.
+
+* pyreverse: add output in PlantUML format.
 
   Closes #4498
 
--- a/doc/whatsnew/2.10.rst
+++ b/doc/whatsnew/2.10.rst
@@ -44,6 +44,8 @@ Extensions
 Other Changes
 =============
 
+* Pyreverse - add option to produce colored output.
+
 * Pyreverse - add output in PlantUML format
 
 * pylint does not crash with a traceback anymore when a file is problematic. It
--- a/doc/whatsnew/2.10.rst
+++ b/doc/whatsnew/2.10.rst
@@ -44,7 +44,7 @@ Extensions
 Other Changes
 =============
 
-* Pyreverse - add option to produce colored output.
+* pyreverse now permit to produce colored generated diagram by using the ``colorized`` option.
 
 * Pyreverse - add output in PlantUML format
 
--- a/pylint/pyreverse/main.py
+++ b/pylint/pyreverse/main.py
@@ -11,6 +11,7 @@
 # Copyright (c) 2020 hippo91 <guillaume.peillex@gmail.com>
 # Copyright (c) 2021 Marc Mueller <30130371+cdce8p@users.noreply.github.com>
 # Copyright (c) 2021 Mark Byrne <31762852+mbyrnepr2@users.noreply.github.com>
+# Copyright (c) 2021 Andreas Finkler <andi.finkler@gmail.com>
 
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
@@ -124,8 +125,7 @@ OPTIONS = (
             short="k",
             action="store_true",
             default=False,
-            help="don't show attributes and methods in the class boxes; \
-this disables -f values",
+            help="don't show attributes and methods in the class boxes; this disables -f values",
         ),
     ),
     (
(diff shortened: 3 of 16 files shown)


ChangeLog
```diff
@@ -12,7 +12,11 @@ Release date: TBA
 ..
   Put bug fixes that should not wait for a new minor version here
 
-+ pyreverse: add output in PlantUML format.
+* pyreverse: add option to produce colored output.
+
+  Closes #4488
+
+* pyreverse: add output in PlantUML format.
```
DudeNr33, line 19: Apparently I placed a "+" instead of "*" in my last MR, fixed this too.

pylint/pyreverse/main.py
```diff
@@ -139,37 +139,56 @@
...
-            "help": "Files or directories to be skipped. They "
-            "should be base names, not paths.",
-        },
+        dict(
+            type="csv",
+            metavar="<file[,file...]>",
+            dest="ignore_list",
+            default=("CVS",),
+            help="Files or directories to be skipped. They should be base names, not paths.",
+        ),
```
DudeNr33, line 170: The formatting for the options was a bit inconsistent. The majority uses ``dict(...)``, but some also used literal dicts. I cleaned those up to be consistent with the remaining options.

tests/pyreverse/test_writer.py
```diff
@@ -140,3 +175,11 @@ def _assert_files_are_equal(generated_file):
...
     )
     assert expected == generated, f"{files}{diff}"
+
+
+def test_color_for_stdlib_module(default_config):
+    writer = DiagramWriter(default_config)
+    obj = Mock()
+    obj.node = Mock()
+    obj.node.qname.return_value = "collections"
+    assert writer.get_color(obj) == "grey"
```
DudeNr33, line 185: I hope this test works well in the CI.
On my local computer it passes in the ``tox`` environment, but fails when running the test directly from VS code. I guess it has something to do with the way the virtual environment is set up. 
If it causes issues we can remove this test, I only included it because otherwise the corresponding could would have no test coverage. But given that the code is simple, having a more stable CI (and dev environments for other contributors) might be more important.

doc/whatsnew/2.10.rst
```diff
@@ -44,6 +44,8 @@ Extensions
 Other Changes
 =============
 
+* Pyreverse - add option to produce colored output.
```
Pierre-Sassoulas, line 47:
```suggestion
* pyreverse now permit to produce colored generated diagram by using the ``colorized`` option.
```

pylint/pyreverse/main.py
Pierre-Sassoulas: Thank you !

pylint/pyreverse/writer.py
```diff
@@ -123,10 +150,26 @@ def get_class_properties(self, obj: ClassEntity) -> NodeProperties:
             attrs=obj.attrs if not self.config.only_classnames else None,
             methods=obj.methods if not self.config.only_classnames else None,
             fontcolor="red" if is_exception(obj.node) else "black",
-            color="black",
+            color=self.get_color(obj) if self.config.colorized else "black",
         )
         return properties
 
+    def get_color(self, obj: DiagramEntity) -> str:
+        """get shape color"""
```
Pierre-Sassoulas, line 158:
```suggestion
    def get_shape_color(self, obj: DiagramEntity) -> str:
```

tests/pyreverse/data/classes_colorized.puml
```diff
@@ -0,0 +1,42 @@
...
+
+get_value()
+set_value(value)
+}
+class "Specialization" as data.clientmodule_test.Specialization #aliceblue {
+TYPE : str
+relation
+relation2
+top : str
+}
```
Pierre-Sassoulas, line 36:
```suggestion
class "Specialization" as data.clientmodule_test.Specialization #aliceblue {
    TYPE : str
    relation
    relation2
    top : str
}
```
I probably missed it in the last review but I think it would be nice to have proper indentation in generated diagrams. I think it might be modified a little after generation so it would help.

Pierre-Sassoulas (approved):
LGTM, I only have minor comments, this is definitely going in 2.10. Thanks again for the huge work you did on this!

tests/pyreverse/test_writer.py
Pierre-Sassoulas: It seems it's passing so probably something due to your VScode environment ? I don't see how this test could be flaky.

tests/pyreverse/data/classes_colorized.puml
DudeNr33: Yes, that would definitely make it more readable! ``VCGPrinter`` already had methods for handling the indentation, I moved them into the base ``Printer`` class and reused them in ``PlantUmlPrinter``.

PR #4851, opened 2021-08-15
Suppress ``consider-using-with`` on return statements

### LINKED ISSUE #4748: False-positive `consider-using-with` when returning a file handler
### Bug description

```python
def func():
    return open("test.py")  # false-positive

text = func().read()
print(text)
```

### (Configuration)

Using the following configuration:
```ini
```

### Command used

```
test.py:2:11: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
```

### (Expected behavior)

The example above might not be consider good code style, but it's enough to reproduce the issue. If the code is replaced by
```py
def func():
    with open("test.py") as fp:
        return fp

text = func().read()
print(text)
```
Python will through an exception
```
ValueError: I/O operation on closed file.
```

### (Version affected)

```shell
pylint 2.9.6-dev0
astroid 2.6.6-dev0
Python 3.10.0b4 (v3.10.0b4:2ba4b20854, Jul  9 2021, 21:56:21) [Clang 12.0.5 (clang-1205.0.22.9)]
```

### (OS / Environment)

_No response_

### (Additional dependencies)

_No response_

DudeNr33:

## Type of Changes

## Description
As discussed in #4748 the ``consider-using-with`` check should be suppressed on ``return`` statements.

The new proposal for ``consider-using-context-manager`` is not included in this MR. I think we should discuss the use cases where this message should - and especially should not - trigger before starting the implementation. 

Closes #4748

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,10 @@ Release date: TBA
 
   Closes #4498
 
+* ``consider-using-with`` is no longer triggered if a context manager is returned from a function.
+
+  Closes #4748
+
 * pylint does not crash with a traceback anymore when a file is problematic. It
   creates a template text file for opening an issue on the bug tracker instead.
   The linting can go on for other non problematic files instead of being impossible.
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):
I agree with you, let's first remove the "false positive" before adding another check.

PR #4853, opened 2021-08-16
Emit consider-using-tuple for in comparisons

cdce8p:
## Type of Changes

## Description
Alternative to #4841
In contrast to #4768, with this PR `consider-using-tuple` will only be emitted for lists in `in` comparisons.

As discussed in #4776, this change **will not** have any impact on performance. CPython does this optimization when converting to bytecode.

```py
# old
for x in [1, 2, 3]:
    ...

# new
for x in (1, 2, 3):
    ...
```


Pierre-Sassoulas:
I thought the point of #4776 was to replace that by ``use-set-for-membership``. Sets are the go to for in membership so if we don't recommend set we should not recommend something else imo. I don't think the problem with hashable being required for set is that huge,  ``use-set-for-membership`` is salvageable, I'm just delaying it in 2.11 so we can think about it and get more opinions.

cdce8p:
The reason I opened this one is that IMO we just shouldn't move forward with `use-set-for-membership`. The minimal performance improvement (if any at all) just isn't worth it. Dev's will overlook that it's a set when they add new items and then wonder why it isn't working.  A False-Positive isn't worth it, not for this one.

If we don't recommend sets, there is still some value in using tuples instead. As the docstring for the checker mentions: even only for style consistency.

Pierre-Sassoulas:
I added more information in #4776 so the discussion can go on.

PR #4656, opened 2021-07-02
Remove references to `astroid.node_classes`

cdce8p:
## Description
Remove references to `astroid.node_classes` in preparation for astroid refactoring.
All symbols are available through import from `astroid` itself.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |
| ✓   | :scroll: Docs          |

## Related Issue
https://github.com/PyCQA/astroid/pull/1057

PR #4855, opened 2021-08-16
Use alias for astroid.nodes 01

cdce8p:
## Type of Changes

## Description

Followup to #4656
Use `an` alias for `astroid.nodes`.

### DIFF
--- a/examples/custom.py
+++ b/examples/custom.py
@@ -1,4 +1,4 @@
-import astroid.nodes as an
+from astroid import nodes
 
 from pylint.checkers import BaseChecker
 from pylint.interfaces import IAstroidChecker
@@ -45,17 +45,17 @@ class MyAstroidChecker(BaseChecker):
         ),
     )
 
-    def visit_call(self, node: an.Call) -> None:
-        """Called when a :class:`.an.Call` node is visited.
+    def visit_call(self, node: nodes.Call) -> None:
+        """Called when a :class:`.nodes.Call` node is visited.
 
         See :mod:`an` for the description of available nodes.
 
         :param node: The node to check.
-        :type node: an.Call
+        :type node: nodes.Call
         """
         if not (
-            isinstance(node.func, an.Attribute)
-            and isinstance(node.func.expr, an.Name)
+            isinstance(node.func, nodes.Attribute)
+            and isinstance(node.func.expr, nodes.Name)
             and node.func.expr.name == self.config.store_locals_indicator
             and node.func.attrname == "create"
         ):
--- a/examples/custom.py
+++ b/examples/custom.py
@@ -1,4 +1,4 @@
-import astroid
+import astroid.nodes as an
 
 from pylint.checkers import BaseChecker
 from pylint.interfaces import IAstroidChecker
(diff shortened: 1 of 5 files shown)


cdce8p:

Pierre-Sassoulas:
Looks more terse ! Should we be using ``ae`` for exceptions too ? To be fair I'm not very satisfied with this solution but astroid first three letter are not very practical for abbreviation.

DanielNoord:
Personally I think this makes the codebase less intuitive for new contributors. Would it be possible to include `nodes` or `node` in the alias?

Pierre-Sassoulas:
@DanielNoord's comment made me realize that using ``from astroid import nodes`` is a better solution. Let's keep it simple. What do you think ?

cdce8p:
Ok, lets use `from astroid import nodes` then. In the end I just think it's important that we are able to distinguish between `List` from typing and `List` from astroid for example. I'll update the PR.

cdce8p:
I updated the PR. Had to fix some name conflicts along the way but nothing major. Let me know what you guys think!

examples/custom.py
```diff
@@ -45,17 +45,17 @@ class MyAstroidChecker(BaseChecker):
...
-    def visit_call(self, node):
-        """Called when a :class:`.astroid.node_classes.Call` node is visited.
+    def visit_call(self, node: nodes.Call) -> None:
+        """Called when a :class:`.nodes.Call` node is visited.
 
         See :mod:`astroid` for the description of available nodes.
 
         :param node: The node to check.
-        :type node: astroid.node_classes.Call
+        :type node: nodes.Call
```
Pierre-Sassoulas, line 54: I'm wondering if we should keep the docstring type when the type is in the signature now.

Pierre-Sassoulas (approved):
👍 Nice convention to use. What do you think @AWhetter ?

examples/custom.py
DanielNoord: I think removing would be fine.

However, I'm wondering if the typing shouldn't be `astroid.nodes.Call`. Not sure what is standard but I always type with whatever `type()` would return. Imo, this makes it easier for new contributors to understand what is going on without remembering all imports and aliases.

AWhetter: I think it's good to remove it from the docstring. Type annotations are usually more expressive than what you can do in a docstring. Plus there's the potential for the two places to go out of sync. Annotations show up in `help()` and in documentation (and in documentation it can even be resolved to a fully qualified name).

pylint/checkers/classes.py
```diff
@@ -1038,25 +1039,25 @@ def _check_attribute_defined_outside_init(self, cnode: astroid.ClassDef) -> None
             return
         defining_methods = self.config.defining_attr_methods
         current_module = cnode.root()
-        for attr, nodes in cnode.instance_attrs.items():
+        for attr, nodes_lst in cnode.instance_attrs.items():
```
AWhetter, line 1042: Trivial: I'd call `nodes_lst` something like `attr_nodes` instead to better describe what it is.

AWhetter (approved):
I love it! Thanks for making this change.
We'll have to be careful about not making more variables called `nodes` in the future (the linting should pick this up), but that's a good thing because it will encourage us to write a more descriptive name for the variable.

cdce8p:
> `astroid.nodes.Call`

@DanielNoord Usually you would use just the class name itself, so `Call` in this case. However that will result in name conflicts for `List` and `Tuple` as those are also in the `typing` module. The idea behind this case then was to find something to resolve that which having to type `astroid.nodes` every time. Thus `an` as first suggestion.
If you like, take a look at this comment here: https://github.com/PyCQA/pylint/pull/4656#issuecomment-873216448

Regarding your concern with new contributors. It might be somewhat true, but consider that most IDEs will help with that. It will also be simpler to just have one convention instead of the patchwork we have at the moment.

DanielNoord:
I understand the need for an alias in the code, but I was wondering about the typing hints.
For example, `node: nodes.Call` or `node: astroid.nodes.Call`. Although the latter adds additional characters to me it seems more helpful for new contributors. They can immediately see that the nodes come from an import outside of the `pylint` package. With `nodes.Call` they will need to scroll up to look at the imports, or have their IDE show it for them. The latter isn't possible (I believe) on GitHub for example.
I totally agree that this is very nit picky and I don't know what is standard practice but if we are going to change all typing hints anyway, this seems like a decision worth considering!

Pierre-Sassoulas:
I understand the concern thank you for voicing it. I think`astroid.nodes` is long enough to have a negative impact greater than the impact of having to check where the typing come from the first time. Also, 'nodes' alone is clearer than 'astroid' (you don't have to know what astroid does).

PR #4856, opened 2021-08-16
pyreverse - Add project modules to sys.path

### LINKED ISSUE #2479: pyreverse:  pyreverse doesn't link sub-classes to an abstract class
### Steps to reproduce
1. Run `git clone git@github.com:pixelogik/NearPy.git`
2. Run `cd nearpy/storage`
3. Run ` pyreverse -o .png .`
4. Open classes_No_Name..png
### Current behavior
You will see a UML diagram like this.This has three classes `RedisStorage`, `MemoryStorage` and an abstract class `Storage`. See the diagram here. 
https://drive.google.com/file/d/1MQ8igLIhNIfVUM0VyXfrcFYdOwDs6V0f/view?usp=sharing

### Expected behavior
classes `RedisStorage` and `MemoryStorage` should inherit from abstract class `Storage` 

### Pylint version 
pylint (1.1.0)

mbyrnepr2:
Closes #2479

## Type of Changes

## Description
The pyreverse 'import from' logic is affected by the sys.path contents on this line: https://github.com/PyCQA/astroid/blob/main/astroid/interpreter/_import/spec.py#L353

Can get to this point by placing a breakpoint() on the following line and following the import logic for one of the import statements: https://github.com/PyCQA/astroid/blob/main/astroid/inference.py#L271

Currently the current directory is placed in sys.path. 
With the change, all modules can be found as they are all placed in sys.path similarly to how pylint itself does.

Closes #2479

### DIFF
--- a/pylint/pyreverse/main.py
+++ b/pylint/pyreverse/main.py
@@ -21,11 +21,11 @@
 
   create UML diagrams for classes and modules in <packages>
 """
-import os
 import sys
 from typing import Iterable
 
 from pylint.config import ConfigurationMixIn
+from pylint.lint.utils import fix_import_path
 from pylint.pyreverse import writer
 from pylint.pyreverse.diadefslib import DiadefsHandler
 from pylint.pyreverse.inspector import Linker, project_from_files
@@ -212,10 +212,7 @@ class Run(ConfigurationMixIn):
         if not args:
             print(self.help())
             return 1
-        # insert current working directory to the python path to recognize
-        # dependencies to local modules even if cwd is not in the PYTHONPATH
-        sys.path.insert(0, os.getcwd())
-        try:
+        with fix_import_path(args):
             project = project_from_files(
                 args,
                 project_name=self.config.project,
@@ -224,9 +221,7 @@ class Run(ConfigurationMixIn):
             linker = Linker(project, tag=True)
             handler = DiadefsHandler(self.config)
             diadefs = handler.get_diadefs(project, linker)
-        finally:
-            sys.path.pop(0)
-        writer.DiagramWriter(self.config).write(diadefs)
+            writer.DiagramWriter(self.config).write(diadefs)
         return 0
 
 


Pierre-Sassoulas (changes requested):
This look great, but would you mind adding a test for it ? I'm aware it's not the easiest thing to test automatically but I'm afraid we could regress easily on this one.

mbyrnepr2:
FYI @Pierre-Sassoulas. I finally got back to it :)

tests/pyreverse/test_writer.py
```diff
@@ -51,6 +51,14 @@
     "output_directory": "",
 }
 
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "..", "data")
```
Pierre-Sassoulas, line 54: Great new variable 👍

tests/pyreverse/test_main.py
```diff
@@ -0,0 +1,25 @@
...
+
+import pytest
+
+from pylint.lint import fix_import_path
+
+TEST_DATA_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "data"))
+PROJECT_ROOT_DIR = os.path.abspath(os.path.join(TEST_DATA_DIR, ".."))
+
+
+@pytest.fixture(name="setup_path", params=[PROJECT_ROOT_DIR, TEST_DATA_DIR])
```
Pierre-Sassoulas, line 12: Why not name the fixture ``setup_path`` directly ?

tests/pyreverse/test_main.py
```diff
@@ -0,0 +1,25 @@
...
+def fixture_setup_path(request):
```
Pierre-Sassoulas, line 13:
```suggestion
def fixture_setup_path(request: ?) -> Iterator[str]:
```
Could you add typing, please ? :)

Pierre-Sassoulas (commented):
To be fair, I don't really understand what is done in the added test, could you explain in a docstring, please ?

tests/pyreverse/test_main.py
mbyrnepr2: That was the plan initially but I see pylint gives: 'W0621: Redefining name 'setup_path' from outer scope (line 13) (redefined-outer-name)'.

From the the pytest docs:
```
If a fixture is used in the same module in which it is defined, the function name of the fixture will be shadowed by the function arg that requests the fixture; one way to resolve this is to name the decorated function fixture_<fixturename> and then use @pytest.fixture(name='<fixturename>').
```
https://docs.pytest.org/en/stable/reference.html#pytest-fixture

tests/pyreverse/test_main.py
mbyrnepr2: Sure.

mbyrnepr2:
@Pierre-Sassoulas if the added test is unclear or seems inaccurate perhaps we could drop it for another approach? I do feel it could be a bit unclear now.

The thinking was that with this issue, the first scenario below produces a file with the inheritance broken, but the second scenario is ok:

1.
cd pylint/tests/data
pyreverse -o .png .

vs.

2.
cd pylint/tests
pyreverse -o .png .

tests/pyreverse/test_main.py
DudeNr33: You could move the fixture definition into ``conftest.py`` to resolve this. 
An alternative (@Pierre-Sassoulas) would be to use ``pylint-pytest``, which suppresses among others this.

Pierre-Sassoulas: Let's do it ! On top of removing false positive, we'll eat our own dog food and this will be a compatibility tests for the latest pylint too.

mbyrnepr2: Thanks for this idea @DudeNr33 . I'm happy to use this approach.
I now realise that the test can have the same result but simplified, here is the update while using the earlier approach from @Pierre-Sassoulas : 37f3a5005771bffa8a6539b4f4b70f56fe933d0b

Pierre-Sassoulas:
Could you rebase on the latest main, please ? We switched to python 3.10 from 3.10-dev so the required pipelines changed.

PR #4860, opened 2021-08-17
Add extension check against use of while loops

### LINKED ISSUE #4367: New check: Ban while-loops
The `while` operator is needed to make Python Turing-complete. But
unless you are implementing something like a theorem prover or an
interpreter, there is very little call to use a true `while` loop. In
practically every case a `while` loop can be converted to a `for` loop
with improved clarity. Seriously, how many web apps have a need for
unbounded computation?

One exception to this is the use of `while` loops at the tippity-top
of applications as a main execution loop. But other than that, genuine
need for `while` is rare.

I would like to see a check that simply flags any use of `while`. This
is an opinionated check, so make it an extension.

nickdrozd:
## Type of Changes

## Description

Closes #4367


pylint/reporters/ureports/__init__.py
```diff
@@ -75,7 +75,7 @@ def get_table_content(self, table):
             cols -= 1
             result[-1].append(cell)
         # fill missing cells
-        while len(result[-1]) < cols:
+        for _ in range(cols - len(result[-1])):
             result[-1].append("")
```
Pierre-Sassoulas, line 79:
```suggestion
        result[-1].append("" * (cols - len(result[-1])))
```

pylint/checkers/utils.py
```diff
@@ -671,7 +671,7 @@ def node_frame_class(node: astroid.node_classes.NodeNG) -> Optional[astroid.Clas
         and not isinstance(klass, astroid.ClassDef)
     ):
         if klass.parent is None:
-            klass = None
+            return None
         else:
             klass = klass.parent.frame()
```
Pierre-Sassoulas, line 676:
```suggestion
        klass = klass.parent.frame()
```

pylint/extensions/typing.py
```diff
@@ -293,11 +293,7 @@ def leave_module(self, node: astroid.Module) -> None:
         """
         if self._py37_plus() and not self._py39_plus():
             msg_future_import = self._msg_postponed_eval_hint(node)
-            while True:
-                try:
-                    msg = self._consider_using_alias_msgs.pop(0)
-                except IndexError:
-                    break
+            for msg in self._consider_using_alias_msgs:
```
Pierre-Sassoulas, line 296: Wow, this one was particularly bad, good job fixing it 👌

Pierre-Sassoulas (changes requested):
Thank you, this is pretty straightforward and the fixes to the existing code are on point too. Could you add an entry in what's new please ?

pylint/extensions/while.py
```diff
@@ -0,0 +1,30 @@
...
+        self.add_message("while-used", node=node)
+
+
+def register(linter):
+    """Required method to auto register this checker.
+
+    :param linter: Main interface object for Pylint plugins
+    :type linter: Pylint object
+    """
+    linter.register_checker(WhileChecker(linter))
```
Pierre-Sassoulas, line 30: We should not register it, or it become a default checker and we don't want that :) But we can add a functional test where the plugin is loaded to check it.

pylint/reporters/ureports/__init__.py
nickdrozd: Should be `result[-1] += "" * (cols - len(result[-1])`, right?

pylint/extensions/while.py
nickdrozd: I think the fact that it is in `extensions` keeps it from being a default.

In any case, removing the register line causes `AttributeError: module 'pylint.extensions.while_used' has no attribute 'register'`.

pylint/reporters/ureports/__init__.py
Pierre-Sassoulas: Hmm, nice catch, it's actually ``result[-1] += [""] * (cols - len(result[-1]))``

pylint/extensions/while.py
Pierre-Sassoulas: Sorry, my bad :)

Pierre-Sassoulas (changes requested):
Could you also add two functional tests, one  with the plugin activated, the other without configuration, please ?

nickdrozd:
> Could you also add two functional tests, one with the plugin activated, the other without configuration, please ?

I added a functional test with the plugin activated, but I'm not sure what you mean about the other one. There are a variety of `while` loops used throughout the other functional tests and none of them are flagged. Is that what you had in mind?

Pierre-Sassoulas (approved):
Thank you for adding the test :) You're right, there's no need for another test, I did not think of that. This will be available in pylint 2.10.0 !

PR #4863, opened 2021-08-17
Use alias for astroid.nodes 02

cdce8p:
## Type of Changes

## Description

Followup to #4656

### DIFF
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -41,6 +41,7 @@ import tokenize
 from typing import TYPE_CHECKING, Iterable
 
 import astroid
+from astroid import nodes
 
 from pylint.checkers import BaseChecker, BaseTokenChecker, utils
 from pylint.checkers.utils import check_messages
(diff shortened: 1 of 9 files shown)


pylint/checkers/strings.py
```diff
@@ -213,13 +214,13 @@
 }
 
 OTHER_NODES = (
```
Pierre-Sassoulas, line 216: Not related to this MR but this variable name is hard to understand.

Pierre-Sassoulas (approved):
Nice work ! Just a nitpick we could remove docstrings and replace them by typing in signature when we have to update docstrings.

cdce8p:
> Nice work ! Just a nitpick we could remove docstrings and replace them by typing in signature when we have to update docstrings.

Doing so would enable type checking in that function which often results in mypy errors. I think it would be better if we do that another time.

Pierre-Sassoulas:
You're right.

PR #4866, opened 2021-08-18
Use alias for astroid.nodes 03

cdce8p:
## Type of Changes

## Description

Followup to #4656

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -73,15 +73,15 @@ from typing import (
 import _string
 import astroid
 import astroid.objects
-from astroid import TooManyLevelsError
+from astroid import TooManyLevelsError, nodes
 
 from pylint.constants import BUILTINS
 
 COMP_NODE_TYPES = (
-    astroid.ListComp,
-    astroid.SetComp,
-    astroid.DictComp,
-    astroid.GeneratorExp,
+    nodes.ListComp,
+    nodes.SetComp,
+    nodes.DictComp,
+    nodes.GeneratorExp,
 )
 EXCEPTIONS_MODULE = "builtins"
 ABC_MODULES = {"abc", "_py_abc"}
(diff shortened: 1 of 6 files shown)


pylint/checkers/utils.py
```diff
@@ -770,65 +767,65 @@ def decorated_with_property(node: astroid.FunctionDef) -> bool:
...
-                    and isinstance(inferred.function, astroid.FunctionDef)
+                    and isinstance(inferred.function, nodes.FunctionDef)
                 ):
                     return decorated_with_property(inferred.function)
     return False
 
 
 def decorated_with(
-    func: Union[astroid.FunctionDef, astroid.BoundMethod, astroid.UnboundMethod],
+    func: Union[nodes.FunctionDef, astroid.BoundMethod, astroid.UnboundMethod],
```
Pierre-Sassoulas, line 822: I was suspicious that astroid.bases was containing nodes too.. `BoundMethod` and `UnboundMethod` could be considered like nodes, right ?

pylint/checkers/utils.py
```diff
@@ -1528,16 +1519,14 @@ def get_subscript_const_value(node: astroid.Subscript) -> astroid.Const:
     :raises InferredTypeError: if the subscript node cannot be inferred as a Const
     """
     inferred = safe_infer(node.slice)
-    if not isinstance(inferred, astroid.Const):
-        raise InferredTypeError(
-            "Subscript.slice cannot be inferred as an astroid.Const"
-        )
+    if not isinstance(inferred, nodes.Const):
+        raise InferredTypeError("Subscript.slice cannot be inferred as an nodes.Const")
```
Pierre-Sassoulas, line 1523:
```suggestion
        raise InferredTypeError("Subscript.slice cannot be inferred as a nodes.Const")
```

pylint/checkers/variables.py
```diff
@@ -713,9 +714,7 @@ def __init__(self, linter=None):
 
     @utils.check_messages("redefined-outer-name")
     def visit_for(self, node):
-        assigned_to = [
-            var.name for var in node.target.nodes_of_class(astroid.AssignName)
-        ]
+        assigned_to = [var.name for var in node.target.nodes_of_class(nodes.AssignName)]
```
Pierre-Sassoulas, line 717:
```suggestion
        assigned_to = [a.name for a in node.target.nodes_of_class(nodes.AssignName)]
```

pylint/checkers/variables.py
```diff
@@ -1656,10 +1644,10 @@ def _loopvar_name(self, node, name):
...
             sequences = (
-                astroid.List,
-                astroid.Tuple,
-                astroid.Dict,
-                astroid.Set,
+                nodes.List,
+                nodes.Tuple,
+                nodes.Dict,
+                nodes.Set,
                 astroid.objects.FrozenSet,
```
Pierre-Sassoulas, line 1651: Look like there's other class too that should go in astroid.nodes ?

Pierre-Sassoulas (approved):
Great cleanup !

pylint/checkers/utils.py
cdce8p: `BoundMethod` and `UnboundMethod` inherit from `Proxy` not `NodeNG`. So I would say no.

pylint/checkers/variables.py
cdce8p: From the `astroid.objects` docstring:
```py
"""
Inference objects are a way to represent composite AST nodes,
which are used only as inference results, so they can't be found in the
original AST tree. [...]
"""
```

As they are only used in inference, I don't think they should move to the `nodes` package. At least not for now.

PR #4868, opened 2021-08-18
``astroid.const.BUILTINS`` use removed for clarity and so astroid can also removes it later

Pierre-Sassoulas:
## Type of Changes

## Description

Replace the constant BUILTINS by the string 'builtins'
    
This make for clearer and also slightly faster code (means time
seems to decrease by 0.68% with this change alone (astroid/pylint)
in the pylint tests benchmarks). Done originally because we were using an
import from astroid from astroid.bases for one of those, which is
kinda messy.

### DIFF
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -70,7 +70,6 @@ from typing import Pattern, Tuple
 
 import astroid
 from astroid import nodes
-from pylint.constants import PY310_PLUS
 
 from pylint.checkers import BaseChecker, utils
 from pylint.checkers.utils import (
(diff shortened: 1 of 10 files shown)


pylint/checkers/typecheck.py
```diff
@@ -70,6 +70,7 @@
 
 import astroid
 from astroid import nodes
+from astroid.const import PY310_PLUS
```
cdce8p, line 73: `PY310` should probably be imported from `pylint.const`

cdce8p (commented):
Just one comment. Aside from it, this looks good!

pylint/checkers/typecheck.py
Pierre-Sassoulas:
```suggestion
from pylint.constants import PY310_PLUS
```

PR #4869, opened 2021-08-18
Use alias for astroid.nodes 04

cdce8p:
## Type of Changes

## Description

Followup to #4656

--
This should be the last PR. There might be a few places left that haven't been updated (especially regarding the functions from `astroid.nodes` -> we might still what to import them just from `astroid` ??? ). In the end though, it's probably best to deal with them once the old way is removed in astroid.

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -126,7 +126,7 @@ def _is_a_return_statement(node: nodes.Call) -> bool:
 
 def _is_part_of_with_items(node: nodes.Call) -> bool:
     """
-    Checks if one of the node's parents is an ``nodes.With`` node and that the node itself is located
+    Checks if one of the node's parents is a ``nodes.With`` node and that the node itself is located
     somewhere under its ``items``.
     """
     frame = node.frame()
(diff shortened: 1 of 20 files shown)


pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -102,44 +103,44 @@ def get_curline_index_start():
...
             return True
         parent = parent.parent
     return False
 
 
-def _is_part_of_with_items(node: astroid.Call) -> bool:
+def _is_part_of_with_items(node: nodes.Call) -> bool:
     """
-    Checks if one of the node's parents is an ``astroid.With`` node and that the node itself is located
+    Checks if one of the node's parents is an ``nodes.With`` node and that the node itself is located
```
Pierre-Sassoulas, line 129:
```suggestion
    Checks if one of the node's parents is a ``nodes.With`` node and that the node itself is located
```

tests/pyreverse/test_utils.py
```diff
@@ -102,7 +103,7 @@ def test_infer_node_2(mock_infer, mock_get_annotation):
 
 
 def test_infer_node_3():
-    """Return a set containing an astroid.ClassDef object when the attribute
+    """Return a set containing an nodes.ClassDef object when the attribute
```
Pierre-Sassoulas, line 106:
```suggestion
    """Return a set containing a nodes.ClassDef object when the attribute
```

Pierre-Sassoulas (approved):
👍

PR #4871, opened 2021-08-19
Add consider-if-expression extension

### LINKED ISSUE #4366: New check: Use ternary for single assignment
A variable gets a value assigned conditionally:

```python
if condition():
    x = 4:
else:
    x = 5
```

I don't like blocks like this because they make the situation look
more complex than it really is. A single variable is getting a value
assigned. What that value is is conditional, but the assignment
itself is not. There's one variable and one assignment, so there
should be just one statement with a ternary expression:

```python
x = 4 if condition() else 5
```

This is a rather opinionated check, so it should be added as an
extension rather than enabled by default.

nickdrozd:
## Type of Changes

## Description

Closes #4366

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -69,7 +69,11 @@ Release date: TBA
 
   Closes #4365
 
-* Added optional extension ``while_used``: Emitted whenever a ``while`` loop is used.
+* Added optional extension ``consider-ternary-expression``: Emitted whenever a variable is assigned in both branches of an if/else block.
+
+  Closes # 4366
+
+* Added optional extension ``while-used``: Emitted whenever a ``while`` loop is used.
 
   Closes # 4367
 
--- a/doc/whatsnew/2.10.rst
+++ b/doc/whatsnew/2.10.rst
@@ -24,7 +24,11 @@ New checkers
 
   Closes #4365
 
-* Added optional extension ``while_used``: Emitted whenever a ``while`` loop is used.
+* Added optional extension ``consider-ternary-expression``: Emitted whenever a variable is assigned in both branches of an if/else block.
+
+  Closes # 4366
+
+* Added optional extension ``while-used``: Emitted whenever a ``while`` loop is used.
 
   Closes # 4367
 
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -358,10 +358,7 @@ def _determine_function_name_type(node: nodes.FunctionDef, config=None):
         # form, treat it like an attribute as well.
         return "attr"
 
-    if node.decorators:
-        decorators = node.decorators.nodes
-    else:
-        decorators = []
+    decorators = node.decorators.nodes if node.decorators else []
     for decorator in decorators:
         # If the function is a property (decorated with @property
         # or @abc.abstractproperty), the name type is 'attr'.
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -151,10 +151,7 @@ def _repr_tree_defs(data, indent_str=None):
     lines = []
     nodes_items = data.items()
     for i, (mod, (sub, files)) in enumerate(sorted(nodes_items, key=lambda x: x[0])):
-        if not files:
-            files = ""
-        else:
-            files = "(%s)" % ",".join(sorted(files))
+        files = "" if not files else "(%s)" % ",".join(sorted(files))
         if indent_str is None:
             lines.append(f"{mod} {files}")
             sub_indent_str = "  "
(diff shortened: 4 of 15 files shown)


nickdrozd:
Something is off with the functional test specification. Am I missing something obvious?

Pierre-Sassoulas (changes requested):
Hey, thank you for implementing this. Could you revert all the ternary changes that do not fit on one line using black ? I think bigger ternary make the code harder to understand.

cdce8p:
> Hey, thank you for implementing this. Could you revert all the ternary changes that do not fit on one line using black ? I think bigger ternary make the code harder to understand.

I agree with that. Ternary ifs should only really be used for short conditions, something like
```py
var = 2 if some_var else 3
```

Pierre-Sassoulas:
To the point that I'm wondering if the checker should not be changed to :
- if it's a ternary and it's longer than 88 characters, tell the user to remove the ternary
- It's it's not a ternary and it's shorter than 88 characters, tell the user to consider a ternary

What do you think @nickdrozd  ?

cdce8p:
> To the point that I'm wondering if the checker should not be changed to :
> 
> * if it's a ternary and it's longer than 88 characters, tell the user to remove the ternary
> * It's it's not a ternary and it's shorter than 88 characters, tell the user to consider a ternary

I'm really not sure we should enforce any of it. IMO it doesn't make the code inherently better to use either.
I.e. even for the existing pylint code I don't think we should do most of these changes.

nickdrozd:
Okay, I took out the changes that don't result in one-liners.

Personally, I would change all of them. It's not about lines of text, it's about complexity -- one single assign statement is less complex than two. It's a shame that Python makes if-expressions so clunky to use. In Lisp and Rust if-expressions are used all over the place with no problems.

As for the check itself, I think it's okay if it suggests some more extreme changes, like if-expressions that would get broken across multiple lines. The check is an extension, disabled by default. So anyone who enables the extension can be safely assumed to be a relatively sophisticated Pylint user, and therefore also someone who can deal with any false positives or iffy cases.

pylint/extensions/consider_if_expression.py
```diff
@@ -0,0 +1,53 @@
...
+from astroid import nodes
+
+from pylint.checkers import BaseChecker
+from pylint.interfaces import IAstroidChecker
+
+
+class ConsiderIfExpressionChecker(BaseChecker):
+
+    __implements__ = (IAstroidChecker,)
+    name = "consider_if_expression"
```
Pierre-Sassoulas, line 12:
```suggestion
    name = "consider_ternary_expression"
```

pylint/extensions/consider_if_expression.py
```diff
@@ -0,0 +1,53 @@
...
+    msgs = {
+        "W0160": (
+            "Consider rewriting as an if-expression",
```
Pierre-Sassoulas, line 15:
```suggestion
            "Consider rewriting as a ternary if-expression",
```

pylint/extensions/consider_if_expression.py
```diff
@@ -0,0 +1,53 @@
...
+            "consider-if-expression",
```
Pierre-Sassoulas, line 16:
```suggestion
            "consider-ternary-expression",
```

pylint/extensions/consider_if_expression.py
```diff
@@ -0,0 +1,53 @@
...
+            "Multiple assign statements spread across if/else blocks can be "
+            "rewritten with a single assingment and an if-expression",
```
Pierre-Sassoulas, line 18:
```suggestion
            "rewritten with a single assignment as a ternary conditional",
```

pylint/extensions/consider_if_expression.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+    }
+
+    def visit_if(self, node):
```
Pierre-Sassoulas, line 22: Could you add the typing, please ? :)

Pierre-Sassoulas (changes requested):
Thank you for cleaning up the MR, I left some comment, mostly I think the expected name is "ternary operator" or "ternary conditional".

Pierre-Sassoulas:
> As for the check itself, I think it's okay if it suggests some more extreme changes,

Yes I agree, I was trying to transform it into something complicated that I would personally use, but if someone want only ternary and have to activate this plugin, then it can be very opinionated without problems.

Pierre-Sassoulas (approved):
Great work, thank you !

zrothberg:
https://nickdrozd.github.io/2021/09/02/new-pylint-checks.html

This dude just used your repo to write a terrible article.

PR #4875, opened 2021-08-19
Improve invalid-metaclass error message for Instances

david-yz-liu:
## Type of Changes

## Description

Ref https://github.com/PyCQA/pylint/pull/4870#discussion_r691830356

Previously, the `invalid-metaclass` message prints out the full class text when given an `Instance` of a `ClassDef`. I added a check so that their repr (`Instance of ...`) would be displayed instead.

As a comment, because `bases.Instance` has some subclasses like `nodes.Const` and `nodes.Dict` where we *do* want to call `as_string`, I used an unidiomatic type check to pull out `Instance` objects directly. Not amazing, but I left a comment explaining why. I feel like astroid should support a separate subclass of `Instance` for "instance of a class with an unknown value" to differentiate it from `Const`/`Dict`/etc.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -160,6 +160,8 @@ Release date: TBA
 
 * Improve performance when inferring ``Call`` nodes, by utilizing caching.
 
+* Improve error message for invalid-metaclass when the node is an Instance.
+
 
 What's New in Pylint 2.9.6?
 ===========================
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Small straightforward fix, thanks a lot 👍

PR #4876, opened 2021-08-19
Add `consider-using-assignment-expr` to `CodeStyleChecker`

### LINKED ISSUE #4862: New check: Walrus operator recommended
### Current problem

Consider this still-common pattern:

```python
match = re.match(regex, string)
if match:
    do_stuff(match)
```

### Desired solution

As of Python 3.8, this can be rewritten with the "walrus":

```python
if (match := re.match(regex, string)):
    do_stuff(match)
```

Pylint should recommend using the walrus operator in situations like this.

### Additional context

See https://www.python.org/dev/peps/pep-0572/

cdce8p:
## Type of Changes

## Description

Add `consider-using-assignment-expr` check to `CodeStyleChecker` extension.

```py
# old
x = 2
if x:
    ...

# new
if (x := 2):
    ...
```
In this basic example brackets aren't really necessary. However I would still like to suggest them, as this will prevent errors in the future if the test changes.

Since the assignment expression requires Python 3.8, I modified the existing `py-version` setting from the `TypingChecker` extension, to be a global setting in the `[MASTER]` section. The current implementation is fully backwards compatible. It doesn't matter in which section `py-version` is defined. If not set, this will default to the current interpreter version.

The config change is already quite isolated. It wouldn't be an issue to create a separate PR for that if we want that.

Closes #4862


Pierre-Sassoulas (commented):
I would replace all reference to 'assignment expression' by 'walrus operator', this is the description I've seen in all python ressources talking about this personally.

cdce8p:
> I would replace all reference to 'assignment expression' by 'walrus operator', this is the description I've seen in all python ressources talking about this personally.

`Assignment expression` is the official name for it. Also the one used in the Python documentation.
https://docs.python.org/3.10/reference/expressions.html?highlight=walrus#assignment-expressions

--
This here is from PEP 572:
> During discussion of this PEP, the operator became informally known as "the walrus operator". The construct's formal name is "Assignment Expressions" (as per the PEP title), but they may also be referred to as "Named Expressions" (e.g. the CPython reference implementation uses that name internally).

https://www.python.org/dev/peps/pep-0572/#abstract

--
The help text mentioned `:=` directly and there is even a suggestion for what to use instead. So I don't think it's necessary to change that.

Pierre-Sassoulas (commented):
The checker itself is nice and could go in 2.11.

pylint/extensions/code_style.py
```diff
@@ -135,6 +152,79 @@ def _check_dict_consider_namedtuple_dataclass(self, node: nodes.Dict) -> None:
...
+
+        prev_sibling = node.previous_sibling()
+        if self._check_prev_sibling_to_if_stmt(prev_sibling, node_name.name):
+            test_str = node.test.as_string().replace(
+                node_name.name,
+                f"({node_name.name} := {prev_sibling.value.as_string()})",  # type: ignore
+                1,
+            )
+            suggestion = f"if {test_str}:"
+            if len(suggestion) > 88:
```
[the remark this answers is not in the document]
cdce8p: I haven't pushed that latest changes yet as I'm still working on it. There I added a new config value so users can change it.

> If we really want to do this I imagine we should behave differently if line-too-long is disabled or not and we should review all the checkers to see if it would also make sense...

I don't think that is necessary. Personally, I consider assignment expressions a special case as they are a truly optional feature that should only be used where it makes sense. It is completely independent from `line-too-long`.

Furthermore, I can't thing of any specific existing and none optional check that could benefit from such a setting. At least not of the top of my head. Once we add it and users would expect it to work for check x / y that doesn't work yet, I'm sure we would see requests for it and can change it then.

> But beside that I think it's counterintuitive to have a message change if the line is shorter or longer. I don't think we want to surprise user like that. [...] I think it's more reasonable to let the user know all the time that they can use a walrus operator, and if the line become too long because of it, they could refactor it so it's shorter.

Again, I would disagree here. A refactor is not always possible (or easy do accomplish). It's much more helpful to only recommend it in cases where there is a clear benefit from using it. It's just a recommendation for an optional feature after all.

One example, I think users won't like it if we recommend the use of `:=`, but it results in an additional line because black reformatted it. Now they will need to add a `pylint: disable`. If they do it two, three times, they will just disable the check completely and don't get anything from it. That doesn't help anyone.

pylintrc
```diff
@@ -36,6 +36,9 @@ unsafe-load-any-extension=no
 # run arbitrary code
 extension-pkg-allow-list=
 
+# Minimum supported python version
+py-version = 3.6
```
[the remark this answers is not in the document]
cdce8p: > Is this done in order to be able to lint py36 code using python 3.8 ?

Exactly. We don't want to recommend features that aren't yet available. I.e. I currently use 3.10 for development and running local pylint checks. Without this settings, pylint would recommend the use of assignment expressions locally. If a users wouldn't know about that, they might only find out about it when they run the CI check on their PR.

> But I'm a little afraid of what would need to be done on other checkers to keep that promise.

The `TypingChecker` extension already uses a similar setting. I would imagine that anything that checks deprecations could benefit as well. Although, those are easily handled due to the long deprecations periods and available alternatives.

cdce8p:
> The checker itself is nice and could go in 2.11.

That is my intention 🙂
I'll continue working on it over the next few days and let you know then.

pylintrc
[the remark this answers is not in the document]
Pierre-Sassoulas: Fair enough :)

pylint/extensions/code_style.py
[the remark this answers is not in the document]
Pierre-Sassoulas: > It is completely independent from line-too-long.

I don't think a checker that change behavior based on line length should be separated from our max-line-length option. If the value is 120 in max-line-length it means the user want to make 120 char lines. It would be strange to not raise a warning because more than 88 char is considered too much by this checker, when the user already explicitly said that 120 is okay.

pylint/extensions/code_style.py
```diff
@@ -41,12 +48,39 @@ class CodeStyleChecker(BaseChecker):
...
+            "Emitted when an if assignment is directly followed by an if statement and "
+            "both can be combined by using an assignment expression ``:=``. "
+            "Requires Python 3.8",
+        ),
     }
+    options = (
+        (
+            "max-line-length-suggestions",
+            {
+                "default": 88,
```
Pierre-Sassoulas, line 63: I think the default should be ``max-line-length`` from the format checker. This is the way to make pylint compatible with black formatting (the line length can be changed in black).

pylint/extensions/code_style.py
```diff
@@ -135,6 +174,131 @@ def _check_dict_consider_namedtuple_dataclass(self, node: nodes.Dict) -> None:
...
+    @staticmethod
+    def _check_ignore_assignment_expr_suggestion(
+        node: nodes.If, name: Optional[str]
+    ) -> bool:
+        """Return True if suggestion for assignment expr should be ignore.
+
+        E.g., in cases where a match statement would be a better fit
+        (multiple conditions).
+        """
+        if isinstance(node.test, nodes.Compare):
```
Pierre-Sassoulas, line 279:
```suggestion
        if not isinstance(node.test, nodes.Compare):
            return False
        ... 
```
Doing a premature return would decrease decrease indentation in the majority if the function.

pylint/extensions/typing.py
```diff
@@ -135,49 +122,38 @@ class TypingChecker(BaseChecker):
...
         self._consider_using_alias_msgs: List[DeprecatedTypingAliasMsg] = []
 
-    @lru_cache()
-    def _py37_plus(self) -> bool:
-        return self.config.py_version >= (3, 7)
-
-    @lru_cache()
-    def _py39_plus(self) -> bool:
-        return self.config.py_version >= (3, 9)
+    def open(self) -> None:
```
Pierre-Sassoulas, line 141: I like this change, we did not need lru cache for that. It could even be faster because we're not making a function call ? Anyway, that's great.

setup.cfg
```diff
@@ -49,6 +49,7 @@ install_requires =
     mccabe>=0.6,<0.7
     toml>=0.7.1
     colorama;sys_platform=="win32"
+    typing-extensions>=3.10.0;python_version<"3.10"
```
Pierre-Sassoulas, line 52: 👍

tests/functional/ext/code_style/code_style_consider_using_assignment_expr.py
```diff
@@ -0,0 +1,148 @@
...
+if d2 is not None:  # [consider-using-assignment-expr]
+    ...
+
+d3 = 2
+if d3 == 2:  # [consider-using-assignment-expr]
+    ...
+
+
+# -----
+# Don't emit warning if match statement would be a better fit
```
Pierre-Sassoulas, line 74: But we're not emitting warning for match statement yet 😄 Is that a glimpse of your future plan ? 😛

tests/functional/ext/code_style/code_style_consider_using_assignment_expr.txt
```diff
@@ -0,0 +1,18 @@
+consider-using-assignment-expr:4:3::"Use 'if (a1 := 2):' instead":HIGH
```
Pierre-Sassoulas, line 1: The message is nice, I like when the expected code is directly given in the message, very beginner friendly.

Pierre-Sassoulas (changes requested):
I only have minor comment, great work !

pylint/extensions/code_style.py
cdce8p: True, however not doing so allow for easier extension in the future. Furthermore, I don't think having a bit more indentations is that bad here as it's still fairly readable.

pylint/extensions/typing.py
cdce8p: Don't really know if it's faster. Probably not by much I imagine.
I didn't know about `open` when I first wrote the cached functions.

setup.cfg
cdce8p: Necessary to be able to import `TypeGuard`

tests/functional/ext/code_style/code_style_consider_using_assignment_expr.py
cdce8p: It might be 😄
But tbh I like it a bit better without `:=` in this case. There is a certain pattern in it that makes it beautiful (and better readable).
```py
o1 = 2
if o1 == 1:
    ...
elif o1 == 2:
    ...
elif o1 == 3:
    ...
```

tests/functional/ext/code_style/code_style_consider_using_assignment_expr.txt
cdce8p: I agree. Just a bit unfortunate that black will remove the brackets in this example since they are technically redundant -> `if a1 := 2:`. Anyway, I still think it's the right choice to include them in the suggestion as it's a lot safer, especially for beginners.

Pierre-Sassoulas (approved):

nickdrozd:
@cdce8p @Pierre-Sassoulas This check is enabled for the Pylint codebase, right? Shouldn't it have flagged a bunch of stuff? I was hoping to see what would turn up, but it looks like it isn't running.

Pierre-Sassoulas:
It's because we're supporting Python 3.6 :)

PR #4887, opened 2021-08-21
Migrate from ``appdirs`` to ``platformdirs``

### LINKED ISSUE #4886: Migrate from ``appdirs`` to ``platformdirs``
### Current problem

> appdirs is effectively unmaintained; platformdirs is the community's forked replacement. Black has already switched, and given this dep was newly added, it would make sense to use the modern, maintained replacement now given its drop-in compatible and fixes numerous long-outstanding bugs with the original, instead of having to switch later if and when things eventually break.

### Desired solution

Migration to platformdirs.

### Additional context

https://github.com/PyCQA/pylint/pull/4661#issuecomment-903009372

Pierre-Sassoulas:
## Type of Changes

## Description

Black already did the migration, the import time also got faster see https://github.com/psf/black/pull/2375. A worthy import time goal for astroid 😄 

Closes #4886

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    appdirs>=1.4.0
+    platformdirs>=2.0.0
     astroid>=2.7.2,<2.8 # (You should also upgrade requirements_test_min.txt)
     isort>=4.2.5,<6
     mccabe>=0.6,<0.7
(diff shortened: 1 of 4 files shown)


cdce8p (approved):
👍🏻

setup.cfg
```diff
@@ -75,15 +75,15 @@ markers =
...
+known_third_party = platformdirs, astroid, sphinx, isort, pytest, mccabe, six, toml
 include_trailing_comma = True
 skip_glob = tests/functional/**,tests/input/**,tests/extensions/data/**,tests/regrtest_data/**,tests/data/**,astroid/**,venv/**
 src_paths = pylint
 
 [mypy]
 scripts_are_modules = True
 
-[mypy-appdirs]
+[mypy-platformdirs]
```
CAM-Gerlach, line 86: @Pierre-Sassoulas FYI, you should be able to remove this, since another benefit of `platformdirs` is it is fully statically typed.

cdce8p: @CAM-Gerlach Thanks for the heads up! We can indeed remove it (#4889), although we need to bump the required version first (#4890).

PR #4889, opened 2021-08-21
Updated mypy checks

cdce8p:
## Type of Changes

## Description
https://github.com/PyCQA/pylint/pull/4887#discussion_r693329085 mentioned that `platformdirs` is now fully typed.

To be able to remove the import ignore from `setup.cfg`, we would need to add `platformdirs` as an additional dependency in `.pre-commit-config.yaml`. Long term it might be better to just run mypy in the local environment instead.

If the PR is accepted, I'll update `astroid` as well.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,8 +72,8 @@ jobs:
           . venv/bin/activate
           pre-commit install --install-hooks
 
-  formatting:
-    name: Run pre-commit checks
+  pylint:
+    name: Run pylint check
     runs-on: ubuntu-latest
     needs: prepare-base
     steps:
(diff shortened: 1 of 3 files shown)


cdce8p:
I've rebased the PR onto #4890

cdce8p:
Seems like pre-commit.ci needs to be updated as well.

Pierre-Sassoulas:
I like the isolation that pre-commit does.  This permit to not have to handle the mypy environnement ourselves 'locally or in github action) and will reduce problem as pre-commit configuration is the source of truth and there would be no human intervention. We can add something in ``additional_dependencies: ["types-pkg_resources==0.1.3", "types-toml==0.1.3"]`` or we could even use the mypy option to auto-install missing stubs.

cdce8p:
> I like the isolation that pre-commit does. This permit to not have to handle the mypy environnement ourselves 'locally or in github action) and will reduce problem as pre-commit configuration is the source of truth and there would be no human intervention.

We already have all additional typing dependencies in `requirement_test.txt` as they are necessary for local runs.

> We can add something in `additional_dependencies: ["types-pkg_resources==0.1.3", "types-toml==0.1.3"]` or we could even use the mypy option to auto-install missing stubs.

That's the way we do it at the moment. One more won't make a difference, but I think in the future we might want to reuse the local environment as it's easier to deal with.

cdce8p:
@Pierre-Sassoulas I updated the PR to only add `platformdirs` as additional dependency. Might be best to squash merge this one.

Pierre-Sassoulas (approved):

PR #4894, opened 2021-08-21
Remove useless minver configuration in functional tests

Pierre-Sassoulas:
## Type of Changes

## Description

This remove the unneeded min_ver configuration in functional tests and rename tests when required. We're only supporting python version > 3.6 so this became useless configuration.


tests/functional/s/singledispatch_functions_py3.rc
```diff
@@ -1,3 +0,0 @@
-[testoptions]
-;https://bugs.python.org/issue10445
```
cdce8p, line 2: Can you add this as a comment in the test file?

tests/functional/a/abstract/abstract_class_instantiated.py
```diff
@@ -126,3 +126,18 @@ def test(self):
...
+    defines abstract methods.
+    """
+    @abc.abstractmethod
+    def test(self):
+        pass
+
+
+def main_two():
+    """ do nothing """
+    BadClass() # [abstract-class-instantiated]
```
cdce8p, line 143:
```suggestion
    BadClassTwo()  # [abstract-class-instantiated]
```
?

cdce8p (approved):
Great cleanup!

tests/functional/s/singledispatch_functions_py3.rc
Pierre-Sassoulas: The content of the link is "python 3.2 and 3.3 receive only security update, we can't backport the fix that is in 3.4 in it", so we can remove the link too.

tests/functional/a/abstract/abstract_class_instantiated.py
Pierre-Sassoulas: Nice catch, 👍

PR #4897, opened 2021-08-22
Add a benchmark github action to have a graph

Pierre-Sassoulas:
## Type of Changes

## Description

Draft for integrated benchmark in CI with a graph that can easily be checked. Following example here: https://github.com/rhysd/github-action-benchmark/tree/master/examples/pytest


Pierre-Sassoulas:
Closing in favor of launching the benchmark twice, once on main, once on the branch of the MR like it would be done for false positive primers in #5364 .

PR #4924, opened 2021-08-27
Emit a correct invalid-name message when using multi-naming style.

yilei:
## Type of Changes

## Description

This PR fixes the `invalid-name` error messages when using multi-naming style.

Previously, given `--function-rgx=(?:(?P<UP>[A-Z]+)|(?P<down>[a-z]+))$` and the code:

```
def FOO():
    pass
def UPPER():
    pass
def lower():
    pass
```

It would emit a message: ``Function name `lower` doesn't conform to '(?:(?P<UP>[A-Z]+)|(?P<down>[a-z]+))$' pattern [invalid-name]``.

The message is misleading as `lower` *does* conform to `(?:(?P<UP>[A-Z]+)|(?P<down>[a-z]+))$`. It's just not the prevalent group "UP".

After this commit, the message becomes: ``Function name `lower` doesn't conform to UP group in the '(?:(?P<UP>[A-Z]+)|(?P<down>[a-z]+))$' pattern [invalid-name]``


Pierre-Sassoulas (changes requested):
Thank you for this change this would make the message clearer indeed. Could you add typing on modified functions, please ? We're trying to add it progressively. Also, it seems that there was nothing to test this before but could you add functional tests for it ? In ``tests/functional/i/invalid_name_rmultinaming_style.py`` by adding a .rc file with multi naming regex in it for example :)

yilei:
I added type annotations and a function test, please take another look, thanks @Pierre-Sassoulas

ChangeLog
```diff
@@ -43,6 +43,7 @@ Release date: TBA
 ..
   Put bug fixes that should not wait for a new minor version here
 
+* Fix ``invalid-name`` error messages when using multiple naming style regexes.
```
Pierre-Sassoulas, line 47:
```suggestion
* The ``invalid-name`` message is now more detailed when using multiple naming style regexes.

```
Could you also add that in docs/whatsnew/2.11, please ?

Pierre-Sassoulas (changes requested):
Thank you for the new tests, I think it's perfect. I forgot about the need to add a little message in whatsnew, this will be released in 2.11 :)

yilei:
Done.

pylint/checkers/base.py
```diff
@@ -2016,10 +2017,21 @@ def _find_name_group(self, node_type):
...
+        warning: str = "invalid-name",
+        prevalent_group: Optional[str] = None,
+    ) -> None:
         type_label = HUMAN_READABLE_TYPES[node_type]
         hint = self._name_hints[node_type]
+        if prevalent_group:
+            # This happens in the multi naming match case. The expected
+            # prevalent group needs to be spelled out to make the message
+            # correct.
+            hint = "the `%s` group in the " % prevalent_group + hint
```
Pierre-Sassoulas, line 2034:
```suggestion
            hint = f"the `{prevalent_group}{hint}` group in the "
```

yilei:
The f-string change is incorrect, I pushed a new commit to fix.

PR #4929, opened 2021-08-28
Implement basic control flow checks for ``consider-using-with``

### LINKED ISSUE #4751: False positive ``consider-using-with`` in ``if``/``else`` block
### Bug description

Given a file `a.py`:

```python
# pylint: disable=missing-docstring, invalid-name
a = True
if a:
    file_handle = open("foo")
else:
    file_handle = open("bar")
with file_handle:
    print(file_handle.read())
```

### (Configuration)

_No response_

### Command used

Result of `pylint a.py`:
```
************* Module a
a.py:4:18: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)

Your code has been rated at 8.33/10 (previous run: 8.33/10, +0.00)
```

### (Expected behavior)

No message ``consider-using-with``, as ``file_handle`` is used in a subsequent ``with`` block.
#4721 implemented changes that already suppress this message for things like:

```python
file_handle = open("foo")
with file_handle:
    print(file_handle.read())
```

So naturally the example code above should also not trigger the message.

### (Version affected)

```shell
pylint 2.9.5
astroid 2.6.5
Python 3.9.6 (default, Jun 28 2021, 19:24:41) 
[Clang 12.0.5 (clang-1205.0.22.9)]
```

### (OS / Environment)

macOS BigSur 11.4

### (Additional dependencies)

_No response_

DudeNr33:

## Type of Changes

## Description
The checker keeps track of context managers that have been assigned to a variable to see if they are later on used in a ``with`` block. 
If a variable is reassigned before it is used in a ``with``, the message will trigger (as intended).
However, the logic to check that was pretty basic, and did not handle cases where e.g. the variable is assigned inside two separate branches of control flow, like in if-else statements or try-except constructs.

This MR solves this. ~~As the logic to determine if two nodes are in different branches of control flow might be useful in other places as well, I put the method inside the ``pylint.checkers.utils`` module.~~

Closes #4751


DudeNr33:
Oh wow, I just noticed that ``astroid`` already has a very similar method ``are_exclusive``. 
It covers not exactly the same cases (especially as it is focused on statements instead of notes, therefore it does not handle ``IfExp``), but duplicating the logic inside ``pylint`` is of course nonsense. 
I will refactor.

Pierre-Sassoulas (approved):
Thank you for recreating an astroid function then refactoring everything to use astroid :D The final result is pretty clean.

PR #4937, opened 2021-08-31
Fix false positive ``protected-access`` in typing 

DanielNoord:

## Type of Changes

## Description

Fix false positive protected-access in typing
Class functions can return protected members which can then be passed as arguments to other functions.
When using type hints in function definitions these raised a message which they shouldn't do.

I did not create an issue for this so no issue is assigned, if needed I can still do this

I also saw some discrepancies between the `Changelog` and the `2.10` & `2.11` files. I fixed these and added the customary styling of the `whatsnew` documents to the `2.11`. I thought this might be helpful, but let me know if I should remove or change that commit!

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -39,6 +39,8 @@ Release date: TBA
 
   Closes #3802
 
+* Fix false positive for ``protected-access`` if a protected member is used in type hints of function definitions
+
 
 What's New in Pylint 2.10.3?
 ============================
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (approved):
Nice fix ! Also thank you for fixing the chanelog, this is something I usually do at release time using `git diff vX.Y.Z-1 ChangeLog doc/whatsnew/` but not having to do it is always nice :D

PR #4943, opened 2021-08-31
Remove dead visit_exec method

cdce8p:
## Type of Changes

## Description
In Python 3 `exec` is just a normal `Call` node. `exec-used` is still checked.

/CC: @DanielNoord


Pierre-Sassoulas (approved):
Nice catch !

PR #4948, opened 2021-09-01
Fix false positive ``superfluous-parens`` for tuples

### LINKED ISSUE #4907: False positive superfluous-parens
### Bug description

```python
tuple(x for x in ((a, str(a)) for a in ()))
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:1: [C0325(superfluous-parens), ] Unnecessary parens after 'in' keyword
```

### Expected behavior

No errors.

### Pylint version

```shell
pylint 2.10.2
astroid 2.7.2
[GCC 9.4.0]
```

### OS / Environment

Ubuntu 20.04.3 LTS

DanielNoord:

## Type of Changes

## Description

Tuples can be created with inner tuples. This creates double parenthesis which we flagged incorrectly.
This closes #4907

### DIFF
--- a/tests/checkers/unittest_format.py
+++ b/tests/checkers/unittest_format.py
@@ -165,11 +165,6 @@ class TestSuperfluousParentheses(CheckerTestCase):
             (Message("superfluous-parens", line=1, args="if"), "if (foo):", 0),
             (Message("superfluous-parens", line=1, args="if"), "if ((foo, bar)):", 0),
             (Message("superfluous-parens", line=1, args="if"), "if (foo(bar)):", 0),
-            (
-                Message("superfluous-parens", line=1, args="return"),
-                "return ((x for x in x))",
-                0,
-            ),
             (Message("superfluous-parens", line=1, args="not"), "not (foo)", 0),
             (Message("superfluous-parens", line=1, args="not"), "if not (foo):", 1),
             (Message("superfluous-parens", line=1, args="if"), "if (not (foo)):", 0),
--- a/tests/functional/s/super/superfluous_parens.py
+++ b/tests/functional/s/super/superfluous_parens.py
@@ -1,6 +1,7 @@
 """Test the superfluous-parens warning."""
-from __future__ import print_function
 # pylint: disable=unneeded-not, unnecessary-comprehension, missing-function-docstring, invalid-name, fixme
+# pylint: disable=import-error, missing-class-docstring, too-few-public-methods
+import numpy as np
 A = 3
 if (A == 5):  # [superfluous-parens]
     pass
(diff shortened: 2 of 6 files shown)


tests/functional/s/super/superfluous_parens.py
```diff
@@ -20,8 +20,8 @@
 
 B = [x for x in ((3, 4))]  # [superfluous-parens]
 C = [x for x in ((3, 4) if 1 > 0 else (5, 6))]
-D = [x for x in ((3, 4) if 1 > 0 else ((5, 6)))]  # [superfluous-parens]
-E = [x for x in ((3, 4) if 1 > 0 else ((((5, 6)))))]  # [superfluous-parens]
+D = [x for x in ((3, 4) if 1 > 0 else ((5, 6)))]
+E = [x for x in ((3, 4) if 1 > 0 else ((((5, 6)))))]
```
Pierre-Sassoulas, line 24: Could we also add the example from the issue ?

DanielNoord: Seems I was to quick to push this. I hoped I could leave some checks for double parenthesis in, but seems like this only creates problems. I'm going to remove them.

Give me a minute!

tests/checkers/unittest_format.py
```diff
@@ -165,11 +165,6 @@ def testCheckKeywordParensHandlesUnnecessaryParens(self):
             (Message("superfluous-parens", line=1, args="if"), "if (foo):", 0),
             (Message("superfluous-parens", line=1, args="if"), "if ((foo, bar)):", 0),
             (Message("superfluous-parens", line=1, args="if"), "if (foo(bar)):", 0),
-            (
```
DanielNoord, line 168: Turns out the unittests also had a false positive.

Pierre-Sassoulas (approved):
Thank you for the lightning fast fix 😉

tests/functional/s/super/superfluous_parens.py
```diff
@@ -18,10 +19,10 @@
 del(DICT['b'])  # [superfluous-parens]
 del DICT['a']
 
-B = [x for x in ((3, 4))]  # [superfluous-parens]
+B = [x for x in ((3, 4))]
 C = [x for x in ((3, 4) if 1 > 0 else (5, 6))]
-D = [x for x in ((3, 4) if 1 > 0 else ((5, 6)))]  # [superfluous-parens]
-E = [x for x in ((3, 4) if 1 > 0 else ((((5, 6)))))]  # [superfluous-parens]
+D = [x for x in ((3, 4) if 1 > 0 else ((5, 6)))]
```
superbobry, line 24: Why are the parens around `(5, 6)` not considered superfluous here?

DanielNoord: See the report in https://github.com/PyCQA/pylint/issues/4907

There is no good way to determine whether the parenthesises are there by accident or if they are intentional.

superbobry: Apologies if I'm missing something obvious, but I think it should be possible if we re-implement the checker to consume the AST instead of tokens. At the AST level we would immediately know that `()` is an empty tuple and thus does require parentheses.

DanielNoord: We already consume the AST so that's not really the issue here. It might be possible to fix this, but in m'n initial attempt I created a lot of false positives so we reverted to not doing this kind of check on nested tuples in comprehension.

I welcome any PR that tries to fix this though!

PR #4949, opened 2021-09-01
fix false positive W0238 with private staticmethod used from classmethod

### LINKED ISSUE #4849: False-positive of unused-private-member with class reference
### Bug description

Consider the following code snippet:
```python
"""p.py"""

class UnusedPrivateMember:
    """unused-private-member (W0238)"""

    @staticmethod
    def __private_method():
        """Is private and does nothing."""

    @classmethod
    def use_private_method(cls):
        """Calls private method."""
        cls.__private_method()
```

### Configuration

_No response_

### Command used

```shell
pylint p.py
```

### Pylint output

```shell
W0238: Unused private member `UnusedPrivateMember.__private_method()` (unused-private-member)
```

### Expected behavior

Calling a private static method only on `cls` (reference to class) in a classmethod produces false positive: `unused-private-member`.

#4782 recently fixed some issues regarding the usage of private members on the class name. This one seems to have slipped through.

Thanks.

### Pylint version

```shell
pylint 2.10.0-dev0
astroid 2.6.6
Python 3.9.6 (default, Jun 30 2021, 10:22:16) 
[GCC 11.1.0]
```
```


### OS / Environment

_No response_

### Additional dependencies

_No response_

### PR DESCRIPTION
**Pull request by yumasheta**
## Type of Changes

## Description

This PR fixes false positive `unusede-private-member` W0238 when using a private staticmethod from a classmethod, ie. `cls.__private_static_method()`.

Also a regression test is added to the functional tests.

Closes #4849

### CODE REVIEW

File: pylint/checkers/classes.py [hunk 1]
```diff
@@ -926,9 +926,12 @@ def _check_unused_private_functions(self, node: nodes.ClassDef) -> None:
                     continue
                 if isinstance(attribute.expr, nodes.Name) and attribute.expr.name in (
                     "self",
+                    "cls",
```
Pierre-Sassoulas, line 929: I think we already fixed something really similar in another MR, we have a duplication between this and ``child.expr.name in ("self", "cls", node.name)`` line 969/972.

yumasheta: I can still reproduce #4849 against master. I don't quite understand what the lines you reference fixed.

yumasheta, line 929: Another option could be:
`"cls" if attribute.scope().type == "classmethod" else "self",` instead of listing both, "cls" and "self".

Pierre-Sassoulas: A previous false positive for ``unused-private-member``. I was wondering if we need to refactor that check to have a single piece of code to check that a name represent the current class. It's probably a little late now 😄

Pierre-Sassoulas: I think it's clear, let me know what you think about factorizing this check with the othe rone, but this could be merged right now :)

yumasheta: Ok. Yes, it seems like the possibility to access attributes/methods on `cls` representing the class is beeing added by parts now.
I'm still unsure of hardcoding "cls" into the checks in general without further specification. While pylint checks, that the convention with "cls" as the first arg of a classmethod is followed, it's still no guarantee. And if this rule is disabled on a case by case basis then all these "fixes" will fail again.
But I guess that's the same for the "self" argument/word.

Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/4645 found the MR I was talking about.

yumasheta: Ah, I see.
I'm definitely in favor of "a single piece of code to check that a name represent the current class". This would address my above concerns.
But I don't think I'm in the position to judge this  -.-

Pierre-Sassoulas: @yushao2 do you have an opinion on this ?

yushao2: @Pierre-Sassoulas kinda rusty, let me refresh my brain for a bit

yushao2: @Pierre-Sassoulas if it can be refactored now, it would be great (since i foresee that it might be refactored again in the future)


Just my two-cents to try to make code more concise and DRY early on rather than doing it later. Could you give it a shot @yumasheta ?

yumasheta: I'm humbled, but as this is my first contribution to pylint source code (and first ever look at it), I reckon there are better qualified people out there.
Also the earliest I could do would be Okt/Nov.

Pierre-Sassoulas: @yushao2 what do you think about merging this then refactoring in #4965 ? Ideally I'd like to include this in 2.11 that should be released soon.

yushao2: @Pierre-Sassoulas lets do that, i will check on the compatibility once this is merged with my branch

yushao2 (approved):
LGTM, will take a shot at refactoring / open for others to take a crack at it too :)

Pierre-Sassoulas (approved):
Thank you for your contribution :) !

PR #4950, opened 2021-09-01
Add typing with `PyAnnotate` to  `./tests`

DanielNoord:

## Description

I thought I would try and see if I could use `pyannote` to do some typing for `./pytest`.

I have split the changes into commits for each directory, sorry for the size of the PR I think this is about the limit of easy reviewing.

### DIFF
--- a/pylint/lint/utils.py
+++ b/pylint/lint/utils.py
@@ -5,7 +5,7 @@ import contextlib
 import sys
 import traceback
 from datetime import datetime
-from pathlib import Path
+from pathlib import Path, PosixPath
 from typing import Union
 
 from pylint.config import PYLINT_HOME
@@ -17,7 +17,7 @@ class ArgumentPreprocessingError(Exception):
 
 
 def prepare_crash_report(
-    ex: Exception, filepath: str, crash_file_path: Union[Path, str]
+    ex: Exception, filepath: PosixPath, crash_file_path: Union[Path, str]
 ) -> Path:
     issue_template_path = (
         Path(PYLINT_HOME) / datetime.now().strftime(str(crash_file_path))
@@ -63,7 +63,7 @@ pylint crashed with a ``{ex.__class__.__name__}`` and with the following stacktr
     return issue_template_path
 
 
-def get_fatal_error_message(filepath: str, issue_template_path: Path) -> str:
+def get_fatal_error_message(filepath: str, issue_template_path: str) -> str:
     return (
         f"Fatal error while checking '{filepath}'. "
         f"Please open an issue in our bug tracker so we address this. "
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -58,10 +58,10 @@ Try updating it with: 'python tests/test_functional.py {UPDATE_OPTION}'"""
 class OutputLine(NamedTuple):
     symbol: str
     lineno: int
-    column: int
+    column: str
     object: Any
     msg: str
-    confidence: str
+    confidence: interfaces.Confidence
 
     @classmethod
     def from_msg(cls, msg):
--- a/tests/extensions/test_check_docs.py
+++ b/tests/extensions/test_check_docs.py
@@ -26,7 +26,7 @@ in particular the parameter documentation checker `DocstringChecker`
 
 import astroid
 import pytest
-from astroid.nodes import ClassDef
+from astroid import nodes
 
 from pylint.extensions.docparams import DocstringParameterChecker
 from pylint.testutils import CheckerTestCase, Message, set_config
@@ -343,7 +343,7 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def _visit_methods_of_class(self, node: ClassDef) -> None:
+    def _visit_methods_of_class(self, node: nodes.ClassDef) -> None:
         """Visit all methods of a class node
 
         :param node: class node
--- a/tests/extensions/test_check_docs.py
+++ b/tests/extensions/test_check_docs.py
@@ -27,6 +27,7 @@ in particular the parameter documentation checker `DocstringChecker`
 import astroid
 import pytest
 from astroid import nodes
+from astroid.nodes.scoped_nodes import ClassDef
 
 from pylint.extensions.docparams import DocstringParameterChecker
 from pylint.testutils import CheckerTestCase, Message, set_config
@@ -42,7 +43,7 @@ class TestParamDocChecker(CheckerTestCase):
         "docstring_min_length": -1,
     }
 
-    def test_missing_func_params_in_sphinx_docstring(self):
+    def test_missing_func_params_in_sphinx_docstring(self) -> None:
         """Example of a function with missing Sphinx parameter documentation in
         the docstring
         """
@@ -64,7 +65,7 @@ class TestParamDocChecker(CheckerTestCase):
         ):
             self.checker.visit_functiondef(node)
 
-    def test_missing_func_params_in_google_docstring(self):
+    def test_missing_func_params_in_google_docstring(self) -> None:
         """Example of a function with missing Google style parameter
         documentation in the docstring
         """
@@ -88,7 +89,7 @@ class TestParamDocChecker(CheckerTestCase):
         ):
             self.checker.visit_functiondef(node)
 
-    def test_missing_type_doc_google_docstring_exempt_kwonly_args(self):
+    def test_missing_type_doc_google_docstring_exempt_kwonly_args(self) -> None:
         node = astroid.extract_node(
             """
         def identifier_kwarg_method(arg1: int, arg2: int, *, value1: str, value2: str):
@@ -106,7 +107,7 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def test_missing_func_params_with_annotations_in_google_docstring(self):
+    def test_missing_func_params_with_annotations_in_google_docstring(self) -> None:
         """Example of a function with missing Google style parameter
         documentation in the docstring.
         """
@@ -128,7 +129,7 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def test_default_arg_with_annotations_in_google_docstring(self):
+    def test_default_arg_with_annotations_in_google_docstring(self) -> None:
         """Example of a function with missing Google style parameter
         documentation in the docstring.
         """
@@ -150,7 +151,9 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def test_missing_func_params_with_partial_annotations_in_google_docstring(self):
+    def test_missing_func_params_with_partial_annotations_in_google_docstring(
+        self,
+    ) -> None:
         """Example of a function with missing Google style parameter
         documentation in the docstring.
         """
@@ -174,7 +177,7 @@ class TestParamDocChecker(CheckerTestCase):
         ):
             self.checker.visit_functiondef(node)
 
-    def test_non_builtin_annotations_in_google_docstring(self):
+    def test_non_builtin_annotations_in_google_docstring(self) -> None:
         """Example of a function with missing Google style parameter
         documentation in the docstring.
         """
@@ -192,7 +195,7 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def test_non_builtin_annotations_for_returntype_in_google_docstring(self):
+    def test_non_builtin_annotations_for_returntype_in_google_docstring(self) -> None:
         """Example of a function with missing Google style parameter
         documentation in the docstring.
         """
@@ -210,7 +213,7 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def test_func_params_and_keyword_params_in_google_docstring(self):
+    def test_func_params_and_keyword_params_in_google_docstring(self) -> None:
         """Example of a function with Google style parameter splitted
         in Args and Keyword Args in the docstring
         """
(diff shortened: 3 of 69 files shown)


tests/test_func.py
```diff
@@ -77,16 +79,18 @@ def _test(self, tocheck):
         except Exception as ex:
             # need finalization to restore a correct state
             self.linter.reporter.finalize()
-            ex.file = tocheck
+            ex.file = tocheck  # type: ignore
```
DanielNoord, line 82: Does this part of the code actually get used? I couldn't get it to trigger when running the tests with a debugger.

If not, can we remove it?
If yes, what is the exact type of `ex` and can we do something to avoid these two `type: ignore`'s?

tests/pyreverse/test_diadefs.py
```diff
@@ -22,25 +22,30 @@
 # pylint: disable=redefined-outer-name
 import sys
 from pathlib import Path
+from typing import Callable, Dict, List, Tuple
 
 import pytest
 from astroid import nodes
+from conftest import PyreverseConfig  # type: ignore #pylint: disable=no-name-in-module
```
Pierre-Sassoulas, line 29: I think we can remove this one and type it properly later.

DanielNoord, line 29: I couldn't get this import to pass for `mypy`.

As this is a test the import from `conftest` works but pylint throws an error.

tests/pyreverse/test_writer.py
```diff
@@ -24,12 +24,14 @@
 import codecs
 import os
 from difflib import unified_diff
+from typing import Callable, Iterator, List
 from unittest.mock import Mock
 
 import pytest
+from conftest import PyreverseConfig  # type: ignore
```
DanielNoord, line 31: Same as above...

tests/test_pragma_parser.py
```diff
@@ -8,93 +8,105 @@
...
+def test_simple_pragma() -> None:
     comment = "#pylint: disable = missing-docstring"
     match = OPTION_PO.search(comment)
-    for pragma_repr in parse_pragma(match.group(2)):
-        assert pragma_repr.action == "disable"
-        assert pragma_repr.messages == ["missing-docstring"]
+    if match:
+        for pragma_repr in parse_pragma(match.group(2)):
+            assert pragma_repr.action == "disable"
+            assert pragma_repr.messages == ["missing-docstring"]
```
Pierre-Sassoulas, line 17:
```suggestion
    assert match
    for pragma_repr in parse_pragma(match.group(2)):
        assert pragma_repr.action == "disable"
        assert pragma_repr.messages == ["missing-docstring"]
```
Or something similar because we want the test to actually fail if there is no match.

DanielNoord:
For some reason `mypy_extensions` is not available on the windows test suite? Is this a known issue?

Pierre-Sassoulas (changes requested):
This looks really good, thank you. There's only this issue with the match change that needs to be addressed.

tests/test_pragma_parser.py
DanielNoord: Give me a minute!

tests/pyreverse/test_diadefs.py
DanielNoord: I would be in favour of letting this in.
We will need to write these lines at some point anyway and the disabled messages serve as a reminder that something needs to be fixed. Removing the typing makes it less obvious that there is still something on the "to-do list".

But obviously, this is up to you if you still want me to remove it from the typing let me know!

Pierre-Sassoulas: No I think you're right. I was wondering why the class definition is in conftest which is probably not something we should explore in this MR.

tests/test_func.py
Pierre-Sassoulas: I think this is only to get an idea of what is failing when your modifying pylint and failing at it. We can sanitize the code:

```python
        except Exception as ex:
            msg = f"Exception: {ex} ({exception_str} while checking {to_check}))
            raise Exception(msg) from ex
        finally:
            self.linter.reporter.finalize()
```

tests/test_func.py
```diff
@@ -77,8 +72,7 @@ def _test(self, tocheck: List[str]) -> None:
         try:
             self.linter.check(tocheck)
         except Exception as ex:
-            msg = f"Exception: {ex} ({exception_str} while checking {tocheck}))"
-            raise Exception(msg) from ex
+            raise Exception from ex
```
DanielNoord, line 75: @Pierre-Sassoulas Is this what you meant?

Pierre-Sassoulas: I'm thinking of:
```diff
diff --git a/tests/test_func.py b/tests/test_func.py
index b57dc7d1..ba13844f 100644
--- a/tests/test_func.py
+++ b/tests/test_func.py
@@ -36,10 +36,6 @@ FILTER_RGX = None
 INFO_TEST_RGX = re.compile(r"^func_i\d\d\d\d$")
 
 
-def exception_str(self, ex):  # pylint: disable=unused-argument
-    """function used to replace default __str__ method of exception instances"""
-    return f"in {ex.file}\n:: {', '.join(ex.args)}"
-
 
 class LintTestUsingModule:
     INPUT_DIR = None
@@ -75,12 +71,11 @@ class LintTestUsingModule:
         try:
             self.linter.check(tocheck)
         except Exception as ex:
+            msg = f"Exception: {ex} in {tocheck}\n:: {', '.join(ex.args)})"
+            raise Exception(msg) from ex
+        finally:
             # need finalization to restore a correct state
             self.linter.reporter.finalize()
-            ex.file = tocheck
-            print(ex)
-            ex.__str__ = exception_str
-            raise
         self._check_result(self.linter.reporter.finalize())
 
     def _has_output(self):
```

DanielNoord: I assumed the ) add the end of the f-string was unintentional, please correct if I'm wrong. Going through tests now.

Pierre-Sassoulas: Yes you're probably right. This is just a piece of code helpful for pylint's contributor during development it does not need to be perfect :)

DanielNoord: Does fail a lot of test though 😢

Pierre-Sassoulas: Is this change making the test fail ? We can just ignore the type error and not change anything if that's the case.

DanielNoord: Yeah, see test history of latest three commits. However, I think it is due to some formatting changes which can be auto solved by pytest. I'm on mobile now so I can't actually run pytest.

Pierre-Sassoulas: Ok I'll try something then.

Pierre-Sassoulas: Hurgh, just realized that those are the legacy functional tests, we're trying to get rid of this code. Let's not do anything at all, ignore all the error and not add typing if it's causing trouble.  I think it would probably be faster to finish the migration to the new functional tests framework than fix this. See :
```python
    assert len(tests) < 13, "Please do not add new test cases here." + "\n".join(
        str(k) for k in tests if not k[2]
    )
```
There's not much left of them too.

DanielNoord: Done!

pylint/lint/utils.py
```diff
@@ -63,7 +63,7 @@ def prepare_crash_report(
     return issue_template_path
 
 
-def get_fatal_error_message(filepath: str, issue_template_path: Path) -> str:
```
Pierre-Sassoulas, line 66: Most of the time the type for path should be ``Union[Path, str]``, I don't know what is the difference with ``PosixPath`` which seems to be the default used by pyannotate.

pylint/testutils/output_line.py
```diff
@@ -58,10 +58,10 @@ def __init__(self, row, exception):
 class OutputLine(NamedTuple):
     symbol: str
     lineno: int
-    column: int
+    column: str
```
Pierre-Sassoulas, line 61:
```suggestion
    column: Union[str, int]
```

tests/test_check_parallel.py
```diff
@@ -56,12 +58,12 @@ class SequentialTestChecker(BaseChecker):
         )
     }
 
-    def __init__(self, linter, *args, **kwargs):
-        super().__init__(linter, *args, **kwargs)
-        self.data = []
+    def __init__(self, linter: PyLinter) -> None:
```
Pierre-Sassoulas, line 61: 👍

tests/test_func.py
```diff
@@ -36,57 +37,60 @@
...
         else:
             self.linter.disable("I")
         try:
             self.linter.check(tocheck)
         except Exception as ex:
-            # need finalization to restore a correct state
-            self.linter.reporter.finalize()
-            ex.file = tocheck
+            print(f"Exception: {ex} in {tocheck}:: {'‚ '.join(ex.args)}")
+            ex.file = tocheck  # type: ignore # This is legacy code we're trying to remove, impossible to type correctly
```
Pierre-Sassoulas, line 82:
```suggestion
            ex.file = tocheck  # type: ignore # This is legacy code we're trying to remove, not worth it to type correctly
```

tests/test_func.py
```diff
@@ -36,57 +37,60 @@
...
+            ex.__str__ = exception_str  # type: ignore # This is legacy code we're trying to remove, impossible to type correctly
             raise
         self._check_result(self.linter.reporter.finalize())
 
-    def _has_output(self):
-        return not self.module.startswith("func_noerror_")
+    def _has_output(self) -> bool:
+        if self.module and not self.module.startswith("func_noerror_"):
+            return True
+        return False
```
Pierre-Sassoulas, line 91:
```suggestion
        return self.module and not self.module.startswith("func_noerror_")
```

tests/test_self.py
```diff
@@ -96,42 +100,53 @@ def _configure_lc_ctype(lc_ctype):
 
 
 class MultiReporter(BaseReporter):
-    def __init__(self, reporters):
+    def __init__(
+        self,
+        reporters: List[
+            Union[JSONReporter, ColorizedTextReporter, TextReporter, BaseReporter]
```
Pierre-Sassoulas, line 106:
```suggestion
            Reporter
```
(Not sure about that but the base class would be clearer imo.)

Pierre-Sassoulas (commented):
Thank you, this is going to help tremendously for putting mypy in place. And it's giving me idea for astroid :)

pylint/lint/utils.py
DanielNoord: PyAnnotate bases this on the test suite so it seems only `str` get passed to this function in all of the test suite. I could add the `Union` but perhaps it is better to only allow `str`'s? The test suite is passing and `mypy` will now throw errors when passing anything else than `str`?

pylint/testutils/output_line.py
DanielNoord: Shouldn't we try and refactor this to be only one of the two? Seems weird to allow both, but perhaps I am missing something obvious here!

tests/test_func.py
DanielNoord: I did this because `mypy` fails to understand that because of `self.module` at the beginning of L89 `self.module` can't be None. I can add a `type: ignore` but thought it would be more elegant to show that the return value will always be `bool`.

DanielNoord:
If I can find the time I would be willing to start this up for `astroid` as well. Although I would first finish doing all of `PyAnnotate` for the `pylint` codebase!

tests/test_func.py
Pierre-Sassoulas: Maybe ``bool(self.module)`` then ?

DanielNoord: Is used `isinstance()`. Going through tests now!

DanielNoord: This works!

tests/extensions/test_check_docs.py
```diff
@@ -26,7 +26,7 @@
 
 import astroid
 import pytest
-from astroid import nodes
+from astroid.nodes import ClassDef
```
DanielNoord, line 29:
```suggestion
from astroid import nodes
```

tests/extensions/test_check_docs.py
```diff
@@ -341,7 +343,7 @@ def function_foo(x, y):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def _visit_methods_of_class(self, node: nodes.ClassDef) -> None:
+    def _visit_methods_of_class(self, node: ClassDef) -> None:
```
DanielNoord, line 346:
```suggestion
    def _visit_methods_of_class(self, node: nodes.ClassDef) -> None:
```

tests/pyreverse/test_diadefs.py
DudeNr33: ``mypy`` cannot deal with modules that have the same name and are inside a "non-package" directory tree (i.e. no ``__init__.py``). I already had to exclude every ``conftest.py`` except the top-level one in the pre-commit configuration for ``mypy`` because of this.
A solution to this would be to add ``__init__.py`` files to the tests folders and turn them into packages.
But this is [warned against in the pytest docs](https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-outside-application-code) for repositories that use a non-src layout in combination with ``tox``, which is exactly the case for pylint.

DanielNoord: So this is the best way to do this? Or would you suggest another solution?

DudeNr33: The only workaround I could think of is extracting the ``PyreverseConfig`` class in a separate module, from which both the ``conftest.py`` and ``test_diadefs.py`` import it. But as this is "just" test code and not production code, I don't think it is worth it. 
So I would say leave it as it is. 
(Funnily enough, the import is only necessary because ``PyreverseConfig`` is itself used as a type hint 😁)

Pierre-Sassoulas: > (Funnily enough, the import is only necessary because PyreverseConfig is itself used as a type hint 😁)

We could use ``if typing.TYPE_CHECKING`` for that. But the issue with conftest.py and mypy is between pytest design and mypy's design and we can't help it. I agree with @DudeNr33 it's not worth a refactor.

pylint/testutils/output_line.py
DanielNoord: Do you want me to commit this suggestion? Or keep as is?

Pierre-Sassoulas:
> If I can find the time I would be willing to start this up for astroid as well. Although I would first finish doing all of PyAnnotate for the pylint codebase!

I started something for astroid, https://github.com/PyCQA/astroid/pull/1163, but there will be a lot of work before being able to add mypy. pyannotate had some bug I had to manually fix the configuration json. We could discuss this in detail on the astroid's MR.

pylint/testutils/output_line.py
Pierre-Sassoulas: Well no you were right we're reading from the functional result file so it's always a string even if it should be castable to int.

DanielNoord:
Going to look at https://github.com/PyCQA/astroid/pull/1163 tomorrow.

Is there anything needed for this PR to be merged?

Pierre-Sassoulas (approved):

PR #4940, opened 2021-08-31
Refactor various typing related issues

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

This PR has two related aims:
1) Add typing to all `visit_...` and `leave_...` calls in the codebase
2) Fix some outstanding mypy issues
As a result of doing the former the number of mypy issues increased so it seemed logical to take another look at some of the open issues (based on comments meant to ignore mypy). I did not add a changelog contribution.

Before merging there are a small number of calls for which I did not find documentation and therefore I could not add the type annotations. I think it would be best to add those before merging this, to make sure this PR catches all calls.
The missing annotations are for:
~`visit_exec`~
~`visit_default`~
`visit_print`
`visit_repr`
`visit_package`
`visit_instance`
~`visit_project`~
Does any of the maintainers know the nodes passed into these calls? And if so, can you give me a list or add them yourself?

PR #4952, opened 2021-09-01
Remove unused ``visit_package`` function

DanielNoord:

## Description

See:
>I'm not too familiar with the `pyreverse` code, but even if it isn't called anymore, we shouldn't remove it here. That should be done in a separate PR.

_Originally posted by @cdce8p in https://github.com/PyCQA/pylint/pull/4940#discussion_r700316157_

### DIFF
--- a/pylint/pyreverse/inspector.py
+++ b/pylint/pyreverse/inspector.py
@@ -123,16 +123,6 @@ class Linker(IdGeneratorMixIn, utils.LocalsVisitor):
         for module in node.modules:
             self.visit(module)
 
-    def visit_package(self, node):
-        """visit an astroid.Package node
-
-        * optionally tag the node with a unique id
-        """
-        if self.tag:
-            node.uid = self.generate_id()
-        for subelmt in node.values():
-            self.visit(subelmt)
-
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
@DudeNr33 what do you think ?

DudeNr33:
It indeed seems like this method is never executed. 
I removed it locally and ran ``pyreverse`` over the codebase without issues.
When adding some logging to the method and running ``pyreverse`` no output appears.
This is probably because the underlying ``ASTWalker`` class determines the method to call for the visit of each node by doing a ``getattr(handler, f"visit_{kid}")``, where ``kid`` is the lowercase class name of the astroid node.
And since ``astroid`` does not have a node class named "Package", this method is never called.

So I say go for it!

As a side note for a potential future cleanup MR, while looking at the code I noticed that the whole tagging logic of the nodes has become obsolete in the ``Linker`` class, as we now use fully qualified names in the output files.

Pierre-Sassoulas (approved):
I'm trusting the current leading expert in pyreverse on that one :)

Issue #1590, opened 2017-07-24
Add error end column to the output then add the information in all the checkers


purpleP:
Right now it's hard to use pylint as linter integrated into text editor, because it isn't showing where error ends and sometimes doesn't even show where error starts (shows zero column for example in unused import)

PCManticore:
You can customize this by using the ``--msg-template`` parameter. How each editor integration is using it though is not something that pylint should be aware of.

purpleP:
@PCManticore According to docs you only have `column` fiend. Which is the column where error starts. I'm talking about outputting where the error ends.

purpleP:
@PCManticore The issue isn't resolved. Please reopen.

pappasam:
I agree, this issue should be re-opened

Pierre-Sassoulas:
This is a huge one though. By the way I read that linked issue in dense analysis and highlighing till the end of the line should indeed be prefered. This can be the default value until we manage to add the real information in all the checkers.

Pierre-Sassoulas:
Closing as completed because it was done in https://github.com/PyCQA/pylint/issues/5336

PR #4954, opened 2021-09-02
Typing for ``./pylint``, batch 1

DanielNoord:

## Type of Changes

## Description

Similar to #4950 but now for `./pylint`, going directory for directory.
For this PR I would like to do 3 directories:
- [x] `./pylint/utils`
- [x] `./pylint/testutils`
- [x] `./pylint/reporters`

### DIFF
--- a/pylint/reporters/base_reporter.py
+++ b/pylint/reporters/base_reporter.py
@@ -23,7 +23,7 @@ class BaseReporter:
 
     extension = ""
 
-    def __init__(self, output: Optional[TextIO] = None):
+    def __init__(self, output: Optional[TextIO] = None) -> None:
         self.linter: "PyLinter"
         self.section = 0
         self.out: TextIO = output or sys.stdout
@@ -45,7 +45,7 @@ class BaseReporter:
         )
         self.out = output or sys.stdout
 
-    def writeln(self, string=""):
+    def writeln(self, string: str = "") -> None:
         """write a line in the output buffer"""
         print(string, file=self.out)
 
--- a/pylint/testutils/tokenize_str.py
+++ b/pylint/testutils/tokenize_str.py
@@ -3,7 +3,9 @@
 
 import tokenize
 from io import StringIO
+from tokenize import TokenInfo
+from typing import List
 
 
-def _tokenize_str(code):
+def _tokenize_str(code: str) -> List[TokenInfo]:
     return list(tokenize.generate_tokens(StringIO(code).readline))
--- a/pylint/utils/pragma_parser.py
+++ b/pylint/utils/pragma_parser.py
@@ -3,7 +3,7 @@
 
 import re
 from collections import namedtuple
-from typing import Generator, List
+from typing import Generator, List, Optional
 
 # Allow stopping after the first semicolon/hash encountered,
 # so that an option can be continued with the reasons
@@ -52,7 +52,7 @@ TOK_REGEX = "|".join(
 )
 
 
-def emit_pragma_representer(action, messages):
+def emit_pragma_representer(action: str, messages: List[str]) -> PragmaRepresenter:
     if not messages and action in MESSAGE_KEYWORDS:
         raise InvalidPragmaError(
             "The keyword is not followed by message identifier", action
@@ -65,7 +65,7 @@ class PragmaParserError(Exception):
     A class for exceptions thrown by pragma_parser module
     """
 
-    def __init__(self, message, token):
+    def __init__(self, message: str, token: str) -> None:
         """
         :args message: explain the reason why the exception has been thrown
         :args token: token concerned by the exception
@@ -88,7 +88,7 @@ class InvalidPragmaError(PragmaParserError):
 
 
 def parse_pragma(pylint_pragma: str) -> Generator[PragmaRepresenter, None, None]:
-    action = None
+    action: Optional[str] = None
     messages: List[str] = []
     assignment_required = False
     previous_token = ""
@@ -113,7 +113,9 @@ def parse_pragma(pylint_pragma: str) -> Generator[PragmaRepresenter, None, None]
                 raise InvalidPragmaError("Missing keyword before assignment", "")
             assignment_required = False
         elif assignment_required:
-            raise InvalidPragmaError("The = sign is missing after the keyword", action)
+            raise InvalidPragmaError(
+                "The = sign is missing after the keyword", action or ""
+            )
         elif kind == "KEYWORD":
             if action:
                 yield emit_pragma_representer(action, messages)


pylint/checkers/similar.py
```diff
@@ -369,11 +371,11 @@ def __init__(
         self.linesets: List["LineSet"] = []
 
     def append_stream(
-        self, streamid: str, stream: TextIOWrapper, encoding=None
+        self, streamid: str, stream: Union[BufferedReader, BytesIO], encoding=None
     ) -> None:
         """append a file to search for similarities"""
         if encoding is None:
-            readlines = stream.readlines
+            readlines: Union[Callable[..., Union[str, bytes]], Any] = stream.readlines
```
DanielNoord, line 378: We might want to refactor this, not the best of typing..

pylint/utils/ast_walker.py
```diff
@@ -3,28 +3,32 @@
...
+
 
 class ASTWalker:
-    def __init__(self, linter):
+    def __init__(self, linter: "PyLinter") -> None:
         # callbacks per node types
         self.nbstatements = 0
-        self.visit_events = collections.defaultdict(list)
-        self.leave_events = collections.defaultdict(list)
+        self.visit_events: collections.defaultdict = collections.defaultdict(list)
```
DanielNoord, line 18: This is missing more specific types, will fix in later commit

pylint/utils/utils.py
```diff
@@ -51,7 +65,7 @@ def diff_string(old, new):
     return diff_str
 
 
-def get_module_and_frameid(node):
+def get_module_and_frameid(node: Any) -> Tuple[str, str]:
```
DanielNoord, line 68: Is this `nodes.NodeNG`?

cdce8p:
I would suggest to wait until after at least #4940 is merged.

pylint/testutils/output_line.py
```diff
@@ -15,16 +16,20 @@ class Message(
     def __new__(cls, msg_id, line=None, node=None, args=None, confidence=None):
         return tuple.__new__(cls, (msg_id, line, node, args, confidence))
 
-    def __eq__(self, other):
+    def __eq__(self, other: Union[object, "Message"]) -> bool:
         if isinstance(other, Message):
             if self.confidence and other.confidence:
                 return super().__eq__(other)
-            return self[:-1] == other[:-1]
+            return self[:-1] == other[:-1]  # type: ignore
```
DanielNoord, line 23: I could not get this to work... Perhaps somebody else wants to take a look at this issue and how to solve it!

DanielNoord:
Tests are failing on the fact that `re.Pattern` is not available in Python 3.6.
Is there a known workaround for this? On two places I use `isinstance(x, re.Pattern)`. Can that be overwritten by something like `isinstance(x, re.BaseClass)` which is available in 3.6?

Pierre-Sassoulas:
> Is there a known workaround for this? On two places I use isinstance(x, re.Pattern). Can that be overwritten by something like isinstance(x, re.BaseClass) which is available in 3.6?

We could wait until the 23rd of December 2021 and drop python 3.6 before adding this typing and removing the ``type: ignore`` ? 😄 Might not be worth our time to fix this issue in 3.6 as very soon we won't have to.

DanielNoord:
I'll do that when I get back to my computer. Need to wait for #4940 anyway.

DanielNoord:
Any idea why the docs build is failing? Should I add `pytest` to the dependencies?

cdce8p:
> Any idea why the docs build is failing? Should I add `pytest` to the dependencies?

Better to guard the `pytest` import behind `if TYPE_CHECKING` and to use strings for the imported types (from pytest).

pylint/checkers/similar.py
Pierre-Sassoulas: I agree, but maybe in another merge request ?

pylint/lint/pylinter.py
```diff
@@ -729,8 +729,10 @@ def enable_fail_on_messages(self):
                         self.fail_on_symbols.append(msg.symbol)
 
     def any_fail_on_issues(self):
-        return self.stats is not None and any(
-            x in self.fail_on_symbols for x in self.stats["by_msg"]
+        return (
+            self.stats is not None
+            and self.stats.get("by_msg")
```
Pierre-Sassoulas, line 734:
```suggestion
            and self.stats.get("by_msg") is not None
```
Not doing that create nasty error if `by_msg == ""` or ``by_msg == 0``

pylint/reporters/reports_handler_mix_in.py
```diff
@@ -33,45 +39,55 @@ def register_report(self, reportid, r_title, r_cb, checker):
...
-                    continue
-                report_sect = Section(r_title)
-                try:
-                    r_cb(report_sect, stats, old_stats)
-                except EmptyReportError:
-                    continue
-                report_sect.report_id = reportid
-                sect.append(report_sect)
-        return sect
+        if isinstance(self, BaseChecker) and isinstance(self, ReportsHandlerMixIn):
```
Pierre-Sassoulas, line 64:
```suggestion
        if not isinstance(self, BaseChecker) or not isinstance(self, ReportsHandlerMixIn):
            raise TypeError
```
Premature return so the following code is simpler to understand with less to keep in mind and less indented.

pylint/reporters/text.py
```diff
@@ -59,7 +64,7 @@
 }
 
 
-def _get_ansi_code(color=None, style=None):
+def _get_ansi_code(color: str = None, style: Optional[str] = None) -> str:
```
Pierre-Sassoulas, line 67:
```suggestion
def _get_ansi_code(color: Optional[str] = None, style: Optional[str] = None) -> str:
```

pylint/testutils/output_line.py
```diff
@@ -78,19 +83,26 @@ def from_msg(cls, msg):
...
 
     @classmethod
-    def from_csv(cls, row):
+    def from_csv(
+        cls, row: Union[List[str], Tuple[str, str, str, str, str], str]
+    ) -> "OutputLine":
         try:
-            confidence = row[5] if len(row) == 6 else interfaces.HIGH.name
+            if len(row) == 6:
+                confidence = row[5]  # type: ignore # mypy does not recognize that this cannot fail
```
Pierre-Sassoulas, line 97: Something we might fix if we do #1590 (By uniformizing all the functional test results)

pylint/checkers/base.py
```diff
@@ -1159,12 +1159,14 @@ def _check_using_constant_test(self, node, test):
 
     def visit_module(self, _: nodes.Module) -> None:
         """check module name, docstring and required arguments"""
+        assert isinstance(self.stats["module"], int)
```
Pierre-Sassoulas, line 1162: In tests we can handle it this way, but we  can't do this in pylint's code. We probably need to remove the possibility that it's not an int altogether by doing a refactor. Crashing with an assert will still be a crash.

Pierre-Sassoulas (changes requested):

pylint/checkers/base.py
DanielNoord: Are you sure? Based on the code on L1163 I think `self.stats["module"]` will always be an `int`. Otherwise it would already produce crashes (and none seem to be reported).
I think this actually asserts something that is already always the case but is needed for `mypy` not to fail (and as a helpful remainder for us that this stat should always be an `int`).

DanielNoord:
There are some `type: ignore`'s in this final version. However, I would be in favour of merging this and adding typing to all other folders and then get back to those comments.
Some of those require extensive refactoring for which I currently do not have the time.

pylint/checkers/base.py
Pierre-Sassoulas: It makes mypy stop warning about the issue but it's the wrong way to fix the issue imo.
- It lower performance because we'll now make an assert at run time
- The root cause of ``self.stats`` not being always properly initialized and ``visit_module`` not being safely callable in badly initialized child class is not addressed (default value for ``self.stats`` is None).
- The error is only marginally better as we'd have a ``TypeError: can only concatenate str (not "int") to str`` already or something similarly clear.
- This is not the case here, but in general for pylint this kind of issue could come from not anticipating a strange code construct that will come from astroid. For example we think we can have only an inner function, but it happens the user defined an inner class. Asserting that we have a ``FunctionDef`` when we have a ``ClassDef`` instead will still crash pylint the same way that if we did nothing at all. It's just silencing the linter with an assumption we made.

I think this is a design issue on the base checker that would require a refactor and we may as well just ignore it until we do the refactor.

pylint/extensions/docparams.py
```diff
@@ -222,8 +223,11 @@ def visit_functiondef(self, node: nodes.FunctionDef) -> None:
         # skip functions smaller than 'docstring-min-length'
         lines = checker_utils.get_node_last_lineno(node) - node.lineno
         max_lines = get_global_option(self, "docstring-min-length")
-        if max_lines > -1 and lines < max_lines:
-            return
+        if isinstance(max_lines, int):
```
DanielNoord, line 226: This is a similar issue to the `assert` I have just removed (due to `linter.stats` having multiple types). Is this okay? Or do you want me to add `type: ignore` here as well?

pylint/extensions/typing.py
```diff
@@ -139,17 +139,21 @@ def __init__(self, linter: PyLinter) -> None:
         self._consider_using_alias_msgs: List[DeprecatedTypingAliasMsg] = []
 
     def open(self) -> None:
-        py_version: Tuple[int, int] = get_global_option(self, "py-version")
-        self._py37_plus = py_version >= (3, 7)
-        self._py39_plus = py_version >= (3, 9)
-        self._py310_plus = py_version >= (3, 10)
+        py_version = get_global_option(self, "py-version")
+        if isinstance(py_version, tuple):
```
DanielNoord, line 143: Same here as above.

pylint/extensions/docparams.py
Pierre-Sassoulas: I think this could be checked once when parsing the option, so we can also ignore.

DanielNoord: Done!

pylint/testutils/lint_module_test.py
```diff
@@ -138,21 +138,21 @@ def multiset_difference(
         return missing, unexpected
 
     # pylint: disable=consider-using-with
-    def _open_expected_file(self):
+    def _open_expected_file(self) -> Union[StringIO, TextIOWrapper, TextIO]:
```
cdce8p, line 141: `IOBase` or `TextIOBase`?

pylint/testutils/lint_module_test.py
```diff
@@ -163,10 +163,10 @@ def _get_expected(self):
             ]
         return expected_msgs, expected_output_lines
 
-    def _get_actual(self):
+    def _get_actual(self) -> Tuple[Counter, List[OutputLine]]:
         messages = self._linter.reporter.messages
         messages.sort(key=lambda m: (m.line, m.symbol, m.msg))
-        received_msgs = Counter()
+        received_msgs: Counter = Counter()
```
cdce8p, line 169: The annotation should not be necessary

pylint/testutils/output_line.py
cdce8p, line 97: Couldn't it fail if `row` is a string of length `6`. `isinstance(row, tuple)` might work

pylint/testutils/output_line.py
```diff
@@ -78,19 +83,26 @@ def from_msg(cls, msg):
...
+            else:
+                confidence = interfaces.HIGH.name
             column = cls.get_column(row[2])
             return cls(row[0], int(row[1]), column, row[3], row[4], confidence)
         except Exception as e:
-            raise MalformedOutputLineException(row, e) from e
+            if isinstance(e, ValueError):
+                raise MalformedOutputLineException(row, e) from e
+            raise e
```
cdce8p, line 105:
```suggestion
            raise
```
That would be enough, but why add the `isinstance` check in the first place?

pylint/testutils/reporter_for_tests.py
```diff
@@ -40,37 +42,44 @@ def handle_message(self, msg: Message) -> None:
...
+        self.messages.append(f"{sigle}:{line:>3}{obj}: {str_message}")  # type: ignore # Probably refactor at some point
 
-    def finalize(self):
+    def finalize(self) -> str:
         self.messages.sort()
         for msg in self.messages:
             print(msg, file=self.out)
-        result = self.out.getvalue()
+        if isinstance(self.out, StringIO):
+            result = self.out.getvalue()
```
cdce8p, line 52: Either this is always true, in which case it shouldn't be there in the first place or `result` is possibly unbound.
For 1: Add `out: StringIO` below the class docstring.

pylint/utils/ast_walker.py
```diff
@@ -3,28 +3,32 @@
...
         if not hasattr(method, "checks_msgs"):
             return True
-        for msg_desc in method.checks_msgs:
+        for msg_desc in method.checks_msgs:  # type: ignore # mypy does not pick up that method always has checks_msgs
             if self.linter.is_message_enabled(msg_desc):
                 return True
         return False
 
-    def add_checker(self, checker):
+    def add_checker(self, checker: Any) -> None:
```
cdce8p, line 31: `BaseChecker`?

pylint/utils/ast_walker.py
```diff
@@ -54,7 +58,7 @@ def add_checker(self, checker):
                     visits[cid].append(visit_default)
         # for now we have no "leave_default" method in Pylint
 
-    def walk(self, astroid):
+    def walk(self, astroid: Any) -> None:
```
cdce8p, line 61: Probably `nodes.NodeNG`

pylint/utils/file_state.py
```diff
@@ -99,8 +125,14 @@ def set_msg_status(self, msg, line, status):
...
     def handle_ignored_message(
-        self, state_scope, msgid, line, node, args, confidence
-    ):  # pylint: disable=unused-argument
+        self,
+        state_scope: int,
+        msgid: str,
+        line: int,
+        _: nodes.NodeNG,
+        __: Any,
+        ___: Confidence,
```
cdce8p, line 134: Please don't remove argument names even if unused

pylint/utils/utils.py
```diff
@@ -101,21 +126,25 @@ def safe_decode(line, encoding, *args, **kwargs):
         return line.decode(sys.getdefaultencoding(), *args, **kwargs)
 
 
-def decoding_stream(stream, encoding, errors="strict"):
+def decoding_stream(
+    stream: Union[BufferedReader, BytesIO, TextIO],
```
cdce8p, line 130: `IOBase`

pylint/utils/utils.py
```diff
@@ -101,21 +126,25 @@ def safe_decode(line, encoding, *args, **kwargs):
...
+    encoding: str,
+    errors: str = "strict",
+) -> Union[latin_1_StreamReader, utf_8_StreamReader, cp1252_StreamReader]:
```
cdce8p, line 133: Can't you use a base class here? `codecs.StreamReader`?

pylint/utils/utils.py
```diff
@@ -243,33 +276,38 @@ def _comment(string):
...
-        value = ".".join(str(item) for item in value)
+        value = ".".join(str(item) for item in value)  # type: ignore
     elif isinstance(value, (list, tuple)):
         value = ",".join(_format_option_value(optdict, item) for item in value)
     elif isinstance(value, dict):
         value = ",".join(f"{k}:{v}" for k, v in value.items())
     elif hasattr(value, "match"):  # optdict.get('type') == 'regexp'
         # compiled regexp
-        value = value.pattern
+        value = value.pattern  # type: ignore
```
cdce8p, line 292:
```suggestion
        value = cast(Pattern, value).pattern
```

tests/test_self.py
```diff
@@ -120,11 +120,11 @@ def _display(self, layout: EvaluationSection) -> None:
         pass
 
     @property
-    def out(self) -> StringIO:
+    def out(self) -> Union[TextIO, StringIO, None]:  # type: ignore # mypy doesn't seem to recognize properties
```
cdce8p, line 123: Properties are usually not an issue for mypy.
`IOBase`?

tests/test_self.py
```diff
@@ -120,11 +120,11 @@ def _display(self, layout: EvaluationSection) -> None:
...
         return self._reporters[0].out
 
-    @property
-    def linter(self) -> Optional[Any]:
+    @property  # type: ignore # mypy doesn't seem to recognize properties
+    def linter(self) -> Optional[PyLinter]:  # type: ignore # See comments above
```
cdce8p, line 127: Same here with the `type: ignore` comments

tests/utils/unittest_ast_walker.py
```diff
@@ -42,7 +42,7 @@ def test_check_messages(self) -> None:
         linter = self.MockLinter(
             {"first-message": True, "second-message": False, "third-message": True}
         )
-        walker = ASTWalker(linter)
+        walker = ASTWalker(linter)  # type: ignore # Can't import MockLinter. This needs a refactor
```
cdce8p, line 45: The issue is that `MockLinter` doesn't extend `PyLinter`

cdce8p (changes requested):
Thanks for your work, but tbh this was really difficult to review. It would be better if you could limit typing changes to much smaller PRs in the future.

One thing that stood out to me was `Union[BufferedReader, BytesIO, TextIO]`. It might be better to use a common base class where possible. Something like `IOBase`. I annotated it a bunch of times, haven't checked it in details though.

--
One last comment. To make the review process a bit easier, I would appreciate if you don't resolve any of the conversations. I won't be able to find comments otherwise. If you agree, simply add a 👍🏻 and change the code, I'll later go through and resolve them. If you like to start a conversation on a resolved item, please unresolve it.

pylint/checkers/similar.py
```diff
@@ -368,10 +371,15 @@ def __init__(
...
+    def append_stream(
+        self,
+        streamid: str,
+        stream: Union[BufferedReader, BytesIO, TextIO],
+        encoding=None,
+    ) -> None:
         """append a file to search for similarities"""
         if encoding is None:
-            readlines = stream.readlines
+            readlines: Union[Callable[..., Union[str, bytes]], Any] = stream.readlines
```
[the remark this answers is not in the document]
DanielNoord: I thought adding typing would be better so we can start moving to adding more mypy flags such as `--disallow-untyped-defs` etc.
I can add a `type: ignore` if you want, but personally I like how `Any` typings serve as a visual reminder that something is left to be done and allow searching for them within the project directory to create a "typing to-do list".

pylint/extensions/code_style.py
```diff
@@ -77,11 +77,11 @@ def __init__(self, linter: PyLinter) -> None:
...
-        self._py38_plus = py_version >= (3, 8)
-        self._max_length: int = (
-            self.config.max_line_length_suggestions
-            or get_global_option(self, "max-line-length")
+        py_version = get_global_option(self, "py-version")
+        self._py38_plus = py_version >= (3, 8)  # type: ignore # This option is always a Tuple[int, int]. Refactor necessary for mypy
+        self._max_length: int = self.config.max_line_length_suggestions or get_global_option(  # type: ignore
+            self,
+            "max-line-length",  # This comment is needed for black to not trip on these lines
         )
```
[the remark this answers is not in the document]
DanielNoord: Oops.. 😄

pylint/reporters/multi_reporter.py
```diff
@@ -88,15 +89,15 @@ def display_messages(self, layout: BaseLayout) -> None:
         for rep in self._sub_reporters:
             rep.display_messages(layout)
 
-    def on_set_current_module(self, module: str, filepath: Optional[AnyPath]) -> None:
+    def on_set_current_module(self, module: str, filepath: PosixPath) -> None:
```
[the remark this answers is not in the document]
DanielNoord: This is based on what the current test suite does. Apparently this function never gets called with `None` or `str`.
However, I think it would be better to change this indeed. I'm going to change this to `Optional[PosixPath, str]` and do a separate commit in which I do `Union[None, PosixPath, str]` -> `Optional[PosixPath, str]` for all other cases.

pylint/reporters/reports_handler_mix_in.py
```diff
@@ -33,24 +39,32 @@ def register_report(self, reportid, r_title, r_cb, checker):
...
+    def make_reports(
+        self: Union[BaseChecker, "ReportsHandlerMixIn"],
+        stats: Dict[str, Any],
+        old_stats: Dict,
+    ) -> Section:
         """render registered reports"""
+        if not isinstance(self, BaseChecker) or not isinstance(
+            self, ReportsHandlerMixIn
+        ):
+            raise TypeError
```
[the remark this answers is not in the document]
DanielNoord: Rather have me put a `type: ignore` here?

pylint/reporters/reports_handler_mix_in.py
```diff
@@ -61,17 +75,20 @@ def make_reports(self, stats, old_stats):
                     r_cb(report_sect, stats, old_stats)
                 except EmptyReportError:
                     continue
-                report_sect.report_id = reportid
+                if hasattr(report_sect, "report_id"):
+                    report_sect.report_id = reportid  # type: ignore
```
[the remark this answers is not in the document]
DanielNoord: The `type: ignore` is because `mypy` doesn't recognise `hasattr`, but I thought it's warning was fair and that we might want a check for it.

pylint/testutils/lint_module_test.py
```diff
@@ -88,26 +88,26 @@ def __str__(self):
...
+    def get_expected_messages(stream: TextIO) -> Counter:
         """Parses a file and get expected messages.
 
         :param stream: File-like input stream.
         :type stream: enumerable
         :returns: A dict mapping line,msg-symbol tuples to the count on this line.
         :rtype: dict
         """
-        messages = Counter()
+        messages: Counter = Counter()
```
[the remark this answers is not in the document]
DanielNoord: `Mypy` complains about this. Want me to add `type ignore`?

pylint/testutils/lint_module_test.py
DanielNoord: `TextIOBase`

pylint/testutils/lint_module_test.py
DanielNoord: `Mypy` complains when it is missing. Want me to add an ignore?

pylint/testutils/output_line.py
DanielNoord: Sorry! Totally forgot about that!

pylint/testutils/output_line.py
DanielNoord: Because `MalformedOutputLineException` requires a `ValueError` and `mypy` complains otherwise.

tests/utils/unittest_ast_walker.py
DanielNoord: Yes, but I also can't type `MockLinter` here. What do you want me to do?

DanielNoord:
I'm truly sorry for making this PR too large. I got a bit carried away when working and forgot to keep a close eye on how truly large this PR had become. If you want I can still split it up based on the three directories and created individual PR's based on your feedback?

I'm working on incorporating all of your feedback. One of the issues I ran into is that `TextIO` doesn't inherit from `IOBase`. It only inherits from `IO`. That class is often too broad for typing and would lead to additional `isinstance` checks to satisfy `mypy`. `TextIO` is returned by some of the functions in the `sys` library and is therefore also needed as a recurring type. Previously I solved this by adding the `Union`'s. I can see why you don't like that solution, but I'm stuck thinking of something better.
TLDR: `IO` creates `mypy` issues because of its broadness and the lack of certain standard attributes and methods of its sub-classes, `TextIO` doesn't inherit from `IOBase` but is returned by `sys` functions, current `Union`'s are too broad and ugly.

I'm sorry for resolving the conversation prematurely. I was used to resolving after incorporating feedback but I can see how this is unhelpful for you. I appreciate the effort you took to still review this PR instead of dismissing it because of its size (which I have seen happen in other projects). Thank you!

cdce8p:
I haven't taken a look at all the review comments yet. Only want to answer some of your questions now.

> If you want I can still split it up based on the three directories and created individual PR's based on your feedback?

In general smaller PRs might be better. Don't know though if directory bases is the best approach. Although I know it will be more work, I would think of it in related topics. Just to name some examples:
* Everything dealing with `self.stats`
* `stream` and `IOBase`
* `get_global_option` and related `type: ignore` comments. (I would appreciate this one especially!)
* `filepath`
* ... (there might be more, but that's a start)

> One of the issues I ran into is that `TextIO` doesn't inherit from `IOBase`.

To be honest with you, I haven't checked it in detail if it would work / where it might not. (There were just too many changes.)
Separating those into a new PR might be a way to better understand the relationships between all of them. In the end, if `Union` is the only thing that works, so be it.

> I'm sorry for resolving the conversation prematurely. I was used to resolving after incorporating feedback but I can see how this is unhelpful for you.

It's nothing personal. With so many different conversations I just felt that otherwise I couldn't follow it properly.

> I appreciate the effort you took to still review this PR instead of dismissing it because of its size (which I have seen happen in other projects). Thank you!

I can understand why it sometimes happens. Most maintainers, I would guess, don't know every inch of their code of the top of their head. So it takes a lot of mental energy to go through it if at all.

Pierre-Sassoulas (approved):
Thanks ! This uncovered so many things that ideally we'd have to refactor, it's definitely a step in the right direction.

Pierre-Sassoulas (approved):
👍

cdce8p (approved):

PR #4971, opened 2021-09-05
Extend consider-using-in for attribute access

cdce8p:
## Type of Changes

## Description
`consider-using-in` doesn't need to be limited to `Name` nodes. `Attribute` once work too.

```py
class A:
    def __init__(self):
        self.a = 2

    def func(self):
        var = self.a
        if var == 2 or var == 3:
            print()
        if self.a == 2 or self.a == 3:  # new
            print()

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -63,6 +63,8 @@ Release date: TBA
 
   Closes #4907
 
+* Extended ``consider-using-in`` check to work for attribute access.
+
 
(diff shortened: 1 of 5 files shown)


DudeNr33 (approved):

DudeNr33:
Yes, that makes sense that attributes should be included in the check!

PR #4973, opened 2021-09-06
Add typing to all calls to ``self.stats``

DanielNoord:

## Type of Changes

## Description

All checkers inherit from a baseclass which has a ``stats`` attribute.
This attribute has a fairly unmanageable type, but the current typing includes all variations of the attribute.
Other changes not directly related to ``self.stats`` are due to ``mypy`` warnings.
This incorporate the feedback received in #4954


pylint/reporters/multi_reporter.py
```diff
@@ -95,8 +95,14 @@ def on_set_current_module(self, module: str, filepath: Optional[AnyPath]) -> Non
 
     def on_close(
         self,
-        stats: Mapping[Any, Any],
-        previous_stats: Mapping[Any, Any],
+        stats: Dict[
```
DanielNoord, line 98: I changed this to `Dict` instead of `Mapping` as its only ever called with `self.stats` and using `Mapping` gave a `Mypy` warning because of an incorrect argument type for one of the calls to `on_close`.

Pierre-Sassoulas (commented):
I was thinking about creating a class so the typing is simpler and we can unit test it easily.

pylint/checkers/__init__.py
```diff
@@ -46,28 +46,38 @@
...
 from pylint.checkers.base_checker import BaseChecker, BaseTokenChecker
 from pylint.checkers.deprecated import DeprecatedMixin
 from pylint.checkers.mapreduce_checker import MapReduceMixin
 from pylint.utils import diff_string, register_plugins
 
 
-def table_lines_from_stats(stats, old_stats, columns):
+def table_lines_from_stats(
+    stats: Dict[
+        str, Union[int, Counter[str], List, Dict[str, Union[int, str, Dict[str, int]]]]
```
DanielNoord, line 59: The `List` included in the type of all `stats` comes from https://github.com/PyCQA/pylint/blob/40cc2ffd7887959157aaf469e09585ec2be7f528/pylint/checkers/imports.py#L469
This is then used in https://github.com/PyCQA/pylint/blob/40cc2ffd7887959157aaf469e09585ec2be7f528/pylint/graph.py#L170-L182

pylint/checkers/__init__.py
```diff
@@ -46,28 +46,38 @@
 
 """
 
+from typing import Counter, Dict, List, Tuple, Union
```
cdce8p, line 49: `typing.Counter` was added in `3.6.1`. Thus to use it, we need to add a guard unfortunately.
https://docs.python.org/3/library/typing.html#typing.Counter

pylint/checkers/__init__.py
```diff
@@ -46,28 +46,38 @@
...
 
 
-def table_lines_from_stats(stats, old_stats, columns):
+def table_lines_from_stats(
+    stats: Dict[
+        str, Union[int, Counter[str], List, Dict[str, Union[int, str, Dict[str, int]]]]
+    ],
+    old_stats: Dict[
+        str, Union[int, Counter[str], List, Dict[str, Union[int, str, Dict[str, int]]]]
+    ],
```
cdce8p, line 63: I'm a bit surprised that the code works with so many different types.
@Pierre-Sassoulas mentioned that we should probably create a `Stats` class or something like it, I agree. However as it would be a larger refactoring for another time, maybe a TypeAlias will work for now.

I would suggest to create `pylint/typing.py` as a helper file to store TypeAliases, Enums, NamedTuple, TypedDicts, and so on. (Only if it doesn't make sense somewhere else of course.)

cdce8p (changes requested):
Not a full review just yet

pylint/checkers/__init__.py
DanielNoord: I borrowed the solution from:
https://github.com/PyCQA/pylint/blob/40cc2ffd7887959157aaf469e09585ec2be7f528/pylint/checkers/strings.py#L51-L52
Let me know if this is correct!

tests/pyreverse/test_writer.py
```diff
@@ -28,7 +28,7 @@
 from unittest.mock import Mock
 
 import pytest
-from conftest import PyreverseConfig  # type: ignore
+from conftest import PyreverseConfig  # type: ignore #pylint: disable=no-name-in-module
```
DanielNoord, line 31: For some reason the import of `conftest` introduced in https://github.com/PyCQA/pylint/commit/25000863761220c1aa5db129e3cd41f5fe51b167 failed the `pre-commit`.

pylint/checkers/__init__.py
```diff
@@ -46,28 +46,35 @@
...
+        new: Union[int, str] = stats[m_type]  # type: ignore
+        old: Union[int, str, None] = old_stats.get(m_type)  # type: ignore
         if old is not None:
             diff_str = diff_string(old, new)
         else:
             old, diff_str = "NC", "NC"
         new = f"{new:.3f}" if isinstance(new, float) else str(new)
         old = f"{old:.3f}" if isinstance(old, float) else str(old)
-        lines += (m_type.replace("_", " "), new, old, diff_str)
+        lines += [m_type.replace("_", " "), new, old, diff_str]
```
cdce8p, line 77: This should probably be `lines.extend( [ ... ] )`, not `lines +=`

pylint/checkers/base.py
```diff
@@ -386,36 +387,49 @@ def _has_abstract_methods(node):
...
                 nice_stats[node_type]["percent_documented"] = "NC"
             try:
-                percent = (stats["badname_" + node_type] * 100.0) / total
+                badname_node: int = stats["badname_" + node_type]  # type: ignore
+                percent = (badname_node * 100.0) / total
                 nice_stats[node_type]["percent_badname"] = f"{percent:.2f}"
             except KeyError:
                 nice_stats[node_type]["percent_badname"] = "NC"
-    lines = ("type", "number", "old number", "difference", "%documented", "%badname")
+    lines: Tuple[str, ...] = (
```
cdce8p, line 422: This should be typed implicitly.

pylint/checkers/raw_metrics.py
```diff
@@ -15,27 +15,32 @@
...
+    old_stats: CheckerStatistics,
+):
     """calculate percentage of code / doc / comment / empty"""
-    total_lines = stats["total_lines"]
+    total_lines: int = stats["total_lines"]  # type: ignore
     if not total_lines:
         raise EmptyReportError()
     sect.description = f"{total_lines} lines have been analyzed"
-    lines = ("type", "number", "%", "previous", "difference")
+    lines: Tuple[str, ...] = ("type", "number", "%", "previous", "difference")
```
cdce8p, line 38: Again, probably not necessary.

pylint/lint/parallel.py
```diff
@@ -3,10 +3,12 @@
 
 import collections
 import functools
+from typing import Counter, Dict, List, Union
```
cdce8p, line 6: Unguarded `Counter` import.

pylint/lint/parallel.py
```diff
@@ -30,20 +32,20 @@ def _get_new_args(message):
     return (message.msg_id, message.symbol, location, message.msg, message.confidence)
 
 
-def _merge_stats(stats):
-    merged = {}
-    by_msg = collections.Counter()
+def _merge_stats(stats: List[CheckerStatistics]):
+    merged: CheckerStatistics = {}
+    by_msg: Counter[str] = collections.Counter()
```
cdce8p, line 37:
```suggestion
    by_msg: "Counter[str]" = collections.Counter()
```

pylint/lint/parallel.py
```diff
@@ -30,20 +32,20 @@ def _get_new_args(message):
...
     for stat in stats:
-        message_stats = stat.pop("by_msg", {})
+        message_stats: Union[Counter, Dict] = stat.pop("by_msg", {})  # type: ignore
```
cdce8p, line 39:
```suggestion
        message_stats: Union["Counter", Dict] = stat.pop("by_msg", {})  # type: ignore
```

pylint/lint/report_functions.py
```diff
@@ -16,33 +22,45 @@ def report_total_messages_stats(sect, stats, previous_stats):
...
+    by_msg_stats: Dict[str, int] = stats["by_msg"]  # type: ignore
+    in_order: List[Tuple[int, str]] = sorted(
         (value, msg_id)
-        for msg_id, value in stats["by_msg"].items()
+        for msg_id, value in by_msg_stats.items()
         if not msg_id.startswith("I")
     )
     in_order.reverse()
-    lines = ("message id", "occurrences")
+    lines: Tuple[str, ...] = ("message id", "occurrences")
```
cdce8p, line 41: Again, implicit.

pylint/typing.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+
+from typing import TYPE_CHECKING, Dict, List, Union
+
+if TYPE_CHECKING:
+    from typing import Counter  # typing.Counter added in Python 3.6.1
+
+
+# The base type of the "stats" attribute of a checker
+CheckerStatistics = Dict[
```
cdce8p, line 14: I'm wondering: Should this be `CheckerStats` instead? It's almost always `self.stats`, so it would make sense to use it here as well.

tests/pyreverse/test_writer.py
cdce8p: Shouldn't this be `from pylint.pyrevers.conftest import PyreverseConfig`?

cdce8p (commented):
I still feel like I need to think about this a bit longer. Left some comments nevertheless.

tests/pyreverse/test_writer.py
DanielNoord: See discussion here: https://github.com/PyCQA/pylint/pull/4950#discussion_r702048576
Seems like we can't fix this without a (big) refactor

pylint/checkers/base.py
DanielNoord: The problem is on line 437. Because the length of the tuple changes there. `mypy` doesn't like that and complains about a difference in typing. I could add an ignore on that line (and on the other cases you mentioned), but `Tuple[str, ...]` might be better in this case.

pylint/checkers/raw_metrics.py
DanielNoord: See above

pylint/lint/report_functions.py
DanielNoord: Similar problem, this time on line 43.
https://github.com/PyCQA/pylint/blob/04706af5d411ef5359fa195565f21d3fef7aed2b/pylint/lint/report_functions.py#L41-L43

DanielNoord:
@Pierre-Sassoulas You might want to change the milestone on this (or other PRs). Same goes for #4980 in order for the current typing efforts of you and me not to become double-work.

Just to let you know, I didn't plan on opening any new typing PRs before the current ones are all merged. Probably best to work on getting the current ones merged first. Perhaps you could help me with a review of #4980 as it requires quite extensive knowledge of `pylint` 😄

pylint/checkers/base.py
cdce8p: If additional items are added later, wouldn't it be better if lines where a `List[str]` instead?

pylint/lint/parallel.py
```diff
@@ -3,10 +3,15 @@
...
 import functools
+from typing import TYPE_CHECKING, Dict, List, Union
 
 from pylint import reporters
 from pylint.lint.utils import _patch_sys_path
 from pylint.message import Message
+from pylint.typing import CheckerStats
+
+if TYPE_CHECKING:
+    from typing import Counter  # typing.Counter added in Python 3.6.1
```
cdce8p, line 14: This could also be a `sys.version_info` guard with import from `typing_extensions`. In that case the quotes around `Counter` wouldn't be necessary. Leave that up to you, if you what to change it. Support for `3.6` will end soon anyway, so might not be that important.

pylint/reporters/reports_handler_mix_in.py
```diff
@@ -65,7 +74,9 @@ def make_reports(self, stats, old_stats):
                 sect.append(report_sect)
         return sect
 
-    def add_stats(self, **kwargs):
+    def add_stats(  # type: ignore # ReportsHandlerMixIn is always mixed with PyLinter
+        self: "PyLinter", **kwargs
```
cdce8p, line 78: Can `**kwargs` be typed? I.e. do you know the type for the dict values (in `kwargs`)?

pylint/reporters/ureports/nodes.py
```diff
@@ -16,6 +16,9 @@
 """
 
 
+from typing import Optional
```
cdce8p, line 19:
```suggestion
from typing import Optional
```
The empty lines are not really needed.

pylint/typing.py
```diff
@@ -0,0 +1,16 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+"""A collection of typing utilities
+"""
```
cdce8p, line 5:
```suggestion
"""A collection of typing utilities."""
```

pylint/typing.py
```diff
@@ -0,0 +1,16 @@
...
+
+from typing import TYPE_CHECKING, Dict, List, Union
```
cdce8p, line 7:
```suggestion
from typing import TYPE_CHECKING, Dict, List, Union
```
Empty line

pylint/typing.py
```diff
@@ -0,0 +1,16 @@
...
+
+if TYPE_CHECKING:
+    from typing import Counter  # typing.Counter added in Python 3.6.1
```
cdce8p, line 10: If you decide to change it, this should be changed as well.

pylint/checkers/base.py
DanielNoord: Yes! I can make all occurrences of `lines` a list and type them as `List[str]`. Want me to do that?
Originally I didn't want to change any code if not necessary and only add typing.

pylint/lint/parallel.py
DanielNoord: Do you know if `typing` imports get skipped during runtime? If not it might be better to keep it under the `if TYPE_CHECKING`?

pylint/reporters/reports_handler_mix_in.py
DanielNoord: Seems to be a mix of `int`, `list`, `dict` and `None`... I guess not? Or I could add a `Union`. What do you prefer?

pylint/checkers/base.py
cdce8p: I think that would make sense. Could you do that in a separate PR? Just the change to `list[str]`. That should be easy to review and fast to merge.

DanielNoord: Yes, give me a couple of minutes!

cdce8p: Tag me, and I'll take a look.

pylint/lint/parallel.py
cdce8p: No, they are not skipped. Basically they are normal imports. However, since `typing` as well as `typing_extensions` are already imported elsewhere, it shouldn't have much of a performance impact. If any at all.

pylint/reporters/reports_handler_mix_in.py
cdce8p: With that mix it might be a case for `Any` 🤔

pylint/checkers/base.py
DanielNoord: See https://github.com/PyCQA/pylint/pull/5009

pylint/lint/parallel.py
DanielNoord: Hmm okay. If you don't have a preference for either I think I will leave it at this.
I actually was thinking about making a checker for imports only being used in typing and a `if TYPE_CHECKING` statement. So this fits nicely!

pylint/reporters/reports_handler_mix_in.py
DanielNoord: Although `mypy` will probably complain about that in the future... I will add the Union 😄

pylint/lint/parallel.py
cdce8p: > I actually was thinking about making a checker for imports only being used in typing and a `if TYPE_CHECKING` statement. So this fits nicely!

Just a gut feeling: Don't know how useful such a checker would really be.

pylint/reporters/reports_handler_mix_in.py
Pierre-Sassoulas: I kinda agree with Marc here, this much type in a Union scream "whatever" or as mypy call it ``Any`` 😄 We'd need to refactor to make this code make sense and we'll probably throw some code if we create a proper class for stats later so we might as well not bother with too much precision in unsalvageable legacy code.

cdce8p: > We'd need to refactor to make this code make sense and we'll probably throw some code if we create a proper class for stats later so we might as well not bother with too much precision in unsalvageable legacy code.

Good point! Although imprecise, sometimes `Any` is the better choice after all. Even if only temporarily.

pylint/lint/parallel.py
DanielNoord: I thought it might be useful for Python 3.9, where people are more likely to not import from `typing`. For example, in some of my own projects I tend to quickly import from a dependency when I am working in a file that only has a helper function. Although performance effects might be negligible, (I think) it fits with the idea of `pylint` suggesting best practice.
But, this will probably be a discussion for when I make the first PR for this checker 😄

pylint/reporters/reports_handler_mix_in.py
DanielNoord: Good points! Add `Any` in latest commit!

pylint/checkers/__init__.py
```diff
@@ -46,28 +46,35 @@
...
 from pylint.checkers.mapreduce_checker import MapReduceMixin
+from pylint.typing import CheckerStats
 from pylint.utils import diff_string, register_plugins
 
 
-def table_lines_from_stats(stats, old_stats, columns):
+def table_lines_from_stats(
+    stats: CheckerStats,
+    old_stats: CheckerStats,
+    columns: Tuple[str, ...],
```
cdce8p, line 61: This could also be `Iterable[str]`. Just a suggestion, not strictly necessary.

cdce8p (commented):
@Pierre-Sassoulas Would you mind taking a look at this? Overall I think it's good. Haven't checked each type tough.

Pierre-Sassoulas (approved):
I like the new class name, and this cleans up a lot of potential problem too 👍

Pierre-Sassoulas:
> Shouldn't this be from pylint.pyrevers.conftest import PyreverseConfig?

Yes, I opened #5010

PR #4976, opened 2021-09-06
Fix #4370: false-positive no-member in class based on typing.NamedTuple

### LINKED ISSUE #4370: False 'no-member' on inner class of NamedTuple
The following produces a 'no-member' error.

```python
from typing import NamedTuple


class Ax(NamedTuple):
    class Bx:
        b = 0


print(Ax.Bx.b)

>>> pylint_no_member.py:9:6: E1101: Class 'Ax' has no 'Bx' member (no-member)

```

However, if Ax doesn't inherent from NamedTuple then it's fine.

```
Ubuntu 20.04.1 LTS
pylint 2.4.4
astroid 2.3.3
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0]
```

dmand:
This PR adds a functional test that's demonstrating #4370. The test should become green once PyCQA/astroid#1166 is merged and astroid is released.

(Oh, and i took the liberty to update my last name in contributors list.)

## Type of Changes

## Description

Closes #4370.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -25,7 +25,7 @@ Current team:
   Added Python 3 check for accessing deprecated methods on the 'string' module,
   various patches.
 
-* Dmitry Pribysh: committer
+* Dmitri Prybysh: committer
 
   multiple-imports, not-iterable, not-a-mapping, various patches.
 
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Thank you for the regression test :) I wonder how your name got added with a typo though 😄

dmand:
I'm from Belarus, we have really weird cyrillic-latin transliteration rules. :)
The old version was my preferred last name at the time, but now I go by the different version, so given the opportunity I've updated the contributors list. (Nobody can pronounce right either version anyway)

PR #4977, opened 2021-09-06
Use ``overload`` to type ``get_global_option``

DanielNoord:

## Type of Changes

## Description

The next step in splitting #4954 in more manageable PR's.

I opted to go for `overload` as it allows us to immediately tell what ``get_global_option`` will return or throw an error if it is unspecified.
The globals added to `pylint/utils/utils.py` could go in `pylint/typing` but since they won't be used outside of this file I thought it would be best to add them here.


DanielNoord:
Turns out `Literal` is only supported >=3.8. Going to think of something new.

PR #4978, opened 2021-09-06
Use ``overload`` to type ``get_global_option``

DanielNoord:

## Type of Changes

## Description

The next step in splitting #4954 in more manageable PR's.

I opted to go for `overload` as it allows us to immediately tell what ``get_global_option`` will return or throw an error if it is unspecified.
The globals added to `pylint/utils/utils.py` could go in `pylint/typing` but since they won't be used outside of this file I thought it would be best to add them here.

Sorry for making a second PR. The fix turned out to be much easier than anticipated.

### DIFF
--- a/pylint/utils/utils.py
+++ b/pylint/utils/utils.py
@@ -17,32 +17,34 @@ import re
 import sys
 import textwrap
 import tokenize
-from typing import TYPE_CHECKING, Any, List, Optional, Pattern, Tuple, Union, overload
+from typing import TYPE_CHECKING, List, Optional, Pattern, Tuple, Union, overload
 
 from astroid import Module, modutils
 
 from pylint.constants import PY_EXTS
 
-if TYPE_CHECKING:
+if sys.version_info >= (3, 8):
     from typing import Literal
+else:
+    from typing_extensions import Literal
 
+if TYPE_CHECKING:
     from pylint.checkers.base_checker import BaseChecker
 
 DEFAULT_LINE_LENGTH = 79
-GLOBAL_OPTION_BOOL = Union[
-    "Literal['ignore-mixin-members']",
-    "Literal['suggestion-mode']",
-    "Literal['analyse-fallback-blocks']",
-    "Literal['allow-global-unused-variables']",
+# These are types used to overload get_global_option() and refer to the options type
+GLOBAL_OPTION_BOOL = Literal[
+    "ignore-mixin-members",
+    "suggestion-mode",
+    "analyse-fallback-blocks",
+    "allow-global-unused-variables",
 ]
-GLOBAL_OPTION_INT = Union[
-    "Literal['max-line-length']", "Literal['docstring-min-length']"
+GLOBAL_OPTION_INT = Literal["max-line-length", "docstring-min-length"]
+GLOBAL_OPTION_LIST = Literal["ignored-modules"]
+GLOBAL_OPTION_PATTERN = Literal[
+    "no-docstring-rgx", "dummy-variables-rgx", "ignored-argument-names"
 ]
-GLOBAL_OPTION_PATTERN = Union[
-    "Literal['no-docstring-rgx']",
-    "Literal['dummy-variables-rgx']",
-    "Literal['ignored-argument-names']",
-]
+GLOBAL_OPTION_TUPLE_INT = Literal["py-version"]
 
 
 def normalize_text(text, line_len=DEFAULT_LINE_LENGTH, indent=""):
@@ -185,7 +187,7 @@ def get_global_option(
 @overload
 def get_global_option(
     checker: "BaseChecker",
-    option: "Literal['ignored-modules']",
+    option: GLOBAL_OPTION_LIST,
     default: Optional[List[str]] = None,
 ) -> List[str]:
     ...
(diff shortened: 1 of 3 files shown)


pylint/utils/utils.py
```diff
@@ -17,12 +17,32 @@
...
 import textwrap
 import tokenize
+from typing import TYPE_CHECKING, Any, List, Optional, Pattern, Tuple, Union, overload
 
 from astroid import Module, modutils
 
 from pylint.constants import PY_EXTS
 
+if TYPE_CHECKING:
+    from typing import Literal
```
cdce8p, line 27: This will fail if someone tries to run mypy on Python 3.7
Instead do something like this:
```py
if sys.version_info >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```

pylint/utils/utils.py
```diff
@@ -17,12 +17,32 @@
...
+
+    from pylint.checkers.base_checker import BaseChecker
+
 DEFAULT_LINE_LENGTH = 79
+GLOBAL_OPTION_BOOL = Union[
+    "Literal['ignore-mixin-members']",
+    "Literal['suggestion-mode']",
+    "Literal['analyse-fallback-blocks']",
+    "Literal['allow-global-unused-variables']",
+]
```
cdce8p, line 37: No need to use `Union` here. You can pass multiple values to `Literal`. Same for the other type aliases.
https://docs.python.org/3/library/typing.html#typing.Literal

pylint/utils/utils.py
```diff
@@ -148,7 +168,50 @@ def register_plugins(linter, directory):
...
+def get_global_option(
+    checker: "BaseChecker", option: GLOBAL_OPTION_INT, default: Optional[int] = None
+) -> int:
+    ...
+
+
+@overload
+def get_global_option(
+    checker: "BaseChecker",
+    option: "Literal['ignored-modules']",
```
cdce8p, line 188: Should we create an alias for it as well?

pylint/utils/utils.py
```diff
@@ -148,7 +168,50 @@ def register_plugins(linter, directory):
...
+    checker: "BaseChecker",
+    option: "Literal['py-version']",
+    default: Tuple[int, ...] = None,
+) -> Tuple[int, ...]:
+    ...
+
+
+def get_global_option(
+    checker: "BaseChecker", option: str, default: Optional[Any] = None
+) -> Any:
```
cdce8p, line 214: `str` and `Any` are technically correct. However we won't catch wrong usages either.
Can you use Unions of the overload types for it?

pylint/utils/utils.py
```diff
@@ -17,12 +17,32 @@
...
 
 from pylint.constants import PY_EXTS
 
+if TYPE_CHECKING:
+    from typing import Literal
+
+    from pylint.checkers.base_checker import BaseChecker
+
 DEFAULT_LINE_LENGTH = 79
+GLOBAL_OPTION_BOOL = Union[
```
cdce8p, line 32: Maybe add a small comment before to mention that these are for `get_global_option`

cdce8p (commented):
I like this a lot 😍

I also though of `overload`. That was the reason I wanted a separate PR for it 😉
Furthermore the change is quite concise which makes it really easy to review.

pylint/utils/utils.py
DanielNoord: I found that for some reason `VAR = "Literal['string']"` doesn't work. But with 
```if sys.version_info >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```
this is again possible!

pylint/utils/utils.py
DanielNoord: Removed because  VAR = "Literal['string']" doesn't work. But will re-add.

pylint/utils/utils.py
DanielNoord: I guess this also means you agree with keeping them here instead of in `pylint/typing`?

cdce8p: Yeah. There isn't really a need to import them anywhere else as they are only required for the overloads of `get_global_option`.

pylint/utils/utils.py
```diff
@@ -148,7 +170,58 @@ def register_plugins(linter, directory):
...
+    default: Optional[Pattern] = None,
+) -> Pattern:
+    ...
+
+
+@overload
+def get_global_option(
+    checker: "BaseChecker",
+    option: GLOBAL_OPTION_TUPLE_INT,
+    default: Tuple[int, ...] = None,
```
cdce8p, line 209:
```suggestion
    default: Optional[Tuple[int, ...]] = None,
```

pylint/utils/utils.py
```diff
@@ -148,7 +170,58 @@ def register_plugins(linter, directory):
...
+    checker: "BaseChecker",
+    option: Union[
+        GLOBAL_OPTION_BOOL,
+        GLOBAL_OPTION_INT,
+        GLOBAL_OPTION_LIST,
+        GLOBAL_OPTION_PATTERN,
+        GLOBAL_OPTION_TUPLE_INT,
+    ],
+    default: Union[None, bool, int, List[str], Pattern, Tuple[int, ...]] = None,
+) -> Union[None, bool, int, List[str], Pattern, Tuple[int, ...]]:
```
cdce8p, line 224: We could define a type alias for it as well.

--
Let's also try something different. I haven't checked if it would work out,  but what about using a `TypeVar`?
```py
from typing import TypeVar

T = TypeVar("T", bool, int, List[str], Pattern, Tuple[int, ...])

def get_global_options(checker: ..., option: ..., default: Optional[T] = None) -> T:
    ...
```

DanielNoord:
Not really sure about the naming of the type alias and `TypeVar`. Went with `GLOBAL_OPTION_RETURN_TYPES` and `GLOBAL_OPTION_NAMES` respectively, but let me know what you think!

pylint/utils/utils.py
```diff
@@ -17,13 +17,55 @@
...
+GLOBAL_OPTION_NAMES = Union[
+    GLOBAL_OPTION_BOOL,
+    GLOBAL_OPTION_INT,
+    GLOBAL_OPTION_LIST,
+    GLOBAL_OPTION_PATTERN,
+    GLOBAL_OPTION_TUPLE_INT,
+]
+GLOBAL_OPTION_RETURN_TYPES = TypeVar(
+    "GLOBAL_OPTION_RETURN_TYPES", bool, int, List[str], Pattern, Tuple[int, ...]
+)
```
cdce8p, line 67: Usually TypeVars are `PascalCase`, similar to class names or even just `T`.
What do you think about `T_GlobalOptionReturnTypes` (best of both worlds)?

cdce8p (commented):
The `TypeVar` seems to work. It will prevent things like this where the return type doesn't match the `default`:
```py
@overload
def get_global_option(
    checker: "BaseChecker", option: GLOBAL_OPTION_BOOL, default: Optional[bool] = None
) -> str:
    ...
```

pylint/utils/utils.py
DanielNoord: Sounds good!

cdce8p (approved):
Looks good! Thanks @DanielNoord 🎉

PR #4980, opened 2021-09-07
Add typing to ``filepath``

DanielNoord:

## Type of Changes

## Description

This fails the test suite, but I think this might be because the test suite is testing something that doesn't happen in normal use of pylint. Please correct me if I am wrong but after investigating the call-chain I believe that almost all `filepath`'s handled by `pylint` are `str` (the only exception being `get_fatal_error_message()`). This presumes that no user directly calls `linter.check()` as we do in the test suite. 
Of note is `lint.py_run()`, which I believe also does no run `check()` before calling `load_command_line_configuration`.

The following reasoning leads me to believe that a normal use of `pylint` always creates a `str` file path.
- `load_command_line_configuration()` in `pylint/config/option_manager_mixin.py` always returns a `List[str]`. Therefore the `args` variable in the `__init__` of the `Run` class in `pylint/lint/run.py` is always a `List[str]`.
- Subsequently in `pylint/lint/pylinter.py` the `files_or_modules` argument to `check()` is always a `List[str]`. Which means that both `check_parallel()` and `_check_files()` only handle `str` representations of paths.
- Therefore `set_current_module()` and `on_set_current_module()` in various other classes only get given `str`.

https://github.com/DanielNoord/pylint/pull/10
This PR shows that with changes to the tests this PR can succeed. These changes are:
- Making sure all initial file paths are a `List[str]`
- Making sure that `test_worker_check_single_file_uninitialised()` catches the new explicitly defined exception
- Making sure that `get_fatal_error_message()` always gets a `Path` as second argument
- Making sure that `check_parallel()` always gets an `Iterable`

These changes seem to follow the standard format of these arguments/variables in normal runs of `pylint` that are being skipped in the tests.


pylint/lint/expand_modules.py
```diff
@@ -1,9 +1,28 @@
...
 
 from astroid import modutils
 
+if sys.version_info >= (3, 8):
+    from typing import Literal, TypedDict
+else:
+    from typing_extensions import Literal, TypedDict
+
+
+class MODULE_DESCRIPTION_DICT(TypedDict):
```
DanielNoord, line 13: After a (potential) merge of #4973 these should be added to `pylint/typing`.
This PR is therefore blocked by #4973

pylint/lint/parallel.py
```diff
@@ -62,9 +63,13 @@ def _worker_initialize(linter, arguments=None):
...
 
 
-def _worker_check_single_file(file_item):
+def _worker_check_single_file(
+    file_item: Tuple[str, str, str]
+) -> Tuple[int, Any, str, Any, List[Tuple[Any, ...]], Any, Any, DefaultDict[Any, List]]:
     name, filepath, modname = file_item
 
+    if not _worker_linter:
+        raise Exception("Worker linter not yet initialised")
```
DanielNoord, line 72: `mpy` complains on the following two lines about the possibility that `_work_linter` might be None.
This is probably the wrong `Exception` type.
We currently already test for this, but do so with catching an `AttributeError` on L73. See:
https://github.com/PyCQA/pylint/blob/fab2de272b3fc36dc4480d89840b8bfe0fbfbd0c/tests/test_check_parallel.py#L177-L182
I think making this error more explicit as `mypy` wants makes sense.

pylint/lint/expand_modules.py
cdce8p, line 13: This should use `PascalCase`. Same for `ERROR_DESCRIPTION_DICT` below.

pylint/lint/parallel.py
```diff
@@ -62,9 +63,13 @@ def _worker_initialize(linter, arguments=None):
     _patch_sys_path(arguments or ())
 
 
-def _worker_check_single_file(file_item):
+def _worker_check_single_file(
+    file_item: Tuple[str, str, str]
```
cdce8p, line 67: Does it make sense to define the type for `file_item` as NamedTuple? It seems to be used in a few places.

cdce8p (commented):
Not a review, just some thoughts.

pylint/lint/parallel.py
```diff
@@ -151,7 +156,7 @@ def check_parallel(linter, jobs, files, arguments=None):
             linter.set_current_module(module, file_path)
             for msg in messages:
                 msg = Message(*msg)
-                linter.reporter.handle_message(msg)
+                linter.reporter.handle_message(msg)  # type: ignore # L134 makes linter always have a reporter attribute
```
cdce8p, line 159: Mentioning line numbers in comments isn't worth it. The code will eventually get updated without the comment.

pylint/lint/parallel.py
DanielNoord: Good call!
Probably after https://github.com/PyCQA/pylint/pull/4973 though to add it to the `typing.py` document.

pylint/lint/parallel.py
```diff
@@ -151,7 +156,7 @@ def check_parallel(linter, jobs, files, arguments=None):
             linter.set_current_module(module, file_path)
             for msg in messages:
                 msg = Message(*msg)
-                linter.reporter.handle_message(msg)
+                linter.reporter.handle_message(msg)  # type: ignore # set_reporter() call above makes linter have a reporter attr
```
cdce8p, line 159:
```suggestion
                linter.reporter.handle_message(msg)  # type: ignore  # linter.set_reporter() call above makes linter have a reporter attr
```

cdce8p (commented):
I wouldn't necessarily wait for #4973 to add `typing.py`. You could also add it here. It shouldn't be too hard to resolve the merge conflicts in the end.

For `Tuple[str, str, str]` it might even make sense to create a separate one. You should you by now that I like small PRs 😄

DanielNoord:
@cdce8p I added the NamedTuple. However, doing this separately was (quite) awful because of the various new `mypy` errors that I already fixed in this PR. (For example raising the Exception on `if not _worker_linter`).
So I added it here!

pylint/lint/expand_modules.py
```diff
@@ -1,9 +1,28 @@
...
+    name: str
+    isarg: bool
+    basepath: str
+    basename: str
+
+
+class ErrorDescriptionTest(TypedDict):
+    key: Literal["fatal"]
+    mod: str
+    ex: Union[ImportError, SyntaxError]
```
cdce8p, line 24: You can move these to `pylint/typing.py` now.

pylint/lint/parallel.py
```diff
@@ -109,7 +115,7 @@ def _merge_mapreduce_data(linter, all_mapreduce_data):
             checker.reduce_map_data(linter, collated_map_reduce_data[checker.name])
 
 
-def check_parallel(linter, jobs, files, arguments=None):
+def check_parallel(linter, jobs, files: Iterator[FileItem], arguments=None):
```
cdce8p, line 118: Does it need to be an `Iterator` or is `Iterable` enough?
https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable

pylint/lint/pylinter.py
```diff
@@ -935,16 +937,13 @@ def initialize(self):
             if not msg.may_be_emitted():
                 self._msgs_state[msg.msgid] = False
 
-    def check(self, files_or_modules):
+    def check(self, files_or_modules: List[str]) -> None:
```
cdce8p, line 940: `Iterable`?

pylint/lint/pylinter.py
```diff
@@ -970,7 +969,7 @@ def check(self, files_or_modules):
                 files_or_modules,
             )
 
-    def check_single_file(self, name, filepath, modname):
+    def check_single_file(self, name: str, filepath: str, modname: str) -> None:
```
cdce8p, line 972: These arguments seems to represent a `FileItem`.

pylint/lint/pylinter.py
```diff
@@ -982,7 +981,11 @@ def check_single_file(self, name, filepath, modname):
                 self.get_ast, check_astroid_module, name, filepath, modname
             )
 
-    def _check_files(self, get_ast, file_descrs):
+    def _check_files(
+        self,
+        get_ast,
+        file_descrs: Union[List[FileItem], Iterator[FileItem]],
```
cdce8p, line 987:
```suggestion
        file_descrs: Iterable[FileItem],
```

pylint/lint/pylinter.py
```diff
@@ -982,7 +981,11 @@ def check_single_file(self, name, filepath, modname):
...
+    ):
```
cdce8p, line 988:
```suggestion
    ) -> None:
```

pylint/lint/pylinter.py
```diff
@@ -1004,12 +1007,14 @@ def _check_files(self, get_ast, file_descrs):
...
-                        msg = (filepath, msg)
+                        self.add_message(symbol, args=(filepath, msg))
                     else:
                         symbol = "fatal"
-                    self.add_message(symbol, args=msg)
+                        self.add_message(symbol, args=msg)
 
-    def _check_file(self, get_ast, check_astroid_module, name, filepath, modname):
+    def _check_file(
+        self, get_ast, check_astroid_module, name: str, filepath: str, modname: str
```
cdce8p, line 1016: These again seem to represent a `FileItem`.

cdce8p (commented):
I haven't commented on that before. A best practice when typing in python is to use the most generic type possible for arguments and the most concrete one for return types. I.e.
```py
def func(var: Iterable[str]) -> List[str]:
    ...
```

If it doesn't matter if a `list`, `tuple` or something else is pass in as long as it's possible to iterate over it, it shouldn't matter to the type checker either.

pylint/lint/pylinter.py
DanielNoord: Needed some refactoring, but added!

pylint/lint/pylinter.py
DanielNoord: Done in own commit!

pylint/lint/parallel.py
```diff
@@ -79,7 +83,7 @@ def _worker_check_single_file(file_item):
     return (
         id(multiprocessing.current_process()),
         _worker_linter.current_name,
-        filepath,
+        file_item[1],
```
cdce8p, line 86:
```suggestion
        file_item.filepath,
```
One advantage of using a NamedTuple.

pylint/lint/pylinter.py
```diff
@@ -935,16 +937,13 @@ def initialize(self):
             if not msg.may_be_emitted():
                 self._msgs_state[msg.msgid] = False
 
-    def check(self, files_or_modules):
+    def check(self, files_or_modules: Sequence[str]) -> None:
```
cdce8p, line 940: `Sequence` is also an excellent choice!

pylint/lint/pylinter.py
```diff
@@ -935,16 +937,13 @@ def initialize(self):
...
         """main checking entry: check a list of files or modules from their name.
 
-        files_or_modules is either a string or list of strings presenting modules to check.
+        files_or_modules is a list of strings presenting modules to check.
```
cdce8p, line 943:
```suggestion
        files_or_modules is a sequence of strings presenting modules to check.
```

pylint/lint/pylinter.py
```diff
@@ -1004,23 +1005,22 @@ def _check_files(self, get_ast, file_descrs):
...
         - filepath: path to the file to check
         - name: Python module name
         :param callable check_astroid_module: callable checking an AST taking the following arguments
         - ast: AST of the module
-        :param str name: full name of the module
-        :param str filepath: path to checked file
-        :param str modname: name of the checked Python module
+        :param FileItem file: data about the file
         """
+        name, filepath, modname = file
```
cdce8p, line 1023: Why do a tuple unpacking. With `NamedTuple` you can access the attributes by name.
It might even make sense to just pass `file` to `set_current_module` / `get_ast` and do the attribute access there. A bit more refactoring though.

cdce8p (commented):
I really like where this is going!

Pierre-Sassoulas:
>  I believe that almost all filepath's handled by pylint are str

We should handle ``Union[Path, str]`` imo.

pylint/lint/pylinter.py
DanielNoord: I left the refactor for another time, since this is becoming quite a large PR.

One thing I noticed is some ambiguity between `modname` and `modpath`. They seem to be used for the same attribute. We currently use `modpath`, is that okay? Or do you want to start using `modname`?

Pierre-Sassoulas: +1 for keeping the refactors for later :)

DanielNoord:
> We should handle `Union[Path, str]` imo.

Not sure if I agree here. This would probably lead us to having to add `Path` to many other typings down the line. The current flow of `pylint` clearly does not allow `Path` to be passed. With a little refactor we can even get the last instance of `Path` to also become a `str`. Having to add `Union[Path, str]` instead of `str` everywhere might clutter the codebase.

@cdce8p & @Pierre-Sassoulas Do you guys thinks your review is far enough for me to add the commit with test fixes as discussed in the PR description and shown in https://github.com/DanielNoord/pylint/pull/10?

Pierre-Sassoulas:
> might clutter the codebase.

Yes, also the migration to Path is also a huge endeavor with not a lot of clear benefit right now. Most ``os.path`` handles ``Union[Path, str]`` so technically that's what we can probably handle on most of the function, but finding out where we can't is probably not worth the time and crashes it will create.

DanielNoord:
This now includes the relevant changes to the tests to make this work. As discussed in PR description I believe the tests check things that cannot happen in real world scenarios. The description includes some additional reasoning for this and a description of the changes to the tests.

pylint/lint/pylinter.py
```diff
@@ -938,16 +939,13 @@ def initialize(self):
...
         """main checking entry: check a list of files or modules from their name.
 
-        files_or_modules is either a string or list of strings presenting modules to check.
+        files_or_modules is a sequence of strings presenting modules to check.
         """
         self.initialize()
 
-        if not isinstance(files_or_modules, (list, tuple)):
-            files_or_modules = (files_or_modules,)
-
```
Pierre-Sassoulas, line 950: I like this, clearer input, it will make the code better. But I think this might be a breaking change.

pylint/lint/pylinter.py
```diff
@@ -1059,19 +1057,19 @@ def _get_file_descr_from_stdin(filepath):
         except ImportError:
             modname = os.path.splitext(os.path.basename(filepath))[0]
 
-        return (modname, filepath, filepath)
+        return FileItem(modname, filepath, filepath)
```
Pierre-Sassoulas, line 1060: 👍

pylint/lint/pylinter.py
cdce8p: Technically all changes of function arguments are breaking ones as well (`check_single_file`, `check`, `_check_files`). It depends if we would consider them part of the public API. Something I'm not sure about.

Pierre-Sassoulas: Well I think the check function can be used by addon relying on pylint's checker. I'm for changing the typing only right now and removing the mechanism in 3.0 as a breaking change (we can also add a deprecation warning).

pylint/lint/pylinter.py
```diff
@@ -938,16 +939,13 @@ def initialize(self):
...
                 self._msgs_state[msg.msgid] = False
 
-    def check(self, files_or_modules):
+    def check(self, files_or_modules: Sequence[str]) -> None:
         """main checking entry: check a list of files or modules from their name.
 
-        files_or_modules is either a string or list of strings presenting modules to check.
+        files_or_modules is a sequence of strings presenting modules to check.
         """
         self.initialize()
```
Pierre-Sassoulas, line 948:
```suggestion
        if not isinstance(files_or_modules, (list, tuple)):
            warnins.warn("In pylint 3.0, the checkers check function will only accept sequence of string")
            files_or_modules = (files_or_modules,)
```

pylint/lint/pylinter.py
```diff
@@ -938,16 +939,17 @@ def initialize(self):
             if not msg.may_be_emitted():
                 self._msgs_state[msg.msgid] = False
 
-    def check(self, files_or_modules):
+    def check(self, files_or_modules: Sequence[str]) -> None:
```
cdce8p, line 942: `Union[Sequence[str], str]` maybe with a `TODO` comment after the deprecation warning to update the type annotation.

pylint/lint/pylinter.py
```diff
@@ -1007,32 +1011,30 @@ def _check_files(self, get_ast, file_descrs):
...
                         symbol = "astroid-error"
-                        msg = (filepath, msg)
+                        self.add_message(symbol, args=(filepath, msg))
                     else:
                         symbol = "fatal"
-                    self.add_message(symbol, args=msg)
+                        self.add_message(symbol, args=msg)
 
-    def _check_file(self, get_ast, check_astroid_module, name, filepath, modname):
+    def _check_file(self, get_ast, check_astroid_module, file: FileItem):
```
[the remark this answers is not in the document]
DanielNoord: As in, keep the `FileItem` or revert to previous state?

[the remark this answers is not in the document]
cdce8p: 😅 Keep the change with `FileItem`

pylint/lint/pylinter.py
```diff
@@ -938,16 +939,18 @@ def initialize(self):
             if not msg.may_be_emitted():
                 self._msgs_state[msg.msgid] = False
 
-    def check(self, files_or_modules):
+    def check(self, files_or_modules: Union[Sequence[str], str]) -> None:
         """main checking entry: check a list of files or modules from their name.
 
-        files_or_modules is either a string or list of strings presenting modules to check.
+        files_or_modules is a sequence of strings presenting modules to check.
```
cdce8p, line 945: This should be reverted

pylint/lint/pylinter.py
```diff
@@ -938,16 +939,18 @@ def initialize(self):
...
         """
         self.initialize()
-
         if not isinstance(files_or_modules, (list, tuple)):
-            files_or_modules = (files_or_modules,)
-
+            warnings.warn(
```
cdce8p, line 949:
```suggestion
            # pylint: disable-next=fixme
            # TODO: Update typing and docstring for 'files_or_modules' when removing the deprecation
            warnings.warn(
```

pylint/lint/pylinter.py
```diff
@@ -938,16 +939,18 @@ def initialize(self):
...
+                "In pylint 3.0, the checkers check function will only accept sequence of string",
+                DeprecationWarning,  # TODO: Update typing of files_or_modules to Sequence[str] # pylint: disable=fixme
```
cdce8p, line 951:
```suggestion
                DeprecationWarning,
```

pylint/lint/pylinter.py
```diff
@@ -973,66 +976,64 @@ def check(self, files_or_modules):
                 files_or_modules,
             )
 
-    def check_single_file(self, name, filepath, modname):
-        """Check single file
+    def check_single_file(self, name: str, filepath: str, modname: str) -> None:
+        warnings.warn(
+            "In pylint 3.0, the checkers check_single_file function will only accept a FileItem",
```
cdce8p, line 981:
```suggestion
            "In pylint 3.0, the checkers check_single_file function will be removed. "
            "Use check_single_file_item instead.",
```
What do you think?

pylint/lint/pylinter.py
```diff
@@ -973,66 +976,64 @@ def check(self, files_or_modules):
...
+            DeprecationWarning,  # TODO: Replace check_single_file with check_single_file_item # pylint: disable=fixme
```
cdce8p, line 982:
```suggestion
            DeprecationWarning,
```

pylint/typing.py
```diff
@@ -2,11 +2,48 @@
...
+    from typing_extensions import Literal, TypedDict
+
+
+class FileItem(NamedTuple):
+    """Represents data about a file handled by pylint
+
+    Each file item has:
+    - name: full name of the module
+    - filepath: path of the file
+    - modname: module name"""
```
cdce8p, line 23:
```suggestion
    - modname: module name
    """
```
PEP8 -> closing quotes should be on a new line if docstring > 1 line.

pylint/lint/pylinter.py
DanielNoord: Don't really have a preference. Guess this works as well and requires less refactoring for `3.0`.

pylint/lint/pylinter.py
DanielNoord: Weird how Github doesn't allow a suggestion here..

cdce8p: I've highlighted both an old an a new line. That's probably why.

Pierre-Sassoulas (approved):

PR #4983, opened 2021-09-08
Add typing to references of ``stream`` and ``stream()``

DanielNoord:

## Type of Changes

## Description

The next instalment of breaking up #4954 into more manageable PR's. 
This PR is based on two related commits which I felt should be part of the same PR as they deal with a similar topic.

The first is annotation of all implementations of `process_module`. This is fairly straightforward and covers adding `nodes.Module` to the typing of the parameter and changing `module`, `astroid` and other parameter names to the new standard of `node`. As `process_module` calls `node.stream()` often this seemed like a good first step to this PR.

The second commit covers annotation of (most) functions with a reference to `stream`, `filestream` or a similar parameter. There are two important comments about this commit.

The first is for `pylint/checkers/similar.py` where I choose to use `STREAM_TYPES = Union[TextIO, BufferedReader, BytesIO]`. I did this because there seems to be a certain particularity with `TextIO` and `IOBase`. See the (small) discussion here: https://github.com/python/typeshed/issues/3225 In practice this means that return types of `open("b")` and `open("t")` are incompatible. This became problematic and is why I opted to introduce `STREAM_TYPES`. This also means that typing is a little more restrictive than using `IOBase` although in practice this probably won't make a real difference.

The second is that in `pylint/utils/utils.py` I don't fully type the `options` parameter and use `Tuple`. This is because trying to determine the type of the second argument of this `tuple` became difficult and I didn't want to overstretch this PR. If we turn on some of the more stricter flags of mypy in the future this will probably show up again and should be solved then.

### DIFF
--- a/examples/custom_raw.py
+++ b/examples/custom_raw.py
@@ -1,3 +1,5 @@
+from astroid import nodes
+
 from pylint.checkers import BaseChecker
 from pylint.interfaces import IRawChecker
 
@@ -22,7 +24,7 @@ class MyRawChecker(BaseChecker):
     }
     options = ()
 
-    def process_module(self, node):
+    def process_module(self, node: nodes.Module) -> None:
         """process a module
 
         the module's content is accessible via node.stream() function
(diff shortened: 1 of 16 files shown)


Pierre-Sassoulas (approved):
Lgtm :)

Pierre-Sassoulas (approved):
👍

PR #4986, opened 2021-09-09
Fix no-self-use and docparams extension for async functions/methods

SupImDos:
## Tasks
- [ ] If it's a new feature, or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.
- [x] Added tests 

## Type of Changes

## Description

Fixes apparent bug with `no-self-use` and the `docparams` extension with async functions/methods.

Minimal example:
https://gist.github.com/SupImDos/6ca27b625b5cbc2aebb8ed30b2995cef

Appears that `visit_asyncfunctiondef()` and `leave_asyncfunctiondef()` were missing for those two checkers.


cdce8p (commented):
Thank you for opening the PR! The change itself looks good already. Just a few points we still need to address before it can be merged.
- Could you add a small changelog entry?
- It seems one test is currently failing `tests/functional/a/assign/assignment_from_no_return_py3.py`. Just add `no-self-use` to the pylint disable comment.
- It would be nice if the change itself would be tested. Here would be a good place for `no-self-use`: https://github.com/PyCQA/pylint/blob/1e85bb755a4f5afcacd3b18397d5c35c820fcca1/tests/functional/n/no/no_self_use.py#L16-L18
Similar the error message needs to be added here: https://github.com/PyCQA/pylint/blob/main/tests/functional/n/no/no_self_use.txt#L1
- For `missing-param-doc` there is: https://github.com/PyCQA/pylint/blob/main/tests/functional/ext/docparams.py

SupImDos:
@cdce8p 

Did those 4 tasks in the last 4 commits. Appears tests are passing.

Let me know if commit `31b4ce0` is not okay, revamped it a bit, but not sure if I have ruined the original intentions/spirit of those tests.

Thanks!

ChangeLog
```diff
@@ -77,6 +77,8 @@ Release date: TBA
 
   Closes #4900
 
+# Fix ``no-self-use`` and ``docparams extension`` for async functions and methods.
```
Pierre-Sassoulas, line 80:
```suggestion
* Fix ``no-self-use`` and ``docparams extension`` for async functions and methods.
```

Pierre-Sassoulas (approved):
The tests are on point, the fix is straightforward, great first contribution !

cdce8p (approved):
Looks good to me!
Thanks @SupImDos 🐬

PR #4992, opened 2021-09-11
Documentation for ``pyreverse`` and ``symilar``

### LINKED ISSUE #4616: Add documentation for ``pyreverse`` and ``symilar``

``pyreverse`` is an awesome tool to visualize the project structure and dependencies.
However, it is not even mentioned once in the documentation.
I think adding at least some basic instructions on how to use it and what the output looks like would make it a bit more widely known.

Probably ``symilar`` should get a few sentences as well, but I must admit that I do not know if it is really intended to be used from the command line directly?

The last console script which gets installed additionally to ``pylint`` is ``epylint``. I do not think that it needs additional documentation, since as I understand it its main purpose is the integration with Emacs, which is already described properly in the docs.

Add documentation pages for at least ``pyreverse``, probably also ``symilar``.

You can assign this issue directly to me, if you want.

DudeNr33:

## Type of Changes

## Description
This adds some documentation for ``pyreverse`` and ``symilar`` in the new section _Additional Commands_.

Closes #4616


doc/additional_commands/index.rst
```diff
@@ -0,0 +1,194 @@
...
+'''''''''''''''''
+
+To run ``pyreverse``, use::
+
+  pyreverse [options] <packages>
+
+<packages> can also be a single Python module.
+The following options are availabe:
+
+.. option:: -h, --help
```
Pierre-Sassoulas, line 27: I think we should advertise the help option only and make it exhaustive. Or we could generate the documentation using the code so there is something in the online doc ?

doc/additional_commands/index.rst
```diff
@@ -0,0 +1,194 @@
...
+
+   Default: none
+
+   set the output directory path.
+
+
+Example Output
+''''''''''''''
+
+Example diagrams generated with the .puml output format are shown below.
```
Pierre-Sassoulas, line 118:
```suggestion
Example diagrams generated with the ``.puml`` output format are shown below.
```

Pierre-Sassoulas (commented):
This is a welcome addition, thank you

doc/additional_commands/index.rst
DudeNr33: I see your point. Describing all the options by hand might lead to doc and implementation becoming out of sync.

We could create a new script beside the existing ones at ``doc/exts`` and generate the documentation similar as in ``pylint_extensions.py``, but that seems a bit overkill as in contrast to the optional checkers ``pyreverse``'s options will remain relatively static. Maybe we could in some future MR implement something similar to ``sphinx-argparse``, but until then I will go with just advertising the ``-h`` option.

Pierre-Sassoulas (approved):
👌

PR #4995, opened 2021-09-13
Remove most `typing.cast()` calls

DanielNoord:

## Type of Changes

## Description

After merge of https://github.com/PyCQA/astroid/pull/1168 I was asked to work on this.
I left in ~three~ two calls.

One and two (because `mypy`/IDE fails to see L120-122):
https://github.com/PyCQA/pylint/blob/ab327e6c9394dcdf9492996865a8b120da6bd4bc/pylint/extensions/code_style.py#L120-L147

~Three (because `NodeNG.scope()` is not typed and we therefore lose typing on `scope`):~
https://github.com/PyCQA/pylint/blob/ab327e6c9394dcdf9492996865a8b120da6bd4bc/pylint/checkers/refactoring/recommendation_checker.py#L122-L124


cdce8p:
The `cast` in `code_checker.py` are necessary and thus cannot be removed unfortunately.

--
Regarding the `recommendation_checker.py` and `scope`:
I would recommend to move the `cast` call in front of the `for` loop. After that, `loop_node` should be detected correctly. I.e.
```py
scope = cast(nodes.NodeNG, node.scope())
```

DanielNoord:
> The `cast` in `code_checker.py` are necessary and thus cannot be removed unfortunately.

I think this is a typo. I didn't change `code_checker.py`, which file are you referring to?

> Regarding the `recommendation_checker.py` and `scope`:

Can we use cast to type `scope` to `Union[Module, FunctionDef, ClassDef, Lambda, GeneratorExpr]`? If we are changing things we might as well be as specific as possible.
I took a quick look at typing `NodeNG.scope()` in `astroid` but this isn't a 2-minute fix because `node.parent` is `Optional[NodeNG]`, thus causing `mypy` issue.

cdce8p:
> I think this is a typo. I didn't change `code_checker.py`, which file are you referring to?

You mentioned it in the PR description. Only wanted to comment on that.

> Can we use cast to type `scope` to `Union[Module, FunctionDef, ClassDef, Lambda, GeneratorExpr]`? If we are changing things we might as well be as specific as possible.

I wouldn't recommend that. My rule for `cast` is usually to choose the most basic type that is still enough. If the implementation changes (it likely won't in this case), it's more likely that the `cast` will still be accurate.

> I took a quick look at typing `NodeNG.scope()` in `astroid` but this isn't a 2-minute fix because `node.parent` is `Optional[NodeNG]`, thus causing `mypy` issue.

Left some comments on https://github.com/PyCQA/astroid/pull/1170.

DanielNoord:
> You mentioned it in the PR description. Only wanted to comment on that.

Ah, that was `code_style.py`. But yeah, can't do anything about it. 

> I wouldn't recommend that. My rule for `cast` is usually to choose the most basic type that is still enough. If the implementation changes (it likely won't in this case), it's more likely that the `cast` will still be accurate.

👍🏻

DanielNoord:
This is (potentially) blocked by https://github.com/PyCQA/astroid/pull/1170 as it would allow removing one additional `cast()` call.

Pierre-Sassoulas:
> This is (potentially) blocked by PyCQA/astroid#1170 as it would allow removing one additional cast() call.

We can wait for astroid 2.8, then, there's only two merge request to handle now : https://github.com/PyCQA/astroid/milestone/16

cdce8p:
Temporarily marked it as draft. Just until the PR is merged

Pierre-Sassoulas (approved):
LGTM !

DanielNoord:
Still blocked by https://github.com/PyCQA/astroid/pull/1170

cdce8p (approved):

DanielNoord:
This does not `astroid` 2.8.1, but that shouldn't be a problem for now :)

PR #4997, opened 2021-09-13
Make ``consider-iterating-dictionary`` consider memb. check

### LINKED ISSUE #4069: consider-dictionary-membership
When testing if a key is contained in a dictionary, it's O(1) to use `in` directly on the dict object as opposed to O(N) through `.keys()`.

So along the lines of `consider-iterating-dictionary`, there should be a check for this that suggests removing `keys()`.

DanielNoord:

## Type of Changes

## Description

This closes #4069

The original issue talked about a new checker but I felt this fitted quite nicely within the description of the original checker.
I also removed a redundant check in the line before the message was added.


tests/functional/c/consider/consider_iterating_dictionary.txt
```diff
@@ -1,15 +1,17 @@
...
-consider-iterating-dictionary:29:24::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:30:29::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:31:11::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:36:24::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:36:55::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:37:31::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:37:61::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:38:30::Consider iterating the dictionary directly instead of calling .keys()
-consider-iterating-dictionary:38:60::Consider iterating the dictionary directly instead of calling .keys()
+consider-iterating-dictionary:23:16::Consider iterating the dictionary directly instead of calling .keys():HIGH
```
DanielNoord, line 1: Sorry for the large diff. I used the command to automatically update the data. Want me to change this?

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -75,7 +75,13 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
             return
         if node.func.attrname != "keys":
             return
-        if not isinstance(node.parent, (nodes.For, nodes.Comprehension)):
+        if not (
+            isinstance(node.parent, (nodes.For, nodes.Comprehension))
+            or (
+                isinstance(node.parent, nodes.Compare)
+                and isinstance(node.parent.parent, nodes.If)
+            )
```
cdce8p, line 83: The brackets are redundant. `and` binds stronger than `or`.

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -75,7 +75,13 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
             return
         if node.func.attrname != "keys":
             return
-        if not isinstance(node.parent, (nodes.For, nodes.Comprehension)):
+        if not (
+            isinstance(node.parent, (nodes.For, nodes.Comprehension))
+            or (
+                isinstance(node.parent, nodes.Compare)
+                and isinstance(node.parent.parent, nodes.If)
```
cdce8p, line 82: What about these examples?
```py
if [1, 2] == {1: "a", 2: "b"}.keys():
    pass

var = 1 in {}.keys()
```

Instead of testing `node.parent.parent`, why not test `ops` of the compare node for `in`?

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -26,7 +26,7 @@ class RecommendationChecker(checkers.BaseChecker):
             "consider-iterating-dictionary",
             "Emitted when the keys of a dictionary are iterated through the .keys() "
             "method. It is enough to just iterate through the dictionary itself, as "
-            'in "for key in dictionary".',
+            'in "for key in dictionary" or "if key in dictionary".',
```
cdce8p, line 29: The added text doesn't really match why it's possible. Maybe it would be better to name that as a separate case for this checker? (e.g. dict lookup is quicker than list comparison)

tests/functional/c/consider/consider_iterating_dictionary.txt
cdce8p: Doesn't really matter. At least not here. With 200+ changes, I would probably evaluate that differently.

cdce8p (changes requested):

pylint/checkers/refactoring/recommendation_checker.py
DanielNoord: Always get these messed up..

tests/functional/c/consider/consider_iterating_dictionary.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=missing-docstring, expression-not-assigned, too-few-public-methods, no-member, import-error, no-self-use, line-too-long, useless-object-inheritance, unnecessary-comprehension
+# pylint: disable=missing-docstring, expression-not-assigned, too-few-public-methods, no-member, import-error, no-self-use, line-too-long, useless-object-inheritance, unnecessary-comprehension, use-dict-literal
```
Pierre-Sassoulas, line 2:
```suggestion
# pylint: disable=missing-docstring, expression-not-assigned, too-few-public-methods
# pylint: disable=no-member, import-error, no-self-use, line-too-long, useless-object-inheritance
# pylint: disable= unnecessary-comprehension, use-dict-literal

```
Nitpick: As we're changing the whole txt result already we might as well reduce the lenght of the disable line too :)

Pierre-Sassoulas (approved):
👍 nice false negative fixed !

tests/functional/c/consider/consider_iterating_dictionary.py
DanielNoord: See latest commit :)

Pierre-Sassoulas (approved):

DanielNoord:
@Pierre-Sassoulas Any idea why your merge-commit made it fail the tests? Seems strange..

Pierre-Sassoulas:
I don't know, we did not change anything for invalid-name on the main branch. This is very strange.

cdce8p:
Really strange. I though a rebase might fix it, but apparently not 🤷🏻‍♂️
Tested the commits on my personal dev branch, and the `invalid-name` doesn't happen.

tests/functional/c/consider/consider_iterating_dictionary.py
```diff
@@ -36,3 +38,30 @@ def keys(self):
...
+    pass
+if 1 in CustomClass().keys():
+    pass
+if 1 in dict():
+    pass
+if 1 in dict().values():
+    pass
+if (1, 1) in dict().items():
+    pass
+if [1] == {}.keys(): # [consider-iterating-dictionary]
```
cdce8p, line 59: This shouldn't be an error.

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -75,16 +77,17 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
...
-        if not isinstance(node.parent, (nodes.For, nodes.Comprehension)):
-            return
-
-        inferred = utils.safe_infer(node.func)
-        if not isinstance(inferred, astroid.BoundMethod) or not isinstance(
-            inferred.bound, nodes.Dict
+        if (
+            isinstance(node.parent, (nodes.For, nodes.Comprehension))
+            or isinstance(node.parent, nodes.Compare)
+            and any(op for op, _ in node.parent.ops if op in {"in", "==", "is"})
```
cdce8p, line 83: This should probably be only for `in`.

cdce8p:
This seems to be caused by https://github.com/PyCQA/astroid/pull/979

tests/functional/c/consider/consider_iterating_dictionary.py
DanielNoord: Ah sorry, I misunderstood you. Fixed!

tests/functional/c/consider/consider_iterating_dictionary.py
```diff
@@ -1,4 +1,6 @@
-# pylint: disable=missing-docstring, expression-not-assigned, too-few-public-methods, no-member, import-error, no-self-use, line-too-long, useless-object-inheritance, unnecessary-comprehension
+# pylint: disable=missing-docstring, expression-not-assigned, too-few-public-methods
+# pylint: disable=no-member, import-error, no-self-use, line-too-long, useless-object-inheritance
+# pylint: disable=unnecessary-comprehension, use-dict-literal
```
Pierre-Sassoulas, line 3:
```suggestion
# pylint: disable=unnecessary-comprehension, use-dict-literal,invalid-name
```

Pierre-Sassoulas (commented):
Maybe a false negative for invalid-name that was fixed in the latest version of astroid ? ``var`` is an invalid name (bad names list), so I think we can disable it or change ``var`` name to something else

tests/functional/c/consider/consider_iterating_dictionary.py
DanielNoord: I think it would be better to change the name than add an additional disable here.
However, I don't see `var` in `bad-names` in `./pylintrc`.

Furthermore, this doesn't explain why this is only being thrown in the CI. The tests pass locally for me (and I think also for @cdce8p). Probably should try and investigate what is exactly causing this before adding a disable.
@cdce8p seems to have an idea, or at least has located the source

nelfin: I was able to replicate the failure when running under tox: (`tox -e py36 -- -k consider_iterating`). Sorry about jumping in, I was just pinged by the mention of PyCQA/astroid#979

DanielNoord: I forgot to add `--recreate`, which you probably did some time before. Can now indeed reproduce.

I can change this to another variable name. Can somebody tell me where `pylint` is getting the `bad-names list` from? I'm probably overlooking something..

nelfin: As to why this shows up after that merge on astroid, my reading of https://github.com/PyCQA/pylint/blob/bc95cd34071ec2e71de5bca8ff95cc9b88e23814/pylint/checkers/base.py#L1975-L1984 is that `invalid-name` is only emitted if the Assign `.value` can be inferred. Since `Compare` nodes are only just now inferred, these popped up.

cdce8p: Seems I mixed up `invalid-name` with `disallowed-name` myself.

`var` is fine to us, but here it's inferred as constant which should be uppercase (like @DanielNoord mentioned).
https://github.com/PyCQA/astroid/pull/979 added additional inference for `compare` nodes which is why these can now correctly be inferred as constants. Thus `invalid-name` is emitted now.

Sorry for the ping @nelfin.

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -25,8 +25,10 @@ class RecommendationChecker(checkers.BaseChecker):
             "Consider iterating the dictionary directly instead of calling .keys()",
             "consider-iterating-dictionary",
             "Emitted when the keys of a dictionary are iterated through the .keys() "
-            "method. It is enough to just iterate through the dictionary itself, as "
-            'in "for key in dictionary".',
+            "method or when .keys() is used for a membership check. "
+            "It is enough to just iterate through the dictionary itself, as "
+            'in "for key in dictionary". For membership checks '
+            '("if key in dictionary") this is actually faster.',
```
cdce8p, line 31:
```suggestion
            "in ``for key in dictionary``. For membership checks, "
            "``if key in dictionary``, is faster.",
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -75,16 +77,17 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
...
-        if not isinstance(node.parent, (nodes.For, nodes.Comprehension)):
-            return
-
-        inferred = utils.safe_infer(node.func)
-        if not isinstance(inferred, astroid.BoundMethod) or not isinstance(
-            inferred.bound, nodes.Dict
+        if (
+            isinstance(node.parent, (nodes.For, nodes.Comprehension))
+            or isinstance(node.parent, nodes.Compare)
+            and any(op for op, _ in node.parent.ops if op == "in")
```
cdce8p, line 83: It's a bit constructed, but technically possible. No warning should be emitted for
```py 
var = [1, 2] == {}.keys() in {False}
```

It should be enough to add an additional condition to the `if` clause, testing that the second value of `ops` is in fact the call `node` itself.

cdce8p (commented):
Some last comments

pylint/checkers/refactoring/recommendation_checker.py
DanielNoord: Some day I want to know as much about Python as you seem to do and be able to come up with these tests 😄

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -25,8 +25,10 @@ class RecommendationChecker(checkers.BaseChecker):
             "Consider iterating the dictionary directly instead of calling .keys()",
             "consider-iterating-dictionary",
             "Emitted when the keys of a dictionary are iterated through the .keys() "
```
cdce8p, line 27:
```suggestion
            "Emitted when the keys of a dictionary are iterated through the ``.keys()`` "
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -25,8 +25,10 @@ class RecommendationChecker(checkers.BaseChecker):
...
-            "method. It is enough to just iterate through the dictionary itself, as "
-            'in "for key in dictionary".',
+            "method or when .keys() is used for a membership check. "
```
cdce8p, line 28:
```suggestion
            "method or when ``.keys()`` is used for a membership check. "
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -25,8 +25,10 @@ class RecommendationChecker(checkers.BaseChecker):
...
+            "It is enough to just iterate through the dictionary itself, as "
+            "in ``for key in dictionary``. For membership checks, "
```
cdce8p, line 30:
```suggestion
            "It is enough to iterate through the dictionary itself, "
            "``for key in dictionary``. For membership checks, "
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -25,8 +25,10 @@ class RecommendationChecker(checkers.BaseChecker):
...
+            "``if key in dictionary``, is faster.",
```
cdce8p, line 31:
```suggestion
            "``if key in dictionary`` is faster.",
```

pylint/checkers/refactoring/recommendation_checker.py
cdce8p: Thanks for the compliment ❤️

cdce8p (approved):

PR #5003, opened 2021-09-14
Pylint reporter cleanup

Pierre-Sassoulas:
## Type of Changes

## Description

This create a file for the ``BaseWriter`` that was hidden in the init file and was creating circular import when trying to add typing in #5004 . Also remove a piece of code that was always failing,(probably a misunderstanding at some point). It should provide slightly better performances by not try-catching an error and calling lower on the class name multiple time..


pylint/reporters/ureports/__init__.py
```diff
@@ -1,98 +1,6 @@
...
-            cols -= 1
-            result[-1].append(cell)
-        # fill missing cells
-        result[-1] += [""] * (cols - len(result[-1]))
-        return result
-
-    def compute_content(self, layout) -> Iterator[str]:
-        """trick to compute the formatting of children layout before actually
-        writing it
+__all__ = ["BaseWriter"]
```
cdce8p, line 4:
```suggestion
__all__ = ("BaseWriter",)
```

cdce8p (approved):
Not an area I know too much about, but the change itself looks good.

PR #5005, opened 2021-09-14
Use node_ancestors to get rid of some while-loops

nickdrozd:

### DIFF
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -1492,14 +1492,14 @@ class BasicChecker(_BasicChecker):
         if not self._tryfinallys:
             return
         # the node could be a grand-grand...-children of the try...finally
+        _parent = node.parent
         _node = node
-        for parent in node.node_ancestors():
-            if isinstance(parent, breaker_classes):
-                break
-            if hasattr(parent, "finalbody") and _node in parent.finalbody:
+        while _parent and not isinstance(_parent, breaker_classes):
+            if hasattr(_parent, "finalbody") and _node in _parent.finalbody:
                 self.add_message("lost-exception", node=node, args=node_name)
                 return
-            _node = parent
+            _node = _parent
+            _parent = _node.parent
 
     def _check_reversed(self, node):
         """check that the argument to `reversed` is a sequence"""
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -238,8 +238,7 @@ def _redefines_import(node):
 
 def in_loop(node):
     """return True if the node is inside a kind of for loop"""
-    parent = node.parent
-    while parent is not None:
+    for parent in node.node_ancestors():
         if isinstance(
             parent,
             (
@@ -251,7 +250,6 @@ def in_loop(node):
             ),
         ):
             return True
-        parent = parent.parent
     return False
 
 
@@ -847,8 +845,7 @@ class BasicErrorChecker(_BasicChecker):
 
     def _check_in_loop(self, node, node_name):
         """check that a node is inside a for or while loop"""
-        _node = node.parent
-        while _node:
+        for _node in node.node_ancestors():
             if isinstance(_node, (nodes.For, nodes.While)):
                 if node not in _node.orelse:
                     return
@@ -861,8 +858,6 @@ class BasicErrorChecker(_BasicChecker):
                 and isinstance(node, nodes.Continue)
             ):
                 self.add_message("continue-in-finally", node=node)
-
-            _node = _node.parent
 
         self.add_message("not-in-loop", node=node, args=node_name)
 
@@ -1497,14 +1492,14 @@ class BasicChecker(_BasicChecker):
         if not self._tryfinallys:
             return
         # the node could be a grand-grand...-children of the try...finally
-        _parent = node.parent
         _node = node
-        while _parent and not isinstance(_parent, breaker_classes):
+        for _parent in node.node_ancestors():
+            if isinstance(_parent, breaker_classes):
+                break
             if hasattr(_parent, "finalbody") and _node in _parent.finalbody:
                 self.add_message("lost-exception", node=node, args=node_name)
                 return
             _node = _parent
-            _parent = _node.parent
 
     def _check_reversed(self, node):
         """check that the argument to `reversed` is a sequence"""
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -845,16 +845,16 @@ class BasicErrorChecker(_BasicChecker):
 
     def _check_in_loop(self, node, node_name):
         """check that a node is inside a for or while loop"""
-        for _node in node.node_ancestors():
-            if isinstance(_node, (nodes.For, nodes.While)):
-                if node not in _node.orelse:
+        for parent in node.node_ancestors():
+            if isinstance(parent, (nodes.For, nodes.While)):
+                if node not in parent.orelse:
                     return
 
-            if isinstance(_node, (nodes.ClassDef, nodes.FunctionDef)):
+            if isinstance(parent, (nodes.ClassDef, nodes.FunctionDef)):
                 break
             if (
-                isinstance(_node, nodes.TryFinally)
-                and node in _node.finalbody
+                isinstance(parent, nodes.TryFinally)
+                and node in parent.finalbody
                 and isinstance(node, nodes.Continue)
             ):
                 self.add_message("continue-in-finally", node=node)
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (changes requested):
Can we keep only the one that are clear cut simplification where we replace 3 lines by a for loops ? Imo it's harder to understand a break in the for loop than the original whiles and the one where we have to still do ``node_origin = parent`` became really confusing.

nickdrozd:
Refactoring these loops turned out to be a little trickier than I anticipated. The logic is broadly similar, but there are lots of small differences and irregularites. My feeling is that these differences are mostly accidental, but I don't know how to deal with them just now.

pylint/checkers/typecheck.py
```diff
@@ -545,7 +546,7 @@ def _emit_no_member(node, owner, owner_name, ignored_mixins=True, ignored_none=T
                 )
             ):
                 return False
-        node_origin, parent = parent, parent.parent
```
nickdrozd, line 548: @Pierre-Sassoulas Is this the one you're talking about? I agree that it's a little complicated, but I would say that was _already_ the case, especially with this double-assignment parent-child walk-up maneuver here. Getting rid of that IMO justifies adding in the break logic. But I can take it out if you feel strongly about it.

Maybe a way to make it clearer would be to return false straightaway instead of breaking the loop.

Pierre-Sassoulas: I don't have strong opinion, let's see what Marc think.

pylint/__pkginfo__.py
```diff
@@ -27,7 +27,7 @@ def get_numversion_from_version(v: str) -> Tuple:
                 version.append(int(num))
             except ValueError:
                 version.append(0)
-    while len(version) != 3:
+    for _ in range(3 - len(version)):
```
cdce8p, line 30: IMO the `while` statement is a bit easier to read and understand.

pylint/checkers/base.py
```diff
@@ -838,8 +836,7 @@ def _check_else_on_loop(self, node):
 
     def _check_in_loop(self, node, node_name):
         """check that a node is inside a for or while loop"""
-        _node = node.parent
-        while _node:
+        for _node in node.node_ancestors():
```
cdce8p, line 839:
```suggestion
        for parent in node.node_ancestors():
```
I would suggest to be consistent with it and use `parent` whenever possible.

pylint/checkers/base.py
```diff
@@ -1488,14 +1483,14 @@ def _check_not_in_finally(self, node, node_name, breaker_classes=()):
         if not self._tryfinallys:
             return
         # the node could be a grand-grand...-children of the try...finally
-        _parent = node.parent
         _node = node
-        while _parent and not isinstance(_parent, breaker_classes):
+        for _parent in node.node_ancestors():
```
cdce8p, line 1487:
```suggestion
        for parent in node.node_ancestors():
```

pylint/checkers/typecheck.py
cdce8p: It works either way, with `while` and with `for`. However I think the previous code was a bit easier to understand in this case. I especially like the tuple assignment as it makes it clear what actually happens.

If I remember correctly, I used something similar in another helper method.

--
A change to an early `return` wouldn't change much. Instead it would only create the possibility to add new errors.

pylint/checkers/utils.py
```diff
@@ -377,11 +375,9 @@ def is_defined_before(var_node: nodes.Name) -> bool:
     (statement_defining ; statement_using).
     """
     varname = var_node.name
-    _node = var_node.parent
-    while _node:
+    for _node in var_node.node_ancestors():
```
cdce8p, line 378:
```suggestion
    for parent in var_node.node_ancestors():
```

cdce8p (changes requested):
> Imo it's harder to understand a break in the for loop than the original whiles

It depends on the situation. For the changes here, I think it's mostly ok to use an early break. In particular something like this is quite easy to understand
```py
for parent in node.node_ancestors():
    if isinstance(parent, breaker_classes):
        break
    ...
```

> Refactoring these loops turned out to be a little trickier than I anticipated. The logic is broadly similar, but there are lots of small differences and irregularites. My feeling is that these differences are mostly accidental, but I don't know how to deal with them just now.

Noticed that as well while looking through the code. Unfortunately, I don't think it's accidental. Each and ever one of the checks / helper methods needs to deal with a very specific set of rules. Something simple like where the loop starts, at the current node or the parent, can have a huge impact and potentially break stuff.

--
The new helper is good, but we don't need to force it upon the existing code. Only using it where it makes sense is probably a good compromise.

Pierre-Sassoulas (approved):
I agree with @cdce8p's remarks, once they're taken into account we can merge.

pylint/__pkginfo__.py
nickdrozd: I tried another way with no loop at all. Compare this with https://github.com/PyCQA/pylint/pull/4860/files#diff-26d7b9f740481d02edafcd71a3f448d5d2f5a7ff9858bfa59622e72cfc87a3ed

pylint/checkers/base.py
nickdrozd: My first thought was to make the changes while disturbing the rest of the code as little as possible. But the variables are renamed now.

pylint/__pkginfo__.py
```diff
@@ -27,8 +27,8 @@ def get_numversion_from_version(v: str) -> Tuple:
                 version.append(int(num))
             except ValueError:
                 version.append(0)
-    while len(version) != 3:
-        version.append(0)
+    # pad out version with 0s
+    version += [0] * (3 - len(version))
```
cdce8p, line 31: Just my opinion, but I would prefer the old way. I find it more intuitive.
Besides, I don't see a good point in changing a method that is only kept for compatibility reasons.

pylint/checkers/base.py
```diff
@@ -1488,14 +1483,14 @@ def _check_not_in_finally(self, node, node_name, breaker_classes=()):
...
-            if hasattr(_parent, "finalbody") and _node in _parent.finalbody:
+        for parent in node.node_ancestors():
+            if isinstance(parent, breaker_classes):
+                break
+            if hasattr(parent, "finalbody") and _node in parent.finalbody:
                 self.add_message("lost-exception", node=node, args=node_name)
                 return
-            _node = _parent
-            _parent = _node.parent
+            _node = parent
```
cdce8p, line 1493: Iterating through `node.parent` and `node.parent.parent` seems to be more common than I thought. Maybe we should add a utils method for it, this time in `pylint/checkers/utils.py`. Something like `walk_node_parents` or `recurse_node_parents`?

pylint/checkers/typecheck.py
cdce8p: --
Obviously, this would be a good spot for the new helper.

pylint/checkers/variables.py
```diff
@@ -1349,7 +1348,6 @@ def _in_lambda_or_comprehension_body(
                 # Furthermore, only the first generator (if multiple) in comprehension should have access.
                 return True
             child = parent
-            parent = parent.parent
```
cdce8p, line 1352: It could be used here as well

cdce8p (changes requested):

pylint/__pkginfo__.py
nickdrozd: Lately I've been exploring the idea of programming using only primitive-recursive operators. Turing-completeness is not often required, and most tasks can be done without it. The `while` operator is awfully powerful, and it's definitely overkill for such a simple task as "pad out this list until it's at least of length 3". Ideally there would be no `while` loops at all except for where they're really needed, like perhaps in inference.

Of course this is an idiosyncratic concern, so I reverted the change.

pylint/checkers/base.py
nickdrozd: There's another common variation where the last ancestor satisfying some condition is returned. I'm not sure if that needs another method or if there's a different pattern for that kind of thing.

pylint/__pkginfo__.py
Pierre-Sassoulas: There's this french author, Perec, who wrote a full novel without the letter e in it. It's an awfully powerful letter - the most used one in french in fact - and apparently, often you don't really need it 😄 Anyway I digress, removing part of a language for the beauty of it is interesting as an exercise, but pylint's users won't benefit from it and may even suffer from it because reviewing opinionated changes takes maintainer time that could be used elsewhere (we have 690 issues currently). Please don't refactor working code without a good reason.

pylint/checkers/base.py
Pierre-Sassoulas: I think a generator ``n for n in node.node_ancestor if satisfy_condition(n)`` is the best solution in term of readability for those.

nickdrozd: Yet another variation is looping over a node's siblings rather than its ancestors.

cdce8p: If we think about this even longer, I'm sure we would be able to come up with further variations.

Personally, I would like to finish this PR soon. It has taken a bit too much time already.
To keep the spirit of @Pierre-Sassoulas last comment https://github.com/PyCQA/pylint/pull/5005#discussion_r710195942, I suggest we keep these three instance as they are now. It works, so why change it. I understand that probably isn't what you were hoping for @nickdrozd, but I believe this will be the easiest (and safest) solution.

--
I'm referring only to those that use child + parent recursively.

nickdrozd:
Okay, I pared it down to just the straightforward cases. Thanks for reviewing!

cdce8p (approved):
LGTM!
Thanks @nickdrozd 🐬

PR #5010, opened 2021-09-15
Fix PyreverseConfig imports in pyreverse's tests

Pierre-Sassoulas:
## Type of Changes

## Description

Fix an import error due to pytest adding all file to the pythonpath before launching tests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -89,6 +89,9 @@ warn_unused_ignores = True
 [mypy-astroid.*]
 ignore_missing_imports = True
 
+[mypy-tests.*]
+ignore_missing_imports = True
+
 [mypy-coverage]
 ignore_missing_imports = True
 
--- a/tests/pyreverse/test_diadefs.py
+++ b/tests/pyreverse/test_diadefs.py
@@ -26,7 +26,7 @@ from typing import Callable, Dict, List, Tuple
 
 import pytest
 from astroid import nodes
-from conftest import PyreverseConfig  # type: ignore #pylint: disable=no-name-in-module
+from tests.pyreverse.conftest import PyreverseConfig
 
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas:
@DudeNr33 fyi, I just realized that pytest is adding all file to the pythonpath so the tests were working but not pylint or mypy :)

DanielNoord:
I don't actually think this fully solved the issue. My IDE can no longer detect `PyreverseConfig` and `Pylance` also complains about this. See screenshots below.
At some point `mypy` is going to complain about this. I would be in favour of reversing this commit.


Schermafbeelding 2021-09-15 om 16 19 43
Schermafbeelding 2021-09-15 om 16 21 48

Without this commit:
Schermafbeelding 2021-09-15 om 16 22 15

Pierre-Sassoulas:
How about we move this class to pylint.pyreverse? It seems it's an interface for optparse.Config that might help with typing for pylint's code too.

DudeNr33:
No, not really - the "real" config class comes from the ``pylint.config.configuration_mixin.ConfigurationMixin``, from which ``pylint.pyreverse.main.Run`` inherits from.
This is just a simplified helper class that represents the config for the tests.
However, we could move it to ``pylint.testutils``.

DanielNoord:
This is actually preventing me from running the test suite locally. I get:
```bash
../../../../.pyenv/versions/3.9.6/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/pyreverse/test_writer.py:31: in <module>
    from tests.pyreverse.conftest import PyreverseConfig
E   ModuleNotFoundError: No module named 'tests'
```
with `tox --recreate  -e py39 `.

PR #5023, opened 2021-09-16
Add typing to `BaseReporter.out`

DanielNoord:

## Type of Changes

| ✓   | 😨 Refactor         |

## Description

Small changes which are needed for further typing of `pylint/reporters`.
I ran the test suite and printed out all types of `BaseReporter.out`. It was only ever `io.TextIO` which also seems fair given what it is doing. However, perhaps somebody with more knowledge of the code base wants to take a look at this.

### DIFF
--- a/pylint/reporters/base_reporter.py
+++ b/pylint/reporters/base_reporter.py
@@ -26,7 +26,7 @@ class BaseReporter:
     def __init__(self, output: TextIO = sys.stdout):
         self.linter: "PyLinter"
         self.section = 0
-        self.out: TextIO = output
+        self.out: TextIO = output or sys.stdout
         self.messages: List[Message] = []
         # Build the path prefix to strip to get relative paths
         self.path_strip_prefix = os.getcwd() + os.sep
--- a/pylint/reporters/base_reporter.py
+++ b/pylint/reporters/base_reporter.py
@@ -3,13 +3,14 @@
 
 import os
 import sys
-from typing import TYPE_CHECKING, List, Optional
+from typing import TYPE_CHECKING, List, Optional, TextIO
 
 from pylint.message import Message
 from pylint.reporters.ureports.nodes import Text
 from pylint.typing import CheckerStats
 
 if TYPE_CHECKING:
+    from pylint.lint.pylinter import PyLinter
     from pylint.reporters.ureports.nodes import Section
 
 
@@ -21,11 +22,10 @@ class BaseReporter:
 
     extension = ""
 
-    def __init__(self, output=None):
-        self.linter = None
+    def __init__(self, output: Optional[TextIO] = None):
+        self.linter: "PyLinter"
         self.section = 0
-        self.out = None
-        self.out_encoding = None
+        self.out: TextIO
         self.set_output(output)
         self.messages: List[Message] = []
         # Build the path prefix to strip to get relative paths
@@ -35,7 +35,7 @@ class BaseReporter:
         """Handle a new message triggered on the current file."""
         self.messages.append(msg)
 
-    def set_output(self, output=None):
+    def set_output(self, output: Optional[TextIO] = None) -> None:
         """set output stream"""
         self.out = output or sys.stdout
 
--- a/pylint/reporters/base_reporter.py
+++ b/pylint/reporters/base_reporter.py
@@ -4,6 +4,7 @@
 import os
 import sys
 from typing import TYPE_CHECKING, List, Optional, TextIO
+from warnings import warn
 
 from pylint.message import Message
 from pylint.reporters.ureports.nodes import Text
@@ -26,7 +27,6 @@ class BaseReporter:
         self.linter: "PyLinter"
         self.section = 0
         self.out: TextIO = output
-        self.set_output(output)
         self.messages: List[Message] = []
         # Build the path prefix to strip to get relative paths
         self.path_strip_prefix = os.getcwd() + os.sep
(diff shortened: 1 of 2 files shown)


tests/test_self.py
```diff
@@ -122,11 +121,11 @@ def _display(self, layout: "Section") -> None:
         pass
 
     @property
-    def out(self) -> StringIO:
+    def out(self) -> TextIO:  # type: ignore # Property is not the same as BaseLinter's attr
```
DanielNoord, line 124: @cdce8p You also commented on this in #4954, see https://github.com/PyCQA/pylint/pull/4954#discussion_r702472666 and https://github.com/PyCQA/pylint/pull/4954#discussion_r702472684
However, this still is an issue. `mypy` maintainers seem to suggest not doing this kind of stuff (see https://github.com/python/mypy/issues/8185). Do we want to refactor (can we?) or is adding the `type: ignore` enough. For some reason for the one below I needed two comments.

cdce8p: I've come across https://github.com/python/mypy/issues/8185 in another project already. Was doing something similar 😅 `type: ignore` would be fine IMO. See also https://github.com/python/mypy/issues/7994 which suggest to use `type: ignore` to silence the error.

```suggestion
    def out(self) -> TextIO:  # type: ignore[override]
```

tests/test_self.py
```diff
@@ -122,11 +121,11 @@ def _display(self, layout: "Section") -> None:
...
         return self._reporters[0].out
 
-    @property
-    def linter(self) -> Optional[Any]:
+    @property  # type: ignore # Property is not the same as BaseLinter's attr
+    def linter(self) -> PyLinter:  # type: ignore
```
cdce8p, line 128: `type: ignore[override]` should only be necessary on the second line.

DanielNoord: I agree, but `mypy` doesn't. Note that we also don't get a "useless suppression" warning. Might want to report this as a bug to `mypy`, but both comments are necessary for `mypy` to pass.

pylint/reporters/base_reporter.py
```diff
@@ -21,11 +22,10 @@ class BaseReporter:
 
     extension = ""
 
-    def __init__(self, output=None):
-        self.linter = None
+    def __init__(self, output: Optional[TextIO] = None):
```
Pierre-Sassoulas, line 25:
```suggestion
    def __init__(self, output: TextIO = sys.stdout):
```

I think we'll have a better time if we get rid of all the optional ``Optional``.

pylint/reporters/base_reporter.py
```diff
@@ -21,11 +22,10 @@ class BaseReporter:
...
+        self.linter: "PyLinter"
         self.section = 0
-        self.out = None
-        self.out_encoding = None
+        self.out: TextIO
```
Pierre-Sassoulas, line 28:
```suggestion
        self.out: TextIO = output
```

pylint/reporters/base_reporter.py
```diff
@@ -35,7 +35,7 @@ def handle_message(self, msg: Message) -> None:
         """Handle a new message triggered on the current file."""
         self.messages.append(msg)
 
-    def set_output(self, output=None):
+    def set_output(self, output: Optional[TextIO] = None) -> None:
         """set output stream"""
         self.out = output or sys.stdout
```
Pierre-Sassoulas, line 40:
```suggestion
    @property
    def out(self) -> TextIO:
        return self.__out


    @out.setter
    def out(self, output: TextIO = sys.stdout) -> None:
        self.__out = output or sys.stdout
        
    def set_output(self, output: Optional[TextIO] = None) -> None:
        """set output stream"""
        warnings.warn("'set_output' will be removed in 3.0, please use reporter.out =stream instead, DeprecationWarning)
        self.out = output
```

Pierre-Sassoulas (changes requested):

pylint/reporters/base_reporter.py
DanielNoord: I don't really understand why we need the `property` and `setter`. Wouldn't adding a depreciation to `set_output` be enough?

Pierre-Sassoulas: You're right, maybe we don't need the ``output or sys.stdout``.

Pierre-Sassoulas: Because what is a false stream anyway ? And why would one assign a false stream to ``reporter.out`` ?

pylint/reporters/base_reporter.py
```diff
@@ -21,11 +22,10 @@ class BaseReporter:
...
 
-    def __init__(self, output=None):
-        self.linter = None
+    def __init__(self, output: TextIO = sys.stdout):
+        self.linter: "PyLinter"
         self.section = 0
-        self.out = None
-        self.out_encoding = None
+        self.out: TextIO = output
         self.set_output(output)
```
Pierre-Sassoulas, line 29:
```suggestion
```

pylint/reporters/base_reporter.py
Pierre-Sassoulas:
```suggestion
    def set_output(self, output: TextIO = sys.stdout) -> None:
        """set output stream"""
        warnings.warn("'set_output' will be removed in 3.0, please use 'reporter.out = stream' instead, DeprecationWarning)
        self.out = output
```

DanielNoord: I don't fully understand your question, but I think this allows extensions to get the output of `pylint` into a different output than `stdout` so they can handle it immediately.
Having this be resolved in the `__init__` is indeed much better, but I wonder why you added the `@property` and `@out.setter`? Keeping `set_output` in with a `DeprecationWarning` seems best to me, as I don't think we fully know how extensions and users might interact with this. You could consider this function to be part of the public API as it is in one of the base classes.

Pierre-Sassoulas: Yes, I agree, I added the property in order to handle the case where output is falsey (so sys.stdout is used instead) to not break compatibility. But if someone is using a falsey stream then if it does not work it's on them.

DanielNoord: See latest commit 😄

Pierre-Sassoulas (approved):

DanielNoord:
Turns out some reporters are initialised while specifically supplying `None`

PR #5028, opened 2021-09-17
Fix typing of `ReportsHandlerMixIn`

DanielNoord:

## Type of Changes

## Description

No better way than to start the Friday morning with a small typing PR.
Small changes to break up #4954.

I moved `ReportsDict` as I felt it fitted much better there. The update to the docstring is because it incorrectly describes what is returned. There is no sorting and there is also no list of reports.

### DIFF
--- a/pylint/reporters/reports_handler_mix_in.py
+++ b/pylint/reporters/reports_handler_mix_in.py
@@ -12,6 +12,8 @@ from pylint.typing import CheckerStats
 if TYPE_CHECKING:
     from pylint.lint.pylinter import PyLinter
 
+ReportsDict = DefaultDict[IChecker, List[Tuple[str, str, Callable]]]
+
 
 class ReportsHandlerMixIn:
     """a mix-in class containing all the reports and stats manipulation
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
Hey, I was trying to reach you by email privately but it seems you're using github's no reply address and are'nt on the discord server either, so let me use this probably not very high traffic issue for a semi private personal message :)

I've been impressed with your tenacity in getting all the typing changes in, as well as the numerous fix in both pylint and astroid for 2.8.0 and 2.11. I was wondering  if you'd be interested in having commit rights to pylint and astroid.

This would allow you a great flexibility in terms of reviewing, triaging or simply getting your code in the repo faster.

Let me know if that would be of any interest to you.

DanielNoord:
Thanks for the compliment! That is really appreciated!

I would gladly accept those rights, although I would be hesitant to use them often. From my latest activity it is also clear that both you and Marc have a much higher knowledge of typing and Python in general. So I would only ever use them after either of you approved a PR. Still, for smaller PRs this would indeed speed up the process!

PR #5041, opened 2021-09-18
Add typing to text reporter

DanielNoord:

## Type of Changes

## Description

Section of #4954

### DIFF
--- a/pylint/reporters/text.py
+++ b/pylint/reporters/text.py
@@ -280,7 +280,7 @@ class ColorizedTextReporter(TextReporter):
                 "In pylint 3.0, the ColoreziedTextReporter will only accept ColorMappingDict as color_mapping parameter",
                 DeprecationWarning,
             )
-            temp_color_mapping = ColorMappingDict()
+            temp_color_mapping: ColorMappingDict = {}
             for key, value in color_mapping.items():
                 color = value[0]
                 style_attrs = tuple(_splitstrip(value[1]))
--- a/pylint/reporters/text.py
+++ b/pylint/reporters/text.py
@@ -26,7 +26,7 @@
 import os
 import sys
 import warnings
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Dict, Optional, Set, TextIO, Tuple
 
 from pylint import utils
 from pylint.interfaces import IReporter
@@ -35,7 +35,10 @@ from pylint.reporters import BaseReporter
 from pylint.reporters.ureports.text_writer import TextWriter
 
 if TYPE_CHECKING:
+    from pylint.lint import PyLinter
     from pylint.reporters.ureports.nodes import Section
+
+COLOR_MAPPING_DICT = Dict[str, Tuple[Optional[str], Optional[str]]]
 
 TITLE_UNDERLINES = ["", "=", "-", "."]
 
@@ -64,22 +67,19 @@ ANSI_COLORS = {
 }
 
 
-def _get_ansi_code(color=None, style=None):
+def _get_ansi_code(color: Optional[str] = None, style: Optional[str] = None) -> str:
     """return ansi escape code corresponding to color and style
 
-    :type color: str or None
     :param color:
       the color name (see `ANSI_COLORS` for available values)
       or the color number when 256 colors are available
 
-    :type style: str or None
     :param style:
       style string (see `ANSI_COLORS` for available values). To get
       several style effects at the same time, use a coma as separator.
 
     :raise KeyError: if an unexistent color or style identifier is given
 
-    :rtype: str
     :return: the built escape code
     """
     ansi_code = []
@@ -98,24 +98,22 @@ def _get_ansi_code(color=None, style=None):
     return ""
 
 
-def colorize_ansi(msg, color=None, style=None):
+def colorize_ansi(
+    msg: str, color: Optional[str] = None, style: Optional[str] = None
+) -> str:
     """colorize message by wrapping it with ansi escape codes
 
-    :type msg: str or unicode
     :param msg: the message string to colorize
 
-    :type color: str or None
     :param color:
       the color identifier (see `ANSI_COLORS` for available values)
 
-    :type style: str or None
     :param style:
       style string (see `ANSI_COLORS` for available values). To get
       several style effects at the same time, use a coma as separator.
 
     :raise KeyError: if an unexistent color or style identifier is given
 
-    :rtype: str or unicode
     :return: the ansi escaped string
     """
     # If both color and style are not defined, then leave the text as is
@@ -136,15 +134,15 @@ class TextReporter(BaseReporter):
     extension = "txt"
     line_format = "{path}:{line}:{column}: {msg_id}: {msg} ({symbol})"
 
-    def __init__(self, output=None):
+    def __init__(self, output: Optional[TextIO] = None) -> None:
         BaseReporter.__init__(self, output)
-        self._modules = set()
+        self._modules: Set[str] = set()
         self._template = self.line_format
 
     def on_set_current_module(self, module: str, filepath: Optional[str]) -> None:
         self._template = str(self.linter.config.msg_template or self._template)
 
-    def write_message(self, msg):
+    def write_message(self, msg: Message) -> None:
         """Convenience method to write a formatted message with class default template"""
         self.writeln(msg.format(self._template))
 
@@ -174,7 +172,7 @@ class ParseableTextReporter(TextReporter):
     name = "parseable"
     line_format = "{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}"
 
-    def __init__(self, output=None):
+    def __init__(self, output: Optional[TextIO] = None) -> None:
         warnings.warn(
             f"{self.name} output format is deprecated. This is equivalent to --msg-template={self.line_format}",
             DeprecationWarning,
@@ -193,7 +191,7 @@ class ColorizedTextReporter(TextReporter):
     """Simple TextReporter that colorizes text output"""
 
     name = "colorized"
-    COLOR_MAPPING = {
+    COLOR_MAPPING: COLOR_MAPPING_DICT = {
         "I": ("green", None),
         "C": (None, "bold"),
         "R": ("magenta", "bold, italic"),
@@ -203,7 +201,11 @@ class ColorizedTextReporter(TextReporter):
         "S": ("yellow", "inverse"),  # S stands for module Separator
     }
 
-    def __init__(self, output=None, color_mapping=None):
+    def __init__(
+        self,
+        output: Optional[TextIO] = None,
+        color_mapping: Optional[COLOR_MAPPING_DICT] = None,
+    ) -> None:
         TextReporter.__init__(self, output)
         self.color_mapping = color_mapping or dict(ColorizedTextReporter.COLOR_MAPPING)
         ansi_terms = ["xterm-16color", "xterm-256color"]
@@ -214,7 +216,7 @@ class ColorizedTextReporter(TextReporter):
 
                 self.out = colorama.AnsiToWin32(self.out)
 
-    def _get_decoration(self, msg_id):
+    def _get_decoration(self, msg_id: str) -> Tuple[Optional[str], Optional[str]]:
         """Returns the tuple color, style associated with msg_id as defined
         in self.color_mapping
         """
@@ -248,7 +250,7 @@ class ColorizedTextReporter(TextReporter):
         self.write_message(msg)
 
 
-def register(linter):
+def register(linter: "PyLinter") -> None:
     """Register the reporter classes with the linter."""
     linter.register_reporter(TextReporter)
     linter.register_reporter(ParseableTextReporter)
--- a/pylint/reporters/text.py
+++ b/pylint/reporters/text.py
@@ -26,7 +26,18 @@
 import os
 import sys
 import warnings
-from typing import TYPE_CHECKING, Dict, NamedTuple, Optional, Set, TextIO, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Dict,
+    NamedTuple,
+    Optional,
+    Set,
+    TextIO,
+    Tuple,
+    Union,
+    cast,
+    overload,
+)
 
 from pylint.interfaces import IReporter
 from pylint.message import Message
@@ -101,6 +112,26 @@ def _get_ansi_code(msg_style: MessageStyle) -> str:
     return ""
 
 
+@overload
+def colorize_ansi(
+    msg: str,
+    msg_style: Optional[MessageStyle] = None,
+) -> str:
+    ...
+
+
+@overload
+def colorize_ansi(
+    msg: str,
+    msg_style: Optional[str] = None,
+    style: Optional[str] = None,
+    *,
+    color: Optional[str] = None,
+) -> str:
+    # Remove for pylint 3.0
+    ...
+
+
 def colorize_ansi(
     msg: str,
     msg_style: Union[MessageStyle, str, None] = None,
@@ -112,6 +143,7 @@ def colorize_ansi(
     :param msg: the message string to colorize
 
     :param msg_style: the message style
+        or color (for backwards compatibility): the color of the message style
 
     :param style: the message's style elements, this will be deprecated
 
@@ -214,6 +246,23 @@ class ColorizedTextReporter(TextReporter):
         "S": MessageStyle("yellow", ("inverse",)),  # S stands for module Separator
     }
 
+    @overload
+    def __init__(
+        self,
+        output: Optional[TextIO] = None,
+        color_mapping: Optional[ColorMappingDict] = None,
+    ) -> None:
+        ...
+
+    @overload
+    def __init__(
+        self,
+        output: Optional[TextIO] = None,
+        color_mapping: Optional[Dict[str, Tuple[Optional[str], Optional[str]]]] = None,
+    ) -> None:
+        # Remove for pylint 3.0
+        ...
+
     def __init__(
         self,
         output: Optional[TextIO] = None,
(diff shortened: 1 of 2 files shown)


pylint/reporters/text.py
```diff
@@ -35,8 +35,11 @@
 from pylint.reporters.ureports.text_writer import TextWriter
 
 if TYPE_CHECKING:
+    from pylint.lint import PyLinter
     from pylint.reporters.ureports.nodes import Section
 
+COLOR_MAPPING_DICT = Dict[str, Tuple[Optional[str], Optional[str]]]
```
DanielNoord, line 41:
```suggestion
ColorMappingDict = Dict[str, Tuple[Optional[str], Optional[str]]]
```

pylint/reporters/text.py
```diff
@@ -193,7 +191,7 @@ class ColorizedTextReporter(TextReporter):
     """Simple TextReporter that colorizes text output"""
 
     name = "colorized"
-    COLOR_MAPPING = {
+    COLOR_MAPPING: COLOR_MAPPING_DICT = {
```
DanielNoord, line 194:
```suggestion
    COLOR_MAPPING: ColorMappingDict = {
```

pylint/reporters/text.py
```diff
@@ -203,7 +201,11 @@ class ColorizedTextReporter(TextReporter):
         "S": ("yellow", "inverse"),  # S stands for module Separator
     }
 
-    def __init__(self, output=None, color_mapping=None):
+    def __init__(
+        self,
+        output: Optional[TextIO] = None,
+        color_mapping: Optional[COLOR_MAPPING_DICT] = None,
```
DanielNoord, line 207:
```suggestion
        color_mapping: Optional[ColorMappingDict] = None,
```

pylint/reporters/text.py
```diff
@@ -35,8 +35,11 @@
 from pylint.reporters.ureports.text_writer import TextWriter
 
 if TYPE_CHECKING:
+    from pylint.lint import PyLinter
     from pylint.reporters.ureports.nodes import Section
 
+ColorMappingDict = Dict[str, Tuple[Optional[str], Optional[str]]]
```
cdce8p, line 41: It might be possible to use a `NamedTuple` for the decorations (instead of `tuple[str | None, str | None]`).

cdce8p, line 41: Additionally, the `style` attribute could be `tuple[str, ...]` instead of `str | None`.
Just need to check if we need to preserve some backwards compatibility.

DanielNoord: Not quite sure how to name the entries of the `NamedTuple` though. The first is obviously `colour`, but the second and third arguments would be something like `style1` and `style2`, right? Or did you have more specific names in mind?

pylint/reporters/text.py
```diff
@@ -35,8 +35,11 @@
 from pylint.reporters.ureports.text_writer import TextWriter
 
 if TYPE_CHECKING:
+    from pylint.lint import PyLinter
     from pylint.reporters.ureports.nodes import Section
 
+ColorMappingDict = Dict[str, Tuple[Optional[str], ...]]
```
cdce8p, line 41: Not exactly what I had in mind 😅
The idea was to change the style definition from `Optional[str]` to `Tuple[str, ...]`. E.g.
```py
# old
("magenta", "bold, italic")

# new
("magenta", ("bold", "italic"))
```

pylint/reporters/text.py
cdce8p: Maybe
```py
class Decoration(NamedTuple):
    color: str | None
    style: tuple[str, ...]
```

pylint/reporters/text.py
DanielNoord: 😂
Let me change that. Then it actually works with a namedtuple as well!

DanielNoord:
Because of the refactor of `MessageStyle` the diff has become quite large. Sorry about that. If you want I can still split, but since Marc already did a first review of the earlier commits I think this should be good.

pylint/reporters/text.py
```diff
@@ -193,17 +190,21 @@ class ColorizedTextReporter(TextReporter):
...
-    COLOR_MAPPING = {
-        "I": ("green", None),
-        "C": (None, "bold"),
-        "R": ("magenta", "bold, italic"),
-        "W": ("magenta", None),
-        "E": ("red", "bold"),
-        "F": ("red", "bold, underline"),
-        "S": ("yellow", "inverse"),  # S stands for module Separator
+    COLOR_MAPPING: ColorMappingDict = {
+        "I": MessageStyle("green", ()),
```
cdce8p, line 194: It might make sense to add `()` as default for `MessageStyle.style`. Only downside, you'll need to import `NamedTuple` from `typing_extensions` instead. Support for default values was only added in `3.6.1`.
https://docs.python.org/3/library/typing.html#typing.NamedTuple

pylint/reporters/text.py
```diff
@@ -193,17 +190,21 @@ class ColorizedTextReporter(TextReporter):
...
+        "C": MessageStyle(None, ("bold",)),
+        "R": MessageStyle("magenta", ("bold", "italic")),
+        "W": MessageStyle("magenta", ()),
```
cdce8p, line 197:
```suggestion
        "W": MessageStyle("magenta"),
```

pylint/reporters/text.py
```diff
@@ -214,41 +215,37 @@ def __init__(self, output=None, color_mapping=None):
...
                 self.out = colorama.AnsiToWin32(self.out)
 
-    def _get_decoration(self, msg_id):
-        """Returns the tuple color, style associated with msg_id as defined
-        in self.color_mapping
-        """
+    def _get_decoration(self, msg_id: str) -> MessageStyle:
+        """Returns the message style as defined in self.color_mapping"""
         try:
             return self.color_mapping[msg_id[0]]
```
cdce8p, line 221: Instead of `try` / `except`, `dict.get` should work as well.

pylint/reporters/text.py
```diff
@@ -214,41 +215,37 @@ def __init__(self, output=None, color_mapping=None):
...
         except KeyError:
-            return None, None
+            return MessageStyle(None, ())
```
cdce8p, line 223:
```suggestion
            return MessageStyle(None)
```

cdce8p (changes requested):
First set of comments. There is some more work required to preserve backwards compatibility. Will leave some more comments for that.

pylint/reporters/text.py
```diff
@@ -64,64 +79,46 @@
...
         else:
-            ansi_code.append(ANSI_COLORS[color])
+            ansi_code.append(ANSI_COLORS[msg_style.colour])
     if ansi_code:
         return ANSI_PREFIX + ";".join(ansi_code) + ANSI_END
     return ""
 
 
-def colorize_ansi(msg, color=None, style=None):
+def colorize_ansi(msg: str, msg_style: MessageStyle) -> str:
```
cdce8p, line 106: This will break backwards compatibility. We might have to use some tricks, but what it comes down to: For the transition phase, users should still be able to use the old function signature.

Only a rough idea:
```py
def colorize_ansi(
    msg: str,
    msg_style: Union[MessageStyle, str, None] = None,  # needs to accept an old 'color' argument
    style: Optional[str] = None,
    **kwargs: Optional[str],  # if 'color' is passed as keyword argument
) -> str:
    if not isinstance(msg_style, MessageStyle):
        warnings.warn("...", DeprecationWarning)
        color = kwargs.get("color")
        style_attrs = tuple(utils._splitstrip(style))
        msg_style = MessageStyle(color or msg_style, style_attrs)
    ...
```

Pierre-Sassoulas (commented):
I never saw color working in pylint, so I wonder if this code is even working in the first place. It makes me want to add colorama to the dependency and refactor.

pylint/reporters/text.py
```diff
@@ -193,17 +190,21 @@ class ColorizedTextReporter(TextReporter):
...
+        "E": MessageStyle("red", ("bold",)),
+        "F": MessageStyle("red", ("bold", "underline")),
+        "S": MessageStyle("yellow", ("inverse",)),  # S stands for module Separator
     }
 
-    def __init__(self, output=None, color_mapping=None):
+    def __init__(
+        self,
+        output: Optional[TextIO] = None,
+        color_mapping: Optional[ColorMappingDict] = None,
```
cdce8p, line 206: Same here `color_mapping` should still accept `Optional[Dict[str, Tuple[Optional[str], Optional[str]]]]`. The conversion should happen in `__init__` when assigning `self.color_mapping`. Together with a deprecation warning.

cdce8p (commented):
To preserve backwards compatibility, these changes would be needed.

DanielNoord:
> I never saw color working in pylint, so I wonder if this code is even working in the first place. It makes me want to add colorama to the dependency and refactor.

Does @cdce8p know anything about this?

pylint/reporters/text.py
DanielNoord: I don't think `NamedTuple` is supported by `typing_extensions`. See their [README](https://github.com/python/typing/blob/master/typing_extensions/README.rst)

cdce8p:
> > I never saw color working in pylint, so I wonder if this code is even working in the first place. It makes me want to add colorama to the dependency and refactor.
> 
> Does @cdce8p know anything about this?

Needed to check myself as I only use the default output. The color output does seem to work when configured
```bash
pylint --output-format colorized
```

pylint/reporters/text.py
cdce8p: 😕 Didn't know that

pylint/reporters/text.py
```diff
@@ -64,64 +79,45 @@
...
-    ansi_code = []
-    if style:
-        style_attrs = utils._splitstrip(style)
-        for effect in style_attrs:
-            ansi_code.append(ANSI_STYLES[effect])
-    if color:
-        if color.isdigit():
+    ansi_code: List[str] = []
+    for effect in msg_style.style:
+        ansi_code.append(ANSI_STYLES[effect])
```
cdce8p, line 93:
```suggestion
    ansi_code: List[str] = [ANSI_STYLES[effect] for effect in msg_style.style]
```

pylint/reporters/text.py
```diff
@@ -193,18 +204,37 @@ class ColorizedTextReporter(TextReporter):
...
+            list(color_mapping.values())[0], MessageStyle
+        ):
+            warnings.warn(
+                "In pylint 3.0, the ColoreziedTextReporter will only accept ColorMappingDict as color_mapping parameter",
+                DeprecationWarning,
+            )
+            for key, value in color_mapping.items():
+                color = value[0]
+                style_attrs = tuple(_splitstrip(value[1]))
+                color_mapping[key] = MessageStyle(color, style_attrs)  # type: ignore
```
cdce8p, line 237: We shouldn't need to add `type: ignore` here. Maybe create a temp dict with type `ColorMappingDict` to assign those to. After the for loop assign the temp one to `color_mapping`.

pylint/reporters/text.py
```diff
@@ -193,18 +204,37 @@ class ColorizedTextReporter(TextReporter):
...
+        color_mapping: Union[
+            ColorMappingDict, Dict[str, Tuple[Optional[str], Optional[str]]], None
+        ] = None,
+    ) -> None:
         TextReporter.__init__(self, output)
+        # pylint: disable-next=fixme
+        # TODO: Remove DeprecationWarning and only accept ColorMappingDict as color_mapping parameter
+        if color_mapping and not isinstance(
+            list(color_mapping.values())[0], MessageStyle
+        ):
```
cdce8p, line 229: Type checker can't narrow the `else` case unfortunately. Something like this should work
```py
if ...:
    ...
else:
    color_mapping = cast(Optional[ColorMappingDict], color_mapping
```

In the end, `self.color_mapping` should be a `ColorMappingDict`.

pylint/reporters/text.py
```diff
@@ -193,18 +204,37 @@ class ColorizedTextReporter(TextReporter):
...
+            warnings.warn(
+                "In pylint 3.0, the ColoreziedTextReporter will only accept ColorMappingDict as color_mapping parameter",
+                DeprecationWarning,
+            )
+            for key, value in color_mapping.items():
+                color = value[0]
+                style_attrs = tuple(_splitstrip(value[1]))
+                color_mapping[key] = MessageStyle(color, style_attrs)  # type: ignore
         self.color_mapping = color_mapping or dict(ColorizedTextReporter.COLOR_MAPPING)
```
cdce8p, line 238: I was wondering why there is the `dict` call here. `self.color_mapping` is written to, but if it should really be safe, maybe `copy` would be a better solution?

pylint/reporters/text.py
```diff
@@ -214,41 +244,37 @@ def __init__(self, output=None, color_mapping=None):
...
-        """
-        try:
-            return self.color_mapping[msg_id[0]]
-        except KeyError:
-            return None, None
+    def _get_decoration(self, msg_id: str) -> MessageStyle:
+        """Returns the message style as defined in self.color_mapping"""
+        style = self.color_mapping.get(msg_id[0])
+        if style:
+            return style  # type: ignore
```
cdce8p, line 251: With `self.color_mapping: ColorMappingDict` this shouldn't be necessary.

pylint/reporters/text.py
```diff
@@ -214,41 +244,37 @@ def __init__(self, output=None, color_mapping=None):
...
+        return MessageStyle(None, ())
```
cdce8p, line 252:
```suggestion
        return self.color_mapping.get(msg_id[0]) or MessageStyle(None, ())
```
That should work

pylint/reporters/text.py
```diff
@@ -64,64 +80,59 @@
...
     return ""
 
 
-def colorize_ansi(msg, color=None, style=None):
+def colorize_ansi(
+    msg: str,
+    msg_style: Union[MessageStyle, str, None] = None,
+    style: Optional[str] = None,
+    **kwargs: Optional[str],
+) -> str:
```
cdce8p, line 109: It would make sense to add overloads here
```py
@overload
def colorize_ansi(
    msg: str,
    msg_style: Optional[MessageStyle] = None,
) -> str:
    ...


@overload
def colorize_ansi(
    msg: str,
    msg_style: Optional[str] = None,
    style: Optional[str] = None,
    *,
    color: Optional[str] = None,
) -> str:
    # Remove for pylint 3.0
    ...
```

pylint/reporters/text.py
```diff
@@ -193,18 +204,37 @@ class ColorizedTextReporter(TextReporter):
...
     }
 
-    def __init__(self, output=None, color_mapping=None):
+    def __init__(
+        self,
+        output: Optional[TextIO] = None,
+        color_mapping: Union[
+            ColorMappingDict, Dict[str, Tuple[Optional[str], Optional[str]]], None
+        ] = None,
+    ) -> None:
```
cdce8p, line 223: Same here
```py
    @overload
    def __init__(
        self,
        output: Optional[TextIO] = None,
        color_mapping: Optional[ColorMappingDict] = None,
    ) -> None:
        ...

    @overload
    def __init__(
        self,
        output: Optional[TextIO] = None,
        color_mapping: Optional[Dict[str, Tuple[Optional[str], Optional[str]]]] = None,
    ) -> None:
        # Remove for pylint 3.0
        ...
```

pylint/reporters/text.py
DanielNoord: I figured I would add a `type: ignore` for the time being and remove after deprecation, but I will change this!

DanielNoord: I tried this, but couldn't get it to work.
With:
```python
if color_mapping and not isinstance(
           color_mapping, ColorMappingDict
        ):
            warnings.warn(
                "In pylint 3.0, the ColoreziedTextReporter will only accept ColorMappingDict as color_mapping parameter",
                DeprecationWarning,
            )
            temp_color_mapping = ColorMappingDict()
            for key, value in color_mapping.items():
                color = value[0]
                style_attrs = tuple(_splitstrip(value[1]))
                temp_color_mapping[key] = MessageStyle(color, style_attrs)
            color_mapping = temp_color_mapping
        self.color_mapping = color_mapping or ColorizedTextReporter.COLOR_MAPPING
```
Schermafbeelding 2021-09-20 om 19 57 32

`color_mapping` keeps getting inferred as possibly being `Dict` even though `color_mapping = temp_color_mapping` makes this impossible. I think the `type: ignore` might be better for the time being.

pylint/reporters/text.py
DanielNoord: No longer an issue now that I use `temp_color_mapping`

pylint/reporters/text.py
```diff
@@ -193,19 +236,59 @@ class ColorizedTextReporter(TextReporter):
...
+        # pylint: disable-next=fixme
+        # TODO: Remove DeprecationWarning and only accept ColorMappingDict as color_mapping parameter
+        if color_mapping and not isinstance(
+            list(color_mapping.values())[0], MessageStyle
+        ):
+            warnings.warn(
+                "In pylint 3.0, the ColoreziedTextReporter will only accept ColorMappingDict as color_mapping parameter",
+                DeprecationWarning,
+            )
+            temp_color_mapping = ColorMappingDict()
```
cdce8p, line 283: Initializing an empty dict with `{}` is a bit faster. For `dict()` there is already a pylint check to detect it. (Not worth to extend it for type aliases though.)
```suggestion
            temp_color_mapping: ColorMappingDict = {}
```

--
Just saw that pylint can't properly deal with it either 😅
```
pylint/reporters/text.py:287:16: E1137: 'temp_color_mapping' does not support item assignment (unsupported-assignment-operation)
```

cdce8p (commented):
Almost there. One last comment

cdce8p (approved):
Looks good!
Thanks @DanielNoord 🚀

DanielNoord:
@cdce8p Thanks for the review 😄

DanielNoord:
I'm so sorry, while squashing I removed the co-author line. Will be more cautious in the future!

cdce8p:
> I'm so sorry, while squashing I removed the co-author line. Will be more cautious in the future!

Don't worry about it! I'm already happy that you squash merged it.

PR #5042, opened 2021-09-18
Rename typing of `global types`

DanielNoord:

## Type of Changes

## Description

While working on #5041 I noticed these did not have the right casing. Global module types should be all caps, right?

### DIFF
--- a/pylint/reporters/reports_handler_mix_in.py
+++ b/pylint/reporters/reports_handler_mix_in.py
@@ -12,7 +12,7 @@ from pylint.typing import CheckerStats
 if TYPE_CHECKING:
     from pylint.lint.pylinter import PyLinter
 
-ReportsDict = DefaultDict[IChecker, List[Tuple[str, str, Callable]]]
+REPORTS_DICT = DefaultDict[IChecker, List[Tuple[str, str, Callable]]]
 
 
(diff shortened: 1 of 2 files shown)


pylint/reporters/reports_handler_mix_in.py
```diff
@@ -12,7 +12,7 @@
 if TYPE_CHECKING:
     from pylint.lint.pylinter import PyLinter
 
-ReportsDict = DefaultDict[IChecker, List[Tuple[str, str, Callable]]]
+REPORTS_DICT = DefaultDict[IChecker, List[Tuple[str, str, Callable]]]
```
cdce8p, line 15: Type aliases are usually PascalCase. So the existing ones are correct.

cdce8p (changes requested):
I would recommend not to move forward with this one.

pylint/reporters/reports_handler_mix_in.py
DanielNoord: My bad, I thought this was only for `TypedDict` and `NamedTuple` as they can use `class` to instantiate. Going to close!

cdce8p:
Just for reference: https://docs.python.org/3/library/typing.html#type-aliases

PR #5043, opened 2021-09-18
Rename `pylint.testutil.Message` to `pylint.testutil.TestMessage`

DanielNoord:

## Type of Changes

## Description

We had two different `Message` classes. This renames one of them to become `OutputMessage` as I was running into conflicts with typing and imports.

### DIFF
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -9,16 +9,16 @@ from pylint.constants import PY38_PLUS
 from pylint.testutils.constants import UPDATE_OPTION
 
 
-class OutputMessage(
+class TestMessage(
     collections.namedtuple(
-        "OutputMessage", ["msg_id", "line", "node", "args", "confidence"]
+        "TestMessage", ["msg_id", "line", "node", "args", "confidence"]
     )
 ):
     def __new__(cls, msg_id, line=None, node=None, args=None, confidence=None):
         return tuple.__new__(cls, (msg_id, line, node, args, confidence))
 
     def __eq__(self, other):
-        if isinstance(other, OutputMessage):
+        if isinstance(other, TestMessage):
             if self.confidence and other.confidence:
                 return super().__eq__(other)
             return self[:-1] == other[:-1]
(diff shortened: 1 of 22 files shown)


pylint/testutils/output_line.py
```diff
@@ -9,14 +9,16 @@
...
 
-class Message(
-    collections.namedtuple("Message", ["msg_id", "line", "node", "args", "confidence"])
+class OutputMessage(
+    collections.namedtuple(
+        "OutputMessage", ["msg_id", "line", "node", "args", "confidence"]
+    )
 ):
     def __new__(cls, msg_id, line=None, node=None, args=None, confidence=None):
         return tuple.__new__(cls, (msg_id, line, node, args, confidence))
```
Pierre-Sassoulas, line 18:
```suggestion
from pylint.message import Message

class TestMessage(Message):

 def __new__(cls, msg_id, line=None, node=None, args=None, confidence=None):
        return tuple.__new__(cls, (msg_id, line, node, args, confidence))
```

```python
 '''
    # super function:
    def __new__(cls, msg_id, symbol, location, msg, confidence):
        return _MsgBase.__new__(
            cls,
            msg_id,
            symbol,
            msg,
            msg_id[0],
            MSG_TYPES[msg_id[0]],
            confidence,
            *location
        )
```
There's ``node`` and args, to add in the child class, and I don't know what ``location`` is exactly but maybe it would make more sense to uniformize this way ? We can put default value for the TestMessage.

DanielNoord: Yeah, let me figure something out!

DanielNoord: I have toyed around for a bit, but I didn't find something that worked.
The problem is that both messages do something different. For example `location` is a tuple of `("abspath", "path", "module", "obj", "line", "colum)`. We could computer the `path` etc. of each `TestMessage` but I don't see how that would provide any benefit other than costing us more time during testing.
At the same time, `args` supplied to `TestMessage` doesn't get handled by `Message` and is needed for the unittests. I suggest renaming `OutputMessage` to `TestMessage` and leave it at that.

Pierre-Sassoulas: Agreed !

Pierre-Sassoulas (approved):

DanielNoord:
@Pierre-Sassoulas Do we want to wait for coveralls to be back up before merging or can we still go ahead? There is no indication about the downtime on their site.

PR #5044, opened 2021-09-20
Fix false-positive undefined-variable with Lambda, IfExp, and :=

cdce8p:
## Type of Changes

## Description

Fix false-positive `undefined-variable` for `y`.

```py
l = lambda x: y if (y := x) else None
```

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1459,7 +1459,7 @@ class VariablesChecker(BaseChecker):
                     # same line as the function definition
                     maybee0601 = False
                 elif (
-                    isinstance(
+                    isinstance(  # pylint: disable=too-many-boolean-expressions
                         defstmt,
                         (
                             nodes.Assign,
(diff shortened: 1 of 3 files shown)


pylint/checkers/variables.py
```diff
@@ -1469,7 +1469,11 @@ def _is_variable_violation(
                             nodes.Return,
                         ),
                     )
-                    and isinstance(defstmt.value, nodes.IfExp)
+                    and (
+                        isinstance(defstmt.value, nodes.IfExp)
+                        or isinstance(defstmt.value, nodes.Lambda)
```
Pierre-Sassoulas, line 1474:
```suggestion
                        isinstance(defstmt.value, (nodes.IfExp, nodes.Lambda)),
```

Pierre-Sassoulas (approved):
Made a small suggestion otherwise 👍

pylint/checkers/variables.py
DanielNoord: Doesn't `and` get precedence over `or`? I think this is checking for:
`IfExp` or (`Lambda` and `IfExp`).

Pierre-Sassoulas: My bad !

PR #5045, opened 2021-09-20
Fix bug with cell-var-from-loop and kw_defaults

### LINKED ISSUE #5012: `cell-var-from-loop` false positive for keyword-only arguments
### Bug description

```python
# pylint: disable=missing-docstring,invalid-name
for i in range(5):
    def f(*, _i=i):
        print(_i)
    f()
    def g(_i=i):
        print(_i)
    g()
```

### Configuration

_No response_

### Command used

```shell
pylint pylint_repro.py
```

### Pylint output

```shell
************* Module pylint_repro
pylint_repro.py:3:16: W0640: Cell variable i defined in loop (cell-var-from-loop)
```

### Expected behavior

I would expect the `cell-var-from-loop` not to be triggered because the value is bound to a keyword argument. The message is only triggered in the presence of the `*` that forces all following arguments to be keyword-only, i.e. it is triggered for `f` but not for `g`.

### Pylint version

```shell
pylint 2.10.2
astroid 2.7.3
Python 3.9.2 (default, Mar 15 2021, 17:11:34)
```

### OS / Environment

Ubuntu 20.04.3 LTS, Python installed with pyenv.

### Additional dependencies

_No response_

david-yz-liu:
## Type of Changes

## Description

The fix in https://github.com/PyCQA/pylint/pull/4827 made use of `pylint.checkers.utils`'s `is_default_argument` function, which currently doesn't handle default values for keyword-only parameters:

https://github.com/PyCQA/pylint/blob/314cb9ffd369b5ab99688c885eb537c901e861eb/pylint/checkers/utils.py#L396-L409

Default values for keyword-only parameters are stored in a separate `args.kw_defaults` attribute.

This led to https://github.com/PyCQA/pylint/issues/5012. I fixed this by modifying the function to check `kw_defaults` as well. Note that since keyword-only parameters can have no default argument, we need to filter out `None` values from this attribute to keep only the astroid nodes representing the actual defaults.

Closes #5012.


pylint/checkers/utils.py
```diff
@@ -402,7 +402,10 @@ def is_default_argument(
     if not scope:
         scope = node.scope()
     if isinstance(scope, (nodes.FunctionDef, nodes.Lambda)):
-        for default_node in scope.args.defaults:
+        all_defaults = scope.args.defaults + [
+            d for d in scope.args.kw_defaults if d is not None
```
Pierre-Sassoulas, line 406: We're calculating the full list of kwargs here (even if the first element of the list would return True) so we could improve performance by using a generator line 408 in ``for default_node in all_defaults:``.

tests/functional/c/cellvar_escaping_loop.py
```diff
@@ -93,6 +93,21 @@ def good_case_issue3107():
         next(filter(lambda j, ix=i[0]: j == ix, [1, 3]))
 
 
+def good_case_issue5102():
+    """Eager binding of cell variable when used as the default value of a keyword-only argument.
+    """
```
Pierre-Sassoulas, line 98:
```suggestion
def good_case_issue5012():
    """Eager binding of cell variable when used as the default value of a keyword-only argument.
    https://github.com/PyCQA/pylint/issues/5012"""
```

pylint/checkers/utils.py
```diff
@@ -402,7 +402,10 @@ def is_default_argument(
     if not scope:
         scope = node.scope()
     if isinstance(scope, (nodes.FunctionDef, nodes.Lambda)):
-        for default_node in scope.args.defaults:
+        all_defaults = scope.args.defaults + [
```
Pierre-Sassoulas, line 405:
```python
    return any(
        default_name_node is node
        for default_name_node in default_node.nodes_of_class(nodes.Name)
        for default_node in scope.args.defaults +  d for d in scope.args.kw_defaults if d is not None
    )
```
Maybe something like this would work ? I'm a little rusty with the imbricated for loop.

Pierre-Sassoulas (changes requested):
Thank you for the fix, I have a minor remark about performance but this is a great addition :)

pylint/checkers/utils.py
david-yz-liu: Nice catch, yeah I was being lazy before. You can't directly concatenate a list and a generator, but I used `itertools.chain` to the same effect.

david-yz-liu:
@Pierre-Sassoulas the "Process test coverage" CI test failed, but I don't think that's because of my changes?

tests/functional/c/cellvar_escaping_loop.py
```diff
@@ -93,6 +93,22 @@ def good_case_issue3107():
         next(filter(lambda j, ix=i[0]: j == ix, [1, 3]))
 
 
+def good_case_issue5102():
```
Pierre-Sassoulas, line 96:
```suggestion
def good_case_issue_5012():
```

pylint/checkers/utils.py
```diff
@@ -402,10 +402,15 @@ def is_default_argument(
     if not scope:
         scope = node.scope()
     if isinstance(scope, (nodes.FunctionDef, nodes.Lambda)):
-        for default_node in scope.args.defaults:
-            for default_name_node in default_node.nodes_of_class(nodes.Name):
-                if default_name_node is node:
-                    return True
+        all_defaults = itertools.chain(
```
Pierre-Sassoulas, line 405: :ok_hand:

Pierre-Sassoulas (approved):
Great ! Don't worry about the coverage, coveralls has been down for a few days we can't do anything about it :)

PR #5046, opened 2021-09-20
Remove `safe_decode()` and `deprecated_option()` from utils

DanielNoord:

## Type of Changes

## Description

While working on the typing of `pylint/utils/utils.py` I noticed these two functions never get called. Therefore they seem like candidates for removal from the codebase.

### DIFF
--- a/pylint/utils/__init__.py
+++ b/pylint/utils/__init__.py
@@ -53,7 +53,6 @@ from pylint.utils.utils import (
     _splitstrip,
     _unquote,
     decoding_stream,
-    deprecated_option,
     diff_string,
     format_section,
     get_global_option,
@@ -62,7 +61,6 @@ from pylint.utils.utils import (
     get_rst_title,
     normalize_text,
     register_plugins,
-    safe_decode,
     tokenize_module,
 )
 
@@ -75,7 +73,6 @@ __all__ = [
     "_splitstrip",
     "_unquote",
     "decoding_stream",
-    "deprecated_option",
     "diff_string",
     "FileState",
     "format_section",
@@ -85,6 +82,5 @@ __all__ = [
     "get_rst_title",
     "normalize_text",
     "register_plugins",
-    "safe_decode",
     "tokenize_module",
 ]
--- a/pylint/utils/utils.py
+++ b/pylint/utils/utils.py
@@ -137,14 +137,6 @@ def get_rst_section(section, options, doc=None):
     return result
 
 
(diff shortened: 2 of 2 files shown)


Pierre-Sassoulas (approved):
Too bad coverall is down, this would have been à pleasant report :D thanks for the cleanup !

cdce8p:
Just a followup, for the future should we deprecate methods first before we remove them?

Pierre-Sassoulas:
Yeah theoretically, I just don't think of ``pylint.utils.utils`` as a public API so imo if it's not used in pylint own code it's not used at all. Maybe I'm wrong and we should have added it to the list of deprecation for 3.0.

DanielNoord:
I was thinking the same as Pierre, but you are indeed right Marc. Sorry about this!

cdce8p:
> Yeah theoretically, I just don't think of `pylint.utils.utils` as a public API so imo if it's not used in pylint own code it's not used at all.

What about `get_global_option`? That is also part of `pylint.utils.utils`. I would consider this at the very least useful for plugins.

> Maybe I'm wrong and we should have added it to the list of deprecation for 3.0.

Regarding the two methods, I do agree that a deprecation probably wasn't necessary.

--
What this makes clear however is that it might be time to have a discussion (or define) which parts of the API are considered public and which private. Especially with the ongoing refactoring effort, this is something to think about. Just as example the changes and deprecations in #5041.

Pierre-Sassoulas:
On the top of my head the public API is:
- Package information (in __pkginfo__.py)
- BaseChecker, DeprecationMixin (in pylint django at least)
- pylint.interfaces
- BaseReporter ?
- pylint.testutil
- Json output for IDE integration
- text output that some tools must be parsing like it's always going to stay that way (including me when I pipe and filter the message to disable when approaching a new project).

I would need to check the major plugin that depends on pylint:
- pylint-django
- prospector
- pylama
- pylint-pycharm (ide integration, see #4741 )
There's probably others but those are the one I know off and probably the main one 😄 

Now that I think of it, the "is it a public API" approach is the wrong one, we should probably ask ourselves what is not instead. And I would say "the small helper functions". The safest way would be to consider everything public API, but then we could never refactor anything without releasing a major and that would make our live really hard, right ?

DanielNoord:
Just to round off this discussion, I think it will be hard to come up with a one solution/rule that fits all cases. We should probably be a little bit more considerate about what not to remove outright, but should not fully constrain ourselves. In the latest refactoring PRs I think we striked the right balance!

PR #5050, opened 2021-09-20
Type `add_message` and add `MessageLocationTuple`

DanielNoord:

## Type of Changes

## Description

This adds typing to `add_message` and adds a new `NamedTuple` to `typing.py`. I add a deprecation warning for `Message`, but I'm not sure if it is even necessary. Is `Message` part of the public API? If not I can remove it.

The `type: ignore` in `tests/testutils/test_output_line.py` will be fixed in a follow-up PR. `OutputLine` needs a little refactor to only accept `int` for `column` and `line`. This will bring it in line with all other code that references `column` or `line`, where those are also `int`.

### DIFF
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -18,7 +18,7 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 import functools
 from inspect import cleandoc
-from typing import Any, Optional, Tuple, Union
+from typing import Any, Optional
 
 from astroid import nodes
 
@@ -115,7 +115,7 @@ class BaseChecker(OptionsProviderMixIn):
         msgid: str,
         line: Optional[int] = None,
         node: Optional[nodes.NodeNG] = None,
-        args: Union[str, Tuple[Union[str, int], ...], None] = None,
+        args: Optional[Any] = None,
         confidence: Optional[Confidence] = None,
         col_offset: Optional[int] = None,
     ) -> None:
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -18,12 +18,14 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 import functools
 from inspect import cleandoc
-from typing import Any
+from typing import Any, Optional, Tuple, Union
+
+from astroid import nodes
 
 from pylint.config import OptionsProviderMixIn
 from pylint.constants import _MSG_ORDER, WarningScope
 from pylint.exceptions import InvalidMessageError
-from pylint.interfaces import UNDEFINED, IRawChecker, ITokenChecker, implements
+from pylint.interfaces import Confidence, IRawChecker, ITokenChecker, implements
 from pylint.message.message_definition import MessageDefinition
 from pylint.typing import CheckerStats
 from pylint.utils import get_rst_section, get_rst_title
@@ -109,10 +111,14 @@ class BaseChecker(OptionsProviderMixIn):
         return result
 
     def add_message(
-        self, msgid, line=None, node=None, args=None, confidence=None, col_offset=None
-    ):
-        if not confidence:
-            confidence = UNDEFINED
+        self,
+        msgid: str,
+        line: Optional[int] = None,
+        node: Optional[nodes.NodeNG] = None,
+        args: Union[str, Tuple[Union[str, int], ...], None] = None,
+        confidence: Optional[Confidence] = None,
+        col_offset: Optional[int] = None,
+    ) -> None:
         self.linter.add_message(msgid, line, node, args, confidence, col_offset)
 
     def check_consistency(self):
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -115,7 +115,7 @@ class BaseChecker(OptionsProviderMixIn):
         msgid: str,
         line: Optional[int] = None,
         node: Optional[nodes.NodeNG] = None,
-        args: Optional[Any] = None,
+        args: Any = None,
         confidence: Optional[Confidence] = None,
         col_offset: Optional[int] = None,
     ) -> None:
--- a/pylint/message/message.py
+++ b/pylint/message/message.py
@@ -3,8 +3,12 @@
 
 
 import collections
+from typing import Optional, Tuple, Union
+from warnings import warn
 
 from pylint.constants import MSG_TYPES
+from pylint.interfaces import Confidence
+from pylint.typing import MessageLocationTuple
 
 _MsgBase = collections.namedtuple(
     "_MsgBase",
@@ -28,7 +32,19 @@ _MsgBase = collections.namedtuple(
 class Message(_MsgBase):
     """This class represent a message to be issued by the reporters"""
 
-    def __new__(cls, msg_id, symbol, location, msg, confidence):
+    def __new__(
+        cls,
+        msg_id: str,
+        symbol: str,
+        location: Union[Tuple[str, str, str, str, int, int], MessageLocationTuple],
+        msg: str,
+        confidence: Optional[Confidence],
+    ):
+        if isinstance(location, tuple):
+            warn(
+                "In pylint 3.0, Messages will only accept a MessageLocationTuple as location parameter",
+                DeprecationWarning,
+            )
         return _MsgBase.__new__(
             cls,
             msg_id,
--- a/pylint/message/message.py
+++ b/pylint/message/message.py
@@ -3,7 +3,7 @@
 
 
 import collections
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, overload
 from warnings import warn
 
 from pylint.constants import MSG_TYPES
@@ -32,6 +32,28 @@ _MsgBase = collections.namedtuple(
 class Message(_MsgBase):
     """This class represent a message to be issued by the reporters"""
 
+    @overload
+    def __new__(
+        cls,
+        msg_id: str,
+        symbol: str,
+        location: MessageLocationTuple,
+        msg: str,
+        confidence: Optional[Confidence],
+    ) -> "Message":
+        ...
+
+    @overload
+    def __new__(
+        cls,
+        msg_id: str,
+        symbol: str,
+        location: Tuple[str, str, str, str, int, int],
+        msg: str,
+        confidence: Optional[Confidence],
+    ) -> "Message":
+        ...
+
     def __new__(
         cls,
         msg_id: str,
@@ -39,8 +61,8 @@ class Message(_MsgBase):
         location: Union[Tuple[str, str, str, str, int, int], MessageLocationTuple],
         msg: str,
         confidence: Optional[Confidence],
-    ):
-        if isinstance(location, tuple):
+    ) -> "Message":
+        if not isinstance(location, MessageLocationTuple):
             warn(
                 "In pylint 3.0, Messages will only accept a MessageLocationTuple as location parameter",
                 DeprecationWarning,
--- a/pylint/message/message.py
+++ b/pylint/message/message.py
@@ -52,6 +52,7 @@ class Message(_MsgBase):
         msg: str,
         confidence: Optional[Confidence],
     ) -> "Message":
+        # Remove for pylint 3.0
         ...
 
     def __new__(
--- a/pylint/message/message_handler_mix_in.py
+++ b/pylint/message/message_handler_mix_in.py
@@ -3,7 +3,7 @@
 
 import sys
 from io import TextIOWrapper
-from typing import TYPE_CHECKING, List, Optional, TextIO, Tuple, Union
+from typing import TYPE_CHECKING, Any, List, Optional, TextIO, Tuple, Union
 
 from astroid import nodes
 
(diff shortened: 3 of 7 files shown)


pylint/message/message_handler_mix_in.py
```diff
@@ -267,14 +279,20 @@ def check_message_definition(message_definition, line, node):
                         f"Message {message_definition.msgid} must provide Node, got None"
                     )
 
-    def add_one_message(
-        self, message_definition, line, node, args, confidence, col_offset
-    ):
+    def add_one_message(  # type: ignore # MessagesHandlesMixIn is always mixed with PyLinter
```
Pierre-Sassoulas, line 282:
```suggestion
    def add_one_message(  # type: ignore # MessagesHandlerMixIn is always mixed with PyLinter
```

pylint/message/message.py
```diff
@@ -28,7 +32,19 @@
...
+        cls,
+        msg_id: str,
+        symbol: str,
+        location: Union[Tuple[str, str, str, str, int, int], MessageLocationTuple],
+        msg: str,
+        confidence: Optional[Confidence],
+    ):
+        if isinstance(location, tuple):
+            warn(
+                "In pylint 3.0, Messages will only accept a MessageLocationTuple as location parameter",
```
Pierre-Sassoulas, line 45: :+1:

pylint/message/message_handler_mix_in.py
```diff
@@ -226,9 +232,15 @@ def is_one_message_enabled(self, msgid, line):
                 return self._msgs_state.get(msgid, fallback)
             return self._msgs_state.get(msgid, True)
 
-    def add_message(
-        self, msgid, line=None, node=None, args=None, confidence=None, col_offset=None
-    ):
+    def add_message(  # type: ignore # MessagesHandlesMixIn is always mixed with PyLinter
+        self: "PyLinter",
+        msgid: str,
```
Pierre-Sassoulas, line 237: `msgid` is a special string with one letter between C W I F R, then 4 digits (C0111)  and `symbol` is a mix of alphabetic character and `-` (missing-docstring). A lot of function in `pylint.message` can takes both but not everything is named correctly right now. You can check `pylint.message.MessageIdStore` & co.


```suggestion
        msgid_or_symbol: MessageId, # MessageId already exists, it seems it would require a refactor or a new type
```

Pierre-Sassoulas (commented):
I like this change. But maybe the msgid/symbol typing is something that should be done prior to this one ?

pylint/message/message_handler_mix_in.py
DanielNoord: You mean create a new type alias called `MessageId` with the specifics discussed in your comment?
I can try and do this, but I think this might create some issues by being too specific. As in, `mypy` will then want us to make sure that every `msgid_or_symbol` conforms to the rules of `msgid_or_symbol` and we might need to add checks all over the code base for this. Still, this is something to try!

Personally I would like to do the following sequence: this PR > `OutputLine` refactor + typing > `msgid_or_symbol` refactor. This will give me the typing of the first two PRs wil working on the `msgid_or_symbol` refactor.

pylint/message/message_handler_mix_in.py
```diff
@@ -226,9 +232,15 @@ def is_one_message_enabled(self, msgid, line):
                 return self._msgs_state.get(msgid, fallback)
             return self._msgs_state.get(msgid, True)
 
-    def add_message(
-        self, msgid, line=None, node=None, args=None, confidence=None, col_offset=None
-    ):
+    def add_message(  # type: ignore # MessagesHandlesMixIn is always mixed with PyLinter
```
DanielNoord, line 235:
```suggestion
    def add_message(  # type: ignore # MessagesHandlerMixIn is always mixed with PyLinter
```

pylint/message/message_handler_mix_in.py
Pierre-Sassoulas: I don't know if mypy can check that a string type is a subset of str. I was thinking of `MessageId` as simply a semantic alias for `str` (The `MessageId` class is both str value (C0111 / missing-docstring) retrieved from the MessageStore using only one of the value and guaranteed to be coherent).

pylint/message/message_handler_mix_in.py
```diff
@@ -267,14 +279,20 @@ def check_message_definition(message_definition, line, node):
...
 
-    def add_one_message(
-        self, message_definition, line, node, args, confidence, col_offset
-    ):
+    def add_one_message(  # type: ignore # MessagesHandlerMixIn is always mixed with PyLinter
+        self: "PyLinter",
+        message_definition: "MessageDefinition",
+        line: Optional[int],
+        node: Optional[nodes.NodeNG],
+        args: Union[str, Tuple[Union[str, int], ...], None],
```
cdce8p, line 287: Is it really limited to only these three types? This is the signature for `str.format`:
```py
def format(self, *args: object, **kwargs: object) -> str: ...
```

pylint/typing.py
```diff
@@ -50,3 +50,13 @@ class ErrorDescriptionDict(TypedDict):
 CheckerStats = Dict[
     str, Union[int, "Counter[str]", List, Dict[str, Union[int, str, Dict[str, int]]]]
 ]
+
+
+# Tuple with information about the location of a to-be-displayed message
```
cdce8p, line 55: Why not use a docstring for it?

tests/message/unittest_message.py
```diff
@@ -1,29 +1,31 @@
...
-                location_value["column"],
-            ],
+            location=(
+                location_value.abspath,
+                location_value.path,
+                location_value.module,
+                location_value.obj,
+                location_value.line,
+                location_value.column,
+            ),
```
cdce8p, line 26: Why is this necessary? Wasn't the whole point so that you could do `location=location_value`?

pylint/message/message_handler_mix_in.py
DanielNoord: Yes, I tested this and `mypy` complains whenever a call has other types for `args`. So I think we cans safely assume this is correct.

tests/message/unittest_message.py
DanielNoord: Sometimes after your reviews I wonder if I was even thinking when submitting a PR... 😅

Pierre-Sassoulas: I missed this one, damn ! 😄

pylint/message/message_handler_mix_in.py
cdce8p: > Not sure if this would actually provide a benefit.

I agree. There are other changes which are much more useful.
So probably nothing we should pursue in the immediate future.

pylint/message/message_handler_mix_in.py
cdce8p: You just need to replace all reference, then it should work.
* `BaseChecker.add_message`
* `MessagesHandlerMixin.add_message`
* `MessagesHandlerMixin.add_one_message`

I would also recommend to use `args: Any` for now. We can always narrow it down later, if we choose to.

tests/testutils/test_output_line.py
```diff
@@ -19,7 +19,7 @@ def inner(confidence: Confidence = HIGH) -> Message:
         return Message(
             symbol="missing-docstring",
             msg_id="C0123",
-            location=[
+            location=[  # type: ignore
```
cdce8p, line 22: This should probably be a `MessageLocationTuple`. Unfortunately, you'll still need `type: ignore` comments for `"line"` and `"column".

DanielNoord: Yes, but I want to change this with an upcoming refactor/typing of `OutputLine`. See the PR description. I can also do that first if you want me to, but I think it might be better to do this first and then do `OutputLine` and its tests.

cdce8p: Sorry. Probably shouldn't review PRs that late 😅

DanielNoord: No worries! As you know it is truly appreciated!

pylint/message/message_handler_mix_in.py
DanielNoord: I don't really understand the first half of your comment. Doesn't this PR already add the `args` typing to all those three places?

I think we can add this typing. `mypy` checks all calls to any of the three functions and checks whether `args` has the correct typing. This was the reason I added `int` to the `Tuple`, `mypy` notified me of that possibility. Perhaps `mypy` is not doing what I think it is, but I believe there is no way for `args` to be anything other than what is now specified in the typing without `mypy` telling us about it.

cdce8p: That would be the case if the codebase was fully type. However `mypy` doesn't check the body of untyped function definitions. So there is the likely case that `Union[str, Tuple[Union[str, int], ...], None]` doesn't cover it.

Additionally, this only covers the our code but what about plugins? Can they for example pass a `float` with `args`.

Thus, I would recommend to replace it with `Any` (at the very least for now).

DanielNoord: I think #4940 made it so all function scopes which contain an `add_message` are typed. But you are right, there is a real possibility that plugins pass other types of `args`. I will change the typing to `Any`.

cdce8p: Sadly, we aren't quite there yet. Adding types there also won't be that easy as it was for #4940. This one here took me < 1min to find:
https://github.com/PyCQA/pylint/blob/b3336c77378f88eb1428d6ff5b6e78bba1177f40/pylint/checkers/base.py#L689-L691

pylint/checkers/base_checker.py
```diff
@@ -109,10 +111,14 @@ def get_full_documentation(self, msgs, options, reports, doc=None, module=None):
...
     def add_message(
-        self, msgid, line=None, node=None, args=None, confidence=None, col_offset=None
-    ):
-        if not confidence:
-            confidence = UNDEFINED
+        self,
+        msgid: str,
+        line: Optional[int] = None,
+        node: Optional[nodes.NodeNG] = None,
+        args: Optional[Any] = None,
```
cdce8p, line 118: `Any` also covers `None`

DanielNoord: Ah, I did this because of this [mypy flag](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional). It didn't mention `Any`. My bad!

cdce8p (approved):
Looks good to me. @Pierre-Sassoulas Would you mind taking another look?

pylint/message/message_handler_mix_in.py
Pierre-Sassoulas: Yes the simple change to do is to keep the string type and rename args ``msgid`` to ``msgid_or_symbol`` (the convention I used in ``pylint.messages``)

pylint/checkers/base_checker.py
```diff
@@ -109,10 +111,14 @@ def get_full_documentation(self, msgs, options, reports, doc=None, module=None):
         return result
 
     def add_message(
-        self, msgid, line=None, node=None, args=None, confidence=None, col_offset=None
-    ):
-        if not confidence:
-            confidence = UNDEFINED
```
Pierre-Sassoulas, line 115: I don't see where the default confidence is set now ? Or did we remove the UNDEFINED value in favor of None ?

pylint/message/message.py
```diff
@@ -28,7 +32,42 @@
...
+        location: Union[Tuple[str, str, str, str, int, int], MessageLocationTuple],
+        msg: str,
+        confidence: Optional[Confidence],
+    ) -> "Message":
+        if not isinstance(location, MessageLocationTuple):
+            warn(
+                "In pylint 3.0, Messages will only accept a MessageLocationTuple as location parameter",
+                DeprecationWarning,
+            )
         return _MsgBase.__new__(
```
Pierre-Sassoulas, line 71:
```suggestion
        confidence = UNDEFINED if confidence is None else confidence
        return _MsgBase.__new__(
```

Pierre-Sassoulas (approved):
Thanks, you're almost single-handedly typing pylint :)

pylint/message/message_handler_mix_in.py
DanielNoord: Couldn't that create problems with extensions explicitly using `msgid`?

pylint/checkers/base_checker.py
DanielNoord: On L116 this functions calls `self.linter.add_message(msgid, line, node, args, confidence, col_offset)` which sets the default confidence here:
https://github.com/PyCQA/pylint/blob/b3336c77378f88eb1428d6ff5b6e78bba1177f40/pylint/message/message_handler_mix_in.py#L229-L246

I thought this was duplicate code.

pylint/message/message_handler_mix_in.py
Pierre-Sassoulas: Yeap, probably a breaking change, let's put in 3.0 or not at all.

pylint/checkers/base_checker.py
cdce8p: Indeed, looks like `lister.add_message` handles that for us already.

pylint/message/message.py
cdce8p: Why set it in the NamedTuple if it's handled by `add_message` already? Am I missing something?

pylint/message/message_handler_mix_in.py
cdce8p: IMO probably not worth doing. It could break too many things.

pylint/message/message.py
DanielNoord: I think this was suggested because Pierre thought I had removed it erroneously.

PR #5051, opened 2021-09-20
Rename `TestMessage` to `MessageTest`

DanielNoord:

## Type of Changes

## Description

The change introduced in #5043 creates numerous `pytest` warnings because the class name begins with `Test`. This causes it to throw this error:
```bash
PytestCollectionWarning: cannot collect test class 'TestingMessage' because it has a __new__ constructor (from: tests/checkers/unittest_deprecated.py)
    class TestingMessage(
```

I don't really like the name `MessageTest` but couldn't come up with anything that doesn't start with `Test`.

### DIFF
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -10,6 +10,9 @@ from pylint.testutils.constants import UPDATE_OPTION
 
 
 class MessageTest(
+    """Used to test messages produced by pylint. Class name cannot start with Test as pytest doesn't allow constructors in test classes.
+    """
+
     collections.namedtuple(
         "MessageTest", ["msg_id", "line", "node", "args", "confidence"]
     )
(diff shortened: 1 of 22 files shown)


pylint/testutils/output_line.py
```diff
@@ -9,16 +9,16 @@
 from pylint.testutils.constants import UPDATE_OPTION
 
 
-class TestMessage(
+class MessageTest(
     collections.namedtuple(
```
Pierre-Sassoulas, line 13:
```suggestion
class MessageTest(
    """This class name cannot start with Test so pytest do not try to launch unittest for it."""

    collections.namedtuple(
```

Pierre-Sassoulas (approved):
Ha, that's unfortunate, thank you for fixing this.

pylint/testutils/output_line.py
DanielNoord:
```suggestion
class MessageTest(
    """Used to test messages produced by pylint. Class name cannot start with Test as pytest doesn't allow constructors in test classes.
    """

    collections.namedtuple(
```
Perhaps this?

Pierre-Sassoulas: Perfect :)

PR #5055, opened 2021-09-20
``Consider-using-f-string`` skips `format()` with backslash

DanielNoord:

## Type of Changes

## Description

As the expression part of f-strings can't include `\` using `format()` or `%` is the preferred way to interpolate `\n` etc.
I ran into this while working on `astroid` and thought I would fix the checker before adding `disable`'s everywhere.


Pierre-Sassoulas:
f-strings can't contain backslashs ?

tests/functional/c/consider/consider_using_f_string.py
```diff
@@ -35,6 +35,11 @@ def print_good():
     print("%(Param_1)s %(Param_2)s" % PARAM_DICT)
     print("%(Param_1)s %(Param_2)s" % return_dict())
     print("{a[Param_1]}{a[Param_2]}".format(a=PARAM_DICT))
+    print("{}".format("\n"))
+    print("{}".format("\n".join(i for i in "string")))
+    print("%s" % "\n")
+    print("%s" % "\n".join(i for i in "string"))
```
Pierre-Sassoulas, line 41:
```suggestion
    print("{}".format("\n"))  # [consider-using-f-string]
    print("{}".format("\n".join(i for i in "string")))
    print("%s" % "\n")  # [consider-using-f-string]
    print("%s" % "\n".join(i for i in "string"))
```
But maybe that would be another checker, "consider-a-string-without-format" ?

Pierre-Sassoulas (commented):
Sorry for the previous comment I misunderstood what you were saying.

tests/functional/c/consider/consider_using_f_string.py
DanielNoord: ? 
See https://stackoverflow.com/questions/44780357/how-to-use-newline-n-in-f-string-to-format-output-in-python-3-6
You can't use \ and thus \n within the expression part. So they shouldn't raise the message.

Pierre-Sassoulas: This looked wrong because it can become ``print("\n")`` but that should definitely be another warning "consider-a-string-without-format" .

Pierre-Sassoulas (approved):

tests/functional/c/consider/consider_using_f_string.py
DanielNoord: Ah yeah. I only kept it minimal to test both direct string interpolation and interpolation of a function containing backlashes.
There does seem to be room for another checker, but that's something for the future 😃

PR #5059, opened 2021-09-21
Add test for ``relative-beyond-top-level`` false positive

DanielNoord:

## Type of Changes

## Description

This builds upon the fix introduced in .

I think running `pylint` on all the files in the test folder might be a bit overkill but I wanted to show my working solution and get feedback first. As said in the `astroid` PR I'm not 100% comfortable with namespace packages and importing, but this seems to fix the issue!


cdce8p:
> > @DanielNoord Can you add the example from [PyCQA/astroid#1200](https://github.com/PyCQA/astroid/issues/1200) here as well?
> 
> Done!

Thanks!

> > Sorry I forgot that we need an entry in the changelog.
> 
> Do we? This was a change in `astroid` without any code changes in `pylint`, only tests. I know for a similar case we didn't add anything.

It's not strictly required, but probably better. Don't know how many users will actually read the `astroid` changelog. Just a one liner + 'closes' should be enough though. It might work to reuse the astroid changelog entry for that. (That's what I did before merging your `requires astroid update` PRs today.)

Pierre-Sassoulas:
I think we'll need to add a regression test for the other issue before 2.12 goes out but we don't have to do it in this PR necesssarily. I don't fear conflicts if we don't merge this immediately though.

DanielNoord:
At some point we might need to check if the changes to the import resolver actually fixed some of the reported issues with `namespace` packages.

PR #5060, opened 2021-09-21
Refactor and typing of OutputLine

DanielNoord:

## Type of Changes

## Description

This brings `OutputLine` in line with the classes dealing with `Message` by making the `line` and `column` attributes `int`. This has increased the size of the diff but I felt that making `line` and `column` `int` throughout the codebase would eventually be worthwhile. `OutputLine` in `csv` form still only consists of `str` (in fact we now make sure this is the case) but whenever we deal with it "internally" we can now assume `line` and `column` are `int`.
The tests have also been changed to accommodate for this.

### DIFF
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -2,10 +2,13 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 import collections
-from typing import Any, NamedTuple
+from typing import Any, Iterable, List, NamedTuple, Optional, Tuple, Union
 
-from pylint import interfaces
+from astroid import nodes
+
 from pylint.constants import PY38_PLUS
+from pylint.interfaces import HIGH, UNDEFINED, Confidence
+from pylint.message.message import Message
 from pylint.testutils.constants import UPDATE_OPTION
 
 
@@ -16,19 +19,30 @@ class MessageTest(
 ):
     """Used to test messages produced by pylint. Class name cannot start with Test as pytest doesn't allow constructors in test classes."""
 
-    def __new__(cls, msg_id, line=None, node=None, args=None, confidence=None):
+    def __new__(
+        cls,
+        msg_id: str,
+        line: Optional[int] = None,
+        node: Optional[nodes.NodeNG] = None,
+        args: Any = None,
+        confidence: Optional[Confidence] = None,
+    ) -> "MessageTest":
         return tuple.__new__(cls, (msg_id, line, node, args, confidence))
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         if isinstance(other, MessageTest):
             if self.confidence and other.confidence:
                 return super().__eq__(other)
-            return self[:-1] == other[:-1]
+            return tuple(self[:-1]) == tuple(other[:-1])
         return NotImplemented  # pragma: no cover
 
 
 class MalformedOutputLineException(Exception):
-    def __init__(self, row, exception):
+    def __init__(
+        self,
+        row: Union[Tuple[str, ...], List[str], str],
+        exception: Exception,
+    ) -> None:
         example = "msg-symbolic-name:42:27:MyClass.my_function:The message"
         other_example = "msg-symbolic-name:7:42::The message"
         expected = [
@@ -62,13 +76,13 @@ Try updating it with: 'python tests/test_functional.py {UPDATE_OPTION}'"""
 class OutputLine(NamedTuple):
     symbol: str
     lineno: int
-    column: str
-    object: Any
+    column: int
+    object: str
     msg: str
-    confidence: interfaces.Confidence
+    confidence: str
 
     @classmethod
-    def from_msg(cls, msg):
+    def from_msg(cls, msg: Message) -> "OutputLine":
         column = cls.get_column(msg.column)
         return cls(
             msg.symbol,
@@ -76,25 +90,28 @@ class OutputLine(NamedTuple):
             column,
             msg.obj or "",
             msg.msg.replace("\r\n", "\n"),
-            msg.confidence.name
-            if msg.confidence != interfaces.UNDEFINED
-            else interfaces.HIGH.name,
+            msg.confidence.name if msg.confidence != UNDEFINED else HIGH.name,
         )
 
     @classmethod
-    def get_column(cls, column):
-        if not PY38_PLUS:
-            return ""  # pragma: no cover
-        return str(column)
+    def get_column(cls, column: str) -> int:
+        if (
+            not PY38_PLUS
+        ):  # This mimicks behaviour of MessagesHandlerMixIn.add_one_message()
+            return 0  # pragma: no cover
+        return int(column)
 
     @classmethod
-    def from_csv(cls, row):
+    def from_csv(cls, row: Union[Tuple[str, ...], List[str], str]) -> "OutputLine":
         try:
-            confidence = row[5] if len(row) == 6 else interfaces.HIGH.name
             column = cls.get_column(row[2])
-            return cls(row[0], int(row[1]), column, row[3], row[4], confidence)
+            if isinstance(row, Iterable) and len(row) == 6:
+                return cls(row[0], int(row[1]), column, row[3], row[4], row[5])
+            if isinstance(row, Iterable) and len(row) == 5:
+                return cls(row[0], int(row[1]), column, row[3], row[4], HIGH.name)
+            raise IndexError
         except Exception as e:
             raise MalformedOutputLineException(row, e) from e
 
-    def to_csv(self):
-        return tuple(self)
+    def to_csv(self) -> Tuple[str, str, str, str, str, str]:
+        return tuple(str(i) for i in self)  # type: ignore # pylint: disable=not-an-iterable
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -2,7 +2,7 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 import collections
-from typing import Any, Iterable, List, NamedTuple, Optional, Tuple, Union
+from typing import Any, Iterable, NamedTuple, Optional, Tuple, Union
 
 from astroid import nodes
 
@@ -95,7 +95,7 @@ class OutputLine(NamedTuple):
 
     @classmethod
     def get_column(cls, column: str) -> int:
-        if not PY38_PLUS:  
+        if not PY38_PLUS:
             # We check the column only for the new better ast parser introduced in python 3.8
             return 0  # pragma: no cover
         return int(column)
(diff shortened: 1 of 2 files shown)


pylint/testutils/output_line.py
```diff
@@ -16,19 +19,30 @@ class MessageTest(
 ):
     """Used to test messages produced by pylint. Class name cannot start with Test as pytest doesn't allow constructors in test classes."""
 
-    def __new__(cls, msg_id, line=None, node=None, args=None, confidence=None):
+    def __new__(
+        cls,
+        msg_id: str,
+        line: Optional[int] = None,
+        node: Optional[nodes.NodeNG] = None,
+        args: Any = None,
```
DanielNoord, line 27: Similar to what we did for `Message` in #5050. This should at some point be fully investigated and changed.

pylint/testutils/output_line.py
```diff
@@ -16,19 +19,30 @@ class MessageTest(
...
+    ) -> "MessageTest":
         return tuple.__new__(cls, (msg_id, line, node, args, confidence))
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         if isinstance(other, MessageTest):
             if self.confidence and other.confidence:
                 return super().__eq__(other)
-            return self[:-1] == other[:-1]
+            return tuple(self[:-1]) == tuple(other[:-1])
```
DanielNoord, line 36: `mypy` complains that on this line the `self` to `__eq__` is no longer a `MessageTest`. Making both sides a tuple solves this issue.

pylint/testutils/output_line.py
```diff
@@ -62,39 +76,42 @@ def __init__(self, row, exception):
 class OutputLine(NamedTuple):
     symbol: str
     lineno: int
-    column: str
-    object: Any
+    column: int
+    object: str
```
DanielNoord, line 80: I have checked the full test suite by printing object in `OutputLine.from_msg()`. This was always `str`, which is to be expected since we are dealing with `csv` and `output lines` here.

pylint/testutils/output_line.py
```diff
@@ -62,39 +76,42 @@ def __init__(self, row, exception):
...
 
     @classmethod
-    def get_column(cls, column):
-        if not PY38_PLUS:
-            return ""  # pragma: no cover
-        return str(column)
+    def get_column(cls, column: str) -> int:
+        if (
+            not PY38_PLUS
+        ):  # This mimicks behaviour of MessagesHandlerMixIn.add_one_message()
```
DanielNoord, line 100: This is referring to the fact that for `not PY38_PLUS` the `column` is always `0`. Perhaps this must be made clearer, or put in a docstring. Let me know what your preferences are.

pylint/testutils/output_line.py
```diff
@@ -62,39 +76,42 @@ def __init__(self, row, exception):
...
+    def from_csv(cls, row: Union[Tuple[str, ...], List[str], str]) -> "OutputLine":
         try:
-            confidence = row[5] if len(row) == 6 else interfaces.HIGH.name
             column = cls.get_column(row[2])
-            return cls(row[0], int(row[1]), column, row[3], row[4], confidence)
+            if isinstance(row, Iterable) and len(row) == 6:
+                return cls(row[0], int(row[1]), column, row[3], row[4], row[5])
+            if isinstance(row, Iterable) and len(row) == 5:
+                return cls(row[0], int(row[1]), column, row[3], row[4], HIGH.name)
+            raise IndexError
```
Pierre-Sassoulas, line 112: Maybe we could uniformize the expected text output by adding the confidence everywhere, so there's only one type of output and we can treat all expected message the same ? (This would prevent big diff where we add the confidence after upgrading the expected output). @cdce8p what do you think, I remember that we had this conversation already at some point.

pylint/testutils/output_line.py
```diff
@@ -62,39 +76,42 @@ def __init__(self, row, exception):
...
         except Exception as e:
             raise MalformedOutputLineException(row, e) from e
 
-    def to_csv(self):
-        return tuple(self)
+    def to_csv(self) -> Tuple[str, str, str, str, str, str]:
+        return tuple(str(i) for i in self)  # type: ignore # pylint: disable=not-an-iterable
```
DanielNoord, line 117: `pylint` finds that `self` is non-iterable, although I think we can assume it is, right?

Furthermore, `mypy` sees the return value as being `Tuple[str, ...]`. I added checks with a temp value and `isinstance(temp_csv, Tuple) and len(temp_csv) == 6`, but it kept denying the correct size. Eventually I opted to add an ignore.

tests/testutils/test_output_line.py
```diff
@@ -56,10 +57,10 @@ def test_output_line_from_message(message: Callable) -> None:
 def test_output_line_to_csv(confidence: Confidence, message: Callable) -> None:
     output_line = OutputLine.from_msg(message(confidence))
     csv = output_line.to_csv()
-    expected_column = "column" if PY38_PLUS else ""
+    expected_column = "2" if PY38_PLUS else "0"
     assert csv == (
         "missing-docstring",
-        "line",
+        "1",
```
DanielNoord, line 63: Note that we are checking the `to_csv()` method here, so `line` and `column` should be `str` again.

pylint/testutils/output_line.py
Pierre-Sassoulas: I think not taking the column into account is only for tests though ? Because below python 3.8 the ast information are different (and inferior) so we're checking only above python 3.8.

DanielNoord: I think you are right! My comment was referring to perhaps making the comment a little bit clearer, although I don't immediately know how.

Pierre-Sassoulas:
```suggestion
        ):  # We check the column only for the new better parser introduced in python 3.8
```

DanielNoord:
```suggestion
        ):  # We check the column only for the new better ast parser introduced in python 3.8
```
😄

pylint/testutils/output_line.py
DanielNoord: This should definitely be part of a separate PR and would then ideally follow after this. If desired I can go and create this. Should be as easy as running `python tests/test_functional.py --update-functional-output` and updating the typing and this function here.

cdce8p: Just my personal opinion: I don't see much value in adding `HIGH` to almost every pylint error in all functional tests. Especially since it's the last value, it would be much more useful to only specify it if it's different (i.e. not `HIGH`).

pylint/testutils/output_line.py
```diff
@@ -62,39 +76,42 @@ def __init__(self, row, exception):
...
+        return int(column)
 
     @classmethod
-    def from_csv(cls, row):
+    def from_csv(cls, row: Union[Tuple[str, ...], List[str], str]) -> "OutputLine":
         try:
-            confidence = row[5] if len(row) == 6 else interfaces.HIGH.name
             column = cls.get_column(row[2])
-            return cls(row[0], int(row[1]), column, row[3], row[4], confidence)
+            if isinstance(row, Iterable) and len(row) == 6:
```
cdce8p, line 108: `isinstance(row, Iterable)` doesn't test if `row` supports `__len__`. You need at least a `Collection` or `Sequence` for that. Since you also use `__getitem__` -> `Sequence`.

cdce8p, line 108: Why not do the `isinstance` check once at the beginning (before the `column` assignment) and move everything else inside the if block?

pylint/testutils/output_line.py
```diff
@@ -16,19 +19,30 @@ class MessageTest(
...
 
-    def __new__(cls, msg_id, line=None, node=None, args=None, confidence=None):
+    def __new__(
+        cls,
+        msg_id: str,
+        line: Optional[int] = None,
+        node: Optional[nodes.NodeNG] = None,
+        args: Any = None,
+        confidence: Optional[Confidence] = None,
+    ) -> "MessageTest":
```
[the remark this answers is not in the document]
DanielNoord: Do you want to block this PR before consensus is reached there?

pylint/testutils/output_line.py
DanielNoord: While I agree with the no added value argument, I think Pierre makes a valid point that when we do this once we will then have less larger diffs after people use the `--update-functional-output` to update tests after they have added new cases. I saw this during my own work as well, where all test data gets updated because of it.
I don't think it would hurt to do this once and would allow simplifying the code here.

pylint/testutils/output_line.py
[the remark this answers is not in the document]
cdce8p: Don't think that would make sense here. Even if we reach a decision, the version bump would likely be in `2.12` at the earliest.

pylint/testutils/output_line.py
cdce8p: We can update / revert `--update-functional-output` to only add different confidence values 😜 Then we would only need to update a few.

Anyway, doesn't matter the decision, this should be part of another PR. (I saw a question about that earlier.)

pylint/testutils/output_line.py
```diff
@@ -62,39 +76,42 @@ def __init__(self, row, exception):
...
 
     @classmethod
-    def from_csv(cls, row):
+    def from_csv(cls, row: Union[Sequence[str], str]) -> "OutputLine":
         try:
-            confidence = row[5] if len(row) == 6 else interfaces.HIGH.name
-            column = cls.get_column(row[2])
-            return cls(row[0], int(row[1]), column, row[3], row[4], confidence)
+            column = cls._get_column(row[2])
+            if isinstance(row, Sequence):
```
cdce8p, line 107:
```suggestion
            if isinstance(row, Sequence):
                column = cls._get_column(row[2])
```
Why not?

pylint/testutils/output_line.py
```diff
@@ -62,39 +76,42 @@ def __init__(self, row, exception):
...
         try:
-            confidence = row[5] if len(row) == 6 else interfaces.HIGH.name
-            column = cls.get_column(row[2])
-            return cls(row[0], int(row[1]), column, row[3], row[4], confidence)
+            if isinstance(row, Sequence):
+                column = cls._get_column(row[2])
+                if len(row) == 6:
+                    return cls(row[0], int(row[1]), column, row[3], row[4], row[5])
+                if len(row) == 5:
+                    return cls(row[0], int(row[1]), column, row[3], row[4], HIGH.name)
```
cdce8p, line 111:
```suggestion
                if len(row) == 5:
                    return cls(row[0], int(row[1]), column, row[3], row[4], HIGH.name)
                if len(row) == 6:
                    return cls(row[0], int(row[1]), column, row[3], row[4], row[5])
```
Doesn't really matter, but I think in order looks a bit better.

cdce8p (commented):
Last comment from me. I'll leave the rest to you @Pierre-Sassoulas

Pierre-Sassoulas (approved):
👍

pylint/testutils/output_line.py
DanielNoord: @cdce8p btw, this is a false positive right? Or am I missing some edge case where this is indeed a valid error?

PR #5062, opened 2021-09-22
Fix loop in too-many-ancestors when an inheritance cycle is inferred

### LINKED ISSUE #4972: 2.10.2 Hangs forever, downgrade to 2.9.6 is working
### Bug description

```python
Installing last version (2.10.2) results in unexpected hangs.
Downgrading to 2.9.6 works as expected.
```

### Configuration

```ini
[MASTER]

# A comma-separated list of package or module names from where C extensions may
# be loaded. Extensions are loading into the active Python interpreter and may
# run arbitrary code.
extension-pkg-whitelist=pycurl,setproctitle

# Specify a score threshold to be exceeded before program exits with error.
fail-under=10.0

# Add files or directories to the blacklist. They should be base names, not
# paths.
ignore=CVS

# Add files or directories matching the regex patterns to the blacklist. The
# regex matches against base names, not paths.
ignore-patterns=

# Python code to execute, usually for sys.path manipulation such as
# pygtk.require().
# init-hook='import sys; sys.path.append(".")'

# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
# number of processors available to use.
jobs=0

# Control the amount of potential inferred values when inferring a single
# object. This can help the performance when dealing with large functions or
# complex, nested conditions.
limit-inference-results=100

# List of plugins (as comma separated values of python module names) to load,
# usually to register additional checkers.
load-plugins=

# Pickle collected data for later comparisons.
persistent=no

# When enabled, pylint would attempt to guess common misconfiguration and emit
# user-friendly hints instead of false-positive error messages.
suggestion-mode=yes

# Allow loading of arbitrary C extensions. Extensions are imported into the
# active Python interpreter and may run arbitrary code.
unsafe-load-any-extension=no


[MESSAGES CONTROL]

# Only show warnings with the listed confidence levels. Leave empty to show
# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED.
confidence=

# Disable the message, report, category or checker with the given id(s). You
# can either give multiple identifiers separated by comma (,) or put this
# option multiple times (only on the command line, not in the configuration
# file where it should appear only once). You can also use "--disable=all" to
# disable everything first and then reenable specific checks. For example, if
# you want to run only the similarities checker, you can use "--disable=all
# --enable=similarities". If you want to run only the classes checker, but have
# no Warning level messages displayed, use "--disable=all --enable=classes
# --disable=W".
disable=print-statement,
        parameter-unpacking,
        unpacking-in-except,
        old-raise-syntax,
        backtick,
        long-suffix,
        old-ne-operator,
        old-octal-literal,
        import-star-module-level,
        non-ascii-bytes-literal,
        raw-checker-failed,
        bad-inline-option,
        locally-disabled,
        file-ignored,
        suppressed-message,
        useless-suppression,
        deprecated-pragma,
        use-symbolic-message-instead,
        apply-builtin,
        basestring-builtin,
        buffer-builtin,
        cmp-builtin,
        coerce-builtin,
        execfile-builtin,
        file-builtin,
        long-builtin,
        raw_input-builtin,
        reduce-builtin,
        standarderror-builtin,
        unicode-builtin,
        xrange-builtin,
        coerce-method,
        delslice-method,
        getslice-method,
        setslice-method,
        no-absolute-import,
        old-division,
        dict-iter-method,
        dict-view-method,
        next-method-called,
        metaclass-assignment,
        indexing-exception,
        raising-string,
        reload-builtin,
        oct-method,
        hex-method,
        nonzero-method,
        cmp-method,
        input-builtin,
        round-builtin,
        intern-builtin,
        unichr-builtin,
        map-builtin-not-iterating,
        zip-builtin-not-iterating,
        range-builtin-not-iterating,
        filter-builtin-not-iterating,
        using-cmp-argument,
        eq-without-hash,
        div-method,
        idiv-method,
        rdiv-method,
        exception-message-attribute,
        invalid-str-codec,
        sys-max-int,
        bad-python3-import,
        deprecated-string-function,
        deprecated-str-translate-call,
        deprecated-itertools-function,
        deprecated-types-field,
        next-method-defined,
        dict-items-not-iterating,
        dict-keys-not-iterating,
        dict-values-not-iterating,
        deprecated-operator-function,
        deprecated-urllib-function,
        xreadlines-attribute,
        deprecated-sys-function,
        exception-escape,
        comprehension-escape,
        too-many-branches,
        too-many-statements,
        C0114, # file docstring
        too-few-public-methods, # useless
        missing-module-docstring # useless too

# Enable the message, report, category or checker with the given id(s). You can
# either give multiple identifier separated by comma (,) or put this option
# multiple time (only on the command line, not in the configuration file where
# it should appear only once). See also the "--disable" option for examples.
enable=c-extension-no-member


[REPORTS]

# Python expression which should return a score less than or equal to 10. You
# have access to the variables 'error', 'warning', 'refactor', and 'convention'
# which contain the number of messages in each category, as well as 'statement'
# which is the total number of statements analyzed. This score is used by the
# global evaluation report (RP0004).
evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / statement) * 10)

# Template used to display messages. This is a python new-style format string
# used to format the message information. See doc for all details.
#msg-template=

# Set the output format. Available formats are text, parseable, colorized, json
# and msvs (visual studio). You can also give a reporter class, e.g.
# mypackage.mymodule.MyReporterClass.
output-format=text

# Tells whether to display a full report or only the messages.
reports=no

# Activate the evaluation score.
score=yes


[REFACTORING]

# Maximum number of nested blocks for function / method body
max-nested-blocks=10

# Complete name of functions that never returns. When checking for
# inconsistent-return-statements if a never returning function is called then
# it will be considered as an explicit return statement and no message will be
# printed.
never-returning-functions=sys.exit


[FORMAT]

# Expected format of line ending, e.g. empty (any line ending), LF or CRLF.
expected-line-ending-format=

# Regexp for a line that is allowed to be longer than the limit.
ignore-long-lines=^\s*(# )?<?https?://\S+>?$

# Number of spaces of indent required inside a hanging or continued line.
indent-after-paren=4

# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
# tab).
indent-string='    '

# Maximum number of characters on a single line.
max-line-length=120

# Maximum number of lines in a module.
max-module-lines=1000

# Allow the body of a class to be on the same line as the declaration if body
# contains single statement.
single-line-class-stmt=no

# Allow the body of an if to be on the same line as the test if there is no
# else.
single-line-if-stmt=no


[SIMILARITIES]

# Ignore comments when computing similarities.
ignore-comments=yes

# Ignore docstrings when computing similarities.
ignore-docstrings=yes

# Ignore imports when computing similarities.
ignore-imports=yes

# Minimum lines number of a similarity.
min-similarity-lines=10


[MISCELLANEOUS]

# List of note tags to take in consideration, separated by a comma.
notes=FIXME,
      XXX,
      TODO

# Regular expression of note tags to take in consideration.
#notes-rgx=


[LOGGING]

# The type of string formatting that logging methods do. `old` means using %
# formatting, `new` is for `{}` formatting.
logging-format-style=old

# Logging modules to check that the string format arguments are in logging
# function parameter format.
logging-modules=logging


[STRING]

# This flag controls whether inconsistent-quotes generates a warning when the
# character used as a quote delimiter is used inconsistently within a module.
check-quote-consistency=no

# This flag controls whether the implicit-str-concat should generate a warning
# on implicit string concatenation in sequences defined over several lines.
check-str-concat-over-line-jumps=no


[VARIABLES]

# List of additional names supposed to be defined in builtins. Remember that
# you should avoid defining new builtins when possible.
additional-builtins=

# Tells whether unused global variables should be treated as a violation.
allow-global-unused-variables=yes

# List of strings which can identify a callback function by name. A callback
# name must start or end with one of those strings.
callbacks=cb_,
          _cb

# A regular expression matching the name of dummy variables (i.e. expected to
# not be used).
dummy-variables-rgx=_+$|(_[a-zA-Z0-9_]*[a-zA-Z0-9]+?$)|dummy|^ignored_|^unused_

# Argument names that match this expression will be ignored. Default to name
# with leading underscore.
ignored-argument-names=_.*|^ignored_|^unused_

# Tells whether we should check for unused import in __init__ files.
init-import=no

# List of qualified module names which can have objects that can redefine
# builtins.
redefining-builtins-modules=six.moves,past.builtins,future.builtins,builtins,io


[BASIC]

# Naming style matching correct argument names.
argument-naming-style=snake_case

# Regular expression matching correct argument names. Overrides argument-
# naming-style.
#argument-rgx=

# Naming style matching correct attribute names.
attr-naming-style=snake_case

# Regular expression matching correct attribute names. Overrides attr-naming-
# style.
#attr-rgx=

# Bad variable names which should always be refused, separated by a comma.
bad-names=foo,
          bar,
          baz,
          toto,
          tutu,
          tata

# Bad variable names regexes, separated by a comma. If names match any regex,
# they will always be refused
bad-names-rgxs=

# Naming style matching correct class attribute names.
class-attribute-naming-style=any

# Regular expression matching correct class attribute names. Overrides class-
# attribute-naming-style.
#class-attribute-rgx=

# Naming style matching correct class names.
class-naming-style=PascalCase

# Regular expression matching correct class names. Overrides class-naming-
# style.
#class-rgx=

# Naming style matching correct constant names.
const-naming-style=UPPER_CASE

# Regular expression matching correct constant names. Overrides const-naming-
# style.
#const-rgx=

# Minimum line length for functions/classes that require docstrings, shorter
# ones are exempt.
docstring-min-length=-1

# Naming style matching correct function names.
function-naming-style=snake_case

# Regular expression matching correct function names. Overrides function-
# naming-style.
#function-rgx=

# Good variable names which should always be accepted, separated by a comma.
good-names=i,
           j,
           k,
           ex,
           Run,
           _

# Good variable names regexes, separated by a comma. If names match any regex,
# they will always be accepted
good-names-rgxs=

# Include a hint for the correct naming format with invalid-name.
include-naming-hint=no

# Naming style matching correct inline iteration names.
inlinevar-naming-style=any

# Regular expression matching correct inline iteration names. Overrides
# inlinevar-naming-style.
#inlinevar-rgx=

# Naming style matching correct method names.
method-naming-style=snake_case

# Regular expression matching correct method names. Overrides method-naming-
# style.
#method-rgx=

# Naming style matching correct module names.
module-naming-style=snake_case

# Regular expression matching correct module names. Overrides module-naming-
# style.
#module-rgx=

# Colon-delimited sets of names that determine each other's naming style when
# the name regexes allow several styles.
name-group=

# Regular expression which should only match function or class names that do
# not require a docstring.
no-docstring-rgx=^_

# List of decorators that produce properties, such as abc.abstractproperty. Add
# to this list to register other decorators that produce valid properties.
# These decorators are taken in consideration only for invalid-name.
property-classes=abc.abstractproperty

# Naming style matching correct variable names.
variable-naming-style=snake_case

# Regular expression matching correct variable names. Overrides variable-
# naming-style.
#variable-rgx=


[SPELLING]

# Limits count of emitted suggestions for spelling mistakes.
max-spelling-suggestions=4

# Spelling dictionary name. Available dictionaries: none. To make it work,
# install the python-enchant package.
spelling-dict=

# List of comma separated words that should not be checked.
spelling-ignore-words=

# A path to a file that contains the private dictionary; one word per line.
spelling-private-dict-file=

# Tells whether to store unknown words to the private dictionary (see the
# --spelling-private-dict-file option) instead of raising a message.
spelling-store-unknown-words=no


[TYPECHECK]

# List of decorators that produce context managers, such as
# contextlib.contextmanager. Add to this list to register other decorators that
# produce valid context managers.
contextmanager-decorators=contextlib.contextmanager

# List of members which are set dynamically and missed by pylint inference
# system, and so shouldn't trigger E1101 when accessed. Python regular
# expressions are accepted.
generated-members=

# Tells whether missing members accessed in mixin class should be ignored. A
# mixin class is detected if its name ends with "mixin" (case insensitive).
ignore-mixin-members=yes

# Tells whether to warn about missing members when the owner of the attribute
# is inferred to be None.
ignore-none=yes

# This flag controls whether pylint should warn about no-member and similar
# checks whenever an opaque object is returned when inferring. The inference
# can return multiple potential results while evaluating a Python object, but
# some branches might not be evaluated, which results in partial inference. In
# that case, it might be useful to still emit no-member and other checks for
# the rest of the inferred objects.
ignore-on-opaque-inference=yes

# List of class names for which member attributes should not be checked (useful
# for classes with dynamically set attributes). This supports the use of
# qualified names.
ignored-classes=optparse.Values,thread._local,_thread._local

# List of module names for which member attributes should not be checked
# (useful for modules/projects where namespaces are manipulated during runtime
# and thus existing member attributes cannot be deduced by static analysis). It
# supports qualified module names, as well as Unix pattern matching.
ignored-modules=

# Show a hint with possible names when a member name was not found. The aspect
# of finding the hint is based on edit distance.
missing-member-hint=yes

# The minimum edit distance a name should have in order to be considered a
# similar match for a missing member name.
missing-member-hint-distance=1

# The total number of similar names that should be taken in consideration when
# showing a hint for a missing member.
missing-member-max-choices=1

# List of decorators that change the signature of a decorated function.
signature-mutators=


[DESIGN]

# Maximum number of arguments for function / method.
max-args=10

# Maximum number of attributes for a class (see R0902).
max-attributes=10

# Maximum number of boolean expressions in an if statement (see R0916).
max-bool-expr=10

# Maximum number of branch for function / method body.
max-branches=12

# Maximum number of locals for function / method body.
max-locals=20

# Maximum number of parents for a class (see R0901).
max-parents=7

# Maximum number of public methods for a class (see R0904).
max-public-methods=20

# Maximum number of return / yield for function / method body.
max-returns=6

# Maximum number of statements in function / method body.
max-statements=50

# Minimum number of public methods for a class (see R0903).
min-public-methods=2


[IMPORTS]

# List of modules that can be imported at any level, not just the top level
# one.
allow-any-import-level=

# Allow wildcard imports from modules that define __all__.
allow-wildcard-with-all=no

# Analyse import fallback blocks. This can be used to support both Python 2 and
# 3 compatible code, which means that the block might have code that exists
# only in one or another interpreter, leading to false positives when analysed.
analyse-fallback-blocks=no

# Deprecated modules which should not be used, separated by a comma.
deprecated-modules=optparse,tkinter.tix

# Create a graph of external dependencies in the given file (report RP0402 must
# not be disabled).
ext-import-graph=

# Create a graph of every (i.e. internal and external) dependencies in the
# given file (report RP0402 must not be disabled).
import-graph=

# Create a graph of internal dependencies in the given file (report RP0402 must
# not be disabled).
int-import-graph=

# Force import order to recognize a module as part of the standard
# compatibility libraries.
known-standard-library=

# Force import order to recognize a module as part of a third party library.
known-third-party=enchant

# Couples of modules and preferred modules, separated by a comma.
preferred-modules=


[CLASSES]

# List of method names used to declare (i.e. assign) instance attributes.
defining-attr-methods=__init__,
                      __new__,
                      setUp,
                      __post_init__

# List of member names, which should be excluded from the protected access
# warning.
exclude-protected=_asdict,
                  _fields,
                  _replace,
                  _source,
                  _make

# List of valid names for the first argument in a class method.
valid-classmethod-first-arg=cls

# List of valid names for the first argument in a metaclass class method.
valid-metaclass-classmethod-first-arg=cls


[EXCEPTIONS]

# Exceptions that will emit a warning when being caught. Defaults to
# "BaseException, Exception".
overgeneral-exceptions=BaseException,
                       Exception
```

### Command used

```shell
pylint
```

### Pylint output

```shell
$ pylint --rcfile=.pylintrc myproject/ tests/
```

### Expected behavior

should passed

### Pylint version

```shell
2.10.2
```

### OS / Environment

Centos 7.9
Python 3.8.6

### Additional dependencies

amqp==5.0.6
aniso8601==9.0.1
astroid==2.7.3
attrs==21.2.0
billiard==3.6.4.0
black==21.8b0
cached-property==1.5.2
celery==5.1.2
celery-redbeat==2.0.0
certifi==2021.5.30
charset-normalizer==2.0.4
click==7.1.2
click-didyoumean==0.0.3
click-plugins==1.1.1
click-repl==0.2.0
dataclasses==0.8
dnspython==2.1.0
dukpy==0.2.3
fakeredis==1.6.0
Flask==2.0.1
flask-restx==0.5.1
gevent==21.8.0
greenlet==1.1.1
idna==3.2
importlib-metadata==4.8.1
iniconfig==1.1.1
isort==5.9.3
itsdangerous==2.0.1
Jinja2==3.0.1
jsonschema==3.2.0
kamene==0.32
kombu==5.1.0
lazy-object-proxy==1.6.0
ldap3==2.9.1
MarkupSafe==2.0.1
mccabe==0.6.1
mypy-extensions==0.4.3
packaging==21.0
pathspec==0.9.0
platformdirs==2.3.0
pluggy==0.13.1
prompt-toolkit==3.0.20
py==1.10.0
pyasn1==0.4.8
pycurl==7.44.1
pylint==2.10.2
pyodbc==4.0.32
PyPAC==0.15.0
pyparsing==2.4.7
pyrsistent==0.18.0
pytest==6.2.4
pytest-pylint==0.18.0
pytest-sugar==0.9.4
python-dateutil==2.8.2
python-dotenv==0.19.0
pytz==2021.1
PyYAML==5.4.1
redis==3.5.3
regex==2021.8.28
requests==2.26.0
setproctitle==1.2.2
six==1.16.0
sortedcontainers==2.4.0
tenacity==8.0.1
termcolor==1.1.0
tld==0.12.6
toml==0.10.2
tomli==1.2.1
typed-ast==1.4.3
typing-extensions==3.10.0.2
urllib3==1.26.6
uWSGI==2.0.19.1
vine==5.0.0
wcwidth==0.2.5
Werkzeug==2.0.1
wrapt==1.12.1
xxhash==2.0.2
zipp==3.5.0
zope.event==4.5.0
zope.interface==5.4.0
zstandard==0.15.2

nelfin:
## Type of Changes

## Description

If a class could be inferred as an ancestor of itself then
the implementation of _get_parents could get caught in an infinite loop,
forever adding itself to to_explore. This could be possible by
monkeypatching an ancestor after class definition like in the following
simplified example:

```python
    class Fake(module.Cls):
        pass

    module.Cls = Fake
```

Reproducing this is pretty tricky, but this integration test shows the behaviour on current `main`:

```sh
#!/bin/sh

HERE=$(readlink -f .)
VENV=$HERE/venv-repro
PIP=$VENV/bin/pip

python -m venv "$VENV"

PYLINT=$VENV/bin/pylint

# assume running in pylint dir
$PIP install -e .
$PIP install flask

mkdir -p repro/flask/
touch repro/__init__.py
cat > repro/flask/__init__.py <<'EOF'
import flask
import repro.flask  # self-import necessary

class Fake(flask.Flask):
    pass

flask.Flask = Fake
EOF

echo +++ this is fine
$PYLINT --rcfile=/dev/null -- repro/flask/
echo +++ finished

echo +++ this loops forever
cd repro/; $PYLINT --rcfile=/dev/null -- flask/
```
--

Closes #4972

### DIFF
--- a/pylint/checkers/design_analysis.py
+++ b/pylint/checkers/design_analysis.py
@@ -267,6 +267,12 @@ def _get_parents_iter(
         if parent.qname() in ignored_parents:
             continue
         if parent not in parents:
+            # This guard might appear to be performing the same function as
+            # adding the resolved parents to a set to eliminate duplicates
+            # (legitimate due to diamond inheritance patterns), but its
+            # additional purpose is to prevent cycles (not normally possible,
+            # but potential due to inference) and thus guarantee termination
+            # of the while-loop
             yield parent
             parents.add(parent)
             to_explore.extend(parent.ancestors(recurs=False))
--- a/pylint/checkers/design_analysis.py
+++ b/pylint/checkers/design_analysis.py
@@ -28,7 +28,7 @@
 
 import re
 from collections import defaultdict
-from typing import FrozenSet, List, Set, cast
+from typing import FrozenSet, Iterator, List, Set, cast
 
 import astroid
 from astroid import nodes
@@ -242,9 +242,9 @@ def _count_methods_in_class(node):
     return all_methods
 
 
-def _get_parents(
+def _get_parents_iter(
     node: nodes.ClassDef, ignored_parents: FrozenSet[str]
-) -> Set[nodes.ClassDef]:
+) -> Iterator[nodes.ClassDef]:
     r"""Get parents of ``node``, excluding ancestors of ``ignored_parents``.
 
     If we have the following inheritance diagram:
@@ -266,9 +266,16 @@ def _get_parents(
         parent = to_explore.pop()
         if parent.qname() in ignored_parents:
             continue
-        parents.add(parent)
-        to_explore.extend(parent.ancestors(recurs=False))
-    return parents
+        if parent not in parents:
+            yield parent
+            parents.add(parent)
+            to_explore.extend(parent.ancestors(recurs=False))
+
+
+def _get_parents(
+    node: nodes.ClassDef, ignored_parents: FrozenSet[str]
+) -> Set[nodes.ClassDef]:
+    return set(_get_parents_iter(node, ignored_parents))
 
 
 class MisdesignChecker(BaseChecker):
--- a/tests/regrtest_data/hang/pkg4972/flask/__init__.py
+++ b/tests/regrtest_data/hang/pkg4972/flask/__init__.py
@@ -1,7 +1,0 @@
-import flask
-import pkg4972.flask  # self-import necessary
-
-class Fake(flask.Flask):
-    pass
-
-flask.Flask = Fake
--- a/tests/regrtest_data/hang/pkg4972/flask/__init__.py
+++ b/tests/regrtest_data/hang/pkg4972/flask/__init__.py
@@ -1,0 +1,7 @@
+import flask
+import pkg4972.flask  # self-import necessary
+
+class Fake(flask.Flask):
+    pass
+
+flask.Flask = Fake
--- a/tests/test_regr.py
+++ b/tests/test_regr.py
@@ -22,7 +22,9 @@ to be incorporated in the automatic functional test framework
 # pylint: disable=redefined-outer-name
 
 import os
+import signal
 import sys
+from contextlib import contextmanager
 from os.path import abspath, dirname, join
 from typing import Iterator
 
@@ -139,3 +141,29 @@ def test_pylint_config_attr() -> None:
     assert len(inferred) == 1
     assert inferred[0].root().name == "optparse"
     assert inferred[0].name == "Values"
+
+
+@contextmanager
+def timeout(timeout_s):
+    def _handle(_signum, _frame):
+        pytest.fail("timed out")
+
+    signal.signal(signal.SIGALRM, _handle)
+    signal.setitimer(signal.ITIMER_REAL, timeout_s)
+    yield
+    signal.setitimer(signal.ITIMER_REAL, 0)
+    signal.signal(signal.SIGALRM, signal.SIG_DFL)
+
+
+@pytest.mark.slow
+@pytest.mark.serial
+@pytest.mark.skipif(not hasattr(signal, "setitimer"), reason="Assumes POSIX signals")
+@pytest.mark.parametrize(
+    "fname,timeout_s",
+    [
+        (join(REGR_DATA, "hang", "pkg4972.flask"), 30.0),
+    ],
+)
+def test_hang(finalize_linter, fname, timeout_s):
+    with timeout(timeout_s):
+        finalize_linter.check(fname)
--- a/tests/test_regr.py
+++ b/tests/test_regr.py
@@ -161,7 +161,7 @@ def timeout(timeout_s):
 @pytest.mark.parametrize(
     "fname,timeout_s",
     [
-        (join(REGR_DATA, "hang", "pkg4972.flask"), 30.0),
+        (join(REGR_DATA, "hang", "pkg4972.string"), 30.0),
     ],
 )
 def test_hang(finalize_linter, fname, timeout_s):
--- a/tests/test_regr.py
+++ b/tests/test_regr.py
@@ -144,7 +144,7 @@ def test_pylint_config_attr() -> None:
 
 
 @contextmanager
-def timeout(timeout_s):
+def timeout(timeout_s: float):
     def _handle(_signum, _frame):
         pytest.fail("timed out")
 
@@ -164,6 +164,6 @@ def timeout(timeout_s):
         (join(REGR_DATA, "hang", "pkg4972.string"), 30.0),
     ],
 )
-def test_hang(finalize_linter, fname, timeout_s):
+def test_hang(finalize_linter, fname: str, timeout_s: float) -> None:
     with timeout(timeout_s):
         finalize_linter.check(fname)
--- a/tests/regrtest_data/hang/pkg4972/string/__init__.py
+++ b/tests/regrtest_data/hang/pkg4972/string/__init__.py
@@ -1,0 +1,7 @@
+import string
+import pkg4972.string  # self-import necessary
+
+class Fake(string.Formatter):
+    pass
+
+string.Formatter = Fake


nelfin:
CC @DanielNoord, @Pierre-Sassoulas. I'd be keen to hear your thoughts on how to improve the commit message (top half of the PR description) and how to create a regression test for this. I struggled to fit it in to the functional test framework given that it caused a hang instead of an exception, and the behaviour differed based on which directory you ran pylint from.

nelfin:
De-specified the type annotation for `_get_parents_iter`

DanielNoord:
Commit message looks good to me.

No idea, if this works, but can we do something with how `benchmark` tests report time? I don't know how those tests are run, but we could make a very basic example and make the test fail if the test takes >1min?

tests/test_regr.py
```diff
@@ -139,3 +141,29 @@ def test_pylint_config_attr() -> None:
...
+    signal.signal(signal.SIGALRM, signal.SIG_DFL)
+
+
+@pytest.mark.slow
+@pytest.mark.serial
+@pytest.mark.skipif(not hasattr(signal, "setitimer"), reason="Assumes POSIX signals")
+@pytest.mark.parametrize(
+    "fname,timeout_s",
+    [
+        (join(REGR_DATA, "hang", "pkg4972.flask"), 30.0),
```
DanielNoord, line 164: Am I missing something or does this PR not include the data this `parametrize` is referring to?

nelfin: Oops

pylint/checkers/design_analysis.py
```diff
@@ -266,9 +266,16 @@ def _get_parents(
         parent = to_explore.pop()
         if parent.qname() in ignored_parents:
             continue
-        parents.add(parent)
-        to_explore.extend(parent.ancestors(recurs=False))
-    return parents
+        if parent not in parents:
+            yield parent
```
DanielNoord, line 270: I think a comment explaining why we do this might be helpful for future reference

DanielNoord, line 270: I think a comment explaining why we do this might be helpful for future reference

tests/regrtest_data/hang/pkg4972/flask/__init__.py
```diff
@@ -0,0 +1,7 @@
+import flask
```
DanielNoord, line 1: Is it necessary to do this with `flask`? It is not currently part of our test dependencies. I haven't tested this and perhaps this works even without the dependency installed but I would prefer a more general case.

tests/test_regr.py
```diff
@@ -139,3 +141,29 @@ def test_pylint_config_attr() -> None:
...
+    ],
+)
+def test_hang(finalize_linter, fname, timeout_s):
```
DanielNoord, line 167: My knowledge of `purest` is limited, but looking at the code and parameter names I guess the `30.0` means this test should pass in 30 seconds? If so, is that reasonable for our CI? Do you know how long this test normally takes?

DanielNoord (changes requested):

DanielNoord:
Btw, I also did this previously but was then (rightly) asked not to rebase after new commits as it makes review more difficult. We normally squash all commits before merging anyway so a "clean" commit history is not necessary for us! 😄

tests/regrtest_data/hang/pkg4972/flask/__init__.py
Pierre-Sassoulas: We can test on "flask looking code" as long as we don't really have to install flask for test ;)

tests/test_regr.py
Pierre-Sassoulas: Yes I think 30s is a long time to add to the total time of every pipeline. If we can't do anything else (because it's hard to reproduce the problem), we might have to parallelize the tests marked slow in github actions. But I guess we could also make the performance of pylint better in the first place ?

DanielNoord: I think it fails whenever the test takes more than 30 seconds? Instead of adding 30 seconds every time.
However, I wonder if there are moments at which GitHub lags too much and the test takes more than 30 seconds and we fail as false positive.

tests/regrtest_data/hang/pkg4972/flask/__init__.py
DanielNoord: Wouldn't it still be better to use "pylintic" imports?

nelfin: I took a few shots at trying to get it working with just standard library imports but eventually gave up. I'll give it another shot.

pylint/checkers/design_analysis.py
nelfin: Good idea

tests/test_regr.py
nelfin: The test fails if it takes more than 30s yes, it only takes a couple of seconds on my 2014-era laptop so I just picked a 10x timeout for safety. Parallelising them wouldn't be compatible with the SIGALRM approach so that would have to change if you start adding `pytest-xdist` or something in the future.

nelfin:
> Btw, I also did this previously but was then (rightly) asked not to rebase after new commits as it makes review more difficult. We normally squash all commits before merging anyway so a "clean" commit history is not necessary for us!

Sounds like a GitHub problem :D. I split that one out into a separate commit for easy testing really, I wanted to make sure it was actually a regression test by checking that it went from failing to passing.

I'll push up the requested changes as new commits without rebasing. For the purposes of the squash, the commit message should come from 80939c0 then.

tests/regrtest_data/hang/pkg4972/flask/__init__.py
nelfin: I got a solution working with `builtins.string`.

> Wouldn't it still be better to use "pylintic" imports?

I'm unclear on what you mean by "pylintic" imports sorry.

pylint/checkers/design_analysis.py
DanielNoord: Perhaps a little expressive, but I like well commented code! 👍

tests/test_regr.py
```diff
@@ -139,3 +141,29 @@ def test_pylint_config_attr() -> None:
     assert len(inferred) == 1
     assert inferred[0].root().name == "optparse"
     assert inferred[0].name == "Values"
+
+
+@contextmanager
+def timeout(timeout_s):
```
DanielNoord, line 147: Could you add typing to these functions? Other than that, this LGTM!

tests/regrtest_data/hang/pkg4972/flask/__init__.py
DanielNoord: Nice work!

Yeah, that's a bit vague. I meant something like "modules/libraries pylint contributors are familiar with", so something within the `pylint` framework or standard lib. Using `string` is a perfect solution imo!

DanielNoord (approved):

Pierre-Sassoulas (approved):
Thank you @nelfin !

Pierre-Sassoulas:
I learned about Hindley-Milner today, thanks 😄

PR #5063, opened 2021-09-22
Fix ``useless-suppression`` for ``wrong-import-order``

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [x] If it's a new feature, or an important bug fix, add a What's New entry in
      `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

This also adds a new method to ``MessagesHandlerMixIn`` which adds messages to the list of the ignored messages without doing anything else. 
This will be useful for some other ``useless-suppression`` false positives.
This closes #2366

PR #5064, opened 2021-09-22
Refactor of ``FileState.handle_ignored_message()``

DanielNoord:

## Type of Changes

## Description

I can understand why the parameters were originally kept similar to `add_message()` but with the added method in  #5063 I think removing them leaves us with a cleaner method and way to call that method.

However, while writing this PR I wonder whether we should deprecate this rather than removing the parameters outright. Opinions?

### DIFF
--- a/pylint/message/message_handler_mix_in.py
+++ b/pylint/message/message_handler_mix_in.py
@@ -302,9 +302,6 @@ class MessagesHandlerMixIn:
                 ),
                 message_definition.msgid,
                 line,
-                node,
-                args,
-                confidence,
             )
             return
         # update stats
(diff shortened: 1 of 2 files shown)


cdce8p (commented):
Don't think a deprecation would be necessary as this should only really be used internally.
I just wonder if we gain much by doing it or if we end up reverting it because we realized that we could use them after all. However, considering that this hasn't been used until now, there likely isn't any use for it in the future either.

Pierre-Sassoulas:
Well I'm going to go with YAGNI here. If we're wrong, we can add what's strictly necessary later.

PR #5067, opened 2021-09-23
Update toml dependency to >=0.9.2

### LINKED ISSUE #5066: TOML dependency declaration is outdated
### Bug description

The `toml` dependency is declared as `>=0.7.1`, however, since this https://github.com/PyCQA/pylint/pull/4176 has been merged the dependency should state `>=0.10`.

I came across this due to a shared dependency on `toml` between two packages, of which one has `toml<0.10,>=0.9` causing `toml==0.9.6` to be installed.

### Configuration

_No response_

### Command used

```shell
pylint --version
```

### Pylint output

```shell
  File "/usr/local/bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
  File "/usr/local/lib/python3.8/site-packages/pylint/__init__.py", line 21, in run_pylint
    from pylint.lint import Run as PylintRun
  File "/usr/local/lib/python3.8/site-packages/pylint/lint/__init__.py", line 76, in <module>
    from pylint.lint.parallel import check_parallel
  File "/usr/local/lib/python3.8/site-packages/pylint/lint/parallel.py", line 9, in <module>
  File "/usr/local/lib/python3.8/site-packages/pylint/lint/utils.py", line 10, in <module>
    from pylint.config import PYLINT_HOME
  File "/usr/local/lib/python3.8/site-packages/pylint/config/__init__.py", line 46, in <module>
    from pylint.config.find_default_config_files import find_default_config_files
  File "/usr/local/lib/python3.8/site-packages/pylint/config/find_default_config_files.py", line 8, in <module>
    from toml.decoder import TomlDecodeError
ModuleNotFoundError: No module named 'toml.decoder'; 'toml' is not a package
```

### Expected behavior

I would expect version numbers to be printed.

### Pylint version

```shell
pylint 2.11.1
```

### OS / Environment

Debian 10 Buster

### Additional dependencies

pylint==2.11.1
toml==0.9.6

jeroenseegers:
## Type of Changes

## Description

Fix the minimum version of the `toml` dependency as described in #5066.

Closes #5066

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -43,6 +43,10 @@ Release date: TBA
 
   Closes #2366
 
+* Fixed ``toml`` dependency issue
+
+  Closes #5066
+
 
 What's New in Pylint 2.11.1?
 ============================
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
     astroid>=2.8.0,<2.9 # (You should also upgrade requirements_test_min.txt)
     isort>=4.2.5,<6
     mccabe>=0.6,<0.7
-    toml>=0.7.1
+    toml>=0.9.2
     colorama;sys_platform=="win32"
     typing-extensions>=3.10.0;python_version<"3.10"
 python_requires = ~=3.6
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -543,3 +543,6 @@ contributors:
 * Hayden Richards (SupImDos): contributor
   - Fixed "no-self-use" for async methods
   - Fixed "docparams" extension for async functions and methods
+
+* Jeroen Seegers (jeroenseegers): contributor
+  - Fixed `toml` dependency issue
--- a/pylint/config/find_default_config_files.py
+++ b/pylint/config/find_default_config_files.py
@@ -5,7 +5,7 @@ import configparser
 import os
 
 import toml
-from toml.decoder import TomlDecodeError
+from toml import TomlDecodeError
 
 
 def _toml_has_config(path):


Pierre-Sassoulas:
> I came across this due to a shared dependency on toml between two packages, of which one has toml<0.10,>=0.9 causing toml==0.9.6 to be installed.

Wouldn't that make pylint incompatible with this package ? Maybe the fix is to handle both import by adding an import guard and keep the compatibility with toml > 0.7 ?

jeroenseegers:
> > I came across this due to a shared dependency on toml between two packages, of which one has toml<0.10,>=0.9 causing toml==0.9.6 to be installed.
> 
> Wouldn't that make pylint incompatible with this package ? Maybe the fix is to handle both import by adding an import guard and keep the compatibility with toml > 0.7 ?

I'm not sure if that's very relevant in this case. It's a tag of a package > 2 years old by now.

https://github.com/stripe/stripe-python/tree/v2.29.4

cdce8p:
From what I can tell, we should be able to import `TomlDecodeError` from `toml` directly even in `0.10.0`. Just need to update the code. With that we shouldn't need to update the dependency.
https://github.com/PyCQA/pylint/blob/09a0b50bfeb473625f8ad586f72fd35d790de6ac/pylint/config/find_default_config_files.py#L8

[`toml/__init__.py` - 0.10.0](https://github.com/uiri/toml/blob/4935f616ef78c35a968b2473e806d7049eba9af1/toml/__init__.py#L15)

jeroenseegers:
> From what I can tell, we should be able to import `TomlDecodeError` from `toml` directly even in `0.10.0`. Just need to update the code. With that we shouldn't need to update the dependency.
> 
> https://github.com/PyCQA/pylint/blob/09a0b50bfeb473625f8ad586f72fd35d790de6ac/pylint/config/find_default_config_files.py#L8
> 
> [`toml/__init__.py` - 0.10.0](https://github.com/uiri/toml/blob/4935f616ef78c35a968b2473e806d7049eba9af1/toml/__init__.py#L15)

The issue is that that import is only available from `0.10.0` onwards, not in earlier versions.

cdce8p:
> The issue is that this import is only available from `0.10.0` onwards, not in earlier versions.
> 
> From `0.9.2` it's available as `from toml import `TomlDecodeError`: https://github.com/uiri/toml/blob/0.9.2/toml.py
> 
> Before that, `TomlDecodeError` does not exist at all.

👍🏻 In that case we should probably require `0.9.2`.

jeroenseegers:
> 👍🏻 In that case we should probably require `0.9.2`.

Shall I then make it `toml>=0.9.2,<0.10` and change
https://github.com/PyCQA/pylint/blob/09a0b50bfeb473625f8ad586f72fd35d790de6ac/pylint/config/find_default_config_files.py#L8
to
```python
from toml import TomlDecodeError
```
?

setup.cfg
```diff
@@ -47,7 +47,7 @@ install_requires =
     astroid>=2.8.0,<2.9 # (You should also upgrade requirements_test_min.txt)
     isort>=4.2.5,<6
     mccabe>=0.6,<0.7
-    toml>=0.7.1
+    toml>=0.9.2,<0.10
```
cdce8p, line 50: Just wondering: Why limit it to `<0.10`? The import should work for `>=0.10` as well.

jeroenseegers: As of `0.10` the `TomlDecodeError` was moved from `toml` to `toml.decoder`.

jeroenseegers: See https://github.com/uiri/toml/blob/0.9.2/toml.py vs https://github.com/uiri/toml/blob/0.10.0/toml/decoder.py

cdce8p: It did, but the import is still possible
https://github.com/uiri/toml/blob/4935f616ef78c35a968b2473e806d7049eba9af1/toml/__init__.py#L15

jeroenseegers: Ah! I missed that part! In that case, we don't need to specify an upper version indeed.

jeroenseegers: Fixed in latest version.

ChangeLog
```diff
@@ -11,6 +11,8 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.12.rst'
 
+* Fixed ``toml`` dependency issue
+
```
cdce8p, line 15: This should be moved to the `2.11.2` section as it's a bugfix.

doc/whatsnew/2.12.rst
```diff
@@ -23,3 +23,5 @@ Extensions
 
 Other Changes
 =============
+
+* Fixed ``toml`` dependency issue
```
cdce8p, line 27: Not necessary -> can be removed.

cdce8p (approved):
LGTM! Thanks @jeroenseegers 🐬

DanielNoord (approved):

Pierre-Sassoulas (approved):
Thank you, great catch !

Issue #5038, opened 2021-09-17
Improve docs for `py-version`


cdce8p:
Followup to #5019 and #5024

The current documentation around the `py-version` setting introduced in `2.11` could be improved.
* Conflicting default values -> Default is autogenerated from `sys.version_info`. Likely requires an exception to not show a default value #5069
* Explain best practices
  * Most users should default to running `pylint` with the oldest Python version their project supports
  * In some cases it's useful to set `py-version` to allow running `pylint` with newer versions. E.g. when `pylint` is used together with a pre-commit hook and  most development happens on a newer Python version.
  * Setting `py-version` will only hide suggestions for features that aren't included in all (by a project supported) Python versions.

PR #5069, opened 2021-09-23
Small py-version improvements

cdce8p:
## Type of Changes

## Description

Followup to review comments here: https://github.com/PyCQA/pylint/pull/5024#discussion_r710739423
Partially addresses #5038

/CC: @Pierre-Sassoulas @scop

#### Changes
* Add `py-version` requirement to checker documentation
* Improve default value documentation for `py-version` (see image below)
* Improve option parser error message for `py-version`

--
Screen Shot 2021-09-23 at 15 43 46
https://pylint--5069.org.readthedocs.build/en/5069/technical_reference/features.html

### DIFF
--- a/pylint/utils/utils.py
+++ b/pylint/utils/utils.py
@@ -311,7 +311,7 @@ def _comment(string):
 def _format_option_value(optdict, value):
     """return the user input's value from a 'compiled' value"""
     if optdict.get("type", None) == "py_version":
-        value = ".".join(str(item) for item in value)
+        value = "current Python interpreter version (e.g., '3.8')"
     elif isinstance(value, (list, tuple)):
         value = ",".join(_format_option_value(optdict, item) for item in value)
     elif isinstance(value, dict):
--- a/pylint/utils/utils.py
+++ b/pylint/utils/utils.py
@@ -131,7 +131,7 @@ def get_rst_section(section, options, doc=None):
         if help_opt:
             formatted_help = normalize_text(help_opt, indent="  ")
             result += f"{formatted_help}\n"
-        if value:
+        if value and optname != "py-version":
             value = str(_format_option_value(optdict, value))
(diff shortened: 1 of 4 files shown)


pylint/utils/utils.py
```diff
@@ -311,7 +311,7 @@ def _comment(string):
 def _format_option_value(optdict, value):
     """return the user input's value from a 'compiled' value"""
     if optdict.get("type", None) == "py_version":
-        value = ".".join(str(item) for item in value)
+        value = "current Python interpreter version (e.g., '3.8')"
```
DanielNoord, line 314: Shouldn't this also return the current value specified by the user? At least that's what the docstring says and what the previous version did.

Pierre-Sassoulas:
```suggestion
        value = f"current Python interpreter version (e.g., '{'.'.join(str(item) for item in value)}')"
```

Pierre-Sassoulas (commented):
👍

pylint/utils/utils.py
cdce8p: > Shouldn't this also return the current value specified by the user? At least that's what the docstring says and what the previous version did.

Yeah 😅
I didn't though about the help text which also uses that method. Thanks for the pointer. Should be fixed now.

cdce8p:
The latest changes hide the `Default: ...` line from the documentation completely.
```
Minimum Python version to use for version dependent checks. Will default to the version used to run pylint.
```

`pylint --help` (run in the pylint repo, `3.6` from `pylintrc`)
```
    --py-version=<py_version>
                        Minimum Python version to use for version dependent
                        checks. Will default to the version used to run
                        pylint. [current: 3.6]
```

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #5070, opened 2021-09-23
Add ``flake8-typing-import`` to the pre-commit configuration

Pierre-Sassoulas:
## Type of Changes

## Description

This permit to check that the typing is valid for all python version. Relates to https://github.com/PyCQA/pylint/pull/5068

It seems we have some trouble with python 3.6.0 right now:
```
pylint/testutils/output_line.py:71:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/testutils/output_line.py:85:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/testutils/output_line.py:91:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/testutils/output_line.py:99:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/pyreverse/printer.py:43:5: TYP005 NamedTuple does not support defaults in 3.6.0
pylint/pyreverse/printer.py:44:5: TYP005 NamedTuple does not support defaults in 3.6.0
pylint/pyreverse/printer.py:45:5: TYP005 NamedTuple does not support defaults in 3.6.0
pylint/pyreverse/printer.py:46:5: TYP005 NamedTuple does not support defaults in 3.6.0
pylint/checkers/similar.py:206:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/checkers/similar.py:211:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/checkers/similar.py:219:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/checkers/similar.py:222:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/checkers/refactoring/refactoring_checker.py:164:5: TYP005 NamedTuple does not support defaults in 3.6.0
pylint/checkers/refactoring/refactoring_checker.py:165:5: TYP005 NamedTuple does not support defaults in 3.6.0
pylint/checkers/refactoring/refactoring_checker.py:166:5: TYP005 NamedTuple does not support defaults in 3.6.0
pylint/checkers/refactoring/refactoring_checker.py:168:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/checkers/refactoring/refactoring_checker.py:171:5: TYP004 NamedTuple does not support methods in 3.6.0
pylint/checkers/refactoring/refactoring_checker.py:181:5: TYP004 NamedTuple does not support methods in 3.6.0

```

 We can fix them in 2.11.0 or upgrade to 3.6.2 right now, what do you think @DanielNoord @cdce8p ?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,7 @@ repos:
     rev: 4.0.1
     hooks:
       - id: flake8
+        additional_dependencies: [flake8-typing-imports==1.10.1]
         exclude: *fixtures
   - repo: local
     hooks:
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,7 +12,7 @@ Release date: TBA
 
 ..
   Insert your changelog randomly, it will reduce merge conflicts
-  (Ie. not necessarilly at the end)
+  (Ie. not necessarily at the end)
 
 
 What's New in Pylint 2.12.1?
@@ -24,7 +24,7 @@ Release date: TBA
 
 ..
   Insert your changelog randomly, it will reduce merge conflicts
-  (Ie. not necessarilly at the end)
+  (Ie. not necessarily at the end)
 
 
 What's New in Pylint 2.12.0?
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -2251,7 +2251,7 @@ class DocStringChecker(_BasicChecker):
                     func.bound, astroid.Instance
                 ):
                     # Strings.
-                    if func.bound.name in ("str", "unicode", "bytes"):
+                    if func.bound.name in {"str", "unicode", "bytes"}:
                         return
             if node_type == "module":
                 message = "missing-module-docstring"
@@ -2386,7 +2386,7 @@ class ComparisonChecker(_BasicChecker):
 
         # True/False singletons have a special-cased message in case the user is
         # mistakenly using == or != to check for truthiness
-        if singleton in (True, False):
+        if singleton in {True, False}:
             suggestion_template = (
                 "{} if checking for the singleton value {}, or {} if testing for {}"
             )
@@ -2440,7 +2440,7 @@ class ComparisonChecker(_BasicChecker):
         def _is_numpy_nan(node):
             if isinstance(node, nodes.Attribute) and node.attrname == "NaN":
                 if isinstance(node.expr, nodes.Name):
-                    return node.expr.name in ("numpy", "nmp", "np")
+                    return node.expr.name in {"numpy", "nmp", "np"}
             return False
 
         def _is_nan(node) -> bool:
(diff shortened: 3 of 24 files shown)


DanielNoord:
If #5068 is scheduled for `2.12` I think should also be added to that milestone. Otherwise we're doing double work.

After that the dependency and python version should probably be updated to 3.6.2.

After that I don't see any problems with this.

cdce8p:
> It seems we have some trouble with python 3.6.0 right now: [...]
> We can fix them in 2.11.0 or upgrade to 3.6.2 right now, what do you think?

I checked to see when these changes were introduced. In the end I'm still be bit surprised that we haven't gotten any bug reports for `3.6.0`.

| File | First version | Commit | Likely issue for pylint |
| ---- | ------------ | -------- | ----------- |
| pylint/testutils/output_line.py | 2.9.0 | https://github.com/PyCQA/pylint/commit/304c87c07df80b0b2b905a3f8e28ceafe660551b | No, used just for tests |
| pylint/pyreverse/printer.py | 2.10.0 | https://github.com/PyCQA/pylint/commit/7bb50436d4bd440c074d1c0d8f9564743b2b56ab | No, just `pyreverse` |
| pylint/checkers/similar.py | 2.10.0 | https://github.com/PyCQA/pylint/commit/1d1619ef913b99b06647d2030bddff4800abdf63 | Yes |
| pylint/checkers/refactoring/refactoring_checker.py | 2.9.4 | https://github.com/PyCQA/pylint/commit/dd54e55265c5b3dc8d99278e77a9c73359fb4ccf | Yes |

--
Usually I would have said that it would be better to fix the issues in `2.11` so we can ship one last fully supported version. However considering these date back to `2.9.4` and we haven't gotten any reports, combined with the nearing EOL for `3.6`. It might be easier to just drop the support now. I.e. moving #5068 up to `2.11.2`. Since it's technically already broken, bumping `python_requires` could be considered a bugfix.

.flake8
```diff
@@ -5,3 +5,4 @@ ignore =
 
 max-line-length=88
 max-complexity=39
+min_python_version = 3.6.0
```
cdce8p, line 8: The `python_requires` option in `setup.cfg` will be used so this isn't necessary.
https://github.com/asottile/flake8-typing-imports#configuration

.pre-commit-config.yaml
```diff
@@ -44,6 +44,7 @@ repos:
     rev: 3.9.2
     hooks:
       - id: flake8
+        additional_dependencies: [flake8-typing-imports==1.10.1]
```
cdce8p, line 47: Can you add `flake8-typing-imports==1.10.1` to `requirements_text_pre_commit.txt`, too?

cdce8p:
Just though about one more thing: It might be the case that Python 3.6.0 still runs. Just the implementation for default values and methods might be broken internally. Can't confirm it though

.flake8
Pierre-Sassoulas: 👍 Right, changing from ``~=`` to ``>=`` is necessary enough. (and this alone does not work because the value in setup.cfg override it.)

Pierre-Sassoulas:
Yes I also think, that the default value is probably not a crash, the method in a testutil class is probably only a problem when launching the test with 3.6.0 or 3.6.1. The only thing then could be the method in ``LineSetStartCouple`` but we never got any issue opened for it afaik ? 

Best practice would be to fix it and release a last working version for 3.6.0 and 3.6.1 but I think there are very few users on those interpreter. If you're on the very first two versions of a dying python interpreter chances are you're also on a dying pylint.

So I agree with
> moving #5068 up to 2.11.2. Since it's technically already broken, bumping python_requires could be considered a bugfix.

DanielNoord:
Do we know if there are any current bugs specific to `3.6.0`/`3.6.1`? If so, we might want to have a look if they are easily fixable and release one final minor version for them. Other than that I agree with what's been said here!

DanielNoord (approved):

Issue #5014, opened 2021-09-15
Refactor of `PyLinter.stats`


DanielNoord:
### Current problem

The current problem is that the `stats` attribute of the main linter/checker classes support a wide array of different items. As of now the typing is:
https://github.com/PyCQA/pylint/blob/d4e61b9024dd6f073739fc6f7e6ecef2dceb9aaa/pylint/typing.py#L12-L14

Problems we are running into is stuff like this in `pylint/lint/pylinter.py`:
```python
def set_current_module(self, modname, filepath: Optional[str] = None):
        """set the name of the currently analyzed module and
        init statistics for it
        """
        if not modname and filepath is None:
            return
        self.reporter.on_set_current_module(modname, filepath)
        self.current_name = modname
        self.current_file = filepath or modname
        self.stats["by_module"][modname] = {}  # type: ignore # Refactor of PyLinter.stats necessary
        self.stats["by_module"][modname]["statement"] = 0  # type: ignore
        for msg_cat in MSG_TYPES.values():
            self.stats["by_module"][modname][msg_cat] = 0  # type: ignore
```
We can't specify the type of `by_module` which then causes issues with `mypy`.

I have tinkered around for a bit, but did not find a good solution/approach to this refactor. (I believe) it is impossible to type certain keys of a `Dict`/`TypedDict` and leave others unspecified. 
We would thus need a complete rework of the `stats` handling. This shouldn't be too big of an issue, but I would like some input on what maintainers/users think will be the best approach to this refactor and how they would ideally see `linter.stats` behave.

### Desired solution

/

Pierre-Sassoulas:
I think we need to look at what is required and include that in a properly initialized and easily usable class. For example there's stat by module, class, methods... I think there might be some work with some design to do first.

DanielNoord:
The most logical approach seems to be to use a `TypeDict` or `NamdeTuple`. However, we tend to loop over `nodes` or `modules` and add stats based on a concatenation of the node-variables and a prefix. This doesn't work with `TypedDict` as it requires string literals, while concatenation always returns `str`. See the following example:

```python
from typing import TypedDict

class Stats(TypedDict):
    badnames_ClassDef: int
    badnames_Module: int

stats = Stats(badnames_ClassDef=1, badnames_Module=1)

for node in ("ClassDef", "Module"):
    stats["badnames_" + node] = 1 # TypedDict key must be a string literal; expected one of ("badnames_ClassDef", "badnames_Module")
```

[...]

PR #5074, opened 2021-09-26
Refactor CheckerStats

DanielNoord:

## Type of Changes

## Description

Ref #5014 

This is a first iteration of a refactor of the `stats` attribute of most checkers.

This current version is able to run `pylint` over our own codebase without any crashes. The biggest change is in `pylint/utils/checkerstats.py`, which introduces a new class which can be used to store stats. 

Before continuing work on this I would like some feedback from maintainers on whether this is the right way forward.

As most tests depend on the previous form of the `stats` attribute we would also need to change most tests. I would like to reach a consensus on the design of the new class before going over all tests to see how they would need to be changed and whether the new class passes them.

Furthermore, as this quite a large change I would like to know how a final PR should be broken down into more manageable pieces for review. I can understand one big PR might be too much to handle.

### DIFF
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -2029,7 +2029,7 @@ class NameChecker(_BasicChecker):
         )
 
         self.add_message(warning, node=node, args=args, confidence=confidence)
-        self.linter.stats.set_bad_name(node_type, 1)
+        self.linter.stats.increase_bad_name(node_type, 1)
 
     def _name_allowed_by_regex(self, name: str) -> bool:
         return name in self.config.good_names or any(
@@ -2059,7 +2059,7 @@ class NameChecker(_BasicChecker):
         if self._name_allowed_by_regex(name=name):
             return
         if self._name_disallowed_by_regex(name=name):
-            self.linter.stats.set_bad_name(node_type, 1)
+            self.linter.stats.increase_bad_name(node_type, 1)
             self.add_message("disallowed-name", node=node, args=name)
             return
         regexp = self._name_regexps[node_type]
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -83,7 +83,7 @@ from pylint.checkers.utils import (
     is_property_setter,
 )
 from pylint.reporters.ureports import nodes as reporter_nodes
-from pylint.utils import CheckerStats
+from pylint.utils import LinterStats
 
 if sys.version_info >= (3, 8):
     from typing import Literal
@@ -394,8 +394,8 @@ def _has_abstract_methods(node):
 
 def report_by_type_stats(
     sect,
-    stats: CheckerStats,
-    old_stats: Optional[CheckerStats],
+    stats: LinterStats,
+    old_stats: Optional[LinterStats],
 ):
     """make a report of
 
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -68,12 +68,12 @@ import collections
 import itertools
 import re
 import sys
-from typing import Any, Dict, Iterator, Optional, Pattern, Union
+from typing import Any, Dict, Iterator, Optional, Pattern, cast
 
 import astroid
 from astroid import nodes
 
-from pylint import checkers, exceptions, interfaces
+from pylint import checkers, interfaces
 from pylint import utils as lint_utils
 from pylint.checkers import utils
 from pylint.checkers.utils import (
@@ -83,7 +83,12 @@ from pylint.checkers.utils import (
     is_property_setter,
 )
 from pylint.reporters.ureports import nodes as reporter_nodes
-from pylint.typing import CheckerStats
+from pylint.utils import CheckerStats
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 
 class NamingStyle:
@@ -390,7 +395,7 @@ def _has_abstract_methods(node):
 def report_by_type_stats(
     sect,
     stats: CheckerStats,
-    old_stats: CheckerStats,
+    old_stats: Optional[CheckerStats],
 ):
     """make a report of
 
@@ -400,38 +405,28 @@ def report_by_type_stats(
     # percentage of different types documented and/or with a bad name
     nice_stats: Dict[str, Dict[str, str]] = {}
     for node_type in ("module", "class", "method", "function"):
-        try:
-            total: int = stats[node_type]  # type: ignore
-        except KeyError as e:
-            raise exceptions.EmptyReportError() from e
+        node_type = cast(Literal["function", "class", "method", "module"], node_type)
+        total = stats.get_node_count(node_type)
         nice_stats[node_type] = {}
         if total != 0:
-            try:
-                undocumented_node: int = stats["undocumented_" + node_type]  # type: ignore
-                documented = total - undocumented_node
-                percent = (documented * 100.0) / total
-                nice_stats[node_type]["percent_documented"] = f"{percent:.2f}"
-            except KeyError:
-                nice_stats[node_type]["percent_documented"] = "NC"
-            try:
-                badname_node: int = stats["badname_" + node_type]  # type: ignore
-                percent = (badname_node * 100.0) / total
-                nice_stats[node_type]["percent_badname"] = f"{percent:.2f}"
-            except KeyError:
-                nice_stats[node_type]["percent_badname"] = "NC"
+            undocumented_node = stats.get_undocumented(node_type)
+            documented = total - undocumented_node
+            percent = (documented * 100.0) / total
+            nice_stats[node_type]["percent_documented"] = f"{percent:.2f}"
+            badname_node = stats.get_bad_names(node_type)
+            percent = (badname_node * 100.0) / total
+            nice_stats[node_type]["percent_badname"] = f"{percent:.2f}"
     lines = ["type", "number", "old number", "difference", "%documented", "%badname"]
     for node_type in ("module", "class", "method", "function"):
-        new = stats[node_type]
-        old: Optional[Union[str, int]] = old_stats.get(node_type, None)  # type: ignore
-        if old is not None:
-            diff_str = lint_utils.diff_string(old, new)
-        else:
-            old, diff_str = "NC", "NC"
+        node_type = cast(Literal["function", "class", "method", "module"], node_type)
+        new = stats.get_node_count(node_type)
+        old = old_stats.get_node_count(node_type) if old_stats else None
+        diff_str = lint_utils.diff_string(old, new) if old else None
         lines += [
             node_type,
             str(new),
-            str(old),
-            diff_str,
+            str(old) if old else "NC",
+            diff_str if diff_str else "NC",
             nice_stats[node_type].get("percent_documented", "0"),
             nice_stats[node_type].get("percent_badname", "0"),
         ]
@@ -1086,13 +1081,12 @@ class BasicChecker(_BasicChecker):
 
     def __init__(self, linter):
         super().__init__(linter)
-        self.stats: CheckerStats = {}
         self._tryfinallys = None
 
     def open(self):
         """initialize visit variables and statistics"""
         self._tryfinallys = []
-        self.stats = self.linter.add_stats(module=0, function=0, method=0, class_=0)
+        self.linter.stats.reset_node_count()
 
     @utils.check_messages("using-constant-test", "missing-parentheses-for-call-in-test")
     def visit_if(self, node: nodes.If) -> None:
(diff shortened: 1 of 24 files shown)


pylint/lint/parallel.py
```diff
@@ -36,21 +34,69 @@ def _get_new_args(message):
 
 
 def _merge_stats(stats: List[CheckerStats]):
-    merged: CheckerStats = {}
```
Pierre-Sassoulas, line 39: This could be a class function in CheckerStats.

Pierre-Sassoulas (commented):
I took a quick look, using a class makes the code better and a lot of places makes more sense now. Regarding the design of the class itself, I don't have a very strong opinion on what should be done, any class is better than what we had before but I don't know what the best design would be. Which means this one with TypedDict is probably good enough. The alternative would be to do something generic by permitting to tag a stat with a string, right ?

pylint/lint/parallel.py
DanielNoord: I thought about this but wasn't sure what was preferred. I was not sure if a function that receives two instances of the same class and returns a third should be a class function or a "external" util. What do you prefer?

DanielNoord:
Yeah, the problem is that you can't just add a new stat and need to add it to one of the `TypedDict`'s (or add a new one). However, in the long run I think this improves maintainability as we will have a much better idea of what kind of stats we are tracking. Figuring that out took quite a bit of time during working on this.

pylint/lint/parallel.py
Pierre-Sassoulas: I did not read in detail and thought it was "merge another stats into this one". An util make more sense if we return a third instance. Also as this is in parallel.py it makes sense that CheckerStats would be immutable, so you're right. Maybe we could move it in the CheckerStats namespace though ?

DanielNoord: Good idea. I'll add it it `checkerstats.py`

Pierre-Sassoulas:
> Figuring that out took quite a bit of time during working on this.

Yes I think the other approach advantages was speed of implementation. And as the work is already done I think it's better to keep the ``TypedDict``. Quite a piece of retro-engineering you did here, congrats !

DanielNoord:
@cdce8p Would you mind taking a look at the current state of this and deciding whether I can go ahead and start working on changing tests to make use of the new `CheckerStats` class? All feedback is welcome of course!

cdce8p:
> @cdce8p Would you mind taking a look at the current state of this and deciding whether I can go ahead and start working on changing tests to make use of the new `CheckerStats` class? All feedback is welcome of course!

Tbh this is not really one of my areas of expertise. Maybe it's because I've disabled the score in my project 🤷🏻‍♂️ If you and @Pierre-Sassoulas think this is a good change than I would say do it.

pylint/checkers/base.py
```diff
@@ -390,7 +395,7 @@ def _has_abstract_methods(node):
 def report_by_type_stats(
```
DanielNoord, line 395: We might also want to move this function to the ``checkerstats`` namespace

pylint/checkers/base.py
```diff
@@ -2210,17 +2190,21 @@ def visit_functiondef(self, node: nodes.FunctionDef) -> None:
                         overridden = True
                         break
                 self._check_docstring(
-                    ftype, node, report_missing=not overridden, confidence=confidence
+                    ftype, node, report_missing=not overridden, confidence=confidence  # type: ignore
```
DanielNoord, line 2193: Although `VS Code` correctly infers that `ftype` is a string literal because of line 2170, `mypy` does not.

pylint/message/message_handler_mix_in.py
```diff
@@ -369,7 +359,7 @@ def add_one_message(  # type: ignore # MessagesHandlerMixIn is always mixed with
             Message(
                 message_definition.msgid,
                 message_definition.symbol,
-                (abspath, path, module, obj, line or 1, col_offset or 0),
+                (abspath, path, module, obj, line or 1, col_offset or 0),  # type: ignore
```
DanielNoord, line 362: The types of `abspath`, `path` and `module` include `Any` and `None`. We need some overloads in `astroid` to fix this. I thought it best to add a `type: ignore` instead of adding multiple `isinstance` checks

DanielNoord:
@Pierre-Sassoulas Let me know how you want to review this and how I can help. The diff is quite large, but I didn't know how to make it any smaller. The `stats` attribute/class is used in various places and I couldn't change it only on some occasions.

I didn't see a lot possibilities for breaking up the PR, but let me know if you find any bits that can be extrapolated into their own PR.

I have tried to keep most tests as they were. This should help reviewing, as the changes do meet our current test suite.

pylint/checkers/base.py
```diff
@@ -1086,13 +1081,12 @@ class BasicChecker(_BasicChecker):
...
     def __init__(self, linter):
         super().__init__(linter)
-        self.stats: CheckerStats = {}
         self._tryfinallys = None
 
     def open(self):
         """initialize visit variables and statistics"""
         self._tryfinallys = []
-        self.stats = self.linter.add_stats(module=0, function=0, method=0, class_=0)
+        self.linter.stats.reset_node_count()
```
Pierre-Sassoulas, line 1089: Neat !

pylint/checkers/base.py
```diff
@@ -2044,7 +2029,7 @@ def _raise_name_warning(
         )
 
         self.add_message(warning, node=node, args=args, confidence=confidence)
-        self.stats["badname_" + node_type] += 1  # type: ignore
+        self.linter.stats.set_bad_name(node_type, 1)
```
Pierre-Sassoulas, line 2032: Are we setting the value of bad_names or are we instead increasing it by one ?

pylint/checkers/imports.py
```diff
@@ -858,7 +856,7 @@ def _add_imported_module(
                 self._module_pkg[context_name] = context_name.rsplit(".", 1)[0]
 
             # handle dependencies
-            dependencies_stat: Dict[str, Union[Set]] = self.stats["dependencies"]  # type: ignore
+            dependencies_stat: Dict[str, Union[Set]] = self.linter.stats.dependencies
```
Pierre-Sassoulas, line 859: Nice !

pylint/checkers/similar.py
```diff
@@ -863,8 +857,8 @@ def close(self):
 
             self.add_message("R0801", args=(len(couples), "\n".join(msg)))
             duplicated += num * (len(couples) - 1)
-        stats["nb_duplicated_lines"] = duplicated
-        stats["percent_duplicated_lines"] = total and duplicated * 100.0 / total
+        stats.nb_duplicated_lines += int(duplicated)
+        stats.percent_duplicated_lines += float(total and duplicated * 100.0 / total)
```
Pierre-Sassoulas, line 861: Unrelated to the change but `total and duplicated` is equivalent to `total + duplicated` ?

pylint/lint/pylinter.py
```diff
@@ -733,8 +734,8 @@ def enable_fail_on_messages(self):
     def any_fail_on_issues(self):
         return (
             self.stats
-            and self.stats.get("by_msg") is not None
-            and any(x in self.fail_on_symbols for x in self.stats["by_msg"])
+            and len(self.stats.by_msg) != 0
```
Pierre-Sassoulas, line 737:
```suggestion
            and self.stats.by_msg
```
?

pylint/lint/pylinter.py
```diff
@@ -1276,21 +1275,29 @@ def _report_evaluation(self):
...
         previous_stats = config.load_results(self.file_state.base_name)
-        if self.stats["statement"] == 0:
+        if self.stats.statement == 0:
             return note
 
         # get a global note for the code
         evaluation = self.config.evaluation
         try:
-            note = eval(evaluation, {}, self.stats)  # pylint: disable=eval-used
+            stats_dict = {
```
Pierre-Sassoulas, line 1284: Should this be a function un checker stats ?

pylint/lint/pylinter.py
```diff
@@ -1276,21 +1275,29 @@ def _report_evaluation(self):
...
+                "error": self.stats.error,
+                "warning": self.stats.warning,
+                "refactor": self.stats.refactor,
+                "convention": self.stats.convention,
+                "statement": self.stats.statement,
+            }
+            note = eval(evaluation, {}, stats_dict)  # pylint: disable=eval-used
```
Pierre-Sassoulas, line 1291: Not linked directly to this MR but there must be a way to not use eval here.

pylint/message/message_handler_mix_in.py
```diff
@@ -332,23 +332,13 @@ def add_one_message(  # type: ignore # MessagesHandlerMixIn is always mixed with
...
-            # This is not true for Linter and Reporter at least
-            # pylint: enable=fixme
-            self.stats = {
-                msg_cat: 0,
-                "by_module": {self.current_name: {msg_cat: 0}},
-                "by_msg": {},
-            }
-        self.stats[msg_cat] += 1  # type: ignore
-        self.stats["by_module"][self.current_name][msg_cat] += 1  # type: ignore
+
```
Pierre-Sassoulas, line 335: :pray:

pylint/utils/checkerstats.py
```diff
@@ -1,30 +1,353 @@
...
+        self.refactor = 0
+        self.statement = 0
+        self.warning = 0
+
+        self.global_note = 0
+        self.nb_duplicated_lines = 0
+        self.percent_duplicated_lines = 0.0
+
+    def __str__(self) -> str:
+        final_string = ""
```
Pierre-Sassoulas, line 120: Maybe a single fstring starting with `f"""` would be easier to maintain and think about here.

pylint/utils/checkerstats.py
```diff
@@ -1,30 +1,353 @@
...
+            "module",
+            "variable",
+        ],
+    ) -> int:
+        """Get a bad names node count"""
+        if node_name == "class":
+            return self.bad_names.get("klass", 0)
+        return self.bad_names.get(node_name, 0)
+
+    def set_bad_name(self, node_name: str, increase: int) -> None:
```
Pierre-Sassoulas, line 161:
```suggestion
    def increase_bad_name_by(self, node_name: str, increase: int) -> None:
```

Pierre-Sassoulas (changes requested):

pylint/checkers/base.py
DanielNoord: I'll rename these methods!

pylint/checkers/similar.py
DanielNoord: No, it seems to take `duplicated` whenever `total == 0`. See [docs](https://docs.python.org/3/library/stdtypes.html)

pylint/lint/pylinter.py
DanielNoord: 😅

DanielNoord: Can be removed altogether, since we now known that `stats` will be initiliazed with `by_msg`.

pylint/lint/pylinter.py
DanielNoord: Don't know. It is only really used here. Might "bloat" the class. Tell me whatever you prefer!

pylint/lint/pylinter.py
DanielNoord: Want me to add a fixme? I think it will be quite difficult though, since the setting in `pylintrc` allows users to create their own "score-function". So we will need to evaluate that function somehow.

pylint/utils/checkerstats.py
DanielNoord: Done for all methods!

pylint/utils/checkerstats.py
DanielNoord: We can't use `\n` in f-strings. We could do something like: 
```python
line_end = "\n"
final_string = f"{stat}{line_end} ..."
```
But I don't think that's much better. 
I thought about doing something with iterating over the fields/attributes. But don't think this is really feasible.

pylint/utils/checkerstats.py
```diff
@@ -1,30 +1,352 @@
...
+    convention: int
+    error: int
+    fatal: int
+    info: int
+    refactor: int
+    statement: int
+    warning: int
+
+
+# pylint: disable-next=too-many-instance-attributes
```
DanielNoord, line 64: Is this a real issue here?

pylint/lint/pylinter.py
Pierre-Sassoulas: Let's keep it that way then.

pylint/lint/pylinter.py
Pierre-Sassoulas: You're right I thought it was for our own function.

pylint/utils/checkerstats.py
Pierre-Sassoulas: I was thinking of a litteral line break:

```python
    return f"""{foo}
{bar}
... 
"""

DanielNoord: Fixed!

DanielNoord:
I have changed the name of the object to `LinterStats` because I noticed that after the refactor we only initialise the object on the linter object.

Pierre-Sassoulas (approved):
Great refactor, I think this will greatly reduce the risk when changing this part of the code.

PR #5079, opened 2021-09-27
Fix ``invalid-name`` for overwritten base ``object`` methods

### LINKED ISSUE #3614: No message for badly styled __init__ method arguments
### Steps to reproduce
1. create file `foo_bar.py`:
```python
class FooBar:
    def __init__(self, fooBar):
        self._foo_bar = fooBar

    def foo_bar(self, foo_bar_arg):
        print(self._foo_bar + foo_bar_arg)

    def bar_bar(self, foo_bar_arg):
        print(self._foo_bar + foo_bar_arg)
```
2. Run: `pylint foo_bar`

### Current behavior

No complaints about the `fooBar` argument (should be `snake_case`).

### Expected behavior

Issues a message that `fooBar` is not in `snake_case` style.

The following example issues correctly `foo_bar.py:6:4: C0103: Argument name "fooBarArg" doesn't conform to `snake_case` naming style (invalid-name)` for the `foo_bar` method, but not for the `__init__` method.

```python
class FooBar:
    def __init__(self, fooBar):
        self._foo_bar = fooBar

    def foo_bar(self, fooBarArg):
        print(self._foo_bar + fooBarArg)

    def bar_bar(self, foo_bar_arg):
        print(self._foo_bar + foo_bar_arg)
```

Edit: it does so with the default settings and with my custom settings with I have added as attachement.
[pylintrc.txt](https://github.com/PyCQA/pylint/files/4616408/pylintrc.txt)

### pylint --version output

```
pylint 2.5.2
astroid 2.4.1
Python 3.6.8 (default, Jan 14 2019, 11:02:34)
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
```

DanielNoord:

## Type of Changes

## Description

`overrides_a_method()` was a little too strict for the `invalid-name` checker as it filtered out overwritten methods of the base `object` class. Therefore, `__init__` would never be checked.

This `utils` function is only used twice, the other instance being here:
https://github.com/PyCQA/pylint/blob/5e24aaa4dde2e3f06027cd0e18fabd96e408f519/pylint/checkers/classes.py#L1409-L1434

Based on the fact that the tests pass and the message that is being checked there I think it is safe to say that excluding overwriting `object` methods is also fine there.

Closes #3614

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -47,6 +47,10 @@ Release date: TBA
 
   Closes #5066
 
+* Fixed ``invalid-name`` not checking parameters of overwritten base ``object`` methods
+
+  Closes #3614
+
 
 What's New in Pylint 2.11.1?
 ============================
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -456,8 +456,11 @@ def assign_parent(node: nodes.NodeNG) -> nodes.NodeNG:
 
 
 def overrides_a_method(class_node: nodes.ClassDef, name: str) -> bool:
-    """return True if <name> is a method overridden from an ancestor"""
+    """return True if <name> is a method overridden from an ancestor
+    which is not the base object class"""
     for ancestor in class_node.ancestors():
+        if ancestor.name == "object":
+            continue
         if name in ancestor and isinstance(ancestor[name], nodes.FunctionDef):
             return True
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas (approved):
👍

PR #5084, opened 2021-09-28
Fix empty ``no_docstring_rgx`` not working correctly

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [x] If it's a new feature, or an important bug fix, add a What's New entry in
      `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |
| ✓   | :hammer: Refactoring   |

## Description

Turns out we did not handle empty regex options correctly. The referenced issue was broken because whenever the `no_docstring_rgx` option was empty it would try to match `re.compile("")` which always matched and would thus never check a parameter.

The change to `get_global_option()` is necessary for it to actually return a `Pattern[str]` for those options. A problem I couldn't solve was that it is impossible (I think) to tell `mypy` that the option belongs to `GLOBAL_OPTION_PATTERN` and should therefore return a `Pattern[str]`. This is because you can't use `isinstance` on `Literal` and there is no good solution to do this any other way (using `cast` also doesn't work).

/CC @cdce8p Because you helped with #4978 and might have a solution for the `mypy` issue that I don't know about 😄 

Closes #4136

PR #5088, opened 2021-09-28
Make test decorator use ``checker.set_option()``

DanielNoord:

## Type of Changes

## Description

While working on #5084 I noticed our current unittests don't actually fully test the `set_option` behaviour and we are therefore not able to correctly test some settings.
We need to call `set_option` to make the settings pass the validators and actually set them like they would have been in normal circumstances.

After adding this I found that we are actually missing a lot of options in the `options` attributes of some of the checkers. `docparams` which is needed for #5084 is missing `no_docstring_rgx` for example. We would need to add those for this to fully work, but I think they shouldn't be missing anyway.

I would like to know if this assertion is correct as I can then start work on adding the missing settings to the checkers.

### DIFF
--- a/pylint/testutils/decorator.py
+++ b/pylint/testutils/decorator.py
@@ -2,12 +2,50 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 import functools
+import optparse  # pylint: disable=deprecated-module
 
+from pylint.lint import PyLinter
 from pylint.testutils.checker_test_case import CheckerTestCase
 
 
 def set_config(**kwargs):
     """Decorator for setting config values on a checker.
+
+    Passing the args and kwargs back to the test function itself
+    allows this decorator to be used on parametrized test cases.
+    """
+
+    def _wrapper(fun):
+        @functools.wraps(fun)
+        def _forward(self, *args, **test_function_kwargs):
+            try:
+                for key, value in kwargs.items():
+                    self.checker.set_option(key.replace("_", "-"), value)
+            except optparse.OptionError:
+                # Check if option is one of the base options of the PyLinter class
+                for key, value in kwargs.items():
+                    self.checker.set_option(
+                        key.replace("_", "-"),
+                        value,
+                        optdict=dict(PyLinter.make_options()),
+                    )
+            if isinstance(self, CheckerTestCase):
+                # reopen checker in case, it may be interested in configuration change
+                self.checker.open()
+            fun(self, *args, **test_function_kwargs)
+
+        return _forward
+
+    return _wrapper
+
+
+def set_config_directly(**kwargs):
+    """Decorator for setting config values on a checker without validation.
+
+    Some options should be declared in two different checkers. This is
+    impossible without duplicating the option key. For example:
+    "no-docstring-rgx" in DocstringParameterChecker & DocStringChecker
+    This decorator allows to directly set such options.
 
     Passing the args and kwargs back to the test function itself
     allows this decorator to be used on parametrized test cases.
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Nice catch !

DanielNoord:
Perhaps the second commit should be part of a separate PR but I thought that with a clean commit history these can be merged directly and it wouldn't be too much of a problem.

Turns out the fix to the failing tests was a little different than initially expected. We know check whether any of the options that are not immediately found on the `checker` being tested is actually part of `PyLinter`'s options. If that's not the case the tests will fail and the contributor will need to rewrite their tests.
One possibility in those cases is to use the `set_option_directly` wrapper, which allows bypassing the validators. This is useful in case the option is located on a base `checker` that's loaded normally but is not during testing.

In #5084 I'm going to work on the two tests that now use `set_option_directly` as I think they should pass compiled regex patterns instead if raw strings. For now, I think this change is fine as it keeps the scope of this PR limited without failing any tests.

pylint/testutils/decorator.py
```diff
@@ -13,6 +15,42 @@ def set_config(**kwargs):
     allows this decorator to be used on parametrized test cases.
     """
 
+    def _wrapper(fun):
```
Pierre-Sassoulas, line 18:
```suggestion
    def _wrapper(fun: typing.Callable) -> typing.Callable:
```
Maybe, it's overkill and we don't need typing for inner function ?

pylintrc
```diff
@@ -27,6 +27,10 @@ load-plugins=
 # Use multiple processes to speed up Pylint.
 jobs=1
 
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
```
Pierre-Sassoulas, line 32: 👍

pylintrc
```diff
@@ -272,7 +276,7 @@ ignore-mixin-members=yes
 
 # List of module names for which member attributes should not be checked
 # (useful for modules/projects where namespaces are manipulated during runtime
-# and thus existing member attributes cannot be deduced by static analysis
+# and thus existing member attributes cannot be deduced by static analysis)
```
Pierre-Sassoulas, line 279: I think this is generated, so it needs to be fixed elsewhere too.

DanielNoord: I think it is generated from this file. I tested ``--generate-rc`` and it just copies the base `pylintrc` found in `./` to `stdout`.

pylint/testutils/decorator.py
DanielNoord: Think we might want to hold off on this when we type `testutils` completely. Adding a general `Callable` doesn't add too much info right now imo.

pylint/testutils/decorator.py
```diff
@@ -2,7 +2,9 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 import functools
+import optparse  # pylint: disable=deprecated-module
```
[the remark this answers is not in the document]
DanielNoord: Yeah, I looked into it already. Requires some work but should be doable.

pylintrc
Pierre-Sassoulas: Wow, I wasn't careful enough with that file.

DanielNoord: Btw, that's also why `suggestion-mode` was missing I think. If it isn't added to `pylintrc` it doesn't get added anywhere.

Although, looking at the code I'm starting to doubt this... Anyway, might as well fix it here 😄 

https://github.com/PyCQA/pylint/blob/20304a6cfb0dedec8ca7878822b8db1fd3cc71cc/pylint/config/option_manager_mixin.py#L191-L224

DanielNoord:
@Pierre-Sassoulas Is that approval still valid? 😄

Pierre-Sassoulas (approved):
Yes :)

Issue #3507, opened 2020-04-24
False positive isinstance-second-argument-not-valid-type when testing typing.Iterable


mthuurne:
### Steps to reproduce

```py
from typing import Iterable

print(isinstance([], Iterable))
```

Similar issues exist for other generic collection types such as `Sized`, `Mapping` etc.

### Current behavior

```
Second argument of isinstance is not a type (isinstance-second-argument-not-valid-type)
```

### Expected behavior

No warning is issued.

When `collections.abc.Iterable` is used instead, there is no warning issued. But the way that class is aliased in the `typing` module is probably too complex for static analysis to deal with and there doesn't seem to be explicit support yet to work around that.

### pylint --version output

```
$ pylint --version
pylint 2.5.0-dev1
astroid 2.4.0
Python 3.8.2 (default, Mar 05 2020, 18:58:42) [GCC]
```
The `astroid` version is a dev snapshot as well, but not reported as such. To be exact, I'm using `pylint` at 0ccf1bca and `astroid` at f6a36d91.

PCManticore:
Thanks, I can confirm the bug. It seems we're inferring `typing.Iterable` as an instance rather than a class.

bertrand-benoit:
I confirm the bug with typing.Dict.

LuisMarquesSerco:
Also with typing.List

Coffeeri:
Also with `typing.Generator`.

eeroniemi:
Also happened with `typing.Match`, pylint version 2.5.2.

cdce8p:
This issue will be fixed with the next release of `pylint`. We improved the handling of typing aliases (`Iterable`, `Generator`, `List`, `Match`, and so on) in `astroid` recently.

kosyachniy:
I have the same issue with `typing.Callable`

cdce8p:
> I have the same issue with `typing.Callable`

@kosyachniy Thanks for the report! This will be fixed in the next release of astroid / pylint.
The PR for reference: https://github.com/PyCQA/astroid/pull/1192

PR #5089, opened 2021-09-28
Add regression test for 3507

cdce8p:
Depends on: https://github.com/PyCQA/astroid/pull/1192
Closes: #3507


tests/functional/r/regression/regression_3507_typing_alias_isinstance.py
```diff
@@ -0,0 +1,24 @@
...
+import collections
+import collections.abc
+import typing
+
+isinstance(42, typing.Dict)
+isinstance(42, typing.Counter)
+isinstance(42, typing.Collection)
+isinstance(42, typing.Iterator)
+isinstance(42, typing.Tuple)
+isinstance(42, typing.Callable)
```
Pierre-Sassoulas, line 15: Maybe also typing.List and some other are missing ? Should this be exhaustive ?

Pierre-Sassoulas (approved):

tests/functional/r/regression/regression_3507_typing_alias_isinstance.py
cdce8p: `typing.List` is similar to `typing.Dict` (which is tested). It wasn't my intention to add all possible aliases here. That would get out of hand quickly. Just a few selected ones.

PR #5093, opened 2021-09-29
Fix empty regex options not converting to regex

DanielNoord:

## Type of Changes

## Description

Whenever a regex option is `None` we do not correctly convert those to patterns.
This also add additional typing and fixes a related issue with one of the tests.

Let me know if you would rather have this be 3 separate commits. It is a bit of a weird collection of `regex option` related stuff.


Pierre-Sassoulas (approved):
👍

PR #5094, opened 2021-09-29
Improve node info for invalid-name

cdce8p:
## Description

At the moment `invalid-name` for a function argument is emitted on the `function` node. That does work, however it limits where `pylint: disable` can be set. Consider the following example:
```py
class SomeClass:
    def func(
        self,
        var,
        invalidName,
    ):
        pass
```

Only `invalidName` is wrong. With this change it's now possible to add `pylint: disable` directly behind it instead of after `def func`.

This appears to be fully backwards compatible. I.e. both with disable `invalid-name`:
```py
# old
class SomeClass:
    def func(  # pylint: disable=invalid-name
        self,
        var,
        invalidName,
    ):
        pass

# new
class SomeClass:
    def func(
        self,
        var,
        invalidName,  # pylint: disable=invalid-name
    ):
        pass
```

### DIFF
--- a/tests/functional/i/invalid/invalid_name.py
+++ b/tests/functional/i/invalid/invalid_name.py
@@ -74,6 +74,7 @@ a_very_very_very_long_function_name_WithCamelCase_to_make_it_sad()
 class FooBar:
     def __init__(self, fooBar) -> None:  # [invalid-name]
         self.foo_bar = fooBar
+        self.foo_bar2 = None
 
     def func1(
         self,
@@ -93,3 +94,12 @@ class FooBar:
 
     def test_disable3(self, fooBar):  # pylint: disable=invalid-name
         self.foo_bar = fooBar
+
+    def test_disable_mixed(
+        self,
+        fooBar,  # pylint: disable=invalid-name
+        fooBar2,  # [invalid-name]
+    ):
+        """Invalid-name will still be raised for other arguments."""
+        self.foo_bar = fooBar
+        self.foo_bar2 = fooBar2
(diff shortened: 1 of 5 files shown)


cdce8p:
I would like to target `2.11.2` for this one if thats ok.

Pierre-Sassoulas:
> I would like to target 2.11.2 for this one if that's ok.

This does not feel like a "fix" but so does all the typing we already added in 2.11.2. I feel like we have already a lot in 2.11.2 and could practically release 2.12 instead at this point.

cdce8p:
> This does not feel like a "fix" but so does all the typing we already added in 2.11.2. I feel like we have already a lot in 2.11.2 and could practically release 2.12 instead at this point.

That might indeed be the better solution 😅
Technically it is a new feature, but since it's fully backwards compatible it could be ok.

tests/functional/i/invalid/invalid_name.py
```diff
@@ -72,5 +72,24 @@ def a_very_very_very_long_function_name_WithCamelCase_to_make_it_sad():  # Shoul
...
+    ):
+        self.foo_bar = fooBar
+
+    # Test disable invalid-name
+    def test_disable1(self, fooBar):  # pylint: disable=invalid-name
+        self.foo_bar = fooBar
+
+    def test_disable2(
+        self,
+        fooBar,  # pylint: disable=invalid-name
```
Pierre-Sassoulas, line 90:
```suggestion
        bar,  # [unused-argument] 
        fooBar,  # pylint: disable=invalid-name
        barFoo,  # [invalid-name, unused-argument]
```

cdce8p: I added another case: `test_disable_mixed`. Let me know if that is what you wanted to test here.

Pierre-Sassoulas: Yes perfect :)

Pierre-Sassoulas (approved):

PR #5095, opened 2021-09-29
Fix empty ``no_docstring_rgx`` not working correctly

### LINKED ISSUE #4136: __init__ arguments are not checked correctly by docparams
Hi there!

I noticed that pylint `docparams` seems not to check whether the __init__ arguments are documented. This seems pretty weird, but here's a code sample to reproduce.

For reference, I'm using this pylintrc file:

```
[MASTER]
disable=
    missing-module-docstring,
    too-few-public-methods,
load-plugins=
    pylint.extensions.docparams,
accept-no-param-doc=no
```

### Code Sample
```
class MyClass:
    """
    My Class Docstring
    """

    def __init__(self, my_param: int, my_second_init_var: str, my_third_init_var: str) -> None:
        """
        My init docstring
        :param my_param: My first param
        """

    def my_method(self, first_param: int, second_param: str) -> None:
        """
        My Docstring
        :param first_param: The first param.
        :param second_param: The second param.
        """

```

### Current behavior
The code is rated 10/10

### Expected behavior
Some warnings should be issued because `my_second_init_var` and `my_third_init_var` are not present in either the __init__ docstring or the class docstring

### pylint --version output
pylint 2.7.0
astroid 2.5
Python 3.8.1 (default, Jun 30 2020, 11:53:47) 
[Clang 11.0.0 (clang-1100.0.33.17)]

DanielNoord:

## Type of Changes

## Description

I wonder why the default value for `no_docstring_rgx` was `^_`.
Perhaps there was a logic behind it, but users such as the one who initially reported this bug will never figure out that they need to add `no_docstring_rgx=` to their `pylintrc` for it to check all functions. I think changing it to be empty is more in line with what users will expect the default value of a regex pattern to be.

I would like confirmation on changing this default value before adding a commit that fixes the tests that fail because of the change.

Furthermore, it seems like the unittests are really prone to incorrectly testing settings. The change to `"no_docstring_rgx": re.compile("")`, in `tests/extensions/test_check_docs.py` shows that we might set settings to certain values with incorrect types and therefore passing/failing tests incorrectly. This is not something we can fix immediately but is something to keep in mind!

Closes #4136


cdce8p:
> I wonder why the default value for `no_docstring_rgx` was `^_`.

You removed the comment explaining it. IMO this does indeed make sense. Docstrings by default should only be **required** by public methods. E.g. to document an API. `private` / `protected` methods are implementation details, thus it's not strictly necessary for them to have a docstring.
```py
# do not require a doc string on private/system methods
```

> Perhaps there was a logic behind it, but users such as the one who initially reported this bug will never figure out that they need to add `no_docstring_rgx=` to their `pylintrc` for it to check all functions.

That's probably something that should be better explained in the docs. Just changing the default here will effect a lot more people for whom this works perfectly fine.

Pierre-Sassoulas:
> I wonder why the default value for no_docstring_rgx was ^_.

It's a convention that you can start a variable name with ``_`` to not have to document. I don't know how many user rely on it, but I'm kinda hesitant to change the default value, because it means the behavior of pylint will change unexpectedly and this could be a problem for those relying on thisbehavior (unknown proportion) and not reading the changelog religiously (probably 98%).

DanielNoord:
I understood the type of functions exempted, but did not know this was a convention. However, shouldn't we perhaps exclude `__init__` from the regex? IMO it makes sense to want documentation of the parameters there. Especially when users go out of their way to add the extension to their settings, they are probably looking for more checkers to enforce stricter do string rules.
On the other hand, excluding certain cases might become too difficult and even less intuitive.

cdce8p:
I think what this comes down to is that we would need to do a better job explaining the different config options.
I.e. a new `User guide` for the `docstring` settings.

Somewhat similar to this one here for NamingStyles: https://pylint.pycqa.org/en/latest/user_guide/options.html
Although I don't really like that the page is titled `Configuration`.

DanielNoord:
I'll work on this a bit more and see if I can update the docs to be clearer. Marking this a draft in the meantime. Could also close as not to clutter the open PR list. Both work for me!

cdce8p:
@DanielNoord I looked at it some more as I was wondering which regex would work. Turns out it doesn't matter. A missing docstring in `__init__` will **never** be marked as technically it's an overwritten method from `object`. I.e. `report_missing` is always `False`.
https://github.com/PyCQA/pylint/blob/8c74530e677ad2e8332f5ab0e7e64516571aae40/pylint/checkers/base.py#L2206-L2214

I disabled `report_missing` just for testing. In that case, this would work:
```ini
[BASIC]
no-docstring-rgx=^(?!__init__$)_
```

DanielNoord:
Hmm, but with the changes to `docparams` this test (added in this PR) does pass:
```python
@set_config_directly(no_docstring_rgx=re.compile(""))
def test_fail_empty_docstring_rgx(self) -> None:
   """Example of a function that matches an empty 'no-docstring-rgx' config option
   An error message is emitted.
   """
   node = astroid.extract_node(
       """
   #pylint disable=missing-module-docstring, too-few-public-methods,
   class MyClass:
       def __init__(self, my_param: int) -> None:
           '''
           My init docstring
           '''
   """
   )
   with self.assertAddsMessages(
       MessageTest(
           msg_id="missing-param-doc", node=node.body[0], args=("my_param",)
       ),
   ):
       self.checker.visit_functiondef(node.body[0])
```

I think this is also why the setting is somewhat problematic. I agree that a private method probably doesn't require a docstring explaining its functionality, but parameter documentation might still be useful/desired (`missing-param-doc`).

cdce8p:
I was only taking about `missing-function-docstring` for `__init__`.

DanielNoord:
Understood, but the `visit_functiodef` of `docparams` depends on the same option. Therefore, by rightfully blocking `missing-function-docstring` on private methods we are also blocking all checkers on `FunctionDef` from `docparams`. 

https://github.com/PyCQA/pylint/blob/8c74530e677ad2e8332f5ab0e7e64516571aae40/pylint/extensions/docparams.py#L211-L222

Note that in our own `pylintrc` we have already overwritten this to be a little less strict:

https://github.com/PyCQA/pylint/blob/8c74530e677ad2e8332f5ab0e7e64516571aae40/pylintrc#L259-L261

We could also create a new setting `no-docstring-rgx-docparams` which is specific to the extension to avoid this double dependency. This would also eliminate the problem with overlapping setting names in two different checkers fixed in #5088 with the addition of the `set_config_directly()` decorator.

cdce8p:
> > Besides that, it could be an idea to document the `no-docstring-rgx=^(?!__init__$)_` solution somewhere.
> 
> I think this is crucial, but I don't know the best approach so I'm open to suggestions. We need a place to tell users that by default `docparams` does not check all functions and a `no-docstring-rgx` setting set to `""` is indeed to force it do so.

Wouldn't this be a good place for it? https://pylint.pycqa.org/en/latest/technical_reference/extensions.html#parameter-documentation-checker
I would also recommend to add the `__init__` example to the `no-docstring-rgx` setting documentation: https://pylint.pycqa.org/en/latest/technical_reference/features.html#basic-checker-options

PR #5097, opened 2021-09-30
Properly identify undocumented parameters and add new message called missing-any-param-doc

### LINKED ISSUE #3799: Parameter Documentation checker only works if at least one parameter is documented properly
### Steps to reproduce
1. Enable `pylint.extensions.docparams`
2. Any of the following functions will be rated 10/10
```python
def foobar1(arg1, arg2):
    """function foobar ...
    """
    print(arg1, arg2)

def foobar2(arg1, arg2):
    """function foobar ...

    Parameters
    ----------
    """
    print(arg1, arg2)

def foobar3(arg1, arg2):
    """function foobar ...

    Parameters
    ----------
    arg1: int
    """
    print(arg1, arg2)
```
3. Run pylint

### Current behavior
The checker doesn't raise any warnings related to the parameters on the documentation
```
```

### Expected behavior
Something similar to this:

```python
def foobar4(arg1, arg2):
    """function foobar ...

    Parameters
    ----------
    arg1: int
        description
    """
    print(arg1, arg2)
```

That will return the warnings for any missing parameter:
```
************* Module pylint_config
pylint_config.py:1:0: W9015: "arg2" missing in parameter documentation (missing-param-doc)
pylint_config.py:1:0: W9016: "arg2" missing in parameter type documentation (missing-type-doc)

```

### pylint --version output
```
pylint 2.6.0
astroid 2.4.2
Python 3.6.4 (default, Jul 22 2019, 12:18:30)
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)]
```

ksaketou:

## Type of Changes

## Description
This PR fixes the problem where undocumented parameters where not being spotted correctly 
by `missing-param-doc` and `missing-type-doc`. Also, a new message has been added called 
`missing-any-param-doc` which is triggered when a function has neither parameter nor parameter
type documentation.

## Related issue

Closes #3799


pylint/extensions/docparams.py
```diff
@@ -145,6 +145,11 @@ class DocstringParameterChecker(BaseChecker):
             "useless-type-doc",
             "Please remove the ignored parameter type documentation.",
         ),
+        "W9021": (
+            'Missing any documentation in "%s"',
+            "missing-any-param-doc",
+            "Please add parameter and/or type documentation.",
+        ),
```
Pierre-Sassoulas, line 152: We could add an old names here, the ideal would be to not force to disable missing-any-param when the old one was already disabled. But they are not really equivalent so maybe you were right to not add it.

tests/functional/ext/docparams.py
```diff
@@ -16,7 +16,7 @@ def _private_func3(param1):  # [missing-raises-doc]
     raise Exception('Example')
 
 
-def public_func1(param1):  # [missing-param-doc, missing-type-doc]
+def public_func1(param1):  # [missing-any-param-doc]
```
Pierre-Sassoulas, line 19: As we can't use old names we should warn in whats new for 2.12 that this can happen.

Pierre-Sassoulas (commented):
I don't have enough time for a full review, just a small remark about documentation :)

pylint/extensions/docparams.py
ksaketou: what does the old names contain exactly?

Pierre-Sassoulas: You can search for C0111 (the old name for missing-function-docstring) for an example :)

ksaketou: So you say that instead of adding an old names, we can add a warning in the 2.12 what's new file?

Pierre-Sassoulas: Yes, something like "A warning might appear if you disabled  ``missing-param-doc``and ``missing-type-doc`` because it became ``missing-any-param-doc`` when there isn't any doc."

ksaketou: I added my changes in the what's new for 2.11 and under 2.11.2 in ChangeLog, should I move them to 2.12 then?

Pierre-Sassoulas: Yes sorry I did not catch that, this is a "new feature" :)

ksaketou:
Thank you!:)

PR #5106, opened 2021-10-03
Add support for `ModuleNotFoundError` Python 3.6+ builtin-exception

HorlogeSkynet:
## Type of Changes

## Description

Dear maintainers, this PR adds (a pretty naive) support for `ModuleNotFoundError` Python 3.6+ builtin-exception.
Currently, `import-error` as well as `no-name-in-module` messages are wrongly raised when a `try..except` block catches this `ImportError` subclass.

Feel free to tweak this branch as you need.

---

Bye :wave:


pylint/checkers/utils.py
```diff
@@ -943,7 +943,12 @@ def is_from_fallback_block(node: nodes.NodeNG) -> bool:
         for import_node in other_body
     )
     ignores_import_error = _except_handlers_ignores_exception(handlers, ImportError)
-    return ignores_import_error or has_fallback_imports
+    ignores_module_not_found_error = _except_handlers_ignores_exception(
```
DanielNoord, line 946: https://github.com/PyCQA/pylint/blob/86c073e49fc099d8111da96cfbb2f30ccd659338/pylint/checkers/utils.py#L949-L953
Looking at `_except_handlers_ignores_exception` the `exception` parameter is actually made to be a `tuple`. I think it makes sense to merge this with `L945` and pass `(ImportError, ModuleNotFoundError)`.

While you're at it it, it would be appreciated if you could add the typing to the `exception` parameter. I think it should be:
`exception: Tuple[Type[ImportError], Type[ModuleNotFoundError]]`

tests/functional/n/no/no_name_in_module.py
```diff
@@ -30,6 +30,11 @@
 except ImportError:
     something = None
 
+try:
+    from collections import something
+except ModuleNotFoundError:
```
DanielNoord, line 35: Could you recheck some of these tests? Some are non-sensical (such as this one) in that they never occur in actual runtime `python`.
Since `collections` can be found this will/should never raise a `ModuleNotFoundError` and only raise a `ImportError`. I think we shouldn't bloat our tests with unfeasible examples.

DanielNoord (changes requested):
Thanks for this fix! This is certainly good for the `import-error` checker. Just two 2 changes that should make this mergeable!

pylint/checkers/utils.py
HorlogeSkynet: You're right, I've missed the `tuple` ! I took the liberty to rename the function as well as its second parameter according to its new behavior.

tests/functional/n/no/no_name_in_module.py
HorlogeSkynet: Indeed, sorry about this one. I was not sure about the other ones so I left them for the time being.

Pierre-Sassoulas (commented):
Let's wait for 2.12 for this one :)

pylint/checkers/utils.py
DanielNoord: Good idea!

tests/functional/n/no/no_name_in_module.py
DanielNoord: I think the others are good!

DanielNoord (approved):
Thanks for the rework! This looks good to me for `2.12`!

DanielNoord:
Thank you for your work @HorlogeSkynet!

PR #5107, opened 2021-10-03
Fix return type checkers calls on ellipses functions

### LINKED ISSUE #4736: invalid-index-returned false positive in stubs
The following file:
```python3
class A:
    def __index__(self) -> int: ...
```

produces the following output:

```python3
$ pylint A.pyi
************* Module A
A.pyi:2: [E0305(invalid-index-returned), A.__index__] __index__ does not return int
```

This looks like a false positive.

DanielNoord:

## Type of Changes

## Description

Closes #4736

### DIFF
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -2168,7 +2168,11 @@ class SpecialMethodsChecker(BaseChecker):
 
         inferred = _safe_infer_call_result(node, node)
         # Only want to check types that we are able to infer
-        if inferred and node.name in self._protocol_map:
+        if (
+            inferred
+            and node.name in self._protocol_map
+            and not self._function_body_is_ellipsis(node)
+        ):
             self._protocol_map[node.name](node, inferred)
 
         if node.name in PYMETHODS:
@@ -2297,6 +2301,17 @@ class SpecialMethodsChecker(BaseChecker):
                     return True
                 except astroid.NotFoundError:
                     pass
+        return False
+
+    @staticmethod
+    def _function_body_is_ellipsis(node: nodes.FunctionDef) -> bool:
+        if (
+            len(node.body) == 1
+            and isinstance(node.body[0], nodes.Expr)
+            and isinstance(node.body[0].value, nodes.Const)
+            and node.body[0].value.value == Ellipsis
+        ):
+            return True
         return False
 
     def _check_iter(self, node, inferred):
--- a/tests/functional/c/class_protocol_ellipsis.py
+++ b/tests/functional/c/class_protocol_ellipsis.py
@@ -1,0 +1,41 @@
+""""Tests for return type checkers for protocol methods with ellipsis function body"""
+# pylint: disable=missing-class-docstring
+from typing import Any, Iterator
+
+
+class MyClass:
+    def __len__(self) -> int:
+        ...
+
+    def __hash__(self) -> int:
+        ...
+
+    def __index__(self) -> int:
+        ...
+
+    def __iter__(self) -> Iterator[Any]:
+        ...
+
+    def __bool__(self) -> bool:
+        ...
+
+    def __repr__(self) -> bool:
+        ...
+
+    def __str__(self) -> str:
+        ...
+
+    def __bytes__(self) -> bytes:
+        ...
+
+    def __length_hint__(self) -> int:
+        ...
+
+    def __format__(self, format_spec: str) -> str:
+        ...
+
+    def __getnewargs__(self) -> tuple:
+        ...
+
+    def __getnewargs_ex__(self) -> tuple:
+        ...
--- a/ChangeLog
+++ b/ChangeLog
@@ -55,6 +55,10 @@ Release date: TBA
 
   Closes #3614
 
+* Prevent return type checkers being called on functions with ellipses as body
+
+  Closes #4736
+
 
 What's New in Pylint 2.11.1?
 ============================


pylint/checkers/classes.py
```diff
@@ -2299,6 +2303,17 @@ def _is_iterator(node):
...
+    @staticmethod
+    def _function_body_is_ellipsis(node: nodes.FunctionDef) -> bool:
+        if (
+            len(node.body) == 1
+            and isinstance(node.body[0], nodes.Expr)
+            and isinstance(node.body[0].value, nodes.Const)
+            and node.body[0].value.value == Ellipsis
+        ):
+            return True
+        return False
```
cdce8p, line 2315:
```suggestion
        return (
            len(node.body) == 1
            and isinstance(node.body[0], nodes.Expr)
            and isinstance(node.body[0].value, nodes.Const)
            and node.body[0].value.value == Ellipsis
        )
```
No need to add an if statement.

pylint/checkers/classes.py
```diff
@@ -2299,6 +2303,17 @@ def _is_iterator(node):
                     pass
         return False
 
+    @staticmethod
+    def _function_body_is_ellipsis(node: nodes.FunctionDef) -> bool:
```
cdce8p, line 2307: What do you think about moving this method to `checkers/utils.py`?

cdce8p, line 2307: I would also suggest to rename it slightly to: `is_function_body_ellipsis`.

tests/functional/c/class_protocol_ellipsis.py
```diff
@@ -0,0 +1,41 @@
+""""Tests for return type checkers for protocol methods with ellipsis function body"""
+# pylint: disable=missing-class-docstring
+from typing import Any, Iterator
+
+
+class MyClass:
```
cdce8p, line 6: Can you add a comment above that line with:
* The link to the Github issue
* Mentioning that `invalid-*-returned` warnings should **not** be emitted for stub functions.

pylint/checkers/classes.py
Pierre-Sassoulas: Shouldn't it be in astroid's FunctionDef instead ?

cdce8p: > Shouldn't it be in astroid's FunctionDef instead ?

Though about this some more. I'm not sure we should move these kinds of helper functions to `astroid`. `astroid` only really deals with the `ast` and inference, everything else should be done by `pylint`.

By the same logic I believe it might have been wrong to move `is_sys_guard` and `is_typing_guard` to `astroid` in the first place. They don't really belong there.

DanielNoord: Couldn't it be argued that these methods are part of the inference? In that they give additional information about the specific type of `FunctionDef` node?
Anyway, I'll leave this up to you and Pierre. I'm fine with either. 
If wanted I can also make a PR to re-move `is_sys_guard` and `is_typing_guard` to `pylint`.

cdce8p: > Couldn't it be argued that these methods are part of the inference? In that they give additional information about the specific type of `FunctionDef` node?

True. Personally I would consider this a case of meta inference. We check multiple conditions to decide wether or not to emit a warning. That's that sort of thing `pylint` does all the time.

One thing to consider with that, it's much easier to iterate (read: fix) such a check in `pylint` alone than to modify something in `astroid` + `pylint`.

What I do consider useful in `astroid` are helpers like this one https://github.com/PyCQA/astroid/pull/1169 which helps dealing with the `NodeNG` tree itself.


> Anyway, I'll leave this up to you and Pierre. I'm fine with either.
> If wanted I can also make a PR to re-move `is_sys_guard` and `is_typing_guard` to `pylint`.

I would be in favor of it, but let's see what @Pierre-Sassoulas thinks first. Unfortunately, we can't outright remove them now, so we would need to deprecate them first (and remove with astroid 3.0).

DanielNoord: I think your second argument might tip the scale (it does for me anyway). Both places seem reasonable but the benefit of being able to tweak these checks by only updating `pylint` is indeed quite good.
If Pierre agrees I would:
1. Fix this PR from code review while keeping method in `pylint`
2. Create PR which re-adds `is_sys_guard` and `is_typing_guard` to `pylint`
3. Create deprecating PR for `is_sys_guard` and `is_typing_guard` to be merged after `2` is merged into a release of `pylint` (so users can actually use the new methods)

cdce8p: Sounds good to me! As it's only a deprecation `3` wouldn't necessarily need to wait for `2`.
In principle I would say all of the helpers mentioned should probably be part of `checkers/utils.py` (if no circular dependency gets in the way).

Pierre-Sassoulas: It make sense. It's easy to see that when something rely on astroid any pylint modification slow down by a factor of 10. Could we find a middle ground and not remove what is already in astroid but not add anything either ? It's less work :smile: (for us but also for dev using astroid as a library)

DanielNoord: I think the middle ground just make the codebase inconsistent. We already have some deprecation warnings in both `astroid` and `pylint` for a next major release so we need to take a look at these anyway (so the extra work is relatively minimal). 
Furthermore, by doing step 3 after step 2 we can direct users to the "new" `pylint` methods if they want to keep using them. That would require minimal work for them as well.

Pierre-Sassoulas: Well you're right about consistency but there are developpers using astroid without using pylint, removing this from astroid could be problematic if they started using the typing guard in their code. Also, I think the main problem is for project that expects any pylint to work with any astroid and are setting astroid manually (prospector being one of them until very recently). Maybe it will be the occasion for them to let pylint handle its own dependencies.

DanielNoord: So, we would leave the situation as is?

Pierre-Sassoulas: I'm not advocating for one solution or the other, consistency is important, we just need to consider the other factors. What do you think @cdce8p ?

cdce8p: I would probably prefer to move these back to pylint, for consistency. If we don't do it, someone else might come along and think it's accepted to add other methods there as well.

Furthermore, as we've only added it recently, there might only be a few projects that use it. Most of them can import it from `pylint` without issues.

DanielNoord: Do you mean first release `2.8.1` and afterwards deprecate, or the other way around?

cdce8p: I guess, let's do the deprecation now. I.e. wait with releasing astroid `2.8.1` until it's done.
https://github.com/PyCQA/astroid/issues/1199

Pierre-Sassoulas: I opened https://github.com/PyCQA/astroid/pull/1202

cdce8p: Opened #5122 to add the guard helper functions back to pylint.

PR #3821, opened 2020-09-07
Less false positives for len-as-conditions (pandas's DataFrame, numpy's Array)

Pierre-Sassoulas:
## Description

`len-as-conditions` is now triggered only for classes that are inheriting directly from list, dict, or set and not implementing the `__bool__` function, or from generators like range or list/dict/set comprehension. This should reduce the false positive for other classes, like pandas's DataFrame or numpy's Array. See discussion in related issue.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |

## Related Issue

#1879

Issue #3031, opened 2019-07-28
Pylint should not report literal-comparison when testing for tuple


chaoflow:
### Steps to reproduce
```python
def foo(bar=()):
    if bar is not ():
        ...
```

### Current behavior

pylint reports "literal-comparison"

### Expected behavior

pylint should regard the "is (not)" comparison correctly.

### pylint --version output
```
pylint 2.3.1
astroid 2.2.5
Python 3.7.3 (default, Apr  3 2019, 19:16:38) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
```

PCManticore:
Yeah, we can probably exempt tuples from the check.

chaoflow:
@PCManticore I think it's noteworthy that only starting with Python 3.7 ``(1,) is (1,)`` before this was only true for the empty tuple.

chaoflow:
Following up on this, I think only the empty tuple should be exempted.

compare

```python
tuple(range(0)) is ()
tuple(range(1)) is (0,)
```

PR #5120, opened 2021-10-05
Use implicit booleanness

### LINKED ISSUE #4774: Add checker for comparing for equality to empty literal
### Current problem

_Originally posted by @cdce8p in https://github.com/PyCQA/pylint/pull/4769#discussion_r679096534_

Best not to check for equality by comparing it to an empty literal.
(This should probably be a pylint check as well.)

### Desired solution

```python
# good
test_list = []
if not test_list:
   pass

# bad
test_list = []
if test_list == []:
   pass
```

### Additional context

_No response_

jaehoonhwang:
## Type of Changes

## Description

Add a new checker; `use-implicit-booleanness` Emitted when using collection  literals for boolean comparisons

Closes #4774


Pierre-Sassoulas (changes requested):
This look really good already as a new check, thank you ! Is there a reason to not check dicts or sets ? I would think it would work the same way.

Also, there's some history in pylint regarding this type of check. Could we merge this with ``len-as-condition`` ? It's an already existing checker that warn to not use ``len(x) > 0`` and is also about implicit booleaness so it could be renamed to ``use-implicit-booleaness`` using the ``old_names`` attribute.

There will be some adaptation to do so it works for both checkers using the same utility function. Taking the functional tests for len-as-condition and adapting it for comparison with a literal would anticipate a lot of issues too. I think that this is one of the main advantages of this approach : a lot of false positives that would be almost similar are already "known" and handled because of this previous check (in particular pandas and numpy data frame do not have implicit booleanness see #3821 for more details.).

jaehoonhwang:
@Pierre-Sassoulas 
Quick question, do you want to the `len_checker` to combine two checks? (original `len` check and new `empty_literal` check?)

I have currently have implemented it as two separate checker inside of `len_check.py` since I couldn't find good explanation to cover both checks. 

Current two check;
1. `use-implicit-booleaness-len`
2. `use-implicit-booleaness-empty-literal`

Also, few thoughts have come into my minds.
* It seems like it might be a good idea to cover empty `String` version as well
Example:
```py
a = "abc"
if a == "":
    pass
```
* I think we might want to cover more than `==` or `!=` and include `>(=)` or `<(=)` and I think at that point we can implement it in two ways
    1. `if operator in ["=="....]`
    2. `if operator not in ["in"....]` 

Please let me know what you think!

Pierre-Sassoulas:
Maybe what make sense is to have common functions like  ``instance_has_bool`` and two (child?) classes for each check. Maybe group that in a implicit_booleaness_checker file ? I also agree that we could add strings to the list of "empty-litteral", as ``len("") == 0`` should raise a ``len-as-condition`` if it was not already. Regarding the use case I think the picture will be clearer once the use case for len-as-condition are transcribed in functional tests.

Regarding the name of the message, I was thinking originally of using ``use-implicit-booleanness`` for both, but I like your proposition to have two message with the same beginning too. Maybe ``use-implicit-booleaness-not-len`` and ``use-implicit-booleaness-not-comparison`` then ? Do you have an opinion about this @cdce8p  ?

jaehoonhwang:
New Commits;
Things to note;
I have tried with string literal checks, but even within `pylint` packages, there were codes that checks both type and the empty string. (Which is more concise, but giving different types of input might not be a good idea in general)

Example:

```py
a = ""
b = 1
c = [1,2,3]
def test(input):
    if input == "": 
          print("empty string")
          return
    print(f"something other than string: {input}")

def test_with_string_check(input):
    if isinstance(input, str) and input: 
          print("empty string")
          return
    print(f"something other than string: {input}")

test(a)
test(b)
test(c)
# versus

test_with_string_check(a)
test_with_string_check(b)
test_with_string_check(c)
```

Please let me know what you think! @Pierre-Sassoulas

jaehoonhwang:
having problem with rebase >_>; don't review just yet

jaehoonhwang:
@Pierre-Sassoulas do you want changes for #3031 in this commit?

I thought I would do another pr for that one

jaehoonhwang:
Updated with test case/documentation; for #3031

I just made it where it ignores `nodes.Tuple`

Pierre-Sassoulas:
It's because we don't install numpy or pandas before tests. It's probably ignored for the other functional test for implicit booleanness, we could do the same here.

jaehoonhwang:
It seems like `use-implicit-booleaness-not-len` is disabling `import-error` check.
Should I move it to the top rather than doing it before the test?

jaehoonhwang:
@Pierre-Sassoulas 
I have rewritten bypass check as per comments; 
now `use-implicit-booleaness-not-comparison` will not check; if a `variable` can't be inferred to `list`, `tuple`, or `dict` **AND** `variable` class has `__bool__` implemented.

`pandas` and `numpy` cases have more explanations added.

example:
```py
class A:
  def __init__(self):  self.a=1

class B:
  def __init__(self):  self.b=1
  def __bool__(self): return False

a, b = A(), B()
if a == []:  # [use-implicit-booleaness-not-comparison]
   print("a")
if b== []:
   print("b")
```

Hey, I also want to thank you for your patience 😄

Pierre-Sassoulas:
Great addition to pylint, this will be the main new feature for 2.12. Some hacktoberfest contributions are a little typo fix, and this one is clearly not that. Thank you for being so persistent :)

jaehoonhwang:
Omg :O It's finally here :partying_face: 
Can't believe this is finally made into main!
Thank you again for your patience and feedbacks! 
It was a great learning opportunity for me to work on a such big python code base.

Please let me know if you have any other hacktoberfest related :eyes:  thing, I have one more pr/mr to make :dancers:

PR #5121, opened 2021-10-05
Use 3.10 for Github actions

cdce8p:
## Description

Replace `3.10-dev` with `3.10` in Github actions.

Astroid PR: https://github.com/PyCQA/astroid/pull/1198

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -149,7 +149,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
     outputs:
       python-key: ${{ steps.generate-python-key.outputs.key }}
     steps:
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
I was going to do that for astroid but my clone is broken locally, this is hard to fix. Anyway, I was going to try using ``3.11-dev`` to see if it's possible yet. Maybe you read the documentation and know it's not ?

cdce8p:
> I was going to do that for astroid but my clone is broken locally, this is hard to fix. Anyway, I was going to try using `3.11-dev` to see if it's possible yet. Maybe you read the documentation and know it's not ?

There haven't been any releases for `3.11` yet. We will need to wait for the first alpha for that.
The tags are for releases in this repo here: https://github.com/actions/python-versions/releases
(You can select to watch only the releases if you like. Quite helpful to know when a new Python version is available.)

cdce8p:
@Pierre-Sassoulas Could you update the required status checks. The names will change a bit. You can also merge this PR, it's ready.

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
> (You can select to watch only the releases if you like. Quite helpful to know when a new Python version is available.)

This is great ! Thanks !

PR #5122, opened 2021-10-06
Add guard helper functions from astroid

cdce8p:
## Description

Followup to the discussion here: https://github.com/PyCQA/pylint/pull/5107#discussion_r721744321

Add `is_typing_guard` and `is_sys_guard` helper functions to `pylint.checkers.utils`.

--
Deprecation astroid: https://github.com/PyCQA/astroid/pull/1202

/CC: @Pierre-Sassoulas @DanielNoord

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1551,12 +1551,46 @@ def get_import_name(
     return modname
 
 
+def is_sys_guard(node: nodes.If) -> bool:
+    """Return True if IF stmt is a sys.version_info guard.
+
+    >>> import sys
+    >>> if sys.version_info > (3, 8):
+    >>>     from typing import Literal
+    >>> else:
+    >>>     from typing_extensions import Literal
+    """
+    if isinstance(node.test, nodes.Compare):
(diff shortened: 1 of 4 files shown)


pylint/checkers/utils.py
```diff
@@ -1551,12 +1551,46 @@ def get_import_name(
...
+    return False
+
+
+def is_typing_guard(node: nodes.If) -> bool:
+    """Return True if IF stmt is a typing guard.
+
+    >>> from typing import TYPE_CHECKING
+    >>> if TYPE_CHECKING:
+    >>>     from xyz import a
+    """
```
Pierre-Sassoulas, line 1582: I think keeping the docstring from astroid with the astroid parsing make sense too. (We just have to change the call from `node.is_...()` to `is_...(node)`). (I started to implement this this morning).

cdce8p: Not sure it's necessary. They are only meant as examples anyway.

Pierre-Sassoulas (approved):

PR #5132, opened 2021-10-08
Rename `len-as-condition` to `use-implicit-booleaness-not-len`

jaehoonhwang:
Rename `len-as-condition` to be more general for new checker
`use-implicit-booleaness-not-comparison`

* Refactor `LenChecker` class -> `ImplicitBooleanessChecker`o
* Rename test files/`len_checker.py`/`__init__.py` to reflect new name.
* Add `len-as-condition` as `old_names` for `use-implicit-booleaness-not-len`

## Type of Changes

## Description

Refactoring `len_checker.py` as part of this PR #5120 

Refer to #4774

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,9 +18,9 @@ Release date: TBA
 
   Closes #3859
 
-* Rename ``len-as-condition`` to ``use-implicit-booleaness-not-len``
-
-  Closes #4774
+* In length checker, ``len-as-condition`` has been renamed as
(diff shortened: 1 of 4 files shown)


jaehoonhwang:
@Pierre-Sassoulas 
I have tried using `len-as-condition` for functional testing, but it seems like test is only accepting new name. 
I couldn't find any example of functional tests that are using one of `old_names` to test.
But, I have found out that disable is working for the `old_names` 

Any suggestions?

Issue #3773, opened 2020-08-10
False positive with useless-super-delegation for keyword args with dict defaults


filip-zyzniewski:
Problem similar to #2022 .

### Steps to reproduce

(I know, mutable default is a bad idea)

```python
# uselesssuper.py
class MyBase:
    def method(self, *, arg={}):
        self.x = arg

class MyChild(MyBase):
    def method(self, *, arg={'foo': 'bar'}):
        super().method(arg=arg)

myobj = MyChild()
myobj.method()
print(myobj.x)
```

### Current behavior

```shell
$ python3 uselesssuperdict.py 
{'foo': 'bar'}
$ pylint uselesssuperdict.py --disable=all --enable=useless-super-delegation
************* Module uselesssuperdict
uselesssuperdict.py:7:4: W0235: Useless super delegation in method 'method' (useless-super-delegation)

------------------------------------------------------------------
Your code has been rated at 8.89/10 (previous run: 8.89/10, +0.00)

$ 
```

### Expected behavior

No issue

### pylint --version output

```shell
$ pylint --version
pylint 2.5.3
astroid 2.4.2
Python 3.8.5 (default, Jul 21 2020, 10:42:08) 
[Clang 11.0.0 (clang-1100.0.33.17)]
$
```

hippo91:
@filip-zyzniewski thanks for the report. I made PR #3778 that should correct this.

PR #5157, opened 2021-10-15
Fix useless-super-delegation false positive when default keyword argument is a variable.

nickpesce:
## Type of Changes

## Description
similar to https://github.com/PyCQA/pylint/issues/3773
### Problem
Minimal example of this issue:
```py
X=1
Y=2
class A:
    def __init__(self, u, v=X):
        pass


class B(A):
    def __init__(self, u, v=Y): #  [useless-super-delegation]
        super().__init__(a, b)
```
variables X an Y are incorrectly assumed to be the same

### Fix
Currently, only a select few types of default values are checked and all others are assumed to be the same. `nodes.Name` was not one of those handled cases, so `X` and `Y` above were assumed to be the same. I could think of 2 solutions
1. If a variable type is not explicitly handled for comparison, assume that they are different values instead of the same
2. Explicitly handle `nodes.Name`

I went with option 2 since it seemed like the least intrusive and less likely to have side effects unknown to me, but I could switch it to option 1 if preferred.

Two variable names are considered the same if the inference system outputs the same possible values. The simpler option would be to always call two variables different but that would give more false negatives. I don't know much about the inference system though, so I'm looking for some input on this approach.

### DIFF
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -186,23 +186,6 @@ def _positional_parameters(method):
     return positional
 
 
-def _get_node_type(node, potential_types):
-    """
-    Return the type of the node if it exists in potential_types.
-
-    Args:
-        node (astroid.node): node to get the type of.
-        potential_types (tuple): potential types of the node.
-
-    Returns:
-        type: type of the node or None.
-    """
-    for potential_type in potential_types:
-        if isinstance(node, potential_type):
-            return potential_type
-    return None
-
-
 def _has_different_parameters_default_value(original, overridden):
     """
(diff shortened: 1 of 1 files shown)


nickpesce:
Thanks for the review @Pierre-Sassoulas! Did you see my note about the Fix in the PR description?

- Do you think that using `infer` is the right way to compare two `nodes.Name`'s?
- Would it be better to default to assuming that two default params are different if they are not one of the handled types?

Pierre-Sassoulas:
In this case, it might be more efficient to check if it's a call of self/cls/ClassName, is that what you meant ?


I did not understand the second part of your question, could you provide an example ?

PR #5195, opened 2021-10-21
Fix test decorator for `PyLinter` options

DanielNoord:

## Type of Changes

## Description

I stumbled upon this while working on #5194.

Without this we don't actually set settings that come from `PyLinter` correctly and therefore we can't test them with `get_global_option`.
This should pass tests immediately, but if not I will fix them!

### DIFF
--- a/pylint/testutils/decorator.py
+++ b/pylint/testutils/decorator.py
@@ -24,11 +24,24 @@ def set_config(**kwargs):
             except optparse.OptionError:
                 # Check if option is one of the base options of the PyLinter class
                 for key, value in kwargs.items():
-                    self.checker.set_option(
-                        key.replace("_", "-"),
-                        value,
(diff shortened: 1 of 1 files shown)


DanielNoord:
The decorator now handles three cases "correctly".

1) If the option is from the checker that is loaded by the unittest class, it sets the setting using the normal validator and flow
2) If the option is from `PyLinter`, it sets the setting using the validator by getting the `optdict` from `PyLinter`
3) If the option is from another checker, it sets the setting using a shortcut. It doesn't create an `optdict` but is allowed to set the setting

The initial change I did some weeks ago was to allow both 2 and 3, but it turned I should have separated those cases more clearly.

Investing more time in 3 and allowing double-loading of checkers during unittests does not seem worth the effort. If we are going to rework the config handling because of `optparse`'s deprecation anyway, we might as well leave this hacky trick. It's only for our own tests and it is already much better than the decorator we used a couple weeks ago.

PR #5210, opened 2021-10-25
Add a regression tests for https://github.com/PyCQA/astroid/pull/1207

Pierre-Sassoulas:
## Type of Changes

## Description

This should fail before we upgrade astroid following the merge of https://github.com/PyCQA/astroid/pull/1207

### DIFF
--- a/tests/lint/test_pylinter.py
+++ b/tests/lint/test_pylinter.py
@@ -1,16 +1,31 @@
 from typing import Any, NoReturn
 from unittest.mock import patch
 
-from astroid import AstroidBuildingError
+from astroid import AstroidBuildingError, extract_node, nodes
 from py._path.local import LocalPath  # type: ignore[import]
 from pytest import CaptureFixture
 
+from pylint.checkers.utils import safe_infer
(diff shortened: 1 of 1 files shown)


DanielNoord:
@Pierre-Sassoulas I would be okay with letting this go. If anything this should probably be tested in `astroid`. This seems to be such a niche case that we can add tests cases when somebody reports problems with this (we will then probably also get more input on what to actually test).

Pierre-Sassoulas:
I'm reluctant to close as we merged in astroid with the implicit assumption that we were going to test in pylint as it would be easier to test in pylint. I totally lost the context of the original change at this point though.

jacobtylerwalls:
I opened an alternative regression test that does fail before the astroid PR (and hopefully is meaningful!) in https://github.com/PyCQA/astroid/pull/1397.

> This does not fail like it should.

I think this was expected, here the author said "when you run the test build within such a no-filesystem environment."

DanielNoord:
Superseded by https://github.com/PyCQA/astroid/pull/1397. Thanks a lot @jacobtylerwalls, this has been a very long open issue.

Issue #2018, opened 2018-04-15
Show why snake_case errors is emitted on single word variables (might be too small)


rmullin7286:
### Steps to reproduce
    no = 'No'

### Current behavior
I'm getting an error saying that 'no' doesn't conform the camel_case style, even though it's only one word.

### Expected behavior
There should be no error.

### pylint --version output
pylint 1.8.4,
astroid 1.6.3
Python 3.6.3 (v3.6.3:2c5fed8, Oct  3 2017, 18:11:49) [MSC v.1900 64 bit (AMD64)]

PCManticore:
Hi, thanks for creating this issue! Can you also post your `pylintrc` file? You could also run pylint with `--include-naming-hint=y`, which will offer more indication on why it considers it a bad name. I think it might be because the name it's too short, by default we expect words of length 3+.

eight04:
Is there a way to configure the minimum length of the variable name?

PCManticore:
@eight04 Not quite, the only thing you could do is to change the `variable-rgx` to conform to the format and length you require. It might be a good idea to just separate the length of a variable/name in pylint from the format of a variable/name.

gitaarik:
I tried setting [`variable-rgx`](http://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=variable-rgx), but the `invalid-name` `snake_case` error doesn't go away. The `variable-rgx` setting does work for variables that really don't conform to the regex, but it doesn't take away the first warning, although the docs say that it should override the `variable-naming-style` setting. When I try to unset it with: `variable-naming-style=` then both errors go away. All by all, it doesn't make any difference for me.

filips123:
Why is variable length important at all?
Variable can have good name even if it is short. Example for this can be `no` and some physical variables and constants.
What else can be done excluding changing `variable-rgx`? One name that conform this style is `var_no`, but is even worse then just `no`.

pedro-w:
I agree with @filips123  - I've been using Python to do physics calculations and I use vars like `m`, `x` and `y` to match the equations. It looks strange and wordy to have to rename them `mass`, `x_coordinate` etc.
One option is to use the `good-names=` config item but it's a bit tedious to maintain in practice.

PCManticore:
No, that's not possible.

filips123:
Why? Can you add support for this? It would be useful if only one file contains that variables.

PCManticore:
[...]

PR #5221, opened 2021-10-27
Fix ``invalid-name`` for ``TypeVar`` and add ``typevar-name-missing-variance`` checker

### LINKED ISSUE #3401: pylint should not use Class name regex for TypeVar names
### Steps to reproduce

1. Consider the following example:

```python
"""Example"""
from typing import TypeVar

T = TypeVar('T')


def identity(arg: T) -> T:
    """Identity function."""
    return arg
```

2. pylint typevar.py

### Current behavior

typevar.py:4:0: C0103: Class name "T" doesn't conform to PascalCase naming style (invalid-name)

### Expected behavior

Either:
- Test should pass
- pylint should use a whitelist of names for TypeVars that is separate from that which is used for Class names. 'T' is a perfectly canonical TypeVar annotation in functional notation, but it is a perfectly rubbish runtime Class name.

See https://docs.python.org/3/library/typing.html#typing.TypeVar for a canonical example of usage from the official python docs.

### pylint --version output

```
pylint 2.3.1
astroid 2.2.5
Python 3.7.6 (default, Dec 19 2019, 22:50:01) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]
```

DanielNoord:
**I'm fine with postponing this till 2.13!**

## Type of Changes

## Description

This supersedes #3463 and closes #3401

This change is very large, sorry about that. We can split it up, but since I took #3463 as a baseline it was very hard to separate the changes logically. As said in the opening line, I'm fine with postponing this until `2.13` and focus on releasing `2.12` first. I just wanted to close one of the stale PRs and create a working version of it.

I noticed that many of the regex patterns in the NamingStyles might not be 100% perfect. It took several tries to create a pattern for PascalCase `TypeVar`s that actually worked and I based it on the current standard pattern...
Furthermore, I don't really like the hack I did with `_test_name_is_correct_for_all_name_types_except_typevar`. However, I didn't want to increase the size of this PR even further and I intend to revisit these checkers in a follow-up PR anyway. I feel like some edge cases are currently passing some of the regex patterns.

I will annotate some of the code to (hopefully) make reviewing this a little bit easier.

### DIFF
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -334,6 +334,11 @@ Other Changes
 
   Closes #5360, #3877
 
+* ``invalid-name`` check now also checks ``TypeVar`` for a pattern set with the new option
+  ``--typevar-rgx`` option or the chosen naming style (``--typevar-naming-style``).
+
+  Closes #3401
+
 * Fixed a false positive (affecting unreleased development) for
   ``used-before-assignment`` involving homonyms between filtered comprehensions
   and assignments in except blocks.
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -128,6 +128,8 @@ class NamingStyle:
             "class_attribute": cls.CLASS_ATTRIBUTE_RGX,
             "class_const": cls.CONST_NAME_RGX,
             "inlinevar": cls.COMP_VAR_RGX,
+            # TODO: Create dedicated TypeVar pattern # pylint: disable=fixme
+            "typevar": cls.CLASS_NAME_RGX,
         }[name_type]
 
 
@@ -1672,6 +1674,7 @@ KNOWN_NAME_TYPES = {
     "class_attribute",
     "class_const",
     "inlinevar",
+    "typevar",
 }
 
 DEFAULT_NAMING_STYLES = {
@@ -1686,6 +1689,7 @@ DEFAULT_NAMING_STYLES = {
     "class_attribute": "any",
     "class_const": "UPPER_CASE",
     "inlinevar": "any",
+    "typevar": "PascalCase",
 }
 
 
@@ -2078,14 +2082,6 @@ class NameChecker(_BasicChecker):
     def _check_name(self, node_type, name, node, confidence=interfaces.HIGH):
         """Check for a name using the type's regexp."""
 
-        # pylint: disable=fixme
-        # TODO: move this down in the function and check TypeVar
-        # for name patterns as well.
-        # Check TypeVar names for variance suffixes
-        if node_type == "typevar":
-            self._check_typevar_variance(name, node)
-            return
-
         def _should_exempt_from_invalid_name(node):
             if node_type == "variable":
                 inferred = utils.safe_infer(node)
@@ -2110,6 +2106,10 @@ class NameChecker(_BasicChecker):
 
         if match is None and not _should_exempt_from_invalid_name(node):
             self._raise_name_warning(None, node, node_type, name, confidence)
+
+        # Check TypeVar names for variance suffixes
+        if node_type == "typevar":
+            self._check_typevar_variance(name, node)
 
     def _check_assign_to_new_keyword_violation(self, name, node):
         keyword_first_version = self._name_became_keyword_in_version(
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -113,6 +113,7 @@ class NamingStyle:
     COMP_VAR_RGX: Pattern[str] = ANY
     DEFAULT_NAME_RGX: Pattern[str] = ANY
     CLASS_ATTRIBUTE_RGX: Pattern[str] = ANY
+    TYPE_VAR_RGX: Pattern[str] = ANY
 
     @classmethod
     def get_regex(cls, name_type):
@@ -128,14 +129,16 @@ class NamingStyle:
             "class_attribute": cls.CLASS_ATTRIBUTE_RGX,
             "class_const": cls.CONST_NAME_RGX,
             "inlinevar": cls.COMP_VAR_RGX,
-            # TODO: Create dedicated TypeVar pattern # pylint: disable=fixme
-            "typevar": cls.CLASS_NAME_RGX,
+            "typevar": cls.TYPE_VAR_RGX,
         }[name_type]
 
 
 class SnakeCaseStyle(NamingStyle):
     """Regex rules for snake_case naming style."""
 
+    TYPE_VAR_RGX = re.compile(
+        r"^_{0,2}(?:[^\W\dA-Z][^\WA-Z]+)+[Tt]?(?<!Type)(?:_co(?:ntra)?)?$"
+    )
     CLASS_NAME_RGX = re.compile(r"[^\W\dA-Z][^\WA-Z]+$")
     MOD_NAME_RGX = re.compile(r"[^\W\dA-Z][^\WA-Z]*$")
     CONST_NAME_RGX = re.compile(r"([^\W\dA-Z][^\WA-Z]*|__.*__)$")
@@ -149,6 +152,9 @@ class SnakeCaseStyle(NamingStyle):
 class CamelCaseStyle(NamingStyle):
     """Regex rules for camelCase naming style."""
 
+    TYPE_VAR_RGX = re.compile(
+        r"^_{0,2}(?:[^\W\dA-Z][^\W_]+)(?:_[Tt])?(?<!Type)(?:_co(?:ntra)?)?$"
+    )
     CLASS_NAME_RGX = re.compile(r"[^\W\dA-Z][^\W_]+$")
     MOD_NAME_RGX = re.compile(r"[^\W\dA-Z][^\W_]*$")
     CONST_NAME_RGX = re.compile(r"([^\W\dA-Z][^\W_]*|__.*__)$")
@@ -160,6 +166,9 @@ class CamelCaseStyle(NamingStyle):
 class PascalCaseStyle(NamingStyle):
     """Regex rules for PascalCase naming style."""
 
+    TYPE_VAR_RGX = re.compile(
+        r"^_{0,2}(?:[^\W\da-z_]+|(?:[^\W\da-z_][^\WA-Z_]+)+T?(?<!Type))(?:_co(?:ntra)?)?$"
+    )
     CLASS_NAME_RGX = re.compile(r"[^\W\da-z][^\W_]+$")
     MOD_NAME_RGX = re.compile(r"[^\W\da-z][^\W_]+$")
     CONST_NAME_RGX = re.compile(r"([^\W\da-z][^\W_]*|__.*__)$")
@@ -171,6 +180,9 @@ class PascalCaseStyle(NamingStyle):
 class UpperCaseStyle(NamingStyle):
     """Regex rules for UPPER_CASE naming style."""
 
+    TYPE_VAR_RGX = re.compile(
+        r"^_{0,2}(?:[^\W\da-z][^\Wa-z]+)*T?(?<!Type)(?:_co(?:ntra)?)?$"
+    )
     CLASS_NAME_RGX = re.compile(r"[^\W\da-z][^\Wa-z]+$")
     MOD_NAME_RGX = re.compile(r"[^\W\da-z][^\Wa-z]+$")
     CONST_NAME_RGX = re.compile(r"([^\W\da-z][^\Wa-z]*|__.*__)$")
--- a/pylintrc
+++ b/pylintrc
@@ -353,6 +353,13 @@ method-rgx=[a-z_][a-z0-9_]{2,}$
 # Naming hint for method names
 method-name-hint=[a-z_][a-z0-9_]{2,}$
 
+# Naming style used for type variables. Default is PascalCase.
+typevar-naming-style=PascalCase
+
+# Regular expression which can overwrite the naming style set by typevar-naming-style.
+# If left empty, type variables will be checked with the set naming style.
+typevar-rgx=
+
 # Regular expression which should only match function or class names that do
 # not require a docstring. Use ^(?!__init__$)_ to also check __init__.
 no-docstring-rgx=__.*__
@@ -363,7 +370,6 @@ docstring-min-length=-1
 
 # List of decorators that define properties, such as abc.abstractproperty.
 property-classes=abc.abstractproperty
-
 
 [TYPECHECK]
 
--- a/tests/functional/t/typevar_naming_style_default.py
+++ b/tests/functional/t/typevar_naming_style_default.py
@@ -1,0 +1,72 @@
+"""Test case for typevar-name-incorrect-variance with default settings"""
+# pylint: disable=too-few-public-methods
+
+from typing import TypeVar
+
+# PascalCase names with prefix
+GoodNameT = TypeVar("GoodNameT")
+_T = TypeVar("_T")
+_GoodNameT = TypeVar("_GoodNameT")
+__GoodNameT = TypeVar("__GoodNameT")
+GoodNameWithoutContra = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodNameWithoutContra", contravariant=True
+)
+GoodNameT_co = TypeVar("GoodNameT_co", covariant=True)
+GoodNameT_contra = TypeVar("GoodNameT_contra", contravariant=True)
+GoodBoundNameT = TypeVar("GoodBoundNameT", bound=int)
+GoodBoundNameT_co = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodBoundNameT_co", bound=int
+)
+GoodBoundNameT_contra = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodBoundNameT_contra", bound=int
+)
+GoodBoundNameT_co = TypeVar("GoodBoundNameT_co", bound=int, covariant=True)
+GoodBoundNameT_contra = TypeVar("GoodBoundNameT_contra", bound=int, contravariant=True)
+
+GoodBoundNameT_co = TypeVar("GoodBoundNameT_co", int, str, covariant=True)
+GoodBoundNameT_co = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodBoundNameT_co", int, str, contravariant=True
+)
+GoodBoundNameT_contra = TypeVar("GoodBoundNameT_contra", int, str, contravariant=True)
+GoodBoundNameT_contra = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodBoundNameT_contra", int, str, covariant=True
+)
+
+# Some of these will create a RunTime error but serve as a regression test
+T = TypeVar(  # [typevar-name-incorrect-variance]
+    "T", covariant=True, contravariant=True
+)
+T = TypeVar("T", covariant=False, contravariant=False)
+T_co = TypeVar("T_co", covariant=True, contravariant=True)
+T_contra = TypeVar(  # [typevar-name-incorrect-variance]
+    "T_contra", covariant=True, contravariant=True
+)
+T_co = TypeVar("T_co", covariant=True, contravariant=False)
+T_contra = TypeVar("T_contra", covariant=False, contravariant=True)
+
+# PascalCase names without prefix
+AnyStr = TypeVar("AnyStr")
+DeviceTypeT = TypeVar("DeviceTypeT")
+CALLABLE_T = TypeVar("CALLABLE_T")  # [invalid-name]
+DeviceType = TypeVar("DeviceType")  # [invalid-name]
+GoodNameWithoutContra = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodNameWithoutContra", contravariant=True
+)
+
+# camelCase names with prefix
+badName = TypeVar("badName")  # [invalid-name]
+badNameWithoutContra = TypeVar(  # [invalid-name, typevar-name-incorrect-variance]]
+    "badNameWithoutContra", contravariant=True
+)
+badName_co = TypeVar("badName_co", covariant=True)  # [invalid-name]
+badName_contra = TypeVar("badName_contra", contravariant=True)  # [invalid-name]
+
+# PascalCase names with lower letter prefix in tuple assignment
+(
+    a_BadName,  # [invalid-name]
+    a_BadNameWithoutContra,  # [invalid-name, typevar-name-incorrect-variance]
+) = TypeVar("a_BadName"), TypeVar("a_BadNameWithoutContra", contravariant=True)
+GoodName_co, a_BadName_contra = TypeVar(  # [invalid-name]
+    "GoodName_co", covariant=True
+), TypeVar("a_BadName_contra", contravariant=True)
+GoodName_co, VAR = TypeVar("GoodName_co", covariant=True), "a string"
--- a/tests/functional/t/typevar_naming_style_rgx.py
+++ b/tests/functional/t/typevar_naming_style_rgx.py
@@ -1,0 +1,15 @@
+"""Test case for typevar-name-missing-variance with non-default settings"""
+
+from typing import TypeVar
+
+# Name set by regex pattern
+TypeVarsShouldBeLikeThis = TypeVar("TypeVarsShouldBeLikeThis")
+TypeVarsShouldBeLikeThis_contra = TypeVar(
+    "TypeVarsShouldBeLikeThis_contra", contravariant=True
+)
+TypeVarsShouldBeLikeThis_co = TypeVar("TypeVarsShouldBeLikeThis_co", covariant=True)
+
+# Name using the standard style
+GoodNameT = TypeVar("GoodNameT")  # [invalid-name]
+GoodNameT_co = TypeVar("GoodNameT_co", covariant=True)  # [invalid-name]
+GoodNameT_contra = TypeVar("GoodNameT_contra", contravariant=True)  # [invalid-name]
--- a/ChangeLog
+++ b/ChangeLog
@@ -450,6 +450,11 @@ Release date: TBA
   Insert your changelog randomly, it will reduce merge conflicts
   (Ie. not necessarily at the end)
 
+* ``invalid-name`` check now also checks ``TypeVar`` for a pattern set with the new option
+  ``--typevar-rgx`` option or the chosen naming style (``--typevar-naming-style``).
+
+  Closes #3401
+
 * Added new checker ``typevar-name-missing-variance``. Emitted when a covariant
   or contravariant ``TypeVar`` does not end with  ``_co`` or ``_contra`` respectively or
   when a ``TypeVar`` is not either but has a suffix.
--- a/doc/user_guide/options.rst
+++ b/doc/user_guide/options.rst
@@ -39,6 +39,8 @@ name is found in, and not the type of object assigned.
 +--------------------+---------------------------------------------------------------------------------------------------+
 | ``inlinevar``      | Loop variables in list comprehensions and generator expressions.                                  |
 +--------------------+---------------------------------------------------------------------------------------------------+
+| ``typevar``        | Type variable declared with ``TypeVar``.                                                          |
++--------------------+---------------------------------------------------------------------------------------------------+
 
 Default behavior
 ~~~~~~~~~~~~~~~~
@@ -82,6 +84,8 @@ Following options are exposed:
 
 .. option:: --inlinevar-naming-style=<style>
 
+.. option:: --typevar-naming-style=<style>
+
 
 Custom regular expressions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -117,6 +121,8 @@ expression will lead to an instance of ``invalid-name``.
 .. option:: --class-const-rgx=<regex>
 
 .. option:: --inlinevar-rgx=<regex>
+
+.. option:: --typevar-rgx=<regex>
 
 Multiple naming styles for custom regular expressions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/pylint/constants.py
+++ b/pylint/constants.py
@@ -71,6 +71,7 @@ HUMAN_READABLE_TYPES = {
     "class_attribute": "class attribute",
     "class_const": "class constant",
     "inlinevar": "inline iteration",
+    "typevar": "type variable",
 }
 
 
--- a/pylint/utils/linterstats.py
+++ b/pylint/utils/linterstats.py
@@ -26,6 +26,7 @@ class BadNames(TypedDict):
     method: int
     module: int
     variable: int
+    typevar: int
 
 
 class CodeTypeCount(TypedDict):
@@ -102,6 +103,7 @@ class LinterStats:
             method=0,
             module=0,
             variable=0,
+            typevar=0,
         )
         self.by_module: Dict[str, ModuleStats] = by_module or {}
         self.by_msg: Dict[str, int] = by_msg or {}
@@ -171,6 +173,7 @@ class LinterStats:
             "method",
             "module",
             "variable",
+            "typevar",
         ],
     ) -> int:
         """Get a bad names node count."""
@@ -192,6 +195,7 @@ class LinterStats:
             "method",
             "module",
             "variable",
+            "typevar",
         }:
             raise ValueError("Node type not part of the bad_names stat")
 
@@ -208,6 +212,7 @@ class LinterStats:
                 "method",
                 "module",
                 "variable",
+                "typevar",
             ],
             node_name,
         )
@@ -230,6 +235,7 @@ class LinterStats:
             method=0,
             module=0,
             variable=0,
+            typevar=0,
         )
 
     def get_code_count(
@@ -317,6 +323,7 @@ def merge_stats(stats: List[LinterStats]):
         merged.bad_names["method"] += stat.bad_names["method"]
         merged.bad_names["module"] += stat.bad_names["module"]
         merged.bad_names["variable"] += stat.bad_names["variable"]
+        merged.bad_names["typevar"] += stat.bad_names["typevar"]
 
         for mod_key, mod_value in stat.by_module.items():
             merged.by_module[mod_key] = mod_value
--- a/tests/checkers/unittest_base.py
+++ b/tests/checkers/unittest_base.py
@@ -210,17 +210,23 @@ class TestNamePresets(unittest.TestCase):
         SNAKE_CASE_NAMES | CAMEL_CASE_NAMES | UPPER_CASE_NAMES | PASCAL_CASE_NAMES
     )
 
-    def _test_name_is_correct_for_all_name_types(
+    def _test_name_is_correct_for_all_name_types_except_typevar(
         self, naming_style: Type[base.NamingStyle], name: str
     ) -> None:
         for name_type in base.KNOWN_NAME_TYPES:
-            self._test_is_correct(naming_style, name, name_type)
+            # pylint: disable-next=fixme
+            # TODO: See if there is a better way to tests these styles
+            if name_type != "typevar":
+                self._test_is_correct(naming_style, name, name_type)
 
-    def _test_name_is_incorrect_for_all_name_types(
+    def _test_name_is_incorrect_for_all_name_types_except_type_var(
         self, naming_style: Type[base.NamingStyle], name: str
     ) -> None:
         for name_type in base.KNOWN_NAME_TYPES:
-            self._test_is_incorrect(naming_style, name, name_type)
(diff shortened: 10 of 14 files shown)


cdce8p:
> The first: I guess? I don't know.. Don't really have a preference.
> The second not for `camelCase` I think? I think it should either start with a `t` or a `s` for `Some`.
> The third I would vote against as wel. They should be all `UpperCase` right?

Good points. So this then?
```py
# snakecase
T_sometypevar_co or t_sometypevar_co  # ?

# PascalCase
T_SomeTypeVar_co

# camelCase
T_someTypeVar_co

# UpperCase
T_SOMETYPEVAR_CO
```

Tbh I also don't care too much about about it. The current `PascalCase` regex works great for me.
@Pierre-Sassoulas Do you have an opinion about it?

DanielNoord:
I think we should make these go through without too many "false positives" and then try to really fix them in #3704

cdce8p:
> I think we should make these go through without too many "false positives" and then try to really fix them in #3704

The default `PascalCase` works great. Already tested it with HomeAssistant.
There are some small issues, `Func` or `FuncT` instead of `T_Func`, `_T_Var` in a module scope, I would however consider these true positives.

Pierre-Sassoulas:
Sorry, I'm late for this discussion and also I did not use TypeVar a lot and never encountered the concept of contra or contra variant before, so I'm bringing more question than answer here. Anyway, I don't know if this is relevant, but instinctively I would try to stay consistent with the casing of the base variable ie:
```python
# snakecase
t_sometypevar_co

# PascalCase
TSomeTypeVarCo

# camelCase
tSomeTypeVarCo

# UpperCase
T_SOMETYPEVAR_CO
```
But to be fair I think of TypeVar as type, so shouldn't it be the same style than the class name all the time ?

pylint/checkers/base.py
```diff
@@ -119,12 +120,14 @@ def get_regex(cls, name_type):
...
             "class_const": cls.CONST_NAME_RGX,
             "inlinevar": cls.COMP_VAR_RGX,
+            "typevar": cls.TYPE_VAR_RGX,
         }[name_type]
 
 
 class SnakeCaseStyle(NamingStyle):
     """Regex rules for snake_case naming style."""
 
+    TYPE_VAR_RGX = re.compile(r"[^\W\dA-Z][^\WA-Z]*$")
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Did you mean ``camelCase`` as opposed to ``snake_case`` ?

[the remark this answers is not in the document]
DanielNoord: Both! The problem with these "styles" is that there are no real authorities on how they should handle `TypeVar`s.

pylint/checkers/base.py
```diff
@@ -2098,6 +2162,52 @@ def _name_became_keyword_in_version(name, rules):
...
+    def _assigns_typevar(node: Optional[nodes.NodeNG]) -> bool:
+        """Check if a node is assigning a TypeVar"""
+        if isinstance(node, astroid.Call):
+            inferred = utils.safe_infer(node.func)
+            if (
+                inferred
+                and isinstance(inferred, astroid.ClassDef)
+                and inferred.qname() == TYPING_TYPE_VAR_QNAME
+            ):
+                return True
```
Pierre-Sassoulas, line 2175:
```suggestion
            return (
                inferred
                and isinstance(inferred, astroid.ClassDef)
                and inferred.qname() == TYPING_TYPE_VAR_QNAME
            )
```

pylint/checkers/base.py
```diff
@@ -2098,6 +2162,52 @@ def _name_became_keyword_in_version(name, rules):
                 return ".".join(str(v) for v in version)
         return None
 
+    @staticmethod
+    def _assigns_typevar(node: Optional[nodes.NodeNG]) -> bool:
+        """Check if a node is assigning a TypeVar"""
+        if isinstance(node, astroid.Call):
```
Pierre-Sassoulas, line 2168:
```suggestion
        if not isinstance(node, astroid.Call):
            return False
```
Would permit to remove the return False at the end ?

DanielNoord:
> ```python
> # snakecase
> t_sometypevar_co # t_some_type_var_co ?
> 
> # PascalCase
> TSomeTypeVarCo
> 
> # camelCase
> tSomeTypeVarCo
> 
> # UpperCase
> T_SOMETYPEVAR_CO # T_SOME_TYPE_VAR_CO ?
> ```

I think the snake case example should be allowed.

The `PascalCase` I think shouldn't as we use `PascalCase` to conform `TypeVar`s to the PEP-guidelines, which mandate using `T_`. I agree it is not strictly `PascalCase`, but I think it should be like this.

I think for `camelCase` using the prefix is also recommended.

The `UpperCase` one is allowed.

> But to be fair I think of TypeVar as type, so shouldn't it be the same style than the class name all the time ?

The PEP gives specific guidelines for them, so I guess we want them to be different?

pylint/checkers/base.py
```diff
@@ -2098,6 +2162,52 @@ def _name_became_keyword_in_version(name, rules):
...
+        if self._typevar_naming_style == "UPPER_CASE":
+            _co, _contra = "_CO", "_CONTRA"
+
+        if isinstance(node.parent, nodes.Assign):
+            keywords = node.assign_type().value.keywords
+        elif isinstance(node.parent, nodes.Tuple):
+            keywords = (
+                node.assign_type().value.elts[node.parent.elts.index(node)].keywords
+            )
+
```
Pierre-Sassoulas, line 2192:
```suggestion
            )
        else:
            return None
```
Seems like we don't handle the case where it's not a Assign not a Tuple. But I guess it never happens ?

pylintrc
```diff
@@ -268,6 +268,13 @@ docstring-min-length=-1
 # List of decorators that define properties, such as abc.abstractproperty.
 property-classes=abc.abstractproperty
 
+# Naming style used for type variables. Default is PascalCase.
+typevar-naming-style=PascalCase
```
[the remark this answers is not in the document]
Pierre-Sassoulas: It's true the pylintrc would become huge but at the same time some options are "hidden". There's also an incomplete ``--help`` hidding a ``--long-help`` with all the options when launching pylint. Maybe having all the options could permit to never miss them even if it does make for a huge wall of text.

tests/functional/t/typevar_name_missing_variance.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import TypeVar
+
+# Type variables without variance
+T = TypeVar("T")
+T_co = TypeVar("T_co")  # [typevar-name-missing-variance]
+T_contra = TypeVar("T_contra")  # [typevar-name-missing-variance]
+
+# Type variables not starting with T_
+N = TypeVar("N")
+N_co = TypeVar("N_co", covariant=True)
```
Pierre-Sassoulas, line 12: I really need to read the documentation about that, my review is not helpful right now. I trust Marc's opinion if you want to merge this in 2.12.

pylintrc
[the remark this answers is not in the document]
DanielNoord: I think we should wait with doing this on the move to `argparse`. We don't know how we might handle these "modular" settings.

Pierre-Sassoulas:
> to conform TypeVars to the PEP-guidelines

Sorry did not know about that, disregard what I said, of course we should follow PEP8.

pylint/checkers/base.py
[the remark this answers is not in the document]
Pierre-Sassoulas: Ok, there's some pressure on us to take a good decision then 😄

pylint/checkers/base.py
DanielNoord: For some reason `mypy` fails on this..
Reverting.

pylint/checkers/base.py
DanielNoord: This creates a `inconsistent-return-statements` warning.

Pierre-Sassoulas: This needs the other suggestion that made mypy fails, nevermind.

pylint/checkers/base.py
```diff
@@ -149,6 +153,7 @@ class CamelCaseStyle(NamingStyle):
 class PascalCaseStyle(NamingStyle):
     """Regex rules for PascalCase naming style."""
 
+    TYPE_VAR_RGX = re.compile(r"[^\W\da-z]+?_?([^\W\da-z_][^\WA-Z_]+)*(_co(ntra)?)?$")
```
cdce8p, line 156: Just got a change to read what PEP8 had to say about it.
https://www.python.org/dev/peps/pep-0008/#type-variable-names

As I understand it, the pattern should be similar to normal class names, except for the `_co` and `_contra` suffixes.
So by default any of these should be allowed
```py
from typing import TypeVar

T = TypeVar("T")
AnyStr = TypeVar("AnyStr")
Num = TypeVar("Num")

VT_co = TypeVar('VT_co', covariant=True)
KT_contra = TypeVar('KT_contra', contravariant=True)
```

We should probably also allow a leading underscore
```py
_T = TypeVar("_T")
_Num = TypeVar("_Num")
```

DanielNoord: The first examples are all supported.

Why would we allow a leading underscore? The examples in the PEP and the PEP introducing `TypeVar` don't seem to use them.
Is this to declare them privately? If so, I'll have to check if that is not something taken care of by the checker at "check-time". The other patterns also don't seem to include leading underscores.

cdce8p: `AnyStr` and `Num` won't work. The first part currently needs to be all upper case: `[^\W\da-z]+?`. This is basically equivalent to `[A-Z_]+?`. Thus a leading `_` is already handled.

> Why would we allow a leading underscore? The examples in the PEP and the PEP introducing TypeVar don't seem to use them. Is this to declare them privately?

Yes. We support it for other names, too.

Pierre-Sassoulas (commented):
I added a comment saying I messed up the previous rebase and did not publish it 😓 Would you mind if I rebase this locally on the latest main @DanielNoord ?

doc/whatsnew/2.13.rst
```diff
@@ -91,6 +94,9 @@ Other Changes
 
   Closes #5360, #3877
 
+* ``invalid-name`` check now also checks ``TypeVar`` for a pattern set with the new option
+  ``--typevar-rgx`` option or the chosen naming style (``--typevar-naming-style``).
+
```
DanielNoord, line 99:
```suggestion

  Closes #3401

```

Pierre-Sassoulas (commented):
I know I should review this but I can't find the time to educate myself about typevar and variant right now. In particular I think we need to take a decision regarding the naming style itself; but I don't know any codebase using variant so it's hard to know what the "consensus" is in the python community right now.

pylint/checkers/base.py
```diff
@@ -157,6 +161,7 @@ class CamelCaseStyle(NamingStyle):
 class PascalCaseStyle(NamingStyle):
     """Regex rules for PascalCase naming style."""
 
+    TYPE_VAR_RGX = re.compile(r"_*[^\W\da-z]*_?([^\W\da-z_][^\WA-Z_]+)*(_co(ntra)?)?$")
```
cdce8p, line 164: I tested the changes against Home Assistant and noticed that even the `PascalCase` style doesn't seem to be perfect just yet. For example `CALLABLE_T` would be allowed currently.

As there aren't any formal guidelines yet, I looked at https://github.com/python/typeshed to see what they are using. This is the regex test list I came up with
```python
_TypeVar
_TBE
_ActionT
_MessageT
AnyStr
_CallableT
_T
_KT
_VT
DeviceTypeT

_T_co
_T_contra
_ActionT_co
_ActionT_contra

# These should NOT match!
DICT_T
CALLABLE_T
ENUM_T
DeviceType
_StrType
```

Some observations
* It's common for TypeVar names to be just one or two chars (at most 3). In these cases all are capitalized.
  `T`, `_KT`
* If the name is a word or multiple words, usually (but not always) it has a `T` suffix.
* `_co` and `_contra` mentioned in PEP 484 is used combined with the first two rules.
* In rare cases `Type` was added as suffix, instead of `T`. `4x - Type` vs `47x - T` in `typeshed`. I would lean towards preventing `Type` as suffix. Only something like `DeviceTypeT` should be allowed. That way we can somewhat unify the TypeVar naming.
* Uppercase names with `_T` suffix should not be allowed. Those should be replaced,
  i.e. `CALLABLE_T` -> `CallableT`

This is the regex I came up with
```py
    TYPE_VAR_RGX = re.compile(
        r"^_{0,2}(?:[^\W\da-z_]+|(?:[^\W\da-z_][^\WA-Z_]+)+T?(?<!Type))(?:_co(?:ntra)?)?$"
    )
```

pylint/checkers/base.py
```diff
@@ -146,6 +149,7 @@ class SnakeCaseStyle(NamingStyle):
 class CamelCaseStyle(NamingStyle):
     """Regex rules for camelCase naming style."""
 
+    TYPE_VAR_RGX = re.compile(r"_*[^\W\dA-Z]+_[^\W_]*(_co(ntra)?)?$")
```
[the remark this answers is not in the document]
cdce8p: I wrote this comment before doing the analysis of the use in `typeshed`. Maybe let's decide on the `PascalCase` version first before coming back to it?

cdce8p (commented):
It took some time, but I finally had time to look at it again. I left some comments below. We should probably start with the `PascalCase` style as this is the default.

To end the guess work, I looked at https://github.com/python/typeshed to see which style they are using. It's the official types repo for the stdlib, so probably a good place to take hints from.

Let me know what you think.

pylint/checkers/base.py
```diff
@@ -1870,6 +1889,7 @@ def open(self):
         self._bad_names_rgxs_compiled = [
             re.compile(rgxp) for rgxp in self.config.bad_names_rgxs
         ]
+        self._typevar_naming_style = get_global_option(self, "typevar_naming_style")
```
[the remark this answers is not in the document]
DanielNoord: I did not know about this, thanks!

pylint/checkers/base.py
DanielNoord: Thanks for the additional work you did here @cdce8p! Much appreciated as always.

This sounds like a very well-thought out regex pattern and list of matches. I checked on https://regex101.com and this works perfectly. I'll go ahead and work on a new version of this PR with this pattern.

cdce8p: I used regex101.com myself 😄
We could debate if we truly want to disallow the `Type` suffix. The more I think about it the less certain I am. What to you do think?

DanielNoord: I came up with one more: what about `T_GoodName`? It is part of the test cases I wrote right now, but no longer passes with this regex. It also doesn't feature in your list.

DanielNoord: > I used regex101.com myself 😄
> We could debate if we truly want to disallow the `Type` suffix. The more I think about it the less certain I am. What to you do think?

It feels a bit arbitrary, but if people want to keep using it they can easily overwrite the regex pattern? I think it does make sense.

cdce8p: > I came up with one more: what about `T_GoodName`? It is part of the test cases I wrote right now, but no longer passes with this regex. It also doesn't feature in your list.

Wasn't it that I suggested it originally? I don't remember anymore.
I've not seen that pattern in typeshed though, only with the Uppercase variant in Home Assistant. I lean towards saying it's invalid. `GoodNameT` would be more consistent.

Pierre-Sassoulas: I wanted to educate myself about typvar with variance, but I did not find the time. I'm going to trust you on this one 😄 I think it's a rather large and non opinionated set of possible names and we can always restrict it later when/if a consensus is reached in the python community.

Pierre-Sassoulas (commented):
I wanted to educate myself about typevar with variance, but I did not find the time. I'm going to trust you on this one 😄 I think it's a rather large and non opinionated set of possible names and we can always restrict it later when/if a consensus is reached in the python community.

pylint/checkers/base.py
cdce8p: > > I used regex101.com myself 😄
> > We could debate if we truly want to disallow the `Type` suffix. The more I think about it the less certain I am. What to you do think?
> 
> It feels a bit arbitrary, but if people want to keep using it they can easily overwrite the regex pattern? I think it does make sense.

Not sure how many would actually choose to overwrite it but you're right, it's at least an option. Nevertheless we have an opportunity to set a good standard here that shouldn't be discarded.

There are a few small arguments in favor of keeping it, at least for now.
* The `Type` suffix is much more common with Type Aliases, e.g. `Union[..., ...]`.
* The suffix `T` could be considered established in typeshed.
* And yes, we can always change it later.

-> Let's leave it then (for now).

cdce8p:
> I wanted to educate myself about typevar with variance, but I did not find the time.

At least from my experience, variance isn't actually all that important. I find `TypeVar("T", bound=<some_type>)` much more useful. (And easier to understand 😄 )

> I think it's a rather large and non opinionated set of possible names and we can always restrict it later when/if a consensus is reached in the python community.

Yeah, at least for the default `PascalCase`, we allow some more options when compared to the "normal" naming style. So far the default TypeVar names do all need to be listed in `BASIC.goood-names` or be combined with `pylint: disable=invalid-name`, like eg. `T`.

cdce8p:
An idea for a followup, `ParamSpec` is quite similar to `TypeSpec` with regards to the naming style used. Just now `P` instead of `T`. We could probably reuse most of the code for `TypeVar` for it.
https://docs.python.org/3/library/typing.html#typing.ParamSpec

It only started to be used recently in typeshed, so there aren't many variations yet. Only `_P` so far.

DanielNoord:
> An idea for a followup, `ParamSpec` is quite similar to `TypeSpec` with regards to the naming style used. Just now `P` instead of `T`. We could probably reuse most of the code for `TypeVar` for it. https://docs.python.org/3/library/typing.html#typing.ParamSpec
> 
> It only started to be used recently in typeshed, so there aren't many variations yet. Only `_P` so far.

I'd like to see if there is any merit in finishing up #3704 after this. If there is, adding `ParamSpec` should be fairly straightforward afterwards.

DanielNoord:
@cdce8p I've updated code and tests for `PascalCase` just so we can agree on that first before tackling the other styles.

One issue was:
```python
GoodBoundNameT_co = TypeVar(  # [typevar-name-incorrect-variance]
    "GoodBoundNameT_co", bound=MyClass
)
```

It matches with `_co` so isn't picked up as invalid. That is obviously debatable, but I have added the `typevar-name-incorrect-variance` warning for these cases to let users know they should probably change the name.
I've also renamed it from `typevar-name-missing-variance` to `typevar-name-incorrect-variance` to reflect this change.

tests/functional/t/typevar_name_missing_variance.py
```diff
@@ -0,0 +1,35 @@
...
+CT_co = TypeVar("CT_co", contravariant=True)  # [typevar-name-incorrect-variance]
+CT_contra = TypeVar("CT_contra")  # [typevar-name-incorrect-variance]
+CT_contra = TypeVar("CT_contra", contravariant=True)
+
+# Tests for combinations with covariance
+VT = TypeVar("VT", covariant=True)  # [typevar-name-incorrect-variance]
+VT_contra = TypeVar("VT_contra", covariant=True)  # [typevar-name-incorrect-variance]
+VT_co = TypeVar("VT_co", covariant=True)
+
+# Tests for combinations with tuple assignment
```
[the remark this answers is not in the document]
cdce8p: There should also be a group with constraints (and `bound`) in combination with variance.
```py
T = TypeVar("T", int, str)
T2 = TypeVar("T2", int, str, covariant=True)
...
T3 = TypeVar("T3", bound=int, contravariant=True)
...
```

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -0,0 +1,53 @@
...
+    "GoodBoundNameT_contra", bound=MyClass
+)
+
+# PascalCase names without prefix
+AnyStr = TypeVar("AnyStr")
+BadNameWithoutContra = TypeVar(  # [typevar-name-incorrect-variance]
+    "BadNameWithoutContra", contravariant=True
+)
+BadName_co = TypeVar("BadName_co", covariant=True)
+BadName_contra = TypeVar("BadName_contra", contravariant=True)
```
cdce8p, line 35: Those are good names now.

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -0,0 +1,53 @@
...
+
+# camelCase names with prefix
+T_goodName = TypeVar("T_goodName")  # [invalid-name]
+T_goodNameWithoutContra = TypeVar(  # [invalid-name, typevar-name-incorrect-variance]]
+    "T_goodNameWithoutContra", contravariant=True
+)
+T_goodName_co = TypeVar("T_goodName_co", covariant=True)  # [invalid-name]
+T_goodName_contra = TypeVar("T_goodName_contra", contravariant=True)  # [invalid-name]
```
cdce8p, line 43: These bad ones. It's probably best to remove the `T_` prefix too.

tests/functional/t/typevar_naming_style_rgx.py
```diff
@@ -0,0 +1,15 @@
...
+TypeVarsShouldBeLikeThis = TypeVar("TypeVarsShouldBeLikeThis")
+TypeVarsShouldBeLikeThis_contra = TypeVar(
+    "TypeVarsShouldBeLikeThis_contra", contravariant=True
+)
+TypeVarsShouldBeLikeThis_co = TypeVar("TypeVarsShouldBeLikeThis_co", covariant=True)
+
+# Name using the standard style
+T_GoodName = TypeVar("T_GoodName")  # [invalid-name]
+T_GoodName_co = TypeVar("T_GoodName_co", covariant=True)  # [invalid-name]
+T_GoodName_contra = TypeVar("T_GoodName_contra", contravariant=True)  # [invalid-name]
```
cdce8p, line 15: Better to remove the `T_` prefix

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -0,0 +1,53 @@
...
+
+# PascalCase names with lower letter prefix in tuple assignment
+(
+    a_BadName,  # [invalid-name]
+    a_BadNameWithoutContra,  # [invalid-name, typevar-name-incorrect-variance]
+) = TypeVar("a_BadName"), TypeVar("a_BadNameWithoutContra", contravariant=True)
+GoodName_co, a_BadName_contra = TypeVar(  # [invalid-name]
+    "GoodName_co", covariant=True
+), TypeVar("a_BadName_contra", contravariant=True)
+GoodName_co, VAR = TypeVar("GoodName_co", covariant=True), "a string"
```
cdce8p, line 53: Maybe at the end. Could you add `CALLABLE_T`, `DeviceTypeT`, and `DeviceType`? The last just so we actual test that part of the regex and know that we change the right thing should we do so.

cdce8p (commented):
> One issue was:
> 
> ```python
> GoodBoundNameT_co = TypeVar(  # [typevar-name-incorrect-variance]
>     "GoodBoundNameT_co", bound=MyClass
> )
> ```
> 
> It matches with `_co` so isn't picked up as invalid. That is obviously debatable, but I have added the `typevar-name-incorrect-variance` warning for these cases to let users know they should probably change the name. I've also renamed it from `typevar-name-missing-variance` to `typevar-name-incorrect-variance` to reflect this change.

Sounds good to me! IMO it's a valid TypeVar name, just like you mentioned the `_co` suffix. So `incorrect-variance` describes it pretty good. I realized that my example with `T_co` wasn't ideal. It was meant as an invalid one that wasn't recognized previously.

I've left some comments about a few more corner case we should consider with the variance check. Besides that the regex matching for `PascalCase` looks about right. We could probably start thinking about the other naming styles soon. Let's finish the variance stuff first though.

cdce8p:
> > An idea for a followup, `ParamSpec` is quite similar to `TypeSpec` with regards to the naming style used. Just now `P` instead of `T`. We could probably reuse most of the code for `TypeVar` for it. https://docs.python.org/3/library/typing.html#typing.ParamSpec
> > It only started to be used recently in typeshed, so there aren't many variations yet. Only `_P` so far.
> 
> I'd like to see if there is any merit in finishing up #3704 after this. If there is, adding `ParamSpec` should be fairly straightforward afterwards.

Tbh I didn't look at #3704 in much detail. From what I've seen though, I'm hesitant about it. It's quite a big change in how we handle these regex without a clear benefit to either use or the enduser. It's much more likely that we accidentally break stuff.

As for `ParamSpec`, what I was thinking about was more along the lines of "copy the style from TypeVar and change the `T` suffix to `P`". Add some tests and done. Even one more step further done the road, there are explicit `TypeAliases` that should use the regular `PascalCase` instead of `UpperCase` at module level. But I'm getting ahead of myself here 😅

DanielNoord:
> Tbh I didn't look at #3704 in much detail. 

Same 😄 Just want to know if we should move it ahead or if we should close.

cdce8p:
> > Tbh I didn't look at #3704 in much detail.
> 
> Same 😄 Just want to know if we should move it ahead or if we should close.

You can ask me again after we finish this one 😄

pylint/checkers/base.py
```diff
@@ -2106,6 +2173,58 @@ def _name_became_keyword_in_version(name, rules):
...
+                node.assign_type().value.elts[node.parent.elts.index(node)].keywords
+            )
+        else:
+            return None
+
+        if keywords:
+            for kw in keywords:
+                if kw.arg == "covariant" and kw.value.value and not name.endswith(_co):
+                    suggest_name = "".join((name.rsplit("_", 1)[0], _co))
+                    return (name, "covariant", suggest_name)
```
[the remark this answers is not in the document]
DanielNoord: https://github.com/PyCQA/pylint/pull/5221/commits/d91d8623c2c312deca2d7d152a606e3354daaccf

pylint/checkers/base.py
```diff
@@ -2106,6 +2173,58 @@ def _name_became_keyword_in_version(name, rules):
...
+                if (
+                    kw.arg == "contravariant"
+                    and kw.value.value
+                    and not name.endswith(_contra)
+                ):
+                    suggest_name = "".join((name.rsplit("_", 1)[0], _contra))
+                    return (name, "contravariant", suggest_name)
+                if kw.arg == "bound" and (name.endswith(_co) or name.endswith(_contra)):
+                    suggest_name = name.rsplit("_", 1)[0]
+                    return (name, "invariant", suggest_name)
```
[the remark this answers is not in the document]
DanielNoord: https://github.com/PyCQA/pylint/pull/5221/commits/269d76e608ccb5090577a38b0a0fc1da345112c2

tests/functional/p/protocol_classes.py
```diff
@@ -12,10 +12,10 @@ def digest(self) -> bytes:
         ...
 
 
-Generic = typing.TypeVar("Generic")
+GenericT = typing.TypeVar("GenericT")
```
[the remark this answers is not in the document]
DanielNoord: Weird. This test failed yesterday, but passes now..

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -0,0 +1,53 @@
+"""Test case for typevar-name-incorrect-variance with default settings"""
+# pylint: disable=too-few-public-methods
+
+from typing import TypeVar
+
+
+class MyClass:
+    """Class to be used to bound TypeVar's"""
```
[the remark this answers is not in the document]
DanielNoord: https://github.com/PyCQA/pylint/pull/5221/commits/5bf36edfabbac8f77037f0b45856e0a2e0c8c799

tests/functional/t/typevar_naming_style_default.py
DanielNoord: https://github.com/PyCQA/pylint/pull/5221/commits/0bde758cc754cd5b81bbb526aa4d055cf33bdc97

tests/functional/t/typevar_naming_style_default.py
DanielNoord: https://github.com/PyCQA/pylint/pull/5221/commits/0bde758cc754cd5b81bbb526aa4d055cf33bdc97

tests/functional/t/typevar_naming_style_default.py
DanielNoord: https://github.com/PyCQA/pylint/pull/5221/commits/cebeeac01b1245a65740006a0af650ebc23005a3

tests/functional/t/typevar_name_missing_variance.py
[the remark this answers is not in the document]
DanielNoord: First comment: https://github.com/PyCQA/pylint/pull/5221/commits/98caba0641cf75f9b7f17c9243ed26bf0464e325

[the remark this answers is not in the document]
DanielNoord: Second comment: https://github.com/PyCQA/pylint/pull/5221/commits/9b627f0e9bd87d2ffa4bc24aceaf5579ad6aa638

cdce8p:
@DanielNoord Could you rebase this one? I would like to try getting this done now, finally 😪 It would really help with Home Assistant. I've been adding to many TypeVars there.

cdce8p:
@DanielNoord Please correct me if I'm wrong, it has been a moment. If I remember correctly, we basically got the the PascalCase regex about right but haven't yet reached a good result for the remaining naming styles.

Unless anyone has a smart idea, I'm not sure there are good solutions for `SnakeCase`, `CamelCase`, and `UpperCase`. What I've asked myself is do we even need any? It's true, currently we provide defaults for all four variations, but does that still need to be the case for more specific names? I suggested similar changes for `ParamSpec` variables and `TypeAliases` already. What should the other defaults be there? It's difficult to find a good solution for something that isn't well established  yet. But maybe that isn't even necessary 🤔

What would you think if we only offer the `--typevar-rgx` option with the default we agreed upon for `PascalCase`? Especially going forward with adding more specific naming patterns, that might be a good option. If we want to do this, we should probably think about a few aspects
* At the moment, an empty regex defaults to the naming style preset. If we don't have any presets, I think it could make sense if it would be equal to `.*` (ie. matching all names and thus disabling the error for that specific category).
* The error message could be improved to mention the `typevar-rgx`. This was one of the reasons that lead me to think about this initially. The message that a TypeVar name doesn't match PascalCase, although it looks that way, could be quite confusing. Here from HomeAssistant. `DeviceType` is valid PascalCase, but for TypeVars `DeviceT` would be better (and match the current regex).
```
Type variable name "DeviceType" doesn't conform to PascalCase naming style
```
* As this is an exception to the usual naming styles, we would need to think about how to document that properly.

--
What do you think? I would also be interest in your opinion @Pierre-Sassoulas. Overall this PR has been dragged on for quite some time already. With the changes mentioned above, it would just be a matter of implementing which should be fairly strait forward. Although, as mentioned, we would depart a bit from the current norm.

Pierre-Sassoulas:
Pretty much agree that choosing something when there is no consensus yet is too much pressure 😄 We can add an option to enforce it without choosing ourselves.

> If we don't have any presets, I think it could make sense if it would be equal to ``.*``

There's a similar issue with module level "constant". It would make sense that it's snake cased or uppercased, but it's still nice to check it's internally consistent. So maybe in those case we could have something more specific than "whatever" for the default preset ie. "whatever... but still pythonnic and internally consistent.". 
- snake_case, PascalCase, camelCase, UPPERCASE_WITH_UNDERSCORE, (T_PascalCase, T_PascalCase_contra ?) = ✔️ 
- BAD_snake_CASE, wHaTEver, BAd_UPPeRCASE, Definitely_Not_Camel_Case = 👎 

I realize now that I wrote it that ``T_PascalCase`` ``T_PascalCase_contra`` do not feel very internally consistent.

cdce8p:
> I realize now that I wrote it that `T_PascalCase` `T_PascalCase_contra` do not feel very internally consistent.

That might have been because of an early suggesting from myself. We did revise that pattern in the meantime, after I checked for the common usage in `typeshed`. Instead of the `T_` prefix a `T` suffix should be used. Ie. `PascalCaseT` and `PascalCaseT_contra` respectively. For more details: https://github.com/PyCQA/pylint/pull/5221#discussion_r776501605

> So maybe in those case we could have something more specific than "whatever" for the default preset ie. "whatever... but still pythonnic and internally consistent.".

So basically if it's a module level constant default to the constant naming style? Although it might be a bit more challenging to implement, it would make sense in case the "specific" regex is `None`.

DanielNoord:
Just to say: very busy this week and this PR is hard to "get back into". I will take a look at this next week!

DanielNoord:
I thought about this a little and don't know what the best way forward with this. (I'm probably also a bit burnt out on this PR..).

One of the problems I foresee with special casing `TypeVar` is that it might become more difficult to understand what is going on. Every special case also increases "maintenance cost" which I would like to avoid.
These messages are currently auto-generated so that would also need to be changed for this, I think?

DanielNoord:
I have added new patterns. They are combinations of the standard patterns used for each style with the one we came up for for `PascalStyle`. It's a compromise between only adding `PascalStyle` and making very intricate styles for the other three. I would say let's launch with these (or improved) patterns and await any feedback before we spend many more hours on this... I don't think that many users will use this anyway and if we get reports it's fairly easy to update the pattern to fix any potential issues.

cdce8p:
> I have added new patterns. They are combinations of the standard patterns used for each style with the one we came up for for `PascalStyle`. It's a compromise between only adding `PascalStyle` and making very intricate styles for the other three.

Thanks for taking the time to look at this again!

> I would say let's launch with these (or improved) patterns and await any feedback before we spend many more hours on this... I don't think that many users will use this anyway and if we get reports it's fairly easy to update the pattern to fix any potential issues.

If it were a general case (e.g. constants, class names) I would agree. However, what I noticed while testing it on Home Assistant is that I think the error message can be problematic. Say you name your type var `DeviceType`, the message would say that it doesn't `"conform to PascalCase naming style"`. As a users who doesn't know about it, I would be confused. It would be much more helpful to say that it doesn't `"conform to the TypeVar regex"`. At least in that case I would know to lookup what acceptable names are. _Just to clarify, using the regex we defined is the correct choice IMO. That should not change. It just isn't `PascalCase` anymore._

You're correct though that we've already spend a lot of time on it. I'm just not sure there is another way to continue using the naming styles for that. It also won't get better for `TypeAliases` and `ParamSpec`.

As a way forward if we choose to go that route, I would suggest starting at the beginning and only implementing a framework to accept regex for specific parts (TypeVar, ParamSpec). That should leave out the variance check. Once that PR is done, this one can be adapted to add the additional check. I would understand if you don't like to do that. In that case I would add it to my Todo list, although I don't know when I'll get to it.

Pierre-Sassoulas:
> Say you name your type var DeviceType, the message would say that it doesn't "conform to PascalCase naming style".

This would be very strange indeed, but wouldn't it be ``"conform to TypeVar naming style"`` instead ?

>  As a users who doesn't know about it, I would be confused. It would be much more helpful to say that it doesn't "conform to the TypeVar regex".

I don't think we should change the way this work (not in this MR at least).

But I agree that although exposing the fact that we internally use a regex is not a good idea in general imo, (because it exposes implementation detail that we should not bother the user with), it would also clear confusion in some case. I'm thinking in particular of the case when the variable name is too short but is valid x naming style anyway ([a very very popular issue](https://github.com/PyCQA/pylint/issues/2018)). But maybe the solution is to provide a better regex or internal mechanism to check those. Maybe a checker for length independent from the checker for ``snake_case``.

So imo, exposing the regex internals of pylint is a short term solution that will alienate our users not well versed in regex and make a later breaking change more problematic (if user start to rely heavily on the regex to control something else than the naming style, like length, they already do of course but they had to read the documentation for it). And I think we'll need that breaking change later on because of #2018

DanielNoord:
> If it were a general case (e.g. constants, class names) I would agree. However, what I noticed while testing it on Home Assistant is that I think the error message can be problematic. Say you name your type var `DeviceType`, the message would say that it doesn't `"conform to PascalCase naming style"`. As a users who doesn't know about it, I would be confused. It would be much more helpful to say that it doesn't `"conform to the TypeVar regex"`. At least in that case I would know to lookup what acceptable names are. _Just to clarify, using the regex we defined is the correct choice IMO. That should not change. It just isn't `PascalCase` anymore._

Would a solution be to say `"conform to the predefined TypeVar PascalCase naming style pattern"`? We could then add the regex pattern to the description of the message so that users who are looking for this predefined pattern can easily find it.

Pierre-Sassoulas:
> Would a solution be to say "conform to the predefined TypeVar PascalCase naming style pattern"?

Sounds good, but we need to remove ``predefined `` as it could be a user defined pattern and we can't easily change the message when it is.

DanielNoord:
Closing this in favour of #5894. Thanks to anybody that participated in this discussion and did reviews of this. We're nearly there 😄

PR #5229, opened 2021-10-29
Unify all `<y or n>` metavars

tusharsadhwani:
## Type of Changes

## Description

Okay, this was bugging me for a while that the metavars for yes/no options were inconsistent across pylint. This simply re-labels all of them.

I looked across the codebase and found that the most common convention was to have spaces in your metavars, instead of dashes or underscores, so I went witht that.


Pierre-Sassoulas (approved):
Thank you for the consistency fix and congratulation on becoming a pylint contributor :)

PR #3300, opened 2019-12-10
Fix "Access to private function in inner class emits protected-access" issue #3066

ikraduya:
## Steps

- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [ ] Add a ChangeLog entry describing what your PR does.
- [ ] If it's a new feature or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Description
Fix #3066 issue about "Access to a private function in inner class emits protected-access". I add an util function for getting outer class given inner class.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |
|    | :sparkles: New feature |
|    | :hammer: Refactoring  |
|    | :scroll: Docs |

## Related Issue

Closes #3066

PR #5232, opened 2021-10-30
Fix ``protected-access`` for attributes and methods of nested classes 

### LINKED ISSUE #3066: Access to a private function in inner class emits protected-access
This issue might be related to #1159 and maybe others.

### Steps to reproduce

The following code emits a W0212:

```python3
# pylint: disable=C0111, R0903


class Aclass:
    a = 42

    @staticmethod
    def foobar(i):
        Aclass._bar(i)

    @staticmethod
    def _bar(i):
        print(i)

    class Bclass:
        @staticmethod
        def foobar(i):
            Aclass.Bclass._bar(i)

        @staticmethod
        def _bar(i):
            print(i)


Aclass.foobar(Aclass.a + 3)
Aclass.Bclass.foobar(Aclass.a)
``` 

### Current behavior
```
************* Module pylint_test
pylint_test.py:19:12: W0212: Access to a protected member _bar of a client class (protected-access)
```

### Expected behavior

A solid 10/10 score.

### pylint --version output

pylint 2.3.1
astroid 2.2.5
Python 3.6.8 (default, Jan 14 2019, 11:02:34)
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]

DanielNoord:

## Type of Changes

## Description

This supersedes #3300
This closes #3066

### DIFF
--- a/tests/functional/a/access/access_to_protected_members.py
+++ b/tests/functional/a/access/access_to_protected_members.py
@@ -215,7 +215,7 @@ class Issue1159Subclass(Issue1159):
 
 class Issue3066:
     """Test for GitHub issue 3066
-    Accessing of atrributes/methods of inner and outer classes
+    Accessing of attributes/methods of inner and outer classes
     https://github.com/PyCQA/pylint/issues/3066"""
 
     attr = 0
--- a/tests/functional/a/access/access_to_protected_members.py
+++ b/tests/functional/a/access/access_to_protected_members.py
@@ -211,3 +211,67 @@ class Issue1159Subclass(Issue1159):
         instance = Issue1159OtherClass()
         instance._bar = 3  # [protected-access]
         _ = instance._foo  # [protected-access]
+
+
+class Issue3066:
+    """Test for GitHub issue 3066
+    Accessing of atrributes/methods of inner and outer classes
+    https://github.com/PyCQA/pylint/issues/3066"""
+
+    attr = 0
+    _attr = 1
+
+    @staticmethod
+    def _bar(i):
+        """Docstring."""
+
+    @staticmethod
+    def foobar(i):
+        """Test access from outer class"""
+        Issue3066._attr = 2
+        Issue3066.Aclass._attr = "y"  # [protected-access]
+        Issue3066.Aclass.Bclass._attr = "b"  # [protected-access]
+
+        Issue3066._bar(i)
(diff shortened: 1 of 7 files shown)


tests/functional/a/access/access_to_protected_members.py
```diff
@@ -211,3 +211,67 @@ def access_other_attr(cls):
         instance = Issue1159OtherClass()
         instance._bar = 3  # [protected-access]
         _ = instance._foo  # [protected-access]
+
+
+class Issue3066:
+    """Test for GitHub issue 3066
+    Accessing of atrributes/methods of inner and outer classes
```
Pierre-Sassoulas, line 218:
```suggestion
    Accessing of attributes/methods of inner and outer classes
```

Pierre-Sassoulas (approved):
Maybe we could cover the new code with tests at 100% too?

DanielNoord:
@Pierre-Sassoulas The non-covered code came from the initial PR. Turns out it was unnecessary. Removing still passes all test and should improve the coverage.

It is going through tests now. Can I merge afterwards?

Pierre-Sassoulas:
I'm not confident about this uncovered code being useless, it's probably here for a reason maybe proper test cases were not added at the time. I would revert the last change of try to understand the background.

DanielNoord:
Yeah I tested this but it is impossible to trigger this.

Basically, if there `get_outer_class` returns `None` we are at the end of a nested class structure and there is no longer a parent to return a frame of. I tried it with creating new classes outside of the nested structure that referenced the nested structure and functions that enclosed the nested classes, but `get_outer_class` will stop at the last parent of a nested class structure.
`parents_callee` is based on the `as_string()` function of the `expr` of the current node. I believe this also stops at the last parent of the nested class structure.

I believe it is impossible to have `get_outer_class` return `None` while the `callee` is not the latest of `parent_callee`. At least I couldn't figure out how to do so after +- 15 minutes and the tests seem to cover every possible case (imo).

Pierre-Sassoulas (approved):
Ok, it's a nice cleanup then, thank you ! 👏

DanielNoord:
Do you want another maintainer (Marc?) to have another look? Or is this good to go?

Pierre-Sassoulas:
Marc would certainely have valid remarks but I think this is good enough to merge :)

PR #5233, opened 2021-10-30
Nographviz fix

sgheppy:
the Install graphviz error is showed even for plantuml message, but for this format no graphiz is needed

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,8 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.12.rst'
 
+* Fix ``install graphiz`` message which isn't needed for puml output format.
+
 * Fix ``simplify-boolean-expression`` when condition can be inferred as False.
 
   Closes #5200
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -572,3 +572,5 @@ contributors:
 * Tushar Sadhwani (tusharsadhwani): contributor
 
 * Ikraduya Edian: contributor
+
+* Antonio Quarta (sgheppy): contributor
--- a/doc/whatsnew/2.12.rst
+++ b/doc/whatsnew/2.12.rst
@@ -64,6 +64,8 @@ Extensions
 Other Changes
 =============
 
+* Fix ``install graphiz`` message which isn't needed for puml output format.
+
 * Fix ``simplify-boolean-expression`` when condition can be inferred as False.
 
   Closes #5200
--- a/pylint/pyreverse/main.py
+++ b/pylint/pyreverse/main.py
@@ -203,7 +203,7 @@ class Run(ConfigurationMixIn):
         super().__init__(usage=__doc__)
         insert_default_options()
         args = self.load_command_line_configuration(args)
-        if self.config.output_format not in ("dot", "vcg"):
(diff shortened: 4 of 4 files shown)


Pierre-Sassoulas (commented):
Plantuml is also based on graphviz though, plantuml.jar won't work without it at least for older version. Did it change recentely ?

sgheppy:
With pyreverse when png out format is selected, that image was generated directly from dot file with graphiz so that tool is needed. If puml or plantuml out format is selected a text file .puml is generate, not an imahe one. So no graphviz tool is needed until this file get converted to image. This pool request is preparatory for another one that add mermaidjs as out format.

ChangeLog
```diff
@@ -11,6 +11,11 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.12.rst'
 
+* Fix ``install graphiz`` message which isn't needed for puml output format.
+
+  Closes #
+
+
```
DudeNr33, line 18: If no ticket exists the ``Closes #`` is not necessary. Also only a single empty line between two ChangeLog entries is the "standard".
As a heads up: it's better if you apply those changes locally on your machine, as the GitHub web UI is not reliable when it comes to adding/removing blank lines. So trying to apply my suggestion directly here on GitHub will probably still be faulty at the end. 

```suggestion
* Fix ``install graphiz`` message which isn't needed for puml output format.

```

DudeNr33 (changes requested):
Thank you for fixing this!
Indeed graphviz is not needed for the PlantUML output, as ``pyreverse`` currently only produces the textual output without further converting it to an image.
This simply slipped through my testing as I have graphviz installed locally. 

Two minor remarks:
* The empty "Closes #" in the ChangeLog can be removed
* I think this should be added to the ``doc/whatsnew/2.12.rst`` as well, as this was probably an annoying bug for users producing PlantUML output without having graphviz installed.

Looking forward to the new output format you mentioned! 🚀

DudeNr33:
@Pierre-Sassoulas if we want to add the feature that also PlantUML output can be converted directly into images through ``pyreverse``, we need to consider the following:

* Currently ``dot`` is implicitly used as an intermediate format when an image format is specified. The user should be able to choose this intermediate format through the CLI.
* PlantUML also has the option to let the rendering be done by a remote server, so installing it locally is not necessary - should we offer both options?
* PlantUML has no direct command line interface (AFAIK), so the user would need to provide the command to use for converting PlantUML files (``java -jar /path/to/plantuml.jar``)

Pierre-Sassoulas:
No no, I was wrong, just generating the plantuml text is more in line with what we want to handle/maintain in pylint, let's not try to do too much here.

sgheppy:
I've updated the pr with your suggestion.
Thank you,
Antonio

DudeNr33 (approved):
Thank you, will merge this as soon as Github Actions have passed!

Pierre-Sassoulas:
Thank you for handling the review and merging @DudeNr33, next time could you also squash if the commits are coherent together and rebase if they're big or make more sense separated, please ? :) This is to have a mot compact official history (we can check the MR on github if we want the details).

DudeNr33:
Will do, thanks for the feedback!

PR #5173, opened 2021-10-17
Primer tests "à la mypy"

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |

## Description

This is a draft for adding primer tests, and discuss about it.

- [x] Stop github actions at the first job not passing (Already the case see : https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f)
- [x] Upgrade the release doc as we're now running acceptance in github actions
- [x] Choose 2/3 good repositories to add
- [x] Improve the message when primer tests fails (This can still be made better but is enough for now)
- [ ] ~~Add more repositories ?~~ #5359 
- [ ] ~~Reduce total pipeline time (possibly by adding more jobs one for each part of the primer)~~  #5359 

Closes #4413

Issue #4413, opened 2021-04-27
Primer acceptance tests like in black or mypy


Pierre-Sassoulas:
Created this ticket following a discord and #4390 discussion with @cdce8p

Pierre-Sassoulas:
> > The difference is that both are focused on maintaining stability whereas we constantly add new things as well.

> Ha, very true. But we can check that pylint is "not crashing". And I think we could check for the presence of error. There's a very slim possibility that there is a real error in a release of those very big libraries but I would bet on a pylint false positive or regression every time. And in particular if suddenly there's a ton of import-error or no-module-error because namespace packages are broken we would detect it.

https://github.com/PyCQA/astroid/pull/1211#issuecomment-943637484

cdce8p:
This issue might be something we should prioritize. Both #5162 and #5170 could have been prevented if detected earlier.

Pierre-Sassoulas:
I was going to say exactly that. Let's do it.

cdce8p:
> I'm already finding crash which is encouraging (#5172). However I think we'll need to install the environment of the project for this to work pylint need to be able to import dependencies. Also we can't easily add project because `pylint .` do not work without an `__init__.py` at the root level.

I haven't looked at it closely yet, but from what I think, you're right: We'll need to install all import dependencies for each project. That's different to what `black_primer` does. Maybe it's an idea to look how [mypy_primer](https://github.com/hauntsaninja/mypy_primer) does it instead? They should also need to install dependencies.

> Maybe we could install with pip and lint the dependencies installed in the tests ? What pip install is a package and is supposed to work, so we would not have this issue.

Not sure about that. `git clone --depth 1` is probably better.

> I'm going to create a draft so we can discuss.

Thanks for doing this 🚀 I'm looking forward to it!

Issue #2648, opened 2018-12-15
Invalid "Module import itself" error for namespace packages.


feluxe:
### Steps to reproduce

You can either clone this repo to get the test file tree:

`git clone https://github.com/feluxe/test_pylint_ns_package_err.git`

`cd` into it and run `pipenv install --dev --pre`

or create the file tree manually with the following steps:

#### Create test file tree manually:

Create this file tree:

```
tmp_pylint_err
├── nspackage
│   ├── mypackage
│   │   └── __init__.py
│   └── sty
│       └── __init__.py
└── test.py

```

Next, install `pylint` and [`sty`](https://github.com/feluxe/sty) from pypi in a Python 3.7 virtualenv:

```
pipenv install --dev --pre pylint sty
```

We add this line to `nspackage/mypackage/__init__.py`:

```python
from nspackage.sty import fg
```

We add this line to `nspackage/sty/__init__.py`:

```python
from sty import *
```

We add these lines to `test.py`:

```python
from nspackage.sty import fg
from nspackage.mypackage import fg as fg2

print(f"{fg.green}hello1{fg.rs}")
print(f"{fg2.green}hello2{fg2.rs}")
```

### Current behavior

Note: As you can see in the project file tree, `nspackage` is a *name space package* omitting `__init__.py`.

If I run `pipenv run pylint nspackage`, I get these errors:

```
************* Module sty
nspackage/sty/__init__.py:1:0: W0406: Module import itself (import-self)
************* Module mypackage
nspackage/mypackage/__init__.py:1:0: E0611: No name 'fg' in module 'nspackage.sty' (no-name-in-module)
```

### Expected behavior

I think the example project is valid Python and pylint should not complain.

You can run `test.py` for validation:

```
pipenv run python test.py
```

which runs just fine.

### pylint --version output
```
pylint 2.2.2
astroid 2.1.0
Python 3.7.1 (default, Nov 11 2018, 02:35:35) 
[GCC 8.2.1 20180831]
```

PCManticore:
Thanks for reporting an issue!

Oberon00:
At opentelemetry/opentelemetry-python, we have the same issue and it's very annoying as we would like to add companion modules for widely-used libraries that are named the same as the corresponding library, but in a different package.

Oberon00:
It's very likely that #2862 is the same problem but that issue has a nicer repro.

Oberon00:
[...]

PR #5235, opened 2021-10-30
Add an option to disable sys.path patching.

### LINKED ISSUE #5226: Support an Option to Disable sys.path patching
### Current problem

I want an argument/config option to disable sys.path patching. Pylint currently modifies sys.path based on `__init__.py` in folders of the linted files. sys.path patching is messy to get right and current logic for it here has a bug for implicit namespace packages that has been open for [years](https://github.com/PyCQA/pylint/issues/3944).  As a side effect a simple package structure of,

```
foo/
--> A.py

bar/
--> foo/
------> A.py
```

where bar is a namespace package breaks if you try to lint both foo.A and bar.foo.A together.

While making implicit namespace packages work with sys.path patching is probably possible, sys.path patching means files are linted in a way different than actual execution of those imports in interpreter. You can have a folder pass because of patching, but running script would fail from that location if uninstalled.

A related issue is importing files in two separate packages sharing a name like in this issue is incompatible with current sys.path [patching](https://github.com/PyCQA/pylint/issues/2966). Any solution for that issue will involve either not patching sys.path or patching more selectively. 

### Desired solution

Support a config argument/command line argument to disable patching. The current patch logic is only done in two places [here](https://github.com/PyCQA/pylint/blob/main/pylint/lint/pylinter.py#L987) and [here](https://github.com/PyCQA/pylint/blob/main/pylint/lint/pylinter.py#L987). An option to disable it entirely would be a small patch to pylint. I'd be happy to make the patch if this change would be supported. 

### Additional context

_No response_

hmc-cs-mdrissi:
## Type of Changes

## Description

Adds an option for disabling sys.path patching. This option is default to False and should have no impact on existing users.

This is helpful for implicit namespace packages where the current patching logic is incorrect and making patch logic correct in general case is difficult. Disabling sys.path patching means that pylint will work if running python on same files will work. Practically this means any files users want to check should either be checked from the package root directory or be installed first.

Closes #5226

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -391,6 +391,7 @@ contributors:
 - Michael Hudson-Doyle <michael.hudson@canonical.com>
 - Michael Giuffrida <mgiuffrida@users.noreply.github.com>
 - Melvin Hazeleger <31448155+melvio@users.noreply.github.com> (melvio)
+- Mehdi Drissi: Added an option to disable sys path patching
 - Matěj Grabovský <mgrabovs@redhat.com>
 - Matthijs Blom <19817960+MatthijsBlom@users.noreply.github.com>
 - Matej Marušák <marusak.matej@gmail.com>
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -536,6 +536,21 @@ class PyLinter(
                     ),
                 },
             ),
+            (
+                "disable-path-patching",
+                {
+                    "type": "yn",
+                    "metavar": "<y or n>",
+                    "default": False,
+                    "help": (
+                        "Disable the patching of sys.path when searching for "
+                        "modules. This is useful for implicit namespace packages where "
+                        "inferring base path of package is ambiguous. The downside is without "
+                        "patching, the files being linted must be importable from the location "
+                        "the script is run in interpreter."
+                    ),
+                },
+            ),
         )
 
     base_option_groups = (
@@ -1058,13 +1073,13 @@ class PyLinter(
                 )
 
             filepath = files_or_modules[0]
-            with fix_import_path(files_or_modules):
+            with fix_import_path(files_or_modules, self.config.disable_path_patching):
                 self._check_files(
                     functools.partial(self.get_ast, data=_read_stdin()),
                     [self._get_file_descr_from_stdin(filepath)],
                 )
         elif self.config.jobs == 1:
-            with fix_import_path(files_or_modules):
+            with fix_import_path(files_or_modules, self.config.disable_path_patching):
                 self._check_files(
                     self.get_ast, self._iterate_file_descrs(files_or_modules)
                 )
--- a/pylint/pyreverse/main.py
+++ b/pylint/pyreverse/main.py
@@ -191,6 +191,21 @@ OPTIONS = (
             help="set the output directory path.",
         ),
     ),
+    (
+        "disable-path-patching",
+        dict(
+            type="yn",
+            metavar="<y or n>",
+            default=False,
+            help=(
+                "Disable the patching of sys.path when searching for "
+                "modules. This is useful for implicit namespace packages where "
+                "inferring base path of package is ambiguous. The downside is without "
+                "patching, the files being linted must be importable from the location "
+                "the script is run in interpreter."
+            ),
+        ),
+    ),
 )
 
 
@@ -217,7 +232,7 @@ class Run(ConfigurationMixIn):
         if not args:
             print(self.help())
             return 1
-        with fix_import_path(args):
+        with fix_import_path(args, self.config.disable_path_patching):
             project = project_from_files(
                 args,
                 project_name=self.config.project,
--- a/pylint/testutils/_fake.py
+++ b/pylint/testutils/_fake.py
@@ -1,0 +1,12 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""This file is intended to only be used for testing namespace package importing.
+
+   Importing a file has a global side effect of adding a module to sys.modules.
+   As namespace packages are sensitive to sys.path, to test them we check
+   an intentional import error and a successful import. To avoid the imports
+   messing up the test environment, we need to make the import error use
+   a file that is never imported by any other test. This serves as that file.
+"""
--- a/ChangeLog
+++ b/ChangeLog
@@ -785,6 +785,12 @@ Release date: 2021-11-24
   longer relies on counting if statements anymore and uses known if statements locations instead.
   It should not crash on badly parsed if statements anymore.
 
+* Add an option ``disable-path-patching`` to better support namespace packages.
+  This option is disabled by default as it has a trade off with being able
+  to lint uninstalled packages.
+
+  Closes #5266
+
 * Fix ``simplify-boolean-expression`` when condition can be inferred as False.
 
   Closes #5200
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -280,12 +280,24 @@ behavior. Likewise, since negative values are still technically supported,
 ``evaluation`` can be set to a version of the above expression that does not
 enforce a floor of zero.
 
-6.2 I think I found a bug in Pylint. What should I do?
+6.2 I have import errors working with a namespace package. How can I fix it?
+------------------------------------------------------------------------------
+
+If you have an implicit namespace package (i.e. a package without ``__init__.py``)
+then pylint may not be able to determine base of the package. Normally when you run
+pylint it adjusts PYTHONPATH to include the parent directory of the package you are
+working in. For namespace packages you can disable path inference with
+``--disable-path-patching`` setting. For this to work you must have the namespace package
+installed in your environment or run pylint from the directory containing the package.
+
+
+
+6.3 I think I found a bug in Pylint. What should I do?
 -------------------------------------------------------
 
 Read :ref:`Bug reports, feedback`
 
-6.3 I have a question about Pylint that isn't answered here.
+6.4 I have a question about Pylint that isn't answered here.
 ------------------------------------------------------------
 
 Read :ref:`Mailing lists`
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -42,6 +42,12 @@ Extensions
 Other Changes
 =============
 
+* Add an option ``disable-path-patching`` to better support namespace packages.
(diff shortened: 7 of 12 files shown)


cdce8p (commented):
I haven't tested any namespace packages, but the change itself looks good. One small comment.

pylint/lint/pylinter.py
```diff
@@ -507,6 +507,19 @@ def make_options():
                     ),
                 },
             ),
+            (
+                "disable-path-patching",
+                {
+                    "action": "store_true",
```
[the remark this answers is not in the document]
hmc-cs-mdrissi: Done for both defaults.

cdce8p (approved):
LGTM! Thanks @hmc-cs-mdrissi 🚀

Pierre-Sassoulas (changes requested):
Thank you for adding this option !I think this need to be documented so we have something to refer to when someone has trouble with ``sys.path`` patching (or even only for discoverability, I'm not sure user often use the``--help``).

hmc-cs-mdrissi:
Which place would you recommend me updating the documentation? The [faq](https://pylint.pycqa.org/en/latest/faq.html) the [run page](https://pylint.pycqa.org/en/latest/user_guide/run.html) with other global options or somewhere else? If I do the faq I could add a question under troubleshooting for namespace packages.

Pierre-Sassoulas:
In the faq is a really good idea :) Also if you have ideas about how to test that automatically with a minimum namespace package so we don't regress later on that would be great, but I'm sure this is quite complicated to do.

hmc-cs-mdrissi:
I wasn't sure if I could write a unit test, but I think I got a reasonable one to work using functional test. It does rely on pylint being installed when running test for disable path patching one to work or running from the repository root directory. So locally testing it with an editable pylint install it worked. I'm unsure if CI installs pylint before testing or not.

If that doesn't work on CI then I think CI will need a tweak.

I'll add documentation tomorrow.

Pierre-Sassoulas:
I'm on mobile, so I can't link the issue but one of the last things we need to do for the 2.12 milestone is (convenientely) adding primer tests. Please feel free to suggest repositories that would be great to add on the relevant issue (you can click the milestone from here).

pylint/pyreverse/main.py
```diff
@@ -191,6 +191,21 @@
             help="set the output directory path.",
         ),
     ),
+    (
+        "disable-path-patching",
+        dict(
```
cdce8p, line 196: Please use `{}` instead of `dict`.

cdce8p:
The issue regarding the primer tests Pierre mentioned
* #4413

pylint/pyreverse/main.py
hmc-cs-mdrissi: I used dict because all of the options in pyreverse also used dict. I can {} if it's fine to mismatch the rest of the file.

hmc-cs-mdrissi:
Test failure is interesting. The two functional tests I wrote pass individually, but fail when run together. Awkwardly once an import happens it impacts global state (sys.modules) and influences next test import resolution. Messy aspect is this can negatively impact an entirely unrelated test. My test that relied on import failure for namespace_package since it used checkers broke generated_members test because it also imported checkers.

After a couple tries I got it working with all tests and left a long comment in _fake.py about it.

pylint/testutils/_fake.py
```diff
@@ -0,0 +1,7 @@
+"""This file is intended to only be used for testing namespace package importing.
```
Pierre-Sassoulas, line 1: I don't think we should add file without purpose in pylint itself. pylint is not a namespace package so maybe we can create a fixture or it's not even required to have one ? Could we use a namespace that does not exists ? Or is this supposed to be in `` tests/functional/n/namespace_package/pylint/testutils/`` ?

Pierre-Sassoulas (changes requested):

pylint/testutils/_fake.py
hmc-cs-mdrissi: I thought testutils was for files used in tests/? Are you fine with test file importing another test file? I don't think that's currently done by any of the tests and a future version of pytest will not allow that by default either.

https://docs.pytest.org/en/6.2.x/pythonpath.html#import-modes
`For this reason this doesn’t require test module names to be unique at all, but also makes test modules non-importable by each other`

So moving it to tests/functional/n/namespace_package will force you to disable an upcoming pytest change. 

There's a second reason why it's better to avoid having the file in tests/. sys path disable mode has a requirement either files you import are installed or you run it from the right directory. tests folder is not included in pytest install. As a side effect if I move the file to tests this new test will only work as expected if run from the repository root directory. If a user tries to run pytest from a different directory for this test it won't work.

Because of those 2 reasons having the _fake file with tests/ will impose restrictions on pytest usage and I'd prefer to avoid. I can move it there, but the tests will become sensitive to where you run them and unsure that's a good trade off.

hmc-cs-mdrissi: Other option is drop this test entirely and make sys path logic is tested in future primer test.

The main issue for making this not directory sensitive is I need module I'm importing to be installed package in repo. The only choice for that is under pylint/ somewhere. I can make a folder _internal/ and actual location doesn't matter as long as it's in pylint.

hmc-cs-mdrissi:
@Pierre-Sassoulas Thoughts on what I should do with the test? Main options I see are drop the test or keep a file somewhere in pylint/. I don't see a good way to test importing works with disable path on and having the files all in tests/ directory.

pylint/testutils/_fake.py
Pierre-Sassoulas: Sorry I forgot to reply. We have some test with import in ``test_func.py`` which is the legacy functional test file. Maybe you could take add a test to it exceptionally ? I'm guessing we'll have to think about a system for those kind of tesst soon if pytest is going to make this impossible. But it will always be possible to call pylint on a directory, as an integration tests, right ? Sorry if this is not in this PR scope. I'm hesitant to drop the test as long as #5173 is not merged. Do you have a list of stable project that we could use to test this ?

DanielNoord: Would putting it in `tests/regrtest_data` work? This is where I have been putting all data for tests that require additional files.
We should probably update the name of that directory though...

hmc-cs-mdrissi: No it would have same issue as anywhere outside pylint. The root issue is this line, https://github.com/PyCQA/pylint/blob/main/setup.cfg#L57

When pylint is installed as a package, it installs pylint directory. It does not install tests directory. Namespace package solution here works under one of 2 conditions, the imported modules are installed or you need to use right directory to run them from. The second condition means that for development tests will break depending on cd and will complicate test scripts. The first condition means I can't test this well if pylint separates test folder from the install.

Some codebases have a test structure of keeping tests in same folder as src and tests are included with a package install. https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-as-part-of-application-code pytest discusses different test folder structures here. The structure of mixing tests with application code would work. It'd need a pr to refactor the test structure. You could in theory mix both test structures but it'd be fairly confusing if some tests lived in pylint/ and other tests lived in tests/. Having every user be forced to install all tests as part of pip install is a downside to this option.

For stable project to test with I'm unsure of repository that uses implicit namespace packages and would fail without this. Most repositories don't use namespace packages. For the ones that do I expect if they use pylint they have workarounds. My internal repo that needs this uses a workaround of never calling pylint on the whole repo and also has multiple setup.py in it.

I can make a toy repository for this that primer could use.

Pierre-Sassoulas: Would this work for primer ? https://github.com/madhavhugar/example-python-namespace-package

hmc-cs-mdrissi: No. Reading the project,

```
When using multiple namespace packages within the same repository:

Because mynamespace doesn’t contain an init.py, setuptools.find_packages() won’t find the sub-package. You must use setuptools.find_namespace_packages() instead or explicitly list all packages in your setup.py. For example:

```

is true in readme, but the repository does not follow that itself.

https://github.com/madhavhugar/example-python-namespace-package/blob/master/setup.py#L11

I don't see any namespace module in that repository. Instead it looks like a repository that has 3 packages all in same repository and controlled by same setup.py, but no namespace that shares them.

I can make a simple example for the primer tomorrow.

Pierre-Sassoulas: Thank you much appreciated !

Pierre-Sassoulas:
> Thoughts on what I should do with the test? Main options I see are drop the test or keep a file somewhere in pylint/. I don't see a good way to test importing works with disable path on and having the files all in tests/ directory.

Sorry for the delay. Imo we can drop the tests if we introduce a namespace repository with #5173

Pierre-Sassoulas:
@hmc-cs-mdrissi did you have the time to find an example of a namespace package ? We're trying to release 2.12 this week, and there's only this issue and the primer tests left :)

Pierre-Sassoulas (changes requested):
LGTM, we still need a namespace package for primer, a very small example should do if we don't find a serious one.

DanielNoord:
What about https://github.com/pypa/sample-namespace-packages?

Pierre-Sassoulas:
> What about https://github.com/pypa/sample-namespace-packages?

Looks good 👍 (We just need to check that it was failing previously)

@hmc-cs-mdrissi do you still want to work on this or should we take over ? I think this MR is very valuable and need to go in 2.13 :)

DanielNoord:
@hmc-cs-mdrissi Sorry for pinging you again. As @Pierre-Sassoulas said previously this is likely a much appreciated function for many users. Do you think you have the time to write the small package for the `primer` you mentioned?

CONTRIBUTORS.txt
```diff
@@ -536,12 +537,3 @@ under this name, or we did not manage to find their commits in the history.
 - Mirko Friedenhagen
 - Nicholas Smith
 - Nuzula H. Yudaka (Nuzhuka)
```
Pierre-Sassoulas, line 539:
```suggestion
- Nuzula H. Yudaka (Nuzhuka)
- Pek Chhan
- Peter Hammond
- Pierre Rouleau
- Richard Goodman: simplifiable-if-expression (with Tomer Chachamu)
- Sebastian Ulrich
- Takashi Hirashima
- Thomas Snowden: fix missing-docstring for inner functions
- Wolfgang Grafen
- Yannick Brehon
```

Pierre-Sassoulas:
Confirmed to fix namespace issue in https://github.com/PyCQA/pylint/issues/2648#issuecomment-1088047226

DanielNoord:
@Pierre-Sassoulas I think I have found a solution for namespaces without this option. Using some of the import logic we have already made at `astroid` we can determine whether something is a namespace package is some cases. I tested #2648 and I could get `astroid` to tell us that we're dealing with a namespace.

Let's wait (again, I know...) with merging this...

Pierre-Sassoulas:
Closing in favor of #6405

PR #5241, opened 2021-11-01
Refactor of ``visit_name`` for ``variables``

DanielNoord:

## Type of Changes


DanielNoord:
The non-covered lines are already non-covered, see the current `main` coverage report here:
https://coveralls.io/builds/43915003/source?filename=pylint%2Fcheckers%2Fvariables.py

pylint/checkers/variables.py
```diff
@@ -976,267 +976,263 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
 
-            # if the name node is used as a function default argument's value or as
-            # a decorator, then start from the parent frame of the function instead
-            # of the function frame - and thus open an inner class scope
-            if (
-                current_consumer.scope_type == "function"
-                and self._defined_in_function_definition(node, current_consumer.node)
-            ):
-                # ignore function scope if is an annotation/default/decorator, as not in the body
+            if not action:
```
Pierre-Sassoulas, line 1019: Maybe we could create an enum for that Action?

Pierre-Sassoulas (commented):
Huge work, and this make this function so much more understandable, thanks 👍

pylint/checkers/variables.py
DanielNoord: I did, but never used `Enum` before. Probably not using it in its intended way 😅 

Please comment if I did not use it like you wanted me to!

pylint/checkers/variables.py
```diff
@@ -976,267 +989,258 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
-                current_consumer.scope_type == "function"
-                and self._defined_in_function_definition(node, current_consumer.node)
-            ):
-                # ignore function scope if is an annotation/default/decorator, as not in the body
+            if not action.value:
                 continue
+            if action.value == 2:
+                current_consumer.mark_as_consumed(node.name, found_nodes)
+            if action.value:
+                return
```
Pierre-Sassoulas, line 1037:
```suggestion
            if action == VariableVisitConsumerAction.CONTINUE:
                continue
            if action == VariableVisitConsumerAction.CONSUME:
                current_consumer.mark_as_consumed(node.name, found_nodes)
            if action == VariableVisitConsumerAction.RETURN:
                return
```
This makes errors less likely, if the code was wrong it would look very wrong here. Also you do not have to know about the underlying value at all.

DanielNoord: The last row doesn't work as `CONSUME` should also return. Should we do `action in {VariableVisitConsumerAction.RETURN, VariableVisitConsumerAction.CONSUME}` or keep `if action.value `?

Pierre-Sassoulas: I would do the former, it's more explicit as evidenced by the fact I just misunderstood what 'action.value' was doing :smile: Also it does not rely on the value of the other enum being falsey.

cdce8p:
I still have this one on my todo list. Might take some time though as I'm still working on my astroid PR.

DanielNoord:
No rush! I think the `column` PR's are far more important now!

pylint/checkers/variables.py
```diff
@@ -976,267 +989,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+            if is_first_level_ref == 2:
+                self.add_message("used-before-assignment", node=node, args=node.name)
+            if is_first_level_ref:
+                return (VariableVisitConsumerAction.RETURN, None)
+
+        elif isinstance(defnode, nodes.NamedExpr):
+            if isinstance(defnode.parent, nodes.IfExp):
+                if self._is_never_evaluated(defnode, defnode.parent):
                     self.add_message("undefined-variable", args=node.name, node=node)
+                    return (VariableVisitConsumerAction.CONSUME, found_nodes)
```
Pierre-Sassoulas, line 1244:
```suggestion
        elif (
            isinstance(defnode, nodes.NamedExpr)
            and isinstance(defnode.parent, nodes.IfExp)
            and self._is_never_evaluated(defnode, defnode.parent)
        ):
            self.add_message("undefined-variable", args=node.name, node=node)
            return (VariableVisitConsumerAction.CONSUME, found_nodes)
```

pylint/checkers/variables.py
```diff
@@ -976,267 +989,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+                    # Handle postponed evaluation of annotations
+                    if not (
+                        self._postponed_evaluation_enabled
+                        and isinstance(
+                            stmt,
+                            (
+                                nodes.AnnAssign,
+                                nodes.FunctionDef,
+                                nodes.Arguments,
+                            ),
```
Pierre-Sassoulas, line 1180: We could create a variable for this for better formatting / understanding thanks to the name of the new variable.

DanielNoord: Any suggestions for a name?

pylint/checkers/variables.py
```diff
@@ -976,267 +989,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+                if not (
+                    self._postponed_evaluation_enabled
+                    and isinstance(stmt, (nodes.AnnAssign, nodes.FunctionDef))
+                ):
+                    self.add_message(
+                        "used-before-assignment", args=node.name, node=node
+                    )
+                    return (VariableVisitConsumerAction.CONSUME, found_nodes)
+
+            if base_scope_type == "lambda":
```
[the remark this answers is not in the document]
DanielNoord: Not sure. Just copied the previous code. I'll de-indent.

pylint/checkers/variables.py
```diff
@@ -976,267 +989,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+                # E0601 can occur in class-level scope in lambdas, as in
+                # the following example:
+                #   class A:
+                #      x = lambda attr: f + attr
+                #      f = 42
+                if isinstance(frame, nodes.ClassDef) and node.name in frame.locals:
+                    if isinstance(node.parent, nodes.Arguments):
+                        if stmt.fromlineno <= defstmt.fromlineno:
```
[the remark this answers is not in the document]
DanielNoord: We can't there is an elision L1220.

pylint/checkers/variables.py
DanielNoord: I did this so we can easily change and add checks for when there is a `NamedExpr`. I prefer to keep these top if's as `elif isinstance(defnode, nodes.XXX):` so we can easily add things later on.

pylint/checkers/variables.py
Pierre-Sassoulas: I'm wondering too. variable definition nodes ?

DanielNoord: Not sure if it increase readability btw. Because then the type variables would be at the top of the file and you would need to scroll to see it.
Perhaps we should leave it as is? Not sure here...

Pierre-Sassoulas (changes requested):
I just realized there's "Submodule flask added at 9486b6" that was added by mistake before primer were introduced.

pylint/checkers/variables.py
Pierre-Sassoulas: Fair enough :)

pylint/checkers/variables.py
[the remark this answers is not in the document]
Pierre-Sassoulas: 👌

pylint/checkers/variables.py
```diff
@@ -1050,9 +1050,9 @@ def _undefined_and_used_before_checker(
                 and isinstance(frame, nodes.FunctionDef)
                 and frame.is_method()
             )
+            and not utils.node_ignores_exception(node, NameError)
         ):
-            if not utils.node_ignores_exception(node, NameError):
-                self.add_message("undefined-variable", args=node.name, node=node)
+            self.add_message("undefined-variable", args=node.name, node=node)
```
DanielNoord, line 1058: @Pierre-Sassoulas For some reason this breaks the tests... I have not had my coffee yet, but I don't understand why this is not the same.

Pierre-Sassoulas (changes requested):
Could we try to cover the 4 lines that are not covered ? It's line that emit messages, so I think it's important to know if they emit properly or remove them if we can't trigger their messages.

pylint/checkers/variables.py
Pierre-Sassoulas: What commit is that, it's hard to know ? c1c7275 ?

DanielNoord: No this one. Moving the `if` on L1054 into the previous `if` by using `and` produces different results.

Pierre-Sassoulas: Could you link the commit in the conversation ? Github display the discussion in the issue directly and I can't find the commit your commenting on 😄

DanielNoord: https://github.com/PyCQA/pylint/pull/5241/commits/06e0712a5e409ebb3eb90d4cbf5a5d49aeb5387b

pylint/checkers/variables.py
```diff
@@ -979,267 +992,258 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+            node.name in nodes.Module.scope_attrs
+            or utils.is_builtin(node.name)
+            or node.name in self.config.additional_builtins
+            or (
+                node.name == "__class__"
+                and isinstance(frame, nodes.FunctionDef)
+                and frame.is_method()
+            )
+            and not utils.node_ignores_exception(node, NameError)
+        ):
```
Pierre-Sassoulas, line 1057: I think I know why : it's due to the ``and`` being before the last parenthesis when it should be after it (adding a parenthesis at the beginning of the if helps).

```suggestion
        if (
            self._undefined_variable_is_enabled and not (
                node.name in nodes.Module.scope_attrs
                or utils.is_builtin(node.name)
                or node.name in self.config.additional_builtins
                or (
                    node.name == "__class__"
                    and isinstance(frame, nodes.FunctionDef)
                    and frame.is_method()
                )
            and not utils.node_ignores_exception(node, NameError)
        ):
```

DanielNoord:
> Could we try to cover the 4 lines that are not covered ? It's line that emit messages, so I think it's important to know if they emit properly or remove them if we can't trigger their messages.

The message after `elif current_consumer.scope_type == "lambda":` was introduced for this issue:
https://github.com/PyCQA/pylint/issues/404 and this commit https://github.com/PyCQA/pylint/commit/5193b7d7fd63b5ddc4077cb5f37c3cd55a9d6561

I'm not quite sure what it is actually trying to find. 

The other one is https://github.com/PyCQA/pylint/commit/05b526d58a9b210a54ea2755eaf5104f74039e5b.

I'm really not sure type of variables are being tested here.

DanielNoord:
We're in luck. We made a spelling error in the message that is emitted after `elif current_consumer.scope_type == "lambda":`.
The `primer` fails because `home assistant` triggers this message. We should be able to find an example in their code base.

jacobtylerwalls:
Great work, I find this readable. No concerns. 👍🏻

pylint/checkers/variables.py
```diff
@@ -979,267 +992,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
-            ):
+            if action is VariableVisitConsumerAction.CONTINUE:
                 continue
+            if action is VariableVisitConsumerAction.CONSUME:
+                current_consumer.mark_as_consumed(node.name, found_nodes)
+            if action in {
+                VariableVisitConsumerAction.RETURN,
+                VariableVisitConsumerAction.CONSUME,
+            }:
+                return
```
jacobtylerwalls, line 1043: I just rebased (locally) #5402 on top of this, and this is the only significant change in the implementation, I just have to plug in here to consume all the uncertain nodes:

```python
            if action in {
                VariableVisitConsumerAction.RETURN,
                VariableVisitConsumerAction.CONSUME,
            }:
                try:
                    uncertain_nodes = current_consumer.consumed_uncertain[node.name]
                except KeyError:
                    uncertain_nodes = []
                current_consumer.mark_as_consumed(node.name, uncertain_nodes)
                return
```

jacobtylerwalls: Ah, there were failing test cases. And this didn't look right. This looks better for my rebase and passes the tests but requires a change in the PR:

```python
            if action is VariableVisitConsumerAction.CONSUME:
                try:
                    uncertain_nodes = current_consumer.consumed_uncertain[node.name]
                except KeyError:
                    uncertain_nodes = []
                current_consumer.mark_as_consumed(node.name, found_nodes + uncertain_nodes)
```

jacobtylerwalls: Ah, no, the change can wait for the #5402. I can just make the change in the previous comment _and_ the following change in #5402:

```python
        if not found_nodes:
            self.add_message("used-before-assignment", args=node.name, node=node)
            if node.name in current_consumer.consumed_uncertain:
                uncertain_nodes = current_consumer.consumed_uncertain[node.name]
                return (VariableVisitConsumerAction.CONSUME, found_nodes + uncertain_nodes)
            return (VariableVisitConsumerAction.RETURN, found_nodes)
```

DanielNoord:
The code that is triggering the warning in `home assistant` can be found here:
https://github.com/home-assistant/core/blob/296b73874054eec31d91f0db1b9a625ad5d47c32/homeassistant/components/sht31/sensor.py#L63
L63.

However, they are actually disabling the message there.. Is this a false positive perhaps?

@cdce8p Do you know if the `undefined-variable` on L63 in the linked file is a false positive? We're trying to write a test case for the line that emits the warning in `pylint` as it is currently uncovered, but perhaps that message shouldn't be emitted anyway.

pylint/checkers/variables.py
```diff
@@ -979,267 +992,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+                        self.add_message(
+                            "used-before-assignment",
+                            args=node.name,
+                            node=node,
+                        )
+                else:
                     self.add_message("undefined-variable", args=node.name, node=node)
+                    return (VariableVisitConsumerAction.CONSUME, found_nodes)
+            elif current_consumer.scope_type == "lambda":
+                self.add_message("undefined-varibale", args=node.name, node=node)
```
cdce8p, line 1229:
```suggestion
                self.add_message("undefined-variable", args=node.name, node=node)
```
Noticed that this crashes when testing Home-Assistant. As our functional tests didn't crash, I would assume this isn't covered so far?

cdce8p (changes requested):

pylint/checkers/variables.py
DanielNoord: See https://github.com/PyCQA/pylint/pull/5241#issuecomment-980041659. I was actually wondering if this should be a message. You seem to be disabling it in Home-Assistant.

cdce8p: I haven't looked at the logic yet, as I only tried to run it on Home Assistant and it crashed before then. Let's fix it first.

jacobtylerwalls:
> Is this a false positive perhaps?

Yes, that looks like valid code.

```python
>>> f = lambda x: y if (y := x) else None
>>> f(0)
>>> f(1)
1
```

jacobtylerwalls:
Actually we have the bug captured in #5360

pylint/checkers/variables.py
```diff
@@ -979,267 +992,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
             # name node from an astroid built from live code, skip
             assert not stmt.root().file.endswith(".py")
             return
 
+        self._undefined_and_used_before_checker(node, stmt)
+        self._loopvar_name(node)
+
+    @utils.check_messages(
+        "undefined-variable", "used-before-assignment", "cell-var-from-loop"
+    )
```
cdce8p, line 1007:
```suggestion
```

The decorator should be around for `def visit_name`, not here.

pylint/checkers/variables.py
```diff
@@ -979,267 +992,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
     def visit_delname(self, node: nodes.DelName) -> None:
         self.visit_name(node)
 
-    # pylint: disable=too-many-branches
     def visit_name(self, node: nodes.Name) -> None:
```
cdce8p, line 995:
```suggestion
    @utils.check_messages(
        "undefined-variable",
        "used-before-assignment",
        "cell-var-from-loop",
    )
    def visit_name(self, node: nodes.Name) -> None:
```

pylint/checkers/variables.py
```diff
@@ -979,267 +992,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
         start_index = len(self._to_consume) - 1
 
-        undefined_variable_is_enabled = self.linter.is_message_enabled(
+        self._undefined_variable_is_enabled = self.linter.is_message_enabled(
             "undefined-variable"
         )
-        used_before_assignment_is_enabled = self.linter.is_message_enabled(
+        self._used_before_assignment_is_enabled = self.linter.is_message_enabled(
             "used-before-assignment"
         )
```
cdce8p, line 1019: This should probably be moved to `def open()`.

pylint/checkers/variables.py
```diff
@@ -979,267 +992,261 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
-                continue
-
-            # Ignore inner class scope for keywords in class definition
-            if (
-                current_consumer.scope_type == "class"
-                and isinstance(node.parent, nodes.Keyword)
-                and isinstance(node.parent.parent, nodes.ClassDef)
-            ):
+            # Certain nodes shouldn't be checked as they get checked another time
+            if self._node_should_be_skipped(current_consumer, node, i == start_index):
```
cdce8p, line 1028: I'm usually not a big fan of comparisons as arguments. I does make somewhat sense here and creating an extra variable for it doesn't make much sense either.

pylint/checkers/variables.py
```diff
@@ -2329,6 +2339,45 @@ def _check_classdef_metaclasses(self, klass, parent_node):
 
         return consumed
 
+    def _node_should_be_skipped(
+        self, consumer: NamesConsumer, node: nodes.Name, is_start_index: bool
```
cdce8p, line 2343:
```suggestion
    def _node_should_be_skipped(
        self, node: nodes.Name, consumer: NamesConsumer, is_start_index: bool
```
I would suggest to put `node` first. That is usually how it's done in the other helper methods.

pylint/checkers/variables.py
cdce8p:
```suggestion
            if self._node_should_be_skipped(node, current_consumer, i == start_index):
```

pylint/checkers/variables.py
```diff
@@ -2329,6 +2339,45 @@ def _check_classdef_metaclasses(self, klass, parent_node):
...
+    ) -> bool:
```
cdce8p, line 2344: Why did you move that method to the bottom? Wouldn't it make sense between `visit_name` and `_check_consumer`?

cdce8p (changes requested):
I didn't get through all of it yet. What I've seen so far though looks promising!

I left a few comments. Personally I would like to suggest to break this up into individual PRs so we can start merging what is done and keep the diff manageable. It will probably be more work rebasing, but I think that would be worth it. What do you think @DanielNoord?

I would start just with the part I've reviewed so far.
* `def open()`
* `_undefined_and_used_before_checker`
* `_node_should_be_skipped`

DanielNoord:
> I left a few comments. Personally I would like to suggest to break this up into individual PRs so we can start merging what is done and keep the diff manageable. It will probably be more work rebasing, but I think that would be worth it. What do you think @DanielNoord?

I'm fine with breaking up, although I must say I left most of the code untouched. I mostly intended this to move around some of the current code and make it more readable. This naturally meant creating some additional methods/functions, but I have kept the actual code and logic mostly untouched. Would you be okay with me accepting your current review, merging this as a base PR and then do some follow up PR's in which we actually improve the code? There are some open issues for this checker and the PR's @jacobtylerwalls has made recently should help with that as well.

pylint/checkers/variables.py
DanielNoord: Shouldn't it be for both? I seem to remember the decorator being used twice for the same message in inner methods.

pylint/checkers/variables.py
DanielNoord: Done!

pylint/checkers/variables.py
DanielNoord: Not sure, moved it to above `_check_consumer`

pylint/checkers/variables.py
Pierre-Sassoulas: It's black magic until we read what the function is doing exactly 😄 
```
def check_messages(*messages: str) -> Callable:
    """decorator to store messages that are handled by a checker method"""

    def store_messages(func):
        func.checks_msgs = messages
        return func

    return store_messages
```    
I was thinking it was a function for optimization purpose (to not launch check if the message is not activated). To be fair I don't know what is the purpose now.

pylint/checkers/variables.py
```diff
@@ -992,6 +992,11 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
     def visit_delname(self, node: nodes.DelName) -> None:
         self.visit_name(node)
 
+    @utils.check_messages(
+        "undefined-variable",
+        "used-before-assignment",
+        "cell-var-from-loop",
+    )
```
DanielNoord, line 1008: @cdce8p @Pierre-Sassoulas This causes a regression as the `unused_import_everything_disabled` functional test no longer passes. I'm not sure what's going on here, but this is crucial apparently.

cdce8p:
> > I left a few comments. Personally I would like to suggest to break this up into individual PRs so we can start merging what is done and keep the diff manageable. It will probably be more work rebasing, but I think that would be worth it. What do you think @DanielNoord?
> 
> I'm fine with breaking up, although I must say I left most of the code untouched. I mostly intended this to move around some of the current code and make it more readable. This naturally meant creating some additional methods/functions, but I have kept the actual code and logic mostly untouched. Would you be okay with me accepting your current review, merging this as a base PR and then do some follow up PR's in which we actually improve the code? There are some open issues for this checker and the PR's @jacobtylerwalls has made recently should help with that as well.

I think I might have some more ideas for `_check_consumer` which is why I suggested to break it up. We can do it all here, but I suspect that it might get difficult to follow.

Any improvement should be in a PR after this one is done anyway.

pylint/checkers/variables.py
cdce8p: It is used here to determine if a `visit_xxx` method needs to be called. Thus the decorator only needs to be placed before `visit` methods. https://github.com/PyCQA/pylint/blob/e33bc6377bbbf6135888cdadecdfc3baa8497a24/pylint/utils/ast_walker.py#L19-L22

DanielNoord:
> I think I might have some more ideas for `_check_consumer` which is why I suggested to break it up. We can do it all here, but I suspect that it might get difficult to follow.
> 
> Any improvement should be in a PR after this one is done anyway.

Shall we merge this then and continue with improving this in other PRs? That would also allow others to contribute to some smaller areas of these large methods.

cdce8p:
> > I think I might have some more ideas for `_check_consumer` which is why I suggested to break it up. We can do it all here, but I suspect that it might get difficult to follow.
> > Any improvement should be in a PR after this one is done anyway.
> 
> Shall we merge this then and continue with improving this in other PRs? That would also allow others to contribute to some smaller areas of these large methods.

No. Only the part I review so far should be easy to merge (after a final quick review). That's why I'm suggesting it in the first place.

Pierre-Sassoulas (approved):
Thank you, it's a lot clearer this way !

cdce8p:
I still think it would be easier to split this up, but if you prefer it, I can review the rest here too. Just let me know.
I hope to do it later today or tomorrow.

DanielNoord:
> I still think it would be easier to split this up, but if you prefer it, I can review the rest here too. Just let me know.
> I hope to do it later today or tomorrow.

Let's do it here as others have also reviewed this too. I'll try and incorporate your review as soon as I get it so @Pierre-Sassoulas can release `2.12.2` 😄

pylint/checkers/variables.py
```diff
@@ -979,267 +1001,296 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
-                or utils.is_builtin(node.name)
-                or node.name in self.config.additional_builtins
-                or (
-                    node.name == "__class__"
-                    and isinstance(frame, nodes.FunctionDef)
-                    and frame.is_method()
-                )
-            ):
-                if not utils.node_ignores_exception(node, NameError):
+            if base_scope_type != "lambda":
```
cdce8p, line 1239:
```suggestion
            elif base_scope_type != "lambda":
```
This needs to be an `elif` as the `return` above only captures the inner most conditional.

pylint/checkers/variables.py
```diff
@@ -979,267 +1001,296 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+                        #      x = 42
+                        #      y = lambda attr=x: attr
+                        self.add_message(
+                            "used-before-assignment",
+                            args=node.name,
+                            node=node,
+                        )
+                else:
+                    self.add_message("undefined-variable", args=node.name, node=node)
+                    return (VariableVisitConsumerAction.CONSUME, found_nodes)
```
cdce8p, line 1271: This should be indented by one extra level and the condition added.
```py
elif base_scope_type == "lambda":
```

pylint/checkers/variables.py
```diff
@@ -979,267 +1001,296 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
     def visit_name(self, node: nodes.Name) -> None:
-        """Check that a name is defined in the current scope"""
         stmt = node.statement()
         if stmt.fromlineno is None:
             # name node from an astroid built from live code, skip
             assert not stmt.root().file.endswith(".py")
             return
 
+        self._undefined_and_used_before_checker(node, stmt)
+        self._loopvar_name(node)
```
cdce8p, line 1018: The first thing `self._loopvar_name` does is check if the message type is enabled. Could you move the check here? That way we won't even have to call the function if it's disabled. (The check itself could also be moved to `def open()`.)

https://github.com/PyCQA/pylint/blob/6caab13816e8098063b6e077ab0044b94aa87557/pylint/checkers/variables.py#L1710-L1711

cdce8p (changes requested):

pylint/checkers/variables.py
DanielNoord: I think we normally don't put these kinds of checks in `open`? For maintainability I think it makes sense to do the if-check at the same place at which the actual check is started. (Your suggestion improves on this)

pylint/checkers/variables.py
DanielNoord: Do you know if this if statement makes sense any way? (Also look at comment below this).
Using `if`, `elif` and even removing the check altogether all pass the tests. I'm (obviously) fine with the `elif` and indenting the other comment, but it seems we're missing something in our tests here. Did you, by any chance, find out what during your review?

cdce8p: I wasn't able to come up with an example on the fly. Might be that we get lucky here.
IMO it doesn't hurt for check conditions to be as precise as possible, so I would recommend to leave it as it was. That seemed to work.

Furthermore, using `elif` just makes sure we don't perform unnecessary work and thus could save some time.

pylint/checkers/variables.py
```diff
@@ -1917,6 +1965,7 @@ def _check_unused_arguments(self, name, node, stmt, argnames):
 
         self.add_message("unused-argument", args=name, node=stmt, confidence=confidence)
 
+    @utils.check_messages("cell-var-from-loop")
```
cdce8p, line 1968:
```suggestion
```
Not a `visit_xxx` method.

Btw. there are a few other `visit_xxx` method in this class that don't have the `check_messages` decorator. If they modify `self._to_consume`, I would say we should leave them alone. (`visit_classdef`, `leave_classdef`, `visit_lambda`, `visit_functiondef`, ...)

cdce8p (changes requested):

pylint/checkers/variables.py
```diff
@@ -717,14 +731,21 @@ class VariablesChecker(BaseChecker):
 
     def __init__(self, linter=None):
         super().__init__(linter)
-        self._to_consume = (
-            None  # list of tuples: (to_consume:dict, consumed:dict, scope_type:str)
-        )
+        self._to_consume: List[NamesConsumer] = None
```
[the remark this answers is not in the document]
DanielNoord: 😅

pylint/checkers/variables.py
DanielNoord: Do we want to remove those now? I'm looking at `visit_import` for example

pylint/checkers/variables.py
DanielNoord: Done!

pylint/checkers/variables.py
cdce8p: `visit_import` is fine. Unless I'm missing something? Usually it does make sense to add the decorator, this class is somewhat special though with the `NamesConsumer`.

DanielNoord: Ah, yeah I was looking at the wrong sub-method. Sorry for the confusion!

pylint/checkers/variables.py
```diff
@@ -979,267 +1003,298 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+                if not (
+                    self._postponed_evaluation_enabled
+                    and isinstance(stmt, (nodes.AnnAssign, nodes.FunctionDef))
+                ):
+                    self.add_message(
+                        "used-before-assignment", args=node.name, node=node
+                    )
+                    return (VariableVisitConsumerAction.CONSUME, found_nodes)
+
+                # E0601 can occur in class-level scope in lambdas, as in
```
cdce8p, line 1254:
```suggestion

            elif base_scope_type == "lambda":
                # E0601 can occur in class-level scope in lambdas, as in
```

DanielNoord: I now know why this wrong in the first place. This `elif` doesn't make a lot sense (I think?) 
https://github.com/PyCQA/pylint/blob/6caab13816e8098063b6e077ab0044b94aa87557/pylint/checkers/variables.py#L1165-L1176

DanielNoord: That's why I removed the second elif, but then the indentation was messed up.

DanielNoord: I think the latest commit fixes this (https://github.com/PyCQA/pylint/pull/5241/commits/5c99a866fc6174f49d0e06105f6dc5037c4aebc2)

Now there is one `elif base_scope_type != "lambda":` and then the `elif`'s that were previously under `elif base_scope_type == "lambda": `.
Using `else` doesn't work because then a `else-if-used` warning is raised.

cdce8p: That still missed the `base_scope_type == "lambda"` condition. So "luck" if it doesn't break.
Using `else` would be wrong, as there is another `elif` case afterwards.

DanielNoord: But if `elif base_scope_type != "lambda":` is `False`  (L1165) isn't `base_scope_type == "lambda":` then always `True`?

cdce8p: Hmm, true. However, adding a separate `elif` makes that more explicit. Especially with the block indentation. I just got confused about it in my last comment when I didn't realize that `elif current_consumer.scope_type == "lambda"` isn't an `elif` on the outer if, instead only for the `== "lambda" case.

DanielNoord: Yeah, looking back at the original version the real mistake was `if isinstance(frame, nodes.ClassDef) ...` instead of `elif isinstance(frame, nodes.ClassDef) ...`

I'm not sure what is better and don't really have a preference. For me the current version works (and has less indentation), but I'm fine with adding the unnecessary `if` back agin. What do you want?

cdce8p: I would prefer to stay as close to the original as possible. So with the `elif base_scope_type == "lambda"`

DanielNoord: Done!

pylint/checkers/variables.py
```diff
@@ -717,14 +731,24 @@ class VariablesChecker(BaseChecker):
...
+        """Called when loading the checker"""
+        self._undefined_variable_is_enabled = self.linter.is_message_enabled(
+            "undefined-variable"
+        )
+        self._used_before_assignment_is_enabled = self.linter.is_message_enabled(
+            "used-before-assignment"
+        )
+        self._undefined_loop_variable_is_enabled = self.linter.is_message_enabled(
+            "undefined-loop-variable"
+        )
```
cdce8p, line 750:
```suggestion
        self._is_undefined_variable_enabled = self.linter.is_message_enabled(
            "undefined-variable"
        )
        self._is_used_before_assignment_enabled = self.linter.is_message_enabled(
            "used-before-assignment"
        )
        self._is_undefined_loop_variable_enabled = self.linter.is_message_enabled(
            "undefined-loop-variable"
        )
```
This might be a bit better. Especially as those names will inevitable be uses as templates.

cdce8p (commented):
This should be the last comment

pylint/checkers/variables.py
```diff
@@ -979,267 +1003,299 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
+                    and frame.is_method()
+                )
+            )
+            and not utils.node_ignores_exception(node, NameError)
+        ):
+            self.add_message("undefined-variable", args=node.name, node=node)
+
+    def _node_should_be_skipped(
+        self, node: nodes.Name, consumer: NamesConsumer, is_start_index: bool
+    ) -> bool:
```
cdce8p, line 1072:
```suggestion
    def _should_node_be_skipped(
        self, node: nodes.Name, consumer: NamesConsumer, is_start_index: bool
    ) -> bool:
```

pylint/checkers/variables.py
```diff
@@ -979,267 +1003,299 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
-
         # iterates through parent scopes, from the inner to the outer
         base_scope_type = self._to_consume[start_index].scope_type
-        # pylint: disable=too-many-nested-blocks; refactoring this block is a pain.
+
         for i in range(start_index, -1, -1):
             current_consumer = self._to_consume[i]
 
+            # Certain nodes shouldn't be checked as they get checked another time
+            if self._node_should_be_skipped(node, current_consumer, i == start_index):
```
cdce8p, line 1035:
```suggestion
            if self._should_node_be_skipped(node, current_consumer, i == start_index):
```

cdce8p (approved):
Really the last one this time 😅
Thanks @DanielNoord! 🚀 Lets squash merge it once all tests pass.

Pierre-Sassoulas (commented):
This sorely needed a refactor, and it was a huge one. Thank you @DanielNoord and all the reviewers, in particular @jacobtylerwalls sorry for the rebase conflcits it'll cost you :)

DanielNoord:
@Pierre-Sassoulas want to do a final review? Or is this good to go after CI passes?

PR #5245, opened 2021-11-02
Inspection for `with threading.Lock():`

### LINKED ISSUE #5208: Add inspection for "with threading.Lock():"
### Current problem

We've had a few instances of
```py
with threading.Lock():
  <statements>
```
escape detection during review and find their way into our codebase; these are never correct or desired.

### Desired solution

Would love for the code to be warned with with something like `looks-like-mutual-exclusion-was-wanted-but-this-code-doesn't-achieve-mutual-exclusion` or... a shorter, nicer warning name. 🙂 

### Additional context

For a file with contents
```py
import threading

def not_a_helpful_function(parameter):
    with threading.Lock():
        print(parameter)
```
`pylint-2.11.1` does not detect any logic problem in the code (it complains about the missing doc strings).

MartinBasti:
Using `with threading.Lock():` directly has no effect.

Correct usage is:
```
lock = threading.Lock()
with lock:
    ...
```

This applies for:
  * threading.Lock
  * threading.RLock
  * threading.Condition
  * threading.Semaphore
  * threading.BoundedSemaphore

Fixes: #5208

Signed-off-by: Martin Basti <mbasti@redhat.com>

Thank you for submitting a PR to pylint!

## Type of Changes

## Description

Closes #5208

### DIFF
--- a/pylint/checkers/threading_checker.py
+++ b/pylint/checkers/threading_checker.py
@@ -1,0 +1,55 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+from astroid import nodes
+
+from pylint import interfaces
+from pylint.checkers import BaseChecker
+from pylint.checkers.utils import check_messages, safe_infer
+
+
+class ThreadingChecker(BaseChecker):
+    """Checks for threading module
+
+    - useless with lock - locking used in wrong way that has no effect (with threading.Lock():)
+    """
+
+    __implements__ = interfaces.IAstroidChecker
+    name = "threading"
+
+    LOCKS = frozenset(
+        (
+            "threading.Lock",
+            "threading.RLock",
+            "threading.Condition",
+            "threading.Semaphore",
+            "threading.BoundedSemaphore",
+        )
+    )
+
+    msgs = {
+        "W2101": (
+            "'%s()' directly created in 'with' has no effect",
+            "useless-with-lock",
+            "Used when a new lock instance is created by using with statement "
+            "which has no effect. Instead, an existing instance should be used to acquire lock.",
+        ),
+    }
+
+    @check_messages("useless-with-lock")
+    def visit_with(self, node: nodes.With) -> None:
+
+        context_managers = (c for c, _ in node.items if isinstance(c, nodes.Call))
+        for context_manager in context_managers:
+            if isinstance(context_manager, nodes.Call):
+                infered_function = safe_infer(context_manager.func)
+                if infered_function is None:
+                    continue
+                qname = infered_function.qname()
+                if qname in self.LOCKS:
+                    self.add_message("useless-with-lock", node=node, args=qname)
+
+
+def register(linter):
+    """required method to auto register this checker"""
+    linter.register_checker(ThreadingChecker(linter))
--- a/tests/functional/u/useless/useless_with_lock.py
+++ b/tests/functional/u/useless/useless_with_lock.py
@@ -1,0 +1,58 @@
+"""Tests for the useless-with-lock message"""
+# pylint: disable=missing-docstring
+import threading
+from threading import Lock, RLock, Condition, Semaphore, BoundedSemaphore
+
+
+with threading.Lock():  # [useless-with-lock]
+    ...
+
+with Lock():  # [useless-with-lock]
+    ...
+
+with threading.Lock() as this_shouldnt_matter:  # [useless-with-lock]
+    ...
+
+with threading.RLock():  # [useless-with-lock]
+    ...
+
+with RLock():  # [useless-with-lock]
+    ...
+
+with threading.Condition():  # [useless-with-lock]
+    ...
+
+with Condition():  # [useless-with-lock]
+    ...
+
+with threading.Semaphore():  # [useless-with-lock]
+    ...
+
+with Semaphore():  # [useless-with-lock]
+    ...
+
+with threading.BoundedSemaphore():  # [useless-with-lock]
+    ...
+
+with BoundedSemaphore():  # [useless-with-lock]
+    ...
+
+lock = threading.Lock()
+with lock:  # this is ok
+    ...
+
+rlock = threading.RLock()
+with rlock:  # this is ok
+    ...
+
+cond = threading.Condition()
+with cond:  # this is ok
+    ...
+
+sem = threading.Semaphore()
+with sem:  # this is ok
+    ...
+
+b_sem = threading.BoundedSemaphore()
+with b_sem:  # this is ok
+    ...
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -154,6 +154,7 @@ contributors:
 
 * Martin Bašti: contributor
   Added new check for shallow copy of os.environ
+  Added new check for useless `with threading.Lock():` statement
 
 * Jacques Kvam: contributor
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -140,6 +140,11 @@ Release date: TBA
 
   Closes #5216
 
+* Added new checker ``useless-with-lock`` to find incorrect usage of with statement and threading module locks.
+  Emitted when ``with threading.Lock():`` is used instead of ``with lock_instance:``.
+
+  Closes #5208
+
 
 What's New in Pylint 2.11.2?
 ============================
--- a/doc/whatsnew/2.12.rst
+++ b/doc/whatsnew/2.12.rst
@@ -48,6 +48,10 @@ New checkers
 
   Closes #3370
 
+* Added new checker ``useless-with-lock`` to find incorrect usage of with statement and threading module locks.
+  Emitted when ``with threading.Lock():`` is used instead of ``with lock_instance:``.
+
+  Closes #5208
 
 Removed checkers
 ================
--- a/tests/functional/u/useless/useless_with_lock.txt
+++ b/tests/functional/u/useless/useless_with_lock.txt
@@ -1,0 +1,11 @@
+useless-with-lock:7:0::'threading.Lock()' directly created in 'with' has no effect:HIGH
+useless-with-lock:10:0::'threading.Lock()' directly created in 'with' has no effect:HIGH
+useless-with-lock:13:0::'threading.Lock()' directly created in 'with' has no effect:HIGH
+useless-with-lock:16:0::'threading.RLock()' directly created in 'with' has no effect:HIGH
+useless-with-lock:19:0::'threading.RLock()' directly created in 'with' has no effect:HIGH
+useless-with-lock:22:0::'threading.Condition()' directly created in 'with' has no effect:HIGH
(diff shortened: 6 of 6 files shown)


pylint/checkers/base.py
```diff
@@ -2593,6 +2594,43 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
 
 
+class ThreadingChecker(_BasicChecker):
+    """Checks for threading module
+
+    - with useless lock - locking used in wrong way that has no effect (with threading.Lock())
+    """
+
+    msgs = {
+        "W9999": (  # TODO
```
MartinBasti, line 2604: I don't know how to choose error code

pylint/checkers/base.py
```diff
@@ -2593,6 +2594,41 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+
+    @utils.check_messages("useless-with-lock")
+    def visit_with(self, node: nodes.With) -> None:
+        locks = {
+            "threading.Lock",
+            "threading.RLock",
+            "threading.Condition",
+            "threading.Semaphore",
+            "threading.BoundedSemaphore",
+        }
```
Pierre-Sassoulas, line 2621: This could be a class constant to be declared only once ?

pylint/checkers/base.py
```diff
@@ -2593,6 +2594,41 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+        pairs = node.items
+        for manager, _ in pairs:
+            if isinstance(manager, nodes.Call):
+                inf = safe_infer(manager.func)
+                if inf is not None:
+                    qname = inf.qname()
+                    if qname in locks:
+                        self.add_message("useless-with-lock", node=node, args=qname)
```
Pierre-Sassoulas, line 2629:
```suggestion
        context_managers = [c for c, _ in node.items if isinstance(c, nodes.Call)
        for context_manager in context_managers:
            infered_function = safe_infer(context_manager.func)
            if infered_function is None:
                continue
            qname = infered_function .qname()
            if qname in locks:
                self.add_message("useless-with-lock", node=node, args=qname)
```
Some nitpick to reduce the indentation level.

pylint/checkers/base.py
```diff
@@ -2593,6 +2594,41 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
 
 
+class ThreadingChecker(_BasicChecker):
+    """Checks for threading module
+
+    - useless with lock - locking used in wrong way that has no effect (with threading.Lock():)
+    """
+
+    msgs = {
+        "W9990": (  # TODO
```
Pierre-Sassoulas, line 2604: You can choose the id yourself it just need to not already exists and to not have the same beginning than another checker to not limit other checkers evolution ([2 first number](https://github.com/PyCQA/pylint/blob/6a7209b9b8b9bbd0fdf08ce22daf325c9d0d28ec/pylint/checkers/base_checker.py#L134)). This is definitely something we should document, I'll open an issue.

DanielNoord, line 2604:
```suggestion
        "W2101": (  # TODO
```

I checked and `21xx` is still "free".

pylint/checkers/base.py
```diff
@@ -2593,6 +2594,41 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+            "Useless usage of with statement and %s()",
```
Pierre-Sassoulas, line 2605:
```suggestion
            "'%s()' directly created in 'with' has no effect",
```
I think the formulation of the extended description is nice and could be used here too.

pylint/checkers/base.py
```diff
@@ -2593,6 +2594,41 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+            "useless-with-lock",
+            "Used when a new lock instance is created by using with statement "
+            "which has no effect, instead of using existing instance to acquire lock. "
+            "Valid for various locking objects: Lock, RLock, Condition, Semaphore, BoundedSemaphore.",
```
Pierre-Sassoulas, line 2609:
```suggestion
```
I think this might not need to be specified here and it also has the potential to become out of date if new lock are added later.
```suggestion
            "Valid for various locking objects: {', '.join(ThreadingChecker.LOCK_CLASSES)}.",
```

tests/functional/u/useless/useless_with_lock.py
```diff
@@ -0,0 +1,57 @@
+# pylint: disable=missing-docstring
```
DanielNoord, line 1:
```suggestion
"""Tests for the useless-with-lock message"""
# pylint: disable=missing-docstring
```

I like adding docstrings at module level to describe what a test is supposed to test. Helps when we revisit them years later.
Will need to update the tests after this because of the added line.

tests/functional/u/useless/useless_with_lock.py
```diff
@@ -0,0 +1,57 @@
...
+with cond:  # this is ok
+    ...
+
+sem = threading.Semaphore()
+with sem:  # this is ok
+    ...
+
+b_sem = threading.BoundedSemaphore()
+with b_sem:  # this is ok
+    ...
```
Pierre-Sassoulas, line 57: The tests are on point 👍

pylint/checkers/base.py
```diff
@@ -2593,6 +2594,41 @@ def _check_type_x_is_y(self, node, left, operator, right):
         self.add_message("unidiomatic-typecheck", node=node)
 
 
+class ThreadingChecker(_BasicChecker):
```
Pierre-Sassoulas, line 2597: Would it be possible to add a new file called ``threading_checker.py``, please ? This file is becoming too huge and we might have to burst it later.

DanielNoord:
Thanks for your contribution! This seems well tested.

I think we could consider moving this out of `base.py` and creating a new file for it. Personally I think having this in a separate file would be better as not too "clutter" `base.py`. If we do so, we might want to make `locks a variable at the module level.

Pierre-Sassoulas (changes requested):
Thank you for the new checker ! I'm not sure if we'll be able to put it in 2.12 but it'll definitely go in the next version otherwise :)

pylint/checkers/base.py
MartinBasti: moved

pylint/checkers/base.py
MartinBasti: Removed

pylint/checkers/base.py
MartinBasti: Updated

pylint/checkers/base.py
MartinBasti: Updated

pylint/checkers/base.py
MartinBasti: Done

tests/functional/u/useless/useless_with_lock.py
MartinBasti: Added

pylint/checkers/base.py
MartinBasti: Thanks, updated

pylint/checkers/threading_checker.py
```diff
@@ -0,0 +1,54 @@
...
+from pylint.checkers.utils import safe_infer
+
+
+class ThreadingChecker(BaseChecker):
+    """Checks for threading module
+
+    - useless with lock - locking used in wrong way that has no effect (with threading.Lock():)
+    """
+    __implements__ = interfaces.IAstroidChecker
+    name = "threading"
```
MartinBasti, line 18: not sure here, maybe just `stdlib` ?

pylint/checkers/__init__.py
```diff
@@ -39,7 +39,8 @@
 15: stdlib
 16: python3
 17: refactoring
-18-50: not yet used: reserved for future internal checkers.
+18: threading
```
MartinBasti, line 42: I couldn't find where those numbers are defined

Pierre-Sassoulas: Could probably be handled in #5248 along with the consistency fixes what do you think @DanielNoord ?

pylint/checkers/threading_checker.py
Pierre-Sassoulas: I think ``threading`` is nice

Pierre-Sassoulas (approved):
LGTM ! Thank you, very nice contribution.

pylint/checkers/__init__.py
DanielNoord: I would think #5255 might be better for this. We could base the test on an update list of checkers and their message id prefixes? #5248 was more meant as an easy script to point (new) contributors to when making new checkers.

pylint/checkers/__init__.py
```diff
@@ -39,7 +39,8 @@
...
+19-50: not yet used: reserved for future internal checkers.
```
DanielNoord, line 43:
```suggestion
18-50: not yet used: reserved for future internal checkers.
```

For now, reverting this might be better. In any case, if we want to change this here it should be `21` as that is the prefix we are now using for this checker.

pylint/checkers/threading_checker.py
```diff
@@ -0,0 +1,55 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+from astroid import nodes
+
+from pylint import interfaces
+from pylint.checkers import BaseChecker, utils
+from pylint.checkers.utils import safe_infer
```
DanielNoord, line 8: You are already importing `utils` here. Can we merge these? I think importing `safe_infer` and `check_messages` from `utils` should work well and is what we normally do for these functions in other files.

pylint/checkers/threading_checker.py
```diff
@@ -0,0 +1,55 @@
...
+            "threading.BoundedSemaphore",
+        )
+    )
+
+    msgs = {
+        "W2101": (
+            "'%s()' directly created in 'with' has no effect",
+            "useless-with-lock",
+            "Used when a new lock instance is created by using with statement "
+            "which has no effect, instead of using existing instance to acquire lock.",
```
DanielNoord, line 35:
```suggestion
            "which has no effect. Instead, an existing instance should be used to acquire lock.",
```

DanielNoord (changes requested):
Looks very good to me!

Just some minor changes left and then we can merge this imo!

pylint/checkers/threading_checker.py
MartinBasti: Merged

pylint/checkers/__init__.py
MartinBasti: Reverted

pylint/checkers/threading_checker.py
MartinBasti: Updated

DanielNoord (approved):

MartinBasti:
Thank you for very nice reviews! :rocket: 

I enjoyed this, can recommend 10/10

Pierre-Sassoulas:
It's going to be in 2.12.0 ! Thank you again and glad you enjoyed the review :)

PR #5248, opened 2021-11-03
Add script to generate new message category id

### LINKED ISSUE #5247: How to choose the msgid for a new check or a new checker ?
### Question

Right now, it's unclear how the message id must be chosen.

The developer documentation should be extended.

### Additional context

https://github.com/PyCQA/pylint/pull/5245#discussion_r741510008

> I don't know how to choose error code

MartinBasti 

https://github.com/PyCQA/pylint/pull/5245#discussion_r741693772

> You can choose the id yourself it just need to not already exists and to not have the same beginning than another checker to not limit other checkers evolution (2 first number). This is definitely something we should document, I'll open an issue.

DanielNoord:

## Type of Changes

## Description

Closes #5247 

What about this? @Pierre-Sassoulas

### DIFF
--- a/doc/development_guide/contribute.rst
+++ b/doc/development_guide/contribute.rst
@@ -194,6 +194,9 @@ Tips for Getting Started with Pylint Development
   message to find where the warning is raised,
   and therefore where the logic for that code exists.
 
+* When adding a new checker class you can use the :file:`get_unused_message_id_category.py`
+  script in :file:`./scripts` to get a message id that is not used by
+  any of the other checkers.
 
 Building the documentation
 ----------------------------
--- a/pylint/extensions/__init__.py
+++ b/pylint/extensions/__init__.py
@@ -1,0 +1,17 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+from typing import TYPE_CHECKING
+
+from pylint.utils import register_plugins
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+def initialize(linter: "PyLinter") -> None:
+    """Initialize linter with checkers in the extensions directory"""
+    register_plugins(linter, __path__[0])  # type: ignore # Fixed in https://github.com/python/mypy/pull/9454
+
+
+__all__ = ["initialize"]
--- a/script/get_unused_message_id_category.py
+++ b/script/get_unused_message_id_category.py
@@ -1,0 +1,33 @@
+"""Small script to get a new unused message id category"""
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+from typing import List
+
+from pylint.checkers import initialize as initialize_checkers
+from pylint.extensions import initialize as initialize_extensions
+from pylint.lint.pylinter import PyLinter
+
+
+def register_all_checkers_and_plugins(linter: "PyLinter") -> None:
+    """Registers all checkers and plugins"""
+    linter.cmdline_parser.set_conflict_handler("resolve")
+    initialize_checkers(linter)
+    initialize_extensions(linter)
+
+
+def get_next_code_category(message_ids: List[str]) -> int:
+    categories = sorted({int(i[:2]) for i in message_ids})
+    for i in categories:
+        if i + 1 not in categories:
+            return i + 1
+    return categories[-1] + 1
+
+
+if __name__ == "__main__":
+    pylinter = PyLinter()
+    register_all_checkers_and_plugins(pylinter)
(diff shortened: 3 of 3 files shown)


Pierre-Sassoulas (approved):
Wow, very nice and helpful script! Should there also be a check in MessageDefinitionStore that no checker impede on another checker possible future evolution by using its first 2 number?

DanielNoord:
Although that would be sensible, that would currently fail. See:

https://github.com/PyCQA/pylint/blob/be9e34722f521608a24a0d00e6991913bbad835a/pylint/checkers/classes.py#L465

&

https://github.com/PyCQA/pylint/blob/be9e34722f521608a24a0d00e6991913bbad835a/pylint/checkers/refactoring/recommendation_checker.py#L18

I guess this was based on:

https://github.com/PyCQA/pylint/blob/main/pylint/checkers/__init__.py#L26

But that seems a little outdated.

Pierre-Sassoulas:
Ha, sadly we did not follow the convention and renaming is not really an option. We could include the letter so F02 and C02 are different?

DanielNoord:
Can messages be disabled by id? If not, can we change msg id's without impact for users? If so, we might want to refactor this. I read the linked docstring as referring to specific checkers instead of "checkers that check classes". So, the docstring is outdated anyway and implies every checker file should have its own "message id category".

I don't think a test accepting different letters would achieve much or prevent any future problems. 
We could still run into problems when a checker evolves and needs to issue a new message level that is being used in another checker, that won't be prevented.
We will be warned when one checker issues `C1201` and another does `C1202`, but if we keep an eye on never creating a checker that does not start at `xx01` we can also avoid this as then the current tests will warn that `xx01` already exists.

Pierre-Sassoulas:
> Can messages be disabled by id? 

Yes even though if we encourage user to use the "symbol" (``use-symbolic-message-instead`` : ``'C0103' is cryptic: use '# pylint: disable=invalid-name' instead``)

Frankly I did not check that new msgid were consistent with this rule during review and this feel bothersome for unclear benefits (who is disabling only C02 and not a list of symbol?). I think the only known/used aspect of the message id is the letter at the beginning. Si imo, we might remove all constraints, except the ``<50`` constraint for pylint internal checks (so we don't impact outside pylint extension by breaking a promess), and remove all the outdated docstring and check for checker consistency in BaseChecker too ? This would only affect new checks because already assigned symbol/msgid would stay the same.

DanielNoord:
I would propose to explore if we can reinforce the consistency and limit a prefix to a specific checker.

If you search for `0101`, `0201`, `0301`, etc. you find that only for 4-5 prefixes multiple files seem to use them. There is even an example of `R0101` being changed to `R1702`. I did a quick (non-exhaustive) search and think the total will be around 20-25 messages. Can't we put this in `3.0` and provide a look-up table in the release notes with old and new codes?

We might as well make `3.0` the "argparse and message settings" release 😄

Pierre-Sassoulas:
There is a mechanism in place for renaming msgid/symbol, but it never free the old name so we can keep compatibility with previously added ``disable``. I think a breaking change for that would be a lot of burden placed on the user for no clear benefit, a consistency change would not put us in a much better situation as a lot of already used message would be used still. We could rename everything to make it consistent, but I'm also comfortable with removing the prefixes meanings altogether and not bothering with it ever again.

Using argparse in contrast makes the maintenance SO much easier and will certainly fix a ton of inconsistencies and bugs just by forcing some sanity on the options parsing.

DanielNoord:
I think the prefixes have an important aim: to make sure that a checker can evolve to new message types with the security that the corresponding ids are not used yet.

Take a checker that only emits F3101 and F3102. At some point we might think of a new message for that checker, which does not go at the Fatal level. We would then want to emit E3101. If we do not "reserve" the "31" prefix for this checker it might actually be the case that another checker is already issuing E3101 and E3102. Thus the first checker would then need to emit F3101, F3102 and E3201.
Although this might not create immediate inconsistenties, illogical numbering of messages like this will (probably) create problems in the future (just like we are seeing now). In any case it will make maintaining `pylint` more "illogical" and thus (probably) more difficult.

I can understand that you do not want to change the id for any of the current messages, but I would really argue against removing the restriction all together. I think I'll be able to create a test that sees if a checker has a unique message id prefix, while ignoring the 4/5 prefixes that are currently non-exclusive. I think that would be the best way forward here!

Pierre-Sassoulas:
> I think I'll be able to create a test that sees if a checker has a unique message id prefix, while ignoring the 4/5 prefixes that are currently non-exclusive. I think that would be the best way forward here!

Yes, ok, 5 messages do not seems that bad to become consistent !

DanielNoord:
> Yes, ok, 5 messages to rename do not seems that bad to become consistent !

4/5 prefixes and 20/25 messages. I do agree with you that renaming might not be the best way forward (for now).

But I was thinking of allowing the non-exclusive prefixes to pass the test if they belong to any of the non-exclusive checker classes while only allowing exclusive prefixes and exclusive checkers for the others!

PR #5252, opened 2021-11-03
Remove ``render: python`` from bug description

DanielNoord:

## Description

This is non intuitive and I always need to change my issues to make them text again.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
@@ -23,7 +23,6 @@ body:
         # Please disable message unrelated to the bug
         # pylint: disable=missing-docstring,
         <a> = b + 1
-      render: python
     validations:
       required: true
   - type: textarea


cdce8p (commented):
Maybe we should split this instead? One section for `Bug description` and one for `Example code`.

DanielNoord:
I don't really like the `render: python` thing.

See:
https://github.com/DanielNoord/pylint/issues/33

I put this in the first box and it completely broke the template.

```python
Test
```

or

```python
Test2
```

I think adding two different example code blocks with ``` yourself should be accepted by the issue form without breaking.
So even if we were to add `Example code` I think rendering it as text might be better (until Github improves these renderers).

Pierre-Sassoulas (approved):
I was going to do that too, I'm also annoyed by it. (I had to edit multiple issues). Marc's proposal could work but it would also add another form and we already have a lot of them. Maybe a non mandatory full Python code form ? But it means another header with 'no response ' in it all the time. We can trust reporter to do better I guess.

DanielNoord:
The problem for me is that it is not really clear that the response will be rendered as a python code block (and my previous example shows that you can't give two examples).

If we want to add another field I would propose to add a hint to use a code block in the text shown before typing.

cdce8p:
Ok. Let's try how this works first, before we think about adding anything else.

PR #5260, opened 2021-11-05
Clarify syntax in pylint.lint example

felixvd:
This makes the call in the documentation non-trivial, so that [idiots like me](https://stackoverflow.com/questions/69851634/how-to-run-pylint-from-inside-python/69851635#69851635) don't get stuck.

If someone could show how to get the output of the command (as asked [here](https://stackoverflow.com/questions/56081299/how-to-get-report-score-from-pylint-when-run-python-code)), that would also be very helpful to add.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -1,4 +1,4 @@
-﻿Contributors
+Contributors
 ------------
 
 Current team:
@@ -577,3 +577,5 @@ contributors:
 * Antonio Quarta (sgheppy): contributor
 
 * Harshil (harshil21): contributor
+
+* Felix von Drigalski (felixvd): contributor
--- a/doc/user_guide/run.rst
+++ b/doc/user_guide/run.rst
@@ -42,7 +42,7 @@ thanks to the ``Run()`` function in the ``pylint.lint`` module
 .. sourcecode:: python
 
   import pylint.lint
-  pylint_opts = ['--version']
+  pylint_opts = ['--disable=line-too-long', 'myfile.py']
   pylint.lint.Run(pylint_opts)
 
 To silently run Pylint on a ``module_name.py`` module,


Pierre-Sassoulas (changes requested):
Thank you for the better example ! I think your commit message `so idiots don't me get stuck.` could be reordered to `so it's harder to get stuck.` More seriously could you add yourself to the contributors, please ? :)

I'll keep the stackoverflow question in mind, I did not manage to make it work myself yet, which is a clue that new documentation will be required.

felixvd:
Done. I assumed it's not worth making an entry in the file for such a small contribution, but if it's policy then sure.

DanielNoord:
> I'll keep the stackoverflow question in mind, I did not manage to make it work myself yet, which is a clue that new documentation will be required.

I have investigated this a bit more. Without changes to `Run` I think this is simply impossible. The way we change `.out` and `._output` doesn't really allow to do this with `Run`.
When using `PyLinter` this might be possible, but I'm not sure we want to create documentation telling people to access it directly. In any case, we could open an issue to track this and close this PR (waiting for a change to `Run` seems incorrect).

Pierre-Sassoulas:
> Without changes to Run I think this is simply impossible.

That was my conclusion too. I had the impression it was a problem with kwargs not being taken into account, but I did not go in depth. Do you want to create the issue @DanielNoord ? 

>  I'm not sure we want to create documentation telling people to access it directly

I pretty much agree, I think we need ``Run`` to be the entry-point. Especially as we will want to make change to Pylinter later as it's so bloated.

DanielNoord:
Opened an issue to track the `output` issue :)

CONTRIBUTORS.txt
```diff
@@ -1,4 +1,4 @@
-﻿Contributors
+Contributors
```
Pierre-Sassoulas, line 1:
```suggestion
Contributors
```

Pierre-Sassoulas (approved):
I'm merging as this is still clearer than what we had before. Follow up to be able to recover the output in a stream is #5295.

PR #5268, opened 2021-11-06
Make ``self-cls-assignment`` check tuple assignment

DanielNoord:

## Type of Changes

## Description

Continuation of #5254 

Basically we only checked single assignments and will now also check tuple assignments.
Furthermore, I have removed some redundant code.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -61,7 +61,7 @@ import os
 import re
 import sys
 from functools import lru_cache
-from typing import DefaultDict, List, Tuple
+from typing import DefaultDict, List, Optional, Set, Tuple
 
 import astroid
 from astroid import nodes
@@ -2024,7 +2024,7 @@ class VariablesChecker(BaseChecker):
 
     def _check_self_cls_assign(self, node: nodes.Assign) -> None:
         """Check that self/cls don't get assigned"""
-        assign_names: Set[Union[str, None]] = set()
+        assign_names: Set[Optional[str]] = set()
         for target in node.targets:
             if isinstance(target, nodes.AssignName):
                 assign_names.add(target.name)
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2024,16 +2024,17 @@ class VariablesChecker(BaseChecker):
 
     def _check_self_cls_assign(self, node):
         """Check that self/cls don't get assigned"""
-        assign_names = {
-            target.name
-            for target in node.targets
-            if isinstance(target, nodes.AssignName)
-        }
+        assign_names = []
+        for target in node.targets:
+            if isinstance(target, nodes.AssignName):
+                assign_names.append(target.name)
+            elif isinstance(target, nodes.Tuple):
+                assign_names.extend(
+                    elt.name for elt in target.elts if isinstance(elt, nodes.AssignName)
+                )
         scope = node.scope()
         nonlocals_with_same_name = any(
-            child
-            for child in scope.body
-            if isinstance(child, nodes.Nonlocal) and assign_names & set(child.names)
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
Nice cleanup on top of the fix ! 👏

pylint/checkers/variables.py
```diff
@@ -2024,16 +2024,17 @@ def _store_type_annotation_names(self, node):
 
     def _check_self_cls_assign(self, node):
```
cdce8p, line 2025:
```suggestion
    def _check_self_cls_assign(self, node: nodes.Assign) -> None:
```
While we are at it, it makes sense to also add type annotations.

pylint/checkers/variables.py
```diff
@@ -2024,16 +2024,17 @@ def _store_type_annotation_names(self, node):
...
         """Check that self/cls don't get assigned"""
-        assign_names = {
-            target.name
-            for target in node.targets
-            if isinstance(target, nodes.AssignName)
-        }
+        assign_names = []
```
cdce8p, line 2027:
```suggestion
        assign_names: Set[Union[str, None]] = set()
```
Why not use a `set` here?
Unfortunately `name` can still be `str | None`, although in practice it will always be `str`. Already deprecated in `astroid`.

pylint/checkers/variables.py
```diff
@@ -2024,16 +2024,17 @@ def _store_type_annotation_names(self, node):
...
+        for target in node.targets:
+            if isinstance(target, nodes.AssignName):
+                assign_names.append(target.name)
```
cdce8p, line 2030:
```suggestion
                assign_names.add(target.name)
```

pylint/checkers/variables.py
```diff
@@ -2024,16 +2024,17 @@ def _store_type_annotation_names(self, node):
...
+            elif isinstance(target, nodes.Tuple):
+                assign_names.extend(
```
cdce8p, line 2032:
```suggestion
                assign_names.update(
```

pylint/checkers/variables.py
```diff
@@ -2024,16 +2024,17 @@ def _store_type_annotation_names(self, node):
...
+                    elt.name for elt in target.elts if isinstance(elt, nodes.AssignName)
+                )
         scope = node.scope()
         nonlocals_with_same_name = any(
-            child
-            for child in scope.body
-            if isinstance(child, nodes.Nonlocal) and assign_names & set(child.names)
+            child for child in scope.body if isinstance(child, nodes.Nonlocal)
```
cdce8p, line 2037: It seems like we also need to do some more work on `node.scope()`.
```
Cannot access member "body" for type "LocalsDictNodeNG"
  Member "body" is unknown
```

pylint/checkers/variables.py
```diff
@@ -2022,18 +2022,19 @@ def _store_type_annotation_names(self, node):
...
 
-    def _check_self_cls_assign(self, node):
+    def _check_self_cls_assign(self, node: nodes.Assign) -> None:
         """Check that self/cls don't get assigned"""
-        assign_names = {
-            target.name
-            for target in node.targets
-            if isinstance(target, nodes.AssignName)
-        }
+        assign_names: Set[Union[str, None]] = set()
```
cdce8p, line 2027: `Set` and `Union` need to be added as imports.

cdce8p (approved):
Looks good! Can be merged once the imports are fixed.

pylint/checkers/variables.py
DanielNoord: Changed to use `Optional`. That should be fine as well right? I'll merge if you give the okay!

cdce8p: Optional works too.

PR #5272, opened 2021-11-07
Add mermaidjs as format output for pyreverse

sgheppy:
This pull request add the [mermaidjs](https://mermaid-js.github.io/mermaid/#/classDiagram) output format (as text like plantuml)

For the packages graph the classDiagram type of diagram is used for mermaidjs, because is the nearest type of graph to the package diagram for this tool.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     rev: v4.0.1
     hooks:
       - id: trailing-whitespace
-        exclude: "tests/functional/t/trailing_whitespaces.py"
+        exclude: "tests/functional/t/trailing_whitespaces.py|tests/pyreverse/data/.*.html"
       - id: end-of-file-fixer
         exclude: "tests/functional/m/missing/missing_final_newline.py|tests/functional/t/trailing_newlines.py"
   - repo: https://github.com/myint/autoflake
@@ -94,3 +94,4 @@ repos:
     hooks:
       - id: prettier
         args: [--prose-wrap=always, --print-width=88]
+        exclude: tests(/.*)*/data
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,8 @@ What's New in Pylint 2.13.0?
 ============================
 Release date: TBA
 
+* Pyreverse - add output in mermaidjs format
+
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
--- a/pylint/pyreverse/mermaidjs_printer.py
+++ b/pylint/pyreverse/mermaidjs_printer.py
@@ -1,0 +1,83 @@
+# Copyright (c) 2021 Antonio Quarta <andi.finkler@gmail.com>
+
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+"""
+Class to generate files in mermaidjs format
+"""
+from typing import Dict, Optional
+
+from pylint.pyreverse.printer import EdgeType, NodeProperties, NodeType, Printer
+from pylint.pyreverse.utils import get_annotation_label
+
+
+class MermaidJSPrinter(Printer):
+    """Printer for MermaidJS diagrams"""
+
+    DEFAULT_COLOR = "black"
+
+    NODES: Dict[NodeType, str] = {
+        NodeType.CLASS: "class",
+        NodeType.INTERFACE: "class",
+        NodeType.PACKAGE: "package",
+    }
+    ARROWS: Dict[EdgeType, str] = {
+        EdgeType.INHERITS: "--|>",
+        EdgeType.IMPLEMENTS: "..|>",
+        EdgeType.ASSOCIATION: "--*",
+        EdgeType.USES: "-->",
+    }
+
+    def _open_graph(self) -> None:
+        """Emit the header lines"""
+        self.emit("classDiagram ")
+        self._inc_indent()
+
+    def emit_node(
+        self,
+        name: str,
+        type_: NodeType,
+        properties: Optional[NodeProperties] = None,
+    ) -> None:
+        """Create a new node. Nodes can be classes, packages, participants etc."""
+        if properties is None:
+            properties = NodeProperties(label=name)
+        stereotype = "~~Interface~~" if type_ is NodeType.INTERFACE else ""
+        nodetype = self.NODES[type_]
+        body = []
+        if properties.attrs:
+            body.extend(properties.attrs)
+        if properties.methods:
+            for func in properties.methods:
+                args = self._get_method_arguments(func)
+                line = f"{func.name}({', '.join(args)})"
+                if func.returns:
+                    line += " -> " + get_annotation_label(func.returns)
+                body.append(line)
+        name = name.split(".")[-1]
+        self.emit(f"{nodetype} {name}{stereotype} {{")
+        self._inc_indent()
+        for line in body:
+            self.emit(line)
+        self._dec_indent()
+        self.emit("}")
+
+    def emit_edge(
+        self,
+        from_node: str,
+        to_node: str,
+        type_: EdgeType,
+        label: Optional[str] = None,
+    ) -> None:
+        """Create an edge from one node to another to display relationships."""
+        from_node = from_node.split(".")[-1]
+        to_node = to_node.split(".")[-1]
+        edge = f"{from_node} {self.ARROWS[type_]} {to_node}"
+        if label:
+            edge += f" : {label}"
+        self.emit(edge)
+
+    def _close_graph(self) -> None:
+        """Emit the lines needed to properly close the graph."""
+        self._dec_indent()
--- a/pylint/pyreverse/mermaidjs_printer.py
+++ b/pylint/pyreverse/mermaidjs_printer.py
@@ -81,3 +81,27 @@ class MermaidJSPrinter(Printer):
     def _close_graph(self) -> None:
         """Emit the lines needed to properly close the graph."""
         self._dec_indent()
+
+
+class HTMLMermaidJSPrinter(MermaidJSPrinter):
+    """Printer for MermaidJS diagrams wrapped in an html boilerplate"""
+
+    HTML_OPEN_BOILERPLATE = """
+    <html>
+        <body>
+            <script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>
+        <div class="mermaid">
+    """
+
+    HTML_CLOSE_BOILERPLATE = """
+            </div>
+        </body>
+    </html>
+"""
+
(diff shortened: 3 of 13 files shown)


DudeNr33:
Very nice, did not know of mermaidjs before. Can you add tests as well please? You can find the existing tests under ``tests/pyreverse/test_writer.py``. Will review the rest of the PR as soon as the tests are added!

pylint/pyreverse/mermaidjs_printer.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2021 Andreas Finkler <andi.finkler@gmail.com>
```
DudeNr33, line 1: I think your name is the one that should appear here! 😊

ChangeLog
```diff
@@ -11,6 +11,8 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.12.rst'
 
+* Pyreverse - add output in PlantUML format
```
DudeNr33, line 14: That should probably be mermaidjs instead of PlantUML.

doc/whatsnew/2.12.rst
```diff
@@ -12,6 +12,10 @@ Summary -- Release highlights
 New checkers
 ============
 
+* Pyreverse - add output in mermaid-js format
+
+* Fix ``install graphiz`` message which isn't needed for puml output format.
+
```
DudeNr33, line 18: This should be moved into the "Other changes" section. There also already is a bullet point for your last fix, so the "Fix ``install graphviz`` message ..." can be removed.

DudeNr33 (commented):
I already left some remarks in the other files, but I will wait for the tests before I start reviewing the new ``MermaidJSPrinter``.

pylint/pyreverse/mermaidjs_printer.py
Pierre-Sassoulas: You can remove the copyright it will be automatically generated before release :)

DudeNr33:
I was thinking if the final output file should already contain the minimal HTML boilerplate necessary to directly open the result in a browser. I think being able to directly view the diagram without having to install Graphviz or PlantUML is one of the advantages of the mermaidjs format. 
It does not require much to get it to work:
```html
<html>

<body>
    <script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>
    <div class="mermaid">
        graph TD
        A[Party] -->|Get money| B(Go shopping!!)
    </div>
</body>

</html>
```
It would be even possible to omit the ``html`` tag, but I think it's cleaner with it.

@Pierre-Sassoulas what are your thoughts on this?

Pierre-Sassoulas:
> I was thinking if the final output file should already contain the minimal HTML boilerplate necessary to directly open the result in a browser.

It depends if you want to use it directly or feed it in another program to be included elsewhere. I think both use are valid, the default could be viewable directly as html, and there could be an option to only output the code without boilerplate ? What do you think @sgheppy ?

sgheppy:
I think is a good idea.
maybe could be an alternative output like png for graphviz.
So if the requested output format is mmd the simple mermaidjs format is returned.
On the other hand, if the requested output is html, the minimal HTML boilerplate is added.

What do you think @Pierre-Sassoulas?

Pierre-Sassoulas:
>  if the requested output is html, the minimal HTML boilerplate is added.

Amazing design idea ! This is very intuitive.

DudeNr33:
I like that idea as well. 
Thanks for adding the tests!
It would be also nice if you could add the new output options to the documentation (``doc/additional_commands/index.rst``).

sgheppy:
the precommit check, edit the html data check file. How i can avoid that?

DudeNr33:
It's best practice to also [install pre-commit in your own development environment](https://pylint.pycqa.org/en/latest/development_guide/contribute.html?highlight=pre-commit#repository).

As you already noticed, in your case that is not enough. You need to exclude the directory for the output files from the pre-commit hook that alters them (my guess would be ``prettier``, but you'll see that when you run it locally). This can be done in ``.pre-commit-config.yaml``.

.pre-commit-config.yaml
```diff
@@ -60,7 +60,7 @@ repos:
             "--load-plugins=pylint.extensions.docparams",
           ]
         # disabled plugins: pylint.extensions.mccabe
-        exclude: tests/functional/|tests/input|tests/extensions/data|tests/regrtest_data/|tests/data/|doc/
+        exclude: tests/functional/|tests/input|tests/extensions/data|tests/regrtest_data/|tests/data/|tests/pyreverse/data/|doc/
```
DudeNr33, line 63: This is not the correct place for the exclude pattern. You can add different excludes for each hook, and my guess is that the HTML reformatting is performed by ``prettier`` (the last hook in the file).
If you install ``pre-commit`` on your own machine, you can see which hook changes which files before you commit.

You can install it by following the pylint docs:

> * install pre-commit using ``pip install pre-commit``
> * then run ``pre-commit install`` in the pylint root directory to enable the git hooks.

sgheppy: thank you for your suggestions, i really appreciate those. I'm using gitpod to edit, and seems that pre-commit has an issue in that behavior. After a (not so) quick and a (really) dirty fix i used pre-commit in this environment.

I hope it's ok now.

Thank you for your support

.pre-commit-config.yaml
```diff
@@ -6,7 +6,7 @@ repos:
     rev: v4.0.1
     hooks:
       - id: trailing-whitespace
-        exclude: "tests/functional/t/trailing_whitespaces.py"
+        exclude: "tests/functional/t/trailing_whitespaces.py|tests(/.*)*/data"
```
Pierre-Sassoulas, line 9: Could we target specifically the html for pyreverse's tests here and for prettier, please ?
```suggestion
        exclude: "tests/functional/t/trailing_whitespaces.py|tests/pyreverse/data/*.html"
```

sgheppy: Sure

doc/whatsnew/2.12.rst
```diff
@@ -12,6 +12,10 @@ Summary -- Release highlights
 New checkers
 ============
 
+* Pyreverse - add output in mermaid-js format and html which is an mermaid js diagram with html boilerplate
+
+* Fix ``install graphiz`` message which isn't needed for puml output format.
+
 * Added ``missing-any-param-doc`` triggered when a function has neither parameter nor parameter type
```
DudeNr33, line 19: Can you please move this to ``doc/whatsnew/2.13.rst``, and put it below "Extensions" instead.

pylint/pyreverse/mermaidjs_printer.py
```diff
@@ -0,0 +1,107 @@
...
+    HTML_CLOSE_BOILERPLATE = """
+            </div>
+        </body>
+    </html>
+"""
+
+    def _open_graph(self) -> None:
+        self.emit(self.HTML_OPEN_BOILERPLATE)
+        super()._open_graph()
+
```
DudeNr33, line 105: A few more ``_inc_indent()`` would be nice so that the mermaid code has a proper indentation under the ``div``.

sgheppy: added indentation

doc/whatsnew/2.12.rst
```diff
@@ -82,6 +86,8 @@ New checkers
 Extensions
 ==========
 
+* Pyreverse - add output in mermaid-js format and html which is an mermaid js diagram with html boilerplate
```
DudeNr33, line 89: This still needs to be moved to ``doc/whatsnew/2.13.rst`` I think, as Pierre set the milestone to 2.13.0 for this PR.
@Pierre-Sassoulas or do you plan to include it in an upcoming 2.12.x release?

ChangeLog
DudeNr33: This still needs to be changed to MermaidJS, and moved to the ``2.13`` section.

doc/whatsnew/2.12.rst
sgheppy: Sorry, it was an oversight. Thank you for your reviews.

Pierre-Sassoulas (approved):
LGTM, thanks a lot @sgheppy !  @DudeNr33 I'll let you merge when it's ok for you :)

Pierre-Sassoulas:
Sorry I forgot that we need when to release pylint 2.12.2 first, I'm adding the blocked label.

sgheppy:
> LGTM, thanks a lot @sgheppy !  @DudeNr33 I'll let you merge when it's ok for you :) 

Thank you all for your support!

sgheppy:
> Sorry I forgot that we need when to release pylint 2.12.2 first, I'm adding the blocked label. 

Do I have to do something about it?

Pierre-Sassoulas:
> Do I have to do something about it?

No, it's maintenance work, we'll merge and rebase if required when it's ready  :)

DudeNr33 (approved):
No more comments from my side, thanks for incorporating all the feedback @sgheppy!

@Pierre-Sassoulas I'll wait until the ``blocked`` label is removed before I merge.
I think "rebase and merge" would be the best option in this case then?

Pierre-Sassoulas:
> I think "rebase and merge" would be the best option in this case then?

Generally when there is a pre-commit automated commit I squash because it means the author did not clean up its tree. I also consider the number of added / removed line (how messy would it be with only one commit?), and quality of messages for the commit. If there are only quality messages I rebase. If there are unclear message and a lot of added OR a lot of removed it's not an issue to have a big commit so I squash. For a lot of added AND a lot of remove with bad commit message I usually ask for a rebase/cleanup from the author. (Right now I think we could squash because  a single commit with almost only addition make sense). I'm just sharing my thought process, to get the better end result it's not a hard "rule" 😄

DudeNr33:
Thanks for the guideline!

DanielNoord:
After this merge `pytest -k test_writer` leaves test files in the repository. This is a bit annoying when running the full test suite. I haven't looked at the code myself yet, but would anyone be able to do a follow up PR so the files created during testing are removed after testing is done?

sgheppy:
> After this merge `pytest -k test_writer` leaves test files in the repository. This is a bit annoying when running the full test suite. I haven't looked at the code myself yet, but would anyone be able to do a follow up PR so the files created during testing are removed after testing is done?

I'll do something for this.
Thank you for your suggestion.

DudeNr33:
I missed that in the review. Currently the ``_setup`` function used by the fixtures takes care of cleaning up, but you need to add the ``MMD_FILES`` and ``HTML_FILES`` lists explicitly in the ``for`` loop after the ``yield``.

sgheppy:
> I missed that in the review. Currently the `_setup` function in `tests/pyreverse/test_writer.py` used by the fixtures takes care of cleaning up, but you need to add the `MMD_FILES` and `HTML_FILES` lists explicitly in the `for` loop after the `yield`.

suggested edits implemented in #5476

PR #5278, opened 2021-11-08
Enable missing-raises-doc to understand class hierarchies

### LINKED ISSUE #4955: docparams check doesn't understand class hierarchies
### Bug description

```python
class MyError(ValueError):
    pass

def foo():
    """Foo method

    Raises:
        ValueError: all the time
    """
    raise MyError()
```

### Configuration

_No response_

### Command used

```shell
pylint x.py --disable all --load-plugins pylint.extensions.docparams --enable missing-raises-doc
```

### Pylint output

```shell
************* Module x
x.py:5:0: W9006: "MyError" not documented as being raised (missing-raises-doc)

Your code has been rated at 7.50/10 (previous run: 2.50/10, +5.00)
```

### Expected behavior

No error;

I understand that checking the hierarchy is more complicated, but I guess there are valid use cases to expose only a general exception

### Pylint version

```shell
pylint: 2.10.2
astroid: 2.7.3
python: 3.7.1
```

### OS / Environment

Linux

### Additional dependencies

_No response_

kasium:
## Type of Changes

## Description

Before, missing-raises-doc could not understand class hierarchies but just compared names.
So, when a method documented a raise of a parent class, but a child exception was raised, the check failed.

With this change, if the name compare doesn't help, the exception class hierarchy is checked.
For that, possible_exc_types was changed to return classes instead of names

Closes #4955


pylint/extensions/docparams.py
```diff
@@ -308,14 +308,28 @@ def visit_raise(self, node: nodes.Raise) -> None:
...
+        missing_excs = set()
+        for expected in expected_excs:
+            for found_exc in found_excs_class_names:
+                if found_exc == expected.name:
+                    break
+                if any(
+                    ancestor
+                    for ancestor in expected.ancestors()
+                    if found_exc == ancestor.name
+                ):
```
areveny, line 328: Could this be
```
                if any(
                    found_exc == ancestor.name 
                    for ancestor in expected.ancestors()
                ):
```
?

kasium: done. Thanks for the idea

tests/extensions/test_check_raise_docs.py
```diff
@@ -105,6 +105,63 @@ def my_func(self):
...
+
+            Raises:
+                NameError: Never
+            """
+            raise CustomError('hi')
+        '''
+        )
+        raise_node = node.body[0]
+        with self.assertNoMessages():
+            self.checker.visit_raise(raise_node)
```
areveny, line 141: How about adding a test that's the opposite of this? If `CustomError` is in the docstring and we raise `NameError`, we emit should emit `missing-raises-doc` with `NameError`.

This test would ensure we don't get `expected` and `found` mixed up and we're checking against the right chain of ancestors.

areveny:
Hi, I looked and debugged the code more thoroughly and it all looks good to me! This is really great.

tests/extensions/test_check_raise_docs.py
kasium: fair point. Added

kasium:
So, can this now be merged?

Pierre-Sassoulas:
> So, can this now be merged?

We're going to release 2.12.2, before being able to review, sorry :)

Meanwhile you can rebase on the latest main and launch the primer tests (locally or in CI), they were added for 2.12 and might catch issues with the code. Also I had a quick look, I think the test could be done by using the functional testing framework with less boilerplate, you can move them to ``tests/functional``. Do not hesitate to ask questions about it, here's an example of MR moving test to functional : #5404

Pierre-Sassoulas:
> After this is merged, I happy to open another one to make the move happen. Is that fine?

Sure, that would be great !

kasium:
Thanks grat. So, can this be merged now or are there other questions. I'm not quite familiar with your release process

Pierre-Sassoulas:
> I'm not quite familiar with your release process

You can follow 2.12.2 progress here: https://github.com/PyCQA/pylint/milestone/54 We're going to start merging 2.13 MR / issues once 2.12.2 is released.

kasium:
Thanks!

DanielNoord:
@kasium `2.12.2` has been released! Could you merge main into your branch and resolve the conflicts? 😄

kasium:
@Pierre-Sassoulas can you please put the changelog to the right place?

```
* ``missing-raises-doc`` will now check the class hierarchy of the raised exceptions

  Closes #4955
```

tests/extensions/test_check_docs_utils.py
```diff
@@ -147,5 +147,7 @@ def my_func():
     ],
 )
 def test_exception(raise_node, expected):
-    found = utils.possible_exc_types(raise_node)
-    assert found == expected
+    founded = utils.possible_exc_types(raise_node)
+    for node in founded:
+        assert isinstance(node, astroid.nodes.ClassDef)
+    assert {found.name for found in founded} == expected
```
Pierre-Sassoulas, line 153:
```suggestion
    found_nodes = utils.possible_exc_types(raise_node)
    for node in found_nodes:
        assert isinstance(node, astroid.nodes.ClassDef)
    assert {node.name for node in found_nodes} == expected
```

pylint/extensions/docparams.py
```diff
@@ -309,14 +309,24 @@ def visit_raise(self, node: nodes.Raise) -> None:
         doc = utils.docstringify(func_node.doc, self.config.default_docstring_type)
         if not doc.is_valid():
             if doc.doc:
-                self._handle_no_raise_doc(expected_excs, func_node)
+                missing = {exc.name for exc in expected_excs}
```
Pierre-Sassoulas, line 312: 👍

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.py
```diff
@@ -0,0 +1,45 @@
...
+
+class CustomError(NameError):
+    pass
+
+
+class CustomChildError(CustomError):
+    pass
+
+
+def test_find_missing_raise_for_parent():
```
Pierre-Sassoulas, line 12: We have an unexpected ``missing-raises-doc`` here. It's something that really need to be fixed as we have documentation for the raises line 15/16.

Pierre-Sassoulas (changes requested):

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.py
kasium: Why does the tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.txt not work?

Pierre-Sassoulas: I'm not sure, the way ``missing-raises-doc`` raise messages probably changed ?

kasium: forgot a code comment...

pylint/extensions/_check_docs_utils.py
```diff
@@ -164,10 +164,17 @@ def possible_exc_types(node):
...
-        return {exc for exc in excs if not utils.node_ignores_exception(node, exc)}
+        final_exceptions = set()
+        for exc in excs:
+            if utils.node_ignores_exception(node, exc.name):
+                continue
+            if not isinstance(exc, nodes.ClassDef):
+                exc = exc.getattr("__class__")[0]
+            final_exceptions.add(exc)
+        return final_exceptions
     except astroid.InferenceError:
```
DanielNoord, line 178: This try-except statement is only needed for `utils.node_ignores_exception` because it is trying to infer the handler of the exception, see https://github.com/PyCQA/pylint/commit/5f048faf002f2af8de11ae50e3567ce872569f1a

I think we should only include that if statement in the try-except and leaving the others out. A comment explaining why this can raise `InferenceError` also wouldn't hurt.

pylint/extensions/_check_docs_utils.py
```diff
@@ -164,10 +164,17 @@ def possible_exc_types(node):
...
-                    excs.append(val.name)
+                    excs.append(val)
 
     try:
-        return {exc for exc in excs if not utils.node_ignores_exception(node, exc)}
+        final_exceptions = set()
+        for exc in excs:
+            if utils.node_ignores_exception(node, exc.name):
+                continue
+            if not isinstance(exc, nodes.ClassDef):
```
DanielNoord, line 174: Why can `exc` not be a `ClassDef` at this point? The coverage report shows that this line is covered by the tests so it seems necessary but I wonder if we shouldn't "fix" this before iterating over `excs`.

pylint/extensions/docparams.py
```diff
@@ -309,14 +309,24 @@ def visit_raise(self, node: nodes.Raise) -> None:
...
+                self._handle_no_raise_doc(missing, func_node)
             return
 
         found_excs_full_names = doc.exceptions()
-
```
DanielNoord, line 316: Could you re-add this? These white lines make the code a little more readable imo.

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.py
```diff
@@ -0,0 +1,45 @@
+"""Tests for missing-raises-doc for class inheritance."""
```
DanielNoord, line 1:
```suggestion
"""Tests for missing-raises-doc for exception class inheritance."""
```

pylint/extensions/docparams.py
```diff
@@ -309,14 +309,24 @@ def visit_raise(self, node: nodes.Raise) -> None:
...
         # Extract just the class name, e.g. "error" from "re.error"
         found_excs_class_names = {exc.split(".")[-1] for exc in found_excs_full_names}
-        missing_excs = expected_excs - found_excs_class_names
+
+        missing_excs = set()
+        for expected in expected_excs:
```
DanielNoord, line 321: Let's use `itertools.zip_longest` here, see [this](https://stackoverflow.com/questions/11318977/zipping-lists-of-unequal-size)

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
```diff
@@ -0,0 +1,2 @@
+[MASTER]
+load-plugins = pylint.extensions.docparams
```
DanielNoord, line 2: What happens if we add:
```ini
[BASIC]
accept-no-raise-doc=no
```

DanielNoord (changes requested):
@kasium Sorry for taking so long to review this!

This is a very sensible change so I'm happy we're working on implementing this. Let me know if you have any questions!

pylint/extensions/_check_docs_utils.py
kasium: I guess there is one corner case. I can't remember which test case but this line is needed. Also this follows the old logic. To be honest I don't feel comfortable to dig too deep, since I'm now expert in this code

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
kasium: This follows all the other tests. Why making another think here?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "python.formatting.provider": "black"
```
DanielNoord, line 2: Could you remove this file from the commit history?

pylint/extensions/_check_docs_utils.py
```diff
@@ -120,7 +120,7 @@ def _split_multiple_exc_types(target: str) -> List[str]:
     return re.split(delimiters, target)
 
 
-def possible_exc_types(node):
+def possible_exc_types(node: nodes.NodeNG) -> Set[nodes.ClassDef]:
```
DanielNoord, line 123: Thanks, also for moving `nodes.NodeNG` 😄

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
DanielNoord: Some of the other tests also use this setting. I think since we're specifically looking at `raise` documentation I think it makes sense to require it.

pylint/extensions/_check_docs_utils.py
```diff
@@ -164,12 +162,20 @@ def possible_exc_types(node):
...
-    except astroid.InferenceError:
-        return set()
+    final_exceptions = set()
+    for exc in excs:
+        try:
+            # can fail because it is trying to infer the handler of the exception
+            if utils.node_ignores_exception(node, exc.name):
+                continue
+        except astroid.InferenceError:
+            return set()
```
DanielNoord, line 174: I wonder if this should be `continue`. I know this is mimicking previous behaviour, but could it be better to continue instead of returning prematurely?

pylint/extensions/_check_docs_utils.py
```diff
@@ -164,12 +162,20 @@ def possible_exc_types(node):
...
+                    excs.append(val)
 
-    try:
-        return {exc for exc in excs if not utils.node_ignores_exception(node, exc)}
-    except astroid.InferenceError:
-        return set()
+    final_exceptions = set()
+    for exc in excs:
+        try:
+            # can fail because it is trying to infer the handler of the exception
```
DanielNoord, line 170:
```suggestion
            # This block can raise InferenceError because it is inferring the handler of the exception
```

pylint/extensions/_check_docs_utils.py
```diff
@@ -164,12 +162,20 @@ def possible_exc_types(node):
...
+            if utils.node_ignores_exception(node, exc.name):
+                continue
+        except astroid.InferenceError:
+            return set()
+        if not isinstance(exc, nodes.ClassDef):
```
DanielNoord, line 175: Since you force-pushed I can't respond to the previous comment, but:

This doesn't really follow the previous logic as this was not part of the original `try ... except`. We need this because of changes to `excs` and thus I think we should handle this corner case earlier.

pylint/extensions/docparams.py
DanielNoord: Something based on:
```suggestion
        for expected, found_exc in itertools.zip_longest(expected_excs, found_excs_class_names)
```

This allows removing an indent with the additional `for` loop.

.vscode/settings.json
Pierre-Sassoulas: You can add this file to your global gitignore.

kasium: thanks, removed

pylint/extensions/_check_docs_utils.py
kasium: Know clue what the side-effects are

pylint/extensions/_check_docs_utils.py
kasium: Sorry for the forced push. I'll check

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
kasium: Now I get
```
self = <pylint.testutils.lint_module_test.LintModuleTest object at 0x7f8c00d26e10>

    def runTest(self) -> None:
>       self._runTest()
E       AssertionError: Wrong results for file "missing_raises_doc_exc_inheritance":
E       
E       Expected in testdata:
E         12: missing-raises-doc

pylint/testutils/lint_module_test.py:88: AssertionError
```

pylint/extensions/_check_docs_utils.py
kasium: So, the code is not longer part of the try-except. What would be the benefit for handling this earlier?

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
DanielNoord: Guess we found a bug then. I think with requiring `raises` doc we should still expect one `missing-raises-doc` message.

pylint/extensions/_check_docs_utils.py
DanielNoord: Would you be willing to try and see what happens with this change?

pylint/extensions/_check_docs_utils.py
DanielNoord: It makes the corner case more specific. You had to add this because of some currently unknown corner case. Documenting and handling such corner cases explicitly is important for future maintainability of the code. If we ever want to refactor this code we won't know what this is doing which would make that much more difficult.

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
kasium: I don't understand the config. Can you please explain what it does and what I need to change in the code?

pylint/extensions/_check_docs_utils.py
kasium: Let's see what the test base says

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
DanielNoord: With `missing-raises-doc=no` docstrings are required to document all exceptions that are found within the function or class definition. Since we are already (supposedly) documenting the exception this shouldn't create a new test failure so this is probably a bug in the current implementation of this feature.

kasium: Sorry, my bad. I have read replace instead of add. Adding these two lines changes nothing. Tests are fine

pylint/extensions/_check_docs_utils.py
kasium: So what is your suggestion? Can you please point me to the line where you see that and in which form. Personally I think the place is fine, but of course I don't have the big picture of pylint in mind

tests/functional/ext/docparams/raise/missing_raises_doc_exc_inheritance.rc
DanielNoord: Yes please! Could you then also update the test file names to include `required`? You can look at the other similar files in the directory.

kasium:
@DanielNoord `itertools.zip_longest(expected_excs, found_excs_class_names):` now leads to None issues...

pylint/extensions/_check_docs_utils.py
DanielNoord: Somewhere in between L137-165 as that is where we're actually determining the possible exception types.
(The original) L169-172  is only used to check for `utils.node_ignores_exception` for each possible node.

I think it would be if we could first determine what the actual corner case is that this is catching and then see where we should "handle" it. Determining the corner case would also allow us to add a comment explaining what the code is trying to do.

DanielNoord:
> @DanielNoord `itertools.zip_longest(expected_excs, found_excs_class_names):` now leads to None issues...

Please look at the docs for itertools ([here](https://docs.python.org/3/library/itertools.html)). Basically `zip_longest` cane be used to merge the two original `for` loops into one. We should also handle the case when both iterables do not have the same length but that should be done inside the `for` loop based on `zip_longest`

kasium:
So, I get new errors. I must admit, I have no clue why the occur. Before everything was fine:

```
  File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/decorators.py", line 142, in raise_if_nothing_inferred
    yield next(generator)
StopIteration

The above exception was the direct cause of the following exception:

  File "/root/git/pylint/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "/root/git/pylint/pylint/extensions/docparams.py", line 298, in visit_raise
    expected_excs = utils.possible_exc_types(node)
  File "/root/git/pylint/pylint/extensions/_check_docs_utils.py", line 148, in possible_exc_types
    excs = [exc for exc in inferred_excs if exc is not astroid.Uninferable]
  File "/root/git/pylint/pylint/extensions/_check_docs_utils.py", line 148, in <listcomp>
    excs = [exc for exc in inferred_excs if exc is not astroid.Uninferable]
  File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/decorators.py", line 142, in raise_if_nothing_inferred
    yield next(generator)
  File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/nodes/node_classes.py", line 88, in unpack_infer
    inferred = next(stmt.infer(context), util.Uninferable)
  File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/nodes/node_ng.py", line 156, in infer
    yield from self._infer(context, **kwargs)
  File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/decorators.py", line 150, in raise_if_nothing_inferred
    ) from error
astroid.exceptions.InferenceError: StopIteration raised without any error information.
```

DanielNoord:
> So, I get new errors. I must admit, I have no clue why this occurs. Before everything was fine:
> 
> ```
> Traceback (most recent call last):
>   File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/decorators.py", line 142, in raise_if_nothing_inferred
>     yield next(generator)
> StopIteration
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "/root/git/pylint/pylint/utils/ast_walker.py", line 72, in walk
>     callback(astroid)
>   File "/root/git/pylint/pylint/extensions/docparams.py", line 298, in visit_raise
>     expected_excs = utils.possible_exc_types(node)
>   File "/root/git/pylint/pylint/extensions/_check_docs_utils.py", line 148, in possible_exc_types
>     excs = [exc for exc in inferred_excs if exc is not astroid.Uninferable]
>   File "/root/git/pylint/pylint/extensions/_check_docs_utils.py", line 148, in <listcomp>
>     excs = [exc for exc in inferred_excs if exc is not astroid.Uninferable]
>   File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/decorators.py", line 142, in raise_if_nothing_inferred
>     yield next(generator)
>   File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/nodes/node_classes.py", line 88, in unpack_infer
>     inferred = next(stmt.infer(context), util.Uninferable)
>   File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/nodes/node_ng.py", line 156, in infer
>     yield from self._infer(context, **kwargs)
>   File "/root/git/pylint/.venv/lib/python3.7/site-packages/astroid/decorators.py", line 150, in raise_if_nothing_inferred
>     ) from error
> astroid.exceptions.InferenceError: StopIteration raised without any error information.
> ```

I can't look into this completely (I'm on mobile) but this seems to be another issue of relying on inference without handling cases where inference fails.
To prevent this and fix this I think it is good to identify the code that is causing this (possibly with `breakpoint()` or the debugger of an IDE) and seeing why that code is throwing an `InferenceError`.

pylint/extensions/_check_docs_utils.py
kasium: I added a comment why this is needed. In fact `pytest tests/test_functional.py -k test_functional[missing_raises_doc]` will fail if the line is not there

kasium:
@DanielNoord I now was able to resolve the issues. Please note, that I reverted the functools commit, because it will not work with the current logic which bases on a for-else loop. Also, I added a comment why the weird `__class__` attribute access is needed. Please check one again.

Also please note, that bringing the CI back to green took some time

pylint/extensions/_check_docs_utils.py
DanielNoord: I have investigated this myself and the following diff would allow us to remove this `if` statement:
We can then revert most of these lines to the previous version and only add `.name`.

```diff
diff --git a/pylint/extensions/_check_docs_utils.py b/pylint/extensions/_check_docs_utils.py
index 1b644cd7..908f8b86 100644
--- a/pylint/extensions/_check_docs_utils.py
+++ b/pylint/extensions/_check_docs_utils.py
@@ -161,12 +161,11 @@ def possible_exc_types(node: nodes.NodeNG) -> Set[nodes.ClassDef]:
                     continue
 
                 val = utils.safe_infer(ret.value)
-                if (
-                    val
-                    and isinstance(val, (astroid.Instance, nodes.ClassDef))
-                    and utils.inherit_from_std_ex(val)
-                ):
-                    excs.append(val)
+                if val and utils.inherit_from_std_ex(val):
+                    if isinstance(val, nodes.ClassDef):
+                        excs.append(val)
+                    elif isinstance(val, astroid.Instance):
+                        excs.append(val.getattr("__class__")[0])
 
     final_exceptions = set()
     for exc in excs:
@@ -176,9 +175,6 @@ def possible_exc_types(node: nodes.NodeNG) -> Set[nodes.ClassDef]:
                 continue
         except astroid.InferenceError:
             continue
-        if not isinstance(exc, nodes.ClassDef):
-            # needed for instances of classes
-            exc = exc.getattr("__class__")[0]
         final_exceptions.add(exc)
     return final_exceptions
 ```

File: pylint/extensions/_check_docs_utils.py [hunk 17]
```diff
@@ -130,28 +130,30 @@ def possible_exc_types(node):
...
 
         if handler and handler.type:
-            inferred_excs = astroid.unpack_infer(handler.type)
-            excs = (exc.name for exc in inferred_excs if exc is not astroid.Uninferable)
+            try:
+                for exc in astroid.unpack_infer(handler.type):
+                    if exc is not astroid.Uninferable:
+                        excs.append(exc)
+            except astroid.InferenceError:
+                return set()
```
DanielNoord, line 152:
```suggestion
                pass
```

This passes `pytest -k missing_raises_doc`. I think it might be better.

DanielNoord (changes requested):
Thanks @kasium! This PR is probably becoming more work than you initially expected, but I think we're getting there.

Sorry for suggesting `itertools`. I misread the code and this does indeed not work. Sorry about the extra work I caused.

Pierre-Sassoulas (approved):
Thank you @kasium for handling this, taking the time to implements that properly was definitely required. It will be a nice addition to 2.13. And Thank you @DanielNoord for reviewing thoroughly :) !

pylint/extensions/_check_docs_utils.py
kasium: done

pylint/extensions/_check_docs_utils.py
```diff
@@ -130,46 +130,53 @@ def possible_exc_types(node):
...
+    final_exceptions = set()
+    for exc in excs:
+        try:
+            # This block can raise InferenceError because it is inferring the handler of the exception
+            if utils.node_ignores_exception(node, exc.name):
+                continue
+        except astroid.InferenceError:
+            continue
+        final_exceptions.add(exc)
+    return final_exceptions
```
DanielNoord, line 179:
```suggestion
     try:
         return {exc for exc in excs if not utils.node_ignores_exception(node, exc.name)}
     except astroid.InferenceError:
         return set()
```

Wouldn't this work? This would limit the diff to only the necessary change.

pylint/extensions/_check_docs_utils.py
```diff
@@ -130,46 +130,53 @@ def possible_exc_types(node):
...
-                    excs.append(val.name)
-
-    try:
-        return {exc for exc in excs if not utils.node_ignores_exception(node, exc)}
-    except astroid.InferenceError:
-        return set()
+                if val and utils.inherit_from_std_ex(val):
+                    if isinstance(val, nodes.ClassDef):
+                        excs.append(val)
+                    elif isinstance(val, astroid.Instance):
```
DanielNoord, line 167: Thanks 😄

DanielNoord (changes requested):
One final change to keep the `diff` a little bit more concise. The rest looks good to me. Thanks again @kasium!

DanielNoord (dismissed):

doc/whatsnew/2.12.rst
```diff
@@ -205,6 +205,20 @@ Other Changes
 
   Closes #5216
 
+* ``missing-raises-doc`` will now check the class hierarchy of the raised exceptions
```
DanielNoord, line 208: This should go in `2.13` 😅 

Sorry, I totally missed this previously..

kasium: I hope the just added commit is fine

ChangeLog
```diff
@@ -91,6 +91,19 @@ Release date: TBA
   Insert your changelog randomly, it will reduce merge conflicts
   (Ie. not necessarily at the end)
 
+* ``missing-raises-doc`` will now check the class hierarchy of the raised exceptions
```
DanielNoord, line 94: Could you move these to just above L90? I can't make a suggestion and commit it myself as it is outside of the scope of the change 😢

DanielNoord (approved):

DanielNoord:
Thanks @kasium :)

PR #5292, opened 2021-11-12
Add documentation for adding ``pylint`` to ``pre-commit``

### LINKED ISSUE #5291: Document setup with pre-commit
### Current problem

I can see the `.pre-commit-hooks.yaml` file in the repository, but I can't find any documentation on how to set it up.

### Desired solution

A short section on how to set up pre-commit like https://flake8.pycqa.org/en/latest/user/using-hooks.html

### Additional context

_No response_

DanielNoord:

## Type of Changes

## Description

**WIP**. 

`readthedocs` doesn't automatically run on forks, so using this PR to see how current iteration renders. I'll convert to finished when ready for review :)

Closes #5291

### DIFF
--- a/doc/user_guide/pre-commit-integration.rst
+++ b/doc/user_guide/pre-commit-integration.rst
@@ -1,0 +1,46 @@
+.. _pre-commit-integration:
+
+Pre-commit integration
+======================
+
+``pylint`` can be used as a `pre-commit <https://pre-commit.com>`_ hook.
+
+Since ``pylint`` needs to import modules and dependencies to work correctly, the
+hook only works with a local installation of ``pylint`` (in your environment).
+If you installed ``pylint`` locally it can be added to ``.pre-commit-config.yaml``
+as follows:
+
+.. sourcecode:: yaml
+
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        args:
+          [
+            "-rn", # Only display messages
+            "-sn", # Don't display the score
+          ]
+
+You can use ``args`` to pass command line arguments as described in the :ref:`tutorial`.
+A hook with more arguments could look something like this:
+
+.. sourcecode:: yaml
+
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        args:
+          [
+            "-rn", # Only display messages
+            "-sn", # Don't display the score
+            "--rcfile=pylintrc", # Link to your config file
+            "--load-plugins=pylint.extensions.docparams", # Load an extension
+          ]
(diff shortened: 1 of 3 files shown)


doc/user_guide/pre-commit-integration.rst
```diff
@@ -0,0 +1,41 @@
...
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        args:
+          [
+            "-rn", # Only display messages
+            "-sn", # Don't display the score
```
Pierre-Sassoulas, line 36: ``rn`` and ``sn`` are pretty much required, because pre-commit launch jobs in parallels so there will be dozen of score (or worst reports) if you don't. So I think we should put them in the simple example too.

Pierre-Sassoulas (approved):
Very nice !

adamchainz:
But what about the ``.pre-commit-hooks.yaml`` file in the repo? https://github.com/PyCQA/pylint/blob/main/.pre-commit-hooks.yaml

Users will be relying on that, and it remains undocumented.

adamchainz:
Also I think it's worth mentioning the tradeoffs. `system` hooks aren't so great with pre-commit as they don't take advantage of its repeatable environments feature. The repo-based hook is easy but it doesn't allow pylint to inspect your depedencies, unless you add them into the pre-commit environment.

Pierre-Sassoulas:
> But what about the ``.pre-commit-hooks.yaml`` file in the repo?
> Users will be relying on that, and it remains undocumented.

I think we need to remove this file as it's misleading. There are too much false positives without the dependencies installed in almost all cases. There's probably a way to make pylint or pre-commit aware of the dependencies it needs to install in order to run correctly, so it's possible to install it in its own venv but this need some reflexion and design.

DanielNoord:
> > But what about the `.pre-commit-hooks.yaml` file in the repo?
> > Users will be relying on that, and it remains undocumented.
> 
> I think we need to remove this file as it's misleading. There are too much false positives without the dependencies installed in almost all cases. There's probably a way to make pylint or pre-commit aware of the dependencies it needs to install in order to run correctly, so it's possible to install it in its own venv but this need some reflexion and design.

@Pierre-Sassoulas Do you still want to remove this? I still have the notification about this as "unread" as I wanted to follow-up to this but never got to it 😄

Pierre-Sassoulas:
I forgot about that :) yes at least temporarily.

DanielNoord:
Will you make the PR?

Pierre-Sassoulas:
> Will you make the PR?

Sure, I'm on mobile so I can't right now; but later.

Pierre-Sassoulas:
Neverming after some consideration the fix is supposed to be a better documentation, see https://github.com/PyCQA/pylint/issues/5293#issuecomment-995587789

PR #5294, opened 2021-11-12
Make y/n validator case insensitive

tusharsadhwani:
## Type of Changes

## Description

I found many people misconfiguring their pylintrc by setting some boolean config options to `True` or `False`, when the correct option would've been `yes` or `no`. This would cause people's CI to crash when trying to run pylint.

This was even weirder when I found out that using the boolean `true` works fine when using a `pyproject.toml` file instead. So I believe `True` should be supported in pylintrc as well. It will also remove unnecessary CI failures.

To achieve this, this patch simply makes the y/n validator case insensitive. Does this sound like a reasonable fix?


Pierre-Sassoulas (changes requested):
This is a welcome change now that we're so far gone in the rabbit hole that soon we'll have to also handle ideogram in 5 languages for yes and no. Could you simply add a changelog for this, please ?

If we make a breaking change for option/conf parsing in the future I will definitely make sure that there is a single possible value for True and False...

tusharsadhwani:
:+1:

Pierre-Sassoulas (approved):
Thanks a lot 👍

Pierre-Sassoulas:
This will be available in 2.12.0 that will release "soon ™️ " ;)

PR #5296, opened 2021-11-12
Fix ``accept-no-yields/return-doc`` for partially correct docstrings

### LINKED ISSUE #5223: pylint.extensions.docparams default options are not working correctly
### Bug description

The [documentation for pylint.extensions.docparams](https://pylint.pycqa.org/en/latest/technical_reference/extensions.html#parameter-documentation-checker-options) describes that by default, `accept-no-yields-doc` and `accept-no-return-doc` are **yes** (along with other configuration options).

However, if I run `pylint myfile.py` with no pylint configuration except for enabling the plugin*, I see lots of entries of:

```
... W9011: Missing return documentation (missing-return-doc)
... W9013: Missing yield documentation (missing-yield-doc)
```

\* my only configuration is:
```toml
[tool.pylint.master]
load-plugins = ["pylint.extensions.docparams"]
```

In all situations, I am deliberately excluding :yields: and :rtype: from the docstring because I am using type annotations.

Am I missing something or is the documented behaviour wrong?

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.4
Python 3.9.5 (default, May  8 2021, 00:04:35) 
[GCC 8.3.0]
```

### OS / Environment
Debian Buster

DanielNoord:

## Type of Changes

## Description

As discussed in the issue it seems that our tests have solidified a bug where we do not break immediately when either of these settings are set to `yes`/`True`. This fixes this and turns off the setting for those tests that relied on the incorrect previous behaviour.

Closes #5223

Edit: I guess I shouldn't be coding this late 😅 Third try is a charm..

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -19,6 +19,11 @@ Release date: TBA
 
 * Fix exception when pyreverse parses ``property function`` of a class.
 
+* Fix ``accept-no-yields-doc`` and ``accept-no-return-doc`` not allowing missing ``yield`` or
+  ``return`` documentation when a docstring is partially correct
+
+  Closes #5223
+
 * Add an optional extension ``consider-using-any-or-all`` : Emitted when a ``for`` loop only
   produces a boolean and could be replaced by ``any`` or ``all`` using a generator. Also suggests
   a suitable any or all statement.
--- a/doc/whatsnew/2.12.rst
+++ b/doc/whatsnew/2.12.rst
@@ -77,6 +77,11 @@ Other Changes
 
   Closes #5178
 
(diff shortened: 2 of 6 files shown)


Pierre-Sassoulas (approved):
LGTM thank you. Very elegant way to solve the issue with legacy tests. Shouldn't we have at least one test for ``accept_no_yields_doc="yes"`` ? Or as it's the default value it's already "all the other tests" ?

DanielNoord:
> LGTM thank you. Very elegant way to solve the issue with legacy tests. Shouldn't we have at least one test for `accept_no_yields_doc="yes"` ? Or as it's the default value it's already "all the other tests" ?

I think "all the other tests" works here? I think you can rebase and merge if you agree :)

PR #5298, opened 2021-11-13
Move ``misplaced-comparison-constant`` to optional extension

### LINKED ISSUE #1064: Disable 'misplaced-comparison-constant' checker by default by moving it to extensions
### Steps to reproduce
1.  Write a comparison such as `if 5 < a:...`
### Current behavior

Pylint complains with 'misplaced-comparison-constant'.
### Expected behavior

No issue.

In other languages, so-called ["Yoda expressions"](https://en.wikipedia.org/wiki/Yoda_conditions) can be used as a code guard to prevent an accident assignment when an equality comparison was indented. (i.e. `if (42 = x) { /* in C, this causes a compile time error error */ }` In Python, we cannot use assignments when a boolean expression is expected, so let's not dwell on this use case any further.

However, that is not my reason for using this style of comparison. I prefer to avoid using the greater-than symbol (`>`). Instead, I'll always use the less-than (`<`) or the less-than-equal (`<=`). I'll yield to Dijkstra's explanation for the best way to describe ranges in his paper ['Why numbering should start at zero'](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html).

Given we are talking Python, that paper is, actually, very interesting because Python is one of the very few languages that the range expression is a valid boolean expression; i.e. `2 <= x < 13`. So, I'll cynically ask, "Is that half-good, half-bad?"

A quick test against pylint tells me that the range expression is handled appropriately.

So, why do we even care?

Readability!? If so, that seems inane to me. Force the nitpickers to enable `misplaced-comparison-constant` rather than ding those that aren't so easily confused.

Other rationale!? I can't think of some use case that may invoke some rare peculiarity as one may find with overloaded operators in C++. That is to say, Python will use reflection to handle equality of unlike objects; assuming the objects make use of `NotImplemented`.

Example:

``` python
"""This example is intentionally absurd."""

class MyClass(object):
    def __eq__(self, other):
        if isintance(other, int):
            return other == 42
        return NotImplemented

class MyOtherClass(object):
    def __eq__(self, other):
        return isinstance(other, MyClass)

MyClass() == 42  # True
42 == MyClass()  # True

MyOtherClass() == MyClass()  # True
MyClass() == MyOtherClass()  # True
```
### pylint --version output

```
$ pylint --version
No config file found, using default configuration
pylint 1.6.4,
astroid 1.4.7
Python 2.7.11 (default, Mar 15 2016, 09:47:58)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)]
```

DanielNoord:

## Type of Changes

## Description

Closes #1064

Consensus was reached that this check should be an extension. I thought about putting it in `code_style`, but since that only has `Recommendation` and this is a `Convention` I thought I might as well create a new extension.
Since the extensions loaded in our `pylintrc` are not automatically loaded in the functional tests this required to update some tests and their results.

### DIFF
--- a/pylint/extensions/comparison_placement.py
+++ b/pylint/extensions/comparison_placement.py
@@ -1,7 +1,10 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
-"""Checks the placement of comparisons"""
+"""
+Checks for yoda comparisons (variable before constant)
+See https://en.wikipedia.org/wiki/Yoda_conditions
+"""
 
 
 from astroid import nodes
@@ -13,7 +16,7 @@ REVERSED_COMPS = {"<": ">", "<=": ">=", ">": "<", ">=": "<="}
 COMPARISON_OPERATORS = frozenset(("==", "!=", "<", ">", "<=", ">="))
 
 
-class ComparisonPlacementChecker(BaseChecker):
+class MisplacedComparisonConstantChecker(BaseChecker):
     """Checks the placement of constants in comparisons"""
 
     __implements__ = (IAstroidChecker,)
--- a/pylint/extensions/comparison_placement.py
+++ b/pylint/extensions/comparison_placement.py
@@ -64,4 +64,4 @@ class MisplacedComparisonConstantChecker(BaseChecker):
 
 def register(linter):
     """Required method to auto register this checker."""
-    linter.register_checker(ComparisonPlacementChecker(linter))
+    linter.register_checker(MisplacedComparisonConstantChecker(linter))
--- a/pylint/extensions/comparison_placement.py
+++ b/pylint/extensions/comparison_placement.py
@@ -1,0 +1,64 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+"""Checks the placement of comparisons"""
+
+
+from astroid import nodes
+
+from pylint.checkers import BaseChecker, utils
+from pylint.interfaces import IAstroidChecker
+
+REVERSED_COMPS = {"<": ">", "<=": ">=", ">": "<", ">=": "<="}
+COMPARISON_OPERATORS = frozenset(("==", "!=", "<", ">", "<=", ">="))
+
+
+class ComparisonPlacementChecker(BaseChecker):
+    """Checks the placement of constants in comparisons"""
+
+    __implements__ = (IAstroidChecker,)
+
+    # configuration section name
+    name = "comparison-placement"
+    msgs = {
+        "C2201": (
+            "Comparison should be %s",
+            "misplaced-comparison-constant",
+            "Used when the constant is placed on the left side "
+            "of a comparison. It is usually clearer in intent to "
+            "place it in the right hand side of the comparison.",
+            {"old_names": [("C0122", "old-misplaced-comparison-constant")]},
+        )
+    }
+
+    options = ()
+
+    def _check_misplaced_constant(
+        self,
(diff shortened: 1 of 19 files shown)


DanielNoord:
The messages that got removed from `CompareToZeroUsedTC` are covered by the functional test. I don't think we need to add a unittest for this extension at this moment.

pylint/extensions/comparison_placement.py
```diff
@@ -0,0 +1,64 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+"""Checks the placement of comparisons"""
```
Pierre-Sassoulas, line 4:
```suggestion
"""
Checks for yoda comparisons (variable before constant)
See https://en.wikipedia.org/wiki/Yoda_conditions
"""
```
I think it's ok to be explicit and use colloquialism

pylint/extensions/comparison_placement.py
```diff
@@ -0,0 +1,64 @@
...
+from astroid import nodes
+
+from pylint.checkers import BaseChecker, utils
+from pylint.interfaces import IAstroidChecker
+
+REVERSED_COMPS = {"<": ">", "<=": ">=", ">": "<", ">=": "<="}
+COMPARISON_OPERATORS = frozenset(("==", "!=", "<", ">", "<=", ">="))
+
+
+class ComparisonPlacementChecker(BaseChecker):
```
Pierre-Sassoulas, line 16: Is this a checker that is supposed to stay with one message forever, we could use the same name than the message if that's the case ?
```suggestion
class MisplacedComparisonConstantChecker(BaseChecker):
```

pylint/extensions/comparison_placement.py
```diff
@@ -0,0 +1,64 @@
...
+    """Checks the placement of constants in comparisons"""
```
Pierre-Sassoulas, line 17: It feels like it's duplicated from the module docstring, we might as well remove one of them.

Pierre-Sassoulas (commented):
Some suggestions otherwise  LGTM :) Thank you for handling the work for a decision taken a long time ago.

pylint/extensions/comparison_placement.py
DanielNoord: We would miss out on IntelliSense on the class then. I also saw that we are waiting to add `extensions.docparams` in our `pylintrc` which would requires us to re-add this docstring in the future.

DanielNoord:
Updated with additional tests to make sure coverage doesn't decrease. Ready for review again!

pylint/extensions/comparison_placement.py
Pierre-Sassoulas: Fair enough.

tests/functional/m/misplaced_comparison_constant.py
```diff
@@ -34,3 +34,14 @@ def good_comparison():
     for i in range(10):
         if i == 5:
             pass
+
+def double_comparison():
+    """Check that we return early for non-binary comparison"""
+    for i in range(10):
+        if i == 1 == 2:
+            pass
```
Pierre-Sassoulas, line 42:
```suggestion
        if i == 1 == 2:
            pass
        if 2 <= i <= 8:
            print("Between 2 and 8 inclusive")
```
I think adding the standard use case could be safe as the other one is kinda theoretical.

Pierre-Sassoulas (approved):
Little nitpick to add the standard test case for double comparison in the functional tests, it's still good to go.

DanielNoord:
@Pierre-Sassoulas Turns out the CI on `main` has been failing since this PR. Not sure if you noticed, just thought I would let you know.

Pierre-Sassoulas:
Ha it look like it's because of my experiment with Github pages. Now I don't even know how to remove it :(

PR #5303, opened 2021-11-14
Fix typo in pylintrc for useless-suppression

### LINKED ISSUE #5040: Typo in top level `pylintrc` and possible bug in `missing-docstring` or `useless-suppression`
### Bug description

The top level `pylintrc` for pylint itself has a typo:

```diff
@@ -54 +54 @@ enable=
-    useless-supression,
+    useless-suppression,
```

Fixing that typo causes a test suite failure, because of one new `useless-suppression` message:
```
 ______________ TestRunTC.test_json_report_does_not_escape_quotes _______________

self = <test_self.TestRunTC object at 0x7f13cd73b5f8>

    def test_json_report_does_not_escape_quotes(self) -> None:
        out = StringIO()
        module = join(HERE, "regrtest_data", "unused_variable.py")
        self._runtest([module], code=4, reporter=JSONReporter(out))
        output = json.loads(out.getvalue())
        assert isinstance(output, list)
>       assert len(output) == 1
E       assert 2 == 1
E        +  where 2 = len([{'column': 4, 'line': 4, 'message': "Unused variable 'variable'", 'message-id': 'W0612', ...}, {'column': 0, 'line': 1, 'message': "Useless suppression of 'missing-class-docstring'", 'message-id': 'I0021', ...}])
```

I'm reporting this here instead of filing a PR, in part because I think it's faster for someone who knows the preferred way to deal with this to just fix it instead of going back and forth with a PR..

But more importantly, the useless suppression message seems to indicate a pylint bug or at least behavior which is weird to me:

`./tests/regrtest_data/unused_variable.py` doesn't even contain a suppression for `missing-class-docstring` which is said to be useless. It just has [one for `missing-docstring`](https://github.com/PyCQA/pylint/blob/e8d112132dbd88e91257cdb8380ddc4e1561e305/tests/regrtest_data/unused_variable.py). I wondered if this could be if the latter was a superset of the former or something, but then again even then the suppression would not seem useless because there just _isn't_ a class docstring in the said file at all, so the warning about the suppression would seem spot on.
### TODO

- [ ] Fix the typo and enable useless suppression
- [ ] Handle the "false positive" / working as intended for useless suppression of missing docstring

### Pylint version

```shell
Current git main
```

DanielNoord:

## Type of Changes

## Description

Closes #5040

See discussion there.


Pierre-Sassoulas (approved):
:thumbsup:

DanielNoord:
@Pierre-Sassoulas Isn't this supposed to show some useless suppressions in our current code? I remember I saw some errors previously. Did we fix those in another pr?

Pierre-Sassoulas:
>  Did we fix those in another pr?

Hmm, I don't think so, there must be something wrong with the way we launch pylint in github actions because:

```bash
pylint pylint/ --enable=all|grep useless
```
```
pylint/utils/utils.py:96:0: I0021: Useless suppression of 'redefined-builtin' (useless-suppression)
pylint/utils/file_state.py:77:0: I0021: Useless suppression of 'not-callable' (useless-suppression)
pylint/utils/file_state.py:79:0: I0021: Useless suppression of 'no-member' (useless-suppression)
pylint/lint/pylinter.py:621:0: I0021: Useless suppression of 'consider-using-with' (useless-suppression)
pylint/lint/pylinter.py:1064:0: I0021: Useless suppression of 'no-member' (useless-suppression)
pylint/lint/run.py:24:0: I0021: Useless suppression of 'not-callable' (useless-suppression)
pylint/config/option.py:184:0: I0021: Useless suppression of 'unsupported-assignment-operation' (useless-suppression)
pylint/reporters/text.py:296:0: I0021: Useless suppression of 'import-error' (useless-suppression)
pylint/checkers/utils.py:1346:0: I0021: Useless suppression of 'redundant-keyword-arg' (useless-suppression)
pylint/checkers/similar.py:30:0: I0021: Useless suppression of 'redefined-builtin' (useless-suppression)
pylint/checkers/format.py:58:0: I0021: Useless suppression of 'redefined-builtin' (useless-suppression)
pylint/checkers/variables.py:1788:0: I0021: Useless suppression of 'too-many-branches' (useless-suppression)
pylint/testutils/reporter_for_tests.py:53:0: I0021: Useless suppression of 'unused-argument' (useless-suppression)
pylint/testutils/unittest_linter.py:17:0: I0021: Useless suppression of 'no-self-use' (useless-suppression)
pylint/testutils/lint_module_test.py:149:0: I0021: Useless suppression of 'consider-using-with' (useless-suppression)
pylint/testutils/lint_module_test.py:156:0: I0021: Useless suppression of 'consider-using-with' (useless-suppression)
```

cdce8p:
> Hmm, I don't think so, there must be something wrong with the way we launch pylint in github actions

Try adding the `-v` flag to the pre-commit pylint run. Normally the log is only shown if the script exits with `!= 0`.
https://pre-commit.com/#pre-commit-run

Pierre-Sassoulas:
Ho is this the issue where ``I`` type message exit with zero even when there are some ? We could change the pylint option when we launch, but this feel wrong, if some message are emitted we should exit with an error code. I can't find the issue about it but I read about that problem somewhere.

cdce8p:
> Ho is this the issue where `I` type message exit with zero even when there are some ? We could change the pylint option when we launch, but this feel wrong, if some message are emitted we should exit with an error code. I can't find the issue about it but I read about that problem somewhere.

We discussed it here: https://github.com/PyCQA/astroid/pull/1217#discussion_r744257353. As I pointed out then we should **not** emit a non-zero exit code for `useless-suppression` messages. At least not without some more investigations.

> The issue is that some errors might be Python version specific. It could create a situation where only one, `pre-commit` or `CI`, will succeed when tested with different versions. The `py-version` option should help, but I don't know if it's enough.
> 
> Additionally, this would prevent us from having the latest `pylint` version installed in a local env as this is likely to have fixed some false-positives. -> Which would then trigger `useless-suppression`.

DanielNoord:
I added a fix with allows failing on `Information` messages.

It requires changing how `fail_under` works. Currently we can get a lower score than 10 by adding `Info` messages to the evaluation, but since that exit code is 0 we won't actually produce a non-zero exit.
A couple of lines above we fixed it likes this, so I imagine this might work here as well?

@cdce8p Our CI takes into account our `pylintrc` which has a `py-version` (3.6). So that shouldn't create problems (I think?). I think the problem of local installations of `pylint` also doesn't affect us here, since `pre-commit` for `pylint` should always run with the locally installed `pylint` anyway? 

This requires careful thought, but I think we might be fine by adding this. If agreed I will fix the outstanding useless-suppression warnings as well so this can go in `2.13`.

Pierre-Sassoulas:
> only one, pre-commit or CI, will succeed when tested with different versions

Isn't this is a good thing and a way to find problem with ``py-version`` ? If we actually check that there is no useless-suppression for any interpreter, then we can cheaply detect inconsistencies ?

cdce8p:
> @cdce8p Our CI takes into account our `pylintrc` which has a `py-version` (3.6). So that shouldn't create problems (I think?). I think the problem of local installations of `pylint` also doesn't affect us here, since `pre-commit` for `pylint` should always run with the locally installed `pylint` anyway?

> Isn't this is a good thing and a way to find problem with `py-version` ? If we actually check that there is no useless-suppression for any interpreter, then we can cheaply detect inconsistencies ?

I guess we'll find out if it works. It was one of the reasons to add such a setting in the first place. Might be that my reservations are unfound.

.pre-commit-config.yaml
```diff
@@ -57,6 +57,9 @@ repos:
             "-rn",
             "-sn",
             "--rcfile=pylintrc",
+            "--fail-under=10",
+            "--evaluation='10.0 - ((float(5 * error + warning + refactor + convention +
+            info) / statement) * 10)'",
```
cdce8p, line 62: Why not use `--fail-on=useless-suppression` instead?
Maybe `--fail-on=I` works 🤷🏻‍♂️ haven't tested it.

DanielNoord: `--fail-on` works, forgot that was a thing!

pylint/lint/run.py
```diff
@@ -402,8 +402,13 @@ def __init__(
...
                 sys.exit(self.linter.msg_status or 1)
-            elif score_value is not None and score_value >= linter.config.fail_under:
-                sys.exit(0)
+            elif score_value is not None:
+                if score_value >= linter.config.fail_under:
+                    sys.exit(0)
+                else:
+                    # We need to make sure we return a failing exit code in this case.
+                    # So we use self.linter.msg_status if that is non-zero, otherwise we just return 1.
+                    sys.exit(self.linter.msg_status or 1)
```
DanielNoord, line 411: This should probably go in another PR. I'll create one.

pylint/utils/file_state.py
```diff
@@ -74,9 +74,9 @@ def _collect_block_lines(
         #
         # 1.   def meth8(self):
         # 2.        """test late disabling"""
-        # 3.        pylint: disable=not-callable
+        # 3.        pylint: disable=not-callable, useless-suppression
```
DanielNoord, line 77: I think this is the most elegant way to allow us to still fail on `I`.

DanielNoord:
This does indeed create issues 😢 

https://github.com/PyCQA/pylint/blob/8a653aab1580d65f30d3575c7a482665598f5016/pylint/lint/run.py#L24-L29

On `3.10` `using-constant-test` is useless but on `3.6` (which the `pre-commit` CI uses) it is not. This begs the question whether we should add `useless-suppression` to our `pylintrc` at all. In this case it will always create a warning on either `3.6` or `3.10`. Although I expect most contributors to be on `>3.8` we can't really say "pylint supports 3.6", but "pylint development should be done on 3.8+"...

Pierre-Sassoulas:
>  "pylint development should be done on 3.8+".

It's kinda true though, as the AST changed in python 3.8, if you need to work on proper column index for messages you need 3.8+ (functional tests will help you but you will need to check the expected column by running github actions and you won't be able to update the column automatically). So maybe we could upgrade the interpreter we use for pre-commit to 3.8 and upgrade the contributor documentation accordingly.

cdce8p:
> So maybe we could upgrade the interpreter we use for pre-commit to 3.8 and upgrade the contributor documentation accordingly.

Up until just recently my goto recommendation would have been to use the oldest supported Python version at least for `mypy` and `pylint`. It might work with 3.8, especially considering our extensive test suite. However there is always a small risk that something goes undetected. I'm thinking especially about `no-member` errors. I.e. we try to add functions that aren't available in all supported versions -> `cached_property` 😉

DanielNoord:
> I'm thinking especially about `no-member` errors. I.e. we try to add functions that aren't available in all supported versions -> `cached_property` 😉

Shouldn't those still fail in the regular test suite?

To make sure this PR doesn't stay open indefinitely I propose to:
1) Update our ``pre-commit`` CI to run on `3.8`
2) Add a changelog entry and doc entry that it is recommended to work on ``pylint`` on 3.8+

I think because of the aforementioned change to the `ast` parser that makes sense. The entry and this change can then go in `2.12` as it is the easiest way to communicate this to potential developers. (If we put it in `2.13` we should officially wait till its release before we can change the CI environment).

Pierre-Sassoulas:
> Shouldn't those still fail in the regular test suite?

Yes, we still run all tests on 3.6 / 3.7 we'd catch the real error. Whatever we do automated tests are still better than static analysis 😄 

>    Update our pre-commit CI to run on 3.8
>    Add a changelog entry and doc entry that it is recommended to work on pylint on 3.8+

👍

pylint/lint/run.py
```diff
@@ -21,7 +21,7 @@
 def _cpu_count() -> int:
     """Use sched_affinity if available for virtualized or containerized environments."""
     sched_getaffinity = getattr(os, "sched_getaffinity", None)
-    # pylint: disable=not-callable,using-constant-test
+    # pylint: disable=not-callable,using-constant-test,useless-suppression
```
DanielNoord, line 24: This is unsolvable atm. The `pre-commit` env does have `sched_getaffinity` so this will always create some sort of error.

doc/whatsnew/2.12.rst
```diff
@@ -195,6 +195,11 @@ Other Changes
 * Added the ``--enable-all-extensions`` command line option. It will load all available extensions
   which can be listed by running ``--list-extensions``
 
+* It is now recommended to do ``pylint`` development on ``Python`` 3.8 or higher. This
+  allows using the latest ``ast`` parser.
+
+* All standard jobs in the ``pylint`` CI now run on ``Python`` 3.8 by default
```
Pierre-Sassoulas, line 201:
```suggestion
* All standard jobs in the ``pylint`` CI now run on ``Python`` 3.8 by default. We still
  support python 3.6 and 3.7 and run tests for those interpreters.
```

pylint/utils/utils.py
```diff
@@ -93,7 +93,7 @@ def normalize_text(text, line_len=DEFAULT_LINE_LENGTH, indent=""):
 
 
 # py3k has no more cmp builtin
-def cmp(a, b):  # pylint: disable=redefined-builtin
```
Pierre-Sassoulas, line 96: This was useless since pylint 2.0 😄

ChangeLog
```diff
@@ -275,6 +275,11 @@ Release date: TBA
   Closes #3507
   Closes #5087
 
+* It is now recommended to do ``pylint`` development on ``Python`` 3.8 or higher. This
+  allows using the latest ``ast`` parser.
+
+* All standard jobs in the ``pylint`` CI now run on ``Python`` 3.8 by default
```
Pierre-Sassoulas, line 281:
```suggestion
* All standard jobs in the ``pylint`` CI now run on ``Python`` 3.8 by default. We still
  support python 3.6 and 3.7 and run tests for those interpreters.
```

Pierre-Sassoulas (approved):
LGTM !

PR #5304, opened 2021-11-14
Add broken NoReturn check

cdce8p:
## Type of Changes

## Description

Response to: https://github.com/PyCQA/astroid/issues/1239

Add a new check to detect broken uses for `typing.NoReturn` in `3.7.0` / `3.7.1`.
I'm open to changing the checker name if something better than `broken-noreturn` is found :)


Pierre-Sassoulas:
Shouldn't this be in [flake8-typing-imports](https://github.com/asottile/flake8-typing-imports) instead ? We were going to add it to our own pre-commit configuration in #5070 (and astroid too)

cdce8p:
> Shouldn't this be in [flake8-typing-imports](https://github.com/asottile/flake8-typing-imports) instead ? We were going to add it to our own pre-commit configuration in #5070 (and astroid too)

Not sure Anthony will add it ;)
Technically, `NoReturn` can be imported and even used, it's just broken in these special cases. So I'm not sure what `flake8-typing-import` should emit. Importing it from `typing_extensions` doesn't fix this issue either.

Pierre-Sassoulas:
> Technically, NoReturn can be imported and even used, it's just broken in these special cases.

Ok, that makes sense. Trying not to add anything in 2.12 so we can release though :)

> Not sure Anthony will add it ;)

Yeah, it's often a struggle.

Pierre-Sassoulas:
> I'm open to changing the checker name if something better than broken-noreturn is found :)

``invalid-no-return-as-typing-argument`` ? It's a little lengthy.

cdce8p:
> > I'm open to changing the checker name if something better than broken-noreturn is found :)
> 
> `invalid-no-return-as-typing-argument` ? It's a little lengthy.

Doesn't really fit IMO. `NoReturn` isn't invalid itself. It's just the specific usage that breaks.

Pierre-Sassoulas:
> It's just the specific usage that breaks.

I thought this usage where it's not valid was when it's used as a typing argument ?

cdce8p:
You can still use it like this
```py
def func() -> NoReturn:
    ...

Alias = Union[None, 'NoReturn']  # note the use of ' '
```
at least if you don't need to call `typing.get_type_annotations`.

cdce8p:
> > It's just the specific usage that breaks.
> 
> I thought this usage where it's not valid was when it's used as a typing argument ?

The uses are valid. It just breaks because of https://bugs.python.org/issue34921

DanielNoord:
@cdce8p Is this ready for review (barring the merge conflicts)?

cdce8p:
> @cdce8p Is this ready for review (barring the merge conflicts)?

Yes. I just resolved the merged conflicts.

tests/functional/ext/typing/typing_broken_noreturn.py
```diff
@@ -0,0 +1,38 @@
...
+If no runtime introspection is required, use string annotations instead.
+"""
+# pylint: disable=missing-docstring
+import sys
+import typing
+from typing import Callable, Union
+
+import typing_extensions
+
+if sys.version_info >= (3, 6, 2):
```
DanielNoord, line 14: Is this still necessary for our tests? According to the `.rc` file we don't even run on `3.6`?

DanielNoord (approved):
One little question. Rest looks good!

tests/functional/ext/typing/typing_broken_noreturn.py
Pierre-Sassoulas: We're incompatible with python 3.6.2 since pylint 2.12.1 (https://github.com/PyCQA/pylint/pull/5068) so I think we can remove all those chekck : >= 3.6.2 it's always true.

Pierre-Sassoulas: My bad I did not see it was in a test, my comment is irrelevant.

cdce8p: > My bad I did not see it was in a test, my comment is irrelevant.

However even the tests require `3.6.2`. So I think the initial comment is still valid.

tests/functional/ext/typing/typing_broken_noreturn.rc
```diff
@@ -0,0 +1,6 @@
+[master]
+py-version=3.7
+load-plugins=pylint.extensions.typing
+
+[testoptions]
+min_pyver=3.7
```
cdce8p, line 6: I should also change that to `3.6`

Pierre-Sassoulas:
```suggestion
min_pyver=3.6
```

Pierre-Sassoulas (approved):
LGTM !

pylint/extensions/typing.py
```diff
@@ -103,6 +109,14 @@ class TypingChecker(BaseChecker):
             "Emitted when 'typing.Union' or 'typing.Optional' is used "
             "instead of the alternative Union syntax 'int | None'.",
         ),
+        "E6004": (
+            "'NoReturn' inside compound types is broken in 3.7.0 / 3.7.1",
+            "broken-noreturn",
+            "``typing.NoReturn`` inside compound types is broken in "
+            "Python 3.7.0 and 3.7.1. If not dependend on runtime introspection, "
```
DanielNoord, line 116:
```suggestion
            "Python 3.7.0 and 3.7.1. If not dependent on runtime introspection, "
```

pylint/extensions/typing.py
```diff
@@ -279,6 +301,27 @@ def leave_module(self, node: nodes.Module) -> None:
...
+                isinstance(inferred, (nodes.FunctionDef, nodes.ClassDef))
+                and inferred.qname() in TYPING_NORETURN
+                # In Python 3.6, NoReturn is alias of '_NoReturn'
+                # In Python 3.7 - 3.8, NoReturn is alias of '_SpecialForm'
+                or isinstance(inferred, astroid.bases.BaseInstance)
+                and isinstance(inferred._proxied, nodes.ClassDef)
+                and inferred._proxied.qname()
+                in {"typing._NoReturn", "typing._SpecialForm"}
+            ):
+                self.add_message("broken-noreturn", node=node)
```
DanielNoord, line 322:
```suggestion
                self.add_message("broken-noreturn", node=node, confidence=INFERENCE)
```

DanielNoord: I can't comment to add the import as well.

DanielNoord (changes requested):

DanielNoord (approved):

cdce8p:
Need to do one more change. So far I've ignored `from __future__ import annotations`.

tests/functional/ext/typing/typing_broken_noreturn_future_import.rc
```diff
@@ -0,0 +1,3 @@
+[master]
+py-version=3.7
+load-plugins=pylint.extensions.typing
```
cdce8p, line 3:
```suggestion
load-plugins=pylint.extensions.typing

[testoptions]
min_pyver=3.7
```

PR #5310, opened 2021-11-15
Add ``future=True`` to all ``NodeNG.statement()`` calls

DanielNoord:

## Type of Changes

## Description

Ref https://github.com/PyCQA/astroid/pull/1235


pylint/checkers/variables.py
```diff
@@ -1392,8 +1392,8 @@ def _in_lambda_or_comprehension_body(
     def _is_variable_violation(
```
Pierre-Sassoulas, line 1392: I think the submodule were added by mistake, but I wondered if we should add them for real. It's probably a waste of time for runner that do not launch the primer.

DanielNoord: Mistake indeed. Not sure why that happened. Don't think we should add them, at least not in this PR.

Pierre-Sassoulas: The proper `.gitignore` for `.pylint_primer_test` is only in the primer test branch right now, so after getting them a `git add .` in another branch will stage them.

DanielNoord:
Opening and closing to run `primer`.

@Pierre-Sassoulas I put this in `2.12.2` but since we don't actually change any code here (only change `AttributeError` to `StatementMissing` + one new `and node_scope.parent`) this could also go in `2.12` without much risk of breaking things.
It would stop the warnings from displaying in our tests.

Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
primers bringing value again 😄

Pierre-Sassoulas:
Moving to 2.13 because of the test fail (could be 2.12.x too).

DanielNoord:
Looking into crash as we speak!

DanielNoord:
Blocked by #5382.

pylint/checkers/variables.py
```diff
@@ -1890,29 +1889,36 @@ def _loopvar_name(self, node: astroid.Name) -> None:
...
-            and astmts[0].statement().parent_of(node)
+            or (
+                astmts[0].parent == astmts[0].root()
+                and astmts[0].parent.parent_of(node)
+            )
+            or (
+                astmts[0].is_statement
+                or not isinstance(astmts[0].parent, nodes.Module)
+                and astmts[0].statement(future=True).parent_of(node)
+            )
```
DanielNoord, line 1913: This was particularly trigger to get right, but this is testing what we want. If `parent` is the root `Module` and the parent of `node` we don't do `_astmts = astmts[:1]`. The other `or` is simply what we did previously.

tests/functional/r/regression_02/regression_node_statement.py
```diff
@@ -0,0 +1,18 @@
...
+    from enum import Enum
+
+    class ellipsis(Enum):
+        Ellipsis = "..."
+
+    Ellipsis = ellipsis.Ellipsis
+
+
+else:
+    ellipsis = type(Ellipsis)
```
DanielNoord, line 18: This is a regression test for code I found while working on this.

pylint/checkers/format.py
```diff
@@ -602,8 +602,10 @@ def visit_default(self, node: nodes.NodeNG) -> None:
             isinstance(node.parent, nodes.TryFinally) and node in node.parent.finalbody
         ):
             prev_line = node.parent.body[0].tolineno + 1
+        elif isinstance(node.parent, nodes.Module):
+            prev_line = 0
```
DanielNoord, line 606: This is the same effect as doing `node.parent.fromlineno` but avoids the `StatementMissing` exception from calling `statement`.

DanielNoord:
@cdce8p This is finally ready for review 🎉

pylint/checkers/variables.py
```diff
@@ -1890,29 +1889,36 @@ def _loopvar_name(self, node: astroid.Name) -> None:
...
-        # filter variables according their respective scope test is_statement
-        # and parent to avoid #74747. This is not a total fix, which would
+        # Filter variables according to their respective scope. Test parent
+        # and statement to avoid #74747. This is not a total fix, which would
         # introduce a mechanism similar to special attribute lookup in
         # modules. Also, in order to get correct inference in this case, the
         # scope lookup rules would need to be changed to return the initial
         # assignment (which does not exist in code per se) as well as any later
         # modifications.
+        # pylint: disable-next=too-many-boolean-expressions
```
Pierre-Sassoulas, line 1902:
```suggestion
        # pylint: disable-next=too-many-boolean-expressions
        parent_is_root_module = astmts[0].parent == astmts[0].root()
```
```suggestion
        # pylint: disable-next=too-many-boolean-expressions
        #  If parent is the root Module and the parent of node we don't do 
        # _astmts = astmts[:1].
```

tests/functional/r/regression_02/regression_node_statement.py
```diff
@@ -0,0 +1,18 @@
+"""Test to see we don't crash on this code in pandas.
+See: https://github.com/pandas-dev/pandas/blob/master/pandas/core/arrays/sparse/array.py
+Code written van G van Rossum here: https://github.com/python/typing/issues/684"""
```
Pierre-Sassoulas, line 3:
```suggestion
Code written by Guido van Rossum here: https://github.com/python/typing/issues/684"""
```

Interesting read!

Pierre-Sassoulas (changes requested):
LGTM, just some small suggestions.

pylint/checkers/variables.py
DanielNoord: First suggestion: Perhaps Github is not working correctly but we can't access `astmts[0]` on L1903. We need the `not astmts` check on L1904 before it.

Second suggestion: Is this necessary? That's basically explaining what the if statement is telling you.

Pierre-Sassoulas: I suggested that because you precised this for the review so we might as well put a permanent comment ?

DanielNoord: Hm yeah, I don't actually specifically known why we don't want to do it.. There is a reference to an issue number of an old numbering system which I can't seem to find and I didn't immediately understand the test that was added in the same commit.

Unless somebody comes in and comes up with an actual explanation for why we don't want to do `_astmts = astmts[:1]` in this case I think we should leave this as is. I added the comment to make reviewing this easier.

Pierre-Sassoulas: Make sense, I'm resolving :)

Pierre-Sassoulas (approved):
LGTM, but I'm going to wait for @cdce8p review to merge.

pylint/checkers/typecheck.py
```diff
@@ -72,7 +72,7 @@
 from typing import Any, Callable, Iterator, List, Optional, Pattern, Tuple
 
 import astroid
-from astroid import bases, nodes
+from astroid import bases, exceptions, nodes
```
cdce8p, line 75: I'm not sure we should use this import here.
There is a `pylint.exceptions` package too which is often imported with
```py
from pylint import exceptions
```
Thus doing the same for astroid will lead to name conflicts.

At some point @Pierre-Sassoulas and I discussed this I think, although we didn't reach a conclusion for a good name. For the time being, I would recommend this
```py
import astroid.exceptions

try:
    ...
except astroid.exceptions...:
    ...
```
It's a bit lengthy but already used elsewhere. Should be good enough util we find something else.

pylint/checkers/typecheck.py
```diff
@@ -1029,10 +1029,12 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
                 if not [
                     n
                     for n in owner.getattr(node.attrname)
-                    if not isinstance(n.statement(), nodes.AugAssign)
+                    if not isinstance(n.statement(future=True), nodes.AugAssign)
                 ]:
                     missingattr.add((owner, name))
                     continue
+            except exceptions.StatementMissing:
+                continue
```
cdce8p, line 1037: After the discussion about not adding the `FrameMissing` error in astroid, I've been thinking if we should deprecate / remove the `StatementMissing` error, too.

Is there a benefit here with it being more specific? I would probably recommend to just catch `ParentMissingError` here.

pylint/checkers/variables.py
cdce8p: Tbh I think you know more about this particular method than I do. It looks like it could work 😅

cdce8p (changes requested):

pylint/checkers/typecheck.py
DanielNoord: I'm not sure if there is any benefit, perhaps when doing something like `[(i.frame(), j.statement()) for i,j in zip(nodes, other_nodes)]` but that is really really convoluted...
On the other hand, we have already gone to the effort of adding it and I don't think it adds too much boilerplate to the `astroid` codebase. We never know when we might need it in the future so I'd say let's just keep it in?

pylint/checkers/typecheck.py
DanielNoord: I have added an `import astroid.exceptions` line. There is already a `import astroid` statement, but `astroid.exceptions.StatementMissing` could not be resolved by VS Code (/ Pylance).. I think because of the wildcard import in `__init__.py` in `astroid`.

pylint/checkers/typecheck.py
cdce8p: Ok, let's keep it. Only because it's already in `astroid` 😉

PR #5316, opened 2021-11-15
Make ``unused-import`` check all ancestors for typing guards

DanielNoord:

## Type of Changes

## Description

See https://github.com/PyCQA/astroid/pull/1235#discussion_r749691526

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -61,7 +61,7 @@ import os
 import re
 import sys
 from functools import lru_cache
-from typing import DefaultDict, List, Optional, Set, Tuple
+from typing import DefaultDict, List, Optional, Set, Tuple, Union
 
 import astroid
 from astroid import nodes
@@ -358,12 +358,13 @@ def _assigned_locally(name_node):
     return any(a.name == name_node.name for a in assign_stmts)
 
 
-def _is_type_checking_import(node):
-    parent = node.parent
-    if not isinstance(parent, nodes.If):
-        return False
-    test = parent.test
-    return test.as_string() in TYPING_TYPE_CHECKS_GUARDS
+def _is_type_checking_import(node: Union[nodes.Import, nodes.ImportFrom]) -> bool:
+    """Check if an import node is guarded by a TYPE_CHECKS guard"""
+    for ancestor in node.node_ancestors():
+        if isinstance(ancestor, nodes.If):
+            if ancestor.test.as_string() in TYPING_TYPE_CHECKS_GUARDS:
+                return True
+    return False
 
 
 def _has_locals_call_after_node(stmt, scope):
--- a/tests/functional/u/unused/unused_import.py
+++ b/tests/functional/u/unused/unused_import.py
@@ -61,3 +61,7 @@ class NonRegr(object):
     def blop(self):
         """yo"""
         print(self, 'blip')
+
+if TYPE_CHECKING:
+    if sys.version_info >= (3, 6, 2):
+        from typing import NoReturn


pylint/checkers/variables.py
```diff
@@ -358,12 +358,13 @@ def _assigned_locally(name_node):
...
-        return False
-    test = parent.test
-    return test.as_string() in TYPING_TYPE_CHECKS_GUARDS
+def _is_type_checking_import(node: Union[nodes.Import, nodes.ImportFrom]) -> bool:
+    """Check if an import node is guarded by a TYPE_CHECKS guard"""
+    for ancestor in node.node_ancestors():
+        if isinstance(ancestor, nodes.If):
+            if ancestor.test.as_string() in TYPING_TYPE_CHECKS_GUARDS:
+                return True
+    return False
```
Pierre-Sassoulas, line 367: This looks like it should be detected as a possible any() by the "consider-any-or-all" checker 

```suggestion
    return any(
        a.test.as_string() in TYPING_TYPE_CHECKS_GUARDS
        for a in node.node_ancestors()
        if isinstance(a, nodes.If)
    )
```

Pierre-Sassoulas (approved):
I'm glad it's for 2.12.2 as 2.12.0 just reach 200 issues, and I like round number 😄

pylint/checkers/variables.py
DanielNoord: Are you sure you find the `any` more readable? In this case I personally prefer the `for` loop, but happy to commit if you prefer this one.

I've also stopped adding to `2.12` but thought that `--enable-all-extensions` would be valuable to try and squeeze in.

Pierre-Sassoulas: Well it's kinda opinionated at this point, in fact I think it should be

```python
    return any(
        isinstance(a, nodes.If) and a.test.as_string() in TYPING_TYPE_CHECKS_GUARDS
        for a in node.node_ancestors()
    )
```   

But arguably it's very dense so readability is not great.

>  that --enable-all-extensions would be valuable to try and squeeze in.

Well, it did, just in time to be number 200 😉

DanielNoord: I'll leave this to you 😄

Pierre-Sassoulas: Let's keep the for loop

DanielNoord: Btw, since I did not open an issue this wouldn't push the 200 limit 👀

Pierre-Sassoulas: I'm actually more confident in a 200 issues/MR release with primer tests and typing than I was for 2.10 and 2.9 that had around 100 issues/MR btw 😄 We're reducing risk drastically !

DanielNoord: Certaintly! Especially after we fixed 3 (4?) crashes already identified by the primer. Our work payed off immediately!

Pierre-Sassoulas: > we fixed 3 (4?) crashes

This did not feel like it, because it felt easy and small compared to the primer tests and configuration test functional framework, but this is huge ! The crash issues that we have to handle fast just after a release are the most stressful one. Looking forward to release being uneventful !

DanielNoord: It helped that I was stuck on an airplane twice last week 😄

DanielNoord:
Closing and opening to run the `primer` on this!

DanielNoord:
No changelog entry? Hadn't added it yet

Pierre-Sassoulas:
> No changelog entry? Hadn't added it yet

I supposed it was voluntary as in "this is a change small enough to not be in the changelog", my bad.

DanielNoord:
Ah no, I think this should go in the changelog. Could you add an entry? I'm away from computer atm.

Pierre-Sassoulas:
Sure, no problem !

Pierre-Sassoulas:
I figured I'd do that in #5397 instead of opening a MR specifically for this.

pylint/checkers/variables.py
```diff
@@ -358,12 +358,13 @@ def _assigned_locally(name_node):
...
-    parent = node.parent
-    if not isinstance(parent, nodes.If):
-        return False
-    test = parent.test
-    return test.as_string() in TYPING_TYPE_CHECKS_GUARDS
+def _is_type_checking_import(node: Union[nodes.Import, nodes.ImportFrom]) -> bool:
+    """Check if an import node is guarded by a TYPE_CHECKS guard"""
+    for ancestor in node.node_ancestors():
+        if isinstance(ancestor, nodes.If):
+            if ancestor.test.as_string() in TYPING_TYPE_CHECKS_GUARDS:
```
cdce8p, line 365: I'm again late, as I'm going through my notifications.
Can't those two be combined? Technically even a `return any(...)` would be possible, but I think that might be more difficult to read.

DanielNoord: Yeah, should have catched that myself. Going to make a PR with this change!

`consider-combining-if` seems like a handy checker for our review process 😄

cdce8p: > `consider-combining-if` seems like a handy checker for our review process 😄

That is also somewhere on my ideas list 😅

DanielNoord: I found the `any` to be okay for readability. See #5410

Pierre-Sassoulas: > consider-combining-if seems like a handy checker for our review process

Why didn't I think of that sooner, It's like half the content of my reviews 😄 Also it should trigger on thing like  ``any(bool(e == 1) for e in elements if int(e) == e))``. Let me open an issue.

Pierre-Sassoulas: https://github.com/PyCQA/pylint/issues/5414

PR #5325, opened 2021-11-16
Add typing to ``PyLinter.reporter``

DanielNoord:

## Type of Changes

## Description

We always expect `reporter` not to be `None`, but we do not initialise it as such. The comment in `load_default_plugins` shows that it is often necessary to have one.
Furthermore, this makes the typing of `PyLinter.reporter` much easier as we won't need to do `if linter.reporter` anymore.

I will add a changelog entry for `2.13` once `2.12` is out. Therefore currently blocked.


tests/test_self.py
```diff
@@ -1259,3 +1259,10 @@ def test_enable_all_extensions() -> None:
             exit=False,
         )
         assert sorted(plugins) == sorted(runner.linter._dynamic_plugins)
+
+    @staticmethod
+    def test_load_text_repoter_if_not_provided() -> None:
+        """Test if PyLinter.reporter is a TextRepoter is no reporter is provided"""
```
Pierre-Sassoulas, line 1265:
```suggestion
        """Test if PyLinter.reporter is a TextReporter if no reporter is provided"""
```

pylint/lint/pylinter.py
```diff
@@ -567,16 +580,10 @@ def __init__(self, options=(), reporter=None, option_groups=(), pylintrc=None):
...
-        if reporter:
-            self.set_reporter(reporter)
 
     def load_default_plugins(self):
         checkers.initialize(self)
         reporters.initialize(self)
-        # Make sure to load the default reporter, because
-        # the option has been set before the plugins had been loaded.
-        if not self.reporter:
-            self._load_reporters()
```
Pierre-Sassoulas, line 579: I like how it already simplify the code by removing handling that is dispersed in place it should not be.

Pierre-Sassoulas (approved):
LGTM !

pylint/lint/pylinter.py
DanielNoord: `MultiReporter` might create some additional problems for us. Have not looked into it fully, but having a reporter "that's not actually a reporter" on `PyLinter.reporter` might be troublesome..

areveny:
This looks like a good change to make the reporter explicit. I think the ``Run`` wrapper introduces a text reporter through the default options, but it's not intuitive how ith appens.

Since I was just looking at the reporters, I tried out some of the command line arguments I was testing with and they have the same functionality as before with this change.

PR #5329, opened 2021-11-17
Add error codes to all ``type: ignore``

DanielNoord:

## Type of Changes

## Description

Similar to the `astroid` PR.


cdce8p (approved):
LGTM! I would be fine with merging this now, what do you think @Pierre-Sassoulas?

Pierre-Sassoulas (approved):
LGTM, Yes, Let's merge. The only blocker for 2.12.0 is the crash in the proxy class, I don't want to add more features (more problems), but we can merge small non problematic MR :)

PR #5330, opened 2021-11-17
use isorts black profile

CarliJoy:
## Type of Changes

## Description

isort has a special black profile, this prevents that isorts changes a line and black changes it back.

See https://pycqa.github.io/isort/docs/configuration/black_compatibility.html

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -75,10 +75,8 @@ markers =
     benchmark: Baseline of pylint performance, if this regress something serious happened
 
 [isort]
-multi_line_output = 3
-line_length = 88
+profile = black
 known_third_party = platformdirs, astroid, sphinx, isort, pytest, mccabe, six, toml
-include_trailing_comma = True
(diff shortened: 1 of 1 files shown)


CarliJoy:
@Pierre-Sassoulas not really a typo but close ;-)

setup.cfg
```diff
@@ -76,9 +76,8 @@ markers =
 
 [isort]
 multi_line_output = 3
```
Pierre-Sassoulas, line 78: I think this is also for black compatibility but I4ve done that a long time ago

```suggestion
```

Pierre-Sassoulas (approved):
Nice 👍

setup.cfg
CarliJoy: I tried it. Result of `pre-commit run --all-files` are the same, so I removed it as well...

Pierre-Sassoulas:
Thank you, and congratulation on becoming a pylint contributor :) You're trusted by github action now 😉

PR #5331, opened 2021-11-17
fix(consider-interating-dictionary): fix false negatives

### LINKED ISSUE #5323: False negative for ``consider-iterating-dictionary`` when using ``not`` or ``list``
### Bug description

In the following sample `a.py`, only the last if raise a `consider-iterating-dictionary`

```python
metadata = {}
if "a" not in list(metadata.keys()):
    print(1)
if "a" not in metadata.keys():
    print(1)
if "a" in list(metadata.keys()):
    print(1)
if "a" in metadata.keys():
    print(1)
```    

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:8:10: C0201: Consider iterating the dictionary directly instead of calling .keys() (consider-iterating-dictionary)

Your code has been rated at 8.89/10 (previous run: 8.89/10, +0.00)
```

### Expected behavior

All four ifs are consider-iterating-dictionary

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.5
Python 3.8.10 (default, Sep 28 2021, 16:10:42) 
[GCC 9.3.0]
```

yushao2:
## Type of Changes

## Description

Added loop logic to check upwards (parents of node) until i hit a ``compare`` node, to circumvent the issue where `list(dict.keys())` results in the `node.parent` to be of type `call`

Also, added operator check for `not in`

Closes #5323


pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -83,21 +83,23 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
...
-                inferred.bound, nodes.Dict
-            ):
-                return
+            return
 
+        if isinstance(node.parent, (nodes.For, nodes.Comprehension)):
+            self.add_message("consider-iterating-dictionary", node=node)
+
+        while not isinstance(node.parent, (nodes.Compare, nodes.Module)):
+            node = node.parent
```
DanielNoord, line 96: Let's use something like this here:
https://github.com/PyCQA/astroid/blob/907818246c89ef938b23448e0cecc52b7c8693e0/astroid/nodes/node_classes.py#L4830-L4835

I think something like this should work:
```python
T_Node = TypeVar("T_Node", bound=NodeNG)

def get_if_node_type_ancestor(node: NodeNG, type: Tuple[Type[T_Node]]) -> Optional[T_Node]:
    """Return the first parent node that is any of the provided types (or None)"""
        if isinstance(ancestor, type):
            return ancestor
    return None
```

Note that `type` is a `tuple` so we can use this function later to also check for multiple types.

We could at it to the checker utils file.

WIth this, we could do one check for ancestors and move all of the logic after that check

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -83,21 +83,23 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
...
             return
-        if (
-            isinstance(node.parent, (nodes.For, nodes.Comprehension))
-            or isinstance(node.parent, nodes.Compare)
-            and any(
-                op
-                for op, comparator in node.parent.ops
-                if op == "in" and comparator is node
-            )
+        inferred = utils.safe_infer(node.func)
```
DanielNoord, line 86: I think it is better if we do the (performance-heavy) inference after we have done the check for ancestors (like we do in the current code).

With the changes in the other comment this should be easily doable.

DanielNoord (changes requested):
Thanks for this contribution!

We have currently closed `2.12` as we're trying to finalise that release, but this should definitely go in `2.13`. We will need to update the place of the changelog entry when we can (after release of `2.12`).

Pierre-Sassoulas:
Nice to see  a new contribution from you @yushao2 :)

pylint/checkers/refactoring/recommendation_checker.py
yushao2: really smart, i like that

pylint/checkers/refactoring/recommendation_checker.py
yushao2: fixed this too

yushao2:
> Nice to see a new contribution from you @yushao2 :)

great to be back, had a hard time adjusting to my new role as devops but slowly getting the hang of it (and thus more room to breathe and to contribute :D)

pylint/checkers/utils.py
```diff
@@ -1686,3 +1687,16 @@ def returns_bool(node: nodes.NodeNG) -> bool:
         and isinstance(node.value, nodes.Const)
         and node.value.value in (True, False)
     )
+
+
+T_Node = TypeVar("T_Node", bound=nodes.NodeNG)
```
DanielNoord, line 1692: Could you move this to the top of the file? We tend to keep our `TypeVar`'s there.

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -83,21 +83,22 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
...
         if (
             isinstance(node.parent, (nodes.For, nodes.Comprehension))
-            or isinstance(node.parent, nodes.Compare)
+            or isinstance(comp_ancestor, nodes.Compare)
             and any(
                 op
-                for op, comparator in node.parent.ops
-                if op == "in" and comparator is node
+                for op, comparator in comp_ancestor.ops
+                if op in ["in", "not in"]
```
DanielNoord, line 93:
```suggestion
                if op in {"in", "not in"}
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -83,21 +83,22 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
...
+                and (comparator in node.node_ancestors() or comparator is node)
```
DanielNoord, line 94: Just to check: is this ever untrue? I'm not sure how the `comparator` of `comp_ancestor.ops` works, but I can imagine this is always true?

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -83,21 +83,22 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
             return
         if node.func.attrname != "keys":
             return
+        comp_ancestor = utils.get_node_first_ancestor_of_type(node, nodes.Compare)
```
DanielNoord, line 86:
```suggestion
        comp_ancestor = utils.get_node_first_ancestor_of_type(node, (nodes.Compare,))
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -83,21 +83,22 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
...
         if (
             isinstance(node.parent, (nodes.For, nodes.Comprehension))
-            or isinstance(node.parent, nodes.Compare)
+            or isinstance(comp_ancestor, nodes.Compare)
```
DanielNoord, line 89:
```suggestion
            or comp_ancestor
```

The `isinstance` is unnecessary now that we use `utils.get_node_first_ancestor_of_type`.

DanielNoord (changes requested):
This is really quite elegant! Thanks for the quick work 😄

pylint/checkers/refactoring/recommendation_checker.py
yushao2: isn't tuple a better idea here?

```suggestion
                if op in ("in", "not in")
```

pylint/checkers/refactoring/recommendation_checker.py
yushao2: it could be a comparison like 

`if a == {}.keys() in {}`

one of the tests fail if this check is omitted, see L68 of test file 

`VAR = [1, 2] == {}.keys() in {False}`

pylint/checkers/refactoring/recommendation_checker.py
Pierre-Sassoulas: When checking for membership sets are faster, we have a check for that but it seems it's not activated right now.

DanielNoord (approved):
Looks good except for the set membership test. Thanks for this!

Pierre-Sassoulas (changes requested):
Nice changes thank you @yushao2 ! I request change so we remember to move the changelog to 2.13 before merging.

DanielNoord (approved):

Pierre-Sassoulas (approved):
Small change, big result 👍

DanielNoord:
@yushao2 Note that this PR was/is blocked by the release of `2.12.2` and the changelog entry in this PR is still in `2.12`. It should either be `2.12.2` or `2.13`. The other PR is also blocked!

Pierre-Sassoulas:
Don't worry about this one, I'll fix the changelog when releasing 2.12.2

yushao2:
That is my bad, i got confused. Sorry for the trouble

pylint/checkers/utils.py
```diff
@@ -1690,3 +1693,13 @@ def returns_bool(node: nodes.NodeNG) -> bool:
         and isinstance(node.value, nodes.Const)
         and node.value.value in {True, False}
     )
+
+
+def get_node_first_ancestor_of_type(
+    node: nodes.NodeNG, ancestor_type: Tuple[Type[T_Node]]
+) -> Optional[T_Node]:
```
cdce8p, line 1700: Don't know if that has been considered. There is no need to limit `ancestor_type` to `Tuple[...]`. `isinstance` works fine with `Type[T_Node]` too. Ie.
```py
ancestor_type: Union[Type[T_Node], Tuple[Type[T_Node]]]
```
This has the advantage that now you can pass both a single type and a tuple of types to it.

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -83,21 +83,22 @@ def _check_consider_iterating_dictionary(self, node: nodes.Call) -> None:
             return
         if node.func.attrname != "keys":
             return
+        comp_ancestor = utils.get_node_first_ancestor_of_type(node, (nodes.Compare,))
```
cdce8p, line 86:
```py
comp_ancestor = utils.get_node_first_ancestor_of_type(node, nodes.Compare)
```
This could then be simplified.

DanielNoord: Added to #5447 as a follow up issue to this.

I initially wanted to keep the typing of the function simple so thought it would be okay to only allow tuples, but we could allow both.

Issue #5321, opened 2021-11-16
Crash on dataclasses attributes within `open()` calls


DanielNoord:
### Bug description

This code should be added to `tests/functional/u/unspecified_encoding_py38.py`:

```python
import dataclasses
from typing import Optional

FILENAME = "test_two.py"


@dataclasses.dataclass
class IOArgs:
    """Dataclass storing information about how to open a file"""

    encoding: Optional[str]
    mode: str


args_good_one = IOArgs(encoding=None, mode="wb")
args_good_two = IOArgs(encoding="utf8", mode="w")
args_bad_one = IOArgs(encoding=None, mode="w")
LOCALE_ENCODING = None
args_bad_two = IOArgs(encoding=LOCALE_ENCODING, mode="w")

file = open(FILENAME, args_good_one.mode, encoding=args_good_one.encoding)
file = open(FILENAME, args_good_two.mode, encoding=args_good_two.encoding)
file = open(FILENAME, args_bad_one.mode, encoding=args_bad_one.encoding)
file = open(FILENAME, args_bad_two.mode, encoding=args_bad_two.encoding)
```

This currently creates a crash.

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
pylint test.py
Exception on node <Call l.21 at 0x10f881e40> in file '/Users/daniel/DocumentenLaptop/Programming/Github/pylint/test.py'
Traceback (most recent call last):
  File "/Users/daniel/DocumentenLaptop/Programming/Github/pylint/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "/Users/daniel/DocumentenLaptop/Programming/Github/pylint/pylint/checkers/stdlib.py", line 534, in visit_call
    self._check_open_encoded(node, inferred.root().name)
  File "/Users/daniel/DocumentenLaptop/Programming/Github/pylint/pylint/checkers/stdlib.py", line 631, in _check_open_encoded
    if not mode_arg or "b" not in mode_arg.value:
  File "/Users/daniel/.pyenv/versions/3.10.0/envs/pylint-3.10.0/lib/python3.10/site-packages/astroid/bases.py", line 133, in __getattr__
    return getattr(self._proxied, name)
AttributeError: 'ClassDef' object has no attribute 'value'
************* Module test
[...]

PR #5332, opened 2021-11-17
Fix crash on ``open()`` calls for non-string ``mode`` arguments

DanielNoord:

## Type of Changes

## Description

Addresses part of #5321


tests/functional/u/unspecified_encoding_py38.py
```diff
@@ -81,3 +83,20 @@
...
+    """Dataclass storing information about how to open a file"""
+
+    encoding: Optional[str]
+    mode: str
+
+
+args_good_one = IOArgs(encoding=None, mode="wb")
+
+# Test for crash reported in https://github.com/PyCQA/pylint/issues/5321
+open(FILENAME, args_good_one.mode, encoding=args_good_one.encoding)
```
Pierre-Sassoulas, line 102: Is it worth to infer the value of ``args_good_one.encoding`` to fix the false negative for unspecified encoding properly ?

DanielNoord: At the moment we can't (I think) as the issue with ``dataclass`` is still here. I used the tests from the issue as I thought it would be easy to add the other tests later on when that issue gets fixed!

cdce8p: > Is it worth to infer the value of `args_good_one.encoding` to fix the false negative for unspecified encoding properly ?

I think you meant the `mode` argument. That will be inferred already. What's missing is just the check that the inferred value is actually a `Const` node and not something else. Only then can we access the `value` attribute. This is what this PR adds.
https://github.com/PyCQA/pylint/blob/2e7b2c20cb08e739dfcc83c8ae1508a96ee8e04d/pylint/checkers/stdlib.py#L627-L628

cdce8p (changes requested):
Could you add three more test cases? The one you added will only test attribute access once the dataclass brain has been fixed.
1. `mode` as constant value
```py
MODE = "w"
open(FILENAME, mode=MODE)
```
2. `mode` as attribute 
```py
class X:
    mode = "w"
open(FILENAME, mode=X.mode)
```
3. Invalid value for `mode`. `pylint` shouldn't crash here.
```py
class X: ...
open(FILENAME, mode=X)
```

DanielNoord:
I have added some additional examples. At some point we might want to check whether these are actually correctly inferred and use `"w"` as mode to see if they produce a warning.
For now I used `"wb"` as that allows only testing the crashes.

tests/functional/u/unspecified_encoding_py38.py
```diff
@@ -81,3 +83,68 @@
 Path(FILENAME).open("w+")  # [unspecified-encoding]
 Path(FILENAME).open("w", encoding=None)  # [unspecified-encoding]
 Path(FILENAME).open("w", encoding=LOCALE_ENCODING)
+
+
+# Tests for storing data about open calls.
+# Most of these are regression tests for a crash
+# reported in https://github.com/PyCQA/pylint/issues/5321
+
+# Constants
```
cdce8p, line 92: Multiline comments seem to be broken at the moment 🤷🏻

I would recommend some formatting changes here.
```py
# -- Constants
MODE = "wb"
open(FILENAME, mode=MODE)


# -- Functions
def return_mode_function():
    """Return a mode for open call"""
    return "wb"

open(FILENAME, mode=return_mode_function())


# -- Classes
class IOData:
    """Class that returns mode strings"""

    mode = "wb"

    def __init__(self):
        self.my_mode = "wb"

    @staticmethod
    def my_mode_method():
        """Returns a pre-defined mode"""
        return "wb"

    @staticmethod
    def my_mode_method_returner(mode: str) -> str:
        """Returns the supplied mode"""
        return mode

open(FILENAME, mode=IOData.mode)
open(FILENAME, mode=IOData().my_mode)
open(FILENAME, mode=IOData().my_mode_method())
open(FILENAME, mode=IOData().my_mode_method_returner("wb"))
# Invalid value but shouldn't crash, reported in https://github.com/PyCQA/pylint/issues/5321
open(FILENAME, mode=IOData)


# -- Dataclasses
@dataclasses.dataclass
class IOArgs:
   ...
```

cdce8p:
> I have added some additional examples. At some point we might want to check whether these are actually correctly inferred and use `"w"` as mode to see if they produce a warning.
> For now I used `"wb"` as that allows only testing the crashes.

Good point!

tests/functional/u/unspecified_encoding_py38.py
```diff
@@ -81,3 +83,64 @@
...
+MODE = "wb"
+open(FILENAME, mode=MODE)
+
+
+# -- Functions
+def return_mode_function():
+    """Return a mode for open call"""
+    return "wb"
+
+
```
DanielNoord, line 102:
```suggestion
```

DanielNoord:
@cdce8p Sorry for pinging so soon, but as this is blocking the primer PR and thus some other PRs (in `astroid`) we need this PR before some things in `pylint` can move along.

Pierre-Sassoulas (approved):
LGTM

cdce8p (approved):

PR #5335, opened 2021-11-18
Update documentation for command-line output options and custom reporters

### LINKED ISSUE #5295: Allow passing an ``output`` to ``Run()`` to capture the output of a run
### Current problem

It seems logical to be able to capture the output of a `pylint` run with code that is similar to this:

```python
from io import StringIO
from pylint.reporters import text
from pylint.lint import Run


def main():
    pylint_output = StringIO()
    reporter = text.TextReporter(pylint_output)
    Run(["test_file.py"], reporter=reporter, do_exit=False)
    print(pylint_output.read()) # Show messages emitted by this pylint run


if __name__ == '__main__':
    main()
```

See for reference:
[This StackOverflow question](https://stackoverflow.com/questions/56081299/how-to-get-report-score-from-pylint-when-run-python-code) and #5260

### Desired solution

A little investigation showed that the `output` argument passed to the `reporter` doesn't actually get used as the `output` of the `PyLinter` class. By setting `PyLinter._output` the expected behaviour might actually be triggerable, but normally we don't want users to directly access or initialise that class.
As @Pierre-Sassoulas noted in https://github.com/PyCQA/pylint/pull/5260#issuecomment-967250512 it might be something to do with kwargs not being taken into account.

To achieve the expected behaviour we could either 1) make the `output` passed to the `reporter` of the `Run` instance be used as the `PyLinter` `output` or 2) allow passing a new `output` argument to `Run` which achieves the same thing.

### Additional context

_No response_

areveny:
## Type of Changes

## Description

This is a documentation improvement to clarify output command line options and provide an example of how to use an open stream wrapped in a custom reporter.

Closes #5295

### DIFF
--- a/doc/user_guide/output.rst
+++ b/doc/user_guide/output.rst
@@ -2,20 +2,42 @@
 Pylint output
 -------------
 
+Output options
+''''''''''''''''''''''''''''
+Output by default is written to stdout. The simplest way to output to a file is
+with the ``--output=<filename>`` option.
+
 The default format for the output is raw text. You can change this by passing
-pylint the ``--output-format=<value>`` option. Possible values are: json,
-parseable, colorized and msvs (visual studio).
+pylint the ``--output-format=<value>`` option. Possible values are: text, json,
+parseable, colorized and msvs (for Visual Studio).
 
 Multiple output formats can be used at the same time by passing
-``--output-format`` a comma-separated list of formats. To change the output file
-for an individual format, specify it after a semicolon. For example, you can
-save a json report to ``somefile`` and print a colorized report to stdout at the
-same time with :
+``--output-format`` a comma-separated list of formats.  
+This output can be redirected to a file by giving a filename after a colon. 
+
+For example, to save a json report to ``somefile.json`` and print 
+a colorized report to stdout at the same time:
 ::
 
-  --output-format=json:somefile,colorized
+  --output-format=json:somefile.json,colorized
 
-Moreover you can customize the exact way information are displayed using the
+Finally, it is possible to invoke pylint programatically with
+reporter initialized with a custom stream:
+
+::
+
+  pylint_output = StringIO() # Custom open stream
+  reporter = text.TextReporter(pylint_output)
+  Run(["test_file.py"], reporter=reporter, do_exit=False)
+  print(pylint_output.getvalue()) # Retrieve and print the text report
+
+This would be useful to capture pylint output in an open stream which
+can be passed onto another program.
+
+Custom message formats
+''''''''''''''''''''''''''''
+
+You can customize the exact way information are displayed using the
 `--msg-template=<format string>` option. The `format string` uses the
 `Python new format syntax`_ and the following fields are available :
 
--- a/doc/user_guide/output.rst
+++ b/doc/user_guide/output.rst
@@ -12,10 +12,10 @@ pylint the ``--output-format=<value>`` option. Possible values are: ``text``, ``
 ``parseable``, ``colorized`` and ``msvs`` (for Visual Studio).
 
 Multiple output formats can be used at the same time by passing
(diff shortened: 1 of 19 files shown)


doc/user_guide/output.rst
```diff
@@ -2,20 +2,42 @@
...
+Output options
+''''''''''''''''''''''''''''
+Output by default is written to stdout. The simplest way to output to a file is
+with the ``--output=<filename>`` option.
+
 The default format for the output is raw text. You can change this by passing
-pylint the ``--output-format=<value>`` option. Possible values are: json,
-parseable, colorized and msvs (visual studio).
+pylint the ``--output-format=<value>`` option. Possible values are: text, json,
+parseable, colorized and msvs (for Visual Studio).
```
DanielNoord, line 12:
```suggestion
pylint the ``--output-format=<value>`` option. Possible values are: ``text``, ``json``,
``parseable``, ``colorized`` and ``msvs`` (for Visual Studio).
```

doc/user_guide/output.rst
```diff
@@ -2,20 +2,42 @@
...
 
 Multiple output formats can be used at the same time by passing
-``--output-format`` a comma-separated list of formats. To change the output file
-for an individual format, specify it after a semicolon. For example, you can
-save a json report to ``somefile`` and print a colorized report to stdout at the
-same time with :
+``--output-format`` a comma-separated list of formats.
```
DanielNoord, line 15:
```suggestion
a comma-separated list of formats to ``--output-format``.
```

doc/user_guide/output.rst
```diff
@@ -2,20 +2,42 @@
...
+This output can be redirected to a file by giving a filename after a colon.
+
+For example, to save a json report to ``somefile.json`` and print
+a colorized report to stdout at the same time:
+::
+
+  --output-format=json:somefile.json,colorized
+
+Finally, it is possible to invoke pylint programatically with
+reporter initialized with a custom stream:
```
DanielNoord, line 25:
```suggestion
Finally, it is possible to invoke pylint programatically with a
reporter initialized with a custom stream:
```

DanielNoord (approved):
Very well written! I think this is a major improvement over what we currently have!!

Pierre-Sassoulas (approved):
Thank you ! The doc is a lot clearer now, and you took the time to search for the right solution 👌

doc/user_guide/output.rst
```diff
@@ -2,20 +2,42 @@
...
+
 ::
 
-  --output-format=json:somefile,colorized
+  pylint_output = StringIO() # Custom open stream
+  reporter = text.TextReporter(pylint_output)
+  Run(["test_file.py"], reporter=reporter, do_exit=False)
+  print(pylint_output.getvalue()) # Retrieve and print the text report
+
```
Pierre-Sassoulas, line 33: Maybe we could also add an example with a context manager ?

DanielNoord (approved):

doc/user_guide/output.rst
```diff
@@ -2,20 +2,51 @@
...
+a comma-separated list of formats to ``--output-format``.
+This output can be redirected to a file by giving a filename after a colon.
+
+For example, to save a json report to ``somefile.json`` and print
+a colorized report to stdout at the same time:
+::
+
+  --output-format=json:somefile.json,colorized
+
+Finally, it is possible to invoke pylint programatically with a
```
jacobtylerwalls, line 24: This one trips me up, too!
```suggestion
Finally, it is possible to invoke pylint programmatically with a
```

areveny: Thanks! Now I'll know for the future too.

Pierre-Sassoulas (approved):
Really nice addition to 2.12, thank you @areveny

Issue #5336, opened 2021-11-18
Include end column / range in message output


cdce8p:
### Current problem

At the moment, `pylint` only include a column offset in the output. That works find for command line or CI uses but doesn't work too well in IDEs. They depend on us to output an error range so they can effectively highlight the issue.
https://github.com/microsoft/vscode-python/issues/18027

### Desired solution

We already have node information for most / all (?) errors. Those include end column information. It should be fairly strait forward to add a new formatting option to the pylint output.

**A Todo list:**
- [x] Modify output formatter to allow outputting end column information -> #5372
- [x] Update documentation -> #5372
- [x] Test changes -> #5349 & #5383
- [x] Update test runner to check end column info ->  #5349 & #5383
- [x] Update existing tests -> #5349 & #5383
- [x] Decide what to output if we don't have an end column. -> end of line, only one char, repeat column offset? -> #5372

### Additional context

Maybe we can release a `beta` version first. With that IDEs and extensions could be updated and then it's up to us.
For that first version we wouldn't need to update the existing tests. It should be enough to enable the output.

--
Side note: IMO, this should be completely optional. It doesn't make much sense to update the default output.

cdce8p:
/CC: @Pierre-Sassoulas @DanielNoord This has come up recently with the last `vscode` update. I believe it would be worth prioritizing it and maybe even to include the `beta` version in `2.12.0`. If we release it this weekend, there might be enough time to get it into `vscode-python` just in time for the last release of the year.

DanielNoord:
Can we use [`end_col_offset`](https://docs.python.org/3/library/ast.html#ast.AST.end_col_offset) and `end_lineno`?

I'm not sure what `vscode` (and other IDEs) need exactly and how they handle multi-line statements, but it seems `ast` might help us here.

cdce8p:
> Can we use [`end_col_offset`](https://docs.python.org/3/library/ast.html#ast.AST.end_col_offset) and `end_lineno`?

That's what I was thinking

Pierre-Sassoulas:
> Decide what to output if we don't have an end column.

I would say only one char as it minimize visual disturbance when we don't know exactly ? There will be a lot of case where a full line would make more sense but it's hard to guess. We'd have to think about it for each message, right ?

DanielNoord:
@cdce8p Are you working on this? If not I can start the process of adding `end_lineno` and `end_col_offset` to `nodes.Statement` in `astroid`. That should be the first step I think.

[...]

PR #5343, opened 2021-11-19
Add ``end_lineno`` and ``end_col_offset`` to ``MessageLocationTuple``

DanielNoord:

## Type of Changes

## Description

Ref: #5336

Preparation for when `astroid` allows us to actually get these attributes. This allows us to start passing the needed attributes to reporters. Some points for consideration:

1) I have opted to type these as `Optional[int]`. I think it might be more valuable to know that a certain message does not have an associated `end_line` or `end_col_offset` instead of an arbitrarily chosen one. If we want to pass an arbitrary number we can still can, but at least now we don't need to do so.

2) I looked online but couldn't find an answer: is it possible to let users using `MessageLocationTuple` know that they should supply a `end_line` and `end_col_offset`? I think it is better to require users to think about whether they want `end_line` to be `None` at creation time (when they still have access to the `node`) then through the default value. Personally I would like these to be required in `pylint` `3.0`, but couldn't find a good way to deprecate `MessageLocationTuple` with 6 values.

3) The necessity of using default values for `MessageLocationTuple` also makes this require `python` > `3.6.0`. I'm not sure what the idea behind `2.12` was: do we allow > `3.6.0` functions in it? Or do we try to minimise them?


Pierre-Sassoulas (commented):
Looks promising, I think there's also the functional tests to take into account and the output txt serialization/deserialization from testutil to upgrade ?  Also maybe it's time to introduce a location class with lineno, column, lineno_end, column_end attributes?

DanielNoord:
> Looks promising, I think there's also the functional tests to take into account and the output txt serialization/deserialization from testutil to upgrade ?

I want to tackle those in another PR after we can actually access these attributes.

> Also maybe it's time to introduce a location class with lineno, column, lineno_end, column_end attributes?

That's what `MessageLocationTuple` is for :) We actually implemented that a couple of weeks ago!

DanielNoord:
Should I fix the `TODO` in this PR and actually access `NodeNG.end_line` and `NodeNG.end_col_offset` or do we want to merge this first and keep the PRs small and manageable?

Pierre-Sassoulas:
> I fix the TODO in this PR and actually access NodeNG.end_line and NodeNG.end_col_offset 

Yes, in order to avoid a double conflict when rebasing on master. We'll have to ask every contributor that touched the functional tests to rebase to get end_line / end_column for some time after merging this.

pylint/message/message.py
```diff
@@ -25,6 +25,8 @@
         "obj",
         "line",
         "column",
+        "end_lineno",
+        "end_col_offset",
```
cdce8p, line 29: _Not a full review yet._

I was wondering, should we rename those to `end_line` and `end_column` respectively? That would match the existing `line` and `column`. Relevant as these are the arguments to `--msg-template`.

DanielNoord: I opted to follow what `ast` uses, but yeah this makes sense. I'll change!

DanielNoord:
I'm not sure what tests we would like to add here. We actually only have one test for the `Message` class which gets updated in this PR.
Let me know if you want to add any additional.

I propose to merge this PR and then make three separate PRs, based on the original to-do list:
1) Modify reporters to accept `end_line` and `end_column`
2) Modify functional test runner to accept `end_line` and `end_column` with a `DeprecationWarning` if not supplied
3) Modify unittests to use `end_line` and `end_column`

Only 1 is necessary for the beta release of this functionality, although if we manage to do so I think it would be nice to put them all in the same release.

Pierre-Sassoulas (approved):
> I propose to merge this PR and then make three separate PRs, based on the original to-do list

👍

pylint/lint/pylinter.py
```diff
@@ -1458,7 +1474,14 @@ def _add_one_message(
...
                 MessageLocationTuple(
-                    abspath, path, module or "", obj, line or 1, col_offset or 0
+                    abspath,
+                    path,
+                    module or "",
+                    obj,
+                    line or 1,
+                    col_offset or 0,
+                    end_lineno,
+                    end_col_offset,
```
cdce8p, line 1484: I was thinking again about this. Although I'm not 100% there yet, setting `end_lineno` and `end_col_offset` to `None` seems like the best we can do.

_We probably shouldn't touch `line or 1` and `col_offset or 0` though as this **will** break tools._

cdce8p (approved):
Ok lets try that! It at least seems to work with `vscode-python`. Although I'm still a bit unsure about the `None` values.

Pierre-Sassoulas:
@DanielNoord I let you merge when it's ready :)

PR #5344, opened 2021-11-19
fix(4716): fix false positive `unnecessary_dict_index_lookup` emitted when `del` is used

### LINKED ISSUE #4716: getting `unnecessary-dict-index-lookup` when deleting an item by value
I was wondering if I'm doing this wrong. I'm trying to remove a dict item when searching the dict's `items()` to re-assign a certain unique value to a new key.

-------------------------------
### NOTES

1. The dict I'm managing is a private class member and will not have multiple keys with the same value.
2. I'm using pylint v2.9.3 (had the same problem with v2.9.0 previously)

------------------------------
I've added `# pylint:` comments around the controversial line
```python
    def _set_address(self, node_id, address, search_by_address=False):
        """Set or change a node_id and network address pair on the master node."""
        for n_id, addr in self._addr_dict.items():
            if not search_by_address:  # search by key
                if n_id == node_id:
                    self._addr_dict[n_id] = address
                    break
            else:  # search by value
                if addr == address:
                    # pylint: disable=unnecessary-dict-index-lookup
                    del self._addr_dict[n_id]
                    # pylint: enable=unnecessary-dict-index-lookup
                    self._addr_dict[node_id] = address
                    break
        if node_id not in self._addr_dict.keys():
            self._addr_dict[node_id] = address
```

The pylint message suggests I use `addr` instead, but that doesn't help when deleting an item by key.

yushao2:
## Type of Changes

## Description

Add simple logic to check subscript's parent to see if it's `Delete` and ignore the rest check if so to prevent false positive emitted when performing index lookup for deletion of an dictionary entry

Closes #4716


tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
```diff
@@ -82,3 +82,15 @@ class Foo:
...
+# Should not be emitted for del
+# (https://github.com/PyCQA/pylint/issues/4716)
+d = {}
+for key, val in d.items():
+    del d[key]
+    break
+
+for item in d.items():
+    del d[item[0]]
+    break
```
DanielNoord, line 96:
```suggestion
    break

outer_dict = {"inner_dict": {}}
for key, val in outer_dict.items():
    for key_two, val_two in val.items():
        del outer_dict[key][key_two]  # [unnecessary-dict-index-lookup]
        break
```

Additional test to see whether we're not ignoring all subscripts on a `del` statement.

tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
```diff
@@ -82,3 +82,15 @@ class Foo:
     if 'V' in d[k]:  # [unnecessary-dict-index-lookup]
         d[k] = "value"
         print(d[k])  # This is fine
+
+# Test false positive described in #4716
+# Should not be emitted for del
+# (https://github.com/PyCQA/pylint/issues/4716)
```
DanielNoord, line 88: Thanks for adding the direct link here 😄

DanielNoord (approved):
Two valuable additions in quick succession 😄 One additional test but rest looks good!

We'll need to move the changelog for `2.13` again, but I'll add the `blocked` label to indicate this!

tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
yushao2: done! it works!

tests/functional/u/unnecessary/unnecessary_dict_index_lookup.txt
```diff
@@ -1,23 +1,24 @@
...
-unnecessary-dict-index-lookup:46:52:46:65::Unnecessary dictionary index lookup, use 'v' instead:UNDEFINED
-unnecessary-dict-index-lookup:48:37:48:50::Unnecessary dictionary index lookup, use 'v' instead:UNDEFINED
-unnecessary-dict-index-lookup:50:1:50:14::Unnecessary dictionary index lookup, use 'v' instead:UNDEFINED
-unnecessary-dict-index-lookup:51:1:51:14::Unnecessary dictionary index lookup, use 'v' instead:UNDEFINED
-unnecessary-dict-index-lookup:51:49:51:62::Unnecessary dictionary index lookup, use 'v' instead:UNDEFINED
-unnecessary-dict-index-lookup:57:10:57:20::Unnecessary dictionary index lookup, use 'item[1]' instead:UNDEFINED
-unnecessary-dict-index-lookup:60:1:60:11::Unnecessary dictionary index lookup, use 'item[1]' instead:UNDEFINED
-unnecessary-dict-index-lookup:65:10:65:20::Unnecessary dictionary index lookup, use 'item[1]' instead:UNDEFINED
-unnecessary-dict-index-lookup:82:14:82:18::Unnecessary dictionary index lookup, use '_' instead:UNDEFINED
+unnecessary-dict-index-lookup:7:10:None:None::Unnecessary dictionary index lookup, use 'v' instead:UNDEFINED
```
DanielNoord, line 1: Are you by any chance working on Python < 3.8? 
The end line and end column information of these messages should be added here, they seem to have been transformed into `None` due to the auto-updater. This is probably also going to fail the test.

@Pierre-Sassoulas Shall we make the updater only accessible on > 3.7? Or make it put out a warning on lower versions?

Pierre-Sassoulas: Yes we need a run time error if the upgrade would be useless in CI. #5418 is ideal for that I guess ?

DanielNoord: Yeah! Seems like a good idea. For now `3.8` is good enough since we don't have any nodes that only have their `end_column` data on 3.9 or 3.10. I think an exception when trying to run and a small explanation on why 3.8 is necessary should do.

Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/5418/commits/1e29198d93f82392b0154ad87908853acde9f0ee

yushao2: my bad let me switch my environments

DanielNoord (approved):

Pierre-Sassoulas (approved):
Very nice light fix, thank you !

PR #5346, opened 2021-11-19
Treat `typing.NamedTuple` self as a sequence

### LINKED ISSUE #5312: typing.NamedTuple self considered non-sequence in methods
### Bug description

When inheriting from `typing.NamedTuple`, the `self` value in methods is not recognized as a sequence. Unpacking `self` results in E0633 (unpacking-non-sequence). This only happens in its own methods, and there is no issue with unpacking elsewhere. Example:

```py
# pylint: disable=C0103,C0114,C0115,C0116
import typing

class V2(typing.NamedTuple):
    x: float
    y: float

    def sum(self):
        x, y = self  # <- pylint complains about unpacking
        return x + y
```

### Configuration

_No response_

### Command used

```shell
pylint example.py
```

### Pylint output

```shell
example.py:9:8: E0633: Attempting to unpack a non-sequence defined at line 4 (unpacking-non-sequence)
```

### Expected behavior

No complaint since unpacking `self` is valid here.

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.5
Python 3.10.0 (default, Oct  5 2021, 10:05:34) [GCC 11.2.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

mbyrnepr2:
Closes #5312

## Type of Changes

## Description

Closes #5312


tests/functional/u/unpacking_non_sequence.py
```diff
@@ -137,3 +138,12 @@ def flow_control_unpacking(var=None):
...
     return None
+
+
+class V2(NamedTuple):
+    x: float
+    y: float
+
+    def sum(self):
+        x, y = self
+        return x + y
```
Pierre-Sassoulas, line 149:
```suggestion
        return x + y
        
   def sum_3d(self):
        x, y, z = self
        return x + y + z        
```
Wondering what happens when we try to unpack more than what we can.

Pierre-Sassoulas (commented):
Looks already good, thank you !

DanielNoord:
@mbyrnepr2 Could you rebase on `main` and update the expected output of the functional tests? For `2.12` we have changed the expected output (end line and end column are now also expected) and they will need to be included for this to pass.
If you run python 3.8+ you can use the auto-updater provided in our testutils.

tests/functional/u/unpacking_non_sequence.py
cdce8p: This can probably be resolve. It will emit `unbalanced-tuple-unpacking`, see test case above.

tests/functional/u/unbalanced_tuple_unpacking.py
```diff
@@ -106,3 +107,12 @@ def test_issue_559():
     from ctypes import c_int
     root_x, root_y, win_x, win_y = [c_int()] * 4
     return root_x, root_y, win_x, win_y
+
+
+class MyClass(NamedTuple):
+    first: float
+    second: float
```
cdce8p, line 114: Could you add / modify one variable to include a default? Just to also test that.

tests/functional/u/unbalanced_tuple_unpacking.py
```diff
@@ -106,3 +107,12 @@ def test_issue_559():
...
+
+    def sum_3d(self):
+        first, second, third = self # [unbalanced-tuple-unpacking]
+        return first + second + third
```
cdce8p, line 118: Could you add two more functions here?
* One that doesn't emit a warning. (Basically the one from the other test file.)
* Unbalanced-unpacking but the other way, ie. trying to unpack 3 into 2 variables.

ChangeLog
```diff
@@ -18,6 +18,10 @@ Release date: TBA
 
   Closes #4716
 
+* False positive - Allow unpacking of ``self`` in a subclass of ``typing.NamedTuple``
```
cdce8p, line 21:
```suggestion
* Fix false positive - Allow unpacking of ``self`` in a subclass of ``typing.NamedTuple``.
```

pylint/checkers/variables.py
```diff
@@ -2130,9 +2130,17 @@ def _check_unpacking(self, inferred, node, targets):
         ):
             # Variable-length argument, we can't determine the length.
             return
+
+        # attempt to check unpacking is properly balanced
```
cdce8p, line 2134:
```suggestion
        # Attempt to check unpacking is properly balanced
```

cdce8p (commented):
Looks good overall, just some small comments.

cdce8p (approved):
LGTM! Thanks @mbyrnepr2 🐬

mbyrnepr2:
Thank you @cdce8p @Pierre-Sassoulas @DanielNoord for your help as always!

Pierre-Sassoulas (approved):
👍

PR #5349, opened 2021-11-21
Update functional test expected output

DanielNoord:

## Type of Changes

## Description

This adds the `confidence` to all expected outputs of functional tests.


pylint/testutils/output_line.py
```diff
@@ -106,6 +107,11 @@ def from_csv(cls, row: Union[Sequence[str], str]) -> "OutputLine":
             if isinstance(row, Sequence):
                 column = cls._get_column(row[2])
                 if len(row) == 5:
+                    warnings.warn(
+                        "In pylint 3.0 tests output should always include the expected confidence"
```
Pierre-Sassoulas, line 111: 👌

tests/functional/a/async_functions.txt
```diff
@@ -1,11 +1,11 @@
...
+redefined-builtin:5:0:next:Redefining built-in 'next':HIGH
+unused-argument:8:30:some_function:Unused argument 'arg2':HIGH
+bad-super-call:22:8:Class.some_method:Bad first argument 'OtherClass' given to super():HIGH
+too-many-arguments:26:0:complex_function:Too many arguments (10/5):HIGH
+too-many-branches:26:0:complex_function:Too many branches (13/12):HIGH
+too-many-return-statements:26:0:complex_function:Too many return statements (10/6):HIGH
+dangerous-default-value:57:0:func:Dangerous default value [] as argument:HIGH
+duplicate-argument-name:57:15:func:Duplicate argument name a in function definition:HIGH
+duplicate-argument-name:57:18:func:Duplicate argument name a in function definition:HIGH
+disallowed-name:62:0:foo:"Disallowed name ""foo""":HIGH
```
Pierre-Sassoulas, line 10:
```suggestion
disallowed-name:62:0:foo:Disallowed name "foo":HIGH
```
There's an issue with string containing ``"`` in the upgrader.

tests/functional/b/bad_open_mode.txt
```diff
@@ -1,6 +1,6 @@
-bad-open-mode:12:0::"""rwx"" is not a valid mode for open."
-bad-open-mode:13:0::"""rr"" is not a valid mode for open."
-bad-open-mode:14:0::"""+"" is not a valid mode for open."
-bad-open-mode:15:0::"""xw"" is not a valid mode for open."
-bad-open-mode:21:0::"""Ua"" is not a valid mode for open."
-bad-open-mode:22:0::"""Ur++"" is not a valid mode for open."
+bad-open-mode:12:0::"""rwx"" is not a valid mode for open.":HIGH
```
Pierre-Sassoulas, line 1: Apparently it was already applied in some place. It's not important.

Pierre-Sassoulas (changes requested):
Approving the first part, thank you ! :) Let's add the end line / end column change on top of it.

tests/functional/b/bad_open_mode.txt
DanielNoord: Should I still apply the other commit?
And do we want to update the upgrader before this PR?

Pierre-Sassoulas: If you know how to easily fix automatically with the upgrader we could do that but let's not do it manually :)

DanielNoord:
Used the new updater! Should be better now!

DanielNoord:
Used the old updater! Shouldn't fail now 😄

DanielNoord:
@Pierre-Sassoulas @cdce8p I'm working on updating `OutputLine`
https://github.com/PyCQA/pylint/blob/16a4e30da9ed738161372c5d12b5fd8f100d0f78/pylint/testutils/output_line.py#L76-L82

It makes sense to have `end_line` and `end_column` go in between `column` and object` in the attributes list. Not only does it makes more sense consistency-wise, but it would also allow the following code to be kept the same:
https://github.com/PyCQA/pylint/blob/16a4e30da9ed738161372c5d12b5fd8f100d0f78/pylint/testutils/output_line.py#L116-L117

Adding `end_line` and `end_column` to the end of `OutputLine`'s attributes would allow using a default value and keeps compatibility with previous versions. However, I wonder if direct access to `OutputLine` is considered part of the public API. I don't think anybody would normally need access to it directly: it should be called either from `from_csv()` or `from_msg()`. We can easily keep compatibility through those methods, even when inserting the attributes in between the others.

TLDR: Is `OutputLine` public or private? And keep we insert `end_line` and `end_column` in between the other attributes?

Pierre-Sassoulas:
> Is OutputLine public or private? And can we insert end_line and end_column in between the other attributes?

It's semi public (as maybe used by plugin developer for their tests ?), but as we provide a quick way to upgrade the expected output let's consider that this is a "breaking change" we can do in a minor.

DanielNoord:
> It's semi public (as maybe used by plugin developer for their tests ?),

Yeah that was what I was wondering. However, looking at the code I don't think they should every instantiate an `OutputClass` directly and only do so via `from_csv()` or `from_msg()`, which makes it less risky to change this!

> but as we provide a quick way to upgrade the expected output let's consider that this is a "breaking change" we can do in a minor.

Fair enough!

DanielNoord:
Latest commit is WIP and blocked by #5343.

pylint/testutils/output_line.py
```diff
@@ -77,41 +82,117 @@ class OutputLine(NamedTuple):
...
         if not PY38_PLUS:
             # We check the column only for the new better ast parser introduced in python 3.8
             return 0  # pragma: no cover
         return int(column)
 
+    @staticmethod
+    def _get_py38_none_value(value: T) -> Optional[T]:
+        """Handle attributes that are always None on pylint < 3.8 similar to _get_column."""
+        if not PY38_PLUS:
+            # We check the column only for the new better ast parser introduced in python 3.8
```
DanielNoord, line 122:
```suggestion
            # We check the value only for the new better ast parser introduced in python 3.8
```

DanielNoord:
I have added additional fixes to the functional test `testutils` as this seemed the right time to update them all at once.

Besides the updating of our own expected output files this PR now includes:
- Allow and add `end_lineno` and `end_column` to the expected output of functional tests
- `DeprecationWarning` for `OutputLine` without `confidence`, `end_lineno` and `end_column`
- Standard `confidence` is now `UNDEFINED` mimicking `PyLinter.add_message()`.
https://github.com/PyCQA/pylint/blob/9e32192fe7bf77281d0dfbc107984b751983e113/pylint/lint/pylinter.py#L1515-L1516
- Removed the note saying that functional test consider `UNDEFINED` and `HIGH` to be the same. See `tests/test_functional.py`
- Made the functional test update run even when the expected output equals the actual output. This makes it easier to update half-correct files and when you're supplying the flag you probably don't care about `pytest` taking 10 secs extra.

Pierre-Sassoulas (approved):
This look amazing. Checking that the column match the expected would take month. I don't think we have a choice but to release as is and wait for issues to get opened.

pylint/testutils/output_line.py
```diff
@@ -77,41 +82,117 @@ class OutputLine(NamedTuple):
...
             return 0  # pragma: no cover
         return int(column)
 
+    @staticmethod
+    def _get_py38_none_value(value: T) -> Optional[T]:
+        """Handle attributes that are always None on pylint < 3.8 similar to _get_column."""
+        if not PY38_PLUS:
+            # We check the value only for the new better ast parser introduced in python 3.8
+            return None  # pragma: no cover
+        return value
```
cdce8p, line 124: That will likely not work for nodes that added `end_lineno` and `end_col_offset` later. E.g. `keyword`, `Slice` (both only in `3.9`). Maybe only compare the values if `end_lineno is not None` or better a new field in `[testoptions]` to overwrite `PY38_PLUS`. I.e. if field is specified, ignore `end_lineno` check for all versions prior.

cdce8p (commented):
A few notes:
1. What is the value in adding `UNDEFINED` basically everywhere? Do we even use the confidence level somewhere?
2. > Checking that the column match the expected would take month. I don't think we have a choice but to release as is and wait for issues to get opened.

It should be much easier once `vscode-python` supports error ranges. Then it's basically just opening the test file and checking that the ranges make sense.

pylint/testutils/output_line.py
DanielNoord: Do we ever create messages on those nodes though? Even `invalid-slice-index` does not pass an `nodes.Slice` as its `node` parameter. Could it be that we're fine if we ignore these?

I'm not sure how the `testoptions` stuff work so I'll need some guidance if want to add a new one.

DanielNoord:
> 1. What is the value in adding `UNDEFINED` basically everywhere? Do we even use the confidence level somewhere?

It allows us to remove the weird ``__eq__`` method. `confidence` is used by the `confidence=` setting so it makes sense to actually include it in our tests. It is a bit dumb since if its `None` we add `UNDEFINED` in `PyLinter.add_message` but I think it makes sense to update this while we're at it.
The fact that we handled `HIGH` as if it is similar to `UNDEFINED` is weird anyway. That is also fixed by this.

Pierre-Sassoulas:
> It should be much easier once vscode-python supports error ranges.

👍 although checking the whole functional test directory might still take a considerable amount of time. The visual result in vscode is motivating though.

Pierre-Sassoulas:
@DanielNoord should we close https://github.com/PyCQA/pylint/issues/5336 when merging this or should we move https://github.com/PyCQA/pylint/issues/5336 to 2.13 ? I'm planning on finally releasing today :)

DanielNoord:
We shouldn't close. `end_column` can't be tested in the unittests right now. That will be the next PR I'm working on.

DanielNoord:
@Pierre-Sassoulas https://github.com/PyCQA/pylint/pull/5349#discussion_r755315222 This was not resolved and might become problematic in the future.

I think we shouldn't release this before fixing this!

Pierre-Sassoulas:
Ha, right. I was just about to release, thank you for bringing this up 😄

pylint/testutils/output_line.py
Pierre-Sassoulas: I think ``only compare the values if end_lineno is not None``  is the simpler because I don't know what we could put in the test_option. A condition that needs to succeed or we return None ? Seems like we'll have to launch an eval for that (with all the implication for job launched from fork via github actions) ?

DanielNoord: I looked into it a bit more and I think `min_pyver_end_line` could work? If we convert it to a boolean and pass it to `_get_py38_none_value` that should make it future proof while testing on versions that it should.

`only compare the values if end_lineno is not None` is what was previously done for `confidence` with the `__eq__`, I don't think that is the way to go as it allows to just never check it.

Pierre-Sassoulas: > min_pyver_end_line

Ho, of course. I did not think of that. Very nice solution ! Should we add one option removing the check for line/end_line/column/end_column (``min_pyver_position`` to check all 4 values properly only in the version that is the most accurate) or should we add 4 options (``min_pyver_end_line``, `` min_pyver_end_column`` ... in order to check line/column if line_end or column_end do not exists in other version) ?

DanielNoord: @cdce8p Knows more about this than I do so I'll let him respond to this but I believe `3.6` has `line` and `col_offset` for all nodes that should have it. `3.8` adds `end_line` and `end_col_offset` for most and `3.9` adds it for some that were missing.

Thus I think we can create one `min_pyver_end_position` (?) which defaults to `3.8.0` and can be increased if needed.

Pierre-Sassoulas: Don't worry 😄 ``min_pyver_end_position`` convinced me, do you agree @cdce8p ? By the way it's probably not strictly necessary to release 2.12.0 as we don't have trouble with it right now. We're probably  never emitting message for Slice/Keyword nodes ?

DanielNoord: Found some time, see #5386

PR #5351, opened 2021-11-21
Update functional test updater to print actual string

DanielNoord:

## Type of Changes

## Description

@Pierre-Sassoulas 

The problem was we're using the `csv` module which tries to escape delimiters and quote characters. See https://docs.python.org/3/library/csv.html
This is problematic for use as we're using `'` and `"` interchangeably in the messages we put out (something to be fixed soon™.

For now I think this also works. I tested with some of the files in the other PR and the results seemed satisfactory. You can use this fix and update `tests/functional/a/async_functions.txt` to see if it works. That output has both `'` and `"`.

### DIFF
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -114,4 +114,11 @@ class OutputLine(NamedTuple):
             raise MalformedOutputLineException(row, e) from e
 
     def to_csv(self) -> Tuple[str, str, str, str, str, str]:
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
Nice, you even removed two warnings :)

DanielNoord:
Oops, look like that `TestDialect` was used in another file..

DanielNoord:
@Pierre-Sassoulas Going to revert this commit. This is not as easy as we thought: some warnings include a `:` so we can't simply split on `:` to get the individual pieces of an OutputLine.

PR #5352, opened 2021-11-21
Revert "Update functional test updater to print actual string"

DanielNoord:
Reverts PyCQA/pylint#5351

See:
https://github.com/PyCQA/pylint/pull/5351#issuecomment-974827473


Pierre-Sassoulas (approved):
I'm ok with letting this rest, this is not very important we're upgrading the expected functional output with the script all the time anyway.

PR #5353, opened 2021-11-21
Add accetpance tests in CI for python 3.8+ instead of launching manually at release

Pierre-Sassoulas:
## Type of Changes

## Description

First step before primer that will be done in #5173. We need to separate both test because they take a long time see https://github.com/PyCQA/pylint/pull/5173#issuecomment-974828609

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -475,3 +475,37 @@ jobs:
         run: |
           . venv/bin/activate
           pytest --benchmark-disable tests/
+
+  pytest-acceptance:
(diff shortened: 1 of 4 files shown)


DanielNoord (approved):

DanielNoord:
Looking at the time of the previous CI this might not have too much impact sadly.. Perhaps making batches out of the external packages is the best option after all.

Pierre-Sassoulas:
I also removed duplicate-code from the stdlib check which is probably the reason why it went from really long to only 3mn. For reference we could not even run the "real" primer in 4+h with duplicate check activated.  Let's see how it goes.

PR #5354, opened 2021-11-21
Add tests `inherit-non-class` with subscriptable class

cdce8p:
## Type of Changes

## Description

Tests for https://github.com/PyCQA/astroid/pull/1266

### DIFF
--- a/tests/functional/i/inherit_non_class.py
+++ b/tests/functional/i/inherit_non_class.py
@@ -84,13 +84,15 @@ class NotInheritableMemoryView(memoryview): # [inherit-non-class]
 # Subscription of parent class that implements __class_getitem__
 # and returns cls should be allowed.
 class ParentGood:
-    def __class_getitem__(cls):
+    def __class_getitem__(cls, item):  # pylint: disable=unused-argument
         return cls
 
 class ParentBad:
-    def __class_getitem__(cls):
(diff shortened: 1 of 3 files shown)


tests/functional/i/inherit_non_class.py
```diff
@@ -79,3 +79,24 @@ class NotInheritableSlice(slice): # [inherit-non-class]
 
 class NotInheritableMemoryView(memoryview): # [inherit-non-class]
     pass
+
+
+# Subscription of parent class that implements __class_getitem__
+# and returns cls should be allowed.
+class ParentGood:
+    def __class_getitem__(cls):
```
mitya57, line 87: This method should accept one argument in addition to `cls`. Otherwise you will get this:
```python
>>> class ParentGood:
...     def __class_getitem__(cls):
...         return cls
... 
>>> class Child1(ParentGood[int]):
...     pass
... 
TypeError: __class_getitem__() takes 1 positional argument but 2 were given
```

mitya57 (changes requested):

tests/functional/i/inherit_non_class.py
cdce8p: Thanks @mitya57! I've updated the test cases.

Pierre-Sassoulas (approved):
👌

PR #5339, opened 2021-11-18
Add regression test for #4982

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [ ] Add a ChangeLog entry describing what your PR does.
- [ ] If it's a new feature, or an important bug fix, add a What's New entry in
      `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Adds a test for #4982, fixed in https://github.com/PyCQA/astroid/pull/1257.
Therefore, closes #4982

@Pierre-Sassoulas Do you want to wait with this for `2.13`? I haven't added a changelog entry because I wanted to know which release you want to put this in. We could also potentially do `2.12.x`, we just need the fix to be available in a released `astroid` version.

PR #5355, opened 2021-11-21
Upgrade astroid to 2.9.0

cdce8p:
## Description

I've included `Closes ...` here as #5339 doesn't have aa Changelog entry (and probably doesn't need one either).


Pierre-Sassoulas (approved):
Thank you !

PR #5357, opened 2021-11-21
Enum identity comparison 

### LINKED ISSUE #5356: Check enums members for identity comparison
### Current problem

Enum members should be compared by identity rather than equality: https://docs.python.org/3/library/enum.html#comparisons
> Enumeration members are compared by identity:

### Desired solution

Extending `singleton-comparison` to cover this would seem to be one possible implementation.

### Additional context

_No response_

scop:
## Type of Changes

## Description

String update and a test case for #5356, just the implementation missing ;)

Closes #5356

### DIFF
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -2329,7 +2329,7 @@ class ComparisonChecker(_BasicChecker):
             "Comparison %s should be %s",
             "singleton-comparison",
             "Used when an expression is compared to singleton "
-            "values like True, False or None.",
+            "values like True, False, None, or an Enum member.",
         ),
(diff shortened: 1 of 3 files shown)


scop:
To be clear, I would prefer if someone would take it from here and do the actual implementation, I eyeballed it a bit and found that due to my unfamiliarity with pylint internals/astroid, it would require more time from me than I can currently invest in this. Please by all means feel free to grab what's in this branch or even finish it there (it's open for contributors to push).

cdce8p:
I'm going to close this for now. If someone decides to pick this up for a custom extension they can do so then.

PR #5365, opened 2021-11-22
Regression test for 3181, add documentation for toml configuration

### LINKED ISSUE #3181: Documentation page for configuration file
Each time I want to try pylint again, I face the same issue: I go to the official documentation to search for information about the configuration file, but can't seem to find anything easily.

Obviously, I go to "User Guide -> Configuration" (https://pylint.readthedocs.io/en/latest/user_guide/options.html) but this is not at all what I expect. There's nothing about the the pylintrc file, or the available configuration options like `max-line-length`. To know if pylint supported pyproject.toml, I had to read the "Running Pylint" section :confused: 

Then I tell myself, "the options names must be the same on the CLI and in the `[tool.pylint]` section", so I write this:

```toml
[tool.pylint]
max-line-length = 120
disable = ["C0330"]
```

...but it doesn't seem to be taken into account. Should I replace `-` with `_`? Should options be in a sub-table? Where do I find this information?

Answers are: I don't know. The search feature on the documentation doesn't help. I eventually went to the FAQ: https://pylint.readthedocs.io/en/latest/faq.html#how-do-i-find-the-option-name-for-pylintrc-corresponding-to-a-specific-command-line-option. So now I know I can get the option name for `--max-line-length` by generating the pylintrc file, and copy-pasting it into pyproject.toml.

But it still does not work :confused:

How one should write the `[tool.pylint]` section?

Am I the only one having a hard time with pylint's documentation structure?

Pierre-Sassoulas:
## Type of Changes

## Description

This add two regression tests for #3181, fix a crash when we can't decode the TOML file, and modify the configuration framework to be able to handle the expected exit code (it can be 1 for fatal too).

Refer to #3181


Pierre-Sassoulas:
@DanielNoord sorry for adding another issue in 2.12, but it felt like a really small doc fix on a popular high priority issue, then I added regression and triggered a crash with an invalid toml conf, and one thing leading to another I had to change the test framework to handle "fatal" exit code 😄

pylint/config/option_manager_mixin.py
```diff
@@ -274,7 +274,10 @@ def read_config_file(self, config_file=None, verbose=None):
             self.set_current_module(config_file)
             parser = self.cfgfile_parser
             if config_file.endswith(".toml"):
-                self._parse_toml(config_file, parser)
+                try:
+                    self._parse_toml(config_file, parser)
+                except toml.TomlDecodeError as e:
+                    self.add_message("parse-error", line=0, args=str(e))
```
DanielNoord, line 280: I don't think `parse-error` should be used here. The description of the error refers to `astroid` not being able to build a representation of the code.

pylint/testutils/configuration_test.py
```diff
@@ -74,18 +70,56 @@ def get_expected_output(
...
+        return [
+            p
+            for p in path.parent.iterdir()
+            if path.stem in str(p) and str(p).endswith(suffix_filter)
+        ]
+
+    exit_code = 0
+    msg = (
+        "we expect a single file of the form "
+        "'filename_dot_expected_error_code_dot_out.32.out'"
```
DanielNoord, line 89: I think this name is a bit difficult to interpret with the `dot`. I read this as expecting a file with the form of `filename.bad-configuration-section.out.32.out`.

pylint/testutils/configuration_test.py
```diff
@@ -74,18 +70,56 @@ def get_expected_output(
     configuration_path: str, user_specific_path: Path
 ) -> Tuple[int, str]:
     """Get the expected output of a functional test."""
-    output = get_expected_or_default(configuration_path, suffix="out", default="")
-    if output:
+
+    def get_related_files(
+        tested_configuration_file: str, suffix_filter: str
+    ) -> List[Path]:
```
DanielNoord, line 78: Let's add a docstring here.

DanielNoord (changes requested):
I have no problem with adding this to `2.12`. Seems to include some good changes again!

However, I think this code is becoming quite difficult to understand when you're a first time contributor to `pylint`. I think we might want to create a short section in the `Contributing` section of our docs on how this part of the test suite works and how new tests should be added.
Once we finish the `toml` configuration parsing we shouldn't have to touch this much in the future, so documenting our work will be useful for ourselves as well.

Also I don't think this should close the issue that's linked. The documentation page that is linked to in the OP of that issue is still quite minimal imo. Once again, let's work on that after `argparse`, but I don't think the changes in this PR satisfy the requirements of that issue.

tests/testutils/test_configuration_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_get_expected_output(caplog: LogCaptureFixture) -> None:
+    caplog.set_level(logging.INFO)
+    exit_code, _ = get_expected_output(DATA_DIRECTORY / "t.toml", USER_SPECIFIC_PATH)
+    assert "Too much .out files" in str(caplog.text)
+    assert exit_code == -1
+    exit_code, _ = get_expected_output(DATA_DIRECTORY / "u.toml", USER_SPECIFIC_PATH)
+    assert exit_code == -1
+    assert "Wrong format for .out file name" in str(caplog.text)
+    exit_code, _ = get_expected_output(DATA_DIRECTORY / "v.toml", USER_SPECIFIC_PATH)
```
DanielNoord, line 21: We might actually want another error here: `.toml doesn't exists`. At least for now I think we should add `v.toml` to `tests/testutils/data`.

pylint/testutils/configuration_test.py
Pierre-Sassoulas: c556a5b

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: Very true, I'm glad you reviewed that ! https://github.com/PyCQA/pylint/pull/5365/commits/b43206f0bbb27bdf3f09f13371d667f44d5781f6

pylint/testutils/configuration_test.py
Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/5365/commits/f884802d20c9c02a14c5ca14c46ef3e7463c4ff3

Pierre-Sassoulas:
>  but I don't think the changes in this PR satisfy the requirements of that issue.

Yes this is quite minimal, but we don't have a lot of documentation for pylintrc either (well we can generate one) and we might want to change the top level section later so it feel like it might be a lot of work to do a full documentation on the top level headers. Maybe we should tell to generate the pylintrc and look at the top level header there ?

tests/testutils/test_configuration_test.py
Pierre-Sassoulas: 7794ba5 . The functional test loops on all the toml/cfg/ini file so the toml is guaranteed to exists if we launch ``get_expected_output``.

pylint/config/option_manager_mixin.py
```diff
@@ -274,7 +274,10 @@ def read_config_file(self, config_file=None, verbose=None):
             self.set_current_module(config_file)
             parser = self.cfgfile_parser
             if config_file.endswith(".toml"):
-                self._parse_toml(config_file, parser)
+                try:
+                    self._parse_toml(config_file, parser)
+                except toml.TomlDecodeError as e:
+                    self.add_message("conf-parse-error", line=0, args=str(e))
```
DanielNoord, line 280:
```suggestion
                    self.add_message("config-parse-error", line=0, args=str(e))
```

Might as well add those two additional letters :)

pylint/lint/pylinter.py
```diff
@@ -98,6 +98,11 @@ def _load_reporter_by_class(reporter_class: str) -> type:
         "Used when an exception occurred while building the Astroid "
         "representation which could be handled by astroid.",
     ),
+    "F0011": (
+        "error while parsing the configuration: %s",
+        "conf-parse-error",
+        "Used when an exception occurred while parsing the configuration given.",
```
DanielNoord, line 104:
```suggestion
        "Used when an exception occurred while parsing a pylint configuration file.",
```

pylint/testutils/configuration_test.py
```diff
@@ -70,22 +68,61 @@ def get_expected_configuration(
...
-    configuration_path: str, user_specific_path: Path
+    configuration_path: Union[str, Path], user_specific_path: Path
 ) -> Tuple[int, str]:
     """Get the expected output of a functional test."""
-    output = get_expected_or_default(configuration_path, suffix="out", default="")
-    if output:
+    exit_code = 0
+    msg = (
+        "we expect a single file of the form 'filename.32.out' with 'filename' the "
+        "stem of the configuration file, and '32' the expected error code."
```
DanielNoord, line 90:
```suggestion
        "We expect a single file of the form 'filename.32.out' where 'filename' represents "
        "the name of the configuration file, and '32' the expected error code."
```

tests/testutils/test_configuration_test.py
DanielNoord: Ah of course!

DanielNoord (changes requested):
Small spelling changes here and there.

I think `config-parse-error` is a little more descriptive and doesn't make the name too long, but I'll let you decide here.

DanielNoord:
> Yes this is quite minimal, but we don't have a lot of documentation for pylintrc either (well we can generate one) and we might want to change the top level section later so it feel like it might be a lot of work to do a full documentation on the top level headers. Maybe we should tell to generate the pylintrc and look at the top level header there ?

I think we should leave this issue open until `argparse` and see what that offers us in terms of automatic configuration file generation and automatic configuration documentation generation. 
I know that right now the auto-generated `pyltinrc` doesn't include all possible options (see https://github.com/PyCQA/pylint/pull/5154). So that would not help for all questions. Furthermore, I think there is also a discussion to be had about whether we should suggest to use `pylintrc` or `pyproject.toml` for new users. 
These are all things to consider when doing a good and full update of the docs, which I think the issue asks for.

DanielNoord (approved):

PR #5366, opened 2021-11-22
Update documentation on writing tests for ``pylint``

DanielNoord:

## Type of Changes

## Description

Thought I shouldn't ask for better documentation while not doing some of it myself. Nobody likes writing docs 😄 

Based on my comment in #5365

### DIFF
--- a/doc/development_guide/testing.rst
+++ b/doc/development_guide/testing.rst
@@ -39,9 +39,9 @@ astroid_ or any of the other dependencies has been published::
 To run only a specific test suite, use a pattern for the test filename
 (**without** the ``.py`` extension), as in::
 
-    python -m tox -e py36 -- -k test_functional
-    python -m tox -e py36 -- -k  \*func\*
-    python -m tox --recreate -e py36 -- -k test_functional # With recreation of the environment
+    python -m tox -e py310 -- -k test_functional
+    python -m tox -e py310 -- -k  \*func\*
+    python -m tox --recreate -e py310 -- -k test_functional # With recreation of the environment
 
 Since we use pytest_ to run the tests, you can also use it on its own.
 We do recommend using the tox_ command though::
@@ -103,15 +103,18 @@ To test the different ways to configure Pylint there is also a small functional 
 for configuration files. These tests can be found in the '/pylint/test/config' directory.
 
 To create a new test create a new file with an unused name in the directory of that type
-of configuration file. Subsequently add a ``.json`` file with the same name which records
+of configuration file. Subsequently add a ``filename.result.json`` file with 'filename'
+being the same name as your configuration file. This file should record
 what the configuration should be **compared to the standard configuration**.
 
-For example, if the configuration should add a warning to the list of disabled messages the
+For example, if the configuration should add a warning to the list of disabled messages
+and you changed the configuration for ``job`` to 10 instead of the default 1 the
 ``.json`` file should include::
 
     "functional_append": {
         "disable": [["a-message-to-be-added"],]
     }
+    "jobs": 10,
 
 Similarly if a message should be removed you can add the following to the ``.json`` file::
 
--- a/doc/development_guide/testing.rst
+++ b/doc/development_guide/testing.rst
@@ -99,7 +99,7 @@ the unittests.
 Writing functional tests for configurations
 -------------------------------------------
 
-To test the different way to configure Pylint there is also a small functional test framework
+To test the different ways to configure Pylint there is also a small functional test framework
 for configuration files. These tests can be found in the '/pylint/test/config' directory.
 
 To create a new test create a new file with an unused name in the directory of that type
--- a/doc/development_guide/testing.rst
+++ b/doc/development_guide/testing.rst
@@ -1,0 +1,131 @@
+.. -*- coding: utf-8 -*-
+.. _testing:
+
+==============
+ Testing
+==============
+
+.. _test_your_code:
+
+Test your code!
+----------------
+
+Pylint is very well tested and has a high code coverage. New contributions are not accepted
+unless they include tests.
+Pylint uses two types of tests: unittests and functional tests.
+
+  - The unittests can be found in the `/pylint/test` directory and they can
+    be used for testing almost anything Pylint related.
+
+  - The functional tests can be found in the `/pylint/test/functional` directory. They are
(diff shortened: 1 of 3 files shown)


doc/development_guide/testing.rst
```diff
@@ -0,0 +1,131 @@
...
+
+To run only a specific test suite, use a pattern for the test filename
+(**without** the ``.py`` extension), as in::
+
+    python -m tox -e py36 -- -k test_functional
+    python -m tox -e py36 -- -k  \*func\*
+    python -m tox --recreate -e py36 -- -k test_functional # With recreation of the environment
+
+Since we use pytest_ to run the tests, you can also use it on its own.
+We do recommend using the tox_ command though::
```
Pierre-Sassoulas, line 47: We should probably suggest ``pytest``... tox is astonishingly slow* (especially with ``--recreate`` !). It was a big barrier to contributing to pylint at first, at least for me. And I had a good computer. The test suite is even longer now, as we added standard test and also a benchmark that is heavy, so the slowness is even more of a problem now. Also I think the CI should prevent the issue that tox prevents anyway.

\* I just tested and tox is a lot faster now. 68s for pytest vs 74s for tox. (Also my computer got a LOT better) But the launch time is really a problem when launching a single test (2s vs 4s).

DanielNoord:
Should probably add `primer` as well after we have merged it. I'll add it to the to-do list of the follow-up issue.

doc/development_guide/testing.rst
```diff
@@ -0,0 +1,131 @@
...
+at some tests that test a similar funcitionality. This can often help write new tests.
+
+If your new test requires any additional files you can put those in the
+`/pylint/test/regrtest_data` directory. This is the directory we use to store any data needed for
+the unittests.
+
+
+Writing functional tests for configurations
+-------------------------------------------
+
```
Pierre-Sassoulas, line 101: You wrote the documentation for the code I made 😍 ?

doc/development_guide/testing.rst
```diff
@@ -0,0 +1,131 @@
...
+To test the different way to configure Pylint there is also a small functional test framework
+for configuration files. These tests can be found in the '/pylint/test/config' directory.
+
+To create a new test create a new file with an unused name in the directory of that type
+of configuration file. Subsequently add a `.json` file with the same name which records
```
Pierre-Sassoulas:
```suggestion
of configuration file. Subsequently add a `filename.result.json` file with 'filename' the same name than your configuration file. This file should records
```

doc/development_guide/testing.rst
```diff
@@ -0,0 +1,131 @@
...
+what the configuration should be **compared to the standard configuration**.
+
+For example, if the configuration should add a warning to the list of disabled messages the
+`.json` file should include::
+
+    "functional_append": {
+        "disable": [["a-message-to-be-added"],]
+    }
+
```
Pierre-Sassoulas:
```suggestion
For example, if the configuration should add a warning to the list of disabled messages
and you changed the configuration for ``job`` to 10 instead of the default 1 the
`.json` file should include::

    "functional_append": {
        "disable": [["a-message-to-be-added"],]
    }
    "jobs": 10,

```

doc/development_guide/testing.rst
```diff
@@ -0,0 +1,131 @@
...
+    python -m tox --recreate # The entire tox environment will be recreated
+    python -m tox --recreate -e py310 # The python 3.10 tox environment will be recreated
+
+
+To run only a specific test suite, use a pattern for the test filename
+(**without** the ``.py`` extension), as in::
+
+    python -m tox -e py36 -- -k test_functional
+    python -m tox -e py36 -- -k  \*func\*
+    python -m tox --recreate -e py36 -- -k test_functional # With recreation of the environment
```
Pierre-Sassoulas, line 44:
```suggestion
    python -m tox -e py39 -- -k test_functional
    python -m tox -e py39 -- -k  \*func\*
    python -m tox --recreate -e py39 -- -k test_functional # With recreation of the environment
    
    # Faster alternative, but you must handle the dependencies with pip editable install
    pytest -k test_functional 
    pytest -k  \*func\* # Faster alternative
    
```
Also upgraded the version so we appear hype :)

Pierre-Sassoulas (commented):
Amazing 👍 I have some small remarks

doc/development_guide/testing.rst
DanielNoord: Seeing as this is mostly read by new contributors I think suggesting `tox` makes sense. The difference in time is negligible and makes sure all dependencies etc. are set correctly for new contributors could help avoid problems. More experienced contributors will use `pytest` anyway probably.

doc/development_guide/testing.rst
DanielNoord: Yeah, felt bad to say you should do it while I could do it myself as well 😄

doc/development_guide/testing.rst
DanielNoord: Should we add those last 3 lines? The 5 lines after this basically cover `pytest` already.

doc/development_guide/testing.rst
DanielNoord: I can't commit this suggestion, but I agree. Can you re-suggest?

doc/development_guide/testing.rst
DanielNoord: 👍

doc/development_guide/testing.rst
Pierre-Sassoulas: Haha, don't feel bad, you're already doing a lot :)

doc/development_guide/testing.rst
Pierre-Sassoulas: Can't commit either, it's strange.

```suggestion
of configuration file. Subsequently add a `filename.result.json` file with 'filename' the same name than your configuration file. This file should records
```

Pierre-Sassoulas: Edit : Just saw you did it manually, thanks !

doc/development_guide/testing.rst
Pierre-Sassoulas: Fair enough.

Pierre-Sassoulas (approved):

Issue #5317, opened 2021-11-15
To-do list for release 2.13.0 


Pierre-Sassoulas:
- [x] #5730
- [x] Refactor the Confidence to be more accurate of what it means see https://github.com/PyCQA/pylint/pull/5369#discussion_r755818326 Done in #5519 
- [x] Release astroid 2.10 and upgrade the version in pylint (#5802)
- [x] 1 issue which needs some further investigation: https://github.com/PyCQA/pylint/issues/5645,
- [x] 1 issue which just need an implementation: https://github.com/PyCQA/pylint/issues/5341
- [x] 1 project maintenance issue: https://github.com/PyCQA/pylint/issues/4701
- [x] 1 project maintenance issue: #5877 
- [x] Another project maintenance issue: https://github.com/PyCQA/pylint/issues/5729
- [x] A lot of issues depending on astroid 2.10.

DanielNoord:
Shall we start the discussion on `check_docs`? The change that deprecated `check_docs` for backward compatibility is 6 years old, see https://github.com/PyCQA/pylint/commit/9423e31126fa2e951d17d1d57d59c1b14bbbd404

Maintaining it is problematic as we can't load the same plugin twice without crashing on the option registering. The change to use `docparams` is also quite easy (just change `check_docs` to `docparams` in your rcfile).

I would vote in favour of removing.

Pierre-Sassoulas:
> I would vote in favor of removing.

Same here, I think it's pretty safe to do at this point.

DanielNoord:
I'm thinking about moving this to 2.14 because of the amount of work involved and the number of duplicate issues we must handle that are already fixed in main and could be released :)

_Originally posted by @Pierre-Sassoulas in https://github.com/PyCQA/pylint/issues/5466#issuecomment-1035517869_

Just my two cents about `2.13`:
Looking at open issues and PRs for `2.13`:
1) There are 4 PRs which are in final stages of review and should be mergeable if we all get together and review and iterate on them
2) 3 issues which needs some further investigation: #5408, #5645,
3) 3 issues which just need an implementation: #5341, #5466 (my proposed hack), #5670
4) 2 project maintenance issues: #4701, #5729
5) A lot of issues depending on `astroid` `2.10`.

From `pylint`'s side that's actually quite manageable, although these are all issues that are less likely to attract contributors due to their difficulty.
[...]

PR #5369, opened 2021-11-22
Fix a crash on psycopg2 for elif used 

Pierre-Sassoulas:
## Type of Changes

## Description

https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/_queries.py#L281

Exception on node <If l.281 at 0x7f2f9141a550> 

pylint crashed with a ``IndexError`` and with the following stacktrace:
```
  File "/home/pierre/pylint/pylint/lint/pylinter.py", line 1035, in _check_files
    self._check_file(get_ast, check_astroid_module, file)
  File "/home/pierre/pylint/pylint/lint/pylinter.py", line 1070, in _check_file
  File "/home/pierre/pylint/pylint/lint/pylinter.py", line 1204, in check_astroid_module
  File "/home/pierre/pylint/pylint/lint/pylinter.py", line 1251, in _check_astroid_module
    walker.walk(node)
  File "/home/pierre/pylint/pylint/utils/ast_walker.py", line 75, in walk
    self.walk(child)
  File "/home/pierre/pylint/pylint/utils/ast_walker.py", line 75, in walk
    self.walk(child)
  File "/home/pierre/pylint/pylint/utils/ast_walker.py", line 75, in walk
    self.walk(child)
  [Previous line repeated 1 more time]
  File "/home/pierre/pylint/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "/home/pierre/pylint/pylint/extensions/check_elif.py", line 69, in visit_if
    if not self._elifs[self._if_counter]:
IndexError: list index out of range
```

The hot fix would be:
```diff
--- a/pylint/extensions/check_elif.py
+++ b/pylint/extensions/check_elif.py
@@ -66,7 +66,7 @@ class ElseifUsedChecker(BaseTokenChecker):
             orelse = node.parent.orelse
             # current if node must directly follow an "else"
             if orelse and orelse == [node]:
-                if not self._elifs[self._if_counter]:
+                if self._if_counter < len(self._elifs) and not self._elifs[self._if_counter]:
                     self.add_message("else-if-used", node=node)
         self._if_counter += 1
```

It seem to me that counting the if/elif properly would take some time. I created a minimal reproduction example:
```
def _if_in_fstring_comprehension():
    order = {}
    if not "false":
        raise TypeError(
            f" {', '.join(sorted(i for i in order or () if i not in vars))}"
        )
    elif "true":
        raise TypeError("d")
```
The issue come from the f string with a comprehension inside it that is not separated for the token processing.        

Detected in #5173

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,11 @@ Release date: TBA
 
 * Fix ``install graphiz`` message which isn't needed for puml output format.
 
+* Fix a crash in the ``check_elif`` extensions where an undetected ifs in comprehension
+  containing an if within a fstring resulted in a out of range error. The checker do not
+  rely on counting ifs anymore and use known ifs locations instead. It should not crash
+  on badly parsed ifs anymore.
+
 * Fix ``simplify-boolean-expression`` when condition can be inferred as False.
 
   Closes #5200
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,10 +17,10 @@ Release date: TBA
 
 * Fix ``install graphiz`` message which isn't needed for puml output format.
 
-* Fix a crash in the ``check_elif`` extensions where an undetected ifs in comprehension
-  containing an if within a fstring resulted in a out of range error. The checker do not
-  rely on counting ifs anymore and use known ifs locations instead. It should not crash
-  on badly parsed ifs anymore.
+* Fix a crash in the ``check_elif`` extensions where an undetected if in a comprehension
+  with an if statement within a f-string resulted in an out of range error. The checker no
+  longer relies on counting if statements anymore and uses known if statements locations instead.
+  It should not crash on badly parsed if statements anymore.
 
 * Fix ``simplify-boolean-expression`` when condition can be inferred as False.
 
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -1848,20 +1848,18 @@ a metaclass class method.",
                     "bad-mcs-method-argument",
                     node.name,
                 )
-        # regular class
-        else:  # pylint: disable=else-if-used
-            # class method
-            if node.type == "classmethod" or node.name == "__class_getitem__":
-                self._check_first_arg_config(
-                    first,
-                    self.config.valid_classmethod_first_arg,
-                    node,
-                    "bad-classmethod-argument",
-                    node.name,
-                )
-            # regular method without self as argument
-            elif first != "self":
-                self.add_message("no-self-argument", node=node)
+        # regular class with class method
+        elif node.type == "classmethod" or node.name == "__class_getitem__":
+            self._check_first_arg_config(
+                first,
+                self.config.valid_classmethod_first_arg,
+                node,
+                "bad-classmethod-argument",
+                node.name,
+            )
+        # regular class with regular method without self as argument
+        elif first != "self":
+            self.add_message("no-self-argument", node=node)
 
     def _check_first_arg_config(self, first, config, node, message, method_name):
         if first not in config:
--- a/pylint/extensions/check_elif.py
+++ b/pylint/extensions/check_elif.py
@@ -38,43 +38,28 @@ class ElseifUsedChecker(BaseTokenChecker):
         self._init()
 
     def _init(self):
-        self._elifs = []
-        self._if_counter = 0
+        self._elifs = {}
 
     def process_tokens(self, tokens):
-        # Process tokens and look for 'if' or 'elif'
-        for _, token, _, _, _ in tokens:
+        """Process tokens and look for 'if' or 'elif'"""
+        for _, token, begin, _, _ in tokens:
             if token == "elif":
-                self._elifs.append(True)
+                self._elifs[begin] = True
             elif token == "if":
-                self._elifs.append(False)
+                self._elifs[begin] = False
 
     def leave_module(self, _: nodes.Module) -> None:
         self._init()
 
-    def visit_ifexp(self, node: nodes.IfExp) -> None:
-        if isinstance(node.parent, nodes.FormattedValue):
-            return
-        self._if_counter += 1
-
-    def visit_comprehension(self, node: nodes.Comprehension) -> None:
-        self._if_counter += len(node.ifs)
-        self._elifs.insert(self._if_counter, True)
-
     @check_messages("else-if-used")
     def visit_if(self, node: nodes.If) -> None:
-        if isinstance(node.parent, nodes.If):
-            orelse = node.parent.orelse
-            # current if node must directly follow an "else"
-            if orelse and orelse == [node]:
-                if (
-                    # Safeguard if we missed an if in self._elifs
-                    # See https://github.com/PyCQA/pylint/pull/5369
-                    self._if_counter < len(self._elifs)
-                    and not self._elifs[self._if_counter]
-                ):
-                    self.add_message("else-if-used", node=node)
-        self._if_counter += 1
+        """Current if node must directly follow an 'else'"""
+        if (
+            isinstance(node.parent, nodes.If)
+            and node.parent.orelse == [node]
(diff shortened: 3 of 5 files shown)


Pierre-Sassoulas:
New warnings are false positives changing to draft.

pylint/extensions/check_elif.py
```diff
@@ -38,37 +38,28 @@ def __init__(self, linter=None):
...
-            # current if node must directly follow an "else"
-            if orelse and orelse == [node]:
-                if not self._elifs[self._if_counter]:
-                    self.add_message("else-if-used", node=node)
-        self._if_counter += 1
+        """Current if node must directly follow an 'else'"""
+        if (
+            isinstance(node.parent, nodes.If)
+            and node.parent.orelse == [node]
+            and not self._elifs[(node.lineno, node.col_offset)]
```
Pierre-Sassoulas, line 60: I don't know if we should guard against ``KeyError`` there, I'll check with the primers.

Pierre-Sassoulas:
@DanielNoord @cdce8p this is ready for review.

ChangeLog
```diff
@@ -17,6 +17,11 @@ Release date: TBA
 
 * Fix ``install graphiz`` message which isn't needed for puml output format.
 
+* Fix a crash in the ``check_elif`` extensions where an undetected ifs in comprehension
+  containing an if within a fstring resulted in a out of range error. The checker do not
+  rely on counting ifs anymore and use known ifs locations instead. It should not crash
+  on badly parsed ifs anymore.
+
```
DanielNoord, line 24:
```suggestion
* Fix a crash in the ``check_elif`` extensions where an undetected if in a comprehension
  with an if statement within a f-string resulted in an out of range error. The checker no
  longer relies on counting if statements anymore and uses known if statements locations instead.
  It should not crash on badly parsed if statements anymore.

```

pylint/checkers/classes.py
```diff
@@ -1848,20 +1848,18 @@ def _check_first_arg_for_type(self, node, metaclass=0):
                     "bad-mcs-method-argument",
                     node.name,
                 )
-        # regular class
-        else:  # pylint: disable=else-if-used
```
DanielNoord, line 1852: 😄

pylint/extensions/check_elif.py
```diff
@@ -38,37 +38,29 @@ def __init__(self, linter=None):
...
-        # Process tokens and look for 'if' or 'elif'
-        for _, token, _, _, _ in tokens:
+        """Process tokens and look for 'if' or 'elif'"""
+        for _, token, begin, _, _ in tokens:
             if token == "elif":
-                self._elifs.append(True)
+                self._elifs[begin] = True
             elif token == "if":
-                self._elifs.append(False)
+                self._elifs[begin] = False
```
DanielNoord, line 49: Do we need these to be `True` or `False` anywhere else than in the `visit_if`? For maintainability we might want to rename them to `"elif"` and `"if"` and just check for string equality.

tests/functional/ext/check_elif/check_elif.txt
```diff
@@ -1,2 +1,4 @@
-else-if-used:10:8:my_function:"Consider using ""elif"" instead of ""else if""":HIGH
-else-if-used:22:20:my_function:"Consider using ""elif"" instead of ""else if""":HIGH
+else-if-used:12:8:my_function:"Consider using ""elif"" instead of ""else if""":HIGH
```
DanielNoord, line 1: These will become `UNDEFINED` with the new updater. Do we want to change that already here?

tests/functional/ext/check_elif/check_elif.py
```diff
@@ -25,3 +27,19 @@ def my_function():
                 if myint:
                     pass
                 myint = 4
+
+
+def _if_in_fstring_comprehension_with_elif():
+    order = {}
+    if "z" not in "false":
+        raise TypeError(
+            f" {', '.join(sorted(i for i in order or () if i not in vars))}"
```
DanielNoord, line 36: Am I mistaken or is `vars` not defined? And if so, this is probably a false negative. We might want to define it here so that we don't need to add the message here when we fix the false negative.

DanielNoord (approved):
Looks good! Small changes here and there which we could do, I'll leave that decision to you!

tests/functional/ext/check_elif/check_elif.py
Pierre-Sassoulas: ``vars`` is a builtin but would need to be called on something (this is the code from psycopg https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/_queries.py#L159)

tests/functional/ext/check_elif/check_elif.txt
Pierre-Sassoulas: Let's start adding the confidence :)

tests/functional/ext/check_elif/check_elif.py
DanielNoord: Ah, yeah learn something new every day 😄

tests/functional/ext/check_elif/check_elif.txt
Pierre-Sassoulas: b2b9545a

pylint/extensions/check_elif.py
Pierre-Sassoulas: e1fe1672

tests/functional/ext/check_elif/check_elif.py
Pierre-Sassoulas: 29cf454f I renamed it to params because it's redefining a builtin. You're right that there should be an error in the previous code as vars is a function not an iterable and running the code would produce:
```
TypeError: 'builtin_function_or_method' object is not iterable
```

pylint/extensions/check_elif.py
```diff
@@ -38,37 +38,27 @@ def __init__(self, linter=None):
...
-                    self.add_message("else-if-used", node=node)
-        self._if_counter += 1
+        """Current if node must directly follow an 'else'"""
+        if (
+            isinstance(node.parent, nodes.If)
+            and node.parent.orelse == [node]
+            and (node.lineno, node.col_offset) in self._elifs
+            and self._elifs[(node.lineno, node.col_offset)] == "if"
+        ):
+            self.add_message("else-if-used", node=node, confidence=HIGH)
```
DanielNoord, line 61: 😄

Pierre-Sassoulas:
I don't understand the ``64 unchanged lines in 10 files lost coverage.`` from coverall. Maybe I need to rebase on main ?

pylint/extensions/check_elif.py
Pierre-Sassoulas: Not having an off by one error when we miscount an If increased the confidence 😄  But really the other option are:
```python
Confidence = namedtuple("Confidence", ["name", "description"])
# Warning Certainties
HIGH = Confidence("HIGH", "No false positive possible.")
INFERENCE = Confidence("INFERENCE", "Warning based on inference result.")
INFERENCE_FAILURE = Confidence(
    "INFERENCE_FAILURE", "Warning based on inference with failures."
)
UNDEFINED = Confidence("UNDEFINED", "Warning without any associated confidence level.")

CONFIDENCE_LEVELS = [HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED]
```
There's no inference here, so it should probably by HIGH. But also, we should probably refactor this (?) It was not used before so we might as well make it make sense. I'm not confident enough to say there will never be any false positive just because there was no inference.

DanielNoord: Seems like we could use a `NO_INFERENCE` level. Maybe add the to the growing list of todo's for `2.13`?

Pierre-Sassoulas: Updated https://github.com/PyCQA/pylint/issues/5317

PR #5376, opened 2021-11-22
Fix checking of ``confidence`` in the unittests

DanielNoord:

## Type of Changes

## Description

Preparation for the PR to update the unittests to use `end_line` and `end_column` > the third PR as mentioned in https://github.com/PyCQA/pylint/pull/5343#issuecomment-975659630.

Updating the unittests is going to take some time and perhaps might be best to do in a later release.
Basically the unittests currently don't really check `confidence` and `col_offset` is ignored completely. I think it makes more sense to update them all at once with `end_line` and `end_col_offset` as we're doing for the functional tests. 

However, I think for this it might be better to not do one large PR but do them in steps. There are some difficult changes here which I would like good review on before merging. Having one large PR that updates all four attributes might nog work well.

Another consideration here is whether this is part of the public API. This might be more "breaking" than the functional test as there is no easy way to update the expected output.
If we do consider this to be public I will need to redo the work here. Just let me know what you guys think. (We might consider making some aspects of our `testutils` private/unstable as to not limit ourselves).


pylint/testutils/output_line.py
```diff
@@ -12,30 +11,15 @@
 from pylint.testutils.constants import UPDATE_OPTION
 
 
-class MessageTest(
-    collections.namedtuple(
-        "MessageTest", ["msg_id", "line", "node", "args", "confidence"]
-    )
-):
+class MessageTest(NamedTuple):
```
Pierre-Sassoulas, line 14: 🙏 My goodness
feelgoodman

But is this already possible in python 3.6 ?

DanielNoord:
@Pierre-Sassoulas Can I add a skip for the failing test on `pypy` or do we want to investigate?

Pierre-Sassoulas (commented):
LGTM

> might be best to do in a later release.

Strongly agree with that, right now there is nothing but we can make it better progressively.

> as there is no easy way to update the expected output.

Could we add a default value to the expected output so the script keeps working until we put the right column later on ? What did I don't understand ?

pylint/testutils/output_line.py
DanielNoord: Yes! Or at least it should. The default value only from `3.6.1` but we are already depending on that in other PRs for `2.12` as well.

DanielNoord:
> > as there is no easy way to update the expected output.
> 
> Could we add a default value to the expected output so the script keeps working until we put the right column later on ? What did I not understand ?

I did do that: I'm using `UNDEFINED`. However, previously if the `confidence` level didn't match we just compared the tuple with `[:-1]`. See the removed `__eq__` method.
So, if a plugin has a warning with a different level than `UNDEFINED` but hasn't added that to the `MessageTest` used in its test, their tests will now fail. I think this is a correct fail: the test is not specific enough and adding `UNDEFINED` as a `confidence` when it is not provided is standard behaviour in `Pylinter.add_message()`. We're just (finally) mimicking that in the tests.
This is technically breaking though.

pylint/testutils/output_line.py
cdce8p: If I remember correctly, the change was planned for `2.12.1`. Would have to look up why though.

DanielNoord: I think because of the default value of `NamedTuple`. However, the `end_line` PR also used those.

DanielNoord:
Took me some tries to fix those test errors. Should have installed `3.6` from the start 😄 

I also added a changelog entry. 

If we agree that we find this to be just under the breaking-limit this is good to go for review/`2.12`.

pylint/testutils/output_line.py
cdce8p: True. With the exception in place, someone with `3.6.0` won't even get that far. `pip install pylint` will fail. It's probably an idea to pick up most of #5068 then too, except for bumping the `python_requires` version and removing the exception. 

https://github.com/PyCQA/pylint/blob/9e32192fe7bf77281d0dfbc107984b751983e113/setup.py#L17-L18

Pierre-Sassoulas (approved):
👍

PR #5383, opened 2021-11-24
Check `node` location attributes in unittests

DanielNoord:

## Type of Changes

## Description

List of unittests which "can't" be moved and are updated to now include location attributes:
- `TestMultiNamingStyle`

### DIFF
--- a/pylint/testutils/unittest_linter.py
+++ b/pylint/testutils/unittest_linter.py
@@ -44,20 +44,12 @@ class UnittestLinter:
         if confidence is None:
             confidence = UNDEFINED
         # pylint: disable=fixme
-        # TODO: Test col_offset
-        # pylint: disable=fixme
-        # TODO: Initialize col_offset on every node (can be None) -> astroid
-        # if col_offset is None and hasattr(node, "col_offset"):
-        #     col_offset = node.col_offset
-        # pylint: disable=fixme
-        # TODO: Test end_lineno and end_col_offset :)
-        # pylint: disable=fixme
-        # TODO: Initialize end_lineno on every node (can be None) -> astroid
+        # TODO: Initialize col_offset, end_lineno and end_col_offset on every node -> astroid
         # See https://github.com/PyCQA/astroid/issues/1273
+        if col_offset is None and node and hasattr(node, "col_offset"):
+            col_offset = node.col_offset
         if not end_lineno and node and hasattr(node, "end_lineno"):
             end_lineno = node.end_lineno
-        # pylint: disable=fixme
-        # TODO: Initialize end_col_offset on every node (can be None) -> astroid
         if not end_col_offset and node and hasattr(node, "end_col_offset"):
             end_col_offset = node.end_col_offset
         self._messages.append(
--- a/pylint/testutils/unittest_linter.py
+++ b/pylint/testutils/unittest_linter.py
@@ -30,6 +30,8 @@ class UnittestLinter:
         self,
         msg_id: str,
         line: Optional[int] = None,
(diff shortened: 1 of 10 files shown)


Pierre-Sassoulas:
> This will be a nightmare to review...

To be fair this is probably a nightmare to write too. How about we move most of this to the functional tests ? It's more understandable for contributor and a lot easier to maintain for us. Because a functional test would be reviewable with vscode + pylint (modulo finding a way to use a dev version in vscode) by opening the file and looking at the column displayed, right ?

DanielNoord:
> > This will be a nightmare to review...
> 
> To be fair this is probably a nightmare to write too. How about we move most of this to the functional tests ? It's more understandable for contributor and a lot easier to maintain for us. Because a functional test would be reviewable with vscode + pylint (modulo finding a way to use a dev version in vscode) by opening the file and looking at the column displayed, right ?

Not sure if converting to functional tests will take less time though.. I have never done it, did it take you much time last time?

Pierre-Sassoulas:
> did it take you much time last time?

Well it's relatively fast, you copy paste code and then add the expected message where required (``# [missing-docstring]``) or remove unwanted message. But it will take longer if the file does not exist yet I guess. Not making a mess and organizing the functional test properly will become important if we massively move unittest to functional though.

DanielNoord:
We're under 100 failed tests on 3.10 now 😄

Pierre-Sassoulas:
Only 36 tests left 🎉  but it look like the functional tests needs updating and there would be less than that.

DanielNoord:
Finally ready for review!

pylint/testutils/unittest_linter.py
```diff
@@ -30,6 +30,8 @@ def add_message(
         self,
         msg_id: str,
         line: Optional[int] = None,
+        # pylint: disable=fixme
+        # TODO: Make node non optional
```
Pierre-Sassoulas, line 34: 👍

Pierre-Sassoulas (approved):
There's still a surprisingly big number of tests that can"t be moved to functional, but this became clearly more manageable, I'm glad we won't have to change the line/col manually for the migrated tests. Amazing work you did there !

pylint/testutils/unittest_linter.py
DanielNoord: Turns out this is actually because node is optional in pylint itself 😅
I guess we can add that to our mental todo list for 3.0

DanielNoord:
> There's still a surprisingly big number of tests that can"t be moved to functional

For some of the remaining tests it was simply just better to keep them as is. Some used parameterization or imported additional data which would be difficult to replicate in functional tests.

pylint/testutils/unittest_linter.py
Pierre-Sassoulas: Yeap, that's probably some work too :)

Pierre-Sassoulas:
> Some used parameterization 

I think unittest works really well with parametrization when applicable. How about adding a little doc about tests : Prefer functional tests, unless there are a lot of similar cases that can be parametrized, or if you're using additional data ?

DanielNoord:
> > Some used parameterization
> 
> I think unittest works really well with parametrization when applicable. How about adding a little doc about tests : Prefer functional tests, unless there are a lot of similar cases that can be parametrized, or if you're using additional data ?

I don't think we should give people reasons to write more unittests, they will probably already do so since it is the most common way of testing.
There are so many difficulties with config parsing, loading of checkers etc that we should really prefer functional test unless truly impossible.

Issue #4761, opened 2021-07-28
False negative `used-before-assignment` when variable is only defined in `except` clause


gcolombeau:
### Bug description

```python
"""used-before-assignment (E0601)"""

def undefined_variable():
    """used-before-assignment (E0601)"""
    try:
        pass
    except ValueError:
        some_message = 'some message'

    if not some_message:
        return 1

    return some_message
```

### Command used

```shell
pylint e0601.py
```

### Pylint output

```shell
------------------------------------
Your code has been rated at 10.00/10
```

### Expected behavior

Pylint should report `used-before-assignment` on lines 10 and 13.

### Pylint version

```shell
pylint 2.9.5
astroid 2.6.5
Python 3.6.10 (default, Jan 31 2020, 10:49:38) 
[GCC 7.4.0]
```

PR #5384, opened 2021-11-24
Fix #85, #2615: Emit `used-before-assignment` in final or except blocks where try statements could have failed

### LINKED ISSUE #85: pylint does not recognize potential undefined variables
Originally reported by: **Mirko Dziadzka (BitBucket: [mirkodziadzka](http://bitbucket.org/mirkodziadzka), GitHub: @mirkodziadzka?)**

---

The following code should generate a warning, that the variable res may be undefined in  the finally clause

```
#!python
try:
    # throw exception here to make res undefined
    res = 42
finally:
    print res
```

---
- Bitbucket: https://bitbucket.org/logilab/pylint/issue/85

### LINKED ISSUE #2615: Not catching use of undefined variables in except clauses
This report is similar to others which mention undetected use of undefined variables after if False or in the finally branch of a try/finally statement. I think use of undefined variables in except clauses is perhaps a higher priority fix. Consider this function:

```
def func(url, headers):
    "doc"
    try:
        resp = requests.get(url, headers)
    except requests.exceptions.SSLError:
        print("SSL error - headers %s", resp.request.headers)
```

It's clear that the variable resp is used while undefined. Missing it is more critical than in most other cases because it is generally going to be executed infrequently. If, as in this case, only one variable is assigned in the try clause and it wasn't defined beforehand, then an undefined-variable message should be emitted. If more than one variable is assigned in the try clause, then a possibly-undefined-variable message should go out over the wire. I'm not suggesting that pylint should be able to do any kind of sophisticated analysis. If it wasn't defined before the try clause, is assigned in the try clause, then referenced in the except clause, it should be flagged.

If the user assigns 37 different variables in the try clause and references fourteen of them in the except clause, the voluminous amount of output will also serve as a good clue that they need to tighten up their try clause (quoting from PEP 8: "Additionally, for all try/except clauses, limit the try clause to the absolute minimum amount of code necessary.") :-)

jacobtylerwalls:
## Type of Changes

## Description

**Previously**
Pylint assumed assignments in the Try block of a Try/Finally or Try/Except pair would succeed. In cases of failure, though, attempting to use such names in a final or except block would raise `UnboundLocalError`.

**Now**
Pylint emits `used-before-assignment` to account for the fact that the statement might have been interrupted by an exception.

Closes #85, Closes #2615

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,11 @@ Release date: TBA
 
   Closes #4761
 
+* ``used-before-assignment`` now considers that assignments in a try block
+  may not have occurred when the except or finally blocks are executed.
+
+  Closes #85, #2615
+
 * ``used-before-assignment`` now checks names in try blocks.
 
 * Some files in ``pylint.testutils`` were deprecated. In the future imports should be done from the
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -405,8 +405,11 @@ MSGS = {
     "E0601": (
         "Using variable %r before assignment",
         "used-before-assignment",
-        "Used when a local variable is accessed before its assignment or within "
-        "except or finally blocks without being defined before the try block.",
+        "Emitted when a local variable is accessed before its assignment took place. "
+        "Assignments in try blocks are assumed not to have occurred when evaluating "
+        "associated except/finally blocks. Assignments in except blocks are assumed "
+        "not to have occurred, except when the associated try block contains a return "
+        "statement.",
     ),
     "E0602": (
         "Undefined variable %r",
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -698,7 +698,6 @@ scope_type : {self._atomic.scope_type}
                     and n.statement(future=True) in n.statement(future=True).parent.body
                     and node.statement(future=True).parent
                     in n.statement(future=True).parent.handlers
-                    and n.statement(future=True) in n.statement(future=True).parent.body
                 )
             ]
             filtered_nodes_set = set(filtered_nodes)
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -408,8 +408,7 @@ MSGS = {
         "Emitted when a local variable is accessed before its assignment took place. "
         "Assignments in try blocks are assumed not to have occurred when evaluating "
         "associated except/finally blocks. Assignments in except blocks are assumed "
-        "not to have occurred, except when the associated try block contains a return "
-        "statement.",
+        "not to have occurred when evaluating statements outside the block.",
     ),
     "E0602": (
         "Undefined variable %r",
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -405,7 +405,8 @@ MSGS = {
     "E0601": (
         "Using variable %r before assignment",
         "used-before-assignment",
-        "Used when a local variable is accessed before its assignment.",
+        "Used when a local variable is accessed before its assignment or in "
+        "except or finally blocks without being defined before the try block.",
     ),
     "E0602": (
         "Undefined variable %r",
@@ -656,6 +657,49 @@ scope_type : {self._atomic.scope_type}
                     )
                 )
                 or n.statement(future=True).parent.parent_of(node)
+            ]
+            filtered_nodes_set = set(filtered_nodes)
+            difference = [n for n in found_nodes if n not in filtered_nodes_set]
+            self.consumed_uncertain[node.name] += difference
+            found_nodes = filtered_nodes
+
+        # If this node is in a Finally block of a Try/Finally,
+        # filter out assignments in the try portion, assuming they may fail
+        if (
+            found_nodes
+            and isinstance(node.statement(future=True).parent, nodes.TryFinally)
+            and node.statement(future=True)
+            in node.statement(future=True).parent.finalbody
+        ):
+            filtered_nodes = [
+                n
+                for n in found_nodes
+                if not (
+                    n.statement(future=True).parent
+                    is node.statement(future=True).parent
+                    and n.statement(future=True) in n.statement(future=True).parent.body
+                )
+            ]
+            filtered_nodes_set = set(filtered_nodes)
+            difference = [n for n in found_nodes if n not in filtered_nodes_set]
+            self.consumed_uncertain[node.name] += difference
+            found_nodes = filtered_nodes
+
+        # If this node is in an ExceptHandler,
+        # filter out assignments in the try portion, assuming they may fail
+        if found_nodes and isinstance(
+            node.statement(future=True).parent, nodes.ExceptHandler
+        ):
+            filtered_nodes = [
+                n
+                for n in found_nodes
+                if not (
+                    isinstance(n.statement(future=True).parent, nodes.TryExcept)
+                    and n.statement(future=True) in n.statement(future=True).parent.body
+                    and node.statement(future=True).parent
+                    in n.statement(future=True).parent.handlers
+                    and n.statement(future=True) in n.statement(future=True).parent.body
+                )
             ]
             filtered_nodes_set = set(filtered_nodes)
             difference = [n for n in found_nodes if n not in filtered_nodes_set]
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -581,12 +581,10 @@ scope_type : {self._atomic.scope_type}
     def consumed_uncertain(self) -> DefaultDict[str, List[nodes.NodeNG]]:
         """
         Retrieves nodes filtered out by get_next_to_consume() that may not
-        have executed, such as statements in except blocks. Checkers that
-        want to treat the statements as executed (e.g. for unused-variable)
-        may need to add them back.
-
-        TODO: A pending PR will extend this to nodes in try blocks when
-        evaluating their corresponding except and finally blocks.
+        have executed, such as statements in except blocks, or statements
+        in try blocks (when evaluating their corresponding except and finally
+        blocks). Checkers that want to treat the statements as executed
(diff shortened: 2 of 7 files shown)


DanielNoord:
I actually had an interesting discussion about this with one of the maintainers of `pylance`:
https://github.com/microsoft/pylance-release/issues/1958

Basically it comes down to whether we want to assume all `try` statement fail or all succeed. 

Personally I'm inclined to say let's assume it is succeeds. As a user I will be more annoyed by having to add `disable`'s everywhere because one of my tools is producing false positives compared to running into exceptions because my tools are too "dumb". I prefer dumbness over potentially incorrect strictness, but this might be personal preference.

jacobtylerwalls:
Thanks for the speedy reply!

> As a user I will be more annoyed by having to add disable's everywhere because one of my tools is producing false positives

So my suggestion here for the user would be to just define the variables before the try rather than add disables.

> Basically it comes down to whether we want to assume all try statement fail or all succeed.

I agree this is a design question. I think this approach splits the difference well. We assume the try statements fail for the sake of finally statements relying on names that might not exist, which is easy for users to fix by initializing their variables, but we assume the try statements succeed for the sake of evaluating `unused-variable`, which I would agree would be quite annoying to disable if the variable is used later in the try block. How does that strike you?

I know it's not surveying opinions on the other side very well, but there's 13 +1's on the ticket. I tend to agree with the Pylance maintainer that if you're depending on a name in the finally block, then that really is an error--either something should be moved out of the try, or else the variable should be initialized before the try.

Pierre-Sassoulas:
>If you are confident that a statement will never raise an exception, you can move it outside of the `try` block

I found this argument from Eric Traut pretty convincing, honestly. By adding a statement in a Try except the user tell us that the code can fail so we should probably treat the code as possibly failing. If something never fail it should be outside the try and the resulting code will be clearer and better.

DanielNoord:
Does it make sense to create `"possibly-used-before-assignment"` for this? Like we have `"possibly-unused-variable"`? I think @Pierre-Sassoulas and I discussed at some point prior.

If not, I think we should go with the majority opinion and continue with this PR!

Pierre-Sassoulas (approved):
LGTM,! Thank you. I think that taking the same decision than pylance is the safe bet here. But I'm putting this in 2.13 so opinion can be heard about the controversial part of the issue if required.

pylint/checkers/variables.py
```diff
@@ -607,6 +609,25 @@ def get_next_to_consume(self, node):
...
+                n
+                for n in found_nodes
+                if not (
+                    isinstance(n.statement().parent, nodes.TryFinally)
+                    and n.statement() in n.statement().parent.body
+                )
+            ]
+            difference = set(found_nodes) - set(filtered_nodes)
+            self.mark_as_consumed(name, difference)
+            found_nodes = filtered_nodes
```
jacobtylerwalls, line 629: Sorry for the additional push after approval, but I think this is safer. I noticed while working on the similar issue #4761.

ChangeLog
```diff
@@ -9,6 +9,11 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+* ``used-before-assignment`` now considers that assignments in the try block
```
DanielNoord, line 12: Can you add a comment about this to the description of the message? I have been working on auto generating documentation for each individual message and it would be helpful if the documentation displayed this.

pylint/checkers/variables.py
```diff
@@ -607,6 +609,25 @@ def get_next_to_consume(self, node):
...
+        if (
+            found_nodes
+            and isinstance(node.statement().parent, nodes.TryFinally)
+            and node.statement() in node.statement().parent.finalbody
+        ):
+            filtered_nodes = [
+                n
+                for n in found_nodes
+                if not (
+                    isinstance(n.statement().parent, nodes.TryFinally)
```
DanielNoord, line 623: Could you use `n.statement(future=True)` here? We will trigger a `DeprecationWarning` otherwise. Same for all the other `statement` calls.

tests/functional/u/use/used_before_assignment_issue85.py
```diff
@@ -0,0 +1,8 @@
+# pylint: disable=missing-module-docstring, missing-function-docstring
```
DanielNoord, line 1: Could you maybe add a link to the issue here? And perhaps even a docstring explaining that we are assuming the try failed? That might be helpful for future reference.

DanielNoord (changes requested):
Small changes here and there! Rest looks good! Glad to see such an old issue finally resolved even if I did not initially agree with the solution. I'm must admit I do think this might be the best solution for this problem going forward. Thanks @jacobtylerwalls for putting this to our attention! 😄

jacobtylerwalls:
Let's not merge until I solve #2615 with a very similar block. I think I can use the same branch.

pylint/checkers/variables.py
```diff
@@ -391,7 +391,8 @@ def _has_locals_call_after_node(stmt, scope):
     "E0601": (
         "Using variable %r before assignment",
         "used-before-assignment",
-        "Used when a local variable is accessed before its assignment.",
+        "Used when a local variable is accessed before its assignment or in "
+        "except or finally blocks without being defined before the try block.",
```
DanielNoord, line 395: Perfect, thanks!

pylint/checkers/variables.py
```diff
@@ -565,6 +566,8 @@ def mark_as_consumed(self, name, consumed_nodes):
         If all of the nodes for the name were consumed, delete the name from
         the to_consume dictionary
         """
+        if name not in self.to_consume:
```
DanielNoord, line 569: Is this change necessary because of the change to `get_next_to_consume`? I do not know why but the docstring of this method does not seem to indicate that the name not being in `self.to_consume` is an option.

I find the code of the consumption checker to be quite fragile so I'm worried this might break something that was working based on an unwritten assumption.
I don't think this change necessarily breaks anything but I wonder if this function should be called in these cases anyway or if we should have returned in an earlier method that calls `mark_as_consumed`.

jacobtylerwalls: > Is this change necessary because of the change to get_next_to_consume? 

Yes, but it will not be necessary after the refactor in #5241 plus the introduction of `consumed_uncertain` in #5402. If you want to hold both PRs for the refactor, that's fine, but if you want to merge this one I will remove the unneeded block after rebasing #5402 on #5241.

The solution in these PRs is that some variables should be partially-consumed: i.e. consumed for the purposes of not being "unused" in the except/finally blocks, or after the try; but still able to emit `unused-before-assignment` if they are used in those blocks (improperly). The old pattern unconditionally marks everything as consumed before exiting the function, so without adding a new attribute like `consumed_uncertain` to store things off, this PR had to let them be marked as consumed twice and added this short-circuit.

DanielNoord: Let's do #5241 > #5402 > #5384 then. That seems to make the most sense 😄

I'll have another good look at #5402 after #5241 has been merged. I really want to make the consumption checker as intuitive as possible seeing the large amount of issues and false positives/negatives we currently have. I believe the code being much harder to read compared to other parts of the code are a large reason why these issues stay open. These PRs are really valuable and are a good way to start improving on it, thanks for that!

Pierre-Sassoulas:
@jacobtylerwalls , we merged #5402 (🎉), but I'm not sure how to resolve the conflict, could you rebase on the latest main, please ?

pylint/checkers/variables.py
```diff
@@ -405,7 +405,8 @@ def _has_locals_call_after_node(stmt, scope):
     "E0601": (
         "Using variable %r before assignment",
         "used-before-assignment",
-        "Used when a local variable is accessed before its assignment.",
+        "Used when a local variable is accessed before its assignment or in "
```
DanielNoord, line 408:
```suggestion
        "Used when a local variable is accessed before its assignment or within "
```

pylint/checkers/variables.py
```diff
@@ -662,6 +661,48 @@ def get_next_to_consume(self, node):
             self.consumed_uncertain[node.name] += difference
             found_nodes = filtered_nodes
 
+        # If this node is in a Finally block of a Try/Finally,
+        # filter out assignments in the try portion, assuming they may fail
+        if (
+            found_nodes
+            and isinstance(node.statement(future=True).parent, nodes.TryFinally)
```
DanielNoord, line 668: Let's store `node.statement` in a local variable. We're calling it 4/5 times in this block.

DanielNoord (changes requested):
Thanks! Most of this already looks good.

I thought about `node.statement` previously but never commented about it. Upon re-review I think it makes sense.

pylint/checkers/variables.py
```diff
@@ -405,7 +405,8 @@ def _has_locals_call_after_node(stmt, scope):
     "E0601": (
         "Using variable %r before assignment",
         "used-before-assignment",
-        "Used when a local variable is accessed before its assignment.",
+        "Used when a local variable is accessed before its assignment or within "
```
DanielNoord, line 408: Let's merge this in the current state and use https://github.com/PyCQA/pylint/pull/5506 to perfect this description 😄

DanielNoord (approved):

pylint/checkers/variables.py
```diff
@@ -405,7 +405,11 @@ def _has_locals_call_after_node(stmt, scope):
     "E0601": (
         "Using variable %r before assignment",
         "used-before-assignment",
-        "Used when a local variable is accessed before its assignment.",
+        "Emitted when a local variable is accessed before its assignment took place. "
+        "Assignments in try blocks are assumed not to have occurred when evaluating "
+        "associated except/finally blocks. Assignments in except blocks are assumed "
+        "not to have occurred, except when the associated try block contains a return "
```
DanielNoord, line 411: This last part is not actually true for this PR right? Might as well remove it and keep the git history accurate.

PR #5386, opened 2021-11-24
Add ``min_pyver_end_position`` option

DanielNoord:

## Type of Changes

## Description

Ref https://github.com/PyCQA/pylint/pull/5349#discussion_r755315222.

### DIFF
--- a/doc/development_guide/testing.rst
+++ b/doc/development_guide/testing.rst
@@ -72,6 +72,15 @@ You can also use ``# +n: [`` with n an integer if the above syntax would make th
 
 If you need special control over Pylint's configuration, you can also create a .rc file, which
 can have sections of Pylint's configuration.
+The .rc file can also contain a section ``[testoptions]`` to pass options for the functional
+test runner. The following options are currently supported:
+
+    "min_pyver": Minimal python version required to run the test
+    "max_pyver": Maximum python version required to run the test
+    "min_pyver_end_position": Minimal python version required to check the end_line and end_column attributes of the message
+    "requires": Packages required to be installed locally to run the test
+    "except_implementations": List of python implementations on which the test should not run
+    "exclude_platforms": List of operating systems on which the test should not run
 
 During development, it's sometimes helpful to run all functional tests in your
 current environment in order to have faster feedback. Run from Pylint root directory with::
--- a/pylint/testutils/lint_module_test.py
+++ b/pylint/testutils/lint_module_test.py
@@ -159,18 +159,23 @@ class LintModuleTest:
             return open(self._test_file.source, encoding="latin1")
         return open(self._test_file.source, encoding="utf8")
 
-    def _get_expected(self) -> Tuple["MessageCounter", List[OutputLine]]:
+    def _get_expected(
+        self, check_end_position: bool
+    ) -> Tuple["MessageCounter", List[OutputLine]]:
         with self._open_source_file() as f:
             expected_msgs = self.get_expected_messages(f)
         if not expected_msgs:
             expected_msgs = Counter()
         with self._open_expected_file() as f:
             expected_output_lines = [
-                OutputLine.from_csv(row) for row in csv.reader(f, "test")
+                OutputLine.from_csv(row, check_end_position)
+                for row in csv.reader(f, "test")
             ]
         return expected_msgs, expected_output_lines
 
-    def _get_actual(self) -> Tuple["MessageCounter", List[OutputLine]]:
+    def _get_actual(
+        self, check_end_position: bool
+    ) -> Tuple["MessageCounter", List[OutputLine]]:
         messages: List[Message] = self._linter.reporter.messages
         messages.sort(key=lambda m: (m.line, m.symbol, m.msg))
         received_msgs: "MessageCounter" = Counter()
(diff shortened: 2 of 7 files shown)


pylint/testutils/lint_module_test.py
```diff
@@ -159,18 +159,23 @@ def _open_source_file(self) -> TextIO:
             return open(self._test_file.source, encoding="latin1")
         return open(self._test_file.source, encoding="utf8")
 
-    def _get_expected(self) -> Tuple["MessageCounter", List[OutputLine]]:
+    def _get_expected(
+        self, check_end_position: bool
+    ) -> Tuple["MessageCounter", List[OutputLine]]:
```
Pierre-Sassoulas, line 164:
```suggestion
    ) -> TestMessages:
```
We could maybe create an alias for `` Tuple["MessageCounter", List[OutputLine]]``. Finding a name is hard.

pylint/testutils/lint_module_test.py
```diff
@@ -180,15 +185,18 @@ def _get_actual(self) -> Tuple["MessageCounter", List[OutputLine]]:
                 msg.symbol != "fatal"
             ), f"Pylint analysis failed because of '{msg.msg}'"
             received_msgs[msg.line, msg.symbol] += 1
-            received_output_lines.append(OutputLine.from_msg(msg))
+            received_output_lines.append(OutputLine.from_msg(msg, check_end_position))
         return received_msgs, received_output_lines
```
Pierre-Sassoulas, line 190:
```suggestion

    def should_check_end_position(self):
        """Should we check the end position ?"""
        return sys.version_info >= self._test_file.options["min_pyver_end_position"]

```

Could we create a function to avoid changing the signature and passing the ``check_end_position`` around ?

Pierre-Sassoulas (commented):
Look good already, I have some minor remarks.

pylint/testutils/lint_module_test.py
DanielNoord: I thought that doing this once when loading the file was better than doing it for each individual `OutputLine`, that's why I pass it around.

However, I will add a default value so this doesn't become a breaking change.

pylint/testutils/lint_module_test.py
DanielNoord: `MessageCounter` is already an alias. Personally I think this is just under the limit of it being too long to read. Especially since it is on a new line anyway.

pylint/testutils/lint_module_test.py
Pierre-Sassoulas: Maybe we can create an attribute in the constructor ? cached property is available only from 3.8 on https://docs.python.org/3.8/library/functools.html?highlight=#functools.cached_property

pylint/testutils/lint_module_test.py
Pierre-Sassoulas: Fair enough.

pylint/testutils/lint_module_test.py
DanielNoord: https://github.com/PyCQA/pylint/pull/5386/commits/602b8ce9abc22c8fb7b88e4cb57be358d5b73a24

doc/development_guide/testing.rst
```diff
@@ -72,6 +72,15 @@ You can also use ``# +n: [`` with n an integer if the above syntax would make th
 
 If you need special control over Pylint's configuration, you can also create a .rc file, which
 can have sections of Pylint's configuration.
+The .rc file can also contain a section ``[testoptions]`` to pass options for the functional
+test runner. The following options are currently supported:
+
+    "min_pyver": Minimal python version required to run the test
+    "max_pyver": Maximum python version required to run the test
```
cdce8p, line 79: This isn't completely true. Whereas `min_pyver` acts inclusively, `max_pyver` is exclusive due to the way tuple comparisons and `sys.version_info` work in Python.

I.e. an alpha release would be `(3, 11, 0, 'alpha', 0) > (3, 11)`. Thus setting  `max_pyver = 3.11` effectively excludes `3.11` and above.

IMO that is the desired behavior as otherwise one would need to say something like `3.10.99` to  exclude `3.11+`. So just the documentation that should get updated.

https://github.com/PyCQA/pylint/blob/1f7f2e96c39ee89ee9fb18ab23f3dfbb9325c870/pylint/testutils/lint_module_test.py#L92-L96

cdce8p (commented):
I didn't had much time the last few days. Just trying to catch up now.

`min_pyver_end_position` was exactly what I though off with my proposal. At the moment I agree that we don't need it. However there are still some AST node types left that don't have position information and just might get  some in a future release.

doc/development_guide/testing.rst
DanielNoord: Thanks @cdce8p for reviewing even after we already merged, much appreciated!

I'll create a PR that updates the doc.

DanielNoord: @cdce8p Any good ideas about how to describe this. I have been trying for 5 minutes but can't come up with a simple but correct sentence 😆

Pierre-Sassoulas: `` "max_pyver": Starting from that python version for the test won't be run`` ?

DanielNoord: See https://github.com/PyCQA/pylint/pull/5411

Issue #5306, opened 2021-11-15
Follow-up to primer tests for pylint 2.12.2


Pierre-Sassoulas:
- [x] #5315 
- [ ] #5364
- [x] #5387 -> Create ``primer`` marker with ``addoption``. This allows removal of ``ignore=primer`` in settings and full discovery of tests in IDEs. See https://docs.pytest.org/en/6.2.x/example/simple.html -> 
- [x] Improve error message and provide information of error location on fail
- [x] Stop running all CI jobs if one jobs fails.
- [x] #5387 -> Add documentation
- [x] #5359 First step done in https://github.com/PyCQA/pylint/pull/5425

Pierre-Sassoulas:
@DanielNoord I would put the better primer in 2.13 at least, even 2.14 for false positive detection, what do you think ?

DanielNoord:
I think we can close this, make quality of life improvements for `2.13` and add false positive detection in `2.14` and beyond. I'll update the `2.13` issue with some other things I thought of!

Pierre-Sassoulas:
Done in #5425

PR #5387, opened 2021-11-24
Add documentation for primer and convert to command line option

DanielNoord:

## Type of Changes

## Description

Based on https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

Because we can remove the standard ``addopts`` IDEs like VS Code will now actually find all tests in our test suite, but will just skip them when running normally. Makes it easier (at least for those on VS Code) to explore all tests.

Ticks of two points from #5306

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -508,7 +508,7 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          pytest --primer-stdlib -n auto
+          pytest -m primer_stdlib --primer-stdlib -n auto
 
   pytest-primer-external:
     name: Run primer tests on external libs Python ${{ matrix.python-version }} (Linux)
@@ -542,4 +542,4 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          pytest --primer-external -n auto
+          pytest -m primer_external --primer-external -n auto
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -508,7 +508,7 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          pytest --primer_stdlib -n auto
+          pytest --primer-stdlib -n auto
 
   pytest-primer-external:
     name: Run primer tests on external libs Python ${{ matrix.python-version }} (Linux)
@@ -542,4 +542,4 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          pytest --primer_external -n auto
+          pytest --primer-external -n auto
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -508,7 +508,7 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          pytest --primer_stdlib
+          pytest --primer_stdlib -n auto
 
   pytest-primer-external:
     name: Run primer tests on external libs Python ${{ matrix.python-version }} (Linux)
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -476,7 +476,7 @@ jobs:
           . venv/bin/activate
           pytest --benchmark-disable tests/
 
-  pytest-primer-stlib:
+  pytest-primer-stdlib:
     name: Run primer tests on stdlib Python ${{ matrix.python-version }} (Linux)
     runs-on: ubuntu-latest
     needs: prepare-tests-linux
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -508,7 +508,7 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          pytest -m primer_stdlib -n auto
+          pytest --primer_stdlib
```
DanielNoord, line 511:
```suggestion
          pytest --primer_stdlib -n auto
```

tests/conftest.py
```diff
@@ -56,3 +56,39 @@ def disable():
 @pytest.fixture(scope="module")
 def reporter():
     return MinimalTestReporter
+
+
+def pytest_addoption(parser) -> None:
+    parser.addoption(
+        "--primer_stdlib",
```
Pierre-Sassoulas, line 63:
```suggestion
        "--primer-stdlib",
```

tests/conftest.py
```diff
@@ -56,3 +56,39 @@ def disable():
...
+        action="store_true",
+        default=False,
+        help="Run primer stdlib tests",
+    )
+    parser.addoption(
+        "--primer_external",
```
Pierre-Sassoulas, line 69:
```suggestion
        "--primer-external",
```
There's a lot of other place to change, but it's hard to do in github.

tests/conftest.py
DanielNoord: You can't mark with `@pytest.mark.primer_stdlib` so I thought I would keep consistency throughout the functions using it. Do you want me to change it anyway?

Pierre-Sassoulas: For CLI option I think the standard is ``-``

doc/development_guide/testing.rst
```diff
@@ -129,6 +129,19 @@ and should exit with exit code 2 the ``.out`` file should be named ``bad_configu
...
 
+Primer tests
+-------------------------------------------
+
+Pylint also uses what we refer to as ``primer`` tests. These are tests that are ran automatically
+in our Continuous Integration and check whether any changes in Pylint lead to crashes or fatal errors
+on the ``stdlib`` and a selection of external repositories.
+This list of repositories is selected on the basis of three criteria: 1) projects need to use a diverse
+range of language features, 2) projects need to be well maintained and 3) projects should not have a codebase
+that is too repeitive. This guarantees a good balance between speed of our CI and finding potential bugs.
```
jacobtylerwalls, line 140:
```suggestion
that is too repetitive. This guarantees a good balance between speed of our CI and finding potential bugs.
```

doc/development_guide/testing.rst
```diff
@@ -129,6 +129,19 @@ and should exit with exit code 2 the ``.out`` file should be named ``bad_configu
 The content of the ``.out`` file should have a similar pattern as a normal Pylint output. Note that the
 module name should be ``{abspath}`` and the file name ``{relpath}``.
 
+Primer tests
+-------------------------------------------
+
+Pylint also uses what we refer to as ``primer`` tests. These are tests that are ran automatically
```
jacobtylerwalls, line 135:
```suggestion
Pylint also uses what we refer to as ``primer`` tests. These are tests that are run automatically
```

doc/development_guide/testing.rst
DanielNoord: 😅

DanielNoord:
We still need to add the mark to the `pytest` command as otherwise the CI job will run all tests including the `primer`. I think the latest commit fixes the increased run time of `primer` in this PR.

Pierre-Sassoulas:
> We still need to add the mark to the pytest command as otherwise the CI job will run all tests including the primer

Wasn't the point of adopt was that primer were not launched by default ? What am I missing ?

DanielNoord:
> We still need to add the mark to the `pytest` command as otherwise the CI job will run all tests including the `primer`. I think the latest commit fixes the increased run time of `primer` in this PR.

This has been fixed with https://github.com/PyCQA/pylint/pull/5387/commits/dab1e030f277d4a894802d5c826ad0ac62053415

> Wasn't the point of adopt that primer were not launched by default ? What am I missing ?

With this implementation ``--primer-external`` is ``False`` by default. It looks in the ``item.keywords`` (decorators) and if ``primer_stdlib`` is in there it adds a ``skip``. If the option is supplied this won't happen so the mark isn't skipped. We need to add `-m primer_external` to then tell it to only run the marked tests again.

Basically instead of not selecting the mark by default (and therefore not showing up in test collection) this makes it so the mark is skipped by default.

PR #5374, opened 2021-11-22
Migrate test for extension to functional tests

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Migrate all the extensions tests that are reading files to the functional tests that permits to test more and are easier to understand.

Refactor prior to #5369

Issue #4574, opened 2021-06-14
Regression: --enable config is being ignored


itamarst:
### Steps to reproduce

Let's say you have `example.py`:

```python
l = []
for x in range(10):
    l.append(lambda: x + 10)
```

I only want to run `cell-var-from-loop` check.

If I run `pylint` with no config, it's fine:

```
$ pylint example.py 
************* Module example
example.py:1:0: C0114: Missing module docstring (missing-module-docstring)
example.py:3:21: W0640: Cell variable x defined in loop (cell-var-from-loop)

-------------------------------------------------------------------
Your code has been rated at 3.33/10 (previous run: 10.00/10, -6.67)
```

### Current behavior

If I run pylint in a way that is supposed to only check `cell-var-from-loop` it reports... no problems!

```
$ pylint --disable=all --enable=cell-var-from-loop example.py 

-------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 3.33/10, +6.67)
```

The `--enable` appears to be ignored.

### Expected behavior

I would expect the `cell-var-from-loop` check to be run. The same thing happens with `W0640`.

### pylint --version output

Result of `pylint --version` output:

```
$ pylint --version
pylint 2.8.3
astroid 2.5.8
Python 3.6.13 (default, Feb 16 2021, 00:00:00) 
[GCC 10.2.1 20201125 (Red Hat 10.2.1-9)]
```

itamarst:
Pylint 2.7, 2.6, 2.5 have the same issue. Pylint 2.4 does the right thing:

```
$ pylint --version
pylint 2.4.4
astroid 2.3.3
Python 3.6.13 (default, Feb 16 2021, 00:00:00) 
[GCC 10.2.1 20201125 (Red Hat 10.2.1-9)]
$ pylint --disable=all --enable=cell-var-from-loop example.py 
************* Module example
example.py:3:21: W0640: Cell variable x defined in loop (cell-var-from-loop)

-------------------------------------------------------------------
Your code has been rated at 6.67/10 (previous run: 10.00/10, -3.33)
```

Pierre-Sassoulas:
Hello, thank you for reporting this issue and starting to analyse the problem.

Pierre-Sassoulas:
I did a git bisect, the regression appeared in c5181886b59799740306829e0e21e548c50db391

jacobtylerwalls:
[...]

PR #5396, opened 2021-11-25
Create separate documentation page for each message

### LINKED ISSUE #1056: Document each rule separately
We can do a better job at documenting our rules. Currently we're employing an automatic approach, in which we generate a restructuredText document from the help messages of the rules, but this means we can't explain them properly, we can't put actual examples in the help messages and we can't version them (when they were added, removed etc)

We can take same inspiration from how the folks from ESLint are doing things: http://eslint.org/docs/rules/

### LINKED ISSUE #2235: Authoritative overview-page for all messages
Whenever I use pylint and encounter a message/code I have not seen before, I like to understand it's background/reasoning.
Therefore I study its short description and if that is not enough, I look it up on a search engine of my choice and mostly find the following:
* Discussions on stackoverflow about this message (if it makes sense at all, why its good/bad, personal opinions)
* [Pylints feature page](http://pylint.pycqa.org/en/latest/technical_reference/features.html)
* Github issues about this message (here)
* Blogposts about similar or this issue
* Description for some messages on http://pylint-messages.wikidot.com/all-messages

Scanning through all the above sources is quite some work and in most cases does not give me what I want exactly and not in a convenient way.
This is because:
* Stackoverflow is too opinion-based and does not cover all messages
* The Github issue on pylint for this message contains too much noise (i.e. discussion on technical implementation of this message)
* Blogposts don't focus on exactly this message and tend to babble too much and is also very different on different messages, because its written by different people at different times (outdated!)
* pylint-messages.wikidot.com is actually quite nice in terms of the description and structure, but lacks some/most of the messages, is sometimes also just the short-description from our source here and is (afaik) not directly managed by this community, which does not make it authoritive. In addition to that it's layout feels outdated.

What I therefore suggest is to add one authoritive, official PyCQA/pylint [github-page](https://pages.github.com/) or a rst/markdown document that has elaborate descriptions/background for each message and is nicely formatted and up-to-date. Maybe even integrated with "msgs".
Something in the direction of [quantifiedcode](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html)

This would especially help Python beginners and people who want to understand these rules/suggestions more thoroughly, instead of just blindly following/accepting them.

If there is a consensus that something like this would be beneficial, I'd offer to start the draft.

If you know of anything that already exists and covers this, please tell.

DanielNoord:

## Type of Changes

## Description

Closes #1056, Closes #2235

Little thing I have worked on. Pages can definitely be improved but I think this is a good start and will help direct users to our documentation via google etc.


Pierre-Sassoulas:
Did you know about https://github.com/vald-phoenix/pylint-errors ? Pretty cool project even if incomplete. Result : https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/C0102. Often it's the first result when searching for a particular message.

@DanielNoord What would you think about integrating this seriously in this MR ? @vald-phoenix what is your opinion about being integrated in pylint directly ?

DanielNoord:
I think there are certainly benefits to be achieved when we can combine the two!

cdce8p (commented):
Some notes, although I didn't review the PR.

* I'm not sure that it makes sense to create separate pages for the different message categories. As a user I would much rather prefer one long list (maybe with sections) that contains all. Then use `search` to find the one I'm looking for.
* This PR does not change it so it can be ignored. I just wanted to mentioned that it probably makes sense to keep the old pages around. That way we don't invalidate existing links.
* I would also recommend some more discussion about the link template for individual messages. It would be good to get them right the first time as they will be difficult to change later. Some things to consider
  * What if we rename the message? It should probably be accessible via the old **and** new one.
  * If the id is used, what if that changes?
  * Should the checker name be part of the url template?
  * Does it make sense for the severity to be part of it?
* I like the use of a custom `sphinx` extension. I believe I've previously seen code related to how settings and error message are displayed in the docs directly in pylint. That might be something to look into when it comes time to refactor those areas. Moving them to an extension as well could be beneficial.

DanielNoord:
> * I'm not sure that it makes sense to create separate pages for the different message categories. As a user I would much rather prefer one long list (maybe with sections) that contains all. Then use `search` to find the one I'm looking for.

I will change this.

> * This PR does not change it so it can be ignored. I just wanted to mentioned that it probably makes sense to keep the old pages around. That way we don't invalidate existing links.

👍 

> * I would also recommend some more discussion about the link template for individual messages. It would be good to get them right the first time as they will be difficult to change later. Some things to consider
>   
>   * What if we rename the message? It should probably be accessible via the old **and** new one.

Can we make redirects in `read the docs`? We could add redirects for old names. Although for `missing-docstring` this might be troublesome as we would need to redirect to 4+ pages. I'll have a look at the possibilities there.

>   * If the id is used, what if that changes?

This would require us to mark id's as being used even if they have changed. I don't think we currently store that information, we only store the name for `old-names`. Do we want to start doing that from now on?

>   * Should the checker name be part of the url template?

I don't think so? Even more things to keep track of when deprecating or changing stuff. Normally as user doesn't know which checker class emitted a warning anyway. I included it mostly for development purposes.

>   * Does it make sense for the severity to be part of it?

I'm not sure. Would it give users much more information besides what they currently have? I don't think it will help them with finding the page quicker than they can now, but it won't allow us to easily change the severity of messages. Including it on the page itself, rather than only as part of the id is something we could do though.

> * I like the use of a custom `sphinx` extension. I believe I've previously seen code related to how settings and error message are displayed in the docs directly in pylint. That might be something to look into when it comes time to refactor those areas. Moving them to an extension as well could be beneficial.

There are two other extensions, although they are not as large as this one. I hope with the move to `argparse` (or another library) we can also easily generate the documentation for the settings as well 😄

Pierre-Sassoulas:
> Can we make redirects in read the docs?

We can but it's manual and I'd rather not have to do manual configuration for each name change.

> If the id is used, what if that changes?

I don't understand the question, if an id existed at any point, it's a current ID or an old names, so 'id' are always "used" once they are added to pylint.

> Should the checker name be part of the url template?

Let's keep this simple for us so we can move messages around internally if we need to for perfirmance as well as cleanliness. Some issues are super hard to fix because of messages being mixed together, we need to be able to burst the checkers (https://github.com/PyCQA/pylint/issues/4574)

> I hope with the move to argparse (or another library) we can also easily generate the documentation for the settings as well

🙏 

Regarding the template I like what @pheanex did here : https://pheanex.github.io/pylint/#R1705 I think it's the basis for  https://github.com/vald-phoenix/pylint-errors too. I think we can add old_names handling to it and the work intensive part will be a proper description for all message. @vald-phoenix how would you feel if we take the current content of your repository for a start ?

DanielNoord:
> > Can we make redirects in read the docs?
> 
> We can but it's manual and I'd rather not have to do manual configuration for each name change.

Can't we pass it to `Sphinx` or `read the docs` during the creating process? That would allow doing to form `old-names`.

> I don't understand the question, if an id existed at any point, it's a current ID or an old names, so 'id' are always "used" once they are added to pylint.

Ah, I forgot id's are also stored in old names. That would solve this problem as we can just make it part of the redirect (if we can make it work).

> Regarding the template I like what @pheanex did here : https://pheanex.github.io/pylint/#R1705 I think it's the basis for https://github.com/vald-phoenix/pylint-errors too. I think we can add old_names handling to it and the work intensive part will be a proper description for all message. @vald-phoenix how would you feel if we take the current content of your repository for a start ?

I'm not sure how easy it would be to transform those `.md` file to `.rst`. I have been thinking about extending this to allow storing information about messages in a separate directory that this extension would read from. For storing longer descriptions or explanations `json` files would work well and can be easily loaded. For code blocks `json` won't work. We could do something with a directory for each message which includes `extra-information.json`, `good-code.rst`, bad-code.rst`. Those good and bad code files can be parsed programatically from @valid-phoenix's project if they agree on merging his project.

Pierre-Sassoulas:
> We could do something with a directory for each message which includes extra-information.json, good-code.rst, bad-code.rst`. 

Wild idea, but maybe we could take the good example / bad example from functional tests and convert it to ``.rst``. It would force (some) standardization for functional test too. Putting the whole functional test content is not desirable but requiring the basic example with good code bad code would nudge us to have some structure:

Ie: having a ``nan-comparison`` message requires a ``tests/functional/n/nan_comparison/`` directory containing:
```
tests/functional/n/nan_comparison/good.py
tests/functional/n/nan_comparison/bad.py
tests/functional/n/nan_comparison/bad.txt
tests/functional/n/nan_comparison/full.py
tests/functional/n/nan_comparison/full.rc
tests/functional/n/nan_comparison/full.txt
tests/functional/n/nan_comparison/issue_49569.py
tests/functional/n/nan_comparison/issue_49569.txt
```
The ``full.py`` or ``issue_49569``  would contains the current functional tests for ``nan-comparison`` but would not be required or used for the documentation, only good and bad would.

The ``tests/functional/ext`` directory is close to that following https://github.com/PyCQA/pylint/pull/5374 (there is a directory with the extension name like ``tests/functional/ext/bad_builtin/``)

DanielNoord:
I like the idea but think it would be better to do another PR for it. Let's focus this PR on:

- [ ] Make one page with links to all messages
- [ ] Approve initial page template with minimal required information
- [ ] Approve links to individual pages
- [ ] Set up system for redirection for old names

cdce8p:
> > We could do something with a directory for each message which includes extra-information.json, good-code.rst, bad-code.rst`.
> 
> Wild idea, but maybe we could take the good example / bad example from functional tests and convert it to `.rst`. It would force (some) standardization for functional test too. Putting the whole functional test content is not desirable but requiring the basic example with good code bad code would nudge us to have some structure:
> 
> Ie: having a `nan-comparison` message requires a `tests/functional/n/nan_comparison/` directory containing:
> 
> ```
> tests/functional/n/nan_comparison/good.py
> tests/functional/n/nan_comparison/bad.py
> tests/functional/n/nan_comparison/bad.txt
> tests/functional/n/nan_comparison/full.py
> tests/functional/n/nan_comparison/full.rc
> tests/functional/n/nan_comparison/full.txt
> tests/functional/n/nan_comparison/issue_49569.py
> tests/functional/n/nan_comparison/issue_49569.txt
> ```

@Pierre-Sassoulas I would **not** recommend this. It just adds too much additional complexity to the tests.

Different idea. In essence, this PR will already create separate page for each message. We could add those pages to git and manually add good examples were needed. The sphinx extension would only need to be able to handle custom sections in existing message rst files and don't delete them when updating.

That would also allow us to add additional explainations if necessary beyond those in the message dict.

DanielNoord:
> Different idea. In essence, this PR will already create separate page for each message. We could add those pages to git and manually add good examples were needed. The sphinx extension would only need to be able to handle custom sections in existing message rst files and don't delete them when updating.
> 
> That would also allow us to add additional explainations if necessary beyond those in the message dict.

I don't think this is a good idea. The other extension does this as well (ie, a base file that's overwritten by the extension) and this is really annoying to work with when improving or working on it. The extensions can be run as a script to see if the `.rst` files they create are correct and this would mean you would need to revert your changes every time you do a test run.
That doesn't sound too time consuming, but while working on this it became annoying pretty fast to have to constantly revert changes to some files while keeping the changes to the extensions files. It's much easier if we can just ignore all files in the `doc/messages/` directory.

If we don't want to add these files to the `tests` directory I would propose to create a new `message_pages_data` directory somewhere in `docs` where we store some type of data to be used by the extension to create the `.rst` files.

Pierre-Sassoulas:
> It's much easier if we can just ignore all files in the doc/messages/ directory.

I agree with @DanielNoord here, I would favor a fully generated documentation based on files that we need to define now, preferably with no duplication. I don't think it's "additional complexity", it's "stricter structure" and "more automation". Forcing to have a functional test with a good/bad example in a per-determined directory enforced by tests will make some work at first (Some messages do not have functional tests.)  then will automate most of the review and guarantee that we'll have an understandable example and a proper documentation. Re-using the functional test will make sure that the example is working and that there is no duplication between doc/tests. Also, I feel like I've spent hours of my life asking for functional tests in reviews 😄 

Saw we want to document ``bad-option-value`` from Pylinter, we'd have.
- ``tests/functional/b/bad-option-value/`` or ``tests/functional/pylinter/bad-option-value/`` and inside:
    - ``good.py`` exists.
    - ``good.txt`` does not exists.
    - ``bad.py`` exists.
    - ``bad.txt`` exists.
    - More complete functional tests (but it's not mandatory)
- ``doc/user_guide/messages/bad-option-value.json`` or ``doc/user_guide/messages/pylinter/bad-option-value.json`` exists and contain information about the message stack-overflow question explaining it, merge request that introduced it, merge request that modified it, others (?)

Then we create a script / sphinx extension to generate the final ``.rst`` using the message description from the message store, the functional test and the doc additional information.

In the long run, the only things that would be more complicated than right now is that you have to create the json and make sure that you have two files in the functional tests containing a small understandable example.

vald-phoenix:
@Pierre-Sassoulas if pylint-errors lib will help pylint then you can do whatever you want with it.
This is a dirty script I used to extract messages.
https://gist.github.com/vald-phoenix/8b05f5ffbf27225c36274d78b9cb930d
I used `pandoc` to convert rst files into markdown. I think you can do the reverse process as well.

DanielNoord:
Sorry for the commit spam. My local environment doesn't really like running `sphinx` for some reason, so I need the `read-the-docs` action to check changes.

I have added a system to redirect for old names. It can be seen in action for `missing-docstring` here:
https://pylint--5396.org.readthedocs.build/en/5396/messages/convention/missing-docstring.html

Since all pages need to be referenced in an index, I have also added the renamed messages to the general messages list.
I wonder if we need the introduction pages anymore. I don't know what else to put on it and right now it is a bit empty. Anyone got any good ideas about what to add to the "messages introduction" page?

As I said, I think this PR should focus on:
- Make one page with links to all messages
-  Approve initial page template with minimal required information
-  Approve links to individual pages
-  Set up system for redirection for old names

These have now all been added so this is ready for review. A follow up issue/PR should then deal with code examples and how we are going to save those.

Pierre-Sassoulas:
>  My local environment doesn't really like running sphinx for some reason

Did you try ``tox -e docs`` ? I'm going to review once pylint 2.12.2 is released.

cdce8p:
Didn't get to it earlier. I wanted to add one more comment about the separate test files.

I've stated to understand your intention for it. Although I would suggest to limit the scale. The existing functional tests have a logic on their own which I would prefer not to change. Furthermore, some files check multiple different, but related messages. Breaking that up would just complicate things.

Instead, what about create a separate test directory for doc examples? That way we could still validate them while keeping the existing once. Another benefit, the doc examples could be more precise and to the point. Eg. there isn't really a need to have regression tests in the there. At first glance this might sound like duplication, but I'm not sure we even need examples for every message. Most messages are pretty self explanatory.

Pierre-Sassoulas:
> what about create a separate test directory for doc examples?

This is a lot clearer than what I had in mind. Let's do this. I started to enforce one message / one functional test directory and there are a lot of exception (pylint / astroid crash type error for example, we hope there are none), and some messages do not have any tests. Which is a problem that we probably want to fix but it's also time consuming. And as you said some functional tests are multiple message mixed together which is not worth fixing.

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,208 @@
...
+def _register_all_checkers_and_extensions(linter: PyLinter) -> None:
+    """Registers all checkers and extensions found in the default folders."""
+    initialize_checkers(linter)
+    initialize_extensions(linter)
+
+
+def _get_all_messages(
+    linter: PyLinter,
+) -> Tuple[MessagesDict, OldMessagesDict]:
+    """Get all messages registered to a linter and return a dictionary indexed by message
```
Pierre-Sassoulas, line 58: We should use the ``MessageDefinitionStore`` or the ``MessageIdStore``, if we need more information about the checker or something else we should add the information in MessageDefinitionStore directly, not in a helper function. (ie : ``for message in linter.msgs_store.messages`` should work)

DanielNoord: https://github.com/PyCQA/pylint/pull/5396/commits/f0414107e6e76f9b2861d0bc9a88417c778419b1

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,207 @@
...
+from pylint.utils import get_rst_title
+
+# PACKAGE/docs/exts/pylint_extensions.py --> PACKAGE/
+PYLINT_BASE_PATH = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+)
+"""Base path to the project folder"""
+
+
+MSG_TYPES = {
```
Pierre-Sassoulas, line 25: ``MSG_TYPES`` can be imported from ``pylint.constants``

doc/exts/pylint_messages.py
Pierre-Sassoulas: There is a protected ``self._msgs_by_category`` in the ``MessageIdStore``, by making it public and maybe handling old names, we can remove most of the helper code in this file.

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,207 @@
...
+        for message in messages:
+            messages_file = os.path.join(category_dir, f"{message.name}.rst")
+            with open(messages_file, "w", encoding="utf-8") as stream:
+                stream.write(f".. _{message.name}:\n\n")
+                stream.write(get_rst_title(f"{message.name} / {message.id}", "="))
+                stream.write("**Message emitted:**\n\n")
+                stream.write(f"{message.definition.msg}\n\n")
+                stream.write("**Description:**\n\n")
+                stream.write(f"*{message.definition.description}*\n\n")
+                stream.write(f"Created by ``{message.checker}`` checker\n")
```
Pierre-Sassoulas, line 109:
```suggestion
                title=get_rst_title(f"{message.name} / {message.id}", "=")
                stream.write(f""".. _{message.name}:

{title}    
**Message emitted:**

{message.definition.msg}

**Description:**

*{message.definition.description}*

                                                                                                                                                                                                                                Created by ``{message.checker}`` checker                                                                                                                                                                                                                                
"""
```
I think it's easier to see the formatting of a big text when using ``"""``

Pierre-Sassoulas (commented):
We can make ``pylint.message`` classes better if you need helpers that do not exists yet. It should contains all the information regarding the message we just need to add new API to get it :)

doc/exts/pylint_messages.py
DanielNoord: `MSG_TYPES` in `pylint.constants` uses `info` instead of `information`. I checked and we tend to use `information-level` in our changelogs and docs so I thought it would be best to do so here as well. That's why I created a new constant.

doc/exts/pylint_messages.py
DanielNoord: I'm not sure if we should do that (as then plugin creators can use it as well and we need to support it). A little extra helper code doesn't hurt right? This is only run once every CI and every push to main. It doesn't take too long.

doc/exts/pylint_messages.py
DanielNoord: Good idea!

DanielNoord: See https://github.com/PyCQA/pylint/pull/5396/commits/0b1e20b78f09c85fa87ee043050714693b242e9e

doc/exts/pylint_messages.py
Pierre-Sassoulas: Maybe we could create the new constant from the old one (or changing the value to information in the old one is reasonable) ?

doc/exts/pylint_messages.py
Pierre-Sassoulas: Fair enough.

doc/exts/pylint_messages.py
DanielNoord: I think its part of the public API? 😅 Changing something called a constant might warrant a deprecation period haha

Pierre-Sassoulas:
```suggestion
MSG_TYPES_DOC = {k:v if v != "info" else "information" for k, v in MSG_TYPES.items()}
```

pylint/utils/utils.py
```diff
@@ -197,6 +197,10 @@ def register_plugins(linter, directory):
                 if hasattr(module, "register"):
                     module.register(linter)
                     imported[base] = 1
+                    # pylint: disable-next=fixme
+                    # TODO: Remove after deprecation of check_docs, see #5322
```
Pierre-Sassoulas, line 201:
```suggestion
                    # TODO: Remove after deprecation of check_docs, see 
                    # https://github.com/PyCQA/pylint/issues/5322
```

doc/messages/messages_list.rst
```diff
@@ -0,0 +1,498 @@
+.. _messages-list:
```
Pierre-Sassoulas, line 1: This file is generated too, right ? Maybe we need to add it to the `.gitignore` ?

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,227 @@
...
+    """Connects the extension to the Sphinx process"""
+    # Register callback at the builder-inited Sphinx event
+    # See https://www.sphinx-doc.org/en/master/extdev/appapi.html
+    app.connect("builder-inited", build_messages_pages)
+
+
+if __name__ == "__main__":
+    pass
+    # Uncomment to allow running this script by your local python interpreter
+    # build_messages_pages(None)
```
Pierre-Sassoulas, line 227: Why do we need to comment it ?

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,227 @@
...
+                    (message.symbol, MSG_TYPES_DOC[message.msgid[0]])
+                )
+
+    return messages_dict, old_messages
+
+
+def _write_message_page(messages_dict: MessagesDict) -> None:
+    """Create or overwrite the file for each message."""
+    for category, messages in messages_dict.items():
+        category_dir = os.path.join(PYLINT_BASE_PATH, "doc", "messages", category)
```
Pierre-Sassoulas, line 91:
```suggestion
        category_dir = DOC_MESSAGES_PATH / category
```

I would create a constant `DOC_MESSAGES_PATH = PYLINT_BASE_PATH / "doc" / "messages"`, this seems to be used multiple time.

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,227 @@
...
+from pylint.lint import PyLinter
+from pylint.message import MessageDefinition
+from pylint.utils import get_rst_title
+
+# PACKAGE/docs/exts/pylint_extensions.py --> PACKAGE/
+PYLINT_BASE_PATH = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+)
+"""Base path to the project folder"""
+
```
Pierre-Sassoulas, line 24: Nitpick here :)

```suggestion
from pathlib import Path

PYLINT_BASE_PATH = Path(__file__).resolve().parent.parent.parent
"""Base path to the project folder"""

DOC_MESSAGES_PATH = PYLINT_BASE_PATH / "doc" / "messages"
"""Base path to the doc folder for messages"""

```

Pierre-Sassoulas (changes requested):
This look pretty good already ! I wonder if we should separate extension's message and optional message from the default one ?

DanielNoord:
> This look pretty good already ! I wonder if we should separate extension's message and optional message from the default one ?

Don't think so. If I'm looking for messages and their description I don't really care where they are coming from I just want to know what they are telling me.

doc/messages/messages_list.rst
DanielNoord: We could, but I thought it would be good to show what the file somewhat looks like instead of only having the code.

doc/exts/pylint_messages.py
Pierre-Sassoulas: Isn't it a good thing as it makes sure that the generated doc is up to date ?

doc/messages/messages_list.rst
Pierre-Sassoulas: Do we need to think about updating it when creating a checker for example ? Or is it simply an example that will stay out of date and be replaced by sphinx ?  If we need to update it's more work, if we do not update it could be confusing. Maybe being able to look at the (previous) result in github pages is good enough ?

DanielNoord: We could update it but it shouldn't be necessary. This is similar to `doc/technical_reference/extensions.rst` which can be outdated as well.

doc/exts/pylint_messages.py
DanielNoord: That is take care of by `doc/Makefile` 😄

doc/exts/pylint_messages.py
DanielNoord: Done 👍

Pierre-Sassoulas (approved):
This is a great first step to have clear example and explanation for each messages !

Pierre-Sassoulas:
Created #5527 to follow-up, I think all point in https://github.com/PyCQA/pylint/pull/5396#issuecomment-981056141 are addressed.

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,226 @@
...
+                    (message.symbol, MSG_TYPES_DOC[message.msgid[0]])
+                )
+
+    return messages_dict, old_messages
+
+
+def _write_message_page(messages_dict: MessagesDict) -> None:
+    """Create or overwrite the file for each message."""
+    for category, messages in messages_dict.items():
+        category_dir = os.path.join(PYLINT_MESSAGES_PATH, category)
```
vald-phoenix, line 92: Maybe it's better to be consistent and use pathlib everywhere, so it will be:
```python
category_dir = PYLINT_MESSAGES_PATH / category
```

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,226 @@
...
+        if not os.path.exists(category_dir):
+            os.makedirs(category_dir)
```
vald-phoenix, line 94: same here
```python
if not category_dir.exists():
    category_dir.mkdir(parents=True, exist_ok=True)
```

doc/exts/pylint_messages.py
```diff
@@ -0,0 +1,226 @@
...
+        for message in messages:
+            messages_file = os.path.join(category_dir, f"{message.name}.rst")
+            with open(messages_file, "w", encoding="utf-8") as stream:
```
vald-phoenix, line 97:
```python
message_file = category_dir / f"{message.name}.rst"
text = f""".. _{message.name}:
{get_rst_title(f"{message.name} / {message.id}", "=")}
**Message emitted:**
{message.definition.msg}
**Description:**
*{message.definition.description}*
Created by ``{message.checker}`` checker
"""
with message_file.open("w", encoding="utf-8") as stream:
    stream.write_text(text)
```

vald-phoenix (commented):
added my two cents about formatting :)

doc/exts/pylint_messages.py
DanielNoord: Done with https://github.com/PyCQA/pylint/pull/5396/commits/8903b9570bbafadd59f7bcd19b34dd441e718d61

doc/exts/pylint_messages.py
DanielNoord: I don't really understand what you're proposing? Perhaps I'm missing something but what's better about declaring text before opening the `message_file`?

DanielNoord:
Did a force push to make sure the documentation generated by this PR is up to date with `main` and doesn't run into any issues.

Pierre-Sassoulas (approved):
Does this still need review ? I think this is a high priority MR because it does not need to wait for a release to be useful, we could be setting up GitHub pages right now.

DanielNoord:
I think @cdce8p wanted to a final check to see if he agreed on the proposed links and how we link to renamed messages. That's why I added the `Needs review` label. Other than that I think this is indeed good to go.

@Pierre-Sassoulas Perhaps this is something to name in the summary you always put at the top of a new release. Would be helpful if we got some feedback from people of what they are missing/looking for on these pages.

Pierre-Sassoulas:
> Perhaps this is something to name in the summary you always put at the top of a new release. 

Sure we're going to advertise it. Especially if we populated it with pylint-errors content it's definitely going to be an highlight ! But once it's the first result when searching for a message it's going to bring contributions too and the sooner we start "SEO" the better imo.

DanielNoord:
@cdce8p Pinging you as I think @Pierre-Sassoulas makes a fair point about SEO and we're nearing the release of `2.13`. Do you know if you'll have some time to review this in the coming days?

DanielNoord:
Blocked by #5730. Afterwards I need to remove the reference to `check_docs` in this PR.

Pierre-Sassoulas:
We merged #5730, removing the blocked label.

DanielNoord:
Code updated!

Pierre-Sassoulas:
The doc job fail becaus eof this warning: ``non-ascii-file-name.rst:12: WARNING: Inline emphasis start-string without end-string.``

DanielNoord:
Fixed! We can't allow `\n` at the end of a message description. The `.rst` becomes invalid. We could separate that into a different PR but I think it is minor enough to stay here.

Pierre-Sassoulas:
I'm going to merge, if something need to change we can deal with it in a follow-up. This is nice enough that I'm pretty sure it's good enough to be super useful and to not delay too much. I'm quite excited to have an official documentation for each message. Amazing work Daniel !

It's lost in the review comment but the follow-up is #5527.

PR #5311, opened 2021-11-15
add a checker for misleading unicode

CarliJoy:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |
| ✓   | :scroll: Docs          |

## Description
Unicode and some other ASCII characters can be used to create programs that run
much different as a human reader would expect it.

[PEP 672](https://www.python.org/dev/peps/pep-0672/) lists some examples.

The added checkers are intended to make users aware of these issues.

Some of the test might be too strict for users programming in other languages as
english. As we assume this is minority than can disable the given tests, this should
be not an issue.

Note that the codes try to be prepared for UTF-16 and UTF-32 even so Python files
can't be encoded in them yet.
There is a checker that would detect UTF-16/UTF-32 files and fail, so even the usage
of an old pylint version would not cause a "security hole" when used with new Python
version that allow these encodings.

Closes #5281

PR #5400, opened 2021-11-25
Move ``TestVariablesChecker`` to functional tests

DanielNoord:

## Type of Changes

## Description

Step one of many.


Pierre-Sassoulas (approved):
I like the fact that you did it for one file only, it makes it more manageable as there are small moving part to takes into account during review. What do you think about the time it took you to do the refactor ? Was it worth it ? If that's the case I'll open an issue to track each file we need to move.

DanielNoord:
This one was okay, but I looked at some of the other files and they might be more difficult. There are tests such as `TestImportsChecker` and `TestDeprecatedChecker`  which are not as straightforward..

Pierre-Sassoulas:
> There are tests such as TestImportsChecker and TestDeprecatedChecker which are not as straightforward..

Let's not focus too much on this then, clearly, it's not a high priority issue. I think more typing in astroid will yield better result

Pierre-Sassoulas:
Although clearly this make pylint more friendly for new contributors and removing almost 100 lines of code is obviously a very good thing..

DanielNoord:
> it's not a high priority issue.

I don't necessarily agree.

See L160 in `tests/checkers/unittest_unicode.py` in #5311. Contributors are running into issues with this and are forced to ignore `end_column`. In the end that means we will only need to update more code. I think this is actually quite high priority and the checking of `end_column` should be turned on sooner than later.

Pierre-Sassoulas:
>  In the end that means we will only need to update more code.

True, true. Also unittest are very hard to understand comparatively. For maintenance work it's pretty high on the list for sure. I was comparing the priority to the json output for 3.0 alpha that will make vscode and pycharm better or other high value issues directly affecting known bugs or highly demanded features.

Issue #4814, opened 2021-08-08
Optimizing speed of the starting and initialisation phase


Hubro:
### Current problem

Pylint is frustratingly slow, taking several seconds to lint even tiny, simple Python files. Pylint is glacially slow compared to any linting tool in any other language I have worked with. This slowness is not too big a problem for CI/CD pipelines, since in this context a few seconds doesn't matter too much and linting of multiple files can be parallelized.

However, for background linting in editors while editing (for example with [ALE](https://github.com/dense-analysis/ale)) this leads to a terrible experience of warnings popping up seconds after I've moved past a line, or warnings staying up seconds after I've resolved them.

Since Python is not a *slow* language, I assume this slowness has to do with one or more linter settings. However, since there are hundreds of settings, I have no idea where to begin with improving the performance.

My goal would be to get a pylint run on a single file to take somewhere in the ballpark of 100 ms.

### Desired solution

I would like a section in the front page README, or a separate guide linked to from the README file, with suggestions for how to improve the performance of pylint, for example disabling certain options that are particularly taxing.

Linting 80 lines of Python should not take 2.5 seconds, at the very most it should take maybe 100 ms.

### Additional context

My pylint config: https://gist.github.com/Hubro/7adba88c42e4df0706067bfe9b2cea53#file-pylintrc-ini

The file I'm currently linting: https://gist.github.com/Hubro/7adba88c42e4df0706067bfe9b2cea53#file-generic_http_mock_server-py

```
➜ \pylint --version
pylint 2.8.3
astroid 2.5.6
Python 3.8.6 (default, Jun 27 2021, 03:28:09)
[GCC 11.1.0]
```

```
➜ time \pylint --rcfile=pylintrc.ini generic_http_mock_server.py
************* Module generic_http_mock_server
generic_http_mock_server.py:28:0: C0115: Missing class docstring (missing-class-docstring)
\pylint --rcfile=pylintrc.ini   2.30s user 0.15s system 101% cpu 2.416 total
```

DudeNr33:
I ran ``pylint`` with different checkers enabled and timed the execution (once with normal output and once while using ``-f pylint.testutils.reporter_for_tests.FunctionalTestReporter`` to suppress the output):
```csv
checkerName   ;user time (with output) ;user time (no output)
async         ;2.37s                   ;2.45s
basic         ;6.02s                   ;6.01s
classes       ;4.08s                   ;4.15s
design        ;2.64s                   ;2.73s
[...]

PR #5401, opened 2021-11-25
Initial implementation of pylintd

matusvalo:
## Type of Changes

## Description

This PR introduces basic implementation of pylintd - pylint deamon running in the background. It is simmilar implementation as mypyd[1] or blackd[2]. The implementation is highly inspired by blackd approach. Pylintd is implemented as simple http server providing simple API for passing files for linting and returning back the result.

The speedup of mainly gained in following areas:
1. avoiding python initialization overhead. But this is only gained when http client is not written in python
2. avoiding pylint initialization overhead.
3. reusing already cached AST trees of already parsed modules in https://github.com/PyCQA/astroid/blob/47e860f7d5c73d53de77e7c450535e709e5ef99e/astroid/manager.py#L78. This is most prominent speedup when repeating linting of file does not trigger parsing other modules since they remain in cache. Re-linting of the same file is usual use case for editors what is the target area of pylintd.

Examples:

Following examples shows comparison between running pylint and pylintd against 2 large files in pylint. The test was executed as follows:
1. fresh instance of pylintd was executed
2. pylint command was executed against the file
3. pylintd was asked to lint the file
4. pylintd was asked again to lint the file without restarting pylintd (this shows speedup when astroid manager cache is filled)

```bash
$ cat test.sh
curl -v -H "X-File-Name:$1" -H accept:application/X-pylint-parseable localhost:8000

$ wc -l pylint/checkers/classes.py
    2415 pylint/checkers/classes.py

$ time pylint pylint/checkers/classes.py

real	0m5.146s
user	0m4.857s
sys	0m0.210s

$ time bash test.sh pylint/checkers/classes.py          # 1st run

real	0m3.920s
user	0m0.008s
sys	0m0.013s

$ time bash test.sh pylint/checkers/classes.py          # 2nd run

real	0m1.053s
user	0m0.008s
sys	0m0.011s


wc -l pylint/checkers/typecheck.py
    2026 pylint/checkers/typecheck.py

$ time pylint pylint/checkers/typecheck.py

real	0m5.243s
user	0m4.837s
sys	0m0.222s

$ time bash test.sh pylint/checkers/typecheck.py       # 1st run

real	0m4.171s
user	0m0.008s
sys	0m0.013s

$ time bash test.sh pylint/checkers/typecheck.py       # 2nd run

real	0m1.000s
user	0m0.009s
sys	0m0.013s
```

[1] https://mypy.readthedocs.io/en/stable/mypy_daemon.html
[2] https://black.readthedocs.io/en/stable/usage_and_configuration/black_as_a_server.html

This PR can partially help closing #4814


matusvalo:
Of course this PR is not finished. It is more as Proof of concept.

Pierre-Sassoulas:
https://github.com/PyCQA/astroid/issues/792 might be a problem if we want to be able to have an efficient ``pylintd``.

matusvalo:
> https://github.com/PyCQA/astroid/issues/792 might be a problem if we want to be able to have an efficient ``pylintd``.

I agree but using separate service can enable users for some workarounds like restarting pylintd service on regular basis. (I know not ideal...)

superbobry:
@Pierre-Sassoulas our experience with running pylint as a daemon suggests that leaks definitely will be a problem. Given that any leaking node retains the full AST, so memory consumption get uncomfortably high very quickly when linting large files.

matusvalo:
I have a question. How we can proceed with this PR (except unittests and fixing the failed CI)? I am willing to continue with this PR but I would like to understand what is missing in order to finilize this PR.

pylint/lint/run.py
```diff
@@ -72,13 +73,11 @@ class Run:
     def _return_one(*args):  # pylint: disable=unused-argument
         return 1
 
-    def __init__(
+    def initialize(
         self,
-        args,
+        args: list,
```
Pierre-Sassoulas, line 78:
```suggestion
        args: List,
```
Unless we import the typing from future :)

pylint/lint/run.py
```diff
@@ -358,10 +357,16 @@ def __init__(
...
             raise
+
+        return linter, args
+
+    def show_help(self, linter: PyLinter, args: list) -> None:
         if not args:
             print(linter.help())
             sys.exit(32)
 
+    def initialize_jobs(self, linter: PyLinter) -> None:
```
Pierre-Sassoulas, line 368: I like the function created here 👌

pylint/pylintd.py
```diff
@@ -0,0 +1,150 @@
...
+            linter.check_single_file_item(file_item)
+        score_value = linter.generate_reports()
+        return score_value
+
+
+if __name__ == "__main__":
+    try:
+        DaemonRun(sys.argv, 8000)
+    except KeyboardInterrupt:
+        pass
```
Pierre-Sassoulas, line 150:
```suggestion
```
I guess it's for manual tests ?

Pierre-Sassoulas (commented):
To be fair I don't know how that would be used or how it would affect pylint's user. I'm not using blackd myself so I don't really know what to expect and how to use a pylint deamon. If needs some automated tests for sure, maybe discussion with potential users ?  

It's probably aimed toward those with a big codebase that launch analysis on single file often ? Might be useful for pre-commit too as pre-commit launch multiple pylint in parallel so if we use pylintd we skip the startup time ? Does the pre-commit use case makes sense to you @matusvalo ?

cdce8p:
I mentioned it in the issue already. I think we should not move forward with it, at this time. Maybe at some point we can pick it up again. https://github.com/PyCQA/pylint/issues/4814#issuecomment-980727214

> To be fair I don't know how that would be used or how it would affect pylint's user. I'm not using blackd myself so I don't really know what to expect and how to use a pylint deamon. If needs some automated tests for sure, maybe discussion with potential users ?

If the startup time isn't too slow, most users (read IDEs, ...) will simply choose the easiest path which is likely launching it direct. They have that implemented already for other tools, no need to change it.

Tbh aside from a few outliers I haven't seen a use case for it yet.

> It's probably aimed toward those with a big codebase that launch analysis on single file often ? Might be useful for pre-commit too as pre-commit launch multiple pylint in parallel so if we use pylintd we skip the startup time ? Does the pre-commit use case makes sense to you @matusvalo ?

For pre-commit it would be much better to use pylints own multiprocessing.

matusvalo:
I agree just partially . Calling pylintd using curl with empty cache gives ~**24%** speedup - 0m5.146s vs 0m3.920s which I consider significant - see my test above. Regarding usage of pylintd, editors can spawn pylintd and use it's HTTP API instead of calling pylint command. 

On the other hand, I think we need also input from developers of editors/plugins whether it make sense or they would use such API.

Pierre-Sassoulas:
>  Regarding usage of pylintd, editors can spawn pylintd and use it's HTTP API instead of calling pylint command.
>  I think we need also input from developers of editors/plugins whether it make sense or they would use such API.

Seems like the potential user are plugin editor then. A 25% perf gain is huge imo. But we'll have to handle caching properly in astroid or it won't last long though. I think we could move this MR to 3.0 alpha version and communicate with pylint-pycharm and vscode about it in order to see what they need and how it should be implemented. Do you want to lead the communication and specification effort @matusvalo ?

matusvalo:
> Do you want to lead the communication and specification effort @matusvalo ?

Yes, I can take this activity on my shoulders. But unfortunately, I was playing more with the implementation and I found out that the leak is pretty serious (~1 GB allocated memory after 10 runs of linting the same file). Moreover it seems that even the performance decreases over time :-/. I suppose, we need to progress with PyCQA/astroid#792 in order to have functional pylintd.

Btw. just side not around the usability of the pylintd - I just came across that blackd has already some plugins [1]. This is kind of usage of pylintd I imagine.

[1] https://plugins.jetbrains.com/plugin/14321-blackconnect

cdce8p:
> > Do you want to lead the communication and specification effort @matusvalo ?
> 
> Yes, I can take this activity on my shoulders. But unfortunately, I was playing more with the implementation and I found out that the leak is pretty serious (~1 GB allocated memory after 10 runs of linting the same file). Moreover it seems that even the performance decreases over time :-/. I suppose, we need to progress with [PyCQA/astroid#792](https://github.com/PyCQA/astroid/issues/792) in order to have functional pylintd.

Not sure it's a good idea to start the conversation now. It's not yet clear if or when we can fix astroid. It also isn't a priority at the moment, IMO rightly so. There are other areas that need attention first.

If you never the less want to start the conversation, please make it clear that this is only experimental (at the moment) and still a long way of (maybe even years).

matusvalo:
@jacobtylerwalls unfortunately even after the fix, there is a leak - consider following example:
```python
from pylint.lint import Run
from pylint.lint import pylinter

while True:
    Run(["pylint/checkers/variables.py"], exit=False)
    pylinter.MANAGER.clear_cache()
```

Following script after ~1 minute allocates around 1 GB of memory and keep raising:

image

Moreover, it removes the performance benefit of avoiding rescanning cached files again and again - hence it yields no or zero perf. improvement.

jacobtylerwalls:
What if you explicitly call garbage collection?

matusvalo:
Unfortunately running gc does not help:
```python
from pylint.lint import Run
from pylint.lint import pylinter
import gc

while True:
    Run(["pylint/checkers/variables.py"], exit=False)
    gc.collect()
    pylinter.MANAGER.clear_cache()
```

image

jacobtylerwalls:
The memory leaks in pylint-dev/astroid#792 have been solved (except for some very small ones, perhaps), but because it relies on clearing the cache to do so, that kind of violates the spirit of this PR, which is to keep the cache.

We probably want something like caching the astroid trees for third-party modules to disk, or having another separate cache for them. We could do that by checking if the third-party module defines a version constant.

DanielNoord:
> The memory leaks in [pylint-dev/astroid#792](https://github.com/pylint-dev/astroid/issues/792) have been solved (except for some very small ones, perhaps), but because it relies on clearing the cache to do so, that kind of violates the spirit of this PR, which is to keep the cache.
> 
> We probably want something like caching the astroid trees for third-party modules to disk, or having another separate cache for them. We could do that by checking if the third-party module defines a version constant.

Do we know how `mypy` does this? Because I feel they also know when you have modified third party code yourself, which I think would be a requirement for a good catch miss/hit.

PR #5402, opened 2021-11-26
Fix #4761: Emit `used-before-assignment` where single assignment only made in except blocks

### LINKED ISSUE #4761: False negative `used-before-assignment` when variable is only defined in `except` clause

jacobtylerwalls:
## Type of Changes

## Description

**Before**
If an assignment only took place in an except block, later use of the name outside the except did not raise `used-before-assignment`.

**Now**
As long as only one assignment for the name is in the except block, the message is raised.

Closes #4761 

Relates to #5384 (we may want to cherry-pick the first commit and rewrite slightly)


tests/functional/u/undefined/undefined_variable.py
```diff
@@ -35,7 +35,7 @@ def bad_default(var, default=unknown2):  # [undefined-variable]
 LMBD2 = lambda x, y: x+z  # [undefined-variable]
 
 try:
-    POUET # don't catch me
+    POUET # [used-before-assignment]
```
jacobtylerwalls, line 38: From the blame it looks like this was tolerated just because of generic uncertainty about the control flow, but I think assuming try blocks _never_ execute is bad style.

DanielNoord (changes requested):
Small comment, rest looks good.

However, I think this should be blocked by #5241. We really need that merged first before changing stuff in `visit_name`. Otherwise we'll need to keep updating the refactor and we might miss newly added lines.

Pierre-Sassoulas:
Hmm yes let me review #5241, I think I messed up with not doing it for 2.12 because this will pretty much change the whole implementation for this once we merge.

Pierre-Sassoulas:
@jacobtylerwalls  as someone who used the previous code, your input would be valuable for #5241. If you have the time to review it too I'd appreciate that.

jacobtylerwalls:
> Small comment, rest looks good.

No problem. I'm having trouble finding the comment, though. Can you post it?

pylint/checkers/variables.py
```diff
@@ -607,6 +616,25 @@ def get_next_to_consume(self, node):
...
             ]
 
+        # Filter out assignments in an Except clause that node is not
+        # contained in, assuming they may fail
+        if found_nodes:
+            filtered_nodes = [
+                n
+                for n in found_nodes
+                if not (
+                    isinstance(n.statement().parent, nodes.ExceptHandler)
```
DanielNoord, line 626: @jacobtylerwalls Sorry about that. Seems like I forgot to actually publish my review..

Comment:
Could you use `statement(future=True)` here and on the other lines? The current behaviour of `statement()` is being deprecated and the `future` keyword can be used to trigger the future behaviour.

ChangeLog
```diff
@@ -9,6 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+* ``used-before-assignment`` now considers that assignments in an except
+  block may not have occurred.
```
```suggestion
* ``used-before-assignment`` now assumes that assignments in except blocks
  may not have occurred and warns accordingly.
```

doc/whatsnew/2.13.rst
```diff
@@ -20,6 +20,11 @@ Extensions
 Other Changes
 =============
 
+* ``used-before-assignment`` now considers that assignments in an except
+  block may not have occurred.
```
DanielNoord, line 24:
```suggestion
* ``used-before-assignment`` now assumes that assignments in except blocks
  may not have occurred and warns accordingly.
```

tests/functional/u/undefined/undefined_variable.py
DanielNoord: Does this warrant an additional entry in the changelog? I know #5384 also deals with this, but this is a change in this PR that is not covered by the changelog. I think we should add something about it.

pylint/checkers/variables.py
```diff
@@ -603,10 +612,31 @@ def get_next_to_consume(self, node):
...
+                        n.statement(future=True).parent.parent, nodes.TryExcept
+                    )
+                )
+                or n.statement(future=True).parent.parent_of(node)
+            ]
+            difference = [n for n in found_nodes if n not in filtered_nodes]
+            try:
+                self.consumed_uncertain[node.name] += difference
+            except KeyError:
+                self.consumed_uncertain[node.name] = difference
```
DanielNoord, line 637: Does a `defaultdict` make sense here? I know we're not using it in the other attributes, but that shouldn't hold us back from using better practices in new code.

pylint/checkers/variables.py
```diff
@@ -1221,7 +1251,14 @@ def visit_name(self, node: nodes.Name) -> None:
...
 
-            current_consumer.mark_as_consumed(node.name, found_nodes)
+            # Some nodes were filtered out because they execute only on success
+            # in try/finally or only on failure in except
+            # Add them back so that they are not unnecessarily marked as unused
+            try:
+                uncertain_nodes = current_consumer.consumed_uncertain[node.name]
+            except KeyError:
+                uncertain_nodes = []
+            current_consumer.mark_as_consumed(node.name, found_nodes + uncertain_nodes)
```
DanielNoord, line 1261:
```suggestion
            if node.name in current_consumer.consumed_uncertain:
                found_nodes  += current_consumer.consumed_uncertain[node.name]
            current_consumer.mark_as_consumed(node.name, found_nodes)
```

Little bit more concise and smaller diff.

pylint/checkers/variables.py
```diff
@@ -1221,7 +1251,14 @@ def visit_name(self, node: nodes.Name) -> None:
                                 "undefined-variable", node=node, args=node.name
                             )
 
-            current_consumer.mark_as_consumed(node.name, found_nodes)
+            # Some nodes were filtered out because they execute only on success
```
DanielNoord, line 1254: I don't particularly like how this works, as any new contributor needs to read about this to understand this. It is not as self-explanatory as `to_consume` and `consumed`. However, I don't know of a better way to do this. Perhaps we could improve on the name? I'm not sure.. I'll think about this some more.

DanielNoord (changes requested):
As said in https://github.com/PyCQA/pylint/pull/5384/files#r757821102 let's do this one before #5384.

Now that the refactor is finally done we can move ahead with this one 😄

pylintrc
```diff
@@ -332,7 +332,7 @@ max-locals=25
 max-returns=11
 
 # Maximum number of branch for function / method body
-max-branches=26
+max-branches=27
```
DanielNoord, line 335: This number is quite arbitrary but I'm not sure if we should change it. I'll leave this decision to others though.

pylint/checkers/variables.py
```diff
@@ -1042,6 +1071,7 @@ def _undefined_and_used_before_checker(
             if action is VariableVisitConsumerAction.CONTINUE:
                 continue
             if action is VariableVisitConsumerAction.CONSUME:
+                found_nodes += current_consumer.consumed_uncertain[node.name]
```
DanielNoord, line 1074: Can't this be done in `_check_consumer`?

pylint/checkers/variables.py
```diff
@@ -544,17 +544,24 @@ class NamesConsumer:
     """
 
     def __init__(self, node, scope_type):
-        self._atomic = ScopeConsumer(copy.copy(node.locals), {}, scope_type)
+        self._atomic = ScopeConsumer(
+            copy.copy(node.locals), {}, collections.defaultdict(list), scope_type
```
DanielNoord, line 548: 👍

DanielNoord (changes requested):
👍 

I think these are my last comments. I'll do a final review after your revision.

I wonder if any other contributors can come up with a more self-explanatory name for `consumed_uncertain`..

pylintrc
jacobtylerwalls: I checked the blame; it was just number of the "branchiest" method at the time.

pylint/checkers/variables.py
jacobtylerwalls: There are so many early returns in `_check_consumer` that I would need to write test cases for each of the branches to see if they all need this statement. I was reluctant to possibly introduce subtle bugs. If you have an idea, please let me know!

DanielNoord: Ah I misunderstood the changes on L1168-1170. I think this might indeed be the best approach. However, perhaps only add a comment to this.
I know I'm asking for a lot of documentation and comments but it is so helpful to be easily understand the code when you're trying to fix a false positive here. As you said, these methods rely heavily on early returns etc. so it can be hard to figure out what is going on.

jacobtylerwalls: No problem, I agree that some more comments are the way to go here.

Pierre-Sassoulas:
This need to be reviewed and merged before #5384

pylint/checkers/variables.py
jacobtylerwalls: How about `consumed_conditionally`?

DanielNoord: Either works for me. @Pierre-Sassoulas do you have an opinion?

pylint/checkers/variables.py
```diff
@@ -590,6 +610,8 @@ def mark_as_consumed(self, name, consumed_nodes):
     def get_next_to_consume(self, node):
         """
         Return a list of the nodes that define `node` from this scope.
+        If it is uncertain whether a node will be consumed, such as statement in
```
DanielNoord, line 613: This and the docstring of `consumed_uncertain` are very clear. This part of the codebase will never be fully self explanatory but this really helps! Thanks a lot for all the iterations and working with me 😄 Much appreciated!

pylint/checkers/variables.py
```diff
@@ -1042,6 +1082,12 @@ def _undefined_and_used_before_checker(
             if action is VariableVisitConsumerAction.CONTINUE:
                 continue
             if action is VariableVisitConsumerAction.CONSUME:
+                assert found_nodes is not None, "Cannot consume an empty list of nodes."
```
DanielNoord, line 1085: Is this necessary? As in, doesn't the code break anyway `if not isinstance(found_nodes, list)` because of L1090? This assert doesn't seem to "add" much. Or is this because of `mypy` complaining?
If the latter, could you add a `TODO` which says to remove this after this has been typed correctly? If `_check_consumer` is typed this shouldn't be a problem anymore.

DanielNoord (changes requested):
Thanks @jacobtylerwalls

Just one final comment! The docstrings and comments really do look good!

pylint/checkers/variables.py
jacobtylerwalls: Yes, it's the latter, mypy complains that if `found_nodes` is None, then the `+=` operation will fail. I'll add a TODO.

pylint/checkers/variables.py
Pierre-Sassoulas: 'maybe_consumed' ? I don't have en opinion either :)

DanielNoord (approved):
Thanks @jacobtylerwalls !

pylint/checkers/variables.py
```diff
@@ -617,9 +639,27 @@ def get_next_to_consume(self, node):
...
+                for n in found_nodes
+                if not (
+                    isinstance(n.statement(future=True).parent, nodes.ExceptHandler)
+                    and isinstance(
+                        n.statement(future=True).parent.parent, nodes.TryExcept
+                    )
+                )
+                or n.statement(future=True).parent.parent_of(node)
             ]
+            difference = [n for n in found_nodes if n not in filtered_nodes]
```
cdce8p, line 660: This is quite inefficient, a list contains check for each node.
I would suggest to create a temporary set from `filtered_nodes` to do the check with.

jacobtylerwalls: Something like this?
```diff
diff --git a/pylint/checkers/variables.py b/pylint/checkers/variables.py
index 6559d07f..0e5a0fd1 100644
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -657,8 +657,8 @@ scope_type : {self._atomic.scope_type}
                 )
                 or n.statement(future=True).parent.parent_of(node)
             ]
-            difference = [n for n in found_nodes if n not in filtered_nodes]
-            self.consumed_uncertain[node.name] += difference
+            difference = set(found_nodes) - set(filtered_nodes)
+            self.consumed_uncertain[node.name] += list(difference)
             found_nodes = filtered_nodes
 
         return found_nodes
```

jacobtylerwalls: The most performant solution would be to build two arrays while the condition above is being checked, but that didn't rhyme with the neighboring patterns, so I didn't pursue it initially. What do you think?

cdce8p: I'm not sure if the order in difference is important. If so, it's lost with your suggestion. I was thinking more along these lines here. It's also not the optimal solution but should be good enough.
```py
            filtered_nodes_set = set(filtered_nodes)
            difference = [n for n in found_nodes if n not in filtered_nodes_set]
            self.consumed_uncertain[node.name] += difference
```

Pierre-Sassoulas (approved):
Very nice MR, well documented, well tested, and it fix the try/except logic to something most user would expect imo.

cdce8p:
Came across a regression case during testing -> #5500

jacobtylerwalls:
Hi @martimlobao, thanks for reporting and for providing a concrete example. I've constructed a couple more examples that I think we'd have to reason about to decide on a way forward.

**1.**
```python
for _ in range(3):
    try:
        [][0]
        error = e  # keep error for final raise
    except:
        pass
    else:
        break
    print("retrying...")
else:
    raise error
```

When executed, gives:
```python
  File "a.py", line 12, in <module>
    raise error
NameError: name 'error' is not defined
```

**2.**
```python
for i in range(3):
    if 'some runtime condition':
        continue
    try:
        1 / 0
        error = e  # keep error for final raise
    else:
        break
    print("retrying...")
else:
    raise error
```
When executed, gives:
```python
  File "b.py", line 12, in <module>
    raise error
NameError: name 'error' is not defined
```
***

Of course, 1 and 2 are very similar: all that's required is some other means of finishing the inner loop than hitting the except handler you contemplated. These are runtime control flow things that I don't think Pylint is in a position to speculate about. I _think_ it's a virtue that as things stand, you either have to define `error` before your loop and then check `is not None` before raising it, or just disable the check if you feel very confident about the design of the code. That's a judgment call: verbosity vs. how likely it is someone someone later turns the code into more like **1** or **2** and causes problems.

How does that strike you?

martimlobao:
> Of course, 1 and 2 are very similar: all that's required is some other means of finishing the inner loop than hitting the except handler you contemplated. These are runtime control flow things that I don't think Pylint is in a position to speculate about. I _think_ it's a virtue that as things stand, you either have to define `error` before your loop and then check `is not None` before raising it, or just disable the check if you feel very confident about the design of the code. That's a judgment call: verbosity vs. how likely it is someone someone later turns the code into more like **1** or **2** and causes problems.
> 
> How does that strike you?

Thanks for the detailed response @jacobtylerwalls! The issue you're trying to fix here isn't an easy one to solve, even conceptually, so I can definitely appreciate the difficulty in balancing false positives and false negatives.

I will say though that from a user's perspective, a false positive error is probably more annoying than a false negative, especially in this case where Pylint classifies this as an actual error and not merely a convention or recommendation. Making a change because of an actual bug is fine, but making it to appease a linter that thinks there's a bug that doesn't exist may feel slightly frustrating.

Your examples illustrate well how hard it is to consider every different edge case. I'm not sure how feasible this is as a solution, but would it be possible to check _every_ "exit branch" for a potentially undefined variable? If at least one exit path does not define the variable, then Pylint should raise the error since there's the possibility for a valid `used-before-assignment` error.

Let me try to illustrate what I mean with a few examples:

```python
def undefined_variable():
    """used-before-assignment (E0601)"""
    try:
        pass  # < exit branch where some_message is *not* defined
    except ValueError:
        some_message = 'some message'  # < exit branch where some_message is defined

    if not some_message:
        return 1

    return some_message
```

```python
for _ in range(3):
    try:
        do_something()  # < exit branch where error is *not* defined, should flag
        error = e  # < exit branch where error is defined
    print("retrying...")
else:
    raise error
```

```python
for _ in range(3):
    try:
        do_something()  # not an exit branch because of `else` below (will only exit here with an exception)
        error = e  # < exit branch where error is defined
    else:
        break  # < exit condition where error is *not* defined, but will skip `raise error` below
    print("retrying...")
else:
    raise error
```

```python
for _ in range(3):
    try:
        do_something()  # not an exit branch because of `else` below (will only exit here with an exception)
        error = e  # < exit branch where error is defined
    except:
        pass  # < exit branch where error is *not* defined, should flag
    else:
        break  # < exit condition where error is *not* defined, but will skip `raise error` below
    print("retrying...")
else:
    raise error
```

```python
for i in range(3):
    if 'some runtime condition':
        continue  # < exit branch where error is *not* defined, should flag
    try:  # < other exit branch, needs to check below if error is defined
        1 / 0
        error = e  # keep error for final raise
    else:
        break
    print("retrying...")
else:
    raise error
```

I have to admit I have no understanding of how feasible it would be to implement this, but I believe this approach should be able to distinguish between (all? most?) scenarios where an unassigned variable exists and those where it does not.

jacobtylerwalls:
Thanks for the well-constructed examples. I'm sure in a moment we'll move this discussion to a new issue.

Of your last examples, this is the one false positive:

```python
for _ in range(3):
    try:
        do_something()  # not an exit branch because of `else` below (will only exit here with an exception)
        error = e  # < exit branch where error is defined
    else:
        break  # < exit condition where error is *not* defined, but will skip `raise error` below
    print("retrying...")
else:
    raise error
```

But I can easily turn it into a true positive with:
```python
for _ in range(3):
    try:
        do_something()  # not an exit branch because of `else` below (will only exit here with an exception)
        error = e  # < exit branch where error is defined
    else:
        if 'runtime condition':
            continue  # < exit condition where error is *not* defined, should flag
        break  # < exit condition where error is *not* defined, but will skip `raise error` below
    print("retrying...")
else:
    raise error
```

So maybe this is the constraint to check for:

- if the use of `error` being tested is guarded by `else` in `for/else` (i.e. the last line of these examples)
- if the definition of `error` in an except block is a descendant of the same `for`
- if that _inner_ `else` breaks
- if there is no `continue` anywhere at all underneath the `for` loop (_other_ than inside another for loop, but _including_ nested conditions and try/excepts)

@martimlobao, could you open an issue to track this? I think this specific scenario might be feasible to implement.

PR #5409, opened 2021-11-27
Fix #3675: `safe_infer()` finds ambiguity among function definitions when number of arguments differ

### LINKED ISSUE #3675: Incorrect number of arguments reported when function is imported over
### Steps to reproduce
1. Create `other_file.py` containing:
```
def f(x, y):
    return x + y
```
2. Run pylint on:
```
"""pylint test module"""
    
def f(x):
    return x + 1

def test_one(param):
    """function defined above is matched first"""
    tmp = f
    if param == 0:
        from other_file import f as tmp  # pylint: disable=import-outside-toplevel
        return tmp(1, 4)  # <-- E1121: Too many positional arguments for function call (too-many-function-args)
    return tmp(1)

print(test_one(0))  # Prints "5"
print(test_one(1))  # Prints "2"
```

### Current behavior
Pylint reports that line 12 was passed "too many positional arguments for function call (too-many-function-args)"

### Expected behavior

Pylint should not report an issue since within this if statement tmp refers to the imported function which can be called with two arguments.

### pylint --version output

pylint 2.5.2
astroid 2.4.1
Python 3.8.2 (default, Apr 27 2020, 15:53:34) 
[GCC 9.3.0]

jacobtylerwalls:
## Type of Changes

## Description

**Before**
`safe_infer` might confuse two function definitions having different numbers of arguments.

**Now**
It won't. <s>But there is an inference problem with `TemporaryFile` being inferred as `NamedTemporaryFile`, which has an extra argument (`delete`), so I special-cased it. Is that an Astroid issue?</s> Update: the way `tempfile` defines the name `TemporaryFile` conditionally based on platform means Astroid finds ambiguity.

Closes #3675

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,11 @@ Release date: TBA
 * Some files in ``pylint.testutils`` were deprecated. In the future imports should be done from the
   ``pylint.testutils.functional`` namespace directly.
 
+* ``safe_infer`` no longer makes an inference given two function
+  definitions with differing numbers of arguments.
+
+  Closes #3675
+
 * Fix ``unnecessary_dict_index_lookup`` false positive when deleting a dictionary's entry.
 
   Closes #4716
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,7 +14,7 @@ Release date: TBA
 * Some files in ``pylint.testutils`` were deprecated. In the future imports should be done from the
   ``pylint.testutils.functional`` namespace directly.
 
-* ``safe_infer`` no longer makes an inference given two function
+* ``safe_infer`` no longer makes an inference when given two function
   definitions with differing numbers of arguments.
 
   Closes #3675
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1260,13 +1260,13 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
-            if isinstance(inferred, nodes.FunctionDef):
-                if (
-                    inferred.args.args is not None
-                    and value.args.args is not None
-                    and len(inferred.args.args) != len(value.args.args)
-                ):
-                    return None  # Different number of arguments indicates ambiguity
+            if (
+                isinstance(inferred, nodes.FunctionDef)
+                and inferred.args.args is not None
+                and value.args.args is not None
+                and len(inferred.args.args) != len(value.args.args)
+            ):
+                return None  # Different number of arguments indicates ambiguity
     except astroid.InferenceError:
         return None  # There is some kind of ambiguity
     except StopIteration:
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1260,6 +1260,19 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
+            if isinstance(inferred, astroid.FunctionDef):
+                # Special case due to unexpected inference ambiguity
+                if (
+                    inferred.name == "TemporaryFile"
+                    and value.name == "NamedTemporaryFile"
+                ):
+                    continue
+                if (
+                    inferred.args.args is not None
+                    and value.args.args is not None
+                    and len(inferred.args.args) != len(value.args.args)
+                ):
+                    return None  # Different number of arguments indicates ambiguity
     except astroid.InferenceError:
         return None  # There is some kind of ambiguity
     except StopIteration:
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1260,7 +1260,7 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
-            if isinstance(inferred, astroid.FunctionDef):
+            if isinstance(inferred, nodes.FunctionDef):
                 if (
                     inferred.args.args is not None
                     and value.args.args is not None
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1261,12 +1261,6 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
             if isinstance(inferred, astroid.FunctionDef):
-                # Special case due to unexpected inference ambiguity
-                if (
-                    inferred.name == "TemporaryFile"
-                    and value.name == "NamedTemporaryFile"
-                ):
-                    continue
                 if (
                     inferred.args.args is not None
                     and value.args.args is not None
--- a/tests/functional/t/too/too_many_function_args.py
+++ b/tests/functional/t/too/too_many_function_args.py
@@ -1,0 +1,19 @@
+"""https://github.com/PyCQA/pylint/issues/3675"""
+
+
+def noop(x):  # pylint: disable=invalid-name
+    """Return value unchanged"""
+    return x
+
+
+def add(x, y):  # pylint: disable=invalid-name
+    """Add two values"""
+    return x + y
+
+
+def main(param):
+    """Should not emit too-many-function-args"""
+    tmp = noop  # matched first
+    if param == 0:
+        tmp = add
+    return tmp(1, 1.01)
--- a/tests/functional/c/consider/consider_using_with.py
+++ b/tests/functional/c/consider/consider_using_with.py
@@ -21,7 +21,7 @@ def test_urlopen():
 
 
 def test_temporary_file():
-    _ = tempfile.TemporaryFile("r")  # [consider-using-with]
+    _ = tempfile.TemporaryFile("r")  # ambiguous with NamedTemporaryFile
 
 
 def test_named_temporary_file():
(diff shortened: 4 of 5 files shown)


pylint/checkers/utils.py
```diff
@@ -1251,6 +1251,19 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
+            if isinstance(inferred, astroid.FunctionDef):
+                # Special case due to unexpected inference ambiguity
+                if (
+                    inferred.name == "TemporaryFile"
```
Pierre-Sassoulas, line 1257: I think this is very specific to the issue with tempfile, probably too much.

jacobtylerwalls: I'll try to diagnose the issue, maybe it's in astroid.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -36,6 +36,7 @@
         "tempfile.NamedTemporaryFile",
         "tempfile.SpooledTemporaryFile",
         "tempfile.TemporaryDirectory",
+        "tempfile.TemporaryFile",
```
DanielNoord, line 39: This is an unrelated change right? Or am I missing something? If so, could you please make another PR for this?

jacobtylerwalls: Yes, I can remove.

jacobtylerwalls: Moved to #5468.

tests/functional/t/too/too_many_function_args.py
```diff
@@ -0,0 +1,19 @@
...
+    """Add two values"""
+    return x + y
+
+
+def main(param):
+    """Should not emit too-many-function-args"""
+    tmp = noop  # matched first
+    if param == 0:
+        tmp = add
+    return tmp(1, 1.01)
```
DanielNoord, line 19: It does seem like we should emit some type of warning for this pattern. Something like `overwriting-with-different-amount-of-function-args` 😅

ChangeLog
```diff
@@ -14,6 +14,11 @@ Release date: TBA
 * Some files in ``pylint.testutils`` were deprecated. In the future imports should be done from the
   ``pylint.testutils.functional`` namespace directly.
 
+* ``safe_infer`` no longer makes an inference given two function
```
DanielNoord, line 17:
```suggestion
* ``safe_infer`` no longer makes an inference when given two function
```

pylint/checkers/utils.py
```diff
@@ -1260,6 +1260,13 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
+            if isinstance(inferred, astroid.FunctionDef):
```
DanielNoord, line 1263:
```suggestion
            if isinstance(inferred, nodes.FunctionDef):
```

DanielNoord (changes requested):
Nice fix! Two minor comments :)

tests/functional/t/too/too_many_function_args.py
jacobtylerwalls: Maybe just extend `arguments-differ`?

DanielNoord: That is not a bad idea!

pylint/checkers/utils.py
```diff
@@ -1260,6 +1260,13 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
+            if isinstance(inferred, nodes.FunctionDef):
+                if (
+                    inferred.args.args is not None
+                    and value.args.args is not None
+                    and len(inferred.args.args) != len(value.args.args)
+                ):
```
DanielNoord, line 1268:
```suggestion
                if len(inferred.args.args) != len(value.args.args):
```

Would this work? `nodes.Arguments.args` is typed as `typing.List[AssignName]` so it shouldn't ever be `None`?

jacobtylerwalls: Maybe the typing in Astroid needs updating?

```
pylint/testutils/lint_module_test.py:181: AssertionError
---------------------------------- Captured stdout call ----------------------------------
Exception on node <Call l.10 at 0x7f7f171e7b20> in file '/Users/.../pylint/tests/functional/m/metaclass_attr_access.py'
---------------------------------- Captured stderr call ----------------------------------
  File "/Users/.../pylint/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "/Users/.../pylint/pylint/checkers/refactoring/refactoring_checker.py", line 996, in visit_call
    self._check_consider_using_with(node)
  File "/Users/.../pylint/pylint/checkers/refactoring/refactoring_checker.py", line 1468, in _check_consider_using_with
  File "/Users/.../pylint/pylint/checkers/utils.py", line 1264, in safe_infer
    if len(inferred.args.args) != len(value.args.args):
TypeError: object of type 'NoneType' has no len()
```

tests/functional/t/too/too_many_function_args.py
jacobtylerwalls: Should we capture this idea in a different issue?

DanielNoord: Wouldn't be a bad idea! Could you open it?

pylint/checkers/utils.py
DanielNoord: This is actually a bug in `astroid`. See:
https://github.com/PyCQA/astroid/blob/6691a9e97373b28e15a76b747f78c041f2064df4/astroid/raw_building.py#L232
This should be:
`func.args.args = []`.

I tested and with this change the additional checks are unnecessary.
Let's block this until this is fixed in `astroid`. If you want you can make the PR there, I haven't come up with a test yet so you would need to do that yourself.

Diff:
```diff
diff --git a/astroid/raw_building.py b/astroid/raw_building.py
index 57e8d354f..e78dae7a2 100644
--- a/astroid/raw_building.py
+++ b/astroid/raw_building.py
@@ -229,7 +229,7 @@ def object_build_methoddescriptor(node, member, localname):
     )
     # set node's arguments to None to notice that we have no information, not
     # and empty argument list
-    func.args.args = None
+    func.args.args = []
     node.add_local_node(func, localname)
     _add_dunder_class(func, member)
```

tests/functional/t/too/too_many_function_args.py
jacobtylerwalls: #5469

pylint/checkers/utils.py
jacobtylerwalls: Wouldn't that introduce a behavior change here?

https://github.com/PyCQA/pylint/blob/3ed7a8e3933d14bc6d07afa8e6ba992b5a4b6926/pylint/checkers/classes.py#L1422

The code comments say it's trying to distinguish a truthful report of zero arguments from a total lack of information.

DanielNoord: Hmm, I have not looked at it well enough but this seems like something that should be fixed. It makes sense that `Arguments.args` is always initialised and allows removing checks such as the one in this PR. 
`astroid` probably shouldn't use that attribute itself to store information about the fact that it can't determine the attribute. If we had enabled `mypy` on `astroid` and had correct typing this wouldn't have been allowed anyway.

If I find the time I'll try and see if I can find a good solution for this.

DanielNoord: Pinging @cdce8p as you often have good ideas about how to represent things in `astroid`. Please disregard if you don't have the time to look into this!

The problem we're dealing with here is that for builtin functions or classes defined in C we can't get a signature and therefore can't get the arguments. In `astroid` we represent this by setting the `args` attribute of the `Arguments` class associated with that `FunctionDef` (set on `FunctionDef.args`) to `None`. See:
https://github.com/PyCQA/astroid/blob/6691a9e97373b28e15a76b747f78c041f2064df4/astroid/raw_building.py#L232

However, `Arguments.args` is typed as `typing.List[AssignName]`. Thus, the linked code is in direct violation of the current typing. Therefore, I don't think we should keep the status quo, but should try and find a solution.

Since there is no way to get the arguments we must indeed have a way to represent that the arguments simply can't be determined. This is also relevant information. For example, the fact that arguments can't be determined is used here:
https://github.com/PyCQA/pylint/blob/3ed7a8e3933d14bc6d07afa8e6ba992b5a4b6926/pylint/checkers/classes.py#L1422
I found about 3/4 other cases within `pylint` that also rely on this. I tried using a `node.parent.name == "builtins"` check, but that doesn't work for stuff like the `_io` module and other C-modules.

How to go forward? I think there are two options here:
1) Change the typing to become `typing.Optional[typing.List[AssignName]]` and lose the possibility of knowing you can iterate over `Arguments.args`.
2) Add a new attribute `missing_signature` (?) to `Arguments` and have the checks that rely on `args is not None` (or `is`) use that new attribute.

@jacobtylerwalls I'm also open to any of your suggestions for this problem. I propose to keep blocking this PR until this `astroid` issue has been resolved as it might impact the code of this PR.

cdce8p: Tbh I wasn't aware of that behavior when I added the type annotation. Even if we wanted to untangle that into a new attribute, we couldn't without deprecating it first. As you mentioned, there is even code in pylint that depends on that behavior so it won't be unreasonable to assume that plugins might do too.

I would suggest we update the type annotation and docstring instead. It's the easiest thing to do here. Link me on the PR and I can take a look.

DanielNoord: See https://github.com/PyCQA/astroid/pull/1283

DanielNoord (approved):
With that discussion solved, this is now good to go!

Thanks @jacobtylerwalls, also for pointing to the code that relied on `args` being `None`!

Pierre-Sassoulas (approved):
Thank you for also making astroid better !

DanielNoord:
@Pierre-Sassoulas I think this can be merged? Or are we waiting on something?

Pierre-Sassoulas:
I probably thought Marc would want to review, but we can merge, most MR do not have 3 approbations.

PR #5411, opened 2021-11-27
Update documentation of ``max_pyver`` test option

DanielNoord:

## Type of Changes

## Description

See https://github.com/PyCQA/pylint/pull/5386#discussion_r756987807.

### DIFF
--- a/doc/development_guide/testing.rst
+++ b/doc/development_guide/testing.rst
@@ -76,7 +76,7 @@ The .rc file can also contain a section ``[testoptions]`` to pass options for th
 test runner. The following options are currently supported:
 
     "min_pyver": Minimal python version required to run the test
-    "max_pyver": Python version from which the test won't be run. Setting to 3.10 will exclude 3.10.1 as well.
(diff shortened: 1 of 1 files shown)


doc/development_guide/testing.rst
```diff
@@ -76,7 +76,7 @@ The .rc file can also contain a section ``[testoptions]`` to pass options for th
 test runner. The following options are currently supported:
 
     "min_pyver": Minimal python version required to run the test
-    "max_pyver": Maximum python version required to run the test
+    "max_pyver": Python version from which the test won't be run. Setting to 3.10 will exclude 3.10.1 as well.
```
cdce8p, line 79:
```
Setting to 3.10 will exclude 3.10.1 as well.
```
I think that is obvious. I believe it would be more helpful to look at it from the other way round.

```
If the last supported version is 3.9 -> set it to 3.10
```

DanielNoord:
```suggestion
    "max_pyver": Python version from which the test won't be run. If the last supported version is 3.9 this setting should be set to 3.10.
```

cdce8p (approved):

PR #5412, opened 2021-11-28
Fix #5371: Correctly count arguments to static methods missing @staticmethod decorator

### LINKED ISSUE #5371: Confusing/false positive arguments-differ on enum _generate_next_value_
### Bug description

```python
# pylint: disable=missing-class-docstring,missing-module-docstring

from enum import Enum, auto
from typing import Any

class MyEnum(Enum):
    @staticmethod
    def _generate_next_value_(name: str, start: int, count: int, last_values: list[Any]) -> Any:
        return 42

class FooEnum(MyEnum):
    BAR = auto()

print(FooEnum.BAR)
```

### Configuration

_No response_

### Command used

```shell
pylint t.py
```

### Pylint output

```shell
************* Module t
t.py:10:4: W0221: Number of parameters was 4 in 'Enum._generate_next_value_' and is now 4 in overridden 'MyEnum._generate_next_value_' method (arguments-differ)
```

### Expected behavior

No error. Or an error message that makes sense, if there is an error :) Currently it says the number is 4 for both which doesn't help with pinpointing the problem.

Somehow, it's the `@staticmethod` that throws pylint off here. That's however necessary for mypy: https://github.com/python/mypy/issues/7591

### Pylint version

```shell
pylint 2.11.1
astroid 2.9.0
Python 3.9.4 (default, Apr 11 2021, 07:48:09) 
[GCC 9.3.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

**Before**
Static methods lacking a `@staticmethod` decorator, such as `Enum._generate_next_value()` in the stdlib would have their first argument stripped when counting their arguments, even though it wasn't `self` or `cls`. This caused comparison issues in `arguments-differ`, including both false negatives and positives.

**Now**
Check if the method is bound before stripping the first argument.

Closes #5371

### DIFF
--- a/tests/functional/a/arguments_differ.py
+++ b/tests/functional/a/arguments_differ.py
@@ -152,7 +152,6 @@ class SuperClass(object):
         return arg1 + arg2
 
     def should_have_been_decorated_as_static(arg1, arg2):  # pylint: disable=no-self-argument
-        # Enum._generate_next_value in stdlib should be static
         return arg1 + arg2
 
 
--- a/tests/functional/a/arguments_differ.py
+++ b/tests/functional/a/arguments_differ.py
@@ -141,7 +141,7 @@ class PropertySetter(Property):
 
 class StaticmethodChild2(Staticmethod):
 
-    def func(self, data):
+    def func(self, data):  # [arguments-differ]
         super().func(data)
 
 
@@ -151,14 +151,23 @@ class SuperClass(object):
     def impl(arg1, arg2, **kwargs):
         return arg1 + arg2
 
+    def should_have_been_decorated_as_static(arg1, arg2):  # pylint: disable=no-self-argument
+        # Enum._generate_next_value in stdlib should be static
+        return arg1 + arg2
+
 
 class MyClass(SuperClass):
 
-    def impl(self, *args, **kwargs):
+    @staticmethod
+    def impl(*args, **kwargs):
         """
         Acceptable use of vararg in subclass because it does not violate LSP.
         """
         super().impl(*args, **kwargs)
+
+    @staticmethod
+    def should_have_been_decorated_as_static(arg1, arg2):
+        return arg1 + arg2
 
 
 class FirstHasArgs(object):
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,11 @@ Release date: TBA
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
 
+* Fixed detection of ``arguments-differ`` when superclass static
+  methods lacked a ``@staticmethod`` decorator.
+
+  Closes #5371
+
 ..
   Insert your changelog randomly, it will reduce merge conflicts
   (Ie. not necessarily at the end)
@@ -21,11 +26,6 @@ Release date: TBA
 
 ..
   Put bug fixes that should not wait for a new minor version here
-
-* Fixed detection of ``arguments-differ`` when superclass static
-  methods lacked a ``@staticmethod`` decorator.
-
-  Closes #5371
 
 * Fixed a false positive for ``unused-import`` where everything
   was not analyzed properly inside typing guards.
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,11 @@ Release date: TBA
 ..
   Put bug fixes that should not wait for a new minor version here
 
+* Fixed detection of ``arguments-differ`` when superclass static
+  methods lacked a ``@staticmethod`` decorator.
+
+  Closes #5371
+
 * Fixed a false positive for ``unused-import`` where everything
   was not analyzed properly inside typing guards.
 
--- a/pylint/checkers/classes.py
+++ b/pylint/checkers/classes.py
@@ -178,7 +178,7 @@ def _definition_equivalent_to_call(definition, call):
 
 def _positional_parameters(method):
     positional = method.args.args
-    if method.type in {"classmethod", "method"}:
+    if method.is_bound() and method.type in {"classmethod", "method"}:
         positional = positional[1:]
     return positional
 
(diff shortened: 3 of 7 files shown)


tests/functional/a/arguments_differ.py
```diff
@@ -151,15 +151,24 @@ class SuperClass(object):
...
+    def impl(*args, **kwargs):
         """
         Acceptable use of vararg in subclass because it does not violate LSP.
         """
         super().impl(*args, **kwargs)
 
+    @staticmethod
+    def should_have_been_decorated_as_static(arg1, arg2):
+        return arg1 + arg2
+
```
jacobtylerwalls, line 171: This is what was reported as generating a false positive.

tests/functional/a/arguments_differ.py
```diff
@@ -141,7 +141,7 @@ def close(self, attr):
 
 class StaticmethodChild2(Staticmethod):
 
-    def func(self, data):
+    def func(self, data):  # [arguments-differ]
         super().func(data)
```
jacobtylerwalls, line 145: Here was a false negative!

Pierre-Sassoulas: Nice fix !

tests/functional/a/arguments_differ.py
```diff
@@ -151,15 +151,24 @@ class SuperClass(object):
...
-    def impl(self, *args, **kwargs):
+    @staticmethod
+    def impl(*args, **kwargs):
         """
         Acceptable use of vararg in subclass because it does not violate LSP.
         """
         super().impl(*args, **kwargs)
 
+    @staticmethod
+    def should_have_been_decorated_as_static(arg1, arg2):
```
Pierre-Sassoulas, line 169: Imo it should emit argument-differ, there are two errors here, the first one lack a self argument but this one's argument differ as there are 3 instead of 2.

Pierre-Sassoulas (changes requested):
Could we add a functional test using the problematic Enum directly ? I think it might be a case where a specific handling of it to not have false positive elsewhere would be warranted.

tests/functional/a/arguments_differ.py
jacobtylerwalls: Interesting, let me see if I can follow. How are you counting three arguments? I think that's a code-style suggestion to say that there should be no static methods inside classes without a decorator, so if someone _has_ done that, we get here and they just have 2 == 2.

Pierre-Sassoulas: Sorry I meant 2 instead of 3. The first function would need self, arg1, arg2 and this one too but as it's static having 2 is equivalent to 3 (like the false positive you found higher up)

Pierre-Sassoulas: In the first function ``arg1`` should be called ``self`` but as long as it's not fixed we don't know if it has 2 or 3 args (will the user rename arg1 to self or add self on top of arg1 arg2 ?). Assuming arg1 will be renamed to self and as the second function is static it means it should only have one arg equivalent to ``arg2`` as the self is implicit. If we assume self will be added before arg1 after the user fix it then no error makes more sense. But we have to detect two problems correctly instead of one for that to work, not making assumption is safer imo.

jacobtylerwalls: > In the first function arg1 should be called self but as long as it's not fixed we don't know if it has 2 or 3 args (will the user rename arg1 to self or add self on top of arg1 arg2 ?).

I guess this is the part I'm having trouble with -- let's say a user has disabled `no-self-use` because they don't care whether their static methods are decorated as such. Now following your idea pylint will be coupling the behavior of `arguments-differ` to the choice of `no-self-use` which will be frustrating for that user. There's nothing strictly wrong with the Enum class in the report, it just makes typing harder. Some users don't type.

> will the user rename arg1 to self or add self on top of arg1 arg2 ?

This is the assumption I would be reluctant to make -- why are we assuming `self` needs to be added at all?

Pierre-Sassoulas: > This is the assumption I would be reluctant to make -- why are we assuming self needs to be added at all?

I agree we should not couple ``no-self-argument`` to ``argument-differ``.

But argument differ here:
```python
class Mother:
   def should_have_been_decorated_as_static(arg1, arg2):  # pylint: disable=no-self-argument
        # Enum._generate_next_value in stdlib should be static
        return arg1 + arg2

class Daugther(Mother):
    @staticmethod
    def should_have_been_decorated_as_static(arg1, arg2): #[argument-differ]
        return arg1 + arg2
```
Mother need 1 explicit args (self is implicit + arg2):
```python
>>> m = Mother()
>>> m.should_have_been_decorated_as_static(1, 2 )
TypeError: should_have_been_decorated_as_static() takes 2 positional arguments but 3 were given
>>> m.should_have_been_decorated_as_static(1)
  File "<stdin>", line 4, in should_have_been_decorated_as_static
TypeError: unsupported operand type(s) for +: 'Mother' and 'int'
```

While Daugter needs 2 explicit args:
```python
>>> d = Daugther()
>>> d.should_have_been_decorated_as_static(1, 2)
3
>>> d.should_have_been_decorated_as_static(1)
TypeError: should_have_been_decorated_as_static() missing 1 required positional argument: 'arg2'
```
Thus argument-differ should be triggered.

jacobtylerwalls: Ah, but only if you call it on the instance!

```python
>>> Mother.should_have_been_decorated_as_static(1, 2)
3
>>> Daugther.should_have_been_decorated_as_static(1, 2)
3
```

I think this was the scenario in the stdlib, which is why I didn't raise a bug against CPython. What do you think?

Pierre-Sassoulas: OK it makes sense now. I think ``@staticmethod`` is a stronger signal than the arg not being called self but there is already a message for the argument not being called self.

Pierre-Sassoulas (changes requested):
Thanks for clarifying the misunderstanding from my previous comment :) Would it be possible to add an example using the enum from the original issue ?

Pierre-Sassoulas (approved):
Very nice, thank you !

Issue #2918, opened 2019-05-16
starargs from lambda result in no-value-for-parameter


aure-olli:
### Steps to reproduce
```python
def foo(one, two, three):
    pass

bar = lambda *args: foo(*args)
```

### Current behavior
```
test.py:4:20: E1120: No value for argument 'one' in function call (no-value-for-parameter)
test.py:4:20: E1120: No value for argument 'two' in function call (no-value-for-parameter)
test.py:4:20: E1120: No value for argument 'three' in function call (no-value-for-parameter)
```

### Expected behavior
No warning

### pylint --version output
This is exactly https://github.com/PyCQA/pylint/issues/722, except I'm using a lambda function.

I hope it can be similarly fixed

PCManticore:
Hi, thank you for the report! I can confirm the bug.

Issue #3563, opened 2020-05-01
False positive `no-value-for-parameter` in Pylint 2.5


selimb:
### Steps to reproduce
In `bug.py`:
```python
"""Whatever"""
import os

def good(*parts):
    """All good"""
    return os.path.join(*parts)

def bad(*parts):
    """False positive below this line"""
    ret = os.path.join(*parts)
    return ret
```
Run `pylint bug.py` (no user config necessary).

### Current behavior
In `Pylint==2.5.0` (see below for detailed versions):
```
************* Module bug
bug.py:10:10: E1120: No value for argument 'a' in function call (no-value-for-parameter)

------------------------------------------------------------------
Your code has been rated at 1.67/10 (previous run: 1.67/10, +0.00)
```

With `Pylint<2.5`:
```

-------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 1.67/10, +8.33)
```

### Expected behavior
`good` and `bad` are effectively the same thing, and both are bug-free. I would expect no errors from Pylint.

Also, I don't get why the argument name is inferred as `a`?
```python
>>> import inspect
>>> inspect.signature(os.path.join)
<Signature (path, *paths)>
```

### pylint --version output
For `Pylint==2.5.0` (first case above):
```
pylint 2.5.0
astroid 2.4.0
Python 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 23 2018, 23:31:17) [MSC v.1916 32 bit (Intel)]
```
For `Pylint<2.5`:
```
pylint 2.4.4
astroid 2.3.3
Python 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 23 2018, 23:31:17) [MSC v.1916 32 bit (Intel)]
```

selimb:
Wow that was quick. Thanks!

DervishD:
Sorry for waking this up, but I'm having the same error with pylint 2.6.2, astroid 2.4.2.

Is this known or do I have to open a new bug?

PR #5417, opened 2021-11-28
typecheck: simplify variadic positional detection, fixing some false positives

### LINKED ISSUE #5416: no-value-for-parameter false positive for variadic function passed as argument
### Bug description

With the above code, `no-value-for-parameter` is emitted for `func_KO`, although nothing is emitted for `func_OK`, therefore I believe it's a false positive.

```python
def func(one, two):
    """A function with two arguments"""
    return one, two


def func_OK(*args):
    """Variadic arguments passed to `func`, assumed to be OK"""
    return func(*args)


def func_KO(*args):
    """When `func` is used as a function argument, False positive is emited"""
    print(func(*args))

```

I was able to fix locally, even though I'm not sure to understand fully why, I'll try to send a pull request

### Configuration

_No response_

### Command used

```shell
pylint --disable=all --enable=no-value-for-parameter repro.py
```

### Pylint output

```shell
************* Module repro
repro.py:13:10: E1120: No value for argument 'one' in function call (no-value-for-parameter)
repro.py:13:10: E1120: No value for argument 'two' in function call (no-value-for-parameter)

Your code has been rated at -6.67/10 (previous run: -6.67/10, +0.00)
```

### Expected behavior

No false positive should be emitted in `func_KO`, just like for `func_OK`.

### Pylint version

```shell
pylint 2.12.2-dev0
astroid 2.9.0
Python 3.7.7 (default, Oct 13 2021, 05:29:46) 
[GCC 8.5.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

perrinjerome:

## Type of Changes

## Description

The isinstance checks were not necessary here and caused several false positives where a function with variadic positional argument was called, like for example when the call is used as a function argument, in a if, while or with statement.

Closes #5416


pylint/checkers/typecheck.py
```diff
@@ -631,7 +631,7 @@ def _no_context_variadic_keywords(node, scope):
 
 def _no_context_variadic_positional(node, scope):
     variadics = ()
-    if isinstance(scope, nodes.Lambda) and not isinstance(scope, nodes.FunctionDef):
+    if isinstance(scope, nodes.Lambda):
```
perrinjerome, line 634: This seem to fix the issue from #5416 without causing regression, but I don't understand the else branch. With this change the else is no longer covered.

Pierre-Sassoulas: Seeing the removed conditional it was for treating ``FunctionDef`` (duh). But it seems it could be treated generically. All tests works well with this change. The really hard question is how to trigger this branch now and is it possible at all ? Did you try running the primer locally with code coverage ? That could help find a covering functional test. (The online code coverage is calculated on standard tests only).

DanielNoord: https://github.com/PyCQA/pylint/blob/e33bc6377bbbf6135888cdadecdfc3baa8497a24/pylint/checkers/typecheck.py#L636-L642

I have looked at the `else` for a bit and in the current tests there is no test where `node != call` and none where `variadics == ()`. Therefore this whole `else` block seems meaningless as doing `variadics = node.starargs + node.kwargs` for all `node`'s works as well. We might be okay by removing both the `if` and the `else`.

The commit that added this was https://github.com/PyCQA/pylint/commit/99ce3cdfd3bc1bf2891fd807f97ee3cd1ec26c21. It mentions a lack of context recognition, that might have been solved now. This commit fixes https://github.com/PyCQA/pylint/issues/2918. @Pierre-Sassoulas do you know anything about this?

The commit prior to that also makes little sense https://github.com/PyCQA/pylint/commit/d95eb088442fad414772b85873c08ac94587ad9a. In `_no_context_variadic_keywords` there is a check for `scope` not being `FunctionDef`. However, the only call to `_no_context_variadic_keywords` is guarded by a check that `scope` is `FunctionDef` on L562.

I'm not really sure what this code is trying to achieve. Based on the expansion of the if statement on L638-640 on the current code compared to the original I feel like this might have been causing false positives and the list of accepted `statement` types was expanded to resolve those without looking at whether the method actually made sense. See for example this issue https://github.com/PyCQA/pylint/issues/3563.

perrinjerome: Thanks for feedback here and sorry for being late replying. I tested a bit more and with that first version of the patch ( for reference it was https://github.com/PyCQA/pylint/pull/5417/commits/4138aa61442f3f636cf183a23c382eb9c43962b6 ) both with unit tests or primer tests the `else:` branch was never used.

I can also confirming that with a simplified version like above all tests are also passing, I think @DanielNoord this is what you suggested.

```python
def _no_context_variadic_positional(node, scope):
    variadics = node.starargs + node.kwargs
    return _no_context_variadic(node, scope.args.vararg, nodes.Starred, variadics)
```

I pushed a new version of the patch using this implementation and also added a few more tests. This test file had text for `*args` and `**kwargs`, so I added the same test for `**kwargs`, even if it was not problematic, I figured it might be good to have it tested as well.

I also added changelog entry etc and I'm marking as ready for review, so if you like, maybe this can be merged. Frankly, I don't fully understand the previous code here, so if you prefer waiting until we get a better understanding of what the code was trying to achieve, that's of course OK with me, but I don't think I'll be able to help with understanding the previous code.

Thanks again for your suggestions

DanielNoord (approved):
Thanks @perrinjerome, this is indeed what I had in mind. I'm approving this but I think @Pierre-Sassoulas or another maintainer might need to take a good look at the previous commits and the rationale for adding that `if ... else` block. I don't understand it as well and think we can remove, but another pair of (more experienced) eyes wouldn't hurt!

@Pierre-Sassoulas I know you want to release `2.12.2` soon, but since the changelog entry for this is already in `2.12.2` and it is a relatively minor fix I think it makes sense to move this. I'll leave that to you though!

perrinjerome:
Thanks @DanielNoord . I just I forgot to edit the title of the pull request to be more meaning full. I just changed from: <del>typecheck: get variadics from node for all Lambda</del> to <ins>typecheck: simplify variadic positional detection, fixing some false positives</ins>.

pylint/checkers/typecheck.py
Pierre-Sassoulas: I did some digging, the else was added in 99ce3cdfd3bc1bf2891fd807f97ee3cd1ec26c21, and the functional test added together with it was:

```python
def func(one, two, three):
    return one + two + three


CALL = lambda *args: func(*args)
```

I'll look into it in more detail tomorrow. I was merging approved MR for 2.13 and this is the last one that is approved and also passing tests :)

Pierre-Sassoulas (approved):
Well, good job fixing a bug by removing almost all the code and replacing it with a one liner :smile: All tests including the original one are passing so maybe the call condition changed since. I'm going to trust the test suite here.

perrinjerome:
Thanks again @Pierre-Sassoulas and @DanielNoord for the help here, it's good that this was merged !

Issue #5359, opened 2021-11-21
Better primer tests (faster, less duplication, more diverse)


Pierre-Sassoulas:
### Current problem

Right now the primer tests take more than double the time of the next slower tests. Also the CI configuration is duplicated between primer for stdlib and primer for external tests. The choice of repository to check is also arbitrary.

### Desired solution

- [x] #5425 > Burst the `primer-external` CI job into multiple jobs to cut down on CI run time
- [x] #5425 >  Remove the duplication in the GitHub action configuration and move `primer` to its own workflow file
- [ ] Clarify the criteria for choosing a repository and update the documentation
- [ ] Remove repositories that are redundant or less useful considering what we already have
- [x] Fixing dependency to a specific commit to be able to cache repository
- [x] Caching of repository for primers
- [x] Remove duplication in primer jobs by using a job template
- [x] #5425 >  Only run when something changes in `pylint` internals or `primer` testutils
- [x] Improve error message on crashes and fatal errors within the primer logs
- [ ] Check and see if CI workflow file can be shortened/improved

### Additional context

#5173

DanielNoord:
@Pierre-Sassoulas Just something I thought of:
I'm not sure if we get unlimited amount of CI minutes from Github, but if we need to be conservative with them we could consider not running the external primer on all three versions.

Pierre-Sassoulas:
We definitely need to cache everything we can too. (Setting the hash of the repository we lint would help).

cdce8p:
> I'm not sure if we get unlimited amount of CI minutes from Github

@DanielNoord AFAIK public repos are only limited in the number of concurrent job runners, not the actual time.

* https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions
* https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

cdce8p:
One suggestion for something I saw recently. If the primer tests are moved to a new workflow, you could also add `concurrency` control. I.e. it doesn't make sense to continue with primer tests if a new commit has been pushed in the meantime.
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

[...]

PR #5425, opened 2021-11-28
Update ``primer`` tests to own file, separate batches and concurrent runs

DanielNoord:

## Type of Changes

## Description

First set of improvements for the `primer`. This adds:
- Concurrent jobs, meaning the jobs will stop if a new action is created.
- `primer` now only runs on changes to `pylint/**` or `tests/primer/**` or pushes to `main`. This should allow updating docs or tests without the primer holding you up.
- The `primer` sits in its own workflow file
- The `primer` has been bursted into two batches based on the first half of the alphabet and the second 😄 

Ref #5359


DanielNoord:
Times between both batches seem fairly equal so I think this division works. Run time is quite high as US is starting to launch GitHub Actions at this time..

Looks good for review!

Pierre-Sassoulas (commented):
I just had a quick look but it look amazing already !I'll do a full review tomorrow.

I was reading the doc for github caching so we an download the repositories once. I think if we choose a particular commit we can even cache reliably between runs as long as we do not update to the latest commits. I'll do that in another MR though, everything here was required, is compatible and will also make things anyway !

.github/workflows/primer-test.yaml
```diff
@@ -0,0 +1,167 @@
...
+        run: |
+          echo "Failed to restore Python venv from cache"
+          exit 1
+      - name: Run pytest
+        run: |
+          . venv/bin/activate
+          pip install -e .
+          pytest -m primer_external_batch_one --primer-external -n auto
+
+  pytest-primer-external-batch-two:
```
Pierre-Sassoulas, line 133: I was trying to find a way to not copy paste so much, did you also search and find nothing acceptable or did you simply implement something fast ?

Pierre-Sassoulas (approved):
I think we can merge as is, but I have a small question.

.github/workflows/primer-test.yaml
DanielNoord: I simply implemented something fast. Do you want me to look if something is possible before merging?

Pierre-Sassoulas: No, let's keep the MR small and do it later on. Also this CI speed up is too valuable to wait, clearly 😄

PR #5428, opened 2021-11-29
Move ``TestLoggingModuleDetection`` to functional tests

DanielNoord:

## Type of Changes

## Description

Some of the most basic test captured in the first methods of the class are already included in some of the functional tests. I did not copy those over.


Pierre-Sassoulas (approved):
Good job ! It would be easier to review if there was multiple commits with each test moved one by one but feel free to keep doing one commit if it's too much work.

DanielNoord:
> Good job ! It would be easier to review if there was multiple commits with each test moved one by one but feel free to keep doing one commit if it's too much work.

I'll try and see if that's possible. I often copy everything out of the unittests into new files and then start fixing the new files as that makes it a bit easier for me, but I can try to do multiple commits if it is not much extra work!

PR #5434, opened 2021-11-29
Move most tests out of ``TestImportsChecker``

DanielNoord:

## Type of Changes

## Description

Left some tests that need data. Those location attributes will need to be added by hand.

### DIFF
--- a/tests/functional/i/import_error.rc
+++ b/tests/functional/i/import_error.rc
@@ -1,5 +1,9 @@
 [Messages Control]
 disable=C,R,W
+enable=multiple-imports
+
+[TYPECHECK]
+ignored-modules=external_module,fake_module.submodule,foo,bar
 
 [testoptions]
 except_implementations=PyPy
--- a/tests/functional/i/import_outside_toplevel.rc
+++ b/tests/functional/i/import_outside_toplevel.rc
@@ -1,0 +1,2 @@
+[IMPORTS]
+allow-any-import-level=astroid
--- a/tests/checkers/unittest_imports.py
+++ b/tests/checkers/unittest_imports.py
@@ -26,7 +26,7 @@ import astroid
 from pylint import epylint as lint
 from pylint.checkers import imports
 from pylint.interfaces import UNDEFINED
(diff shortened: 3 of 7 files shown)


tests/functional/i/import_error.rc
```diff
@@ -1,5 +1,9 @@
 [Messages Control]
 disable=C,R,W
+enable=multiple-imports
```
Pierre-Sassoulas, line 3: Maybe we should create a specific file for this one as it affect the whole pre-existing file.

tests/functional/i/import_outside_toplevel.rc
```diff
@@ -0,0 +1,2 @@
+[IMPORTS]
+allow-any-import-level=astroid
```
Pierre-Sassoulas, line 2: I think this option is specific enough to be added in the same file.

Pierre-Sassoulas (approved):
Amazing work once again, I have a nitpick, you can disregard it if not applicable.

tests/functional/i/import_error.rc
DanielNoord: I just added this because the original test also checked this message. I think this was mostly a by-product of the actual thing it was trying to test. because all messages are enabled I needed to re-enable it. There are actual tests for `multiple-imports`. I can either remove this and the test will pass without showing that this message is also raised or I can leave as is. Moving to other file doesn't make a lot of sense since the actual message that is being tested is `import-error`.

tests/functional/i/import_outside_toplevel.rc
DanielNoord: Will do!

tests/functional/i/import_error.rc
Pierre-Sassoulas: Ok, if you thought about it let's keep it that way !

tests/functional/i/import_outside_toplevel.rc
Pierre-Sassoulas: I meant it's ok there's nothing to change here 😄

PR #5435, opened 2021-11-29
Move ``TestDocstringCheckerYield`` to functional tests

DanielNoord:

## Type of Changes

## Description

@Pierre-Sassoulas I've tried to keep a somewhat sane commit history..
This was quite a lot of work and there are 4/5 files like this 😢


Pierre-Sassoulas (changes requested):
> I've tried to keep a somewhat sane commit history. This was quite a lot of work

Yeah, I can see that, my bad, I'll trust you for the next one 😄  It's really easy to check with a commit for each thank you. On the other hand copy pasting code is not something with a high probability of mistake either (the  'touchy' part is the config and if it's important and forgotten the test will fail without it).

I just have a remark regarding the place where the functional test were added. Right now extension's functional tests are in ``tests/functional/ext/extension_name``, could you move it in docparams ? I know it's not really organized right now. In particular it's not intuitive that extension should be in ``ext`` and not their first letter dir right now. This is something we'll likely discuss in #5396

DanielNoord:
> > I've tried to keep a somewhat sane commit history. This was quite a lot of work
> 
> Yeah, I can see that, my bad, I'll trust you for the next one 😄 It's really easy to check with a commit for each thank you. On the other hand copy pasting code is not something with a high probability of mistake either (the 'touchy' part is the config and if it's important and forgotten the test will fail without it).

No I think it was actually a fair comment. I just should have split the test into the different styles from the beginning and not at the end. That would have made the history a lot clearer. I will keep that in mind for the other `docparams` files 😄 

> I just have a remark regarding the place where the functional test were added. Right now extension's functional tests are in `tests/functional/ext/extension_name`, could you move it in docparams ? I know it's not really organized right now. In particular it's not intuitive that extension should be in `ext` and not their first letter dir right now. This is something we'll likely discuss in #5396

Will do! I guess you approve of the rest? I'll merge after the move then!

Pierre-Sassoulas:
>  I guess you approve of the rest?

Yeap :) !

PR #5447, opened 2021-11-30
Deprecate ``is_inside_lambda`` from utils

DanielNoord:

## Type of Changes

## Description

Follow up to #5331. We might as well use this new `util` for all similar type of functions


pylint/checkers/utils.py
```diff
@@ -294,6 +295,11 @@ class InferredTypeError(Exception):
 
 def is_inside_lambda(node: nodes.NodeNG) -> bool:
     """Return whether the given node is inside a lambda"""
+    warnings.warn(
+        "utils.is_inside_lambda will be removed in favour of calling "
+        "utils.get_node_first_ancestor_of_type(x, nodes.Lambda)",
```
Pierre-Sassoulas, line 300:
```suggestion
        "utils.get_node_first_ancestor_of_type(x, nodes.Lambda) in pylint 3.0",
```

Pierre-Sassoulas (changes requested):
A non trivial conflict appeared following the merge of the visit_name refactor, and it would also be nice to have a test for the deprecation warning.

Pierre-Sassoulas (approved):
👍

cdce8p:
It isn't a huge deal, but technically this PR adds a new `DeprecationWarning`. Going forward, I think it would make sense to do those changes only for minor releases. I.e. `2.13.0`.

Anyway, `2.12.2` is already "special" ;)

DanielNoord:
Ah sorry, little too trigger happy 😄

PR #5454, opened 2021-12-02
Fix false positive for ``used-before-assignment`` for vars in methods

### LINKED ISSUE #5342: Regression `used-before-assignment` with type annotation
### Bug description

```py
from typing import Optional

class A:
    def func(self) -> None:
        _node: Optional[A] = self  # false-positive: used-before-assignment
```

Noticed this in `astroid` first -> [astroid/nodes/node_ng.py](https://github.com/PyCQA/astroid/blob/96ca4dc75aacaba4d1f4e9a66a72699c9fc93349/astroid/nodes/node_ng.py#L435). The example above is valid as the function body isn't evaluated until called at witch time the class is already defined. This is a regression in `main` compared to `v2.11.1`.

/CC: @DanielNoord Haven't looked at it yet, but I would guess this might be related to the work you've been doing in this area.

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
E0601: Using variable 'A' before assignment (used-before-assignment)
```

### Expected behavior

No error

### Pylint version

```shell
pylint 2.11.2-dev0
astroid 2.8.6-dev0
Python 3.10.0 (v3.10.0:b494f5935c, Oct  4 2021, 14:59:20) [Clang 12.0.5 (clang-1205.0.22.11)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

/CC: @cdce8p Finally fixed that false positive. Turned out it was a minor fix after all.

Closes #5342

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,11 @@ Release date: TBA
 * Fixed a false positive for ``unused-import`` where everything
   was not analyzed properly inside typing guards.
 
+* Fixed a false-positive regression for ``used-before-assignment`` for
+  typed variables in the body of class methods that reference the same class
+
+  Closes #5342
+
 * Specified that the ``ignore-paths`` option considers "\" to represent a
   windows directory delimiter instead of a regular expression escape
   character.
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (approved):
👍

PR #5456, opened 2021-12-02
Add endLine and endColumn keys to JSONReporter

### LINKED ISSUE #5380: Add ``end_line`` and ``end_column`` to the output of the ``JSONReporter``
### Current problem

These attributes were added in `astroid` `2.9.0` to most nodes and should be accessible from the output of the ``JSONReporter``.

### Desired solution

Basically all changes in this reverted commit should be committed to `3.0`:

https://github.com/PyCQA/pylint/pull/5372/commits/6e912254dbc1e2e8b89f4e80cd859df42c07aa14

### Additional context

_No response_

cdce8p:
## Type of Changes

## Description

Based on the work of @DanielNoord in https://github.com/PyCQA/pylint/commit/6e912254dbc1e2e8b89f4e80cd859df42c07aa14, I've added the `endLine` and `endColumn` keys to the output of the `JSONReporter`. These are necessary for support in vscode-python.

I've explicitly chosen `endLine` (not `end_line`) and `endColumn` as those are better compatible with javascript syntax.

Closes #5380


Pierre-Sassoulas (commented):
Looks good, isn't it a small breaking change dependending on the way the json is read ? Thinking about  ``Message(**json.load(..))`` here. Regarding breaking change we may as well do them properly in a major, where they are expected instead of multiple small one released with minors where they are not ? We already had multiple downstream bug following breaking change in things we considered internal. We don't want downstream libs to not have enough trust to recover our minors automatically right ?

cdce8p:
> Looks good, isn't it a small breaking change depending on the way the json is read ?

In theory, it could be. The problem is that we need this change for error range support in vscode-python. At least I've not been able to come up with an alternative implementation that's also backwards compatible.

With the json output, it a) only effects a limited audience to begin with, b) should, for the most part, be backwards compatible with existing use cases, and c) if it does break, users should be able to work around it.

I'm open to suggestions, if anyone has a better idea.

--
Usually I too would prefer such changes in a major version (eg. 3.0), I just don't think that makes much sense here. Neither does waiting until we are ready for it as the vscode-python PR is currently blocked. It isn't ideal, but thats the unfortunate tradeoff for working with an old codebase.

Pierre-Sassoulas (approved):
Let's make the tradeoff then, having range in vscode now as opposed to when 3.0 is released months or years is a powerful motivator.

PR #5457, opened 2021-12-02
Fix assigning-non-slot false positive with setattr

### LINKED ISSUE #3793: False positive assigning-non-slot (E0237) in a class with __slots__ + __setattr__
### Steps to reproduce
```py
"Demonstrate __slots__ + __setattr__ false positive"


class Foo:
    "Simple class with slots and setattr"

    __slots__ = [
        "foo",
    ]

    def __init__(self, foo, bar):
        self.foo = foo
        self.bar = bar

    def __setattr__(self, key, value):
        if key in self.__slots__:
            getattr(type(self), key).__set__(self, value)
        else:
            print(key, value)
```

### Current behavior
```
E0237: Assigning to attribute 'bar' not defined in class slots (assigning-non-slot)
```

### Expected behavior
No error.

### pylint --version output
pylint 2.6.0
astroid 2.4.2
Python 3.7.6 (default, Jan  8 2020, 19:59:22) 
[GCC 7.3.0]

jakelishman:
- [N/A] If it's a new feature, or an important bug fix, add a What's New entry in

## Type of Changes

## Description

Previously, if a class was slotted and overrode `__setattr__`,
`assigning-non-slot` would be issued when assigning to attributes.  With
`__setattr__` defined, we cannot infer if it is an error to assign to an
attribute, so we suppress the error.

Fix #3793

I came across that same behaviour in Qiskit/qiskit-terra#7347, and found the old issue #3793 describing it.  That's marked "high effort" whereas this is a very simple patch, so apologies if I've missed some complicating circumstances here.


jakelishman:
I've fixed the merge conflict in `CONTRIBUTORS.rst`, and moved the change log entry I added into the 2.13.0 release section, to match the label applied to this PR.

Pierre-Sassoulas:
Thank you ! There will be a lot of conflicts with the first other MR being merged in main (there's only one place to put the changelog right now so conflicts are inevitable), please be mindful of your time and wait for a first review before rebasing again 😄

jakelishman:
Ah sure, no problem!

Pierre-Sassoulas (approved):
It look like the fix was indeed simpler than anticipated at first. Thank you for the very straightforward fix ! Small change-set, well tested, and a proper changelog 👌

ChangeLog
```diff
@@ -16,6 +16,11 @@ Release date: TBA
 
   Closes #85, #2615
 
+* Fixed a false positive for ``assigning-non-slot`` when the slotted class
```
DanielNoord, line 19: Could you add an entry to the changelog stored in the `2.13` file?

DanielNoord (changes requested):
Small change to fix the changelog. Rest looks good to me, thanks!

ChangeLog
jakelishman: To check: the `docs/whatsnew/2.13.rst` file?

DanielNoord: Yes! Sorry, should have been clearer about that.

jakelishman: No problem, done in b78b0278!

DanielNoord (approved):

DanielNoord:
Thanks for the quick turnaround @jakelishman!

Pierre-Sassoulas:
Congratulation on becoming a contributor too :wink:

jakelishman:
Thanks!  I've used pylint a lot through general development, so it's nice I finally had a chance to give something back to it, no matter how small!

PR #5464, opened 2021-12-03
Require ``\`` for asterisks in Sphinx-style parameter docstrings

### LINKED ISSUE #5406: [v2.12 Regression] Asterisk argument `mising-param-doc` vs `anomalous-backslash-in-string`
### Bug description

Take the following function:

```py
def echo(*args):
    """
    Print the given arguments.
    """
    print(*args)
```

The `args` parameter can be added to the docstring (in Sphinx style) in 3 ways:

1. `:param args: Arguments to print.`
    This works with sphinx and pylint 2.11.1 and earlier (before #5175).  With pylint 2.12.0 and later it causes the following errors
   > echo.py:4:0: W9015: "*args" missing in parameter documentation (missing-param-doc)
   > echo.py:4:0: W9017: "args" differing in parameter documentation (differing-param-doc)
2. `:param *args: Arguments to print.`
    This works with pylint, but causes the following error with sphinx:
    > echo.py:docstring of echo.echo:3:Inline emphasis start-string without end-string.
3. `:param \*args: Arguments to print.`
    This works with sphinx, but causes the following error with pylint:
   > W1401: Anomalous backslash in string: '\*'. String constant might be missing an r prefix. (anomalous-backslash-in-string)

### Configuration

```ini
[MASTER]
```

### Command used

```shell
pylint echo.py
```

### Pylint output

For method 1 (with pylint 2.12.1):

    echo.py:4:0: W9015: "*args" missing in parameter documentation (missing-param-doc)
    echo.py:4:0: W9017: "args" differing in parameter documentation (differing-param-doc)

For method 2:

No errors.

For method 3:

    echo.py:8:11: W1401: Anomalous backslash in string: '\*'. String constant might be missing an r prefix. (anomalous-backslash-in-string)

### Expected behavior

I would expect pylint to accept the syntaxes that Sphinx does (1 and 3 from above).

I'd prefer if pylint rejected syntaxes which Sphinx rejects (2 from above).  However, since Sphinx accepts asterisk-without-backslash for Google- and NumPy-style docstrings (as requested in #3733) if these can't be easily distinguished by pylint, it'd probably be preferable to accept them all.

Note:  This issue has affected other linters and it was noted that sphinx.ext.napoleon (which adds support for NumPy and Google style docstrings) auto-escapes \* at the beginning of arguments:  https://github.com/peterjc/flake8-rst-docstrings/issues/18#issuecomment-661040265  This is likely why those styles differ from Sphinx style in what is accepted.

### Pylint version

```shell
pylint 2.12.1
Sphinx 4.3.0
```

### OS / Environment

Debian GNU/Linux bookworm/sid

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #5406 

This took way longer than it should have.. Never do regex on a Friday..


Pierre-Sassoulas:
> Never do regex on a Friday..

``s/' on a Friday.'/''/g`` 😄

Pierre-Sassoulas (approved):
I was joking about regex, I like them fondly. But the approach with regex might have been an original design mistake. Good job managing with that though. The fix is quite dandy, Also the more tests we have the more manageable a migration from regex in the future will be if hard to fix bugs keep piling up. Chances are most case are handled and we won't have to change this too often.

DanielNoord:
> I was joking about regex, I like them fondly. But the approach with regex might have been an original design mistake. Good job managing with that though. The fix is quite dandy, Also the more tests we have the more manageable a migration from regex in the future will be if hard to fix bugs keep piling up. Chances are most case are handled and we won't have to change this too often.

I do to, but what is really annoying is how these patterns are written. It doesn't allow you to easily copy them over to regex101.com or a similar engine. That has really frustrated my process of working on these..

pylint/extensions/_check_docs_utils.py
```diff
@@ -377,6 +377,8 @@ def match_param_docs(self):
 
         for match in re.finditer(self.re_param_in_docstring, self.doc):
             name = match.group(2)
+            # Remove espace characters necessary for asterisks
```
jacobtylerwalls, line 380:
```suggestion
            # Remove escape characters necessary for asterisks
```

Of course it is now mandatory to the lovely piece of electronic music [Espace/Escape](https://www.youtube.com/watch?v=HaRsnJ1DZmA) by Francis Dhomont.

PR #5470, opened 2021-12-04
New checker - Detect use of unnecessary ellipsis

### LINKED ISSUE #5460: Add checker for unnecessary ellipsis
### Current problem

Add a checker for unnecessary ellipsis.

```
def my_function():
    """docstring & elipses"""
    ...  # nothing emitted

def my_second_function():
    """docstring & pass"""
    pass  # pylint emits: `my_test.py:7:4: W0107: Unnecessary pass statement (unnecessary-pass)`
```

### Desired solution

Emit a warning for this scenario, similar to when a docstring is immediately followed by `pass`.

### Additional context

_No response_

mbyrnepr2:
Closes #5460

Emit a warning when `...` is used and could be omitted. Similar to the `unnecessary-pass` checker.

## Type of Changes

## Description

Closes #5460


Pierre-Sassoulas (approved):
LGTM !

Pierre-Sassoulas:
Well except the fact it's in the string checker, it's not really related to strings, right ?

DanielNoord (changes requested):
Just had a quick look and want to interrupt before @Pierre-Sassoulas might merge this:

Expected output of the tests is incorrect as it is missing the end line and column. This is because `node` is not being passed to `add_message`. This should be done!

@Pierre-Sassoulas This is something for future reviews as well: `node` should be made mandatory (is on my todo list) but that requires some work. In the meantime we should require it for most new PRs.

Pierre-Sassoulas:
> node should be made mandatory (is on my todo list) but that requires some work.

You're right, my bad. Isn't that another breaking change for 3.0 ?

DanielNoord:
> You're right, my bad. Isn't that another breaking change for 3.0 ?

Yeah! But adding a `DeprecationWarning` now would make us trigger it ourself. I'd like to fix that before doing so.

Just to let you know, I currently plan to first fix unittests and then fix statement(). After this it's back to typing `PyLinter` in terms of maintenance. 
I'll do some simpler crash/bug fixes in between. :)

DanielNoord:
> You're right, my bad. Isn't that another breaking change for 3.0 ?

Yeah! But adding a `DeprecationWarning` now would make us trigger it ourself. I'd like to fix that before doing so.

Just to let you know, I currently plan to first fix unittests and then fix statement(). After this it's back to typing `PyLinter` in terms of maintenance. 
I'll do some simpler crash/bug fixes in between. :)

Pierre-Sassoulas:
>  adding a DeprecationWarning now would make us trigger it ourself.

Sounds good !

pylint/checkers/base.py
```diff
@@ -2270,6 +2270,32 @@ def _check_docstring(
             )
 
 
+class EllipsisChecker(_BasicChecker):
```
Pierre-Sassoulas, line 2273: Please create a new file ``pylint/checker/ellipsis_checker.py``, as we want to burst base.py in #5489

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas (approved):
Thanks for separating the checker 👍

DanielNoord:
Note that I haven't done a full review but I think the 18 prefix is already being used. I'm on mobile so I can't link the file, but there is a script in the scripts folder that can point you to a new unused prefix.

Pierre-Sassoulas:
> I think the 18 prefix is already being used

We could add a pre-commit script to check that, what do you think @DanielNoord ? Maybe based on the existing script for unused prefix.

DanielNoord:
There is an open issue about this to which I assigned myself. Haven't gotten around to it because the unittests move take so much time and I think they are more important.

mbyrnepr2:
Thanks for the advice @Pierre-Sassoulas & @DanielNoord 👍

ChangeLog
```diff
@@ -47,6 +47,10 @@ Release date: TBA
 
   Closes #5323
 
+* Add checker ``unnecessary-ellipsis``
```
DanielNoord, line 50:
```suggestion
* Add checker ``unnecessary-ellipsis``:  Emmitted when the ellipsis constant is used unnecessarily.
```

pylint/checkers/ellipsis_checker.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class EllipsisChecker(BaseChecker):
+    __implements__ = (IAstroidChecker,)
+    name = "unnecessary_ellipsis"
+    msgs = {
+        "W2301": (
+            "Unnecessary ellipsis constant",
+            "unnecessary-ellipsis",
+            "Used when the ellipsis constant is encountered and can be avoided.",
```
DanielNoord, line 18: Let's add an explanation as to why these are unnecessary. New python users might not know why they are and this message description doesn't necessarily help them avoid future warnings.

pylint/checkers/ellipsis_checker.py
```diff
@@ -0,0 +1,41 @@
...
+        )
+    }
+
+    @check_messages("unnecessary-ellipsis")
+    def visit_const(self, node: nodes.Const) -> None:
+        """Check if the ellipsis constant is used unnecessarily"""
```
DanielNoord, line 24: Could you add an explanation of when an ellipsis is unnecessary? That would help potential contributors understand what these `if` statements are trying to determine.

tests/functional/u/unnecessary/unnecessary_ellipsis.txt
```diff
@@ -0,0 +1,3 @@
+unnecessary-ellipsis:9:4:9:7::Unnecessary ellipsis constant:UNDEFINED
+unnecessary-ellipsis:43:4:43:7:docstring_and_ellipsis:Unnecessary ellipsis constant:UNDEFINED
+unnecessary-ellipsis:59:4:59:7:DocstringAndEllipsis:Unnecessary ellipsis constant:UNDEFINED
```
DanielNoord, line 3: @Pierre-Sassoulas This makes me start to wonder if we can add confidence to this message? Since it isn't based on inference?

DanielNoord (changes requested):
Code itself looks good, I would only add some additional comments and explanations.

I'm also wondering if we should add a confidence level to this message.

tests/functional/u/unnecessary/unnecessary_ellipsis.txt
Pierre-Sassoulas: Shouldn't we first add a new ``NO_INFERENCE`` value ?

DanielNoord: Probably! Let's leave this for now.

pylint/checkers/ellipsis_checker.py
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @check_messages("unnecessary-ellipsis")
+    def visit_const(self, node: nodes.Const) -> None:
+        """Check if the ellipsis constant is used unnecessarily.
+        Emit a warning when:
+         - A line consisting of an ellipsis is preceded by a docstring.
+         - A statement exists in the same scope as the ellipsis.
+           For example: A function consisting of an ellipsis followed by a
+           return statement on the next line."""
```
DanielNoord, line 32:
```suggestion
           return statement on the next line.
       """
```

DanielNoord (changes requested):
Thanks for these changes, much better!

I just thought of one additional test case I think we should add by default: overloads.
Could you add an example? The [mypy documentation](https://mypy.readthedocs.io/en/stable/more_types.html) should help with coming up with a simple but effective example.

DanielNoord (approved):
👍

PR #5473, opened 2021-12-05
Revert renaming extension tests

cdce8p:
## Description
Revert parts of #5374

The test file names for the `code_style` and `typing` extensions where intentionally prefixed to allow for easier testing.

```
pytest tests/test_functional -k typing
pytest tests/test_functional -k cs
```


Pierre-Sassoulas (approved):
Sorry, I thought that the file path was taken into account when using ``-k``, It's not the case anymore as we're using the name and not the file path when parametrizing. Good job on separating python 3.5 / 3.8 too !

PR #5476, opened 2021-12-05
fix deleting files for pyreverse tests

sgheppy:
This pr fix the annoying behavior introduced in pr #5272.

After this pr when pyreverse tests are executed some tests files aren't removed after execution.


Pierre-Sassoulas (approved):
Thank you doe the quick fix :)

DudeNr33 (approved):
Thanks for fixing this!

PR #5484, opened 2021-12-06
Move tests for Google docstrings from ``TestParamDocChecker`` to functional tests

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/tests/extensions/test_check_docs.py
+++ b/tests/extensions/test_check_docs.py
@@ -916,25 +916,6 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def test_ignores_ignored_argument_names_google(self) -> None:
-        """Example of a method documenting the return type that an
-        implementation should return.
-        """
-        node = astroid.extract_node(
-            """
-        class Foo(object):
-            def foo(self, arg, _): #@
-                '''docstring ...
-
-                Args:
-                    arg (int): An argument.
-                '''
-                pass
-        """
-        )
-        with self.assertNoMessages():
-            self.checker.visit_functiondef(node)
-
     def test_ignores_ignored_argument_names_numpy(self) -> None:
         """Example of a method documenting the return type that an
         implementation should return.
--- a/tests/extensions/test_check_docs.py
+++ b/tests/extensions/test_check_docs.py
@@ -696,40 +696,6 @@ class TestParamDocChecker(CheckerTestCase):
         with self.assertNoMessages():
             self.checker.visit_functiondef(node)
 
-    def test_finds_missing_raises_from_setter_google(self) -> None:
-        """Example of a setter having missing raises documentation in
-        the Google style docstring of the property
-        """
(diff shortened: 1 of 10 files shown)


tests/functional/ext/docparams/parameter/missing_param_doc_required_Google.py
```diff
@@ -16,3 +16,18 @@ def test_multi_line_parameters(param: int) -> None:
...
+    """Example of a function with missing Google style parameter
+    documentation in the docstring
+
+    Args:
+        x: bla
+        z (int): bar
+
+    some other stuff
+    """
+    pass
```
Pierre-Sassoulas, line 36: Wondering if we should do `print(x, y, z)` here and remove the disable. I'm not going to suggest for each commit as suggestion don't work in commit and this is an open question.

tests/functional/ext/docparams/parameter/missing_param_doc_required_Google.py
```diff
@@ -103,3 +103,13 @@ def test_non_builtin_annotations_in_google_docstring(
             topright: top right point of rectangle
     """
     pass
+
+
+def get_midpoint(bottomleft: Point, topright: Point) -> Point:
```
Pierre-Sassoulas, line 111: test_non_builtin_annotations_for_returntype_in_google_docstring

tests/extensions/test_check_docs.py
```diff
@@ -613,25 +610,6 @@ def my_func(named_arg, **kwargs):
         "tuple(int or str) or list(int or str)",
     ]
 
-    @pytest.mark.parametrize("complex_type", COMPLEX_TYPES)
```
Pierre-Sassoulas, line 1136: Maybe keeping the parametrize make sense here ? Good thing we can revert this commit in particular if you agree

Pierre-Sassoulas (commented):
Great ! I have some open questions before merging.

tests/functional/ext/docparams/parameter/missing_param_doc_required_Google.py
DanielNoord: I removed all `pass`'s

tests/extensions/test_check_docs.py
DanielNoord: I don't think we should keep tests in two places. I have merged them in:
https://github.com/PyCQA/pylint/runs/4445138080?check_suite_focus=true

Pierre-Sassoulas: > I don't think we should keep tests in two places.

Fair enough, I asked because the parametrize was quite elegant imo and I think removed a lot of duplication.

Pierre-Sassoulas: Your solution is very elegant :+1:

Pierre-Sassoulas (approved):

tests/extensions/test_check_docs.py
DanielNoord: 😄

PR #5489, opened 2021-12-07
Burst ``pylint.checkers base`` into multiple file

Pierre-Sassoulas:
## Type of Changes

## Description

``pylint.checker.base`` was 2600 lines long which was impacting analysis performance in IDE (more than 4s to get a new analysis after a save in Pycharm). This also permitted to find unused code and to remove it.


cdce8p:
We should do #5221 first. I still need to look at that though.

PR #5492, opened 2021-12-08
Move some of the Numpy tests out of ``TestParamDocChecker``

DanielNoord:

## Type of Changes

## Description

This does not cover all numpy tests as there are some issues with those that check `raise` documentation (it seems as if tests are incorrectly passing currently).
Just to make the other PR easier to review, let's merge this first!

### DIFF
--- a/tests/extensions/test_check_docs.py
+++ b/tests/extensions/test_check_docs.py
@@ -169,37 +169,6 @@ class TestParamDocChecker(CheckerTestCase):
         ):
             self.checker.visit_functiondef(node)
 
-    def test_finds_args_without_type_numpy(self) -> None:
-        node = astroid.extract_node(
-            '''
-        def my_func(named_arg, typed_arg: bool, untyped_arg, *args):
-            """The docstring
-
-            Args
-            ----
-            named_arg : object
-                Returned
-            typed_arg
-                Other argument without numpy type annotation
-            untyped_arg
-                Other argument without any type annotation
-            *args :
-                Optional Arguments
-
-            Returns
-            -------
-                object or None
-                    Maybe named_arg
-            """
-            if args:
-                return named_arg
-        '''
-        )
-        with self.assertAddsMessages(
-            MessageTest(msg_id="missing-type-doc", node=node, args=("untyped_arg",))
-        ):
-            self.checker.visit_functiondef(node)
-
     def test_finds_args_with_xref_type_numpy(self) -> None:
         node = astroid.extract_node(
             '''
--- a/tests/extensions/test_check_docs.py
+++ b/tests/extensions/test_check_docs.py
@@ -360,29 +360,6 @@ class TestParamDocChecker(CheckerTestCase):
         ):
             self.checker.visit_return(node)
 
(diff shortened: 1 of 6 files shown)


tests/functional/ext/docparams/parameter/missing_param_doc_required_Numpy.py
```diff
@@ -0,0 +1,6 @@
+"""Tests for missing-param-doc and missing-type-doc for Numpy style docstrings
```
Pierre-Sassoulas, line 1: I'm curious, do you create this commit at the very end by rebasing and taking the value you had after moving everything ?

tests/extensions/test_check_docs.py
```diff
@@ -25,7 +25,6 @@
 in particular the parameter documentation checker `DocstringChecker`
 """
 
-# pylint: disable=too-many-public-methods
```
Pierre-Sassoulas, line 28: 👌

Pierre-Sassoulas (approved):
🎉 100 lines less to maintain, again  ! Great work !

tests/functional/ext/docparams/parameter/missing_param_doc_required_Numpy.py
DanielNoord: Taken from `missing_param_doc_required_Google` as they tend to have the same tests and thus similar ignores. I don't think there are any unnecessary ignores in here (and it doesn't really hurt if there are as long as they are not the messages we want to test)

Pierre-Sassoulas: Ho alright, thank you for explaining :)

Pierre-Sassoulas:
Nice job finding the problem in tests ! Moving then fixing would permit to have a better history by just squashing both, but if it takes too much time I'm okay with mixing too.

DanielNoord:
Moving>fixing is (nearly) impossible as I can't let the tests pass without fixing or accepting false positives.

I'll have a look somewhere this week what the best effort/reward balance would be.

Pierre-Sassoulas:
Ok, I see the problem now, let's do both at the same time then, there's no point in fixing then moving it's more work for only theoretical benefit.

DanielNoord:
@Pierre-Sassoulas The issue is more troublesome than I thought:
https://github.com/PyCQA/pylint/blob/a51a5486ebff7543ae4fb6943fac2558947fa974/pylint/extensions/_check_docs_utils.py#L529-L536

This is a pretty naive approach at validating the docstrings. What happens for the remaining tests is that the `property_returns_lines` matches the `Google` style as it is similar to `Numpy`. Therefore this method returns `True`, even though the raises sections does not match the `Google` style.

https://github.com/PyCQA/pylint/blob/a51a5486ebff7543ae4fb6943fac2558947fa974/pylint/extensions/_check_docs_utils.py#L175-L184
Because of the order of this tuple we match with `Google` before `Numpy` and thus never actually see that the docstring we're testing is not `Google`, but `Numpy.

To fix this we need to change how we validate docstrings. Something that takes into account when we match multiple types and then fixes the conflict between them.

Issue #5479, opened 2021-12-06
Unexpected crash when inferring attribute on node


tushar-deepsource:
### Bug description

It's hard to reproduce this bug. But it happened when I was trying to run pylint over [this repository](https://github.com/School-Simplified/Timmy-SchoolSimplified)

Here's the logs, they should be fairly helpful in tracking down the issue:

```log
Exception on node <AssignAttr.read_messages l.1027 at 0x11ed3bf70> in file '/tmp/Timmy-SchoolSimplified/utils/events/TicketDropdown.py'
Traceback (most recent call last):
  File "/private/tmp/Timmy-SchoolSimplified/venv/lib/python3.10/site-packages/pylint/utils/ast_walker.py", line 75, in walk
    callback(astroid)
  File "/private/tmp/Timmy-SchoolSimplified/venv/lib/python3.10/site-packages/pylint/checkers/classes.py", line 1459, in visit_assignattr
    self._check_in_slots(node)
  File "/private/tmp/Timmy-SchoolSimplified/venv/lib/python3.10/site-packages/pylint/checkers/classes.py", line 1511, in _check_in_slots
    if node.attrname in klass.locals and _has_data_descriptor(
  File "/private/tmp/Timmy-SchoolSimplified/venv/lib/python3.10/site-packages/pylint/checkers/classes.py", line 385, in _has_data_descriptor
    attributes = cls.getattr(attr)
  File "/private/tmp/Timmy-SchoolSimplified/venv/lib/python3.10/site-packages/astroid/nodes/scoped_nodes.py", line 2596, in getattr
    raise AttributeInferenceError(
astroid.exceptions.AttributeInferenceError: 'read_messages' not found on <ClassDef.PermissionOverwrite l.602 at 0x112e2d720>.
```

### Command used

```shell
pylint [spaces separated list of all python files in the repo]
```

### Pylint output

```shell
Exception on node <AssignAttr.read_messages l.1027 at 0x11ed3bf70> in file '/tmp/Timmy-SchoolSimplified/utils/events/TicketDropdown.py'
Traceback (most recent call last):
  File "/private/tmp/Timmy-SchoolSimplified/venv/lib/python3.10/site-packages/pylint/utils/ast_walker.py", line 75, in walk
    callback(astroid)
  File "/private/tmp/Timmy-SchoolSimplified/venv/lib/python3.10/site-packages/pylint/checkers/classes.py", line 1459, in visit_assignattr
    self._check_in_slots(node)
[...]

PR #5494, opened 2021-12-08
Fix crash with slots and annotated assignments

DanielNoord:

## Type of Changes

## Description

This fixes the issue reported in https://github.com/PyCQA/pylint/issues/5479#issuecomment-988798391

I'm not actually sure if this is the expected output. 
Should this code raise `assigning-non-slot`? If not, the code should be changed. If just added the test code from the issue + a fix to get initial feedback on it.


DanielNoord:
@mbyrnepr2 Seeing as you opened #5617 and seem to be experienced with `slots`, could I ask you to take a look at this? Does this change look okay? You can look at the CI output to see what the current output for the added test is.

Pierre-Sassoulas (changes requested):
This need to be rebased following #5497, sorry. You could be able to cherry-pick the commit onto main even if a rebase would conflict..

DanielNoord:
That's an interesting Github bug. `git` saw no issue merging main into this branch. Didn't need to make any changes for this to work..

mbyrnepr2:
Hi @DanielNoord I think your change looks good.

Regarding your question above about whether `assigning-non-slot` should be emitted, I think it also behaves as expected - I.e: `assigning-non-slot` is **not emitted** in your test case.

If I understand this checker, and if it were to emit a warning, then running Python on the example would also result in something similar to: 
`AttributeError: 'Connector' object has no attribute '__reader'`.

As a simplified example of your case, have no issues when running Python:
```
class MyClass:
    __slots__ = ("a", "b")

    @classmethod
    def mymethod(cls):
        cls.c = "x"

MyClass.mymethod()
```

In the following alternative case get an AttributeError:
```
class MyClass2:
    __slots__ = ("a", "b")

    def mymethod(self):
        self.c = "x"

MyClass2().mymethod()
```
`AttributeError: 'MyClass2' object has no attribute 'c'`

Let me know if I've missed anything!

DanielNoord:
@mbyrnepr2 Thanks for your review! Your examples make sense indeed.

I have updated the test options so they should no longer fail on pypy and 3.6. This is ready for review for others!

pylint/checkers/classes/class_checker.py
```diff
@@ -1495,11 +1495,17 @@ def _check_in_slots(self, node):
                     # Properties circumvent the slots mechanism,
                     # so we should not emit a warning for them.
                     return
-                if node.attrname in klass.locals and _has_data_descriptor(
-                    klass, node.attrname
-                ):
-                    # Descriptors circumvent the slots mechanism as well.
-                    return
+                if node.attrname in klass.locals:
+                    for local_name in klass.locals.get(node.attrname, []):
```
Pierre-Sassoulas, line 1499:
```suggestion
                if node.attrname in klass.locals:
                    for local_name in klass.locals.get(node.attrname):
```
```suggestion
                for local_name in klass.locals.get(node.attrname, []):
```
We know `node.attrname` is in `klass.locals` at this point

Pierre-Sassoulas (approved):
LGTM !

DanielNoord:
@Pierre-Sassoulas  I think that change broke the tests :(

pylint/checkers/classes/class_checker.py
```diff
@@ -1495,11 +1495,17 @@ def _check_in_slots(self, node):
...
+                        statement = local_name.statement(future=True)
+                        if (
+                            isinstance(statement, nodes.AnnAssign)
+                            and not statement.value
+                        ):
+                            return
+                    if _has_data_descriptor(klass, node.attrname):
+                        # Descriptors circumvent the slots mechanism as well.
```
Pierre-Sassoulas, line 1507:
```suggestion             
                for local_name in klass.locals.get(node.attrname, []):
                    statement = local_name.statement(future=True)
                    if (
                        isinstance(statement, nodes.AnnAssign)
                        and not statement.value
                    ):
                        return
                if node.attrname in klass.locals and _has_data_descriptor(klass, node.attrname):
                    # Descriptors circumvent the slots mechanism as well.
```
Or the first suggestion in the previous comment (no default value)

PR #5497, opened 2021-12-09
Burst pylint.checker.classes

Pierre-Sassoulas:
## Type of Changes

## Description

pylint.checker.classes was huge (mainly because the Class Checker is itself huge). It's impacting analysis performance in IDE. Follow-up to an effort to have one Checker class per file started in #5489


pylintrc
```diff
@@ -171,7 +171,7 @@ single-line-if-stmt=no
 no-space-check=trailing-comma,dict-separator
 
 # Maximum number of lines in a module
-max-module-lines=2000
+max-module-lines=1500
```
Pierre-Sassoulas, line 174: This does not seem to work.

Pierre-Sassoulas:
I want to merge this fast because a single modification on it would create a conflict during rebase/merge. @cdce8p I see you self-requested a review do you mind if I merge now  and we handle remarks in a subsequent MR later ?

cdce8p:
> I want to merge this fast because a single modification on it would create a conflict during rebase/merge. @cdce8p I see you self-requested a review do you mind if I merge now and we handle remarks in a subsequent MR later ?

Could you give me time until this evening? I'll try to look at it as soon as I have time later today.

pylint/checkers/classes/class_checker.py
```diff
@@ -49,19 +49,33 @@
...
+    _has_data_descriptor,
+    _has_different_parameters_default_value,
+    _has_same_layout_slots,
+    _is_attribute_property,
+    _is_invalid_base_class,
+    _signature_from_arguments,
+    _signature_from_call,
+)
+from pylint.checkers.classes.different_parameters import _different_parameters
+from pylint.checkers.classes.scope_access_map import ScopeAccessMap
```
cdce8p, line 76: All these are only use in this file. I think it would be better to keep them together.
Otherwise we just end up with a lot of util files everywhere.

pylintrc
cdce8p: I'm not a huge fan of hard limits. Sometimes it just makes more sense to keep things together even if that would be a few lines over the limit.

pylint/checkers/classes/class_checker.py
```diff
@@ -49,19 +49,33 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
-"""classes checker for Python code
-"""
-import collections
-from itertools import chain, zip_longest
-from typing import List, Pattern
+"""classes checker for Python code """
```
cdce8p, line 52:
```suggestion
"""Classes checker for Python code"""
```

cdce8p (changes requested):
I understand that you would like to split up these large files and it does somewhat make sense here.

However, I don't think we should create different helper files just to move a few hundred lines. If functions and classes are only used in one checker, in general we should try to keep them together. Otherwise there is a strong risk that we end up with too many different util files.

My recommendation would be a similar structure to `pylint/checkers/refactoring/`.

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: Yeah I wondered about that too. There's a practical aspect I considered: the parsing time of the file when you work on it. The main pain point is the ClassChecker itself being so big though and creating util file does not solve the main design problem. I'll revert.

Pierre-Sassoulas:
Changing to draft as the order of helper functions changed.

pylint/checkers/classes/class_checker.py
cdce8p: At least in VS Code, I don't find the parsing time to be an issue.
Of course it takes a few seconds until the pylint warnings show up but that isn't a problem for me.

Pierre-Sassoulas: For 2000+ line file Pycharm analysis takes more than 5s on my setup (which is not high end but not that bad either). I guess low budget work station would have even more lag with our biggest files.

Pierre-Sassoulas: (Pycharm is a JetBrain IDE, not using pylint as linter)

pylint/checkers/classes/different_parameters.py
```diff
@@ -0,0 +1,121 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+"""Deal with parameters overriding in two methods."""
```
cdce8p, line 4: The functions here should be part of `class_checker.py`. They aren't used anywhere else.

pylint/checkers/classes/scope_access_map.py
```diff
@@ -0,0 +1,31 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+
+import collections
+
+from pylint.checkers.utils import node_frame_class
```
cdce8p, line 7: The function and class here should be part of `class_checker.py`. They aren't used anywhere else.

pylintrc
cdce8p: Can you remove the change here? If at all, it's probably better to do that separately.

cdce8p (changes requested):

pylint/checkers/classes/class_checker.py
cdce8p: You should really try VS Code someday 😄 I use a few years old, 2 core MacBook Air.

Pierre-Sassoulas: I'm also in a VM that do not have enough disk space 😄 Help to commiserate with developer without a good equipment :)

PR #5498, opened 2021-12-10
Fix incorrect classification of property docstrings in Numpy-style

DanielNoord:

## Type of Changes

## Description

See short discussion in https://github.com/PyCQA/pylint/pull/5492#issuecomment-990986957

The test that has been moved to the functional tests wouldn't pass without this change. Because of the property description in the docstring this gets considered to be a `Google` style and then the `Raises` section is not correctly recognised.
The use of `is_valid` is pretty naive and I think this fix (although also pretty naive) is an improvement: it selects the style for which most sections get recognised.

I wonder if this is a breaking change: removing `is_valid` and requiring `matching_sections` on `DocString` might be troublesome for plugins that define their own styles. Do we know if there are actually plugins that do so?
We could think of finding a way to make this backwards compatible, but I could not think of something right away and wanted to see whether it is actually needed.

### DIFF
--- a/pylint/extensions/_check_docs_utils.py
+++ b/pylint/extensions/_check_docs_utils.py
@@ -172,7 +172,8 @@ def possible_exc_types(node):
         return set()
 
 
-def docstringify(docstring, default_type="default"):
+def docstringify(docstring: str, default_type: str = "default") -> "Docstring":
+    best_match = (0, DOCSTRING_TYPES.get(default_type, Docstring)(docstring))
     for docstring_type in (
         SphinxDocstring,
         EpytextDocstring,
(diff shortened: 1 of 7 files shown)


pylint/extensions/_check_docs_utils.py
```diff
@@ -210,8 +211,9 @@ def __init__(self, doc):
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__}:'''{self.doc}'''>"
 
-    def is_valid(self):
-        return False
+    def matching_sections(self) -> int:
```
Pierre-Sassoulas, line 214:
```suggestion
    def is_valid(self) -> bool:
        return self.matching_sections() > 0

    def matching_sections(self) -> int:
```
To avoid the breaking change ?

Pierre-Sassoulas (commented):
The new mechanism sounds better than the previous one. It seems we never had complaints about that so it must not happen very often to mix style in the real world.

pylint/extensions/_check_docs_utils.py
DanielNoord: I thought about this, but then the method name doesn't match its behaviour..

However, this is an extension so is this actually part of the public API?

Pierre-Sassoulas: Yeah you're right it would not be what someone expect from "is_valid". Let's consider it's not a public API.

Pierre-Sassoulas (approved):

PR #5505, opened 2021-12-11
Fix #5504: Fix crash if the output of items() is assigned to a 1-tuple

### LINKED ISSUE #5504: Crash ``list index out of range`` on simple and wrong python
### Bug description

I tried to find this example but was unable to.  Please note that pylint some times crashes on this example that I collapsed from a larger example, as I have run it multiple times.  I am submitting this as pylint shouldn't crash on badly written python code and this is such a small example.

```python
""" Python example """

mydict={
    "one":"1",
    "two":"2",
    "three": "3"
}

for x, in mydict.items():
    {
    print(x,mydict[x])
    }

```

pylint crashed with a ``IndexError`` and with the following stacktrace:
```
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/lint/pylinter.py", line 1034, in _check_files
    self._check_file(get_ast, check_astroid_module, file)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/lint/pylinter.py", line 1069, in _check_file
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/lint/pylinter.py", line 1203, in check_astroid_module
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/lint/pylinter.py", line 1250, in _check_astroid_module
    walker.walk(node)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/utils/ast_walker.py", line 75, in walk
    self.walk(child)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/checkers/refactoring/refactoring_checker.py", line 637, in visit_for
    self._check_unnecessary_dict_index_lookup(node)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/checkers/refactoring/refactoring_checker.py", line 1899, in _check_unnecessary_dict_index_lookup
    args=(node.target.elts[1].as_string()),
IndexError: list index out of range
```

### Configuration

_No response_

### Command used

```shell
pylint t001.py
```

### Pylint output

```shell
Exception on node <For l.9 at 0x7fbbb9ebd640> in file '/Users/Rob 1/Programming/python/openshift-acct-mgt/t001.py'
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/checkers/refactoring/refactoring_checker.py", line 637, in visit_for
    self._check_unnecessary_dict_index_lookup(node)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pylint/checkers/refactoring/refactoring_checker.py", line 1899, in _check_unnecessary_dict_index_lookup
    args=(node.target.elts[1].as_string()),
IndexError: list index out of range
************* Module t001
t001.py:1:0: F0001: Fatal error while checking 't001.py'. Please open an issue in our bug tracker so we address this. There is a pre-filled template that you can use in '/Users/Rob 1/Library/Caches/pylint/pylint-crash-2021-12-10-20.txt'. (fatal)

```

### Expected behavior

I would expect something other than a "Fatal Error", perhaps something indicating that C programmers shouldn't be programming in Python :).

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.9.0 (v3.9.0:9cf6752276, Oct  5 2020, 11:29:23) 
[Clang 6.0 (clang-600.0.57)]
```

### OS / Environment

Mac OSX 11.6.1

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Fixes crash if the dict view returned by `.items()` is partially unpacked to a 1-tuple, e.g. `for k, in d.items()` instead of `for k, _ in d.items()`

Closes #5504

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1882,6 +1882,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
                     if isinstance(value, nodes.Name):
                         if (
                             not isinstance(node.target, nodes.Tuple)
+                            # Ignore 1-tuples: for k, in d.items()
                             or len(node.target.elts) < 2
                             or value.name != node.target.elts[0].name
                             or iterating_object_name != subscript.value.as_string()
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1882,6 +1882,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
                     if isinstance(value, nodes.Name):
                         if (
                             not isinstance(node.target, nodes.Tuple)
+                            or len(node.target.elts) < 2
                             or value.name != node.target.elts[0].name
                             or iterating_object_name != subscript.value.as_string()
                         ):
--- a/tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
+++ b/tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
@@ -100,3 +100,10 @@ for key, val in outer_dict.items():
     for key_two, val_two in val.items():
         del outer_dict[key][key_two]  # [unnecessary-dict-index-lookup]
         break
+
+# Test partial unpacking of items
+# https://github.com/PyCQA/pylint/issues/5504
+
+d = {}
+for key, in d.items():
+    print(d[key])
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,11 @@ Release date: TBA
 
   Closes #4716
 
+* Fix crash in ``unnecessary-dict-index-lookup`` checker if the output of
+  ``items()`` is assigned to a 1-tuple.
+
+  Closes #5504
+
 * The ``PyLinter`` class will now be initialiazed with a ``TextReporter``
   as its reporter if none is provided.
 
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -26,6 +26,11 @@ Other Changes
 
   Closes #4716
 
+* Fix crash in ``unnecessary-dict-index-lookup`` checker if the output of
+  ``items()`` is assigned to a 1-tuple.
+
+  Closes #5504
+
 * Fix false negative for ``consider-iterating-dictionary`` during membership checks encapsulated in iterables
   or ``not in`` checks
 


pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1882,6 +1882,7 @@ def _check_unnecessary_dict_index_lookup(
                     if isinstance(value, nodes.Name):
```
Pierre-Sassoulas, line 1882: The comment above is outdated but I can't suggest on it directly. 

```diff
- # Case where .items is assigned to k,v (i.e., for k, v in d.items())
+ # Case where .items is assigned (i.e., for k, v in d.items() or for k, in d.items())
```

tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
```diff
@@ -100,3 +100,10 @@ class Foo:
     for key_two, val_two in val.items():
         del outer_dict[key][key_two]  # [unnecessary-dict-index-lookup]
         break
+
+# Test partial unpacking of items
+# https://github.com/PyCQA/pylint/issues/5504
+
+d = {}
+for key, in d.items():
+    print(d[key])
```
Pierre-Sassoulas, line 109: We should probably still raise an ``unnecessary-dict-index-lookup`` here but if that's too hard to do I'd merge a hot fix for the crash. 

```suggestion
for key, in d.items():
    print(d[key]) # [unnecessary-dict-index-lookup]
```

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas:
Thank you for working on this. It makes me want to implement a better release flow to be able to release hot fix for crashes. Because right now, as we did not catch it before the last 12.x release went out, this will have to wait for 2.13 to be released.

tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
jacobtylerwalls: Let me see if I follow. The dict index lookup is still necessary, because we don't have a `v` variable for the value, but you're saying you still want it to raise a msg? That would be more like `poor-unpacking-of-dict-items` more than `unnecessary-dict-index-lookup`, right? The index lookup is still needed.

pylint/checkers/refactoring/refactoring_checker.py
jacobtylerwalls: Isn't the comment still correct? Anything that's not assignment to a 2-tuple is still filtered out, i.e. I read this comment as applying to everything after the `continue`. We're just writing a better `continue` condition :-)

tests/functional/u/unnecessary/unnecessary_dict_index_lookup.py
Pierre-Sassoulas: I checked the functional tests and we have something similar:
```python
d = {'key': 'value'}

for k, _ in d.items():
    d[k] += 'VALUE'
    if 'V' in d[k]:  # This is fine, if d[k] is replaced with _, the semantics change
        print('found V')
```
Where we do not raise a warning. So you're right it should be another message entirely to stay coherent. (Use .value() instead of item(), or something.)

pylint/checkers/refactoring/refactoring_checker.py
Pierre-Sassoulas: Hmm, ok I did not see it like that but you're right. It will still be hard to tell what `` len(node.target.elts) < 2`` is doing without the context of this MR though. Maybe we could add a comment or create a named intermediary boolean ?

Pierre-Sassoulas (approved):
👍

PR #5506, opened 2021-12-11
Fix #5500: Fix false-positive `used-before-assignment` for assignments in except blocks following try blocks that return

### LINKED ISSUE #5500: Regression `used-before-assignment` with try-except
### Bug description

#5402 added the `used-before-assignment` error in cases an assignment only happens in `except` blocks.
While testing, I came across the example below. It's a special case, but I believe it should still be possible to exclude it.

* The `try` block should return
* Each `except` block defines the variable

```py
def func():
    try:
        value = 1 / 0
        return
    except ZeroDivisionError:
        data = 1
    except ValueError:
        data = 2
    print(data)  # used-before-assignment
```

/CC: @jacobtylerwalls

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
E0601: Using variable 'data' before assignment (used-before-assignment)
```

### Expected behavior

_No error_

### Pylint version

```shell
Current main
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Fixes false positive for `used-before-assignment` when evaluating statements after except handlers where the corresponding `try` returned. In this scenario, we should assume one of the except handlers _did_ execute.

There is the potential for a false-negative where only ONE of the except handlers defines the variable in question, but to check that, we would need to permit scenarios where the other except handlers merely raise or return--that case shouldn't emit a warning. I think the feasibility of that should be evaluated separately.

Closes #5500


Pierre-Sassoulas (approved):
LGTM, I think it's a clever fix to bring better control flow.

DanielNoord:
Fix itself seems fine, but this should be documented in the description of the `used-before-assignment` message. I'm not sure how to do this properly, especially since the original issue is also discussing another exception to the general "assume tries don't execute" rule. I think whatever we end up assuming should be clearly documented for end-users without the need of looking at our actual code or previous changelogs.

pylint/checkers/variables.py
```diff
@@ -405,7 +405,9 @@ def _has_locals_call_after_node(stmt, scope):
     "E0601": (
         "Using variable %r before assignment",
         "used-before-assignment",
-        "Used when a local variable is accessed before its assignment.",
+        "Used when a local variable is accessed before its assignment or "
+        "accessed only after an assignment in except/finally blocks if the "
+        "try block does not return.",
```
DanielNoord, line 410:
```suggestion
        "Emitted when a local variable is accessed before its assignment took place. "
        "Assignments in try blocks are assumed not to have occurred. Assignments "
        "in except/finally blocks are assumed to have occurred when the associated "
        "try block contains a return statement.",
```

Pretty lengthy, but this might be a bit more informative for end-users? I'm not sure..

tests/functional/u/use/used_before_assignment_issue5500.py
```diff
@@ -0,0 +1,9 @@
+"""used-before-assignment (E0601)"""
```
DanielNoord, line 1:
```suggestion
"""Tests for used-before-assignment with assignments in except handlers after
try blocks with return statements.
See: https://github.com/PyCQA/pylint/issues/5500.
"""
```

tests/functional/u/use/used_before_assignment_issue5500.py
```diff
@@ -0,0 +1,9 @@
...
+def function():
+    """Assume except blocks execute if the try block returns."""
+    try:
+        return "success message"
+    except ValueError:
```
DanielNoord, line 6: Let's also add a tests for the finally (as mentioned in the description of the message).

And, could you rename the test file? Personally I think we should try and make the file names descriptive as well. `issue5500` doesn't tell new contributors much about what tests will be included.

pylint/checkers/variables.py
jacobtylerwalls: Assignments in try blocks *are* assumed to have occurred when evaluating statements _after any_ except/finally handling. They are merely not assumed to have occurred when evaluating the **except/finally** statement, but not until #5384 is merged, so I didn't mention here.

And the suggested version isn't quite clear about the except/finally assumptions--they are assumed to not have executed in most cases, better to explain this than the exception.

jacobtylerwalls: Glad to have a wordsmith partner though! 💯   Let me know your thoughts on phrasing.

tests/functional/u/use/used_before_assignment_issue5500.py
jacobtylerwalls: Smart! (was just trying to figure out the norms!)

pylint/checkers/variables.py
DanielNoord: Second try 😄 

```suggestion
        "Emitted when a local variable is accessed before its assignment took place. "
        "Assignments in try blocks are assumed not to have occurred when evaluating "
        "associated except/finally blocks. Assignments in except/finally blocks are "
        "assumed not to have occurred, except when the associated try block "
        "contains a return statement.",
```

And agreed! You're previous docstrings and comments were very clear, no doubt we'll come up with something that's spot on for this one!

tests/functional/u/use/used_before_assignment_issue5500.py
DanielNoord: I don't think there are many norms for file names (see the regression folder...) but that doesn't stop us from starting to improve on them 😄

pylint/checkers/variables.py
jacobtylerwalls: This one looks great--except that the last part just applies to except, not finally. I just want to be clear it also reflects changes from #5384, so we should decide whether to merge that one first.

DanielNoord: Could you rephrase the last part and then apply? I approved #5384 knowing that this will get merged before `2.13` gets out. However, it would be even better if you could copy some of the text here and add it to that PR. Not sure how much work that will be for you, but would be appreciated I think! 😄

jacobtylerwalls: Yeah I've already committed my fix to your proposed change, just waiting to push until I address the comments below. Coming right up!

tests/functional/u/use/used_before_assignment_issue5500.py
jacobtylerwalls: > Let's also add a tests for the finally (as mentioned in the description of the message).

Good catch, this found an issue in the implementation. The finally executes before the try returns, so we need to continue to emit a message.

jacobtylerwalls: Fixed in 3b28e7d.

Pierre-Sassoulas: > was just trying to figure out the norms!

We're also doing that 😄 any ideas ? ;)

jacobtylerwalls:
Sorry for the extra force push, I pushed a comment to the wrong branch and removed it when I discovered it to avoid cluttering the branch history.

cdce8p (approved):
The change looks good and indeed fixes the false-positives.

One thing I noticed though is that the `filtered_nodes` list comprehension has gotten rather complicated with multiple nested ifs and comprehensions itself. This should probably be a function instead.

That would also allow `n.statement()` to be cached witch would improve performance at least a tiny bit.

Could you create a followup PR for it?

jacobtylerwalls:
> Could you create a followup PR for it?

Sure thing--I agree, the comprehension is quite ugly now.

PR #5509, opened 2021-12-13
Move tests from ``TestParamDocChecker`` to functional tests

DanielNoord:

## Type of Changes

## Description

Last move for ``TestParamDocChecker`` 😄


DanielNoord:
@Pierre-Sassoulas Do you want me to remove these unused `testutils`? Or do we want to keep them for future uses?

tests/functional/ext/docparams/raise/missing_raises_doc.py
```diff
@@ -96,3 +97,12 @@ def test_no_error_notimplemented_documented():
         NotImplementedError: When called.
     """
     raise NotImplementedError
+
+
+def do_somtest_finds_short_name_exceptionething():
```
Pierre-Sassoulas, line 102:
```suggestion
def test_finds_short_name_exception():
```

Pierre-Sassoulas (approved):
👍

tests/functional/ext/docparams/raise/missing_raises_doc.py
DanielNoord: 😄

Pierre-Sassoulas:
> Do you want me to remove these unused testutils?

Let's keep them so it's not a breaking change for plugin developer.

DanielNoord:
> Let's keep them so it's not a breaking change for plugin developer.

Shall I add a `DeprecationWarning`? `set_config_directly` was added in `2.12` (perhaps `2.11`) by me as a hacky solution to an issue with multiple checkers using the same config names. If you're using it you should probably redesign your checker 😄

Pierre-Sassoulas:
Sorry I missed that we just added them in 2.12.2 I think we could remove, I doubt anyone jumped on it.

DanielNoord:
> Sorry I missed that we just added them in 2.12.2 I think we could remove, I doubt anyone jumped on it.

It was `2.12`, see https://github.com/PyCQA/pylint/pull/5195

Still okay with removing that decorator?

Pierre-Sassoulas:
Do you want to revert 5195 ? I hope in the future there won't be the same argument in multiple checker in the configuration so removing it would not be a problem anymore. Meanwhile using functional test all the time has a lot of advantages.

DanielNoord:
I don't think reverting makes sense as #5195 actually built upon other changes to these decorators.

For now I would propose to deprecate `set_config_directly` in `2.13` and remove it in `2.14`:
If plugin developers are actively developing plugins that they jumped on this after `2.12` I think they should also notice this deprecation. Normally we should obviously remove it in `3.0` but since this is actually allowing badly designed checkers to be tested in a hacky way I don't think we should wait for it.
If you agree I'll merge this and then do a follow-up with the deprecation.

Pierre-Sassoulas:
The minor version deprecation makes a lot of sense, let's do that 👍

PR #5519, opened 2021-12-13
Modify the description of HIGH confidence in pylint.interfaces

Pierre-Sassoulas:
## Type of Changes

## Description

Pycharm raises a warning when we import something that is not in ``__all__``. We also need a new confidence level clearer than ``HIGH``.

### DIFF
--- a/pylint/interfaces.py
+++ b/pylint/interfaces.py
@@ -26,6 +26,19 @@ from astroid import nodes
 if TYPE_CHECKING:
     from pylint.reporters.ureports.nodes import Section
 
+__all__ = (
+    "IRawChecker",
+    "IAstroidChecker",
+    "ITokenChecker",
+    "IReporter",
+    "IChecker",
+    "HIGH",
+    "INFERENCE",
+    "INFERENCE_FAILURE",
+    "UNDEFINED",
+    "CONFIDENCE_LEVELS",
+)
+
 Confidence = namedtuple("Confidence", ["name", "description"])
 # Warning Certainties
 HIGH = Confidence("HIGH", "No false positive possible.")
@@ -102,6 +115,3 @@ class IReporter(Interface):
 
     def display_reports(self, layout: "Section") -> None:
         """display results encapsulated in the layout tree"""
-
-
-__all__ = ("IRawChecker", "IAstroidChecker", "ITokenChecker", "IReporter", "IChecker")
--- a/pylint/interfaces.py
+++ b/pylint/interfaces.py
@@ -42,7 +42,7 @@ __all__ = (
 
 Confidence = namedtuple("Confidence", ["name", "description"])
 # Warning Certainties
-HIGH = Confidence("HIGH", "No false positive possible.")
+HIGH = Confidence("HIGH", "Warning that is not based on inference result.")
 INFERENCE = Confidence("INFERENCE", "Warning based on inference result.")
 INFERENCE_FAILURE = Confidence(
     "INFERENCE_FAILURE", "Warning based on inference with failures."
--- a/pylint/interfaces.py
+++ b/pylint/interfaces.py
@@ -75,10 +75,10 @@ class IChecker(Interface):
     """
 
     def open(self):
-        """called before visiting project (i.e set of modules)"""
+        """called before visiting project (i.e. set of modules)"""
 
     def close(self):
-        """called after visiting project (i.e set of modules)"""
+        """called after visiting project (i.e. set of modules)"""
 
 
 class IRawChecker(IChecker):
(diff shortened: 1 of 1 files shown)


pylint/interfaces.py
```diff
@@ -26,16 +26,34 @@
...
+    "Confidence",
+    "NO_INFERENCE",
+    "HIGH",
+    "INFERENCE",
+    "INFERENCE_FAILURE",
+    "UNDEFINED",
+    "CONFIDENCE_LEVELS",
+)
+
 Confidence = namedtuple("Confidence", ["name", "description"])
```
Pierre-Sassoulas, line 44: I'd want to do an enum here, but it's probably not worth the breaking change. On the other hand maybe addint it to ``__all__`` send the message that it's a public API.

pylint/interfaces.py
```diff
@@ -26,16 +26,34 @@
...
 # Warning Certainties
+NO_INFERENCE = Confidence(
+    "NO_INFERENCE", "Warning that is not based on inference result."
+)
 HIGH = Confidence("HIGH", "No false positive possible.")
```
DanielNoord, line 49: What's the difference between `NO_INFERENCE` and `HIGH`?

pylint/interfaces.py
DanielNoord: Should `Confidence` be added here? Isn't that something we only use internally?

Pierre-Sassoulas: Right it's a remnant of the refactor to replace all constant by an enum containing the constants.

pylint/interfaces.py
Pierre-Sassoulas: HIGH is the old value I don't want to remove so there is not breaking change, but it's the same. I'm never confident enough to say there won't be any false positive.

Pierre-Sassoulas: I would have renamed HIGH but that would be a breaking change.

DanielNoord: Can't we just rephrase the description of `HIGH` then?

Pierre-Sassoulas: You're right.

pylint/interfaces.py
```diff
@@ -46,7 +59,7 @@ def is_implemented_by(cls, instance):
         return implements(instance, cls)
 
 
-def implements(obj: "Interface", interface: Tuple[type, type]) -> bool:
+def implements(obj: "Interface", interface: Type) -> bool:
```
DanielNoord, line 62: I can see one case of `implements` in the code in which this is indeed a `Tuple`. Would this work:
```suggestion
def implements(obj: "Interface", interface: Union[Type["Interface"], Tuple["Interface", ...]]) -> bool:
```

DanielNoord (changes requested):
If we squash these commits we'll need to think about a better name than the current PR title, as it does much more than that right now :)

pylint/interfaces.py
Pierre-Sassoulas: Thanks it's a lot better. I guess the idea is to work like is_instance. Another thing is that obj should be Any, right ? We're trying to tell if it's an interface.

pylint/interfaces.py
```diff
@@ -46,7 +59,9 @@ def is_implemented_by(cls, instance):
         return implements(instance, cls)
 
 
-def implements(obj: "Interface", interface: Tuple[type, type]) -> bool:
+def implements(
+    obj: "Interface", interface: Union[Type["Interface"], Tuple[Type["Interface"], ...]]
```
DanielNoord, line 63:
```suggestion
    obj: BaseChecker, interface: Union[Type["Interface"], Tuple[Type["Interface"], ...]]
```

@Pierre-Sassoulas Since the previous comment was merged I can't suggest on it.

I think it should be BaseChecker actually. We use this to see if checkers implement interfaces. However, since the checkers don't actually subclass `Interface` this will start throwing errors when we add typing to checkers that call `implements`.

The use of `__implements__` seems to be quite outdated anyway, see this rejected [pep](https://www.python.org/dev/peps/pep-0245/)

Pierre-Sassoulas: I added that typing change because it felt like a quick typing fix, but it seems it's becoming a bigger refactor now. It's going to be a problem later on though, so I guess we could merge as is.

When the pip has functions in CamelCase you know it's old 😄. Would you suggest to change the implementation of interfaces later ? Maybe we could we use duck typing and ``has_attr`` or similar instead ? Do you want me to open an issue ?

DanielNoord: Yeah I think with my latest suggestion we have fixed the issue here! PRs like this (small change + small typing update) are actually really useful in our typing efforts.

I think opening an issue doesn't make a lot of sense right now: it will just clutter the issue list and we probably won't have any actual useful discussion going on. Let's add this to our mental todo list 😄

DanielNoord (approved):

PR #5521, opened 2021-12-13
Fix #5451: Produce a score of 0 for fatal errors

### LINKED ISSUE #5451: wrong exit code when linting a mix of existing and non-existing modules
### Bug description

Pylint ignores the fatal error that one module is not found, if another one is.

### Configuration

_No response_

### Command used

```shell
poetry run pylint myfile.py moduel_not_exists.py && echo $?
```

### Pylint output

```shell
************* Module foo.py
foo.py:1:0: F0001: No module named foo.py (fatal)

0
```

### Expected behavior

The exit code should be 1, as an fatal error occured.

### Pylint version

```shell
pylint 2.12.1
astroid 2.9.0
Python 3.9.8 (main, Nov  7 2021, 15:47:09) 
[GCC 11.2.0]
```

### OS / Environment

Kali (XFCE) with latest VS Code

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

**Before**
Modules without statements (such as typo'd paths) don't generate scores. When a linter is run over both valid and invalid paths, this can lead to a perfect score. Scores meeting the ``--fail-under`` threshold exit with a 0 code (passing), which is counterintuitive for the case just described.

**Now**
<s>As long as ``--fail-under`` is 10.0, this PR fails runs with perfect scores if they had any non-zero exit codes.

Of course, the other approach would be to start emitting non-perfect scores for modules without statements, but this seemed less invasive.</s>

Update: new approach is to simply produce a score of 0 for fatal errors.

Closes #5451

### DIFF
--- a/examples/pylintrc
+++ b/examples/pylintrc
@@ -90,7 +90,7 @@ enable=c-extension-no-member
 [REPORTS]
 
 # Python expression which should return a score less than or equal to 10. You
-# have access to the variables 'fatal', 'error', 'warning', 'refactor', and 'convention'
+# have access to the variables 'fatal', 'error', 'warning', 'refactor', 'convention' and 'info'
 # which contain the number of messages in each category, as well as 'statement'
 # which is the total number of statements analyzed. This score is used by the
 # global evaluation report (RP0004).
--- a/examples/pylintrc
+++ b/examples/pylintrc
@@ -90,11 +90,11 @@ enable=c-extension-no-member
 [REPORTS]
 
 # Python expression which should return a score less than or equal to 10. You
-# have access to the variables 'error', 'warning', 'refactor', and 'convention'
+# have access to the variables 'fatal', 'error', 'warning', 'refactor', and 'convention'
 # which contain the number of messages in each category, as well as 'statement'
 # which is the total number of statements analyzed. This score is used by the
 # global evaluation report (RP0004).
-evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / statement) * 10)
+evaluation=0 if fatal else 10.0 - ((float(5 * error + warning + refactor + convention) / statement) * 10)
 
 # Template used to display messages. This is a python new-style format string
 # used to format the message information. See doc for all details.
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -1326,6 +1326,8 @@ class PyLinter(
         except Exception as ex:  # pylint: disable=broad-except
             msg = f"An exception occurred while rating: {ex}"
         else:
+            if self.stats.by_msg.get('fatal', 0):
+                note = 0.0
             self.stats.global_note = note
             msg = f"Your code has been rated at {note:.2f}/10"
             if previous_stats:
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -1326,7 +1326,7 @@ class PyLinter(
         except Exception as ex:  # pylint: disable=broad-except
             msg = f"An exception occurred while rating: {ex}"
         else:
-            if self.stats.by_msg.get('fatal', 0):
+            if self.stats.by_msg.get("fatal", 0):
                 note = 0.0
             self.stats.global_note = note
             msg = f"Your code has been rated at {note:.2f}/10"
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -1315,6 +1315,7 @@ class PyLinter(
         evaluation = self.config.evaluation
         try:
             stats_dict = {
+                "fatal": self.stats.fatal,
                 "error": self.stats.error,
                 "warning": self.stats.warning,
                 "refactor": self.stats.refactor,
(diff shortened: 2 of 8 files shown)


Pierre-Sassoulas:
>  the other approach would be to start emitting non-perfect scores for modules without statements, but this seemed less invasive.

I think this makes more sense. We won't have to add a special case and the corresponding tests this way. The score should be 10 if there isn't any error or 0 if there is at least an error.

jacobtylerwalls:
Sure, happy to go in the other direction.

> the user has taken extra care to setup pylint how they want it to behave (ie., change from the default 10 value).

Just to be clear, the case was when the user has not changed from the default 10. We were emitting 10 anyway.

pylint/lint/pylinter.py
```diff
@@ -1326,6 +1326,8 @@ def _report_evaluation(self):
         except Exception as ex:  # pylint: disable=broad-except
             msg = f"An exception occurred while rating: {ex}"
         else:
+            if self.stats.by_msg.get("fatal", 0):
+                note = 0.0
```
Pierre-Sassoulas, line 1330: I think the code to change is in ``Pylinter.generate_reports()``, we could add typing and change the type from ``Optional[float]`` to ``float``.

jacobtylerwalls: Sorry, I'm having trouble following the suggestion. Can you add more details? I may have my own solution in mind too specifically to see the suggestion.

Pierre-Sassoulas: I think the returned score should be modified at the source where it's calculated, not modified later on. Seems like it's here that we sometime return None: https://github.com/PyCQA/pylint/blob/main/pylint/lint/pylinter.py#L1311

`` note = eval(evaluation, {}, stats_dict)  `` will raise a ``DivideByZero`` error if we do not handle the special case with 0 statement.

jacobtylerwalls: When running my regression test, `self.stats.statement` is 50, not 0, so that's why I'm not sure how to get closer to the "source". If I can get closer, I will! ;)

Pierre-Sassoulas: Sorry I thought the score was not calculated properly when there is 0 statement. What is the issue when there is a fatal and > 0 statements ?

jacobtylerwalls: I see better now. We would have to change the `evaluation` expression. Fatal is not even taken into account in the score, only error:

https://github.com/PyCQA/pylint/blob/35254c29eb3a0f1c7847cfcb3451501a4180373d/pylintrc#L105

Pierre-Sassoulas (approved):
I was wondering where the actual code change was, then I realized that the changed default value is valid python code. That's pretty clever !

examples/pylintrc
```diff
@@ -90,11 +90,11 @@ enable=c-extension-no-member
 [REPORTS]
 
 # Python expression which should return a score less than or equal to 10. You
-# have access to the variables 'error', 'warning', 'refactor', and 'convention'
+# have access to the variables 'fatal', 'error', 'warning', 'refactor', and 'convention'
```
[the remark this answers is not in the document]
DanielNoord: This my own mistake. I added `info` in a previous PR (https://github.com/PyCQA/pylint/pull/5318) and didn't add this here.

pylintrc
```diff
@@ -98,11 +98,11 @@ files-output=no
 reports=no
 
 # Python expression which should return a note less than 10 (10 is the highest
-# note). You have access to the variables errors warning, statement which
-# respectively contain the number of errors / warnings messages and the total
-# number of statements analyzed. This is used by the global evaluation report
-# (RP0004).
-evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / statement) * 10)
+# note). You have access to the variables 'fatal', 'error', 'warning', 'refactor',
+# and 'convention', which contain the number of messages in each category, as
```
DanielNoord, line 102:
```suggestion
# note). You have access to the variables 'fatal', 'error', 'warning', 'refactor', 'convention'
# and 'info', which contain the number of messages in each category, as
```

tests/test_self.py
```diff
@@ -1154,6 +1154,14 @@ def test_fail_on_exit_code(self, args, expected):
         # and errors that are generated they don't affect the exit code.
         self._runtest([path, "--fail-under=-10"] + args, code=expected)
 
+    def test_one_module_fatal_error(self):
+        """
+        Fatal errors in one of several modules linted still exits non-zero.
+        """
+        valid_path = join(HERE, "conftest.py")
+        invalid_path = join(HERE, "garbagePath.py")
+        self._runtest([valid_path, invalid_path], code=1)
```
DanielNoord, line 1163: Isn't expected code `0` here? I haven't looked closely at `_runtest` yet so I might be missing something.

DanielNoord (changes requested):
Small changes, mostly because I forgot to do stuff in https://github.com/PyCQA/pylint/pull/5318 😅 

Note to self: copy `pylintrc` to `examples/pylintrc` in a follow-up PR.

tests/test_self.py
jacobtylerwalls: 0 would indicate success. A fatal error is raised linting `garbagePath.py`, so now (with this PR) we exit with 1.

DanielNoord: My bad.. Sorry! Mixed up 0 and 1. Thanks!

DanielNoord (approved):
Thanks for the quick work @jacobtylerwalls!

I'll try and think of the `pylintrc` move tomorrow.

PR #5523, opened 2021-12-13
Add additional test cases used-before-assignment with try-except

cdce8p:
## Description
Followup to #5500 and #5506

Add additional tests for `used-before-assignment` cases with `try-except`.
`used-before-assignment` should **not** be emitted for variables after the `try-except` block if
* All ExceptHandlers define it
* OR all ExceptHandlers, that don't `raise` an Exception or `return`, define it.

/CC: @jacobtylerwalls I created these while trying to isolate which case you mentioned https://github.com/PyCQA/pylint/issues/5500#issuecomment-991695275 isn't yet covered.


Pierre-Sassoulas (approved):
👍

PR #5525, opened 2021-12-14
Update typing of reporter attributes in ``PyLinter``

DanielNoord:

## Type of Changes

## Description

Small change to these attributes and methods. `_reporter_names` was only used by `_load_reporters` so it makes more sense to just pass that option value to the method instead of creating an additional attribute.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -5,7 +5,6 @@ import collections
 import contextlib
 import functools
 import operator
-import optparse  # pylint: disable=deprecated-module
 import os
 import sys
 import tokenize
@@ -693,10 +692,9 @@ class PyLinter(
                     meth(value)
                 return  # no need to call set_option, disable/enable methods do it
         elif optname == "output-format":
-            if not isinstance(value, str):
-                raise optparse.OptionValueError(
-                    "'output-format' should be a comma separated string of reporters"
-                )
+            assert isinstance(
+                value, str
+            ), "'output-format' should be a comma separated string of reporters"
             self._load_reporters(value)
         try:
             checkers.BaseTokenChecker.set_option(self, optname, value, action, optdict)
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -5,13 +5,14 @@ import collections
 import contextlib
 import functools
 import operator
+import optparse  # pylint: disable=deprecated-module
 import os
 import sys
 import tokenize
 import traceback
 import warnings
 from io import TextIOWrapper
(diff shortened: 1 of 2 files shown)


pylint/lint/pylinter.py
```diff
@@ -619,19 +620,21 @@ def load_plugin_configuration(self):
             except ModuleNotFoundError as e:
                 self.add_message("bad-plugin-value", args=(modname, e), line=0)
 
-    def _load_reporters(self) -> None:
+    def _load_reporters(self, reporter_names: str) -> None:
```
Pierre-Sassoulas, line 623: 👌

pylint/reporters/base_reporter.py
```diff
@@ -23,6 +23,9 @@ class BaseReporter:
 
     extension = ""
 
+    name = "base"
```
Pierre-Sassoulas, line 26:
```suggestion
    name: str = "base"
```
Or is that implicit because we assign a string to it ?

Pierre-Sassoulas (approved):
LGTM !

pylint/lint/pylinter.py
```diff
@@ -690,18 +693,18 @@ def set_option(self, optname, value, action=None, optdict=None):
...
                 return  # no need to call set_option, disable/enable methods do it
         elif optname == "output-format":
-            self._reporter_names = value
-            # If the reporters are already available, load
-            # the reporter class.
-            if self._reporters:
-                self._load_reporters()
-
+            if not isinstance(value, str):
+                raise optparse.OptionValueError(
```
Pierre-Sassoulas, line 697: This line is not tested.

DanielNoord: I'm not sure how to handle this:
Whenever we encounter this option it should always be `str`. Only if you manipulate it by setting the option directly and supplying a `list` (for example) this would raise the warning. In normal use cases that shouldn't be possible. However, `mypy` will complain if we don't assert that it is a `str`...
I could add a test but I foresee that we will then need to start adding tests for every option + incorrect value type. That also seems counter-productive..

pylint/reporters/base_reporter.py
DanielNoord: Yes, it is! 😄

pylint/lint/pylinter.py
Pierre-Sassoulas: Should we use ``assert isinstance(value, str)`` instead then ?

DanielNoord: 👍

DanielNoord: Added!

Pierre-Sassoulas (approved):

PR #5532, opened 2021-12-15
Fix ``used-before-assignment`` for conditional self-referential typing

### LINKED ISSUE #5499: Regression `used-before-assignment` with type annotation in conditional
### Bug description

```py
import random

class A:
    def f(self) -> None:
        if random.random() < 0.5:
            _x: A = self  # false-positive: used-before-assignment

    def g(self) -> None:
        _x: A = self  # no warning
```

This is the same issue reported and fixed in #5342 but it still occurs if the assignment is inside a conditional.

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
test.py:6:16: E0601: Using variable 'A' before assignment (used-before-assignment)
```

### Expected behavior

No error

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.9.6 (default, Aug 16 2021, 15:13:47) 
[GCC 9.3.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #5499.

I should have thought of this fix in #5342. This is much better..

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1805,7 +1805,7 @@ class VariablesChecker(BaseChecker):
         """
         if (
             node.frame().parent == defstmt
-            and node.statement(future=True) not in node.frame().body
+            and node.statement(future=True) == node.frame()
         ):
             # Check if used as type annotation
             # Break but don't emit message if postponed evaluation is enabled
--- a/ChangeLog
+++ b/ChangeLog
@@ -21,6 +21,11 @@ Release date: TBA
 
   Closes #3793
 
+* Fixed false positive for ``used-before-assignment`` with self-referential type
+  annotation in conditional statements within class methods.
+
+  Closes #5499
+
 * ``used-before-assignment`` now assumes that assignments in except blocks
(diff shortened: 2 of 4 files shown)


pylint/checkers/variables.py
```diff
@@ -1805,7 +1805,7 @@ def _is_first_level_self_reference(
         """
         if (
             node.frame().parent == defstmt
-            and node.statement(future=True) not in node.frame().body
+            and node.statement(future=True) == node.frame()
```
Pierre-Sassoulas, line 1808: I don't understand how this work, could you explain ?

DanielNoord: Basically for arguments in function definitions their statement is also their frame (the function definition). For nodes within the body or even within sub-bodies (conditionals) the statement will be different from the frame (the latter still being the function definition).

Pierre-Sassoulas: That makes sense, thank you.

Pierre-Sassoulas (approved):
👍

Issue #1730, opened 2017-11-07
not-callable reported when calling a callable returned by non-trivial property


jendagroovy:
### Steps to reproduce

```python
class ClsA(object):
    pass


class ClsB(object):

    def __init__(self):
        self._some_cls = None

    @property
    def some_cls(self):
        if self._some_cls:
            return self._some_cls

        self._some_cls = ClsA
        return ClsA

    def get_object(self):
        return self.some_cls()
```

### Current behavior

```
pylint -E /tmp/x.py
************* Module x
E: 19,15: self.some_cls is not callable (not-callable)
```
### Expected behavior

```
pylint -E /tmp/x.py
```
No output as the property actually returns a callable. Error should only be reported when there is a certainty, that the property can't return a callable.

### pylint --version output
```
pylint 1.7.4,
astroid 1.5.3
Python 2.7.10 (default, Feb  7 2017, 00:08:15)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)]
```

acatton:
Here's a real life example, with type annotations:

```python
import bz2
import enum
import gzip
import typing


class Algo(enum.Enum):
    BZIP = 'bzip'
    GZIP = 'gzip'

    @property
    def compress(self) -> typing.Callable[[bytes], bytes]:
        return {
            self.BZIP: bz2.compress,
            self.GZIP: gzip.compress,
        }[self]


def compress(algo: str, data: bytes) -> bytes:
    return Algo(algo).compress(data)
```

```
$ pylint x.py
No config file found, using default configuration
************* Module x
[…]
E: 20,11: Algo(algo).compress is not callable (not-callable)
[…]
```

heyakyra:
Just came across this as well, and managed to find one SO question on it: https://stackoverflow.com/questions/53598374/pylint-giving-not-callable-error-for-object-property-that-is-callable

Also, is it expected that even without using `@property` instance variables are not callable? 

Could this be related to #1493 ?

PCManticore:
@acatton I can no longer reproduce the bug you were having using the latest not released yet version.

@heyakyra Do you have a reproduction case for which you get this error? The reason this wasn't inferrable for the original poster was the fact that `pylint` does not proper support for control flow understanding, which is why it had an issue with `get_object` and `self.some_class` but other than that `pylint` should be able to infer a callable coming from a property. An example would be great.

heyakyra:
[...]

PR #5537, opened 2021-12-16
Fix ``not-callable`` for attributes that alias ``NamedTuple``

DanielNoord:

## Type of Changes

## Description

Closes part of #1730, specifically this comment: https://github.com/PyCQA/pylint/issues/1730#issuecomment-489572303

I tried to destructure the `if` a little bit because I got a `too-many-boolean-expressions` warning and it became really unreadable.. This did lead to some code duplication though..

### DIFF
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1559,25 +1559,28 @@ accessed. Python regular expressions are accepted.",
         Only functions, generators and objects defining __call__ are "callable"
         We ignore instances of descriptors since astroid cannot properly handle them yet
         """
-        if inferred_call and not inferred_call.callable():
-            if isinstance(inferred_call, astroid.Instance):
-                # Don't emit if we can't make sure this object is callable.
-                if not has_known_bases(inferred_call):
-                    return
-
-                if inferred_call.parent and isinstance(
-                    inferred_call.scope(), nodes.ClassDef
-                ):
-                    # Ignore descriptor instances
-                    if "__get__" in inferred_call.locals:
-                        return
-                    # NamedTuple instances are callable
-                    if inferred_call.qname() == "typing.NamedTuple":
-                        return
-
-            self.add_message("not-callable", node=node, args=node.func.as_string())
-        else:
+        # Handle uninferrable calls
+        if not inferred_call or inferred_call.callable():
             self._check_uninferable_call(node)
+            return
+
+        # Ignore non instances
+        if not isinstance(inferred_call, astroid.Instance):
+            return
+
+        # Don't emit if we can't make sure this object is callable.
+        if not has_known_bases(inferred_call):
+            return
+
+        if inferred_call.parent and isinstance(inferred_call.scope(), nodes.ClassDef):
+            # Ignore descriptor instances
+            if "__get__" in inferred_call.locals:
+                return
+            # NamedTuple instances are callable
+            if inferred_call.qname() == "typing.NamedTuple":
+                return
+
+        self.add_message("not-callable", node=node, args=node.func.as_string())
 
     @check_messages("invalid-sequence-index")
     def visit_extslice(self, node: nodes.ExtSlice) -> None:
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1564,6 +1564,10 @@ accessed. Python regular expressions are accepted.",
             self._check_uninferable_call(node)
             return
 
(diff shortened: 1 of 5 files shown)


pylint/checkers/typecheck.py
```diff
@@ -1294,16 +1294,23 @@ def visit_call(self, node: nodes.Call) -> None:
         # Ignore instances of descriptors since astroid cannot properly handle them
         # yet
         if called and not called.callable():
```
Pierre-Sassoulas, line 1296: Maybe we can create a function from line 1296 to line 1318 and use premature return to simplify this expression ?

```python
 if not isinstance(called, astroid.Instance):
    self._check_uninferable_call(node)
    return
if not has_known_bases(called):
    return
....        
```

Pierre-Sassoulas (commented):
I think this would benefit from a refactor even if that's a little outside of the scope of the current MR. Let me know what you think :)

DanielNoord:
Makes sense! I'll get to it tomorrow 😄

pylint/checkers/typecheck.py
```diff
@@ -1567,6 +1551,34 @@ def _check_invalid_sequence_index(self, subscript: nodes.Subscript):
...
+                    # Ignore descriptor instances
+                    if "__get__" in inferred_call.locals:
+                        return
+                    # NamedTuple instances are callable
+                    if inferred_call.qname() == "typing.NamedTuple":
+                        return
+
+            self.add_message("not-callable", node=node, args=node.func.as_string())
+        else:
+            self._check_uninferable_call(node)
```
Pierre-Sassoulas, line 1580:
```suggestion
if not inferred_call or inferred_call.callable():
    return self._check_uninferable_call(node)
if not isinstance(inferred_call, astroid.Instance):
    return
# Don't emit if we can't make sure this object is callable.
if not has_known_bases(inferred_call):
    return
if inferred_call.parent and isinstance(
        inferred_call.scope(), nodes.ClassDef
):
    # Ignore descriptor instances
    if "__get__" in inferred_call.locals:
        return
    # NamedTuple instances are callable
    if inferred_call.qname() == "typing.NamedTuple":
        return
self.add_message("not-callable", node=node, args=node.func.as_string())
```
One step further :)

DanielNoord: Slightly adapted this suggestion and added some extra white lines + comments in latest commit.

pylint/checkers/typecheck.py
```diff
@@ -1559,25 +1559,28 @@ def _check_not_callable(
...
-
-            self.add_message("not-callable", node=node, args=node.func.as_string())
-        else:
+        # Handle uninferrable calls
+        if not inferred_call or inferred_call.callable():
             self._check_uninferable_call(node)
+            return
+
+        # Ignore non instances
+        if not isinstance(inferred_call, astroid.Instance):
```
Pierre-Sassoulas, line 1568: Sorry I was wrong in my first proposition I edited it. I think there's a guaranteed message for non instances. (The add_message is in the same scope)

DanielNoord: This doesn't fail any tests.. The coveralls comments seems broken, but are we actually testing this?

DanielNoord: Relevant commit:
https://github.com/PyCQA/pylint/commit/ea14cad3187aa2efdd0ae7b520b2ba44797c0863

Seems like we can remove it? All `inferred_call`'s that we test are `Instance`'s

Pierre-Sassoulas: I think you're right, but it's really hard to reconstruct code example that would trigger a particular piece of code though. (Or "prove" that it's impossible). I hope primer++ will fix this.

Pierre-Sassoulas (approved):
I approve with our without the instance test, you decide :smile:

DanielNoord:
I've removed it for now. `called` is the inference of `nodes.Call.func`, seems like that should always be an `Instance` (considering `nodes.Const` subclasses `Instance` as well). The original commit also doesn't provide any good reasons to think otherwise, I feel the `if` was added simply because the issue referred to instances originally.

DanielNoord:
Thank our test suite: this change was not trivial. Reverting and adding the message back in, it should actually be called see:
https://github.com/PyCQA/pylint/blob/566a377154ce68e795aa3f952bb70bd987045228/tests/functional/d/deprecated/deprecated_methods_py38.py#L30

PR #5539, opened 2021-12-16
Make ``arguments-differ`` check extra parameters for default values

### LINKED ISSUE #1556: Allow adding parameters in an overridden method, without tripping `arguments-differ`
The `arguments-differ` check will complain about a method which *loses* arguments that the superclass's method will allow. Such code violates the [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle), so IMO should trigger `arguments-differ`.

What is not desirable, IMO, is to have the same check complain about:

```python
class Ipsum:
    def dolor(self, sit):
        pass


class LoremIpsum(Ipsum):
    def dolor(self, sit, amet=None):
        handle(amet)
        super().dolor(sit)

```

That code does not violate the LSP; an instance of `LoremIpsum` will function just fine as an instance of `Ipsum`. Its methods can be called in the same way as methods of `Ipsum`.

How can I have PyLint not complain about code like this which AFAICT does not violate LSP, while still correctly flagging LSP violations?

### LINKED ISSUE #5338: False positive `arguments-differ` with defaults
### Bug description

```python
class Base:
    def method(self, arg: str):
        print(arg)


class Derived(Base):
    def method(self, arg: str, param1: int = 42, *, param2: int = 42):
        print(arg, param1, param2)


xxx: Base = Derived()
xxx.method('blah')
```

### Pylint output

```shell
 W0221: Number of parameters was 2 in 'Base.method' and is now 4 in overridden 'Derived.method' method (arguments-differ)
```

### Expected behavior

no arguments-differ error in the case.

### Pylint version

```shell
pylint 2.9.6
astroid 2.6.6
Python 3.9.7 (default, Aug 30 2021, 00:00:00) 
[GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
```

### OS / Environment

Fedora 34

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #1556, Closes #5338

Whenever a new value has a default argument the child class can be replaced by the base class and thus this does not violate LSP.


pylint/checkers/classes.py
```diff
@@ -240,10 +240,17 @@ def _has_different_parameters(
...
             return ["Number of parameters "]
 
+        if not original_param:
+            try:
+                overridden_param.parent.default_value(overridden_param.name)
+                continue
+            except astroid.NoDefault:
+                return ["Number of parameters "]
+
+        params = (original_param, overridden_param)
```
Pierre-Sassoulas, line 253:
```suggestion
```
To save an assignment ?

pylint/checkers/classes.py
```diff
@@ -240,10 +240,17 @@ def _has_different_parameters(
...
         # check for the arguments' name
         names = [param.name for param in params]
```
Pierre-Sassoulas, line 255:
```suggestion
        names = [param.name for param in (original_param, overridden_param)]
```

Pierre-Sassoulas (approved):
LGTM, I have a small suggestion.

Pierre-Sassoulas:
In #1556 the idea was to add a new message instead. What do you think ? It would prevent us to have to handle the reverse false positive "I have an argument with a default I should get an argument-differ". We did that already for ``argument-renamed``.

pylint/checkers/classes.py
DanielNoord: Good catch!

DanielNoord:
I'm not sure.. I updated the `arguments-differ` description to reflect the change. I think we could add a more restrictive message but that the default message should reflect what is proposed in this PR.
I don't think this actually that much of an anti-pattern and `pylint` shouldn't complain about it as a standard. In fact,  by allowing to resolve this message by adding default arguments instead of completely disabling the message I think we actually point users to start writing better code.

Pierre-Sassoulas:
> In fact, by allowing to resolve this message by adding default arguments instead of completely disabling the message I think we actually point users to start writing better code.

Fair point !

cdce8p:
Please don't merge it yet. I've been thinking about the `arguments-differ` error a bit and would like to take a look first.

doc/whatsnew/2.13.rst
```diff
@@ -86,6 +86,11 @@ Other Changes
 * The ``testutils`` for unittests now accept ``end_lineno`` and ``end_column``. Tests
   without these will trigger a ``DeprecationWarning``.
 
+* ``arguments-differ`` will no longer complain about method redefintions with extra parameters
```
jacobtylerwalls, line 89:
```suggestion
* ``arguments-differ`` will no longer complain about method redefinitions with extra parameters
```

ChangeLog
```diff
@@ -92,6 +92,11 @@ Release date: TBA
 * The ``testutils`` for unittests now accept ``end_lineno`` and ``end_column``. Tests
   without these will trigger a ``DeprecationWarning``.
 
+* ``arguments-differ`` will no longer complain about method redefintions with extra parameters
```
jacobtylerwalls, line 95:
```suggestion
* ``arguments-differ`` will no longer complain about method redefinitions with extra parameters
```

Pierre-Sassoulas:
@cdce8p do you think you'll have time to review this or should we merge ? I suppose you had other thing on your mind since mid december :D

Pierre-Sassoulas:
I'm going to move that PR and associated issues to 2.14 if it's the last one remaining for 2.13.0 :)

DanielNoord:
I hope we can get this in. The change itself is pretty minor, although I agree a complete review of `arguments-differ` might be necessary in the future.

Pierre-Sassoulas:
What do you think about merging this in 2.13.0 and refactoring later @cdce8p ?

jacobtylerwalls (changes requested):
Would you mind updating the message description? I think this is a substantial enough change to warrant it.

pylint/checkers/classes/class_checker.py
```diff
@@ -529,7 +535,8 @@ def _has_same_layout_slots(slots, assigned_value):
         "%s %s %r method",
         "arguments-differ",
         "Used when a method has a different number of arguments than in "
-        "the implemented interface or in an overridden method.",
+        "the implemented interface or in an overridden method. Extra arguments "
+        "with default values are ignored.",
```
jacobtylerwalls, line 539: 🤦🏻‍♂️

jacobtylerwalls:
@DanielNoord what do you think about this edge case, is it in scope?

I would expect this to continue to raise `arguments-differ` because `arg2` was added with no default, but now it raises `arguments-renamed`:

```python
class A:
    def z(self, *, arg1): ...

class B(A):
    def z(self, *, arg2, arg1=None): ...
```
**main**
```
Number of parameters was 2 in 'A.z' and is now 3 in overridden 'B.z' method (arguments-differ)
```
***PR***
```
Parameter 'arg1' has been renamed to 'arg2' in overridden 'B.z' method (arguments-renamed)
```

DanielNoord:
That's probably in scope (although without your review we would have likely considered it a new false positive).

I'll have a look!

DanielNoord:
@jacobtylerwalls Fixed!

The message could certainly be improved but I'd like to wait on that for @cdce8p's look at the general message and a possible refactor. This should be enough for now.

tests/functional/a/arguments_renamed.py
```diff
@@ -40,15 +40,15 @@ class Child(Parent):
     def test(self, arg1): # [arguments-renamed]
         return arg1 + 1
 
-    def kwargs_test(self, arg, *, value1, var2): #[arguments-renamed]
```
DanielNoord, line 43: What do we want to do here? We go from:
`def kwargs_test(self, arg, *, var1, var2):`
to
`def kwargs_test(self, arg, *, value1, var2):`

Is the `arguments-renamed` or `arguments-differ`?

tests/functional/a/arguments_renamed.py
```diff
@@ -40,15 +40,15 @@ class Child(Parent):
...
+    def kwargs_test(self, arg, *, value1, var2): #[arguments-differ]
         print(f"keyword parameters are {value1} and {var2}.")
 
 class Child2(Parent):
 
     def test(self, var): # [arguments-renamed]
         return var + 1
 
-    def kwargs_test(self, *, var1, kw2): #[arguments-renamed, arguments-differ]
+    def kwargs_test(self, *, var1, kw2): #[arguments-differ]
```
DanielNoord, line 51: What do we want to do here? We go from:
`def kwargs_test(self, arg, *, var1, var2):`
to
`def kwargs_test(self, *, var1, kw2):`

Is the `arguments-renamed` or `arguments-differ`?

jacobtylerwalls: I think arguments-differ is an improvement in this case. For kw-only args, an argument cannot truly be "renamed".

tests/functional/a/arguments_renamed.py
jacobtylerwalls: Same as below, this is an improvement (kwonly args can't be "renamed")

tests/functional/a/arguments_renamed.txt
```diff
@@ -2,10 +2,9 @@ arguments-renamed:17:4:17:12:Orange.brew:Parameter 'fruit_name' has been renamed
 arguments-renamed:20:4:20:26:Orange.eat_with_condiment:Parameter 'fruit_name' has been renamed to 'orange_name' in overridden 'Orange.eat_with_condiment' method:UNDEFINED
 arguments-differ:27:4:27:26:Banana.eat_with_condiment:Number of parameters was 3 in 'Fruit.eat_with_condiment' and is now 4 in overridden 'Banana.eat_with_condiment' method:UNDEFINED
 arguments-renamed:40:4:40:12:Child.test:Parameter 'arg' has been renamed to 'arg1' in overridden 'Child.test' method:UNDEFINED
-arguments-renamed:43:4:43:19:Child.kwargs_test:Parameter 'var1' has been renamed to 'value1' in overridden 'Child.kwargs_test' method:UNDEFINED
+arguments-differ:43:4:43:19:Child.kwargs_test:Number of parameters was 4 in 'Parent.kwargs_test' and is now 4 in overridden 'Child.kwargs_test' method:UNDEFINED
```
jacobtylerwalls, line 5: For other reviewers, we have an issue to track this already: #5793

tests/functional/a/arguments_differ.py
```diff
@@ -326,3 +326,34 @@ class Foo2(AbstractFoo):
...
+
+
+class DerivedClassWithoutAnnotation(BaseClass):
+    def method(self, arg, param1=42, *, param2=42):
+        print(arg, param1, param2)
+
+
+class AClass:
+    def method(self, *, arg1):  # [no-self-use]
+        ...
```
jacobtylerwalls, line 354: Is it the usual practice to mix in unrelated messages instead of using correct style or suppressing? A `print(self)` is not _that_ much uglier than `...`

pylint/checkers/classes/class_checker.py
```diff
@@ -258,6 +264,30 @@ def _has_different_parameters(
...
+    if any(name not in overridden_names for name in original_names):
+        return ["Number of parameters "]
+
+    for name in overridden_names:
+        if name in original_names:
+            continue
+
+        try:
+            overridden[0].parent.default_value(name)
+            continue
```
jacobtylerwalls, line 284: This is straight-up bikeshedding, but what's your view on removing the `continue` here and in the similar place above?

pylint/checkers/classes/class_checker.py
```diff
@@ -258,6 +264,30 @@ def _has_different_parameters(
     return result
 
 
+def _has_different_keyword_parameters(
```
jacobtylerwalls, line 267: To match the docstring?
```suggestion
def _has_different_keyword_only_parameters(
```

jacobtylerwalls (approved):

jacobtylerwalls:
> Whenever a new value has a default argument the base class can be substituted for the child class and thus this does not violate LSP.

oh btw, isn't it talked of as subbing child for base?

DanielNoord:
> > Whenever a new value has a default argument the base class can be substituted for the child class and thus this does not violate LSP.
> 
> oh btw, isn't it talked of as subbing child for base?

Yeah! Changed!

pylint/checkers/classes/class_checker.py
DanielNoord: I removed this one, the other leads to a crash. We need the `if not original_param:` check and without the `continue` it becomes meaningless.

jacobtylerwalls:
> Edit: Wait, that's what's there right? Base becomes child?

I think the child is substituted for the base, and because the child's extra arguments have defaults, calls written according to the base signature without knowledge of the child's signature still work. So child stands in for base, not base stands in for child (which would be the sense of "base class substituted for the child class")

tests/functional/a/arguments_differ.py
Pierre-Sassoulas: I think generally it's a balance between functional tests that are unitary and having a lot of code construct and example without too much duplication. For no-self-use I don't think we need more tests for it. Also we're probably going to remove it from the default configuration in #3512 so I think using print(self) will still work then but we'd get a useless-suppression if we disable or a "expected a message" from the test with the current code.

DanielNoord: Yeah, I was just lazy here. Added a print.

DanielNoord:
> > Edit: Wait, that's what's there right? Base becomes child?
> 
> I think the child is substituted for the base, and because the child's extra arguments have defaults, calls written according to the base signature without knowledge of the child's signature still work. So child stands in for base, not base stands in for child (which would be the sense of "base class substituted for the child class")

Ah yeah, my Dutch and English is missing with me here. In Dutch this would be the order to say this, but in English it is the other way around. I'll change.

PR #5542, opened 2021-12-17
Add tests for NewType false positives

### LINKED ISSUE #2296: False positive not-an-iterable
### Steps to reproduce
1. Put this code in test.py:
```python
from typing import List, NewType

a = NewType("a", List[int])

def fun() -> a:
    """Some function."""
    data = [1, 2, 3]

    return a(data)

def fun1():
    """some function."""
    for x in fun():
        pass
```
2. Run  `pylint test.py`

### Current behavior
`test.py:14:13: E1133: Non-iterable value fun() is used in an iterating context (not-an-iterable)`

### pylint --version output
pylint 2.0.0
astroid 2.0.0.dev4
Python 3.6.5 (default, May  3 2018, 10:08:28) 
[GCC 5.4.0 20160609]

### LINKED ISSUE #3162: no-member false positive from NewType
This is a weird interaction between the typing library and pylint, so I'm not sure if this is the right place to file the bug, but here goes.

### Steps to reproduce
```python 
class A:
    def __init__(self, value: int):
        self.value = value

a = A(5)
a.value  # no no-member error

B = NewType("B", A)

b = B(a)
b.value  # Instance of 'B' has no 'value' memberpylint(no-member)
```

### Current behavior
Pylint flags a no-member error when accessing members of objects with a NewType class.

### Expected behavior
Pylint flags no errors.

### pylint --version output
pylint 2.4.2
astroid 2.3.1
Python 3.7.4 (default, Sep 20 2019, 11:46:08) 
[GCC 8.3.0]

colatkinson:
## Type of Changes

## Description

Adds tests for false positives related to `typing.NewType`. Specifically, in the contexts of member checking and iterability checking.

Note that the fix itself is at https://github.com/PyCQA/astroid/pull/1301 -- these won't pass until that's merged.

Closes #2296 
Closes #3162


DanielNoord:
Since the PR in `astroid` is clsoed we can also close this :)

Issue #3167, opened 2019-10-06
False no-member positive on defaultless instance variables on subclasses


scop:
t.py:
```python
class A:
    myfield: int

class B(A):
    pass

a = A()
print(a.myfield)

b = B()
print(b.myfield)
```
```
# pylint 2.4.2, astroid 2.3.1
$ pylint t.py | grep no-member
t.py:11:6: E1101: Instance of 'B' has no 'myfield' member (no-member)
```
Pylint doesn't appear to "see" the subclass `myfield` member declared in the parent class (reports the error against `B`), but does see it in the parent class (doesn't report against `A`).

(Technically I suppose the member isn't actually there in the parent class either as it's an instance variable without a default [1], and I believe they kind of spring to life only when actually set to a value which doesn't happen in this example -- for this reason the example code actually raises when run. Pylint seems to "know" that it's (or will be) there nevertheless which I think is great, but it should also know that it's there in the subclass.
[1]  https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations)

PCManticore:
Yeah, I think `pylint` should accept subclasses as well and maybe metaclasses. The reason we don't emit these is that variable annotations can be used to specify attributes that might exist at runtime somehow.

nattyg93:
@PCManticore Is there any movement on resolving the scenario in the original issue? We are running into this issue on a daily basis. 🙂

frenck:
This issue causes us to either use improper typing and assign a None default and constantly make sure that it is asserted to satisfy type checking, or ignore about every line to either satisfy type checking or pylint.

It is rather sad, as it really degrades the whole goal of these tools in the first place. It has been a frustrating experience. Really hope that this can be picked up by someone 😢

cdce8p:
I opened #4194 to fix this issue

PR #5540, opened 2021-12-17
Fix typos

kianmeng:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :scroll: Docs          |

PR #4194, opened 2021-03-05
Fix false-positive no-member for typed annotations without default value

cdce8p:
## Steps

- [ ] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [ ] If it's a new feature or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Description
Fix false-positive `no-member` for typed annotations without default value.

```py
class A:
    myfield: int

class B(A):
    pass

a = A()
print(a.myfield)

b = B()
print(b.myfield)  # No longer emits `no-member` warning 
```

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |

## Related Issue
Closes #3167

Issue #2945, opened 2019-06-03
Better support for PEP 526 instance and classvar annotations


belm0:
### Steps to reproduce
A class which doesn't have `__init__` (it may set attributes in `__new__`, for example) will yield no-member when instance attributes are accessed.

I tried to resolve it with PEP 526 instance var annotations which don't have assignment (i.e. type only), but still get no-member error.

```
class Foo:
    bar:  int
    ...

x = Foo()
x.bar
```

### Current behavior
```
E1101: Instance of 'Foo' has no 'bar' member (no-member)
```

### Expected behavior
no warnings

### pylint --version output
```
pylint 2.3.1
astroid 2.3.0
Python 3.7.0b3 (default, Aug 25 2018, 16:41:39)
[Clang 9.0.0 (clang-900.0.39.2)]
```

PCManticore:
The check is correct, that annotation is not accessible as an attribute in Python. You'll most likely get an error as the following if you attempt to run the code:
```
Traceback (most recent call last):
  File "a.py", line 6, in <module>
    x.bar
AttributeError: 'Foo' object has no attribute 'bar'
```

belm0:
@PCManticore please let me clarify with a complete example:

```
class Foo(BaseException):
    bar:  int

    def __new__(cls):
        o = BaseException.__new__(cls)
        o.bar = 10
        return o

x = Foo()
print(x.bar)
```
```
E1101: Instance of 'Foo' has no 'bar' member (no-member)
```

From my understanding of PEP 526, one of the points is to enable typing in complex initialization scenarios.

I'll note that if Foo simply derives from `object`, then pylint does not emit the error, even if the annotation is removed.  So it seems this has something to do with type derivation.

PCManticore:
Ah, thanks, I definitely misunderstood the purpose of the PEP.

PR #5544, opened 2021-12-17
Fix #1555: Fix false negatives for `no-member` from self-referencing assignments

### LINKED ISSUE #1555: Missed detection of undefined variable inside class
### Steps to reproduce
Run pylint on the following:
```
class SomeClass():
    '''This class does nothing'''
    def __init__(self):
        self.val = self.val
```

### Current behavior
No errors are thrown

### Expected behavior
It should warn that self.val is undefined. 
Note that it works correctly inside functions and at module level. 

### pylint --version output
pylint 1.7.1, 
astroid 1.5.3
Python 3.6.1 (default, Mar 27 2017, 00:27:06) 
[GCC 6.3.1 20170306]

jacobtylerwalls:
## Type of Changes

## Description

Emits `no-member` when attempting to assign an instance attribute to itself before it has been defined, since at the moment of attempted assignment, there is no such member.

Closes #1555

### DIFF
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1007,15 +1007,7 @@ accessed. Python regular expressions are accepted.",
                 return
 
             try:
-                if not [
-                    n
-                    for n in owner.getattr(node.attrname)
-                    if not isinstance(n.statement(future=True), nodes.AugAssign)
-                ]:
-                    missingattr.add((owner, name))
-                    continue
-            except astroid.exceptions.StatementMissing:
-                continue
+                attr_nodes = owner.getattr(node.attrname)
             except AttributeError:
                 continue
             except astroid.DuplicateBasesError:
@@ -1039,6 +1031,24 @@ accessed. Python regular expressions are accepted.",
                     continue
                 missingattr.add((owner, name))
                 continue
+            else:
+                for attr_node in attr_nodes:
+                    attr_parent = attr_node.parent
+                    # Skip augmented assignments
+                    try:
+                        if isinstance(
+                            attr_node.statement(future=True), nodes.AugAssign
+                        ):
+                            continue
+                    except astroid.exceptions.StatementMissing:
+                        break
+                    # Skip self-referencing assignments
+                    if attr_parent is node.parent:
+                        continue
+                    break
+                else:
+                    missingattr.add((owner, name))
+                    continue
             # stop on the first found
             break
         else:
--- a/tests/functional/n/no/no_member_assign_same_line.py
+++ b/tests/functional/n/no/no_member_assign_same_line.py
@@ -1,0 +1,37 @@
+"""Tests for no-member for self-referencing instance attributes
+See https://github.com/PyCQA/pylint/issues/1555
+"""
+# pylint: disable=too-few-public-methods
+
+
+class ClassWithMember:
+    """Member defined in superclass."""
+    def __init__(self):
+        self.member = True
+
+
+class AssignMemberInSameLine:
+    """This class attempts to assign and access a member in the same line."""
+    def __init__(self):
+        self.member = self.member  # [no-member]
+
+
+class AssignMemberInSameLineAfterTypeAnnotation:
+    """This might emit a message like `maybe-no-member` in the future."""
+    def __init__(self):
+        self.member: bool
+        self.member = self.member
+
+
+class AssignMemberFromSuper1(ClassWithMember):
+    """This assignment is valid due to inheritance."""
+    def __init__(self):
+        self.member = self.member
+        super().__init__()
+
+
+class AssignMemberFromSuper2(ClassWithMember):
+    """This assignment is valid due to inheritance."""
+    def __init__(self):
+        super().__init__()
+        self.member = self.member
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,7 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
+* Fix false negative for ``no-member`` when attempting to assign an instance
+  attribute to itself without any prior assignment.
 
+  Closes #1555
 
 ..
   Insert your changelog randomly, it will reduce merge conflicts
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -23,3 +23,8 @@ Extensions
 
 Other Changes
 =============
+
+* Fix false negative for ``no-member`` when attempting to assign an instance
+  attribute to itself without any prior assignment.
+
+  Closes #1555
--- a/tests/functional/n/no/no_member_assign_same_line.txt
+++ b/tests/functional/n/no/no_member_assign_same_line.txt
@@ -1,0 +1,1 @@
+no-member:16:22:16:33:AssignMemberInSameLine.__init__:Instance of 'AssignMemberInSameLine' has no 'member' member:INFERENCE


Pierre-Sassoulas (approved):
LGTM, thank you :)

tests/functional/n/no/no_member_assign_same_line.py
```diff
@@ -0,0 +1,21 @@
...
+    """This class attempts to assign and access a member in the same line."""
+    def __init__(self):
+        self.member = self.member  # [no-member]
+
+
+class AssignMemberFromSuper(ClassWithMember):
+    """This assignment is valid due to inheritance."""
+    def __init__(self):
+        super().__init__()
+        self.member = self.member
```
DanielNoord, line 21: I would like some additional test cases just to be sure we are not emitting false postives/negatives.

What about:
```python
class AssignMemberFromSuper(ClassWithMember):
     """This assignment is valid due to inheritance."""
     def __init__(self):
         self.member = self.member
         super().__init__()
```
and
```python
class MyClass:
     """This assignment is valid due to inheritance."""
     def __init__(self):
         self.member: bool
         self.member = self.member
```

There might be more variations, but this is just what I came up with quickly.

I also wonder what happens when the assignment in the base class is conditional.
Something like:
```python
class ClassWithMember:
     """Member defined in superclass."""
     def __init__(self):
         if False:
             self.member = True

class AssignMemberFromSuper(ClassWithMember):
     """This assignment is valid due to inheritance."""
     def __init__(self):
         super().__init__()
         self.member = self.member
```
That would require control flow, but would we emit the message with the current code? Or do we assume the assignment succeeded?
If we still emit we might want to open an issue about it to track future progress.

DanielNoord (changes requested):

tests/functional/n/no/no_member_assign_same_line.py
jacobtylerwalls: Good cases.

- variation in placement of call to super ✅ 
- type annotation ❌ (false negative)
- control flow with constant false values ❌ (false negative)

I've updated the PR.

DanielNoord: > * variation in placement of call to super ✅

👍  Sign of a well written checker 😄 

> * type annotation ❌ (false negative)

Do you think we can fix this in this PR? Seems to fit the general idea of the original issue and would prevent future issues being opened.

> * control flow with constant false values ❌ (false negative)

I think the standard is not to include tests that we do not currently pass in the test suite, but rather to open new issues for them. I've gotten this comment on some of my PRs previously. Perhaps it's best to remove these from the tests and create a new well defined issue for them. Already writing down the tests often helps with eventual implementation.

jacobtylerwalls: > I think the standard is not to include tests that we do not currently pass in the test suite

Makes sense, since we don't have a way to decorate with expectedFailure given how things are set up.

> Do you think we can fix this in this PR?

I'll have a look. 🔦  Don't know much about how type annotations are tracked just yet.

pylint/checkers/typecheck.py
```diff
@@ -1030,7 +1031,21 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                        # Skip self-referencing assignments
+                        and n.parent is not node.parent
+                        # Skip assignments that are only type annotations
+                        and not (
+                            isinstance(n.parent, nodes.AnnAssign)
+                            and n.parent.value is None
+                        )
+                        # But allow dataclasses
+                        or brain_dataclasses.is_decorated_with_dataclass(
+                            n.parent.parent
```
jacobtylerwalls, line 1046: I can remove the comments if they seem self-evident. We're nested inside a negated list comprehension, so I erred on the side of more help.

pylint/reporters/text.py
```diff
@@ -190,6 +190,7 @@ def __init__(self, output: Optional[TextIO] = None) -> None:
 
     def on_set_current_module(self, module: str, filepath: Optional[str]) -> None:
         """Set the format template to be used and check for unrecognized arguments."""
+        # pylint: disable-next=no-member
```
jacobtylerwalls, line 193: `linter` is set by `PyLinter.set_reporter()`

pylint/checkers/typecheck.py
DanielNoord: Do you think it makes sense to create a for loop here? We're calling `n.parent` quite a lot and the readability of this `if` within the comprehension is becoming questionable 😄 

I like the comments personally, makes it easier to understand what's going on.

pylint/reporters/text.py
DanielNoord:
```suggestion
        # pylint: disable-next=no-member # linter is set by PyLinter.set_reporter()
```

tests/functional/n/no/no_member_assign_same_line.py
```diff
@@ -0,0 +1,35 @@
+"""https://github.com/PyCQA/pylint/issues/1555"""
```
DanielNoord, line 1:
```suggestion
"""Tests for no-member for self-referencing instance attributes
See https://github.com/PyCQA/pylint/issues/1555
"""
```

pylint/reporters/text.py
Pierre-Sassoulas:
```suggestion
        # linter is set by PyLinter.set_reporter()
        # pylint: disable-next=no-member
```

pylint/checkers/typecheck.py
```diff
@@ -74,6 +74,7 @@
 import astroid
 import astroid.exceptions
 from astroid import bases, nodes
+from astroid.brain import brain_dataclasses
```
DanielNoord, line 77: I'm wondering if we should add this to a `utils` module in `astroid`. Porting the util to `pylint` doesn't make sense as we would need to duplicate the globals that are being used in the function, but importing from `brain` also feels weird.

@Pierre-Sassoulas Do you have an opinion?

DanielNoord (changes requested):

Pierre-Sassoulas (commented):
LGTM, but there's something strange with the coverage report : ``86 unchanged lines in 4 files lost coverage.``, do you know why ?

pylint/checkers/typecheck.py
Pierre-Sassoulas: Importing the brain do feel weird, we're accessing astroid internals.

DanielNoord:
@jacobtylerwalls Thanks for your work again! Looks like you found some additional cases as well. Should we add a test for dataclasses? If we change the comprehension to a for loop that if statement would become uncovered as well (I think).

@Pierre-Sassoulas `coveralls` has been having problems recently. I noticed these comments as well, even on trivial changes that touched almost no code. The comment might not be correct.

pylint/reporters/text.py
Pierre-Sassoulas: I made the suggestion independentely @DanielNoord, I have nothing against yours 😄

pylint/checkers/typecheck.py
Pierre-Sassoulas: Maybe this should be moved in pylint, like we did for ``is_typing_guard``, but I did not check if it's used in astroid itself.

DanielNoord: It's used in `astroid` and the function would be impacted by https://github.com/PyCQA/astroid/pull/1298 for example.

We need to decide where we want to store our "list of dataclass-like decorators". I think `astroid` makes the most sense (also because it is already there). However, I think a refactor into a `utils` module which is explicitly public API makes sense.

Pierre-Sassoulas: Quoting my review in https://github.com/PyCQA/astroid/pull/1298#pullrequestreview-834665567

> There might be a problem with our design if we need to add the name of all dataclasses modules to this constant. I don't have a better solution as it's astroid and not pylint we can't add an option to let user add their own data-classes.

I don't know about the feasibility of moving this to pylint though.

Maybe we should keep it in astroid but provide an argument to be able to add user defined dataclasses (for example from a pylint option) ? @cdce8p what do you think ?

jacobtylerwalls:
> Should we add a test for dataclasses? 

That line resulted from fixing failing tests in `no_member_dataclasses.py`, so I think we're good on test coverage there.

> even on trivial changes that touched almost no code.

See for example #5540.

pylint/checkers/typecheck.py
```diff
@@ -1027,11 +1028,26 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                            isinstance(n_parent, nodes.AnnAssign)
+                            and n_parent.value is None
+                        )
+                        # But allow dataclasses
+                        or brain_dataclasses.is_decorated_with_dataclass(
+                            n_parent.parent
+                        )
+                    )
+                    if found_attribute:
+                        break
```
DanielNoord, line 1049:
```suggestion
                    # Skip augmented assignments
                    if isinstance(n.statement(future=True), nodes.AugAssign):
                        break
                    # Skip self-referencing assignments
                    if not n_parent is not node.parent:
                        break
                    # Skip assignments that are only type annotations
                    if isinstance(n_parent, nodes.AnnAssign) and n_parent.value is None:
                        break
                    # Allow dataclasses
                    if brain_dataclasses.is_decorated_with_dataclass(n_parent.parent):
                        break
```

Would this work? Refactoring for loops on Github is hard so I might have misplaced a not or if, but I think this should work right? Personally I think using premature break is easier on the eyes, but opinions might differ obviously.

jacobtylerwalls: Ah I don't think so, because `found_attribute` is three conditions, and they all need to pass, then we have to mix `continue` and `break`, like this:

```python
            try:
                for n in owner.getattr(node.attrname):
                    n_parent = n.parent
                    # Skip augmented assignments
                    if isinstance(n.statement(future=True), nodes.AugAssign):
                        continue
                    # Skip self-referencing assignments
                    if n_parent is node.parent:
                        continue
                    # Skip assignments that are only type annotations and not dataclasses
                    if (
                        isinstance(n_parent, nodes.AnnAssign)
                        and n_parent.value is None
                        and not brain_dataclasses.is_decorated_with_dataclass(
                            n_parent.parent
                        )
                    ):
                        continue
                    break
                else:
                    missingattr.add((owner, name))
                    continue
```

I'm 50/50 on whether that's any better, since we end up with different nesting levels of `continue`s.


***

However, thank you for having me look at this again, I've noticed missing parentheses at the end to make what's now 4 conditions into what was meant to be the 3 conditions.

DanielNoord: Personally I think this is better and we can even improve on it some more!

For example, the `StatementMissing` catcher can be move to the `statement()` line. Perhaps that's a bit beyond the scope of this PR but I think using this broken down version allows for more refactoring.

jacobtylerwalls: I'll push with what I quoted above, and let's look into moving the excepts closer in a follow-up PR, because it's possible there could be a behavior change -- we could simply continue searching nodes when handling an exception rather than bail out entirely, as before.

pylint/checkers/typecheck.py
```diff
@@ -1027,11 +1028,25 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
                 return
 
             try:
-                if not [
-                    n
-                    for n in owner.getattr(node.attrname)
-                    if not isinstance(n.statement(future=True), nodes.AugAssign)
-                ]:
+                for n in owner.getattr(node.attrname):
```
DanielNoord, line 1031:
```suggestion
                for attr_node in owner.getattr(node.attrname):
```

pylint/checkers/typecheck.py
```diff
@@ -1027,11 +1028,25 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                    n_parent = n.parent
```
DanielNoord, line 1032:
```suggestion
                    attr_parent = attr_node.parent
```

pylint/checkers/typecheck.py
```diff
@@ -1027,11 +1028,25 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                    # Skip augmented assignments
+                    if isinstance(n.statement(future=True), nodes.AugAssign):
```
DanielNoord, line 1034:
```suggestion
                    if isinstance(attr_node.statement(future=True), nodes.AugAssign):
```

pylint/checkers/typecheck.py
```diff
@@ -1027,11 +1028,25 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                        continue
+                    # Skip self-referencing assignments
+                    if n_parent is node.parent:
```
DanielNoord, line 1037:
```suggestion
                    if attr_parent is node.parent:
```

pylint/checkers/typecheck.py
```diff
@@ -1027,11 +1028,25 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                        continue
+                    # Skip assignments that are only type annotations and not dataclasses
+                    if (
+                        isinstance(n_parent, nodes.AnnAssign)
+                        and n_parent.value is None
+                        and not brain_dataclasses.is_decorated_with_dataclass(
+                            n_parent.parent
```
DanielNoord, line 1044:
```suggestion
                        isinstance(attr_parent, nodes.AnnAssign)
                        and attr_parent.value is None
                        and not brain_dataclasses.is_decorated_with_dataclass(
                            attr_parent.parent
```

DanielNoord (changes requested):
@jacobtylerwalls Do you want to do the exception handler move in this PR? Or do it in a later PR?

We should also wait for a resolution on the `astroid` `utils` location discussion.

pylint/checkers/typecheck.py
cdce8p: I haven't had time to look at it in detail yet, but I'm not sure it's even necessary here. The goal of inference tips in astroid, as we have for `dataclasses` is to "hide" the magic and convert it to a format we normally expect.

If you need to check that it's not a decorator in a dataclass, probably the astroid brain would need to be improved. In this case I would think however, it's just not necessary. @jacobtylerwalls Could you add a test case for it?

jacobtylerwalls: Hi, just checking I'm understanding -- you're asking for a quick demo that the line is needed? When I comment out the check I get these seven failures:

```
====================================================== short test summary info ======================================================
FAILED tests/test_functional.py::test_functional[unpacking_non_sequence_py37] - AssertionError: Wrong results for file "unpacking_...
FAILED tests/test_functional.py::test_functional[unspecified_encoding_py38] - AssertionError: Wrong results for file "unspecified_...
FAILED tests/test_functional.py::test_functional[no_member_dataclasses] - AssertionError: Wrong results for file "no_member_datacl...
FAILED tests/test_functional.py::test_functional[no_member_subclassed_dataclasses] - AssertionError: Wrong results for file "no_me...
FAILED tests/test_functional.py::test_functional[too_many_instance_attributes_py37] - AssertionError: Wrong results for file "too_...
FAILED tests/test_functional.py::test_functional[too_few_public_methods_37] - AssertionError: Wrong results for file "too_few_publ...
FAILED tests/test_functional.py::test_functional[dataclass_typecheck] - AssertionError: Wrong results for file "dataclass_typecheck":
============================================ 7 failed, 635 passed, 21 skipped in 33.58s =============================================
```

Which makes sense, because dataclasses define members with what looks like unused type annotations. For instance, several dozen here:

```
_______________________________________________ test_functional[dataclass_typecheck] ________________________________________________

self = <pylint.testutils.lint_module_test.LintModuleTest object at 0x7f9d0ba10f70>

    def runTest(self) -> None:
>       self._runTest()
E       AssertionError: Wrong results for file "dataclass_typecheck":
E       
E         30: no-member
E         31: no-member
E         32: no-member
E         34: no-member
E         35: no-member
E         36: no-member
E         38: no-member
E         39: no-member
E         40: no-member
E         42: no-member
E         43: no-member
...
```

DanielNoord:
I'm removing this from `2.13` as it is blocked by discussion in `astroid`. Namely: https://github.com/PyCQA/astroid/pull/1306 and subsequently whether we wan't to create a special `Instance` for `dataclass`. If we did we wouldn't need the `astroid` util as we can just do `isinstance(node, bases.Dataclass)`.
Also adding `requires Astroid update`

DanielNoord:
Possible fix for the problems identified here in https://github.com/PyCQA/astroid/pull/1391

DanielNoord:
@jacobtylerwalls `dev` of `astroid` now has `is_dataclass` on all `ClassDef` nodes. You could change the code in this PR to use it so that when we release `2.10` it's just a matter of upgrading the dependency and rebasing.

jacobtylerwalls:
Great, thank you for doing the astroid PR!

jacobtylerwalls:
@DanielNoord I made the changes, expecting CI to fail, and was greeted by a very unwanted ✅ .

**What happened?**
This PR makes no changes to the silencing of `AttributeError`, which is why this didn't fail by depending on the unreleased astroid changes. We discussed cleaning up those excepts [here](https://github.com/PyCQA/pylint/pull/5544#discussion_r771851873). At this point I'm inclined to clean them up in this PR, since I am learning about the edge cases after all.

So with that cleaned up, this would crash again before updating astroid (good), but I was still expecting more substantive failures. At first, it looked like I didn't need the newfangled `is_dataclass` attribute, because checking `isinstance(attr_parent.parent, nodes.ClassDef)` was enough. But _why_?  Well, with non-dataclasses, these blocks are short-circuited[1], and jump to `_emit_no_member()`, which silences the warning with this explanation:

https://github.com/PyCQA/pylint/blob/b31dd5623966069386b9df3ccc08e2834aca621d/pylint/checkers/typecheck.py#L515-L518

This goes against what you and I discussed [here](https://github.com/PyCQA/pylint/pull/5544#discussion_r771456956), so it's worth a double-check. 

**Question:** Do you think we should move forward with emitting `no-member` for possibly never-assigned type annotations? (Currently `undefined-variable` is _not_ raised by the changes in #5158, because they are in different scopes.) Or reduce the scope of this PR back to just dealing with self-referencing assignments?
```python
class PossibleNoMember:
    def __init__(self):
        self.member: bool

    def set_member(member):
        self.member = member

    def depend_on_member():
        print(self.member)  # should this raise no-member? or just rely on undefined-variable?
```

[1] -- These patterns are very intricate -- so after cleaning up the excepts we may be able to write a test case that will require using `is_dataclass` after all.

jacobtylerwalls:
I have a feeling the answer is going to be "create `possibly-no-member`".

jacobtylerwalls:
Well, from the blame there's a good argument to leave these alone, see discussion at #2945 and #3167.

DanielNoord:
#4194 is interesting as well as it fixes #3167, but there is no discussion about this potentially being undefined.

I'm not sure. I think we actually use this pattern ourself:
We create a `PyLinter()` with some attributes typed but not assigned.
We call some methods (but not the `__init__`) that set the unassigned attributes.
We call some methods that use the previously unassigned attributes.

Personally I feel that that is worth a message, although as you said perhaps it should be `possibly-no-member`. This pattern is very error-prone and only because I started to type the `PyLinter` class these potential issues arose. Warning users about this, and having them think about not assigning attributes in the `__init__` is a good thing imo. However, this is probably used a little too often to suddenly start emitting `no-member` warnings everywhere...
Perhaps it would indeed be best to re-rescope this PR back to the original intent and create a new issue in which we discuss adding `possibly-no-member`? We could make `2.14` the `possible-...` update 😄

pylint/checkers/typecheck.py
```diff
@@ -1055,15 +1055,9 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
-                ]:
-                    missingattr.add((owner, name))
-                    continue
-            except astroid.exceptions.StatementMissing:
-                continue
+                attr_nodes = owner.getattr(node.attrname)
+            # pylint: disable-next=fixme
+            # TODO: remove except AttributeError when astroid updated to 2.10
             except AttributeError:
                 continue
```
jacobtylerwalls, line 1062: We will no longer need to catch `AttributeError` thanks to @areveny fixing it in https://github.com/PyCQA/astroid/pull/1366.

Test case:
```python
from astroid import extract_node
code = """
def f():
    pass
a = property(f)
a.fget
"""
node = extract_node(code)
prop = list(node.expr.infer())[0]
prop.getattr('fget')
```

pylint/checkers/typecheck.py
```diff
@@ -1055,15 +1055,7 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
-                    n
-                    for n in owner.getattr(node.attrname)
-                    if not isinstance(n.statement(future=True), nodes.AugAssign)
-                ]:
-                    missingattr.add((owner, name))
-                    continue
-            except astroid.exceptions.StatementMissing:
-                continue
+                attr_nodes = owner.getattr(node.attrname)
             except AttributeError:
```
DanielNoord, line 1059: I believe this is already uncovered due to the upgrade. Can you remove it?

jacobtylerwalls: I tried, but had to revert:
```python
AttributeError: 'Lambda' object has no attribute 'getattr'
```

I was planning to have a look this morning at the cause in `astroid`.

DanielNoord: Strange how it is not covered then. I'll await your investigation 😄

DanielNoord (approved):

pylint/checkers/typecheck.py
jacobtylerwalls: Do the primer tests count toward coverage? That would explain it, if not.

DanielNoord: Ah no, they don't!

Pierre-Sassoulas (changes requested):
Need a rebase and moving the changelog to 2.14

Pierre-Sassoulas (approved):

PR #5547, opened 2021-12-17
Check if decorator returns use keyword (``unexpected-keyword-arg``)

### LINKED ISSUE #258: False positive ``unexpected-keyword-arg`` (E1123) for decorators
Originally reported by: **the mulhern (BitBucket: [the_mulhern](http://bitbucket.org/the_mulhern))**

---

For the following example,  error 
"Passing unexpected keyword argument 'foo' in function call (unexpected-keyword-arg)"
is reported, correctly.

```
#!python

def junk1(junk=None):
    print("%s" % junk)

def main():
    junk1(foo=2)
```

For the example below, no such error is reported, due to the use of keyword args parameter. This is also correct.

```
#!python

def junk1(junk=None, *args, **kwargs):
    print("%s" % junk)

def main():
    junk1(foo=2)
```

In the following example, the decorator supplies the foo argument, but the analysis still reports
"Passing unexpected keyword argument 'foo' in function call (unexpected-keyword-arg)"
which is incorrect. It should be able to detect that foo is now supplied, and give no error.

```
#!python

def decorator(f):

    def new_func(foo=3):
        f(junk=foo)

    return new_func

@decorator
def junk1(junk=None):
    print("%s" % junk)

def main():
    junk1(foo=2)
```

And for the following example it still reports 
"Passing unexpected keyword argument 'foo' in function call (unexpected-keyword-arg)"
although in this case it should give up due to the use of keyword arguments param, as
in the second example.

```
#!python

def decorator(f):

    def new_func(**kwargs):
        f(junk=kwargs["foo"])

    return new_func

@decorator
def junk1(junk=None):
    print("%s" % junk)

def main():
    junk1(foo=2)
```

---
- Bitbucket: https://bitbucket.org/logilab/pylint/issue/258

DanielNoord:

## Type of Changes

## Description

Closes #258 (🎉)

I'd like to see the coverage report on this. I might need to add some additional tests.


DanielNoord:
Blocked by #5549. Rest of the tests and coverage look good so this is ready for review.

mbyrnepr2 (approved):

tests/functional/u/unexpected_keyword_arg.py
```diff
@@ -0,0 +1,118 @@
+"""Tests for unexpected-keyword-arg"""
```
Pierre-Sassoulas, line 1: The tests are on point !

pylint/checkers/typecheck.py
```diff
@@ -1496,6 +1500,46 @@ def visit_call(self, node: nodes.Call) -> None:
...
+        """
+        if not func.decorators:
+            return False
+
+        for decorator in func.decorators.nodes:
+            inferred = safe_infer(decorator)
+
+            # If we can't infer the decorator we assume it satisfies consumes
+            # the keyword, so we don't raise false positives
+            if not inferred:
```
Pierre-Sassoulas, line 1518: Not really related to this MR but if we have to handle confidence level, this kind of return where we do not exactly know if inference was used or not in the calling code will be a headache later. Makes me thing that implementing confidence properly will be an enormous work, maybe not even worth it.

Pierre-Sassoulas (approved):
It's heart-warming to see really old issues getting addressed 🎉

pylint/checkers/typecheck.py
DanielNoord: Thankfully we have `UNDEFINED` 😄

martimlobao:
@DanielNoord maybe I should open a new ticket, but just tried this out with the latest version of the `main` branch and there seems to be a case where this is still raising a false positive `unexpected-keyword-arg` error. Here's a MWE:

```python
def example_decorator(*new_func_args, **new_func_kwargs):
    def decorator(func):
        def wrapper(*args, **kwargs):
            print_me = kwargs.pop("print_me", None)
            def inner(*inner_args, **inner_kwargs):
                print(print_me)
                return func(*inner_args, **inner_kwargs)

            return inner(*args, **kwargs)

        return wrapper

    if len(new_func_args) == 1 and len(new_func_kwargs) == 0 and callable(new_func_args[0]):
        return decorator(new_func_args[0])
    return decorator

@example_decorator
def example_func(num):
    return num ** 2

example_func(6, print_me="hello")
```

Using pylint on commit hash e75e37a2908ef1c003ea5816c1e234523bdd38be, this is the error that's printed out:

```
E1123: Unexpected keyword argument 'print_me' in function call (unexpected-keyword-arg)
```

Calling `example_func` works fine:

```
>>> example_task(6, print_me="hello")
hello
36
```

DanielNoord:
Ah yes, this won't be catched. Not sure how easy it would be to implement looking for inner return values.
In any case, I think this warrants its own issue. Could you open it?

martimlobao:
> Ah yes, this won't be catched. Not sure how easy it would be to implement looking for inner return values. In any case, I think this warrants its own issue. Could you open it?

Sorry for taking forever, but finally opened the ticket :)

DanielNoord:
Thanks @martimlobao! Totally forgot about this myself.

martimlobao:
> Thanks @martimlobao! Totally forgot about this myself.

lol sorry for the ridiculously incomplete title, i started writing it and then forgot to finish 🤦‍♂️

PR #5553, opened 2021-12-19
Update default evaluation formula to match that in default pylintrc

jacobtylerwalls:
## Type of Changes

## Description
Missed in #5521 -- when updating the default score evaluation formula in example pylintrc, should have updated the default value when there is no pylintrc. Missed this when performing find and replace due to the line break.


Pierre-Sassoulas (approved):
Thanks !

PR #5554, opened 2021-12-19
Ignore files with name that starts like an emacs lock files

### LINKED ISSUE #367: tries to open emacs locks
Originally reported by: **BitBucket: [sandrotosi](http://bitbucket.org/sandrotosi), GitHub: @sandrotosi?**

---

Hello,
this is bug http://www.logilab.org/19517 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=563000 which is still happening:

```
When a file is modified but not saved in emacs, some sort of
dangling-symlink lockfile is created:

 .#HexBoard.py -> username@hostname.domain.tld.9285:1262073654

This confuses pylint:

************* Module Omaha.Boards..#HexBoard
F0002:  1: <class 'logilab.astng._exceptions.ASTNGBuildingException'>: Unable to load module Omaha.Boards..#HexBoard (No module named #HexBoard)


I suppose it would be enough to just ignore dotfiles, I guess they
would make invalid python module names anyway.  Similarly, the #
character is most probaby forbidden here as well - all in all, pylint
should probably ignore such files as impossible to properly use from
python, and possibly issue a warning about their presence.
```

```
> > Is this still happening with the latest version of pylint in sid?

The problem still happens today:

************* Module Omaha.Games.abstract..#PlayerPools
F:  1, 0: error while code parsing: Unable to load file '/work/yann/games/omaha2/Omaha/Games/abstract/.#PlayerPools.py' ([Errno 2] No such file or directory: '/work/yann/games/omaha2/Omaha/Games/abstract/.#PlayerPools.py') (parse-error)
```

---
- Bitbucket: https://bitbucket.org/logilab/pylint/issue/367

Pierre-Sassoulas:
## Type of Changes

## Description

This ignore the file starting like emacs file lock. It's a problem in emacs not in pylint but this kind of file name can almost exclusively come from emacs as it's not a valid module name so we may as well ease the life of our emacs users.

Closes #367

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,12 @@ Release date: TBA
 
   Closes #5499
 
+* pylint do not take into account files starting with ``.#`` anymore. Those are
+  considered `emacs file lock`. See
+  https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html.
+
+  Closes #367
+
 * ``used-before-assignment`` now assumes that assignments in except blocks
   may not have occurred and warns accordingly.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,8 +30,8 @@ Release date: TBA
 
   Closes #5499
 
-* By default, pylint do not take into account files starting with ``.#`` anymore. Those are
-  considered `emacs file lock`. See
+* By default, pylint does no longer take files starting with ``.#`` into account. Those are
+  considered `emacs file locks`. See
   https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html.
   This behavior can be reverted by redefining the ``ignore-patterns`` option.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,9 +30,10 @@ Release date: TBA
 
   Closes #5499
 
-* pylint do not take into account files starting with ``.#`` anymore. Those are
+* By default, pylint do not take into account files starting with ``.#`` anymore. Those are
   considered `emacs file lock`. See
   https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html.
+  This behavior can be reverted by redefining the ``ignore-patterns`` option.
 
   Closes #367
 
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -32,6 +32,13 @@ Extensions
 Other Changes
 =============
 
+* pylint do not take into account files starting with ``.#`` anymore. Those are
+  considered `emacs file lock`_.
+
+  Closes #367
+
+.. _`emacs file lock`: https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html
+
 * Fixed extremely long processing of long lines with comma's.
 
   Closes #5483
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -32,13 +32,13 @@ Extensions
 Other Changes
 =============
 
-* By default, pylint do not take into account files starting with ``.#`` anymore. Those are
-  considered `emacs file lock`_. This behavior can be reverted by redefining the
+* By default, pylint does no longer take files starting with ``.#`` into account. Those are
+  considered `emacs file locks`_. This behavior can be reverted by redefining the
   ``ignore-patterns`` option.
 
   Closes #367
 
-.. _`emacs file lock`: https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html
+.. _`emacs file locks`: https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html
 
 * Fixed extremely long processing of long lines with comma's.
 
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -32,8 +32,9 @@ Extensions
 Other Changes
 =============
 
-* pylint do not take into account files starting with ``.#`` anymore. Those are
-  considered `emacs file lock`_.
+* By default, pylint do not take into account files starting with ``.#`` anymore. Those are
+  considered `emacs file lock`_. This behavior can be reverted by redefining the
+  ``ignore-patterns`` option.
 
   Closes #367
 
--- a/pylint/lint/expand_modules.py
+++ b/pylint/lint/expand_modules.py
@@ -52,8 +52,6 @@ def expand_modules(
         basename = os.path.basename(something)
         if (
             basename in ignore_list
-            # Emacs file locks see #367
-            or basename.startswith(".#")
             or _is_in_ignore_list_re(os.path.basename(something), ignore_list_re)
             or _is_in_ignore_list_re(something, ignore_list_paths_re)
         ):
--- a/pylint/lint/expand_modules.py
+++ b/pylint/lint/expand_modules.py
@@ -8,8 +8,8 @@ from pylint.typing import ErrorDescriptionDict, ModuleDescriptionDict
 
 
 def _modpath_from_file(filename, is_namespace, path=None):
-    def _is_package_cb(path, parts):
-        return modutils.check_modpath_has_init(path, parts) or is_namespace
+    def _is_package_cb(inner_path, parts):
+        return modutils.check_modpath_has_init(inner_path, parts) or is_namespace
 
     return modutils.modpath_from_file_with_callback(
         filename, path=path, is_package_cb=_is_package_cb
--- a/pylint/lint/expand_modules.py
+++ b/pylint/lint/expand_modules.py
@@ -48,11 +48,12 @@ def expand_modules(
     result: List[ModuleDescriptionDict] = []
     errors: List[ErrorDescriptionDict] = []
     path = sys.path.copy()
-
     for something in files_or_modules:
         basename = os.path.basename(something)
         if (
             basename in ignore_list
+            # Emacs file locks see #367
+            or basename.startswith(".#")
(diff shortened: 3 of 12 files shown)


ChangeLog
```diff
@@ -30,6 +30,11 @@ Release date: TBA
 
   Closes #5499
 
+* pylint do not take into account files starting with ``.#`` anymore. Those are
+  considered broken emacs symlinks.
```
Pierre-Sassoulas, line 34:
```suggestion
  considered [emacs file lock](https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html).
```

Pierre-Sassoulas:
@DanielNoord I think it's rather safe and reasonable to do. Let me know what you think.

DanielNoord:
Looking at the place of the code: does it make sense to add this to the default value of ignore path? I'm not a big fan of hard-coding an exception.

Pierre-Sassoulas:
Thank you, this is a lot better when handled in the configuration. I added functional tests accordingly.

Pierre-Sassoulas:
It's ready for review again.

tests/functional/e/.#emacs_file_lock.py
```diff
@@ -0,0 +1,4 @@
+# The name is invalid, but we should not analyse this file
+# Because it starts like an emacs file lock ignored by default
```
DanielNoord, line 2:
```suggestion
# Because its filename reseambles an emacs file lock ignored by default
```

tests/functional/e/.#emacs_file_lock.py
```diff
@@ -0,0 +1,4 @@
...
+# https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html
+# See #367
```
DanielNoord, line 4:
```suggestion
# See https://github.com/PyCQA/pylint/issues/367
```

tests/functional/e/.#emacs_file_lock_by_conf.py
```diff
@@ -0,0 +1,2 @@
+# The name is invalid, but we should not analyse this file
+# Because it starts like an emacs file lock ignored in the configuration
```
DanielNoord, line 2:
```suggestion
# Because it resembles an emacs file lock and is ignored by the configuration
# See https://github.com/PyCQA/pylint/issues/367
```

tests/functional/e/.#emacs_file_lock_redefined_conf.py
```diff
@@ -0,0 +1,3 @@
+# [invalid-name]
+# The name is invalid, and we should analyse this file
+# as the ignore-patterns is redefined in the configuration
```
DanielNoord, line 3:
```suggestion
# The name is invalid and we should analyse this file
# as ignore-patterns is redefined in the configuration
```

pylint/testutils/configuration_test.py
```diff
@@ -31,7 +38,7 @@ def get_expected_or_default(
         with open(expected_result_path, encoding="utf8") as f:
             expected = f.read()
         # logging is helpful to realize your file is not taken into
-        # account after a misspell of the file name. The output of the
+        # account after a misspelling of the file name. The output of the
```
DanielNoord, line 41:
```suggestion
        # account after a misspelling of the file name. The output of the
```

😄

pylint/testutils/configuration_test.py
```diff
@@ -18,6 +20,11 @@
 PylintConfiguration = Dict[str, ConfigurationValue]
 
 
+if not PY38_PLUS:
+    # See https://stackoverflow.com/a/56935186
+    copy._deepcopy_dispatch[type(re.compile(""))] = lambda r, _: r  # type: ignore[attr-defined]
```
DanielNoord, line 25: Is this relevant for this change? I don't see `copy` being used here?

pylint/lint/pylinter.py
```diff
@@ -223,7 +223,7 @@ def make_options():
                     "type": "regexp_csv",
                     "metavar": "<pattern>[,<pattern>...]",
                     "dest": "black_list_re",
-                    "default": (),
+                    "default": (r"^\.#",),
                     "help": "Files or directories matching the regex patterns are"
                     " skipped. The regex matches against base names, not paths.",
```
DanielNoord, line 228: Let's add something about the new default value here. Something along the lines of: the default value ignores emacs file locks.

pylint/lint/expand_modules.py
```diff
@@ -48,7 +48,6 @@ def expand_modules(
     result: List[ModuleDescriptionDict] = []
     errors: List[ErrorDescriptionDict] = []
     path = sys.path.copy()
-
```
DanielNoord, line 51: Personally I like these white lines for readability, but I think that is personal preference.

doc/whatsnew/2.13.rst
```diff
@@ -32,6 +32,14 @@ Extensions
 Other Changes
 =============
 
+* By default, pylint do not take into account files starting with ``.#`` anymore. Those are
+  considered `emacs file lock`_. This behavior can be reverted by redefining the
+  ``ignore-patterns`` option.
+
+  Closes #367
+
+.. _`emacs file lock`: https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html
```
DanielNoord, line 41:
```suggestion
* By default, pylint does no longer take files starting with ``.#`` into account. Those are
  considered `emacs file locks`_. This behavior can be reverted by redefining the
  ``ignore-patterns`` option.

  Closes #367

.. _`emacs file locks`: https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Locks.html
```

ChangeLog
```diff
@@ -30,6 +30,13 @@ Release date: TBA
 
   Closes #5499
 
+* By default, pylint do not take into account files starting with ``.#`` anymore. Those are
+  considered `emacs file lock`. See
```
DanielNoord, line 34:
```suggestion
* By default, pylint does no longer take files starting with ``.#`` into account. Those are
  considered `emacs file locks`. See
```

pylint/lint/pylinter.py
```diff
@@ -223,7 +223,7 @@ def make_options():
                     "type": "regexp_csv",
                     "metavar": "<pattern>[,<pattern>...]",
                     "dest": "black_list_re",
-                    "default": (),
+                    "default": (r"^\.#",),
```
DanielNoord, line 226: Should we add this to our own `pylintrc`? The `ignore-paths` option is currently not included.

DanielNoord (changes requested):

pylint/testutils/configuration_test.py
Pierre-Sassoulas: Yes, we're starting to deepcopy a regex pattern in the functional configuration test which does not work in python 3.6 or 3.7.

pylint/lint/pylinter.py
Pierre-Sassoulas: Probably something to do in the upgrade of our pylintrc so it matches the one generated by pylint more closely ?

pylint/lint/pylinter.py
Pierre-Sassoulas:
```suggestion
                    " skipped. The regex matches against base names, not paths. The default value "
                    "ignores emacs file locks",
```

pylint/testutils/configuration_test.py
Pierre-Sassoulas: I don't know if it's the joke, but your suggestion does not change anything ? 😄

pylint/lint/expand_modules.py
```diff
@@ -48,7 +48,6 @@ def expand_modules(
     result: List[ModuleDescriptionDict] = []
     errors: List[ErrorDescriptionDict] = []
     path = sys.path.copy()
```
Pierre-Sassoulas, line 50:
```suggestion
    path = sys.path.copy()

```

pylint/lint/expand_modules.py
Pierre-Sassoulas: Added back :)

pylint/testutils/configuration_test.py
DanielNoord: Oh sorry! I wanted to add a 😄 for the fact that misspell was misspelled

pylint/lint/pylinter.py
DanielNoord: Okay!

pylint/testutils/configuration_test.py
DanielNoord: Could you add a comment as to why this is needed? Because it is 25 lines detached from where we need this this isn't immediately clear.

DanielNoord (changes requested):

pylint/testutils/configuration_test.py
```diff
@@ -18,6 +20,11 @@
 PylintConfiguration = Dict[str, ConfigurationValue]
 
 
+if not PY38_PLUS:
+    # See https://stackoverflow.com/a/56935186
```
Pierre-Sassoulas, line 24:
```suggestion
    # We need to deepcopy a compiled regex pattern
    # In python 3.6 and 3.7 this require a hack
    # See https://stackoverflow.com/a/56935186
```

Pierre-Sassoulas:
> Could you add a comment as to why this is needed?

https://github.com/PyCQA/pylint/pull/5554/commits/f7c49bd2c40724a154a51cf817d11ccc151494d7

DanielNoord (approved):

PR #5556, opened 2021-12-19
Add missing settings to pylintrc

DanielNoord:

## Type of Changes

## Description

There were many settings documented in `examples/pylintrc` that were not used in our own `pylintrc`. Similarly, there were commands in our `pylintrc` that were not in the example. This brings both documents in line with each other. 

As new users may look at our own configuration I think it is good to have all options in here, even if they are commented out. Please also look at the values of the new settings, as they essentially become default/suggested values.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -371,7 +371,7 @@ property-classes=abc.abstractproperty
 # A class is considered mixin if its name matches the mixin-class-rgx option.
 ignore-mixin-members=yes
 
-# Regex pattern to define which classes are considered mixins ignore-mixin-
+# Regex pattern to define which classes are considered mixins if ignore-mixin-
 # members is set to 'yes'
 mixin-class-rgx=.*MixIn
 
--- a/pylintrc
+++ b/pylintrc
@@ -11,6 +11,14 @@
 # paths.
 ignore=CVS
 
+# Add files or directories matching the regex patterns to the ignore-list. The
+# regex matches against paths and can be in Posix or Windows format.
+ignore-paths=
+
+# Files or directories matching the regex patterns are skipped. The regex
+# matches against base names, not paths.
+ignore-patterns=^\.#
+
 # Pickle collected data for later comparisons.
 persistent=yes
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
This seems a nice change to make but it's going to be very hard to review. How did you do that ? Did you use ``--generate-rc-file`` ? Might be a good thing if we checked that consistency in our continuous integration so it does not drift again.

DanielNoord:
I just looked at all settings that were in `examples` and moved them over. Then I copied `pylintrc` to `examples/pylintrc`. The actual number of uncommented new settings is limited (less than 10) as I commented some out.

Pierre-Sassoulas (changes requested):
I just checked and it seems there are inconsistencies between the text in our pylintrc and the generated text from ``pylint --generate-rcfile`` (not only the conf itself which is only natural, the descriptions and help texts too). The generated text should be as good and preferably identical as the one in our own pylintrc.  I think we should minimize the diff we get when we do ``pylint --generate-rcfile>pylintrc``.

DanielNoord:
Good point. Let's do it the other way around then > use --generate-rcfile the create a `pylintrc` and then add missing options to our current file.

I'll get to this somewhere this week

Pierre-Sassoulas (approved):
The diff looks good but the command to generate the rcfile is in a different order, right ? What's the plan ? :)

pylintrc
```diff
@@ -277,16 +371,19 @@ property-classes=abc.abstractproperty
 # A class is considered mixin if its name matches the mixin-class-rgx option.
 ignore-mixin-members=yes
 
-# Regex pattern to define which classes are considered mixins.
+# Regex pattern to define which classes are considered mixins ignore-mixin-
+# members is set to 'yes'
```
Pierre-Sassoulas, line 375:
```suggestion
# Regex pattern to define which classes are considered mixins if ignore-mixin-
# members is set to 'yes'
```
?

DanielNoord:
I didn't want to make the diff too large so I just copied what was missing.. Also I think we're not always using the default settings ourself, so it doesn't make sense to completely copy the generated file. (I don't think we should change default values right now).

Personally I think just adding them for visibility's sake is good enough and we shouldn't worry too much about the order.

Pierre-Sassoulas:
> we shouldn't worry too much about the order.

It's not the order per-se that worries me, it's more the consistencies between our pylintrc and the generated one. For example, we fixed a typo here : https://github.com/PyCQA/pylint/pull/5556#discussion_r772655286. But we fixed it only in our pylintrc and not in the generating code. I'm going to make another MR later to test the generation while still keeping our non default setting. Let's merge this one, it is a first step and has the benefit of being reviewable (and will make next MR more reviewable too). I realized that we have duplicated value when generating too (disallow list / black list which are the same thing are both created)

PR #5558, opened 2021-12-20
Add typing and uniformize the checker registering in Pylinter

Pierre-Sassoulas:
## Type of Changes

## Description

Add typing and uniformize the checker registering in Pylinter. Also fix the typing affected by it in PyLinter and co..

### DIFF
--- a/doc/how_tos/custom_checkers.rst
+++ b/doc/how_tos/custom_checkers.rst
@@ -201,7 +201,11 @@ Add the ``register`` function to the top level of the file.
 
 .. code-block:: python
 
-  def register(linter):
+  def register(linter: "PyLinter") -> None:
+      """This required method auto registers the checker during initialization.
+
+      :param linter: The linter to register the checker to.
+      """
       linter.register_checker(UniqueReturnChecker(linter))
 
 We are now ready to debug and test our checker!
--- a/doc/how_tos/custom_checkers.rst
+++ b/doc/how_tos/custom_checkers.rst
@@ -201,6 +201,14 @@ Add the ``register`` function to the top level of the file.
 
 .. code-block:: python
 
+  from typing import TYPE_CHECKING
+
+  import astroid
+
+  if TYPE_CHECKING:
+      from pylint.lint import PyLinter
+
+
   def register(linter: "PyLinter") -> None:
       """This required method auto registers the checker during initialization.
 
--- a/doc/how_tos/plugins.rst
+++ b/doc/how_tos/plugins.rst
@@ -25,6 +25,14 @@ So a basic hello-world plugin can be implemented as:
 .. sourcecode:: python
 
   # Inside hello_plugin.py
+  from typing import TYPE_CHECKING
+
+  import astroid
+
+  if TYPE_CHECKING:
+      from pylint.lint import PyLinter
+
+
   def register(linter: "PyLinter") -> None:
     """This required method auto registers the checker during initialization.
 
@@ -47,6 +55,14 @@ We can extend hello-world plugin to ignore some specific names using
 .. sourcecode:: python
 
   # Inside hello_plugin.py
+  from typing import TYPE_CHECKING
+
+  import astroid
+
+  if TYPE_CHECKING:
+      from pylint.lint import PyLinter
+
+
   def register(linter: "PyLinter") -> None:
     """This required method auto registers the checker during initialization.
 
--- a/doc/how_tos/plugins.rst
+++ b/doc/how_tos/plugins.rst
@@ -25,7 +25,11 @@ So a basic hello-world plugin can be implemented as:
 .. sourcecode:: python
 
   # Inside hello_plugin.py
-  def register(linter):
+  def register(linter: "PyLinter") -> None:
+    """This required method auto registers the checker during initialization.
+
+    :param linter: The linter to register the checker to.
+    """
     print('Hello world')
 
 
@@ -43,7 +47,11 @@ We can extend hello-world plugin to ignore some specific names using
 .. sourcecode:: python
 
   # Inside hello_plugin.py
-  def register(linter):
+  def register(linter: "PyLinter") -> None:
+    """This required method auto registers the checker during initialization.
+
+    :param linter: The linter to register the checker to.
+    """
     print('Hello world')
 
   def load_configuration(linter):
--- a/doc/how_tos/transform_plugins.rst
+++ b/doc/how_tos/transform_plugins.rst
@@ -66,7 +66,13 @@ Module, Class, Function etc. In our case we need to transform a class. It can be
 
 .. sourcecode:: python
 
+  from typing import TYPE_CHECKING
+
   import astroid
+
+  if TYPE_CHECKING:
+      from pylint.lint import PyLinter
+
 
   def register(linter: "PyLinter") -> None:
     """This required method auto registers the checker during initialization.
--- a/doc/how_tos/transform_plugins.rst
+++ b/doc/how_tos/transform_plugins.rst
@@ -68,8 +68,11 @@ Module, Class, Function etc. In our case we need to transform a class. It can be
 
   import astroid
 
-  def register(linter):
-    # Needed for registering the plugin.
+  def register(linter: "PyLinter") -> None:
+    """This required method auto registers the checker during initialization.
+
+    :param linter: The linter to register the checker to.
+    """
     pass
 
   def transform(cls):
--- a/examples/custom.py
+++ b/examples/custom.py
@@ -1,11 +1,16 @@
+from typing import TYPE_CHECKING
+
 from astroid import nodes
 
 from pylint.checkers import BaseChecker
 from pylint.interfaces import IAstroidChecker
 
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
 
-# This is our checker class.
 # Checkers should always inherit from `BaseChecker`.
+
+
 class MyAstroidChecker(BaseChecker):
     """Add class member attributes to the class local's dictionary."""
 
@@ -60,10 +65,9 @@ class MyAstroidChecker(BaseChecker):
             in_class.locals[param.name] = node
 
 
-def register(linter):
-    """This required method auto registers the checker.
+def register(linter: "PyLinter") -> None:
+    """This required method auto registers the checker during initialization.
 
     :param linter: The linter to register the checker to.
-    :type linter: pylint.lint.PyLinter
     """
     linter.register_checker(MyAstroidChecker(linter))
--- a/pylint/checkers/async.py
+++ b/pylint/checkers/async.py
@@ -11,6 +11,7 @@
 """Checker for anything related to the async protocol (PEP 492)."""
 
 import sys
+from typing import TYPE_CHECKING
 
 import astroid
 from astroid import nodes
@@ -18,6 +19,9 @@ from astroid import nodes
 from pylint import checkers, interfaces, utils
 from pylint.checkers import utils as checker_utils
 from pylint.checkers.utils import decorated_with
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
 
 
 class AsyncChecker(checkers.BaseChecker):
@@ -94,6 +98,9 @@ class AsyncChecker(checkers.BaseChecker):
             )
 
 
-def register(linter):
-    """required method to auto register this checker"""
+def register(linter: "PyLinter") -> None:
+    """This required method auto registers the checker during initialization.
+
+    :param linter: The linter to register the checker to.
+    """
     linter.register_checker(AsyncChecker(linter))
--- a/pylint/checkers/async.py
+++ b/pylint/checkers/async.py
@@ -99,8 +99,4 @@ class AsyncChecker(checkers.BaseChecker):
 
 
 def register(linter: "PyLinter") -> None:
-    """This required method auto registers the checker during initialization.
-
-    :param linter: The linter to register the checker to.
-    """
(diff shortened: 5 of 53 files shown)


doc/how_tos/transform_plugins.rst
```diff
@@ -68,8 +68,7 @@ Module, Class, Function etc. In our case we need to transform a class. It can be
 
   import astroid
 
-  def register(linter):
-    # Needed for registering the plugin.
+  def register(linter: "PyLinter") -> None:
```
DanielNoord, line 71: Why do we remove the doc string? Doesn't it make sense to keep some (updated) form?

examples/custom.py
```diff
@@ -1,11 +1,16 @@
...
+
 from astroid import nodes
 
 from pylint.checkers import BaseChecker
 from pylint.interfaces import IAstroidChecker
 
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
 
-# This is our checker class.
```
DanielNoord, line 7: Why remove this?

examples/custom.py
```diff
@@ -60,10 +65,5 @@ def visit_call(self, node: nodes.Call) -> None:
             in_class.locals[param.name] = node
 
 
-def register(linter):
```
DanielNoord, line 63: Not going to comment about this more, but I think there is benefit to the doc string. Certainly if they include a `param` line.

doc/how_tos/transform_plugins.rst
Pierre-Sassoulas: I was thinking maybe the docstring in PyLinter was enough. What about:

```suggestion
  def register(linter: "PyLinter") -> None:
      """Register the plugin inside the PyLinter."""
```
? (Obviousely the same docstring for each function)

examples/custom.py
Pierre-Sassoulas: The class is called `` MyAstroidChecker`` so I thought it was self explanatory without the comment.

examples/custom.py
Pierre-Sassoulas: Let's discuss it in https://github.com/PyCQA/pylint/pull/5558/files#r772717062

doc/how_tos/transform_plugins.rst
DanielNoord: Yeah, but when you're looking at a checker-file (as a new contributor) you're not necessarily looking at the `PyLinter` file as well. Especially since it is not a class method these docstring are not necessarily "linked".

What about "Register the plugin with a linter. Function called during initialisation to register checkers." Can't hurt to be as explicit as possible.

examples/custom.py
DanielNoord: Hm yeah. Since it is "documentation" it doesn't hurt to be as explicit as possible, but I don't have a preference.

doc/how_tos/transform_plugins.rst
DanielNoord: We could even include the `param` line as well if we're going to make them uniform. I think `param` is used by `sphinx`.

Pierre-Sassoulas: Done, it's a lot of duplication imo but this is an issue with a design we cannot change.

pylint/checkers/async.py
```diff
@@ -94,6 +98,9 @@ def visit_asyncwith(self, node: nodes.AsyncWith) -> None:
             )
 
 
-def register(linter):
-    """required method to auto register this checker"""
+def register(linter: "PyLinter") -> None:
+    """This required method auto registers the checker during initialization.
+
+    :param linter: The linter to register the checker to.
+    """
```
cdce8p, line 105: Just my opinion, but after seeing this in every checker, I don't see much value in having a docstring from `register` aside from the examples in the documentation. It's unlikely someone will need to modify it once the checker is created.

My recommendation: Remove the docstring from all builtin checkers instead.

pylint/checkers/imports.py
```diff
@@ -69,10 +69,12 @@
 from pylint.exceptions import EmptyReportError
 from pylint.graph import DotBackend, get_cycles
 from pylint.interfaces import IAstroidChecker
-from pylint.lint import PyLinter
 from pylint.reporters.ureports.nodes import Paragraph, Section, VerbatimText
 from pylint.utils import IsortDriver, get_global_option
 
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
```
cdce8p, line 76: Any reason? The direct import seems to have worked previously

pylint/checkers/stdlib.py
```diff
@@ -39,14 +39,16 @@
...
 
 import astroid
 from astroid import nodes
 
 from pylint.checkers import BaseChecker, DeprecatedMixin, utils
 from pylint.interfaces import IAstroidChecker
-from pylint.lint import PyLinter
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
```
cdce8p, line 51: Same here

pylint/extensions/code_style.py
```diff
@@ -72,7 +73,7 @@ class CodeStyleChecker(BaseChecker):
         ),
     )
 
-    def __init__(self, linter: PyLinter) -> None:
+    def __init__(self, linter: "PyLinter") -> None:
```
cdce8p, line 76: Again, why?

pylint/reporters/reports_handler_mix_in.py
```diff
@@ -24,12 +24,12 @@ def __init__(self) -> None:
         self._reports: ReportsDict = collections.defaultdict(list)
         self._reports_state: Dict[str, bool] = {}
 
-    def report_order(self) -> List[checkers.BaseChecker]:
+    def report_order(self) -> List["BaseChecker"]:
```
cdce8p, line 27: The changes in this file are the only once not directly related to `PyLinter`. It would be better to deal with them separately. Please create another PR for it, especially considering the size fo this one such a change would get lost otherwise.

cdce8p (changes requested):

pylint/checkers/async.py
DanielNoord: This was my recommendation as I felt it was unproductive to remove docstrings that had already been written.
If (for whatever reason) we ever want to build documentation or anything from docstrings like we do in `astroid` it will be nice not to need to add them again. Besides, personally I don't mind duplication of docstrings as I find it helps attract contributors (and myself) to code. I find it better than having to look for `def register(` in the repository in the hope that some other similar function does have a docstring.

cdce8p: IMO documentation should mainly be used  **public** classes and methods. For whom would you want to publish implementation details? At the very least, remove the `:param linter:` line. A one liner is more than enough. For everything more, one should read the documentation anyway.

DanielNoord: Hmm I don't fully agree. Considering the size and age of our codebase I think some internal documentation is often quite useful. Also because we don't really have a steady group of employees that work with the internal code, but an ever-changing group of outside contributors. Besides implementation details we would ideally also publish contributing details, of which documentation in functions such as these should function.

I agree this might be a bit overkill, but looking at the typing effort and the difficulty we sometimes face to see how and why certain internal functions behave in a certain way I believe internal documentation is also a good step towards easier contributions.

cdce8p: It's already difficult enough to keep the public documentation up to date. In my experience no documentation is still better than outdated one. It's just one more thing that needs to be maintained, for just not enough benefit.

examples/custom.py
```diff
@@ -1,11 +1,16 @@
+from typing import TYPE_CHECKING
+
 from astroid import nodes
 
 from pylint.checkers import BaseChecker
 from pylint.interfaces import IAstroidChecker
 
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Well it could result in circular dependency later. If something is only used for typing I think using a typing guard is the way to go (especially with PyLinter that is used "everywhere" and use "everything")

doc/how_tos/custom_checkers.rst
```diff
@@ -201,7 +201,11 @@ Add the ``register`` function to the top level of the file.
 
 .. code-block:: python
 
-  def register(linter):
+  def register(linter: "PyLinter") -> None:
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done in 755b5c5c4d8243f5fc5de002a4d8ea6a93be01ec

doc/how_tos/plugins.rst
```diff
@@ -25,7 +25,11 @@ So a basic hello-world plugin can be implemented as:
 .. sourcecode:: python
 
   # Inside hello_plugin.py
-  def register(linter):
+  def register(linter: "PyLinter") -> None:
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done in 755b5c5c4d8243f5fc5de002a4d8ea6a93be01ec

doc/how_tos/transform_plugins.rst
```diff
@@ -68,8 +68,11 @@ Module, Class, Function etc. In our case we need to transform a class. It can be
...
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Done in 755b5c5c4d8243f5fc5de002a4d8ea6a93be01ec

pylint/checkers/async.py
Pierre-Sassoulas: What about keeping it in the doc examples, and removing it in the code ?

pylint/checkers/imports.py
Pierre-Sassoulas: I'm going to resolve all the issue with TYPE_CHECKING in favor of https://github.com/PyCQA/pylint/pull/5558#discussion_r773948413

pylint/reporters/reports_handler_mix_in.py
Pierre-Sassoulas: Originally it was necessary to make mypy works, we did multiple prior MR already (#5564 #5574).

examples/custom.py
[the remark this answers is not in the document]
cdce8p: I don't really think that's going to be an issue here. The checkers itself are already quite isolated by design. IMO guarding the import is unnecessary work. _We can always change it later if it should become necessary. Although I don't believe it will._

pylint/checkers/async.py
cdce8p: > What about keeping it in the doc examples, and removing it in the code ?

👍🏻 Works for me.

pylint/reporters/reports_handler_mix_in.py
cdce8p: Overlooked that it was just an import change. Btw. here the `if TYPE_CHECKING` guard does make sense as it's a mixin module for `pylinter` itself.

cdce8p, line 27: The return type should probably be `MutableSequence["BaseChecker"]` instead of `List`. The later is invariant, so technically the list can only contain `BaseChecker` classes. `MutableSequence` is covariant, thus subclasses are also allowed.

Pierre-Sassoulas: 8b00b6e

pylint/checkers/async.py
Pierre-Sassoulas: 4a1c99a

examples/custom.py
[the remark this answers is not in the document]
Pierre-Sassoulas: > guarding the import is unnecessary work.

The work is already done and reverting = more work.

pylint/checkers/similar.py
```diff
@@ -75,6 +76,8 @@
 from pylint.reporters.ureports.nodes import Table
 from pylint.utils import LinterStats, decoding_stream
 
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
 DEFAULT_MIN_SIMILARITY_LINE = 4
```
cdce8p, line 81:
```suggestion
if TYPE_CHECKING:
    from pylint.lint import PyLinter

DEFAULT_MIN_SIMILARITY_LINE = 4
```

pylint/checkers/typecheck.py
```diff
@@ -100,6 +110,8 @@
 from pylint.interfaces import INFERENCE, IAstroidChecker
 from pylint.utils import get_global_option
 
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
 CallableObjects = Union[
```
cdce8p, line 115:
```suggestion
if TYPE_CHECKING:
    from pylint.lint import PyLinter

CallableObjects = Union[
```

pylint/extensions/code_style.py
```diff
@@ -1,14 +1,15 @@
...
 
 from pylint.checkers import BaseChecker, utils
 from pylint.checkers.utils import check_messages, safe_infer
 from pylint.interfaces import IAstroidChecker
-from pylint.lint import PyLinter
 from pylint.utils.utils import get_global_option
 
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
 if sys.version_info >= (3, 10):
```
cdce8p, line 13:
```suggestion
if TYPE_CHECKING:
    from pylint.lint import PyLinter

if sys.version_info >= (3, 10):
```

cdce8p (approved):
Just some small formatting comments. Seems like black doesn't care about it. Looks good otherwise!

cdce8p:
@Pierre-Sassoulas All checks pass now

PR #5563, opened 2021-12-20
Fix #5557: Don't emit `comparison-with-callable` if the callable raises

### LINKED ISSUE #5557: False positive comparison-with-callable when comparing to Any in python 3.10
### Bug description

When I run pylint on the following code (saved as `a.py`):
```python
from typing import Any

def check_any(type_) -> bool:
    return type_ == Any
```
I get an warning:
```
a.py:4:11: W0143: Comparing against a callable, did you omit the parenthesis? (comparison-with-callable)
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:3:0: C0116: Missing function or method docstring (missing-function-docstring)
a.py:4:11: W0143: Comparing against a callable, did you omit the parenthesis? (comparison-with-callable)
```

### Expected behavior

No `comparison-with-callable` warning.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.10.0 | packaged by conda-forge | (default, Nov 10 2021, 13:20:59) [MSC v.1916 64 bit (AMD64)]
```

### OS / Environment

WIndows 10 (FYI: I can also reproduce this bug in Ubuntu 20.04)

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description
Typing constants such as `typing.Any` raise when called, which means they should not emit `comparison-with-callable`.

This PR skips emitting that message when a callable raises in its body (or if it's a typing constant, since some typing constants only raise via their decorator, not via their body).

Closes #5557

### DIFF
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -2514,12 +2514,12 @@ class ComparisonChecker(_BasicChecker):
         number_of_bare_callables = 0
         for operand in left_operand, right_operand:
             inferred = utils.safe_infer(operand)
-            if isinstance(inferred, bare_callables) and not any(
-                # Ignore callables that raise
-                isinstance(x, nodes.Raise)
-                # Or typing constants
-                or "typing._SpecialForm" in inferred.decoratornames()
-                for x in inferred.body
+            # Ignore callables that raise, as well as typing constants
+            # implemented as functions (that raise via their decorator)
+            if (
+                isinstance(inferred, bare_callables)
+                and "typing._SpecialForm" not in inferred.decoratornames()
+                and not any(isinstance(x, nodes.Raise) for x in inferred.body)
             ):
                 number_of_bare_callables += 1
         if number_of_bare_callables == 1:
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -2511,14 +2511,18 @@ class ComparisonChecker(_BasicChecker):
         left_operand, right_operand = node.left, node.ops[0][1]
         # this message should be emitted only when there is comparison of bare callable
         # with non bare callable.
-        if (
-            sum(
-                1
-                for operand in (left_operand, right_operand)
-                if isinstance(utils.safe_infer(operand), bare_callables)
-            )
-            == 1
-        ):
+        number_of_bare_callables: int = 0
+        for operand in left_operand, right_operand:
+            inferred = utils.safe_infer(operand)
+            if isinstance(inferred, bare_callables) and not any(
+                # Ignore callables that raise
+                isinstance(x, nodes.Raise)
+                # Or typing constants
+                or "typing._SpecialForm" in inferred.decoratornames()
+                for x in utils.safe_infer(operand).body
+            ):
+                number_of_bare_callables += 1
+        if number_of_bare_callables == 1:
             self.add_message("comparison-with-callable", node=node)
 
     @utils.check_messages(
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -2511,7 +2511,7 @@ class ComparisonChecker(_BasicChecker):
         left_operand, right_operand = node.left, node.ops[0][1]
         # this message should be emitted only when there is comparison of bare callable
         # with non bare callable.
-        number_of_bare_callables: int = 0
+        number_of_bare_callables = 0
         for operand in left_operand, right_operand:
             inferred = utils.safe_infer(operand)
             if isinstance(inferred, bare_callables) and not any(
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -2519,7 +2519,7 @@ class ComparisonChecker(_BasicChecker):
                 isinstance(x, nodes.Raise)
                 # Or typing constants
                 or "typing._SpecialForm" in inferred.decoratornames()
-                for x in utils.safe_infer(operand).body
+                for x in inferred.body
             ):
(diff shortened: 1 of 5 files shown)


pylint/checkers/base.py
```diff
@@ -2511,14 +2511,18 @@ def _check_callable_comparison(self, node):
...
         # with non bare callable.
-        if (
-            sum(
-                1
-                for operand in (left_operand, right_operand)
-                if isinstance(utils.safe_infer(operand), bare_callables)
-            )
-            == 1
-        ):
+        number_of_bare_callables: int = 0
```
Pierre-Sassoulas, line 2514:
```suggestion
        number_of_bare_callables = 0
```
Nitpick, I think this is implicit when assigning an integer.

tests/functional/c/comparison_with_callable.py
```diff
@@ -58,3 +58,12 @@ def fake_property(self, prop):
...
 if a == b:
     pass
+
+
+def eventually_raise():
+    print()
+    raise Exception
+
+
+if a == eventually_raise:
```
Pierre-Sassoulas, line 68: Nice test !

Pierre-Sassoulas (approved):
LGTM !

Pierre-Sassoulas:
The issue with the pipeline was due to #5561, it's reverted now, you should rebase to fix the primers :)

pylint/checkers/base.py
```diff
@@ -2511,14 +2511,18 @@ def _check_callable_comparison(self, node):
...
-        ):
+        number_of_bare_callables = 0
+        for operand in left_operand, right_operand:
+            inferred = utils.safe_infer(operand)
+            if isinstance(inferred, bare_callables) and not any(
+                # Ignore callables that raise
+                isinstance(x, nodes.Raise)
+                # Or typing constants
+                or "typing._SpecialForm" in inferred.decoratornames()
+                for x in utils.safe_infer(operand).body
```
DanielNoord, line 2522:
```suggestion
                for x in inferred.body
```

Can't we safe an inference call here?

tests/functional/c/comparison_with_callable.py
DanielNoord, line 68: Could you add a little docstring about why this shouldn't raise? (I feel like I'm becoming the "docstring guy" 😄 )

tests/functional/c/comparison_with_callable_typing_constants.py
```diff
@@ -0,0 +1,19 @@
+"""
+Typing constants are actually implemented as functions, but they
```
DanielNoord, line 2:
```suggestion
"""Typing constants are actually implemented as functions, but they
```

Well documented tests, thanks!

tests/functional/c/comparison_with_callable_typing_constants.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def check_any(type_) -> bool:
+    """See https://github.com/PyCQA/pylint/issues/5557"""
+    return type_ == Any
+
+
+def check_optional(type_) -> bool:
+    """
+    Unlike Any, Optional does not raise in its body.
```
DanielNoord, line 16:
```suggestion
    """Unlike Any, Optional does not raise in its body.
```

Good job catching this as well!

pylint/checkers/base.py
```diff
@@ -2511,14 +2511,18 @@ def _check_callable_comparison(self, node):
...
-        if (
-            sum(
-                1
-                for operand in (left_operand, right_operand)
-                if isinstance(utils.safe_infer(operand), bare_callables)
-            )
-            == 1
-        ):
+        number_of_bare_callables = 0
+        for operand in left_operand, right_operand:
```
DanielNoord, line 2515: I did not know you do not need to cast to tuple here. Learn something new every day :)

DanielNoord (changes requested):
Good job finding the special case for `Optional`. Small changes but looks good!

pylint/checkers/base.py
jacobtylerwalls: Yes, that was the whole reason for rewriting this as a for loop. I guess between rebases or something I lost it :D

tests/functional/c/comparison_with_callable_typing_constants.py
jacobtylerwalls: Fun fact: all of the typing constants raise first via the decorator rather than via their body, but when I saw that, I realized we ought to take care of detecting raises in the body, since that's easier.

pylint/checkers/base.py
jacobtylerwalls: Just like `for k, v in d.items()`! Or its evil twin `for k, in d.items()` that I worked on in #5505

DanielNoord: 😄

DanielNoord (approved):

pylint/checkers/base.py
```diff
@@ -2511,14 +2511,18 @@ def _check_callable_comparison(self, node):
...
+            inferred = utils.safe_infer(operand)
+            if isinstance(inferred, bare_callables) and not any(
+                # Ignore callables that raise
+                isinstance(x, nodes.Raise)
+                # Or typing constants
+                or "typing._SpecialForm" in inferred.decoratornames()
```
jacobtylerwalls, line 2521: This should be outside the generator.

PR #5564, opened 2021-12-20
Typo and small fixes before 5558

Pierre-Sassoulas:
## Type of Changes

## Description

Some small unrelated change before #5558

### DIFF
--- a/examples/custom.py
+++ b/examples/custom.py
@@ -7,7 +7,7 @@ from pylint.interfaces import IAstroidChecker
 # This is our checker class.
 # Checkers should always inherit from `BaseChecker`.
 class MyAstroidChecker(BaseChecker):
-    """Add class member attributes to the class locals dictionary."""
+    """Add class member attributes to the class local's dictionary."""
 
     # This class variable defines the type of checker that we are implementing.
     # In this case, we are implementing an AST checker.
@@ -31,8 +31,7 @@ class MyAstroidChecker(BaseChecker):
     options = (
         # Each option definition has a name which is used on the command line
         # and in config files, and a dictionary of arguments
-        # (similar to those to those to
-        # argparse.ArgumentParser.add_argument).
+        # (similar to argparse.ArgumentParser.add_argument).
         (
             "store-locals-indicator",
             {
@@ -48,10 +47,7 @@ class MyAstroidChecker(BaseChecker):
     def visit_call(self, node: nodes.Call) -> None:
         """Called when a :class:`.nodes.Call` node is visited.
 
-        See :mod:`astroid` for the description of available nodes.
-
-        :param node: The node to check.
-        """
+        See :mod:`astroid` for the description of available nodes."""
         if not (
             isinstance(node.func, nodes.Attribute)
             and isinstance(node.func.expr, nodes.Name)
--- a/pylint/extensions/broad_try_clause.py
+++ b/pylint/extensions/broad_try_clause.py
@@ -10,6 +10,7 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 """Looks for try/except statements with too much code in the try clause."""
+from typing import Union
 
 from astroid import nodes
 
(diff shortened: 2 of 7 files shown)


examples/custom.py
```diff
@@ -48,10 +47,7 @@ class MyAstroidChecker(BaseChecker):
     def visit_call(self, node: nodes.Call) -> None:
         """Called when a :class:`.nodes.Call` node is visited.
 
-        See :mod:`astroid` for the description of available nodes.
-
-        :param node: The node to check.
```
DanielNoord, line 53: I think this might still be relevant for `sphinx`. We tend to remove `type` in favour of in-line typing, but the `param` lines don't hurt.

pylint/extensions/broad_try_clause.py
```diff
@@ -59,7 +60,7 @@ def _count_statements(self, try_node):
 
         return statement_count
 
-    def visit_tryexcept(self, node: nodes.TryExcept) -> None:
+    def visit_tryexcept(self, node: Union[nodes.TryExcept, nodes.TryFinally]) -> None:
```
DanielNoord, line 63: Really? That might have an effect in other places as well. When I added this typing I assumed the method name reflects the node being visited..

pylint/extensions/docparams.py
```diff
@@ -589,7 +589,7 @@ class constructor.
             ):
                 self.add_message(
                     "missing-any-param-doc",
-                    args=(warning_node.name),
+                    args=warning_node.name,
```
DanielNoord, line 592: Could you revert this. This is building upon making `args` a tuple by default (one of my longer term projects).

pylint/interfaces.py
```diff
@@ -64,19 +64,15 @@ def implements(
...
     if not isinstance(kimplements, (list, tuple)):
         kimplements = (kimplements,)
     return any(issubclass(i, interface) for i in kimplements)
 
 
 class IChecker(Interface):
-    """This is a base interface, not designed to be used elsewhere than for
-    sub interfaces definition.
-    """
+    """Base interface, not to be used elsewhere than for sub interfaces definition."""
```
DanielNoord, line 75: "not to be used elsewhere other than"

pylint/interfaces.py
```diff
@@ -64,19 +64,15 @@ def implements(
     obj: "BaseChecker",
     interface: Union[Type["Interface"], Tuple[Type["Interface"], ...]],
 ) -> bool:
-    """Return whether the given object (maybe an instance or class) implements
-    the interface.
-    """
+    """Does the given object (maybe an instance or class) implements the interface."""
```
DanielNoord, line 67: "implements" should be "implement"

DanielNoord (commented):
My Github mails turned up in my spam yesterday so I didn't see this PR. I have some comments about this, perhaps we should make a follow-up PR?

pylint/extensions/broad_try_clause.py
DanielNoord: Never mind, I just saw how this checker does this.

examples/custom.py
Pierre-Sassoulas: I thought it does not add any information compared to just the having the name.

pylint/extensions/docparams.py
Pierre-Sassoulas:
```suggestion
                    args=(warning_node.name,),
```
Shouldn't it be this instead then ?

DanielNoord: I think so!

examples/custom.py
DanielNoord: I don't think we're using it but we do in `astroid`. For example see https://pylint.pycqa.org/projects/astroid/en/latest/api/base_nodes.html#astroid.nodes.NodeNG.scope and how it uses `returns`. I think we should keep these `param` lines in: somebody went through the effort of adding them and we never know when we might want to use it for our own documentation. Doesn't hurt to keep them in.

pylint/interfaces.py
Pierre-Sassoulas: Figured we might as well go all in on the typo front : https://github.com/PyCQA/pylint/pull/5575

pylint/extensions/docparams.py
Pierre-Sassoulas: Done in https://github.com/PyCQA/pylint/pull/5575/commits/3cdadeb3c28d613db0e46f4d5e7d13ffc4f23c4b

PR #5566, opened 2021-12-20
Add requirement files to list of files to run primer against

DanielNoord:

## Type of Changes

## Description

Seems like one of the new dependencies broke something in the primer.

This makes sure primer is also ran against changes to the dependencies, so we no longer unexpectedly fail on `main`.


Pierre-Sassoulas (approved):
👍

PR #5567, opened 2021-12-20
Revert "Update tbump requirement from ~=6.6.0 to ~=6.6.1"

Pierre-Sassoulas:
Reverts PyCQA/pylint#5561, we need to fix the primer but we can't block all current MR in order to do that.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -3,7 +3,7 @@
 coveralls~=3.3
 coverage~=6.2
 pre-commit~=2.16;python_full_version>="3.6.2"
-tbump~=6.6.1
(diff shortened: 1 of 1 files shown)


DanielNoord:
Do we understand why this is breaking the primer?

Pierre-Sassoulas:
Not yet, and it's not obvious for me after reading the logs. I intended to revert the revert and fix it later.

Pierre-Sassoulas:
We could also add tbump to the primers. I think it's a small project. I realized some project takes a lot of time like django, but we can add small project without too much drawbacks.

PR #5576, opened 2021-12-21
Add typing to file visiting attributes of ``PyLinter``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -540,11 +540,13 @@ class PyLinter(
         self.msgs_store = MessageDefinitionStore()
         self._checkers = collections.defaultdict(list)
         self._pragma_lineno = {}
-        self._ignore_file = False
-        # visit variables
+
+        # Attributes related to visiting files
         self.file_state = FileState()
         self.current_name: Optional[str] = None
-        self.current_file = None
+        self.current_file: Optional[str] = None
+        self._ignore_file = False
+
         self.stats = LinterStats()
         self.fail_on_symbols = []
         # init options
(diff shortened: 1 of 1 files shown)


pylint/lint/pylinter.py
```diff
@@ -1491,7 +1493,7 @@ def _add_one_message(
                 message_definition.msgid,
                 message_definition.symbol,
                 MessageLocationTuple(
-                    abspath,
+                    abspath or "",
```
Pierre-Sassoulas, line 1496: It means we can raise a message without telling the user what file is affected, right ? I suppose there was a crash somewhere before if we passed `None` ?

Pierre-Sassoulas (changes requested):
I think the code should fail if `current_file` is not set, raising an error without the file affected is not user friendly.

pylint/lint/pylinter.py
DanielNoord: Yes!

There is probably a check somewhere for `if abspath`. That would fail both with `""` and `None`. However, when I made `MessageLocationTuple` I typed `abspath` and `module` to be `str`. In the commit that added `MessageLocationTuple` I also added the `or ""` to `module`. I think it makes sense to force `abspath` to be a string here.

DanielNoord:
> I think the code should fail if `current_file` is not set, raising an error without the file affected is not user friendly.

I'm not sure `current_file` is every actually `None`. It think it's get set in a similar way to `set_reporter` and `set_current_module`. Refactoring this is a much larger task. I think the current fix should be fine for now.

Pierre-Sassoulas (approved):
> Refactoring this is a much larger task. I think the current fix should be fine for now.

As there are no changes in behavior with `abspath or ""` I agree :)

DanielNoord:
I double-checked and `_check_file` sets `linter.current_file` so as long as `nodes.Module.file` is not `None` `current_file` shouldn't ever be `None`. We could refactor to actually reflect this, but for now I think this is fine.

PR #5577, opened 2021-12-21
Make an implicitely abtract class explicitely abstract and remove disable

Pierre-Sassoulas:
## Type of Changes

## Description

It's better if abstractness is explicit and enforced by an error at instanciation.

### DIFF
--- a/pylint/config/man_help_formatter.py
+++ b/pylint/config/man_help_formatter.py
@@ -6,7 +6,6 @@ import sys
 import time
 
 
-# pylint: disable=abstract-method; by design?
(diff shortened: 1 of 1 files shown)


DanielNoord:
I force pushed another change to your branch. The class is not abstract but was missing an abstract method from the class it implements. The commit I added fixes that 😄

Pierre-Sassoulas:
> The class is not abstract but was missing an abstract method from the class it implements. 

😄 I wonder how it worked before, good job checking that !

DanielNoord:
The method is never used by us, but we should have implemented it (that's what the `abstract-method` checker is for).

PR #5580, opened 2021-12-21
Add ``DeprecationWarning`` for non-string ``current_name`` & ``file``

DanielNoord:

## Type of Changes

## Description

Turned out to be quite easy.

@Pierre-Sassoulas I wonder if this breaks anything. I don't think it shouldn't, but I'm hesitant about it...

### DIFF
--- a/tests/lint/unittest_lint.py
+++ b/tests/lint/unittest_lint.py
@@ -434,8 +434,8 @@ def test_set_unsupported_reporter(linter: PyLinter) -> None:
         linter.set_option("output-format", "missing.module.Class")
 
 
-def test_set_option_1(init_linter: PyLinter) -> None:
-    linter = init_linter
+def test_set_option_1(initialized_linter: PyLinter) -> None:
+    linter = initialized_linter
     linter.set_option("disable", "C0111,W0234")
     assert not linter.is_message_enabled("C0111")
     assert not linter.is_message_enabled("W0234")
@@ -444,8 +444,8 @@ def test_set_option_1(init_linter: PyLinter) -> None:
     assert not linter.is_message_enabled("non-iterator-returned")
 
 
-def test_set_option_2(init_linter: PyLinter) -> None:
-    linter = init_linter
+def test_set_option_2(initialized_linter: PyLinter) -> None:
+    linter = initialized_linter
     linter.set_option("disable", ("C0111", "W0234"))
     assert not linter.is_message_enabled("C0111")
     assert not linter.is_message_enabled("W0234")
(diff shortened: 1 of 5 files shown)


tests/lint/unittest_lint.py
```diff
@@ -251,7 +251,7 @@ def reporter():
 @pytest.fixture
 def init_linter(linter: PyLinter) -> PyLinter:
```
Pierre-Sassoulas, line 252: I don't like this fixture name, should be a noun like `initialized_linter` ? But it's outside of the scope of this MR.

Pierre-Sassoulas (approved):
:ok_hand:

DanielNoord:
What do you think about possible breakage of external tools? As shown if you `enable` or `disable` on the `linter` object directly this can break stuff... However, I'm not sure if that is supported in the first place.

tests/lint/unittest_lint.py
DanielNoord: I'll do a follow-up.

Pierre-Sassoulas:
>  possible breakage of external tools

Yeah, this will potentially breaks plugins that did not give the file prevousely. We need a deprecation period first, and we can only merge in 3.0.

DanielNoord:
Difficult to deprecate to.. Not sure how we could warn people.

Pierre-Sassoulas:
> Not sure how we could warn people.

I'm thinking a condition in ``MessageLocation`` or ``PyLinter.add_message`` like:

```python
if not abspath: # if not self.current_file: ...
    warning.warn("'current_file' will need to be a valid path in 3.0...")
```

tests/test_check_parallel.py
```diff
@@ -297,7 +297,6 @@ def test_sequential_checkers_work(self) -> None:
 
         # Add a sequential checker to ensure it records data against some streams
         linter.register_checker(SequentialTestChecker(linter))
-        linter.enable("R9999")
```
Pierre-Sassoulas, line 300: 👍

Pierre-Sassoulas (commented):
LGTM but maybe we need some tests for the deprecation warning ?

Pierre-Sassoulas:
Also should it be in 2.13 as it's only deprecation warning and no actual breaking changes ?

DanielNoord:
> LGTM but maybe we need some tests for the deprecation warning ?

We could. Imo, not really necessary. I don't think this is warning against any real code. There is so much depending on `current_name` being set, so this probably only raises against testing code.



> Also should it be in 2.13 as it's only deprecation warning and no actual breaking changes ?

Yeah, the milestone is from the previous version!

Pierre-Sassoulas (approved):

Pierre-Sassoulas (approved):
I let this go stale when it could be merged, my bad.

DanielNoord:
No worries! I often leave stuff open after I or you have reviewed to give others a chance to review but forget to revisit at a later point.

Pierre-Sassoulas:
>  I often leave stuff open after I or you have reviewed to give others a chance to review

Same here. As long as there's not too much conflicts there's no downside and we can merge a bunch of approved MR when we close up a release.

PR #5583, opened 2021-12-21
Fix false negative for `used-before-assignment` when an Except intervenes between Try/Finally

jacobtylerwalls:
## Type of Changes

## Description

Fixes false negative for `used-before-assignment` when evaluating statements in a finally block, there is a possibly failing statement in the try block, and an except handler intervenes between.

Found while working on #5582. Unhandled case from #5384 -- I had thought of this case, but didn't know how to solve it: a finally that is not the direct descendant of a try handler because we have try/except/finally. (Since we were just incrementally solving false negatives, that wasn't such a problem at the time.) After #5582, I now know how to handle it!

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -797,11 +797,25 @@ scope_type : {self._atomic.scope_type}
             ) = utils.get_node_first_ancestor_of_type_and_its_child(
                 other_node_statement, nodes.TryFinally
             )
-            if other_node_try_finally_ancestor is not closest_try_finally_ancestor:
-                continue
+            # other_node needs to descend from the try of a try/finally.
             if (
                 child_of_other_node_try_finally_ancestor
                 not in other_node_try_finally_ancestor.body
+            ):
+                continue
+            # If the two try/finally ancestors are not the same, then
+            # node_statement's closest try/finally ancestor needs to be in
+            # the final body of other_node's try/finally ancestor, or
+            # descend from one of the statements in that final body.
+            if (
+                other_node_try_finally_ancestor is not closest_try_finally_ancestor
+                and not any(
+                    other_node_final_statement is closest_try_finally_ancestor
+                    or other_node_final_statement.parent_of(
+                        closest_try_finally_ancestor
+                    )
+                    for other_node_final_statement in other_node_try_finally_ancestor.finalbody
+                )
             ):
                 continue
             # Passed all tests for uncertain execution
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -797,6 +797,8 @@ scope_type : {self._atomic.scope_type}
             ) = utils.get_node_first_ancestor_of_type_and_its_child(
                 other_node_statement, nodes.TryFinally
             )
+            if other_node_try_finally_ancestor is None:
+                continue
             # other_node needs to descend from the try of a try/finally.
             if (
                 child_of_other_node_try_finally_ancestor
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -652,23 +652,15 @@ scope_type : {self._atomic.scope_type}
 
         # If this node is in a Finally block of a Try/Finally,
         # filter out assignments in the try portion, assuming they may fail
-        if (
-            found_nodes
-            and isinstance(node_statement.parent, nodes.TryFinally)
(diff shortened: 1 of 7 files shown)


jacobtylerwalls:
Need to test this case -- newly failing on this branch:
```python
try:
    pass
finally:
    try:
        times = 1
    except TypeError:
        pass
    print(times)  # emits. shouldn't. and won't if the outer try goes away.
```

DanielNoord:
@jacobtylerwalls I haven't look at this but in #5582 we discussed doing some of the changes for `used-before-assignment` in `2.13`, get feedback, and then do some changes in `2.14`. Does this fall under the same category of "perhaps wait till 2.14" or should this be reviewed for `2.13`?

jacobtylerwalls:
Hi @DanielNoord thanks for looking into this. I think this is the same category ("no reason to wait; fixes false negatives for assumptions we've already doc'd and merged to master") as #5582, see https://github.com/PyCQA/pylint/pull/5582#issuecomment-1009142789. Happy to hear your views on this.

pylint/checkers/variables.py
```diff
@@ -741,6 +733,43 @@ def _uncertain_nodes_in_except_blocks(found_nodes, node, node_statement):
             uncertain_nodes.append(other_node)
         return uncertain_nodes
 
+    @staticmethod
+    def _uncertain_nodes_in_try_blocks_when_evaluating_finally_blocks(
+        found_nodes: List[nodes.NodeNG], node_statement: nodes.Statement
+    ):
```
DanielNoord, line 739: This is missing return type annotation

DanielNoord (approved):
@jacobtylerwalls misunderstood! Only one comment 😄

DanielNoord (approved):

Pierre-Sassoulas (changes requested):
Looks good, but the rebase for this is non trivial, and there's a strange conflict in the changelog and whatsnew. Could you rebase yourself @jacobtylerwalls ?

tests/functional/u/use/used_before_assignment_issue85.py
```diff
@@ -7,3 +7,61 @@ def main():
...
+    """
+    try:
+        pass
+    finally:
+        try:
+            times = 1
+        except TypeError:
+            pass
+        # Don't emit: only assume the outer try failed
+        print(times)
```
Pierre-Sassoulas, line 67: I don't understand ``Don't confuse assignments in different finally statements``. Shouldn't we raise or ``times`` be defined somewhere in the other finally statement?

tests/functional/c/consider/consider_using_with_open.py
```diff
@@ -137,7 +137,7 @@ def test_defined_in_try_and_finally(self):
         except FileNotFoundError:
             Path("foo").touch()
         finally:
-            file_handle.open("foo", encoding="utf")  # must not trigger
+            file_handle.open("foo", encoding="utf")  # [used-before-assignment] # must not trigger
```
Pierre-Sassoulas, line 140: Is it a new false positive or should the comment be removed ?

Pierre-Sassoulas (commented):
Looks good already, I have two questions regarding the tests.

tests/functional/c/consider/consider_using_with_open.py
jacobtylerwalls: I think `used-before-assignment` is correct, but `consider-using-with` is apparently not supposed to be triggered. I can update the comment!

Pierre-Sassoulas: Ha, I get it
```suggestion
            # +1: [used-before-assignment] 
            file_handle.open("foo", encoding="utf")  # must not trigger consider-using-with
```

tests/functional/u/use/used_before_assignment_issue85.py
jacobtylerwalls: The docstring looks like it was lazily copied from above and not really updated. I'll push again and explain in the docstring how I see what is being tested and where to go from here.

jacobtylerwalls: In rewriting the tests I discovered I could handle nesting better. So there are more tests now :-)

Pierre-Sassoulas (approved):
Amazing work, thank you @jacobtylerwalls !

Issue #5392, opened 2021-11-25
Migration to argparse from optparse


Pierre-Sassoulas:
### Current problem

``optparse`` is deprecated since python 3.2 and is not developed anymore. Also, the parsing are handled in a strange way with callback that created stupid bug that could probably be avoided by refactoring the code (https://github.com/PyCQA/pylint/pull/3581) We need to migrate to another CLI parser, the most logical being ``argparse`` or ``click``, but ``confuse`` is also a possibility because it permits to have default option in a configuration file that could be one of the configuration we already handle and make option / configuration consistent. I'm open to other libraries too. 

### Desired solution

No more optparse used in the code, small and modular handling of CLI argument parsing, no more special case, or minimal.

### Additional context

https://github.com/PyCQA/pylint/pull/5152#issuecomment-960890607
https://github.com/PyCQA/pylint/issues/5259#issuecomment-960879430

DanielNoord:
I think `argparse` makes the most sense as it is the direct successor to `optparse`.

To do this I think it makes the most sense to do some initial preparation, namely convert all options to a new class (we could use the one I proposed in #5152) and move them all to `PyLinter`. This avoids the current problem we're having with double settings etc and will probably also be faster (?) as we only need to parse arguments once instead for each individual checker.

saroad2:
Even though `argparse` is built-in library, and a way better solution than `optparse`, I **strongly recommend** using [Click](https://click.palletsprojects.com/en/8.0.x/) as your CLI library:

* It makes a shorter and cleaner code
* It has a lot of neat features that makes it so much easy to write CLI with.
* It is a living project that still publish new abilities and features
* It is supported by its developers
* It's widely used in the industry (personally, I use it all the time and in all of my projects)

Pierre-Sassoulas:
I like click, but my concern is it adds a new dependency. Also [argparse is compatible with confuse](https://confuse.readthedocs.io/en/latest/usage.html#command-line-options) which would be very nice to help  with consistency between options and configuration file enforced in a simple way.

saroad2:
In my humble opinion, adding a dependency is not that of a problem. In the end it's all matter of "buy what you won't implement", using a library instead of developing something out of your domain. I think that `click` answers to this definition.

[...]

PR #5584, opened 2021-12-21
Create an `Argument` class and allow convertion of optdict into them

DanielNoord:

## Type of Changes

## Description

This is my first proposed step towards #5392.

We could build on this in a separate branch, as I intend not to make this interfere too much with any current code. That should keep merge conflicts low.

All of this is based on the documentation of `argparse` found here:
https://docs.python.org/3/library/argparse.html

The idea of this PR is to showcase how we can transform our current `optdicts` into a `Argument` class that we can then at some point start passing to the `ArgumentsManager`. I think that second part would be the next step, to make this in a MVP. After that we can start working on adding new checkers.
I chose the logging module as it is a fairly easy module without only two options.

Don't hold back! Let me know what you think of this. All criticism is valuable!

**Note:** Ideally these commits wouldn't need to be squashed in a final merge. That makes cherry-picking later on much easier (see `2.0` branch in `astroid`). So after final review I'll revisit the commit history.

### DIFF
--- a/pylint/config/argument.py
+++ b/pylint/config/argument.py
@@ -1,0 +1,76 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""Definition of an Argument class and validators for various argument types.
+
+An Argument instance represents a pylint option to be handled by an argparse.ArgumentParser
+"""
+
+
+from typing import Any, Callable, Dict, List, Optional, Union
+
+from pylint import utils as pylint_utils
+
+_ArgumentTypes = Union[str, List[str]]
+"""List of possible argument types."""
+
+
+def _csv_validator(value: Union[str, List[str]]) -> List[str]:
+    """Validates a comma separated string."""
+    return pylint_utils._check_csv(value)
+
+
+_ASSIGNMENT_VALIDATORS: Dict[str, Callable[[Any], _ArgumentTypes]] = {
+    "choice": str,
+    "csv": _csv_validator,
+}
+"""Validators for all assignment types."""
+
+
+class _Argument:
+    """Class representing an argument to be passed by an argparse.ArgumentsParser.
+
+    This is based on the parameters passed to argparse.ArgumentsParser.add_message.
+    See:
+    https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument
+    """
+
+    def __init__(
+        self,
+        flags: List[str],
+        action: str,
+        default: _ArgumentTypes,
+        arg_type: str,
+        choices: Optional[List[str]],
+        arg_help: str,
+        metavar: str,
+    ) -> None:
+        self.flags = flags
+        """The name of the argument."""
+
+        self.action = action
+        """The action to perform with the argument."""
+
+        self.type = _ASSIGNMENT_VALIDATORS[arg_type]
+        """A validator function that returns and checks the type of the argument."""
+
+        self.default = self.type(default)
+        """The default value of the argument."""
+
+        self.choices = choices
+        """A list of possible choices for the argument.
+
+        None if there are no restrictions.
+        """
+
+        # argparse uses % formatting on help strings, so a % needs to be escaped
+        self.help = arg_help.replace("%", "%%")
+        """The description of the argument."""
+
+        self.metavar = metavar
+        """The metavar of the argument.
+
+        See:
+        https://docs.python.org/3/library/argparse.html#metavar
+        """
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -1,0 +1,107 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""Arguments manager class used to handle command-line arguments and options."""
+
+import argparse
+import configparser
+from typing import TYPE_CHECKING, Dict, List
+
+from pylint.config.argument import _Argument
+from pylint.config.exceptions import UnrecognizedArgumentAction
+from pylint.config.utils import _convert_option_to_argument
+
+if TYPE_CHECKING:
+    from pylint import checkers
+
+
+class _ArgumentsManager:
+    """Arguments manager class used to handle command-line arguments and options."""
+
+    def __init__(self) -> None:
+        self.namespace = argparse.Namespace()
+        """Namespace for all options."""
+
+        self._arg_parser = argparse.ArgumentParser(prog="pylint")
+        """The command line argument parser."""
+
+        self._argument_groups_dict: Dict[str, argparse._ArgumentGroup] = {}
+        """Dictionary of all the argument groups."""
+
+    def _register_options_provider(self, provider: "checkers.BaseChecker") -> None:
+        """Register an options provider and load its defaults."""
+        # pylint: disable-next=fixme
+        # TODO: Do something own_group parameter (see OptionsManagerMixIn.register_options_provider)
+        for opt, optdict in provider.options:
+            argument = _convert_option_to_argument(opt, optdict)
+            self._add_arguments_to_parser(provider.name, argument)
+
+        # pylint: disable-next=fixme
+        # TODO: Do something with option provider priorities (see OptionsManagerMixIn.register_options_provider)
+        # pylint: disable-next=fixme
+        # TODO: Do something with option groups within optdicts (see OptionsManagerMixIn.register_options_provider)
+
+        # pylint: disable-next=fixme
+        # TODO: Investigate performance impact of loading default arguments on every call
+        self._load_default_argument_values()
+
+    def _add_arguments_to_parser(self, section: str, argument: _Argument) -> None:
+        """Iterates over all argument sections and add them to the parser object."""
+        try:
+            section_group = self._argument_groups_dict[section]
+        except KeyError:
+            section_group = self._arg_parser.add_argument_group(title=section)
+            self._argument_groups_dict[section] = section_group
+        self._add_parser_option(section_group, argument)
+
+    @staticmethod
+    def _add_parser_option(
+        section_group: argparse._ArgumentGroup, argument: _Argument
+    ) -> None:
+        """Add an argument."""
+        if argument.action == "store":
+            section_group.add_argument(
+                *argument.flags,
+                action=argument.action,
+                default=argument.default,
+                type=argument.type,  # type: ignore[arg-type] # incorrect typing in typeshed
+                help=argument.help,
+                metavar=argument.metavar,
+                choices=argument.choices,
+            )
+        else:
+            raise UnrecognizedArgumentAction
+
+    def _load_default_argument_values(self) -> None:
+        """Loads the default values of all registered options."""
+        self.namespace = self._arg_parser.parse_args([], self.namespace)
+
+    def _parse_configuration_file(
+        self, config_parser: configparser.ConfigParser
+    ) -> None:
+        """Parse the arguments found in a configuration file into the namespace."""
+        arguments = []
+        for section in config_parser.sections():
+            for opt, value in config_parser.items(section):
+                arguments.extend([f"--{opt}", value])
+        # pylint: disable-next=fixme
+        # TODO: This should parse_args instead of parse_known_args
+        self.namespace = self._arg_parser.parse_known_args(arguments, self.namespace)[0]
+
+    def _parse_command_line_configuration(self, arguments: List[str]) -> None:
+        """Parse the arguments found on the command line into the namespace."""
+        # pylint: disable-next=fixme
+        # TODO: This should parse_args instead of parse_known_args
+        self.namespace = self._arg_parser.parse_known_args(arguments, self.namespace)[0]
+
+        # pylint: disable-next=fixme
+        # TODO: This should return a list of arguments with the option arguments removed
+        # just as PyLinter.load_command_line_configuration does
+
+    def _parse_plugin_configuration(self) -> None:
+        # pylint: disable-next=fixme
+        # TODO: This is not currently implemented.
+        # Perhaps we should also change the name to distuingish it better?
+        # See PyLinter.load_plugin_configuration
+        pass
--- a/pylint/config/config_initialization.py
+++ b/pylint/config/config_initialization.py
@@ -55,7 +55,7 @@ def _config_initialization(
 
     # Load command line arguments
     try:
-        args_list = linter.load_command_line_configuration(args_list)
+        parsed_args_list = linter.load_command_line_configuration(args_list)
     except SystemExit as exc:
         if exc.code == 2:  # bad options
             exc.code = 32
@@ -63,7 +63,7 @@ def _config_initialization(
 
     # args_list should now only be a list of files/directories to lint. All options have
     # been removed from the list
-    if not args_list:
+    if not parsed_args_list:
         print(linter.help())
         sys.exit(32)
 
@@ -71,7 +71,21 @@ def _config_initialization(
     # load plugin specific configuration.
     linter.load_plugin_configuration()
 
+    # pylint: disable-next=fixme
+    # TODO: Start of the implemenation of argument parsing with argparse
+    # When finished this should replace the implementation based on optparse
+
+    # First we parse any options from a configuration file
+    linter._parse_configuration_file(config_parser)
+
+    # Second we parse any options from the command line, so they can override
+    # the configuration file
+    linter._parse_command_line_configuration(args_list)
+
+    # Lastly we parse any options from plugins
+    linter._parse_plugin_configuration()
+
     # Now that plugins are loaded, get list of all fail_on messages, and enable them
     linter.enable_fail_on_messages()
 
-    return args_list
+    return parsed_args_list
--- a/pylint/config/utils.py
+++ b/pylint/config/utils.py
@@ -1,0 +1,34 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""Utils for arguments/options parsing and handling."""
+
+
+from typing import Any, Dict
+
+from pylint.config.argument import _Argument
+
+IMPLEMENTED_OPTDICT_KEYS = {"action", "default", "type", "choices", "help", "metavar"}
+"""This is used to track our progress on accepting all optdict keys."""
+
+
+def _convert_option_to_argument(opt: str, optdict: Dict[str, Any]) -> _Argument:
+    """Convert an optdict to an Argument class instance."""
+    # See if the optdict contains any keys we don't yet implement
+    # pylint: disable-next=fixme
+    # TODO: This should be removed once the migration to argparse is finished
+    for key, value in optdict.items():
+        if key not in IMPLEMENTED_OPTDICT_KEYS:
+            print("Unhandled key found in Argument creation:", key)  # pragma: no cover
+            print("It's value is:", value)  # pragma: no cover
+
+    return _Argument(
+        flags=[f"--{opt}"],
+        action=optdict.get("action", "store"),
+        default=optdict["default"],
+        arg_type=optdict["type"],
+        choices=optdict.get("choices", None),
+        arg_help=optdict["help"],
+        metavar=optdict["metavar"],
+    )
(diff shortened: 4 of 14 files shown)


DanielNoord:
If anybody could help me figure out why this fails on Windows that would be much appreciated!

jacobtylerwalls:
I think it's that [subroutines can't be pickled on Windows](https://stackoverflow.com/a/8805244/14818839). Any way we can avoid pickling `ArgumentParser`?

DanielNoord:
I'll try and come up with something!

DanielNoord:
I have fixed the issue with `argparse.ArgumentsParser` pickling. It is annoying as it means we'll need to instantiate an `ArgumentsParser` and extract any relevant information during the init of the `ArgumentsManager`. 
On the other hand, that does force us to think clearly about how we handle arguments parsing and might help us avoid any redundant calls 😄

DanielNoord:
Converting this to draft as I found issues with the current implementation. In retrospect I think it doesn't make sense to break this into smaller PRs, as we might run into problems after earlier parts have been merged.

Therefore, I'll keep working on this until we can completely use `argparse` to parse the options for the `logging` module. That should be a good test that at least basic functionality is guaranteed.

Pierre-Sassoulas (commented):
Do we really need new classes ? Or could we `use argparse.Namespace` directly ? optparse is more complex so if we try to do a "small refactor" we'll keep the optparse complexity in our implementation even when using argparse.

One thing I had in mind for this refactor was separating parsing option / checking. Maybe  creating NamedTuple or DataClasses and gives them to the checker pre-parsed instead of parsing the option inside each checker (in order to not have breaking change we could parse the option if the object is not given ?).

DanielNoord:
> Do we really need new classes ? Or could we `use argparse.Namespace` directly ? optparse is more complex so if we try to do a "small refactor" we'll keep the optparse complexity in our implementation even when using argparse.

I think it is better to create new classes as it allows us to more easily work on this. If we start to inject into the pre-existing `optparse` based classes we will create merge conflicts very easily.

> One thing I had in mind for this refactor was separating parsing option / checking. Maybe creating NamedTuple or DataClasses and gives them to the checker pre-parsed instead of parsing the option inside each checker (in order to not have breaking change we could parse the option if the object is not given ?).

I'm not sure what you mean here.

I think the flow would be 1) register options on parser object, 2) parse options with parser object, 3) set namespace object as config namespace of linter object.

DanielNoord:
This might be more difficult than I thought. I had a pretty good idea of how to not require `argparse.ArgumentParser` as subclass of `PyLinter`, but we need some way to store the `ArgumentParser`. Otherwise we would need to register all options every time we want to change any of them.

The problem is that even `Run` gets pickled, so I don't know a good place to register an `ArgumentParser` and store it.. I'll keep thinking about this, but this is a very large limitation of `argparse` which I'm not sure we can overcome..

Pierre-Sassoulas:
> 3) set namespace object as config namespace of linter object.
 
 I agree with 1) and 2) but I don't understand this part.
  
> we need some way to store the ArgumentParser. Otherwise we would need to register all options every time we want to change any of them.

Couldn't we store the result of the parsing, i.e. an ``argparse.Namespace`` ?

DanielNoord:
>  I agree with 1) and 2) but I don't understand this part.

After the parsing is done (ideally in the `Run` init) the resulting namespace needs to be made accessible to the linter class. So it can be accessed there.


> Couldn't we store the result of the parsing, i.e. an ``argparse.Namespace`` ?

There is a difference between the Namespace and the storage of arguments. The namespace only stores the current value whereas the parser objects also stores stuff like default value and description. I think it's important that information is not lost after initial parsing of arguments is done, so we'll preferably need to save both the parser and the namespace. Currently I don't see a way to store the parser, as it seems everything within Run is pickled..

Pierre-Sassoulas:
>  the resulting namespace needs to be made accessible to the linter class. 

Ok I agree with that too.

> I think it's important that information is not lost after initial parsing of arguments is done,

Why ? Where could we use the default value and the description after the parsing is done ?

DanielNoord:
> Why ? Where could we use the default value and the description after the parsing is done ?

We would need it to "reparse" for sub directories and for help messages. We would then need to recollect and reinstantiate all options for every subdirectory. That's not really good performance wise..,

Pierre-Sassoulas:
Can't we parse all the options for each directories first and store them ? (By the way the configuration by directory feature was removed from flake8 because it was bringing "90% of the bugs"  we might want to reconsider that one)

DanielNoord:
> Can't we parse all the options for each directories first and store them ? (By the way the configuration by directory feature was removed from flake8 because it was bringing "90% of the bugs" we might want to reconsider that one)

That might be possible, but that would definitely come later.

However, I think I might have found a way to do it. We won't store information about the arguments this way, but I think we could design everything in such a way that that is not necessary.

DanielNoord:
@Pierre-Sassoulas Would you consider using [`dill`](https://github.com/uqfoundation/dill) to pickle the `PyLinter` object for the multiprocessing handling? I'm basing this on this SO question which seems to suggest it should be able to fix our problem:
https://stackoverflow.com/questions/8804830/python-multiprocessing-picklingerror-cant-pickle-type-function

According to their README it has been download 200M+ times via Pypi and supports almost all python versions (`Python >=2.7, !=3.0.*`). Being able to pickle the input to `multiprocessing.Pool` would save all of the issues and allow us to make a more intuitive system.

---------------

Some additional arguments for why I'm exploring fixing the `multiprocessing` issue instead of working around: we can't be sure that all plugins and tools (such as prospector) create a `Run` class. In our own tests we don't do so, but after some additional thoughts I don't think we should do stuff in the `Run` `__init__` that is needed for `PyLinter` to work. (That's counterintuitive and bad design).
Thus, my third (fourth, fifth?) attempt at this doesn't really work. I hoped that by doing the argument parsing in `Run` we could skip the pickling.

Pierre-Sassoulas:
Look like ``dill`` do not have any dependencies of its own and it's 3-clause BSD license seems compatible with pylint. We would more than double their download per day by adding them as a dependencies but it's already used a lot. 

> I don't think we should do stuff in the Run __init__ that is needed for PyLinter to work. (That's counterintuitive and bad design).

Depends on what we want in PyLinter, right now it's doing almost everything. There's a problem with breaking change in downstream plugins to consider for sure, but let's put that aside for a minute If I understood what we're talking about and the current state of the code, PyLinter (each checker even) is parsing the options right now ? Parsing the options first **before** launching anything and especially the parallel run actually make more sense imo ? i.e. PyLinter could be the class that analyses a set of file given a set of parsed options and have less responsibilities. Alternatively we could include multiprocessing handling in PyLinter if we want PyLinter and not Run to be the entrypoint ?

DanielNoord:
> Look like `dill` do not have any dependencies of its own and it's 3-clause BSD license seems compatible with pylint. We would more than double their download per day by adding them as a dependencies but it's already used a lot.

I'll start work on a PR implementing it for our `parallel` run then, so we can continue discussion about actually doing this in a separate PR 😄 

> Depends on what we want in PyLinter, right now it's doing almost everything. There's a problem with breaking change in downstream plugins to consider for sure, but let's put that aside for a minute If I understood what we're talking about and the current state of the code, PyLinter (each checker even) is parsing the options right now ?

Since every `checker` is an `OptionsProvider` each plugin or checker added after initialising `PyLinter` can add new options that need to be read, "added to the parser" and then "parsed into the Namespace". The flow is as follows:
1. Initialise `Run`, 2. Initialise `PyLinter` and its associated `parser`, 3. "Add" all options from the `PyLinter` class, 4. "Parse" `PyLinter` options, 5. "Add" all options from default checkers and plugins, 6. "Parse" all options from default checkers and plugins, 7. "Add" all options from extra plugins, 8. "Parse" all options from extra plugins.

Note `load_defaults` in `register_checker` here:
https://github.com/PyCQA/pylint/blob/e815843293adf100662fe4e183c34f3040529a15/pylint/lint/pylinter.py#L754-L763

The problem is thus that at step 5 and step 7 we need to re-initialise an `ArgumentsParser` if we can't store the parser as an attribute or subclass of `PyLinter`. Adding the parser to `Run` doesn't help as for step 7 we don't have access to that step anymore.
I have tested what would happen if we had to re-initialise an `ArgumentsParser` 100 times and it became a significant impact on runtime.

> Parsing the options first **before** launching anything and especially the parallel run actually make more sense imo ? i.e. PyLinter could be the class that analyses a set of file given a set of parsed options and have less responsibilities. Alternatively we could include multiprocessing handling in PyLinter if we want PyLinter and not Run to be the entrypoint ?

We can't really do this as we need to parse the `load-plugins` option before knowing all the options we have to parse.

Pierre-Sassoulas:
> We can't really do this as we need to parse the load-plugins option before knowing all the options we have to parse.

I'm probably off base here but what if we first parse ``load-plugin`` alone, then add their options in 5 at the same time than default checker and plugins ? And launch multiprocessing only after that. Because right now multiprocessing happens in 2., right ? Does'nt it happens too soon and it would be more efficient to do it later when we know exactly what "work" we'll have to do ?

DanielNoord:
> I'm probably off base here but what if we first parse `load-plugin` alone, then add their options in 5 at the same time than default checker and plugins ?

We already kind of do this, but only in `Run`. If we want to have the `PyLinter` class work on its own (like it needs to do in `prospector`) we will need to the argument parsing in `PyLinter` and thus create an `ArgumentParser` there. If we want to allow plugins to add new options (which we should) we will still need to store (or re-initialize) a parser each time we register one.

I think I looked into loading all plugins, "adding" all options and then "parsing" them but stopped exploring it further as it would mean a complete rewrite of the `__init__` of `PyLinter`. I can look into it again though.

> And launch multiprocessing only after that. Because right now multiprocessing happens in 2., right ? Does'nt it happens too soon and it would be more efficient to do it later when we know exactly what "work" we'll have to do ?

I believe multiprocessing happens after step 6 now.

Another hack that is the result of the need to pickling (I believe) is that we set `linter.reporter` to `None` here:
https://github.com/PyCQA/pylint/blob/532be9afa75bd090a389db0f172c47ba9d37b411/pylint/lint/parallel.py#L106-L113
See this commit: https://github.com/Pierre-Sassoulas/pylint/commit/8babeffde54a696e7ba84d69f8fc1f871e432950


I'm going to see if I can "fix" the multiprocessing first, as it would allow for a much cleaner design and worry about the order of argument registering and parsing after. If that's okay with you of course!

Pierre-Sassoulas:
> I'm going to see if I can "fix" the multiprocessing first, as it would allow for a much cleaner design and worry about the order of argument registering and parsing after

Sounds good, the order of arguments seems to be optimal right now anyway, thank you for clarifying

DanielNoord:
Another iteration of this... But this has passed CI on my own fork already, so should be more robust.
Some points:

1. This is obviously blocked by #5609 
2. Originally I intended to work on this in a separate branch. However, after looking at the performance impact of the current version I think this might not be necessary. We spend less than 15 ms in the `arguments_manager` file per run, where 7 ms comes from `_parse_configuration_file` which should not take too much extra time as we add more checkers to the new parsing method (its runtime is not currently linearly dependent on the amount of checkers). Besides, at some point we could also looking into not registering "argparse-using" checkers in the optparse manager, which would reduce performance impact of migration as well.
3. The way the migration works is that each checker can pass `future_option_parsing=True` to the `__init__` of `BaseChecker` to start parsing with `argparse`. As such we can do the change checker per checker, tackling any issues as we go along. For example, in the current version not all `optdict` keys are supported and they can be added when we get to them. There are some additional TODO's that should be fixed as well, but aren't necessary for the base functionality which is shown here.
4. In the current version we don't need a `OptionsProviderMixIn` as everything is handled within 6 lines in `BaseChecker` and the actual `ArgumentsManager`. I hope this can stay true as it would remove complexity from the code.
5. I have opted to rewrite all validators for options. This gives me (and others) some extra work, but make sure we can completely remove the `optparse` files when we're done and the deprecation-period is over and allows me to correctly type everthing. One of the issues with the current system is that we have no good way to know what type everything can be, this can be avoided by rewriting everything from the ground up.

Please don't hold back on any criticism. I think this is the right way forward, but we only get one opportunity to fix this so we should be really considerate of any potential problems and issues.

tests/config/data/logging_format_interpolation_style.py
```diff
@@ -0,0 +1,16 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
DanielNoord, line 1: @Pierre-Sassoulas Can we exclude the copyright hook for this directory?

Pierre-Sassoulas: Sure !

.pre-commit-config.yaml
```diff
@@ -69,7 +69,7 @@ repos:
             "--load-plugins=pylint.extensions.docparams",
           ]
         # disabled plugins: pylint.extensions.mccabe
-        exclude: tests/functional/|tests/input|tests/extensions/data|tests/regrtest_data/|tests/data/|doc/
+        exclude: tests/functional/|tests/input|tests/regrtest_data/|tests(/\w*)*/data/|doc/
```
Pierre-Sassoulas, line 72: That's great, maybe we could add it line 16 in the "fixtures" variable too ?

pylint/config/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+def _convert_option_to_argument(opt: str, optdict: Dict[str, Any]) -> _Argument:
+    """Convert an optdict to an Argument class instance."""
+    # See if the optdict contains any keys we don't yet implement
+    # pylint: disable-next=fixme
+    # TODO: This should be removed once the migration to argparse is finished
+    for key, value in optdict.items():
+        if key not in IMPLEMENTED_OPTDICT_KEYS:
+            print("Unhandled key found in Argument creation:", key)  # pragma: no cover
+            print("It's value is:", value)  # pragma: no cover
```
Pierre-Sassoulas, line 24: Can this be seen by end user with custom plugin ? Maybe we could add a link to our issue tracker ?

Pierre-Sassoulas (commented):
Look good already. I only skimmed because I'm not an expert in argparse custom parser and even less so in optparse so it's a hard review for me.

pylint/config/utils.py
DanielNoord: I don't think so (right now). Plugins will need to start using the `argparse` system and as long as we don't tell about them they probably won't.

I hope to make `argparse` the standard for `pylint's` own checkers in `2.14` and then start telling people about them. That would be after we fix all these open issues though.

.pre-commit-config.yaml
DanielNoord: I'll do a separate PR for the pre-commit changers

DanielNoord: See #5974.

Pierre-Sassoulas (approved):

DanielNoord:
@Pierre-Sassoulas Do you want others to review this as well before merging? @jacobtylerwalls possibly?

Pierre-Sassoulas:
More reviews is always a good thing if we can afford the luxury :)

pylint/config/config_initialization.py
```diff
@@ -55,23 +55,37 @@ def _config_initialization(
...
     # args_list should now only be a list of files/directories to lint. All options have
     # been removed from the list
-    if not args_list:
+    if not parsed_args_list:
         print(linter.help())
         sys.exit(32)
 
     # We have loaded configuration from config file and command line. Now, we can
     # load plugin specific configuration.
     linter.load_plugin_configuration()
```
jacobtylerwalls, line 72: Why is this still needed (given that it was moved below)?

pylintrc
```diff
@@ -393,7 +393,7 @@ ignored-classes=SQLObject, optparse.Values, thread._local, _thread._local
 # List of members which are set dynamically and missed by pylint inference
 # system, and so shouldn't trigger E1101 when accessed. Python regular
 # expressions are accepted.
-generated-members=REQUEST,acl_users,aq_parent
+generated-members=REQUEST,acl_users,aq_parent,argparse.Namespace
```
jacobtylerwalls, line 396: When I revert this change I don't see a problem (locally). Problem on CI?

tests/config/test_argparse_config.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_namespace_creation(self) -> None:
+        """Test that the linter object has a namespace attribute and that it is not empty"""
+
+        assert self.base_run.linter.namespace
+        assert self.base_run.linter.namespace._get_kwargs()
+
+
+class TestArgparseOptionsProviderMixin:
+    """Tests for the argparse implementation of OptionsProviderMixin.
```
jacobtylerwalls, line 32:
```suggestion
    """Tests for the argparse implementation of OptionsProviderMixIn.
```

tests/config/test_argparse_config.py
```diff
@@ -0,0 +1,56 @@
...
+
+    The logger checker is used as an example checker for this implementation.
+    """
+
+    @staticmethod
+    def test_logger_without_options() -> None:
+        """Check that we raises messages when we do not supply any options."""
```
jacobtylerwalls, line 39:
```suggestion
        """Check that we raise messages when we do not supply any options."""
```

pylint/config/argument.py
```diff
@@ -0,0 +1,76 @@
...
+    """Class representing an argument to be passed by an argparse.ArgumentsParser.
+
+    This is based on the parameters passed to argparse.ArgumentsParser.add_message.
+    See:
+    https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument
+    """
+
+    def __init__(
+        self,
+        name: List[str],
```
jacobtylerwalls, line 41: Maybe `flags`? or `names`? It's odd for `name` to be a list. (And most often a list of one name...)

pylint/config/arguments_manager.py
```diff
@@ -0,0 +1,107 @@
...
+        """Namespace for all options."""
+
+        self._arg_parser = argparse.ArgumentParser(prog="pylint")
+        """The command line argument parser."""
+
+        self._argument_groups_dict: Dict[str, argparse._ArgumentGroup] = {}
+        """Dictionary of all the argument groups."""
+
+    def _register_options_provider(self, provider: "checkers.BaseChecker") -> None:
+        """Register an options provider and loads its defaults."""
```
jacobtylerwalls, line 33:
```suggestion
        """Register an options provider and load its defaults."""
```

pylint/config/arguments_manager.py
```diff
@@ -0,0 +1,107 @@
...
+                *argument.name,
+                action=argument.action,
+                default=argument.default,
+                type=argument.type,  # type: ignore[arg-type] # incorrect typing in typeshed
+                help=argument.help,
+                metavar=argument.metavar,
+                choices=argument.choices,
+            )
+        else:
+            raise UnrecognizedArgumentAction
```
jacobtylerwalls, line 74: May I ask if there is a reason we can't just permit other actions like `store_true` and send them along? As in, is this something a plugin writer could take advantage of? (I'm used to `store_true` and the docs/tutorial suggest it also...)

pylint/config/arguments_manager.py
```diff
@@ -0,0 +1,107 @@
...
+        self.namespace = self._arg_parser.parse_args([], self.namespace)
+
+    def _parse_configuration_file(
+        self, config_parser: configparser.ConfigParser
+    ) -> None:
+        """Parse the arguments found in a configuration file into the namespace."""
+        arguments = []
+        for section in config_parser.sections():
+            for opt, value in config_parser.items(section):
+                arguments.extend([f"--{opt}", value])
```
jacobtylerwalls, line 87: This is probably a nothing. But, since you know the parsing flow very well, I'll ask -- is there a chance `opt` is already in `arguments`? And if so, does `argparse` pick the right one later?

jacobtylerwalls (commented):
Terrific, thanks for moving this migration forward. A few questions, some for my own understanding.

pylint/config/config_initialization.py
DanielNoord: I think you're referring to `linter._parse_plugin_configuration()`?

That has not been implemented yet. The code below is just the order in which things should occur. This is "captured" by `# TODO: Start of the implemenation of argument parsing with argparse` and the `TODO` in the method definition.

pylintrc
DanielNoord: I think this might become problematic at some point. Perhaps it was in an earlier version but doing `namespace.setting` will raise a `no-member` for `setting` on `namespace`. I could remove this for now, but we'll likely need it at some point.

pylint/config/argument.py
DanielNoord: The documentation refers to it as `name or flags`. I will change it to flags.

pylint/config/arguments_manager.py
DanielNoord: We should at some point. For now we only need `store` as logging only has `store` options (that's why I chose to implement there first).

The `_Argument` class also does not support of parameters that can be passed to `add_argument()`. I want to add these on a if-needed basis when I convert more checkers.

pylint/config/arguments_manager.py
DanielNoord: Hm, I don't actually know how double arguments are picked up by the `config_parser`. Although I think that it will probably raise a `configparser.DoubleKeyError` or something. (`configparser` is not our own code).

The arguments in the namespace get overwritten if they get re-parsed. That's why the order in `config_initialization` is important.

jacobtylerwalls (approved):

jacobtylerwalls:
Thanks for answering my questions. Is now a good time to edit the branch history?

DanielNoord:
I guess so. I'll do a rebase.

DanielNoord:
@Pierre-Sassoulas Please merge if you're okay with this and the number of reviews 😄

Pierre-Sassoulas (approved):
Great refactor @DanielNoord !

PR #5591, opened 2021-12-23
Create ``_config_initialization`` util and use in functional tests

DanielNoord:

## Type of Changes

## Description

I'm not sure about the location of this `util`, but I think it makes sense here.

It turns out the functional tests had their own way of loading the config. This became problematic while working on `argparse` as I would need to update config loading in two places and there is a possibility that they diverge. This add a config initialisation method which is used both by `Run` and the functional tests. The functional test should therefore better mimick an actual run of `pylint`. Other benefits include that we now have access to the `init-hook` option and that there is less duplication of code.


pylint/config/utils.py
```diff
@@ -0,0 +1,75 @@
...
+    args_list: List[str],
+    reporter: Union[reporters.BaseReporter, reporters.MultiReporter, None] = None,
+    config_file: Optional[str] = None,
+    verbose_mode: Optional[bool] = None,
+) -> List[str]:
+    """Parse all available options, read config files and command line arguments and
+    set options accordingly.
+    """
+
+    # Read the config file. The parser is stored on linter.cfgfile_parser
```
Pierre-Sassoulas, line 25: I does not make much sense, right ? Why isn't it in the ``pylint.config`` package ? (Refactor for later, or maybe even prior to this one ?)

pylint/config/utils.py
```diff
@@ -0,0 +1,75 @@
...
+        linter.read_config_file(config_file=config_file, verbose=verbose_mode)
+    except OSError as ex:
+        print(ex, file=sys.stderr)
+        sys.exit(32)
+    config_parser = linter.cfgfile_parser
+
+    # Run init hook, if present, before loading plugins
+    if config_parser.has_option("MASTER", "init-hook"):
+        exec(  # pylint: disable=exec-used
+            utils._unquote(config_parser.get("MASTER", "init-hook"))
```
Pierre-Sassoulas, line 36: The init-hook is not tested as it's just been added, maybe we could add a small functional test for it ?

pylint/lint/run.py
```diff
@@ -327,40 +328,8 @@ def __init__(
...
-        try:
-            args = linter.load_command_line_configuration(args)
-        except SystemExit as exc:
-            if exc.code == 2:  # bad options
-                exc.code = 32
-            raise
-        if not args:
-            print(linter.help())
-            sys.exit(32)
+        args = _config_initialization(linter, args, reporter, verbose_mode=self.verbose)
```
Pierre-Sassoulas, line 332: 👌

Pierre-Sassoulas (changes requested):
The factorization of this code is amazing 👍 I would call the new file ``config_initialization.py``,

pylint/config/utils.py
DanielNoord: I'm not sure. But it is kinda nice: we have access to it everywhere we have access to the `linter`. (Exactly the opposite of the problem I'm facing with `argparse` right now). So for now I'd said let's leave it as is.

pylint/config/utils.py
DanielNoord: Got any good ideas? Not really sure what people use it for anyway.

DanielNoord:
> The factorization of this code is amazing 👍 I would call the new file `config_initialization.py`,

I'm adding other `utils` in the `argparse` PR as well so I thought it made sense to already create that file.

Although now that I think about it, perhaps we should separate them.

pylint/config/utils.py
DanielNoord: Btw, it needs to be a functional test with an `init-hook`. So just `raise RuntimError` like we do in another test won't work sadly...

Pierre-Sassoulas: I know that people used it to modify the sys path, in order to increase max inference in astroid, the maximum recursion depth for genuinely large recursion trees... Well that's the one I hear about in issues anyway (all in order to avoid problem with pylint as you see). There probably others.

DanielNoord: I found a way to do it :)

tests/testutils/data/init_hook.rc
```diff
@@ -0,0 +1,3 @@
+[MASTER]
+
+init-hook=raise RuntimeError
```
Pierre-Sassoulas, line 3: Huhu, that's clever.

Pierre-Sassoulas (approved):
This init hook test is hacky and fun, I like it.

tests/testutils/data/init_hook.rc
DanielNoord: Not actually my idea, found in the tests for `init-hook` as a command line argument. Works quite well though 😄

PR #5594, opened 2021-12-23
Make functional tests always go through config initialization

DanielNoord:

## Type of Changes

## Description

Follow up to #5591. Diff should be much smaller after that has been merged (adding blocked in the meantime).

This makes it so that whenever a functional test does not have a `.rc` file we still pass a `pylintrc` file to the configuration initialiser.
This does not fix any current problems but allows me (and others) to assume that `_config_initialization` is always ran, also for functional tests that skip instantiating a `Run` object. That allows me to then work within `_config_initialization` for the implementation of `argparse` and make `argparse` and `optparse` work along side each other while we're still working on it.

### DIFF
--- a/pylint/testutils/lint_module_test.py
+++ b/pylint/testutils/lint_module_test.py
@@ -54,12 +54,12 @@ class LintModuleTest:
             pass
 
         try:
-            sources = [test_file.source]
+            args = [test_file.source]
         except NoFileError:
             # If we're still raising NoFileError the actual source file doesn't exist
-            sources = [""]
+            args = [""]
         _config_initialization(
-            self._linter, sources, config_file=rc_file, reporter=_test_reporter
+            self._linter, args_list=args, config_file=rc_file, reporter=_test_reporter
         )
         self._test_file = test_file
         self._config = config
--- a/pylint/testutils/lint_module_test.py
+++ b/pylint/testutils/lint_module_test.py
@@ -9,7 +9,7 @@ from collections import Counter
 from io import StringIO
 from pathlib import Path
 from typing import Counter as CounterType
-from typing import Dict, List, Optional, TextIO, Tuple
+from typing import Dict, List, Optional, TextIO, Tuple, Union
 
 import pytest
 from _pytest.config import Config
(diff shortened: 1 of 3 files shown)


DanielNoord:
As to why this is needed: in the current form we don't enter `_config_initialization` when there is not `.rc` file. That doesn't work as I intend to use it to construct the `linter.namespace` attribute.
However, if we run `_config_initialization` without specifying an `.rc` file we will take the default values. The functional tests have not been written with those in mind, so that creates (a lot of) fails. I think this is the better solution.

pylint/testutils/lint_module_test.py
```diff
@@ -37,20 +40,39 @@ def __init__(
     ) -> None:
         _test_reporter = FunctionalTestReporter()
         self._linter = PyLinter()
-        self._linter.set_reporter(_test_reporter)
         self._linter.config.persistent = 0
         checkers.initialize(self._linter)
-        self._linter.disable("suppressed-message")
-        self._linter.disable("locally-disabled")
-        self._linter.disable("useless-suppression")
+
```
Pierre-Sassoulas, line 45:
```python
if "there is an rc file":
    rc_file = ...
else:
   rc_file = default
...
_config_initialization( # with less try except
```

Pierre-Sassoulas (changes requested):
I think a little refactor in the handling would make the code clearer.

Pierre-Sassoulas (approved):
I made suggestion with a commit directly because it was too hard to do in github interface, I'll let you check that before merging :)

DanielNoord:
Primer is failing after you're latest commit. Could you look into that?
Furthermore, I don't think that the variable should be source, but rather "arguments", but you should recheck that as I'm not 100% sure.

Pierre-Sassoulas:
> Primer is failing after you're latest commit. Could you look into that?

It seems there are a lot of connectivity issue on a lot of pipelines (timeout when recovering the primer repository but not only). I'll change the variable name :)

Pierre-Sassoulas (approved):

PR #5597, opened 2021-12-24
Add documentation about profiling and performance analysis

### LINKED ISSUE #4067: Create documentation for performance analysis on pylint
In my company we have a number of Python projects, and on one of them (220 KB, 4200 lines of code) `pylint` is tremendously slow to run, it takes about 30 minutes on a 4-core machine. The other projects are of comparable size, functionality and complexity, and yet `pylint` runs reasonably fast on them. All the projects are "clean" meaning the code is working fine and all pylint errors and warnings have been cleaned up.

My purpose here is to identify the aspects of my code that cause `pylint` too long to process, and modify my code so that the problem would go away. I tried different things, like excluding some imports, splitting files and classes in half, etc., but the root of the problem still eludes me. Profiler also is not of much help as it just shows that time gets consumed in some `pylint` internals.

What I'm looking for is some ideas what to check or how to identify the problem. Maybe there are some internal mechanisms in `pylint` that could be used to find out which parts of the code beings checked are taking a lot of time to check?

Thanks!

DanielNoord:

## Type of Changes

## Description

Closes #4067

While by no means perfect, this might be a good step towards better documentation on performance analysis. At least it shows users how to generate a useful profiler report.

### DIFF
--- a/doc/development_guide/profiling.rst
+++ b/doc/development_guide/profiling.rst
@@ -17,7 +17,7 @@ The documentation for both modules can be found at cProfile_.
 To profile a run of Pylint over itself you can use the following code and run it from the base directory.
 Note that ``cProfile`` will create a document called ``stats`` that is then read by ``pstats``. The
 human-readable output will be stored by ``pstats`` in ``./profiler_stats``. It will be sorted by
-``cumulative time``::
+``cumulative time``:
 
 .. sourcecode:: python
 
@@ -35,7 +35,7 @@ human-readable output will be stored by ``pstats`` in ``./profiler_stats``. It w
 
 You can also interact with the stats object by sorting or restricting the output.
 For example, to only print functions from the ``pylint`` module and sort by cumulative time you could
-use::
+use:
 
 .. sourcecode:: python
 
@@ -51,7 +51,7 @@ use::
         stats.sort_stats("cumtime")
         stats.print_stats("pylint/pylint")
 
-Lastly, to profile a run over your own module or code you can use::
+Lastly, to profile a run over your own module or code you can use:
 
 .. sourcecode:: python
 
--- a/doc/development_guide/profiling.rst
+++ b/doc/development_guide/profiling.rst
@@ -15,28 +15,33 @@ you need to profile a Pylint run and see which functions take how long to run.
 The documentation for both modules can be found at cProfile_.
 
 To profile a run of Pylint over itself you can use the following code and run it from the base directory.
-It will store the profiling data in ``./profiler_stats``::
+Note that ``cProfile`` will create a document called ``stats`` that is then read by ``pstats``. The
+human-readable output will be stored by ``pstats`` in ``./profiler_stats``. It will be sorted by
+``cumulative time``::
+
+.. sourcecode:: python
 
     import cProfile
     import pstats
     import sys
-    from pstats import SortKey
 
     sys.argv = ["pylint", "pylint"]
     cProfile.run("from pylint import __main__", "stats")
 
     with open("profiler_stats", "w", encoding="utf-8") as file:
         stats = pstats.Stats("stats", stream=file)
+        stats.sort_stats("cumtime")
         stats.print_stats()
 
 You can also interact with the stats object by sorting or restricting the output.
 For example, to only print functions from the ``pylint`` module and sort by cumulative time you could
 use::
 
+.. sourcecode:: python
+
     import cProfile
     import pstats
     import sys
-    from pstats import SortKey
 
     sys.argv = ["pylint", "pylint"]
     cProfile.run("from pylint import __main__", "stats")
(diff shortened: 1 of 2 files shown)


doc/development_guide/profiling.rst
```diff
@@ -0,0 +1,67 @@
...
+.. _profiling:
+
+===================================
+ Profiling and performance analysis
+===================================
+
+Performance analysis for Pylint
+-------------------------------
+
+To analysis the performance of Pylint we recommend to use the ``cProfile`` module
```
DanielNoord, line 11:
```suggestion
To analyse the performance of Pylint we recommend to use the ``cProfile`` module
```

DanielNoord:
@DudeNr33 I wrote a little something just so we have this information somewhere and can refer to it. Please let me know what you think of it!

doc/development_guide/profiling.rst
```diff
@@ -0,0 +1,67 @@
...
+    import pstats
+    import sys
+    from pstats import SortKey
+
+    sys.argv = ["pylint", "pylint"]
+    cProfile.run("from pylint import __main__", "stats")
+
+    with open("profiler_stats", "w", encoding="utf-8") as file:
+        stats = pstats.Stats("stats", stream=file)
+        stats.print_stats()
```
DudeNr33, line 30: Is there a call to ``stats.sort_stats()`` missing? 
You import ``SortKey`` but do not use it. It would make sense imo to sort by cumulative time.

I think the script should also delete the ``stats`` file that was created. Alternatively, the documentation should mention that it will be created, and can be loaded for interactive use if desired.

doc/development_guide/profiling.rst
```diff
@@ -0,0 +1,67 @@
...
+
+    sys.argv = ["pylint", "pylint"]
+    cProfile.run("from pylint import __main__", "stats")
+
+    with open("profiler_stats", "w", encoding="utf-8") as file:
+        stats = pstats.Stats("stats", stream=file)
+        stats.sort_stats("cumtime")
+        stats.print_stats("pylint/pylint")
+
+Lastly, the profile a run over your own module or code you can use::
```
DudeNr33, line 49:
```suggestion
Lastly, to profile a run over your own module or code you can use::
```

doc/development_guide/profiling.rst
DanielNoord: Ah yeah, I copied the script from the more advanced script and forgot to remove the import.

I'll add a comment about `stats` being created!

DudeNr33 (commented):
This is a great idea to document this for future reference!
Maybe a short summary what to look for in the results could be helpful.

I think besides the "general" profiling, a contributor might be interested in optimising a new checker they wrote. They are probably interested which method is called how many times, and what the most taxing portions of their code (both in total and cumulative time) are.

DudeNr33 (approved):
Looks good! Thank you for documenting this!

doc/development_guide/profiling.rst
```diff
@@ -0,0 +1,104 @@
...
+the profiling output.
+
+
+Performance analysis of a specific checker
+------------------------------------------
+
+To analyse the performance of specific checker within Pylint we can use the human-readable output
+created by ``pstats``.
+
+If you search in the ``profiler_stats`` file for the file name of the checker you will find all functional
```
Pierre-Sassoulas, line 80: Maybe we could also talk about some tools that interpret it like ``gprof2dot`` or ``snakeviz`` (you don't have to analyses the text output yourself even if you can) ?

Pierre-Sassoulas (commented):
I think only talking about the specific of pylint is good. We could add possible tools in the last part or remove the part where we explains how cprofile works altogether and link to [the documentation](https://docs.python.org/3/library/profile.html) instead. Performance inclined contributor are unlikely to need basic advises like this imo.

doc/development_guide/profiling.rst
DudeNr33: That's a good idea, but I would keep some explanation (or link to existing documentation) on how to interact, sort and filter the ``pstats`` file as well. From the discussion in #4814 we saw that the graph for ``pylint`` can become very large and hard to follow. If I were to analyse performance issues in a checker, I would probably work more with the ``Stats`` object than the graph.

Pierre-Sassoulas: Yes pre-sorting with pstat is really useful.

DanielNoord: @Pierre-Sassoulas The original issue was created by an user wondering why their program took so long too lint. That was what prompted me to include basic advice as well. Do you want me to remove this?

I'll create a section about interacting with `gprof2dot` and pre-sorting. I don't want this to be too long: if people want to learn how to profile they should look elsewhere, but this documentation should set them on the right track with regards to any `pylint` specifics.

DanielNoord: @DudeNr33 Did you figure out how to use `print_stats("pylint/pylint")` with `gprof2dot`? I don't think `print_stats` changes the `pstats` `stats` object itself, so I can't restrict the file passed to `gprof2dot`

DudeNr33: I don't know if this would make sense, as the diagram is a call graph and filtering would lead to incomplete or maybe even broken graphs. 
You could probably directly delete entries from the ``stats`` attribute and then use [``dump_stats()``](https://stackoverflow.com/questions/51536411/saving-cprofile-results-to-readable-external-file) to save it. I don't know however how the resulting output with ``gprof2dot`` looks, and if it is even able to process it properly if some nodes in the graph are missing.

DanielNoord: Hmm okay, so a short section pointing to `gprof2dot` would be enough?

DudeNr33: I think so. As Pierre pointed out ``snakeviz`` is an alternative visualisation tool ([this article](https://medium.com/@narenandu/profiling-and-visualization-tools-in-python-89a46f578989) provides a brief comparison of the tools) which provides a different representation together with filtering and sorting of the output, so I think this should be included, too.

Pierre-Sassoulas: > @Pierre-Sassoulas The original issue was created by an user wondering why their program took so long too lint. That was what prompted me to include basic advice as well. Do you want me to remove this?

> if people want to learn how to profile they should look elsewhere

Il agree let's keep the basic, but also let's keep it basic :) a link to popular tools' docs are enough imo.

Pierre-Sassoulas (approved):
Great work, the basis for pylint getting faster 👌

PR #5601, opened 2021-12-27
Fix reported node for `unnecessary-comprehension`

tushar-deepsource:
## Type of Changes

## Description

The issue location of `unnecessary-comprehension` was being raised at `line:0`, even if the comprehension wasn't present at column 0.

This would give the wrong location, for example:

```python
item = [(k, v) for k, v in a_dict.items()]
```

```text
************* Module mytest
mytest.py:1:0: R1721: Unnecessary use of a comprehension, use list(a_dict.items()) instead. (unnecessary-comprehension)
```

The expected position would be `1:7`, not `1:0`.

Weirdly, the test cases had zero examples of a comprehension that wasn't at column zero, so I tweaked some of those as well.

### DIFF
--- a/tests/functional/u/unnecessary/unnecessary_comprehension.py
+++ b/tests/functional/u/unnecessary/unnecessary_comprehension.py
@@ -14,14 +14,14 @@
 [(x, y, 1) for x, y in iterable]  # exclude useful comprehensions
 # Test case for issue #4499
 a_dict = {}
-[(k, v) for k, v in a_dict.items()]  # [unnecessary-comprehension]
+item = [(k, v) for k, v in a_dict.items()]  # [unnecessary-comprehension]
 
 # Set comprehensions
 {x for x in iterable}  # [unnecessary-comprehension]
 {y for x in iterable}  # expression != target_list
 {x for x,y,z in iterable}  # expression != target_list
 {(x,y,z) for x,y,z in iterable}  # [unnecessary-comprehension]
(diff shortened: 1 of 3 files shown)


tests/functional/u/unnecessary/unnecessary_comprehension.py
```diff
@@ -44,6 +44,6 @@
 my_set = set()
 
 [elem for elem in my_list]  # [unnecessary-comprehension]
-{k: v for k, v in my_dict.items()} # [unnecessary-comprehension]
+items = {k: v for k, v in my_dict.items()} # [unnecessary-comprehension]
 {k: my_dict[k] for k in my_dict} # [consider-using-dict-items]
```
Pierre-Sassoulas, line 48: Do you know why the node is not correct on this line ?

Pierre-Sassoulas (commented):
LGTM. Nice fix ;) I have a small remark but we could merge as is, it's already a great improvement. Github actions is not working well at the moment, don't worry about the pipeline failing.

tests/functional/u/unnecessary/unnecessary_comprehension.py
tusharsadhwani: there's a `Comprehension` node inside each `SetComp`, `DictComp` etc, in the `generators=` attribute. That one seems to have incorrect/incomplete location information.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #5602, opened 2021-12-27
Bump CI cache version

cdce8p:
## Description
Bump the cache version. Will fix the CI issues seen in #5600 where it could create the venv.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
   pull_request: ~
 
 env:
-  CACHE_VERSION: 3
+  CACHE_VERSION: 4
   DEFAULT_PYTHON: 3.8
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
You're lucky, no issue with github actions 😄

cdce8p:
> You're lucky, no issue with github actions 😄

Define lucky 😅
Took me a while to figure out that this was enough to fix it. Had the same issue on one of my personal repos.

DanielNoord:
@cdce8p For future reference, we also store `CACHE_VERSION` in the `primer` workflow. I'm adding a commit + comments that they should be changed in both places.

cdce8p:
Found a Python bug report that looks like the root cause for our issues: https://bugs.python.org/issue46197

DanielNoord:
@cdce8p Good catch! That indeed seems to eb the culprit.

PR #5605, opened 2021-12-28
Add caching to bottlenecks in the message store

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

This add caching to bottlenecks in the MessageStore. Next step would be to generate the code of an immutable data structure from the checkers instead of creating and checking it at runtime and then using this structure directly, but I think caching is fine for now. (Especially as we can't do everything this way: For plugin there are still unexpected message id and symbol.)

``get_active_msgids`` was calling .lower() / is_digit on every message id we gave it every time caching it save a lot of call.
There's some function we can't cache because of the way PyLinter is designed.

Tested by linting django (03cadb912c78b769d6bf4a943a2a35fc1d952960) with ``python3 -m cProfile -m pylint .pylint_primer_tests/django|grep message``

### Before:
```
  1494493    0.257    0.000    0.257    0.000 <frozen importlib._bootstrap>:222(_verbose_message)
        3    0.000    0.000    0.000    0.000 _enchant.py:120(find_message)
    67884    0.064    0.000    2.472    0.000 base_checker.py:112(add_message)
      678    0.001    0.000    0.003    0.000 base_checker.py:147(create_message_definition_from_tuple)
       86    0.000    0.000    0.004    0.000 base_checker.py:171(messages)
        1    0.000    0.000    0.000    0.000 base_reporter.py:69(display_messages)
    41158    0.014    0.000    0.014    0.000 file_state.py:132(handle_ignored_message)
     2722    0.007    0.000    0.008    0.000 file_state.py:148(iter_spurious_suppression_messages)
    26728    0.021    0.000    0.032    0.000 linterstats.py:288(increase_single_message_count)
    26728    0.020    0.000    0.020    0.000 linterstats.py:292(increase_single_module_message_count)
        1    0.000    0.000    0.000    0.000 linterstats.py:298(reset_message_count)
        1    0.000    0.000    0.000    0.000 message.py:34(Message)
        1    0.000    0.000    0.001    0.001 message.py:5(<module>)
    26728    0.056    0.000    0.080    0.000 message.py:60(__new__)
        1    0.000    0.000    0.000    0.000 message_definition.py:17(MessageDefinition)
      678    0.000    0.000    0.001    0.000 message_definition.py:18(__init__)
        1    0.000    0.000    0.000    0.000 message_definition.py:4(<module>)
[...]

PR #5606, opened 2021-12-28
Get the msgids directly from the MessageIdStore

Pierre-Sassoulas:
## Type of Changes

## Description

We don't need to create message definitions to get their msgids.

The level of indirection may suggest we need to refactor a
little here. Maybe we actually need message definition only
when we want to display the messages.

Follow-up to #5605

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -1435,8 +1435,7 @@ class PyLinter(
             if confidence.name not in self.config.confidence:
                 return False
         try:
-            message_definitions = self.msgs_store.get_message_definitions(msg_descr)
-            msgids = [md.msgid for md in message_definitions]
+            msgids = self.msgs_store.message_id_store.get_active_msgids(msg_descr)
         except exceptions.UnknownMessageError:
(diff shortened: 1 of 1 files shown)


DanielNoord (commented):
Is there a difference between the message definitions and the "active" message definitions?

Pierre-Sassoulas:
I'll let you judge, here's the implementation of ``get_message_definitions`` 😄 
```python
    def get_message_definitions(self, msgid_or_symbol: str) -> List[MessageDefinition]:
        """Returns the Message definition for either a numeric or symbolic id."""
        return [
            self._messages_definitions[m]
            for m in self.message_id_store.get_active_msgids(msgid_or_symbol)
        ]
```        
 I think we discussed typing msgid at some point as msgid are string with a letter and 4 digits and decided against it. This might be a big refactor but using msgid everywhere instead of message definition might be a good idea. We could even transform symbol to msgid early on when it's user input.

DanielNoord:
😄 

The refactor should probably be left for a later point, but this is indeed a good fix!

DanielNoord (approved):

PR #5609, opened 2021-12-29
Use ``dill`` to pickle when run in parallel mode

DanielNoord:

## Type of Changes

## Description

For reference see my earlier comment here:
https://github.com/PyCQA/pylint/pull/5584#issuecomment-1002470174

`multiprocessing.Pool` pickles the arguments passed in `initargs`. This was noted in the original commit that added this: https://github.com/PyCQA/pylint/commit/8babeffde54a696e7ba84d69f8fc1f871e432950
One of the effects of this was that it was decided to set `reporter` to `None` as then the reporter doesn't need to be pickled. We're running into a similar issue with the migration to `argparse` since `argparse.ArgumentsParser` isn't pickle-able by `stdlib`.

To overcome this I looked at the solution proposed on StackOverflow here:
https://stackoverflow.com/questions/8804830/python-multiprocessing-picklingerror-cant-pickle-type-function

The `dill` package is able to pickle objects such as `Argumentsparser`. Therefore, by pickling the `linter` object with `dill` before passing to `multiprocessing.Pool` we can avoid the error ran into with the `stdlib`.
As @Pierre-Sassoulas said in https://github.com/PyCQA/pylint/issues/5584#issuecomment-1002476948_:

> Look like ``dill`` do not have any dependencies of its own and its 3-clause BSD license seems compatible with pylint. We would more than double their download per day by adding them as a dependencies but it's already used a lot. 

@cdce8p Is the way I added `dill` to the `mypy` config correct?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    dill>=0.3.4
+    dill>=0.2
     platformdirs>=2.2.0
     astroid>=2.9.0,<2.10 # (You should also upgrade requirements_test_min.txt)
     isort>=4.2.5,<6
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,6 +43,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
+    dill>=0.3.4
     platformdirs>=2.2.0
     astroid>=2.9.0,<2.10 # (You should also upgrade requirements_test_min.txt)
     isort>=4.2.5,<6
@@ -122,4 +123,7 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 
 [mypy-git.*]
+ignore_missing_imports = True
+
+[mypy-dill]
 ignore_missing_imports = True
--- a/tests/test_check_parallel.py
+++ b/tests/test_check_parallel.py
@@ -11,9 +11,12 @@
 
 # pylint: disable=protected-access,missing-function-docstring,no-self-use
 
+import argparse
+import multiprocessing
 import os
 from typing import List
 
+import dill
 import pytest
 from astroid import nodes
 
@@ -174,8 +177,22 @@ class TestCheckParallelFramework:
 
     def test_worker_initialize(self) -> None:
         linter = PyLinter(reporter=Reporter())
-        worker_initialize(linter=linter)
-        assert pylint.lint.parallel._worker_linter == linter
+        worker_initialize(linter=dill.dumps(linter))
+        assert isinstance(pylint.lint.parallel._worker_linter, type(linter))
+
+    def test_worker_initialize_pickling(self) -> None:
+        """Test that we can pickle objects that standard pickling in multiprocessing can't.
+
+        See:
+        https://stackoverflow.com/questions/8804830/python-multiprocessing-picklingerror-cant-pickle-type-function
+        https://github.com/PyCQA/pylint/pull/5584
+        """
+        linter = PyLinter(reporter=Reporter())
+        linter.attribute = argparse.ArgumentParser()  # type: ignore[attr-defined]
+        pool = multiprocessing.Pool(  # pylint: disable=consider-using-with
+            2, initializer=worker_initialize, initargs=[dill.dumps(linter)]
+        )
+        pool.imap_unordered(print, [1, 2])
 
     def test_worker_check_single_file_uninitialised(self) -> None:
         pylint.lint.parallel._worker_linter = None
@@ -186,7 +203,7 @@ class TestCheckParallelFramework:
 
     def test_worker_check_single_file_no_checkers(self) -> None:
         linter = PyLinter(reporter=Reporter())
-        worker_initialize(linter=linter)
+        worker_initialize(linter=dill.dumps(linter))
 
         (
             _,  # proc-id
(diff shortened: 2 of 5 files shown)


DanielNoord:
@Pierre-Sassoulas Do you know why the `primer` is failing to download dependencies?

tests/test_check_parallel.py
```diff
@@ -174,8 +177,22 @@ def teardown_class(self):
...
+    def test_worker_initialize_pickling(self) -> None:
+        """Test that we can pickle objects that standard pickling in multiprocessing can't.
+
+        See:
+        https://stackoverflow.com/questions/8804830/python-multiprocessing-picklingerror-cant-pickle-type-function
+        https://github.com/PyCQA/pylint/pull/5584
+        """
+        linter = PyLinter(reporter=Reporter())
+        linter.attribute = argparse.ArgumentParser()  # type: ignore[attr-defined]
+        pool = multiprocessing.Pool(  # pylint: disable=consider-using-with
```
Pierre-Sassoulas, line 192: You don't want to use `with` ? :)

tests/test_check_parallel.py
```diff
@@ -174,8 +177,22 @@ def teardown_class(self):
...
+            2, initializer=worker_initialize, initargs=[dill.dumps(linter)]
+        )
+        pool.imap_unordered(print, [1, 2])
```
Pierre-Sassoulas, line 195: Shouldn't we assert something in this test ?

Pierre-Sassoulas (changes requested):
LGTM, thank you for the improved typing ! (Every typing is magic :smile: )

tests/test_check_parallel.py
DanielNoord: This test will fail without using `dill` as `argparse.ArgumentParser` won't be able to be pickled. This is a bit of a silly test as the test uses `dill.dumps` itself, but should serve as a guard against anybody trying to remove `dill` as a dependency in the future without thinking of the impact on "pickle-bility".

tests/test_check_parallel.py
DanielNoord: This was just directly copied from `parallel.py`. We can change, but I think by keeping it similar it shows better what we are testing here.

DanielNoord: Actually I'll do a follow-up PR where I'll add a `with` to `parallel.py` and this test (if this gets merged).

Pierre-Sassoulas (approved):

DanielNoord:
> @Pierre-Sassoulas Do you know why the `primer` is failing to download dependencies?

@Pierre-Sassoulas I think you might have missed this. Any idea?

Pierre-Sassoulas:
> Do you know why the primer is failing to download dependencies?

I think there's still some instability with github actions right now.

DanielNoord:
> > Do you know why the primer is failing to download dependencies?
> 
> I think there's still some instability with github actions right now.

It's weird as this only seems to trigger on this PR. The `CI` workflow uses the same commands (I think) so I don't understand why this keeps failing (I've re-run the action 4 times now)... I think we should investigate before merging.

Pierre-Sassoulas:
> I've re-run the action 4 times now

Hmm, this is strange, you're right. Maybe the new dependency force some cache to be regenerated and this is what fails ?

setup.cfg
```diff
@@ -43,6 +43,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
+    dill>=0.3.4
```
cdce8p, line 46: Does it need to be version `0.3.4` or would it work with an older version, too?
If we can, we should try to keep the requirements as open as possible.

setup.cfg
```diff
@@ -123,3 +124,6 @@ ignore_missing_imports = True
 
 [mypy-git.*]
 ignore_missing_imports = True
+
+[mypy-dill]
+ignore_missing_imports = True
```
cdce8p, line 129: That looks about right

setup.cfg
DanielNoord: I wasn't sure about this.

I'll go back and see if this works on other versions as well.

DanielNoord: https://pypi.org/project/dill/#history

0.2b1 doesn't work, 0.2 does. Updated!

cdce8p: 👍🏻 I couldn't find a good changelog which does make it difficult to know what exactly changed. `0.2` should be broad enough though.

cdce8p:
This is not really an area I know too much about. I trust you guys there.

setup.cfg
DanielNoord: I think they were missing a dependency, or at least an error popped up that from a google search seemed related to incorrect dependency issues in other projects as well.
Based on their release history `0.2` seems to be the first "full" version.

DanielNoord:
> This is not really an area I know too much about. I trust you guys there.

@Pierre-Sassoulas I think it makes sense to put this in `2.13` and wait with merging any PRs that depend on it. That way if we run into conflicting dependency issues for other tools we will know about them before we're totally depending on it ourselves.
The current change is easily reversible.

setup.cfg
cdce8p: > Based on their release history `0.2` seems to be the first "full" version.

I agree. Although a broad dependency range is better, we don't need to encourage old beta releases. No need to go there.

PR #5612, opened 2021-12-30
Use ``with`` statement in ``parallel.py``

DanielNoord:

## Type of Changes

## Description

Follow up to #5609.

Not sure why this was not done in the original PR, but this can be a `with` statement.

### DIFF
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -133,11 +133,10 @@ def check_parallel(
     # is identical to the linter object here. This is required so that
     # a custom PyLinter object can be used.
     initializer = functools.partial(_worker_initialize, arguments=arguments)
-    pool = multiprocessing.Pool(  # pylint: disable=consider-using-with
+    with multiprocessing.Pool(
         jobs, initializer=initializer, initargs=[dill.dumps(linter)]
-    )
-    linter.open()
-    try:
+    ) as pool:
+        linter.open()
         all_stats = []
         all_mapreduce_data = collections.defaultdict(list)
 
@@ -164,9 +163,6 @@ def check_parallel(
             all_stats.append(stats)
             all_mapreduce_data[worker_idx].append(mapreduce_data)
             linter.msg_status |= msg_status
-    finally:
-        pool.close()
-        pool.join()
 
     _merge_mapreduce_data(linter, all_mapreduce_data)
     linter.stats = merge_stats([linter.stats] + all_stats)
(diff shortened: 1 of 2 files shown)


pylint/lint/parallel.py
```diff
@@ -164,9 +163,6 @@ def check_parallel(
             all_stats.append(stats)
             all_mapreduce_data[worker_idx].append(mapreduce_data)
             linter.msg_status |= msg_status
-    finally:
-        pool.close()
-        pool.join()
```
Pierre-Sassoulas, line 169: I'm wondering if the join is important. If could help to have more contexte in case of exception in the thread. Maybe it was the reason we could not use the with ? It seems ``Pool.join`` is not called in ``__exit__`` : https://github.com/python/cpython/blob/1b37268ef10bd20c30d349b8401c88215c8a6be8/Lib/multiprocessing/pool.py#L651.

Pierre-Sassoulas (commented):
Maybe we found a false positive for ``consider-using-with``; let me know what you think :)

pylint/lint/parallel.py
DanielNoord: Are you sure? I haven't looked into it fully but I see some `join()` calls in `self._terminate_pool`.

@DudeNr33 Was there a reason for skipping this statement in https://github.com/PyCQA/pylint/commit/922f38969c326826344e2d25af499cf8c5f80d8c?

DudeNr33: I can't really recall, maybe I was just not confident enough to make change to this file. 

It is not entirely clear to me if there could be instances where you would have to call ``pool.join()`` explicitly.
But this here should not be one of them. While ``poo.terminate()`` [terminates the processes without completing outstanding work](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.terminate), ``imap_unordered`` [blocks until the result is ready](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.terminate).
In fact only the [async methods](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.map_async) in the ``Pool`` class seem to not block automatically, and ``consider-using-with`` does not apply to ``async with`` blocks which would then be necessary.

I guess this construct here was from the pre Python 3.3 days, where the ``Pool`` class did not support the context manager protocol yet.
Also, all examples given in the official documentation use the Pool in a ``with`` statement.

So, to summarise: I'd say the ``pool.join()`` is not necessary and this is not a false positive.

Pierre-Sassoulas (approved):

PR #5614, opened 2021-12-30
Fix `useless-else-on-loop` false positive

### LINKED ISSUE #5598: ``useless-else-on-loop`` should not be raised on `return`
### Bug description

Here's a typical example of W0120 being raised:

```python
def f():
    while True:
        if condition:
            pass
    else:
        pass
```

And this code **doesn't** raise W0120 (correctly):

```python
def f():
    while True:
        if condition:
            break
    else:
        pass
```

But, this code is pretty much identical in function and it **does** raise W0120 (bug):

```python
def f():
    while True:
        if condition:
            return
    else:
        pass
```

The check for W0120 should check for both `break` and `return` in the while block.

### Configuration

_No response_

### Command used

```shell
pylint mytest.py
```

### Pylint output

```shell
************* Module mytest
mytest.py:5:4: W0120: Else clause on loop without a break statement (useless-else-on-loop)
```

### Expected behavior

Issue not raised

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.6-dev0
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

tushar-deepsource:
## Type of Changes

## Description

Closes #5598

### DIFF
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -298,16 +298,17 @@ def _get_break_loop_node(break_node):
 
 def _loop_exits_early(loop):
     """
-    Returns true if a loop may end with a break statement.
+    Returns true if a loop may end with a break or return statement.
 
     Args:
         loop (astroid.For, astroid.While): the loop node inspected.
 
     Returns:
-        bool: True if the loop may end with a break statement, False otherwise.
+        bool: True if the loop may end with a break or return statement, False otherwise.
     """
     loop_nodes = (nodes.For, nodes.While)
     definition_nodes = (nodes.FunctionDef, nodes.ClassDef)
+    break_nodes = (nodes.Break, nodes.Return)
     inner_loop_nodes = [
         _node
         for _node in loop.nodes_of_class(loop_nodes, skip_klass=definition_nodes)
@@ -315,7 +316,7 @@ def _loop_exits_early(loop):
     ]
     return any(
         _node
-        for _node in loop.nodes_of_class(nodes.Break, skip_klass=definition_nodes)
+        for _node in loop.nodes_of_class(break_nodes, skip_klass=definition_nodes)
         if _get_break_loop_node(_node) not in inner_loop_nodes
     )
 
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -276,7 +276,7 @@ def in_nested_list(nested_list, obj):
 
 def _get_break_loop_node(break_node):
     """
-    Returns the loop node that holds the break node in arguments.
+    Returns the loop node that holds the break or return node in arguments.
 
     Args:
         break_node (astroid.Break): the break node of interest.
@@ -318,8 +318,18 @@ def _loop_exits_early(loop):
         _node
         for _node in loop.nodes_of_class(break_nodes, skip_klass=definition_nodes)
         if _get_break_loop_node(_node) not in inner_loop_nodes
+        and not _node_in_orelse(loop, _node)
     )
 
+
+def _node_in_orelse(loop, node):
+    for child in getattr(loop, "orelse", []):
+        if child == node:
+            return True
+        if any(kid == node for kid in child.get_children()):
+            return True
+    
+    return False
 
 def _is_multi_naming_match(match, node_type, confidence):
     return (
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -328,8 +328,9 @@ def _node_in_orelse(loop, node):
             return True
         if any(kid == node for kid in child.get_children()):
             return True
-    
+
     return False
+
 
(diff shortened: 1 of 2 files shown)


tushar-deepsource:
There's just one test case that is failing right now. It's when there's a `break` statement present in the `else` clause of a nested for-loop. I'm not taking that into account, and right now I can't think of how I would. Feel free to suggest ideas.

pylint/checkers/base.py
```diff
@@ -276,7 +276,7 @@ def in_nested_list(nested_list, obj):
 
 def _get_break_loop_node(break_node):
```
DanielNoord, line 277: Can the code that uses this be changed to use `checkers.utils.get_node_first_ancestor_of_type`? 
This function is basically doing the same I believe.

pylint/checkers/base.py
```diff
@@ -298,28 +298,40 @@ def _get_break_loop_node(break_node):
 
 def _loop_exits_early(loop):
```
DanielNoord, line 299:
```suggestion
def _loop_exits_early(loop: Union[nodes.While, nodes.For]) -> bool:
```

pylint/checkers/base.py
```diff
@@ -298,28 +298,40 @@ def _get_break_loop_node(break_node):
...
+    Returns true if a loop may end with a break or return statement.
 
     Args:
         loop (astroid.For, astroid.While): the loop node inspected.
 
     Returns:
-        bool: True if the loop may end with a break statement, False otherwise.
+        bool: True if the loop may end with a break or return statement, False otherwise.
     """
     loop_nodes = (nodes.For, nodes.While)
```
DanielNoord, line 309: Personally I think this whole function could do with a rewrite. Using a `for` loop with premature returns (and potentially breaks) would improve readability and could probably remove unnecessary assignments such as `definition_nodes` and `break_nodes`.

pylint/checkers/base.py
```diff
@@ -298,28 +298,40 @@ def _get_break_loop_node(break_node):
...
     return any(
         _node
-        for _node in loop.nodes_of_class(nodes.Break, skip_klass=definition_nodes)
+        for _node in loop.nodes_of_class(break_nodes, skip_klass=definition_nodes)
         if _get_break_loop_node(_node) not in inner_loop_nodes
+        and not _node_in_orelse(loop, _node)
     )
 
 
+def _node_in_orelse(loop, node):
```
DanielNoord, line 325:
```suggestion
def _node_in_orelse(loop:Union[nodes.While, nodes.For], node: Union[nodes.Break, nodes.Return]) -> bool:
```

Perhaps even add a docstring? I like those even for internal methods, but I know opinions differ.

DanielNoord (changes requested):
As for the breaking test: can't you iterate over `nodes.Else` children and call the same method recursively?

pylint/checkers/base.py
Pierre-Sassoulas: I'm also for docstring if there is more information than in the function name and argument names themselves. Here if the docstring was """Is the node in an or else inside the given loop ?""" I think it's the same info that function name and arg name, only duplicated.

jacobtylerwalls:
I'm not very positive on this change, see my comment on the issue. If this goes forward, I think the description of the error will need updating.

tushar-deepsource:
I'll work on a rewrite if I get a green signal that it would go in. IMO, pylint raising this issue on a code snippet where it shouldn't is more than enough of a reason to fix it.

jacobtylerwalls:
Hey sorry if my other message sounded short! I'm glad you're raising this for discussion. I'm just saying I'm not seeing the bug: I think Pylint is raising a correct style nitpick here. And a long while back Pylint devs of yesteryear had this same discussion and decided on the current behavior, so I'm not seeing why to change it? 🤔

tusharsadhwani:
@jacobtylerwalls yeah, I would agree if it were  a style nitpick, but right now it's not really that, right?

The short code is `useless-else-on-loop`, which to me implies the else-block is essentially dead code and should be removed. That makes the `return` edge case important.

Any way to mitigate that would be alright. Either changing the message/shortcode, or adding return support. Otherwise it feels incomplete to me.

tusharsadhwani:
oh my bad. you're totally right, I was misinterpreting the two cases.

Now with that clarity, I think there isn't a false positive at all

Pierre-Sassoulas:
I originally had the same problem of misunderstanding this message myself. I saw question on stackoverflow with the same misunderstanding. The fact that MR get opened to "fix" it is an even stronger indicator that the message is hard to understand and I think a clearer message would help everyone. What about "remove the else and de-indent all the code inside it" ?

tushar-deepsource:
I'll close the PR and we can continue discussion in a single thread: #5598

PR #5615, opened 2021-12-30
Add test for exclusivity of id names

### LINKED ISSUE #5255: Create test for exclusivity of message id prefixes
I think the prefixes have an important aim: to make sure that a checker can evolve to new message types with the security that the corresponding ids are not used yet.

Take a checker that only emits F3101 and F3102. At some point we might think of a new message for that checker, which does not go at the Fatal level. We would then want to emit E3101. If we do not "reserve" the "31" prefix for this checker it might actually be the case that another checker is already issuing E3101 and E3102. Thus the first checker would then need to emit F3101, F3102 and E3201.
Although this might not create immediate inconsistenties, illogical numbering of messages like this will (probably) create problems in the future (just like we are seeing now). In any case it will make maintaining `pylint` more "illogical" and thus (probably) more difficult.

I can understand that you do not want to change the id for any of the current messages, but I would really argue against removing the restriction all together. I think I'll be able to create a test that sees if a checker has a unique message id prefix, while ignoring the 4/5 prefixes that are currently non-exclusive. I think that would be the best way forward here!

_Originally posted by @DanielNoord in https://github.com/PyCQA/pylint/issues/5248#issuecomment-959866640_

Further discussion in that PR.

DanielNoord:

## Type of Changes

## Description

Closes #5255 

This is a very quick and dirty approach, but I guess it works? Too bad that I see some recent additions in here that could have been avoided if this test had been in place earlier 😢

### DIFF
--- a/tests/message/unittest_message_id_store.py
+++ b/tests/message/unittest_message_id_store.py
@@ -2,8 +2,6 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 from pathlib import Path
-
-EMPTY_FILE = str(Path(__file__).parent.parent.resolve() / "regrtest_data" / "empty.py")
 from typing import Dict, ValuesView
 
 import pytest
@@ -13,7 +11,7 @@ from pylint.exceptions import InvalidMessageError, UnknownMessageError
 from pylint.message.message_definition import MessageDefinition
 from pylint.message.message_id_store import MessageIdStore
 
-HERE = abspath(dirname(__file__))
+EMPTY_FILE = str(Path(__file__).parent.parent.resolve() / "regrtest_data" / "empty.py")
 
 
 def test_len_str(msgid_store: MessageIdStore, msgids: Dict[str, str]) -> None:
--- a/tests/message/unittest_message_id_store.py
+++ b/tests/message/unittest_message_id_store.py
@@ -1,13 +1,17 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
+from os.path import abspath, dirname, join
 from typing import Dict, ValuesView
 
 import pytest
 
+from pylint import lint
 from pylint.exceptions import InvalidMessageError, UnknownMessageError
 from pylint.message.message_definition import MessageDefinition
 from pylint.message.message_id_store import MessageIdStore
+
+HERE = abspath(dirname(__file__))
 
 
 def test_len_str(msgid_store: MessageIdStore, msgids: Dict[str, str]) -> None:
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
Shouldn't this be a script instead so we can use it with pre-commit later ?

DanielNoord:
Not sure? This will still fail the CI if checkers are not exclusive.

tests/message/unittest_message_id_store.py
```diff
@@ -90,3 +94,47 @@ def test_duplicate_msgid(msgid_store: MessageIdStore) -> None:
         "Message symbol 'warning-symbol' cannot be used for 'W1234' and 'W1235'"
         in str(error.value)
     )
+
+
+def test_exclusivity_of_msg_ids() -> None:
```
Pierre-Sassoulas, line 99:
```suggestion
def test_exclusivity_of_msgids() -> None:
```
This is the convention in ``pylint.messages``.

tests/message/unittest_message_id_store.py
```diff
@@ -90,3 +94,47 @@ def test_duplicate_msgid(msgid_store: MessageIdStore) -> None:
...
+    """Test to see if all checkers have an exclusive message id prefix"""
+    err_msg = (
+        "{} has the same prefix ('{}') as the '{}' checker. Please make sure the prefix "
+        + "is unique for each checker. You can use 'script/get_unused_message_id_category.py' "
+        + "to get an unique id."
```
Pierre-Sassoulas, line 104:
```suggestion
        "{} has the same prefix ('{}') as the '{}' checker. Please make sure the prefix "
        "is unique for each checker. You can use 'script/get_unused_message_id_category.py' "
        "to get an unique id."
```
Implicit concatenation of string would work.

tests/message/unittest_message_id_store.py
```diff
@@ -1,14 +1,18 @@
...
 from typing import Dict, ValuesView
 
 import pytest
 
+from pylint import lint
 from pylint.exceptions import InvalidMessageError, UnknownMessageError
 from pylint.message.message_definition import MessageDefinition
 from pylint.message.message_id_store import MessageIdStore
 
+HERE = abspath(dirname(__file__))
```
Pierre-Sassoulas, line 14:
```suggestion
EMPTY_FILE = join(abspath(dirname(__file__)),  "..", "regrtest_data", "empty.py")
```

tests/message/unittest_message_id_store.py
```diff
@@ -90,3 +94,47 @@ def test_duplicate_msgid(msgid_store: MessageIdStore) -> None:
...
+def test_exclusivity_of_msg_ids() -> None:
+    """Test to see if all checkers have an exclusive message id prefix"""
+    err_msg = (
+        "{} has the same prefix ('{}') as the '{}' checker. Please make sure the prefix "
+        + "is unique for each checker. You can use 'script/get_unused_message_id_category.py' "
+        + "to get an unique id."
+    )
+
+    runner = lint.Run(
+        ["--enable-all-extensions", join(HERE, "..", "regrtest_data", "empty.py")],
```
Pierre-Sassoulas, line 108:
```suggestion
        ["--enable-all-extensions", EMPTY_FILE],
```

tests/message/unittest_message_id_store.py
```diff
@@ -90,3 +94,47 @@ def test_duplicate_msgid(msgid_store: MessageIdStore) -> None:
...
+        exit=False,
+    )
+
+    # Some pairs are hard-coded as they are pre-existing and non-exclusive
+    # and we (currently) don't want to rename them
```
Pierre-Sassoulas, line 113:
```suggestion
    # Some pairs are hard-coded as they are pre-existing and non-exclusive
    # and we don't want to rename them and make old msgid doc obsoletes
```
I think this will never be worth it. It would make sooooo much old names and break every question asking about a specific msgid on stackoverflow.

tests/message/unittest_message_id_store.py
```diff
@@ -90,3 +94,47 @@ def test_duplicate_msgid(msgid_store: MessageIdStore) -> None:
...
+        "20": ("compare-to-zero", "refactoring"),
+    }
+
+    for msg_id, definition in runner.linter.msgs_store._messages_definitions.items():
+        if msg_id[1:3] in checker_id_pairs:
+            assert (
+                definition.checker_name in checker_id_pairs[msg_id[1:3]]
+            ), err_msg.format(msg_id, msg_id[1:3], checker_id_pairs[msg_id[1:3]][0])
+        else:
+            checker_id_pairs[msg_id[1:3]] = (definition.checker_name,)
```
Pierre-Sassoulas, line 140:
```suggestion
    for msgid, definition in runner.linter.msgs_store._messages_definitions.items():
        if msgid[1:3] in checker_id_pairs:
            assert (
                definition.checker_name in checker_id_pairs[msgid[1:3]]
            ), err_msg.format(msgid, msgid[1:3], checker_id_pairs[msgid[1:3]][0])
        else:
            checker_id_pairs[msgid[1:3]] = (definition.checker_name,)
```

tests/message/unittest_message_id_store.py
```diff
@@ -1,14 +1,18 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
+from os.path import abspath, dirname, join
```
Pierre-Sassoulas, line 4:
```suggestion
from pathlib import Path

EMPTY_FILE = str(Path(__file__).parent.parent.resolve() / "regrtest_data" / "empty.py")
```
I don't want to be ideological about but pathlib.Path is a really nice class 😛

Pierre-Sassoulas (changes requested):
Fair enough, there's no need for pre-commit as can wait for a pipeline to run. And code inside script is not properly tested and would need to redo the whole initialization we already have for the tests' pylinter.

Pierre-Sassoulas (approved):

PR #5618, opened 2021-12-31
Rename CI job names

DanielNoord:

## Type of Changes

## Description

Taking inspiration from `pip` as seen here https://github.com/pypa/pip/pull/10720 I really like their clean CI jobs names.
This is probably personal preference, but would we consider changing them? This is my suggestion based on their design.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -227,7 +227,7 @@ jobs:
           path: .coverage
 
   coverage:
-    name: tests / process coverage
+    name: tests / process / coverage
     runs-on: ubuntu-latest
     needs: ["prepare-tests-linux", "pytest-linux"]
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -74,7 +74,7 @@ jobs:
           pre-commit install --install-hooks
 
   formatting:
-    name: Run pre-commit checks
+    name: checks / pre-commit
     runs-on: ubuntu-latest
     needs: prepare-base
     steps:
@@ -116,7 +116,7 @@ jobs:
           pre-commit run pylint --all-files
 
   spelling:
-    name: Run spelling checks
+    name: checks / spelling
     runs-on: ubuntu-latest
     needs: prepare-base
     steps:
@@ -146,7 +146,7 @@ jobs:
           pytest tests/ -k unittest_spelling
 
   prepare-tests-linux:
-    name: Prepare tests for Python ${{ matrix.python-version }} (Linux)
+    name: tests / prepare / ${{ matrix.python-version }} / Linux
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -188,7 +188,7 @@ jobs:
           pip install -U -r requirements_test.txt
 
   pytest-linux:
-    name: Run tests Python ${{ matrix.python-version }} (Linux)
+    name: tests / run / ${{ matrix.python-version }} / Linux
     runs-on: ubuntu-latest
     needs: prepare-tests-linux
     strategy:
@@ -227,7 +227,7 @@ jobs:
           path: .coverage
 
   coverage:
-    name: Process test coverage
+    name: tests / process coverage
     runs-on: ubuntu-latest
     needs: ["prepare-tests-linux", "pytest-linux"]
     strategy:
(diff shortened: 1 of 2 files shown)


DanielNoord:
We probably need to change some setting about which actions are required for PRs to mergeable for this to work.

cdce8p (commented):
Tbh at first glance I though this looked quite good, but the CI test summery is much more readable with the old style. Could be the multiple slashes that make it more difficult to read.

I would suggest to keep the old style.

.github/workflows/ci.yaml
```diff
@@ -227,7 +227,7 @@ jobs:
           path: .coverage
 
   coverage:
-    name: Process test coverage
+    name: tests / process coverage
```
Pierre-Sassoulas, line 230:
```suggestion
    name: tests / process / coverage
```

.github/workflows/primer-test.yaml
```diff
@@ -22,7 +22,7 @@ concurrency:
 
 jobs:
   prepare-tests-linux:
-    name: Prepare tests for Python ${{ matrix.python-version }} (Linux)
+    name: prepare / ${{ matrix.python-version }} / Linux
```
Pierre-Sassoulas, line 25:
```suggestion
    name: primer / prepare / ${{ matrix.python-version }} / Linux
```

.github/workflows/primer-test.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
           pip install -U -r requirements_test.txt
 
   pytest-primer-stdlib:
-    name: Run primer tests on stdlib Python ${{ matrix.python-version }} (Linux)
+    name: run on stdlib / ${{ matrix.python-version }} / Linux
```
Pierre-Sassoulas, line 67:
```suggestion
    name: primer / stdlib / ${{ matrix.python-version }} / Linux
```

.github/workflows/primer-test.yaml
```diff
@@ -98,9 +98,7 @@ jobs:
           pytest -m primer_stdlib --primer-stdlib -n auto
 
   pytest-primer-external-batch-one:
-    name:
-      Run primer tests batch one on external libs Python ${{ matrix.python-version }}
-      (Linux)
+    name: run on batch one / ${{ matrix.python-version }} / Linux
```
Pierre-Sassoulas, line 101:
```suggestion
    name: primer / external batch one / ${{ matrix.python-version }} / Linux
```

.github/workflows/primer-test.yaml
```diff
@@ -134,9 +132,7 @@ jobs:
           pytest -m primer_external_batch_one --primer-external -n auto
 
   pytest-primer-external-batch-two:
-    name:
-      Run primer tests batch two on external libs Python ${{ matrix.python-version }}
-      (Linux)
+    name: run on batch two / ${{ matrix.python-version }} / Linux
```
Pierre-Sassoulas, line 135:
```suggestion
    name: primer / external batch two / ${{ matrix.python-version }} / Linux
```

Pierre-Sassoulas (changes requested):
LGTM ! Some suggestions

.github/workflows/primer-test.yaml
DanielNoord: Primer is already included in the workflow name, see the names in the checks of this PR

.github/workflows/primer-test.yaml
DanielNoord: See above

.github/workflows/primer-test.yaml
DanielNoord: See above

Pierre-Sassoulas (approved):
LGTM, but I think it requires that I change the settings of the repositories for required jobs, so it might be a mess during the time where both old job exists on out of date branch and new jobs on up to date branch. Maybe it's time to remove required job because all mainteners are responsible individual that will only merge responsibly anyway ?

cdce8p:
> Maybe it's time to remove required job because all maintainers are responsible individual that will only merge responsibly anyway ?

I think that should be fine. That had been the default a while back anyway.

> (Also if Marc think the old one are better and we still require some job to pass, it would creates work for an opinionated change and we may choose to close)

Just my opinion, in the PR view I definitely think the old style looked better. The only advantage I see is in the actions side panel, but just ever so slightly. I mentioned it earlier already, it could be the multiple slashes for each entry which make it much more difficult to read. Overall I'm somewhere between -0.5 and -1 on this one. However if you both think it's a definite improvement, I guess I could live with it too.

Pierre-Sassoulas:
I removed the required job and I think this is making the test name easier to parse by giving predictability. As this is not a very important change let's go with the majority opinion here.

Issue #2955, opened 2019-06-12
Unable to import 'distutils.util'


crazyscientist:
There are quite a few related issues (e.g. #1691), but none mention this particular issue. Somehow `distutils.util` cannot be processed by pylint.

### Steps to reproduce
1. Import something from `distutils.util`
2. Run `pylint`

### Current behavior
Pylint produces two errors for the import statement:

```
# In the file
from distutils.util import strtobool
```

```
# In the output
************* Module [...].run_agents
[...]run_agents.py:33:0: E0611: No name 'util' in module 'distutils' (no-name-in-module)
[...]run_agents.py:33:0: E0401: Unable to import 'distutils.util' (import-error)
```

### Expected behavior
None of the above error messages

### pylint --version output
```
pylint 2.3.1
astroid 2.2.5
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0]
```

crazyscientist:
Looks like this might be a duplicate of #73...

Pierre-Sassoulas:
If it is, it has been fixed 21 days ago in https://github.com/PyCQA/astroid/pull/672 and might be available in the next release.

crazyscientist:
OK, just checked out the master branch and ran `pylint` on the problematic file.

Looks like the reason for the errors is the same as for #73, but it was not fixed by said pull request :(

chingc:
I'm having the same issue with `import distutils.dir_util`

PCManticore:
Thanks for confirming it's not working for you, we'll check why is that.

AWhetter:
I can't reproduce this.
I'm using virtualenv 16.1.0 with

```
pylint 2.4.0-dev0
astroid 2.3.0
Python 3.6.8 (default, May 21 2019, 02:17:32) 
[GCC 8.3.0]
```

I have no custom config set on pylint and I'm running it on a file containing

```python
from distutils.util import strtobool
```

If you look at the fix in https://github.com/PyCQA/astroid/pull/672, is the condition that was added getting triggered? Are you able to find why not?

crazyscientist:
> If you look at the fix in PyCQA/astroid#672, is the condition that was added getting triggered? Are you able to find why not?

Found it, actually using this unpublished version of astroid containing the fix helps :)

PR #5620, opened 2021-12-31
Regression tests for 73 (distutil import error)

Pierre-Sassoulas:
## Type of Changes

## Description

Refer to https://github.com/PyCQA/astroid/pull/1321

### DIFF
--- a/pylint/typing.py
+++ b/pylint/typing.py
@@ -57,7 +57,7 @@ class MessageLocationTuple(NamedTuple):
 
 
 class ManagedMessage(NamedTuple):
-    """Tuple with information ahout a managed message of the linter"""
+    """Tuple with information about a managed message of the linter"""
 
     name: Optional[str]
     msgid: str
--- a/tests/functional/r/regression_02/regression_distutil_import_error_73.py
+++ b/tests/functional/r/regression_02/regression_distutil_import_error_73.py
@@ -1,0 +1,15 @@
+"""
+Regression test to check that distutils can be imported
+See https://github.com/PyCQA/pylint/issues/73
+
+See also:
+https://github.com/PyCQA/pylint/issues/2955
+https://github.com/PyCQA/astroid/pull/1321
+"""
+
+# pylint: disable=unused-import
+
+import distutils.version
+from distutils.util import strtobool
+from distutils import doesnottexists # [no-name-in-module]
+from distutils.doesnottexists import nope # [no-name-in-module, import-error]
--- a/tests/functional/r/regression_02/regression_distutil_import_error_73.txt
+++ b/tests/functional/r/regression_02/regression_distutil_import_error_73.txt
@@ -1,0 +1,3 @@
+no-name-in-module:14:0:14:36::No name 'doesnottexists' in module 'distutils':UNDEFINED
+import-error:15:0:15:41::Unable to import 'distutils.doesnottexists':UNDEFINED
+no-name-in-module:15:0:15:41::No name 'doesnottexists' in module 'distutils':UNDEFINED


tests/functional/d/distutil_regression_import_error.py
```diff
@@ -0,0 +1,4 @@
+# pylint: disable=missing-docstring
```
DanielNoord, line 1: We might as well add a module docstring + link to some of the relevant issue 😄 

Saves us the trouble of looking them up again when we need to revisit

cdce8p: I agree. These once could be useful
* https://github.com/PyCQA/pylint/issues/73
* https://github.com/PyCQA/pylint/issues/2955
* https://github.com/PyCQA/astroid/pull/1321

Pierre-Sassoulas:
```suggestion
"""
Regression test to check that distutils can be imported
See https://github.com/PyCQA/pylint/issues/73

See also:
https://github.com/PyCQA/pylint/issues/2955
https://github.com/PyCQA/astroid/pull/1321
"""
```

tests/functional/d/distutil_regression_import_error.py
```diff
@@ -0,0 +1,4 @@
...
+import distutils.version
+
+print(distutils.version)
```
Pierre-Sassoulas, line 4:
```suggestion
import distutils.version
from distutils.util import strtobool

print(distutils.version)
print(strtobool)
```

cdce8p (approved):

cdce8p (commented):
Was a bit quick there with the approval. Would it make sense to rename the test file? Most other regression tests start with `regression_` I believe.

tests/functional/d/distutil_regression_import_error.py
```diff
@@ -0,0 +1,13 @@
...
+
+See also:
+https://github.com/PyCQA/pylint/issues/2955
+https://github.com/PyCQA/astroid/pull/1321
+"""
+import distutils.version
+from distutils.util import strtobool
+
+print(distutils.version)
+print(strtobool)
```
DanielNoord, line 13: DIdn't we also want to have a test to see that `distutils` still emitted `no-member` for incorrect imports? We might want to add one case here.

tests/functional/d/distutil_regression_import_error.py
```diff
@@ -0,0 +1,13 @@
+"""
+Regression test to check that distutils can be imported
```
DanielNoord, line 2:
```suggestion
"""Regression test to check that distutils can be imported
```

We need to find some pre-commit hook that starts to enforce this.
Perhaps we can look at something for `Path` at the same time? 😄

DanielNoord (changes requested):

tests/functional/d/distutil_regression_import_error.py
Pierre-Sassoulas: Maybe https://github.com/PyCQA/pydocstyle ?

Pierre-Sassoulas: Better yet https://github.com/myint/docformatter

cdce8p: For module docstrings I think both are fine. Those are tests after all.

> Perhaps we can look at something for Path at the same time? 😄

I wouldn't recommend that. It's not really worth it and might limit us more than it actually would help.

DanielNoord: > I wouldn't recommend that. It's not really worth it and might limit us more than it actually would help.

This was a little joke as @Pierre-Sassoulas keeps having to remind me to use it on the PRs I open 😄 
Don't think we should enforce it all the time indeed.

Perhaps those packages allow us to exclude modules? Oh well, something to explore another time.

Pierre-Sassoulas: Yes we definitely don't want to refactor all the old code that uses os.path.

tests/functional/d/distutil_regression_import_error.py
Pierre-Sassoulas: 22524d2

pylint/typing.py
```diff
@@ -57,7 +57,7 @@ class MessageLocationTuple(NamedTuple):
 
 
 class ManagedMessage(NamedTuple):
-    """Tuple with information ahout a managed message of the linter"""
+    """Tuple with information about a managed message of the linter"""
```
DanielNoord, line 60: Perhaps clean up the commit history and don't squash?

Pierre-Sassoulas:
> Would it make sense to rename the test file?

I renamed :) Although I think having a nice explicit name is better than "regression_1234" so I kept the explicit part.

DanielNoord (approved):

pylint/typing.py
Pierre-Sassoulas: You're right I wanted to rebase originally but there are other commits that prevents that now.

PR #5621, opened 2021-12-31
Add ``future=True`` to all ``NodeNG.frame`` calls

DanielNoord:

## Type of Changes

## Description

Follow up to changes in `astroid` `2.9.1`.


DanielNoord:
Not sure if this should be the PR to update the `astroid` version, but it could be.

pylint/checkers/classes/class_checker.py
```diff
@@ -1659,7 +1659,7 @@ def _is_called_inside_special_method(node: nodes.NodeNG) -> bool:
         Returns true if the node is located inside a special (aka dunder) method
         """
         try:
-            frame_name = node.frame().name
+            frame_name = node.frame(future=True).name
         except AttributeError:
             return False
```
cdce8p, line 1664: Should we catch `ParentMissingError` here now too? Usually it should never happen unless we or a plugin screws up. In that case we would need to catch it for every other call as well which is just unnecessary.

cdce8p (approved):
LGTM! Thanks @DanielNoord 🐬

pylint/checkers/classes/class_checker.py
DanielNoord: We could also just remove it?

We don't cover it right now: https://coveralls.io/builds/45308622/source?filename=pylint%2Fcheckers%2Fclasses%2Fclass_checker.py

The commit that added it also doesn't talk about a plugin doing this wrong or anything: https://github.com/PyCQA/pylint/commit/5309e1d6dacbcbc30f2a4b077903e551dedc2385
The code from the original issue did raise an `AttributeError` but the fact that we no longer cover it and those tests had been added means (I think) that we handle those cases better in an earlier function and we never reach this.

Pierre-Sassoulas (approved):
Neat !

DanielNoord:
@Pierre-Sassoulas I don't think this was completely ready to merge. https://github.com/PyCQA/pylint/pull/5621#discussion_r777069180 wasn't resolved yet 😉

Pierre-Sassoulas:
My bad I got overexcited with all those approved MR :D maybe we don't need to revert everything and we can do a follow-up instead ?

DanielNoord:
Yeah no problem. Do we want to remove the `except` indeed?

DanielNoord:
See #5664 in case we want to go ahead with this!

pylint/checkers/classes/class_checker.py
cdce8p: Yeah, let's remove it! If a plugin screws up, it's better to surface the error than to ignore it silently.

Issue #5627, opened 2022-01-01
Update arg to mypy pre-commit checker


orSolocate:
### Bug description

Running pre-commit checkers on my environment shows `mypy` error related to `missing-imports` .

[link ](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
)to mypy docs.

The error is gone when updating `.pre-commit-config.yaml` adding:

`args: [--ignore-missing-imports]`

Also in pre-commit `mirrors-mypy` [docs](https://github.com/pre-commit/mirrors-mypy) there's a recommendation to is this argument.

Should this be added to pylint too?

### Configuration

```ini
Python 3.8.10
```

### Command used

```shell
running pre-commit's script
```

### Pylint output

```shell
Cannot find implementation or library stub for module named "colorama" (https://mypy.readthedocs.io/en/stable/ files)

http://pylint/reporters/text.py:323:%20note:%20See%20https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

### Expected behavior

mypy passed

### Pylint version

```shell
pylint 2.13.0-dev0
astroid 2.9.1
Python 3.8.10
```

### OS / Environment

Windows 10 Pro 64 bit

DanielNoord:
I think we do this on a per-package basis in `setup.cfg`. @orSolocate Could you try and see if adding colorma there works for Windows? I think most currently active contributors work on Mac or Linux..

orSolocate:
@DanielNoord  - I seem to have the same message again, and I see my fix in `setup.cfg` file was removed. do you have any idea where can I add the `ignore_missing_imports` attribute now?

PR #5628, opened 2022-01-01
Add `iterating-modified-list` checker for modified lists

### LINKED ISSUE #5348: Add a checker for expanding list during for iteration
### Current problem

I would like Pylint to address a classic python bug using `for` loops, when iterating through an expanding list or a dictionary:

For example running:
```py
 item_list = [1, 2, 3]
 for item in item_list:
      item_list.append(1)
     print(item)
```

results in an infinite loop.

For dictionary we have a runtime error raised when this happens:

```py
my_dict = {"1": 1, "2": 2, "3": 3}
i = 1
for item in my_dict:
    my_dict[i] = 1
    i += 1
    print(i)
```

results in: `RuntimeError: dictionary changed size during iteration`

Pylint does not detect both errors.

### Desired solution

I suggest a new checker that detects this structure and suggests using a copy of the list.
in the list example: change `for item in item_list:`  to  `for item in item_list.copy():`
with a message like:
` Expanding list during iteration, consider using <list_name>.copy() instead`
and a name like:
`iteration_expanding_list`

If the community agrees I could implement it :)

### Additional context

See MR #5628

_No response_

orSolocate:
Add `iterating-modified-list` checker for modified lists during loop iteration.

Implemented only for lists, `iterating-modified-dict` and `iterating-modified-set` messages added but without implementation.

## Type of Changes

## Description

Closes #5348


pylint/checkers/base.py
```diff
@@ -63,7 +63,7 @@
 # Copyright (c) 2021 Lorena B <46202743+lorena-b@users.noreply.github.com>
 # Copyright (c) 2021 David Liu <david@cs.toronto.edu>
 # Copyright (c) 2021 Andreas Finkler <andi.finkler@gmail.com>
-# Copyright (c) 2021 Or Bahari <orbahari@mail.tau.ac.il>
+# Copyright (c) 2021-2022 Or Bahari <or.ba402@gmail.com>
```
Pierre-Sassoulas, line 66: You need to modify the copyrite aliases so it's done automatically.

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
         self.add_message("unidiomatic-typecheck", node=node)
 
 
+class ModifiedIterationChecker(_BasicChecker):
```
Pierre-Sassoulas, line 2591: Please create a new file we're trying to limit the number of very big file in the codebase and this one is already gigantic :) Could be ``modified_iteration_checker.py``

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+            "Used when iterating through a set with items added to or removed from during a loop iteration."
+            "Doing so results in RuntimeError",
+        ),
+    }
+
+    options = ()
+    priority = -2
+
+    def __init__(self, linter=None):
+        _BasicChecker.__init__(self, linter)
```
Pierre-Sassoulas, line 2623:
```suggestion
```

It's the same, right ?

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+                    and body_node.value.func.attrname in {"append", "remove"}
+                    and body_node.value.func.expr.name == iter_list_obj.name
+                ):
+                    self.add_message(
+                        "iterating-modified-list",
+                        node=body_node,
+                        args=(iter_list_obj.name,),
+                    )
+        except astroid.InferenceError:
+            pass
```
Pierre-Sassoulas, line 2647: I think the try except should be around the ``infer()`` call

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+
+    @staticmethod
+    def _is_attribute_call_expr(body_node):
```
Pierre-Sassoulas, line 2650:
```suggestion
    def _is_attribute_call_expr(body_node: ?) -> bool:
```
The work to have typing on legacy is enormous but we're trying to have typing on all new functions at least.

tests/checkers/unittest_base.py
```diff
@@ -299,3 +299,81 @@ def test_no_six(self):
             has_six = False
 
         self.assertFalse(has_six, "pylint must be able to run without six")
+
```
Pierre-Sassoulas, line 302: There's no need to duplicate the tests here, functional tests are better and enough. You can move all the example that are different to functional.

Pierre-Sassoulas (changes requested):
This looks nice already I have some remarks but they're mostly about the form and not the substance.

.pre-commit-config.yaml
```diff
@@ -84,7 +84,7 @@ repos:
         entry: mypy
         language: python
         types: [python]
-        args: []
+        args: [--ignore-missing-imports]
```
DanielNoord, line 87: This should be done in a separate PR. See my comment in #5627

doc/whatsnew/2.13.rst
```diff
@@ -14,6 +14,10 @@ New checkers
 
   Closes #5460
 
+* ``iterating-modified-list``: Emitted when a list adds or removes its items while being iterated through
```
DanielNoord, line 17:
```suggestion
* ``iterating-modified-list``: Emitted when items are added or removed to a list being iterated through.
```

ChangeLog
```diff
@@ -9,6 +9,11 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+* Add ``iterating-modified-list`` checker for lists which items are added or removed
+  during iteration while they are iterating through.
```
```suggestion
* Add ``iterating-modified-list``, emitted when items are added or removed to a list being iterated through.
```

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
 
+class ModifiedIterationChecker(_BasicChecker):
+    """Checks for modified for loops iterations
+    Currently supports `for` loops for Sets, Dictionaries and Lists.
+    """
+
+    __implements__ = interfaces.IAstroidChecker
+
+    msgs = {
+        "W4714": (
```
DanielNoord, line 2599: Not sure if @Pierre-Sassoulas agrees but I think the errors can go in `BasicErrorChecker`. Although W4714 is not technically an error I think it makes sense to keep them together, so I'd suggest to move it there as well.

Otherwise I think we can move the two error messages to `BasicErrorChecker` and put the warning in `MisdesignChecker`.

That would also invalidate [this comment](https://github.com/PyCQA/pylint/pull/5628/files#r777143397)

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+            "Iterated list '%s' is being modified inside loop body, consider iterating through a copy of it instead.",
+            "iterating-modified-list",
+            "Used when iterating through a list with items added to or removed from during a loop iteration."
+            "Doing so can result in unexpected behaviour, that is why it is always preferred using a copy of the list.",
```
DanielNoord, line 2603:
```suggestion
            "Emitted when items are added or removed to a list being iterated through. "
            "Doing so can result in unexpected behaviour, that is why it is preferred to use a copy of the list.",
```

ChangeLog
DanielNoord: The other messages should also be listed here and in the other changelog

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+        ),
+        "E4714": (
+            "Iterated dict '%s' is being modified inside loop body, iterate through a copy of it instead.",
+            "iterating-modified-dict",
+            "Used when iterating through a dict with items added to or removed from during a loop iteration."
+            "Doing so results in RuntimeError",
```
DanielNoord, line 2609:
```suggestion
            "Emitted when items are added or removed to a dict being iterated through. "
            "Doing so raises a RuntimeError",
```

pylint/checkers/base.py
```diff
@@ -2588,10 +2588,78 @@ def _check_type_x_is_y(self, node, left, operator, right):
...
+        ),
+        "E4715": (
+            "Iterated set '%s' is being modified inside loop body, iterate through a copy of it instead.",
+            "iterating-modified-set",
+            "Used when iterating through a set with items added to or removed from during a loop iteration."
+            "Doing so results in RuntimeError",
```
DanielNoord, line 2615:
```suggestion
            "Emitted when items are added or removed to a set being iterated through. "
            "Doing so raises a RuntimeError",
```

tests/functional/i/iterating_modified.rc
```diff
@@ -0,0 +1,2 @@
+[Messages Control]
+disable=locally-disabled,useless-suppression,missing-module-docstring
```
DanielNoord, line 2: Can you add the necessary disables to the test file itself? You can add a `pylint: disable` line after the module docstring. We place disables there by convention in the test files so we know what is disabled for each test when opening the file.

tests/functional/i/iterating_modified.py
```diff
@@ -0,0 +1,28 @@
+item_list = [1, 2, 3]
```
DanielNoord, line 1:
```suggestion
"""Tests for iterating-modified messages"""
item_list = [1, 2, 3]
```

tests/functional/i/iterating_modified.py
```diff
@@ -0,0 +1,28 @@
...
+for item in item_list:
+    item_list.append(item)  # [iterating-modified-list]
```
DanielNoord, line 3: Could you also add tests for `iterating-modified-set` and `dict`?

DanielNoord (commented):
Thanks @orSolocate! This is a very nice addition to `2.13` which currently does not include many new checkers and focus mostly on internal clean-up. This makes the release feel much more "complete".

I left some comments, let me know if anything is unclear! 

Edit: I just saw that the primer tests are failing. Something with `nodes.Subscripts`. Perhaps it would be good to also add the code that is currently failing as a test.

pylint/checkers/base.py
Pierre-Sassoulas: Sure, it make sense. We're going to burst base.py into file before releasing 2.13 in #5489 anyway so having the file temporarily be bigger because old checker already inside it get bigger is not a problem.

pylint/checkers/base.py
orSolocate: yeah could remove it

ChangeLog
orSolocate: Although they're not implemented yet?

tests/functional/i/iterating_modified.rc
orSolocate: The rcfiles in functional tests are new for me.. I am actually used to the old method like you said with:

` # pylint disable=`

So what should I put in the rc file?

tests/functional/i/iterating_modified.py
orSolocate: once they are implemented obviously I will

tests/functional/i/iterating_modified.rc
jacobtylerwalls: > I would change the test but what should I put in the rc file?

The rc file is optional, so you can simply delete it. You only need it if there are other settings required for your test (an example would be if your PR changed the default evaluation formula and you needed to test a file with a different evaluation formula).

pylint/checkers/base.py
orSolocate: I understood so I would update `.copyrite_aliases` and keep those manual changes too?

pylint/checkers/base.py
orSolocate: you mean inside the for loop i guess, the infer is part of a big `if` statement

ChangeLog
DanielNoord: Oh sorry, I didn't see that before writing this.

Do you plan on adding them as well in this PR? If not, you should probably also remove their messages and only add those in the relevant PRs

pylint/checkers/base.py
Pierre-Sassoulas: Yes, you can only update the config file, I upgrade all code files automatically during release

ChangeLog
orSolocate: @DanielNoord  Just added them :)

pylint/checkers/base.py
orSolocate: I didn't understand what you guys mean.. I am not familiar with those 2 classes...

pylint/checkers/base.py
```diff
@@ -261,7 +261,7 @@ def _redefines_import(node):
 
 
 def in_loop(node: nodes.NodeNG) -> bool:
-    """Return whether the node is inside a kind of for loop"""
+    """Return whether the node is inside a kind of for loop."""
```
DanielNoord, line 264: Could you remove these changes introduced by `pydocstringformatter`? You have probably installed the latest `main` commit of the tool which is not currently being used by `pylint`.

ChangeLog
```diff
@@ -9,6 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+* * Add ``iterating-modified-list``, emitted when items are added or removed to a list being iterated through.
```
DanielNoord, line 12: Could you re-add the other messages as well?

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+
+    __implements__ = interfaces.IAstroidChecker
+
+    name = "modified_iteration"
+
+    msgs = {
+        "W4714": (
+            "Iterated list '%s' is being modified inside loop body, consider iterating through a copy of it instead.",
+            "iterating-modified-list",
+            "Emitted when items are added or removed to a list being iterated through."
```
DanielNoord, line 26:
```suggestion
            "Emitted when items are added or removed to a list being iterated through. "
```

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+            "Doing so can result in unexpected behaviour, that is why it is preferred to use a copy of the list.",
+        ),
+        "E4714": (
```
DanielNoord, line 29:
```suggestion
        "E4702": (
```

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+    """Checks for modified for loops iterations
+    Currently supports `for` loops for Sets, Dictionaries and Lists.
+    """
+
+    __implements__ = interfaces.IAstroidChecker
+
+    name = "modified_iteration"
+
+    msgs = {
+        "W4714": (
```
DanielNoord, line 23:
```suggestion
        "W4701": (
```

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+            "Emitted when items are added or removed to a list being iterated through."
+            "Doing so can result in unexpected behaviour, that is why it is preferred to use a copy of the list.",
+        ),
+        "E4714": (
+            "Iterated dict '%s' is being modified inside loop body, iterate through a copy of it instead.",
+            "iterating-modified-dict",
+            "Emitted when items are added or removed to a dict being iterated through."
+            "Doing so raises a RuntimeError",
+        ),
+        "E4715": (
```
DanielNoord, line 35:
```suggestion
        "E4703": (
```

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+        "W4714": (
+            "Iterated list '%s' is being modified inside loop body, consider iterating through a copy of it instead.",
+            "iterating-modified-list",
+            "Emitted when items are added or removed to a list being iterated through."
+            "Doing so can result in unexpected behaviour, that is why it is preferred to use a copy of the list.",
+        ),
+        "E4714": (
+            "Iterated dict '%s' is being modified inside loop body, iterate through a copy of it instead.",
+            "iterating-modified-dict",
+            "Emitted when items are added or removed to a dict being iterated through."
```
DanielNoord, line 32:
```suggestion
            "Emitted when items are added or removed to a dict being iterated through. "
```

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+            "Doing so raises a RuntimeError",
+        ),
+        "E4715": (
+            "Iterated set '%s' is being modified inside loop body, iterate through a copy of it instead.",
+            "iterating-modified-set",
+            "Emitted when items are added or removed to a set being iterated through."
```
DanielNoord, line 38:
```suggestion
            "Emitted when items are added or removed to a set being iterated through. "
```

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+    options = ()
+    priority = -2
+
+    @utils.check_messages(
+        "iterating-modified-list", "iterating-modified-dict", "iterating-modified-set"
+    )
+    def visit_for(self, node: nodes.For) -> None:
+        iter_obj = node.iter
+        for body_node in node.body:
+            if self._iterating_modified_list_cond(body_node, iter_obj):
```
DanielNoord, line 52: These names might need to be updated if we change the name of the messages.

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+                )
+            elif self._iterating_modified_set_cond(body_node, iter_obj):
+                self.add_message(
+                    "iterating-modified-set",
+                    node=body_node,
+                    args=(iter_obj.name,),
+                )
+
+    @staticmethod
+    def _is_attribute_call_expr(body_node) -> bool:
```
DanielNoord, line 72: Could you add parameter typing here and to all other methods?

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+            isinstance(body_node, astroid.Expr)
+            and isinstance(body_node.value, astroid.Call)
+            and isinstance(body_node.value.func, astroid.Attribute)
+        )
+
+    @classmethod
+    def _common_cond_list_set(cls, node, list_obj, infer_val):
+        return (
+            cls._is_attribute_call_expr(node)
+            and (infer_val == list_obj.inferred()[0])
```
DanielNoord, line 83: Please use `safe_infer` from `checkers.utils` here.

pylint/checkers/modified_iteration_checker.py
```diff
@@ -0,0 +1,132 @@
...
+    @classmethod
+    def _iterating_modified_list_cond(cls, node, list_obj):
+        try:
+            infer_val = next(node.value.func.expr.infer())
+            return (
+                cls._common_cond_list_set(node, list_obj, infer_val)
+                and (infer_val.pytype() == "builtins.list")
+                and node.value.func.attrname in {"append", "remove"}
+            )
+        except (astroid.InferenceError, AttributeError):
```
DanielNoord, line 102: I'd like for these exceptions to be caught by the correct `isinstance` checks rather than catching them all and returning `False`. We don't know what is failing here and this might actually be exposing real bugs in `pylint` or `astroid`.
Same goes for the others.

tests/functional/c/consider/consider_using_dict_items.py
```diff
@@ -1,5 +1,5 @@
 """Emit a message for iteration through dict keys and subscripting dict with key."""
-# pylint: disable=line-too-long,missing-docstring,unsubscriptable-object,too-few-public-methods,redefined-outer-name,use-dict-literal
+# pylint: disable=locally-disabled,useless-suppression,suppressed-message,line-too-long,missing-docstring,unsubscriptable-object,too-few-public-methods,redefined-outer-name,use-dict-literal,iterating-modified-dict
```
DanielNoord, line 2: Are those necessary? I feel like `useless-suppression` could be avoided here?

tests/functional/i/iterating_modified.py
```diff
@@ -0,0 +1,41 @@
+"""Tests for iterating-modified messages"""
+# pylint: disable=locally-disabled,not-callable,unnecessary-comprehension,suppressed-message,useless-suppression
```
DanielNoord, line 2:
```suggestion
# pylint: disable=ot-callable, unnecessary-comprehension,
```

DanielNoord (changes requested):
Thanks @orSolocate! This is really useful 😄 

I'm unsure about the message name though, what about `modified-iterating-list`? You're modifying the iterating list, that's what we're raising message about.

@Pierre-Sassoulas  Now that the checker is larger it might indeed be better to keep this separate from `BaseChecker`.

.copyrite_aliases
```diff
@@ -113,5 +113,10 @@
     "mails": ["tushar.sadhwani000@gmail.com", "tushar@deepsource.io"],
     "authoritative_mail": "tushar.sadhwani000@gmail.com",
     "name": "Tushar Sadhwani"
+  },
+  {
+    "mails": ["or.ba402@gmail.com"],
+    "name": "Or Bahari",
+    "authoritative_mail": "or.ba402@gmail.com"
```
Pierre-Sassoulas, line 120:
```suggestion
  {
    "mails": ["or.ba402@gmail.com", "orbahari@mail.tau.ac.il"],
    "name": "Or Bahari",
    "authoritative_mail": "or.ba402@gmail.com"
```

Pierre-Sassoulas (changes requested):
I'm not going to review right now, too much will change. Once @DanielNoord remarks are taken into account, please re-request a review :)

pylint/checkers/base.py
orSolocate: yes I did, and okay I will

orSolocate:
Okay I like the name correction I will change it

tests/functional/i/iterating_modified.py
orSolocate: when I run pytest locally it says it's necessary...

tests/functional/c/consider/consider_using_dict_items.py
orSolocate: maybe the tests env is different than testing locally? when I run locally I get a message these are needed... e.g.

```
 Unexpected in testdata:
E          2: useless-suppression
```

pylint/checkers/modified_iteration_checker.py
orSolocate: okay  I will move the `isinstance`  conditions before any infer, so we try to infer only when we are on the correct expression type

tests/functional/c/consider/consider_using_dict_items.py
jacobtylerwalls: Hmm, it says line 2, so isn't that raising a warning _about_ the warnings Daniel doesn't think you need to suppress? I think this passes with just adding `iterating-modified-dict`.

DanielNoord: All functional test should run with `suppressed-message`, `locally-disabled` and `useless-suppression` enabled. I've recently updated the test runner so if this is not the case this is a bug and not working as intended.
As a result of this, these shouldn't need to be added to any individual test file.

Test runner and how it loads configuration can be found at `pylint/testutils/lint_module_test.py`.

orSolocate: @jacobtylerwalls  - This is the messages raised when I run pytest without adding the `useless-suppression` to the disabled list..

This is how I run the test from terminal when I get this message:

`python tests/test_functional.py -k "test_functional[modified_iterating]"`

It seems in the file you sent me that the constructor which disables these messages run in `tests_functional`.
I debugged it and it seems like I get the error because I don't have the `pylintrc` file under `\testutils\testing_pylintrc`

Why there is no empty file by default in the repo? is it a bug? Anyways I would have solved it by adding the message disables to a `finally` block :)

code:

```
        rc_file: Union[Path, str] = PYLINTRC
        try:
            rc_file = test_file.option_file
        except NoFileError:
            pass
```

DanielNoord: Ah, `testing_pylintrc` was a fairly recent addition. See https://github.com/PyCQA/pylint/commit/efe59ca44b9ed350c9790d79a39a7343af5e7ea7
You might have created your local branch when this commit was not yet on `main`. If you pull and merge `main` into your local branch this should be fixed!

orSolocate: it happens when I install pylint as a package. This is when it doesn't create the file. maybe a bug?
(in my local repo I have it)

@DanielNoord 
Also, I think there is a bug in the code I attached, because even if the `.rc`  file doesn't exist we still want to have these disables...

I could upload a quick fix on another branch if you think it's a good idea

DanielNoord: Do you mean installing with `pip install pylint` or with `pip install -e .`? 

Hmm, really?
This is the relevant code:
https://github.com/PyCQA/pylint/blob/fdd71ee0cad6dc965c4a7bd7bc3d16e2ab564f70/pylint/testutils/lint_module_test.py#L47-L63
It sets `rc_file` to `testing_pylintrc` which includes the disables. Then it checks if there is a specific `.rc` file for the tests and overwrite `rc_file`. To make sure we still disable these three messages we then disable them manually on L50-52. I think this should disable all three in all cases, but I would obviously welcome a fix if I'm missing something!

orSolocate:
see `consider_using_dict_items.py` [discussion](https://github.com/PyCQA/pylint/pull/5628#discussion_r785347490),
If you like it I opened a PR for fixing this issue (if it's a real issue).

#5687

@Pierre-Sassoulas  - Windows Python 3.10 test failed due to permission issue, could you please rerun the test maybe it should pass? otherwise could you help me solve it I have no idea where this error came from..

Pierre-Sassoulas:
>  Windows Python 3.10 test failed due to permission issue, could you please rerun the test maybe it should pass? otherwise could you help me solve it I have no idea where this error came from..

I suppose @DanielNoord handled it as the tests are green now, thanks Daniel :)

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,138 @@
...
+        try:
+            infer_val = utils.safe_infer(node.targets[0].value)
+            return (
+                (infer_val is not None)
+                and (infer_val.pytype() == "builtins.dict")
+                and (infer_val == list_obj.inferred()[0])
+                and (node.targets[0].value.name == list_obj.name)
+            )
+        except (astroid.InferenceError, AttributeError):
+            return False
```
Pierre-Sassoulas, line 119: In this part of the code the except seems to be covered by the test... It makes me reconsider my other comment. Maybe the inference error come from ``list_obj.inferred()`` and the code can't raise an inference error for the 2 other functions ? If we can have an inference error for the other, we should add a test case where the inference fail and I think the try/except should be around the minimal amount of code (not safe_infer). If it can't raise an ``InferenceError`` we should remove the try.

```suggestion
        infer_val = utils.safe_infer(node.targets[0].value)
        try:
            return (
                (infer_val is not None)
                and (infer_val.pytype() == "builtins.dict")
                and (infer_val == list_obj.inferred()[0])
                and (node.targets[0].value.name == list_obj.name)
            )
        except (astroid.InferenceError, AttributeError):
            return False
```

Pierre-Sassoulas (changes requested):
Thank you, @orSolocate, this looks great and polished already.

pylint/checkers/modified_iterating_checker.py
DanielNoord: I think a cleaner solution would be to split the if statements. Do the first two checks in a separate if and then do a `safe_infer` on `list_obj` which can then lead to either `False` or `True`. We'll need to add 2 or 3 extra return statements but I think it reads much nicer and allows the removal of those `try...except`'s.

Pierre-Sassoulas: Something like this ?
```suggestion
        infer_val = utils.safe_infer(node.targets[0].value)
        if infer_val is None or infer_val.pytype() != "builtins.dict":
            # Uninferable or not a dict
            return False
        if node.targets[0].value.name != list_obj.name:
            return False
        try:
            return infer_val == list_obj.inferred()[0]
        except (astroid.InferenceError, AttributeError):
            return False
```

DanielNoord: Yes, but then only with a `safe_infer` on L117 to remove the `try...except` (I think that should be possible).

orSolocate:
One thing I wasn't sure about: the message is for the line inside the `for` loop and now the for loop itself.
I think I will modify it to be the `for` loop and if there is more than 1 message for a specific `for` only 1 message will be raised

ChangeLog
```diff
@@ -9,6 +9,11 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+* Add ``modified-iterating-list``, emitted when items are added to or removed from a list being iterated through.
+* Add ``modified-iterating-dict``, emitted when items are added to or removed from a dict being iterated through.
```
```suggestion
* Add ``modified-iterating-dict``, emitted when items are added to or removed from a dictionary being iterated through.
```

doc/whatsnew/2.13.rst
```diff
@@ -37,6 +37,10 @@ New checkers
 
   Closes #5460
 
+* ``modified-iterating-list/dict/set``: Emitted when items are added to or removed from a list/dict/set being iterated through.
+
+  Closes #5348
```
DanielNoord, line 42:
```suggestion
* Add ``modified-iterating-list``, emitted when items are added to or removed from a list being iterated through.
* Add ``modified-iterating-dict``, emitted when items are added to or removed from a dictionary being iterated through.
* Add ``modified-iterating-set``, emitted when items are added to or removed from a set being iterated through.

  Closes #5348
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+                    msg_id,
+                    node=node,
+                    args=(iter_obj.name,),
+                )
+                break  # since the msg is raised for the `for` node no further check is needed
+
+    @staticmethod
+    def _is_attribute_call_expr(body_node) -> bool:
+        return (
+            isinstance(body_node, astroid.Expr)
```
DanielNoord, line 73:
```suggestion
            isinstance(body_node, nodes.Expr)
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            and isinstance(body_node.value, astroid.Call)
```
DanielNoord, line 74:
```suggestion
            and isinstance(body_node.value, nodes.Call)
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            and isinstance(body_node.value.func, astroid.Attribute)
```
DanielNoord, line 75:
```suggestion
            and isinstance(body_node.value.func, nodes.Attribute)
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+        try:
+            return (infer_val == utils.safe_infer(list_obj)) and (
+                node.value.func.expr.name == list_obj.name
+            )
+        except AttributeError:
+            return False
+
+    @staticmethod
+    def _dict_node_cond(node) -> bool:
+        return isinstance(node, astroid.Assign) and (
```
DanielNoord, line 89:
```suggestion
        return isinstance(node, nodes.Assign) and (
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            isinstance(node.targets[0], astroid.Subscript)
```
DanielNoord, line 90:
```suggestion
            isinstance(node.targets[0], nodes.Subscript)
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+
+from typing import TYPE_CHECKING
+
+import astroid
```
DanielNoord, line 6:
```suggestion
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+from pylint import checkers, interfaces
+from pylint.checkers import utils
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class ModifiedIterationChecker(checkers.BaseChecker):
+    """Checks for modified for loops iterations
+    Currently supports `for` loops for Sets, Dictionaries and Lists.
```
DanielNoord, line 18:
```suggestion
    """Checks for modified for loops iterations.
    
    Currently supports `for` loops for Sets, Dictionaries and Lists.
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+    __implements__ = interfaces.IAstroidChecker
+
+    name = "modified_iteration"
+
+    msgs = {
+        "W4701": (
+            "Iterated list '%s' is being modified inside loop body, consider iterating through a copy of it instead.",
+            "modified-iterating-list",
+            "Emitted when items are added or removed to a list being iterated through. "
+            "Doing so ccan result in unexpected behaviour, that is why it is preferred to use a copy of the list.",
```
DanielNoord, line 30:
```suggestion
            "Doing so can result in unexpected behaviour, that is why it is preferred to use a copy of the list.",
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+        ),
+        "E4702": (
+            "Iterated dict '%s' is being modified inside loop body, iterate through a copy of it instead.",
+            "modified-iterating-dict",
+            "Emitted when items are added or removed to a dict being iterated through. "
+            "Doing so raises a RuntimeError",
```
DanielNoord, line 36:
```suggestion
            "Doing so raises a RuntimeError.",
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+        ),
+        "E4703": (
+            "Iterated set '%s' is being modified inside loop body, iterate through a copy of it instead.",
+            "modified-iterating-set",
+            "Emitted when items are added or removed to a set being iterated through. "
+            "Doing so raises a RuntimeError",
```
DanielNoord, line 42:
```suggestion
            "Doing so raises a RuntimeError.",
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+                msg_id = "modified-iterating-dict"
+            elif self._modified_iterating_set_cond(body_node, iter_obj):
+                msg_id = "modified-iterating-set"
+            if msg_id is not None:
+                self.add_message(
+                    msg_id,
+                    node=node,
+                    args=(iter_obj.name,),
+                )
+                break  # since the msg is raised for the `for` node no further check is needed
```
DanielNoord, line 68: I don't particularly like this: If `modified-iterating-list` is being ignored in `pylintrc` and `modified-iterating-set` should also be raised it won't raise the message any longer and there will be no messages at all.

Personally I always prefer raising every correct message as you never know how users disable their code. If both a `dict` and a `set` are being modified we should warn accordingly.
I wonder if this is needed though: how can you iterate over a `set` and a `dict` at the same time?

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @staticmethod
+    def _is_attribute_call_expr(body_node) -> bool:
```
DanielNoord, line 71: Could you add typing here? I guess `body_node: nodes.NodeNG`

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+        return (
+            isinstance(body_node, astroid.Expr)
+            and isinstance(body_node.value, astroid.Call)
+            and isinstance(body_node.value.func, astroid.Attribute)
+        )
+
+    @classmethod
+    def _common_cond_list_set(cls, node, list_obj, infer_val) -> bool:
```
DanielNoord, line 79: Please add argument typing here

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+        try:
+            return (infer_val == utils.safe_infer(list_obj)) and (
+                node.value.func.expr.name == list_obj.name
+            )
+        except AttributeError:
```
DanielNoord, line 84: Although not really pythonic perhaps, we like to use `isinstance` checks to avoid `AttributeErrors`. Could you refactor this?

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            return False
+
+    @staticmethod
+    def _dict_node_cond(node) -> bool:
```
DanielNoord, line 88:
```suggestion
    def _dict_node_cond(node: nodes.NodeNG) -> bool:
```

I think?
And could you add a one line docstring or make the method name a little more expressive?

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+        return isinstance(node, astroid.Assign) and (
+            isinstance(node.targets[0], astroid.Subscript)
+        )
+
+    @classmethod
+    def _modified_iterating_list_cond(cls, node, list_obj) -> bool:
```
DanielNoord, line 94: I'm not going to make any more comments, but please add parameter typing to all methods.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+        if not cls._is_attribute_call_expr(node):
+            return False
+        infer_val = utils.safe_infer(node.value.func.expr)
+        if infer_val is None or infer_val.pytype() != "builtins.list":
```
DanielNoord, line 98:
```suggestion
        if not isinstance(infer_val, nodes.List):
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            # Uninferable or not a list
+            return False
+        return cls._common_cond_list_set(
+            node, list_obj, infer_val
+        ) and node.value.func.attrname in {"append", "remove"}
```
DanielNoord, line 103: Please use a module constant for this defined after the imports.
Something like:
`LIST_CHANGE_METHODS = {"append", "remove"}`

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @classmethod
+    def _modified_iterating_dict_cond(cls, node, list_obj) -> bool:
+        if not cls._dict_node_cond(node):
+            return False
+        infer_val = utils.safe_infer(node.targets[0].value)
+        if infer_val is None or infer_val.pytype() != "builtins.dict":
```
DanielNoord, line 110:
```suggestion
        if not isinstance(infer_val, nodes.Dict):
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            # Uninferable or not a dict
+            return False
+        if infer_val != utils.safe_infer(list_obj):
+            return False
+        try:
+            return node.targets[0].value.name == list_obj.name
+        except AttributeError:
```
DanielNoord, line 117: Please use an `isinstance` here

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            return False
+
+    @classmethod
+    def _modified_iterating_set_cond(cls, node, list_obj) -> bool:
+        if not cls._is_attribute_call_expr(node):
+            return False
+        infer_val = utils.safe_infer(node.value.func.expr)
+        if infer_val is None or infer_val.pytype() != "builtins.set":
```
DanielNoord, line 125:
```suggestion
        if not isinstnace(infer_val, nodes.Set):
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+            # Uninferable or not a set
+            return False
+        return cls._common_cond_list_set(
+            node, list_obj, infer_val
+        ) and node.value.func.attrname in {"add", "remove"}
```
DanielNoord, line 130: Use a module global here.

tests/functional/m/modified_iterating.py
```diff
@@ -0,0 +1,42 @@
...
+for item in my_dict.copy():
+    my_dict[i] = 1
+    i += 1
+
+item_set = {1, 2, 3}
+for item in item_set:  # [modified-iterating-set]
+    item_set.add(item + 10)
+
+for item in item_set.copy():
+    item_set.add(item + 10)
```
DanielNoord, line 42: Could you add some tests with nested for loops?

And if you can think of one, a test where two types are modified like you mentioned in your last comment.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,134 @@
...
+                msg_id = "modified-iterating-list"
+            elif self._modified_iterating_dict_cond(body_node, iter_obj):
+                msg_id = "modified-iterating-dict"
+            elif self._modified_iterating_set_cond(body_node, iter_obj):
+                msg_id = "modified-iterating-set"
+            if msg_id is not None:
+                self.add_message(
+                    msg_id,
+                    node=node,
+                    args=(iter_obj.name,),
```
DanielNoord, line 66:
```suggestion
                    args=(iter_obj.name,),
                    confidence=interfaces.INFERENCE
```

DanielNoord (changes requested):
Some comments about some the code style and some potential refactors. The checks themselves look very promising!

tests/functional/m/modified_iterating.py
```diff
@@ -0,0 +1,42 @@
...
+
+import copy
+
+item_list = [1, 2, 3]
+for item in item_list:  # [modified-iterating-list]
+    item_list.append(item)
+
+for item in item_list:  # [modified-iterating-list]
+    item_list.remove(item)
+    item_list.append(1)
```
orSolocate, line 12: @DanielNoord  I added this line to cover that.

tests/functional/m/modified_iterating.py
orSolocate: https://github.com/PyCQA/pylint/pull/5628/files/7be0f88cf01f91a0dc72c18e77d9ab65c4301341#r790330656

DanielNoord: I'm not sure I understand what you're referring to. That isn't a nested for loop, I meant:
```python
for x in y:
    for a in z:
        pass
```

It also doesn't change two types?

orSolocate: added

orSolocate:
Got rid of all `try-except` blocks(!)
Thanks for the detailed code review guys! I really appreciate it and learned a lot about this repo's conventions and style. I feel this checker is now of a really high quality 🗡️ 
Should I squash all to a single commit and (force) push before merging to main branch?

Pierre-Sassoulas:
> Should I squash all to a single commit and (force) push before merging to main branch?

Don't worry we'll squash if required. You could reorganize the commits yourself as there is a commit where you modify the ``copyright_aliases`` that would make more sense alone (or open another MR so it get merged first).

I don't have time to review right now but @DanielNoord did such a good job that some of his comment are hidden by default and were not taken into account. You should click on "X hidden conversations" to see / apply them 😄

pylint/checkers/modified_iterating_checker.py
orSolocate: @DanielNoord  - I understand what you mean but I think you are right, every `for` loop is considered separately and every `for` loop can contain only 1 list/dict/set.
and for the rare case that we iterate a `list` and a `set` or `dict` in the SAME `for` loop, if it exist, it wouldn't be detected anyway probably (because the structure of such `for` loop is not supported in my code).

orSolocate:
@Pierre-Sassoulas  great i cleared all other suggestions in my lastcommit.
Also made a separate PR for the aliases at #5721

doc/whatsnew/2.13.rst
```diff
@@ -37,6 +37,12 @@ New checkers
 
   Closes #5460
 
+* ``modified-iterating-list``: Emitted when items are added to or removed from a list being iterated through.
+* ``modified-iterating-dict``: Emitted when items are added to or removed from a dict being iterated through.
```
DanielNoord, line 41:
```suggestion
* ``modified-iterating-dict``: Emitted when items are added to or removed from a dictionary being iterated through.
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,145 @@
...
+from astroid import nodes
+
+from pylint import checkers, interfaces
+from pylint.checkers import utils
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+__LIST_MODIFIER_METHODS__ = {"append", "remove"}
```
DanielNoord, line 15:
```suggestion
_LIST_MODIFIER_METHODS = {"append", "remove"}
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,145 @@
...
+__SET_MODIFIER_METHODS__ = {"add", "remove"}
```
DanielNoord, line 16:
```suggestion
_SET_MODIFIER_METHODS = {"add", "remove"}
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class ModifiedIterationChecker(checkers.BaseChecker):
+    """Checks for modified for loops iterations
```
DanielNoord, line 20:
```suggestion
    """Checks for modified iterators in for loops iterations.
```

pylint/checkers/modified_iterating_checker.py
DanielNoord: Do we need the `break` statement here then? If we don't consider the case where we iterate over two types of iterators this `break` doesn't do much for us right?

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,145 @@
...
+                self.add_message(
+                    msg_id,
+                    node=node,
+                    args=(iter_obj.name,),
+                    confidence=interfaces.INFERENCE,
+                )
+                break  # since the msg is raised for the `for` node no further check is needed
+
+    @staticmethod
+    def _is_node_expr_calls_attribute_name(node: nodes.NodeNG) -> bool:
```
DanielNoord, line 76:
```suggestion
    def _is_node_expr_that_calls_attribute_name(node: nodes.NodeNG) -> bool:
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,145 @@
...
+            isinstance(node, nodes.Expr)
+            and isinstance(node.value, nodes.Call)
+            and isinstance(node.value.func, nodes.Attribute)
+            and isinstance(node.value.func.expr, nodes.Name)
+        )
+
+    @classmethod
+    def _common_cond_list_set(
+        cls,
+        node: nodes.NodeNG,
```
DanielNoord, line 87: This `node` can be narrowed due to the callers of this function.
`node.value.func.expr.name` can't be called on a simple `nodes.NodeNG` so we need to narrow down this node. (I don't know why `mypy` doesn't pick this up.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,145 @@
...
+        cls, node: nodes.NodeNG, list_obj: nodes.NodeNG
+    ) -> bool:
+        if not cls._is_node_expr_calls_attribute_name(node):
+            return False
+        infer_val = utils.safe_infer(node.value.func.expr)
+        if not isinstance(infer_val, nodes.List):
+            return False
+        return (
+            cls._common_cond_list_set(node, list_obj, infer_val)
+            and node.value.func.attrname in __LIST_MODIFIER_METHODS__
```
DanielNoord, line 113:
```suggestion
            and node.value.func.attrname in _LIST_MODIFIER_METHODS
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @classmethod
+    def _modified_iterating_dict_cond(
+        cls, node: nodes.NodeNG, list_obj: nodes.NodeNG
+    ) -> bool:
+        if not cls._is_node_assigns_subscript_name(node):
+            return False
+        infer_val = utils.safe_infer(node.targets[0].value)
+        if not isinstance(infer_val, nodes.Dict):
+            return False
```
DanielNoord, line 124: This line is uncovered. Can we make a test case that hits this line?

tests/functional/m/modified_iterating.py
```diff
@@ -0,0 +1,47 @@
...
+
+item_set = {1, 2, 3}
+for item in item_set:  # [modified-iterating-set]
+    item_set.add(item + 10)
+
+for item in item_set.copy():
+    item_set.add(item + 10)
+
+for l in item_list:  # [modified-iterating-list]
+    for s in item_set:  # [modified-iterating-set]
```
DanielNoord, line 44: Perfect, thanks!

DanielNoord (changes requested):
Sorry for coming with this so late into the review process:
I think we need to raise the message on the `append` or `remove`.
If you have a very long `for` loop you won't know what line is actually causing the message if the message is being raised on the `for` loop itself.

I think we can also add some more cases such as `pop` and `update`, but perhaps we should do a follow up PR for that?

Pierre-Sassoulas (commented):
I started a review that I left pending but I see Daniel also reviewed, so I'm just going to re-review later :) Could you rebase on the latest main so the MR for copyryte alias is taken into account, please ?

doc/whatsnew/2.13.rst
```diff
@@ -37,6 +37,12 @@ New checkers
...
+* ``modified-iterating-set``: Emitted when items are added to or removed from a set being iterated through.
```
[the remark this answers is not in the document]
orSolocate: @DanielNoord  - The message was at the `remove/add` before but I decided to change it to the `for` line. The reason is that the actual thing you need to fix is the `for` line. I agree with you though that the user wouldn't know for sure why this message was raised for long `for` loops, so I suggest modifying the message so that the `add/remove` line/line number is shown there. also we can get rid of the "only one message per `for` loop" logic and raise more than 1 message if necessary since after adding the 'problematic' `add/remove` line makes the messages different.

* for pop/update - we can open an issue for extending this checker after this is merged.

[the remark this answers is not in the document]
orSolocate: well removing the `add/remove` changes the logic of your code and you don't want to suggest the user to do that in your message. iterating through a copy of the `list` in the `for` loop keeps the logic the same.
Let's play with an example, say you have a list of numbers and for each nonzero-1-digit number in the list you want to modify it to a 2-digit number and have copies of the new number times the value of the original, so you'd write something like this:

```
lst = [1, 2, 3]
for item in lst:
    if item//10 == 0:
        lst.remove(item)
        for i in range(item):
            lst.append((item * 10) + item)
```
You expect:

`lst = [11, 22, 22, 33, 33, 33]`

but you get:
`lst = [2, 11, 33, 33, 33]`

 The code has unexpected behavior because of our bug. you need the `add/remove` lines because this is what you want to do.. iterating through the copy of the list gives the expected value in `lst`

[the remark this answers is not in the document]
DanielNoord: I agree that his is the bug we're trying to solve here. However, I still don't think we should raise on the `for` loop:
1) The actual offending line is the `add/remove`, I believe we should always raise on what is wrong instead of where to solve.
2) If we raise on the `add/remove` line it's trivial to find the `for` loop (indentation), the other way around is more difficult. There are situations where providing the `lineno` doesn't really help in finding the offending line (see `Zen` mode in VS Code)
3) _It is also more in line with the message name, the list/dict/set gets modified on the `add/remove` line._

I can't come up with an example for this, but I think there are other examples within `pylint` where the solution to a message is not necessarily on the line where the message is raised. It also helps python beginners why something is raised, which is probably also the target for this message. (More experienced users aren't likely to do this).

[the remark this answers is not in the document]
Pierre-Sassoulas: I agree with Daniel here, pylint always point to the problem, not to the solution. It sometime suggest a solution in its message but it would be counterintuitive to start doing this.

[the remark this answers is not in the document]
orSolocate: Okay I would change to the `add/remove` line.. then we could have more than 1 message for the same `for` loop.
I still think the solution suggestion is helpful and should be included in the message.

A small suggestion to improve the message, adding the word `for`:

Iterated list '%s' is being modified inside **for** loop body, consider iterating through a copy of it instead.

pylint/checkers/modified_iterating_checker.py
orSolocate: Sure, I found something and added to the test  :)

doc/whatsnew/2.13.rst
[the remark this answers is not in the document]
Pierre-Sassoulas: Very nice message 👍

ChangeLog
```diff
@@ -9,6 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+* Add ``modified-iterating-list``, ``modified-iterating-dict`` and ``modified-iterating-set`` messages emitted when
+  items are added to or removed from a list being iterated through.
```
Pierre-Sassoulas, line 13:
```suggestion
* Add ``modified-iterating-list``, ``modified-iterating-dict``, and ``modified-iterating-set``,
  emitted when items are added to or removed from respectively a list, dictionary or
  set being iterated through.
```
I don't think this comment was applied previousely

doc/whatsnew/2.13.rst
```diff
@@ -37,6 +37,12 @@ New checkers
 
   Closes #5460
 
+* ``modified-iterating-list``: Emitted when items are added to or removed from a list being iterated through.
+* ``modified-iterating-dict``: Emitted when items are added to or removed from a dictionary being iterated through.
+* ``modified-iterating-set``: Emitted when items are added to or removed from a set being iterated through.
+
```
Pierre-Sassoulas, line 43:
```suggestion
* Add ``modified-iterating-list``, ``modified-iterating-dict``, and ``modified-iterating-set``,
  emitted when items are added to or removed from respectively a list, dictionary or
  set being iterated through.

```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+    def visit_for(self, node: nodes.For) -> None:
+        iter_obj = node.iter
+        for body_node in node.body:
+            message_raised = self._modified_iterating_check(body_node, iter_obj)
+            if not message_raised:  # needs further check when message not raised
+                for child in body_node.get_children():
+                    self._modified_iterating_check(child, iter_obj)
+
+    def _modified_iterating_check(self, node, iter_obj):
```
Pierre-Sassoulas, line 65:
```suggestion
    def _modified_iterating_check(self, node: ?, iter_obj: ?) -> bool:
```
I don't know about the type of ``body_node`` maybe NodeNG ? for iter_obj there's probably a particular node class.

Pierre-Sassoulas (changes requested):
Thank you for taking all the remarks into account @orSolocate (and thank you @DanielNoord for the great review!). This look really good, I have three minor remarks then it's good to go.

pylint/checkers/modified_iterating_checker.py
orSolocate: In my simple that it can be all of: `ListComp`, `Name`, `Cal` .. so I will stick with `NodeNG`..

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,155 @@
...
+    priority = -2
+
+    @utils.check_messages(
+        "modified-iterating-list", "modified-iterating-dict", "modified-iterating-set"
+    )
+    def visit_for(self, node: nodes.For) -> None:
+        iter_obj = node.iter
+        for body_node in node.body:
+            message_raised = self._modified_iterating_check(body_node, iter_obj)
+            if not message_raised:  # needs further check when message not raised
```
DanielNoord, line 61: This is me being stupid, but what case is this covering? It's covered by `coveralls` so clearly it's doing something correctly but I can't immediately see what test case this covers.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,155 @@
...
+            and isinstance(node.value, nodes.Call)
+            and isinstance(node.value.func, nodes.Attribute)
+            and isinstance(node.value.func.expr, nodes.Name)
+        )
+
+    @classmethod
+    def _common_cond_list_set(
+        cls,
+        node: nodes.Expr,
+        list_obj: nodes.NodeNG,
```
DanielNoord, line 98: Could you change this parameter name? It can be a `set` as well right? Perhaps `iter_obj`?

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        infer_val: Optional[nodes.NodeNG],
```
DanielNoord, line 99:
```suggestion
        infer_val: Union[nodes.Set, nodes.List],
```

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,155 @@
...
+    @staticmethod
+    def _is_node_assigns_subscript_name(node: nodes.NodeNG) -> bool:
+        return isinstance(node, nodes.Assign) and (
+            isinstance(node.targets[0], nodes.Subscript)
+            and (isinstance(node.targets[0].value, nodes.Name))
+        )
+
+    @classmethod
+    def _modified_iterating_list_cond(
+        cls, node: nodes.NodeNG, list_obj: nodes.NodeNG
```
DanielNoord, line 114:
```suggestion
        cls, node: nodes.NodeNG, iter_obj: nodes.NodeNG
```

We don't actually know if this is a list yet.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        if not isinstance(infer_val, nodes.List):
+            return False
+        return (
+            cls._common_cond_list_set(node, list_obj, infer_val)
+            and node.value.func.attrname in _LIST_MODIFIER_METHODS
+        )
+
+    @classmethod
+    def _modified_iterating_dict_cond(
+        cls, node: nodes.NodeNG, list_obj: nodes.NodeNG
```
DanielNoord, line 128:
```suggestion
        cls, node: nodes.NodeNG, iter_obj: nodes.NodeNG
```

Similarly.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -0,0 +1,155 @@
...
+        infer_val = utils.safe_infer(node.targets[0].value)
+        if not isinstance(infer_val, nodes.Dict):
+            return False
+        if infer_val != utils.safe_infer(list_obj):
+            return False
+        return node.targets[0].value.name == list_obj.name
+
+    @classmethod
+    def _modified_iterating_set_cond(
+        cls, node: nodes.NodeNG, list_obj: nodes.NodeNG
```
DanielNoord, line 141:
```suggestion
        cls, node: nodes.NodeNG, iter_obj: nodes.NodeNG
```

Similarly.

DanielNoord (changes requested):
I'm really nitpicky here, but this seems like code we're not likely to revisit often (as it is quite robust) so I like to get everything spot on the first time.

pylint/checkers/modified_iterating_checker.py
orSolocate: the `_modified_iterating_check` checks if there is 1 of the 3 message to raise on each node in the `for` loop.
If not: it checks also it's children nodes. this covers the "for inside for" test when we want to raise a message for the `append` inside the inner for loop regarding the outer for loop node

pylint/checkers/modified_iterating_checker.py
orSolocate: yeah sure okay

pylint/checkers/modified_iterating_checker.py
orSolocate: + `None` for failed infer

pylint/checkers/modified_iterating_checker.py
DanielNoord: Ah understood!

What about:
```python
for l in item_list:
     for s in item_set:
         for x in iterator:
             print()
             item_list.append(1)  # [modified-iterating-list]
```
Do we catch that as well? It seems like we can only go down one level right now?

orSolocate: I think so too but since it is a false-nagative and more of a rare case I suggest to merge this MR and open a new issue for it.

DanielNoord: Yes, but the current structure of the code doesn't allow for this. Therefore, I think we should change it now. The following diff fixes the issue:
```diff
diff --git a/pylint/checkers/modified_iterating_checker.py b/pylint/checkers/modified_iterating_checker.py
index 55aeaf518..d745a9606 100644
--- a/pylint/checkers/modified_iterating_checker.py
+++ b/pylint/checkers/modified_iterating_checker.py
@@ -57,10 +57,15 @@ class ModifiedIterationChecker(checkers.BaseChecker):
     def visit_for(self, node: nodes.For) -> None:
         iter_obj = node.iter
         for body_node in node.body:
-            message_raised = self._modified_iterating_check(body_node, iter_obj)
-            if not message_raised:  # needs further check when message not raised
-                for child in body_node.get_children():
-                    self._modified_iterating_check(child, iter_obj)
+            self._check_node_and_children_for_changes_to_iterator(body_node, iter_obj)
+
+    def _check_node_and_children_for_changes_to_iterator(
+        self, body_node: nodes.NodeNG, iter_obj: nodes.NodeNG
+    ) -> None:
+        """See if node or any of its children interact with iterator node."""
+        self._modified_iterating_check(body_node, iter_obj)
+        for child in body_node.get_children():
+            self._check_node_and_children_for_changes_to_iterator(child, iter_obj)
 
     def _modified_iterating_check(
         self, node: nodes.NodeNG, iter_obj: nodes.NodeNG
```

With that change the following test also passes:
```python
# Check for nested for loops and changes to iterators
for l in item_list:
    item_list.append(1)  # [modified-iterating-list]
    for _ in []:
        item_list.append(1)  # [modified-iterating-list]
        for _ in []:
            item_list.append(1)  # [modified-iterating-list]
            for _ in []:
                item_list.append(1)  # [modified-iterating-list]
```

pylint/checkers/modified_iterating_checker.py
orSolocate: fixed with `Optional[nodes.Set, nodes.List]`

Pierre-Sassoulas (approved):
Thank you for the new checker and thank you for taking all the remark into account @orSolocate !

pylint/checkers/modified_iterating_checker.py
orSolocate: nice. verified it works and added this test

DanielNoord:
Suggested edit:
```diff
diff --git a/pylint/checkers/modified_iterating_checker.py b/pylint/checkers/modified_iterating_checker.py
--- pylint/checkers/modified_iterating_checker.py
+++ pylint/checkers/modified_iterating_checker.py
@@ -68,25 +68,23 @@
             self._modified_iterating_check_on_node_and_children(child, iter_obj)
 
     def _modified_iterating_check(
         self, node: nodes.NodeNG, iter_obj: nodes.NodeNG
-    ) -> bool:
+    ) -> None:
         msg_id = None
         if self._modified_iterating_list_cond(node, iter_obj):
             msg_id = "modified-iterating-list"
         elif self._modified_iterating_dict_cond(node, iter_obj):
             msg_id = "modified-iterating-dict"
         elif self._modified_iterating_set_cond(node, iter_obj):
             msg_id = "modified-iterating-set"
-        if msg_id is not None:
+        if msg_id:
             self.add_message(
                 msg_id,
                 node=node,
                 args=(iter_obj.name,),
                 confidence=interfaces.INFERENCE,
             )
-            return True
-        return False
 
     @staticmethod
     def _is_node_expr_that_calls_attribute_name(node: nodes.NodeNG) -> bool:
         return (
@@ -95,11 +93,10 @@
             and isinstance(node.value.func, nodes.Attribute)
             and isinstance(node.value.func.expr, nodes.Name)
         )
 
-    @classmethod
+    @staticmethod
     def _common_cond_list_set(
-        cls,
         node: nodes.Expr,
         iter_obj: nodes.NodeNG,
         infer_val: Union[nodes.List, nodes.Set],
     ) -> bool:
@@ -113,46 +110,46 @@
             isinstance(node.targets[0], nodes.Subscript)
             and (isinstance(node.targets[0].value, nodes.Name))
         )
 
-    @classmethod
+    @staticmethod
     def _modified_iterating_list_cond(
-        cls, node: nodes.NodeNG, iter_obj: nodes.NodeNG
+        self, node: nodes.NodeNG, iter_obj: nodes.NodeNG
     ) -> bool:
-        if not cls._is_node_expr_that_calls_attribute_name(node):
+        if not self._is_node_expr_that_calls_attribute_name(node):
             return False
         infer_val = utils.safe_infer(node.value.func.expr)
         if not isinstance(infer_val, nodes.List):
             return False
         return (
-            cls._common_cond_list_set(node, iter_obj, infer_val)
+            self._common_cond_list_set(node, iter_obj, infer_val)
             and node.value.func.attrname in _LIST_MODIFIER_METHODS
         )
 
-    @classmethod
+    @staticmethod
     def _modified_iterating_dict_cond(
-        cls, node: nodes.NodeNG, iter_obj: nodes.NodeNG
+        self, node: nodes.NodeNG, iter_obj: nodes.NodeNG
     ) -> bool:
-        if not cls._is_node_assigns_subscript_name(node):
+        if not self._is_node_assigns_subscript_name(node):
             return False
         infer_val = utils.safe_infer(node.targets[0].value)
         if not isinstance(infer_val, nodes.Dict):
             return False
         if infer_val != utils.safe_infer(iter_obj):
             return False
         return node.targets[0].value.name == iter_obj.name
 
-    @classmethod
+    @staticmethod
     def _modified_iterating_set_cond(
-        cls, node: nodes.NodeNG, iter_obj: nodes.NodeNG
+        self, node: nodes.NodeNG, iter_obj: nodes.NodeNG
     ) -> bool:
-        if not cls._is_node_expr_that_calls_attribute_name(node):
+        if not self._is_node_expr_that_calls_attribute_name(node):
             return False
         infer_val = utils.safe_infer(node.value.func.expr)
         if not isinstance(infer_val, nodes.Set):
             return False
         return (
-            cls._common_cond_list_set(node, iter_obj, infer_val)
+            self._common_cond_list_set(node, iter_obj, infer_val)
             and node.value.func.attrname in _SET_MODIFIER_METHODS
         )
 
 

```

orSolocate:
You're welcome guys! uploaded the changes

DanielNoord (approved):

Pierre-Sassoulas:
Great addition to 2.13.0 @orSolocate, thank you for being so persistent !

PR #5640, opened 2022-01-05
Add checker `redefined-slots-in-subclass`

### LINKED ISSUE #5617: Add new checker `redefined-slots-in-subclass`
### Current problem

According to the Python documentation, slots defined in a subclass should contain the names of **additional** slots only.

Example - ok:
```
class Base:
    __slots__ = "a", "b"

class Subclass(Base):
    __slots__ = "c", "d"
```

Example - Not ok:
```
class Base:
    __slots__ = "a", "b"

class Subclass(Base):
    __slots__ = "a", "c"
```

### Desired solution

Emit a warning when a subclass is encountered which redefines a slot which is already defined in one of its parent classes;
For example: `redefined-slots-in-subclass`

### Additional context

Quoting from the Python documentation: https://docs.python.org/3/reference/datamodel.html:

_The action of a __slots__ declaration is not limited to the class where it is defined. __slots__ declared in parents are available in child classes. However, child subclasses will get a __dict__ and __weakref__ unless they also define __slots__ (**which should only contain names of any additional slots**)._

_If a class defines a slot also defined in a base class, the instance variable defined by the base class slot is inaccessible (except by retrieving its descriptor directly from the base class). This renders the meaning of the program undefined. In the future, a check may be added to prevent this._

mbyrnepr2:
## Type of Changes

## Description

Closes #5617

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1374,7 +1374,9 @@ a metaclass class method.",
         slots_list: List[nodes.NodeNG],
     ) -> None:
         """Check if `node` redefines a slot which is defined in an ancestor class"""
-        slots_names = [getattr(slot, "value", "") for slot in slots_list]
+        slots_names = [
+            slot.value for slot in slots_list if isinstance(slot, nodes.Const)
+        ]
 
         # Slots of all parent classes
         ancestors_slots_names = {
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1375,7 +1375,8 @@ a metaclass class method.",
     ) -> None:
         """Check if `node` redefines a slot which is defined in an ancestor class"""
         slots_names = [
-            slot.value for slot in slots_list if isinstance(slot, nodes.Const)
+            slot.value if isinstance(slot, nodes.Const) else next(slot.infer()).value
+            for slot in slots_list
         ]
 
         # Slots of all parent classes
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -588,7 +588,7 @@ MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
         "subclassed-final-class",
         "Used when a class decorated with typing.final has been subclassed.",
     ),
-    "W0241": (
+    "W0244": (
         "Redefined slots %r in subclass",
         "redefined-slots-in-subclass",
         "Used when a slot is re-defined in a subclass.",
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -52,7 +52,7 @@
 """Classes checker for Python code"""
 import collections
 from itertools import chain, zip_longest
-from typing import List, Pattern, Set
+from typing import List, Pattern
 
 import astroid
 from astroid import nodes
@@ -1332,7 +1332,7 @@ a metaclass class method.",
                 node=function_node,
             )
 
-    def _check_slots(self, node: nodes.ClassDef):
+    def _check_slots(self, node: nodes.ClassDef) -> None:
         if "__slots__" not in node.locals:
             return
         for slots in node.igetattr("__slots__"):
@@ -1371,17 +1371,17 @@ a metaclass class method.",
         slots_list: List[nodes.NodeNG],
     ) -> None:
         """Check if `node` redefines a slot which is defined in an ancestor class"""
-        slots_names: List[str] = [getattr(slot, "value", "") for slot in slots_list]
+        slots_names = [getattr(slot, "value", "") for slot in slots_list]
 
         # Slots of all parent classes
-        ancestors_slots_names: Set[str] = {
+        ancestors_slots_names = {
             slot.value
             for ancestor in node.local_attr_ancestors("__slots__")
             for slot in ancestor.slots() or []
         }
 
         # Slots which are common to `node` and its parent classes
-        redefined_slots: Set[str] = ancestors_slots_names.intersection(slots_names)
+        redefined_slots = ancestors_slots_names.intersection(slots_names)
 
         if redefined_slots:
             self.add_message(
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1375,7 +1375,9 @@ a metaclass class method.",
     ) -> None:
         """Check if `node` redefines a slot which is defined in an ancestor class"""
         slots_names = [
-            slot.value if isinstance(slot, nodes.Const) else safe_infer(slot).value
+            slot.value
+            if isinstance(slot, nodes.Const)
+            else getattr(safe_infer(slot), "value", "")
             for slot in slots_list
         ]
 
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1375,7 +1375,7 @@ a metaclass class method.",
     ) -> None:
         """Check if `node` redefines a slot which is defined in an ancestor class"""
         slots_names = [
-            slot.value if isinstance(slot, nodes.Const) else next(slot.infer()).value
+            slot.value if isinstance(slot, nodes.Const) else safe_infer(slot).value
             for slot in slots_list
         ]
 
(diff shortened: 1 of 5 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -1356,6 +1362,33 @@ def _check_slots(self, node):
...
+            self._check_redefined_slots(node, slots, values)
+
+    def _check_redefined_slots(
+        self,
+        node: nodes.ClassDef,
+        slots_node: nodes.NodeNG,
+        slots_list: List[nodes.NodeNG],
+    ) -> None:
+        """Check if `node` redefines a slot which is defined in an ancestor class"""
+        slots_names = [getattr(slot, "value", None) for slot in slots_list]
```
mbyrnepr2, line 1374: Using `getattr` here instead of `slot.value` as a test scenario fails which uses `deque.__name__` as a slot.

pylint/checkers/classes/class_checker.py
```diff
@@ -1326,7 +1332,7 @@ def _check_invalid_overridden_method(self, function_node, parent_function_node):
                 node=function_node,
             )
 
-    def _check_slots(self, node):
+    def _check_slots(self, node: nodes.ClassDef):
```
DanielNoord, line 1335:
```suggestion
    def _check_slots(self, node: nodes.ClassDef) -> None:
```

pylint/checkers/classes/class_checker.py
```diff
@@ -1356,6 +1362,33 @@ def _check_slots(self, node):
...
+            self._check_redefined_slots(node, slots, values)
+
+    def _check_redefined_slots(
+        self,
+        node: nodes.ClassDef,
+        slots_node: nodes.NodeNG,
+        slots_list: List[nodes.NodeNG],
+    ) -> None:
+        """Check if `node` redefines a slot which is defined in an ancestor class"""
+        slots_names: List[str] = [getattr(slot, "value", "") for slot in slots_list]
```
DanielNoord, line 1374:
```suggestion
        slots_names = [getattr(slot, "value", "") for slot in slots_list]
```

pylint/checkers/classes/class_checker.py
```diff
@@ -1356,6 +1362,33 @@ def _check_slots(self, node):
...
+
+        # Slots of all parent classes
+        ancestors_slots_names: Set[str] = {
```
DanielNoord, line 1377:
```suggestion
        ancestors_slots_names = {
```

pylint/checkers/classes/class_checker.py
```diff
@@ -1356,6 +1362,33 @@ def _check_slots(self, node):
...
+
+    def _check_redefined_slots(
+        self,
+        node: nodes.ClassDef,
+        slots_node: nodes.NodeNG,
+        slots_list: List[nodes.NodeNG],
+    ) -> None:
+        """Check if `node` redefines a slot which is defined in an ancestor class"""
+        slots_names: List[str] = [getattr(slot, "value", "") for slot in slots_list]
+
```
DanielNoord, line 1375: I haven't looked at the internals, but how can `slot.value` be nothing?

pylint/checkers/classes/class_checker.py
```diff
@@ -1356,6 +1362,33 @@ def _check_slots(self, node):
...
+        # Slots of all parent classes
+        ancestors_slots_names: Set[str] = {
+            slot.value
+            for ancestor in node.local_attr_ancestors("__slots__")
+            for slot in ancestor.slots() or []
+        }
+
+        # Slots which are common to `node` and its parent classes
+        redefined_slots: Set[str] = ancestors_slots_names.intersection(slots_names)
```
DanielNoord, line 1384:
```suggestion
        redefined_slots = ancestors_slots_names.intersection(slots_names)
```

DanielNoord (changes requested):
Small changes about implicit typing and one question.

Rest looks good, thanks @mbyrnepr2

pylint/checkers/classes/class_checker.py
mbyrnepr2: Thanks for the question @DanielNoord. Here is a demo showing when `deque.__name__` has no value attribute:

```
(venv310) cat demo.py 
from collections import deque

class Base:
    __slots__ = "a", "b", "__dict__", deque.__name__
(venv310)
(venv310) pylint demo.py 
> /Users/markbyrne/programming/pylint/pylint/checkers/classes/class_checker.py(1375)_check_redefined_slots()
-> slots_names = [getattr(slot, "value", "") for slot in slots_list]
(Pdb) slots_list
[<Const.str l.3 at 0x10a4a52a0>, <Const.str l.3 at 0x10a4a50f0>, <Const.str l.3 at 0x10a4a5210>, <Attribute.__name__ l.3 at 0x10a4a5240>]
(Pdb) slots_list[3]
<Attribute.__name__ l.3 at 0x10a4a5240>
(Pdb) slots_list[3].value
*** AttributeError: 'Attribute' object has no attribute 'value'
(Pdb) 
```

DanielNoord: This does answer my question, but creates a follow-up 😄 

Shouldn't we do something with `deque.__name__`? As in, extract `__name__` or something from it? If a subclass redefines `deque.__name__` it should raise as well right?

This `getattr` is a different way of doing `isinstance(slot, nodes.Const)` which I find a bit more explicit. So whatever we decide on for `deque.__name__` personally I would prefer rewriting as a `Const` check.

pylint/checkers/classes/class_checker.py
```diff
@@ -1375,7 +1375,8 @@ def _check_redefined_slots(
     ) -> None:
         """Check if `node` redefines a slot which is defined in an ancestor class"""
         slots_names = [
-            slot.value for slot in slots_list if isinstance(slot, nodes.Const)
+            slot.value if isinstance(slot, nodes.Const) else next(slot.infer()).value
```
DanielNoord, line 1378:
```suggestion
            slot.value if isinstance(slot, nodes.Const) else utils.safe_infer(slot).value
```

mbyrnepr2: Using `getattr` since have the following from mypy otherwise:
`pylint/checkers/classes/class_checker.py:1378: error: Item "None" of "Optional[Any]" has no attribute "value"  [union-attr]`

pylint/checkers/classes/class_checker.py
```diff
@@ -1359,6 +1365,38 @@ def _check_slots(self, node):
...
+        slots_node: nodes.NodeNG,
+        slots_list: List[nodes.NodeNG],
+    ) -> None:
+        """Check if `node` redefines a slot which is defined in an ancestor class"""
+        slots_names = [
+            slot.value
+            if isinstance(slot, nodes.Const)
+            else getattr(safe_infer(slot), "value", "")
+            for slot in slots_list
+        ]
```
DanielNoord, line 1382:
```suggestion
        slots_names: List[str] = []
        for slot in slots_list:
            if isinstance(slot, nodes.Const):
                slots_names.append(slot.value)
            else:
                inferred_slot = safe_infer(slot)
                if inferred_slot:
                    slots_names.append(inferred_slot.value)
```

It's a bit longer, but I prefer how explicit and readable this is. Let me know what you think!

pylint/checkers/classes/class_checker.py
```diff
@@ -588,6 +588,11 @@ def _has_same_layout_slots(slots, assigned_value):
         "subclassed-final-class",
         "Used when a class decorated with typing.final has been subclassed.",
     ),
+    "W0241": (
```
DanielNoord, line 591:
```suggestion
    "W0244": (
```

I just noticed that there is `E0241`. I know we don't really enforce this right now, but I think we're going to move to making the last two digits exclusive within a checker class.

DanielNoord (approved):
One final change which I should have caught earlier.

Thanks, @mbyrnepr2! Sorry for the back and forth on that list comprehension, but I think we got a good final result 👍

pylint/checkers/classes/class_checker.py
mbyrnepr2: I'm also fine with that. Thanks @DanielNoord

mbyrnepr2:
> One final change which I should have caught earlier.
Thanks, @mbyrnepr2! Sorry for the back and forth on that list comprehension, but I think we got a good final result 👍

No, thank you for your attention to detail! Updated the number btw.

Pierre-Sassoulas (approved):
👍 Good to go in 2.13.0 :)

PR #5641, opened 2022-01-05
Fix #5638: Allow for encoding to be supplied as a positional argument

### LINKED ISSUE #5638: unspecified-encoding error when the encoding is specified with positional argument
### Bug description

*test.py*
```py
from pathlib import Path
Path("test.txt").write_text("中文")
Path("test2.txt").write_text("中文", "utf8")
Path("test3.txt").write_text("中文", encoding="utf8")
```
Get an error on L3 even the encoding is already specified. I think this also applies to `read_text()`.

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:2:0: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
test.py:3:0: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
```

### Expected behavior

L3 should not raise `unspecified-encoding`.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.2
Python 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 22:39:24) [MSC v.1916 32 bit (Intel)]
```

### OS / Environment

Windows 10 64-bit / CMD

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Allows `encoding` to be supplied as a positional argument to calls opening files without triggering `unspecified-encoding`. The positional argument was identified correctly only for `read_text()` (see line 63 of test file).

Closes #5638

### DIFF
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -640,9 +640,17 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
                     encoding_arg = utils.get_argument_from_call(
                         node, position=0, keyword="encoding"
                     )
+                elif open_module == "pathlib" and node.func.attrname == "write_text":
+                    encoding_arg = utils.get_argument_from_call(
+                        node, position=1, keyword="encoding"
+                    )
+                elif open_module == "pathlib":
+                    encoding_arg = utils.get_argument_from_call(
+                        node, position=2, keyword="encoding"
+                    )
                 else:
                     encoding_arg = utils.get_argument_from_call(
-                        node, position=None, keyword="encoding"
+                        node, position=3, keyword="encoding"
                     )
             except utils.NoSuchArgumentError:
                 self.add_message("unspecified-encoding", node=node)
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
Tests are on point :thumbsup: !

pylint/checkers/stdlib.py
```diff
@@ -640,9 +640,17 @@ def _check_open_encoded(self, node: nodes.Call, open_module: str) -> None:
                     encoding_arg = utils.get_argument_from_call(
                         node, position=0, keyword="encoding"
                     )
+                elif open_module == "pathlib" and node.func.attrname == "write_text":
+                    encoding_arg = utils.get_argument_from_call(
```
DanielNoord, line 644: Shall we merge these `if` for `path lib` and then do the `attrname` one. I'm trying to count `if`-calls and its getting late but I think we can reduce the number checks needed to get into L648 from 3 to 2 if you understand what I mean 😅

DanielNoord (approved):

Issue #5527, opened 2021-12-14
Integrate content from ``pylint-errors`` into the new pages for message


Pierre-Sassoulas:
### Current problem

We'll have a single documentation page for each message following #5396 but we need to add the content.

### Desired solution

Start by taking what exists in ``pylint-errors``, the owner is okay with it and [the license is MIT](https://github.com/vald-phoenix/pylint-errors/blob/master/LICENSE)

- [x] Proof of concept of a structure and linking for the individual pages (#5396)
- [x] Fix small issue with two titles that are the same (#5740)
- [ ] Merging git history of pylnt-error (or simply copy pasting) and moving all the relevant content in  ``doc/data/messages/original``
- [ ] Create a script to migrate pylint-errors format to our structure
- [ ] Add and check messages by manageable batch until there is nothing left of the "original" pylint-errors files
- [ ] Add functional tests for ``good.py``/``bad.py``

### Additional context

https://github.com/PyCQA/pylint/pull/5396#issuecomment-983427450

DanielNoord:
#5396 is more about creating a structure and linking for the individual pages. I think this issue and associated PRs should cover how we are going to store the data for the messages pages. Later PRs can then cover writing the actual data.

Based on the work of @vald-phoenix for [pylint-errors](https://github.com/vald-phoenix/pylint-errors) I think each page should (ideally) cover the following sections:
1. Rationale
2. Good code example
3. Bad code example
4. Development relevant information (checker name, file link, etc.)
5. Related sources

The first message on `pylint-errors` already contains most of this data for example, see [example](https://github.com/vald-phoenix/pylint-errors/blob/master/plerr/errors/async/E1700.md). We should definitely base the content of our pages on the data already collected there.

However, before doing the "data collection" part we should consider how to store these sections. I thought some more about integrating the code examples with the functional tests, but I think that is likely going to become a big mess. Taking the message I linked as an example: the good and bad code takes up 7 lines, whereas in our test suite this is 17 lines. For other messages this will be much higher. The file for `arguments-differ` is 320+ lines at the moment. I don't see how we can dim this done effectively while also maintaining a good test suite without substantial time investment.
Thus, I don't think we should mix the tests with the documentation code.

What I would propose is to create a new folder `doc/data/messages`. Within these each message name (so `yield-inside-async-function`, not `E1700`) get its own directory. This directory can contain 4 files:
[...]

Issue #5466, opened 2021-12-03
Better error ranges for classes and functions


cdce8p:
The error ranges for `Class` and `Function` nodes should be improved. At the moment they span the whole length of the corresponding block which isn't ideal, especially once error range support in VSCode is added.

Ideally, those should only highlight the actual class or function name.

```py
def func2():
    return

class SomeCls:
    def __init__(self):
        """"""
        pass


def some_decorator_function():
    """some docstring"""

class invalid_class_name:
    pass

@some_decorator_function()
def invalidFunctionName():
    pass
```

<img width="369" alt="Screen Shot 2021-12-03 at 19 14 28" src="https://user-images.githubusercontent.com/30130371/144652403-d078e8ea-3188-46b5-8d83-f7e06f102f55.png">

<img width="588" alt="Screen Shot 2021-12-03 at 19 17 01" src="https://user-images.githubusercontent.com/30130371/144652757-a9c85e04-4676-4ea8-b4fe-7365b80a903a.png">

DanielNoord:
@cdce8p Do you want to fix this by changing the `end_line` for `nodes.ClassDef` in `astroid` or do a check in `add_message` like:
```python
if isinstance(node, nodes.ClassDef):
    end_lineno = lineno
```

Pierre-Sassoulas:
I think it should be done in pylint: the end lineno from astroid is accurate and useful (maybe used in other libs toi). The current issue is linked to the way pylint needs to display errors for class and function.

cdce8p:
I agree, we shouldn't change the `end_line` attributes itself. They are still correct. What we need is some better way to indicate at which range the actual class / function name is defined. We can't use the AST for it, unfortunately.

Thinking out loud for a moment, it should be somewhat strait forward to do some basic string parsing. Maybe it would make sense to modify the `astroid/rebuilder.py`, ie. add an additional field for it. That's a long shoot though.

For a start, it would already be an improvement to modify `add_message` so the error is only emitted for the first line. It won't be perfect but definitely an improvement.

kovla:
Highlighting entire code block is very distracting, to be honest. A lot of unnecessary highlighting is done to communicate something very simple or something that the developer choses not to change.  

Old behavior (highlight just the beginning of the affected line) was concise and effective enough. Please at least provide an option to change this behavior via settings. Thanks!

Pierre-Sassoulas:
[...]

PR #5643, opened 2022-01-06
Improve non ascii checker

CarliJoy:

## Type of Changes

## Description

Add ``non-ascii-identifier`` as replacement ``non-ascii-name``   to ensure really __all__ Python names are ASCII.   Checker now checks properly the names of imports (``non-ascii-module-import``) as  well for as of file names (``non-ascii-file-name``).  Non ASCII characters could be homoglyphs (look alike characters) and hard to  enter on a non specialized keyboard.

->  See [Confusable Characters in PEP 672](https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,14 @@ Release date: TBA
 
   Closes #5588
 
+* Add ``non-ascii-identifier`` as replacement ``non-ascii-name``
+   to ensure really __all__ Python names are ASCII.
+   Checker now checks properly the names of imports (``non-ascii-module-import``) as
+   well for as of file names (``non-ascii-file-name``).
+   Non ASCII characters could be homoglyphs (look alike characters) and hard to
+   enter on a non specialized keyboard.
+   See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_
+
 * When run in parallel mode ``pylint`` now pickles the data passed to subprocesses with
   the ``dill`` package. The ``dill`` package has therefore been added as a dependency.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,7 +14,7 @@ Release date: TBA
 
   Closes #5588
 
-* Renamed  ``non-ascii-name`` to ``non-ascii-identifier`` and rewrote the checker.
+* Rewrote checker for ``non-ascii-name``.
    It now ensures __all__ Python names are ASCII and also properly
    checks the names of imports (``non-ascii-module-import``) as
    well as file names (``non-ascii-file-name``) and emits their respective new warnings.
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,10 +14,11 @@ Release date: TBA
 
   Closes #5588
 
-* Add ``non-ascii-identifier`` as replacement ``non-ascii-name``
-   to ensure really __all__ Python names are ASCII.
-   Checker now checks properly the names of imports (``non-ascii-module-import``) as
-   well for as of file names (``non-ascii-file-name``).
+* Renamed  ``non-ascii-name`` to ``non-ascii-identifier`` and rewrote the checker.
+   It now ensures __all__ Python names are ASCII and also properly
+   checks the names of imports (``non-ascii-module-import``) as
+   well as file names (``non-ascii-file-name``) and emits their respective new warnings.
+
    Non ASCII characters could be homoglyphs (look alike characters) and hard to
    enter on a non specialized keyboard.
    See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -25,7 +25,6 @@ New checkers
 
 Removed checkers
 ================
-* ``non-ascii-name`` has been renamed to ``non-ascii-identifier``
 
 Extensions
 ==========
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -14,8 +14,17 @@ New checkers
 
   Closes #5460
 
+* Add ``non-ascii-identifier`` as replacement ``non-ascii-name``
+   to ensure really __all__ Python names are ASCII.
+   Checker now checks properly the names of imports (``non-ascii-module-import``) as
+   well for as of file names (``non-ascii-file-name``).
+   Non ASCII characters could be homoglyphs (look alike characters) and hard to
+   enter on a non specialized keyboard.
+   See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_
+
 Removed checkers
 ================
+* ``non-ascii-name`` has replaced by ``non-ascii-identifier``
 
 Extensions
 ==========
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -14,17 +14,18 @@ New checkers
 
   Closes #5460
 
-* Add ``non-ascii-identifier`` as replacement ``non-ascii-name``
-   to ensure really __all__ Python names are ASCII.
-   Checker now checks properly the names of imports (``non-ascii-module-import``) as
-   well for as of file names (``non-ascii-file-name``).
+* Renamed  ``non-ascii-name`` to ``non-ascii-identifier`` and rewrote the checker.
+   It now ensures __all__ Python names are ASCII and also properly
+   checks the names of imports (``non-ascii-module-import``) as
+   well as file names (``non-ascii-file-name``) and emits their respective new warnings.
+
    Non ASCII characters could be homoglyphs (look alike characters) and hard to
    enter on a non specialized keyboard.
    See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_
 
 Removed checkers
 ================
-* ``non-ascii-name`` has replaced by ``non-ascii-identifier``
+* ``non-ascii-name`` has been renamed to ``non-ascii-identifier``
 
 Extensions
 ==========
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -14,7 +14,7 @@ New checkers
 
   Closes #5460
 
-* Renamed  ``non-ascii-name`` to ``non-ascii-identifier`` and rewrote the checker.
+* Rewrote Checker of ``non-ascii-name``.
    It now ensures __all__ Python names are ASCII and also properly
    checks the names of imports (``non-ascii-module-import``) as
    well as file names (``non-ascii-file-name``) and emits their respective new warnings.
--- a/pylint/checkers/__init__.py
+++ b/pylint/checkers/__init__.py
@@ -39,7 +39,14 @@ Base id of standard checkers (used in msg and report ids):
 15: stdlib
 16: python3
 17: refactoring
-18-50: not yet used: reserved for future internal checkers.
+.
+.
+.
+25: non-ascii-names
+25-50: not yet used: reserved for future internal checkers.
+This file is not updated. Use
+   script/get_unused_message_id_category.py
+to get the next free checker id.
 51-99: perhaps used: reserved for external checkers
 
 The raw_metrics checker has no number associated since it doesn't emit any
--- a/pylint/checkers/__init__.py
+++ b/pylint/checkers/__init__.py
@@ -42,7 +42,7 @@ Base id of standard checkers (used in msg and report ids):
 .
 .
 .
-25: non-ascii-names
+24: non-ascii-names
 25-50: not yet used: reserved for future internal checkers.
 This file is not updated. Use
    script/get_unused_message_id_category.py
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -73,7 +73,7 @@ import collections
 import itertools
 import re
 import sys
-from typing import TYPE_CHECKING, Any, Dict, Iterator, Optional, Pattern, cast
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, Optional, Pattern, cast
 
 import astroid
 from astroid import nodes
@@ -1719,7 +1719,31 @@ def _create_naming_options():
     return tuple(name_options)
 
 
-class NameChecker(_BasicChecker):
+class NameCheckerHelper:
+    """Class containing functions required by NameChecker and NonAsciiNamesChecker"""
+
+    def _check_name(
+        self, node_type: str, name: str, node: nodes.NodeNG, optional_kwarg: Any = None
+    ):
+        """Only Dummy function will be overwritten by implementing classes
+
+        Note: kwarg arguments will be different in implementing classes
+        """
+        raise NotImplementedError
+
+    def _recursive_check_names(self, args: Iterable[nodes.AssignName]):
+        """check names in a possibly recursive list <arg>"""
+        for arg in args:
+            if isinstance(arg, nodes.AssignName):
+                self._check_name("argument", arg.name, arg)
+            else:
+                # pylint: disable-next=fixme
+                # TODO: Check if we can remove this if branch because of
+                #       the up to date astroid version used
+                self._recursive_check_names(arg.elts)
+
+
+class NameChecker(_BasicChecker, NameCheckerHelper):
     msgs = {
         "C0103": (
             '%s name "%s" doesn\'t conform to %s',
@@ -1979,14 +2003,6 @@ class NameChecker(_BasicChecker):
                 else:
                     self._check_name("class_attribute", node.name, node)
 
-    def _recursive_check_names(self, args):
-        """check names in a possibly recursive list <arg>"""
-        for arg in args:
-            if isinstance(arg, nodes.AssignName):
-                self._check_name("argument", arg.name, arg)
-            else:
-                self._recursive_check_names(arg.elts)
-
     def _find_name_group(self, node_type):
         return self._name_group.get(node_type, node_type)
 
@@ -2027,6 +2043,7 @@ class NameChecker(_BasicChecker):
             pattern.match(name) for pattern in self._bad_names_rgxs_compiled
         )
 
+    # pylint: disable-next=arguments-renamed
     def _check_name(self, node_type, name, node, confidence=interfaces.HIGH):
         """check for a name using the type's regexp"""
 
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -1737,11 +1737,6 @@ class NameChecker(_BasicChecker):
                 ]
             },
         ),
-        "C0144": (
-            '%s name "%s" contains a non-ASCII unicode character',
-            "non-ascii-name",
-            "Used when the name contains at least one non-ASCII unicode character.",
-        ),
         "W0111": (
             "Name %s will become a keyword in Python %s",
             "assign-to-new-keyword",
@@ -1838,7 +1833,6 @@ class NameChecker(_BasicChecker):
         self._name_hints = {}
         self._good_names_rgxs_compiled = []
         self._bad_names_rgxs_compiled = []
-        self._non_ascii_rgx_compiled = re.compile("[^\u0000-\u007F]")
 
     def open(self):
         self.linter.stats.reset_bad_names()
@@ -1878,7 +1872,7 @@ class NameChecker(_BasicChecker):
 
         return regexps, hints
 
-    @utils.check_messages("disallowed-name", "invalid-name", "non-ascii-name")
+    @utils.check_messages("disallowed-name", "invalid-name")
     def visit_module(self, node: nodes.Module) -> None:
         self._check_name("module", node.name.split(".")[-1], node)
         self._bad_names = {}
@@ -1904,9 +1898,7 @@ class NameChecker(_BasicChecker):
             for args in warnings:
                 self._raise_name_warning(prevalent_group, *args)
 
-    @utils.check_messages(
-        "disallowed-name", "invalid-name", "assign-to-new-keyword", "non-ascii-name"
-    )
+    @utils.check_messages("disallowed-name", "invalid-name", "assign-to-new-keyword")
     def visit_classdef(self, node: nodes.ClassDef) -> None:
         self._check_assign_to_new_keyword_violation(node.name, node)
         self._check_name("class", node.name, node)
@@ -1914,9 +1906,7 @@ class NameChecker(_BasicChecker):
             if not any(node.instance_attr_ancestors(attr)):
                 self._check_name("attr", attr, anodes[0])
 
-    @utils.check_messages(
-        "disallowed-name", "invalid-name", "assign-to-new-keyword", "non-ascii-name"
-    )
+    @utils.check_messages("disallowed-name", "invalid-name", "assign-to-new-keyword")
     def visit_functiondef(self, node: nodes.FunctionDef) -> None:
         # Do not emit any warnings if the method is just an implementation
         # of a base class method.
@@ -1944,14 +1934,12 @@ class NameChecker(_BasicChecker):
 
     visit_asyncfunctiondef = visit_functiondef
 
-    @utils.check_messages("disallowed-name", "invalid-name", "non-ascii-name")
+    @utils.check_messages("disallowed-name", "invalid-name")
     def visit_global(self, node: nodes.Global) -> None:
         for name in node.names:
             self._check_name("const", name, node)
 
-    @utils.check_messages(
-        "disallowed-name", "invalid-name", "assign-to-new-keyword", "non-ascii-name"
-    )
+    @utils.check_messages("disallowed-name", "invalid-name", "assign-to-new-keyword")
     def visit_assignname(self, node: nodes.AssignName) -> None:
         """check module level assigned names"""
         self._check_assign_to_new_keyword_violation(node.name, node)
@@ -2041,11 +2029,6 @@ class NameChecker(_BasicChecker):
 
     def _check_name(self, node_type, name, node, confidence=interfaces.HIGH):
         """check for a name using the type's regexp"""
-        non_ascii_match = self._non_ascii_rgx_compiled.match(name)
-        if non_ascii_match is not None:
-            self._raise_name_warning(
-                None, node, node_type, name, confidence, warning="non-ascii-name"
-            )
 
         def _should_exempt_from_invalid_name(node):
             if node_type == "variable":
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -73,14 +73,14 @@ import collections
 import itertools
 import re
 import sys
-from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, Optional, Pattern, cast
+from typing import TYPE_CHECKING, Any, Dict, Iterator, Optional, Pattern, cast
 
 import astroid
 from astroid import nodes
 
-from pylint import checkers, constants, interfaces
+from pylint import constants, interfaces
 from pylint import utils as lint_utils
-from pylint.checkers import utils
+from pylint.checkers import base_checker, utils
 from pylint.checkers.utils import (
     infer_all,
     is_overload_stub,
@@ -456,12 +456,7 @@ def redefined_by_decorator(node):
     return False
 
 
-class _BasicChecker(checkers.BaseChecker):
-    __implements__ = interfaces.IAstroidChecker
-    name = "basic"
-
-
-class BasicErrorChecker(_BasicChecker):
+class BasicErrorChecker(base_checker._BasicChecker):
     msgs = {
         "E0100": (
             "__init__ method is a generator",
@@ -937,7 +932,7 @@ class BasicErrorChecker(_BasicChecker):
             )
 
 
-class BasicChecker(_BasicChecker):
+class BasicChecker(base_checker._BasicChecker):
     """checks for :
     * doc strings
     * number of arguments, local variables, branches, returns and statements in
@@ -1719,31 +1714,7 @@ def _create_naming_options():
     return tuple(name_options)
 
 
-class NameCheckerHelper:
-    """Class containing functions required by NameChecker and NonAsciiNamesChecker"""
-
-    def _check_name(
-        self, node_type: str, name: str, node: nodes.NodeNG, optional_kwarg: Any = None
-    ):
-        """Only Dummy function will be overwritten by implementing classes
-
-        Note: kwarg arguments will be different in implementing classes
-        """
-        raise NotImplementedError
-
-    def _recursive_check_names(self, args: Iterable[nodes.AssignName]):
-        """Check names in a possibly recursive list <arg>"""
-        for arg in args:
-            if isinstance(arg, nodes.AssignName):
-                self._check_name("argument", arg.name, arg)
-            else:
-                # pylint: disable-next=fixme
-                # TODO: Check if we can remove this if branch because of
-                #       the up to date astroid version used
-                self._recursive_check_names(arg.elts)
-
-
-class NameChecker(_BasicChecker, NameCheckerHelper):
+class NameChecker(base_checker._NameCheckerBase):
     msgs = {
         "C0103": (
             '%s name "%s" doesn\'t conform to %s',
@@ -2043,7 +2014,6 @@ class NameChecker(_BasicChecker, NameCheckerHelper):
             pattern.match(name) for pattern in self._bad_names_rgxs_compiled
         )
 
-    # pylint: disable-next=arguments-renamed
     def _check_name(self, node_type, name, node, confidence=interfaces.HIGH):
         """check for a name using the type's regexp"""
 
@@ -2092,7 +2062,7 @@ class NameChecker(_BasicChecker, NameCheckerHelper):
         return None
 
 
-class DocStringChecker(_BasicChecker):
+class DocStringChecker(base_checker._BasicChecker):
     msgs = {
         "C0112": (
             "Empty %s docstring",
@@ -2258,7 +2228,7 @@ class DocStringChecker(_BasicChecker):
             )
 
 
-class PassChecker(_BasicChecker):
+class PassChecker(base_checker._BasicChecker):
     """check if the pass statement is really necessary"""
 
     msgs = {
@@ -2300,7 +2270,7 @@ def _infer_dunder_doc_attribute(node):
     return docstring.value
 
 
-class ComparisonChecker(_BasicChecker):
+class ComparisonChecker(base_checker._BasicChecker):
     """Checks for comparisons
 
     - singleton comparison: 'expr == True', 'expr == False' and 'expr == None'
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -78,9 +78,9 @@ from typing import TYPE_CHECKING, Any, Dict, Iterator, Optional, Pattern, cast
 import astroid
 from astroid import nodes
 
-from pylint import constants, interfaces
+from pylint import checkers, constants, interfaces
 from pylint import utils as lint_utils
-from pylint.checkers import base_checker, utils
+from pylint.checkers import utils
 from pylint.checkers.utils import (
     infer_all,
     is_overload_stub,
@@ -456,7 +456,12 @@ def redefined_by_decorator(node):
     return False
 
 
-class BasicErrorChecker(base_checker._BasicChecker):
+class _BasicChecker(checkers.BaseChecker):
+    __implements__ = interfaces.IAstroidChecker
+    name = "basic"
+
+
+class BasicErrorChecker(_BasicChecker):
     msgs = {
         "E0100": (
             "__init__ method is a generator",
@@ -932,7 +937,7 @@ class BasicErrorChecker(base_checker._BasicChecker):
             )
 
 
-class BasicChecker(base_checker._BasicChecker):
+class BasicChecker(_BasicChecker):
     """checks for :
     * doc strings
     * number of arguments, local variables, branches, returns and statements in
@@ -1714,7 +1719,7 @@ def _create_naming_options():
     return tuple(name_options)
 
 
-class NameChecker(base_checker._NameCheckerBase):
+class NameChecker(_BasicChecker):
     msgs = {
         "C0103": (
             '%s name "%s" doesn\'t conform to %s',
@@ -1974,6 +1979,14 @@ class NameChecker(base_checker._NameCheckerBase):
                 else:
                     self._check_name("class_attribute", node.name, node)
 
+    def _recursive_check_names(self, args):
+        """check names in a possibly recursive list <arg>"""
+        for arg in args:
+            if isinstance(arg, nodes.AssignName):
+                self._check_name("argument", arg.name, arg)
+            else:
+                self._recursive_check_names(arg.elts)
+
     def _find_name_group(self, node_type):
         return self._name_group.get(node_type, node_type)
 
@@ -2062,7 +2075,7 @@ class NameChecker(base_checker._NameCheckerBase):
         return None
 
 
-class DocStringChecker(base_checker._BasicChecker):
+class DocStringChecker(_BasicChecker):
     msgs = {
         "C0112": (
             "Empty %s docstring",
@@ -2228,7 +2241,7 @@ class DocStringChecker(base_checker._BasicChecker):
             )
 
 
-class PassChecker(base_checker._BasicChecker):
+class PassChecker(_BasicChecker):
     """check if the pass statement is really necessary"""
 
     msgs = {
@@ -2270,7 +2283,7 @@ def _infer_dunder_doc_attribute(node):
     return docstring.value
 
 
-class ComparisonChecker(base_checker._BasicChecker):
+class ComparisonChecker(_BasicChecker):
     """Checks for comparisons
 
     - singleton comparison: 'expr == True', 'expr == False' and 'expr == None'
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -1732,7 +1732,7 @@ class NameCheckerHelper:
         raise NotImplementedError
 
     def _recursive_check_names(self, args: Iterable[nodes.AssignName]):
-        """check names in a possibly recursive list <arg>"""
+        """Check names in a possibly recursive list <arg>"""
         for arg in args:
             if isinstance(arg, nodes.AssignName):
                 self._check_name("argument", arg.name, arg)
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -19,10 +19,11 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 import functools
 from inspect import cleandoc
-from typing import Any, Optional
+from typing import Any, Iterable, Optional
 
 from astroid import nodes
 
+from pylint import interfaces
 from pylint.config import OptionsProviderMixIn
 from pylint.constants import _MSG_ORDER, WarningScope
 from pylint.exceptions import InvalidMessageError
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
     def process_tokens(self, tokens):
         """Should be overridden by subclasses."""
         raise NotImplementedError()
+
+
+class _BasicChecker(BaseChecker):
+    __implements__ = interfaces.IAstroidChecker
+    name = "basic"
+
+
+class _NameCheckerBase(_BasicChecker):
+    """Class containing functions required by NameChecker and NonAsciiNameChecker"""
+
+    def _check_name(
+        self, node_type: str, name: str, node: nodes.NodeNG, confidence=interfaces.HIGH
+    ):
+        """Only Dummy function will be overwritten by implementing classes
+
+        Note: kwarg arguments will be different in implementing classes
+        """
+        raise NotImplementedError
+
+    def _recursive_check_names(self, args: Iterable[nodes.AssignName]):
+        """Check names in a possibly recursive list <arg>"""
+        for arg in args:
+            if isinstance(arg, nodes.AssignName):
+                self._check_name("argument", arg.name, arg)
+            else:
+                # pylint: disable-next=fixme
+                # TODO: Check if we can remove this if branch because of
+                #       the up to date astroid version used
+                self._recursive_check_names(arg.elts)
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -19,11 +19,10 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 import functools
 from inspect import cleandoc
-from typing import Any, Iterable, Optional
+from typing import Any, Optional
 
 from astroid import nodes
 
-from pylint import interfaces
 from pylint.config import OptionsProviderMixIn
 from pylint.constants import _MSG_ORDER, WarningScope
 from pylint.exceptions import InvalidMessageError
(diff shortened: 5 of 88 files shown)


CarliJoy:
@DanielNoord and @Pierre-Sassoulas here the splitted version of the ascii name checker.
I did some requested (and some non requested) improvements.
These improvements are within the second commit to allow easier reviewing.
The first commit is simple copy/paste from #5311 

About the `_is_ascii_only` attribute for the nodes. I know it is a bit of a hack but I don't know a better way.

Only good solution would be to spend a lot more time into figuring out why some nodes are hit multiple times. But sorry I spent to much time already sorry, I won't do that.

If something should be added within Astroid, I would recommend something like at `was_checked_by: Set[str]`  to NodesNG, just to make it simple.

I would kindly ask you to this for me.

Alternative would be that I keep a list of all checked nodes within the parser and just check "if node in self._checked".
But that feels much more wronge and inefficient as the current hack.

CarliJoy:
BTW: Some test always fail locally for me. So I have no way to check the coverage.
So I don't know which line lost coverage in  [https://coveralls.io/builds/45430446/source?filename=pylint%2Ftestutils%2Flint_module_test.py#L9]

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,264 @@
...
+    msgs = {
+        "C2401": (
+            '%s name "%s" contains a non-ASCII character, rename it.',
+            "non-ascii-identifier",
+            (
+                "Used when the name contains at least one non-ASCII unicode character."
+                "See https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers"
+                " for a background why this could be bad."
+            ),
+            {"old_names": [("C0144", "non-ascii-name")]},
```
Pierre-Sassoulas, line 72:
```suggestion
            "non-ascii-name",
            (
                "Used when the name contains at least one non-ASCII unicode character."
                "See https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers"
                " for a background why this could be bad."
            ),
            {"old_names": [("C0144", "old-non-ascii-name")]},
```
We can rename the msgid and not the symbol with this one simple trick (and avoid the new symbol that is just a synonym altogether)

Pierre-Sassoulas (changes requested):

tests/functional/e/.#emacs_file_lock_redefined_conf.py
```diff
@@ -1,3 +1,3 @@
-# [invalid-name]
+# [invalid-name, non-ascii-file-name]
```
DanielNoord, line 1: What is non-ascii about this file name? The `#`?

tests/functional/n/non/non_ascii_name.py
```diff
@@ -1,6 +1,6 @@
 """ Tests for non-ascii-name checker. """
 
-áéíóú = 4444 # [non-ascii-name]
+áéíóú = 4444 # [non-ascii-identifier]
```
DanielNoord, line 3: Could you rename these files? They no longer reflect the message name.

tests/functional/n/non/non_ascii_name_backward_test_code.py
```diff
@@ -0,0 +1,7 @@
+""" Tests for non-ascii-name checker working as expected """
```
DanielNoord, line 1:
```suggestion
""" Tests for non-ascii-name checker working as expected after a major refactor """
```

tests/functional/n/non/non_ascii_name_backward_test_msg.py
```diff
@@ -0,0 +1,7 @@
+""" Tests for non-ascii-name checker working as expected """
```
DanielNoord, line 1:
```suggestion
""" Tests for non-ascii-name checker working as expected after a major refactor """
```

ChangeLog
```diff
@@ -14,6 +14,14 @@ Release date: TBA
...
   Closes #5588
 
+* Add ``non-ascii-identifier`` as replacement ``non-ascii-name``
+   to ensure really __all__ Python names are ASCII.
+   Checker now checks properly the names of imports (``non-ascii-module-import``) as
+   well for as of file names (``non-ascii-file-name``).
+   Non ASCII characters could be homoglyphs (look alike characters) and hard to
+   enter on a non specialized keyboard.
+   See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_
+
```
DanielNoord, line 24:
```suggestion
* Renamed  ``non-ascii-name`` to ``non-ascii-identifier`` and rewrote the checker.
   It now ensures __all__ Python names are ASCII and also properly
   checks the names of imports (``non-ascii-module-import``) as
   well as file names (``non-ascii-file-name``) and emits their respective new warnings.
   
   Non ASCII characters could be homoglyphs (look alike characters) and hard to
   enter on a non specialized keyboard.
   See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_

```

doc/whatsnew/2.13.rst
```diff
@@ -14,8 +14,17 @@ New checkers
...
   Closes #5460
 
+* Add ``non-ascii-identifier`` as replacement ``non-ascii-name``
+   to ensure really __all__ Python names are ASCII.
+   Checker now checks properly the names of imports (``non-ascii-module-import``) as
+   well for as of file names (``non-ascii-file-name``).
+   Non ASCII characters could be homoglyphs (look alike characters) and hard to
+   enter on a non specialized keyboard.
+   See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_
+
```
DanielNoord, line 24:
```suggestion
* Renamed  ``non-ascii-name`` to ``non-ascii-identifier`` and rewrote the checker.
   It now ensures __all__ Python names are ASCII and also properly
   checks the names of imports (``non-ascii-module-import``) as
   well as file names (``non-ascii-file-name``) and emits their respective new warnings.
   
   Non ASCII characters could be homoglyphs (look alike characters) and hard to
   enter on a non specialized keyboard.
   See `Confusable Characters in PEP 672 <https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers>`_

```

doc/whatsnew/2.13.rst
```diff
@@ -14,8 +14,17 @@ New checkers
...
 Removed checkers
 ================
+* ``non-ascii-name`` has replaced by ``non-ascii-identifier``
```
DanielNoord, line 27:
```suggestion
* ``non-ascii-name`` has been renamed to ``non-ascii-identifier``
```

Pierre-Sassoulas:
> So I have no way to check the coverage. So I don't know which line lost coverage

You need to connect with your github account on coveralls.io, they need your token to display the file or they would get rate limited.

pylint/checkers/base.py
```diff
@@ -1719,7 +1719,31 @@ def _create_naming_options():
     return tuple(name_options)
 
 
-class NameChecker(_BasicChecker):
+class NameCheckerHelper:
```
DanielNoord, line 1722: Not sure about the location of this class. Does `pylint/checkers/base_checker.py` perhaps makes more sense? We need a place to define sub-class of `BaseChecker` that get inherited by other classes in other files..

pylint/checkers/base.py
```diff
@@ -1719,7 +1719,31 @@ def _create_naming_options():
...
+    """Class containing functions required by NameChecker and NonAsciiNamesChecker"""
+
+    def _check_name(
```
DanielNoord, line 1725: I'm not that experienced with this, so @Pierre-Sassoulas probably know better, but shouldn't this be decorated by `abc.abstractmethod`?

pylint/checkers/base.py
```diff
@@ -1719,7 +1719,31 @@ def _create_naming_options():
...
+        self, node_type: str, name: str, node: nodes.NodeNG, optional_kwarg: Any = None
+    ):
+        """Only Dummy function will be overwritten by implementing classes
+
+        Note: kwarg arguments will be different in implementing classes
+        """
+        raise NotImplementedError
+
+    def _recursive_check_names(self, args: Iterable[nodes.AssignName]):
+        """check names in a possibly recursive list <arg>"""
```
DanielNoord, line 1735:
```suggestion
        """Check names in a possibly recursive list <arg>"""
```

pylint/checkers/base.py
```diff
@@ -1719,7 +1719,31 @@ def _create_naming_options():
...
+        for arg in args:
+            if isinstance(arg, nodes.AssignName):
+                self._check_name("argument", arg.name, arg)
+            else:
+                # pylint: disable-next=fixme
+                # TODO: Check if we can remove this if branch because of
```
DanielNoord, line 1741: Which version are you referring to?

pylint/checkers/base.py
```diff
@@ -1719,7 +1719,31 @@ def _create_naming_options():
...
+
+        Note: kwarg arguments will be different in implementing classes
+        """
+        raise NotImplementedError
+
+    def _recursive_check_names(self, args: Iterable[nodes.AssignName]):
+        """check names in a possibly recursive list <arg>"""
+        for arg in args:
+            if isinstance(arg, nodes.AssignName):
+                self._check_name("argument", arg.name, arg)
```
DanielNoord, line 1738: This is always true right?

pylint/checkers/base.py
```diff
@@ -2039,13 +2043,9 @@ def _name_disallowed_by_regex(self, name: str) -> bool:
             pattern.match(name) for pattern in self._bad_names_rgxs_compiled
         )
 
+    # pylint: disable-next=arguments-renamed
```
DanielNoord, line 2046: If we make `confidence` a keyword only parameter on `_check_name` can't we avoid this warning and allow us to add typing to it and the parameter of the other implementation of `_check_name`?

DanielNoord (changes requested):
Is there a reason for adding the `tests/functional/n/non_ascii_name_class/__init__.py` and `tests/functional/n/non_ascii_name/__init__.py` files? I think they can be removed tight?

Could you add a docstring to `tests/functional/n/non_ascii_import/foobar.py` to say what it is testing. I can't make a suggestion because it is empty 😢 

I haven't looked at `pylint/checkers/non_ascii_names.py` and the unittest this time (hopefully I'll get time this evening), but this should help this along again. I'm at 47/69 files viewed and approved right now so we're certainly getting there 😄

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,264 @@
...
+        for name in node.names:
+            self._check_name("const", name, node)
+
+    @pylint.checkers.utils.check_messages("non-ascii-name")
+    def visit_assignname(self, node: nodes.AssignName) -> None:
+        """check module level assigned names"""
+        # The NameChecker from which this Checker originates knows a lot of different
+        # versions of variables, i.e. constants, inline variables etc.
+        # To simplify we use only `variable` here, as we don't need to apply different
+        # rules to different types of variables.
```
DanielNoord, line 203: Thanks for the updated comment, much better!

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,264 @@
...
+            self._check_name("variable", node.name, node)
+        elif isinstance(frame, nodes.FunctionDef):
+            if not hasattr(node, "_is_ascii_only"):
+                # only check if not already done
+                self._check_name("variable", node.name, node)
+        elif isinstance(frame, nodes.ClassDef):
+            self._check_name("attr", node.name, node)
+        else:
+            # Just to make sure we check EVERYTHING (!)
+            self._check_name("variable", node.name, node)
```
DanielNoord, line 219: We seem to covering this according to `coveralls`. Might be worth adding a `breakpoint()` locally and seeing what type of `node` triggers this so we can add another case for it.

pylint/testutils/checker_test_case.py
```diff
@@ -53,10 +59,15 @@ def assertAddsMessages(self, *messages: MessageTest) -> Generator[None, None, No
 
         for expected_msg, gotten_msg in zip(messages, got):
             assert expected_msg.msg_id == gotten_msg.msg_id, msg
-            assert expected_msg.line == gotten_msg.line, msg
             assert expected_msg.node == gotten_msg.node, msg
             assert expected_msg.args == gotten_msg.args, msg
             assert expected_msg.confidence == gotten_msg.confidence, msg
+
+            if ignore_position:
```
DanielNoord, line 66: Is it possible for us to only check `line` and `col_offset` and then skip the other two when `ignore_position` is given?
I think we need at least a `line` and `col_offset` for messages to be displayable in IDEs.

tests/checkers/unittest_non_ascii_name.py
```diff
@@ -0,0 +1,312 @@
...
+                    p_or_kw_not_økay=None,
+                    *,
+                    kw_arg_okay,
+                    kw_arg_not_økay,
+                ):
+                    ...
+            """
+        )
+        assert isinstance(node, nodes.FunctionDef)
+        arguments: nodes.Arguments = node.args
```
DanielNoord, line 39:
```suggestion
        arguments = node.args
```

tests/checkers/unittest_non_ascii_name.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+class TestNonAsciiChecker(pylint.testutils.CheckerTestCase):
+    CHECKER_CLASS = pylint.checkers.non_ascii_names.NonAsciiNamesChecker
+    checker: pylint.checkers.non_ascii_names.NonAsciiNamesChecker
+
+    @pytest.mark.skipif(
+        sys.version_info < (3, 8), reason="requires python3.8 or higher"
+    )
+    def test_kwargs_and_position_only(self):
```
DanielNoord, line 20: Isn't this a duplicate of the the functional test in which we also test the location parameters of the message? If we have access to all four I don't think we also need a test whether we have the correct `node` (we don't for 99% of the other messages).

tests/checkers/unittest_non_ascii_name.py
```diff
@@ -0,0 +1,312 @@
...
+            self.checker.visit_classdef(node)
+
+    def test_import_ignore_star(self):
+        """Special case from xyz import *, where '*' is not A-Za-z0-9 but still valid"""
+        node = astroid.parse("from urllib.parse import *")
+
+        with self.assertAddsMessages():
+            self.walk(node)
+
+    @pytest.mark.parametrize(
```
DanielNoord, line 192: I think it would make sense to convert this into a functional test. Imo, those are a lot better to read and I know they actually test "more" than the unittests. (They almost recreate an instance of `pylint` and run it over the file, whereas the unittests takes some shortcuts to initialise faster. This makes it more prone to missing crucial setup phases).

tests/checkers/unittest_non_ascii_name.py
```diff
@@ -0,0 +1,312 @@
...
+                msg_id="non-ascii-name",
+                node=node,
+                args=("Class", "FooBär"),
+                confidence=pylint.interfaces.HIGH,
+                line=2,
+                col_offset=0,
+                # pylint: disable-next=fixme
+                # TODO: Maybe we have to select something different
+                #       here, as line 3 is not that thing we expect
+                end_line=3,
```
DanielNoord, line 179: I wonder what's happening here.. Is the Ellipsis considered part of the `ClassDef` node?

tests/checkers/unittest_non_ascii_name.py
```diff
@@ -0,0 +1,312 @@
...
+                msg_id="non-ascii-name",
+                node=kwargs[1],
+                args=("Argument", "kw_arg_not_økay"),
+                confidence=pylint.interfaces.HIGH,
+            ),
+            ignore_position=True,
+        ):
+            self.checker.visit_functiondef(node)
+
+    @pytest.mark.parametrize(
```
DanielNoord, line 74: I'm probably missing something here, but what's the added benefit of these over the functional tests you wrote?

DanielNoord (changes requested):
Got to the other files as well!

@CarliJoy Tests look very good and robust. 🎉  I just wonder if we can move some of the remaining unittests to the functional tests as well. I'm asking this not only because I find them more readable, but also since I know they are actually better "tests". They way those are invoked better resembles the way `pylint` is invoked by normal users than the unittests are, which has recently caught a bug that the unittests simply could not in their current setup. This is something we as maintainers/contributors should fix, but in the meantime functional tests are often the way to go.

tests/functional/n/non_ascii_name/non_ascii_name_pos_and_kwonly_function.rc
```diff
@@ -0,0 +1,2 @@
+[testoptions]
+min_pyver=3.8
```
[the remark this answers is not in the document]
CarliJoy: The columns aren't detected correctly in Python 3.8.
Or at least that was the reason for doing it in the other tests.

`foobar.py` was a test file to check imports. It was added by accident.

tests/functional/e/.#emacs_file_lock_redefined_conf.py
CarliJoy: Yep the # is not valid for the regex.

But this might be duplicated to the invalid-name checker.
Maybe we can remove the regex all togehter and only use the Python internal .ascii checker if that's what's wanted. I am not sure.
But looking at it, it makes kind of sense...

CarliJoy: @DanielNoord 
The regex used is `self._non_ascii_rgx_compiled = re.compile("[^A-Za-z0-9_]")`

Maybe we can only use: `Py37Str(check_string).isascii()` -> that is actually what we want to check. Everything else should be catched by `invalid name` probably.
But to be honest looking at the source code for `invalid_name` I don't get it.
You have more background knowledge @DanielNoord, so let me know if we drop the regex for the `non-ascii-*` checker.

pylint/checkers/base.py
CarliJoy: Makes sense to use a linear inheritance and to put in withing base_checker.
I will do this.

pylint/checkers/base.py
CarliJoy: I don't know. I haven't written the code, I just moved it. But it seems to be not up to date anymore, that why I put the TODO in it.
Never change a running system, that you not responsible for and don't understand ;-).

pylint/checkers/base.py
CarliJoy: Again I did not write this code I just moved it, to be able to use it.
I won't change anything there, because I don't understand the impacts. This should be an Issue for people that have written the code/understand it correctly.

pylint/checkers/base.py
CarliJoy: in `non_ascii_checker` you or @Pierre-Sassoulas suggested to remove the confidence parameter and hard code it. Also I needed to introduce there another parameter optional `check_string`, so I can check the names of modules, without getting problems with the "." within the name.

If we only check for `.isascii` it actually would simplify everyhting a lot, because this would not be needed anymore.

tests/functional/n/non/non_ascii_name.py
CarliJoy: > Got to the other files as well!
> 
> @CarliJoy Tests look very good and robust. tada I just wonder if we can move some of the remaining unittests to the functional tests as well. I'm asking this not only because I find them more readable, but also since I know they are actually better "tests". They way those are invoked better resembles the way `pylint` is invoked by normal users than the unittests are, which has recently caught a bug that the unittests simply could not in their current setup. This is something we as maintainers/contributors should fix, but in the meantime functional tests are often the way to go.

@DanielNoord which unit test are you referring to?
Because I added functional test for the most use cases. I escpecially use the unit tests for two things:
a) To check that actually got the correct node and make sure that I understood the logic correctly (internal logic checks)
b) To use parameterize to check a function for all possible outcomes. This would be very cumbersome to this manually in funcrtional tests, if there is not equivalent of parameterize. Some cases I use there are still tested within the functional tests. So testing all possiblites with unit test and a bunch of real world examples should actually cover everything IMO.

CarliJoy:
> Got to the other files as well!
> 
> @CarliJoy Tests look very good and robust. tada I just wonder if we can move some of the remaining unittests to the functional tests as well. I'm asking this not only because I find them more readable, but also since I know they are actually better "tests". They way those are invoked better resembles the way `pylint` is invoked by normal users than the unittests are, which has recently caught a bug that the unittests simply could not in their current setup. This is something we as maintainers/contributors should fix, but in the meantime functional tests are often the way to go.

@DanielNoord which unit test are you referring to?
Because I added functional test for the most use cases. I escpecially use the unit tests for two things:
a) To check that actually got the correct node and make sure that I understood the logic correctly (internal logic checks)
b) To use parameterize to check a function for all possible outcomes. This would be very cumbersome to this manually in funcrtional tests, if there is not equivalent of parameterize. Some cases I use there are still tested within the functional tests. So testing all possiblites with unit test and a bunch of real world examples should actually cover everything IMO.

PS: Very

tests/functional/e/.#emacs_file_lock_redefined_conf.py
DanielNoord: I would be fine with using `isascii()` and dropping the `regex`. Seems to be less cumbersome.

@Pierre-Sassoulas What's your opinion?

pylint/checkers/base.py
DanielNoord: Ah makes sense! I'll add this to my personal to-do list for a follow-up 😄

pylint/checkers/base.py
DanielNoord: I'll do a follow-up after this has been merged!

pylint/checkers/base.py
DanielNoord: What about changing the signature of `check_name` to:
```python
self, node_type: str, name: str, node: nodes.NodeNG, confidence: Confidence = interfaces.UNKNOWN, check_string: Optional[str] = None,
```
Wouldn't that avoid these warnings?

DanielNoord:
> @DanielNoord which unit test are you referring to? Because I added functional test for the most use cases. I escpecially use the unit tests for two things: a) To check that actually got the correct node and make sure that I understood the logic correctly (internal logic checks) b) To use parameterize to check a function for all possible outcomes. This would be very cumbersome to this manually in funcrtional tests, if there is not equivalent of parameterize. Some cases I use there are still tested within the functional tests. So testing all possiblites with unit test and a bunch of real world examples should actually cover everything IMO.

I was thinking of `test_check_import` specifically, but the others might fit as well. I agree that testing some examples for `nodes` makes sense, but can't we just put all those import statements in one big functional test file and see if they test correctly?

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,264 @@
...
+    pylint.checkers.BaseChecker, pylint.checkers.base.NameCheckerHelper
+):
+    """A strict name checker only allowing ASCII
+
+    If your programming guideline defines that you are programming in English,
+    then there should be no need for non ASCII characters in Python Names.
+    Everybody else can simply not use this check.
+
+    Note: This check only checks Names, so it ignores the content of
+          docstrings and comments!
```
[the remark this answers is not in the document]
CarliJoy: This should go in the doucmentation.
Actually what I am missing is a part of the User documentation that explain/lists all checkers/messages and possibly all configuration Options...
(Best of course auto generated)
Or did I miss this @DanielNoord ?


Currently using pylint is a bit overwhelming in this regards.

I looked into the message and I think they are to the point. This is really just an explaination for the programmers why we need this checker at all.

pylint/checkers/base.py
CarliJoy: I guess once it is within an linear inheritance it doesn't need to be.
Also it makes no sense to declare something abstractmethod if it does not has an AbstractMetaClass. Because only with an AbstractMetaClass this can also be checked.

pylint/checkers/non_ascii_names.py
CarliJoy: I kind of like *identifier* more and think it is more fitting as the *name*.
Name is more generic IMO. Also considering that now a file__name__ is an extra message.

@DanielNoord and @Pierre-Sassoulas could you agree about this. Because changing it back would mean not renaming some files but instead adopting the messages everywhere again (okay not that much of work, that to `sed`).

Just need to know this before going forward with some issues mentioned here.

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,264 @@
...
+        check_string: Optional[str] = None,
+    ) -> None:
+        """Check whether a name is using non-ASCII characters.
+
+        Also set the dynamic attribute ``_is_ascii_only`` that is used to
+        determine if a node has been already checked, so we don't have to handle
+        too many edge cases.
+        """
+
+        current_state = getattr(node, "_is_ascii_only", True)
```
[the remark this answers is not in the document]
CarliJoy: The lru_cache idea would work only if this is always the same node_type and as I can rember this is not always the case. One are assigments (=variable) and the other are the arguments of the function call.

The only idea would be to keep a list of nodes already checked. 
This would be bad memory wise i.e. once we use daemon etc.

So currently I really don't know a better way :( (besides modifying astroid to add an extra arguments `checked_by`)

pylint/checkers/non_ascii_names.py
CarliJoy: I will simplify that.

pylint/testutils/checker_test_case.py
CarliJoy: line and col_offset should be calculated from the given node.
The `ignore_position` is intended to create unit test in the sense of "Check only this unit/function", so this should not depend on the functionality of extracting the nodes/col.

Also this makes the unit test much smaller/easier to read.

tests/checkers/unittest_non_ascii_name.py
CarliJoy: For people like me that aren't fluent in AST this helps a lot to see what is going on.
Also in opposit to the functional test I define my expectations first and see if they match.

Do you really look in every functional test, compare the line endings and colums and see if they make sense?
I doubt that.
Especially as they generated automatically.

For me the functional test fell much more like Regression and Integration Tests, that IMO still do not replace Unit test.

But honestly -> are we really discussing now that I add **too many** tests 😉  ?

tests/checkers/unittest_non_ascii_name.py
CarliJoy: See https://github.com/PyCQA/pylint/pull/5643/files/7c0411a83caccef2059580bbd195ec93ca1b0dcd#r780088668

tests/checkers/unittest_non_ascii_name.py
CarliJoy: Yes I think so. Probably the complete class (defintion) is considered.
Maybe we can define the end_line manually to line, as a Class Defintion (at least the part containing the name) shouldn't exceed one line? Is there an easy way to do this?

tests/checkers/unittest_non_ascii_name.py
CarliJoy: As explained earlier: I still think they are different.
And sorry too much work. Won't do.

pylint/checkers/base.py
CarliJoy: Yep it would. Worte that a different point already.

CarliJoy:
> > @DanielNoord which unit test are you referring to? Because I added functional test for the most use cases. I escpecially use the unit tests for two things: a) To check that actually got the correct node and make sure that I understood the logic correctly (internal logic checks) b) To use parameterize to check a function for all possible outcomes. This would be very cumbersome to this manually in funcrtional tests, if there is not equivalent of parameterize. Some cases I use there are still tested within the functional tests. So testing all possiblites with unit test and a bunch of real world examples should actually cover everything IMO.
> 
> I was thinking of `test_check_import` specifically, but the others might fit as well. I agree that testing some examples for `nodes` makes sense, but can't we just put all those import statements in one big functional test file and see if they test correctly?

Then they are integration tests and I have to consider too many other side effects: i.e. nothing marked duplicated, modules should really exist, you name it. 
Just gets more complicated and additional work for no real benefit IMO.

I again was just glad about my unit tests, because I am in the process of figuring out how to remove `._is_ascii` attribute. For this unit test are just a charm, because they run fast and are easy to debug and don't have side effects.

pylint/checkers/non_ascii_names.py
CarliJoy: Kept the old name.

tests/functional/e/.#emacs_file_lock_redefined_conf.py
CarliJoy: I just removed it. Too much complexity that is not needed.

tests/functional/n/non/non_ascii_name.py
CarliJoy: Kept the original name for the message and freed me from renaming every file.... (was easier)

pylint/checkers/base.py
CarliJoy: moved

pylint/checkers/base.py
CarliJoy: Just did it anyway and didn't wait for @Pierre-Sassoulas, because I just want to finish it ASAP.

CarliJoy:
Using Unit Tests I was able to checkout and debug a number of things and so could simplify the Checker a good amount.

Also recognized, that I used Plural for the checker name and changed that.

The Checker is now almost dead simple 🎉 

Hope you like it :-)

pylint/checkers/non_ascii_names.py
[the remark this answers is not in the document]
CarliJoy: Fixed 🎉

tests/checkers/unittest_non_ascii_name.py
DanielNoord: This is something we're thinking about in #5466.

We'll likely change it so that `add_message` detects class and function nodes and handles them a bit better. For now this is fine.

(Thinking about this, this is another example of where the unittests do not represent actual pylint. The tests use a different `add_message` function which is defined independently of the actual function. So, if we change anything to `PyLinter.add_message` this doesn't get reflected in the unittests. Imo, this is a major design flaw of the unittests that has probably been here for years, but also one of the reasons why I often ask people to write functional tests.)

pylint/checkers/base_checker.py
```diff
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
     def process_tokens(self, tokens):
         """Should be overridden by subclasses."""
         raise NotImplementedError()
+
+
+class _BasicChecker(BaseChecker):
```
DanielNoord, line 207: As we're likely going to squash merge this PR it would be good to do this change and associated changes in `pylint/checkers/base.py` in a separate PR. I can this for you if you want?

pylint/checkers/__init__.py
```diff
@@ -39,7 +39,14 @@
 15: stdlib
 16: python3
 17: refactoring
-18-50: not yet used: reserved for future internal checkers.
+.
+.
+.
+25: non-ascii-names
```
DanielNoord, line 45:
```suggestion
24: non-ascii-names
```

pylint/checkers/base_checker.py
```diff
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
...
+
+class _NameCheckerBase(_BasicChecker):
+    """Class containing functions required by NameChecker and NonAsciiNameChecker"""
+
+    def _check_name(
+        self, node_type: str, name: str, node: nodes.NodeNG, confidence=interfaces.HIGH
+    ):
+        """Only Dummy function will be overwritten by implementing classes
+
+        Note: kwarg arguments will be different in implementing classes
```
DanielNoord, line 220:
```suggestion
```

pylint/checkers/base_checker.py
```diff
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
...
+        """
+        raise NotImplementedError
```
DanielNoord, line 222: Regarding your other comment, should this then inherit `AbstractClass`?

pylint/checkers/base_checker.py
```diff
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
...
+    name = "basic"
+
+
+class _NameCheckerBase(_BasicChecker):
+    """Class containing functions required by NameChecker and NonAsciiNameChecker"""
+
+    def _check_name(
+        self, node_type: str, name: str, node: nodes.NodeNG, confidence=interfaces.HIGH
+    ):
+        """Only Dummy function will be overwritten by implementing classes
```
DanielNoord, line 218:
```suggestion
        """Check whether the name associated with a node passes a certain check.
```

pylint/checkers/base_checker.py
```diff
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
...
+    __implements__ = interfaces.IAstroidChecker
+    name = "basic"
+
+
+class _NameCheckerBase(_BasicChecker):
+    """Class containing functions required by NameChecker and NonAsciiNameChecker"""
+
+    def _check_name(
+        self, node_type: str, name: str, node: nodes.NodeNG, confidence=interfaces.HIGH
+    ):
```
DanielNoord, line 217: What is the return type of this function?

pylint/checkers/non_ascii_names.py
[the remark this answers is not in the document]
DanielNoord: I have been working on this in #5396.

Currently the only way to store extra data per message is to store it in the message description, which is what that PR takes to populate each message's page. Also see follow-up #5527 which discuss better ways to store additional data per message. Perhaps for now we can add it to the message description as it would then be added to the documentation when #5396 gets merged?

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,208 @@
...
+            (
+                "Used when the name contains at least one non-ASCII unicode character. "
+                "See https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers"
+                " for a background why this could be bad."
+            ),
+            {"old_names": [("C0144", "old-non-ascii-name")]},
+        ),
+        # First %s will always be "file"
+        "W2402": (
+            '%s name "%s" contains a non-ASCII character. PEP 3131 only allows non-ascii identifiers.',
```
DanielNoord, line 60:
```suggestion
            '%s name "%s" contains a non-ASCII character. PEP 3131 only allows non-ascii identifiers, not file names.',
```

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,208 @@
...
+            "non-ascii-file-name",
+            (
+                # Some = PyCharm at the time of writing didn't display the non_ascii_name_loł
+                # files. Probably only a bug shows the problem quite good.
+                # That's also why this is a warning and not only a convention!
+                "Some editors don't support non-ASCII file names properly. "
+                "Even so Python supports UTF-8 files since Python 3.5 this isn't recommended for "
```
DanielNoord, line 67:
```suggestion
                "Even though Python supports UTF-8 files since Python 3.5 this isn't recommended for "
```

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,208 @@
...
+        node_type: str,
+        name: Optional[str],
+        node: nodes.NodeNG,
+        confidence=interfaces.HIGH,
+    ) -> None:
+        """Check whether a name is using non-ASCII characters.
+
+        Also set the dynamic attribute ``_is_ascii_only`` that is used to
+        determine if a node has been already checked, so we don't have to handle
+        too many edge cases.
```
DanielNoord, line 99:
```suggestion
```

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,208 @@
...
+            self._check_name("variable", node.name, node)
+
+    @utils.check_messages("non-ascii-name")
+    def visit_classdef(self, node: nodes.ClassDef) -> None:
+        self._check_name("class", node.name, node)
+        for attr, anodes in node.instance_attrs.items():
+            if not any(node.instance_attr_ancestors(attr)):
+                self._check_name("attr", attr, anodes[0])
+
+    def _check_module_import(self, node: Union[nodes.ImportFrom, nodes.Import]):
```
DanielNoord, line 187:
```suggestion
    def _check_module_import(self, node: Union[nodes.ImportFrom, nodes.Import]) -> None:
```

DanielNoord (changes requested):
@CarliJoy Thanks again! See my comments about unittests and functional tests about why I was so insistent on them. You're right I shouldn't have complained about too many tests 😄 

I think we're really getting there with this PR. The final checker is becoming really quite elegant and concise.

One questions that couldn't really be commented anywhere:
Do we need `tests/functional/n/non_ascii_name_class/__init__.py`? And the other `__init__` files in the functional test directories?

tests/checkers/unittest_non_ascii_name.py
Pierre-Sassoulas: > Do you really look in every functional test, compare the line endings and columns and see if they make sense? I doubt that.

This is something we have to do only once. When we update the functional output we check that the line/column is correct, for a single node, maybe multiple node if there was a particular issue opened about line/column values. The column is very seldom wrong because the information come from the AST and if the ast is wrong, well, we can't do much about it anyway. And if one node is correct the chance that all node are correct is very very very high.

Defining the expected line / column in a ``CheckerTestCase`` is very tedious, at least as tedious as checking the expected line / column in the functional test. But we can upgrade the expected output automatically with functional test once we know that the value are correct. Also reading a functional test is easier to understand by far. By the way ``CheckerTestCase`` is not a unit test, it's a... checker test. Basically the same thing than functional test but with 15 lines of plumbing for each expected message and only one checker instead of multiple (and you can disable other checker's message in functional if you want). Imo the only valid use if if you use parametrize it's the only way to be denser than functional tests.

> But honestly -> are we really discussing now that I add too many tests 😉 ?

There's such a thing as too many tests: if the tests are duplicated or test the same code construct over and over, then we might as well remove the duplicate, save time in our test suite, and save time if we're going to need to update the test later.

pylint/checkers/base_checker.py
```diff
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
...
         """Should be overridden by subclasses."""
         raise NotImplementedError()
+
+
+class _BasicChecker(BaseChecker):
+    __implements__ = interfaces.IAstroidChecker
+    name = "basic"
+
+
+class _NameCheckerBase(_BasicChecker):
```
Pierre-Sassoulas, line 212: I think this should be in a name_* checker file and not inherit from ``_BasicChecker``.

pylint/checkers/base_checker.py
```diff
@@ -201,3 +202,32 @@ class BaseTokenChecker(BaseChecker):
...
+    """Class containing functions required by NameChecker and NonAsciiNameChecker"""
+
+    def _check_name(
+        self, node_type: str, name: str, node: nodes.NodeNG, confidence=interfaces.HIGH
```
Pierre-Sassoulas, line 216: Do we really need the confidence here ? It's always HIGH because there is no inference when we visit a name.

Pierre-Sassoulas (changes requested):
Thank you for splitting the unicode MR, this is easier to review. The tests are also getting exhaustive and this is very nice. I think we're converging and will be able to merge this first part soon.

CarliJoy:
@DanielNoord wrote:
> @CarliJoy Thanks again! See my comments about unittests and functional tests about why I was so insistent on them. You're right I shouldn't have complained about too many tests smile
> 
> I think we're really getting there with this PR. The final checker is becoming really quite elegant and concise.
> 
> One questions that couldn't really be commented anywhere: Do we need `tests/functional/n/non_ascii_name_class/__init__.py`? And the other `__init__` files in the functional test directories?

I oriented myself on the existing functional tests and all main folders (a-z) and some of the subfolders contain a `__init__.py`.
I am not sure if that is required. I guess for testing the local import functions this is actually required.
Anyway, as this seems to be an issues of the existing codebase I kindly ask to address this issues outside of this PR.

pylint/checkers/base_checker.py
CarliJoy: This comes from the existing code base. But I will remove it.

pylint/checkers/base_checker.py
CarliJoy: I agreed with @DanielNoord to use linear inheritance for this file. Please do not ask for contradicting changes.

tests/checkers/unittest_non_ascii_name.py
CarliJoy: As I see it:
There are function tests, that test the complete pylint architecture works as expected, i.e. configuration settings messages with the correct line, col, etc....

Each of these tests takes approx 45ms on my machine. They are hard to use for non regular pylint developers like me, as helping Tools like shell completion and "Click to run on test" within an IDE like PyCharm don't work (or other useful features like going to the executed code if the test fails).

On the other hand there are the normal pytest `CheckerTestCases`, all helpers (i.e. code completion in the shell, IDE functions) work as expected. 
Using them, not all things of pylint (settings, lines, etc) work correctly/are considered. Which means they have little side effects and run fast a blink (all unittest together for my checker take approx 1ms on my Machine, a fraction of just a single function test!).
So I am testing only a method, that is kind of mocked and not a complete run -> feels damn much like a unittest to me.

For me as a contributor not fluent in the internals of neither pylint nor (and espcially) of AST/astroid they help a big deal understanding what is going on, without setting a number of brakepoints everywhere:
Looking at the unittests I know exactly which methods I would need to adopt and how the parameters are extracted.

So of course I can remove them (besides the import unit tests), as "duplicated" functional tests exist.

But I hope you (@Pierre-Sassoulas and @DanielNoord ) understand that from my view as a contributor it just feels wrong to remove these tests.

Still I removed two tests that were really just duplicates (import star and class). 

But I really would like keep the remainig checks as I they are very helpful for contributor IMO.

tests/checkers/unittest_non_ascii_name.py
CarliJoy: Removed test, as seems to be  more of an integration issues = functional test

pylint/checkers/base_checker.py
CarliJoy: reverted

pylint/checkers/base_checker.py
CarliJoy: reverted

pylint/checkers/base_checker.py
CarliJoy: reverted

pylint/checkers/base_checker.py
CarliJoy: reverted

CarliJoy:
@DanielNoord and @Pierre-Sassoulas 

I reverted the changes to the Base Checkers and now use a simpler args checks.
I don't know if this really catches all cases but I don't want to invest more time in this and contradicting statements on how to have the new checker use functions of the old one, especially as this are only 4 lines of code.

I also removed two unittests but kept the other. See https://github.com/PyCQA/pylint/pull/5643#discussion_r781062311 for my view on them.
If you still want the rest removed, submit a change.
For all of them (besides the imports) a functional test exists as well.

pylint/checkers/non_ascii_names.py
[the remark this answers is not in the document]
CarliJoy: Integrated

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # Some node types have customized messages
+        if node_type == "file":
+            msg = "non-ascii-file-name"
+        elif node_type == "module":
+            msg = "non-ascii-module-import"
+
+        self.add_message(msg, node=node, args=args, confidence=interfaces.HIGH)
+
+    # pylint: disable-next=arguments-renamed
```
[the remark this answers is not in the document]
CarliJoy: Removed.

tests/checkers/unittest_non_ascii_name.py
```diff
@@ -0,0 +1,279 @@
...
+            pytest.param(
+                "from foo.bar import functiøn as good, bäd",
+                "bäd",
+                id="from_okay_module_import_bad_as_good_and_bad",
+            ),
+            pytest.param(
+                "from foo.bar import functiøn as good, *",
+                # We still have functiøn within our namespace and could detect this
+                # But to do this properly we would need to check all `*` imports
+                # -> Too much efford!
```
DanielNoord, line 249:
```suggestion
                # -> Too much effort!
```

tests/checkers/unittest_non_ascii_name.py
```diff
@@ -0,0 +1,279 @@
...
+                "functiøn",
+                id="from_okay_module_import_bad_as_good_and_star",
+                marks=pytest.mark.xfail(
+                    reason="We don't what is imported when using star"
```
DanielNoord, line 253:
```suggestion
                    reason="We don't know what is imported when using star"
```

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,201 @@
...
+            return all("\u0000" <= x <= "\u007F" for x in self)
+
+
+NON_ASCII_HELP = (
+    "Used when the name contains at least one non-ASCII unicode character. "
+    "See https://www.python.org/dev/peps/pep-0672/#confusable-characters-in-identifiers"
+    " for a background why this could be bad. \n"
+    "If your programming guideline defines that you are programming in "
+    "English, then there should be no need for non ASCII characters in "
+    "Python Names. Everybody else can simply disable this check."
```
DanielNoord, line 39:
```suggestion
    "Python Names. If not you can simply disable this check."
```

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,201 @@
...
+        self, node_type: str, name: Optional[str], node: nodes.NodeNG
+    ) -> None:
+        """Check whether a name is using non-ASCII characters."""
+
+        if name is None:
+            # For some nodes i.e. *kwargs from a dict, the name will be empty
+            return
+
+        if not (Py37Str(name).isascii()):
+            type_label = constants.HUMAN_READABLE_TYPES.get(node_type, node_type)
```
DanielNoord, line 100:
```suggestion
            type_label = constants.HUMAN_READABLE_TYPES[node_type]
```

If we're missing any this will be a good check to tell us we need to update `constants.HUMAN_READABLE_TYPES` 😄

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,201 @@
...
+
+        # Check position only arguments
+        if arguments.posonlyargs:
+            for pos_only_arg in arguments.posonlyargs:
+                self._check_name("argument", pos_only_arg.name, pos_only_arg)
+
+        # Check "normal" arguments
+        args = arguments.args or tuple()
+        for arg in args:
+            self._check_name("argument", arg.name, arg)
```
DanielNoord, line 134:
```suggestion
        if arguments.args:
            for arg in arguments.args:
                self._check_name("argument", arg.name, arg)
```

pylint/checkers/non_ascii_names.py
```diff
@@ -0,0 +1,201 @@
...
+        """check module level assigned names"""
+        # The NameChecker from which this Checker originates knows a lot of different
+        # versions of variables, i.e. constants, inline variables etc.
+        # To simplify we use only `variable` here, as we don't need to apply different
+        # rules to different types of variables.
+        frame = node.frame()
+
+        if isinstance(frame, nodes.FunctionDef):
+            if node.parent in frame.body:
+                # Only perform the check if the assigment was done in within the body
```
DanielNoord, line 159:
```suggestion
                # Only perform the check if the assignment was done in within the body
```

DanielNoord (approved):
Approving with my last recommendations.

I think we could keep going back and forth about `pylint`'s testing framework, but it doesn't feel too productive. I just want to say that I do see and agree with some of the points you're making or made previously, but that within the context of the current code we can't easily fix those. That's the result of working with such an old codebase I guess.

With regard to conflicting suggestions, I think that is to be expected in open source projects without strict hierarchy and review procedures on large PRs such as these. Of course we try to minimise them, but I don't think it is strange that reviewers can have conflicting opinions about certain decisions. With PRs with 100+ comments it's easy to miss comments from previous reviewers (or even your own) about the same topic. I know that that is frustrating and it was something I ran into when I first started contributing to `pylint` but I think that's one of the downsides of these open source projects. That doesn't really make this less annoying but I hope you can see how such things can occur.

tests/checkers/unittest_non_ascii_name.py
Pierre-Sassoulas: Thank for bringing the speed issue to my attention. Maybe related to https://github.com/PyCQA/astroid/issues/1320. Anyway, I'll resolve because arguing about this is longer than the time it'll take to move to functional if we need to modify it later.

Pierre-Sassoulas (approved):
Thank you for the new checker @CarliJoy !

To add to what @DanielNoord said about conflicting reviews: the smaller and atomic the MR is, the less likely it's going to happen. Making our lives easier by making small easy reviewable MR will help tremendously. If we need to review 1000 lines repeatedly with hundred of comments then we might not have the time to read all comments from everyone. Especially if something that should be discussed in another MR in detail is lost in the hundred of comments and makes us feel that the code change a lot and that we need to reread everything every-time as a result.

Also remember there's other interactions at the same time on the repository (repositories!) we're maintaining, we don't have a single MR / issue to take care of and we have jobs and lifes on top of it.

PR #5649, opened 2022-01-08
[optional ext] Emit `redefined-loop-name` for redefinitions of loop variables in body

### LINKED ISSUE #5072: New check: mutated loop variable
### Current problem

Here's one I came across in the wild the other day:

```python
for i in range(5):
    print(i)
    i += 1
```

Pylint doesn't have any problem with it. In the real-world case I saw, it didn't cause a change in behavior, but the variable mutation line was unintended and unneeded.

### Desired solution

Mutating the variable of a `for`-loop is most likely a mistake, and in any case it's not a very good practice. I would like Pylint to flag it, and I think such a check could be enabled by default without too much difficulty.

### Additional context

_No response_

### LINKED ISSUE #5608: redefined-outer-name emitted for variables in the same namespace
### Bug description

pylint emits `redefined-outer-name` for loop variables defined in nested for statements. For example,

```
for y in range(10):
    for y in range(10):
        ...
```

Technically, both definitions of `y` are in the same namespace, and while this code does look fishy, the `redefined-outer-name` message seems irrelevant. The message is **not** emitted for

```
x = 42
x = 24
```

nor 

```
for z in range(10):
    ...

for z in range(10):
    ...
```

### Configuration

_No response_

### Command used

```shell
python3 -m pylint /tmp/foo.py
```

### Pylint output

```shell
************* Module foo
/tmp/foo.py:7:4: W0621: Redefining name 'y' from outer scope (line 6) (redefined-outer-name)
```

### Expected behavior

`redefined-outer-name` is not emitted.

### Pylint version

```shell
pylint 2.8.3
astroid 2.5.6
Python 3.9.9 (main, Nov 16 2021, 10:24:31) 
[GCC 11.2.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #5072 -- Adds new message in optional extension
Closes #5608 -- Uncouples the nested loop redefinition case from `redefined-outer-name` so the check can be enabled/disabled appropriately

```python
# 1
for y in iterable:
    for y in iterable:  # [already emitted redefined-outer-name, now emits redefined-loop-name]
        ...

# 2
for y in iterable:
    y = y.lower()  # [with this PR emits redefined-loop-name]
```
<s>Ultimately I'm open to the idea of making this into a new message, but then we should get a decision on whether to bring example 1 into the new message also.</s> done!


.copyrite_aliases
```diff
@@ -113,5 +113,10 @@
     "mails": ["tushar.sadhwani000@gmail.com", "tushar@deepsource.io"],
     "authoritative_mail": "tushar.sadhwani000@gmail.com",
     "name": "Tushar Sadhwani"
+  },
+  {
+    "mails": ["jacobtylerwalls@gmail.com"],
```
Pierre-Sassoulas, line 118: This is used to recognize two mails as the same person. Unless you use other emails sometime this is not necessary.

Pierre-Sassoulas (commented):
I'm not very clear on the best way between extending redefined outer name and creating a new checker. We could create a new message without creating a new checker too. I hope for a consensus to be reached in the discussion.

tests/functional/c/cellvar_escaping_loop.py
```diff
@@ -76,7 +76,7 @@ def good_case10():
     lst = []
     for i in range(10):  # pylint: disable=unused-variable
         def func():
-            i = 100
+            i = 100  # pylint: disable=redefined-outer-name
```
jacobtylerwalls, line 79: This looks like a false positive.

jacobtylerwalls: Fixed in 91e118ae71b9991fb1e672ef9eb3a783ca9ccb30

pylint/checkers/variables.py
```diff
@@ -1351,6 +1352,22 @@ def visit_global(self, node: nodes.Global) -> None:
             self.add_message("global-statement", node=node)
 
     def visit_assignname(self, node: nodes.AssignName) -> None:
+        if self.linter.is_message_enabled("redefined-outer-name") and isinstance(
```
DanielNoord, line 1355: Is `self.linter.is_message_enabled` needed? This tends to interact badly with `useless-suppression` as message don't get flagged as being normally emitted.

tests/functional/r/redefined/redefined_loop_name.py
```diff
@@ -0,0 +1,18 @@
...
+    line = line.strip()  # [redefined-outer-name]
+
+lines = [(1, "1\t"), (2, "2\t")]
+for i, line in lines:
+    line = line.strip()  # [redefined-outer-name]
+
+line = "no warning"
+
+for i in range(10):
+    i += 1  # [redefined-outer-name]
```
DanielNoord, line 18: Can you add a test for nested for loops?

DanielNoord (changes requested):

Pierre-Sassoulas (changes requested):
Look pretty good already !

pylint/checkers/variables.py
```diff
@@ -460,9 +460,10 @@ def _has_locals_call_after_node(stmt, scope):
         "`'from X import *'` style import.",
     ),
     "W0621": (
-        "Redefining name %r from outer scope (line %s)",
+        "Redefining name %r from outer scope or loop (line %s)",
```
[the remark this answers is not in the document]
jacobtylerwalls: The reporter of #5608 felt that it wasn't -- he saw "scope" and reasoned "then it shouldn't emit for variables in the same namespace".

pylint/checkers/variables.py
jacobtylerwalls: Wow, thanks for teaching me about that. I must have picked this up from other examples. That's an audit opportunity. I guess it's used for performance in some places but at the expense of false negatives for `useless-suppression`?

pylint/checkers/variables.py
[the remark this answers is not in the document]
Pierre-Sassoulas: 👍 let's keep it explicit then

pylint/checkers/variables.py
Pierre-Sassoulas: To add to what @DanielNoord said, I think the most "bang for the buck" issue regarding performance is the startup time ([importing astroid before launching pylint](https://github.com/PyCQA/astroid/issues/1320)). Sure the inference is what takes the most time overall. But I think it's rare that pylint is run on an entire code base by a user, it's done in CI where responsiveness is less of an issue. When a user launch pylint on a single file or on a small number of files he or she just modified then the startup time is taking more than 50% of the total time and make pylint feel very slow.

DanielNoord:
@jacobtylerwalls We might need to take a look at the failing `pylint` run here.

jacobtylerwalls:
Yes, I'm inclined toward that also, but I really do think we need to ask whether #1 from the top of the thread should become part of the new checker then.

jacobtylerwalls:
@superbobry do you have thoughts on this approach? This would allow you to disable `redefined-loop-name` and also be disabling the nested loop case, uncoupling it from `redefined-outer-name`.

Also, do you have thoughts on making this optional or default? I will probably disable this in my own projects.

jacobtylerwalls:
Updated to make this an optional extension.

jacobtylerwalls:
Note to self: test uncovered line?

DanielNoord:
Is this ready for review again?

jacobtylerwalls:
Yep!

jacobtylerwalls:
Failing docs check is not relevant, see #6470

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,85 @@
...
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""Optional checker to warn when loop variables are overwritten in the loop's body."""
+
+from astroid import nodes
+
+from pylint import checkers, interfaces
+from pylint.checkers import utils
+from pylint.checkers.variables import in_for_else_branch
```
DanielNoord, line 11: This should probably move to `utils`. Checkers importing functions from each other seems like a good way to cause headaches 10/15 years later on 😄

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,85 @@
...
+from pylint.interfaces import HIGH
+from pylint.lint import PyLinter
+from pylint.utils import utils as pylint_utils
+
+
+class RedefinedLoopName(checkers.BaseChecker):
+
+    __implements__ = interfaces.IAstroidChecker
```
DanielNoord, line 19: This is no longer necessary and should in fact raise a `DeprecationWarning`.

```suggestion
```

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,85 @@
...
+
+    msgs = {
+        "W2901": (
+            "Redefining %r from loop (line %s)",
+            "redefined-loop-name",
+            "Used when a loop variable is overwritten in the loop body.",
+        ),
+    }
+
+    def __init__(self, linter=None):
```
DanielNoord, line 30:
```suggestion
    def __init__(self, linter: PyLinter) -> None:
```

Weird that `mypy` doesn't complain about this. `linter` is no longer allowed to be `None`.

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,85 @@
...
+        super().__init__(linter)
+        self._loop_variables = []
```
DanielNoord, line 32: Could you add typing here? Probably need a `from __future__ ` as well.

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,85 @@
...
+                    node=node,
+                    confidence=HIGH,
+                )
+                break
+
+    @utils.check_messages("redefined-loop-name")
+    def visit_for(self, node: nodes.For) -> None:
+        assigned_to = [a.name for a in node.target.nodes_of_class(nodes.AssignName)]
+        # Only check variables that are used
+        dummy_rgx = pylint_utils.get_global_option(
```
DanielNoord, line 56: This util has been deprecated. `self.linter.config.dummy_variable_rgx` should work.

And if not, we found a `beta` blocker 😄

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,85 @@
...
+        if not isinstance(assign_type, (nodes.Assign, nodes.AugAssign)):
+            return
+        node_scope = node.scope()
+        for outer_for, outer_variables in self._loop_variables:
+            if node_scope is not outer_for.scope():
+                continue
+            if node.name in outer_variables and not in_for_else_branch(outer_for, node):
+                self.add_message(
+                    "redefined-loop-name",
+                    args=(node.name, outer_for.fromlineno),
```
DanielNoord, line 46: Is there a reason for using `fromlineno` instead of `lineno`?

tests/functional/ext/redefined_loop_name/redefined_loop_name.py
```diff
@@ -0,0 +1,53 @@
+"""Tests for redefinitions of loop variables inside the loop body.
```
DanielNoord, line 1: Like these tests very much 😄

tests/functional/ext/redefined_loop_name/reused_outer_loop_variable.rc
```diff
@@ -0,0 +1,2 @@
+[MASTER]
+load-plugins=pylint.extensions.redefined_loop_name,
```
DanielNoord, line 2: Just a thought, do we want to turn the extension on ourselves?

tests/functional/r/redefined/redefined_loop_name.txt
```diff
@@ -0,0 +1,6 @@
+redefined-loop-name:9:4:9:8::Redefining 'line' from loop (line 8):HIGH
```
DanielNoord, line 1: I think the related test file is missing here?

DanielNoord (changes requested):
As I said, I really like the test. Sorry for letting this sit so long, that causes a number of change-requests due to my own aggressive deprecation PRs related to `argparse` and `mypy` 😅  Sorry about that!

pylint/extensions/redefined_loop_name.py
jacobtylerwalls: Thanks for the heads up. BTW the deprecation warning is being silenced here:
https://github.com/jacobtylerwalls/pylint/blob/0e35511ea2f55da8eaa3272468f320195b5f8006/pylint/checkers/base_checker.py#L189

tests/functional/r/redefined/redefined_loop_name.txt
jacobtylerwalls: Oh, it got moved under `ext/`, and I forgot to delete this one. Thanks.

pylint/extensions/redefined_loop_name.py
jacobtylerwalls: Just copying from the prior implementation. I'm not very clear on the difference: mostly for class/function defs to avoid getting decorators?

Pierre-Sassoulas (approved):
I like the creation of a new checker from the VariableChecker. Some checks definitely need to be disentangled to make maintenance easier.

tests/functional/ext/redefined_loop_name/reused_outer_loop_variable.rc
jacobtylerwalls: I don't know. There are [a lot of changes](https://github.com/PyCQA/pylint/runs/5737015716?check_suite_focus=true) to make/suppress. It's a perfectly fine pattern to overwrite a loop variable when you're cleaning/stripping it. So I wouldn't necessary volunteer for it right now.

pylint/extensions/redefined_loop_name.py
jacobtylerwalls: nvm, I was confused, the warning did appear here: https://github.com/PyCQA/pylint/runs/6203141616?check_suite_focus=true

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,87 @@
...
+        "W2901": (
+            "Redefining %r from loop (line %s)",
+            "redefined-loop-name",
+            "Used when a loop variable is overwritten in the loop body.",
+        ),
+    }
+
+    def __init__(self, linter: PyLinter) -> None:
+        super().__init__(linter)
+        self._loop_variables: list[tuple[nodes.For, list[str | None]]] = []
```
DanielNoord, line 31:
```suggestion
        self._loop_variables: list[tuple[nodes.For, list[str]]] = []
```

I think. `mypy` will complain because `.name` is typed as `str | None` I think, but that is a mistake. An `Assign` node should always have a `str` `name` attribute. Generally, whenever typing in `astroid` is incorrect we add `type: ignore` to offending lines instead of "trickling-down" the incorrect typing to `pylint`.

_I have also stopped using assertions in these cases as it is difficult to keep track of those and remove them when we update `astroid`. We get warned about unused ignores, so that's often the better option imo._

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @utils.only_required_for_messages("redefined-loop-name")
+    def visit_assignname(self, node: nodes.AssignName) -> None:
+        assign_type = node.assign_type()
+        if not isinstance(assign_type, (nodes.Assign, nodes.AugAssign)):
+            return
+        node_scope = node.scope()
+        for outer_for, outer_variables in self._loop_variables:
+            if node_scope is not outer_for.scope():
```
DanielNoord, line 40: Should we add the `scope` to `_loop_variables`? That could be expensive to call every time?

DanielNoord (changes requested):

pylint/extensions/redefined_loop_name.py
jacobtylerwalls: Hmm. Parent lookups are O(1), right? That doesn't seem too bad. As opposed to iterating over a large body of statements or all children.

DanielNoord: Yeah it is definitely not the most expensive call, but since we're doing it on every pass I thought it might make sense. I'll leave this to you though!

pylint/checkers/utils.py
```diff
@@ -1743,3 +1743,10 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
         and ancestor.test.as_string() in TYPING_TYPE_CHECKS_GUARDS
         for ancestor in node.node_ancestors()
     )
+
+
```
Pierre-Sassoulas, line 1747: Should this be cached ? (As it's used in two checkers now).

pylint/checkers/variables.py
```diff
@@ -1078,31 +1071,7 @@ def open(self) -> None:
...
-                        args=(variable, outer_for.fromlineno),
-                        node=node,
-                    )
-                    break
-
-        self._loop_variables.append((node, assigned_to))
-
-    @utils.only_required_for_messages("redefined-outer-name")
     def leave_for(self, node: nodes.For) -> None:
-        self._loop_variables.pop()
```
Pierre-Sassoulas, line 1105: I don't see the initialization of ``_loop_variables`` being removed, was it used for something else too ?

jacobtylerwalls: The `.pop()` was moved to the optional checker, which still makes use of the _loop_variables (initialized somewhere else)

jacobtylerwalls: Essentially, `redefined-outer-name` was doing two things, including one special case for nested loops. I moved the logic for that special-nested-loops case into the new, more robust optional checker.

DanielNoord (approved):

pylint/extensions/redefined_loop_name.py
jacobtylerwalls: Looking at this more closely I think you're right -- thanks for the idea 👍🏻

pylint/extensions/redefined_loop_name.py
```diff
@@ -0,0 +1,87 @@
...
+        "W2901": (
+            "Redefining %r from loop (line %s)",
+            "redefined-loop-name",
+            "Used when a loop variable is overwritten in the loop body.",
+        ),
+    }
+
+    def __init__(self, linter: PyLinter) -> None:
+        super().__init__(linter)
+        self._loop_variables: list[tuple[nodes.For, list[str], nodes.NodeNG]] = []
```
DanielNoord, line 31:
```suggestion
        self._loop_variables: list[tuple[nodes.For, list[str], nodes.LocalsDictNodeNG]] = []
```

jacobtylerwalls: ooh, right!

DanielNoord (approved):

pylint/checkers/variables.py
Pierre-Sassoulas: So ``_loop_variables `` is used in both checkers now? Is it populated the same way in both? Won't the lack of pop cause problem in the old checker ?

jacobtylerwalls: No, it's no longer used in the old variables checker. Sorry this MR took a lot of twists and turns. It's only needed for `redefined-loop-name` so it sits in the new optional checker for it. That new optional checker now has `visit_assignname` and `visit_for` implemented, both of which need access to `_loop_variables`. One of those two methods is new functionality (#5072), the second one is moved from the variables checker where it was mixed in with `redefined-outer-name` and confusing the reporter of #5608

jacobtylerwalls: Oh! you mean the `= []`. Duh. Yeah I can get rid of it.

Pierre-Sassoulas: Sorry I could have been clearer here 😄

Pierre-Sassoulas (approved):

PR #5651, opened 2022-01-08
Fix #5326: Fix false positive for `unused-variable` for a comprehension variable matching a type annotation

### LINKED ISSUE #5326: False positive `undefined-variable` for variable in inner scope matching outer scope type annotation
### Bug description

```python
def func():
    a: bool
    _ = [print(kwarg1=a, kwarg2=a) for a in range(10)]
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
a.py:3: [E0602(undefined-variable), func] Undefined variable 'a'
```

### Expected behavior

No messages

### Pylint version

```shell
pylint 2.11.2-dev0
astroid 2.8.5
Python 3.9.6 (v3.9.6:db3ff76da1, Jun 28 2021, 11:49:53)

Does NOT occur on 2.11.1, possible regression in dev version?
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Fixes the regression described in #5326 where `undefined-variable` is emitted when a comprehension variable matches a type annotation.

One test result is edited to show that after these changes we will need better detection of walrus operators.

Fixes #5326

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1803,7 +1803,7 @@ class VariablesChecker(BaseChecker):
                 # Local refs are ordered, so we break.
                 #     print(var)
                 #     var = 1  # <- irrelevant
-                if defstmt_frame == node_frame and not ref_node.lineno < node.lineno:
+                if defstmt_frame == node_frame and ref_node.lineno > node.lineno:
                     break
 
                 # If the parent of the local reference is anything but an AnnAssign
--- a/ChangeLog
+++ b/ChangeLog
@@ -83,7 +83,7 @@ Release date: TBA
 * Fix false positive for ``unused-variable`` for a comprehension variable matching
   an outer scope type annotation.
 
-  Refs #5326
+  Closes #5326
 
 * Some files in ``pylint.testutils`` were deprecated. In the future imports should be done from the
   ``pylint.testutils.functional`` namespace directly.
--- a/ChangeLog
+++ b/ChangeLog
@@ -80,7 +80,7 @@ Release date: TBA
 
   Closes #5461
 
-* Fix false positive for `unused-variable` for a comprehension variable matching
+* Fix false positive for ``unused-variable`` for a comprehension variable matching
   an outer scope type annotation.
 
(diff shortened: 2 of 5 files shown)


DanielNoord:
Knowing you you have probably already investigated this, but it was impossible to fix this without creating the false negative?

jacobtylerwalls:
> Knowing you you have probably already investigated this, but it was impossible to fix this without creating the false negative?

Ha. So my thought there was that it was wrong to make `_is_only_type_assignment()` responsible for finding undefined variables in self-referencing walrus operator expressions. Pylint already has trouble with that, and it has nothing to do with with type annotations at all. I'll open a new issue.

```shell
$ cat test.py
'''Should raise undefined-variable'''
if (var := var):
    print(var)
$ pylint test.py

Your code has been rated at 10.00/10 (previous run: 5.00/10, +5.00)
```

I'm assuming that's the false negative you meant, since it was "created" instead of "left alone." (But "created" only insofar as we were getting lucky 🎲 .)

If you meant the false negative I started to describe in https://github.com/PyCQA/pylint/issues/5326#issuecomment-970491922, then to keep things sane I'll just break that out into a new issue. I think they can be addressed separately.

DanielNoord (approved):

pylint/checkers/variables.py
```diff
@@ -1803,7 +1803,7 @@ def _is_only_type_assignment(node: nodes.Name, defstmt: nodes.Statement) -> bool
                 # Local refs are ordered, so we break.
                 #     print(var)
                 #     var = 1  # <- irrelevant
-                if defstmt_frame == node_frame and not ref_node.lineno < node.lineno:
+                if defstmt_frame == node_frame and ref_node.lineno > node.lineno:
```
Pierre-Sassoulas, line 1806: I like the small optimization to save a "not" on top of the clever fix !

Pierre-Sassoulas (approved):
Thanks a lot @jacobtylerwalls, very elegant fix !

PR #5663, opened 2022-01-11
Fix remaining typing issues for ``PyLinter`` and related functions

DanielNoord:

## Type of Changes

## Description

These are the final changes necessary to add typing to the ``__init__`` of `PyLinter`. 
I'll comment on some of the changes as these are a bit all over the place.

### DIFF
--- a/pylint/constants.py
+++ b/pylint/constants.py
@@ -8,6 +8,12 @@ import platformdirs
 
 from pylint.__pkginfo__ import __version__
 
+if sys.version_info >= (3, 8):
+    from typing import Final
+else:
+    from typing_extensions import Final
+
+
 PY37_PLUS = sys.version_info[:2] >= (3, 7)
 PY38_PLUS = sys.version_info[:2] >= (3, 8)
 PY39_PLUS = sys.version_info[:2] >= (3, 9)
@@ -24,7 +30,7 @@ MSG_STATE_SCOPE_MODULE = 1
 # The line/node distinction does not apply to fatal errors and reports.
 _SCOPE_EXEMPT = "FR"
 
-MSG_TYPES = {
+MSG_TYPES: Final = {
     "I": "info",
     "C": "convention",
     "R": "refactor",
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -537,7 +537,7 @@ class PyLinter(
     def __init__(
         self,
         options=(),
-        reporter=None,
+        reporter: Union[reporters.BaseReporter, reporters.MultiReporter, None] = None,
         option_groups=(),
         pylintrc=None,
     ):
@@ -1496,7 +1496,11 @@ class PyLinter(
         msg_cat = MSG_TYPES[message_definition.msgid[0]]
         self.msg_status |= MSG_TYPES_STATUS[message_definition.msgid[0]]
         self.stats.increase_single_message_count(msg_cat, 1)
-        self.stats.increase_single_module_message_count(self.current_name, msg_cat, 1)
+        self.stats.increase_single_module_message_count(
+            self.current_name,  # type: ignore[arg-type] # Should be removable after https://github.com/PyCQA/pylint/pull/5580
+            msg_cat,  # type: ignore[arg-type] # Mypy doesn't see a Final dict as dict of literals
+            1,
+        )
         try:
             self.stats.by_msg[message_definition.symbol] += 1
         except KeyError:
@@ -1613,6 +1617,9 @@ class PyLinter(
     ) -> None:
         """Set the status of an individual message"""
         if scope == "module":
+            if not isinstance(line, int):
+                raise ValueError
+
             self.file_state.set_msg_status(msg, line, enable)
             if not enable and msg.symbol != "locally-disabled":
                 self.add_message(
@@ -1641,7 +1648,9 @@ class PyLinter(
         else:
             category_id_formatted = category_id
         if category_id_formatted is not None:
-            for _msgid in self.msgs_store._msgs_by_category.get(category_id_formatted):
+            for _msgid in self.msgs_store._msgs_by_category.get(
+                category_id_formatted, []
+            ):
                 message_definitions.extend(
                     self._get_messages_to_set(_msgid, enable, ignore_unknown)
                 )
--- a/pylint/utils/file_state.py
+++ b/pylint/utils/file_state.py
@@ -130,7 +130,7 @@ class FileState:
             self._module_msgs_state[msg.msgid] = {line: status}
 
     def handle_ignored_message(
-        self, state_scope: Optional[Literal[0, 1, 2]], msgid: str, line: int
+        self, state_scope: Optional[Literal[0, 1, 2]], msgid: str, line: Optional[int]
     ) -> None:
         """Report an ignored message.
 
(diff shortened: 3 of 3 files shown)


pylint/constants.py
```diff
@@ -24,7 +30,7 @@
 # The line/node distinction does not apply to fatal errors and reports.
 _SCOPE_EXEMPT = "FR"
 
-MSG_TYPES = {
+MSG_TYPES: Final = {
```
DanielNoord, line 33: Although `mypy` doesn't recognise that because this `dict` is final its values could be seen as `Literal` instead of `str` it does "allow" us to add an ignore later on.
We will be warned if `MSG_TYPES` is being altered and we can thus be sure that any of the values obtained from it is one of the values already in there.

pylint/lint/pylinter.py
```diff
@@ -537,7 +537,7 @@ def make_options() -> Tuple[Tuple[str, OptionDict], ...]:
     def __init__(
         self,
         options=(),
-        reporter=None,
+        reporter: Union[reporters.BaseReporter, reporters.MultiReporter, None] = None,
```
DanielNoord, line 540: I wanted to do this in a separate PR but some of the issues fixed here didn't show if I didn't add typing to at least one of these parameters.

pylint/lint/pylinter.py
```diff
@@ -1496,7 +1496,11 @@ def _add_one_message(
         msg_cat = MSG_TYPES[message_definition.msgid[0]]
         self.msg_status |= MSG_TYPES_STATUS[message_definition.msgid[0]]
         self.stats.increase_single_message_count(msg_cat, 1)
-        self.stats.increase_single_module_message_count(self.current_name, msg_cat, 1)
+        self.stats.increase_single_module_message_count(
+            self.current_name,  # type: ignore[arg-type] # Should be removable after https://github.com/PyCQA/pylint/pull/5580
```
DanielNoord, line 1500: `current_name` shouldn't be `None`...

pylint/lint/pylinter.py
```diff
@@ -1496,7 +1496,11 @@ def _add_one_message(
...
+            msg_cat,  # type: ignore[arg-type] # Mypy doesn't see a Final dict as dict of literals
```
DanielNoord, line 1501: See earlier comment about `Final`

pylint/lint/pylinter.py
```diff
@@ -1613,6 +1617,9 @@ def _set_one_msg_status(
     ) -> None:
         """Set the status of an individual message"""
         if scope == "module":
+            if not isinstance(line, int):
```
DanielNoord, line 1620: `int` is never `None` if the scope is `module`. I'm not sure what the best way to do this is. If it is not an `int` we should indeed raise a `ValueError`, but in our current code that's impossible...
Perhaps an `assert`? Let me know!

pylint/utils/file_state.py
```diff
@@ -139,6 +139,8 @@ def handle_ignored_message(
         or globally.
         """
         if state_scope == MSG_STATE_SCOPE_MODULE:
+            if not isinstance(line, int):
```
DanielNoord, line 142: Similar problem here: if `state_scope == MSG_STATE_SCOPE_MODULE` then `line` is never not an `int`. How to fix this...?

pylint/lint/pylinter.py
Pierre-Sassoulas: I think asserts are the canonical way to convey "I'm doing that to appease the mypy god".

pylint/utils/file_state.py
Pierre-Sassoulas: Isn't it an issue to fix in astroid ?

DanielNoord: No this is a pylint issue. If we disable something via `pylintrc` its disabled on the "package" scope, that's why it doesn't have a line number. If we disable something in a module itself it will always have an assosciated line number

pylint/lint/pylinter.py
DanielNoord: Yeah, although both `assert` and `isinstance` will actually decrease performance. I wonder if there is a way to do this without impacting performance..

pylint/utils/file_state.py
Pierre-Sassoulas: All right, thank you for clearing that up.

Maybe we could create a different structure to handle module and package disable, but I'm not very clear on the state of the code right now. I'll have a look before 2.13.0 is released. (Or I trust you if you find a solution that you think is good enough in the meantime).

pylint/lint/pylinter.py
Pierre-Sassoulas: Some typing can be declared beforehand at no cost (I think we used that for the message id store). But it only work when it's actually impossible for line to not be an int. I think the only way is to change the design which might be costly (and/or overkill).

Pierre-Sassoulas (approved):
I think we can merge, the isinstance won't be the main contributor to pylint being slow and a refactor of the design would take a long time. We can benefit from  the typing before that.

DanielNoord:
Perhaps @cdce8p knows a good trick for this?

pylint/lint/pylinter.py
cdce8p: If you know that it's never `None`, just use `cast`.
In case it's **supposed** to always be `int`, `assert isinstance(line, int)` is probably the better choice. The assert basically checks that some other code didn't mess up.

`if isinstance(...)` is more designed to select a code branch in case of true alternatives.

pylint/utils/file_state.py
cdce8p: Same here. Either `cast` or possibly `assert`.

cdce8p:
> Perhaps @cdce8p knows a good trick for this?

I left some comments. Check the conversations above 🔝

pylint/constants.py
DanielNoord: @cdce8p Do you know if there is any plans or previous discussion about the concept of final dicts? I saw a similar issue about final frozen sets I think, which would also be beneficial to us.

I can change this to `TypedDict`, but if there are plans to (at some point) support final dicts I think it makes sense to keep it like this as it "better" represents what this actually is.

cdce8p: > @cdce8p Do you know if there is any plans or previous discussion about the concept of final dicts? I saw a similar issue about final frozen sets I think, which would also be beneficial to us.
> 
> I can change this to `TypedDict`, but if there are plans to (at some point) support final dicts I think it makes sense to keep it like this as it "better" represents what this actually is.

I don't know of any plans. If `TypedDicts` work, they might be the easiest solution here.

pylint/lint/pylinter.py
```diff
@@ -1631,6 +1635,8 @@ def _set_one_msg_status(
     ) -> None:
         """Set the status of an individual message"""
         if scope == "module":
+            assert isinstance(line, int)
```
cdce8p, line 1638: I was wondering, would it make sense to add a comment along the lines
```py
assert ...  # line should always be int inside module scope
```

pylint/utils/file_state.py
```diff
@@ -139,6 +139,8 @@ def handle_ignored_message(
         or globally.
         """
         if state_scope == MSG_STATE_SCOPE_MODULE:
+            assert isinstance(line, int)
```
cdce8p, line 142: Same here

DanielNoord:
`TypedDict` creates a lot of problems because you can only use the correct keys and its difficult to `assert` that the first letter of each message id is in the keys of the `TypedDict`. I have opted to leave it as is.

PR #5665, opened 2022-01-11
Make sure to split non-separated csv ``OuputLine's``

DanielNoord:

## Type of Changes

## Description

https://github.com/PyCQA/pylint/blob/2631583577a3b2228af66214a364e4e284ba55dd/tests/testutils/test_output_line.py#L126-L130

This example was not actually handled correctly. For some reason `issubclass(str, Sequence)` equals `True`, so the `isinstance` check didn't make any sense. This meant that the above line had a `len` of `51`. This changes it so that a string will first be separated on the comma's and then parsed.
The diff is quite large as the indentation could then be removed. 

I also hope this removes some `DeprecationWarnings` that were showing up in the test suite on `3.7` and `3.8`.

### DIFF
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -117,49 +117,50 @@ class OutputLine(NamedTuple):
         """Create an OutputLine from a comma separated list (the functional tests expected
         output .txt files).
         """
+        if isinstance(row, str):
+            row = row.split(",")
         try:
-            if isinstance(row, Sequence):
-                column = cls._get_column(row[2])
-                if len(row) == 5:
-                    warnings.warn(
-                        "In pylint 3.0 functional tests expected output should always include the "
-                        "expected confidence level, expected end_line and expected end_column. "
-                        "An OutputLine should thus have a length of 8.",
-                        DeprecationWarning,
-                    )
-                    return cls(
-                        row[0],
-                        int(row[1]),
-                        column,
-                        None,
-                        None,
-                        row[3],
-                        row[4],
-                        UNDEFINED.name,
-                    )
-                if len(row) == 6:
-                    warnings.warn(
-                        "In pylint 3.0 functional tests expected output should always include the "
-                        "expected end_line and expected end_column. An OutputLine should thus have "
-                        "a length of 8.",
-                        DeprecationWarning,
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
Maybe we could remove the error handling ? There's no index error anymore. It could still happen with malformed output file.

DanielNoord:
I think the `IndexError` mimics refers to the fact that the `Sequence` or `list` does not have the correct length. Do you want me to remove it?

Pierre-Sassoulas:
Let's keep that un mind for 3.0, it's a breaking change for plugin development after all. Would you mind adding a todo comment close to the deprecation warnings ?

Pierre-Sassoulas:
No I meant a todo to remove this particular index error handling. Or do you think it's obvious ? The coverage decreased right now si it's easy to see the code became unused but we might not realize it later. I don't think we should add Todos for all deprecation warning. 

Sorry I'm on mobile so I can't be very detailled. I think the index error can happen on plugin code using our functional tests without enough elements in the output.

DanielNoord:
Oh wait, there is another `IndexError` that has lost coverage. I didn't catch that. I thought you were referring to the `IndexError` in `from_csv`.

Looking into it I don't think the test is testing what it used to test. We raising a `ValueError` on the first two cases because it tries to turn a `"'1'"` into an integer, which is impossible. This creates the `MalformedOutputLineException`. I feel like previously the tests were used to test different lengths, but oh well. They still pass...
I have added a test to make sure the `IndexError` is covered again. Basically we allow any `OutputLine` that is too long there so that we can still generate the nice error message even if it is too long. I don't think we should disallow that. If you misplace a `:` or `,` when hand-writing the `OutputLine` it would be nice to still get a formatted error message instead of a new `IndexError`. This test tests that behaviour.

Pierre-Sassoulas (approved):
> I have added a test to make sure the IndexError is covered again. Basically we allow any OutputLine that is too long there so that we can still generate the nice error message even if it is too long. 

I remember know. I wrote that code, I'm the one responsible for the lack of tests.. 😓

Thanks for covering that !

PR #5666, opened 2022-01-11
Fix #5586: False positive for `used-before-assignment` with homonyms in filtered comprehensions and except blocks

### LINKED ISSUE #5586: False positive `used-before-assignment` for filtered comprehensions in try blocks
### Bug description

```python
def func():
    try:
        print(value for value in range(1 / 0) if isinstance(value, int))  # BUG: emits used-before-assignment
    except ZeroDivisionError:
        value = 1
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
/Users/.../a.py:3:60: E0601: Using variable 'value' before assignment (used-before-assignment)
```

### Expected behavior

No message.

Here, `found_nodes` for the node inside the generator expression includes the assignment of `value` in the except block.
https://github.com/PyCQA/pylint/blob/c0fbbb7adbc39cbd67bbd0e0b1ca4e9421c12643/pylint/checkers/variables.py#L611-L620

I'm not sure if the responsibility for filtering that out lies in `to_consume()`; if not, I suppose the new `used-before-assignment` logic will have to deal with it.

### Pylint version

```shell
Caused in bd55b27d41542e3ca1f031f986b6151f6cac457f (2.13.0 unreleased)
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #5586

With a small adjustment, @DanielNoord's suggestion in https://github.com/PyCQA/pylint/issues/5586#issuecomment-1010155471 passes the suite, so I just opened a PR. I'm not certain this is the best way to do this, but I tried an alternative diff in `get_next_to_consume()`, and it turned out to be more questionable.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1316,6 +1316,9 @@ class VariablesChecker(BaseChecker):
             if utils.is_func_decorator(current_consumer.node) or not (
                 current_consumer.scope_type == "comprehension"
                 and self._has_homonym_in_upper_function_scope(node, consumer_level)
+                # But don't catch homonyms against the filter of a comprehension,
+                # "if x" in [x for x in expr() if x]
+                # https://github.com/PyCQA/pylint/issues/5586
                 and not (
                     isinstance(node.parent.parent, nodes.Comprehension)
                     and node.parent in node.parent.parent.ifs
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1317,7 +1317,7 @@ class VariablesChecker(BaseChecker):
                 current_consumer.scope_type == "comprehension"
                 and self._has_homonym_in_upper_function_scope(node, consumer_level)
                 # But don't catch homonyms against the filter of a comprehension,
-                # "if x" in [x for x in expr() if x]
+                # (like "if x" in "[x for x in expr() if x]")
                 # https://github.com/PyCQA/pylint/issues/5586
                 and not (
                     isinstance(node.parent.parent, nodes.Comprehension)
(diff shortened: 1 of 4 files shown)


DanielNoord (approved):

DanielNoord:
Thanks for picking this up immediately @jacobtylerwalls!

Not 100% about the place as well, but since this is what was making the method exit early this seem to be okay!

pylint/checkers/variables.py
```diff
@@ -1316,6 +1316,13 @@ def _check_consumer(
             if utils.is_func_decorator(current_consumer.node) or not (
                 current_consumer.scope_type == "comprehension"
                 and self._has_homonym_in_upper_function_scope(node, consumer_level)
+                # But don't catch homonyms against the filter of a comprehension,
+                # "if x" in [x for x in expr() if x]
```
Pierre-Sassoulas, line 1320:
```suggestion
                # if x in [x for x in expr() if x]
```

Pierre-Sassoulas (approved):
:thumbsup:

pylint/checkers/variables.py
DanielNoord: I like the quotes here as it shows that the `if x` is not part of the code but refers to the `if x` in the code example after the first `in`.

Pierre-Sassoulas: ~~I don't understand at all 😄 ~~ nevermind it took me 30s. 

```suggestion
                # (like "if x" in "[x for x in expr() if x]")
```

DanielNoord: This suggestion I can get behind 😄

PR #5669, opened 2022-01-13
Update `pandas-dev` testing primer package branch to `main`

areveny:
## Type of Changes

## Description

`pandas-dev` just renamed their main branch from `master` to `main` as seen [here](https://github.com/pandas-dev/pandas/pull/45336). This is one of our primer repositories so we need to update accordingly or else batch two tests fail.


DanielNoord:
Thanks @areveny!

Issue #5670, opened 2022-01-13
Investigate use of ``lru_cache`` on methods


DanielNoord:
### Problem

I stumbled upon this video which explains problems with using `lru_cache` on methods as it leaks `self`. I haven't watched the full video yet but watched the previous video and saw the PR at `pytest` here: https://github.com/pytest-dev/pytest/pull/9148

This is something we're doing ourselves as well and could be a potential source of memory leaks and other problems. I think this warrants an investigation and probably a refactor.

I have also opened https://github.com/PyCQA/astroid/issues/1344 to track this.

--------

I think after the investigation of internal use it might make sense to perhaps turn this into a checker? I'll report after I have finished my investigation whether that might be useful.

asottile:
you probably also want to ban `@functools.cache` on methods -- glad you enjoyed my video :)

DanielNoord:
Thanks @asottile! Wasn't sure whether if you would appreciate being tagged but I did wonder if there were other troublesome functions. I'll open an issue for `cache` in a couple of minutes.
I do wonder though, in the [docs of `functools`](https://docs.python.org/3/library/functools.html) there is an example of turning `cached_property` into a stacking of `property` and `cache`. Wouldn't that create the same issue?

asottile:
yeah that would likely have the same issues

mbyrnepr2:
@DanielNoord good job on your work here.
I was wondering, however, does this checker give a counter-intuitive warning to users?

The lru_cache is core Python functionality & the Python developers seem to be saying here that this is expected behaviour of the lru cache: https://bugs.python.org/issue19859.
My feeling is this checker would mean Pylint is now warning users who are using it correctly against doing so; and there are ways to misuse any part of the language.

DanielNoord:
Hmm perhaps you are right. It is at `Warning` level now. Would you want to completely remove the checker, or downgrade it to either `Convention` or `Refactor`?

mbyrnepr2:
I think especially the line in the warning - `lru_cache shouldn't be used on a method as it creates memory leaks` contradicts the views of the core Python developers in the issue linked above. Because of that, it feels controversial to say it shouldn't be used. 
So, I think that wording could change to something like `Method arguments are kept alive for the duration of the program when decorating a method with lru_cache` etc. This would simply be stating a fact and a user may be more comfortable knowing that nothing is wrong with that per-se and that it is simply expected behaviour that Pylint is giving a heads-up about.

[...]

PR #5672, opened 2022-01-13
Remove ``lru_cache`` from ``get_active_msgids``

DanielNoord:

## Type of Changes

## Description

Ref #5670.

Instead of using `lru_cache` we can use a basic cache defined on the instance to achieve a similar performance boost.

Tested on `pylint` over `pylint`. 
With new form of cache:
```console
54436    0.029    0.000    0.030    0.000 message_id_store.py:107(get_active_msgids)
```

Without new cache:
```console
54429    0.099    0.000    0.132    0.000 message_id_store.py:107(get_active_msgids)
```

Problem introduced in https://github.com/PyCQA/pylint/pull/5605

### DIFF
--- a/pylint/message/message_id_store.py
+++ b/pylint/message/message_id_store.py
@@ -1,6 +1,5 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
-import functools
 from typing import Dict, List, NoReturn, Optional, Tuple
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Looks good, I did not understand the performance report in the description though. No cache was actually faster than using lru cache ?!

DanielNoord:
Seems I miswrote the description. Updated.

The first is with this change, the second is without any form of cache. I haven't included `lru_cache` as that's not really a viable option.

Issue #5607, opened 2021-12-28
Refactor the message handling to use a smaller data structure


Pierre-Sassoulas:
### Current problem

The message store use message definition, which are big data structure only required when we need to create the messages or explains the message. This is minor as it's probably a small performance gain. Also, we dynamically check and generate the data structure from our code which is always the same for a version.

### Desired solution

We could use the msgid internally and simplify the code (and even improve performance). A translation from symbol to msgid when a symbol is used need to be done upfront. We could use typing to enforce that as msgid are string starting with a char betwee E W C or R and then 4 to 5 digits. , We could also generate an optimized data structure for Pylint's own message and only check the message for user plugin and extension.

### Additional context

#5606 #5605

cdce8p:
Worth remembering that the `msgid` might not be unique. AFAIK we don't enforce it currently. (It might not even be practical with custom plugins.)

Pierre-Sassoulas:
The msgid (``C0111``) and associated symbol (``missing-docstring``) are unique by definition, or it means that the same msgid refer to different checks/warning, which would be very problematic. It's also checked at run time right now (in the MessageIdStore) for both pylint's code, loaded extensions and user extensions.

PR #5673, opened 2022-01-13
Remove ``lru_cache`` from ``get_message_definitions``

DanielNoord:

## Type of Changes

## Description

Ref #5670, #5605 

This cache isn't actually that necessary. As long as `message_id_store.get_active_msgids` is reasonably fast (or cached) this function is also quite fast.

I tested using a similar pattern as in #5672 but actually setting values to a `dict` instance attribute and retrieving them takes longer than just doing the list comprehension for the return values of `get_active_msgids`. I think this is because if `get_active_msgids` returns just one `msgid` it's just one dict lookup whereas with a primitive cache it would be one dict lookup but also one dict assignment.
Anyway, a cache seems over engineered looking at the changes in these run times as long as `get_active_msgids` is cached.

Tested on `pylint` over `pylint`. 
With lru_cache:
```console
105    0.000    0.000    0.001    0.000 message_definition_store.py:50(get_message_definitions)
```

With new cache:
```console
5069    0.008    0.000    0.011    0.000 message_definition_store.py:51(get_message_definitions)
```

Without new cache:
```console
5069    0.005    0.000    0.008    0.000 message_definition_store.py:49(get_message_definitions)
```

Problem introduced in https://github.com/PyCQA/pylint/pull/5605

### DIFF
--- a/pylint/message/message_definition_store.py
+++ b/pylint/message/message_definition_store.py
@@ -2,7 +2,6 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 import collections
-import functools
 from typing import TYPE_CHECKING, Dict, List, Tuple, ValuesView
 
 from pylint.exceptions import UnknownMessageError
@@ -47,14 +46,8 @@ class MessageDefinitionStore:
         self._messages_definitions[message.msgid] = message
         self._msgs_by_category[message.msgid[0]].append(message.msgid)
 
-    @functools.lru_cache()
     def get_message_definitions(self, msgid_or_symbol: str) -> List[MessageDefinition]:
-        """Returns the Message definition for either a numeric or symbolic id.
-
-        The cache has no limit as its size will likely stay minimal. For each message we store
-        about 1000 characters, so even if we would have 1000 messages the cache would only
-        take up ~= 1 Mb.
-        """
+        """Returns the Message definition for either a numeric or symbolic id."""
         return [
             self._messages_definitions[m]
             for m in self.message_id_store.get_active_msgids(msgid_or_symbol)


Pierre-Sassoulas:
Are you sure that caching 'self' is an issue ? Messages stores are singletons. 0,01s for each pylint startup is significant imo. (Well it should be, and it will once astroid's import is faster).

DanielNoord:
`get_message_definitions` itself only takes 0.005s with this change (n=1 obviously, so take exact values with grain of salt).

I'm not sure whether this creates any real trouble immediately, but I think it would be good to enforce `lru-cache-decorating-method`. I really have no idea how the garbage collection works for `multiprocessing`, but can also imagine the other tools spawn multiple instances of `PyLinter` to do their own implementation of running `PyLinter` in parallel. 

I did a quick test for running `pylint` over an empty file, with this change.
Sorted with `cumtime` `get_message_definitions` is the 685th function.
Sorted with `tottime` `get_message_definitions` is the 375th function.

Same test with `pylint` over `pylint`.
Sorted with `cumtime` `get_message_definitions` is the 1060th function.
Sorted with `tottime` `get_message_definitions` is the 685th function.

I don't think this will be the first function that needs performance optimisation 😄 
Personally I'm fine with taking the 0.005-0.01s hit for the security that we're not creating any hard to debug memory leaks here.

Pierre-Sassoulas:
We never change the content of the cache and the value for a given pylint version are always the same. Also there is always only one instance.  As we said in the original MR the RAM used is negligible and deterministic. I don't see how it could become a problem. 

For multiprocessing the cache is not shared but actually generated for each thread see [this comment](https://github.com/kvangorkom/pandeia-coronagraphy/issues/22#issuecomment-401552446).

(It makes me want to generate the message store code into an optimized data structure.)

DanielNoord:
> We never change the content of the cache and the value for a given pylint version are always the same. 

Don't plugin messages get registered on the `MessageDefinitionStore` as well?

> Also there is always only one instance. As we said in the original MR the RAM used is negligible and deterministic. I don't see how it could become a problem.

You're right but I'd think I'd prefer to follow our new message here rather than adding a disable for it. The impact of this optimisations would be minimal:
For a run over `pylint` `cumtime` is 0.009s on a total runtime of 54.281s.
For a run over an empty file it is 0.000s (😅) on a total of 1.034s.
So for the `pylint` run we're discussing a 54,281/54,272 - 1 = 0,00016583% improvement here.

For me that warrants readability/maintainability over performance. Not adding the `disable` removes the need to explain why we're disabling here. Like I said, your argument is true but I don't think this small improvement warrants the "extra complexity" of special casing this method.

Pierre-Sassoulas:
> Don't plugin messages get registered on the MessageDefinitionStore as well?

Yes it's discussed in #5607, we can't just remove the code doing the check entirely.

>  0,00016583% improvement here

I agree there's more pressing optimization to do, and this function scale with run time (contrary to other optimization that have a fixed cost that we pay each time pylint is launched). But pylint is also downloaded 250 000 time a day, used maybe 10 time for each download, if we take into account an average use time of 10s, it adds up to more than an hour saved each day if we just disable this one false positive that we know will lower performance ever so slightly. We also have worst complexity to care about, the disable + short explanation won't be what make pylint unmaintainable imo 😄

DanielNoord:
> I agree there's more pressing optimization to do, and this function scale with run time (contrary to other optimization that have a fixed cost that we pay each time pylint is launched). But pylint is also downloaded 250 000 time a day, used maybe 10 time for each download, if we take into account an average use time of 10s, it adds up to more than an hour saved each day if we just disable this one false positive that we know will lower performance ever so slightly. We also have worst complexity to care about, the disable + short explanation won't be what make pylint unmaintainable imo 😄

I'll give in. I'll add a disable in #5674 and close this.

Do you have any good ideas on how to allow this behaviour without adding a `disable`? Perhaps we should create it before committing the checker to `main`?

Pierre-Sassoulas:
> Do you have any good ideas on how to allow this behaviour without adding a disable? 

I think singleton should be exempted, but I don't think there is a consensus on a pattern to indicate that an instance is a singleton in python. (See https://stackoverflow.com/questions/6760685/creating-a-singleton-in-python)

Issue #4701, opened 2021-07-11
Document a better release process than the current one


Pierre-Sassoulas:
See #4651

yushao2:
@Pierre-Sassoulas i think this can be closed now? 🎉

Pierre-Sassoulas:
I don't think we implemented git flow we have a simpler flow right now, [maybe it's enough](https://www.endoflineblog.com/gitflow-considered-harmful). I'm not sure that we're strictly following OneFlow either, but we don't have major issue with how we do thing right now. Thoughts @cdce8p ?

cdce8p:
From what I can tell, it mostly works how we need it to. So I don't really see a reason to change it just to conform to some specification. Btw. I always though of it as close to the GithubFlow, although we don't have a separate release branch ;)

Pierre-Sassoulas:
I'm going to try a 2.13 branch where we back-port bug fixes and crash fixes to be released as patch this time. So we do not have crash fixes waiting for month while we finalize a minor release There's a "Needs Backport" label let's start using this and squashing the MR that needs backport for easy cherry-picking. This is going to massively increase the release cadence for bug/crash fixes. Putting this in 2.13.0 milestone so I remember to document the process.

DanielNoord:
@Pierre-Sassoulas For clarity, it might be good to add a "backported" label which can replace the "needs backport" label after a PR has been back ported. That allows us to easily see if all recently merged PRs that need back porting have had it.

cdce8p:
Home Assistant has a label named `cherry-picked`. Every bugfix PR gets assigned to the  corresponding milestone and once it's "backported" `cherry-picked` is added.

PR #5674, opened 2022-01-13
Add ``lru-cache-decorating-method`` checker

### LINKED ISSUE #5670: Investigate use of ``lru_cache`` on methods

DanielNoord:

## Type of Changes

## Description

Blocked by #5672 and #5673.

Closes #5670.

`classmethod` and `staticmethod` have been excluded as I don't think they generate problems. We could ping asottile to see if this is correct, but I'm not sure how happy he would be to be pinged for this.


pylint/checkers/stdlib.py
```diff
@@ -571,6 +584,20 @@ def visit_boolop(self, node: nodes.BoolOp) -> None:
         for value in node.values:
             self._check_datetime(value)
 
+    @utils.check_messages("lru-cache-decorating-method")
+    def visit_functiondef(self, node: nodes.FunctionDef) -> None:
+        if node.decorators and isinstance(node.parent, nodes.ClassDef):
+            decorator_names = node.decoratornames()
+            if any(i in LRU_CACHE for i in decorator_names) and not (
```
Pierre-Sassoulas, line 591:
```suggestion
            if any(d in LRU_CACHE for d in decorator_names) and not (
```

Pierre-Sassoulas (commented):
Looks pretty good already. I wonder is singleton are a false positive, but it's hard to tell a class is a singleton by static analysis in python.

tests/functional/l/lru_cache_decorating_method.py
```diff
@@ -0,0 +1,48 @@
+"""Tests for lru-cache-decorating-method"""
+# pylint: disable=no-self-use, missing-function-docstring, reimported, too-few-public-methods
+# pylint: disable=missing-class-docstring, function-redefined
+
+import functools
```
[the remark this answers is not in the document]
DanielNoord: I'll add a change!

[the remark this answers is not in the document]
DanielNoord: See latest commit!

pylint/message/message_definition_store.py
```diff
@@ -47,8 +47,14 @@ def register_message(self, message: MessageDefinition) -> None:
         self._messages_definitions[message.msgid] = message
         self._msgs_by_category[message.msgid[0]].append(message.msgid)
 
+    # We disable the message here because MessageDefinitionStore is only
+    # initialized once and due to the size of the class does not run the
+    # risk of creating a large memory leak.
+    # See discussion in: https://github.com/PyCQA/pylint/pull/5673
     @functools.lru_cache()
-    def get_message_definitions(self, msgid_or_symbol: str) -> List[MessageDefinition]:
+    def get_message_definitions(  # pylint: disable=lru-cache-decorating-method
```
DanielNoord, line 55: Do we want to raise on the function or on the decorator?

Pierre-Sassoulas: I would say the decorator.

pylint/checkers/stdlib.py
```diff
@@ -446,6 +453,13 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             "Calls to breakpoint(), sys.breakpointhook() and pdb.set_trace() should be removed "
             "from code that is not actively being debugged.",
         ),
+        "W1516": (
+            "lru_cache shouldn't be used on a method as it creates memory leaks",
+            "lru-cache-decorating-method",
+            "By decorating a method with lru_cache the 'self' argument will be linked to "
+            "to the lru_cache function and therefore never garbage collected. It is recommended "
+            "to refactor code to avoid this pattern.",
```
Pierre-Sassoulas, line 461:
```suggestion
            "to the lru_cache function and therefore never garbage collected. Unless your instance"
            "will never need to be garbage collected (singleton) it is recommended to refactor"
            "code to avoid this pattern.",
```

pylint/checkers/stdlib.py
```diff
@@ -571,6 +585,29 @@ def visit_boolop(self, node: nodes.BoolOp) -> None:
...
+    @utils.check_messages("lru-cache-decorating-method")
+    def visit_functiondef(self, node: nodes.FunctionDef) -> None:
+        if node.decorators and isinstance(node.parent, nodes.ClassDef):
+            self._check_lru_cache_decorators(node.decorators)
+
+    def _check_lru_cache_decorators(self, decorators: nodes.Decorators) -> None:
+        """Check if instance methods are decorated with functools.lru_cache."""
+        lru_cache_nodes: List[nodes.NodeNG] = []
+        for d_node in decorators.nodes:
+            for infered_node in d_node.infer():
```
Pierre-Sassoulas, line 597: Shouldn't we use ``safe_infer`` or catch InferenceError here ?

pylint/message/message_definition_store.py
Pierre-Sassoulas: Wow, this refactor was a lot of work, sorry about that !

Pierre-Sassoulas (changes requested):
Tests are on point, thank you for doing the refactor you suggested, I think the message on the decorator is a really nice touch and will help user understand the issue easier.

pylint/checkers/stdlib.py
DanielNoord: Yes we should.. See tests 😅

pylint/message/message_definition_store.py
DanielNoord: No problem! This is indeed much better.

pylint/checkers/stdlib.py
DanielNoord: I didn't use `safe_infer` as then we only get one value or `Uninferable` for `lru_cache()` decorators.
I did add a `try...except` though.

Pierre-Sassoulas (approved):
Great job, a lightning fast new checker 😄

DanielNoord:
Now to get this to work with `astroid`...

DanielNoord:
Btw, @Pierre-Sassoulas not sure if this is the right place to discuss this but I'm trying to round off `2.13`. This was relatively easy to implement but I think we should try and close off the `2.13` milestone for crash fixes only.

Pierre-Sassoulas:
> I'm trying to round off 2.13

Let's do this :) I'm aiming to change the git workflow to release bug/crash fix faster see https://github.com/PyCQA/pylint/issues/4701#issuecomment-1012927593

PR #5681, opened 2022-01-14
Add typing to ``get_next_to_consume`` and ``_check_consumer``

DanielNoord:

## Type of Changes

## Description

Tagging @jacobtylerwalls again as the expert.

Instead of doing an assert I updated the typing and used an ignore. `mypy` is dumb here and doesn't recognise that `action is VariableVisitConsumerAction.CONSUME` says something about the second value of the tuple. `pylance` does recognise this. Hopefully `mypy` will start to recognise this and we'll get a useless ignore warning here.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -619,7 +619,7 @@ scope_type : {self._atomic.scope_type}
         else:
             del self.to_consume[name]
 
-    def get_next_to_consume(self, node):
+    def get_next_to_consume(self, node: nodes.Name) -> Optional[List[nodes.NodeNG]]:
         """Return a list of the nodes that define `node` from this scope. If it is
         uncertain whether a node will be consumed, such as for statements in
         except blocks, add it to self.consumed_uncertain instead of returning it.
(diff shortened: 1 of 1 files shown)


pylint/checkers/variables.py
```diff
@@ -1321,7 +1317,16 @@ def _check_consumer(
...
+    ) -> Union[
+        Tuple[
+            Union[
+                Literal[VariableVisitConsumerAction.CONTINUE],
+                Literal[VariableVisitConsumerAction.RETURN],
+            ],
+            None,
+        ],
+        Tuple[Literal[VariableVisitConsumerAction.CONSUME], List[nodes.NodeNG]],
+    ]:
```
Pierre-Sassoulas, line 1329:
```suggestion
    ) ->Tuple[VariableVisitConsumerAction, Optional[List[nodes.NodeNG]]]:
```
Would this be enough ? I know the other one is more specific but it's also very verbose.

Pierre-Sassoulas (approved):
LGTM, I have a small comment but I approve either way.

pylint/checkers/variables.py
DanielNoord: We could but then we'll need the assert again. The current version because if `CONSUME` is the first value then the second is restricted to be a `list`. That's why `pylance` infers but `mypy` fails to do.

jacobtylerwalls (approved):

PR #5676, opened 2022-01-13
Lint all files in a directory

matusvalo:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
|     | :bug: Bug fix          |
| ✓   | :sparkles: New feature |
|     | :hammer: Refactoring   |
|     | :scroll: Docs          |

## Description

This PR adds new parameter `--discover-files` which switches pylint into mode when it iterates over FS subtree and executes pylint on all `.py` files within this subtree regardless they are in package or not. But in the other hand this PR honours package name in the final output.

Example consider following subtree:
```
- test
     `+- a.py
      +- b.py
      +- c
          `+- __init__.py
           +- d.py 
```
The pylint output with no switch will be following:
```
$ pylint test
************* Module test
test/__init__.py:1:0: F0010: error while code parsing: Unable to load file test/__init__.py:
[Errno 2] No such file or directory: 'test/__init__.py' (parse-error)
```
The pylint output with switch will be following:
> Note: Each file contains single line simulating error `1 == 1`
```
$ pylint --discover-files test
************* Module a
test/a.py:1:0: W0104: Statement seems to have no effect (pointless-statement)
test/a.py:1:0: R0124: Redundant comparison - 1 == 1 (comparison-with-itself)
************* Module b
test/b.py:1:0: W0104: Statement seems to have no effect (pointless-statement)
test/b.py:1:0: R0124: Redundant comparison - 1 == 1 (comparison-with-itself)
************* Module c
test/c/__init__.py:1:0: W0104: Statement seems to have no effect (pointless-statement)
test/c/__init__.py:1:0: R0124: Redundant comparison - 1 == 1 (comparison-with-itself)
************* Module c.d
test/c/d.py:1:0: W0104: Statement seems to have no effect (pointless-statement)
test/c/d.py:1:0: R0124: Redundant comparison - 1 == 1 (comparison-with-itself)

------------------------------------------------------------------
Your code has been rated at 0.00/10 (previous run: 0.00/10, +0.00)
```
Closes #352

PR #5682, opened 2022-01-14
Lint all files in a directory by expanding arguments

### LINKED ISSUE #352: lint all files in a directory
Originally reported by: **Buck Evan (BitBucket: [bukzor](http://bitbucket.org/bukzor), GitHub: @bukzor?)**

---

It would make things easier for me to use pylint in CI if I could simply say `pylint .` in order to tell pylint to look at all my files.

flake8/pep8 already have this behavior. They make it configurable via a filepatterns config (default `*.py`) and an exclude config (default `.svn,CVS,.bzr,.hg,.git,__pycache__`).

Discussion from #pylint:

```
buck1: jcristau: in flake8, I can run `flake8 .` and it has a set of configured match/exclude patterns to check everything in the directory
buck1: currently pylint assumes that any directory is a package, and fails if not
buck1: how would you feel about including the flake8 behavior
buck1: would make my life 100% easier when trying to run pylint against everything during CI
jcristau: doesn't sound crazy, though there might be backwards compat concerns...  please float the idea in a ticket or on the list, i'm not the maintainer :)
buck1: jcristau: i'll make a feature-request ticket. Sounds like something I might be able to implement. Just wanted to see your thoughts first.
```

---
- Bitbucket: https://bitbucket.org/logilab/pylint/issue/352

matusvalo:
## Type of Changes

## Description

This PR is alternative to #5676 trying to resolve `sys.path` adjustment issue mentioned in https://github.com/PyCQA/pylint/issues/352#issuecomment-544220427.

The main rationale of this PR is to introduce recursive discovery of python modules/packages transparently to rest of pylint components. This is done by
expanding arguments containing directories with all modules and packages inside this directory. This approach should be safe with https://github.com/PyCQA/pylint/issues/352#issuecomment-544220427 since this change is totally transparent to `sys.path` adjustments in pylint. Hence for following subtree:

```
- test
     `+- a.py
      +- b.py
      +- c
          `+- __init__.py
           +- d.py
```

Executing command `pylint --recursive test` is equal to executing command `pylint test/a.py test/b.py test/c`.

Closes #352


pylint/lint/pylinter.py
```diff
@@ -1019,6 +1042,8 @@ def check(self, files_or_modules: Union[Sequence[str], str]) -> None:
                 DeprecationWarning,
             )
             files_or_modules = (files_or_modules,)  # type: ignore[assignment]
+        if self.config.recursive:
+            files_or_modules = tuple(self._discover_files(files_or_modules))
```
matusvalo, line 1052: Maybe we can even move calling `_discover_files` even to more early stage of pylint

matusvalo:
Marking as draft, since still missing polishing, tests and documentation. I would like to ask just for review of approach.

Pierre-Sassoulas (commented):
Yeah, I think this is pretty clever. In fact this is what was done currently (in bash or pre-commit by each users) when launching pylint with a long list of files, so imo this is pretty safe too. We're probably missing some optimization of doing the sys path manipulation less internally but the alternative is a very complex change and waiting even longer for what has been the most wanted feature of the project for years. Goob job !

DanielNoord:
I think this is also pretty clever! 

I wonder if we could make it so that `pylint .` is equal to `pylint . -r`. I think most users would want us to do that as well. Perhaps a bit out of the scope for this PR, but I think if we want to go ahead and do that we should include a PR that does so in the same release so it works as intended from the start.

Pierre-Sassoulas:
> I wonder if we could make it so that ``pylint .`` is equal to ``pylint . -r``.

``pylint`` should also be equal to ``pylint . -r`` as well. I also think ``-r`` should become the default option in 3.0.

DanielNoord:
> > I wonder if we could make it so that `pylint .` is equal to `pylint . -r`.
> 
> `pylint` should also be equal to `pylint . -r` as well. I also think `-r` should become the default option in 3.0.

Agreed! But I was wondering if for `2.x` we can already make `pylint .` work like `-r` was provided. Although I wonder how many projects use `pylint .` already... Hm perhaps we shouldn't do this and just add it to the `3.0` list.

Pierre-Sassoulas:
> already make ``pylint .`` work like -r was provided.

The intent is pretty clear when doing that imo, so I'd say it's safe to put in 2.x.

In fact my first instinct was to also make recursive the default in 2.x. But there's probably projects somewhere that use this "non feature" to exclude files from the linting,

matusvalo:
Thank you for your feedback. Let me finish the PR. I will close the previous one.

matusvalo:
Test added, code cleaned up, added type annotations and docstrings.

I have 2 questions:
1. About `--recursive` option. Right with no option there is old behaviour and later we should have `--recursive` option as default. But how will user then force old behaviour? New option will be added or we want to use `--recursive=y/n` option (in this case we won't need new option after)

2. Should this be documented in pylint userguide (running pylint chapter)? It is useful to have there but soon it will be default if I understand that correctly.

Pierre-Sassoulas:
> About --recursive option. Right with no option there is old behaviour and later we should have --recursive option as default. But how will user then force old behaviour after the switch? New option will be added or we want to use --recursive=y/n option (in this case we won't need new option after)

If this was a new project, I would add a ``--no-recursive`` option, but in pylint the ``--recursive=y/n`` is clearly the thing user would expect. I don't think we'll break this convention even if we switch to argparse and are forced to do *some* breaking changes for 3.0, what do you think @DanielNoord ?

DanielNoord:
> > About --recursive option. Right with no option there is old behaviour and later we should have --recursive option as default. But how will user then force old behaviour after the switch? New option will be added or we want to use --recursive=y/n option (in this case we won't need new option after)
> 
> If this was a new project, I would add a `--no-recursive` option, but in pylint the `--recursive=y/n` is clearly the thing user would expect. I don't think we'll break this convention even if we switch to argparse and are forced to do _some_ breaking changes for 3.0, what do you think @DanielNoord ?

Would using `--recursive=y/n` allow equating `--recursive` with `--recursive=y`? I think being able to just use `--recursive ` would be a highly request feature otherwise. Perhaps something with making all of these go to the same `dest` and making the options mutually exclusive? Like you did for `pydocstringformatter`.

> 2. Should this be documented in pylint userguide (running pylint chapter)? It is useful to have there but soon it will be default if I understand that correctly.

Yes, I would appreciate some documentation about it. If we make this default we can just update the documentation to reflect this!

matusvalo:
Documentation added to this PR.

pylint/lint/pylinter.py
```diff
@@ -1019,6 +1053,8 @@ def check(self, files_or_modules: Union[Sequence[str], str]) -> None:
                 DeprecationWarning,
             )
             files_or_modules = (files_or_modules,)  # type: ignore[assignment]
+        if self.config.recursive:
```
Pierre-Sassoulas, line 1056: It seems to happen before in the initialization but should we make recursive true, if pylint is called without directory (``pylint --rc-file=...``), or with ``.`` (``pylint . --rc-file=...``) and if ``__init__.py``does not exists in the current directory ? I think the intent to use recursive is clear for those two as it would only work if there was an ``__init__.py`` present. So we could fix it right now (before releasing pylint 3.0). On the other hand modifying behavior based on other argument than the one controlling the behavior feels wrong. @DanielNoord, thought ?

Pierre-Sassoulas (commented):
LGTM. I'm not very clear on what we should do in 2.x. The current change are very consistent, but I'm wondering if we can anticipate 3.0 behavior a little without breaking already existing workflows.

pylint/lint/pylinter.py
```diff
@@ -515,6 +515,13 @@ def make_options() -> Tuple[Tuple[str, OptionDict], ...]:
                     ),
                 },
             ),
+            (
+                "recursive",
```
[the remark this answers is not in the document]
matusvalo: I am fine also with yes/no option. @DanielNoord?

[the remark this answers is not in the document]
DanielNoord: Yeah, perhaps adding yes / no is better. Like I said previously, did we explore whether `--recursive` can mean `--recursive=y`?
And do we want a short name? `--rec`?

pylint/lint/pylinter.py
DanielNoord: > It seems to happen before in the initialization but should we make recursive true, if pylint is called without directory (`pylint --rc-file=...`), or with `.` (`pylint . --rc-file=...`) and if `__init__.py`does not exists in the current directory ?

I don't think `pylint --rc-file=...` should be allowed. It should either be `pylint . --rc-file=...` or with another directory.
We could discuss allowing `pylint --rc-file=...` if we add support for a `files` option in `rc-file`, but running `pylint` without any clear intent on what directory to lint wouldn't be good imo. How would we display the help message then?

> I think the intent to use recursive is clear for those two as it would only work if there was an `__init__.py` present. So we could fix it right now (before releasing pylint 3.0). On the other hand modifying behavior based on other argument than the one controlling the behavior feels wrong. @DanielNoord, thought ?

I agree that `recursive=y` should be set when called with `.` as directory. The same probably goes for missing `__init__.py`. However, those would be breaking changes and I wonder how much depends on this currently not being supported. I'd be hesitant to change this in a minor version.

Pierre-Sassoulas: > I don't think pylint --rc-file=... should be allowed. 


pytest does this I think it's a nice default value. It was asked [here and has currently 12 upvotes](https://github.com/PyCQA/pylint/issues/352#issuecomment-381300520)


> How would we display the help message then?


`pylint --help`, or `pylint -h`,  which I think is the expectation for CLI tool. (also ``pylint --long-help`` because we're special :smile:)


> I agree that recursive=y should be set when called with . as directory. The same probably goes for missing __init__.py. 


I meant both `.` and no `__init__.py` in the current directory at the same time. In this configuration the current result would be a fatal and nothing working so I guess it would not be a real breaking change (although there's probably someone somewhere who check for pylint exiting with a fatal instead of checking for an `__init__.py` :smile: ).

You made me realize that in fact we can make recursive true when there is no `__init__.py` in any directory we lint and it does not matter if it's `.` the argument that it was not working at all before is still true. But changing the recusrive option for `.` would be a breaking change as pylint could be working before if there was an `__init__.py` in the top level directory, but would start to lint new files unexpectedly.

DanielNoord: > pytest does this I think it's a nice default value. It was asked [here and has currently 12 upvotes](https://github.com/PyCQA/pylint/issues/352#issuecomment-381300520)

`Mypy` also allows this but only when you set `files` in your config. I really think we shouldn't allow this without setting `files` first. 

> `pylint --help`, or `pylint -h`, which I think is the expectation for CLI tool. (also `pylint --long-help` because we're special 😄)

My expectation is that just running the command also gives a short help message. I know so many CLI tool that do this.

pylint/lint/pylinter.py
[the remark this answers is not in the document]
Pierre-Sassoulas: If the obvious one letter (r) is already taken by report I think the full  explicit value is the next best thing ? Rational being that typing `ursive` (maybe with auto-completion once we do #5392 ) is easier than remembering if it's `--re` `--rec` or `--recu`.

> did we explore whether --recursive can mean --recursive=y?

It's probably a change to do in optparsse's `"metavar": "<y or n>",`. Maybe not worth it if we migrate to argparse ?

pylint/lint/pylinter.py
Pierre-Sassoulas: > My expectation is that just running the command also gives a short help message. I know so many CLI tool that do this.

Yes, this is what happen when the tool can't work without providing arguments. For example `ls` or `pwd`, `pytest` or `flake8` can run without args so they do. `mv` or `scp` can't, so they display the help (or something telling to how to display the help in the case of mv):
```
$scp
usage: scp [-346BCpqrTv] [-c cipher] [-F ssh_config] [-i identity_file]
            [-J destination] [-l limit] [-o ssh_option] [-P port]
            [-S program] source ... target
```

Imo, pylint should work without any argument (like flake8). But I wonder why mypy do not though. Do you know the rational behind it ?

DanielNoord: I looked in their repository for issues with "run", "parameter", "arguments" etc. but couldn't find anything.

I mean this boils down to one issue: can `pylint` run without arguments? I'd say no, as we need to know which directory to lint. `mypy` has answered that question similarly and says "as long as you don't provide an `files` argument I can't run". `flake8` has responded by assuming that if there is no `files` argument the user probably wants to run on `curdir`. Both are valid, but I like `mypy`'s response more as it makes no assumptions about what the user intends while still offering the desired behaviour (you just need to set `files` once in your `pyproject.toml`).

Besides this discussion I would vote against this as it breaks a core behaviour of `pylint`, namely how it is invoked. I don't think we should do such changes if there are no large benefits. In this case the benefit would be allowing to write `pylint` instead of `pylint .`. You know how much I like breaking changes and `DeprecationWarnings` (😄) but for me this is not enough to do this.
_Setting `--recursive=y` as the default is different to me as it is so much longer compared to writing ` .`._

Pierre-Sassoulas: > I looked in their repository for issues with "run", "parameter", "arguments" etc. but couldn't find anything.

Yeah neither do I.

> I like mypy's response more as it makes no assumptions about what the user intends while still offering the desired behavior

I think to default to current directory with recursive search is rather intuitive (and popular as it gathered more than 10 up votes in the original issue). But let's defer it to pylint 3.0 as it's only slightly related to the original problem and is basically adding a default value which we can do it easily when a consensus is reached. I'm going to create another issue and we'll be able to direct the original issue watchers to it and get their opinion.

Are we agreeing on recursive being activated automatically if there is no `__init__.py` in the given directory ?

DanielNoord: Yes I think so!

pylint/lint/pylinter.py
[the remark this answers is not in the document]
matusvalo: `--recursive` parameter changed to `--recursive=y/n`

pylint/lint/pylinter.py
matusvalo: > > I looked in their repository for issues with "run", "parameter", "arguments" etc. but couldn't find anything.
> 
> Yeah neither do I.
> 
> > I like mypy's response more as it makes no assumptions about what the user intends while still offering the desired behavior
> 
> I think to default to current directory with recursive search is rather intuitive (and popular as it gathered more than 10 up votes in the original issue). But let's defer it to pylint 3.0 as it's only slightly related to the original problem and is basically adding a default value which we can do easily when a consensus is reached. I'm going to create another issue and we'll be able to direct the original issue watchers to it and get their opinion.
> 
> Are we agreeing on recursive being activated automatically if there is no `__init__.py` in the given directory ?

If I read it correctly, we are not going to change default behaviour of pylint? Is there anything pending it this PR?

Pierre-Sassoulas: I would say we change the behavior a little: when the user want to lint a directory without an ``__init__.py`` inside it, we make ``recursive`` to be true. I think the tests needs to be adapted for a directory without init file and one test for a file that does not exists could be added.

matusvalo: It means 
1. executing command `pylint` inside a directory `spam` without `__init__.py` file will trigger also recursive behaviour
2. executing command `pylint .` inside a directory `spam` without `__init__.py` file will trigger also recursive behaviour 

Is that correct?

Pierre-Sassoulas: For 1, no, it should show the help like before, we did not reach a consensus on this with @DanielNoord  and I'm going to create an issue to settle it by popular vote, once we merge this one. (By the way, what do you think of defaulting to "." ? :) )

For 2. yes exactly.

DanielNoord: Just for myself but perhaps helpful for others. I think for now this is the summary:
For `2.x`:
When in `spam` without `__init__.py`:
`pylint .` > Recursive
`pylint` > Exit without doing anything.
`pylint ../spam` > ??? (Recursive I guess)?

When in `spam` with `__init__.py`:
`pylint .` > not recursive
`pylint` >  Exit without doing anything.
`pylint ../spam` > ??? (Not recursive I guess)?

When in parent without `__init__.py` of `spam` without `__init__.py`:
`pylint .` > Recursive, finds `spam`
`pylint spam` > Recursive
`pylint` > Exit without doing anything.
`pylint ../.` > ??? (Recursive I guess)?

When in parent without `__init__.py` of `spam` with `__init__.py`:
`pylint .` > Recursive, finds `spam`
`pylint spam` > Not recursive
`pylint` > Exit without doing anything.
`pylint ../.` > ??? (Recursive I guess)?

-----------------
For `3.0`. In bold is those that changed compared to `2.x`:
When in `spam` without `__init__.py`:
`pylint .` > Recursive
`pylint` > **TBD.**
`pylint ../spam` > **Recursive**

When in `spam` with `__init__.py`:
`pylint .` > **Recursive**
`pylint` > **TBD.**
`pylint ../spam` > **Recursive**

When in parent without `__init__.py` of `spam` without `__init__.py`:
`pylint .` > Recursive, finds `spam`
`pylint spam` > Recursive
`pylint` > **TBD.**
`pylint ../.` > **Recursive**

When in parent without `__init__.py` of `spam` with `__init__.py`:
`pylint .` > Recursive, finds `spam`
`pylint spam` > **Recursive**
`pylint` > **TBD**
`pylint ../.` > **Recursive**

Pierre-Sassoulas: > ??? (Recursive I guess)?

I would say those should indeed be recursive.

matusvalo:
@Pierre-Sassoulas @DanielNoord The last commit implements default recursive mode when current directory (`.`) is passed to pylint. I did not add cases like `../.`, etc because this can be tricky. Can we go for now with more simple solution? If you insist to have also cases like `../.`, my humble opinion is to create new PR for that to keep his one simple and straightforward. is there anything in this PR missing or it can be accepted?

pylint/lint/pylinter.py
matusvalo: Last commit implements default recursive mode when `.` directory is passed to pylint.

Pierre-Sassoulas:
I thought the change for changing the recursive option was "simple" but seeing that:

- the test are not passing anymore so it means there must be some more adaptation and it's trickier than I though
- and being generic seems hard and we can only handle ``.``
- The main change has been the most asked feature in pylint for years and we already have a follow up with #5701

I'm in favor of reverting the last commit, merging this and doing a follow-up MR for "little" ameliorations that aren't that small after all. What do you think @DanielNoord ?

matusvalo:
> the test are not passing anymore so it means there must be some more adaptation and it's trickier than I though

Oh I was not expecting this. I checked only test_self.py passing and expecting that this change wont break anything.

> I'm in favor of reverting the last commit, merging this and doing a follow-up MR for "little" ameliorations that aren't that small after all.

+1. I am for keeping this PR as straightforward as possible and add rest later.

pylint/lint/pylinter.py
DanielNoord: Let's leave this comment as unresolved as it is quite important for future people looking back at this PR and it might get easily lost.

DanielNoord:
+1 on keeping this simple.
 
Let's revert. @matusvalo would you be willing to write a follow-up issue to lay out what needs to be done to get default recursive behaviour as discussed in `2.x` and `3.0`? That way we don't lose what we've been working on here!

matusvalo:
> Let's revert. 

Last commit reverted.

> @matusvalo would you be willing to write a follow-up issue to lay out what needs to be done to get default recursive behaviour as discussed in `2.x` and `3.0`? That way we don't lose what we've been working on here!

Yes, I can create issues for that but it can take days to be done.

Pierre-Sassoulas (approved):
LGTM, thank you @matusvalo, this has been the most wanted pylint feature since more than 7 years and I'm pretty sure it's done now :)

doc/faq.rst
```diff
@@ -116,6 +116,27 @@ For example::
 
 Much probably. Read :ref:`ide-integration`
 
+3.5 I need to run pylint over all modules and packages in my project directory.
+-------------------------------------------------------------------------------
+
+``pylint`` command by default accepts only list of python modules and packages. Putting
```
DanielNoord, line 122:
```suggestion
By default the ``pylint`` command only accepts a list of python modules and packages. Using a
```

doc/faq.rst
```diff
@@ -116,6 +116,27 @@ For example::
...
+directory which is not a package results in an error::
+
+    pylint mydir
+    ************* Module mydir
+    mydir/__init__.py:1:0: F0010: error while code parsing: Unable to load file mydir/__init__.py:
+    [Errno 2] No such file or directory: 'mydir/__init__.py' (parse-error)
+
+To execute pylint over all modules and packages under the directory, ``--recursive=y`` option must
```
DanielNoord, line 130:
```suggestion
To execute pylint over all modules and packages under the directory, the ``--recursive=y`` option must
```

doc/faq.rst
```diff
@@ -116,6 +116,27 @@ For example::
...
+be provided. This option causes that pylint tries to discover all modules (files ending with ``.py`` extension)
```
DanielNoord, line 131:
```suggestion
be provided. This option makes ``pylint`` attempt to discover all modules (files ending with ``.py`` extension)
```

doc/faq.rst
```diff
@@ -116,6 +116,27 @@ For example::
...
+and all packages (all directories containing ``__init__.py`` file).
```
DanielNoord, line 132:
```suggestion
and all packages (all directories containing a ``__init__.py`` file).
```

doc/faq.rst
```diff
@@ -116,6 +116,27 @@ For example::
...
+Those modules and packages are then subject for analysis::
```
DanielNoord, line 133:
```suggestion
Those modules and packages are then analyzed::
```

doc/faq.rst
```diff
@@ -116,6 +116,27 @@ For example::
...
+
+    pylint --recursive=y mydir
+
+When ``--recursive=y`` option is used, also module and packages are accepted as parameters::
```
DanielNoord, line 137:
```suggestion
When ``--recursive=y`` option is used, modules and packages are also accepted as parameters::
```

ChangeLog
```diff
@@ -9,6 +9,9 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+* Add ``--recursive`` option to allow linting recursively all modules and packages in subtree.
```
DanielNoord, line 12:
```suggestion
* Add ``--recursive`` option to allow recursive discovery of all modules and packages in subtree.
```

doc/whatsnew/2.13.rst
```diff
@@ -77,6 +77,12 @@ Extensions
 Other Changes
 =============
 
+* Add ``--recursive`` option to allow linting recursively all modules and packages in subtree. Running pylint with
```
DanielNoord, line 80:
```suggestion
* Add ``--recursive`` option to allow recursive discovery of all modules and packages in subtree. Running pylint with
```

doc/whatsnew/2.13.rst
```diff
@@ -77,6 +77,12 @@ Extensions
...
+  ``--recursive`` option will check all discovered ``.py`` files and packages found inside subtree of directory provided
```
DanielNoord, line 81:
```suggestion
  ``--recursive=y`` option will check all discovered ``.py`` files and packages found inside subtree of directory provided
```

doc/whatsnew/2.13.rst
```diff
@@ -77,6 +77,12 @@ Extensions
...
+  as parameter to pylint.
+
+  Closes #352
```
DanielNoord, line 84: This should be copied to the `Changelog` file. ie., `2.13.rst` and `Changelog` should include the same changes.

doc/user_guide/run.rst
```diff
@@ -33,6 +33,10 @@ will work if ``directory`` is a python package (i.e. has an __init__.py
 file or it is an implicit namespace package) or if "directory" is in the
 python path.
 
+By default, pylint will result with an error when one of the arguments is directory which is not
```
DanielNoord, line 36:
```suggestion
By default, pylint will exits an error when one of the arguments is a directory which is not
```

doc/user_guide/run.rst
```diff
@@ -33,6 +33,10 @@ will work if ``directory`` is a python package (i.e. has an __init__.py
...
+a python package. In order to run pylint over all modules and packages within file system
```
DanielNoord, line 37:
```suggestion
a python package. In order to run pylint over all modules and packages within the provided
```

doc/user_guide/run.rst
```diff
@@ -33,6 +33,10 @@ will work if ``directory`` is a python package (i.e. has an __init__.py
...
+subtree of directory, ``--recursive=y`` option must be provided.
```
DanielNoord, line 38:
```suggestion
subtree of a directory, the ``--recursive=y`` option must be provided.
```

pylint/lint/pylinter.py
```diff
@@ -515,6 +515,15 @@ def make_options() -> Tuple[Tuple[str, OptionDict], ...]:
...
+                {
+                    "type": "yn",
+                    "metavar": "<yn>",
+                    "default": False,
+                    "help": "Discover python modules and packages in file system subtree.",
```
DanielNoord, line 524:
```suggestion
                    "help": "Discover python modules and packages in the file system subtree.",
```

tests/test_self.py
```diff
@@ -1290,3 +1290,15 @@ def test_regex_paths_csv_validator() -> None:
...
+
+    def test_regression_recursive(self):
+        self._test_output(
+            [join(HERE, "regrtest_data", "directory", "subdirectory"), "--recursive=n"],
+            expected_output="No such file or directory",
+        )
+
+    def test_recursive(self):
+        self._runtest(
+            [join(HERE, "regrtest_data", "directory", "subdirectory"), "--recursive=y"],
```
DanielNoord, line 1302: Could you add a test for the `pylint .` behaviour? I think there are some tests already where we change the `curdir` at the start of the test and then move back to the old `curdir` afterwards.

DanielNoord (changes requested):
Sorry for taking so long to review this!

The change itself looks good, but I think we would benefit from some additional tests and I added some spelling comments.

tests/test_self.py
matusvalo: done.

doc/whatsnew/2.13.rst
matusvalo: done

matusvalo:
Tests are failing for  TestRunTC.test_regression_recursive_current_dir but it is working on my local computer.

DanielNoord:
@matusvalo Does this pass for you locally?

tests/test_self.py
```diff
@@ -100,6 +100,24 @@ def _configure_lc_ctype(lc_ctype: str) -> Iterator:
             os.environ[lc_ctype_env] = original_lctype
 
 
+@contextlib.contextmanager
```
Pierre-Sassoulas, line 103: This could probably even go in ``tests/conftest.py`` to be used everywhere ? I don't know how frequent it is. But we might want to do that in another MR, this one need to be merged 😄

matusvalo:
> @matusvalo Does this pass for you locally?

Yes and no. the tests are failing only when executed by `pytest tests` as mentioned before. When you execute the file directly (`pytest tests/test_self.py`), it is working fine.

The reason is because pytest in case of executing tests via command `pytest tests` is adding to the `sys.path` also `/Users/matus/dev/pylint/tests/regrtest_data` path. Having this path causes that here:
https://github.com/PyCQA/pylint/blob/40ef1bebe9ee47ef0beef9b3ec53628d954710fa/pylint/lint/expand_modules.py#L65-L70
exception is not raised and `modname` is used not from exception handler but directly from return value of function. This is causing that pylint is not raising error for `TestRunTC.test_regression_recursive_current_dir ` and hence failing the unittest.

But as said before only in case of running `pytest tests`. I have pushed the possible fix in last commit.

matusvalo:
I have merged main to fix conflict. @DanielNoord is there anything pending in this PR?

doc/user_guide/run.rst
```diff
@@ -33,6 +33,10 @@ will work if ``directory`` is a python package (i.e. has an __init__.py
 file or it is an implicit namespace package) or if "directory" is in the
 python path.
 
+By default, pylint will exits an error when one of the arguments is a directory which is not
```
DanielNoord, line 36:
```suggestion
By default, pylint will exit with an error when one of the arguments is a directory which is not
```

pylint/lint/pylinter.py
```diff
@@ -1005,6 +1014,33 @@ def initialize(self):
             if not msg.may_be_emitted():
                 self._msgs_state[msg.msgid] = False
 
+    @staticmethod
+    def _discover_files(files_or_modules: Sequence[str]) -> Iterator[str]:
```
DanielNoord, line 1018: Shouldn't `yields` be typed with `Generator`? That's what I always do.

tests/test_self.py
```diff
@@ -1284,3 +1294,47 @@ def test_regex_paths_csv_validator() -> None:
...
+    def test_recursive(self):
+        self._runtest(
+            [join(HERE, "regrtest_data", "directory", "subdirectory"), "--recursive=y"],
+            code=0,
+        )
+
+    def test_recursive_current_dir(self):
+        with _test_sys_path():
+            # pytest is including directory HERE/regrtest_data to sys.path which causes
+            # astroid believe that directory is a package.
```
DanielNoord, line 1313:
```suggestion
            # astroid believe to that directory is a package.
```

tests/test_self.py
```diff
@@ -1284,3 +1294,47 @@ def test_regex_paths_csv_validator() -> None:
...
+                os.chdir(join(HERE, "regrtest_data", "directory"))
+                self._runtest(
+                    [".", "--recursive=y"],
+                    code=0,
+                )
+
+    def test_regression_recursive_current_dir(self):
+        with _test_sys_path():
+            # pytest is including directory HERE/regrtest_data to sys.path which causes
+            # astroid believe that directory is a package.
```
DanielNoord, line 1329:
```suggestion
            # astroid believe to that directory is a package.
```

DanielNoord (changes requested):
Sorry @matusvalo, was swarmed at work and didn't really have the time to look into this.

One final question (and some spelling fixes). 😄

pylint/lint/pylinter.py
matusvalo: It it is just yielding it can be also Iterator - see https://docs.python.org/3/library/typing.html#typing.Generator. But if you insist to have here Generator type, I am not against it - please let me know.

matusvalo:
All suggestions resolved. Only point is just Iterator vs. Generator question. Let me know about it.

tests/test_self.py
```diff
@@ -1284,3 +1294,47 @@ def test_regex_paths_csv_validator() -> None:
...
+    def test_recursive(self):
+        self._runtest(
+            [join(HERE, "regrtest_data", "directory", "subdirectory"), "--recursive=y"],
+            code=0,
+        )
+
+    def test_recursive_current_dir(self):
+        with _test_sys_path():
+            # pytest is including directory HERE/regrtest_data to sys.path which causes
+            # astroid believe to that directory is a package.
```
jacobtylerwalls, line 1313:
```suggestion
            # astroid to believe that directory is a package.
```

tests/test_self.py
```diff
@@ -1284,3 +1294,47 @@ def test_regex_paths_csv_validator() -> None:
...
+                os.chdir(join(HERE, "regrtest_data", "directory"))
+                self._runtest(
+                    [".", "--recursive=y"],
+                    code=0,
+                )
+
+    def test_regression_recursive_current_dir(self):
+        with _test_sys_path():
+            # pytest is including directory HERE/regrtest_data to sys.path which causes
+            # astroid believe to that directory is a package.
```
jacobtylerwalls, line 1329:
```suggestion
            # astroid to believe that directory is a package.
```

jacobtylerwalls (commented):
Daniel spotted these errors but was working just a bit quickly when suggesting the fix :-)

matusvalo:
Thank you @jacobtylerwalls !

pylint/lint/pylinter.py
DanielNoord: TIL! Thanks 😄 

Fine with keeping like this

DanielNoord (approved):

Pierre-Sassoulas:
This is a huge improvement to pylint, thank you @matusvalo  !

PR #5684, opened 2022-01-15
Fix #5683: False positive ``used-before-assignment`` in loop `else` where the only non-break exit is through `except`

### LINKED ISSUE #5683: False positive for `used-before-assignment` in `else` clause when the only non-`break` exit from a loop is an except block
### Bug description

```python
def do_something():
    pass

for _ in range(3):
    try:
        do_something()  # not an exit branch because of `else` below (will only exit here with an exception)
        error = e  # < exit branch where error is defined
    else:
        break  # < exit condition where error is *not* defined, but will skip `raise error` below
    print("retrying...")
else:
    raise error
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:13:10: E0601: Using variable 'error' before assignment (used-before-assignment)
```

### Expected behavior

No message. Identified by @martimlobao and discussed in detail https://github.com/PyCQA/pylint/pull/5402#issuecomment-1012791780.

Quoting from my [suggested](https://github.com/PyCQA/pylint/pull/5402#issuecomment-1013209351) solution:

So maybe this is the constraint to check for:

- if the use of `error` being tested is guarded by `else` in `for/else` (or `while/else`)
- if the definition of `error` in an `except` block is a descendant of the same `for/while`
- if that inner `else` breaks
- if there is no `continue` anywhere at all underneath the loop and before the break (other than inside another loop, but including nested conditions and try/excepts)

### Pylint version

```shell
pylint 2.13.0-dev0
astroid 2.9.2
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #5683

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -741,9 +741,107 @@ scope_type : {self._atomic.scope_type}
                     # if one of the except blocks does not define the name in question,
                     # raise, or return. See: https://github.com/PyCQA/pylint/issues/5524.
                     continue
+
+            if NamesConsumer._check_loop_finishes_via_except(
+                node, other_node_statement.parent.parent
+            ):
+                continue
+
             # Passed all tests for uncertain execution
             uncertain_nodes.append(other_node)
         return uncertain_nodes
+
+    @staticmethod
+    def _check_loop_finishes_via_except(
+        node: nodes.NodeNG, other_node_try_except: nodes.TryExcept
+    ) -> bool:
+        """Check for a case described in https://github.com/PyCQA/pylint/issues/5683.
+        It consists of a specific control flow scenario where the only
+        non-break exit from a loop consists of the very except handler we are
+        examining, such that code in the `else` branch of the loop can depend on it
+        being assigned.
+
+        Example:
+
+        for _ in range(3):
+            try:
+                do_something()
+            except:
+                name = 1  <-- only non-break exit from loop
+            else:
+                break
+        else:
+            print(name)
+        """
+        if not other_node_try_except.orelse:
+            return False
+        closest_loop: Optional[
+            Union[nodes.For, nodes.While]
+        ] = utils.get_node_first_ancestor_of_type(node, (nodes.For, nodes.While))
+        if closest_loop is None:
+            return False
+        if not any(
+            else_statement is node or else_statement.parent_of(node)
+            for else_statement in closest_loop.orelse
+        ):
+            # `node` not guarded by `else`
+            return False
+        for inner_else_statement in other_node_try_except.orelse:
+            if isinstance(inner_else_statement, nodes.Break):
+                break_stmt = inner_else_statement
+                break
+        else:
+            # No break statement
+            return False
+
+        def _try_in_loop_body(
+            other_node_try_except: nodes.TryExcept, loop: Union[nodes.For, nodes.While]
+        ) -> bool:
+            """Return True if `other_node_try_except` is a descendant of `loop`."""
+            return any(
+                loop_body_statement is other_node_try_except
+                or loop_body_statement.parent_of(other_node_try_except)
+                for loop_body_statement in loop.body
+            )
+
+        if not _try_in_loop_body(other_node_try_except, closest_loop):
+            for ancestor in closest_loop.node_ancestors():
+                if isinstance(ancestor, (nodes.For, nodes.While)):
+                    if _try_in_loop_body(other_node_try_except, ancestor):
+                        break
+            else:
+                # `other_node_try_except` didn't have a shared ancestor loop
+                return False
+
+        for loop_stmt in closest_loop.body:
+            if NamesConsumer._recursive_search_for_continue_before_break(
+                loop_stmt, break_stmt
+            ):
+                break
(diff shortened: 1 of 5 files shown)


pylint/checkers/variables.py
```diff
@@ -743,10 +743,109 @@ def _uncertain_nodes_in_except_blocks(
...
+                break
+        else:
+            print(name)
+        """
+        if not other_node_try_except.orelse:
+            return False
+        closest_loop: Optional[
+            Union[nodes.For, nodes.While]
+        ] = utils.get_node_first_ancestor_of_type(
+            node, (nodes.For, nodes.While)  # type: ignore
```
DanielNoord, line 783: Can you add the error this is ignoring? `mypy` has not yet implemented the feature, but we're trying not to do general type ignores.

Btw, isn't this returning the correct typing? DO we need to update `get_node_first_ancestor_of_type`?

pylint/checkers/variables.py
```diff
@@ -743,10 +743,109 @@ def _uncertain_nodes_in_except_blocks(
...
+            if isinstance(inner_else_statement, nodes.Break):
+                break_stmt = inner_else_statement
+                break
+        else:
+            # No break statement
+            return False
+
+        def _try_in_loop_body(loop: Union[nodes.For, nodes.While]) -> bool:
+            """Return True if `other_node_try_except` is a descendent of `loop`."""
+            nonlocal other_node_try_except
```
DanielNoord, line 803: I'd prefer passing `other_node_try_except` as parameter. Can we do that?

DanielNoord (changes requested):
Thanks @jacobtylerwalls! That's quite a lot of code!

pylint/checkers/variables.py
Pierre-Sassoulas: I agree typing ``get_node_first_ancestor_of_type`` in a prior MR would probably make sense here. But it seems it's node_ancestors from astroid's ``NodeNG`` that should be typed ?
```python
def get_node_first_ancestor_of_type(
    node: nodes.NodeNG, ancestor_type: Union[Type[T_Node], Tuple[Type[T_Node]]]
) -> Optional[T_Node]:
    """Return the first parent node that is any of the provided types (or None)"""
        if isinstance(ancestor, ancestor_type):
            return ancestor
    return None
```

pylint/checkers/variables.py
```diff
@@ -743,10 +743,109 @@ def _uncertain_nodes_in_except_blocks(
...
+                # `other_node_try_except` didn't have a shared ancestor loop
+                return False
+
+        for loop_stmt in closest_loop.body:
+            if NamesConsumer._recursive_search_for_continue_before_break(
+                loop_stmt, break_stmt
+            ):
+                break
+        else:
+            # No continue found, so we arrived at our special case!
```
Pierre-Sassoulas, line 825: That was a whole journey 😄 !

tests/functional/u/use/used_before_assignment_issue4761.py
```diff
@@ -10,3 +10,169 @@ def function():
...
+        except ValueError as verr:
+            error = verr
+        else:
+            break
+    else:
+        while function():
+            print('The time is:')
+            break
+        else:
+            raise error
```
Pierre-Sassoulas, line 178: I'm pretty sure we're going to still be surprised by the code that exists in the wild, but I'm also pretty confident we won't be surprised often. Those tests are thorough !

Pierre-Sassoulas (approved):
LGTM, the typing issue is minor even if we might want to dig into it before merging (fix it then rebase on it to remove the typing that became uneeded?). Thank you @jacobtylerwalls , great addition to 2.13 !

pylint/checkers/variables.py
jacobtylerwalls: Ah, it needs a `...`! I'll fix in another PR (and the similar function just below it)

DanielNoord: Ah, that's a nasty oversight from me in the review. Good catch!

DanielNoord (approved):
Sorry! Should have changed my review to approved after the last comments were integrated!

DanielNoord:
@jacobtylerwalls Just something I thought of, I see the confidence of `used-before-assignment` is `UNDEFINED`. I believe this message will likely always require some form of inference (right?). Should we update the confidence of the message to be `INFERENCE`?

jacobtylerwalls:
What about a new level `CONTROL_FLOW`? It's not really node-inference dependent.

DanielNoord:
I like that a lot!

jacobtylerwalls:
Cool, I'll PR it unless you've already started 🏃🏻‍♂️

DanielNoord:
No I haven't, would certainly welcome a PR 😄

DanielNoord:
@jacobtylerwalls This seems to have landed in stale PR heaven 😄  If you could update the test output I will merge this asap!

DanielNoord (approved):

DanielNoord:
Thanks!

PR #5694, opened 2022-01-17
Cache results when checking for valid attributes

### LINKED ISSUE #5695: Performance improvement with reusing checks for attribute validity
### Current problem

While triaging #4562 I noticed two occurrences of `obj.value` causes two calls to `typecheck.visit_attribute` which is a fairly substantial method.

```
val = 1 + obj.value
print(obj.value)
```

These two `visit_attribute` calls should produce the same result from the same scope, so it seems natural to cache the result.

I did a quick implementation of the general concept in the linked PR and there are some promising results.

Profiling on `pylint` current:
```
        1    0.000    0.000   22.432   22.432 __init__.py:21(run_pylint)
    12172    0.035    0.000    0.654    0.000 typecheck.py:1007(visit_attribute)
```

```
        1    0.000    0.000   22.297   22.297 __init__.py:21(run_pylint)
    12185    0.027    0.000    0.325    0.000 typecheck.py:1009(visit_attribute)
```

1.4% less run time.

Django has a lot of attributes. Profiling Django before:
```
        1    0.000    0.000  229.853  229.853 __init__.py:21(run_pylint)
   189671    0.656    0.000   23.609    0.000 typecheck.py:1007(visit_attribute)
  1327697    0.262    0.000    0.434    0.000 typecheck.py:1016(<genexpr>)
   169909    0.052    0.000    0.061    0.000 typecheck.py:1031(<listcomp>)
   393684    0.063    0.000    0.105    0.000 typecheck.py:1052(<genexpr>)
    46297    0.030    0.000    0.066    0.000 typecheck.py:1058(<listcomp>)
     1812    0.002    0.000    3.504    0.002 typecheck.py:1113(_get_nomember_msgid_hint)
```

Profiling Django after:
```
        1    0.000    0.000  215.938  215.938 __init__.py:21(run_pylint)
   189671    0.408    0.000    7.123    0.000 typecheck.py:1009(visit_attribute)
   322385    0.070    0.000    0.115    0.000 typecheck.py:1032(<genexpr>)
    44869    0.015    0.000    0.018    0.000 typecheck.py:1047(<listcomp>)
   118548    0.020    0.000    0.033    0.000 typecheck.py:1068(<genexpr>)
     9157    0.006    0.000    0.014    0.000 typecheck.py:1074(<listcomp>)
      605    0.001    0.000    0.957    0.002 typecheck.py:1130(_get_nomember_msgid_hint)
```

7.1% less runtime. We only execute the body of the function about 1/3 as often, so some 2/3 of attributes were already checked.

### Desired solution

Only one functional test, `no_member_dataclasses` fails with the mock-up implementation of this improvement because we define the same name twice across two different classes.

If this change looks sensible, we:

* Have the cache store and check against the scope to prevent this test failure
* Add additional tests to try and break this cache

This will probably reduce the performance improvement slightly as cache lookups become more expensive.

### Additional context

_No response_

areveny:
## Type of Changes

## Description

Early development to give reference for discussion.

Closes #5695


pylint/checkers/typecheck.py
```diff
@@ -1013,6 +1016,24 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        if isinstance(expr, nodes.Call):
+            expr = expr.func
+            outer = "()"
+        if isinstance(expr, nodes.Name):
+            outer = expr.name + outer
+        elif isinstance(expr, nodes.Attribute):
+            outer = expr.attrname + outer
+        name_with_expr = f"{outer}.{node.attrname}"
+        if (
+            outer != ""
```
Pierre-Sassoulas, line 1031:
```suggestion
            outer

File: pylint/checkers/typecheck.py [hunk 2]
```diff
@@ -1013,6 +1016,24 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+            and name_with_expr in self.node_exists
+            and self.node_exists[name_with_expr]
```
Pierre-Sassoulas, line 1033:
```suggestion
            and self.node_exists.get(name_with_expr) 
```

pylint/checkers/typecheck.py
```diff
@@ -1013,6 +1016,24 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
 
         function/method, super call and metaclasses are ignored
         """
+
```
Pierre-Sassoulas, line 1019: I think it would be a good thing to isolate the cache updating in a specifically named function si it's explicit that this part is about caching.

Pierre-Sassoulas (commented):
I liked the analysis, 2 to 7 percent better performance is huge too.

areveny:
Added the scope in the cache key so existing tests pass. The performance improvement is smaller but still good, on Django we hit the cache 60% of the time and save 5% runtime.

Currently working on some unittests to try and break this and ensure caches hits/misses are correct.

areveny:
Incorporated existing feedback, added unit tests and a ChangeLog message in a provisional place since this is targeting 2.14.

I think there are obscure cases where the cache breaks, for example conditional imports/class definitions in `if` statements, but I am not sure if we would/can pick those up now.

Pierre-Sassoulas:
> I think there are obscure cases where the cache breaks

By "break" do you mean "a cache miss", or "a crash" ? 😄

By the way, I should have asked during my first review (sorry), but any reason why we are not using ``functool.lru_cache`` ?

areveny:
Break meaning a false negative.

Below using the cache hides an error. `counter.b` is used twice in the same function. The cache assumes that all attributes within a scope with the same name have the same result. The cache returns that `counter.b` is valid in the second access, the current code infers on each access to ensure it is valid to catch this situation.

```
class RootCounter:
    elements = [1, 2, 3]
    b = [1, 2, 3]

class RootCounter2:
    elements = [1, 2, 3]

class CounterContainer1:

    counter = RootCounter()

    def fn(self):
        print(counter.elements)
        print(counter.b)
        counter = RootCounter2()
        print(counter.b) # no-member
```



I'm not using a cache from `functools` because the function accepts a different Attribute node object each time so it wouldn't hit the cache. We could write a custom decorator to pull out the cache key from the node but that's essentially what's happening now.

Also I don't think cache eviction is necessary, the memory impact of storing this cache map should be small compared to the AST.

DanielNoord:
@areveny I kinda forgot what the status of this PR is. Is it ready for review? Or is the issue mentioned above still there?

Pierre-Sassoulas:
> making assumptions about attributes staying either valid or invalid in a given scope 

Do you have an example where it's not the case or detail the reasoning here in case someone else want to dig into it ? I don't have an example of the validity of an attribute changing inside a scope right now 😄

areveny:
Yeah, consider this:

counter.b is used twice in the same function. The cache assumes that all attributes within a scope with the same name have the same result. The cache returns that counter.b is valid in the second access, the current code infers on each access to ensure it is valid to catch this situation.

```
class RootCounter:
    elements = [1, 2, 3]
    b = [1, 2, 3]

class RootCounter2:
    elements = [1, 2, 3]

class CounterContainer1:

    counter = RootCounter()

    def fn(self):
        print(counter.elements)
        print(counter.b)
        counter = RootCounter2()
        print(counter.b) # no-member
```

PR #5696, opened 2022-01-18
Add tests for #4826

### LINKED ISSUE #4826: False positive with operator+ and lists (no-member)
### Bug description

When using `operator+` to merge two lists (in the context of a class method?), pylint incorrectly thinks that the objects in the combined list are also of type `list`.

The simplest example I can construct is the following:

```python
# pylint: disable=C0103,C0114,C0115,C0116,R0903

class A:
    def __init__(self):
        self.Alist = []
        self.data = []

    def run(self):
        for a in [self] + self.Alist:
            for _ in a.data:  # # <-- causes "no-member" warning
                pass
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:12:21: E1101: Instance of 'list' has no 'data' member (no-member)
```

### Expected behavior

I would expect there to be no warning, since both `self` and `self.Alist` are (nominally) of type `A`, which does have a member called `data`.

If instead of using `[self] + self.Alist`, any of the following will _not_ result in a warning:
* `for a in [self]`
* `for a in self.Alist`

### Pylint version

```shell
pylint 2.9.6
astroid 2.6.5
Python 3.7.4 (default, Aug 13 2019, 20:35:49) 
[GCC 7.3.0]
```

### OS / Environment

CentOS 8.4

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Ref https://github.com/PyCQA/astroid/pull/1360

Closes #4826.


Pierre-Sassoulas:
Do we want to release astroid 2.10 before pylint 2.13 ?

DanielNoord:
Preferably, but it shouldn't be a blocker imo. There are a lot of PRs stuck in review limbo at `astroid` (in `pylint` as well..) so I don't think we should for those.

Pierre-Sassoulas (approved):
Changes can be merged as soon as astroid is updated and test run.

Pierre-Sassoulas:
> There are a lot of PRs stuck in review limbo at astroid (in pylint as well..)

Yes, it's hard to keep up, I can't even find the time to finish my own MR 😄 

>  so I don't think we should for those.

I did not get that part very well. Do you think we should not release astroid and handle the MR that are currentely in limbo ? Or release despite them ?

DanielNoord:
> I did not get that part very well. Do you think we should not release astroid and handle the MR that are currentely in limbo ? Or release despite them ?

I think we should at least merge/finish some of the open PRs, otherwise `2.10` will be very empty. However, we should also probably not wait for all PRs to be closed.

I think we should try to make sure that `2.10` includes the `distutils` fixes, the change (or rejection) of my `NamedTuple` PR, and the caching PR(s). I'm obviously biased but with the approval of `NamedTuple` I could continue working on actually making use of the new instance and doing the same of dataclasses, which would also benefit from their own instance I believe.

PR #5700, opened 2022-01-19
``unused-variable`` when `nonlocal` name in a multiple-assignment

### LINKED ISSUE #3781: False positive for unused variable (W0612) on tuple unpacking
### Steps to reproduce
```python3
def function():
    a, b = 0, []

    print(b)

    def inner():
        nonlocal a
        b.append(a)
        a += 1

    return inner()
```

### Current behavior
```
file.py:2:4: W0612: Unused variable 'a' (unused-variable)
```
Interestingly, if you remove the 9th line (`a += 1`), the error goes away.

### Expected behavior
No errors.

### pylint --version output
```
pylint 2.5.2
astroid 2.4.1
Python 3.8.5 (default, Jul 31 2020, 00:00:00) 
[GCC 10.2.1 20200723 (Red Hat 10.2.1-1)]
```

mbyrnepr2:
The `nodes.Tuple` type is currently not considered in the logic to determine if the name is a member of the nonlocal set;
With this merge-request, if the parent of the name which is assigned is of type `nodes.Tuple` (as is the case with tuple-unpacking) then a `return` is reached which prevents emitting the warning message.

## Type of Changes

## Description

Closes #3781

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -48,7 +48,7 @@ Release date: TBA
 
 * Fixed false positive for ``unused-variable`` when a ``nonlocal`` name is assigned as part of a multi-name assignment.
 
-  Cloases #3781
+  Closes #3781
 
 * Added ``lru-cache-decorating-method`` checker with checks for the use of ``functools.lru_cache``
   on class methods. This is unrecommended as it creates memory leaks by never letting the instance
--- a/ChangeLog
+++ b/ChangeLog
@@ -46,6 +46,10 @@ Release date: TBA
 
   Closes #3088
 
+* Fixed false positive for ``unused-variable`` when a `nonlocal` name is assigned as part of a multi-name assignment.
+
+  Cloases #3781
+
 * Added ``lru-cache-decorating-method`` checker with checks for the use of ``functools.lru_cache``
   on class methods. This is unrecommended as it creates memory leaks by never letting the instance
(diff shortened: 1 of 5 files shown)


tests/functional/u/undefined/undefined_variable.py
```diff
@@ -440,3 +440,16 @@ def my_method(self) -> Coords:
             pass
 
     print(MyObject)
+
```
mbyrnepr2, line 443: This test should probably live somewhere else

Pierre-Sassoulas (approved):
LGTM, thank you @mbyrnepr2 !

ChangeLog
```diff
@@ -46,6 +46,10 @@ Release date: TBA
 
   Closes #3088
 
+* Fixed false positive for ``unused-variable`` when a ``nonlocal`` name is assigned as part of a multi-name assignment.
+
+  Cloases #3781
```
jacobtylerwalls, line 51:
```suggestion
  Closes #3781
```
:-)

DanielNoord (approved):
If only all fixes could be this easy 😄

ChangeLog
mbyrnepr2: Drat how did I miss that! Thank you!

PR #5708, opened 2022-01-23
False negative - tuple unpacking

### LINKED ISSUE #5707: False negative - tuple unpacking & empty list
### Bug description

As mentioned in https://github.com/PyCQA/pylint/issues/5671, the Pylint 2.13 logic introduced the false negative with tuple unpacking.

The following example would have emitted an `unbalanced-tuple-unpacking` warning before the change of https://github.com/PyCQA/pylint/pull/5346:

mytest.py:
```
def f(x):
    l = []
    for i in x:
        l.append(i)
    return l

a, b, c, d, e, f, g = f('12')
```

In 2.13 it does not emit any warning. Would be nice to fix the unintended change to the behaviour introduced in https://github.com/PyCQA/pylint/pull/5346.

### Configuration

_No response_

### Command used

```shell
pylint mytest.py
```

### Pylint output

```shell
```

### Expected behavior

mytest.py:7:0: W0632: Possible unbalanced tuple unpacking with sequence defined at line 2: left side has 7 label(s), right side has 0 value(s) (unbalanced-tuple-unpacking)

### Pylint version

```shell
pylint 2.13.0-dev0
astroid 2.9.3
Python 3.10.0b2 (v3.10.0b2:317314165a, May 31 2021, 10:02:22) [Clang 12.0.5 (clang-1205.0.22.9)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

mbyrnepr2:
## Type of Changes

## Description

Closes #5707


tests/functional/u/unbalanced_tuple_unpacking.py
```diff
@@ -121,10 +137,21 @@ def my_sum(self):
...
         return first + second + third + fourth
 
     def sum_unpack_3_into_2(self):
         """Attempt to unpack 3 variables into 2"""
-        first, second = self # [unbalanced-tuple-unpacking]
+        first, second = self  # [unbalanced-tuple-unpacking]
         return first + second
+
+
+def my_function(mystring):
```
mbyrnepr2, line 149: The added test; The other changed content is from running `black` on the module.

pylint/checkers/variables.py
```diff
@@ -2400,15 +2400,8 @@ def _check_unpacking(self, inferred, node, targets):
...
-        if isinstance(inferred, (nodes.Tuple, nodes.List)):
-            values = inferred.itered()
-        elif isinstance(inferred, astroid.Instance) and any(
-            ancestor.qname() == "typing.NamedTuple" for ancestor in inferred.ancestors()
-        ):
-            values = [i for i in inferred.values() if isinstance(i, nodes.AssignName)]
-
-        if values:
+        values = self._get_values_to_unpack(inferred)
+        if values is not None:
```
DanielNoord, line 2404: Can't we return an empty list and then just compare the lengths? Thereby removing this `if`.

pylint/checkers/variables.py
```diff
@@ -2430,6 +2423,17 @@ def _check_unpacking(self, inferred, node, targets):
...
 
+    @staticmethod
+    def _get_values_to_unpack(node: nodes.Assign) -> Optional[List]:
+        """Return the list of values of the `Assign` node"""
+        if isinstance(node, (nodes.Tuple, nodes.List)):
+            return node.itered()
+        if isinstance(node, astroid.Instance) and any(
+            ancestor.qname() == "typing.NamedTuple" for ancestor in node.ancestors()
+        ):
+            return [i for i in node.values() if isinstance(i, nodes.AssignName)]
```
DanielNoord, line 2434: Shouldn't this also cover `AnnAssign`?

ChangeLog
```diff
@@ -37,6 +37,10 @@ Release date: TBA
 
   Closes #5281
 
+* Fixed false negative ``unpacking-non-sequence`` when value is an empty list.
```
DanielNoord, line 40: The changelog entry in `2.13` is missing

pylint/checkers/variables.py
DanielNoord: Never mind, I see this is just copied from previous code. Perhaps a follow-up issue worthy?

DanielNoord (changes requested):
Small question(s) 😄

pylint/checkers/variables.py
mbyrnepr2: It would be nice to remove the indented block but it changes the logic to do that.

Returning None indicates that the right-hand-side of the assignment is not something that can be unpacked.
So, if something like `a, b = 1` is encountered we want it to go to the `elif not utils.is_iterable(inferred):` block to trigger `unpacking-non-sequence`.

ChangeLog
mbyrnepr2: Forgive me if I'm missing something obvious but this is the 2.13 Changelog, right?

pylint/checkers/variables.py
mbyrnepr2: Is that situation possible without running into E0001?
For example:
`a, b: List = [1,2]` # only single target (not tuple) can be annotated)?

ChangeLog
DanielNoord: We're using two separate documents to track changelog entries (don't ask me why.. 😅).
There is `Changelog` and `doc/whatsnew/2.13`, I think the latter is used to update the changelogs on the documentation pages.

pylint/checkers/variables.py
DanielNoord: Ah my bad, should have spotted that!

pylint/checkers/variables.py
DanielNoord: Hmm, seems I wasn't the sharpest when I reviewed this previously. You're right 😄

DanielNoord (changes requested):

pylint/checkers/variables.py
mbyrnepr2: I know how you feel!

ChangeLog
Pierre-Sassoulas: > We're using two separate documents to track changelog entries (don't ask me why.. 😅).

Well that's a good question though 😄 Unifying the two would take some time but the benefit of having only one place to modify would be great. Black has pretty great check on changelog entry, we may take example. Do you want to open the issue as you had the idea @DanielNoord or should I do it ?

DanielNoord: Feel free to open the issue @Pierre-Sassoulas

doc/whatsnew/2.13.rst
```diff
@@ -279,6 +279,11 @@ Other Changes
 
   Closes #3781
 
+* Fix false positive - Allow unpacking of ``self`` in a subclass of ``typing.NamedTuple``.
+
+  Closes #5312
+  Closes #5671
```
DanielNoord, line 285: Hmm I wonder if we should add a changelog entry for the change introduce in this one as well. I'll leave these changelog fixes to @Pierre-Sassoulas's discretion. 

I'm approving the rest of it!

DanielNoord (approved):

ChangeLog
Pierre-Sassoulas: Opened #5728

ChangeLog
```diff
@@ -37,6 +37,10 @@ Release date: TBA
...
+
+ Closes #5707
+
```
Pierre-Sassoulas, line 43:
```suggestion

* Fix false positive - Allow unpacking of ``self`` in a subclass of ``typing.NamedTuple``.

  Closes #5312
  Closes #5671
```

doc/whatsnew/2.13.rst
```diff
@@ -279,6 +279,11 @@ Other Changes
...
+
```
Pierre-Sassoulas, line 286:
```suggestion

* Fixed false negative ``unpacking-non-sequence`` when value is an empty list.

  Closes #5707
```

tests/functional/u/unbalanced_tuple_unpacking.py
```diff
@@ -121,10 +137,21 @@ def my_sum(self):
...
+    """The number of items on the right-hand-side of the assignment to this function is not known"""
+    mylist = []
+    for item in mystring:
+        mylist.append(item)
+    return mylist
+
+
+a, b = my_function("12")  # [unbalanced-tuple-unpacking]
```
Pierre-Sassoulas, line 157:
```suggestion
a, b = my_function("12")  # [unbalanced-tuple-unpacking]
c = my_function("12")
d, *_ = my_function("12") 
```
Checking that there are no false positives, I suppose those should not raise either, right ?

Pierre-Sassoulas (changes requested):
Thank you, this looks great already ! I suggested some more tests cases, let me know what you think @mbyrnepr2

ChangeLog
mbyrnepr2: Updated. Note the 5671 should never have been there; removed.

tests/functional/u/unbalanced_tuple_unpacking.py
mbyrnepr2: Looks sensible; will get back to this asap.

pylint/checkers/variables.py
```diff
@@ -2430,6 +2423,17 @@ def _check_unpacking(self, inferred, node, targets):
                 args=(_get_unpacking_extra_info(node, inferred),),
             )
 
+    @staticmethod
+    def _get_values_to_unpack(node: nodes.Assign) -> Optional[List]:
```
DanielNoord, line 2427: Could you add typing to the List? I guess:
```suggestion
    def _get_values_to_unpack(node: nodes.Assign) -> Optional[List[nodes.AssignName]]:
```
Or:
```suggestion
    def _get_values_to_unpack(node: nodes.Assign) -> Optional[List[nodes.NodeNG]]:
```

DanielNoord (changes requested):

pylint/checkers/variables.py
mbyrnepr2: Good point; I think it is more accurate to use `Optional[List[nodes.NodeNG]]` since the nodes which are unpacked won't necessarily be of type `nodes.AssignName`.
Moreover we can make the function name more accurate (`_nodes_to_unpack`?) & the type of the parameter should be `nodes.NodeNG` since this value originates from the returned value of the `safe_infer` method.

ChangeLog
```diff
@@ -42,6 +42,14 @@ Release date: TBA
 
   Closes #5281
 
+* Fix false positive - Allow unpacking of ``self`` in a subclass of ``typing.NamedTuple``.
```
DanielNoord, line 45: Just for the commit history I would be better to leave this where it currently is, right? This is not a blocking issue for me though.

DanielNoord (approved):

Pierre-Sassoulas (approved):
Thank you @mbyrnepr2 , great addition to 2.13 !

PR #5709, opened 2022-01-23
Add confidence level `CONTROL_FLOW`

jacobtylerwalls:
## Type of Changes

## Description

Add a new confidence level `CONTROL_FLOW` for messages emitted that depend on assumptions about control flow execution. Discussed at https://github.com/PyCQA/pylint/pull/5684#issuecomment-1019552360.


pylint/checkers/variables.py
```diff
@@ -1400,7 +1406,17 @@ def _check_consumer(
         if found_nodes is None:
             return (VariableVisitConsumerAction.CONTINUE, None)
         if not found_nodes:
-            self.add_message("used-before-assignment", args=node.name, node=node)
+            if node.name in current_consumer.consumed_uncertain:
+                confidence = CONTROL_FLOW
+            else:
+                # Set to HIGH in future?
+                confidence = None
```
Pierre-Sassoulas, line 1413: ``HIGH`` only means we do not use inference, I think we can use ``HIGH``.
```suggestion
                confidence = HIGH
```

Pierre-Sassoulas (commented):
Looks good thank you !

The multiple place to modify to add a value makes me want to refactor Confidence to be an Enum instead because namedtuple are slow to import (as it's an exec). But this would make a breaking change and adding a confidence value will not happen that often if we're being honest. But maybe we need to do this breaking change before using confidence more, what do you think @DanielNoord  ?

Pierre-Sassoulas (approved):

pylint/checkers/variables.py
```diff
@@ -1400,7 +1406,16 @@ def _check_consumer(
         if found_nodes is None:
             return (VariableVisitConsumerAction.CONTINUE, None)
         if not found_nodes:
-            self.add_message("used-before-assignment", args=node.name, node=node)
+            if node.name in current_consumer.consumed_uncertain:
+                confidence = CONTROL_FLOW
+            else:
+                confidence = HIGH
```
DanielNoord, line 1412: @jacobtylerwalls You're the expert here, but are we sure we never use inference in any of the method used in control flow "inference"?

pylint/checkers/variables.py
```diff
@@ -1400,7 +1406,16 @@ def _check_consumer(
...
+            self.add_message(
+                "used-before-assignment",
```
DanielNoord, line 1414: Do we want to add the level to the four other places where this message gets raised?

pylint/checkers/variables.py
jacobtylerwalls: Right, I wondered that, too. I'm sort of sure, but I'll take another look tomorrow.

jacobtylerwalls: All I can see so far is inference being used for some of the cases for `undefined-variable`.

jacobtylerwalls: I added a failing assertion in my local editable install of `astroid` to crash when inferring, and then pylinted some files from the test suite with --disable=all, --enable=used-before-assignment. I only saw some boilerplate inferences when loading the original ast trees or pylint itself.

DanielNoord: That brings up another point: do we want to use this for `undefined-variable` as well? And if so, do we want to change `CONTROL_FLOW` to also include inference? ie., `CONTROL_FLOW` represents control flow assumptions + inference.

jacobtylerwalls: I added confidence `HIGH` for `undefined-variable`. I couldn't find any cases where `INFERRED` and `CONTROL_FLOW` are both used: the one edge case of `undefined-variable` that uses inference does not use control flow.

Pierre-Sassoulas:
I fixed the conflict and merged main but I think this could be squashed then rebased for clarity.

DanielNoord (approved):

pylint/checkers/variables.py
DanielNoord: Thanks!

PR #5711, opened 2022-01-24
Fix #5654: Fix false negative for `undefined-variable` for unused type annotation after the same name appears multiple times in a comprehension

### LINKED ISSUE #5654: Should emit `used-before-assignment` for unused type annotation after the same name appears multiple times in a comprehension
### Bug description

```python
"""https://github.com/PyCQA/pylint/issues/5654"""
from typing import Any
def func():
    """Should raise used-before-assignment on the last line"""
    my_none: Any
    _ = [print(sep=my_none, end=my_none) for my_none in (None, None)]
    print(my_none)
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
On `main`, it's raised on the wrong line (the comprehension line):
test.py:5:37: E0602: Undefined variable 'my_none' (undefined-variable)

Your code has been rated at 0.00/10 (previous run: 10.00/10, -10.00)

After #5651, no warnings at all:

$ pylint ../test.py

Your code has been rated at 10.00/10 (previous run: 0.00/10, +10.00)
```

### Expected behavior

test.py:7:37: E0601: Using variable 'my_none' before assignment (used-before-assignment)

### Pylint version

```shell
pylint 2.13.0-dev0
astroid 2.9.2
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #5654

Added another exception to the "return early if a name already consumed is encountered again" logic, again having to do with comprehensions. This is getting special-case heavy, so comprehension handling might need to be factored differently going forward, but the challenge was not breaking the `unused-argument` warning raised here: https://github.com/PyCQA/pylint/blob/7611ec443509888b96087d1519be9d73cf3b4c79/tests/functional/u/unused/unused_argument.py#L43-L48

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1393,7 +1393,7 @@ class VariablesChecker(BaseChecker):
                         and node.parent in node.parent.parent.ifs
                     )
                     # Or homonyms against values to keyword arguments
-                    # (like "var" in [func(arg=var) for var in expr()])
+                    # (like "var" in "[func(arg=var) for var in expr()]")
                     or (
                         isinstance(node.scope(), nodes.ComprehensionScope)
                         and isinstance(node.parent, (nodes.Call, nodes.Keyword))
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1388,8 +1388,16 @@ class VariablesChecker(BaseChecker):
                 # (like "if x" in "[x for x in expr() if x]")
                 # https://github.com/PyCQA/pylint/issues/5586
                 and not (
-                    isinstance(node.parent.parent, nodes.Comprehension)
-                    and node.parent in node.parent.parent.ifs
+                    (
+                        isinstance(node.parent.parent, nodes.Comprehension)
+                        and node.parent in node.parent.parent.ifs
+                    )
+                    # Or homonyms against values to keyword arguments
+                    # (like "var" in [func(arg=var) for var in expr()])
+                    or (
+                        isinstance(node.scope(), nodes.ComprehensionScope)
+                        and isinstance(node.parent, (nodes.Call, nodes.Keyword))
+                    )
                 )
             ):
                 self._check_late_binding_closure(node)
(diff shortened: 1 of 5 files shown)


jacobtylerwalls:
Force-pushed to correct typo in commit msg (also fixed in PR title)

pylint/checkers/variables.py
```diff
@@ -1388,8 +1388,16 @@ def _check_consumer(
...
                 # https://github.com/PyCQA/pylint/issues/5586
                 and not (
-                    isinstance(node.parent.parent, nodes.Comprehension)
-                    and node.parent in node.parent.parent.ifs
+                    (
+                        isinstance(node.parent.parent, nodes.Comprehension)
+                        and node.parent in node.parent.parent.ifs
+                    )
+                    # Or homonyms against values to keyword arguments
+                    # (like "var" in [func(arg=var) for var in expr()])
```
Pierre-Sassoulas, line 1396:
```suggestion
                    # (like "var" in "[func(arg=var) for var in expr()]")
```

tests/functional/u/undefined/undefined_variable_py38.py
```diff
@@ -123,5 +123,19 @@ def type_annotation_used_after_comprehension():
 
 def type_annotation_unused_after_comprehension():
     """https://github.com/PyCQA/pylint/issues/5326"""
+    my_int: int  # [unused-variable]
```
Pierre-Sassoulas, line 126:
```suggestion
    # TODO follow up in https://github.com/PyCQA/pylint/issues/5713
    my_int: int  # [unused-variable]
```

Pierre-Sassoulas (approved):
LGTM

tests/functional/u/undefined/undefined_variable_py38.py
jacobtylerwalls: I'll just go ahead and open the PR for that; it's not 100% slam-dunk that it's worth changing the message, but if there's a time to do it, it's probably soon before more existing disables out there have to be changed.

DanielNoord (approved):

DanielNoord:
One test for this commit failed on `main`. @jacobtylerwalls could you perhaps take a look at this?

Pierre-Sassoulas:
The tests were both green before the merge, maybe the two last merge requests are incompatible together ?

PR #5718, opened 2022-01-25
Fix #5713: Emit `used-before-assignment` instead of `undefined-variable` when accessing unused type annotations

### LINKED ISSUE #5713: Emit `used-before-assignment` instead of `undefined-variable` for unused type annotations
### Bug description

I think we can raise a more appropriate message for cases like:

```python
def type_annotation_used_improperly_after_comprehension():
    """https://github.com/PyCQA/pylint/issues/5654"""
    my_int: int
    _ = [print(sep=my_int, end=my_int) for my_int in range(10)]  <--should be used-before-assignment
    print(my_int)  # [undefined-variable]
```

Because the python interpreter produces:

```
UnboundLocalError: local variable 'my_int' referenced before assignment
```

This feature was added in #5158, so perhaps recent enough to change, but not sure if it's too much churn/worth it to bother.

### Expected behavior

Raise `used-before-assignment` instead of `unused-variable` for the above test from `undefined_variable_py38.py`

### Pylint version

```shell
pylint 2.13.0-dev0
astroid 2.9.3
```

jacobtylerwalls:
## Type of Changes

## Description

This one is somewhat pedantic, sorry! I simply noticed that we could attempt to raise more targeted messages so that `undefined-variable` corresponds generally with `NameError` and `used-before-assignment` corresponds generally with `UnboundLocalError`. The only use of the message affected by this PR is new in 2.12, so hopefully not breaking too many existing disables. But I suppose that should be balanced against people suppressing `undefined-variable` that would be glad to get `used-before-assignment` coming through now.

See example shell session drawn from existing unit tests adjusted in this PR:

```python
In [1]: G: int
   ...: 
   ...: 
   ...: def a():
   ...:     global G
   ...:     print(G)
   ...: 
   ...: 
   ...: def bound_value():
   ...:     variable: int
   ...:     print(variable)
   ...: 
   ...: 
   ...: def decorator_returning_incorrect_function():
   ...:     """A decorator that returns a wrapper function with decoupled typing"""
   ...: 
   ...:     def wrapper_with_type_and_no_value():
   ...:         # This emits NameError, so undefined-variable is okay
   ...:         # even though the traceback refers to
   ...:         print(var)  # [undefined-variable]
   ...: 
   ...:     var: int
   ...:     return wrapper_with_type_and_no_value

In [2]: a()
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Input In [2], in <module>
----> 1 a()

Input In [1], in a()
      4 def a():
      5     global G
----> 6     print(G)

NameError: name 'G' is not defined

In [3]: bound_value()
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
Input In [3], in <module>
----> 1 bound_value()

Input In [1], in bound_value()
      9 def bound_value():
     10     variable: int
---> 11     print(variable)

UnboundLocalError: local variable 'variable' referenced before assignment

In [4]: decorator_returning_incorrect_function()()
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Input In [4], in <module>
----> 1 decorator_returning_incorrect_function()()

Input In [1], in decorator_returning_incorrect_function.<locals>.wrapper_with_type_and_no_value()
     17 def wrapper_with_type_and_no_value():
     18     # This emits NameError, so undefined-variable is okay
     19     # even though the traceback refers to "used before assignment"
---> 20     print(var)

NameError: free variable 'var' referenced before assignment in enclosing scope

```

Closes #5713

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1544,9 +1544,13 @@ class VariablesChecker(BaseChecker):
 
         elif self._is_only_type_assignment(node, defstmt):
             if node.scope().locals.get(node.name):
-                self.add_message("used-before-assignment", args=node.name, node=node, confidence=HIGH)
+                self.add_message(
+                    "used-before-assignment", args=node.name, node=node, confidence=HIGH
+                )
             else:
-                self.add_message("undefined-variable", args=node.name, node=node, confidence=HIGH)
+                self.add_message(
+                    "undefined-variable", args=node.name, node=node, confidence=HIGH
+                )
             return (VariableVisitConsumerAction.CONSUME, found_nodes)
 
         elif isinstance(defstmt, nodes.ClassDef):
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1544,9 +1544,9 @@ class VariablesChecker(BaseChecker):
 
         elif self._is_only_type_assignment(node, defstmt):
             if node.scope().locals.get(node.name):
-                self.add_message("used-before-assignment", args=node.name, node=node)
+                self.add_message("used-before-assignment", args=node.name, node=node, confidence=HIGH)
             else:
-                self.add_message("undefined-variable", args=node.name, node=node)
+                self.add_message("undefined-variable", args=node.name, node=node, confidence=HIGH)
             return (VariableVisitConsumerAction.CONSUME, found_nodes)
 
         elif isinstance(defstmt, nodes.ClassDef):
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1543,7 +1543,10 @@ class VariablesChecker(BaseChecker):
                     )
 
         elif self._is_only_type_assignment(node, defstmt):
-            self.add_message("undefined-variable", args=node.name, node=node)
+            if node.scope().locals.get(node.name):
+                self.add_message("used-before-assignment", args=node.name, node=node)
+            else:
+                self.add_message("undefined-variable", args=node.name, node=node)
             return (VariableVisitConsumerAction.CONSUME, found_nodes)
 
         elif isinstance(defstmt, nodes.ClassDef):
--- a/ChangeLog
+++ b/ChangeLog
@@ -89,6 +89,11 @@ Release date: TBA
 
 * Pyreverse - add output in mermaidjs format
 
+* Emit ``used-before-assignment`` instead of ``undefined-variable`` when attempting
+  to access unused type annotations.
+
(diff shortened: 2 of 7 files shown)


pylint/checkers/variables.py
```diff
@@ -1543,7 +1543,10 @@ def _check_consumer(
                     )
 
         elif self._is_only_type_assignment(node, defstmt):
-            self.add_message("undefined-variable", args=node.name, node=node)
+            if node.scope().locals.get(node.name):
```
Pierre-Sassoulas, line 1546: I did not look in detail so it's an open question but performance wise would it make sense to create another function than `_is_only_type_assignment`  according to you? It seems it handle the two use case then we have to do another check to separate the use cases.

pylint/checkers/variables.py
```diff
@@ -1543,7 +1543,10 @@ def _check_consumer(
...
+                self.add_message("used-before-assignment", args=node.name, node=node)
+            else:
+                self.add_message("undefined-variable", args=node.name, node=node)
```
Pierre-Sassoulas, line 1549: Maybe we could start adding the confidence if it's easy enough to determine, what do you think ?

Pierre-Sassoulas (commented):
It's a nice consistency refactor, thank you :+1: I think we can do it as the number of false positives for those two message is probably low and of disable of those message even lower.

pylint/checkers/variables.py
jacobtylerwalls: I don't think it would be worth the effort, the two checkers would look pretty much identical, since they both would have to keep looking in outer scopes. I don't think it's much of a performance penalty to do one more hit to `scope()` since it's only when emitting a message.

Pierre-Sassoulas (approved):
👌

DanielNoord (approved):
Thanks @jacobtylerwalls for looking into this. I think this makes sense (even though somewhat breaking)!

DanielNoord:
This fails on `main` as well. I think it is related to the fail of #5711. Perhaps @jacobtylerwalls can take a look at them together and do a follow-up PR?

jacobtylerwalls:
@DanielNoord actually this failure looks "random" --> 

```
>               _winapi.TerminateProcess(int(self._handle), TERMINATE)
E               PermissionError: [WinError 5] Access is denied
```

Note the test suite passed on all other runners. 🤔

Pierre-Sassoulas:
> this failure looks "random" 

I came to the same conclusion when looking at it (I'm notified if what I merge fail on main).

PR #5726, opened 2022-01-26
Fix test conflicts

jacobtylerwalls:
## Type of Changes

## Description

After f85fb8d975d17f411e5f91ed2d5f21ba999e121e, certain unused type annotations raise `used-before-assignment` rather than `undefined-variable`. The subsequent commit 0fb6a120c6f0fbc94e305542fe29e790274be096 was drafted before the prior change was made, and it addressed a related false negative, so this commit just applies the intended change of message type.

Sorry for not pointing this out during review; I was figuring I would have a short interval to fix this when one of them got merged!


Pierre-Sassoulas (approved):
Thank you for the fast fix ! 

This kind of thing are rare enough that forcing a rebase on the main branch before merging would not be worth it imo. I should be more mindful when the change seems linked in the future though :smile:

PR #5727, opened 2022-01-26
Fix #5399: Fix false negatives for further variable messages for invalid type annotations or default arguments

### LINKED ISSUE #5399: Inconsistent undefined-variable reporting
### Bug description

The following code:
```python3
# pylint: disable=multiple-statements, no-self-use, unused-argument, used-before-assignment
class A:
    def x(self, o: A) -> bool: ...
    def z(self) -> A: ...
    def t(self) -> A: ...
```
produces only 1 error report instead of 3.

### Command used

```shell
pylint test.pyi
```

### Pylint output

```shell
test.pyi:5: [E0602(undefined-variable), A.z] Undefined variable 'A'
```

### Expected behavior

Error should be reported in lines 3, 4 and 5.

### Pylint version

```shell
pylint 2.12.1
astroid 2.9.0
Python 3.8.10 (default, Sep 28 2021, 16:10:42) 
[GCC 9.3.0]
```

### OS / Environment

Ubuntu 20.04.3 LTS \n \l

Linux ubuntu 5.4.0-90-generic #101-Ubuntu SMP Fri Oct 15 20:00:55 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

jacobtylerwalls:
## Type of Changes

## Description
There were two _short-circuit-and-consume-names_ paths in the variables checker preventing reporting further errors:
 - undefined variable as default function argument
 - undefined variable as type annotation

Removing some of these "special early consume" messages has the benefit that going through the regular path raises the correct flavor of `used-before-assignment` versus `undefined-variable`.

Closes #5399

***

Will need a rebase after #5709. (edit: done!)

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1528,10 +1528,7 @@ class VariablesChecker(BaseChecker):
                         )
                         and node.name in node.root().locals
                     ):
-                        self.add_message(
-                            "undefined-variable", args=node.name, node=node
-                        )
-                        return (VariableVisitConsumerAction.CONSUME, found_nodes)
+                        return (VariableVisitConsumerAction.CONTINUE, None)
 
             elif base_scope_type != "lambda":
                 # E0601 may *not* occurs in lambda scope.
@@ -1585,9 +1582,7 @@ class VariablesChecker(BaseChecker):
         elif isinstance(defstmt, nodes.ClassDef):
             is_first_level_ref = self._is_first_level_self_reference(node, defstmt)
             if is_first_level_ref == 2:
-                self.add_message(
-                    "used-before-assignment", node=node, args=node.name, confidence=HIGH
-                )
+                return (VariableVisitConsumerAction.CONTINUE, None)
             if is_first_level_ref:
                 return (VariableVisitConsumerAction.RETURN, None)
 
@@ -1988,15 +1983,15 @@ class VariablesChecker(BaseChecker):
         refers to its own class.
 
         Return values correspond to:
-            0 = Continue
+            0 = Continue and consume names
             1 = Break
-            2 = Break + emit message
+            2 = Continue without consuming names
         """
         if node.frame(future=True).parent == defstmt and node.statement(
             future=True
         ) == node.frame(future=True):
             # Check if used as type annotation
-            # Break but don't emit message if postponed evaluation is enabled
+            # Break if postponed evaluation is enabled
             if utils.is_node_in_type_annotation_context(node):
                 if not utils.is_postponed_evaluation_enabled(node):
                     return 2
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1532,6 +1532,8 @@ class VariablesChecker(BaseChecker):
                         )
                         and node.name in node.root().locals
                     ):
+                        if defined_by_stmt:
+                            current_consumer.mark_as_consumed(node.name, [node])
                         return (VariableVisitConsumerAction.CONTINUE, None)
 
             elif base_scope_type != "lambda":
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -170,7 +170,7 @@ TYPING_NAMES = frozenset(
 
 
 class VariableVisitConsumerAction(Enum):
-    """Used after _visit_consumer to determine the action to be taken
+    """Used after _check_consumer to determine the action to be taken
 
     Continue -> continue loop to next consumer
     Return -> return and thereby break the loop
@@ -180,6 +180,18 @@ class VariableVisitConsumerAction(Enum):
     CONTINUE = 0
     RETURN = 1
     CONSUME = 2
+
+
+VariableVisitConsumerActionAndOptionalNodesType = Union[
+    Tuple[
+        Union[
+            Literal[VariableVisitConsumerAction.CONTINUE],
+            Literal[VariableVisitConsumerAction.RETURN],
+        ],
+        None,
+    ],
+    Tuple[Literal[VariableVisitConsumerAction.CONSUME], List[nodes.NodeNG]],
+]
 
 
 def _is_from_future_import(stmt, name):
@@ -1371,16 +1383,7 @@ class VariablesChecker(BaseChecker):
         current_consumer: NamesConsumer,
         consumer_level: int,
         base_scope_type: Any,
-    ) -> Union[
-        Tuple[
-            Union[
-                Literal[VariableVisitConsumerAction.CONTINUE],
-                Literal[VariableVisitConsumerAction.RETURN],
-            ],
-            None,
-        ],
-        Tuple[Literal[VariableVisitConsumerAction.CONSUME], List[nodes.NodeNG]],
-    ]:
+    ) -> VariableVisitConsumerActionAndOptionalNodesType:
         """Checks a consumer for conditions that should trigger messages"""
         # If the name has already been consumed, only check it's not a loop
         # variable used outside the loop.
@@ -1580,11 +1583,7 @@ class VariablesChecker(BaseChecker):
             return (VariableVisitConsumerAction.CONSUME, found_nodes)
 
         elif isinstance(defstmt, nodes.ClassDef):
-            is_first_level_ref = self._is_first_level_self_reference(node, defstmt)
-            if is_first_level_ref == 2:
-                return (VariableVisitConsumerAction.CONTINUE, None)
-            if is_first_level_ref:
-                return (VariableVisitConsumerAction.RETURN, None)
+            return self._is_first_level_self_reference(node, defstmt, found_nodes)
 
         elif isinstance(defnode, nodes.NamedExpr):
             if isinstance(defnode.parent, nodes.IfExp):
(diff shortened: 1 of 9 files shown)


tests/functional/u/undefined/undefined_variable.py
```diff
@@ -133,7 +133,7 @@ class Ancestor1(object):
     """ No op """
 
 NANA = BAT # [undefined-variable]
-del BAT
+del BAT  # [undefined-variable]
```
Pierre-Sassoulas, line 136: This is already warned against on the previous line, isn't it a "false positive" or more accurately a little annoying to have multiple warning for the same problem ?

tests/functional/u/use/used_before_assignment_py37.py
```diff
@@ -14,7 +14,7 @@ def second_correct_typing_method(self, other: List[MyClass]) -> bool:
         return self == other[0]
 
     def incorrect_default_method(
-        self, other=MyClass() # [used-before-assignment]
+        self, other=MyClass() # [undefined-variable]
```
Pierre-Sassoulas, line 17: :+1: It matches the real error netter as it's `NameError: name 'MyClass' is not defined`

pylint/checkers/variables.py
```diff
@@ -1988,15 +1983,15 @@ def _is_first_level_self_reference(
         refers to its own class.
 
         Return values correspond to:
-            0 = Continue
+            0 = Continue and consume names
```
Pierre-Sassoulas, line 1986: Maybe we could rename the enum attribute and document this in the enum directly.

Pierre-Sassoulas (commented):
LGTM. I like the simplification when simply removing the add_message makes the code simpler and removes an inconsistency.

tests/functional/u/undefined/undefined_variable.py
jacobtylerwalls: I figured it was the same as the original report in #5399. On consecutive lines it could be a little annoying, but forty lines down it could be informative!

jacobtylerwalls: Also, it's consistent with the general scenario on `main`:

```
% cat a.py
class A():
  jazz
  jazz


% pylint a.py
************* Module a
a.py:2:2: W0104: Statement seems to have no effect (pointless-statement)
a.py:2:2: E0602: Undefined variable 'jazz' (undefined-variable)
a.py:3:2: W0104: Statement seems to have no effect (pointless-statement)
a.py:3:2: E0602: Undefined variable 'jazz' (undefined-variable)

```

Pierre-Sassoulas: > On consecutive lines it could be a little annoying, but forty lines down it could be informative!

Very fair point, I'm resolving

Pierre-Sassoulas (approved):
Thank you @jacobtylerwalls  !

tests/functional/u/undefined/undefined_variable.py
```diff
@@ -28,8 +28,8 @@ def bad_default(var, default=unknown2):  # [undefined-variable]
     """function with default arg's value set to an nonexistent name"""
     print(var, default)
     print(xxxx)  # [undefined-variable]
-    augvar += 1  # [undefined-variable]
-    del vardel  # [undefined-variable]
+    augvar += 1  # [undefined-variable, unused-variable]
+    del vardel  # [undefined-variable, unused-variable]
```
DanielNoord, line 32: Isn't this a false positive? Why are we reporting `unused-variable` on a `del` operation with a `undefined-variable`.

pylint/checkers/variables.py
```diff
@@ -170,7 +170,7 @@
 
 
 class VariableVisitConsumerAction(Enum):
-    """Used after _visit_consumer to determine the action to be taken
+    """Used after _check_consumer to determine the action to be taken
```
DanielNoord, line 173: This is out of date right? Now that `_is_first_level_self_reference` uses it as well? Perhaps we should make the docstring more general and not refer to method names that could change.

pylint/checkers/variables.py
```diff
@@ -182,6 +182,18 @@ class VariableVisitConsumerAction(Enum):
     CONSUME = 2
 
 
+VariableVisitConsumerActionAndOptionalNodesType = Union[
```
DanielNoord, line 185: Should have done this when I added this 😅

pylint/checkers/variables.py
DanielNoord: Hmm, on further thought. It is still used in `check_consumer` but also in "sub-methods". I would still think a more general docstring makes more sense here.

pylint/checkers/variables.py
```diff
@@ -1982,31 +1976,26 @@ def _is_only_type_assignment(node: nodes.Name, defstmt: nodes.Statement) -> bool
 
     @staticmethod
     def _is_first_level_self_reference(
-        node: nodes.Name, defstmt: nodes.ClassDef
-    ) -> Literal[0, 1, 2]:
+        node: nodes.Name, defstmt: nodes.ClassDef, found_nodes: List[nodes.NodeNG]
+    ) -> VariableVisitConsumerActionAndOptionalNodesType:
```
DanielNoord, line 1980: Nice change here 👍

DanielNoord (commented):
Just some questions. Change itself is very nice! 
Stuff like this make me happy about all the work we've been putting in typing and make the code more readable over the last couple of months. The change to `_is_first_level_self_reference` shows how because of earlier simplification we can now actually improve the messages by building on that previous groundwork. 🎉

tests/functional/u/undefined/undefined_variable.py
Pierre-Sassoulas: If I understand correctly we're going to merge as it is and then open another MR to handle this comment ? I'm ok with this plan.

Pierre-Sassoulas (approved):
👍

tests/functional/u/undefined/undefined_variable.py
DanielNoord: I don't think so, and if @jacobtylerwalls wanted to do so I would advise against.
Nothing personal, but we never know what might come up in any of our personal lives and I don't think it's good to add code we know raises a false-positive instead of correctly warning previously. I have 100% confidence @jacobtylerwalls will get to this, but I don't think we should merge the "faulty" code.

jacobtylerwalls: Addressed this in 53c5200, happy to hear your thoughts. These are statements used and defined in the same line, and we have a variable already to track that, so the approach is to just check that condition and mark as consumed.

pylint/checkers/variables.py
```diff
@@ -1528,10 +1532,9 @@ def _check_consumer(
                         )
                         and node.name in node.root().locals
                     ):
-                        self.add_message(
-                            "undefined-variable", args=node.name, node=node
-                        )
-                        return (VariableVisitConsumerAction.CONSUME, found_nodes)
+                        if defined_by_stmt:
+                            current_consumer.mark_as_consumed(node.name, [node])
+                        return (VariableVisitConsumerAction.CONTINUE, None)
```
DanielNoord, line 1537: I'm probably missing something here, but why can't we return `Consume` with `[node]` instead of `found_nodes`?

jacobtylerwalls: CONTINUE is the only action that will allow further messages to be emitted in the same scope (the original issue being fixed), but when leaving the scope, if it's a `defined_by_stmt` line we want to make sure the name is consumed and does not emit `undefined-variable`.

Otherwise, we would need a fourth VariableVisitConsumerAction action along the lines of CONSUME_AND_CONTINUE.

DanielNoord: Hm, what do you thinks makes most sense? `CONSUME_AND_CONTINUE` sounds the most future-proof to me? I feel like having a `mark_as_consumed` somewhere in this method makes the consumption harder to comprehend again.

jacobtylerwalls: An alternative might be to just reduce the actions to CONTINUE and RETURN and then permit the second element of the tuple to have elements to mark as consumed no matter the first value?

DanielNoord: Hmm that's not bad actually. Performance wise it would be similar, as `if action == CONSUME` or `if consume_nodes` should be quite similar.
Perhaps that should be done in another PR though?

_Look at me thinking this PR showed how good and future-proof our previous code was 😅_

jacobtylerwalls: Follow up PR sounds right.

Just making sure I follow: did you want to avoid adding a fourth consumer action for now?

DanielNoord: Yes, if we know we're going to reduce down to two actions in the foreseeable future we might as well keep it like this.

DanielNoord (approved):
Approving, on the condition that we don't forget about that follow up PR 😄

jacobtylerwalls:
> Approving, on the condition that we don't forget about that follow up PR 😄

Easy! Will rebase #5745 on this when merged.

Pierre-Sassoulas (approved):
Thank you @jacobtylerwalls !

PR #5736, opened 2022-01-28
Clearer error message for useless-else-x type messages

### LINKED ISSUE #5598: ``useless-else-on-loop`` should not be raised on `return`

Pierre-Sassoulas:
## Type of Changes

## Description

See https://github.com/PyCQA/pylint/pull/5614#issuecomment-1023614434 for rationale.

Closes #5598

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,6 +37,10 @@ Release date: TBA
 
   Closes #5281
 
+* Better warning messages for useless else or elif when a function return early.
+
+  Closes #5614
+
 * Fixed false positive ``consider-using-dict-comprehension`` when creating a dict
   using a list of tuples where key AND value vary depending on the same condition.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,7 +37,7 @@ Release date: TBA
 
   Closes #5281
 
-* Better warning messages for useless else or elif when a function return early.
+* Better warning messages for useless else or elif when a function returns early.
 
   Closes #5614
 
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -86,6 +86,10 @@ Other Changes
 
   Closes #5648
 
+* Better warning messages for useless else or elif when a function return early.
+
+  Closes #5614
+
 * Fixed false positive ``consider-using-dict-comprehension`` when creating a dict
   using a list of tuples where key AND value vary depending on the same condition.
 
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -86,7 +86,7 @@ Other Changes
 
   Closes #5648
 
-* Better warning messages for useless else or elif when a function return early.
+* Better warning messages for useless else or elif when a function returns early.
 
   Closes #5614
 
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -239,7 +239,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "with statement assignment and exception handler assignment.",
         ),
         "R1705": (
-            'Unnecessary "%s" after "return"',
+            'Unnecessary "%s" after "return" remove it and de-indent all the code inside it',
             "no-else-return",
             "Used in order to highlight an unnecessary block of "
             "code following an if containing a return statement. "
@@ -339,7 +339,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "Used when an if expression can be replaced with 'bool(test)'. ",
         ),
         "R1720": (
-            'Unnecessary "%s" after "raise"',
+            'Unnecessary "%s" after "raise" remove it and de-indent all the code inside it',
             "no-else-raise",
             "Used in order to highlight an unnecessary block of "
             "code following an if containing a raise statement. "
@@ -360,7 +360,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "Instead of using exit() or quit(), consider using the sys.exit().",
         ),
         "R1723": (
-            'Unnecessary "%s" after "break"',
+            'Unnecessary "%s" after "break" remove it and de-indent all the code inside it',
             "no-else-break",
             "Used in order to highlight an unnecessary block of "
             "code following an if containing a break statement. "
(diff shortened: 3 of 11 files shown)


Pierre-Sassoulas:
@tushar-deepsource what do you think ?

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -239,7 +239,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "with statement assignment and exception handler assignment.",
         ),
         "R1705": (
-            'Unnecessary "%s" after "return"',
+            'Unnecessary "%s" after "return" remove it and de-indent all the code inside it',
```
tusharsadhwani, line 242: I think here should be a comma before remove, in all the cases

Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/5736/commits/a281fadc650072b7737f3b0bb6485011c5f78db4

Pierre-Sassoulas:
Converting to draft because this need a separate message for ``else`` or ``elif``, when it's ``elif`` the fix is to simply remove the ``el`` in ``elif``

Pierre-Sassoulas:
@jacobtylerwalls let me know what you think, this is a follow up to #5614

ChangeLog
```diff
@@ -37,6 +37,10 @@ Release date: TBA
 
   Closes #5281
 
+* Better warning messages for useless else or elif when a function return early.
```
jacobtylerwalls, line 40:
```suggestion
* Better warning messages for useless else or elif when a function returns early.
```

doc/whatsnew/2.13.rst
```diff
@@ -86,6 +86,10 @@ Other Changes
 
   Closes #5648
 
+* Better warning messages for useless else or elif when a function return early.
```
jacobtylerwalls, line 89:
```suggestion
* Better warning messages for useless else or elif when a function returns early.
```

pylint/extensions/check_elif.py
```diff
@@ -31,7 +31,8 @@ class ElseifUsedChecker(BaseTokenChecker):
     name = "else_if_used"
     msgs = {
         "R5501": (
-            'Consider using "elif" instead of "else if"',
+            'Consider using "elif" instead of "else", replace the "else" by an "elif",'
+            ' then de-indent the code and switch from "if" to "elif" inside the else',
```
jacobtylerwalls, line 35: I find this suggestion hard to follow. If I try to follow it on the first example in the unittests, I get something like this.

source
```python
    if myint > 5:
        pass
    else:
        if myint <= 5:  # [else-if-used]
            pass
        else:
            myint = 3
            if myint > 2:
                if myint > 3:
                    pass
```

"replace the "else" by an "elif"...
```python
    if myint > 5:
        pass
    elif myint <= 5:
            pass
        else:
            myint = 3
            if myint > 2:
                if myint > 3:
                    pass
```

" then de-indent the code..."

```python
    if myint > 5:
        pass
    elif myint <= 5
            pass
    else:
        myint = 3
        if myint > 2:
            if myint > 3:
                pass
```

" and switch from "if" to "elif" inside the else"
```python
    if myint > 5:
        pass
    elif myint <= 5
            pass  <-- wait, wasn't this to be unindented?
    else: <-- turn this to elif?
        myint = 3
        if myint > 2:  <-- turn this to elif?
            if myint > 3:
                pass
```

I appreciate what we want to do here to help folks out, but I think it might not do any good in some cases. Sorry for going through it so step by step but just wanted to show if we're trying to help folks maybe less is more.

jacobtylerwalls (commented):
Great, this is very good for the original `no-else-return` and friends; I just have a slight concern about being misleadingly prescriptive with the `else-if-used` refactoring instructions.

pylint/extensions/check_elif.py
Pierre-Sassoulas: Ha, I made a slight change before seeing this review, let me try to do it step by step again.

Pierre-Sassoulas: > remove the ""else"", then de-indent the code and switch from ""if"" to ""elif"" inside the else"

```python
   if myint > 5:
        pass
    else:
        if myint <= 5:  # [else-if-used]
            pass
        else:
            myint = 3
            if myint > 2:
                if myint > 3:
                    pass
```     
 remove the ""else"",

```python
   if myint > 5:
        pass
        if myint <= 5:  # [else-if-used]
            pass
        else:
            myint = 3
            if myint > 2:
                if myint > 3:
                    pass
``` 
 then de-indent the code 
```python
   if myint > 5:
        pass
    if myint <= 5:  # [else-if-used]
        pass
    else:
        myint = 3
        if myint > 2:
            if myint > 3:
                pass
```      
switch from ""if"" to ""elif"" inside the else"
```python
   if myint > 5:
        pass
  elif myint <= 5:
        pass
  else:
        myint = 3
        if myint > 2:
            if myint > 3:
                pass
```         
Seems to work, but then again I'm the one that wrote the message so of course I understand it correctly 😄

jacobtylerwalls: Yeah, this is better, but it's still hard because "switch from "if" to "elif" inside the else" is referring to an else that's already gone.

What about:
`consider using elif instead of else to remove one indentation level`

Pierre-Sassoulas: Yeap, that's nicer let's use this.

jacobtylerwalls: Maybe:
Consider consolidating if after else to elif

Pierre-Sassoulas: I don't understand "consolidating" so I went with ``'Consider using "elif" instead of "else" then "if" to remove one indentation level'``, which is nice and small imo.

jacobtylerwalls (approved):

tusharsadhwani (approved):

PR #5738, opened 2022-01-28
Fix assignment-from-none false negative case using list.sort()

### LINKED ISSUE #5722: assign list.sort() instead of sorted(list)
### Current problem

when you use the `sort()` method on a list object, it modifies the list and returns None.
A common mistake is to assign the return value of this assignment (i.e. None) to a variable thinking you assigned the sorted list instead.

```
lst=[3,2,1]
A=lst.sort()
```

Results with: 
None in `A` and [1,2,3] in `lst`.

### Desired solution

The checker should raise the WARNING message:
“A sort method of `list` object is assigned to a variable and equals None, consider using the `sorted()` function instead”.

### Additional context

When i run pylint with all extensions on the example above I don't get any message.
But PyCharm's linter has a message for it so I am not sure if it was already implemented.

If not I could write it.

orSolocate:
## Type of Changes

## Description
Fix `assignment-from-none` false negative case using `list.sort()`

Closes #5722

### DIFF
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1141,10 +1141,8 @@ accessed. Python regular expressions are accepted.",
         self._check_assignment_from_function_call(node)
         self._check_dundername_is_string(node)
 
-    def _check_assignment_from_function_call(self, node):
-        """check that if assigning to a function call, the function is
-        possibly returning something valuable
-        """
+    def _check_assignment_from_function_call(self, node) -> None:
+        """When assigning to a function call, check that the function returns a valuable value"""
         if not isinstance(node.value, nodes.Call):
             return
 
@@ -1153,7 +1151,7 @@ accessed. Python regular expressions are accepted.",
         if not isinstance(function_node, funcs):
             return
 
-        # Unwrap to get the actual function object
+        # Unwrap to get the actual function node object
         if isinstance(function_node, astroid.BoundMethod) and isinstance(
             function_node._proxied, astroid.UnboundMethod
         ):
@@ -1162,33 +1160,31 @@ accessed. Python regular expressions are accepted.",
         # Make sure that it's a valid function that we can analyze.
         # Ordered from less expensive to more expensive checks.
         if (
-            function_node.decorators is not None
+            not function_node.is_function
+            or function_node.decorators is not None
             or self._is_ignored_function(function_node)
-            or isinstance(function_node, nodes.AsyncFunctionDef)
-            or utils.is_error(function_node)
         ):
             return
 
-        if self._list_sort_method(
-            node.value
-        ):  # false-negative case fix, see issue #5722
+        # Fix a false-negative for list.sort(), see issue #5722
+        if self._list_sort_method(node.value):
             self.add_message("assignment-from-none", node=node)
             return
 
         if not function_node.root().fully_defined():
             return
 
-        returns = list(
+        return_nodes = list(
             function_node.nodes_of_class(nodes.Return, skip_klass=nodes.FunctionDef)
         )
-        if not returns:
+        if not return_nodes:
             self.add_message("assignment-from-no-return", node=node)
         else:
-            for rnode in returns:
+            for ret_node in return_nodes:
                 if not (
-                    isinstance(rnode.value, nodes.Const)
-                    and rnode.value.value is None
-                    or rnode.value is None
+                    isinstance(ret_node.value, nodes.Const)
+                    and ret_node.value.value is None
+                    or ret_node.value is None
                 ):
                     break
             else:
@@ -1199,22 +1195,23 @@ accessed. Python regular expressions are accepted.",
         function_node: Union[
             nodes.FunctionDef, astroid.UnboundMethod, astroid.BoundMethod
         ]
-    ):
+    ) -> bool:
         return (
-            not function_node.is_function
-            or function_node.is_generator()
+            function_node.is_generator()
             or function_node.is_abstract(pass_is_abstract=False)
+            or isinstance(function_node, nodes.AsyncFunctionDef)
+            or utils.is_error(function_node)
         )
 
     @staticmethod
-    def _list_sort_method(node: nodes.Call):
+    def _list_sort_method(node: nodes.Call) -> bool:
         return (
             isinstance(node.func, nodes.Attribute)
             and node.func.attrname == "sort"
             and isinstance(utils.safe_infer(node.func.expr), nodes.List)
         )
 
-    def _check_dundername_is_string(self, node):
+    def _check_dundername_is_string(self, node) -> None:
         """Check a string is assigned to self.__name__"""
 
         # Check the left-hand side of the assignment is <something>.__name__
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1142,7 +1142,7 @@ accessed. Python regular expressions are accepted.",
         self._check_dundername_is_string(node)
 
     def _check_assignment_from_function_call(self, node: nodes.Assign) -> None:
-        """When assigning to a function call, check that the function returns a valuable value."""
+        """When assigning to a function call, check that the function returns a valid value."""
         if not isinstance(node.value, nodes.Call):
             return
 
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1161,16 +1161,21 @@ accessed. Python regular expressions are accepted.",
 
         # Make sure that it's a valid function that we can analyze.
         # Ordered from less expensive to more expensive checks.
-        # pylint: disable=too-many-boolean-expressions
         if (
-            not function_node.is_function
+            function_node.decorators is not None
+            or self._is_ignored_function(function_node)
             or isinstance(function_node, nodes.AsyncFunctionDef)
-            or function_node.decorators
-            or function_node.is_generator()
-            or function_node.is_abstract(pass_is_abstract=False)
             or utils.is_error(function_node)
-            or not function_node.root().fully_defined()
         ):
+            return
+
+        if self._list_sort_method(
+            node.value
+        ):  # false-negative case fix, see issue #5722
+            self.add_message("assignment-from-none", node=node)
+            return
+
+        if not function_node.root().fully_defined():
             return
 
         returns = list(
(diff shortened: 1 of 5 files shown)


pylint/checkers/typecheck.py
```diff
@@ -1189,6 +1194,26 @@ def _check_assignment_from_function_call(self, node):
             else:
                 self.add_message("assignment-from-none", node=node)
 
+    @staticmethod
+    def _is_ignored_function(
+        function_node: Union[
+            nodes.FunctionDef, astroid.UnboundMethod, astroid.BoundMethod
+        ]
+    ):
```
Pierre-Sassoulas, line 1202:
```suggestion
    ) -> bool:
    """Ordered from the least to most expensive check."""
```

pylint/checkers/typecheck.py
```diff
@@ -1189,6 +1194,26 @@ def _check_assignment_from_function_call(self, node):
...
+        return (
+            not function_node.is_function
+            or function_node.is_generator()
+            or function_node.is_abstract(pass_is_abstract=False)
+        )
+
+    @staticmethod
+    def _list_sort_method(node: nodes.Call):
```
Pierre-Sassoulas, line 1210:
```suggestion
    def _list_sort_method(node: nodes.Call) -> bool:
```

pylint/checkers/typecheck.py
```diff
@@ -1161,18 +1161,23 @@ def _check_assignment_from_function_call(self, node):
...
             or utils.is_error(function_node)
-            or not function_node.root().fully_defined()
         ):
             return
 
+        if self._list_sort_method(
+            node.value
+        ):  # false-negative case fix, see issue #5722
+            self.add_message("assignment-from-none", node=node)
+            return
```
Pierre-Sassoulas, line 1176:
```suggestion
        # Fix a false-negative for list.sort(), see issue #5722
        if self._list_sort_method(node.value):  
            self.add_message("assignment-from-none", node=node)
            return
```

Pierre-Sassoulas (changes requested):
Thank you ! LGTM, I only have some typing and formatting nitpicks

pylint/checkers/typecheck.py
```diff
@@ -1161,18 +1161,23 @@ def _check_assignment_from_function_call(self, node):
...
+
+        if not function_node.root().fully_defined():
+            return
+
         returns = list(
             function_node.nodes_of_class(nodes.Return, skip_klass=nodes.FunctionDef)
         )
```
Pierre-Sassoulas, line 1183: I'm wondering why we could not make the generic part also work for list.sort() though, what went wrong ?

orSolocate: I know why.. I debugged it.
It fails because `function_node.root().fully_defined()` returns `False`.
It returns `False` because `astroid` searches for a return value in the **abstract** definition of the `sort` function, in `builtins.py`:

```
    def sort(self, *args, **kwargs): # real signature unknown
        """
        Sort the list in ascending order and return None.
        
        The sort is in-place (i.e. the list itself is modified) and stable (i.e. the
        order of two equal elements is maintained).
        
        If a key function is given, apply it once to each list item and sort them,
        ascending or descending, according to their function values.
        
        The reverse flag can be set to sort in descending order.
        """
        pass
```

I don't think it's possible to retrieve the `astroid.FunctionDef` of the correct function..
If it is possible then we should fix the redirection itself.. 
It might raise some false-positives though,..

Anyone knows if it's possible to do so in `astroid`?

orSolocate:
@Pierre-Sassoulas  Thanks for the quick review!  Will clear the pipeline errors and apply your suggestions.

pylint/checkers/typecheck.py
orSolocate: seems from [this SO answer ](https://stackoverflow.com/questions/52081316/how-to-view-the-implementation-of-pythons-built-in-functions-in-pycharm) that there is there is no actual python implementation, that the C implementation already built-in inside the interpreter...

DanielNoord: Thanks @orSolocate for this thorough search!

pylint/checkers/typecheck.py
```diff
@@ -1141,10 +1141,8 @@ def visit_assign(self, node: nodes.Assign) -> None:
         self._check_assignment_from_function_call(node)
         self._check_dundername_is_string(node)
 
-    def _check_assignment_from_function_call(self, node):
-        """check that if assigning to a function call, the function is
-        possibly returning something valuable
-        """
+    def _check_assignment_from_function_call(self, node) -> None:
+        """When assigning to a function call, check that the function returns a valuable value"""
```
DanielNoord, line 1145:
```suggestion
        """When assigning to a function call, check that the function returns a valuable value."""
```

pylint/checkers/typecheck.py
```diff
@@ -1153,43 +1151,67 @@ def _check_assignment_from_function_call(self, node):
...
                 ):
                     break
             else:
                 self.add_message("assignment-from-none", node=node)
 
-    def _check_dundername_is_string(self, node):
+    @staticmethod
+    def _is_ignored_function(
+        function_node: Union[
+            nodes.FunctionDef, astroid.UnboundMethod, astroid.BoundMethod
```
DanielNoord, line 1196: `BoundMethod` is a sub class of `UnboundMethod` so it can be removed. Could you import these from `bases` instead of `astroid` directly?

pylint/checkers/typecheck.py
DanielNoord: Could you add typing to this function? It's probably `node: nodes.Call`?

pylint/checkers/typecheck.py
```diff
@@ -1153,43 +1151,67 @@ def _check_assignment_from_function_call(self, node):
...
-            or not function_node.root().fully_defined()
+            or function_node.decorators is not None
+            or self._is_ignored_function(function_node)
         ):
             return
 
-        returns = list(
+        # Fix a false-negative for list.sort(), see issue #5722
+        if self._list_sort_method(node.value):
+            self.add_message("assignment-from-none", node=node)
```
DanielNoord, line 1171:
```suggestion
            self.add_message("assignment-from-none", node=node, confidence=INFERENCE)
```

DanielNoord: You'll need to import this from `interfaces`.

pylint/checkers/typecheck.py
```diff
@@ -1153,43 +1151,67 @@ def _check_assignment_from_function_call(self, node):
...
-            or utils.is_error(function_node)
-            or not function_node.root().fully_defined()
+            or function_node.decorators is not None
+            or self._is_ignored_function(function_node)
         ):
             return
 
-        returns = list(
+        # Fix a false-negative for list.sort(), see issue #5722
+        if self._list_sort_method(node.value):
```
DanielNoord, line 1170:
```suggestion
        if self._is_list_sort_method(node.value):
```

pylint/checkers/typecheck.py
```diff
@@ -1153,43 +1151,67 @@ def _check_assignment_from_function_call(self, node):
...
+    ) -> bool:
+        return (
+            function_node.is_generator()
+            or function_node.is_abstract(pass_is_abstract=False)
+            or isinstance(function_node, nodes.AsyncFunctionDef)
+            or utils.is_error(function_node)
+        )
+
+    @staticmethod
+    def _list_sort_method(node: nodes.Call) -> bool:
```
DanielNoord, line 1207:
```suggestion
    def _is_list_sort_method(node: nodes.Call) -> bool:
```

pylint/checkers/typecheck.py
```diff
@@ -1153,43 +1151,67 @@ def _check_assignment_from_function_call(self, node):
...
+    def _is_ignored_function(
+        function_node: Union[
+            nodes.FunctionDef, astroid.UnboundMethod, astroid.BoundMethod
+        ]
+    ) -> bool:
+        return (
+            function_node.is_generator()
+            or function_node.is_abstract(pass_is_abstract=False)
+            or isinstance(function_node, nodes.AsyncFunctionDef)
+            or utils.is_error(function_node)
```
DanielNoord, line 1203:
```suggestion
            isinstance(function_node, nodes.AsyncFunctionDef)
            or utils.is_error(function_node)
            or function_node.is_generator()
            or function_node.is_abstract(pass_is_abstract=False)
```

I believe this is the order of the most expensive function calls (least>most)

pylint/checkers/typecheck.py
```diff
@@ -1153,43 +1151,67 @@ def _check_assignment_from_function_call(self, node):
...
-        # pylint: disable=too-many-boolean-expressions
         if (
             not function_node.is_function
-            or isinstance(function_node, nodes.AsyncFunctionDef)
-            or function_node.decorators
-            or function_node.is_generator()
-            or function_node.is_abstract(pass_is_abstract=False)
-            or utils.is_error(function_node)
-            or not function_node.root().fully_defined()
+            or function_node.decorators is not None
```
DanielNoord, line 1164: In the original code we do `function_node.decorators`. Is there a difference? I believe decorators is `Union[None, nodes.Decorators]` so we could also just use `or function_node.decorators`

DanielNoord (changes requested):
Thanks @orSolocate! Another great addition. Just some comments about style and order. Fix itself is quite elegant 😄

pylint/checkers/typecheck.py
orSolocate: I just maintained the previous implementation and moved it to a function but sure I will fix that in the code when assigning to `function_node`

pylint/checkers/typecheck.py
orSolocate: @DanielNoord  okay. I just tried to maintain the order it was because in the comment they left in the code they say it was already ordered this way...

DanielNoord: Ah okay. I looked at all the implementation, I think this should be slightly better.

pylint/checkers/typecheck.py
DanielNoord: Thanks! 👍

pylint/checkers/typecheck.py
```diff
@@ -1141,10 +1141,8 @@ def visit_assign(self, node: nodes.Assign) -> None:
         self._check_assignment_from_function_call(node)
         self._check_dundername_is_string(node)
 
-    def _check_assignment_from_function_call(self, node):
-        """check that if assigning to a function call, the function is
-        possibly returning something valuable
-        """
+    def _check_assignment_from_function_call(self, node: nodes.Assign) -> None:
+        """When assigning to a function call, check that the function returns a valuable value."""
```
Pierre-Sassoulas, line 1145:
```suggestion
        """When assigning to a function call, check that the function returns a valid value."""
```

pylint/checkers/typecheck.py
```diff
@@ -1153,43 +1151,65 @@ def _check_assignment_from_function_call(self, node):
...
+
+        return_nodes = list(
             function_node.nodes_of_class(nodes.Return, skip_klass=nodes.FunctionDef)
         )
-        if not returns:
+        if not return_nodes:
             self.add_message("assignment-from-no-return", node=node)
         else:
-            for rnode in returns:
+            for ret_node in return_nodes:
```
Pierre-Sassoulas, line 1183: We could even rename it to ``return_node``,  I don't suggest because an IDE refactor should do it instantly.

Pierre-Sassoulas (approved):
LGTM, thank you for this, it will definitely help newbies and distracted veterans alike.

DanielNoord (approved):

pylint/checkers/typecheck.py
orSolocate: You added it yourself already while merging?

Pierre-Sassoulas: It wasn't that important, so I merged :)

PR #5745, opened 2022-01-29
Remove VariableVisitConsumerAction.CONSUME

jacobtylerwalls:
## Type of Changes

## Description

See discussion in https://github.com/PyCQA/pylint/pull/5727#discussion_r795078071. Wait for #5727, then rebase and open for review.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1517,10 +1517,8 @@ class VariablesChecker(BaseChecker):
                 node=node,
                 confidence=confidence,
             )
-            # If there are nodes added to consumed_uncertain by
-            # get_next_to_consume() because they might not have executed,
-            # return a CONSUME action so that _undefined_and_used_before_checker()
-            # will mark them as used
+            # Mark for consumption any nodes added to consumed_uncertain by
+            # get_next_to_consume() because they might not have executed.
             return (
                 VariableVisitConsumerAction.RETURN,
                 current_consumer.consumed_uncertain[node.name],
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -95,9 +95,9 @@ if TYPE_CHECKING:
     from pylint.lint import PyLinter
 
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    pass
 else:
-    from typing_extensions import Literal
+    pass
 
 SPECIAL_OBJ = re.compile("^_{2}[a-z]+_{2}$")
 FUTURE = "__future__"
@@ -182,8 +182,7 @@ class VariableVisitConsumerAction(Enum):
 
 
 VariableVisitConsumerActionAndOptionalNodesType = Tuple[
-    VariableVisitConsumerAction,
-    Optional[List[nodes.NodeNG]]
+    VariableVisitConsumerAction, Optional[List[nodes.NodeNG]]
 ]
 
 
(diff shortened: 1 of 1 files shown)


pylint/checkers/variables.py
```diff
@@ -1522,21 +1514,22 @@ def _check_consumer(
...
-                # If there are nodes added to consumed_uncertain by
-                # get_next_to_consume() because they might not have executed,
-                # return a CONSUME action so that _undefined_and_used_before_checker()
-                # will mark them as used
-                return (VariableVisitConsumerAction.CONSUME, found_nodes)
-            return (VariableVisitConsumerAction.RETURN, None)
+            # If there are nodes added to consumed_uncertain by
+            # get_next_to_consume() because they might not have executed,
+            # return a CONSUME action so that _undefined_and_used_before_checker()
+            # will mark them as used
```
jacobtylerwalls, line 1520: Note to self: when rebasing, fix this comment

pylint/checkers/variables.py
```diff
@@ -1483,7 +1465,6 @@ def _check_consumer(
         consumer_level: int,
         base_scope_type: Any,
     ) -> VariableVisitConsumerActionAndOptionalNodesType:
```
DanielNoord, line 1467: Why did you remove the docstring here?

pylint/checkers/variables.py
```diff
@@ -175,23 +169,14 @@ class VariableVisitConsumerAction(Enum):
...
-        Union[
-            Literal[VariableVisitConsumerAction.CONTINUE],
-            Literal[VariableVisitConsumerAction.RETURN],
-        ],
-        None,
-    ],
-    Tuple[Literal[VariableVisitConsumerAction.CONSUME], List[nodes.NodeNG]],
+
+
+VariableVisitConsumerActionAndOptionalNodesType = Tuple[
```
DanielNoord, line 178: I wonder if this is necessary. With the simplification wouldn't it be better to use this directly instead of an alias?

pylint/checkers/variables.py
jacobtylerwalls: Oops, that was an oversight. Thanks for catching.

pylint/checkers/variables.py
jacobtylerwalls: Yes, I was starting to think that myself.

pylint/checkers/variables.py
```diff
@@ -1526,21 +1503,20 @@ def _check_consumer(
...
-                # get_next_to_consume() because they might not have executed,
-                # return a CONSUME action so that _undefined_and_used_before_checker()
-                # will mark them as used
-                return (VariableVisitConsumerAction.CONSUME, found_nodes)
-            return (VariableVisitConsumerAction.RETURN, None)
+            # Mark for consumption any nodes added to consumed_uncertain by
+            # get_next_to_consume() because they might not have executed.
+            return (
+                VariableVisitConsumerAction.RETURN,
+                current_consumer.consumed_uncertain[node.name],
```
DanielNoord, line 1510: Is this the same as `current_consumer.get_next_to_consume(node)`?

jacobtylerwalls: I don't think so: that function has already run and actually was responsible for _populating_ the dict entry being accessed here. Now having received a RETURN action, it's time to finish consuming all the names, including what maybe-executed.

I'm realizing `consumed_uncertain` is less descriptive as a name now that we're getting rid of CONSUME as an action. It now refers more to "execution uncertain" or something.

DanielNoord (approved):

Pierre-Sassoulas (approved):
Nice refactoring !

PR #5752, opened 2022-01-31
Add timeouts for CI jobs

cdce8p:
## Description

I recently had a job run for 6h just to fail. Usually they all complete in <10min (CI) / <30min (Primer). So anything that is taking longer than that is probably going to fail anyway.

This PR adds `timeout-minutes` to the individual jobs to abort them after a reasonable amount of time. Usually 2x what they actually need. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

/CC: @DanielNoord Is 60min enough for Primer jobs? From what I've seen an individual one usually takes 20min.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,7 @@ jobs:
   prepare-base:
     name: Prepare base dependencies
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     outputs:
       python-key: ${{ steps.generate-python-key.outputs.key }}
       pre-commit-key: ${{ steps.generate-pre-commit-key.outputs.key }}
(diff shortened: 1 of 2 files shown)


DanielNoord:
60 minutes is more than enough! I would even consider setting it to 45 or 50, that's more than twice they usually take.

cdce8p:
> 60 minutes is more than enough! I would even consider setting it to 45 or 50, that's more than twice they usually take.

👍🏻 We can leave some headroom. It's still much better than the default 6h timeout. Usually those are issues with Github Actions anyway.

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Great catch @cdce8p  ! Should we change to 45mn before merging ?

cdce8p:
> Should we change to 45mn before merging ?

Don't think that's necessary. The primer tests are somewhat separate from the rest already. 1h also isn't that long. Still an improvement over 6h.

PR #5753, opened 2022-01-31
Add test for functional tests directories structure

DanielNoord:

## Type of Changes

## Description

Based on the discussion in #5474.

I'm not sure this is the best way to test this, but it ensures new contributors will run it as they are likely to run the `pytest` command at least once.

The test checks that a functional test should not be move to a sub-directory (`use_test` should go in `u/use`) and that it is not in the incorrect directory (`using_test` should not go in `u/use`).
I made two directories as the tests in those directories really do not follow this pattern and I don't think it makes too much sense to change all the tests names there.


Pierre-Sassoulas:
Actually I had some code ready but I delayed pushing it as there was a misnamed symlink that was not very straightforward to rename. I modified the pylint/testutils/functional code. Let me open a MR with the current code so as to discuss this.

DanielNoord:
For some reason this test passes locally... So I need to push to see what's wrong 😓

DanielNoord:
So, this finally passes..

Pierre-Sassoulas (changes requested):
I think we could either:
- add your check in the check function in #5757

or

- move the [assert from testutil](https://github.com/PyCQA/pylint/pull/5757/files#diff-32d4d27d17ed6921c4540efe5c6829c55d9b5067634611a6c77abcc407c53af1R38) into this test function instead

But right now 5757 lacks a feature and there is a dissemination of structure check in this one. I think I prefer option 1 because it would permits to provide the structure to users of ``pylint/testutil``.

Pierre-Sassoulas:
I rebased Daniel works into the two first commits then moved the test to ``pylint/testutils/`` and took what was in the other MR, then renamed the symlink. Locally it was working, but I'm going to work on that / fix it tomorrow.

Pierre-Sassoulas:
I'm still working on it :)

DanielNoord:
@Pierre-Sassoulas Still working on this? Or want me to try and get this ready?

DanielNoord:
> I'm still working on it but could push the half-baked work in progress I have locally. There's an issue with moving some of the functional tests that have 48+ files in their directory. They are not entirely independents as they import each others so moving them break the tests. The easy fix is to change the imports, the real fix would be to make them independent. I often don't have a long enough uninterrupted period to do the real fix.

If you push the in progress work I'll take a look!

Pierre-Sassoulas:
I just did :) The 3 last commits are temporary garbage.

DanielNoord:
I think this is now good to go!

**NOTE: PLEASE DON'T SQUASH**.

We can use https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame to ignore the moving commit from git blame so that we don't lose easy access to relevant commits in git blame viewers (such as the one in VS Code).

Pierre-Sassoulas (approved):
Sorry for the delay in reviewing. LGTM but a conflict appeared and we can't squash so we should rebase on latest main. Let me do it then.

PR #5757, opened 2022-01-31
Add assert for functional tests directories structure

Pierre-Sassoulas:
## Type of Changes

## Description

Similar to #5753 but another implementation


DanielNoord:
I don't have a preference opinion wise. The only thing my implementation adds is that also checks whether any tests should be moved to a sub-directory. So `u/used_test.py` should be in `u/used/...`.
The PR itself became quite large sadly as there are a lot of directories that turned out to be wrong... Not sure in which PR we should continue the discussion, but let me know what you think!

Pierre-Sassoulas:
Let's keep the other one, better general state and more work done already.

Issue #2835, opened 2019-03-25
Catch potential UnboundLocalError in try/except 


lbenezriravin:
Potentially related to #2615. Not sure if these would both be fixed by the same patch, so I'm making a new issue.

### Is your feature request related to a problem? Please describe
The following code passes pylint:
```
try:
   foo = 1 / 0
except:
   pass
print(foo)
```
But raises an `UnboundLocalError` when run. 

### Describe the solution you'd like
Pylint already has an `undefined-loop-variable` warning for a similar case that could result in the same error:
```
for foo in []:
    pass
print(foo)
```
I feel that the former code should raise a warning if that latter code does. More generally, I don't expect `UnboundLocalError` from any code that passes pylint, but I'm not sure if that guarantee is possible.

### Additional context
Thanks for your work on this tool!

PCManticore:
Thanks @lbenezriravin This makes sense.

Pierre-Sassoulas:
Closing as duplicate of #5524

DanielNoord:
This was actually not a duplicate see: https://github.com/PyCQA/pylint/pull/5764#issuecomment-1032803372.

A very relevant comment was made by @cdce8p in #5524 [here](https://github.com/PyCQA/pylint/issues/5524#issuecomment-1025262104):

> Another idea. I like how pyright names it. Maybe `possibly-unbound`?
> 
> ```
> "a" is possibly unbound
> ```
> 
>  > Maybe `consider-using-try-else`
> 
> We shouldn't really "force" users to add an `else` case. There are just to many alternatives. You could for example assign a default value before you enter the `try` block.

jacobtylerwalls:
Implementation tip: when checking whether the except handlers define a name, raise, or return, consider using `NamesConsumer._defines_name_raises_or_returns()` from #5764

Pierre-Sassoulas:
pyright has ``possibly unbound``, mypy has [``possibly-undefined``](https://mypy.readthedocs.io/en/stable/error_code_list2.html#warn-about-variables-that-are-defined-only-in-some-execution-paths-possibly-undefined), I think we should introduce either one of those two names in pylint on top of ``undefined-variable`` (maybe ``possibly-undefined-variable`` for consistency)

jacobtylerwalls:
When I first implemented try/except fathoming in used-before-assignment, the decision we made was to just check the except blocks and let users treat try blocks as succeeding. Even this was a huge change for a minor release, so we stopped there.

[...]

Issue #5645, opened 2022-01-07
[2.12.x] false positives: no-name-in-module, import-error


paride:
### Bug description

Pylint 2.12.x emits `no-name-in-module` and `import-error` false positives with Python 3.6. This doesn't happen with pylint 2.11.x, or with newer Python versions.

### Command used

```shell
The issue is trivial to reproduce with a 1-line Python file:

$ cat test.py
import distutils.util

$ python3 test.py
$

$ python3 -m pylint test.py
```

### Pylint output

```shell
$ python3 -m pylint test.py
************* Module test
test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
test.py:1:0: E0611: No name 'util' in module 'distutils' (no-name-in-module)
test.py:1:0: E0401: Unable to import 'distutils.util' (import-error)
test.py:1:0: W0611: Unused import distutils.util (unused-import)
```

### Expected behavior

No `E` tags. 

### Pylint version

```shell
$ python3 -m pylint --version
pylint 2.12.2
astroid 2.9.2
Python 3.6.9 (default, Dec  8 2021, 21:08:43) 
[GCC 8.4.0]
```

### OS / Environment

Ubuntu 18.04.

Pierre-Sassoulas:
Thank you for opening the issue. Sorry but we're not going to fix this unless someone open a MR fixing it before pylint 2.13.0 is released as pylint 2.13.0 is the latest version supporting python 3.6 and as python 3.6 is past end of life now.

DanielNoord:
@Pierre-Sassoulas Didn't we add a test specifically for this https://github.com/PyCQA/pylint/commit/11eba03a9de2cf2293f0d952f0c8988b9bfd870d as an regression test against changes in `astroid` `2.9.2`?

@paride Just to be sure, could you try this while pinning `astroid` to `2.9.0`?

Pierre-Sassoulas:
I thougth so but the version reported is:
```
pylint 2.12.2
astroid 2.9.2
```
As we did not change anything in pylint and the astroid used is the latest I don't think the bug is gone away.

paride:
@DanielNoord no false positive when pinning `astroid==2.9.0`.

DanielNoord:
Note that https://github.com/DanielNoord/pylint/pull/74 does not report a false positive here...

I'm not sure what is going wrong here. It probably has to do with recent changes in `astroid` but apparently it does not fall on all setups. @paride Are you using a `venv` or anything?

[...]

PR #5764, opened 2022-02-02
Fix false negative for `used-before-assignment` when some except handlers don't define a name

### LINKED ISSUE #5524: False-negative `used-before-assignment` after try-except block
Followup to https://github.com/PyCQA/pylint/issues/5500#issuecomment-991695275

```py
def func(var):
    try:
        return 1 / var.some_other_func()
    except AttributeError:
        pass
    except ZeroDivisionError:
        msg = "Devision by 0"

    print(msg)  # should emit 'used-before-assignment'
```

The code above should raise a `used-before-assignment` error since `msg` is not assigned in the `AttributeError` ExceptHandler.

jacobtylerwalls:
## Type of Changes

## Description

Fixes #5524.

<s>In discussion I suggested closing #2835 as a duplicate, but if folks like we could "unlink" the two and reopen the older ticket.

The older ticket is just suggesting "always assume try statements fail", and @cdce8p suggested a new message for that (`possibly-unbound`). Probably a 2.14 or later thing. By contrast this PR is just fixing a quick false negative. Sorry for not seeing that we were more likely to tackle these separately!</s> Edit: done!

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -703,17 +703,18 @@ scope_type : {self._atomic.scope_type}
         for other_node in found_nodes:
             other_node_statement = other_node.statement(future=True)
             # Only testing for statements in the except block of TryExcept
-            if not (
-                isinstance(other_node_statement.parent, nodes.ExceptHandler)
-                and isinstance(other_node_statement.parent.parent, nodes.TryExcept)
-            ):
+            closest_except_handler = utils.get_node_first_ancestor_of_type(
+                other_node_statement, nodes.ExceptHandler
+            )
+            if not closest_except_handler:
                 continue
+            closest_try_except: nodes.TryExcept = closest_except_handler.parent
             # If the other node is in the same scope as this node, assume it executes
-            if other_node_statement.parent.parent_of(node):
+            if closest_except_handler.parent_of(node):
                 continue
             try_block_returns = any(
                 isinstance(try_statement, nodes.Return)
-                for try_statement in other_node_statement.parent.parent.body
+                for try_statement in closest_try_except.body
             )
             # If the try block returns, assume the except blocks execute.
             if try_block_returns:
@@ -722,24 +723,18 @@ scope_type : {self._atomic.scope_type}
                 if (
                     isinstance(node_statement.parent, nodes.TryFinally)
                     and node_statement in node_statement.parent.finalbody
-                    # We have already tested that other_node_statement has two parents
-                    # and it was TryExcept, so getting one more parent is safe.
-                    and other_node_statement.parent.parent.parent.parent_of(
-                        node_statement
-                    )
+                    and closest_try_except.parent.parent_of(node_statement)
                 ):
                     uncertain_nodes.append(other_node)
                 # Assume the except blocks execute, so long as each handler
                 # defines the name, raises, or returns.
                 elif all(
                     NamesConsumer._defines_name_raises_or_returns(node.name, handler)
-                    for handler in other_node_statement.parent.parent.handlers
+                    for handler in closest_try_except.handlers
                 ):
                     continue
 
-            if NamesConsumer._check_loop_finishes_via_except(
-                node, other_node_statement.parent.parent
-            ):
+            if NamesConsumer._check_loop_finishes_via_except(node, closest_try_except):
                 continue
 
             # Passed all tests for uncertain execution
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -751,7 +751,14 @@ scope_type : {self._atomic.scope_type}
         for stmt in handler.get_children():
             if isinstance(stmt, (nodes.Raise, nodes.Return)):
                 return True
-            if isinstance(stmt, nodes.Assign) and stmt.targets[0].as_string() == name:
+            if isinstance(stmt, nodes.Assign):
+                 for target in stmt.targets:
+                     if isinstance(target, nodes.AssignName) and target.name == name:
+                         return True
+                     if isinstance(target, (nodes.Tuple, nodes.List)):
+                         for elt in utils.get_all_elements(target):
+                             if isinstance(elt, nodes.Const) and elt.value == name:
+                                 return True
                 return True
             if isinstance(stmt, nodes.If):
                 if (
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -749,7 +749,7 @@ scope_type : {self._atomic.scope_type}
         raises, or returns.
         """
 
-        def _define_raise_or_return(stmt) -> bool:
+        def _define_raise_or_return(stmt: nodes.NodeNG) -> bool:
             if isinstance(stmt, (nodes.Raise, nodes.Return)):
                 return True
             if isinstance(stmt, nodes.Assign):
@@ -758,6 +758,7 @@ scope_type : {self._atomic.scope_type}
                         if isinstance(elt, nodes.AssignName) and elt.name == name:
                             return True
             if isinstance(stmt, nodes.If):
+                # Check for assignments inside the test
                 if (
                     isinstance(stmt.test, nodes.NamedExpr)
                     and stmt.test.target.name == name
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -756,10 +756,11 @@ scope_type : {self._atomic.scope_type}
                     if isinstance(target, nodes.AssignName) and target.name == name:
                         return True
                     if isinstance(target, (nodes.Tuple, nodes.List)):
-                        for elt in utils.get_all_elements(target):
-                            if isinstance(elt, nodes.Const) and elt.value == name:
-                                return True
-                return True
+                        if any(
+                            isinstance(elt, nodes.Const) and elt.value == name
+                            for elt in utils.get_all_elements(target)
+                        ):
+                            return True
             if isinstance(stmt, nodes.If):
                 if (
                     isinstance(stmt.test, nodes.NamedExpr)
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -767,7 +767,9 @@ scope_type : {self._atomic.scope_type}
                 ):
                     return True
                 if isinstance(stmt.test, nodes.Call):
-                    for arg_or_kwarg in stmt.test.args + stmt.test.keywords:
+                    for arg_or_kwarg in stmt.test.args + [
+                        kw.value for kw in stmt.test.keywords
+                    ]:
                         if (
                             isinstance(arg_or_kwarg, nodes.NamedExpr)
                             and arg_or_kwarg.target.name == name
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -752,13 +752,13 @@ scope_type : {self._atomic.scope_type}
             if isinstance(stmt, (nodes.Raise, nodes.Return)):
                 return True
             if isinstance(stmt, nodes.Assign):
-                 for target in stmt.targets:
-                     if isinstance(target, nodes.AssignName) and target.name == name:
+                for target in stmt.targets:
+                    if isinstance(target, nodes.AssignName) and target.name == name:
                          return True
-                     if isinstance(target, (nodes.Tuple, nodes.List)):
-                         for elt in utils.get_all_elements(target):
-                             if isinstance(elt, nodes.Const) and elt.value == name:
-                                 return True
+                    if isinstance(target, (nodes.Tuple, nodes.List)):
+                        for elt in utils.get_all_elements(target):
+                            if isinstance(elt, nodes.Const) and elt.value == name:
+                                return True
                 return True
             if isinstance(stmt, nodes.If):
                 if (
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -753,13 +753,8 @@ scope_type : {self._atomic.scope_type}
                 return True
             if isinstance(stmt, nodes.Assign):
                 for target in stmt.targets:
-                    if isinstance(target, nodes.AssignName) and target.name == name:
-                        return True
-                    if isinstance(target, (nodes.Tuple, nodes.List)):
-                        if any(
-                            isinstance(elt, nodes.Const) and elt.value == name
-                            for elt in utils.get_all_elements(target)
-                        ):
+                    for elt in utils.get_all_elements(target):
+                        if isinstance(elt, nodes.AssignName) and elt.name == name:
                             return True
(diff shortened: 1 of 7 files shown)


pylint/checkers/variables.py
```diff
@@ -729,10 +729,12 @@ def _uncertain_nodes_in_except_blocks(
...
                     uncertain_nodes.append(other_node)
-                else:
-                    # Assume the except blocks execute. Possibility for a false negative
-                    # if one of the except blocks does not define the name in question,
-                    # raise, or return. See: https://github.com/PyCQA/pylint/issues/5524.
+                # Assume the except blocks execute, so long as each handler
+                # defines the name, raises, or returns.
+                elif all(
+                    NamesConsumer._defines_name_raises_or_returns(node.name, handler)
+                    for handler in other_node_statement.parent.parent.handlers
```
areveny, line 736: Hey, I see you're still adding commits but some of your test failures might be from this `.parent.parent.handlers` pattern, which might not be resilient to cases where there are parent statements between the name and the `try`/`except`.

It looks like some failures are coming from [ntpath](https://github.com/python/cpython/blob/main/Lib/ntpath.py) line 467 where an entire function is defined under an `except`.

jacobtylerwalls: Hey, thanks for having a look. I _should_ be okay with the two-parent pattern, since it was just checked earlier in this function. But at the same time I share your suspicion that something could be up with that assumption -- maybe we might be missing some more false negatives?

areveny:
```
def func_invalid3(var):
    """'msg' is not defined in one handler."""
    err_message = False
    try:
        return 1 / var.some_other_func()
    except AttributeError:
        msg: str
    except ZeroDivisionError:
        if err_message:
            msg = "Division by 0"
        else:
            msg = None
    print(msg)  # [used-before-assignment]
```

Yeah, this would be a false negative with the check.

jacobtylerwalls: You're right, this is another "don't assume direct parentage" false negative like #5582. I can address it on this branch.

jacobtylerwalls: Test case added in 054fb68

pylint/checkers/variables.py
```diff
@@ -722,28 +723,51 @@ def _uncertain_nodes_in_except_blocks(
...
+    def _defines_name_raises_or_returns(
+        name: str, handler: nodes.ExceptHandler
+    ) -> bool:
+        """Return True if some child of `handler` defines the name `name`,
+        raises, or returns.
+        """
+        for stmt in handler.get_children():
+            if isinstance(stmt, (nodes.Raise, nodes.Return)):
+                return True
+            if isinstance(stmt, nodes.Assign) and stmt.targets[0].as_string() == name:
```
areveny, line 754:
```suggestion
            if isinstance(stmt, nodes.Assign):
                 for target in stmt.targets:
                     if isinstance(target, nodes.AssignName) and target.name == name:
                         return True
                     if isinstance(target, (nodes.Tuple, nodes.List)):
                         for elt in utils.get_all_elements(target):
                             if isinstance(elt, nodes.Const) and elt.value == name:
                                 return True
```

Three parts to this suggestion. 

1) I figured out the case where `len(stmt.targets) > 1`. Looping over `stmt.targets` catches this case:

```
except ZeroDivisionError:
   bsg = msg = "Division by 0"
print(msg)
``` 

2) `as_string` is primarily used to format message parameters, and it seems better to avoid it in comparisons. 

3) Check target unpacking to `List` and `Tuple`, which can recursively contain more `List` or `Tuple` to catch this case:

```
asd, msg = ("asd", "Division by 0")
[asd, msg] = ["asd", "Division by 0"]
```

pylint/checkers/variables.py
```diff
@@ -722,28 +723,51 @@ def _uncertain_nodes_in_except_blocks(
...
+                return True
+            if isinstance(stmt, nodes.If):
+                if (
+                    isinstance(stmt.test, nodes.NamedExpr)
+                    and stmt.test.target.name == name
+                ):
+                    return True
+                if isinstance(stmt.test, nodes.Call):
+                    for arg_or_kwarg in stmt.test.args + stmt.test.keywords:
```
areveny, line 763:
```suggestion
                    for arg_or_kwarg in stmt.test.args + [keyword.value for key word in stmt.test.keywords]:
```

`keywords` is an array of `nodes.Keyword` wrapping `nodes.NamedExpr`.

```
    except AttributeError:
        if print(bsg := var.get_msg(), sep=(msg:=var.get_bsg())):
            pass
```

pylint/checkers/variables.py
jacobtylerwalls: Thank you, this is fantastic help.

areveny (approved):
Great! Thanks for the changes and really nice simplification with the last commit.

I'm satisfied with this, if we could get a second approval I think this is ready to merge. Thanks!

tests/functional/u/use/used_before_assignment_except_handler_for_try_with_return.py
```diff
@@ -49,3 +49,66 @@ def func_ok3(var):
...
+    try:
+        return 1 / var.some_other_func()
+    except AttributeError:
+        msg: str
+    except ZeroDivisionError:
+        msg = "Division by 0"
+    print(msg)  # [used-before-assignment]
+
+
+def func_invalid3(var):
```
DanielNoord, line 100: These tests are great! Just one remark: `func_invalid3` is not actually testing anything new compared to `func_invalid2`. I think we want to test the assignment in `if` statements here, which would be best tested by removing the `AttributeError` except handler and checking if not message is being raised. Currently this just tests that `msg: str` works as expected, if the handling of the `if` statement is incorrect we wouldn't notice.

pylint/checkers/variables.py
```diff
@@ -703,17 +703,18 @@ def _uncertain_nodes_in_except_blocks(
         for other_node in found_nodes:
             other_node_statement = other_node.statement(future=True)
             # Only testing for statements in the except block of TryExcept
-            if not (
-                isinstance(other_node_statement.parent, nodes.ExceptHandler)
-                and isinstance(other_node_statement.parent.parent, nodes.TryExcept)
-            ):
+            closest_except_handler = utils.get_node_first_ancestor_of_type(
```
DanielNoord, line 706: Gotta love this `util` 😄

pylint/checkers/variables.py
```diff
@@ -703,17 +703,18 @@ def _uncertain_nodes_in_except_blocks(
...
+                other_node_statement, nodes.ExceptHandler
+            )
+            if not closest_except_handler:
                 continue
             # If the other node is in the same scope as this node, assume it executes
-            if other_node_statement.parent.parent_of(node):
+            if closest_except_handler.parent_of(node):
                 continue
+            closest_try_except: nodes.TryExcept = closest_except_handler.parent
```
DanielNoord, line 714: I think it would be better to add an ignore wherever that is necessary and then wait for `astroid`'s typing to be updated. If we update `parent` of `ExceptHandler` at some point we wouldn't notice that this typing is redundant.

DanielNoord (changes requested):
Small changes, but this looks great!

@jacobtylerwalls just to get this right, what do you propose in terms of issue management. What should be opened/closed after this has been merged?

tests/functional/u/use/used_before_assignment_except_handler_for_try_with_return.py
jacobtylerwalls: Actually, I think it does test something different. The first implementation didn't catch this case (by @areveny), and we discussed it [here](https://github.com/PyCQA/pylint/pull/5764#discussion_r798116392).

Should I flesh out the docstring to clarify?

pylint/checkers/variables.py
jacobtylerwalls: `mypy` didn't make me do this, I just thought I was helping readability! I'll change this to an assert instead to make more explicit that I'm depending on the parent being a TryExcept.

DanielNoord: Ah okay, never mind then. I'm not sure about performance impact but typing is probably better than asserts? So we should probably revert 😅

tests/functional/u/use/used_before_assignment_except_handler_for_try_with_return.py
DanielNoord: Hmm could it be that something changed in the implementation that made that test different? 

```python
def func_invalid3(var):
    """'msg' is not defined in one handler."""
    err_message = False
    try:
        return 1 / var.some_other_func()
    except ZeroDivisionError:
        if err_message:
            msg = "Division by 0"
        else:
            msg = None
    print(msg)
```

This shouldn't raise right?

jacobtylerwalls: Sorry, I don't think I'm following you very well. 😅 . I thought you were proposing I take `func_invalid3` and rewrite it like you just did in the comment directly above this one. But I'm saying that `func_invalid3` as written catches a false negative that this PR didn't address until @areveny pointed it out and I fixed it (to stop assuming direct parentage), so I think the test has value as written. But I think I may be missing your point still, perhaps? Would you like to me _add_ this as another test, so it becomes `func_ok6`?

DanielNoord: Yeah I'm not making many sense. Let me rephrase:
I don't really understand why the third function needs to be tested in combination with the `msg: str` statement. Why would the behaviour of the message change when there is a combination of `msg: str` and `ifs` in two expect handlers. Imo it makes more sense to test those separately.
However, you have discussed this a lot with @areveny so perhaps there is something difficult going on when both of those patterns are present and we should test them together? Then perhaps the docstring should be added.
While trying to find out if I could find this relatedness between `msg: str` and the `ifs` I stumbled upon the example I gave above. I think the current version raises on this code but it shouldn't, right? That is a separate issue from the one with which I started this comment thread, but it might be related.

Sorry for being so unclear 😅 Long day..

jacobtylerwalls: No problem, I can explain.

> However, you have discussed this a lot with @areveny so perhaps there is something difficult going on when both of those patterns are present and we should test them together?

In a word, yup.

I can change `msg: str` to `pass` if that would be more clear that it doesn't have to do with type annotations. However, having two handlers where the one with the name is nested under if or with is the essence of the test case. All of the tests for correct positives require multiple except handlers. This one in involves nesting under an if or with just to catch the faulty assumption the code first made about direct parentage. 

"Round 1" of most of my control flow PRs adding methods like `_uncertain_nodes_in...` just assumed direct parentage of the node being assigned by its containing try/except. After you pointed me to `utils.get_node_first_ancestor_of_type` I did #5582 to stop assuming this and catch more false negatives, but that PR didn't touch this method. (I guess I was trying to take small bites! 😅 ) @areveny spotted the false negative, and I decided it should be done in this PR. So if you check out the first commit on the branch, you'll see it fails the test case as written.

***
And then you found a false positive for me to investigate! I'll see what's up with that.

DanielNoord: > I can change `msg: str` to `pass` if that would be more clear that it doesn't have to do with type annotations. However, having two handlers where the one with the name is nested under if or with is the essence of the test case. All of the tests for correct positives require multiple except handlers. This one in involves nesting under an if or with just to catch the faulty assumption the code first made about direct parentage.

But if there was only a single expect handler with nested if and the message isn't raised, won't the functionality of nested ifs be tested as well? If it does not work (ie., assignment in nested ifs is not recognised) the message would be raised incorrectly. 

> And then you found a false positive for me to investigate! I'll see what's up with that.

👍

jacobtylerwalls: > But if there was only a single expect handler with nested if and the message isn't raised, won't the functionality of nested ifs be tested as well? If it does not work (ie., assignment in nested ifs is not recognised) the message would be raised incorrectly.

Ah, no, I don't think testing a correct positive and correct negative will necessarily go down the same code path. You're talking about testing a correct negative (which is good to test!), but that will short-circuit a lot earlier (see `"# If the other node is in the same scope as this node, assume it executes"`), which is why it was probably passing.

pylint/checkers/variables.py
```diff
@@ -722,28 +723,68 @@ def _uncertain_nodes_in_except_blocks(
...
 
+    @staticmethod
+    def _defines_name_raises_or_returns(
+        name: str, handler: nodes.ExceptHandler
+    ) -> bool:
+        """Return True if some child of `handler` defines the name `name`,
+        raises, or returns.
+        """
+
+        def _define_raise_or_return(stmt) -> bool:
```
DanielNoord, line 752: Should we add typing for `stmt` here?

DanielNoord (commented):
One final question!

jacobtylerwalls:
@areveny @DanielNoord thanks for the reviews and bugspotting. I'm glad we're getting a real v1.0 of control flow for try/except into something that should _just work_!

> in terms of issue management

Let's pretend I never called the two issues duplicates! Specifically:
- reopen #2835 
- paste a link there to Marc's "possibly unbound" suggestion in https://github.com/PyCQA/pylint/issues/5524#issuecomment-1025262104
- I'll leave a comment there that the implementation can reuse `NamesConsumer._defines_name_raises_or_returns`

And then I can adjust this PR to say "closes" instead of "partially closes" #5524. Sorry for the extra work, but now I see from discussion we're likely to address the bigger-picture issue with a new message, so better to keep it all separate.

DanielNoord:
I have opened and made the comments as you suggested. I'm going to ask for a review from Pierre as this is quite a large PR again.

Just FYI: I think this should be the last new PR to go into `2.13`. We really (really really really) need to release as we're hitting almost 300 issues/PRs closed with a lot of useful features as well. The release is also blocking some important maintenance work such as the `argparse` refactor.
Any help with any of the open `2.13` issues would be much appreciated. Especially with #5645 as I don't have the necessary hardware to work on it myself...

DanielNoord (approved):

Pierre-Sassoulas (commented):
LGTM, the tests are really nice, and this will be a nice addition to 2.13. Also thank you for providing early feedback @areveny, :)

Small question, I noticed that there could be more tests if we do the permutation for "defined in one handler only" with unpacked iterable, or chained assignment. I suppose it's because it's not necessary to check this again as it's already checked in funcok_4 and func_ok5, and not doing all permutation would speed the functional tests up ?

jacobtylerwalls:
I went ahead and added the tests. After all, just a few comments above I was very keen to tell the world that true positives and true negatives might hit different code paths! 😄

PR #5766, opened 2022-02-03
Avoid using get on a defaultdict

tushar-deepsource:
## Type of Changes

## Description

I was poking around Pylint internals, trying to add new issue categories to the codebase, and stumbled upon this bug.
Since `MessageDefinitionStore._msgs_by_category` is a `defaultdict(list)`, using `.get()` here can break the code on bad input, because it will return `None`, which is not iterable.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -1659,7 +1659,7 @@ class PyLinter(
         else:
             category_id_formatted = category_id
         if category_id_formatted is not None:
-            for _msgid in self.msgs_store._msgs_by_category.get(category_id_formatted):
+            for _msgid in self.msgs_store._msgs_by_category[category_id_formatted]:
                 message_definitions.extend(
                     self._get_messages_to_set(_msgid, enable, ignore_unknown)
                 )


DanielNoord (approved):

pylint/lint/pylinter.py
```diff
@@ -1659,7 +1659,7 @@ def _get_messages_to_set(
         else:
             category_id_formatted = category_id
         if category_id_formatted is not None:
-            for _msgid in self.msgs_store._msgs_by_category.get(category_id_formatted):
+            for _msgid in self.msgs_store._msgs_by_category[category_id_formatted]:
```
Pierre-Sassoulas, line 1662:
```suggestion
            for _msgid in self.msgs_store._msgs_by_category.get(category_id_formatted, []):
```

Pierre-Sassoulas (changes requested):
But the proposed change would also result in a KeyError crash, right ?

pylint/lint/pylinter.py
tusharsadhwani: It won't, as it's a defaultdict. If the key doesn't exist it'll make one by the default factory (`list()`) in this case.

tusharsadhwani:
It won't, as it's a defaultdict. If the key doesn't exist it'll make one by the default factory (`list()` in this case).

Pierre-Sassoulas (approved):
I missed that it was a defaultdict, thank you for the fix !

PR #5774, opened 2022-02-06
wrong-import-position is equivalent to E402

robin-wayve:
## Type of Changes

## Description

As far as I can tell, Pylint's `wrong-import-position` (C0413) and [pycodestyle's E402](https://www.flake8rules.com/rules/E402.html) are checking for the same thing.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -201,7 +201,7 @@ You can see the plugin you need to explicitly `load in the technical reference`_
 4.8 I am using another popular linter alongside pylint. Which messages should I disable to avoid duplicates?
 ------------------------------------------------------------------------------------------------------------
 
-pycodestyle_: unneeded-not, line-too-long, unnecessary-semicolon, trailing-whitespace, missing-final-newline, bad-indentation, multiple-statements, bare-except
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
Thank you for the proposition. Pylint's check is not the same. It means sorted aphabetically and semantically (bultin, then lib, then internal) according to isort.

robin-wayve:
Should this check be listed next to `wrong-import-order` as redundant if you are using isort instead?

I saw an example with output similar to this and wasn't clear if there would be any difference between the two checks.
```py
import os

os.environ['FOO'] = 'bar'

import sys
```

pylint
```
example.py:5:0: C0413: Import "import sys" should be placed at the top of the module (wrong-import-position)
```

flake8
```
example.py:5:1: E402 module level import not at top of file
```

robin-wayve:
Is it only redundant if you have both isort and pycodestyle, perhaps? Anyway, it wasn't immediately clear to me and I hoped either I could improve the docs or be set straight, so thanks for taking the time to reply 😄 .

Pierre-Sassoulas (approved):
I mistaken ``wrong-import position``with ``wrong-import-order``, you're right, thank you for making the doc better 😄 !

PR #5775, opened 2022-02-06
Fix unsafe `utils.safe_infer`

cdce8p:
## Type of Changes

## Description
Came across an error while testing Home Assistant. `safe_infer` here assumes that `value` is a `FunctionDef` to be able to access `value.args.args`. The assumption is never fully checked however. `value` could also be an `<Instance of builtins.function ...>` which would cause an `AttributeError`.

I wasn't able to write a dedicated regression test unfortunately. Home Assistant only managed to break it with some combination of patching a builtin function and custom plugins which import parts of the checked files itself.

The condition was originally added in #5409

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1263,6 +1263,7 @@ def safe_infer(node: nodes.NodeNG, context=None) -> Optional[nodes.NodeNG]:
             if (
                 isinstance(inferred, nodes.FunctionDef)
                 and inferred.args.args is not None
+                and isinstance(value, nodes.FunctionDef)
                 and value.args.args is not None
                 and len(inferred.args.args) != len(value.args.args)
             ):


DanielNoord (approved):

Pierre-Sassoulas (approved):
Nice catch ! home-assistant is in the primer though: https://github.com/PyCQA/pylint/blob/main/tests/primer/packages_to_lint_batch_one.json#L26?, was it only on a branch (not yet merged maybe) ?

cdce8p:
> Nice catch ! home-assistant is in the primer though: https://github.com/PyCQA/pylint/blob/main/tests/primer/packages_to_lint_batch_one.json#L26?, was it only on a branch (not yet merged maybe) ?

It was in `dev`, but only merged a few days ago. The issue was only visible because of a custom plugin and then only if the `time` module was imported in a certain way. Pretty strange tbh. Might be the reason why it didn't show up in the primer tests. I did check astroid, but that seemed to do what was expected.

In the meantime, a fix for Home Assistant has been merged. So it definitely won't show up in the tests anymore.
https://github.com/home-assistant/core/pull/65911

The fix here is never the less useful. If for nothing else then type safety (and mypy).

PR #5786, opened 2022-02-10
Fix parsing of long lines when ``missing-final-newline`` is enabled

### LINKED ISSUE #5724: 'missing-final-newline' is surprisingly slow on a large codebase
### Bug description

I'm trying to enable the 'missing-final-newline' checker in [this PR](https://github.com/diofant/diofant/pull/1202).  Unfortunately, the Github CI run doesn't finish in a reasonable time (i.e. default job timeout).

The codebase of the [Diofant](https://github.com/diofant/diofant/) is big, however I think that this check should be reasonably cheap (shouldn't involve AST parsing etc).  Unfortunately, it's not true and I think it's an issue for pylint.  At least, I would appreciate any workaround.

AFAIR, same is true for 'mixed-line-endings', 'trailing-newlines' and 'trailing-whitespace' checkers.

### Configuration

```ini
see the project's pyproject.toml (on the PR's branch)
```

### Command used

```shell
pylint diofant
```

### Pylint output

```shell
Nothing (timeout)
```

### Expected behavior

Tests pass.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.10.0+ (heads/3.10:045f205ba4, Nov  8 2021, 05:21:54) [GCC 10.2.1 20210110]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #5724.

The stuff at the beginning of this regex pattern isn't actually needed for us to extract our disable comments. This massively speeds up the parsing and unstucks us from the code in the issue.
I haven't added a test case as it is a bit difficult to test "don't look for things we don't need"...

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -129,6 +129,10 @@ Release date: TBA
 
   Closes #5569
 
+* Fix parsing of long lines when ``missing-final-newline`` is enabled.
+
+  Closes #5724
+
 * Fix false positives for ``used-before-assignment`` from using named
   expressions in a ternary operator test and using that expression as
   a call argument.
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -180,6 +180,10 @@ Other Changes
 
   Closes #5177, #5212
 
+* Fix parsing of long lines when ``missing-final-newline`` is enabled.
+
+  Closes #5724
+
 * Fix ``unnecessary_dict_index_lookup`` false positive when deleting a dictionary's entry.
 
   Closes #4716
--- a/pylint/utils/pragma_parser.py
+++ b/pylint/utils/pragma_parser.py
@@ -9,10 +9,7 @@ from typing import Generator, List, Optional
 # so that an option can be continued with the reasons
 # why it is active or disabled.
 OPTION_RGX = r"""
-    \s*                # Any number of whitespace
-    \#?                # One or zero hash
-    .*                 # Anything (as much as possible)
-    (\s*               # Beginning of first matched group and any number of whitespaces
+    (                  # Beginning of first matched group and any number of whitespaces
     \#                 # Beginning of comment
     .*?                # Anything (as little as possible)
     \bpylint:          # pylint word and column
--- a/pylint/utils/pragma_parser.py
+++ b/pylint/utils/pragma_parser.py
@@ -11,7 +11,7 @@ from typing import Generator, List, Optional
 OPTION_RGX = r"""
     (                  # Beginning of first matched group and any number of whitespaces
     \#                 # Beginning of comment
-    .*?                # Anything (as little as possible)
+    \s*?               # Any whitespaces (as little as possible)
     \bpylint:          # pylint word and column
     \s*                # Any number of whitespaces
     ([^;#\n]+))        # Anything except semicolon or hash or newline (it is the second matched group)


DanielNoord:
The first pattern does not allow:
`# a comment # pylint: disable=this`.
This can be fixed with the second pattern, which disallows:
`# a comment # something in between pylint: disable=this`.

However, do we really want to support `something in between`? Is there any use case for adding words in between the `#` and `pylint`?

DanielNoord:
Seems like we're using the second pattern ourselves in:
`# -*- pylint: disable=no-init,too-few-public-methods, useless-object-inheritance`.

I think it makes sense to disallow this, but this would probably need to be deprecated. That would also mean that the performance fix can only go in `3.x`.
1) Do we want to disallow any non-whitespace characters between the `#` and `pylint` for disable comments with the knowledge that this would boost the performance of the regex pattern that looks for these?
2) If yes, do we want to make this change in `3.x` or in `2.x`?

ChangeLog
```diff
@@ -129,6 +129,10 @@ Release date: TBA
 
   Closes #5569
 
+* Fix parsing of long lines when ``missing-final-newline`` is enabled.
```
Pierre-Sassoulas, line 132:
```suggestion
* Optimize parsing of long lines when ``missing-final-newline`` is enabled.
```

doc/whatsnew/2.13.rst
```diff
@@ -180,6 +180,10 @@ Other Changes
 
   Closes #5177, #5212
 
+* Fix parsing of long lines when ``missing-final-newline`` is enabled.
```
Pierre-Sassoulas, line 183:
```suggestion
* Optimize parsing of long lines when ``missing-final-newline`` is enabled.
```

pylint/utils/pragma_parser.py
```diff
@@ -9,12 +9,9 @@
...
 # why it is active or disabled.
 OPTION_RGX = r"""
-    \s*                # Any number of whitespace
-    \#?                # One or zero hash
-    .*                 # Anything (as much as possible)
-    (\s*               # Beginning of first matched group and any number of whitespaces
+    (                  # Beginning of first matched group and any number of whitespaces
     \#                 # Beginning of comment
-    .*?                # Anything (as little as possible)
+    \s*?               # Any whitespaces (as little as possible)
```
Pierre-Sassoulas, line 14: This is a breaking change, because every changes is a breaking change (*heavy sigh*), but it's probably worth it. 

We could do:
```suggestion
    .*?               # Anything (as little as possible)
```
Until we release 3.0, but then we need a way to make the deprecation known. So should we warn the user if we match a pylint comment with ``.*?`` but not with ``\s*?`` ? it would actually decrease performance in 2.X, possibly by a lot [citation required].

Another solution is to consider that it was an unintended feature and just consider this a performance fix. How many users are going to be pissed off by this decision... ? I wish I knew.

DanielNoord: I don't think there is a real performance change if we make `.*?` a capturing group and check if it contains anything other than spaces whenever we use this regex pattern. That should be relatively straightforward.

Pierre-Sassoulas: Wouldn't it solve the breaking change issue ?

DanielNoord: I don't think so. The breaking change would be to remove `.*` completely and replace it with nothing or with `\s*`. To deprecate this we can use a capturing group and emit a warning when we match anything other than spaces.

skirpichev:
This works for me:
```diff
--- pragma_parser.py.old        2022-02-11 15:35:07.864685538 +0300
+++ pragma_parser.py    2022-02-11 15:35:14.156481194 +0300
@@ -9,11 +9,8 @@
 # so that an option can be continued with the reasons
 # why it is active or disabled.
 OPTION_RGX = r"""
-    \s*                # Any number of whithespace
-    \#?                # One or zero hash
-    .*                 # Anything (as much as possible)
-    (\s*               # Beginning of first matched group and any number of whitespaces
-    \#                 # Beginning of comment
+    (?:\s*|\s*\#.*(?=\#.*?\bpylint:))
+    (\#                 # Beginning of comment
     .*?                # Anything (as little as possible)
     \bpylint:          # pylint word and column
     \s*                # Any number of whitespaces
```

DanielNoord:
> This works for me:
> 
> ```diff
> --- pragma_parser.py.old        2022-02-11 15:35:07.864685538 +0300
> +++ pragma_parser.py    2022-02-11 15:35:14.156481194 +0300
> @@ -9,11 +9,8 @@
>  # so that an option can be continued with the reasons
>  # why it is active or disabled.
>  OPTION_RGX = r"""
> -    \s*                # Any number of whithespace
> -    \#?                # One or zero hash
> -    .*                 # Anything (as much as possible)
> -    (\s*               # Beginning of first matched group and any number of whitespaces
> -    \#                 # Beginning of comment
> +    (?:\s*|\s*\#.*(?=\#.*?\bpylint:))
> +    (\#                 # Beginning of comment
>      .*?                # Anything (as little as possible)
>      \bpylint:          # pylint word and column
>      \s*                # Any number of whitespaces
> ```

Have you tested this against `tests/functional/l/line_too_long.py`? Locally that fails with this change for me.

skirpichev:
Hmm, I see.  Probably, we should have another alternative to match comments, i.e. `^\s*\#.*`.  With that change - tests pass for me.

DanielNoord:
```diff
diff --git a/pylint/utils/pragma_parser.py b/pylint/utils/pragma_parser.py
index 5ef4ef481..482e295ef 100644
--- a/pylint/utils/pragma_parser.py
+++ b/pylint/utils/pragma_parser.py
@@ -12,8 +12,8 @@ OPTION_RGX = r"""
-    \s*                # Any number of whitespace
-    \#?                # One or zero hash
-    .*                 # Anything (as much as possible)
-    (\s*               # Beginning of first matched group and any number of whitespaces
-    \#                 # Beginning of comment
+    (?:\s*|\s*\#.*|^\s*\#.*(?=\#.*?\bpylint:))
+    (\#                 # Beginning of comment
     .*?                # Anything (as little as possible)
     \bpylint:          # pylint word and column
     \s*                # Any number of whitespaces
```

Nice job! I think that actually works. If you want you can create a PR yourself @skirpichev, you are the one that came up with it!

skirpichev:
> If you want you can create a PR yourself

No, I think you did almost everything.

BTW, probably different conditions in the `(?:)` group require comments.  And there should be a regression test, isn't?  I don't know how to do this for the pylint.  (I use pytest-timeout for the diofant for similar performance issues.)

pylint/utils/pragma_parser.py
Pierre-Sassoulas: But why would we be doing the breaking change if there is no performance issues if we keep the current behavior ?

Pierre-Sassoulas:
>  (I use pytest-timeout for the diofant for similar performance issues.)

We're using.. nothing I guess. We have a global timeout for the tests in CI, and a benchmark using pytest-benchmark that we do not look at (so it's basically a time waster in our CI right now, that we hope to use later). Is pytest-timeout is something that would solve this issue ?

skirpichev:
> Is pytest-timeout is something that would solve this issue?

Yes, that's [how I address](https://github.com/diofant/diofant/blob/0677d240eb5de697f851c6c844fefc8039754edc/diofant/tests/polys/test_factortools.py#L913-L933) such issues (_local_ timeouts for tests).   It's a little tricky, because you need to find out a timeout, that triggers problem on different machines (e.g. locally for you and on the Github Actions CI).

DanielNoord:
Superseded by #5925, thanks to @skirpichev  😄

Issue #2913, opened 2019-05-10
disable=too-few-public-methods doesn't work for empty class


roelvandergoot:
### Steps to reproduce
1. Create file base.py with the following content:

<pre>
"""Module doc string."""

class Base:
    """My base class."""
    # pylint: disable=too-few-public-methods
</pre>

2. Run pylint base.py

### Current behavior
<pre>
************* Module base
base.py:3:0: R0903: Too few public methods (0/2) (too-few-public-methods)

---------------------------------------------------------------------
Your code has been rated at 0.00/10
</pre>

### Expected behavior

<pre>

---------------------------------------------------------------------
Your code has been rated at 10.00/10
</pre>

### pylint --version output

<pre>
pylint 2.3.1
astroid 2.2.5
Python 3.7.3 (default, Apr  3 2019, 05:39:12)
[GCC 8.3.0]
</pre>

PCManticore:
Thanks for the report.

PCManticore:
Alright, the following should work:
```
class Base: # pylint: disable=too-few-public-methods
    """My base class."""
```

The given example doesn't work because the message pragma need to incorporate the object for which you are disabling / enabling a message. When used inside the body of the class, the pragma  control does not get applied as the message is emitted on the class itself (thus the line of `class Base` definition) rather than the body of the class (which would be a different line altogether).

roelvandergoot:
Claudio,

Why can I put said pragma inside a non-empty class then?

Cheers :),
Roel.

Pierre-Sassoulas:
The error happens on the class, in  the code of the class it's too late to disable it.

There is two way to do this:
```python
# pylint: disable=too-few-public-methods
class Base:
    """My base class."""
# pylint: enable=too-few-public-methods


class BaseCorrect:  # pylint: disable=too-few-public-methods
    """My base class."""
```

PR #5802, opened 2022-02-13
Fixes to tests due to (possible) new ``tolineno`` behaviour

DanielNoord:

## Type of Changes

## Description

These are the changes necessary for the tests to allow https://github.com/PyCQA/astroid/pull/1351 to pass the `pylint` test suite.

I'll annotate the changes to explain why something changed and why I think this is not a bad thing.

Obviously, this is all dependent on https://github.com/PyCQA/astroid/pull/1351 actually landing, which will need to be discussed in the `astroid` repository.

_Expect a push to https://github.com/PyCQA/astroid/pull/1351 in a couple of minutes to undraft that PR._

### DIFF
--- a/tests/checkers/unittest_stdlib.py
+++ b/tests/checkers/unittest_stdlib.py
@@ -49,10 +49,8 @@ class TestStdlibChecker(CheckerTestCase):
         """
 
         def infer_func(
-            node: Name, context: Optional[Any] = None
-        ) -> Iterator[
-            Union[Iterator, Iterator[AssignAttr]]
-        ]:  # pylint: disable=unused-argument
+            node: Name, context: Optional[Any] = None  # pylint: disable=unused-argument
+        ) -> Iterator[Union[Iterator, Iterator[AssignAttr]]]:
             new_node = nodes.AssignAttr(attrname="alpha", parent=node)
             yield new_node
 
--- a/tests/checkers/unittest_unicode/unittest_bad_chars.py
+++ b/tests/checkers/unittest_unicode/unittest_bad_chars.py
@@ -130,6 +130,7 @@ class TestBadCharsChecker(pylint.testutils.CheckerTestCase):
             # string
             module = astroid.MANAGER.ast_from_string(file)
         except AstroidBuildingError:
+            # pylint: disable-next=redefined-variable-type
             module = cast(nodes.Module, FakeNode(file.read_bytes()))
 
         expected = [
--- a/tests/functional/ext/broad_try_clause/broad_try_clause_extension.txt
+++ b/tests/functional/ext/broad_try_clause/broad_try_clause_extension.txt
@@ -1,4 +1,4 @@
 too-many-try-statements:5:0:10:8::try clause contains 3 statements, expected at most 1:UNDEFINED
 too-many-try-statements:12:0:17:8::try clause contains 3 statements, expected at most 1:UNDEFINED
-too-many-try-statements:19:0:27:8::try clause contains 4 statements, expected at most 1:UNDEFINED
+too-many-try-statements:19:0:25:8::try clause contains 4 statements, expected at most 1:UNDEFINED
 too-many-try-statements:29:0:44:8::try clause contains 7 statements, expected at most 1:UNDEFINED
--- a/tests/functional/i/invalid/invalid_name.py
+++ b/tests/functional/i/invalid/invalid_name.py
@@ -58,13 +58,8 @@ def dummy_decorator(aaabc, bbbcd):
     return real_decorator
 
 
-@dummy_decorator(1, [
-    0  # Fixing #119 should make this go away
-# C0103 always points here - line 61  # [invalid-name]
-
-
-])
-def a_very_very_very_long_function_name_WithCamelCase_to_make_it_sad():  # Should be line 65
+@dummy_decorator(1, [0])
+def a_very_very_very_long_function_name_WithCamelCase_to_make_it_sad():  # [invalid-name]
     """Docstring"""
     print('LOL')
 
--- a/tests/functional/r/regression/regression_2913.py
+++ b/tests/functional/r/regression/regression_2913.py
@@ -14,6 +14,6 @@ class BaseCorrect:  # pylint: disable=too-few-public-methods
     """My base class."""
 
 
-class BaseInner:  # [too-few-public-methods]
+class BaseInner:
     # pylint: disable=too-few-public-methods
     """My base class."""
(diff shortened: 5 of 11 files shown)


tests/functional/ext/broad_try_clause/broad_try_clause_extension.txt
```diff
@@ -1,4 +1,4 @@
 too-many-try-statements:5:0:10:8::try clause contains 3 statements, expected at most 1:UNDEFINED
 too-many-try-statements:12:0:17:8::try clause contains 3 statements, expected at most 1:UNDEFINED
-too-many-try-statements:19:0:27:8::try clause contains 4 statements, expected at most 1:UNDEFINED
+too-many-try-statements:19:0:25:8::try clause contains 4 statements, expected at most 1:UNDEFINED
```
DanielNoord, line 3: This is due to the change of no longer adding the `finally:` child to the end line number calculation of a `TryExcept` node.

tests/functional/i/invalid/invalid_name.py
```diff
@@ -59,12 +59,12 @@ def wrapper(*args, **kwargs):
...
 @dummy_decorator(1, [
-    0  # Fixing #119 should make this go away
-# C0103 always points here - line 61  # [invalid-name]
+    0
+
 
 
 ])
-def a_very_very_very_long_function_name_WithCamelCase_to_make_it_sad():  # Should be line 65
+def a_very_very_very_long_function_name_WithCamelCase_to_make_it_sad():  # [invalid-name]
```
DanielNoord, line 67: This actually fixes the long standing issue that is mentioned in the comment here.

tests/functional/r/regression/regression_2913.py
```diff
@@ -14,6 +14,6 @@ class BaseCorrect:  # pylint: disable=too-few-public-methods
     """My base class."""
 
 
-class BaseInner:  # [too-few-public-methods]
+class BaseInner:
```
DanielNoord, line 17: This was actually inconsistent with how we describe this should work, see:
https://github.com/PyCQA/pylint/blob/b31dd5623966069386b9df3ccc08e2834aca621d/pylint/utils/file_state.py#L70-L85

According to this, a disable on L18 should affect L17. Apparently it did not, and this got fixed with this change.

The original issue actually discusses this as well:
https://github.com/PyCQA/pylint/issues/2913

The reporter wonders why this isn't behaving like our internal comment is telling it should. I think this is thus a good change (and perhaps worthy of a changelog).

tests/functional/i/invalid/invalid_name.py
Pierre-Sassoulas: #119 is closed as 'won't fix' but apparently it was not specific to python 3.6 ?

DanielNoord: I think it's a combination of `parser` changes and how we do `tolineno` right now (as in, without `end_lineno`). In any case, this change clearly fixes it 😄

DanielNoord:
@Pierre-Sassoulas This can be rerun after the release of `astroid` and then merged. We will need to merge this first as all test suites will fail without these changes.

Pierre-Sassoulas (approved):
Thank you for fixing this, astroid 2.10.0 is released let's merge 😄

tests/checkers/unittest_stdlib.py
```diff
@@ -49,10 +49,8 @@ def test_deprecated_no_qname_on_unexpected_nodes(self) -> None:
         """
 
         def infer_func(
-            node: Name, context: Optional[Any] = None
-        ) -> Iterator[
-            Union[Iterator, Iterator[AssignAttr]]
-        ]:  # pylint: disable=unused-argument
+            node: Name, context: Optional[Any] = None  # pylint: disable=unused-argument
+        ) -> Iterator[Union[Iterator, Iterator[AssignAttr]]]:
```
Pierre-Sassoulas, line 53: This actually make the formatting better, I like this. But it also means some work when upgrading 😿

DanielNoord:
This is actually pretty problematic. Since `tolineno` is now version dependent, disables have become version dependent as well...
We might need a hotfix for `astroid` or change how we determine the position of disables. I'll investigate.

DanielNoord:
Pushed a PR to `astroid` to fix the underlying issue. Tested against the problem here and that was resolved 😄

z3ntu:
What's the status on this? On Alpine Linux edge pylint is currently broken because astroid 2.10 is in the repos but pylint isn't yet compatible with that. 
Is this patch ready to be picked into Alpine or is something still holding this up?

Pierre-Sassoulas:
> What's the status on this? 

We're working on it, we're going to release astroid 2.11 and this is the first MR we'll merge once this is done.

> pylint is currently broken because astroid 2.10

You should use pylint 2.12.2 with the astroid version defined in our packaging (<2.10, but that should not matter). Don't try to set the version astroid yourself and pylint won't break.

tests/checkers/unittest_unicode/unittest_bad_chars.py
```diff
@@ -130,6 +130,7 @@ def test_find_bad_chars(
             # string
             module = astroid.MANAGER.ast_from_string(file)
         except AstroidBuildingError:
+            # pylint: disable-next=redefined-variable-type
```
DanielNoord, line 133: This is potentially a regression, but it could also be something we did not pick up previously and are now incorrectly picking up (if you understand what I mean).

The warning makes _some_ sense, but perhaps we should allow this use of ``cast``

Pierre-Sassoulas: Yeap, I think astroid got better so a new false positive appeared in pylint (or is it ? It seems to make sense). Happens often when upgrading astroid in fact 😄

Pierre-Sassoulas (approved):
Happy to finally be upgrading astroid 🎉 Thank you @DanielNoord !

tests/checkers/unittest_unicode/unittest_bad_chars.py
DanielNoord: Yeah, but I don't understand why it happened `2.10>2.11`. Like, we did nothing for this in `2.11`, right? Well anyway, doesn't hurt too much.

_And I don't think I can take waiting for a 2.11.1 hotfix anymore_

Pierre-Sassoulas: >And I don't think I can take waiting for a 2.11.1 hotfix anymore

We definitely need to release soon, there are hundred of issues fixed in 2.13, some of them get reopened as duplicate frequently.

PR #5812, opened 2022-02-16
Fix #5112: Prevent `used-before-assignment` if named expression found first in container

### LINKED ISSUE #5112: Invalid used-before-assignment
### Bug description

```python
from random import randint

asd2 = {
    **({'param': x} if (x := randint(0, 1)) else {}),
}
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
E0601: Using variable 'x' before assignment (used-before-assignment)
```

### Expected behavior

no error.

### Pylint version

```shell
pylint 2.9.6
astroid 2.6.6
Python 3.9.7 (default, Aug 30 2021, 00:00:00) 
[GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
```

### OS / Environment

Fedora 34

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Prevent a false positive for `used-before-assignment` when a named expression (walrus operator) is found first in a container.

For instance, this passed:

```python
assert x := True
```
But this didn't (it's a tuple):
```python
assert x := True, x
```
Notice this should not pass:
```python
assert x, x := True
```
Closes #5112

Of course, this can wait for 2.14, no problem.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1971,7 +1971,7 @@ class VariablesChecker(BaseChecker):
                     # same line as the function definition
                     maybe_before_assign = False
                 elif (
-                    isinstance(  # pylint: disable=too-many-boolean-expressions
+                    isinstance(
                         defstmt,
                         (
                             nodes.Assign,
@@ -1981,26 +1981,7 @@ class VariablesChecker(BaseChecker):
                             nodes.Return,
                         ),
                     )
-                    and (
-                        isinstance(defstmt.value, nodes.IfExp)
-                        or (
-                            isinstance(defstmt.value, nodes.Lambda)
-                            and isinstance(defstmt.value.body, nodes.IfExp)
-                        )
-                        or (
-                            isinstance(defstmt.value, nodes.Call)
-                            and (
-                                any(
-                                    isinstance(kwarg.value, nodes.IfExp)
-                                    for kwarg in defstmt.value.keywords
-                                )
-                                or any(
-                                    isinstance(arg, nodes.IfExp)
-                                    for arg in defstmt.value.args
-                                )
-                            )
-                        )
-                    )
+                    and VariablesChecker._maybe_used_and_assigned_at_once(defstmt)
                     and frame is defframe
                     and defframe.parent_of(node)
                     and stmt is defstmt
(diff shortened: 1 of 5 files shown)


pylint/checkers/variables.py
```diff
@@ -2074,6 +2055,29 @@ def _is_variable_violation(
...
+        if isinstance(value, nodes.IfExp):
+            return True
+        if isinstance(value, nodes.Lambda) and isinstance(value.body, nodes.IfExp):
+            return True
+        if isinstance(value, nodes.Call) and (
+            any(isinstance(kwarg.value, nodes.IfExp) for kwarg in value.keywords)
+            or any(isinstance(arg, nodes.IfExp) for arg in value.args)
+        ):
+            return True
+        return False
```
Pierre-Sassoulas, line 2079: We could return directly the content of the if, but I guess you did it that way for better extendability.
```suggestion
        return isinstance(value, nodes.Call) and (
            any(isinstance(kwarg.value, nodes.IfExp) for kwarg in value.keywords)
            or any(isinstance(arg, nodes.IfExp) for arg in value.args)
        )
```

tests/functional/u/undefined/undefined_variable_py38.py
```diff
@@ -155,3 +155,12 @@ def __init__(self, value):
 
 
 dummy = Dummy(value=val if (val := 'something') else 'anything')
+
+def expression_in_ternary_operator_inside_container():
+    """Named expression in ternary operator: inside container"""
+    return [val2 if (val2 := 'something') else 'anything']
+
```
Pierre-Sassoulas, line 162:
```suggestion

def expression_in_ternary_operator_inside_container_tuple():
    """Same case, using a tuple"""
    return [(val3, val3) if (val3 := 'something') else 'anything']
```
This test is not strictly necessary but I have to admit understanding why the next test case is failing took me way too long, so I hope this help with future reader understanding 😄

Pierre-Sassoulas (approved):
I don't know about where to put this. On one hand this is a bug fix, it feels wrong not to put it in if the work is done. On the other hand we have a LOT of thing in 2.13 already and we need to focus on actually releasing at some point. We can probably put this in 2.13 as I took the time to review now.

tests/functional/u/undefined/undefined_variable_py38.py
```diff
@@ -155,3 +155,17 @@ def __init__(self, value):
...
+
+def expression_in_ternary_operator_inside_container_tuple():
+    """Same case, using a tuple inside a 1-element list"""
+    return [(val3, val3) if (val3 := 'something') else 'anything']
+
+
+def expression_in_ternary_operator_inside_container_wrong_position():
+    """2-element list where named expression comes too late"""
+    return [val3, val3 if (val3 := 'something') else 'anything']  # [used-before-assignment]
```
Pierre-Sassoulas, line 171:
```suggestion
    return [val3, val3 if (val3 := 'something') else 'anything']  # [used-before-assignment]
    

def regression_5817():
    """Test case from https://github.com/PyCQA/pylint/issues/5817."""
    try:
        some_list = [e for e in (1, 2, 3) if e]  # Using variable 'e' before assignment!
    except Exception as e:
        pass    
        
```

Pierre-Sassoulas (changes requested):
There's been a new issue opened regarding this problem, I think we could add this regression test, what do you think ?

tests/functional/u/undefined/undefined_variable_py38.py
jacobtylerwalls: I'm opening a separate PR :-)

Pierre-Sassoulas (approved):
Sounds good to me.

tests/functional/u/undefined/undefined_variable_py38.py
DanielNoord, line 171: Pff... Man that's convoluted. Can't we raise an `don't-write-code-like-this` warning here? 😅

DanielNoord (approved):

tests/functional/u/undefined/undefined_variable_py38.py
Pierre-Sassoulas: Haha, I had to take 2-3mn and run the code to understand myself 😄 But do we really want to make the kind of person that would write this, flock to our bug tracker to complain 😉 ?

Issue #5809, opened 2022-02-15
🔍 Improve CPU count detection in cgroup environments


connorbrinton:
### Current problem

PyLint [currently uses](https://github.com/PyCQA/pylint/blob/8c0062f5ac0cfd80a74568e36d1e68e5a128c7f5/pylint/lint/run.py#L22-L30) `sched_getaffinity` or `multiprocessing.cpu_count` to identify the number of CPUs available to the system, falling back to `1` if neither function is available. This doesn't work properly in cgroup environments, where CPU limits may be imposed as a fraction of total CPU time available to the set of processes.

### Desired solution

It would be great if PyLint were to automatically detect the fraction of CPUs available to it when determining how many subprocesses to launch. Currently in our Kubernetes-based CI system, PyLint is launching 16 processes which all compete for a single CPU's worth of time. This ultimately slows down the checking process, which we want to be fast 🙂 

Alternatively, it would be great to be able to set an environment variable to control the number of subprocesses to launch. This would make it possible for us set the number of subprocesses in the same place where we set the number of CPUs available to the CI process.

As a workaround, we're going to modify our launching script to compute the fraction of CPUs available to the process and pass that using the `-j` argument to PyLint.

### Additional context

Determining the amount of CPU available to a Python program is a somewhat widespread problem, here are some related resources:
* Python issue requesting this ability in the standard library: https://bugs.python.org/issue36054
* Message showing how to manually query cgroup quotas: https://bugs.python.org/msg353690
* Conan PR using similar logic to load cgroup quota information: https://github.com/conan-io/conan/pull/5466/files
* Utility using `LD_PRELOAD` to hijack CPU count requests to return computed CPU availability: https://github.com/agile6v/container_cpu_detection

PR #5816, opened 2022-02-17
Don't assume test runners have more than 2 cores available

DanielNoord:

## Type of Changes

## Description

This fixes the CI issues we were seeing in #5778 and #5805.
The second argument of these tuples is used to determine the number of jobs. However, the Linux runner only has two cores available. I'm not sure why this is different on Linux and Windows, but I think because we were trying to use more processes than available cores the test became either very slow or completely locked up.
You can see that we these changes both #5778 and #5805 pass the tests in expected times on all environments.

### DIFF
--- a/tests/test_check_parallel.py
+++ b/tests/test_check_parallel.py
@@ -413,15 +413,9 @@ class TestCheckParallel:
             (3, 1, 1),
             (3, 1, 2),
             (3, 1, 3),
-            (3, 5, 1),
-            (3, 5, 2),
-            (3, 5, 3),
             (10, 2, 1),
             (10, 2, 2),
             (10, 2, 3),
-            (2, 10, 1),
-            (2, 10, 2),
-            (2, 10, 3),
         ],
     )
     def test_compare_workers_to_single_proc(self, num_files, num_jobs, num_checkers):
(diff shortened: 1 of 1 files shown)


cdce8p:
Should we can consider reverting #5755 then?

DanielNoord:
> Should we can consider reverting #5755 then?

We could I think. It might make sense, although I'm not sure if there is a difference between organisations' and individual runners and if those could affect run time (I don't think so).

cdce8p:
> > Should we can consider reverting #5755 then?
> 
> We could I think. It might make sense, although I'm not sure if there is a difference between organisations' and individual runners and if those could affect run time (I don't think so).

AFAIK the main / only difference is the number of concurrent runners.

Pierre-Sassoulas:
So we're closing this one and reverting #5755 then ? Just a note, it seems the underlying issue would be fixed by https://github.com/PyCQA/pylint/issues/5809

DanielNoord:
@Pierre-Sassoulas No! This needs to be merged first as this is what's causing all the time outs.

Haven't looked at the code yet, but I don't think that PR fixes this. I believe the way this test is invoked sets the number of jobs automatically to `10` (or `5`). That PR is referring to when `jobs` is set to `0`, which makes `pylint` try to figure it out itself. That's a different case I believe.

DanielNoord:
Btw, thinking about this: should we check if there are two cores available as well? I feel like that's a bare minimum for any test runner nowadays, but that could still crash/hang these tests..

PR #5825, opened 2022-02-21
Add ``typevar-name-missing-variance`` checker

DanielNoord:

## Type of Changes

## Description

Attempt to make #5221 actually reviewable. It extracts the new checker from the PR.


Pierre-Sassoulas (approved):
Thank you for separating the two concerns @DanielNoord, this indeed make the review less intimidating. (The choice of a new naming style was overwhelming by itself already 😄)

LGTM !

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,68 @@ def _name_became_keyword_in_version(name, rules):
...
+                    suggest_name = re.sub("_co$", "", name)
+                    suggest_name += "_contra"
+                    args_to_display.append((name, "contravariant", suggest_name))
+                if (
+                    kw.arg == "bound"
+                    and len(keywords) == 1
+                    and (name.endswith("_co") or name.endswith("_contra"))
+                ):
+                    suggest_name = re.sub("_contra$|_co$", "", name)
+                    args_to_display.append((name, "invariant", suggest_name))
```
Pierre-Sassoulas, line 2206: Look like this part is not covered by the tests (according to coverall)

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,68 @@ def _name_became_keyword_in_version(name, rules):
...
+        args_to_display: List[Tuple[str, str, str]] = []
+
+        if isinstance(node.parent, nodes.Assign):
+            keywords = node.assign_type().value.keywords
+        elif isinstance(node.parent, nodes.Tuple):
+            keywords = (
+                node.assign_type().value.elts[node.parent.elts.index(node)].keywords
+            )
+        else:
+            return args_to_display
```
Pierre-Sassoulas, line 2180: This line is not tested either.

pylint/checkers/base.py
DanielNoord: I just submitted a commit to fix this 😄

pylint/checkers/base.py
DanielNoord: https://github.com/PyCQA/pylint/pull/5221#discussion_r739526643

😄 It was actually you who suggested to add this. But removed in latest commit.

DanielNoord:
@Pierre-Sassoulas Would you consider changing the `coveralls` setting to "fail" the CI if coverage drops? I think that would make sense. In large and "much discussed" PRs the coveralls comment can get lost and it would be a good check for us. If we don't make it a required CI step we can still merge PRs that do not need to be covered (if there are very any).

DanielNoord:
@Pierre-Sassoulas @cdce8p 

See https://github.com/PyCQA/pylint/pull/5221#discussion_r776884745.

Pylance complains whenever you use `bound` and any of the variants together. Ar we sure they can be used together?

DanielNoord:
@Pierre-Sassoulas Tests are timing out.. 😢

pylint/checkers/base.py
Pierre-Sassoulas: Haha sorry. But yeah if we can't find a way to cover it then it's safe to remove :)

Pierre-Sassoulas:
>  Tests are timing out.. 

I'm out of idea on how to fix this, could we just make the limit higher at least until we release 2.13 ?

Pierre-Sassoulas (approved):
LGTM :) !

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,64 @@ def _name_became_keyword_in_version(name, rules):
...
+
+        if keywords:
+            for kw in keywords:
+                if (
+                    kw.arg == "covariant"
+                    and kw.value.value
+                    and not name.endswith("_co")
+                ):
+                    suggest_name = re.sub("_contra$", "", name)
+                    suggest_name += "_co"
```
cdce8p, line 2188: Why not use an f-string?
```py
f"{re.sub('_contra$', '', name)}_co"
```

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,64 @@ def _name_became_keyword_in_version(name, rules):
...
+                    args_to_display.append((name, "covariant", suggest_name))
+                if (
+                    kw.arg == "contravariant"
+                    and kw.value.value
+                    and not name.endswith("_contra")
+                ):
+                    suggest_name = re.sub("_co$", "", name)
+                    suggest_name += "_contra"
```
cdce8p, line 2196: Same here

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,64 @@ def _name_became_keyword_in_version(name, rules):
...
+            for kw in keywords:
+                if (
+                    kw.arg == "covariant"
+                    and kw.value.value
+                    and not name.endswith("_co")
+                ):
+                    suggest_name = re.sub("_contra$", "", name)
+                    suggest_name += "_co"
+                    args_to_display.append((name, "covariant", suggest_name))
+                if (
```
cdce8p, line 2190: I would suggest adding an empty line between these block. That should make it more readable.

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,64 @@ def _name_became_keyword_in_version(name, rules):
...
+                    kw.arg == "contravariant"
+                    and kw.value.value
+                    and not name.endswith("_contra")
+                ):
+                    suggest_name = re.sub("_co$", "", name)
+                    suggest_name += "_contra"
+                    args_to_display.append((name, "contravariant", suggest_name))
+                if kw.arg == "bound" and (
```
cdce8p, line 2198: Same here

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,64 @@ def _name_became_keyword_in_version(name, rules):
...
+                    name.endswith("_co") or name.endswith("_contra")
+                ):
+                    suggest_name = re.sub("_contra$|_co$", "", name)
+                    args_to_display.append((name, "invariant", suggest_name))
+        elif name.endswith("_co") or name.endswith("_contra"):
```
cdce8p, line 2203: Maybe even one here

pylint/checkers/base.py
```diff
@@ -2075,6 +2147,64 @@ def _name_became_keyword_in_version(name, rules):
...
+            suggest_name = re.sub("_contra$|_co$", "", name)
+            args_to_display.append((name, "invariant", suggest_name))
```
cdce8p, line 2205: With my suggestion from above, instead of creating an arbitrary list, try calling `self.add_message` directly. Same for the other ones here.

tests/functional/t/typevar_name_missing_variance.py
```diff
@@ -0,0 +1,41 @@
...
+
+# Tests for combinations with covariance
+VT = TypeVar("VT", covariant=True)  # [typevar-name-incorrect-variance]
+VT_contra = TypeVar("VT_contra", covariant=True)  # [typevar-name-incorrect-variance]
+VT_co = TypeVar("VT_co", covariant=True)
+
+# Tests for combinations with bound
+VT = TypeVar("VT", bound=True)
+VT_co = TypeVar("VT_co", bound=True)  # [typevar-name-incorrect-variance]
+VT_contra = TypeVar("VT_contra", bound=True)  # [typevar-name-incorrect-variance]
```
cdce8p, line 29:
```suggestion
VT = TypeVar("VT", bound=int)
VT_co = TypeVar("VT_co", bound=int)  # [typevar-name-incorrect-variance]
VT_contra = TypeVar("VT_contra", bound=int)  # [typevar-name-incorrect-variance]
```
`True` is not a valid type

cdce8p (changes requested):
Thanks for doing this @DanielNoord!
Initially I thought we would need to do the naming style first, but I guess your workaround with excluding TypeVars from name checks will work, too. Let's just be sure that the other one is (almost) ready too before we merge it. Just so both land in the same release.

_It might make sense to create an entirely new PR for the naming style. Rebasing it now would likely mess with the PR history. Besides it has gotten difficult to follow anyway. We can close the current one once both parts are done._

Let's focus on this one first, though.

> Pylance complains whenever you use `bound` and any of the variants together. Ar we sure they can be used together?

Are you sure? It works for me. Make sure the argument for `bound` is an actual type.
```py
from typing import TypeVar

T = TypeVar("T", bound=int, covariant=True)
```

DanielNoord:
> Thanks for doing this @DanielNoord! Initially I thought we would need to do the naming style first, but I guess your workaround with excluding TypeVars from name checks will work, too. Let's just be sure that the other one is (almost) ready too before we merge it. Just so both land in the same release.

We can make the other one a blocker for `2.13`.

> Are you sure? It works for me. Make sure the argument for `bound` is an actual type.

Yeah, I messed up there. Sadly, this required quite a big refactor.

DanielNoord:
Btw, the new iteration does not handle this:
```python
T = TypeVar("T", covariant=True, contravariant=True) # [typevar-name-incorrect-variance, typevar-name-incorrect-variance]
T_co = TypeVar("T_co", covariant=True, contravariant=True) # [typevar-name-incorrect-variance]
```

It will now be:
```python
T = TypeVar("T", covariant=True, contravariant=True) # [typevar-name-incorrect-variance]
T_co = TypeVar("T_co", covariant=True, contravariant=True)
```

However, since this creates a ``RunTime`` error I think it makes more sense to leave this as an unhandled edge case. We can create a new ``typevar-double-variance`` checker to warn about this.

cdce8p (commented):
Some last comments.

Regarding the edge case, I think it's fine. Not something a name checker should handle. _There is also another one with `covariant=1`, but fixing it would create other issues which is why I didn't mention it in my comments. That should be handled by a separate checker._

pylint/checkers/base.py
```diff
@@ -2075,6 +2130,62 @@ def _name_became_keyword_in_version(name, rules):
...
+
+        for kw in keywords:
+            if kw.arg == "covariant" and kw.value.value:
+                if not name.endswith("_co"):
+                    suggest_name = f"{re.sub('_contra$', '', name)}_co"
+                    self.add_message(
+                        "typevar-name-incorrect-variance",
+                        node=node,
+                        args=(name, "covariant", suggest_name),
+                        confidence=interfaces.HIGH,
```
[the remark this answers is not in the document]
DanielNoord: Ah, that's a mistake. Should be `INFERENCE` indeed. Changing.

tests/functional/t/typevar_name_incorrect_variance.py
```diff
@@ -0,0 +1,55 @@
+"""Test case for typevar-name-incorrect-variance."""
+
+from typing import TypeVar
+
+# Type variables without variance
+T = TypeVar("T")
+T_co = TypeVar("T_co")  # [typevar-name-incorrect-variance]
+T_contra = TypeVar("T_contra")  # [typevar-name-incorrect-variance]
+Tscores_contra = TypeVar("Tscores_contra")  # [typevar-name-incorrect-variance]
```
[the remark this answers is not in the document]
DanielNoord: I was unsure. Will change

cdce8p (approved):
Ok. If we work on the other ones next, I guess we can merge this one. There is still some time, considering that we need a new astroid release first.

PR #3473, opened 2020-04-03
benchmark| Potential solution for performance regressions

doublethefish:
## Steps

- [x] (na) Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] (na) Add a ChangeLog entry describing what your PR does.
- [x] (na) If it's a new feature or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Description
These commits generate comparable performance data for a run of benchmarked code.

This is a potential 1st step in establishing performance regression tests. Later work would be to actually do the comparison against the target-branch somehow.

We use pytest-benchmark here, rather than asv, because pytest-benchmark was easier to integrate into the existing tests.

### Use-case
A use-case for this functionality is:

1. Write a good benchmark test:
  - See  `tests/benchmark/test_baseline_benchmarks.py` for an example
2. Run `tox -e benchmark`
  - This generates benchmark .json files in the .tox directory
3. Apply some change that you suspect makes the code faster
4. Re-run `tox -e benchmark`
  - This generates another benchmark .json file, probably `00002-something.json`
5. Compare stats:
  - In `.tox/` run `py.test-benchmark compare 0001 0002` - the numbers are used as a glob to find files that match, use the numbers that refer to the `tox -e benchmark` runs you want to compare.
6. Look for statistically significant differences. You will need to use your common sense here.
7. Profit.

### Baseline benchmark for pylint
The baseline benchmark in this PR attempts to benchmark basic runs:
- with no-files
- a single empty file
- lots of empty files
- a single empty checker
- many empty checkers
- a checker that mimics doing expensive work
- across workers and a single-processed

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :hammer: Refactoring (support of)  |
| ✓  | :scroll: Docs (new performance metrics) |

PR #5827, opened 2022-02-22
Don't assume runners have more than 2 cores available for benchmarking

DanielNoord:

## Type of Changes

## Description

Based on testing in https://github.com/DanielNoord/pylint/pull/87.
I missed these in #5816 as tests were not failing on these initially, but know they are. It is logical as they make the same incorrect assumption about test runners as the others were.

_I do wonder why we even have these tests. It's not as if we're doing anything with them 😅 _

### DIFF
--- a/tests/benchmark/test_baseline_benchmarks.py
+++ b/tests/benchmark/test_baseline_benchmarks.py
@@ -142,37 +142,37 @@ class TestEstablishBaselineBenchmarks:
             linter.msg_status == 0
         ), f"Expected no errors to be thrown: {pprint.pformat(linter.reporter.messages)}"
 
-    def test_baseline_benchmark_j10(self, benchmark):
+    def test_baseline_benchmark_j2(self, benchmark):
         """Establish a baseline of pylint performance with no work across threads.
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Change itself looks good. Those benchmark were introduced in one of those : https://github.com/PyCQA/pylint/pulls?q=is%3Apr+is%3Aclosed+author%3Adoublethefish, probably #3473

DanielNoord:
> Yes, it's just a bunch of wasted runner time right now. They were added to follow performance at some point but we're not using the benchmarks right now.

> Change itself looks good. Those benchmark were introduced in one of those : https://github.com/PyCQA/pylint/pulls?q=is%3Apr+is%3Aclosed+author%3Adoublethefish, probably #3473

Looking at that PR I don't even understand why this is part of the test suite in the first place. They seem like instructions to generate benchmark data, but that should probably be included in the development guidelines and as an additional script. If I understand their purpose correctly there is (currently) no need to ever run them in CI as they are never getting compared to something.

Pierre-Sassoulas:
> If I understand their purpose correctly there is (currently) no need to ever run them in CI as they are never getting compared to something.

Yes, they could, potentially, but this is not what is happening right now.

PR #5841, opened 2022-02-26
Add some flags showing information to the CI pytest runs

DanielNoord:

## Type of Changes

## Description

This should help narrow down issues with the CI if they keep popping up.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -224,7 +224,7 @@ jobs:
       - name: Run pytest
         run: |
           . venv/bin/activate
-          pytest --benchmark-disable --cov --cov-report= tests/
+          pytest -vv --durations=20 --benchmark-disable --cov --cov-report= tests/
       - name: Upload coverage artifact
         uses: actions/upload-artifact@v2.3.1
         with:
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
👌

Pierre-Sassoulas:
Look like the verbose option is not displayed when the timeout happens : https://github.com/PyCQA/pylint/runs/5344145948?check_suite_focus=true as the operation is canceled. We have strange spike on the coverage job sometime, might be due to coveralls and not github actions ? Should we separate 3.8 and increase the timeout ? Just increase the timeout ?

DanielNoord:
It is showing, see L19 of the action. We're just failing on the second test.
I have done some testing again myself and there seem to be issue with all test that use 2 or more cores on Linux. I'm really starting to wonder if something is up with the runners or if Python jobs can only use 1 core.. The support for GitHub Actions is a bit dispersed though, so I haven't found the right place to ask questions/support about this..

Pierre-Sassoulas:
> It is showing, see L19 of the action.

Ho wow, you're right, it's stuck very early !

PR #5843, opened 2022-02-26
Restore the useful part of the python3 checker

### LINKED ISSUE #5025: eq-without-hash check removed
### Bug description

```python
class A:
    def __init__(self) -> None:
        self.x = 5

    def __eq__(self, other: object) -> bool:
        return isinstance(other, A) and other.x == self.x
```

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
empty
```

### Expected behavior

Formerly `eq-without-hash` error was issued, but not more since pylint 2.11.

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.0
[GCC 9.4.0]
```

### OS / Environment

Ubuntu 20.04.3 LTS \n \l

### Additional dependencies

_No response_

Pierre-Sassoulas:
## Type of Changes

## Description

~~This make a check that was optional, non-optional.~~

Closes #5025


pylint/checkers/python3.py
```diff
@@ -0,0 +1,40 @@
...
+
+from pylint import checkers, interfaces
+from pylint.checkers import utils
+from pylint.lint import PyLinter
+
+
+class Python3Checker(checkers.BaseChecker):
+
+    __implements__ = interfaces.IAstroidChecker
+    enabled = True
```
Pierre-Sassoulas, line 20:
```suggestion
    enabled = False
```
If we want to make this check optional we could do this.

Pierre-Sassoulas:
@jolaf FYI :)

jolaf:
@Pierre-Sassoulas Thanks!

DanielNoord:
We should probably also rename `Python3Checker`? Perhaps move this checker into the `extensions` and rename it?

Pierre-Sassoulas:
> We should probably also rename Python3Checker?

I think not touching it as the benefit of clarity for those that used the python3 checker before. How would you rename it ? I'm afraid if more checks are brought back and we named it ``eq_without_hash`` then the name do not make sense anymore with another check inside it. We could put the check in the codestyle checker though?

DanielNoord:
> > We should probably also rename Python3Checker?
> 
> I think not touching it as the benefit of clarity for those that used the python3 checker before. How would you rename it ? I'm afraid if more checks are brought back and we named it `eq_without_hash` then the name do not make sense anymore with another check inside it. We could put the check in the codestyle checker though?

There have been multiple checks of the previous checker in #5025 and nobody found any useful checkers except this one so I think we'll be fine. Either way, how many people will be looking for the specific line `class Python3Checker` compared to the number of future people that will wonder why this class is called `Python3Checker`.

I think it makes sense to just make this an extension and name it as such `EqualWithoutHashChecker`.

pylint/constants.py
```diff
@@ -80,8 +80,9 @@ class DeletedMessage(NamedTuple):
     old_names: List[Tuple[str, str]] = []
 
 
-DELETED_MSGID_PREFIXES = [
-    16,  # the PY3K+ checker, see https://github.com/PyCQA/pylint/pull/4942
+DELETED_MSGID_PREFIXES: List[int] = [
```
DanielNoord, line 83: I don't think we need the comment here. What does this tell us a year from now? Just instantiating an empty list and being able to add later to it is fine for now I think.

pylint/extensions/code_style.py
```diff
@@ -28,7 +28,7 @@ class CodeStyleChecker(BaseChecker):
        i.e. detect a common issue or improve performance
        => it should probably be part of the core checker classes
     2. Is it something that would improve code consistency,
-       maybe because it's slightly better with regards to performance
+       maybe because it's slightly better with regard to performance
```
DanielNoord, line 31: I think it should be `regards` actually, but I'm doubting myself.

pylint/extensions/eq_without_hash.py
```diff
@@ -0,0 +1,39 @@
...
+            "Used when a class implements __eq__ but not __hash__. Objects get "
+            "None as their default __hash__ implementation if they also implement __eq__.",
+        ),
+    }
+
+    @utils.check_messages("eq-without-hash")
+    def visit_classdef(self, node: nodes.ClassDef) -> None:
+        locals_and_methods = set(node.locals).union(x.name for x in node.mymethods())
+        if "__eq__" in locals_and_methods and "__hash__" not in locals_and_methods:
+            self.add_message("eq-without-hash", node=node)
```
DanielNoord, line 35: Do we want to add a confidence while we're at it?
Does `mymethods()` use inference? If not I think this should be `HIGH`.

tests/functional/ext/eq_without_hash/eq_without_hash.py
```diff
@@ -0,0 +1,11 @@
+"""Regression test for #5025"""
+
+# pylint: disable=invalid-name,missing-docstring, too-few-public-methods
```
DanielNoord, line 3:
```suggestion
# pylint: disable=invalid-name, missing-docstring, too-few-public-methods
```

DanielNoord (changes requested):
Last couple of questions!

pylint/extensions/code_style.py
jacobtylerwalls: from google: regard is the original "correct" one but regards is catching up quickly, so not a big deal either way.

pylint/extensions/eq_without_hash.py
Pierre-Sassoulas: It's only yielding FunctionDef:
```python
    def mymethods(self):
        """Iterate over all of the method defined in this class only.

        :returns: The methods defined on the class.
        :rtype: iterable(FunctionDef)
        """
        for member in self.values():
            if isinstance(member, FunctionDef):
                yield member
```
```suggestion
            self.add_message("eq-without-hash", node=node, confidence=interfaces.HIGH)
```

pylint/constants.py
```diff
@@ -80,8 +80,9 @@ class DeletedMessage(NamedTuple):
...
+    # The PY3K+ checker was deleted see https://github.com/PyCQA/pylint/pull/4942
+    # And then reinstated see https://github.com/PyCQA/pylint/issues/5025
 ]
```
Pierre-Sassoulas, line 86:
```suggestion
DELETED_MSGID_PREFIXES: List[int] = []
```

PR #5847, opened 2022-02-27
Enum: Add support for self-defined members

### LINKED ISSUE #5138: Support Enum with self defined members
### Current problem

False positive E1101 are reported for self defined member in Enums. 

Example:
- [Enum `Layout` w/ self defined members `col` & `row`](https://github.com/OldIMP/volksschule/blob/ba0dc09ad4590439d5ae8c3c7b7639b45869db70/r_quiz.py#L139)
- [Here](https://github.com/OldIMP/volksschule/blob/ba0dc09ad4590439d5ae8c3c7b7639b45869db70/r_quiz.py#L174) will then be reported falsely as E1101:
  > r_quiz.py:174:12: E1101: Instance of 'SIMPLE' has no '`row`' member (no-member)
  > r_quiz.py:174:12: E1101: Instance of 'SCHRIFTLICH' has no '`row`' member (no-member)
  > r_quiz.py:174:25: E1101: Instance of 'SIMPLE' has no '`col`' member (no-member)
  > r_quiz.py:174:25: E1101: Instance of 'SCHRIFTLICH' has no '`col`' member (no-member)

#### pylint --version
pylint 2.11.1
astroid 2.8.2
Python 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:15:42) [MSC v.1916 64 bit (AMD64)]

### Desired solution

According to [Python doc](https://docs.python.org/3/library/enum.html#planet), self defined members in Enums are allowed. Therefore no issues should be reported for the example above. 

### Additional context

_No response_

huwcbjones:
## Type of Changes

## Description
As per title, this PR adds an enhancement to the typechecker to find _simple_ self-defined members

This change **does not** support:
* conditionally defining members (it will assume all members are valid even if the validity is conditional)
* defining members via a function call in `__new__` or `__init__`

Closes #5138

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -249,6 +249,10 @@ Release date: TBA
 
   Closes #5219
 
+* Fixed false positive ``no-member`` for Enums with self-defined members.
+
+  Closes #5138
+
 * Fix false negative for ``no-member`` when attempting to assign an instance
   attribute to itself without any prior assignment.
 
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -260,6 +260,10 @@ Other Changes
 
   Closes #5219
 
+* Fixed false positive ``no-member`` for Enums with self-defined members.
+
+  Closes #5138
+
 * Fix false negative for ``no-member`` when attempting to assign an instance
   attribute to itself without any prior assignment.
 
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -526,22 +526,22 @@ def _emit_no_member(
 
 
 def _get_all_attribute_assignments(
-    node: astroid.FunctionDef, name: str | None = None
+    node: nodes.FunctionDef, name: str | None = None
 ) -> set[str]:
     attributes: set[str] = set()
-    for child in node.nodes_of_class((astroid.Assign, astroid.AnnAssign)):
+    for child in node.nodes_of_class((nodes.Assign, nodes.AnnAssign)):
         targets = []
-        if isinstance(child, astroid.Assign):
+        if isinstance(child, nodes.Assign):
             targets = child.targets
-        elif isinstance(child, astroid.AnnAssign):
+        elif isinstance(child, nodes.AnnAssign):
             targets = [child.target]
         for assign_target in targets:
-            if isinstance(assign_target, astroid.Tuple):
+            if isinstance(assign_target, nodes.Tuple):
                 targets.extend(assign_target.elts)
                 continue
             if (
-                isinstance(assign_target, astroid.AssignAttr)
-                and isinstance(assign_target.expr, astroid.Name)
+                isinstance(assign_target, nodes.AssignAttr)
+                and isinstance(assign_target.expr, nodes.Name)
                 and (name is None or assign_target.expr.name == name)
             ):
                 attributes.add(assign_target.attrname)
@@ -556,16 +556,13 @@ def _enum_has_attribute(
             (b.parent for b in owner.bases if isinstance(b.parent, nodes.ClassDef)),
             None,
         )
-        if enum_def is None:
-            return False
+        assert enum_def is not None, "enum_def unexpectedly None"
     else:
         enum_def = owner
 
     # Find __new__ and __init__
     dunder_new = next((m for m in enum_def.methods() if m.name == "__new__"), None)
     dunder_init = next((m for m in enum_def.methods() if m.name == "__init__"), None)
-    if dunder_new is None and dunder_init is None:
-        return False
 
     enum_attributes: set[str] = set()
 
@@ -573,11 +570,7 @@ def _enum_has_attribute(
     if dunder_new:
         # Get the object returned in __new__
         returned_obj_name = next(
-            (
-                c.value
-                for c in dunder_new.get_children()
-                if isinstance(c, astroid.Return)
-            ),
+            (c.value for c in dunder_new.get_children() if isinstance(c, nodes.Return)),
             None,
         )
         if returned_obj_name is not None:
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -525,6 +525,21 @@ def _emit_no_member(
     return True
 
 
+def _get_all_attribute_assignments(
+    node: astroid.FunctionDef, name: str | None = None
+) -> set[str]:
+    attributes = set()
+    for child in node.nodes_of_class(astroid.Assign):
+        for assign_target in child.targets:
+            if not isinstance(assign_target, astroid.AssignAttr):
+                continue
+            if not isinstance(assign_target.expr, astroid.Name):
+                continue
+            if name is None or assign_target.expr.name == name:
+                attributes.add(assign_target.attrname)
+    return attributes
+
+
 def _enum_has_attribute(
     owner: Union[astroid.Instance, nodes.ClassDef], node: nodes.Attribute
 ) -> bool:
@@ -533,31 +548,38 @@ def _enum_has_attribute(
     while not isinstance(enum_def, astroid.ClassDef):
         enum_def = enum_def.parent
 
-    # Find __new__
+    # Find __new__ and __init__
     dunder_new = next((m for m in enum_def.methods() if m.name == "__new__"), None)
-    if dunder_new is None:
-        return False
-
-    # Get the returned object
-    returned_obj_name = next(
-        (c.value for c in dunder_new.get_children() if isinstance(c, astroid.Return)),
-        None,
-    )
-    if returned_obj_name is None:
+    dunder_init = next((m for m in enum_def.methods() if m.name == "__init__"), None)
+    if dunder_new is None and dunder_init is None:
         return False
 
     enum_attributes = set()
-    # Iterate over children and find Enum attr assignments
-    for child in dunder_new.get_children():
-        if not isinstance(child, astroid.Assign):
-            continue
-        for assign_target in child.targets:
-            if not isinstance(assign_target, astroid.AssignAttr):
-                continue
-            if not isinstance(assign_target.expr, astroid.Name):
-                continue
-            if assign_target.expr.name == returned_obj_name.name:
-                enum_attributes.add(assign_target.attrname)
+
+    # Find attributes defined in __new__
+    if dunder_new:
+        # Get the object returned in __new__
+        returned_obj_name = next(
+            (
+                c.value
+                for c in dunder_new.get_children()
+                if isinstance(c, astroid.Return)
+            ),
+            None,
+        )
+        if returned_obj_name is not None:
+            # Find all attribute assignments to the returned object
+            enum_attributes |= _get_all_attribute_assignments(
+                dunder_new, returned_obj_name.name
+            )
+
+    # Find attributes defined in __init__
+    if dunder_init and dunder_init.body and dunder_init.args:
+        # Grab the name referring to `self` from the function def
+        enum_attributes |= _get_all_attribute_assignments(
+            dunder_init, dunder_init.args.arguments[0].name
+        )
+
     return node.attrname in enum_attributes
 
 
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -451,6 +451,8 @@ def _emit_no_member(
                 return False
             if metaclass:
                 # Renamed in Python 3.10 to `EnumType`
+                if _enum_has_attribute(owner, node):
+                    return False
                 return metaclass.qname() in {"enum.EnumMeta", "enum.EnumType"}
             return False
         if not has_known_bases(owner):
@@ -521,6 +523,40 @@ def _emit_no_member(
         node_origin, parent = parent, parent.parent
 
     return True
+
+
+def _enum_has_attribute(owner, node) -> bool:
+    enum_def = owner
+    # Traverse the AST to find the Enum ClassDef
+    while not isinstance(enum_def, astroid.ClassDef):
+        enum_def = enum_def.parent
+
+    # Find __new__
+    dunder_new = next((m for m in enum_def.methods() if m.name == "__new__"), None)
+    if dunder_new is None:
+        return False
+
+    # Get the returned object
+    returned_obj_name = next(
+        (c.value for c in dunder_new.get_children() if isinstance(c, astroid.Return)),
+        None,
+    )
+    if returned_obj_name is None:
+        return False
+
+    enum_attributes = set()
+    # Iterate over children and find Enum attr assignments
+    for child in dunder_new.get_children():
+        if not isinstance(child, astroid.Assign):
+            continue
+        for assign_target in child.targets:
+            if not isinstance(assign_target, astroid.AssignAttr):
+                continue
+            if not isinstance(assign_target.expr, astroid.Name):
+                continue
+            if assign_target.expr.name == returned_obj_name.name:
+                enum_attributes.add(assign_target.attrname)
+    return node.attrname in enum_attributes
 
 
 def _determine_callable(
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -528,7 +528,7 @@ def _emit_no_member(
 def _get_all_attribute_assignments(
     node: astroid.FunctionDef, name: str | None = None
 ) -> set[str]:
-    attributes = set()
+    attributes: set[str] = set()
     for child in node.nodes_of_class((astroid.Assign, astroid.AnnAssign)):
         targets = []
         if isinstance(child, astroid.Assign):
@@ -536,11 +536,14 @@ def _get_all_attribute_assignments(
         elif isinstance(child, astroid.AnnAssign):
             targets = [child.target]
         for assign_target in targets:
-            if not isinstance(assign_target, astroid.AssignAttr):
+            if isinstance(assign_target, astroid.Tuple):
+                targets.extend(assign_target.elts)
                 continue
-            if not isinstance(assign_target.expr, astroid.Name):
-                continue
-            if name is None or assign_target.expr.name == name:
+            if (
+                isinstance(assign_target, astroid.AssignAttr)
+                and isinstance(assign_target.expr, astroid.Name)
+                and (name is None or assign_target.expr.name == name)
+            ):
                 attributes.add(assign_target.attrname)
     return attributes
 
@@ -557,9 +560,6 @@ def _enum_has_attribute(
             return False
     else:
         enum_def = owner
-    # Traverse the AST to find the Enum ClassDef
-    while not isinstance(enum_def, astroid.ClassDef):
-        enum_def = enum_def.parent
 
     # Find __new__ and __init__
     dunder_new = next((m for m in enum_def.methods() if m.name == "__new__"), None)
@@ -567,7 +567,7 @@ def _enum_has_attribute(
     if dunder_new is None and dunder_init is None:
         return False
 
-    enum_attributes = set()
+    enum_attributes: set[str] = set()
 
     # Find attributes defined in __new__
     if dunder_new:
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -543,7 +543,15 @@ def _get_all_attribute_assignments(
 def _enum_has_attribute(
     owner: Union[astroid.Instance, nodes.ClassDef], node: nodes.Attribute
 ) -> bool:
-    enum_def = owner
+    if isinstance(owner, astroid.Instance):
+        enum_def = next(
+            (b.parent for b in owner.bases if isinstance(b.parent, nodes.ClassDef)),
+            None,
+        )
+        if enum_def is None:
+            return False
+    else:
+        enum_def = owner
     # Traverse the AST to find the Enum ClassDef
     while not isinstance(enum_def, astroid.ClassDef):
         enum_def = enum_def.parent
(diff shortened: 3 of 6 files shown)


pylint/checkers/typecheck.py
```diff
@@ -507,6 +507,8 @@ def _emit_no_member(
                 return False
             if metaclass:
                 # Renamed in Python 3.10 to `EnumType`
+                if _enum_has_attribute(owner, node):
+                    return False
```
Pierre-Sassoulas, line 511:
```suggestion
                if _enum_has_attribute(owner, node):
                    return False
                # Renamed in Python 3.10 to `EnumType`
```
I think the comment we related to line 512.

tests/functional/e/enum_self_defined_member_5138.py
```diff
@@ -0,0 +1,22 @@
+# [missing-module-docstring]
```
Pierre-Sassoulas, line 1:
```suggestion
# pylint: disable=missing-docstring
```

tests/functional/e/enum_self_defined_member_5138.py
```diff
@@ -0,0 +1,22 @@
...
+from enum import IntEnum
+
+
+class Day(IntEnum):  # [missing-class-docstring]
```
Pierre-Sassoulas, line 5:
```suggestion
class Day(IntEnum):
```

tests/functional/e/enum_self_defined_member_5138.txt
```diff
@@ -0,0 +1,3 @@
+missing-module-docstring:1:0:None:None::Missing module docstring:HIGH
+missing-class-docstring:5:0:18:18:Day:Missing class docstring:HIGH
+no-member:22:6:22:20::Instance of 'FRIDAY' has no 'foo' member:INFERENCE
```
Pierre-Sassoulas, line 3:
```suggestion
no-member:22:6:22:20::Instance of 'FRIDAY' has no 'foo' member:INFERENCE
```

pylint/checkers/typecheck.py
```diff
@@ -579,6 +581,40 @@ def _emit_no_member(
     return True
 
 
+def _enum_has_attribute(owner, node) -> bool:
```
Pierre-Sassoulas, line 584:
```suggestion
def _enum_has_attribute(owner: Union[astroid.Instance, nodes.ClassDef], node: nodes.Attribute) -> bool:
```
As far as I know this seems to be the typing here.

pylint/checkers/typecheck.py
```diff
@@ -579,6 +581,40 @@ def _emit_no_member(
...
+        None,
+    )
+    if returned_obj_name is None:
+        return False
+
+    enum_attributes = set()
+    # Iterate over children and find Enum attr assignments
+    for child in dunder_new.get_children():
+        if not isinstance(child, astroid.Assign):
+            continue
```
Pierre-Sassoulas, line 607:
```suggestion
    for child in dunder_new.nodes_of_class(astroid.Assign):
```
Maybe something to consider ? This is not equivalent as it's also searching in children of children.

Pierre-Sassoulas (changes requested):
Thank you for opening the PR @huwcbjones. This looks great already. I think this is going to go in 2.14, we're trying to release 2.13.

pylint/checkers/typecheck.py
huwcbjones: Ta

pylint/checkers/typecheck.py
huwcbjones: Thanks, didn't know this method existed!
I think it _is_ better to check the children as it would catch members defined anywhere in the function def tree.
If someone were to define a method within `__init__` or `__new__` then the checker could do the wrong thing. However, my response would be "don't do that" as it's kinda nasty.

pylint/checkers/typecheck.py
```diff
@@ -51,6 +51,7 @@
 # Copyright (c) 2021 Andrew Haigh <nelfin@gmail.com>
 # Copyright (c) 2021 Jens H. Nielsen <Jens.Nielsen@microsoft.com>
 # Copyright (c) 2021 Ikraduya Edian <ikraduya@gmail.com>
+# Copyright (c) 2022 Huw Jones <huw@huwcbjones.co.uk>
```
Pierre-Sassoulas, line 54: Don't worry about that it's automated at release time :)

pylint/checkers/typecheck.py
```diff
@@ -579,6 +582,72 @@ def _emit_no_member(
     return True
 
 
+def _get_all_attribute_assignments(
+    node: astroid.FunctionDef, name: Optional[str] = None
+) -> Set[str]:
+    attributes = set()
+    for child in node.nodes_of_class(astroid.Assign):
```
mbyrnepr2, line 589: There could be a false positive if an instance variable is annotated in the `__init__` method, for example:
```python
def __init__(unit):
    self.unit: str = unit
```

huwcbjones: Thanks for catching that, have pushed a fix :)

huwcbjones:
Apologies, made a mess of the rebase wrt to imports/pylint

pylint/checkers/typecheck.py
mbyrnepr2: Nice!

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
...
+                continue
+            if not isinstance(assign_target.expr, astroid.Name):
+                continue
+            if name is None or assign_target.expr.name == name:
+                attributes.add(assign_target.attrname)
+    return attributes
+
+
+def _enum_has_attribute(
+    owner: astroid.Instance | nodes.ClassDef, node: nodes.Attribute
```
DanielNoord, line 549: We should probably import this from `objects` instead of `astroid`.

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
...
+            (b.parent for b in owner.bases if isinstance(b.parent, nodes.ClassDef)),
+            None,
+        )
+        if enum_def is None:
+            return False
+    else:
+        enum_def = owner
+    # Traverse the AST to find the Enum ClassDef
+    while not isinstance(enum_def, astroid.ClassDef):
+        enum_def = enum_def.parent
```
DanielNoord, line 562: This is unnecessary right? It is also uncovered.

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
...
+
+    # Find __new__ and __init__
+    dunder_new = next((m for m in enum_def.methods() if m.name == "__new__"), None)
+    dunder_init = next((m for m in enum_def.methods() if m.name == "__init__"), None)
+    if dunder_new is None and dunder_init is None:
+        return False
```
DanielNoord, line 568: The last line is uncovered. I also wonder if we can do this without iterating over all methods. Doesn't `getattr` work?

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
...
+
+    enum_attributes = set()
```
DanielNoord, line 570:
```suggestion
    enum_attributes: set[...] = set()
```

Could you add typing here?

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
     return True
 
 
+def _get_all_attribute_assignments(
+    node: astroid.FunctionDef, name: str | None = None
+) -> set[str]:
+    attributes = set()
```
DanielNoord, line 531:
```suggestion
    attributes: set[str] = set()
```

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
...
+        targets = []
+        if isinstance(child, astroid.Assign):
+            targets = child.targets
+        elif isinstance(child, astroid.AnnAssign):
+            targets = [child.target]
+        for assign_target in targets:
+            if not isinstance(assign_target, astroid.AssignAttr):
+                continue
+            if not isinstance(assign_target.expr, astroid.Name):
+                continue
```
DanielNoord, line 542: This is uncovered. Can we cover it?

tests/functional/e/enum_self_defined_member_5138.py
```diff
@@ -0,0 +1,45 @@
+# pylint: disable=missing-docstring
```
DanielNoord, line 1:
```suggestion
"""Tests for https://github.com/PyCQA/pylint/pull/5847."""

# pylint: disable=missing-docstring
```

DanielNoord (changes requested):
Thanks @huwcbjones! Let me know if you have any questions!

pylint/checkers/typecheck.py
huwcbjones: Sorry, don't understand what you're talking about.

Assuming you're referencing the type hints (albeit unsure which exact one), I copied the existing style in the file:
* `Instance`: mostly imported from `astroid`
* `ClassDef`/`Attribute`: mostly imported from `nodes`

DanielNoord: Ah okay, then the style on the rest of the file is messed up sadly. `Instance` comes from `astroid.objects` but is also imported in the ``__init__`` of `astroid`.

Let's keep as is then!

ChangeLog
```diff
@@ -238,9 +238,16 @@ Release date: TBA
...
+* Fixed false positive ``no-member`` for Enums with self-defined members.
+
+  Closes #5138
+
+* Fix false positive for ``unsubscriptable-object`` in Python 3.8 and below for
+  statements guarded by ``if TYPE_CHECKING``.
 
   Closes #5219
+  
+* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.
```
Pierre-Sassoulas, line 250:
```suggestion
```

ChangeLog
```diff
@@ -238,9 +238,16 @@ Release date: TBA
 
   Closes #4525
 
-* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.
+* Fixed false positive ``no-member`` for Enums with self-defined members.
+
+  Closes #5138
+
+* Fix false positive for ``unsubscriptable-object`` in Python 3.8 and below for
+  statements guarded by ``if TYPE_CHECKING``.
```
Pierre-Sassoulas, line 246:
```suggestion
```

doc/whatsnew/2.14.rst
```diff
@@ -249,10 +249,17 @@ Other Changes
 
   Closes #5205
 
-* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.
+* Fixed false positive ``no-member`` for Enums with self-defined members.
+
+  Closes #5138
+
+* Fix false positive for ``unsubscriptable-object`` in Python 3.8 and below for
+  statements guarded by ``if TYPE_CHECKING``.
```
Pierre-Sassoulas, line 257:
```suggestion
```

doc/whatsnew/2.14.rst
```diff
@@ -249,10 +249,17 @@ Other Changes
...
 
   Closes #5219
 
+* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.
+
```
Pierre-Sassoulas, line 262:
```suggestion
```

ChangeLog
```diff
@@ -238,10 +238,17 @@ Release date: TBA
...
 
   Closes #5219
 
+* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.
+
```
Pierre-Sassoulas, line 251:
```suggestion
```

ChangeLog
```diff
@@ -238,7 +238,9 @@ Release date: TBA
 
   Closes #4525
 
-* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.
+* Fixed false positive ``no-member`` for Enums with self-defined members.
+
+  Closes #5138
```
Pierre-Sassoulas, line 243:
```suggestion
* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.

* Fixed false positive ``no-member`` for Enums with self-defined members.

  Closes #5138
```

doc/whatsnew/2.14.rst
```diff
@@ -249,7 +249,10 @@ Other Changes
 
   Closes #5205
 
-* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.
+* Fixed false positive ``no-member`` for Enums with self-defined members.
+
+  Closes #5138
+
```
Pierre-Sassoulas, line 255:
```suggestion
* Fix falsely issuing ``useless-suppression`` on the ``wrong-import-position`` checker.

* Fixed false positive ``no-member`` for Enums with self-defined members.

  Closes #5138

```

tests/functional/e/enum_self_defined_member_5138.py
huwcbjones: I think this is superfluous because the file name contains the issue number and the purpose of the tests, but I'll shove some English in a docsttring

pylint/checkers/typecheck.py
huwcbjones: Astroid docs says that [AssignAttr.expr is  `Optional[NodeNG]`](https://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.AssignAttr.expr), hence the check.
This was more a "defensive coding" thing because it could happen in the wild but I don't know how to trigger it.

pylint/checkers/typecheck.py
huwcbjones: yeah, forgot the type checker isn't clever enough to pick that up

tests/functional/e/enum_self_defined_member_5138.py
DanielNoord: Somewhat agree, but having a direct link is always easier to revisit an issue imo. Especially considering the age of the project. There are multiple references to "issue xx" which don't correspond to a Github issue and then be issues on `gitlab` or any of the previous hosting servers of the project. That's why I like explicit links if possible.

pylint/checkers/typecheck.py
DanielNoord: Let's test this on L546. I highly suspect this is due to incorrect typing in `astroid` (of which there is a lot for node attributes), so it is likely we will never be able to cover this.

huwcbjones: Sure thing

pylint/checkers/typecheck.py
huwcbjones: With getattr, I came up with something like this
```py
    dunder_new: astroid.FunctionDef | None = (enum_def.getattr("__new__") or [None])[0]
    dunder_init: astroid.FunctionDef | None = (enum_def.getattr("__init__") or [None])[0]
```

In practice, I'd expect there to always be an `__init__` or `__new__` on an enum, so we can probably drop the if None check (and the or `[None]` too)

huwcbjones: With getattr, the code could now explode with an uncaught `AttributeInferenceError` if someone were to write some truly awful code.

DanielNoord: Yeah I think that should be fine. Especially since we will always end up at `object`.

huwcbjones: The current way I've implemented it makes me feel more comfortable from a known types PoV

DanielNoord: Hmm okay. I'll keep this on my mental todo list. Iterating over all methods instead of doing something like `get()` seems like we could optimise this. Thanks for trying the other solutions, much appreciated 😄

huwcbjones: nw :)
And I agree that somelike like `.get(name, default=_, type=_)` would be quite nice, if no default is provided explode, if a type is provided return the type you expect, etc

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
     return True
 
 
+def _get_all_attribute_assignments(
+    node: astroid.FunctionDef, name: str | None = None
```
DanielNoord, line 529: Sorry I just noticed, we seem to be mixing `astroid` and `nodes` in this function. See L552 for example where we do `nodes.ClassDef`. Sorry for asking so late in the process, but could you get all nodes from `nodes.`? That's how we normally do this. I could also do this for you if you want to!

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
...
+def _enum_has_attribute(
+    owner: astroid.Instance | nodes.ClassDef, node: nodes.Attribute
+) -> bool:
+    if isinstance(owner, astroid.Instance):
+        enum_def = next(
+            (b.parent for b in owner.bases if isinstance(b.parent, nodes.ClassDef)),
+            None,
+        )
+        if enum_def is None:
+            return False
```
DanielNoord, line 560: Let's do:
```suggestion
        assert enum_def
```

I think this will never be `None` right?

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,77 @@ def _emit_no_member(
...
+    else:
+        enum_def = owner
+
+    # Find __new__ and __init__
+    dunder_new = next((m for m in enum_def.methods() if m.name == "__new__"), None)
+    dunder_init = next((m for m in enum_def.methods() if m.name == "__init__"), None)
+    if dunder_new is None and dunder_init is None:
+        return False
```
DanielNoord, line 568:
```suggestion
```

This `if` is captured by the next two if-blocks and as discussed it is likely this will never be the case anyway.

Pierre-Sassoulas (approved):
LGTM, thank you @huwcbjones !

DanielNoord (changes requested):
Some final remarks! Sorry for the first comment, should have seen that earlier.. 😅

pylint/checkers/typecheck.py
huwcbjones: nw, better to finish this off than starting the XML processing that's next in my queue 😛

DanielNoord:
@huwcbjones Seems like we have actually found an issue with the assertion in the primer. Are you able to read the test output? That should probably allow us to create a test that fixes the previously missing coverage 🎉

huwcbjones:
> @huwcbjones Seems like we have actually found an issue with the assertion in the primer. Are you able to read the test output? That should probably allow us to create a test that fixes the previously missing coverage 🎉

ha, it appears that we tried to infer the the Enum definition of Enum and crashed doing so 😅 

```
astroid.exceptions.AttributeInferenceError: '_name_' not found on <ClassDef.Enum l.625 at 0x7f547d674bb0>.
AssertionError: enum_def unexpectedly None
```

pylint/checkers/typecheck.py
```diff
@@ -556,6 +556,15 @@ def _enum_has_attribute(
             (b.parent for b in owner.bases if isinstance(b.parent, nodes.ClassDef)),
             None,
         )
+
+        if enum_def is None:
+            # We don't inherit from anything, so try to find the parent
+            # class definition and roll with that
+            enum_def = node
+            while enum_def is not None and not isinstance(enum_def, astroid.ClassDef):
+                enum_def = enum_def.parent
```
huwcbjones, line 565: @DanielNoord how's this look?
It feels like a bit of a bodge, but makes the stdlib primer not crash on `enum.py`

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,79 @@ def _emit_no_member(
...
+        enum_def = next(
+            (b.parent for b in owner.bases if isinstance(b.parent, nodes.ClassDef)),
+            None,
+        )
+
+        if enum_def is None:
+            # We don't inherit from anything, so try to find the parent
+            # class definition and roll with that
+            enum_def = node
+            while enum_def is not None and not isinstance(enum_def, astroid.ClassDef):
```
DanielNoord, line 564:
```suggestion
            while not isinstance(enum_def, nodes.ClassDef):
```

DanielNoord: This should also work right? And I added `nodes.`.

DanielNoord (approved):
One final change. Yeah I think this works, although I'll let the `primer` decide that.

@Pierre-Sassoulas Luckily this will be in the `beta` as well, so we might catch additional early errors 😄

pylint/checkers/typecheck.py
huwcbjones: From a correctness PoV, we need the `not None` check because if there is no classdef and we hit the module node, then `.parent` will be `None` and we'll crash out with a "`NoneType` doesn't have attribute parent" exception. However in reality that "shouldn't happen™" because a class has to have a definition node somewhere.

And thanks, I need to programme my brain to use `nodes` instead of `astroid` 😬

DanielNoord: Oh of course.. I mixed up my `not None` and `is not`'s... We should probably keep this in then 😅

pylint/checkers/typecheck.py
```diff
@@ -523,6 +525,79 @@ def _emit_no_member(
...
+        enum_def = next(
+            (b.parent for b in owner.bases if isinstance(b.parent, nodes.ClassDef)),
+            None,
+        )
+
+        if enum_def is None:
+            # We don't inherit from anything, so try to find the parent
+            # class definition and roll with that
+            enum_def = node
+            while not isinstance(enum_def, nodes.ClassDef):
```
DanielNoord, line 564:
```suggestion
            while enum_def is not None and not isinstance(enum_def, nodes.ClassDef):
```

DanielNoord:
Thanks @huwcbjones for your work! And congratulations for becoming a `pylint` contributor! 🎉

PR #5853, opened 2022-03-01
Also test on PyPy 3.7

DanielNoord:

## Type of Changes

## Description

See https://github.com/PyCQA/pylint/pull/5849#discussion_r816766177.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -417,7 +417,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        python-version: ["pypy-3.6"]
+        python-version: ["pypy-3.6", "pypy-3.7"]
     outputs:
       python-key: ${{ steps.generate-python-key.outputs.key }}
     steps:
@@ -462,7 +462,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["pypy-3.6"]
+        python-version: ["pypy-3.6", "pypy-3.7"]
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.0


Pierre-Sassoulas (approved):
I'm ok with fixing the issue with the line/col either in 2.13 or 2.14 :)

.github/workflows/ci.yaml
```diff
@@ -476,7 +476,7 @@ jobs:
         with:
           path: venv
           key:
-            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{
+            ${{ runner.os }}-${{ matrix.python-version }}-${{
```
cdce8p, line 479: This shouldn't be necessary. The `Set up Python` version will output `${{ matrix.python-version }}`. As far as I can tell, the tests didn't crash.

DanielNoord: Hmm, this doesn't work as well.

Please see the version of `pytest` in the `pypy-3.7` run. That's the `3.6` version. You can see they are using the same cache in the `Restore Python virtual environment` job. The `key` is the same.

However, my fix also doesn't work. I guess something weird is going on with `python-version` and `PyPy`? For the `cpython` jobs the keys do work..

DanielNoord:
There is an issue with the current caching on `PyPy`. For some reason `steps.python.outputs.python-version` is empty on the `PyPy` jobs. I'm not sure why.
You can see in the first commit checks that `pypy-3.7` actually uses `3.6` for `pytest`. They share the same cache. For `3.6` and `3.7` this isn't a problem, but it will be when we add `3.8` and it was also the causes of the issues I had in #5849.

I could do a separate PR for the fix, but I think it should be fine to squash here, or to add two commits and rebase here.

-------

For the future: we should remove `3.6` when we officially drop support of `3.6`. After that we should look into testing `PyPy` on `3.8` as well and seeing what that requires from us.

.github/workflows/ci.yaml
cdce8p: This pretty much looks like a bug in the `setup-python` actions. If I had to guess a result of https://github.com/actions/setup-python/pull/342.

cdce8p: I've included the fix in #5868
Not sure when the bug will be addressed. We can always revert it, once a new version is released.

After #5868 is done, we should be able to move forward here.

cdce8p (approved):
I've rebased the PR to only include the addition of `pypy-3.7`.
Don't know if you wanted to do anything else, so will  leave it as draft for now. Feel free to merge it though if it's finished.

DanielNoord:
Thanks @cdce8p, I'm going to merge this!

PR #5834, opened 2022-02-23
New checker `unnecessary-list-index-lookup` (#4525)

timmartin:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
|    | :bug: Bug fix          |
| ✓   | :sparkles: New feature |
|    | :hammer: Refactoring   |
|    | :scroll: Docs          |

## Description

Added a new checker for catching the case of iterating over a list using `enumerate()`:

    for index, value in enumerate(mylist):
        print(mylist[index]) # Bad, should use value instead

This will now report

    Unnecessary list index lookup, use 'value' instead

Closes #4525

PR #5856, opened 2022-03-02
Use value directly instead of index in ``enumerate`` contexts

timmartin:
Refactoring to prevent warnings being issued on these lines from a new
checker.

## Type of Changes

## Description

The new `unnecessary-list-lookup` checker being added in PR #5834 will cause regressions, refactoring to avoid this.


DanielNoord (approved):
Thanks @timmartin 👍

PR #5863, opened 2022-03-05
Add stdlib `xml.etree.cElementTree` to deprecated modules

### LINKED ISSUE #5862: Warn if deprecated stdlib module `cElementTree` is imported
### Current problem

The stdlib module `cElementTree` has been deprecated in the docs since Python 3.3, and if imported just aliases `xml.etree.ElementTree`.

Core Python devs seem reluctant to issue `DeprecationWarning`s at runtime because of the longstanding idiom for attempting to import it and falling back to the python implementation, back when the C-accelerator was not on-by-default:

```python
try:
    import xml.etree.cElementTree as ETree
except ImportError:
    import xml.etree.ElementTree as ETree
```

This is the entire content of cElementTree:
```python
# Deprecated alias for xml.etree.ElementTree

from xml.etree.ElementTree import *
```

Which means the first example, although harmless [*], should just be:
```python
import xml.etree.ElementTree as ETree
```

[*] not _totally_ harmless, because someday the winds might blow the other way and a consensus develops to emit warnings and remove the module. Who knows?

There are murmurs on the Python issue ([bpo-36543](https://bugs.python.org/issue36543)) that maybe linters could help spread the deprecation notice to prevent people copy-pasting this dead code into the future.

> IMHO we don't have to spend extra work on an import hook. It might be a good idea to raise the problem with https://github.com/PyCQA/, though. Large projects tend to use code checkers and linters.

### Desired solution

I suggest importing `xml.etree.cElementTree` emits a `deprecated-module` warning.

### Additional context

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #5862, see full write-up there. 

`cElementTree` has been deprecated since Python 3.3, but does not raise `DeprecationWarning` to avoid pestering folks using really old idioms, from before the C-accelerator was on by default.

`pylint` can (and should!) be a bit more aggressive than that, and it was even floated on the python issue that this is a job for a linter.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -395,6 +395,10 @@ Release date: TBA
 
   Closes #258
 
+* Importing the deprecated stdlib module ``xml.etree.cElementTree`` now emits ``deprecated_module``.
+
+  Closes #5862
+
 * ``missing-raises-doc`` will now check the class hierarchy of the raised exceptions
 
   .. code-block:: python
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -382,6 +382,10 @@ Other Changes
 * The ``testutils`` for unittests now accept ``end_lineno`` and ``end_column``. Tests
   without these will trigger a ``DeprecationWarning``.
 
+* Importing the deprecated stdlib module ``xml.etree.cElementTree`` now emits ``deprecated_module``.
+
+  Closes #5862
+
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas (approved):
Good thing you're lurking in python's discussion group 😉

DanielNoord (approved):

PR #5864, opened 2022-03-07
Add `distutils` to deprecated modules

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |

## Description

`distutils` was deprecated in Python 3.10. Adding it (and some other tiny methods and decorators that were missed) to the list of deprecations.

PR #5874, opened 2022-03-07
Removed unused detection of site-packages directory

jacobtylerwalls:
## Type of Changes

## Description

<s>Remove use of deprecated module `distutils` by relying on astroid's determination of the location of site packages. It seems error-prone to have two places for this logic.</s>

EDIT: remove unused functionality for determining the location of site-packages (unused since 9812c5a49ae430945e9ad5dde7ec523904498805).

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -443,7 +443,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
 
     @staticmethod
     def _compute_site_packages() -> Set[str]:
-        def _normalized_path(path):
+        def _normalized_path(path: str) -> str:
             return os.path.normcase(os.path.abspath(path))
 
         return {_normalized_path(path) for path in astroid.modutils.STD_LIB_DIRS}
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -442,7 +442,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
         self._site_packages = self._compute_site_packages()
 
     @staticmethod
-    def _compute_site_packages():
+    def _compute_site_packages() -> Set[str]:
         def _normalized_path(path):
             return os.path.normcase(os.path.abspath(path))
 
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -50,8 +50,6 @@
 import collections
 import copy
 import os
-import sys
-from distutils import sysconfig
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
 
 import astroid
@@ -448,25 +446,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
         def _normalized_path(path):
             return os.path.normcase(os.path.abspath(path))
 
-        paths = set()
-        real_prefix = getattr(sys, "real_prefix", None)
-        for prefix in filter(None, (real_prefix, sys.prefix)):
-            path = sysconfig.get_python_lib(prefix=prefix)
-            path = _normalized_path(path)
-            paths.add(path)
-
-        # Handle Debian's derivatives /usr/local.
-        if os.path.isfile("/etc/debian_version"):
-            for prefix in filter(None, (real_prefix, sys.prefix)):
-                libpython = os.path.join(
-                    prefix,
-                    "local",
-                    "lib",
-                    "python" + sysconfig.get_python_version(),
-                    "dist-packages",
-                )
-                paths.add(libpython)
-        return paths
+        return {_normalized_path(path) for path in astroid.modutils.STD_LIB_DIRS}
 
     def open(self):
         """Called before visiting project (i.e set of modules)."""
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -439,15 +439,6 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
             ("RP0402", "Modules dependencies graph", self._report_dependencies_graph),
         )
 
-        self._site_packages = self._compute_site_packages()
-
-    @staticmethod
-    def _compute_site_packages() -> Set[str]:
-        def _normalized_path(path: str) -> str:
-            return os.path.normcase(os.path.abspath(path))
-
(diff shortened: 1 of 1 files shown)


pylint/checkers/imports.py
```diff
@@ -448,25 +446,7 @@ def _compute_site_packages():
...
-                libpython = os.path.join(
-                    prefix,
-                    "local",
-                    "lib",
-                    "python" + sysconfig.get_python_version(),
-                    "dist-packages",
-                )
-                paths.add(libpython)
-        return paths
+        return {_normalized_path(path) for path in astroid.modutils.STD_LIB_DIRS}
```
DanielNoord, line 449: Could you add typing to this method? 😏

jacobtylerwalls: Sure thing. The normalization is probably already done in astroid, but I didn't audit it over there, so I was reluctant to change it here.

DanielNoord: I don't think it actually is. At least `normcase` might be necessary/new here.

pylint/checkers/imports.py
```diff
@@ -444,29 +442,11 @@ def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         self._site_packages = self._compute_site_packages()
 
     @staticmethod
-    def _compute_site_packages():
+    def _compute_site_packages() -> Set[str]:
         def _normalized_path(path):
```
DanielNoord, line 446:
```suggestion
        def _normalized_path(path: str) -> str:
```

I think `mypy` complains about this with `--strict`.

DanielNoord (approved):

pylint/checkers/imports.py
jacobtylerwalls: And it should, because it would accept bytes also. But we only get strings from the astroid constant.

jacobtylerwalls:
I agree it's wise to put this in 2.14, but I suggest doing it right after 2.13 so it doesn't block contributors who have python 3.10 interpreters for pre-commit.

(Pylint's pre-commit hook doesn't specify a version, correct? 3.8 is running on CI. I had to `--no-verify` my changes on #5864 when removing the refactor.)

DanielNoord:
> (Pylint's pre-commit hook doesn't specify a version, correct? 3.8 is running on CI. I had to `--no-verify` my changes on #5864 when removing the refactor.)

On mobile, so I can't check. But: does deprecated-module do anything with the python-version setting? If your python-version is set to 3.6, should we emit deprecated-module warnings for distutils?

jacobtylerwalls:
> On mobile, so I can't check. But: does deprecated-module do anything with the python-version setting? If your python-version is set to 3.6, should we emit deprecated-module warnings for distutils?

Good question. The deprecated checker uses a map keyed on python versions so that warnings won't be emitted for earlier versions, e.g. `distutils` on python 3.9. That is why CI was passing for months with `pylint` using a deprecated module without a disable; CI uses python 3.8 for pre-commit.

I suppose we should just specify python3.8 in the pre-commit config?
```yaml
language_version: python3.8
```

cdce8p:
> On mobile, so I can't check. But: does deprecated-module do anything with the python-version setting? If your python-version is set to 3.6, should we emit deprecated-module warnings for distutils?

That was my question too. However it seems, the checker doesn't respect the `py-version` setting currently. Don't know though if that has been intentional or simply because `py-version` was added later.

It's definitely something worth exploring further IMO.

> I suppose we should just specify python3.8 in the pre-commit config?
> 
> ```yaml
> language_version: python3.8
> ```

`pylint` is defined as a system hook, since we need the custom environment with all dependencies. So that wouldn't work.

pylint/checkers/imports.py
```diff
@@ -444,29 +442,11 @@ def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         self._site_packages = self._compute_site_packages()
 
     @staticmethod
-    def _compute_site_packages():
-        def _normalized_path(path):
+    def _compute_site_packages() -> Set[str]:
+        def _normalized_path(path: str) -> str:
```
cdce8p, line 446: Another question: Where is it still used? I wanted to check if `STD_LIB_DIRS` is actual the correct replacement but couldn't find any code that uses `self._site_packges`.

pylint/checkers/imports.py
```diff
@@ -444,29 +442,11 @@ def __init__(self, linter: Optional["PyLinter"] = None) -> None:
...
-                libpython = os.path.join(
-                    prefix,
-                    "local",
-                    "lib",
-                    "python" + sysconfig.get_python_version(),
-                    "dist-packages",
-                )
-                paths.add(libpython)
-        return paths
+        return {_normalized_path(path) for path in astroid.modutils.STD_LIB_DIRS}
```
cdce8p, line 449:
```suggestion
        return {_normalized_path(path) for path in astroid.modutils.EXT_LIB_DIRS}
```
If these should be the paths to the `site-packages` dirs, it should be `EXT_LIB_DIRS` instead.
See https://github.com/PyCQA/astroid/pull/1322

cdce8p (changes requested):

pylint/checkers/imports.py
jacobtylerwalls: Yep, it looks unused. `pylint` used to rely on `EXT_LIB_DIRS`, until these special cases were added in 6e7037479b79b28ed8db4c74df7bff193f4574c4. Then the use of `_site_packages` was removed in 9812c5a49ae430945e9ad5dde7ec523904498805. I think this can just be removed without a replacement.

pylint/checkers/imports.py
jacobtylerwalls: Thanks, bad mistake. That was the PR I was looking at yesterday when I started down this path, and then I mistook the recent astroid changes for being relevant. 🤕

DanielNoord (approved):

Pierre-Sassoulas (approved):
Nice catch that it was unused ! In the end if we're just removing I think it can go in 2.13 ? (Btw we just reached 94% coverage, this kind of cleanup helps a lot to make the code base easier to maintain).

cdce8p (approved):

cdce8p:
Thanks @jacobtylerwalls 🐬

PR #5876, opened 2022-03-08
Prevent `useless-suppression` on disables for stdlib deprecation checker

jacobtylerwalls:
## Type of Changes

## Description
**Before**
<s>The deprecated names checker ignored the `py_version` setting, raising warnings (or not) depending on the python interpreter version.</s>
`useless-suppression` is raised on disables of `deprecated-module` and sim. on lower versions where the module is not yet deprecated. This can make it challenging to get a project to pass pylint on multiple versions with a disable for a deprecated module.

**Now**
<s>Deprecated name warnings are emitted according to `py_version` (which falls back to the interpreter version).

Discussed at https://github.com/PyCQA/pylint/pull/5874#issuecomment-1061261766. </s>
`useless-suppression` is treated as incompatible with the stdlib deprecation checks--it is not emitted.

**Tangent**
As a follow up, we could consider adding ~`--py-version=3.8` to the pre-commit config to match CI?~ After 2.13, I suppose it should be `--py-version=3.7`.

### DIFF
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -36,7 +36,7 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 
 """Checkers for various standard library functions."""
-
+import sys
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple
 
@@ -46,7 +46,6 @@ from astroid import nodes
 from pylint import interfaces
 from pylint.checkers import BaseChecker, DeprecatedMixin, utils
 from pylint.interfaces import IAstroidChecker
-from pylint.utils import get_global_option
 
 if TYPE_CHECKING:
     from pylint.lint import PyLinter
@@ -474,25 +473,20 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
         self._deprecated_modules: Set[str] = set()
         self._deprecated_decorators: Set[str] = set()
 
-    def open(self):
-        # In all versions
-        self._deprecated_methods.update(DEPRECATED_METHODS[0])
-        # Pad (X, Y) -> (X, Y, 0) since the constants are 3-tuples
-        py_version = (*get_global_option(self, "py-version"), 0)
-        for since_vers, func_list in DEPRECATED_METHODS[py_version[0]].items():
-            if since_vers <= py_version:
+        for since_vers, func_list in DEPRECATED_METHODS[sys.version_info[0]].items():
+            if since_vers <= sys.version_info:
                 self._deprecated_methods.update(func_list)
         for since_vers, func_list in DEPRECATED_ARGUMENTS.items():
-            if since_vers <= py_version:
+            if since_vers <= sys.version_info:
                 self._deprecated_arguments.update(func_list)
         for since_vers, class_list in DEPRECATED_CLASSES.items():
-            if since_vers <= py_version:
+            if since_vers <= sys.version_info:
                 self._deprecated_classes.update(class_list)
         for since_vers, mod_list in DEPRECATED_MODULES.items():
-            if since_vers <= py_version:
+            if since_vers <= sys.version_info:
                 self._deprecated_modules.update(mod_list)
         for since_vers, decorator_list in DEPRECATED_DECORATORS.items():
-            if since_vers <= py_version:
+            if since_vers <= sys.version_info:
                 self._deprecated_decorators.update(decorator_list)
 
     def _check_bad_thread_instantiation(self, node):
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -466,11 +466,11 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
 
     def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         BaseChecker.__init__(self, linter)
-        self._deprecated_methods: Set[Any] = set()
-        self._deprecated_attributes: Dict = {}
-        self._deprecated_classes: Dict = {}
-        self._deprecated_modules: Set[Any] = set()
-        self._deprecated_decorators: Set[Any] = set()
+        self._deprecated_methods: Set[str] = set()
+        self._deprecated_attributes: Dict[str, Tuple[Tuple[Optional[int], str], ...]] = {}
+        self._deprecated_classes: Dict[str, Set[str]] = {}
+        self._deprecated_modules: Set[str] = set()
+        self._deprecated_decorators: Set[str] = set()
 
     def open(self):
         # In all versions
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -467,7 +467,7 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
     def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         BaseChecker.__init__(self, linter)
         self._deprecated_methods: Set[str] = set()
-        self._deprecated_attributes: Dict[
+        self._deprecated_arguments: Dict[
             str, Tuple[Tuple[Optional[int], str], ...]
         ] = {}
         self._deprecated_classes: Dict[str, Set[str]] = {}
@@ -484,7 +484,7 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
                 self._deprecated_methods.update(func_list)
         for since_vers, func_list in DEPRECATED_ARGUMENTS.items():
             if since_vers <= py_version:
-                self._deprecated_attributes.update(func_list)
+                self._deprecated_arguments.update(func_list)
         for since_vers, class_list in DEPRECATED_CLASSES.items():
             if since_vers <= py_version:
                 self._deprecated_classes.update(class_list)
@@ -788,7 +788,7 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
         return self._deprecated_methods
 
     def deprecated_arguments(self, method: str):
-        return self._deprecated_attributes.get(method, ())
+        return self._deprecated_arguments.get(method, ())
 
     def deprecated_classes(self, module: str):
         return self._deprecated_classes.get(module, ())
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -466,19 +466,16 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
 
     def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         BaseChecker.__init__(self, linter)
-        self._deprecated_names_populated = False
         self._deprecated_methods: Set[Any] = set()
         self._deprecated_attributes: Dict = {}
         self._deprecated_classes: Dict = {}
         self._deprecated_modules: Set[Any] = set()
         self._deprecated_decorators: Set[Any] = set()
 
-    def _ensure_deprecated_names_populated(self):
-        if self._deprecated_names_populated:
-            return
+    def open(self):
         # In all versions
         self._deprecated_methods.update(DEPRECATED_METHODS[0])
-        # Pad (X.Y) -> (X.Y.0), since the constants are 3-tuples
+        # Pad (X, Y) -> (X, Y, 0) since the constants are 3-tuples
         py_version = (*get_global_option(self, "py-version"), 0)
         for since_vers, func_list in DEPRECATED_METHODS[py_version[0]].items():
             if since_vers <= py_version:
@@ -495,7 +492,6 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
         for since_vers, decorator_list in DEPRECATED_DECORATORS.items():
             if since_vers <= py_version:
                 self._deprecated_decorators.update(decorator_list)
-        self._deprecated_names_populated = True
 
     def _check_bad_thread_instantiation(self, node):
         if not node.kwargs and not node.keywords and len(node.args) <= 1:
@@ -784,23 +780,18 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
 
     def deprecated_modules(self):
         """Callback returning the deprecated modules."""
-        self._ensure_deprecated_names_populated()
         return self._deprecated_modules
 
     def deprecated_methods(self):
-        self._ensure_deprecated_names_populated()
         return self._deprecated_methods
 
     def deprecated_arguments(self, method: str):
-        self._ensure_deprecated_names_populated()
         return self._deprecated_attributes.get(method, ())
 
     def deprecated_classes(self, module: str):
-        self._ensure_deprecated_names_populated()
         return self._deprecated_classes.get(module, ())
 
     def deprecated_decorators(self) -> Iterable:
-        self._ensure_deprecated_names_populated()
         return self._deprecated_decorators
 
 
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -38,7 +38,7 @@
 """Checkers for various standard library functions."""
 
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
+from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple
 
 import astroid
 from astroid import nodes
@@ -467,7 +467,9 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
     def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         BaseChecker.__init__(self, linter)
         self._deprecated_methods: Set[str] = set()
-        self._deprecated_attributes: Dict[str, Tuple[Tuple[Optional[int], str], ...]] = {}
+        self._deprecated_attributes: Dict[
+            str, Tuple[Tuple[Optional[int], str], ...]
+        ] = {}
         self._deprecated_classes: Dict[str, Set[str]] = {}
         self._deprecated_modules: Set[str] = set()
         self._deprecated_decorators: Set[str] = set()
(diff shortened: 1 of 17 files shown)


pylint/checkers/stdlib.py
```diff
@@ -466,27 +466,36 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
 
     def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         BaseChecker.__init__(self, linter)
+        self._deprecated_names_populated = False
         self._deprecated_methods: Set[Any] = set()
+        self._deprecated_attributes: Dict = {}
+        self._deprecated_classes: Dict = {}
+        self._deprecated_modules: Set[Any] = set()
+        self._deprecated_decorators: Set[Any] = set()
```
Pierre-Sassoulas, line 474:
```suggestion
        self._deprecated_decorators: Set[Any] = set()
        self._ensure_deprecated_names_populated()
```
Would that work instead of calling it in each class function?

Pierre-Sassoulas (changes requested):
👍 Nice job, this is definitely a checker where the ``py-version`` is very useful.

pylint/checkers/stdlib.py
jacobtylerwalls: Unfortunately no, since`__init__()` runs before all option providers are parsed.

**Suggestion**
```
% python3.9 -m pylint tests/functional/d/deprecated/deprecated_methods_py39.py --py-version=3.8
************* Module deprecated_methods_py39
tests/functional/d/deprecated/deprecated_methods_py39.py:4:0: W1505: Using deprecated method b2a_hqx() (deprecated-method)

```
**PR**
```
% python3.9 -m pylint tests/functional/d/deprecated/deprecated_methods_py39.py --py-version=3.8

Your code has been rated at 10.00/10 (previous run: 5.00/10, +5.00)
```

jacobtylerwalls: Added a test case 9822f2f2eb66a9ebe596c8a5124d1d33e07bfa73

cdce8p: You can use the `open` method for it.
https://github.com/PyCQA/pylint/blob/6c0ba130730e82fc21063f138ffd05347bea0b9e/pylint/extensions/typing.py#L143-L144

pylint/checkers/stdlib.py
```diff
@@ -466,27 +466,36 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
...
+
+    def _ensure_deprecated_names_populated(self):
+        if self._deprecated_names_populated:
+            return
+        # In all versions
         self._deprecated_methods.update(DEPRECATED_METHODS[0])
-        for since_vers, func_list in DEPRECATED_METHODS[sys.version_info[0]].items():
-            if since_vers <= sys.version_info:
+        # Pad (X.Y) -> (X.Y.0), since the constants are 3-tuples
+        py_version = (*get_global_option(self, "py-version"), 0)
```
cdce8p, line 482: I don't think padding is necessary. Tuple comparison works even with tuples of different length.
```py
>>> (3, 8, 0) >= (3, 8)
True
```

cdce8p (changes requested):

pylint/checkers/stdlib.py
jacobtylerwalls: This is what I was running into:
```python
>>> since_vers = (3, 9, 0)
>>> py_version = (3, 9)
>>> since_vers <= py_version
False
>>> py_version_padded = (3, 9, 0)
>>> since_vers <= py_version_padded
True
```

jacobtylerwalls: Ah, but maybe I could rewrite the inequalities instead?

cdce8p: I see. In that case it's probably easiest to leave it as is, with padding. Just keep in mind that `py-version = 3.6.2` would  also be valid. AFAICT padding would work here too, though.

pylint/checkers/stdlib.py
```diff
@@ -467,25 +467,30 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
     def __init__(self, linter: Optional["PyLinter"] = None) -> None:
         BaseChecker.__init__(self, linter)
         self._deprecated_methods: Set[Any] = set()
+        self._deprecated_attributes: Dict = {}
```
cdce8p, line 470:
```suggestion
        self._deprecated_attributes: Dict[str, Tuple[Tuple[Optional[int], str], ...]] = {}
```

pylint/checkers/stdlib.py
```diff
@@ -467,25 +467,30 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
...
+        self._deprecated_classes: Dict = {}
```
cdce8p, line 471:
```suggestion
        self._deprecated_classes: Dict[str, Set[str]] = {}
```

pylint/checkers/stdlib.py
```diff
@@ -467,25 +467,30 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
...
+        self._deprecated_modules: Set[Any] = set()
```
cdce8p, line 472:
```suggestion
        self._deprecated_modules: Set[str] = set()
```

pylint/checkers/stdlib.py
```diff
@@ -467,25 +467,30 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
...
+        self._deprecated_decorators: Set[Any] = set()
```
cdce8p, line 473:
```suggestion
        self._deprecated_decorators: Set[str] = set()
```

pylint/checkers/stdlib.py
```diff
@@ -37,7 +37,6 @@
 
 """Checkers for various standard library functions."""
 
-import sys
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set
```
cdce8p, line 41:
```suggestion
from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
```

cdce8p (approved):
Looks good. I left a few comments with some more detailed types.

Pierre-Sassoulas (approved):

DanielNoord:
Bit late to the party here, but are we sure this should consider `py-version`? I was thinking the same thing earlier, but if we do we lose the "timely warning" effect of `deprecated-module`. Taking `distutils` as an example, we would only start to warn when `py-version==3.10`. What effect does `depracted-module` then have compared to the normal `DeprectationWarnings` that get emitted?

jacobtylerwalls:
> What effect does depracted-module then have compared to the normal DeprectationWarnings that get emitted?

My reply would be: it's better to divorce these things from your python interpreter. If you haven't (or won't be) upgrading your python interpreter from 3.8 to 3.10, then you can run with `--py-version=3.10` to get an early warning.

The docs should probably be updated to suggest running `--py-version` for highest and lowest supported versions, not just lowest.

jacobtylerwalls:
> What effect does depracted-module then have compared to the normal DeprectationWarnings that get emitted?

Well, if you have poor test coverage and then never trigger a warning during tests, pylint comes to the rescue! 😉

Pierre-Sassoulas:
Wow, I think Daniel is right. We probably want to do the opposite change and warn for all versions regardless of the interpreter used.

DanielNoord:
> My reply would be: it's better to divorce these things from your python interpreter. If you haven't (or won't be) upgrading your python interpreter from 3.8 to 3.10, then you can run with `--py-version=3.10` to get an early warning.

Agreed, but at the same time it kind of defeats the purpose of this message? A completely (and perhaps equally flawed approach) could be to remove `deprecated-...` from the `useless-suppression` warning. I believe that's the only thing that is really troublesome about this, right?

> The docs should probably be updated to suggest running `--py-version` for highest and lowest supported versions, not just lowest.

That runs a bit counter to what that setting is supposed to represent (both in `pylint` and in other tools that are slowly starting to support the setting).

jacobtylerwalls:
> That runs a bit counter to what that setting is supposed to represent (both in pylint and in other tools that are slowly starting to support the setting).

Got it, fair enough. But I wonder about other things besides the deprecated checker. What about the `consider-using-alias` that only operates on 3.7 and 3.8. If there's a `disable` for that in a codebase being linted against py 3.7, won't a python 3.10 interpreter flag that as `useless-suppression`? And now back to the same problem with pre-commit blocking people with different interpreters.

> Agreed, but at the same time it kind of defeats the purpose of this message? A completely (and perhaps equally flawed approach) could be to remove deprecated-... from the useless-suppression warning. I believe that's the only thing that is really troublesome about this, right?

If we warn on all versions for, say, `distutils` deprecation, projects that might not support py 3.10 for another two years or maybe _never_ (more likely, with say, 4.0 than 3.10...), then have to add disables everywhere they use the particular modules deprecated in 3.10. They can't do a global disable without losing warnings they want on 3.8 and below (e.g.). So they're wading into picking and choosing all the deprecated names.

***
(I'm open to whatever; interested in this for its own sake, not some passionate commitment to any direction.)

Pierre-Sassoulas:
I think it also makes sense to follow the logic used elsewhere (in the typing checker for example).

>  A completely (and perhaps equally flawed approach) could be to remove deprecated-... from the useless-suppression warning. 

I like this that way we could both have minimum false negative and false positive. Maybe more generally if a message uses the ``py-version`` at all we disable useless-suppression for it ? I think it make sense, the potential for false positive is high for multi-interpreter project. (The other false positives for useless-suppression I can think off are various versions of the same lib --numpy 1.19 vs 1.20, for example-- but we do not have this level of support for lib in astroid brain.)

I don't have a strong opinion about this. I think @cdce8p, as the creator of ``py-version``, will though  😄

cdce8p:
> What about the `consider-using-alias` that only operates on 3.7 and 3.8. If there's a `disable` for that in a codebase being linted against py 3.7, won't a python 3.10 interpreter flag that as `useless-suppression`?

The reason why I added `py-version` in the first place was to deal with these exact issues. Personally, I'm using Python 3.10. If I use that for pylint, I would end up with `suggestions` for typing which don't work in all supported version.
For your example, if the project sets `py-version = 3.7`, anyone linting it would get the same result. Regardless if they are using 3.7 or 3.10. Thus `useless-suppression` isn't an issue.

Some points to take away. `py-version` should always indicate the **oldest** supported version and it's used to emit **suggestions**, not warnings. _At least I think of it that way._

--
I'm not completely sure anymore it can be applied that way to the deprecation checker. Both arguments are valid here, use `py-version` - errors should only be emitted when **all** supported versions would emit one - and use `sys.version_info` at least that way maintainers know about upcoming deprecations earlier.

_It could even be that exact conflict why we haven't changed it already._

So how do we resolve this? Tbh I don't have a good answer here. One moment I like `py-version` and the next `sys.version_info`. Maybe the later is slightly better, considering that projects can actively choose to resolve deprecation warnings earlier?

--
> Agreed, but at the same time it kind of defeats the purpose of this message? A completely (and perhaps equally flawed approach) could be to remove `deprecated-...` from the `useless-suppression` warning. I believe that's the only thing that is really troublesome about this, right?

Not sure. Wouldn't that defeat the purpose of the `useless-suppression` check?
IMO `useless-suppression` just shouldn't be enabled by default. Just using it from time to time to cleanup messages is enough. That would resolve the whole issue. _I believe that's also what I argued for when it was added to pylintrc, but I might be wrong._

jacobtylerwalls:
Thanks, that's helpful. I think now `--py-version` is the wrong way to go, and this PR shouldn't tie it to the deprecated checker.

I looked at the original deprecated checker, no discussion about multiple-interpreter projects.

_Warn for all versions_ and _don't emit useless-suppression for the deprecated checker_ are not incompatible with each other, but nobody's asking for the former ATM, so I suggest just doing the smaller change to deal with the little bump I ran into. :-)

DanielNoord:
> Not sure. Wouldn't that defeat the purpose of the `useless-suppression` check? IMO `useless-suppression` just shouldn't be enabled by default. Just using it from time to time to cleanup messages is enough. That would resolve the whole issue. _I believe that's also what I argued for when it was added to pylintrc, but I might be wrong._

I think that just comes down to personal preference.

Fact remains that a project which uses `distutils` and runs its CI both on `3.9` and `3.10` will be unable to pass CI when both `useless-suppression` and `deprecated-module` are turned on. Having two messages be incompatible with each other seems like something we should try and avoid (and fix).

Another (far-fetched) idea would be to change `deprecated-module` to the `Information` category? That way it's easier to let CIs pass even though the message is emitted?

Pierre-Sassoulas:
> Some points to take away. py-version should always indicate the oldest supported version and it's used to emit suggestions, not warnings

Thank you for the reminder ! Taking that into account if ``py-version`` is 3.6 it means you can be supporting 3.10 (what we currently do in pylint). So if something deprecated in 3.7 is removed in 3.10 you can have 3.10 code not working because the deprecated function was removed and pylint not raising a warning (yet) because you still support 3.6.

By "emit suggestions, not warnings" are you saying we should decrease the level of the message like @DanielNoord suggested ? I think information message are deactivated by default so if we decrease to ``I`` most user won't see it at all.

cdce8p:
> By "emit suggestions, not warnings" are you saying we should decrease the level of the message like @DanielNoord suggested ? I think information message are deactivated by default so if we decrease to `I` most user won't see it at all.

No, **suggestions** is just the mental model I use here. As you said, information messages likely wouldn't be seen at all. So those don't help.

> _Warn for all versions_ and _don't emit useless-suppression for the deprecated checker_ are not incompatible with each other, but nobody's asking for the former ATM, so I suggest just doing the smaller change to deal with the little bump I ran into. :-)

Sounds good to me. In your changes, I saw that we already exclude `cyclic-import` messages. So adding a few more wouldn't be that big of a change. If @Pierre-Sassoulas and @DanielNoord I would suggest to move forward with that.

pylint/constants.py
```diff
@@ -173,3 +173,13 @@ class DeletedMessage(NamedTuple):
     # https://github.com/PyCQA/pylint/pull/3571
     DeletedMessage("C0330", "bad-continuation"),
 ]
+
+
+INCOMPATIBLE_WITH_USELESS_SUPPRESSION = [
```
cdce8p, line 178: This could be a set / frozenset. That should help with the lookups.

pylint/utils/file_state.py
```diff
@@ -162,7 +166,10 @@ def iter_spurious_suppression_messages(
                 if not enable and (warning, line) not in self._ignored_msgs:
                     # ignore cyclic-import check which can show false positives
                     # here due to incomplete context
-                    if warning != "R0401":
+                    # and deprecated-{module, argument, class, method, decorator}
+                    # which can cause false positives for multi-interpreter projects
+                    # when linting with an interpreter on a lower python version
```
cdce8p, line 171: Can you move the explanation to the constant file? I think it would make more sense there.

pylint/utils/file_state.py
```diff
@@ -162,7 +166,10 @@ def iter_spurious_suppression_messages(
...
+                    if warning not in INCOMPATIBLE_WITH_USELESS_SUPPRESSION:
```
cdce8p, line 172: The if statement can probably be combined with the previous one.

tests/functional/d/deprecated/deprecated_module_py39.rc
```diff
@@ -0,0 +1,5 @@
+[testoptions]
+min_pyver=3.9
+
+[Messages Control]
+enable=deprecated-module
```
cdce8p, line 5: Isn't it enabled by default?

cdce8p (changes requested):

pylint/constants.py
```diff
@@ -173,3 +173,13 @@ class DeletedMessage(NamedTuple):
...
+    "R0401",  # cyclic-import
+    "W0402",  # deprecated-module
+    "W1505",  # deprecated-method
+    "W1511",  # deprecated-argument
+    "W1512",  # deprecated-class
+    "W1513",  # deprecated-decorator
```
Pierre-Sassoulas, line 184: I'm wondering how hard it would be to still raise useless-suppression if the function/class/module/decorator... is not deprecated in any python version. I don't think this checker is very performance intensive, maybe relaunching with the highest python version (3.10 right now) is possible ?

cdce8p: Though about something similar, but in the context of a bare `disable` once the deprecation has been removed.

> I don't think this checker is very performance intensive, maybe relaunching with the highest python version (3.10 right now) is possible ?

Not sure that is practical. It would almost certainly change major parts of the application flow.
Maybe it's simpler to add a new config option to show **all** `useless-suppression` messages? We should mention however, that it should only be used temporarily and not added to `pylintrc`.

DanielNoord: @cdce8p I like your last suggestion!

tests/functional/d/deprecated/deprecated_module_py39.rc
jacobtylerwalls: Thanks, this was a left-over artifact from investigating unexpected test results.

jacobtylerwalls:
Thanks for the reviews -- they helped point this in the right direction and get it polished.

> Maybe it's simpler to add a new config option to show all useless-suppression messages? We should mention however, that it should only be used temporarily and not added to pylintrc.

Shall we defer that to another issue?

cdce8p:
> > Maybe it's simpler to add a new config option to show all useless-suppression messages? We should mention however, that it should only be used temporarily and not added to pylintrc.
> 
> Shall we defer that to another issue?

👍🏻

cdce8p (approved):
Looks good to me. I'll leave the final call for @DanielNoord and @Pierre-Sassoulas.
Thanks @jacobtylerwalls 🐬

Pierre-Sassoulas (approved):

DanielNoord (approved):
Thanks @jacobtylerwalls!

Could you make that follow-up issue about finding an elegant way to force `useless-suppression` to show all warnings?

PR #5877, opened 2022-03-08
Simplify hard to maintain copyright notice 

Pierre-Sassoulas:
## Description

git is the source of truth for the copyright, copyrite (the tool) was taking exponentially longer with each release, and it's
polluting the code with sometime as much as 50 lines of names. I replaced it with a single line linking to the contributors shown by github.

Also added a pre-commit hook to check and fix the copyright notice, and fixed the existing files so they have a notice

## Type of Changes

# Context

Counterpart of https://github.com/PyCQA/astroid/pull/1441, this is still a draft I'll take the remark from the astroid MR into account here.

Originally included in https://github.com/PyCQA/pylint/pull/5858


examples/custom.py
```diff
@@ -1,3 +1,7 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/graphs/contributors
+
```
cdce8p, line 4: I would suggest to exclude `examples\/*` too.

pylint/__main__.py
```diff
@@ -2,6 +2,7 @@
 
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/graphs/contributors
```
cdce8p, line 5: Do we want to replace this with a link to `CONTRIBUTORS.txt`?

Pierre-Sassoulas: Yes, it's the old code (draft) before taking into account the remarks from the astroid MR :)

DanielNoord:
@Pierre-Sassoulas I went ahead a separated some of the less controversial changes (changing the notice in `doc` and `pylint` to a separate PR) to keep this moving along.

PR #5881, opened 2022-03-08
Fix pyreverse type hinting for class methods

teobouvard:
This commit fixes the alignment of arguments and their type annotations
in pyreverse printer output.

Not a feature, is it an important bug fix ?

## Type of Changes

## Description

Until now, the following class

```python
class Example:
    def method(self, arg: int):
        pass
    
    @classmethod
    def class_method(cls, arg: int):
        pass
    
    @staticmethod
    def static_method(arg: int):
        pass
```

would generate the following PlantUML diagram when running `pyreverse -o puml example.py`

```puml
@startuml example
class Example {
  class_method(cls: int, arg)
  method(arg: int)
  static_method(arg)
}
@enduml
```

This output displays wrong type hints because the alignment of arguments and their type annotations is based on the first argument having name `self`. This is not the case for class methods, where the first argument is usually `cls` and static methods where the first argument is not related to the class.

This commit fixes the issue by performing the alignment using the method type rather than the parameters names. This generates the correct class diagram shown below.

```puml
@startuml example
class Example {
  class_method(arg: int)
  method(arg: int)
  static_method(arg: int)
}
@enduml
```

### DIFF
--- a/pylint/pyreverse/printer.py
+++ b/pylint/pyreverse/printer.py
@@ -100,7 +100,7 @@ class Printer(ABC):
     @staticmethod
     def _get_method_arguments(method: nodes.FunctionDef) -> List[str]:
         if method.args.args:
-            first_arg = 0 if method.type in ("function", "staticmethod") else 1
+            first_arg = 0 if method.type in {"function", "staticmethod"} else 1
             arguments: List[nodes.AssignName] = method.args.args[first_arg:]
         else:
             first_arg = 0
--- a/pylint/pyreverse/printer.py
+++ b/pylint/pyreverse/printer.py
@@ -100,13 +100,13 @@ class Printer(ABC):
     @staticmethod
     def _get_method_arguments(method: nodes.FunctionDef) -> List[str]:
         if method.args.args:
-            arguments: List[nodes.AssignName] = [
-                arg for arg in method.args.args if arg.name != "self"
-            ]
+            first_arg = 0 if method.type in ("function", "staticmethod") else 1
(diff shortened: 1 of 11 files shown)


teobouvard:
I'm going to try to cover the `else` branch to prevent a decrease of code coverage

DanielNoord:
@Pierre-Sassoulas, did you change the coveralls setting to disallow decreasing coverage PRs? The CI seems to have passed on that commit that did not have full coverage.

@teobouvard Thanks for the contribution! I have requested a review from @DudeNr33 as they are the `pyreverse` expert I believe.

Pierre-Sassoulas:
>  did you change the coveralls setting to disallow decreasing coverage PRs? 

No, I think it can be problematic, for example in https://github.com/PyCQA/pylint/pull/5874 we removed code and the coverage decreased. I don't want to (have to) deal with this kind of false positive. I think the easier thing to do would be to check the coverage before approving.

pylint/pyreverse/printer.py
```diff
@@ -99,14 +99,13 @@ def emit_edge(
...
-            arguments: List[nodes.AssignName] = [
-                arg for arg in method.args.args if arg.name != "self"
-            ]
-        else:
-            arguments = []
-
-        annotations = dict(zip(arguments, method.args.annotations[1:]))
+        if method.args.args is None:
+            return []
+
```
DudeNr33, line 104: Nice cleanup and lowering the indentation level!

DudeNr33 (approved):
Nice catch, thank you for the fix. Looks good to me, well tested with both relevant cases added to the example test data. 👍

PR #5891, opened 2022-03-10
Add broken Callable check

cdce8p:
## Type of Changes

## Description
Similar to the broken NoReturn check - #5304, however this time for `Callable`.
Using `collections.abc.Callable` inside `Optional` and `Union` is broken in Python 3.9.0 and 3.9.1.

This one is a bit more difficult as we currently currently suggest replacing the deprecated `typing.Callable` with `collections.abc.Callable`. Doing that can result in unexpected errors. The PR will change that behavior and **not** emit `deprecated-typing-alias` for `typing.Callable` if `py-version < (3, 9, 2)`. In addition it adds a new check to warn about potential errors.

**Example broken code**
```py
from collections.abc import Callable
from typing import Optional, Union

Alias1 = Optional[Callable[[int], None]
Alias2 = Union[Callable[[int], None], None]
```

**When exactly is it broken?**
* Python 3.9.0 or 3.9.1
* And `collections.abc.Callable` as part of `Optional` or `Union`
* And argument list for `Callable`, e.g. `[int]`.

https://bugs.python.org/issue42965

### DIFF
--- a/pylint/extensions/typing.py
+++ b/pylint/extensions/typing.py
@@ -81,7 +81,7 @@ class DeprecatedTypingAliasMsg(NamedTuple):
     node: Union[nodes.Name, nodes.Attribute]
     qname: str
     alias: str
-    parent_subscript: bool
+    parent_subscript: bool = False
 
 
 class TypingChecker(BaseChecker):
@@ -118,6 +118,14 @@ class TypingChecker(BaseChecker):
             "use string annotation instead. E.g. "
             "``Callable[..., 'NoReturn']``. https://bugs.python.org/issue34921",
         ),
+        "E6005": (
+            "'collections.abc.Callable' inside Optional and Union is broken in "
+            "3.9.0 / 3.9.1 (use 'typing.Callable' instead)",
+            "broken-collections-callable",
+            "``collections.abc.Callable`` inside Optional and Union is broken in "
+            "Python 3.9.0 and 3.9.1. Use ``typing.Callable`` for these cases instead. "
+            "https://bugs.python.org/issue42965",
+        ),
     }
     options = (
         (
@@ -152,7 +160,9 @@ class TypingChecker(BaseChecker):
     def __init__(self, linter: "PyLinter") -> None:
         """Initialize checker instance."""
         super().__init__(linter=linter)
+        self._found_broken_callable_location: bool = False
         self._alias_name_collisions: Set[str] = set()
+        self._deprecated_typing_alias_msgs: List[DeprecatedTypingAliasMsg] = []
         self._consider_using_alias_msgs: List[DeprecatedTypingAliasMsg] = []
 
     def open(self) -> None:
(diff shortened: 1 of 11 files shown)


pylint/extensions/typing.py
```diff
@@ -328,6 +367,57 @@ def _check_broken_noreturn(self, node: Union[nodes.Name, nodes.Attribute]) -> No
...
+
+        # Check nested inside Optional or Union
+        parent_subscript = node.parent.parent
+        if isinstance(parent_subscript, nodes.BaseContainer):
+            parent_subscript = parent_subscript.parent
+        if not (
+            isinstance(parent_subscript, nodes.Subscript)
+            and isinstance(parent_subscript.value, (nodes.Name, nodes.Attribute))
+        ):
+            return False
```
Pierre-Sassoulas, line 408: This line is not covered by the tests.

pylint/extensions/typing.py
```diff
@@ -169,8 +179,9 @@ def open(self) -> None:
         )
 
         self._should_check_noreturn = py_version < (3, 7, 2)
+        self._should_check_callable = py_version < (3, 9, 2)
 
-    def _msg_postponed_eval_hint(self, node) -> str:
+    def _msg_postponed_eval_hint(self, node: nodes.NodeNG) -> str:
```
Pierre-Sassoulas, line 184: 👍

pylint/extensions/typing.py
```diff
@@ -169,8 +179,9 @@ def open(self) -> None:
         )
 
         self._should_check_noreturn = py_version < (3, 7, 2)
+        self._should_check_callable = py_version < (3, 9, 2)
```
Pierre-Sassoulas, line 182: I like the optimization of calculating the boolean once here.

pylint/extensions/typing.py
```diff
@@ -284,7 +307,20 @@ def leave_module(self, node: nodes.Module) -> None:
...
+            for msg in self._deprecated_typing_alias_msgs:
+                if (
+                    self._found_broken_callable_location
+                    and msg.qname == "typing.Callable"
+                ):
+                    continue
+                self.add_message(
+                    "deprecated-typing-alias",
+                    node=msg.node,
+                    args=(msg.qname, msg.alias),
```
Pierre-Sassoulas, line 320:
```suggestion
                    args=(msg.qname, msg.alias),
                    confidence=interfaces.INFERENCE,
```
AFAIK because of ``inferred.qname()`` line 275 / 282 ?

Pierre-Sassoulas (changes requested):
Looks good already!

pylint/extensions/typing.py
cdce8p: Let's do that separately. It would change existing tests.

pylint/extensions/typing.py
cdce8p: Saw that too. Seems like my previous filter steps are quite good already. Took me a moment to find something that would run this line.

Pierre-Sassoulas (approved):
👌

PR #5894, opened 2022-03-11
Create a ``TypeVar`` style for ``invalid-name``

### LINKED ISSUE #3401: pylint should not use Class name regex for TypeVar names

DanielNoord:

## Type of Changes

## Description

Not sure if we want to do this, but:
After merging https://github.com/PyCQA/pylint/pull/5825, we can merge this. It only adds the necessary boilerplate for `typevar` name checking, but uses the pattern as set for classes.
This makes the diff of https://github.com/PyCQA/pylint/pull/5221 even smaller and should help getting that merged asap after #5825 has been merged.

Closes #3401.

### DIFF
--- a/doc/user_guide/options.rst
+++ b/doc/user_guide/options.rst
@@ -39,6 +39,8 @@ name is found in, and not the type of object assigned.
 +--------------------+---------------------------------------------------------------------------------------------------+
 | ``inlinevar``      | Loop variables in list comprehensions and generator expressions.                                  |
 +--------------------+---------------------------------------------------------------------------------------------------+
+| ``typevar``        | Type variable declared with ``TypeVar``.                                                          |
++--------------------+---------------------------------------------------------------------------------------------------+
 
 Default behavior
 ~~~~~~~~~~~~~~~~
@@ -117,6 +119,8 @@ expression will lead to an instance of ``invalid-name``.
 .. option:: --class-const-rgx=<regex>
 
 .. option:: --inlinevar-rgx=<regex>
+
+.. option:: --typevar-rgx=<regex>
 
 Multiple naming styles for custom regular expressions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/user_guide/options.rst
+++ b/doc/user_guide/options.rst
@@ -95,7 +95,7 @@ The following type of names are checked with a predefined pattern:
 +--------------------+---------------------------------------------------+------------------------------------------------------------+
 | Name type          | Good names                                        | Bad names                                                  |
 +====================+===================================================+============================================================+
-| ``TypeVar``        |`T`, `_CallableT`, `_T_co`, `AnyStr`, `DeviceTypeT`| `DICT_T`, `CALLABLE_T`, `ENUM_T`, `DeviceType`, `_StrType` |
+| ``typevar``        |`T`, `_CallableT`, `_T_co`, `AnyStr`, `DeviceTypeT`| `DICT_T`, `CALLABLE_T`, `ENUM_T`, `DeviceType`, `_StrType` |
 +--------------------+---------------------------------------------------+------------------------------------------------------------+
 
 Custom regular expressions
--- a/doc/user_guide/options.rst
+++ b/doc/user_guide/options.rst
@@ -84,6 +84,15 @@ Following options are exposed:
 
 .. option:: --inlinevar-naming-style=<style>
 
+Predefined Naming Patterns
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Pylint provides predefined naming patterns for some names. These patterns are often
+based on a Naming Style but there is no option to choose one of the styles mentioned above.
+The pattern can be overwritten with the options discussed below.
+
+The following type of names are checked with a predefined pattern:
+
+- TypeVars. Accepted names include: `T`, `TypeVar`, `_CallableT`, `_T_co`.
 
 Custom regular expressions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/user_guide/options.rst
+++ b/doc/user_guide/options.rst
@@ -92,7 +92,11 @@ The pattern can be overwritten with the options discussed below.
 
 The following type of names are checked with a predefined pattern:
 
-- TypeVars. Accepted names include: `T`, `TypeVar`, `_CallableT`, `_T_co`.
++--------------------+---------------------------------------------------+------------------------------------------------------------+
+| Name type          | Good names                                        | Bad names                                                  |
++====================+===================================================+============================================================+
+| ``TypeVar``        |`T`, `_CallableT`, `_T_co`, `AnyStr`, `DeviceTypeT`| `DICT_T`, `CALLABLE_T`, `ENUM_T`, `DeviceType`, `_StrType` |
++--------------------+---------------------------------------------------+------------------------------------------------------------+
 
 Custom regular expressions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -334,6 +334,11 @@ Other Changes
 
   Closes #5360, #3877
 
+* ``invalid-name`` check now also checks ``TypeVar`` for a pattern set with the new option
+  ``--typevar-rgx`` option or a default ``PascalCase`` based style.
+
+  Closes #3401
+
 * Fixed a false positive (affecting unreleased development) for
   ``used-before-assignment`` involving homonyms between filtered comprehensions
   and assignments in except blocks.
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -334,8 +334,8 @@ Other Changes
 
   Closes #5360, #3877
 
-* ``invalid-name`` check now also checks ``TypeVar`` for a pattern set with the new option
-  ``--typevar-rgx`` option or a default ``PascalCase`` based style.
+* Improve ``invalid-name`` check for ``TypeVar`` names.
+  The accepted pattern can be customized with ``--typevar-rgx``.
 
   Closes #3401
 
--- a/pylint/checkers/base.py
+++ b/pylint/checkers/base.py
@@ -1708,11 +1708,13 @@ def _create_naming_options():
     for name_type in sorted(KNOWN_NAME_TYPES):
         human_readable_name = constants.HUMAN_READABLE_TYPES[name_type]
         name_type_hyphened = name_type.replace("_", "-")
-        help_msg = f"Regular expression matching correct {human_readable_name} names."
+        help_msg = f"Regular expression matching correct {human_readable_name} names. "
+        if name_type in KNOWN_NAME_TYPES_WITH_STYLE:
+            help_msg += f"Overrides {name_type_hyphened}-naming-style. "
+        help_msg += f"If left empty, {human_readable_name} will be checked with the set naming style."
 
         if name_type in KNOWN_NAME_TYPES_WITH_STYLE:
             default_style = DEFAULT_NAMING_STYLES[name_type]
-            help_msg += f" Overrides {name_type_hyphened}-naming-style."
             name_options.append(
                 (
                     f"{name_type_hyphened}-naming-style",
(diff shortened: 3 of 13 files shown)


pylintrc
```diff
@@ -353,6 +353,13 @@ method-rgx=[a-z_][a-z0-9_]{2,}$
 # Naming hint for method names
 method-name-hint=[a-z_][a-z0-9_]{2,}$
 
+# Naming style used for type variables. Default is PascalCase.
+typevar-naming-style=PascalCase
```
cdce8p, line 357: I believe this was the most recent discussion we had: Should we add a `typevar-naming-style` add all or only have a `typevar-rgx` option?

The arguments so far
* If we say the default is `PascalCase`, it will be difficult to explain why `PathType` is invalid and `PathT` valid when both are PascalCase.
* It's unclear what the regex patterns for the other styles should be and how they should interact with the `_co` and `_contra` suffixes.
* Having only the the rgx option (with default) would still allow for customizability if wanted.
* What should be done in similar cases, e.g. for ParamSpec names or Type aliases?

Changing it should be possible if we modify `_create_naming_rules` and `_create_naming_options` to allow for rgx only options.

DanielNoord: I was hesitant to do this at first, but turned out to be relatively easy to implement. I have pushed an iteration which does this to this branch. This basically supersedes #5221. I'm also fine with pushing it to that branch to keep the discussion in one place, but perhaps a relatively clean slate is helpful here.

tests/checkers/unittest_base.py
```diff
@@ -210,17 +210,23 @@ class TestNamePresets(unittest.TestCase):
         SNAKE_CASE_NAMES | CAMEL_CASE_NAMES | UPPER_CASE_NAMES | PASCAL_CASE_NAMES
     )
 
-    def _test_name_is_correct_for_all_name_types(
+    def _test_name_is_correct_for_all_name_types_except_typevar(
         self, naming_style: Type[base.NamingStyle], name: str
     ) -> None:
         for name_type in base.KNOWN_NAME_TYPES:
```
cdce8p, line 216: Why not iterate only over those name types which accept naming styles?

cdce8p (changes requested):
This looks like the solution I had in mind. 👍🏻
I think it's a good idea to do the changes here and not in the old PR.

doc/whatsnew/2.13.rst
```diff
@@ -334,6 +334,11 @@ Other Changes
 
   Closes #5360, #3877
 
+* ``invalid-name`` check now also checks ``TypeVar`` for a pattern set with the new option
+  ``--typevar-rgx`` option or a default ``PascalCase`` based style.
```
DanielNoord, line 338:
```suggestion
* Improve ``invalid-name`` check for ``TypeVar`` names.
  The accepted pattern can be customized with ``--typevar-rgx``.
```

pylint/checkers/base.py
```diff
@@ -1693,28 +1707,36 @@ def _create_naming_options():
     name_options = []
     for name_type in sorted(KNOWN_NAME_TYPES):
```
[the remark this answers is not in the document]
DanielNoord: That would lead to quite a bit of code duplication though. I have added some line breaks and comments, perhaps that helps? If not, I'll change it. Let me know!

pylint/checkers/base.py
```diff
@@ -1862,15 +1884,20 @@ def open(self):
...
+    def _create_naming_rules(self) -> Tuple[Dict[str, Pattern[str]], Dict[str, str]]:
+        regexps: Dict[str, Pattern[str]] = {}
+        hints: Dict[str, str] = {}
 
         for name_type in KNOWN_NAME_TYPES:
-            naming_style_option_name = f"{name_type}_naming_style"
-            naming_style_name = getattr(self.config, naming_style_option_name)
-
-            regexps[name_type] = NAMING_STYLES[naming_style_name].get_regex(name_type)
+            if name_type in KNOWN_NAME_TYPES_WITH_STYLE:
```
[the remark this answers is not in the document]
DanielNoord: Changed some stuff around to make the `if...else` more similar.

pylintrc
```diff
@@ -353,6 +353,10 @@ method-rgx=[a-z_][a-z0-9_]{2,}$
 # Naming hint for method names
 method-name-hint=[a-z_][a-z0-9_]{2,}$
 
+# Regular expression which can overwrite the naming style set by typevar-naming-style.
+# If left empty, type variables will be checked with the set naming style.
+typevar-rgx=
```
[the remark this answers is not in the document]
DanielNoord: I checked. If it is empty the setting will be `None` (I think this is due to how the config file is parsed). The following lines make sure that the pattern is only picked up if the pattern is set. (Note that this is also how this works for all other patterns).
```python
custom_regex = getattr(self.config, custom_regex_setting_name, None)
if custom_regex is not None:
    regexps[name_type] = custom_regex
```

tests/checkers/unittest_base.py
DanielNoord: Ah that's a good point. This is from the old version, but that's a much better suggestion.

doc/user_guide/options.rst
```diff
@@ -82,6 +84,15 @@ Following options are exposed:
...
 
+Predefined Naming Patterns
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Pylint provides predefined naming patterns for some names. These patterns are often
+based on a Naming Style but there is no option to choose one of the styles mentioned above.
+The pattern can be overwritten with the options discussed below.
+
+The following type of names are checked with a predefined pattern:
+
+- TypeVars. Accepted names include: `T`, `TypeVar`, `_CallableT`, `_T_co`.
```
cdce8p, line 95: I wonder if a table would make sense here? With `Name type`, `Good names`, and `Bad names`.

I would prefer not to suggest `TypeVar`. True it's valid, but it shouldn't be used IMO. Let's use something like `MyClassT` to highlight PascalCase. For bad names, we should definitely mention something like `T_Class`, and `MyClassType`.

Some other names from the tests: `AnyStr`, `DeviceTypeT`, `DeviceType`, `CALLABLE_T`.

pylint/checkers/base.py
```diff
@@ -1693,28 +1707,37 @@ def _create_naming_options():
...
-                    "choices": list(NAMING_STYLES.keys()),
-                    "metavar": "<style>",
-                    "help": f"Naming style matching correct {human_readable_name} names.",
-                },
+        name_type_hyphened = name_type.replace("_", "-")
+
+        help_msg = f"Regular expression matching correct {human_readable_name} names. "
+        if name_type in KNOWN_NAME_TYPES_WITH_STYLE:
+            help_msg += f"Overrides {name_type_hyphened}-naming-style. "
+        help_msg += f"If left empty, {human_readable_name} will be checked with the set naming style."
```
cdce8p, line 1715: That's much better 👍🏻

cdce8p, line 1715:
```suggestion
        help_msg += f"If left empty, {human_readable_name} names will be checked with the set naming style."
```
This maybe?

pylintrc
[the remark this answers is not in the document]
cdce8p: An empty setting will be `re.compile('')` and thus overwrite `regexps[name_type]`. That isn't bad, we just can't leave it empty in `pylintrc`. Otherwise TypeVar names in pylint itself won't be checked. That's similar to the other regex settings in pylintrc, eg. `class-rgx`.

Alternatively, an possibly my preference would be the comment out that line. Similar to `#class-const-rgx=`.

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -0,0 +1,72 @@
...
+GoodBoundNameT_contra = TypeVar("GoodBoundNameT_contra", bound=int, contravariant=True)
+
+GoodBoundNameT_co = TypeVar("GoodBoundNameT_co", int, str, covariant=True)
+GoodBoundNameT_co = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodBoundNameT_co", int, str, contravariant=True
+)
+GoodBoundNameT_contra = TypeVar("GoodBoundNameT_contra", int, str, contravariant=True)
+GoodBoundNameT_contra = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodBoundNameT_contra", int, str, covariant=True
+)
```
cdce8p, line 33: Can we remove some of the `GoodName` TypeVars here? We already have dedicated tests for the `typevar-name-incorrect-variance` so I don't think we need to test this again here.

Basically just keep
```py
# PascalCase names with prefix
GoodNameT = TypeVar("GoodNameT")
_T = TypeVar("_T")
_GoodNameT = TypeVar("_GoodNameT")
__GoodNameT = TypeVar("__GoodNameT")
GoodNameT_co = TypeVar("GoodNameT_co", covariant=True)
GoodNameT_contra = TypeVar("GoodNameT_contra", contravariant=True)
```

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -0,0 +1,72 @@
...
+T_contra = TypeVar("T_contra", covariant=False, contravariant=True)
+
+# PascalCase names without prefix
+AnyStr = TypeVar("AnyStr")
+DeviceTypeT = TypeVar("DeviceTypeT")
+CALLABLE_T = TypeVar("CALLABLE_T")  # [invalid-name]
+DeviceType = TypeVar("DeviceType")  # [invalid-name]
+GoodNameWithoutContra = TypeVar(  # [typevar-name-incorrect-variance]
+    "GoodNameWithoutContra", contravariant=True
+)
```
cdce8p, line 54:
```suggestion
```

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -0,0 +1,72 @@
...
+
+# camelCase names with prefix
+badName = TypeVar("badName")  # [invalid-name]
+badNameWithoutContra = TypeVar(  # [invalid-name, typevar-name-incorrect-variance]]
+    "badNameWithoutContra", contravariant=True
+)
```
cdce8p, line 60:
```suggestion
```

cdce8p (changes requested):

Pierre-Sassoulas (approved):
Looks good !

pylintrc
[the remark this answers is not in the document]
DanielNoord: Ah I understand why I was mistaken. The functional test don't actually use this `pylintrc` so they passed even with this uncommented.

Commented it out!

doc/user_guide/options.rst
```diff
@@ -82,6 +84,19 @@ Following options are exposed:
...
+Pylint provides predefined naming patterns for some names. These patterns are often
+based on a Naming Style but there is no option to choose one of the styles mentioned above.
+The pattern can be overwritten with the options discussed below.
+
+The following type of names are checked with a predefined pattern:
+
++--------------------+---------------------------------------------------+------------------------------------------------------------+
+| Name type          | Good names                                        | Bad names                                                  |
++====================+===================================================+============================================================+
+| ``TypeVar``        |`T`, `_CallableT`, `_T_co`, `AnyStr`, `DeviceTypeT`| `DICT_T`, `CALLABLE_T`, `ENUM_T`, `DeviceType`, `_StrType` |
```
DanielNoord, line 98:
```suggestion
| ``typevar``        |`T`, `_CallableT`, `_T_co`, `AnyStr`, `DeviceTypeT`| `DICT_T`, `CALLABLE_T`, `ENUM_T`, `DeviceType`, `_StrType` |
```

PR #5897, opened 2022-03-11
Use ``node.position`` in ``add_message`` if available

### LINKED ISSUE #5466: Better error ranges for classes and functions

DanielNoord:

## Type of Changes

## Description

Closes #5466.

I'd like to get approval on the code before I do the automatic updating of the tests. So please review this already even if tests fail 😄


Pierre-Sassoulas (commented):
The current change LGTM :)

cdce8p (approved):
> The current change LGTM :)

👍🏻

DanielNoord:
@Pierre-Sassoulas Let's merge this first before we add all the regression test PRs. Otherwise we need to keep rebasing this.

cdce8p (commented):
Missed that earlier, do we have an option for error messages to either opt out of or opt into using the `position` attribute? Say for example one message should be emitted for the whole function and not just the `def <name>`. Maybe a new parameter for `add_message`?

DanielNoord:
> Missed that earlier, do we have an option for error messages to either opt out of or opt into using the `position` attribute? Say for example one message should be emitted for the whole function and not just the `def <name>`. Maybe a new parameter for `add_message`?

If you supply `line` etc. when you add the message nothing will be overwritten. Is that enough?

Note: apparently 3.8 is broken... I'll merge the other test PRs first while working on this.

Pierre-Sassoulas (approved):
Thank you @DanielNoord !

cdce8p:
> > Missed that earlier, do we have an option for error messages to either opt out of or opt into using the `position` attribute? Say for example one message should be emitted for the whole function and not just the `def <name>`. Maybe a new parameter for `add_message`?
> 
> If you supply `line` etc. when you add the message nothing will be overwritten. Is that enough?

My initial idea was a parameter to toggle between both options. Not passing each value individually. However, after comparing the changes here, I think it might no longer be necessary. So far I haven't found a case were it would be noticeable better to mark the whole code block. Let's leave it as it is for now.

PR #5901, opened 2022-03-12
Fix false positive for `invalid-class-object` when inference fails

jacobtylerwalls:
## Type of Changes

## Description
Fix a false positive for `invalid-class-object` when the object being assigned is not inferable. (`None` is one of the ambiguous cases returned by `safe_infer()`.)

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1482,8 +1482,12 @@ a metaclass class method.",
         if not node.attrname == "__class__":
             return
         inferred = safe_infer(node.parent.value)
-        if isinstance(inferred, nodes.ClassDef) or inferred is astroid.Uninferable:
-            # If is uninferrable, we allow it to prevent false positives
+        if (
+            isinstance(inferred, nodes.ClassDef)
+            or inferred is astroid.Uninferable
+            or inferred is None
+        ):
+            # If is uninferable, we allow it to prevent false positives
             return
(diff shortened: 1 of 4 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -1482,8 +1482,12 @@ def _check_invalid_class_object(self, node: nodes.AssignAttr) -> None:
         if not node.attrname == "__class__":
             return
         inferred = safe_infer(node.parent.value)
-        if isinstance(inferred, nodes.ClassDef) or inferred is astroid.Uninferable:
-            # If is uninferrable, we allow it to prevent false positives
+        if (
+            isinstance(inferred, nodes.ClassDef)
+            or inferred is astroid.Uninferable
+            or inferred is None
```
Pierre-Sassoulas, line 1488:
```suggestion
            or inferred in {astroid.Uninferable, None}
```

Pierre-Sassoulas (approved):
LGTM !

pylint/checkers/classes/class_checker.py
jacobtylerwalls: Happy to change, but it will be a bit slower, unless we make it a frozenset outside the condition. Do you think it's more readable on one line?

Pierre-Sassoulas: It's just a suggestion for readability, I did not think the performance would be worse. It's probably not worth creating a constant for.

PR #2338, opened 2018-07-23
annotations changes on new branch.

sushobhit27:
### Fixes / new features
- #2079

PR #5902, opened 2022-03-12
Fix typing of ``safe_infer``

DanielNoord:

## Type of Changes

## Description

While looking at #5901 I noticed that the typing on `safe_infer` is incorrect. `Uninferable` doesn't subclass `NodeNG` and thus the return annotation is incorrect. I also added typing to related functions.

Change was introduced in https://github.com/PyCQA/pylint/pull/2338/. I haven't fully investigated but I think this was just a mistake (or `Uninferable` was added in between these 4 years).

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1231,7 +1231,7 @@ def supports_delitem(value: nodes.NodeNG, _: nodes.NodeNG) -> bool:
     return _supports_protocol(value, _supports_delitem_protocol)
 
 
-def _get_python_type_of_node(node):
+def _get_python_type_of_node(node: nodes.NodeNG) -> Optional[str]:
     pytype = getattr(node, "pytype", None)
     if callable(pytype):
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
👌 I'm surprised it's not raising more mypy issue. Not taking into account the uninferable sure is making mypy less informative. I guess our tests are covering well enough so we catch the issue anyway 😓 .

DanielNoord:
> 👌 I'm surprised it's not raising more mypy issue. Not taking into account the uninferable sure is making mypy less informative. I guess our tests are covering well enough so we catch the issue anyway 😓 .

I think this is because `infer()` is untyped (something I'd like to fix at some point). This means that `value` becomes `Any` and `mypy` just accepts whatever the annotated return type is.
Then there is also the issue of `mypy` not understanding `Uninferable` correctly 😅

cdce8p (approved):
Looks good!

Issue #2366, opened 2018-07-30
useless-suppression false positive with ungrouped-imports


mthuurne:
### Steps to reproduce

Run pylint on the following code:

```python
# pylint: enable=useless-suppression
from pylint import run_pylint
import astroid
from pylint import run_pyreverse # pylint: disable=ungrouped-imports
```

### Current behavior

This message is issued:
```
4:0: I0021: Useless suppression of 'ungrouped-imports' (useless-suppression)
```
However, if the comment is removed from line 4, this message is issued:
```
4:0: C0412: Imports from package pylint are not grouped (ungrouped-imports)
```

### Expected behavior

Since there is an actual message being suppressed, I would expect to get no `useless-suppression` message.

### pylint --version output

```
pylint 2.0.1
astroid 2.0.1
Python 3.4.6 (default, Mar 22 2017, 12:26:13) [GCC]
```

PCManticore:
Thanks, I can reproduce this bug. While `useless-supression` is inoffensive in the sense that it is not consider towards any score or not considered as an error per se, this bug shows that there is some inconsistency somewhere.

brycepg:
So this particular check as a quit early conditional before adding the message:
```python
if not self.linter.is_message_enabled('ungrouped-imports', import_node.fromlineno):
      continue
```
Removing the line fixes the issue. Do we want to remove these guards to allow `add_message` to go through or disable `useless-supression` for that check if locally disabled?

bersbersbers:
This is still an issue with
```
pylint==2.5.0
astroid==2.4.0
```
on Python 3.8.2.

I believe my example code is functionally identical to the OP's:

```
"""False-positive useless-suppression."""
# pylint: enable=useless-suppression
from numpy import array
from pandas import DataFrame

if True and True:
    from numpy import zeros  # pylint:disable=ungrouped-imports

print(array, DataFrame, zeros)
```

This code gives
```
bug.py:7:0: I0021: Useless suppression of 'ungrouped-imports' (useless-suppression)
```

If you remove `# pylint:disable=ungrouped-imports`, you get
```
bug.py:7:4: C0412: Imports from package numpy are not grouped (ungrouped-imports)
```

bersbersbers:
[...]

PR #5903, opened 2022-03-12
Fix disabling of ``ungrouped-imports``

DanielNoord:

## Type of Changes

## Description

Needed for a PR that stops using the deprecated ``cachedproperty`` decorator from `astroid`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -73,6 +73,8 @@ Release date: TBA
 * Fixed an issue where ``ungrouped-imports`` could not be disabled without raising
   ``useless-suppression``.
 
+  Ref #2366
+
 * Added several checkers to deal with unicode security issues
   (see `Trojan Sources <https://trojansource.codes/>`_ and
   `PEP 672 <https://www.python.org/dev/peps/pep-0672/>`_ for details) that also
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
👍

jacobtylerwalls:
Cool. So it looks like #2366 wasn't truly fixed, as the regression test added in #5063 didn't repro -- it just lacked the final line from the original report! Is it worth referencing the original issue in the changelog or not?

DanielNoord:
@jacobtylerwalls I totally forgot I added that method. Anyway, I think that this solution is better (instead of using the method added in #5063). I think I did not fully look into `ungrouped-imports` when I did that issue and only focused on the message in the OP.

I could add the issue, but since it was "fixed" in 2.12 might be a bit weird. @Pierre-Sassoulas You decide 😄

jacobtylerwalls:
> I could add the issue, but since it was "fixed" in 2.12 might be a bit weird.

There are several cases where we just say "refers to" or "refs #..." right? best of both worlds.

Pierre-Sassoulas:
I think we don't need to create issue when the problem is fixed (We can ``Refs #The PR`` so those wanting detail can find them).

jacobtylerwalls (approved):

PR #5907, opened 2022-03-12
Use ``functools.cached_property`` on 3.8+

DanielNoord:

## Type of Changes

## Description

Last PR to fix all deprecations as a result of `astroid` upgrade 🎉

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -4,9 +4,19 @@
 import collections
 import copy
 import itertools
+import sys
 import tokenize
 from functools import reduce
-from typing import Dict, Iterator, List, NamedTuple, Optional, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Dict,
+    Iterator,
+    List,
+    NamedTuple,
+    Optional,
+    Tuple,
+    Union,
+)
 
 import astroid
 from astroid import nodes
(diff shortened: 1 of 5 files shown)


jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
👌

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -17,6 +27,13 @@
 from pylint.checkers import utils
 from pylint.checkers.utils import node_frame_class
 
+if sys.version_info >= (3, 8) or TYPE_CHECKING:
+    # pylint: disable-next=ungrouped-imports
+    from functools import cached_property
+else:
+    # pylint: disable-next=ungrouped-imports
+    from astroid.decorators import cachedproperty as cached_property
```
cdce8p, line 35: I thought about this some more, especially since I wanted to see what it would take to resolve / hide the pylint error.

First, the reason it's emitted is that we don't check `BoolOp` nodes (in `is_node_in_guarded_import_block`) when we determine if `ungrouped-imports` should be emitted.
https://github.com/PyCQA/pylint/blob/dee07af04df93c4b5edeb2492eaa5c46ebcdb7b7/pylint/checkers/imports.py#L548-L550

The question then becomes if we should do that. To answer that it's worth taking a look at the condition itself and how `cached_property` is defined in `typeshed`. Basically `typeshed` also uses version guards, in particular `if sys.version_info >= (3, 8)`. That means the `or TYPE_CHECKING` I suggested in astroid is basically useless. If we run mypy with `--python-version 3.8` or above, the `sys.version_info` check is enough, whereas with  `3.7` we just get an `attr-defined` error anyway.
```
Module "functools" has no attribute "cached_property"  [attr-defined]
```

In the end `or TYPE_CHECKING` is useless and causes the unnecessary errors. So it might be best / easiest if we remove them.

PR #5908, opened 2022-03-13
Add regression test for #5771

### LINKED ISSUE #5771: False positive W0613 with keyword only argument
### Bug description

Using a keyword-only argument is triggering a W0613(unused-argument) warning where it is not occurring if the (unused) argument is a keyword argument. Below, `B.add` does not issue the warning, but `D.add` does.

kwoa.py
```python
# pylint: disable=C*,R*
class A:
    def add(self, a, b=0):
        return a + b

class B(A):
    def add(self, a, b=0):
        return None

class C:
    def add(self, a, *, b=0):
        return a + b

class D(C):
    def add(self, a, *, b=0):
        return None
```

### Configuration

_No response_

### Command used

```shell
pylint kwoa.py
```

### Pylint output

```shell
************* Module kwoa
kwoa.py:16:24: W0613: Unused argument 'b' (unused-argument)
```

### Expected behavior

I believe there are two possibilities here but I had been operating under the assumption that this warning would not occur for `D.add` given that is not the behavior for `B.add`. The other option is that a warning should also be issued `B.add`.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.9.5 (default, May 18 2021, 12:31:01) 
[Clang 10.0.0 ]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #5771 
Refs PyCQA/astroid#1382

Pardon me if this shouldn't go in 2.13! I just noticed no regression test PR for it yet, even though it's already fixed. Feel free to bump the milestone.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2310,9 +2310,7 @@ class VariablesChecker(BaseChecker):
             if global_names and _import_name_is_global(stmt, global_names):
                 return
 
-        argnames = list(
-            itertools.chain(node.argnames(), [arg.name for arg in node.args.kwonlyargs])
-        )
+        argnames = node.argnames()
         # Care about functions with unknown argument (builtins)
         if name in argnames:
             self._check_unused_arguments(name, node, stmt, argnames)
--- a/ChangeLog
+++ b/ChangeLog
@@ -154,6 +154,12 @@ Release date: TBA
   Closes #5408
   Ref PyCQA/astroid#1392
 
+* Fixed false positive for ``unused-argument`` when a method overridden in a subclass
+  does nothing with the value of a keyword-only argument.
+
+  Closes #5771
+  Ref PyCQA/astroid#1382
+
 * The issue template for crashes is now created for crashes which were previously not covered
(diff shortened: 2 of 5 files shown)


pylint/checkers/variables.py
```diff
@@ -2310,9 +2310,7 @@ def _check_is_unused(self, name, node, stmt, global_names, nonlocal_names):
             if global_names and _import_name_is_global(stmt, global_names):
                 return
 
-        argnames = list(
-            itertools.chain(node.argnames(), [arg.name for arg in node.args.kwonlyargs])
-        )
+        argnames = node.argnames()
```
jacobtylerwalls, line 2313: We discussed leveraging the changes in astroid to simplify this here: https://github.com/PyCQA/pylint/issues/5771#issuecomment-1030703199

Pierre-Sassoulas: Pretty nice simplification !

Pierre-Sassoulas (approved):

PR #5916, opened 2022-03-14
Fix pyreverse type hints for methods returning None

teobouvard:

## Type of Changes

## Description

Previously, the following class

```python
class Example:
    def method(self, arg: int) -> None:
        pass
```

would generate the following PlantUML diagram when running `pyreverse -o puml example.py`

```puml
@startuml example
class Example {
  method(arg: int) -> 
}
@enduml
```

This does not show that the method was annotated with `None` as return type.
This commit fixes the issue by using the `as_string()` method for all `NodeNG` arguments to `get_annotation_label` rather than limiting it to `Subscript` arguments. This generates the diagram below.

```puml
@startuml example
class Example {
  method(arg: int) -> None
}
@enduml
```


teobouvard:
@DudeNr33 I'm not sure the unit tests I added are really the right way to thoroughly test the `get_annotation_label()` method, feel free to suggest alternatives.

ChangeLog
```diff
@@ -17,6 +17,8 @@ Release date: TBA
 
 * Fix pyreverse diagrams type hinting for classmethods and staticmethods.
 
+* Fix pyreverse diagrams type hinting for methods returning None.
```
DanielNoord, line 20: Could you als add a changelog entry to the `2.13` file? It's in `doc/whatsnew`

I checked and it is also missing for you previous PR. Perhaps you could add that one in this PR as well?

DanielNoord (changes requested):

DudeNr33 (commented):
I looked at the tests and thought a bit about what we want to test here.

The ``get_annotation_label`` function is called from two places:
* from within ``get_annotation``
* from within the individual concrete ``Printer`` classes' ``emit_node`` methods

The first usage is already well covered with the existing functional tests of ``pyreverse``.
The second usage is also considered in your updated test data (``clientmodule_test``) 👍 

For the unit tests I would say we do not need to cover the first usage scenario, as ``get_annotation_label`` is only called "under the hood" of the already unit tested ``get_annotation`` function.

So I would suggest to focus the unit tests for ``get_annotation_label`` more on the second usage (getting the label for a ``nodes.FunctionDef.returns`` attribute), and also including a case of ``nodes.Subscript`` to ensure there are no regressions:

```python
@pytest.mark.parametrize(
    "node_text, expected_label",
    [
        ("def foo() -> None:            pass", "None"),
        ("def foo() -> int:             pass", "int"),
        ("def foo(a) -> Optional[int]:  return 1 if a else None", "Optional[int]"),
        ("def foo() -> 'MyType':        pass", "'MyType'"),
    ],
)
def test_get_annotation_label_of_return_type(node_text: str, expected_label: str) -> None:
    func = astroid.extract_node(node_text)
    assert get_annotation_label(func.returns) == expected_label
```

This new test should include your current unit tests and makes it - in my opinion - a bit clearer to understand the intention of this test. I don't have a strong opinion on this, so feel free to incorporate this feedback or leave the tests as they are.

Daniel already pointed out the missing ``whatsnew`` entry.
Otherwise the PR looks good to me!

ChangeLog
DudeNr33: The Milestone is currently set to 2.14.0 though (however, the last PR was terminated for 2.14 as well initially and then merged into 2.13.0). 
Should we start including a section for 2.14 in the ``ChangeLog`` already, to avoid confusion?

teobouvard: I added it to the 2.13, tell me if you would rather target 2.14 for this change.

teobouvard:
> This new test should include your current unit tests and makes it - in my opinion - a bit clearer to understand the intention of this test. I don't have a strong opinion on this, so feel free to incorporate this feedback or leave the tests as they are.

Thank you for this suggestion, this is indeed a lot clearer.

DudeNr33 (approved):
All good from my side now! Thanks for the fix!

ChangeLog
DanielNoord: Let's do `2.13` for this as the review is done and we can merge it :)

doc/whatsnew/2.13.rst
```diff
@@ -502,3 +502,5 @@ Other Changes
 * Fixed false positive for ``global-variable-not-assigned`` when the ``del`` statement is used
 
   Closes #5333
+
+* Fix type hints in class diagrams generated by pyreverse for class methods and methods returning None.
```
DanielNoord, line 506: Could you add the note for your previous fix as well?

DanielNoord (changes requested):

doc/whatsnew/2.13.rst
teobouvard: @DanielNoord This sentence includes both fixes, do you want me to split it into two bullet points ?

DanielNoord: Ah shit, I didn't notice that. Hmm, I guess this is fine! Thanks!

Issue #4683, opened 2021-07-07
Cleanup after we drop support for Python 3.6


Pierre-Sassoulas:
When we do drop support because 3.6 has reached its end of life, we'll need to do some cleanup:

- [x] remove the code for Python 3.6 only and `PY36_PLUS` conditionals
- [x] clean up the minimum required version in the tests and functional tests RC files (this has not been done previously) 
- [x] Use postponed evaluation of annotations (with from ``__future__`` import annotations) -> #6326
- [x] Use dataclasses to simplify the code where it makes the most sense
- [x] Bump `py-version` in `pylintrc` -> #6326
- [ ] Look at coverage drop in https://github.com/PyCQA/pylint/pull/5921 and remove dead code > #6421

cdce8p:
With the removal, we could also start using features that have been added in `3.7`
The most relevant to us IMO would be:
- Postponed evaluation of annotations (with `from __future__ import annotations`)
- Dataclasses

https://www.python.org/downloads/release/python-370/
https://docs.python.org/3.7/whatsnew/3.7.html#new-features

cdce8p:
Just for reference: #4310 (Issue discussing when the removal of 3.6 should happen)

Pierre-Sassoulas:
Thank for the suggestion I added postponed evaluation to the list. Regarding dataclasses, could you explain the advantages compared to keeping NamedTuple for example ?

cdce8p:
Dataclasses are essentially just classes. So compared to `NamedTuple` you can change attributes without creating a new instance of it. Where dataclasses really shine is for classes that just implement some boilerplate code (`__init__` to assign all attributes to instance attrs, basic `__str__`, `__repr__`, `__eg__`, ...). Basically, they are meant for classes that mostly store data (hence the name). https://docs.python.org/3/library/dataclasses.html

Good examples where to use it can be found in the PR for the Similarity checker. 
https://github.com/PyCQA/pylint/blob/65a20b6c5075e1f1756c1e6033c20b289b6edbbf/pylint/checkers/similar.py#L97-L114

https://github.com/PyCQA/pylint/blob/65a20b6c5075e1f1756c1e6033c20b289b6edbbf/pylint/checkers/similar.py#L163-L189

Pierre-Sassoulas:
I added dataclasses to the list :) I think it should be done in good intelligence though (ie: not everywhere for refactor's sake)

cdce8p:
[...]

PR #5921, opened 2022-03-15
Require python ``3.7.2+``

### LINKED ISSUE #4310: Remove support for Python 3.6 and future Python support
Python 3.6 was release on the **2016-12-23** and is nearing the end of support on 2021-12-23.
Especially while working on support for the `typing` module, it has become apparent that a lot of special case handling is needed for `3.6`. I don't think it's worth it to support `3.6` in new releases. The majority of users does use at least `3.8` as seen from the recent bug reports.

**My suggestion**
Remove support for **Python 3.6** in `pylint` and `astroid` after the next minor release of `pylint`.

**Moving forward**
It might not be reasonable for us to always support all official supported Python version. The scientific python community has adopted an IMO good standard of supporting all minor version of Python that have been released in the last **42 months** or at least the **last two**. See [NEP 029](https://numpy.org/neps/nep-0029-deprecation_policy.html).
`pylint` might have a boarder audience, so widening this window to the **last three** minor releases might make sense.

/CC: @PyCQA/pylint-dev

DanielNoord:

## Type of Changes

## Description

Two additional follow up PRs would be necessary after this:
1) Remove all tests options for `min_pyver=3.7`. That would be quite a large diff so I have excluded it from this PR.
2) Update `py-version` in `pylintrc`. This causes a number of warnings about using `from __future__ import annotations`. So that warrants its own PR.
3) _We might want to run https://github.com/cdce8p/python-typing-update before doing step 2 just to see what can be automated_

Just a draft to see if I got everything that needs to be changed for an initial PR.

Closes #4310
First step toward #4683


DanielNoord:
With respect to coverage. The first decrease seems "correct", it is no longer covered. Do we want to remove it?

The second coverage decline comes from `assign-to-new-keyword` become irrelevant. There are no new keywords introduced after 3.7 (`match` and `case` are soft keywords), so the message will never be emitted for anything. Again, do we want to remove this? Or just keep it in the codebase without any coverage?

Pierre-Sassoulas (commented):
We could wait a little for this one to avoid issue with Python 3.6 when backporting fixes on 2.13. What do you think?

Pierre-Sassoulas:
I think we should remove dead code no longer accessed.

DanielNoord:
Fine with me! Shall we say first week of April?

Pierre-Sassoulas:
Alternatively we could merge this release 2.14.0 without python 3.6 right now, and move everything that is in 2.14 right now to 2.15.0.

cdce8p:
I would suggest to wait a bit longer before dropping 3.6, even if it's only in a dev branch. Once it's committed, there sure will be a lot of cleanup and update PRs which will make bugfixes / backports even more difficult. A few weeks more or even a month doesn't make much of a difference IMO.

> Alternatively we could merge this release 2.14.0 without python 3.6 right now

Additionally, it would be better not to release new minor versions with so little time after one another. The fast bugfix releases are much appreciated though!

Pierre-Sassoulas:
> Once it's committed, there sure will be a lot of cleanup and update PRs which will make bugfixes / backports even more difficult.

The idea I had (following #5988 where the test fail for python 3.6) is to not do any bug fixes for python 3.6 / pylint 2.13 and just support pylint 2.14 / python 3.7.

cdce8p:
> > Once it's committed, there sure will be a lot of cleanup and update PRs which will make bugfixes / backports even more difficult.
> 
> The idea I had (following #5988 where the test fail for python 3.6) is to not do any bug fixes for python 3.6 / pylint 2.13 and just support pylint 2.14 / python 3.7.

🤔 Not sure about that. I would still recommend to wait. I agree pylint 2.14.0 should be the one which requires 3.7, but it would make more sense to include the whole cleanup there as well.

Until we're ready, bugfixes for 2.13.x still need to be tested on 3.6.

After all, the particular issue with 3.6 wasn't a big deal and easy to resolve. It should be possible to support it for a few more weeks now, too.

pylint/extensions/typing.py
```diff
@@ -361,12 +361,9 @@ def _check_broken_noreturn(self, node: Union[nodes.Name, nodes.Attribute]) -> No
             if (
                 isinstance(inferred, (nodes.FunctionDef, nodes.ClassDef))
                 and inferred.qname() in TYPING_NORETURN
-                # In Python 3.6, NoReturn is alias of '_NoReturn'
-                # In Python 3.7 - 3.8, NoReturn is alias of '_SpecialForm'
```
cdce8p, line 365: I would leave the second line.

cdce8p (commented):
Not a full review, but from what I've seen this looks about right.

Some notes
* We didn't add CI tests for `pypy-3.8`, right?
* This should definitely be done separately, but we should make sure to update `py-version`. Just a warning, that might trigger a lot a pylint typing warning 😄
https://github.com/PyCQA/pylint/blob/96d7e898cc863a6854cd0870ba4069c8f7d7851b/pylintrc#L56-L57

DanielNoord:
> * We didn't add CI tests for `pypy-3.8`, right?

No we need to exclude line number checking. Something for another PR :)

> * This should definitely be done separately, but we should make sure to update `py-version`. Just a warning, that might trigger a lot a pylint typing warning 😄
>   https://github.com/PyCQA/pylint/blob/96d7e898cc863a6854cd0870ba4069c8f7d7851b/pylintrc#L56-L57

I tried, and yeah that's also something for a follow-up 😄

Pierre-Sassoulas:
> we should make sure to update py-version

There's a checklist in #4683 I assumed we were going to do it later on, but doing the whole checklist here would make sense.

cdce8p:
> > we should make sure to update py-version
> 
> There's a checklist in #4683 I assumed we were going to do it later on, but doing the whole checklist here would make sense.

Not so sure about that. It might be easier to review if they are split up into multiple PRs.

DanielNoord:
@Pierre-Sassoulas Shall we merge this Monday 11th? We could do one final hotfix this weekend if needed without the trouble of having the hotfix supporting `3.6` and `main` `3.7`. After that we will probably need to be a little more picky when choosing whether something needs to be back ported.

Pierre-Sassoulas:
Yes let's do it. I think we can (try to) release 2.14 fairly soon after. This way we don't have to backport to 2.13 and think about python 3.6 anymore

DanielNoord:
> Yes let's do it. I think we can (try to) release 2.14 fairly soon after. This way we don't have to backport to 2.13 and think about python 3.6 anymore

There is quite a lot still on the `argparse` to-do list, but I agree. We should be able to do `2.14` much sooner than the gap between `2.12` and `2.13`. Especially with the speed we have been merging lately.

pylint/extensions/typing.py
```diff
@@ -367,12 +367,9 @@ def _check_broken_noreturn(self, node: Union[nodes.Name, nodes.Attribute]) -> No
             if (
                 isinstance(inferred, (nodes.FunctionDef, nodes.ClassDef))
                 and inferred.qname() in TYPING_NORETURN
```
DanielNoord, line 369:
```suggestion
                and inferred.qname() in TYPING_NORETURN
                # In Python 3.7 - 3.8, NoReturn is alias of '_SpecialForm'
```

jacobtylerwalls:
The plan is still to merge this after a likely 2.13.6? Or are we cancelling that milestone?

DanielNoord:
> The plan is still to merge this after a likely 2.13.6? Or are we cancelling that milestone?

I think so, but I think we can already finish the review of the changes here. I'll fix the merge conflict first though.

Pierre-Sassoulas:
> The plan is still to merge this after a likely 2.13.6?

I need to create a test for the namespace changes but I've been quite busy lately and the main blocker for 2.13.6 is https://github.com/PyCQA/pylint/pull/6141 that needs to be reviewed by Marc ideally. We could release 2.13.6 without this issue in the worst case scenario. And the issue with 3.6 would be manageable anyway. I think we can merge this one.

tests/functional/d/deprecated/deprecated_method_getmoduleinfo.txt
```diff
@@ -1 +0,0 @@
-deprecated-method:4:::Using deprecated method getmoduleinfo()
```
jacobtylerwalls, line 1: Why not leave this?

jacobtylerwalls (changes requested):

tests/functional/d/deprecated/deprecated_method_getmoduleinfo.txt
DanielNoord: The maximum test version is 3.6, so when we remove the 3.6 runner this test will never be run again.

jacobtylerwalls: Sorry, I wasn't clear. I'm questioning why the better solution is not to remove the maximum test version, which looks like it should never have been there in the first place. (pylint warns forever about deprecated stdlib materials, right?)

DanielNoord:
```console
>       self._runTest()
E       AssertionError: Wrong results for file "deprecated_method_getmoduleinfo":
E       
E       Expected in testdata:
E          4: deprecated-method
E       
E          4: no-member

pylint/testutils/lint_module_test.py:95: AssertionError
```

This is with python 3.10. After `3.6` the method was actually removed so we run into `no-member` messages. I guess that has the same effect as `deprecated-method`?

DanielNoord: This does sound like we could improve this somewhat indeed!

Pierre-Sassoulas (approved):

PR #5923, opened 2022-03-15
Handle ``ComprehensionScope`` and their behaviour better

DanielNoord:

## Type of Changes

## Description

Refactoring necessary for https://github.com/PyCQA/astroid/pull/1475 to be "acceptable" for `pylint`.

Note that the change to `names_hint = ", ".join(names)` might be unrelated, but I couldn't get `pre-commit` to pass without it and I wanted to show that if you clone https://github.com/PyCQA/astroid/pull/1475 and these changes then both test suites should pass.

### DIFF
--- a/pylint/checkers/classes/special_methods_checker.py
+++ b/pylint/checkers/classes/special_methods_checker.py
@@ -289,6 +289,7 @@ class SpecialMethodsChecker(BaseChecker):
         if isinstance(node, astroid.bases.Generator):
             # Generators can be iterated.
             return True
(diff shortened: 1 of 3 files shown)


DanielNoord:
> I think we could merge directly for 2.13.0, after investigation of the coverage change ? Imo the code in pylint never considered that the comprehension scope could be something else than uninferrable because it never was, but this is not a voluntary design choice.

The coverage change is just two `if` statements for `ComprehensionScope` that will be covered when the change lands in `astroid`. So they are as expected.

Pierre-Sassoulas (approved):

PR #5927, opened 2022-03-16
Better configuration for duplicate email/name in the contributors

Pierre-Sassoulas:
## Type of Changes

## Description

Separating the work from #5877 in two MR. @DanielNoord 

We need to find duplicate of email and name in the shortlog.

For example 'frostming' has a duplicate as I open this MR

 ``- Frost Ming <mianghong@gmail.com> (frostming <frostming@tencent.com>)`` 

Needs a conf like:

```yaml
  "mianghong@gmail.com": {
    "mails": ["mianghong@gmail.com", "frostming@tencent.com"],
    "name": "Frost Ming"
  }
```

So we treat it properly.


DanielNoord:
@Pierre-Sassoulas I just changed the `script/.contributors_aliases.json` format for now to use the correct `key:value` pairs.

Had you thought about how you want to store the information from the `* .......` lines? Will that be in the aliases file? Currently if you regenerate CONTRIBUTORS it will be removed.

Pierre-Sassoulas:
At first I assumed name was unique, but it's not, we have a "Carl" for example. So I use email now and it permits to change the name if we have the email. At least the email should be unique unless someone is not using a real email (it happens too). I'm afraid we're going to have to throw away some contributions with a bad name and no email (or root@none.com). To uniformize the config file I use : ``contributors-txt-normalize-configuration -a scripts/.contributors_aliases.json -o  scripts/.contributors_aliases.json`` (then pre-commit). There a check for unicity of email here : https://github.com/Pierre-Sassoulas/contributors-txt/blob/main/contributors_txt/create_content.py#L78.

DanielNoord:
Ah okay, I'll let you handle this then. No point in having two captains on one ship 😄

PR #5928, opened 2022-03-16
Add new message called duplicate-value

### LINKED ISSUE #5880: Check sets for duplicate entries
### Current problem

Pylint check dictionaries for duplicates, but not sets.

```python
DICTIONARY = {
    1: 2,
    2: 3,
    1: 2,  # duplicate flagged
}

SET = {
    1,
    2,
    1,  # duplicate not flagged
}
```

### Desired solution

Flag duplicate set entries just as with dictionaries.

### Additional context

_No response_

ksaketou:
## Type of Changes

## Description

This PR creates a new error message called `duplicate-value` which identifies duplicate values inside sets.  

Its functionality is similar to the `duplicate-key` message which finds duplicate values in dictionaries.

Closes #5880


nickdrozd:
Tested locally; works! Thanks @ksaketou!

pylint/checkers/base.py
```diff
@@ -1467,6 +1472,19 @@ def visit_dict(self, node: nodes.Dict) -> None:
...
             keys.add(key)
 
+    @utils.check_messages("duplicate-value")
+    def visit_set(self, node: nodes.Set) -> None:
+        """Check duplicate value in set."""
+        values = set()
+        for v in node.elts:
+            if isinstance(v, nodes.Const):
+                value = v.value
+            else:
```
areveny, line 1482: Could we catch some more cases?

Matching `nodes.Name` would pick up:

```
a = 1
b = 2
s = {a, b, b}
```

This would be really useful. Another case could be Tuples of constants, e.g. `{(12, 15), (12, 15), (12)}`.

pylint/checkers/base.py
```diff
@@ -1467,6 +1472,19 @@ def visit_dict(self, node: nodes.Dict) -> None:
                 self.add_message("duplicate-key", node=node, args=key)
             keys.add(key)
 
+    @utils.check_messages("duplicate-value")
+    def visit_set(self, node: nodes.Set) -> None:
```
areveny, line 1476: Can we add a `visit_call` to pick up the `set([1, 2, 2, 3])` initialization syntax?

areveny (changes requested):
This looks pretty good. I think some additions would pick up many more common cases.

Also surfacing the failure on the Pylint presubmit: https://github.com/PyCQA/pylint/runs/5571708705?check_suite_focus=true

```
************* Module pylint.checkers.format
pylint/checkers/format.py:721:23: W0130: Duplicate value '\x0b' in set (duplicate-value)
pylint/checkers/format.py:721:23: W0130: Duplicate value '\x0c' in set (duplicate-value)
```

Which is this:

```
        unsplit_ends = {
            "\v",
            "\x0b",
            "\f",
            "\x0c",
            "\x1c",
            "\x1d",
            "\x1e",
            "\x85",
            "\u2028",
            "\u2029",
        }
```

https://stackoverflow.com/questions/26184100/how-does-v-differ-from-x0b-or-x0c

It seems `\v == \x0b` and `\x0c == \f`. I think we remove duplicates in our code. It can benefit clarity to explicitly write both synonym escape sequences, but that clarification is better in a comment.

pylint/checkers/base.py
DanielNoord: We might want to move that to a follow-up issue. I can see that becoming quite difficult and could inflate the size (and review burden) of this PR.

areveny: Because of inferring iterables?

I was thinking just checking for `set` called on a list literal and then running the logic of `visit_set` on that list would capture a really common pattern and feasible for this PR.

nickdrozd: That sounds like a good check to add, but it's distinct from this one.

If someone writes `set([1, 2, 2, 3])`, they could get a hint to use set notation instead: `{1, 2, 2, 3}`. The next run would then trigger this check and suggest getting rid of the duplicate.

Maybe that could be collapsed to get the optimal result the first time, but I agree that it's outside the scope of this PR.

pylint/checkers/base.py
nickdrozd: Instead of opting in for certain types, what about opting out instead? What kinds of things should be ignored?

pylint/checkers/base.py
areveny: Ah! Okay, my thinking was that `set([items])` and `{items}` are both supported ways of initializing a set. The [Python docs](https://docs.python.org/3/tutorial/datastructures.html#sets) aren't a style guide but are unopiniated on initializing sets: "Curly braces or the [set()](https://docs.python.org/3/library/stdtypes.html#set) function can be used to create sets."

The default check that prefers `{}` over `dict()` is also supported by the docs: "A pair of braces creates an empty dictionary: {}."

So my preference is to support both, but if we want to point people to `{items}` and not extend this PR to `set` I'll resolve this comment.

ksaketou:
> It seems \v == \x0b and \x0c == \f. I think we remove duplicates in our code. It can benefit clarity to explicitly write both 
> synonym escape sequences, but that clarification is better in a comment.  

So regarding that case should we modify `unsplit_sets` by keeping one of the two cases and referring to the other one in a comment? Which one should we keep in comments?

pylint/checkers/base.py
DanielNoord: Yeah, I think we should definitely support it, but in terms of reviewing it's just a lot easier to do that as a follow up.

DanielNoord:
> > It seems \v == \x0b and \x0c == \f. I think we remove duplicates in our code. It can benefit clarity to explicitly write both
> > synonym escape sequences, but that clarification is better in a comment.
> 
> So regarding that case should we modify `unsplit_ends` by keeping one of the two cases and referring to the other one in a comment? Which one should we keep in comments?

Let's keep the unicode one and add `\v` in the comments.

DanielNoord:
Hi @ksaketou,

We have been working on the documentation of the messages that pylint emits. As this PR adds a new message could I ask you to rebase on `main` and add two examples for bad and good code like explained in https://github.com/PyCQA/pylint/issues/5953? That way we immediately help users understand what this new message will do 😄

Pierre-Sassoulas:
Hello @ksaketou, sorry the basic checker got split up on main in #5489 so the conflict resolution might be a little rough when rebasing on main. It's only code moving around so you can copy paste the code you added in the new files that were created.

doc/whatsnew/2.13.rst
```diff
@@ -66,6 +66,10 @@ New checkers
 
   Closes #5670
 
+* Added new message called ``duplicate-value`` which identifies duplicate values inside sets.
```
DanielNoord, line 69: This should be moved to `2.14`.

pylint/checkers/base.py
DanielNoord: @ksaketou This comes down to you wanting to put more time in. I would be fine with merging this as a very bare-bones checker and expanding later.

pylint/checkers/base.py
```diff
@@ -1474,6 +1479,19 @@ def visit_dict(self, node: nodes.Dict) -> None:
...
+        """Check duplicate value in set."""
+        values = set()
+        for v in node.elts:
+            if isinstance(v, nodes.Const):
+                value = v.value
+            else:
+                continue
+            if value in values:
+                self.add_message("duplicate-value", node=node, args=value)
```
DanielNoord, line 1492:
```suggestion
                self.add_message("duplicate-value", node=node, args=value, confidence=HIGH)
```

You'll probably need to important that.

DanielNoord (changes requested):

ksaketou:
Hello @DanielNoord @Pierre-Sassoulas , sorry for the late reply. Unfortunately, I don't currently have enough time to expand the checker's functionality (https://github.com/PyCQA/pylint/pull/5928#discussion_r837243931) . However, I can work on the documentation and the other minor suggested changes above during the following days if that's okay.

ksaketou:
Quick question, should I add the documentation on a separate PR? @DanielNoord

ksaketou:
Did the refactoring affect the test execution at all? because I get errors regarding the `setup.cfg`.  
I suppose this checker now goes into `basic_checker.py`?

Pierre-Sassoulas:
> Did the refactoring affect the test execution at all? because I get errors regarding the setup.cfg.

No it should be the same behavior, only less code per file.

> I suppose this checker now goes into basic_checker.py?

Yes, I checked the diff it was in the BasicChecker class.

> When I add the timeout option into the markers config option the test passes, but I dont think I should modify this file.

I had the same issue, I just fixed it in #6054, it should work if you rebase on latest main

ChangeLog
```diff
@@ -19,6 +19,9 @@ Release date: TBA
 
 * Fix pyreverse diagrams type hinting for classmethods and staticmethods.
 
+* Add new message ``duplicate-value`` for identifying duplicate values inside sets.
```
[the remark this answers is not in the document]
DanielNoord: It should now be removed here 😄

doc/data/messages/d/duplicate-value/details.rst
```diff
@@ -0,0 +1,2 @@
+This error message is emitted when a set contains the same value two or more times.
```
DanielNoord, line 1: Information like this should probably included in the message description in its declaration.

doc/whatsnew/2.13.rst
```diff
@@ -84,6 +84,7 @@ New checkers
 
   Closes #5670
 
+
```
DanielNoord, line 87:
```suggestion
```

DanielNoord (changes requested):
Almost there! Most of this looks good to go in my opinion!

ChangeLog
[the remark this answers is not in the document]
ksaketou: I dont understand the change here

Pierre-Sassoulas (approved):
LGTM, good job @ksaketou !

ChangeLog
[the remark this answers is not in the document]
Pierre-Sassoulas: Resolving in favor of https://github.com/PyCQA/pylint/pull/5928#discussion_r840771737 :)

doc/data/messages/d/duplicate-value/details.rst
DanielNoord: Only need to move this to `pylint/checkers/base/basic_checker.py`

DanielNoord (changes requested):

DanielNoord (approved):

areveny (approved):

PR #5929, opened 2022-03-16
spellcheck will skip the rule names of mypy inline directives

eli88fine:
## Type of Changes

## Description

The spellchecker was a bit overaggressive in flagging the Mypy rule names inside of `ignore` directives.  This allows those to be ignored by the spellchecker.


eli88fine:
@Pierre-Sassoulas - you were the one who took a look at my previous PR regarding the spellchecker (https://github.com/PyCQA/pylint/pull/4330), are you still the right person to review this?

Pierre-Sassoulas (commented):
Sorry for the time it took to review. It looks pretty good already. I wanted to ask you to add a functional tests but I'm struggling to add one myself in #6137. I'll upgrade your own MR when I figure out how to do that.

DanielNoord:
@eli88fine Sorry for letting this rest so long. I think @Pierre-Sassoulas and I agreed that this shouldn't be blocked by our internal changes to how we test the spelling checker.

That said, would you be willing to explore whether this `regex` matching can be done in the same pattern as above? `regex` matching and compiling is often quite expensive (it is on of our main costs during start-up) so I'd like to integrate most patterns where possible.
If not, please let me know and I might have some time to tinker with this myself!

pylint/checkers/spelling.py
```diff
@@ -368,6 +384,7 @@ def _check_spelling(self, msgid, line, line_num):
             starts_with_comment = False
 
         line = _strip_code_flanked_in_backticks(line)
```
DanielNoord, line 370: @eli88fine Sorry should have been clearer. I was referring to this function!

eli88fine: ah, you're saying see if I can somehow combine the regular expressions into a single one.  I understand now.  hmm, I'll take a crack at it and see how it goes. that would definitely be the most complex thing I've ever tried with regex

eli88fine: hmm, I'm worried about accidentally creating a bunch of weird edge cases by trying to combine them.

What if I had a "pre-filter" that just checked the line for either backticks OR a `type: ignore`.  Then if that hit it would run both of the existing regex statements

`if '`' in line or 'type: ignore' in line`:`

DanielNoord: Perhaps you are right. Let's leave as is. I'm not sure if the performance benefits of skipping `regex` matching outweigh the downsides of extra `ifs`.

pylint/checkers/spelling.py
```diff
@@ -185,6 +185,7 @@ def _next(self):
 
 
 CODE_FLANKED_IN_BACKTICK_REGEX = re.compile(r"(\s|^)(`{1,2})([^`]+)(\2)([^`]|$)")
+MYPY_IGNORE_DIRECTIVE_RULE_REGEX = re.compile(r"(\s|^)(type\: ignore\[)([^\]]+)(\].*)")
```
DanielNoord, line 188:
```suggestion
MYPY_IGNORE_DIRECTIVE_RULE_REGEX = re.compile(r"(\s|^)(type\: ignore\[[^\]]+\])(.*)")
```
Schermafbeelding 2022-05-05 om 21 17 30

This creates the following three groups. I don't think we want to check anything in between `type` until the last bracket.

pylint/checkers/spelling.py
```diff
@@ -202,6 +203,21 @@ def replace_code_but_leave_surrounding_characters(match_obj) -> str:
...
 
+def _strip_mypy_ignore_directive_rule(line: str) -> str:
+    """Alter line so mypy rule name is ignored.
+
+    Pyenchant parses anything flanked by spaces as an individual token,
+    so this cannot be done at the individual filter level.
+    """
+
+    def replace_rule_name_but_leave_surrounding_characters(match_obj) -> str:
+        return match_obj.group(1) + match_obj.group(2) + match_obj.group(4)
```
DanielNoord, line 214:
```suggestion
        return match_obj.group(1) + match_obj.group(3)
```

ChangeLog
```diff
@@ -495,6 +495,8 @@ Release date: TBA
 
   Closes #4955
 
+* Disable spellchecking of mypy rule names in ignore directives.
```
DanielNoord, line 498: Could you also add an entry in `doc/whatsnew/2.14`

DanielNoord (changes requested):
Let's get this merged 😄

eli88fine:
trying to follow the timestamps on all the different threads.  Are you saying "leave the code alone, just add an entry in the doc/whatsnew" ?

Or are there actual code changes you think would be good?

DanielNoord:
> trying to follow the timestamps on all the different threads.  Are you saying "leave the code alone, just add an entry in the doc/whatsnew" ?
> 
> 
> 
> Or are there actual code changes you think would be good?

Sorry! Being indecisive today 😅

The last comments were all recent. So let's change the pattern slightly and add the changelog entry!

eli88fine:
regex has been changed and docs updated

ChangeLog
```diff
@@ -1022,6 +1022,8 @@ Release date: 2022-03-24
...
+
```
DanielNoord, line 1026:
```suggestion
  Closes #5929

```

DanielNoord (approved):
One final change and then LGTM!

Pierre-Sassoulas (approved):
Thank you for your contribution @eli88fine ! It's going to be released in 2.14.

PR #5934, opened 2022-03-18
Set up system of code examples and details for message documentation

### LINKED ISSUE #5527: Integrate content from ``pylint-errors`` into the new pages for message

DanielNoord:

## Type of Changes

## Description

This closes #5527. I'd like to create a separate issue to track the progress of migrating the code examples of `pylint-errors`.

This allows for each message to have a `good.py`, `bad.py`, `related.rst` and `details.rst` section. I have added two examples which can be viewed [here](https://pylint--5934.org.readthedocs.build/en/5934/messages/error/yield-inside-async-function.html) and [here](https://pylint--5934.org.readthedocs.build/en/5934/messages/convention/empty-docstring.html). Based on [this](https://vald-phoenix.github.io/pylint-errors/plerr/errors/async/E1700) and [this](https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/C0112). Note that the original example is actually incorrect.
The `bad.py` is written in the same style as the functional test so that at one point we could test against them. For now, I have considered that out of scope. (I also want to find a way to not do that every PR as not to inflate test time).

I think the structure works but let me know what you think!

### DIFF
--- a/doc/data/messages/e/empty-docstring/bad.py
+++ b/doc/data/messages/e/empty-docstring/bad.py
@@ -1,0 +1,2 @@
+def foo():
+    pass  # [emtpy-docstring]
--- a/doc/data/messages/e/empty-docstring/good.py
+++ b/doc/data/messages/e/empty-docstring/good.py
@@ -1,0 +1,2 @@
+def foo():
+    """A dummy description."""
--- a/doc/data/messages/y/yield-inside-async-function/bad.py
+++ b/doc/data/messages/y/yield-inside-async-function/bad.py
@@ -1,0 +1,2 @@
+async def foo():
+    yield from [1, 2, 3]  # [yield-inside-async-function]
--- a/doc/data/messages/y/yield-inside-async-function/good.py
+++ b/doc/data/messages/y/yield-inside-async-function/good.py
@@ -1,0 +1,7 @@
+async def foo():
+    def _inner_foo():
+        yield from [1, 2, 3]
+
+
+async def foo():
+    yield 42
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -24,7 +24,7 @@ PYLINT_MESSAGES_PATH = PYLINT_BASE_PATH / "doc" / "messages"
 """Path to the messages documentation folder."""
 
 PYLINT_MESSAGES_DATA_PATH = PYLINT_BASE_PATH / "doc" / "data" / "messages"
-"""Path to the messages documentation folder."""
+"""Path to the folder with data for the messages documentation."""
 
 MSG_TYPES_DOC = {k: v if v != "info" else "information" for k, v in MSG_TYPES.items()}
 
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -23,6 +23,8 @@ PYLINT_BASE_PATH = Path(__file__).resolve().parent.parent.parent
 PYLINT_MESSAGES_PATH = PYLINT_BASE_PATH / "doc" / "messages"
 """Path to the messages documentation folder."""
 
+PYLINT_MESSAGES_DATA_PATH = PYLINT_BASE_PATH / "doc" / "data" / "messages"
+"""Path to the messages documentation folder."""
 
 MSG_TYPES_DOC = {k: v if v != "info" else "information" for k, v in MSG_TYPES.items()}
 
@@ -32,6 +34,10 @@ class MessageData(NamedTuple):
     id: str
     name: str
     definition: MessageDefinition
+    good_code: str
+    bad_code: str
+    details: str
+    related_links: str
 
 
 MessagesDict = Dict[str, List[MessageData]]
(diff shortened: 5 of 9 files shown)


doc/exts/pylint_messages.py
```diff
@@ -23,6 +23,8 @@
 PYLINT_MESSAGES_PATH = PYLINT_BASE_PATH / "doc" / "messages"
 """Path to the messages documentation folder."""
 
+PYLINT_MESSAGES_DATA_PATH = PYLINT_BASE_PATH / "doc" / "data" / "messages"
+"""Path to the messages documentation folder."""
```
jacobtylerwalls, line 27: Can we distinguish these two docstrings?

jacobtylerwalls (commented):
Great!

jacobtylerwalls (approved):

doc/exts/pylint_messages.py
DanielNoord: Oops, that's a mistake. Fixing!

doc/data/messages/e/empty-docstring/good.py
```diff
@@ -0,0 +1,2 @@
+def foo():
+    """A dummy description."""
```
Pierre-Sassoulas, line 2:
```suggestion
def add(a: Any, b: Any) -> Any:
    """Add a and b."""
    return a + b
```

doc/data/messages/e/empty-docstring/bad.py
```diff
@@ -0,0 +1,2 @@
+def foo():
+    pass  # [emtpy-docstring]
```
Pierre-Sassoulas, line 2: Not related to the actual MR, but I think code example could be 'real' and 'exciting' with animal, vegetal and real concepts or code example to be more rememberable. Also typing by default would influence begginner to use it later.


```suggestion
def add(a: Any, b: Any) -> Any:
    return a + b  # [emtpy-docstring]
```
Foos and bars are traditional and have the benefit of not introducing more concept than the message itself, but also a little boring and I think human learn better with (fun) examples than with generic examples.

doc/data/messages/y/yield-inside-async-function/good.py
```diff
@@ -0,0 +1,7 @@
+async def foo():
+    def _inner_foo():
+        yield from [1, 2, 3]
+
+
+async def foo():
+    yield 42
```
Pierre-Sassoulas, line 7:
```suggestion
async def count_to_three() -> Iterator[int]:
    def _inner_count_to_three() -> Iterator[int]:
        yield from [1, 2, 3]


async def the_answer() -> Iterator[int]:
    yield 42
```

doc/data/messages/y/yield-inside-async-function/bad.py
```diff
@@ -0,0 +1,2 @@
+async def foo():
+    yield from [1, 2, 3]  # [yield-inside-async-function]
```
Pierre-Sassoulas, line 2:
```suggestion
async def count_to_three()  -> Iterator[int]:
    yield from [1, 2, 3]  # [yield-inside-async-function]
```

Pierre-Sassoulas (approved):
LGTM ! Thank you. Glad this is getting some traction again. This is important, yet we have a lot to do for 2.13 already. I made some suggestions about redactional style for the documentation, let me know what you think.

doc/data/messages/e/empty-docstring/good.py
DanielNoord: I mean, what does adding `Any` really provide here? Without any information `mypy` and others will already infer everything as `Any`.

doc/data/messages/e/empty-docstring/bad.py
Pierre-Sassoulas: You're right about "any", maybe we could another example ? (``def do_nothing() -> None:``).

Pierre-Sassoulas: I think you're also right about barebone being a good thing, but I dislike foo/bar for documentation. For a beginner the first thing they have to do is to read https://en.wikipedia.org/wiki/Foobar and learn about our little software programmer internal meme. We can probably make barebone examples that a school children would understand.

DanielNoord: Ah shit, I completely forgot about this.

I'll keep it in mind for future PRs 👍

Pierre-Sassoulas: Don't worry we'll have to review hundred of examples that were in pylint-error, there's no use nitpicking about the style right now 😄

PR #5938, opened 2022-03-19
C2801: New check for manual __magic__ methods

### LINKED ISSUE #5936: Warning when manually invoking __magic__ methods
### Current problem

It is almost always considered a bad practice to manually invoke a `__magic__` method on a class.
```py
# Bad
my_str = my_object.__str__()
# Good
my_str = str(my_object)
```
A warning for this type of issue could be nice and instructive for people who are new to Python.

### Desired solution

A warning is given when Pylint is run on code that invokes `__magic__` methods on a class rather than using the builtin functions for it.

### Additional context

_No response_

jpy-git:
## Type of Changes

## Description

Closes #5936. Adds new checker to detect the use of manual __magic__ methods.

Design decisions:
- I'm only flagging magic methods not attributes since they usually have genuine use cases. e.g. `my_instance_name = x.__class__.__name__` or `my_pkg_version = pkg.__version__`.
- `super().__init__(self, ...)`, `Foo.__init__(self, ...)`, `super().__new__(cls, ...)`, or `Foo.__new__(cls, ...)` are all pretty common use cases where people manually call the magic method (`.__init__(...)` shows up 65 times in functional and `.__new__(...)` shows up 13 times). Therefore I've excluded them from this check.
- [`cls.__subclasses__()`](https://github.com/PyCQA/pylint/blob/2a33b6742ec5cc0bc718ef6cc691ed16684b8cab/tests/functional/r/regression_02/regression_4982.py#L9) and [`.__init_subclass__`](https://github.com/PyCQA/pylint/blob/2a33b6742ec5cc0bc718ef6cc691ed16684b8cab/tests/functional/i/init_subclass_classmethod.py#L6-L8) are cases where there is no alternative to the dunder method so I've also excluded these.

- I've disabled this rule in a couple of the existing tests since they use `__setattr__`, `__get__`, or `__iter__`. Which should be replaced with `setattr(...)`, `.get(...)/[...]`, or `iter(...)` respectively. Happy to update those tests if we'd rather do that instead of ignoring?


jpy-git:
Copying @MiguelGuthridge's comment here for reference:
>One thing that could be worth consideration is whether to use an inclusion list or an exclusion list. I feel like using a whitelist might be better in terms of reducing false positives if there are new __magic__ methods added. It might be also useful to provide the function that should be used instead within the error message, so that people know what to use instead.

It's something I considered and I'm happy to go with maintainer preference on inclusion vs exclusion 😄
Either:
- includelist specific magic methods to guarantee no false positives but risks rule becoming outdated as more are added.
- Identify by dunder naming convention (starts and ends with `__`). Risks a couple false positives on methods similar to `__subclasses__` that don't have an alternative method/function/operator, but we can exclude these as they come up.

As @MiguelGuthridge suggested, I do like the benefit of the includelist in that we could make it a dict like:
```python
includelist = {
    "__lt__": "< operator",
    "__del__": "del function",
    "__repr__": "repr function",
    ...
}
```
and then our warning message can be like:
`"Manually invokes magic method %s. Use %s instead."`
which just fills in the corresponding matched value from includelist: `Manually invokes magic method __str__. Use str function instead.`

jpy-git:
We already have a big list of `__dunder__` methods [here](https://github.com/PyCQA/pylint/blob/0481a7356140bb9506833d93149e1d219a3ec79f/pylint/checkers/utils.py#L125-L232) in the code so I could probably refactor that slightly to make the method name the key and have information about number of args and whether there is an alternative method/function/operator. Would save us having to maintain 2 separate whitelists.

Also that would be useful for a similar issue (#5940) I've just opened about preventing users from creating non-official `__dunder__` methods since we just check is the func def name have a `__dunder__` style name and is it not in the whitelist. 😄

Pierre-Sassoulas:
> I do like the benefit of the includelist in that we could make it a dict

I also prefer the incudelist, false positives are more problematic than false negatives as maintainers. The dict is nice too. I don't think we should try to use the existing  dunder lists as it's going to introduce too much coupling with other checks.

This already looks good but I'm going to review this properly only after 2.13 is released :)

pylint/checkers/magic_methods.py
```diff
@@ -0,0 +1,144 @@
...
+
+    __implements__ = IAstroidChecker
+
+    # Docs: https://docs.python.org/3/reference/datamodel.html#basic-customization
+    # We exclude __init__, __new__, __subclasses__, __init_subclass__,
+    # __set_name__, __class_getitem__, __missing__, __exit__, __await__,
+    # __del__, __aexit__, __getnewargs_ex__, __getnewargs__, __getstate__,
+    # __setstate__, __reduce__, __reduce_ex__
+    # since these either have no alternative method of being called or
+    # have a genuine use case for being called manually.
```
jpy-git, line 23: As mentioned in the comment some of these (`__init__`, `__new__`, ...) have genuine use cases for being manually called.

On the other hand some are just excluded since they don't have an alternative method/function/operator for me to suggest.
e.g. you probably shouldn't be calling the `__del__` destructor directly and something like `__exit__` never needs to be manually called it just gets automatically executed when exiting a context manager.

If you'd like we could just have a message for these like `This is called automatically when exiting a context manager`, so we suggesting they remove it rather than replace it. I just excluded them for now as there could always be some weird unknown use case for these being called that we don't know about so playing it safe wrt false positives.

Easy either way on these ones 😄

jpy-git:
> I also prefer the incudelist, false positives are more problematic than false negatives as maintainers. The dict is nice too. I don't think we should try to use the existing dunder lists as it's going to introduce too much coupling with other checks.

Perfect, I've updated the code now with it's own includelist and custom lint messages 👍 

> This already looks good but I'm going to review this properly only after 2.13 is released :)

Thanks, no rush 😄

DanielNoord:
Hi @jpy-git ,

I haven't gotten around reviewing this, but I wanted to ask you in advance:
We have been working on the documentation of the messages that pylint emits. As this PR adds a new message could I ask you to rebase on `main` and add two examples for bad and good code like explained in https://github.com/PyCQA/pylint/issues/5953? That way we immediately help users understand what this new message will do 😄

CONTRIBUTORS.txt
```diff
@@ -536,12 +536,3 @@ under this name, or we did not manage to find their commits in the history.
 - Mirko Friedenhagen
 - Nicholas Smith
 - Nuzula H. Yudaka (Nuzhuka)
```
Pierre-Sassoulas, line 538:
```suggestion
- Nuzula H. Yudaka (Nuzhuka)
- Pek Chhan
- Peter Hammond
- Pierre Rouleau
- Richard Goodman: simplifiable-if-expression (with Tomer Chachamu)
- Sebastian Ulrich
- Takashi Hirashima
- Thomas Snowden: fix missing-docstring for inner functions
- Wolfgang Grafen
- Yannick Brehon
```

ChangeLog
```diff
@@ -2,13 +2,23 @@
...
+
+
 What's New in Pylint 2.13.0?
 ============================
 Release date: TBA
 
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
+
```
Pierre-Sassoulas, line 21:
```suggestion
```

pylint/checkers/magic_methods.py
```diff
@@ -0,0 +1,147 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
```
Pierre-Sassoulas, line 3:
```suggestion
# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt

```

Pierre-Sassoulas:
I'm going to release 2.13 before reviewing but I wanted to fix merge conflicts caused by #5877 :)

jpy-git:
> As this PR adds a new message could I ask you to rebase on `main` and add two examples for bad and good code like explained in #5953?

Sure thing, will take a look this weekend 😄

jpy-git:
@DanielNoord I've added an example to the docs folder. Lmk if that's enough or if you need more examples there 😄

doc/data/messages/m/manual-magic-methods/bad.py
```diff
@@ -0,0 +1 @@
+var_str = var.__str__() # [manual-magic-methods]
```
Pierre-Sassoulas, line 1:
```suggestion
three = 3.0.__str__()  # [manual-magic-methods]
twelve = "1".__add__("2")  # [manual-magic-methods]

def is_bigger_than_two(x):
    return x.__gt__(2)  # [manual-magic-methods]
```

doc/data/messages/m/manual-magic-methods/good.py
```diff
@@ -0,0 +1 @@
+var_str = str(var)
```
Pierre-Sassoulas, line 1:
```suggestion
three = str(3.0)
twelve = "1" + "2"  # [manual-magic-methods]


def is_bigger_than_two(x):
    return x > 2
```

pylint/checkers/magic_methods.py
```diff
@@ -0,0 +1,148 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
Pierre-Sassoulas, line 1: I think this should be an optional check in pylint/extensions instead. We try to not add opinionated new check so installing a new version is not too painful.

pylint/checkers/magic_methods.py
```diff
@@ -0,0 +1,148 @@
...
+
+    __implements__ = IAstroidChecker
+
+    # Docs: https://docs.python.org/3/reference/datamodel.html#basic-customization
+    # We exclude __init__, __new__, __subclasses__, __init_subclass__,
+    # __set_name__, __class_getitem__, __missing__, __exit__, __await__,
+    # __del__, __aexit__, __getnewargs_ex__, __getnewargs__, __getstate__,
+    # __setstate__, __reduce__, __reduce_ex__
+    # since these either have no alternative method of being called or
+    # have a genuine use case for being called manually.
```
Pierre-Sassoulas, line 27:
```suggestion
    """Check for manual __magic__ method calls.
    
    Docs: https://docs.python.org/3/reference/datamodel.html#basic-customization
    We exclude __init__, __new__, __subclasses__, __init_subclass__,
    __set_name__, __class_getitem__, __missing__, __exit__, __await__,
    __del__, __aexit__, __getnewargs_ex__, __getnewargs__, __getstate__,
    __setstate__, __reduce__, __reduce_ex__
    since these either have no alternative method of being called or
    have a genuine use case for being called manually.
    """

    __implements__ = IAstroidChecker
```

Pierre-Sassoulas (changes requested):
Looks good already, thank you @jpy-git  !

DanielNoord:
@jpy-git Still haven't gotten the time, but just thought about the name of the message. I actually misinterpreted it and only noticed after seeing the test case.

What about `magic-method-call`? Or `manual-magic-method-call`? The plural of `method` feels incorrect and I think it would be good to have something about "calling" in the message name. For some reason I initially interpreted the message name as "manually defining" magic methods.

pylint/checkers/magic_methods.py
DanielNoord: I haven't looked at the actual list yet, but shouldn't we be able to come up with an un-opinionated list of methods? The example of `str()` vs. `__str__()` seems like a non controversial lint message.

jpy-git: I've already excluded all the dunder methods that don't have an alternative way of being called so the remaining ones should all be ones with an obvious alternative function/operator.

doc/data/messages/m/manual-magic-methods/good.py
```diff
@@ -0,0 +1,6 @@
+three = str(3.0)
+twelve = "1" + "2"  # [manual-magic-methods]
```
Pierre-Sassoulas, line 2:
```suggestion
twelve = "1" + "2"
```

pylint/checkers/magic_methods.py
Pierre-Sassoulas: Fair enough !

jpy-git:
> What about `magic-method-call`?

Yeah that sounds much better 👍

Pierre-Sassoulas:
Maybe "manual-dunder-call" ? (I never encountered 'magic' before)

DanielNoord:
> > What about `magic-method-call`?
> 
> Yeah that sounds much better 👍

Or, another thought, `unnecessary-magic-method-call`? This message name kind of implies we should flag all magic method, which we don't want to. So adding unnecessary might be good.

Pierre-Sassoulas:
Another option would be  "consider-using-builtin" ?

jpy-git:
let's compromise on `unnecessary-dunder-call`? 😆

Pierre-Sassoulas:
> unnecessary-dunder-call

Sounds great !

jpy-git:
I've updated all the references to magic with dunder and added the new message name 👍

pylint/checkers/dunder_methods.py
```diff
@@ -0,0 +1,148 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+from typing import TYPE_CHECKING
+
+from astroid import nodes
+
+from pylint.checkers import BaseChecker
+from pylint.interfaces import IAstroidChecker
```
Pierre-Sassoulas, line 10:
```suggestion
from pylint.interfaces import IAstroidChecker, HIGH
```

pylint/checkers/dunder_methods.py
```diff
@@ -0,0 +1,148 @@
...
+    def visit_call(self, node: nodes.Call) -> None:
+        """Check if method being called is an unnecessary dunder method."""
+        if (
+            isinstance(node.func, nodes.Attribute)
+            and node.func.attrname in self.includedict
+        ):
+            self.add_message(
+                "unnecessary-dunder-call",
+                node=node,
+                args=(node.func.attrname, self.includedict[node.func.attrname]),
```
Pierre-Sassoulas, line 143:
```suggestion
                args=(node.func.attrname, self.includedict[node.func.attrname]),
                confidence=HIGH,
```
We can use that confidence as there is not inference.

tests/functional/u/unnecessary/unnecessary_dunder_call.txt
```diff
@@ -0,0 +1,3 @@
+unnecessary-dunder-call:4:10:4:28::Unnecessarily calls dunder method __str__. Use str built-in function.:UNDEFINED
+unnecessary-dunder-call:5:11:5:30::Unnecessarily calls dunder method __add__. Use + operator.:UNDEFINED
+unnecessary-dunder-call:6:10:6:40::Unnecessarily calls dunder method __repr__. Use repr built-in function.:UNDEFINED
```
Pierre-Sassoulas, line 3:
```suggestion
unnecessary-dunder-call:4:10:4:28::Unnecessarily calls dunder method __str__. Use str built-in function.:HIGH
unnecessary-dunder-call:5:11:5:30::Unnecessarily calls dunder method __add__. Use + operator.:HIGH
unnecessary-dunder-call:6:10:6:40::Unnecessarily calls dunder method __repr__. Use repr built-in function.:HIGH
```

pylint/checkers/dunder_methods.py
```diff
@@ -0,0 +1,148 @@
...
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class DunderCallChecker(BaseChecker):
+    """Check for unnecessary dunder method calls."""
+
+    __implements__ = IAstroidChecker
+
+    # Docs: https://docs.python.org/3/reference/datamodel.html#basic-customization
```
Pierre-Sassoulas, line 21: Is there a reason this was reverted from being in the docstring ? I don't mind, but just to make sure other changes were not removed too :)

Pierre-Sassoulas (changes requested):

pylint/checkers/dunder_methods.py
jpy-git: Nice!

pylint/checkers/dunder_methods.py
jpy-git: whoops my bad, messed up when rebasing after the 2.13 release and forgot to add this back in. I've moved it back to the docstring and the other changes are still fine 😄

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Thank you for the new checker and for the fruitful discussion during review @jpy-git :)

jpy-git:
Thanks, I appreciate the review and the great work you guys do on this project 😄

ChangeLog
```diff
@@ -10,6 +10,10 @@ Release date: TBA
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
 
+* Add new check (C2801) for unnecessary dunder method calls.
```
```suggestion
* Add new check ``unnecessary-dunder-call`` for unnecessary dunder method calls.
```

doc/whatsnew/2.14.rst
```diff
@@ -12,6 +12,10 @@ Summary -- Release highlights
 New checkers
 ============
 
+* Add new check (C2801) for unnecessary dunder method calls.
```
DanielNoord, line 15:
```suggestion
* Add new check ``unnecessary-dunder-call`` for unnecessary dunder method calls.
```

pylint/checkers/base_checker.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, linter=None):
 
     def __gt__(self, other):
         """Permit to sort a list of Checker by name."""
-        return f"{self.name}{self.msgs}".__gt__(f"{other.name}{other.msgs}")
+        return f"{self.name}{self.msgs}" > (f"{other.name}{other.msgs}")
```
DanielNoord, line 47: This is uncovered. We might want to do a follow up issue/PR and see if this is necessary anyway.

tests/functional/u/unnecessary/unnecessary_dunder_call.py
```diff
@@ -0,0 +1,44 @@
+# pylint: disable=too-few-public-methods, missing-docstring, undefined-variable, useless-object-inheritance
```
DanielNoord, line 1:
```suggestion
"""Checks for unnecessary-dunder-call."""
# pylint: disable=too-few-public-methods, undefined-variable, useless-object-inheritance
```

DanielNoord (changes requested):
Small changes! Rest looks very good 😄

pylint/checkers/base_checker.py
jpy-git: Noted this in #6045 👍

DanielNoord: Thanks! And thank you for being so proactive these last couple of days! Real breeze of fresh air in the project 😄

DanielNoord (approved):

Pierre-Sassoulas:
Great work, thank you @jpy-git !

MiguelGuthridge:
AYYYYYYY awesome job!!! It's so awesome to see how this all came together!

jacobtylerwalls:
Hi @jpy-git, would you have a chance to look at #6074 ? I would appreciate your input 🙏🏻

PR #5941, opened 2022-03-20
Add missing __magic__ methods to `_SPECIAL_METHODS_PARAMS`

jpy-git:
## Type of Changes

## Description

Noticed this whilst working on #5938.

These magic methods were missing from the `_SPECIAL_METHODS_PARAMS` variable used for several checks so I've added them in:
- [`__subclasses__`](https://docs.python.org/3/library/stdtypes.html?highlight=__subclasses#class.__subclasses__)
- [`__init_subclass__`](https://docs.python.org/3/library/stdtypes.html?highlight=__subclasses#class.__subclasses__)

This variable gets used in these checkers:
- variables.py
- class_checker.py
- special_methods_checker.py

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -356,3 +356,4 @@ contributors:
 - Téo Bouvard
 - Konrad Weihmann
 - Sergey B Kirpichev
+- Joseph Young (jpy-git)
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,8 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.13.rst'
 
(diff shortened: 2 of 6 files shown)


Pierre-Sassoulas (changes requested):
Thank you for opening this merge request. I think we need some functional test cases for the checker that would be affected.

jpy-git:
I've added the functional test for `special_methods_checker.py` to `tests/functional/u/unexpected_special_method_signature.py`

for `variables.py` and `class_checker.py` they're just excluding all the dunder methods from their checks and these two new methods don't really make sense in that context anyway so no need to add those.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #5942, opened 2022-03-20
Add missing messages to docs

jpy-git:
## Type of Changes

## Description

Another quick PR. Noticed that when I ran `tox -e docs` it generated some missing messages to the docs. These are all from recent PRs so would be good to get them added to the docs before the next release.

### DIFF
--- a/doc/messages/messages_list.rst
+++ b/doc/messages/messages_list.rst
@@ -59,6 +59,8 @@ All messages in the error category:
    error/bad-string-format-type.rst
    error/bad-super-call.rst
    error/bidirectional-unicode.rst
+   error/broken-collections-callable.rst
+   error/broken-noreturn.rst
    error/catching-non-exception.rst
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
This file is autogerated before creating the doc it's not always up to date but it does not affect the final published result. I think we could add a little pre-commit hook script to check it's up to date, remove the file entirely or update the comment line 7. (We had this discussion with @DanielNoord during initial implementation).

DanielNoord:
Yeah, I think we discussed we would update the `docs` before each release in order to keep the diffs of other PRs smaller. We could create a hook that checks whether the docs have been built, but it's not a priority for me atm.

Pierre-Sassoulas:
Thank you @DanielNoord I forgot about this: https://github.com/PyCQA/pylint/blob/main/tbump.toml#L32

It's upgraded during releases.

PR #5956, opened 2022-03-24
Add CI test for message documentation

DanielNoord:

## Type of Changes

## Description

Well, I immediately spotted a mistake...

We could also think about moving the message of `empty-docstring` to the `doc_node` attribute of the definition instead of on the definition line but that's for another PR.

Code is based on the functional test runner but I have removed everything that was redundant. I couldn't inherit the class because there was a lot of stuff to handle the `.rc` and `.txt` files. This seemed like a simpler approach.

I did not put this in the normal test suite as I didn't want to unnecessarily inflate local test runs. This is really something that only should be run before merging so the CI seemed a better fit.

### DIFF
--- a/doc/test_messages_documentation.py
+++ b/doc/test_messages_documentation.py
@@ -1,0 +1,110 @@
+"""Functional tests for the code examples in the messages documentation."""
+
+from collections import Counter
+from pathlib import Path
+from typing import Counter as CounterType
+from typing import List, TextIO, Tuple
+
+import pytest
+
+from pylint import checkers
+from pylint.config.config_initialization import _config_initialization
+from pylint.lint import PyLinter
+from pylint.message.message import Message
+from pylint.testutils.constants import _EXPECTED_RE
+from pylint.testutils.reporter_for_tests import FunctionalTestReporter
+
+MessageCounter = CounterType[Tuple[int, str]]
+
+
+def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str, Path]]:
+    """Get all functional tests in the input_dir."""
+    suite: List[Tuple[str, Path]] = []
+
+    for subdirectory in input_dir.iterdir():
+        for message_dir in subdirectory.iterdir():
+            if (message_dir / "good.py").exists():
+                suite.append(
+                    (message_dir.stem, message_dir / "good.py"),
+                )
+            if (message_dir / "bad.py").exists():
+                suite.append(
+                    (message_dir.stem, message_dir / "bad.py"),
+                )
+    return suite
+
+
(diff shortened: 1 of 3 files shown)


doc/test_messages_documentation.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+    return suite
+
+
+TESTS_DIR = Path(__file__).parent.resolve() / "data" / "messages"
+TESTS = get_functional_test_files_from_directory(TESTS_DIR)
+TESTS_NAMES = [f"{t[0]}-{t[1].stem}" for t in TESTS]
+
+
+class LintModuleTest:
```
Pierre-Sassoulas, line 42: Can't we reuse https://github.com/PyCQA/pylint/blob/main/pylint/testutils/lint_module_test.py#L35 ?

doc/test_messages_documentation.py
```diff
@@ -0,0 +1,110 @@
...
+from pylint.config.config_initialization import _config_initialization
+from pylint.lint import PyLinter
+from pylint.message.message import Message
+from pylint.testutils.constants import _EXPECTED_RE
+from pylint.testutils.reporter_for_tests import FunctionalTestReporter
+
+MessageCounter = CounterType[Tuple[int, str]]
+
+
+def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str, Path]]:
```
Pierre-Sassoulas, line 20: We could use https://github.com/PyCQA/pylint/blob/main/pylint/testutils/functional/find_functional_tests.py#L31, right ? ( yield from get_functional_test_files_from_directory if the file is good.py or bad.py ? Maybe we don't even need to check that it's good.py or bad.py)

Pierre-Sassoulas (changes requested):
Looks great already ! I think reusing code would make it easier to maintain.

doc/test_messages_documentation.py
DanielNoord: No, that returns `FunctionalTestFile`:
https://github.com/PyCQA/pylint/blob/80838744c6f739fbf2f49bd1bfb70c98a84531aa/pylint/testutils/functional/test_file.py#L45

That class has a lot of additional methods that do not work with the file structure of the messages documentation. Stuff like `self._parse_options()`.
For the purpose of these tests we only need a tuple of test name and path so I don't think it makes much sense to try and change `FunctionalTestFile` to also allow a completely different directory structure.

doc/test_messages_documentation.py
DanielNoord: No, as that automatically disables some messages (that we might want to test) and can only accept `FunctionalTestFile` as `test_file` while we (I think) only want to pass a simple tuple.

Pierre-Sassoulas (approved):

Issue #5953, opened 2022-03-23
HELP WANTED: Help copy content from ``pylint-errors`` over to our own documentation


DanielNoord:
A longstanding issue of `pylint` has been that there is little documentation about some of the messages and why they are emitted. For some messages the short message that is displayed is not enough to make explicit what needs to be changed or what is considered wrong.
With the closure of #5527 and the merge of #5934 we have now set up a system that allows us to do so!
We have also received the okay from @vald-phoenix to base this additional documentation on their wonderful [pylint-errors](https://github.com/vald-phoenix/pylint-errors) project.

However, now **we need some community help to write this documentation**. If you'd like to help out and help future users of `pylint` understand what `pylint` is telling them, please continue reading.
<details>
Two examples of how the new documentation is supposed to look like can be seen [here](https://pylint.pycqa.org/en/latest/messages/error/yield-inside-async-function.html) and [here](https://pylint.pycqa.org/en/latest/messages/convention/empty-docstring.html).

How to create such pages? The process is quite straightforward.
Within the `doc/data/messages` directory you can find various subdirectories. These are the letters of the alphabet. Within these directories you will find directories for each message that `pylint` can emit. These directories contain the data from which these pages are build.

To help build a new page:

1. Create a new directory for the message you are working on. For example `doc/data/messages/a/abstract-class-instantiated`.
2. Look at the page for the message in the `pylint-errors` repo. For example [here](https://github.com/vald-phoenix/pylint-errors/blob/master/plerr/errors/basic/E0110.md). If the message is not documented yet you will need to come up with examples yourself.
3. Create a `good.py` document in which you include a short code example that shows of good code that won't trigger the message. You can base this on `pylint-error`.
4. Create a `bad.py` document in which you include a short bad code example. **On the line that is supposed to emit the message please add a `# [abstract-class-instantiated]` comment.** See these [examples](https://github.com/PyCQA/pylint/blob/main/doc/data/messages/y/yield-inside-async-function/bad.py).
5. If necessary create a `pylintrc` file. Here you can include configuration options for `pylint` like loading of optional checkers. See `doc/data/messages/c/confusing-consecutive-elif/pylintrc` for example.
6. If necessary create a `details.rst` file. Here you can place additional details in .rst format that will be displayed on the page. Note that some of the details included on the `pylint-error` pages are already included in our `Message` or `Description` sections.
[...]

PR #5960, opened 2022-03-24
Add documentation for messages with assert in their name

DanielNoord:

## Type of Changes

## Description

Code examples based on [pylint-errors](https://github.com/vald-phoenix/pylint-errors) and my own work.

Ref. #5953.

These are a bit silly, but oh well... At least it shows how to avoid them.

### DIFF
--- a/doc/data/messages/a/assert-on-string-literal/good.py
+++ b/doc/data/messages/a/assert-on-string-literal/good.py
@@ -1,2 +1,2 @@
-def test(param):
+def test():
     # Nothing, as an assert of a string literal will always pass


DanielNoord:
> The example are a bit silly here as it's hard to tell what te code should be. It should probably be "nothing" because the assert is useless ? What about `# nothing, the assert was always true`

Yeah I'm going to change it to say something like that.

doc/data/messages/a/assert-on-string-literal/good.py
```diff
@@ -0,0 +1,2 @@
+def test(param):
```
Pierre-Sassoulas, line 1:
```suggestion
def test():
```

Pierre-Sassoulas (approved):

Issue #5361, opened 2021-11-22
False negatives superfluous-parens with more than one superfluous parenthesis


KotlinIsland:
### Current problem

```py
for (i) in [1]:  # superfluous-parens
    print(i)

for ((i)) in [1]:  # no warning
    print(i)
```

### Desired solution

warning on both

jacobtylerwalls:
See [discussion](https://github.com/PyCQA/pylint/pull/5964#discussion_r835280442) -- we might be able to revert most of the changes from #4948 to solve this. Just check #4907 for any legitimate cases.

clavedeluna:
Investigating this and trying to translate the context into an actionable. Is the new consensus that any test case **except  the one with startswith((** in `tests/functional/s/superfluous_parens.py` which contains `((` and does not report `superfluous-parens` should now do so?

PR #5964, opened 2022-03-24
Fix false positive for `superfluous-parens` for `return (a or b) in iterable`

### LINKED ISSUE #5803: False alarm for superfluous-parens
### Bug description

Simple code gives false warning:

```    
def f():
   return (base or dist_name) in ("Debian", "Ubuntu") 
```

As you can see, there are no braces around the return value, but I still get `superfluous-parens   Unnecessary parens after 'return' keyword`

### Configuration

```ini
Default config too.
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
nuitka/utils/Utils.py:107 C0325 superfluous-parens  Unnecessary parens after 'return' keyword
```

### Expected behavior

Should detect the parens usage properly.

### Pylint version

```shell
pylint 2.12.1
astroid 2.9.0
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
```

### OS / Environment

Debian Bullseye, but pip installed pylint

### Additional dependencies

Nothing

jacobtylerwalls:
## Type of Changes

## Description

Fix false positive for ``superfluous-parens`` for patterns like `return (a or b) in iterable`.

Closes #5803


pylint/checkers/format.py
```diff
@@ -382,19 +383,15 @@ def _check_keyword_parentheses(
...
-                    elif not found_and_or and keyword_token != "in":
-                        self.add_message(
-                            "superfluous-parens", line=line_num, args=keyword_token
-                        )
+                    if found_and_or:
+                        return
+                    if keyword_token == "in":
+                        # This special case was added in https://github.com/PyCQA/pylint/pull/4948
+                        # but it could be removed in the future. Avoid churn for now.
+                        return
```
jacobtylerwalls, line 391: @DanielNoord this is here to preserve the special case added in #4948, but when I looked into it, I had trouble figuring out why it was deemed a bugfix. (Looks like unnecessary parentheses to me!) Do you recall the thought behind it?

Pierre-Sassoulas (approved):
LGTM, I'm thinking the changelog should be in 2.13.1 in ChangeLog and in 2.14 in whatsnew if we backport this but let me know what you think.

ChangeLog
```diff
@@ -9,6 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
+* Fix false positive for ``superfluous-parens`` for patterns like
```
DanielNoord, line 12: Might as well put this in `2.13.1`

doc/whatsnew/2.14.rst
```diff
@@ -23,3 +23,8 @@ Extensions
 
 Other Changes
 =============
+
+* Fix false positive for ``superfluous-parens`` for patterns like
```
DanielNoord, line 27: See above.

tests/functional/s/superfluous_parens.py
```diff
@@ -44,8 +44,11 @@ def function_A():
 def function_B(var):
     return (var.startswith(('A', 'B', 'C')) or var == 'D')
 
+def function_C(first, second):
+    return (first or second) in (0, 1)
+
 # TODO: Test string combinations, see https://github.com/PyCQA/pylint/issues/4792
-# Lines 45, 46 & 47 should raise the superfluous-parens message
+# The lines with "+" should raise the superfluous-parens message
```
DanielNoord, line 51: Smart

pylint/checkers/format.py
DanielNoord: Does this still pass:
`D = [x for x in ((3, 4) if 1 > 0 else ((5, 6)))]`

See https://github.com/PyCQA/pylint/issues/4907#issuecomment-907640922

I think sometimes people want to create inner tuples and this was added so we don't flag those.

DanielNoord (changes requested):

pylint/checkers/format.py
jacobtylerwalls: Yes, it still passes now, but I'm essentially suggesting we later revert #4948.

> I think sometimes people want to create inner tuples and this was added so we don't flag those.

I'm not sure what _inner tuple_ means here. `((5,6),)` is a tuple of a tuple. `((5,6))` is superfluous parentheses, so it could just be disabled if people really prefer the extra parentheses. But I think maybe there's a nuance here I'm not getting?

jacobtylerwalls: > Does this still pass:
D = [x for x in ((3, 4) if 1 > 0 else ((5, 6)))]

(This test is still in the suite)

DanielNoord: Hmm perhaps I was wrong to re-allow those.

I thought there might be use cases where you would want to create `Tuple[Tuple[Tuple[int, ...]]]` instead of `Tuple[int, ...]`?

jacobtylerwalls: Sure, but that's `(((5,),),)` rather than `(((5,)))`.

Parentheses increase readability, so I understand why people use them, belt and suspenders and all, but I'm just saying those folks might not want the check enabled if they're in the practice of adding them.

DanielNoord: Hm, I'm fine with reverting then. But we should check #4907 for any legitimate use cases.

jacobtylerwalls: I'll capture this discussion on #5361

DanielNoord:
@jacobtylerwalls Do we understand the coverage report?

jacobtylerwalls:
> @jacobtylerwalls Do we understand the coverage report?

From the edit history of the comment the lines lost coverage when I merged main 10 hours ago, so ... yes? :D  (I don't really have a mental model of how Coveralls handles merges into PR branches.)

DanielNoord (approved):

PR #5968, opened 2022-03-25
Don't emit `raising-bad-type` when there is ambiguity

### LINKED ISSUE #2793: false positive raising-bad-type for value known to be not None
### Steps to reproduce
1. cat test.py
```
class Foo(object):
  def __init__(self):
    self._exc = None

  def fail(self):
    self._exc = Exception("")

  def doit(self):
    if self._exc:
      exc = self._exc
      self._exc = None
      raise exc
```
2. pylint -E test.py
3.

### Current behavior
```
************* Module test
test.py:12:6: E0702: Raising NoneType while only classes or instances are allowed (raising-bad-type)
```

### Expected behavior
pylint should be smart enough not to warn about this.

### pylint --version output
pylint 2.3.1
astroid 2.2.3
Python 3.7.2 (default, Feb 13 2019, 15:31:10) 
[Clang 10.0.0 (clang-1000.11.45.5)]

jacobtylerwalls:
## Type of Changes

## Description

Avoid false positives for `raising-bad-type` when there is ambiguity on the inferred node. (It might _not_ be a bad type!)

Closes #2793


ChangeLog
```diff
@@ -9,6 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
+* Avoid emitting ``raising-bad-type`` when there is inference ambiguity on
```
DanielNoord, line 12: `2.13.1`... 😄

tests/functional/i/invalid/invalid_exceptions/invalid_exceptions_raised.py
```diff
@@ -108,3 +108,21 @@ class Error(Exception):
...
+        try:
+            1 / 0
+        except ZeroDivisionError as zde:
+            self.stored_exception = zde
+
+    def raise_stored_exception(self):
+        if self.stored_exception is not None:
+            exc = self.stored_exception
+            self.stored_exception = None
+            raise exc
```
DanielNoord, line 128: I'm wondering what will happen here if we merge the `is None` constraint PR in `astroid`, but that's an issue for another time!

DanielNoord (approved):

doc/whatsnew/2.14.rst
```diff
@@ -23,3 +23,8 @@ Extensions
 
 Other Changes
 =============
+
+* Avoid emitting ``raising-bad-type`` when there is inference ambiguity on
+  the variable being raised.
+
+  Closes #2793
```
Pierre-Sassoulas, line 30: Should be in 2.13.rst too, right ?

Pierre-Sassoulas (approved):

PR #5984, opened 2022-03-26
Don't emit broken typing errors inside TYPE_CHECKING blocks

cdce8p:
## Description
`broken-noreturn` and `broken-collections-callable` errors should not be emitted inside `if TYPE_CHECKING` blocks. At runtime `TYPE_CHECKING` is false and thus the block is never executed and cannot fail.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -51,6 +51,9 @@ Release date: TBA
 
   Closes #5187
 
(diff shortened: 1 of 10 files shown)


Pierre-Sassoulas (approved):
👍 I like the little refactor too !

PR #5985, opened 2022-03-26
Cleanup of the USER_HOME variable in ``pylint.config``

Pierre-Sassoulas:
## Type of Changes

## Description

Remove what I believe is dead code. There's more to do in ``pylint.config.__init__``  but I'm taking it slow as it's untested and hard to test right now. The refactor aims to make this more testable.

### DIFF
--- a/pylint/config/__init__.py
+++ b/pylint/config/__init__.py
@@ -11,6 +11,7 @@ from datetime import datetime
 from pylint.config.argument import _Argument
 from pylint.config.arguments_manager import _ArgumentsManager
 from pylint.config.configuration_mixin import ConfigurationMixIn
+from pylint.config.environment_variable import PYLINTRC
 from pylint.config.find_default_config_files import (
     find_default_config_files,
     find_pylintrc,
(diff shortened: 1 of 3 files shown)


pylint/config/__init__.py
```diff
@@ -18,27 +19,26 @@
...
     "Option",
     "OptionsManagerMixIn",
     "OptionParser",
     "OptionsProviderMixIn",
     "UnsupportedAction",
+    "PYLINTRC",
+    "USER_HOME",
 ]
 
-USER_HOME = os.path.expanduser("~")
```
Pierre-Sassoulas, line 35: USER_HOME equal ``os.path.expanduser("~")`` so it's never equal to ``~``.

pylint/config/environement_variable.py
```diff
@@ -0,0 +1,7 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+from pylint.config.find_default_config_files import find_pylintrc
```
DanielNoord, line 5: The name of this file is misspelled 😄

DanielNoord (changes requested):
Need to change the name

pylint/config/environement_variable.py
Pierre-Sassoulas: Do you mean there should not be an "s" ?

DanielNoord: Isn't it `environment`?

Pierre-Sassoulas: Ha right, I thought it was about ``find_default_config_files `` !

DanielNoord (approved):

PR #5662, opened 2022-01-11
Fix #5569: Fix a crash in `unused-private-member` when `type(self)` used in bound methods

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #5569

PR #5990, opened 2022-03-26
Fix false negative for `protected-access` on functions

### LINKED ISSUE #5989: Regression protected-access not recognized
```py
class Light:
    @property
    def _light_internal(self) -> set[str]:
        return set()


def func(light: Light) -> None:
    print(light._light_internal)  # should emit 'protected-access' error
```

Error in pylint 2.12.2, not recognized in 2.13.0
```
W0212: Access to a protected member _light_internal of a client class (protected-access)
```

The issue is caused by a change in #5662. In particular the one in `_is_mandatory_method_param` which prevents the `protected-access` check from running.
https://github.com/PyCQA/pylint/blob/556304eb1e54c67940d0f9d0497ffa368b50a22a/pylint/checkers/classes/class_checker.py#L2105-L2122

https://github.com/PyCQA/pylint/blob/556304eb1e54c67940d0f9d0497ffa368b50a22a/pylint/checkers/classes/class_checker.py#L1470-L1476

/CC: @jacobtylerwalls

jacobtylerwalls:
## Type of Changes

## Description

Fixes a false negative regression in #5662.

Closes #5989

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -980,7 +980,7 @@ a metaclass class method.",
                 if attribute.attrname != assign_attr.attrname:
                     continue
 
-                if self._is_type_self_call(attribute.expr):
+                if isinstance(attribute.expr, nodes.Call):
                     continue
 
                 if assign_attr.expr.name in {
@@ -2106,6 +2106,7 @@ a metaclass class method.",
         """Check if nodes.Name corresponds to first attribute variable name.
 
         Name is `self` for method, `cls` for classmethod and `mcs` for metaclass.
+        Static methods return False.
         """
         if self._first_attrs:
             first_attr = self._first_attrs[-1]
@@ -2115,6 +2116,8 @@ a metaclass class method.",
                 node, nodes.FunctionDef
             )
             if closest_func is None:
+                return False
+            if not closest_func.is_bound():
                 return False
             if not closest_func.args.args:
                 return False
--- a/tests/functional/p/protected_access.py
+++ b/tests/functional/p/protected_access.py
@@ -27,3 +27,17 @@ class MC:
 class Application(metaclass=MC):
     def __no_special__(cls):
         nargs = obj._nargs  # [protected-access]
+
+
+class Light:
+    @property
+    def _light_internal(self) -> None:
+        return None
+
+    @staticmethod
+    def func(light) -> None:
+        print(light._light_internal)  # [protected-access]
+
+
+def func(light: Light) -> None:
+    print(light._light_internal)  # [protected-access]
--- a/ChangeLog
+++ b/ChangeLog
@@ -41,6 +41,11 @@ Release date: TBA
 
   Closes #5803
 
+* Fix a false negative regression in 2.13.0 where ``protected-access`` was not
+  raised on functions.
+
+  Fixes #5989
+
 
 What's New in Pylint 2.13.1?
 ============================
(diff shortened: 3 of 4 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -2116,6 +2117,8 @@ def _is_mandatory_method_param(self, node: nodes.NodeNG) -> bool:
             )
             if closest_func is None:
                 return False
+            if not closest_func.is_bound():
+                return False
```
jacobtylerwalls, line 2121: Resolves the false negative.

tests/functional/p/protected_access.py
```diff
@@ -27,3 +27,17 @@ def nargs(self):
...
+
+
+class Light:
+    @property
+    def _light_internal(self) -> set[str]:
+        return set()
+
+    @staticmethod
+    def func(light) -> None:
+        print(light._light_internal)  # [protected-access]
```
jacobtylerwalls, line 39: I added another test.

pylint/checkers/classes/class_checker.py
```diff
@@ -980,7 +980,7 @@ def _check_unused_private_attributes(self, node: nodes.ClassDef) -> None:
                 if attribute.attrname != assign_attr.attrname:
                     continue
 
-                if self._is_type_self_call(attribute.expr):
+                if isinstance(attribute.expr, nodes.Call):
```
jacobtylerwalls, line 983: Resolving the false negative then caused a failure here in another test -- we really just need to test if it's a Call and short-circuit, because `Call` doesn't have a `name` attribute and will crash below.

tests/functional/p/protected_access.py
```diff
@@ -27,3 +27,17 @@ def nargs(self):
 class Application(metaclass=MC):
     def __no_special__(cls):
         nargs = obj._nargs  # [protected-access]
+
+
+class Light:
+    @property
+    def _light_internal(self) -> set[str]:
+        return set()
```
cdce8p, line 35:
```suggestion
    def _light_internal(self):
        return 42
```

`set[str]` requires Python 3.9. Best to just remove it.

cdce8p (commented):
I tested the change against Home Assistant. It does restore the state before the regression. Thanks @jacobtylerwalls for the quick fix!

DanielNoord (approved):
Good to see the speed at which these hot fixes get PRed. Great work from all involved 😄

cdce8p:
> Good to see the speed at which these hot fixes get PRed. Great work from all involved 😄

I completely agree 🚀 From what I've seen, it also looks that the initial release was much more stable than some of the past ones. The primer tests seem to help a lot!

Pierre-Sassoulas:
> the initial release was much more stable than some of the past ones.

Yes, we had average of 5 crashes opened for 100 issues (before primer from pylint 2.6 to 2.11), we just released 2.13 with 400 issues and we had 3 crashes total (one being in pylint_django a downstream library). This is night and day. It will get even better with the github actions to show the change on primer repositories that @DanielNoord has experimented with. We'll have fewer false positives too then.

jacobtylerwalls:
> The primer tests seem to help a lot!

Time permitting this summer I was hoping to look at the baseline functionality. Adding baselines for the primer repos (and triggering PR comments etc) would have probably helped us catch the false negative regressions like this one.

jacobtylerwalls:
> the github actions to show the change on primer repositories that @DanielNoord has experimented with.

wooooo!!! 🕺🏻

DanielNoord:
> > The primer tests seem to help a lot!
> 
> Time permitting this summer I was hoping to look at the baseline functionality. Adding baselines for the primer repos (and triggering PR comments etc) would have probably helped us catch the false negative regressions like this one.

This discussion should probably take place somewhere else, but both for baseline and primer we just need to fix the `json` output. There has been some discussion about that already.
With a good and easily parseable `json` output it is trivial to create a primer for false negatives. The biggest problem (after parsable output) for the primer is getting it to run somewhat decently fast. `home-assistant` is awesome as a primer project but we would need to run the primer over it twice to catch false negatives and that would really inflate CI runs.

Pierre-Sassoulas (approved):

PR #5992, opened 2022-03-26
Added assignment-from-no-return message example

matusvalo:
## Type of Changes

## Description

Closes partially #5953

### DIFF
--- a/doc/data/messages/a/assignment-from-no-return/bad.py
+++ b/doc/data/messages/a/assignment-from-no-return/bad.py
@@ -3,4 +3,3 @@ def add(x, y):
 
 
 value = add(10, 10)  # [assignment-from-no-return]
-print(value)
--- a/doc/data/messages/a/assignment-from-no-return/bad.py
+++ b/doc/data/messages/a/assignment-from-no-return/bad.py
@@ -1,0 +1,6 @@
+def add(x, y):
+    print(x + y)
+
+
+value = add(10, 10)  # [assignment-from-no-return]
+print(value)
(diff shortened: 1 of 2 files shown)


doc/data/messages/a/assignment-from-no-return/bad.py
```diff
@@ -0,0 +1,6 @@
+def add(x, y):
+    print(x + y)
+
+
+value = add(10, 10)  # [assignment-from-no-return]
+print(value)
```
DanielNoord, line 6:
```suggestion
```

doc/data/messages/a/assignment-from-no-return/good.py
```diff
@@ -0,0 +1,6 @@
+def add(x, y):
+    return x + y
+
+
+value = add(10, 10)
+print(value)
```
DanielNoord, line 6:
```suggestion
```

DanielNoord (approved):
Bit nit picky, but let's try to keep this examples as short as possible. For a beginner, every additional line might be interpreted as necessary for the code to work or the message to go away.

matusvalo:
> Bit nit picky, but let's try to keep this examples as short as possible.

I put print to leave example having just one single error - the one which is documented. (if print is missing there is error about not used variable). But this is just matter of taste.

DanielNoord:
> > Bit nit picky, but let's try to keep this examples as short as possible.
> 
> I put print to leave example having just one single error - the one which is documented. (if print is missing there is error about not used variable). But this is just matter of taste.

I agree with the principle but I think it will be unmanageable. For stuff which involves classes we would need to satisfy `too-few-public-methods`, all the `missing-docstring` messages, the `no-self-use` etc.
At that point we might as well write a complete module 😄 

The tester has been specifically set up to only enable the message that is in the name of the directory so any other pylint messages are no issue for these.

Pierre-Sassoulas (approved):

PR #5997, opened 2022-03-27
Added example for bad-except-order

matusvalo:
## Type of Changes

## Description

Closes partially #5953

### DIFF
--- a/doc/data/messages/b/bad-except-order/good.py
+++ b/doc/data/messages/b/bad-except-order/good.py
@@ -1,0 +1,6 @@
+try:
+    print(int(input()))
+except TypeError:
+    raise
+except Exception:
+    raise
--- a/doc/data/messages/b/bad-except-order/bad.py
+++ b/doc/data/messages/b/bad-except-order/bad.py
@@ -1,0 +1,8 @@
+try:
+    print(int(input()))
+except Exception:
+    raise
+except TypeError:  # [bad-except-order]
+    # This block cannot be reached since TypeError exception
+    # is caught by previous exception handler.
+    raise


Pierre-Sassoulas (approved):

doc/data/messages/b/bad-except-order/good.py
```diff
@@ -0,0 +1,6 @@
+try:
+    print(int(input()))
+except TypeError:
+    raise
+except Exception:
+    raise
```
jacobtylerwalls, line 6: This would emit `try-except-raise`, I think, so maybe we should model doing something with the exception?

matusvalo: Based on this discussion https://github.com/PyCQA/pylint/pull/5992#issuecomment-1079794958 we should prefer simplicity/readability over code cleanness.

jacobtylerwalls: That's fair, but I think a flexible approach to this is worth considering, because I suspect we'll get an issue someday "I rewrote my code the way pylint said to but then I got another error".

matusvalo: I leave the decision to @Pierre-Sassoulas  and @DanielNoord

Pierre-Sassoulas: How about a ``pass`` instead of a raise ? Still simple, but no ``try-except-raise`` ?

jacobtylerwalls: I guess it's a losing battle because bare-exception is also in play (probably?). So I'm willing to leave this be 😄

DanielNoord (approved):

PR #5999, opened 2022-03-27
Fix crash on unbalanced tuple unpacking

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #5998

PR #6004, opened 2022-03-27
C2901 & C2902: New checks for unnecessary lambda expression usage

### LINKED ISSUE #5976: New lint: Use `def` instead of lambda expression for bare assignments and calls
### Current problem

Currently these uses of lambda expressions don't raise any lint.
A)
```python
a = lambda x: x
```
(^ Same as E731 in flake8)

B)
```python
(lambda x: x)("hello world")
```
(not flagged by flake8 or pylint)

### Desired solution

A)
```python
a = lambda x: x
```
raises lint to tell user to create a normal function instead of assigning a lambda.
Should be replaced with
```python
def a(x)
    return x
```

B)
```python
y = (lambda x: x.upper())("hello world")
```
raises lint telling user to just execute the lambda body since there's no need to define a function or lambda expression to compute something once.
Should be replaced with
```python
x = "hello world"
y = x.upper()
```
### Additional context

_No response_

jpy-git:
## Type of Changes

## Description

Closes #5976.

Adds new check `unnecessary-lambda-assignment`: `Lambda expression assigned to a variable. Define a function using the "def" keyword instead.`

From [PEP8](https://peps.python.org/pep-0008/):
```
Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier:
# Correct:
def f(x): return 2*x
# Wrong:
f = lambda x: 2*x
```

This is equivalent to `E731` in flake8 but additionally addresses several false negatives that are missed by that implementation: https://github.com/PyCQA/pycodestyle/issues/1061

Also adds a new check `unnecessary-direct-lambda-call`: `Lambda expression called directly. Execute the expression inline instead.`
i.e. `a**2` not `(lambda x: x**2)(a)`
This prevents a weird redundant usage of lambda functions that I've seen some junior devs do before.


tests/functional/u/unnecessary/unnecessary_lambda_assignment.py
```diff
@@ -0,0 +1,34 @@
...
+    lambda x: x,  # [unnecessary-lambda-assignment]
+    lambda y: y,  # [unnecessary-lambda-assignment]
+    lambda z: z,  # This isn't assigned so don't flag.
+)
+# Example below is currently commented out due to
+# https://github.com/PyCQA/pylint/issues/5998
+# N.B. should raise 'unnecessary-lambda-assignment'
+# I've just removed the inline comment for now as
+# it's confusing the tester whilst commented out!
+# a, b, c = lambda x: x, lambda y: y
```
jpy-git, line 30: Will uncomment this test case once #5999 gets merged in

jpy-git:
There were a lot of existing tests using that I had to add ignores to in that last commit, so look at the commit before that for easier reviewing

pylint/checkers/lambda_expressions.py
```diff
@@ -0,0 +1,81 @@
...
+            'Define a function using the "def" keyword instead.',
+            "unnecessary-lambda-assignment",
+            "Used when a lambda expression is assigned to variable "
+            'rather than defining a standard function with the "def" keyword.',
+        ),
+        "C2902": (
+            "Lambda expression called directly. Execute the expression inline instead.",
+            "unnecessary-direct-lambda-call",
+            "Used when a lambda expression is directly called "
+            "rather than executing it's contents inline.",
```
jacobtylerwalls, line 36:
```suggestion
            "rather than executing its contents inline.",
```

pylint/checkers/lambda_expressions.py
```diff
@@ -0,0 +1,81 @@
...
+        ):
+            self.add_message(
+                "unnecessary-lambda-assignment",
+                node=node.value,
+            )
+        elif isinstance(node.targets[0], nodes.Tuple) and isinstance(
+            node.value, (nodes.Tuple, nodes.List)
+        ):
+            # Iterate over tuple unpacking assignment elements and
+            # see if any lambdas are assinged to a variable.
```
jacobtylerwalls, line 54:
```suggestion
            # see if any lambdas are assigned to a variable.
```

jacobtylerwalls (commented):
This is hilariously unlikely to happen, but what do you think about attempting to catch this case? (It doesn't emit a warning even if you remove the disable):
https://github.com/PyCQA/pylint/blob/edf753e6864224ac0144c0ac12c8d89baa93e097/tests/functional/a/assignment/assignment_expression.py#L44

jpy-git:
@jacobtylerwalls I've extended the check to flag assignment via named expressions as well 👍 
Currently the tests for python<=3.7 are failing because walrus operators were only introduced in 3.8. Is there a way to only run some lines/files for certain versions?

Pierre-Sassoulas:
> Currently the tests for python<=3.7 are failing because walrus operators were only introduced in 3.8. Is there a way to only run some lines/files for certain versions?

Yes, look at a .rc files in `tests/functional/`` there should be examples of use for ``min_version``

jpy-git:
Thanks!

jacobtylerwalls (approved):

DanielNoord:
@Pierre-Sassoulas Not sure if this is an option, but as a collaborator I can re-run actions. Can we allow this for contributors as well? That might make this more workable.

tests/functional/l/lambda_use_before_assign.py
```diff
@@ -1,4 +1,5 @@
 """https://www.logilab.net/elo/ticket/18862"""
+# pylint: disable=unnecessary-lambda-assignment
 from __future__ import print_function
```
DanielNoord, line 3:
```suggestion

from __future__ import print_function
```

tests/functional/u/undefined/undefined_variable.py
```diff
@@ -1,6 +1,7 @@
 # pylint: disable=missing-docstring, multiple-statements, useless-object-inheritance, import-outside-toplevel
 # pylint: disable=too-few-public-methods, no-init, no-self-use, bare-except, broad-except
 # pylint: disable=using-constant-test, import-error, global-variable-not-assigned, unnecessary-comprehension
+# pylint: disable=unnecessary-lambda-assignment
```
DanielNoord, line 4:
```suggestion
# pylint: disable=unnecessary-lambda-assignment

```

tests/functional/u/unnecessary/unnecessary_direct_lambda_call.py
```diff
@@ -0,0 +1,4 @@
+"""Test unnecessary direct calls to lambda expressions."""
+# pylint: disable=undefined-variable,line-too-long
+y = (lambda x: x**2 + 2*x + 1)(a)  # [unnecessary-direct-lambda-call]
```
DanielNoord, line 3:
```suggestion

y = (lambda x: x**2 + 2*x + 1)(a)  # [unnecessary-direct-lambda-call]
```

tests/functional/u/unnecessary/unnecessary_lambda_assignment.py
```diff
@@ -0,0 +1,29 @@
+"""Test assignment of lambda expressions to a variable."""
+# pylint: disable=unbalanced-tuple-unpacking,undefined-variable,line-too-long
+a = lambda x: x  # [unnecessary-lambda-assignment]
```
DanielNoord, line 3:
```suggestion

a = lambda x: x  # [unnecessary-lambda-assignment]
```

tests/functional/u/unnecessary/unnecessary_lambda_assignment.py
```diff
@@ -0,0 +1,29 @@
+"""Test assignment of lambda expressions to a variable."""
```
DanielNoord, line 1: Could you add the following test case:
```python
DICT = {1: lambda x: x, 2: lambda x: x + 1}

for key, value in DICT.items():
    print(value(key))
```

Similar patterns are reasons why I use the `lambda x: x` pattern and I want to make sure we don't accidentally regress and flag this in the future.

DanielNoord (changes requested):

Pierre-Sassoulas:
> Can we allow this for contributors as well?

I don't think so, but the trick to open/close the issue should do most of the time.

jpy-git:
is there a reason why we don't increase the `timeout-minutes` settings in the yaml file?

Pierre-Sassoulas:
From what I understand the pipeline that are stuck are stuck because the runner used by github has only one core and we try to use two. So it will fail even if we wait one hour. 15mn permit to free the worker sooner and have overall better availability.

tests/functional/u/unnecessary/unnecessary_lambda_assignment.py
jpy-git: Yeah this is fine, I've added it in to the unit test 😄

Pierre-Sassoulas (approved):
LGTM, but I think it would be a nice check to inaugurate the false positive github action primer when it's available. It's a default checker and it's going to be run in a lot of code once it lands in pylint.

DanielNoord:
@Pierre-Sassoulas @jpy-git For some reason the latest merge of `main` broke this. Could either of you look into this?

After that, I think we can go ahead and merge this as we shouldn't let this be blocked by the eventual update to the `primer` imo.

Pierre-Sassoulas:
Sure let me fix this, I should have merged that a long time ago. I think the issue is only with the checker number that has been claimed by another new checker since.

Pierre-Sassoulas:
pypy has column/line variation compared to cpython. This is something we do not control and we're probably never going to change the parsing to have the right column in both. Should we still take the column/line info for under PYPY ?

DanielNoord:
> @DanielNoord pypy has column/line variation compared to cpython. This is something we do not control and we're probably never going to change the parsing to have the right column in both. Should we still take the column/line into account under PYPY ?

Let's ignore this test on `PyPy`. This is a very rare (and only) case where the `col_offset` (and not `end_col_offset`) differ. Actually I think the column offset of `PyPy` is more "correct", but that's just my opinion :smile. I don't think we should stop checking `col_offset` on `PyPy` all together as this is the only case where it differs.

DanielNoord:
Thanks @jpy-git for contributing the new checker and thanks to all reviewers 🎉

PR #6009, opened 2022-03-28
Add documentation and code examples for ``consider-using-with``

DudeNr33:

## Type of Changes

## Description

Good and bad examples for ``consider-using-with``.
I decided to stick with the most obvious example. 
The original PR and list of stdlib callables that are checked is based on [this blog post](https://johnlekberg.com/blog/2020-10-11-ctx-manage.html).
Should we include that in ``related.rst``, or should we stick to the official documentation only?

Ref: #5953


Pierre-Sassoulas (approved):
I think the article you used would be great in the related links.

DudeNr33:
Thanks for the feedback, added it.

PR #6010, opened 2022-03-28
Split checks and tests in CI

DanielNoord:

## Type of Changes

## Description

By splitting these we can easily tell the CI not to run the tests on any PR that only adds message data to the codebase.

This should significantly reduce the number of runners these PRs use up.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,0 +1,180 @@
+name: Checks
+
+on:
+  push:
+    branches:
+      - main
+      - 2.*
+  pull_request: ~
+
+env:
+  # Also change CACHE_VERSION in the other workflows
+  CACHE_VERSION: 5
+  DEFAULT_PYTHON: 3.8
+  PRE_COMMIT_CACHE: ~/.cache/pre-commit
+
+jobs:
+  prepare-base:
+    name: Prepare base dependencies
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      python-key: ${{ steps.generate-python-key.outputs.key }}
+      pre-commit-key: ${{ steps.generate-pre-commit-key.outputs.key }}
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.0.0
+        with:
+          fetch-depth: 0
+      - name: Set up Python ${{ env.DEFAULT_PYTHON }}
+        id: python
+        uses: actions/setup-python@v3.0.0
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON }}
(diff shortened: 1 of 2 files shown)


DanielNoord:
I have also added concurrency to the checks workflow as I don't think they need to keep running after a new commit has been pushed.

.github/workflows/checks.yaml
```diff
@@ -0,0 +1,184 @@
...
+  CACHE_VERSION: 5
+  DEFAULT_PYTHON: 3.8
+  PRE_COMMIT_CACHE: ~/.cache/pre-commit
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  prepare-base:
```
Pierre-Sassoulas, line 21: Shouldn't prepare-base be in its own file as it's used in checks as well as in tests ?

.github/workflows/checks.yaml
```diff
@@ -0,0 +1,184 @@
...
+          restore-keys: |
+            ${{ runner.os }}-pre-commit-${{ env.CACHE_VERSION }}-
+      - name: Install pre-commit dependencies
+        if: steps.cache-precommit.outputs.cache-hit != 'true'
+        run: |
+          . venv/bin/activate
+          pre-commit install --install-hooks
+
+  pre-commit:
+    name: pre-commit
```
Pierre-Sassoulas, line 82: This is in fact pylint, the pre-commit jobs is something else and cannot run pylint because it does not install dependencies that pylint need.
```suggestion
  pylint:
    name: pylint
```

.github/workflows/tests.yaml
```diff
@@ -1,186 +1,21 @@
...
 
 on:
   push:
     branches:
       - main
       - 2.*
-  pull_request: ~
+  pull_request:
+    paths-ignore:
+      - doc/data/messages/**
```
Pierre-Sassoulas, line 10: 👍 It will save a lot of time as you said.

Pierre-Sassoulas (changes requested):
That's pretty nice, I have some nitpicks.

.github/workflows/checks.yaml
DanielNoord: It is isn't used in tests 😄 

I thought so as well, but it isn't. Which makes this separation so easy.

.github/workflows/checks.yaml
DanielNoord: Ah I messed up `astroid` and `pylint` here. I believe in `astroid` there are other steps in this as well. I'll change it.

Pierre-Sassoulas (approved):

Issue #5224, opened 2021-10-27
Checker for `TypeVar` name and first argument


DanielNoord:
### Current problem

Maybe more an idea for a separate checker. `TypeVar`, `NewType`, and `ParamSpec` have special name requirements: The first argument should match the variable name. Something like this is discouraged
```py
T = TypeVar("T_Something")
```

Would be nice if we could also detect it, too.

_Originally posted by @cdce8p in https://github.com/PyCQA/pylint/pull/5221#discussion_r737677291_

### Desired solution

Basically:

```python
T = TypeVar("T", ...) # good
K = TypeVar("T", ...) # bad
K_contra = TypeVar("T", ...) # bad
```

jpy-git:
_From #6011_
```python
T = TypeVar('A')
```
Suggested name: `typevar-name-mismatch`
Suggested message: `TypeVar must be assigned to a variable named "A"`

PR #6015, opened 2022-03-28
New check C0106: typevar-double-invariance

### LINKED ISSUE #5895: Create ``typevar-double-variance`` checker
### Current problem

The following code should raise a ``typevar-double-variance`` warning.

```python
T = TypeVar("T", covariant=True, contravariant=True) # [typevar-double-variance]
```

### Desired solution

Some test cases that can be used:

```python
T_co = TypeVar("T", covariant=True, contravariant=True) # [typevar-double-variance]
T_co = TypeVar("T", covariant=True, contravariant=False)
T_contra = TypeVar("T", covariant=True, contravariant=True) # [typevar-double-variance]
T_contra = TypeVar("T", covariant=False, contravariant=True)
T = TypeVar("T", covariant=False, contravariant=False)
```

### Additional context

_No response_

jpy-git:
## Type of Changes

## Description

Closes #5895. Adds a check to ensure that `TypeVar cannot be both covariant and contravariant`.


jpy-git:
For context here are the changes for #5224 that will build on top of the work in this PR: https://github.com/jpy-git/pylint/pull/1

pylint/checkers/base/name_checker/checker.py
```diff
@@ -150,6 +158,12 @@ class NameChecker(_BasicChecker):
             "do not require a suffix. The message is also emitted when invariant "
             "variables do have a suffix.",
         ),
+        "C0106": (
```
DanielNoord, line 161: Could you use another number? The `0106` is already emitted as `E0106`. We have a check for reusing numbers, but for these lower numbers there are already conflicting numbers. (There is also `W0106`) Therefore, it is turned off for `0106`.

pylint/checkers/base/name_checker/checker.py
```diff
@@ -545,42 +560,60 @@ def _assigns_typevar(node: Optional[nodes.NodeNG]) -> bool:
...
-                    suggest_name = f"{re.sub('_co$', '', name)}_contra"
-                    self.add_message(
-                        "typevar-name-incorrect-variance",
-                        node=node,
-                        args=(name, "contravariant", suggest_name),
-                        confidence=interfaces.INFERENCE,
+                variance = (
+                    TypeVarVariance.covariant
+                    if variance
+                    not in (
```
DanielNoord, line 578: This feels so nitpicky, but in the GitHub `diff` viewer the current code is a little unreadable imo:
Can we move the check for `TypeVarVariance.double_variant` to the if that checks the `kw.arg`? I think we can do it as the first check and skip the complete if statement if we already encounter a double_variant.

tests/functional/t/typevar_naming_style_default.py
```diff
@@ -16,12 +16,12 @@
 GoodBoundNameT = TypeVar("GoodBoundNameT", bound=int)
 
 # Some of these will create a RunTime error but serve as a regression test
-T = TypeVar(  # [typevar-name-incorrect-variance]
+T = TypeVar(  # [typevar-double-variance]
```
DanielNoord, line 19: I wondered what we should do in this case. I don't really like how we suppress the `name-incorrect` warning here. Ie., if a user removes on of the variances they would need to do a new pylint run to see that name is still incorrect.
For users that rely on CI to run `pylint` this can be frustrating as it might take some time for that message to reach them.
Can we think of a way to display both messages?

DanielNoord (changes requested):
Thanks @jpy-git! Let me know if you have any questions about my comments.

I do really like the use of the `Enum`. I never use those myself and I learned something new here 😄

tests/functional/t/typevar_naming_style_default.py
Pierre-Sassoulas: I agree we should emit both messages.

pylint/checkers/base/name_checker/checker.py
```diff
@@ -545,42 +560,60 @@ def _assigns_typevar(node: Optional[nodes.NodeNG]) -> bool:
...
+                node=node,
+                confidence=interfaces.INFERENCE,
+            )
+        elif variance == TypeVarVariance.covariant and not name.endswith("_co"):
+            suggest_name = f"{re.sub('_contra$', '', name)}_co"
+            self.add_message(
+                "typevar-name-incorrect-variance",
+                node=node,
+                args=(name, "covariant", suggest_name),
+                confidence=interfaces.INFERENCE,
```
Pierre-Sassoulas, line 604: Is it really inference ? I think this is the kind of check where we don't need it because we check the name directly. (I don't see where we inferred something either). I'm not doing suggestions everywhere in case I'm wrong.

```suggestion
                confidence=interfaces.HIGH,
```

Pierre-Sassoulas (changes requested):
I think @DanielNoord remarks can be done together, instead of refactoring the existing messages we can add the ``typevar-double-variance`` check on top of it. As far as I know pylint display all the messages at once even if fixing one would also fix the others, so it would be counter-intuitive to have a new message appear after fixing a double variance.

pylint/checkers/base/name_checker/checker.py
DanielNoord: No it is. We use inference to determine that we're dealing with a typevar and not with another type of name.

pylint/checkers/base/name_checker/checker.py
jpy-git: will do, FWIW the checks immediately above this line (C0103-C0105) also clash with E and W error codes

tests/functional/t/typevar_naming_style_default.py
jpy-git: I actually removed this message intentionally because it's not really appropriate to suggest a correct name in these double invariance cases since we don't know which type of variance they're intending.

Referring to current main branch:
image

The only reason we were suggesting "T_co" is just because the old code saw `covariant=True` and immediately returned, but you can't tell here if the user actually wants "T_contra"?

These examples below aren't necessarily wrong names (actually in context they're likely correct). We need the double-variance to be addressed first before we can say whether they are wrong or not and therefore why distract from the user with a false positive wrong name?
```python
T_co = TypeVar("T_co", covariant=True, contravariant=True)
T_contra = TypeVar("T_contra", covariant=True, contravariant=False)
```

DanielNoord: I understand that issue, but I don't like how we need two runs to fix all linting issues.

Can we do something like `Type variable name does not reflect variance.` and then pass `"T" is covariant, use "T_co" instead` as arg to the message if we know the expected variance?

jpy-git: Yeah i'm happy to do the generic message approach in this case then 👍 

(just noting details below for future ref)
- generic message prompts the user to consider fixing the typevar name as well as double-variance. Strictly speaking they would still have to run the linter a second time after fixing double-variance to get a suggestion on what exactly to change the name to, but users familiar with the naming convention will likely know to check/fix the name if needed after the first generic message.
```python
T = TypeVar("T", covariant=True, contravariant=True)
```

- Wording of `Type variable name does not reflect variance.` is correct here, even if they do probably want this name. If they fix the double-invariance and the name turns out to be correct then the message wouldn't show up in subsequent runs so no overhead of prompting the user to consider potential bad naming.
```python
T_co = TypeVar("T_co", covariant=True, contravariant=True)
```

Pierre-Sassoulas (approved):
LGTM, the new message with more detail is very nice.

pylint/checkers/base/name_checker/checker.py
```diff
@@ -545,46 +560,67 @@ def _assigns_typevar(node: Optional[nodes.NodeNG]) -> bool:
...
-                    suggest_name = f"{re.sub('_co$', '', name)}_contra"
-                    self.add_message(
-                        "typevar-name-incorrect-variance",
-                        node=node,
-                        args=(name, "contravariant", suggest_name),
-                        confidence=interfaces.INFERENCE,
-                    )
-                return
+            if variance == TypeVarVariance.double_variant:
+                pass
```
DanielNoord, line 575:
```suggestion
                break
```

I guess?

tests/functional/t/typevar_naming_style_default.txt
```diff
@@ -1,6 +1,10 @@
-typevar-name-incorrect-variance:11:0:11:21::"Type variable ""GoodNameWithoutContra"" is contravariant, use ""GoodNameWithoutContra_contra"" instead":INFERENCE
-typevar-name-incorrect-variance:19:0:19:1::"Type variable ""T"" is covariant, use ""T_co"" instead":INFERENCE
-typevar-name-incorrect-variance:24:0:24:8::"Type variable ""T_contra"" is covariant, use ""T_co"" instead":INFERENCE
+typevar-name-incorrect-variance:11:0:11:21::"Type variable name does not reflect variance. ""GoodNameWithoutContra"" is contravariant, use ""GoodNameWithoutContra_contra"" instead":INFERENCE
+typevar-double-variance:19:0:19:1::TypeVar cannot be both covariant and contravariant:INFERENCE
+typevar-name-incorrect-variance:19:0:19:1::Type variable name does not reflect variance%s:INFERENCE
+typevar-double-variance:23:0:23:4::TypeVar cannot be both covariant and contravariant:INFERENCE
+typevar-name-incorrect-variance:23:0:23:4::Type variable name does not reflect variance%s:INFERENCE
```
DanielNoord, line 5: The %s here doesn't look correct? Can we remove it?

DanielNoord (changes requested):
Small changes! Thanks for taking the time to also improve the other message!

pylint/checkers/base/name_checker/checker.py
jpy-git: This should be `pass`. The reason makes more sense if you look at the follow up PR https://github.com/jpy-git/pylint/pull/1/files

Basically to check the typevar name we need to scan all keyword args since a user could feasibly write
```python
_T = TypeVar(covariant=True, contravariant=True, name="_S")
```
If we broke on discovering double variance then we wouldn't find the name in the last kwarg 😄

tests/functional/t/typevar_naming_style_default.txt
jpy-git: Good spot, that should be filled by an empty string so let me take a look at what's going on there!

DanielNoord: Thanks for that fix. Let's see if that crashes any other tests 😄

DanielNoord (approved):

pylint/lint/pylinter.py
```diff
@@ -1567,7 +1567,7 @@ def _add_one_message(
             self.stats.by_msg[message_definition.symbol] = 1
         # Interpolate arguments into message string
         msg = message_definition.msg
-        if args:
+        if args is not None:
```
jpy-git, line 1570: @DanielNoord RE this [comment](https://github.com/PyCQA/pylint/pull/6015#discussion_r841270789):

This is a classic Falsey bug. When a function takes `None` as it's default arg then it's always best to compare the argument to `None` otherwise you risk catching other Falsey objects. In this case it was unintentionally catching the empty string I passed 😄

Pierre-Sassoulas: I lost more than two days of work to 0 and 0.0 being falsey in the last year, I send my heartfelt condolences

PR #6016, opened 2022-03-28
Add examples for ``await-outside-async`` and fix activation of message

DudeNr33:

## Type of Changes

## Description

This adds examples and link to the relevant PEP for ``await-outside-async``. 
The new test case for checking the message docs actually found a bug (which cost me more time to find than I'd be willing to admit 😀 ): 
Because the relevant ``visit_await`` method was implemented in the wrong checker class (there are two in the module), it was not possible to disable all messages and then just enable ``await-outside-async`` again.
As both the impact and the fix of this are so minor I decided to directly fix this in this PR and only add a ChangeLog entry, without separate issue or whatsnew entry.
Let me know if you'd rather want to split this into two PRs.

Ref: #5953

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -26,6 +26,7 @@ Release date: TBA
 
   Closes #1555
 
+* Fix bug where specifically enabling just ``await-outside-async`` was not possible.
 
 ..
   Insert your changelog randomly, it will reduce merge conflicts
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas:
> Because the relevant visit_await method was implemented in the wrong checker class (there are two in the module), it was not possible to disable all messages and then just enable await-outside-async again.

Sounds like something that is supposed to take some time to find out 😄 Thanks for the debug !

PR #6029, opened 2022-03-29
Include ``testing_pylintrc`` in source and wheel distributions

### LINKED ISSUE #6028: Make sure to package the ``pylintrc`` from testutils
### Bug description

We have discussed this in the ``pylint-django`` repo but I haven't got the time to get to it.

If somebody else knows how to do this please go ahead. I'm no expert in `setuptools`.

Basically, there is a pylintrc file in the testutils that should be packaged with `pylint` as the `testutils` depend on it at runtime.

jpy-git:
## Type of Changes

## Description

Closes #6028. `testing_pylintrc` wasn't being included in package distributions. 
This should fix running tests via `tox` (this is where I encountered it) and any uses of `pylint.testutils`

image

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,7 +37,9 @@ What's New in Pylint 2.13.4?
 ============================
 Release date: TBA
 
+* Include ``testing_pylintrc`` in source and wheel distributions.
 
+  Closes #6028
 
 
 What's New in Pylint 2.13.3?
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -128,6 +128,10 @@ Extensions
 Other Changes
 =============
 
+* Include ``testing_pylintrc`` in source and wheel distributions.
+
+  Closes #6028
+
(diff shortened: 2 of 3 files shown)


DanielNoord (approved):
Ah, it was that easy...

Thanks @jpy-git

Pierre-Sassoulas (commented):
Copy pasting the comment I put on the wrong issue: 

There's another solution: it's to generate the default value at runtime in the testutil if the file do not exists. As those tests class are used only by downstream library maintainer this file is very seldom required so not adding it to the packaging at all will be better 99,95% of the time.

DanielNoord:
> Copy pasting the comment I put on the wrong issue:
> 
> There's another solution: it's to generate the default value at runtime in the testutil if the file do not exists. As those tests class are used only by downstream library maintainer this file is very seldom required so not adding it to the packaging at all will be better 99,95% of the time.

The file is relatively small so won't add much. I think having it as a separate file makes the code much cleaner.
If we really want to worry about distrubution size we should make all `testutils` optional.

Pierre-Sassoulas (approved):

Issue #5462, opened 2021-12-03
Provide an auto-upgrade option / migration tool for pylint configurations


Pierre-Sassoulas:
### Current problem

There are breaking changes that could impact the configuration file and make our live as maintainer easier. Asking the user to upgrade their configuration files make most breaking chance impossible. I we can upgrade the configuration file automatically it's easier to make the configuration consistent and easier to maintain. For example a migration to pyproject.toml only instead of maintaining setup.cfg, pylintrc (ini) and pyrpoejct.toml.

### Desired solution

New subcommand ``pylint --auto-upgrade-conf`` to upgrade the configuration file.

### Additional context

https://github.com/PyCQA/pylint/issues/4741#issuecomment-984887726

The new framework for configuration file would help to implement this greatly, we would add an auto-upgraded expected result.

Pierre-Sassoulas:
The auto-upgrade should remove pylint messages that do not exists anymore, or that can't be raised for interpreter below  ``py-version`` from disable/enable.

Pierre-Sassoulas:
The command that generate the toml config is doing that based on the current configuration. So we have a basis to improve upon. It's also going to need to have some tests before being deployed, and some specification after that to provide in-place upgrade.

Pierre-Sassoulas:
Required for breaking changes like #6870. I'm going to start slow and build on it.

DanielNoord:
Don't forget to integrate this in `pylint-config`!

Pierre-Sassoulas:
I checked pylint-config, I think we used it to be able to reuse all of pylint options in order to generate the config from the existing one. There's a lot of complexity we still have following the refactor from optparse (pre-parsing etc.). It seems it would be a lot simpler to start from scratch and have only an argument for input conf / output conf  / (maybe target pylint version ? Though I'd prefer not over-engineering the shit out of this upgrade tool and simply make all the upgrade for the latest pylint version at time of release).  What do you think ? Is there an advantage in using the common parsing I don't see ?

DanielNoord:
[...]

PR #6032, opened 2022-03-29
Use the ``argparse`` config handler in ``imports.py``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -7,7 +7,7 @@
 import collections
 import copy
 import os
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Set, Tuple, Union
 
 import astroid
 from astroid import nodes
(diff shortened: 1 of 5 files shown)


DanielNoord:
Retracted this PR as I noticed that I need to fix some other issues first before tackling this.

Most notably, parsing of the configuration files. Turns out that's quite _special_.

Pierre-Sassoulas:
We don't have a lot of tests for configuration right now but we have a conf file test framework so we could add more tests and do a full refactor to make it less *special*

DanielNoord:
Yeah, one thing I didn't notice until now is that the configuration parser is based on the `ini` format and `.toml` types get converted to that type. However, that loses the rich-format `toml` can have and requires additional validation steps.

I think I'm just going to write it from the ground up just like I did with this arguments processing. That makes it much easier to keep types etc. correct.

Pierre-Sassoulas:
We can make the assumption that in the end we'll use toml and ``pyproject.toml`` as the main configuration file long term. There's also this issue: #5462 that might be easier to do than accommodating 4 different formats, and would also make sure that we reach "the end" faster.

DanielNoord:
Yeah I'm thinking of using tomlib as the main parser and using the ini parser as secondary parser.

Pierre-Sassoulas:
> using tomlib as the main parser

Sounds great 👍 What about a standalone migration script from ``pylintrc`` / ``setup.cfg`` to ``pyproject.toml`` so we can separate the ini parser right now and eventually remove it easily by just removing the migration script ?

DanielNoord:
> > using tomlib as the main parser
> 
> 
> 
> Sounds great 👍 What about a standalone migration script from ``pylintrc`` / ``setup.cfg`` to ``pyproject.toml`` so we can separate the ini parser right now and eventually remove it easily by just removing the migration script ?

I think that should be part of the issue you previously mentioned. 
Note that it was a conscious decision not to add a toml writer (only a parser) to the stdlib. So we'll need to investigate which toml writer is most widely used.

_Our previous choice for the parser was incorrect as tomli has "won" over toml._

Issue #5604, opened 2021-12-28
Next step in pylint's gamification and user experience


Pierre-Sassoulas:
This is a follow up to #2399 and relate to #5403. It was discussed in detail in https://github.com/PyCQA/pylint/pull/4538#issuecomment-853355683

Right now when starting to use pylint the score can be really bad and we might be dumping thousands of warning on users. This is demoralizing, even for code quality oriented persons and hurting adoption for legacy projects.

> The experience I have right now in starting to use pylint on a code base is overwhelming, I have to use ``pylint .|grep -o "([a-z\-]*)$"|cut -d "(" -f2|sort|uniq|sed s/\)/,/`` to disable all the existing warning and start from somewhere. Then I need to think hard before enabling them again to know where to start. The way I see it the pylint guide would add the disable automatically, maybe in pylintrc, maybe in code directly, then it would rank the warning and tell you exactly what to do next and provide a feeling of progression with level up and "encouragements". Congratulation you do not have any error level warning, Congratulation the code follow SOLID principles, Congratulation we do not have any performance improvement to provide, Congratulation your code did not regress after upgrading to pylint 2.10.1, you were already that good, Congratulation you've got a perfect code according to pylint's opinion etc. A lot of variables to follow that could be in the pylintrc.

The first step is Implementing baseline so only new message are raised : #5403. This way pylint can be adopted easily on a legacy codebase.

Then I think we could create a pylint "guide". What I mean by that is upgrading the pylintrc step by step to fix the legacy messages starting by the more significant one (error, then vicious warning, then warning, then important convention, then opinionated convention, then adding extensions...). This way there would be a clear path to clean code and nothing overwhelming along the way. This would permit to add "levels" and "achievements" and improve gamification. The score do not need to be bad, it can be a good score for this level (i.e. good jobs there are no errors, do you want to increase the quality of your code ?). Pylint would be friendlier and it would help user to increase their code base quality.

Hornwitser:
The experience I had using Pylint on an existing codebase was 1500 messages being reported and no way to sort through it, no way to determine what has changed or compare with previous runs and no way to filter by the messages or modules I want to work on. I ended up [writing an html interface to display the reports based on the JSON output that has all of these features.](https://www.hornwitser.no/x/discord/lint/#ref=lint-2&cmp=860d6a9&types=all&modules=discord.client&hide=reports)

nickdrozd:
[...]

PR #6034, opened 2022-03-29
Remove the concept of checker priority

DanielNoord:

## Type of Changes

## Description

I don't really see what this does? There is this issue https://github.com/PyCQA/pylint/issues/229.
But it seems more like a "let's make a function that uses something that is already there" than something that actually adds something. When is it ever useful to have some checker be executed 50ms earlier than another?

I'd vote for removal. I don't consider this a breaking change because you can still define `priority` without it breaking anything.
And we can easily revert this commit if something does depend on it.

### DIFF
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -26,6 +26,7 @@ New checkers
 Removed checkers
 ================
 
+* The concept of checker priority has been removed.
 
 Extensions
 ==========
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -26,7 +26,6 @@ New checkers
 Removed checkers
 ================
 
-* The concept of checker priority has been removed.
 
 Extensions
 ==========
@@ -34,6 +33,8 @@ Extensions
 
 Other Changes
 =============
+
+* The concept of checker priority has been removed.
 
 * Fix false negative for ``no-member`` when attempting to assign an instance
   attribute to itself without any prior assignment.
--- a/pylint/config/option_manager_mixin.py
+++ b/pylint/config/option_manager_mixin.py
@@ -87,13 +87,7 @@ class OptionsManagerMixIn:
 
     def register_options_provider(self, provider, own_group=True):
         """Register an options provider."""
-        assert provider.priority <= 0, "provider's priority can't be >= 0"
-        for i, options_provider in enumerate(self.options_providers):
-            if provider.priority > options_provider.priority:
-                self.options_providers.insert(i, provider)
-                break
-        else:
-            self.options_providers.append(provider)
+        self.options_providers.append(provider)
         non_group_spec_options = [
             option for option in provider.options if "group" not in option[1]
         ]
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -5,7 +5,6 @@
 import collections
 import contextlib
 import functools
-import operator
 import os
 import sys
 import tokenize
@@ -214,7 +213,6 @@ class PyLinter(
     __implements__ = (interfaces.ITokenChecker,)
 
     name = MAIN_CHECKER_NAME
-    priority = 0
     level = 0
     msgs = MSGS
     # Will be used like this : datetime.now().strftime(crash_file_path)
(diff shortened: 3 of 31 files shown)


DanielNoord:
> Yeap, I don't see what the use case could be. Did you search for the rational in the history ? Nice cleanup imo.

No I can't access the bitbucket issue, so I can't see a discussion in the original PR. After working on this some more, I feel like a lot of the `optparse` code was directly copied from logilab "patterns". There seems to be some functionality that was just copied from other projects. That might be what happened here as well, and then later somebody actually made this functionality work.
I wonder if there is any real functionality though.

@jacobtylerwalls Thoughts?

doc/whatsnew/2.14.rst
```diff
@@ -26,6 +26,7 @@ New checkers
 Removed checkers
 ================
 
+* The concept of checker priority has been removed.
```
jacobtylerwalls, line 29: I would move this to other changes.

pylint/lint/pylinter.py
```diff
@@ -974,10 +971,6 @@ def prepare_checkers(self):
             messages = {msg for msg in checker.msgs if self.is_message_enabled(msg)}
             if messages or any(self.report_is_enabled(r[0]) for r in checker.reports):
                 needed_checkers.append(checker)
-        # Sort checkers by priority
-        needed_checkers = sorted(
-            needed_checkers, key=operator.attrgetter("priority"), reverse=True
-        )
```
jacobtylerwalls, line 980: Is anyone aware of pylint ever having a `failfast` mode? That's all I can think of for the usefulness of sorting checkers. If you want the more important checkers to run first when failing fast.

pylint/config/option_manager_mixin.py
```diff
@@ -87,13 +87,7 @@ def reset_parsers(self, usage=""):
...
     def register_options_provider(self, provider, own_group=True):
         """Register an options provider."""
-        assert provider.priority <= 0, "provider's priority can't be >= 0"
-        for i, options_provider in enumerate(self.options_providers):
-            if provider.priority > options_provider.priority:
-                self.options_providers.insert(i, provider)
-                break
-        else:
-            self.options_providers.append(provider)
+        self.options_providers.append(provider)
```
jacobtylerwalls, line 90: Option provider priority is different than checker priority, right? We should document that as going away also. 

I'm a little nervous that a behavior change could be hiding here, but maybe you can reassure me :D. I tried to trace upstream from this where this could be used and noticed `ConfigurationMixin` (subclass of `OptionsManagerMixIn`) is used in pyreverse, which is way outside my involvement.

Also, apparently checkers are option providers? So could removing the checker priority alter the status quo for what order checker-provided-optoins get overwritten?  But if this is part of removing complexity during the argparse migration that's fine...

jacobtylerwalls (commented):
I couldn't quickly assure myself this isn't being used somehow. But I can definitely see it helping the argparse migration to reduce complexity. I just am not familiar with in what sense a checker provides options, unfortunately.

pylint/config/option_manager_mixin.py
DanielNoord: Well, since each check is a option provider they share the same priority. Thus, we're basically doing the same checks here as we're doing at that other place.

I think the only two "functions" that come from these priorities is that:
1) Checkers will be run sooner, but when is that ever needed?
2) Options _could_ be overwritten by providers with higher priority. But then again, why would we allow that? That creates a lot of mess and duplicate options have actually caused me multiple headaches before.

pylint/lint/pylinter.py
Pierre-Sassoulas: I don't think we have a failfast mode but you're right it's the probably a use case. We could want to order by warnings significance instead of checkers later (i.e. to raise most significant messages first) see #5604

jacobtylerwalls: We don't have a failfast mode, but this wasn't really documented or implemented, so we can always put one back if we want it and document it.

jacobtylerwalls (approved):
As we discussed there could be subtle option-overwriting behavior changes with this change, but the changes are probably for the better and not anything we're going to support after migrating to argparse. Thanks for clearing out old cruft!

Issue #5817, opened 2022-02-17
False positive used-before-assignment with try-except


TTKuLa:
### Bug description

The following code will raise a `used-before-assignment` error:

```
def some_method():
    try:
        some_list = [e for e in (1, 2, 3) if e]  # Using variable 'e' before assignment!
    except Exception as e:
        pass
```

### Command used

```shell
pylint used_before_assignment.py -E
```

### Pylint output

```shell
************* Module used_before_assignment
used_before_assignment.py:3:45: E0601: Using variable 'e' before assignment (used-before-assignment)
```

### Expected behavior

Expected no error, since the variable `e` in the try-except and the `e` in the list comprehension use different scopes.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AMD64)]
```

Pierre-Sassoulas:
Probably fixed in https://github.com/PyCQA/pylint/pull/5812 (duplicate of #5112 ?)

DanielNoord:
I checked. Still emitting on `main` 😢

jacobtylerwalls:
No, it's unrelated! Pretty much the only except-related case in that method that I had not altered yet :-D

PR #6040, opened 2022-03-29
Remove assumption of direct parentage for `used-before-assignment` homonym handling for try/except blocks and filtered comprehensions

### LINKED ISSUE #6035: `used-before-assignment` false positive with list comprehension and try/except
### Bug description

Somewhat similarly to #5965,

```python
def github_upload(filename, all_assets):
    assets = [asset for asset in all_assets if asset.name == filename]

    try:
        pass
    except ValueError:
        asset = assets[0]
        asset.delete()
```

leads to an `used-before-assignment` false-positive for the second line, yet `asset` is always defined there.

I bisected this to bd55b27d41542e3ca1f031f986b6151f6cac457f, part of #5402 ("Fix #4761: Emit used-before-assignment where single assignment only made in except blocks")

### Configuration

_No response_

### Command used

```shell
pylint -E repro1.py
```

### Pylint output

```shell
************* Module repro1
repro1.py:2:47: E0601: Using variable 'asset' before assignment (used-before-assignment)
```

### Expected behavior

No error.

### Pylint version

```shell
pylint 2.13.2
astroid 2.11.2
Python 3.8.12 (default, Mar  3 2022, 18:19:28) 
[GCC 11.2.0]
```

(environment in which I originally encountered the bug - output above is from a git install of both astroid and pylint)

Also happens with latest git main (astroid b18c7828, pylint 0bc45e90). Does *not* happen on pylint 2.12.2 and astroid 2.9.3 which I used previously.

### OS / Environment

Archlinux

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Fixes #6035

The prior fixes for extremely similar scenarios in #5586 and #5817 still assumed that the comprehension would be the direct parent of the name being checked. The reported bug violated that assumption: there, the parent is the attribute, not the comprehension.

EDIT: I don't think this is the _ideal_ place to be doing this logic, but it plugs a hole. Long-term, see Daniel's [comment](https://github.com/PyCQA/pylint/issues/5586#issuecomment-1010155471):
> The fix should thus probably be in _check_consumer and how it handles the ifs of comprehensions.

### DIFF
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -575,3 +575,9 @@ Other Changes
   separated (e.g. by parens).
 
   Closes #5769
+
+* Fix false positive regression in 2.13.0 for ``used-before-assignment`` for
+  homonyms between variable assignments in except blocks and earlier attribute
+  access in the test of a filtered comprehension.
+
+  Closes #6035
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -575,9 +575,3 @@ Other Changes
   separated (e.g. by parens).
 
   Closes #5769
-
-* Fix false positive regression in 2.13.0 for ``used-before-assignment`` for
-  homonyms between variable assignments in except blocks and earlier attribute
-  access in the test of a filtered comprehension.
-
-  Closes #6035
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2480,11 +2480,10 @@ class VariablesChecker(BaseChecker):
 
     @staticmethod
     def _has_homonym_in_comprehension_test(node: nodes.Name) -> bool:
-        """
-        Return True if `node`'s frame contains a comprehension having
+        """Return True if `node`'s frame contains a comprehension employing an
+        identical name in a test.
 
-        an "if" test employing an identical name. The use of the name
-        in the if test could come in several levels of nesting:
+        The name in the test could appear at varying depths:
 
         Examples:
             [x for x in range(3) if name]
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -638,7 +638,7 @@ scope_type : {self._atomic.scope_type}
         )
         if (
             closest_comprehension
-            and node.frame().parent_of(closest_comprehension)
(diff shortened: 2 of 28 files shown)


doc/whatsnew/2.13.rst
```diff
@@ -575,3 +575,9 @@ Other Changes
   separated (e.g. by parens).
 
   Closes #5769
+
+* Fix false positive regression in 2.13.0 for ``used-before-assignment`` for
+  homonyms between variable assignments in except blocks and earlier attribute
+  access in the test of a filtered comprehension.
+
+  Closes #6035
```
Pierre-Sassoulas, line 583:
```suggestion
```
If the regression was from 2.13.0 we can avoid adding this in the whatsnew.

Pierre-Sassoulas (approved):

pylint/checkers/variables.py
```diff
@@ -629,13 +629,25 @@ def get_next_to_consume(self, node: nodes.Name) -> Optional[List[nodes.NodeNG]]:
...
+        # And is not part of a test in a filtered comprehension
+        # Examples:
+        #   [e for e in range(3) if e] <--- followed by except e:
+        #   [e for e in range(3) if e.num == 1] <--- followed by except e:
+        closest_comprehension = utils.get_node_first_ancestor_of_type(
+            node, nodes.Comprehension
+        )
+        if (
+            closest_comprehension
+            and node.frame().parent_of(closest_comprehension)
```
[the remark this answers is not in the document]
jacobtylerwalls: Yes, thanks for catching. I think the deprecation warning should have been louder to catch this use. I hope to follow up.

pylint/checkers/variables.py
```diff
@@ -2494,6 +2478,31 @@ def _has_homonym_in_upper_function_scope(
...
+        Return True if `node`'s frame contains a comprehension having
+
+        an "if" test employing an identical name. The use of the name
+        in the if test could come in several levels of nesting:
+
+        Examples:
+            [x for x in range(3) if name]
+            [x for x in range(3) if name.num == 1]
+            [x for x in range(3)] if call(name.num)]
+        """
```
jacobtylerwalls, line 2493: `pydocstringformatter` won't let me remove this blank line. i'm github-screengrabbing-it with this comment.

PR #6052, opened 2022-03-30
Some refactors of configuration file discovery

DanielNoord:

## Type of Changes

## Description

Sorry for this being a bit all over the place...

These are some basic refactors all related to `read_config_file` to make the parameters that get passed to that function a little more logical. That makes it easier to write a new function that doesn't need to handle some of the non-sensical parameters.

I also removed the configuration file discovery out of `PyLinter` into `Run`. Since the configuration is done in `Run` it makes no sense to do that discovery in `PyLinter`.

### DIFF
--- a/pylint/config/config_initialization.py
+++ b/pylint/config/config_initialization.py
@@ -4,7 +4,7 @@
 
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING, List, Optional, Union
+from typing import TYPE_CHECKING, List, Union
 
 from pylint import reporters
 from pylint.utils import utils
@@ -18,18 +18,23 @@ def _config_initialization(
     args_list: List[str],
     reporter: Union[reporters.BaseReporter, reporters.MultiReporter, None] = None,
     config_file: Union[None, str, Path] = None,
-    verbose_mode: Optional[bool] = None,
+    verbose_mode: bool = False,
 ) -> List[str]:
     """Parse all available options, read config files and command line arguments and
     set options accordingly.
     """
+    # Convert config_file in a Path and string
+    config_file_path = Path(config_file) if config_file else None
+    config_file = str(config_file) if config_file else None
+
     # Set the current module to the configuration file
     # to allow raising messages on the configuration file.
-    linter.set_current_module(linter.config_file)
+    linter.set_current_module(config_file)
 
-    # Read the config file. The parser is stored on linter.cfgfile_parser
+    # Read the configuration file
     try:
-        linter.read_config_file(config_file=config_file, verbose=verbose_mode)
+        # The parser is stored on linter.cfgfile_parser
+        linter.read_config_file(config_file=config_file_path, verbose=verbose_mode)
     except OSError as ex:
         print(ex, file=sys.stderr)
         sys.exit(32)
--- a/pylint/config/option_manager_mixin.py
+++ b/pylint/config/option_manager_mixin.py
@@ -11,7 +11,7 @@ import os
 import sys
 from pathlib import Path
 from types import ModuleType
-from typing import Dict, List, Optional, TextIO, Tuple, Union
+from typing import Dict, List, Optional, TextIO, Tuple
 
 from pylint import utils
 from pylint.config.man_help_formatter import _ManHelpFormatter
@@ -61,8 +61,7 @@ def _patch_optparse():
 class OptionsManagerMixIn:
     """Handle configuration from both a configuration file and command line options."""
 
-    def __init__(self, usage, config_file=None):
-        self.config_file = config_file
+    def __init__(self, usage):
         self.reset_parsers(usage)
         # list of registered options providers
         self.options_providers = []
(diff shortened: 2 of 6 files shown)


DanielNoord:
Initial test failure was due to incomplete call to `_config_initialization`. However, since that is private it shouldn't matter too much.

pylint/config/option_manager_mixin.py
```diff
@@ -245,25 +244,23 @@ def load_provider_defaults(self):
...
-        if config_file is None:
-            config_file = self.config_file
-        if config_file is not None:
-            config_file = os.path.expandvars(os.path.expanduser(config_file))
-            if not os.path.exists(config_file):
-                raise OSError(f"The config file {config_file} doesn't exist!")
-
-        use_config_file = config_file and os.path.exists(config_file)
-        if use_config_file:
+        if config_file:
```
Pierre-Sassoulas, line 253:
```suggestion
        if not config_file:
            if verbose:
                msg = "No config file found, using default configuration"
                print(msg, file=sys.stderr)
            return
       ...
```
Pre-ma-ture re-turn 🎉  This one actually made you buffer 30 lines and 5+ ifs/else that even github hide as it's too long to show

pylint/config/config_initialization.py
```diff
@@ -18,18 +18,23 @@ def _config_initialization(
...
     config_file: Union[None, str, Path] = None,
-    verbose_mode: Optional[bool] = None,
+    verbose_mode: bool = False,
 ) -> List[str]:
     """Parse all available options, read config files and command line arguments and
     set options accordingly.
     """
+    # Convert config_file in a Path and string
+    config_file_path = Path(config_file) if config_file else None
+    config_file = str(config_file) if config_file else None
```
Pierre-Sassoulas, line 28: Why do we need both ?  
```suggestion
    config_file = Path(config_file) if config_file else None
```
The we cast to string when/if required ?

pylint/config/config_initialization.py
```diff
@@ -18,18 +18,23 @@ def _config_initialization(
...
+
     # Set the current module to the configuration file
     # to allow raising messages on the configuration file.
-    linter.set_current_module(linter.config_file)
+    linter.set_current_module(config_file)
```
Pierre-Sassoulas, line 32:
```suggestion
    linter.set_current_module(str(config_file))
```

Or we could make ``set_current_module`` accepts Path too.

Pierre-Sassoulas (changes requested):
I started to modify with the online editor but this is risky as it's impossible to launch tests 😄

pylint/config/config_initialization.py
DanielNoord: I opted to do it like this to give some flexibility in later PRs and keep those diffs smaller.

pylint/config/option_manager_mixin.py
DanielNoord: I do this in the final version but didn't want to touch the rest of the code too much. It's more about accepting different types of parameters correctly than actually improving this method.

pylint/config/option_manager_mixin.py
```diff
@@ -274,17 +271,16 @@ def read_config_file(self, config_file=None, verbose=None):
                         sect = sect[len("pylint.") :]
                     if not sect.isupper() and values:
                         parser._sections[sect.upper()] = values
+
         if not verbose:
             return
-        if use_config_file:
-            msg = f"Using config file {os.path.abspath(config_file)}"
+        if config_file and config_file.exists():
+            msg = f"Using config file {str(config_file)}"
```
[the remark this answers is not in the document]
DanielNoord: Do `Paths` automatically get converted to string in f-strings?

pylint/config/config_initialization.py
Pierre-Sassoulas: All right !

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
Pierre-Sassoulas: If we don't change the initialization it's already a string and not a Path, but yes:
```python
>>> from pathlib import Path
>>> f"{Path('tests').resolve()}"
'/home/pierre/pylint/tests'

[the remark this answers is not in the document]
Pierre-Sassoulas: Everything get cast to string in f-strings :)

pylint/config/option_manager_mixin.py
Pierre-Sassoulas: Fair enough !

pylint/config/option_manager_mixin.py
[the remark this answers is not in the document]
DanielNoord: Yeah I wondered if it was `'/home/pierre/pylint/tests'` or `Path('/home/pierre/pylint/tests')`. I think the `repr()` has the `Path()` in it.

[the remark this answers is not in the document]
Pierre-Sassoulas: Ha yes I did the test because of that too. We probably printed lists of Paths before

Pierre-Sassoulas:
Sorry for butchering the history with my edits and revert.

PR #6055, opened 2022-03-30
Create ``_ConfigurationFileParser`` class

DanielNoord:

## Type of Changes

## Description

Blocked by #6052.

Last commit introduces a new configuration parser. After some debating with myself I have opted to retain the idea of transforming all rich types into strings. `argparse` can only handle strings. We could inject the rich types into the `Namespace` object but I decided against it. Stuff like that is what makes the current configuration handling so difficult to understand. Let's separate things clearly: a file parser parses a file and an argument parser parses arguments.

The commit also introduces other changes that followed from these changes.

One major change is that we will now allow configuration in `.toml` files without the sections. I have not yet implemented this for the `ini` files (and wonder if I ever will) but in `.toml` you can now do:
```toml
[tool.pylint]
missing-member-hint = true

[tool.pylint."TYPECHECK"]
missing-member-hint = true

[tool.pylint."I-ORDER-MY-SECTIONS-THE-WAY-I-WANT-TO"]
missing-member-hint = true
```

### DIFF
--- a/pylint/config/config_file_parser.py
+++ b/pylint/config/config_file_parser.py
@@ -44,8 +44,6 @@ class _ConfigurationFileParser:
     @staticmethod
     def _parse_toml_value(value: Any) -> str:
         """Parse rich toml types into strings."""
-        if isinstance(value, bool):
-            return "True" if value else "False"
         if isinstance(value, list):
             return ",".join(value)
         return str(value)
--- a/pylint/config/config_file_parser.py
+++ b/pylint/config/config_file_parser.py
@@ -1,0 +1,94 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""Configuration file parser class."""
+
+import configparser
+import os
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Dict, Optional
+
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    import tomli as tomllib
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class _ConfigurationFileParser:
+    """Class to parse various formats of configuration files."""
+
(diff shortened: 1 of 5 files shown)


pylint/config/option_manager_mixin.py
```diff
@@ -259,8 +259,8 @@ def read_config_file(
             if config_file.suffix == ".toml":
                 try:
                     self._parse_toml(config_file, parser)
-                except tomllib.TOMLDecodeError as e:
-                    self.add_message("config-parse-error", line=0, args=str(e))  # type: ignore[attr-defined]
+                except tomllib.TOMLDecodeError:
+                    pass
```
DanielNoord, line 263: This is now raised by the new parser and would lead to double messages otherwise.

pylint/config/config_file_parser.py
```diff
@@ -0,0 +1,94 @@
...
+        for section in parser.sections():
+            for opt in parser[section]:
+                config_content[opt] = parser[section][opt]
+        return config_content
+
+    @staticmethod
+    def _parse_toml_value(value: Any) -> str:
+        """Parse rich toml types into strings."""
+        if isinstance(value, bool):
+            return "True" if value else "False"
```
Pierre-Sassoulas, line 48:
```suggestion
```
```
>>> str(True)
'True'
>>> str(False)
'False'
```

pylint/config/config_file_parser.py
```diff
@@ -0,0 +1,94 @@
...
+            with open(file_path, mode="rb") as fp:
+                content = tomllib.load(fp)
+        except tomllib.TOMLDecodeError as e:
+            self.linter.add_message("config-parse-error", line=0, args=str(e))
+            return {}
+
+        try:
+            sections_values = content["tool"]["pylint"]
+        except KeyError:
+            return {}
```
Pierre-Sassoulas, line 65:
```suggestion
        if content.get("tool", {}).get("pylint") is None:
            return {}
```

Pierre-Sassoulas (approved):
LGTM !

pylint/config/config_file_parser.py
DanielNoord: I prefer the other syntax. I'll let you decide and commit 😄

Pierre-Sassoulas: It's not worth having to re-run the pipeline 😄

PR #6056, opened 2022-03-30
More tests for config file finding, handle malformatted setup.cfg

Pierre-Sassoulas:
## Type of Changes

## Description

I added some tests for our config finding functions. I'm wondering if there's not something to do regarding ``find_pylintrc``. It's none if there is no pylintrc but there could be setup.cfg or pyproject.toml with valid configuration.

```python
def find_pylintrc() -> Optional[str]:
    """Search the pylint rc file and return its path if it finds it, else return None."""
    for config_file in find_default_config_files():
        if config_file.endswith("pylintrc"):
            return config_file
    return None
```    

Imo it could be nice to have a function : "recover configuration from wherever" and return a configuration object (which is why I made tests for the two important function to reach this goal). What do you think @DanielNoord ?

### DIFF
--- a/tests/config/test_find_default_config_files.py
+++ b/tests/config/test_find_default_config_files.py
@@ -1,0 +1,68 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+from pathlib import Path
+
+import pytest
+
+from pylint.config.find_default_config_files import _cfg_has_config, _toml_has_config
+
+
+@pytest.mark.parametrize(
+    "content,expected",
+    [
+        ["", False],
+        ["(not toml valid)", False],
+        [
+            """
+[build-system]
+requires = ["setuptools ~= 58.0", "cython ~= 0.29.0"]
+""",
+            False,
+        ],
+        [
+            """
+[tool.pylint]
+missing-member-hint = true
+""",
+            True,
+        ],
+    ],
+)
+def test_toml_has_config(content: str, expected: bool, tmp_path: Path) -> None:
+    """Test that a toml file has a pylint config."""
+    fake_toml = tmp_path / "fake.toml"
+    with open(fake_toml, "w", encoding="utf8") as f:
+        f.write(content)
+    assert _toml_has_config(fake_toml) == expected
+
+
+@pytest.mark.parametrize(
+    "content,expected",
+    [
+        ["", False],
+        [
+            """
+[metadata]
+name = pylint
+""",
+            False,
+        ],
+        [
+            """
+[metadata]
+name = pylint
+
+[pylint.messages control]
+disable = logging-not-lazy,logging-format-interpolation
+""",
+            True,
+        ],
+    ],
+)
+def test_cfg_has_config(content: str, expected: str, tmp_path: Path) -> None:
+    """Test that a cfg file has a pylint config."""
+    fake_cfg = tmp_path / "fake.cfg"
+    with open(fake_cfg, "w", encoding="utf8") as f:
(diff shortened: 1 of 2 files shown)


tests/config/test_find_default_config_files.py
```diff
@@ -0,0 +1,68 @@
...
+    fake_toml = tmp_path / "fake.toml"
+    with open(fake_toml, "w", encoding="utf8") as f:
+        f.write(content)
+    assert _toml_has_config(fake_toml) == expected
+
+
+@pytest.mark.parametrize(
+    "content,expected",
+    [
+        ["", False],
```
Pierre-Sassoulas, line 43: We do not handle parsing error in setup.cfg, so pylint will crash if a setup.cfg is malformed. (We had the issue for toml, and we handle it.)

DanielNoord:
This looks like `find_default_config_files` might actually benefit from a refactor. Why are we scanning `.toml` files for the `pylint` section if then in the file parsing we're doing the same?

tests/config/test_find_default_config_files.py
```diff
@@ -0,0 +1,69 @@
...
+        [
+            """
+[tool.pylint]
+missing-member-hint = true
+""",
+            True,
+        ],
+    ],
+)
+def test_toml_has_config(content, expected, tmp_path):
```
DanielNoord, line 32:
```suggestion
def test_toml_has_config(content: str, expected: bool, tmp_path: Path) -> None:
```

tests/config/test_find_default_config_files.py
```diff
@@ -0,0 +1,69 @@
...
+name = pylint
+
+[pylint.messages control]
+disable = logging-not-lazy,logging-format-interpolation
+""",
+            True,
+        ],
+    ],
+)
+def test_cfg_has_config(content, expected, tmp_path):
```
DanielNoord, line 64:
```suggestion
def test_cfg_has_config(content: str, expected: str, tmp_path: Path) -> None:
```

tests/config/test_find_default_config_files.py
DanielNoord: I'm still not really sure if we should even allow `pylint` to run with malformed `setup.cfg` and `pyproject.toml`'s. I know we have added that new message but looking at it again it seems kind of weird. We're not a `.toml` linter so raising issues about them will always be a bit strange. Why don't we require users to fix their bad configuration before their tools can run?

Pierre-Sassoulas: The mere existence of a malformed setup.cfg / pyproject.toml would crash pylint here even if we do not use it at all. This function is searching for a pylint section in a configuration, but the pylintrc could exists. If the malformed setup.cfg / or pyproject.toml is the main configuration file then I think we should report the issue immediately and exit.

DanielNoord: +1 for passing here.

But do we then agree that the `config-parse-error` message should be removed?

DanielNoord (approved):

tests/config/test_find_default_config_files.py
Pierre-Sassoulas: > Why don't we require users to fix their bad configuration before their tools can run?

I think it's because we did not fix our badly designed and documented configuration files (yet) 😄 so error messages helps in the meantime. Removing the need to have section header and having clear error message for option that do not exists or have impossible values, and a good documentation would be better.

DanielNoord: Okay, let's pick this up in the Parser refactor then.

Lots to take into account there still 😄

Pierre-Sassoulas: Yeah maybe looking at all the issues talking about configuration would be a (labor-intensive) good start. I'm going to do that.

DanielNoord: @Pierre-Sassoulas You can look at the migration project I openend. I think I put all relevant issues in there already.

PR #3769, opened 2020-08-08
feat: Allow all config formats in all config locations

danrneal:
## Steps

- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [ ] If it's a new feature or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :sparkles: New feature |
| ✓  | :hammer: Refactoring  |
| ✓  | :scroll: Docs |

This PR would allow a `pylintrc`, `.pylintrc`, `pyproject.toml`, or `setup.cfg` in any location that config files are stored.  For example, currently `pyproject.toml` will only be read from the project root but not from the home directory.  This PR would allow it to be read from the home directory.  This is helpful when creating a global default config as other projects (such as isort) allow this config format in the home director.  Let me know if this is a welcome change with or without revision or an unwelcome one.

PR #6057, opened 2022-03-30
Deprecate ``find_pylintrc``

DanielNoord:

## Type of Changes

## Description

I wonder if this is considered external. It is in the `__all__` but there is no reason for anybody to use this, I think? We don't even use it as the logic for finding the configuration file has moved to `find_default_config_files`.

### DIFF
--- a/pylint/config/environment_variable.py
+++ b/pylint/config/environment_variable.py
@@ -2,6 +2,10 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
+import warnings
+
 from pylint.config.find_default_config_files import find_pylintrc
 
-PYLINTRC = find_pylintrc()
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    PYLINTRC = find_pylintrc()
--- a/pylint/config/find_default_config_files.py
+++ b/pylint/config/find_default_config_files.py
@@ -5,6 +5,7 @@
 import configparser
 import os
 import sys
+import warnings
 from typing import Iterator, Optional
 
 if sys.version_info >= (3, 11):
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
The coverage decreased because there were tests to check that we could browse a python package to find the pylintrc in parent directory too.

I would be in favor of searching at the root of the project in pyproject.toml, setup.cfg, pylintrc and .pylintrc and only that.  It's already a lot more than what flake8 is doing for example.  Guess which project has less issues to handle ? 😄  I think it's the way pylint is used most of the time (or should be used). And we added the possibility to do ``pylint .`` recently.But it's definitely a breaking change and I also think it's contrary to the per-directory configuration that was a big project of @AWhetter. 

This is related to https://github.com/PyCQA/pylint/pull/3769 where they were discussion about configuration.

DanielNoord:
> The coverage decreased because there were tests to check that we could browse a python package to find the pylintrc in parent directory too.
> 
> I would be in favor of searching at the root of the project in pyproject.toml, setup.cfg, pylintrc and .pylintrc and only that. It's already a lot more than what flake8 is doing for example. Guess which project has less issues to handle ? 😄 I think it's the way pylint is used most of the time (or should be used). And we added the possibility to do `pylint .` recently.But it's definitely a breaking change and I also think it's contrary to the per-directory configuration that was a big project of @AWhetter.
> 
> This is related to #3769 where they were discussion about configuration.

The issue is that these tests don't test that.. I removed these tests because the only asserts are on `find_pylintrc` which we don't use internally.
I have a refactor of `find_default_config_files` ready to go, so I would be more than happy to add similar tests for that function in a follow-up PR.

Pierre-Sassoulas (changes requested):
I think we should deprecate ``find_pylintrc`` first, before removing in 3.0, it's an API. We can add the new tests in this MR too.

Pierre-Sassoulas (approved):
Lgtm, thank you.

PR #6061, opened 2022-03-30
Fix crash in ``unnecessary-list-index-lookup`` checker (#6049)

### LINKED ISSUE #6049: TypeError in `unnecessary-list-index-lookup`
### Bug description

```python
pairs = [(0, 0)]
for i, (a, b) in enumerate(pairs):
    print(pairs[i][0])
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
Exception on node <For l.2 at 0x106f22e00> in file '/Users/.../.../a.py'
  File "/Users/.../pylint/pylint/utils/ast_walker.py", line 73, in walk
    callback(astroid)
  File "/Users/.../pylint/pylint/checkers/refactoring/refactoring_checker.py", line 651, in visit_for
    self._check_unnecessary_list_index_lookup(node)
  File "/Users/.../pylint/pylint/checkers/refactoring/refactoring_checker.py", line 2037, in _check_unnecessary_list_index_lookup
    and index.lookup(value_variable.name)[1][-1].lineno
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

### Expected behavior

no crash, and potentially, a message to use `a` and not `[i][0]`

### Pylint version

```shell
pylint 2.14.0-dev0
astroid 2.12.0-dev0
Python 3.10.2 (v3.10.2:a58ebcc701, Jan 13 2022, 14:50:16) [Clang 13.0.0 (clang-1300.0.29.30)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

timmartin:
## Type of Changes

## Description

This fixes a crash in the `unnecessary-list-index-lookup` checker where the value from `enumerate()` is being destructured, and thus no variable name is available. There are conceivably some cases where we could still salvage a useful check here, but the possible benefit seems minor and the original motivation was to catch the simple cases, so I hope we can just ignore this case for now.

Closes #6049

### DIFF
--- a/tests/functional/u/unnecessary/unnecessary_list_index_lookup.py
+++ b/tests/functional/u/unnecessary/unnecessary_list_index_lookup.py
@@ -44,3 +44,8 @@ result = [val for idx, val in enumerate(my_list) if my_list[idx] == 'a'] # [unne
 result = [val for idx, val in enumerate(my_list) if idx > 0 and my_list[idx - 1] == 'a']
 result = [val for idx, val in enumerate(my_list) if other_list[idx] == 'a']
 result = [my_list[idx] for idx, val in enumerate(my_list)] # [unnecessary-list-index-lookup]
+
+# Regression test for https://github.com/PyCQA/pylint/issues/6049
+pairs = [(0, 0)]
+for i, (a, b) in enumerate(pairs):
+    print(pairs[i][0])
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1994,6 +1994,11 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             # enumerate() result is being assigned without destructuring
             return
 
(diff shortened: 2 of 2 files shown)


tests/functional/u/unnecessary/unnecessary_list_index_lookup.py
```diff
@@ -44,3 +44,8 @@ def process_list_again(data):
 result = [val for idx, val in enumerate(my_list) if idx > 0 and my_list[idx - 1] == 'a']
 result = [val for idx, val in enumerate(my_list) if other_list[idx] == 'a']
 result = [my_list[idx] for idx, val in enumerate(my_list)] # [unnecessary-list-index-lookup]
+
+# Issue #6409
```
DanielNoord, line 48:
```suggestion
# Regression test for https://github.com/PyCQA/pylint/issues/6049
```

DanielNoord (approved):

jacobtylerwalls:
This checker is not released yet, it's only in 2.14. So I'll remove the backport labels.

PR #5593, opened 2021-12-23
False positive - not callable

mbyrnepr2:
## Description
Add a functional test. The issue itself has been fixed in https://github.com/PyCQA/astroid/pull/1311

Refs #5113

PR #6068, opened 2022-03-31
Fix E1102 / ``not-callable`` false positive for property that returns a lambda function conditionally

### LINKED ISSUE #5931: False positive: `not-callable` for property that returns a lambda function conditionally
### Bug description

```python
class C:
    def __init__(self):
        self._x = None

    @property
    def foo(self):
        if self._x is None:
            self._x = lambda: None
        return self._x

c = C()
c.foo()
```

### Configuration

_No response_

### Command used

```shell
pylint myfile.py | grep E1102
```

### Pylint output

```shell
myfile.py:12:0: E1102: c.foo is not callable (not-callable)
```

### Expected behavior

No error. The code runs when run under python.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.10.1 (main, Dec  6 2021, 22:25:40) [Clang 13.0.0 (clang-1300.0.29.3)]
```

### OS / Environment

MacOS Monterey

### Additional dependencies

_No response_

tushar-deepsource:
## Type of Changes

## Description

Update E1102 to only be raised when no inferred value is callable.

Resolves #5931

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,10 +9,6 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
-* Only raise ``not-callable`` when all the inferred values of a property are not callable.
-
-  Closes #5931
-
 
 * Add new check ``unnecessary-dunder-call`` for unnecessary dunder method calls.
 
@@ -53,6 +49,9 @@ Release date: TBA
   * functions & classes which contain both a docstring and an ellipsis.
   * A body which contains an ellipsis ``nodes.Expr`` node & at least one other statement.
 
+* Only raise ``not-callable`` when all the inferred values of a property are not callable.
+
+  Closes #5931
 
 
 What's New in Pylint 2.13.4?
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,7 +9,7 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
-* Only raise ``not-callable`` when all the inferred values of an expression are not callable.
+* Only raise ``not-callable`` when all the inferred values of a property are not callable.
 
   Closes #5931
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
+* Only raise ``not-callable`` when all the inferred values of an expression are not callable.
+
+  Closes #5931
+
 
 * Add new check ``unnecessary-dunder-call`` for unnecessary dunder method calls.
 
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -38,6 +38,10 @@ Extensions
 Other Changes
 =============
 
+* Only raise ``not-callable`` when all the inferred values of an expression are not callable.
+
+  Closes #5931
+
 * Fix false negative for ``no-member`` when attempting to assign an instance
   attribute to itself without any prior assignment.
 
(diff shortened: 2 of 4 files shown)


tushar-deepsource:
https://github.com/PyCQA/pylint/pull/6068/files#diff-9b76776694be02a09ca25a56adee0b3087ad5b86dfafd318923c4bf008c131c3R1233-R1238

This is the part that isn't covered, but it's hard to come up with a case for this, that won't break in the future.

pylint/checkers/typecheck.py
```diff
@@ -1227,20 +1227,25 @@ def _check_uninferable_call(self, node):
...
             if decorated_with_property(attr):
-
                 try:
-                    all_returns_are_callable = all(
-                        return_node.callable() or return_node is astroid.Uninferable
-                        for return_node in attr.infer_call_result(node)
-                    )
-                except astroid.InferenceError:
-                    continue
+                    call_results = list(attr.infer_call_result(node))
```
jacobtylerwalls, line 1231: Can you place the except directly after this so that the try only wraps what might fail?

srijan-paul:
zamn :^)

pylint/checkers/typecheck.py
```diff
@@ -1227,20 +1227,22 @@ def _check_uninferable_call(self, node):
...
                     continue
 
-                if not all_returns_are_callable:
-                    self.add_message(
-                        "not-callable", node=node, args=node.func.as_string()
-                    )
-                    break
+                if all(
+                    return_node is astroid.Uninferable for return_node in call_results
+                ):
```
jacobtylerwalls, line 1237:
```suggestion
                if astroid.Uninferable in call_results:
```

pylint/checkers/typecheck.py
```diff
@@ -1227,20 +1227,22 @@ def _check_uninferable_call(self, node):
...
+                    # We were unable to infer return values of the call, skipping
```
jacobtylerwalls, line 1238:
```suggestion
                    # We were unable to infer some return value of the call, skipping
```

jacobtylerwalls (commented):
nice! one suggestion to loosen this even more.

jacobtylerwalls:
Thanks for looking into it. Darn.

I still think long-term we could consider making that change, though. There's no restriction on what you can set with the property. You can set something that's callable, and then `not-callable` becomes sort of wrong. But no one is asking for that change right now :-)

DanielNoord:
@tushar-deepsource Coming into this pretty late but I was taking a look at stale PRs and found #5593. Could you quickly check if this solves that issue?

tushar-deepsource:
@DanielNoord This change only affects properties, so no it doesn't solve that issue.

DanielNoord (changes requested):
Only needs a changelog entry!

Didn't see this was only for `properties`. But that makes sense :)

tushar-deepsource:
@DanielNoord hopefully this is correct?

ChangeLog
```diff
@@ -9,6 +9,10 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
+* Only raise ``not-callable`` when all the inferred values of an expression are not callable.
```
DanielNoord, line 12:
```suggestion
* Only raise ``not-callable`` when all the inferred values of a property are not callable.
```

doc/whatsnew/2.14.rst
```diff
@@ -38,6 +38,10 @@ Extensions
 Other Changes
 =============
 
+* Only raise ``not-callable`` when all the inferred values of an expression are not callable.
```
DanielNoord, line 41:
```suggestion
* Only raise ``not-callable`` when all the inferred values of a property are not callable.
```

DanielNoord (approved):
Almost 😄 

But thanks @tushar-deepsource! Wonderful fix!

jacobtylerwalls:
Are we moving the changelog entry to 2.13.5?

PR #6075, opened 2022-03-31
Fix false-positive for unnecessary-ellipsis checker on comparisons

### LINKED ISSUE #6071: False positive `unnecessary-ellipsis` when multiple conditions in `if` statement
### Bug description

```python
# test.py
def func(val1, val2):
   if val1 is not ... and val2:
        pass
```

Above code raises error, but the following does not:
```python
def func(val1, _):
   if val1 is not ...:
        pass
```

So, it probably is something to do with multiple conditions here.

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:2:18: W2301: Unnecessary ellipsis constant (unnecessary-ellipsis)
```

### Expected behavior

No false-positives

### Pylint version

```shell
pylint 2.13.4
astroid 2.11.2
Python 3.10.0 (default, Oct 17 2021, 20:19:40) [Clang 12.0.5 (clang-1205.0.22.9)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

skshetry:
Closes #6071.

## Type of Changes

## Description

Fix false positive for `unnecessary-ellipsis` checker in comparisons.

Closes #6071.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -48,6 +48,7 @@ Release date: TBA
   * A body which contains an ellipsis ``nodes.Expr`` node & at least one other statement.
 
 
+
 What's New in Pylint 2.13.4?
 ============================
 Release date: 2022-03-31
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,8 @@ Release date: TBA
 
   Closes #1555
 
+* ``duplicate-argument-name`` now only raises once for each set of duplicated arguments.
+
 * Fix bug where specifically enabling just ``await-outside-async`` was not possible.
 
 ..
@@ -41,9 +43,9 @@ What's New in Pylint 2.13.5?
 ============================
 Release date: TBA
 
-* Fix false positive for ``unnecessary-ellipsis`` checker in comparisons.
-
-  Closes #6071.
+* Narrow the scope of the ``unnecessary-ellipsis`` checker to:
+  * functions & classes which contain both a docstring and an ellipsis.
+  * A body which contains an ellipsis ``nodes.Expr`` node & at least one other statement.
 
 
 What's New in Pylint 2.13.4?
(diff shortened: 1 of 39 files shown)


DanielNoord (approved):

Pierre-Sassoulas:
Maybe we should switch from disable list to enable list for this checker, we already had 3 issues with false positives for it. I.e. we list nodes where it should be raised instead of listing nodes where it should not be raised.

mbyrnepr2:
> Maybe we should switch from disable list to enable list for this checker, we already had 3 issues with false positives for it. I.e. we list nodes where it should be raised instead of listing nodes where it should not be raised.

I'm OK with this idea. I think it's the pragmatic thing to do. In hindsight the scope of this checker is quite complex. Perhaps we can limit the scope to functions and classes which define a docsting followed by an ellipsis, and then expand the scope from there as desired.

mbyrnepr2:
> > Maybe we should switch from disable list to enable list for this checker, we already had 3 issues with false positives for it. I.e. we list nodes where it should be raised instead of listing nodes where it should not be raised.
> 
> I'm OK with this idea. I think it's the pragmatic thing to do. In hindsight the scope of this checker is quite complex. Perhaps we can limit the scope to functions and classes which define a docstring followed by an ellipsis, and then expand the scope from there as desired.

Here is a MR with the more narrow scope for the checker: https://github.com/PyCQA/pylint/pull/6081; If this is ok with everyone perhaps we can review/merge that + the contents of the tests from the current MR? Otherwise please feel free to make any changes required in the current MR and ignore the one I've created.

ChangeLog
```diff
@@ -48,7 +48,6 @@ Release date: TBA
   * A body which contains an ellipsis ``nodes.Expr`` node & at least one other statement.
```
Pierre-Sassoulas, line 50:
```suggestion


```

Pierre-Sassoulas (approved):
Thank you for the contribution @skshetry , the fix itself disappeared because we took a safer /more conservative approach, but the tests will stay forever :)

PR #6078, opened 2022-03-31
Add the perflint extension to the pylintrc

Pierre-Sassoulas:
## Type of Changes

## Description

Trying out [perflint](https://pypi.org/project/perflint/).

Apparently my preferred way to import things is also faster see  ``dotted-import-in-loop``.

We could add it disable everything and then remove disable on a case by case basis.

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1280,7 +1280,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             if isinstance(comparison_node, nodes.Compare):
                 _find_lower_upper_bounds(comparison_node, uses)
 
-        for _, bounds in uses.items():
+        for bounds in uses.values():
             num_shared = len(bounds["lower_bound"].intersection(bounds["upper_bound"]))
             num_lower_bounds = len(bounds["lower_bound"])
             num_upper_bounds = len(bounds["upper_bound"])
--- a/pylintrc
+++ b/pylintrc
@@ -113,7 +113,6 @@ disable=
     loop-invariant-global-usage,
     use-list-comprehension,
     use-list-copy,
-    incorrect-dictionary-iterator,
     use-dict-comprehension,
 
 
--- a/pylintrc
+++ b/pylintrc
@@ -35,6 +35,7 @@ load-plugins=
     pylint.extensions.typing,
     pylint.extensions.redefined_variable_type,
     pylint.extensions.comparison_placement,
+    perflint,
 
 # Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
 # number of processors available to use.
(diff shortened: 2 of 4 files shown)


DanielNoord:
I haven't looked at what this does exactly, but before we get into the oldest debate after the chicken and egg problem:
```console
❯ cat test.py
import os


def dotted_import():
    for _ in range(100_000):
        return os.path.exists("/")
❯ cat test2.py
from os.path import exists


def direct_import():
    for _ in range(100_000):
        return exists("/")
❯ python -m timeit -s "from test import dotted_import; dotted_import()"
50000000 loops, best of 5: 7.24 nsec per loop
❯ python -m timeit -s "from test2 import direct_import; direct_import()"
50000000 loops, best of 5: 7.29 nsec per loop
```

😈 

There are so many articles about this online that I don't think we should get into this too much. I really prefer not losing namespaces as it makes code much more readable without having to worry where `dotted_import` is defined.
_To give an example: our test code. I always wonder whether parameter or functions are 1) pytest fixtures, 2) pytest fixtures defined in contest, 3) testutils imported directly, 4) utils imported directly, 5) functions defined in the test file, etc. Namespaces have a real benefit imo._

DanielNoord:
```python
❯ cat test.py
DICT = {i: i + 1 for i in range(100000)}


def iter_items():
    for _, k in DICT.items():
        x = k
❯ cat test2.py
DICT = {i: i + 1 for i in range(100000)}


def iter_values():
    for k in DICT.values():
        x = k
❯ python -m timeit -s "from test import iter_items; iter_items()"
50000000 loops, best of 5: 7.32 nsec per loop
❯ python -m timeit -s "from test2 import iter_values; iter_values()"
50000000 loops, best of 5: 7.48 nsec per loop
```

I was curious why this wasn't optimised and turns out on my local machine this difference is actually not true.
Obviously this "evidence" is anecdotal but I think we should consider this a style rather than a performance linter.

Pierre-Sassoulas:
> I think we should consider this a style rather than a performance linter.

True, I did not check the performance because the code style is also better. I think the "early beta" phrasing in the doc of perflint is true, there is a lot of false positive right now. ``use-tuple-over-list`` in particular, is already in pylint we discussed it with Marc, and came to the conclusion that is was really opinionated.

tonybaloney:
See https://github.com/tonybaloney/anti-patterns for a benchmark of each of the patterns. 

wrt to dotted imports. the rule is specific to loops because the difference is on the extra LOAD_ATTR instruction inside the loop https://github.com/tonybaloney/anti-patterns/blob/master/bench_imports.py#L1-L14

perflint is still in early development and is quite noisy on most code bases.

DanielNoord:
> See https://github.com/tonybaloney/anti-patterns for a benchmark of each of the patterns.
> 
> wrt to dotted imports. the rule is specific to loops because the difference is on the extra LOAD_ATTR instruction inside the loop https://github.com/tonybaloney/anti-patterns/blob/master/bench_imports.py#L1-L14
> 
> perflint is still in early development and is quite noisy on most code bases.

Like I said, I think we can debate this endlessly without reaching an agreement. I know there is a difference in instructions but I agree with many online that say that such optimisations are probably not worth it.  How often do you use an imported function anyway? Perhaps a 1000 times? That would be 1000 instructions extra, that won't make the difference for us. I think namespaces offer us much more. 

But as I said, this is mostly personal preference I think.

Pierre-Sassoulas:
Closing, as perflint is style in early development and we integrated the change that made sense.

PR #6081, opened 2022-03-31
Narrow the scope of the ``unnecessary-ellipsis`` checker 

mbyrnepr2:
Narrow the scope of the ``unnecessary-ellipsis`` checker to functions & classes which contain both a docstring and an ellipsis.

## Type of Changes

## Description

Closes #XXX

### DIFF
--- a/tests/functional/u/unnecessary/unnecessary_ellipsis.py
+++ b/tests/functional/u/unnecessary/unnecessary_ellipsis.py
@@ -9,10 +9,10 @@ try:
     A = 2
 except ValueError:
     A = 24
-    ... # [unnecessary-ellipsis]
+    ...
 
 def ellipsis_and_subsequent_statement():
-    ... # [unnecessary-ellipsis]
+    ...
     return 0
 
 # The parent of ellipsis is an assignment
--- a/tests/functional/u/unnecessary/unnecessary_ellipsis.py
+++ b/tests/functional/u/unnecessary/unnecessary_ellipsis.py
@@ -9,10 +9,10 @@ try:
     A = 2
 except ValueError:
     A = 24
-    ...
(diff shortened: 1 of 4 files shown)


tests/functional/u/unnecessary/unnecessary_ellipsis.py
```diff
@@ -9,10 +9,10 @@
     A = 2
 except ValueError:
     A = 24
-    ... # [unnecessary-ellipsis]
+    ...
```
DanielNoord, line 12: Hm, don't we think these are unnecessary?

mbyrnepr2: True. I simplified the checker to an extreme to prefer false negatives over false positives in order to make things as safe as possible to users. There will likely be other scenarios we can emit it also. We can allow this case in the current MR or in another one.

DanielNoord: Would `node.parent.parent.doc_node or len(node.parent.parent.body) > 1 work? Or something like that? Every `FunctionDef` or `ClassDef` that has a docstring or a body that has more than 1 node? That also reflects the description of the message a little better.

mbyrnepr2: That works well with the tests; lets try this. Thanks @DanielNoord.

DanielNoord (approved):
Perfect! Glad my suggestion worked!

DanielNoord:
@Pierre-Sassoulas Does this need an entry in `2.13.rst`?

Pierre-Sassoulas (approved):
It's ok, I think it's only a refactor.

PR #6082, opened 2022-04-01
Add example for `duplicate-argument-name` message documentation

gunungpw:

## Type of Changes

## Description

add code example for `duplicate-argument-name`

Ref #5953

### DIFF
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -4,7 +4,6 @@
 
 """Basic Error checker from the basic checker."""
 
-import collections
 import itertools
 from typing import Any, Iterator
 
@@ -15,6 +14,7 @@ from pylint import utils as lint_utils
 from pylint.checkers import utils
 from pylint.checkers.base.basic_checker import _BasicChecker
 from pylint.checkers.utils import infer_all
+from pylint.interfaces import HIGH
 
 ABC_METACLASSES = {"_py_abc.ABCMeta", "abc.ABCMeta"}  # Python 3.7+,
 # List of methods which can be redefined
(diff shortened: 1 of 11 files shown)


doc/data/messages/d/duplicate-argument-name/details.rst
```diff
@@ -0,0 +1,7 @@
+.. code:: python
+
+    >>> def get_fruits(apple, banana, apple):  # [duplicate-argument-name]
+    ...     pass
+    ...
+    File "<stdin>", line 1
```
DanielNoord, line 6: Schermafbeelding 2022-04-01 om 09 06 33

This is kind of duplicating what we already state above. We note that these raise SyntaxErrors and give this code example. Would you be okay with excluding this?

The `pylint-errors` project combines some of the data we originally provided so things that we have in our `Description` section can add up in their `Details` section.

doc/data/messages/d/duplicate-argument-name/related.rst
```diff
@@ -0,0 +1,2 @@
+`Testcases <https://github.com/PyCQA/pylint/blob/master/tests/functional/d/duplicate_argument_name.py>`_
```
DanielNoord, line 1: I understand why `pylint-error` has these in their examples, but I'm a bit scared that these links can easily change if we move tests around.
The Issue Tracker link also doesn't really capture all issues about this message if people use other names in their issues so I think it is not fool-proof.

DanielNoord (commented):
@gunungpw Thanks for you first contribution to `pylint`!

I made some comments. I should probably have addressed this in the guide in the original issue. I will update it with these comments as well.

doc/data/messages/d/duplicate-argument-name/related.rst
gunungpw: Understandable 👍, I will remove it

doc/data/messages/d/duplicate-argument-name/details.rst
gunungpw: Okay, thanks for the advice. I will also remove it 😄

DanielNoord (approved):

pylint/checkers/base/basic_error_checker.py
```diff
@@ -285,14 +285,17 @@ def visit_functiondef(self, node: nodes.FunctionDef) -> None:
 
         # provide detailed report about each repeated argument
         for argument_duplicates in arg_clusters.values():
```
DanielNoord, line 287: @Pierre-Sassoulas Can't we just merge this with for loop above? Add `arg.name` to a list. If `name` is already in that list raise a message on the current `arg` node being checked?

Pierre-Sassoulas: Yeah thank you, I was in the mindset of extending the existing code, but a refactor is a lot better.

PR #6091, opened 2022-04-01
Remove a cosmetic/semantic base class in the basic checker nebula

Pierre-Sassoulas:
## Type of Changes

## Description

See https://github.com/PyCQA/pylint/issues/6060#issuecomment-1085648352

### DIFF
--- a/pylint/checkers/base/name_checker/checker.py
+++ b/pylint/checkers/base/name_checker/checker.py
@@ -13,8 +13,7 @@ import astroid
 from astroid import nodes
 
 from pylint import constants, interfaces
-from pylint.checkers import utils
-from pylint.checkers.base.basic_checker import _BasicChecker
+from pylint.checkers import BaseChecker, utils
 from pylint.checkers.base.name_checker.naming_style import (
     KNOWN_NAME_TYPES,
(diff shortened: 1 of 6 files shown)


pylint/checkers/base/name_checker/checker.py
```diff
@@ -24,6 +23,8 @@
 from pylint.checkers.utils import is_property_deleter, is_property_setter
 
 # Default patterns for name types that do not have styles
+from pylint.interfaces import IAstroidChecker
```
DanielNoord, line 26: This seems incorrectly placed.

Pierre-Sassoulas:
Closing as it was just done to check if the inheritance from the same base class was required for it to work.

DanielNoord:
Don't we want to go ahead with this?

Pierre-Sassoulas:
I don't know if it's worth it, there's an explicitness in inheriting from the same base class that makes it easy to see those 6 checkers are linked and the name is important. I'm neutral on the change, so adding a commit to the repository for this seems like noise. If you think we should do it then let's do it :)

DanielNoord:
Fine to keep as is. We need to something about that `__implements__` anyway/

PR #6095, opened 2022-04-01
Close and join pools correctly in ``parallel.py``

DanielNoord:

## Type of Changes

## Description

Another attempt. Probable culprit of our CI issues would be https://github.com/PyCQA/pylint/pull/5612.

Turns out `.join()` is necessary. I checked locally and with the `join()` statement here there are more calls to internal `join()` methods than when only relying on the termination as the context manager exits. So, although I don't really understand the internals of `multiprocessing` and `threading` the change to the contextmanager has changed behaviour.

The difference in behaviour comes from `terminate()` only joining the `worker_handler` and `join()` also closing the pools.

### DIFF
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -175,6 +175,9 @@ def check_parallel(
             all_mapreduce_data[worker_idx].append(mapreduce_data)
             linter.msg_status |= msg_status
 
+        pool.close()
+        pool.join()
+
     _merge_mapreduce_data(linter, all_mapreduce_data)
     linter.stats = merge_stats([linter.stats] + all_stats)
 


Pierre-Sassoulas:
@doublethefish would you mind reviewing this, please ? :)

DanielNoord:
It's incorrect. But, I have finally, finally, finally found a way to skip these tests based on some metric. PR coming up :)

doublethefish:
I think the change is good, so I would reopen and merge.

`multiprocess.Pool.close/join` should ensure that the memory and other allocs are tidied up between invocations of the call (don't ask me why this is needed or the deeper implications of it). I am as surprised by this as you, I suspect - I am going to have to have a good look at the `multiprocess.Pool` implementation I guess.

The original idea, IIRC, was that the pool would be used once, do the work, and then the main proc would collate and analyse the returned data returned. Meaning the pool itself is disposable/closable/joinable. So, the problem I imagine you're seeing won't be an issue during normal "production" runs but it would probably affect unittests and CI/CD where `check_parallel()` might be called a lot. 

NB: `check_parallel()` should be replaced with a better, more integrated solutions, say optionally using MP.Pool in `PyLinter._check_files()`.

DanielNoord:
> I think the change is good, so I would reopen and merge.

Thanks for the review! I have reopened!

> `multiprocess.Pool.close/join` should ensure that the memory and other allocs are tidied up between invocations of the call. Don't ask me why this is needed or the deeper implications of it as I am as surprised by this as you, I suspect - I am going to have to have a good look at the `multiprocess.Pool` implementation I guess.

I spent some time debugging and didn't get too far. It seems like `terminate()` does send a signal to terminate to all `_pool` members but then net all of them get terminated before `terminate()` gets to it `join()` calls. Thus, it only sends a `join()` call for one of the two pool members (in our test function), while the other pool remains unjoined.

I don't know what the implications of that are but that's what I could figure out: a difference between calls to `join()` between `join()` and `terminate()`.

> NB: `check_parallel()` should be replaced with a better, more integrated solutions, say optionally using MP.Pool in `PyLinter._check_files()`.

I'm hesitant to change anything here, but obviously open to discussion. One of the issues with the parallelisation is stuff like `cyclic-import` and changing attributes of modules that then get put into the `astroid` cache. There are a number of open issues that discuss this.

DanielNoord:
Going to merge this as is. We don't really know how to test this and a changelog entry doesn't seem to be needed.

Pierre-Sassoulas:
Thank you for the guidance @doublethefish, much appreciated !

doublethefish:
On Fri, 1 Apr 2022 at 15:31, Daniël van Noord ***@***.***>
wrote:

> I think the change is good, so I would reopen and merge.
>
> Thanks for the review! I have reopened!
>
It’s a very interesting patch with some interesting implications, so I’m
glad to have been invited to look at it. I don’t like surprises like this
in what should be very solid APIs.

> multiprocess.Pool.close/join should ensure that the memory and other
> allocs are tidied up between invocations of the call. Don't ask me why this
> is needed or the deeper implications of it as I am as surprised by this as
> you, I suspect - I am going to have to have a good look at the
> multiprocess.Pool implementation I guess.
>
> I spent some time debugging and didn't get too far. It seems like
> terminate() does send a signal to terminate to all _pool members but then
> net all of them get terminated before terminate() gets to it join()
> calls. Thus, it only sends a join() call for one of the two pool members
> (in our test function), while the other pool remains unjoined.
>
Were exceptions being thrown in the sub-proc? You want to sync error
handling across procs, so that might have had implications for the rest of
pylint (without this patch?).

> I don't know what the implications of that are but that's what I could
> figure out: a difference between calls to join() between join() and
> terminate().
>
> NB: check_parallel() should be replaced with a better, more integrated
> solutions, say optionally using MP.Pool in PyLinter._check_files().
>
> I'm hesitant to change anything here, but obviously open to discussion.
> One of the issues with the parallelisation is stuff like cyclic-import
> and changing attributes of modules that then get put into the astroid
> cache. There are a number of open issues that discuss this.
>
That would be a huge change ;). I wasn’t able to get astroid to play nice
with the experiments I made doing just this.

DanielNoord:
> Were exceptions being thrown in the sub-proc? You want to sync error handling across procs, so that might have had implications for the rest of pylint (without this patch?).

I don't think so. It were just normal "successful" pylint runs.

> That would be a huge change ;). I wasn’t able to get astroid to play nice with the experiments I made doing just this.

Let me know if you need any help with testing something 😄 I quite enjoyed my little deep dive into `multiprocessing` today.

PR #6096, opened 2022-04-01
Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``

### LINKED ISSUE #6089: False positive unused-import (W0611) when assigning class properties with the same name
### Bug description

This code snippet
```python
from other import x, y  # raises unused-import (W0611) for x but not for y


class MyClass:
    x = float(x)
    y = y
```
will start raising a false positive from ea13058b9fde38698515c93bf67cc6018ed0064e up until the current main branch (6505c743564452f1b298b2a7ce2507a0b280ea25).

Related: #4993.

### Configuration

_No response_

### Command used

A false positive is only raised in combination with `--disable=all`:

```
$ python -m pylint --disable=all --enable=W0611 <path_to_file>
```
But when invoked as 
```
$ python -m pylint <path_to_file>
```
no false positive is raised, even though `unused-import (W0611)` is enabled by default.

### Pylint output

```shell
************* Module mycodesnippet
/path/to/mycodesnippet.py:1:0: W0611: Unused x imported from other (unused-import)

Your code has been rated at 7.50/10 (previous run: 7.50/10, +0.00)
```

### Expected behavior

No false positive should be raised.

### Pylint version

```shell
pylint 2.14.0-dev0
astroid 2.11.2
Python 3.6.15 | packaged by conda-forge | (default, Dec  3 2021, 18:49:43) 
[GCC Clang 11.1.0]
```

### OS / Environment

macOS 12.3

### Additional dependencies

_No response_

DudeNr33:

## Type of Changes

## Description
This should fix the false positive reported in #6089 (yes I named my branch wrong 😉).
But as I am not very familar with this part of the code base, I would like to ask @jacobtylerwalls for a review if moving this check can have other side effects than just a small performance loss.
Test suite is passing locally.

Closes #6089

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -71,6 +71,10 @@ Release date: TBA
   Closes #6069
   Closes #6136
 
+* Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``.
+
+  Closes #6089
+
 * Narrow the scope of the ``unnecessary-ellipsis`` checker to:
   * functions & classes which contain both a docstring and an ellipsis.
   * A body which contains an ellipsis ``nodes.Expr`` node & at least one other statement.
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,10 @@ Release date: TBA
 
   Closes #5936
 
+* Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``.
+
+  Closes #6089
+
 * ``potential-index-error``: Emitted when the index of a list or tuple exceeds its length.
   This checker is currently quite conservative to avoid false positives. We welcome
   suggestions for improvements.
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,10 +23,6 @@ Release date: TBA
 
   Closes #5936
 
-* Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``.
-
-  Closes #6089
-
 * ``potential-index-error``: Emitted when the index of a list or tuple exceeds its length.
   This checker is currently quite conservative to avoid false positives. We welcome
   suggestions for improvements.
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -59,3 +59,7 @@ Other Changes
   attribute to itself without any prior assignment.
 
   Closes #1555
+
+* Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``.
+
+  Closes #6089
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -59,7 +59,3 @@ Other Changes
   attribute to itself without any prior assignment.
 
   Closes #1555
-
-* Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``.
-
-  Closes #6089
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1555,12 +1555,6 @@ class VariablesChecker(BaseChecker):
 
         self._check_late_binding_closure(node)
 
-        if not (
-            self._is_undefined_variable_enabled
-            or self._is_used_before_assignment_enabled
-        ):
-            return (VariableVisitConsumerAction.RETURN, found_nodes)
-
         defnode = utils.assign_parent(found_nodes[0])
         defstmt = defnode.statement(future=True)
         defframe = defstmt.frame(future=True)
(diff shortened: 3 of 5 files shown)


pylint/checkers/variables.py
```diff
@@ -1604,6 +1598,12 @@ def _check_consumer(
         if use_outer_definition:
             return (VariableVisitConsumerAction.CONTINUE, None)
 
+        if not (
+            self._is_undefined_variable_enabled
+            or self._is_used_before_assignment_enabled
+        ):
+            return (VariableVisitConsumerAction.RETURN, found_nodes)
+
```
jacobtylerwalls, line 1606: Let's just nuke this in a follow up PR. It's only going to cause problems. I'm worried about similar problems with the `CONTINUE` that is still underneath this.

jacobtylerwalls (approved):
Great analysis. I think this condition and the instance variables it manages should just be deleted. Does it make sense to do it in a follow up PR or here?

If the check is moved even lower to leapfrog over the last `CONTINUE` action, then it's not accomplishing much. (Even `_is_only_type_annotation()` is good about short-circuiting.)

pylint/checkers/variables.py
DudeNr33: I think we can remove this bit entirely in this PR then. 
As I do not know exactly what code you are referring to with the ``CONTINUE``, I did not change anything else.

jacobtylerwalls: Sorry for not being clear -- I was referring to this on L1653:
https://github.com/PyCQA/pylint/blob/5e1926e77fb5956f3776f7a75b18e809bce56a62/pylint/checkers/variables.py#L1653

My thought was that a similar test case (whatever should be hitting that CONTINUE) might fail if we don't remove this bit altogether. 👍🏻

jacobtylerwalls:
I think this is unused now ⬇️ 

https://github.com/PyCQA/pylint/blob/5e1926e77fb5956f3776f7a75b18e809bce56a62/pylint/checkers/variables.py#L1093-L1095

DudeNr33:
``self._is_used_before_assignment_enabled`` has been removed with the latest commit.

DudeNr33:
@Pierre-Sassoulas how do you handle backports? 
Merge the current PR into main and then merge it manually also into the maintenance branch for 2.13? 
Should I update the ``ChangeLog`` and ``whatsnew`` entry to include the change in the sections for 2.13.5 in this PR already?

Pierre-Sassoulas:
> how do you handle backports?

[The full doc](https://github.com/PyCQA/pylint/blob/main/doc/release.md#backporting-a-fix-from-main-to-the-maintenance-branch)

TLDR we squash on the main branch so we only have to cherry-pick one commit on the maintenance branch later. (What is not said is that I cherry-pick everything in chronological order just before the release, so you can just label "Need backport", We could do it just after merging too)

> Should I update the ChangeLog and whatsnew entry to include the change in the sections for 2.13.5 in this PR already?

Yes, in the current patch version 2.13.5 for the changelog and the whatsnew for 2.13.

DudeNr33:
Thanks for the info! I have updated the ``ChangeLog`` and ``whatsnew/2.13.rst``. 
But I fear that something went wrong when I tried to merge the current ``main`` branch, as it now displays an awful lot of changed files...

Pierre-Sassoulas (changes requested):
I think it should be a rebase instead of a merge, it's probably fixable with ``git rebase origin/main``.

Pierre-Sassoulas:
You need to disregard what git is telling you when you're pushing and actually rewrite the history by using ``git push --force ``

DudeNr33:
Thanks a ton, that did the trick. I'm still struggling with ``git`` sometimes... 😀

ChangeLog
```diff
@@ -23,6 +23,10 @@ Release date: TBA
 
   Closes #5936
 
+* Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``.
+
+  Closes #6089
+
```
jacobtylerwalls, line 29:
```suggestion
```

doc/whatsnew/2.14.rst
```diff
@@ -59,3 +59,7 @@ Other Changes
   attribute to itself without any prior assignment.
 
   Closes #1555
+
+* Fix false positive for ``unused-import`` when disabling both ``used-before-assignment`` and ``undefined-variable``.
+
+  Closes #6089
```
jacobtylerwalls, line 65:
```suggestion
```

DudeNr33:
Thanks @jacobtylerwalls, removed the entries for 2.14.

Pierre-Sassoulas (approved):
👍

PR #6098, opened 2022-04-01
Improve CPU count detection in cgroup environments and fix CI

### LINKED ISSUE #5809: 🔍 Improve CPU count detection in cgroup environments

DanielNoord:

## Type of Changes

## Description

After a rough Friday afternoon I think I have finally found the issue: our CI runs in a docker container.
#5809 already raised an issue about this and our approach at determining the number of available CPUs. I have adopted a suggestion to fix this issue from the Python bug report found at https://bugs.python.org/issue36054.

I have tested this and `_query_cpu` returns `1` for us in the CI, both on Linux <s>and on Windows</s>. As you will see in this CI run the test marked with the new `needs_two_cores` mark are indeed skipped.

This would thus close #5809 as well as fix our CI issues.

<s>One issue that results from this is that these tests won't be run on our CI. They will still run locally (I checked this on my own local Mac), but won't be run in the CI. I don't immediately know a solution for this. The Github Actions runners simply don't really allow us to test our approach at parallelisation I think...</s>

Edit: I misread and on Windows the jobs do keep running. Perhaps that was just a fluke and it changes based on the runner? That would explain some of the variance we have been seeing. Anyway, seems as if the test will run at least some of the time. So that is probably good enough for now?


DanielNoord:
I'm trying to turn on `3.11` as there is finally a release available in https://github.com/DanielNoord/pylint/runs/6037333512?check_suite_focus=true but I think we're getting stuck here even more often.

Probably can't start testing on `3.11` until this lands.

Pierre-Sassoulas (commented):
I'm not qualified to review query_cpu and I don't think I'll be able to be in a reasonable time, but I guess you did your research properly and that detecting CPU's inside a docker is can't be done without some hacks 😄 The principle of the mark when two cores are required is consensual so 👍 for that. I guess the worst that can happen if qury_cpu do not work is that it use the wrong number of CPU when inside a docker ?

DanielNoord:
> I'm not qualified to review query_cpu and I don't think I'll be able to be in a reasonable time, but I guess you did your research properly and that detecting CPU's inside a docker is can't be done without some hacks 😄 

Do we have somebody who would be better suited to do this?

> The principle of the mark when two cores are required is consensual so 👍 for that. I guess the worst that can happen if qury_cpu do not work is that it use the wrong number of CPU when inside a docker ?

Yeah! Or any other environment that uses these files.

Pierre-Sassoulas (approved):

pylint/lint/run.py
```diff
@@ -36,15 +37,51 @@
     multiprocessing = None  # type: ignore[assignment]
 
 
+def _query_cpu() -> int | None:
+    """Try to determine number of CPUs alloted in a docker container.
```
jacobtylerwalls, line 41:
```suggestion
    """Try to determine number of CPUs allotted in a docker container.
```

pylint/lint/run.py
```diff
@@ -36,15 +37,51 @@
...
+
+    This is based on discussion and coped from suggestions in
```
jacobtylerwalls, line 43:
```suggestion
    This is based on discussion and copied from suggestions in
```

jacobtylerwalls (approved):

jacobtylerwalls:
Thanks! If this ever gets brittle/stops working because the file locations change, we could look into the 3rd party packages on the BPO issue.

PR #6109, opened 2022-04-01
Use the ``argparse`` config handler in ``class_checker.py``

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

PR #6110, opened 2022-04-01
Use the ``argparse`` config handler in ``imports.py``

DanielNoord:

## Type of Changes

## Description

Blocked by #6109.

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -7,7 +7,7 @@
 import collections
 import copy
 import os
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Set, Tuple, Union
 
 import astroid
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
Maybe we could rename ``linter.namespace`` to ``linter.config``  or ``linter.option`` ?

DanielNoord:
> Maybe we could rename `linter.namespace` to `linter.config` or `linter.option` ?

Shall we do that once I have done all checkers? I can imagine that we might not actually want to put it on the `linter` but set it as `config` on each `checker` for example.

Pierre-Sassoulas:
It depends if cleaning up later costs more than doing it since the beginning, I'll let you be the judge of that 😄

DanielNoord:
> It depends if cleaning up later costs more than doing it since the beginning, I'll let you be the judge of that 😄

I'll probably use a replace all for that. I think for now I'll continue like this.

PR #6115, opened 2022-04-02
add example for `missing-format-argument-key` message documentation

gunungpw:

## Type of Changes

## Description

- add code example for `missing-format-argument-key`
- add PEP 3101 and Python string formatting docs to `related.rst`

Ref #5953


Pierre-Sassoulas (approved):
Thanks ! All your example are so... american 😄

gunungpw:
🤣  I am afraid if the example is not a popular/normal phrase or word, a beginner like me who reads the code will confuse about what actually the code does. it's often the case if the code uses `foo` `bar`. 
I am actually not from English speaking country (I am still learning English my grammar is bad :)) so the only reference I got from there was popular culture like movie 😄.

Pierre-Sassoulas:
Not saying it's a bad thing at all. But you can also integrate things from your own culture if you want.

>  a beginner like me who reads the code will confuse about what actually the code does. it's often the case if the code uses foo bar.

Yes I think so too, example that can be understood by school children with no prior knowledge is what we should be aiming at. Mainly because actual school children could read the examples, but also because everyone benefit from more engaging examples.

doc/data/messages/m/missing-format-argument-key/related.rst
```diff
@@ -0,0 +1,2 @@
+`PEP 3101 <https://peps.python.org/pep-3101/>`_
```
DanielNoord, line 1: @Pierre-Sassoulas  @gunungpw I didn't check this myself, but the `related.rst` documents should preferably be a list using `- ...`. That gets formatted a little nicer.

Pierre-Sassoulas: There's an issue with lists in something we merged recently see additional detail : https://pylint.pycqa.org/en/latest/messages/warning/arguments-differ.html. I think because it's rst formatting we need ``*`` instead of ``-``.

PR #6117, opened 2022-04-02
Use ``argparse`` config handler in ``imports.py`` & ``exceptions.py``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/config/argument.py
+++ b/pylint/config/argument.py
@@ -39,10 +39,18 @@ def _yn_transformer(value: str) -> bool:
     )
 
 
+def _non_empty_string_transformer(value: str) -> str:
+    """Check that a string is not empty and remove quotes."""
+    if not value:
+        raise argparse.ArgumentTypeError("Option cannot be an empty string.")
+    return pylint_utils._unquote(value)
+
+
 _TYPE_TRANSFORMERS: Dict[str, Callable[[str], _ArgumentTypes]] = {
     "choice": str,
     "csv": _csv_transformer,
     "int": int,
+    "non_empty_string": _non_empty_string_transformer,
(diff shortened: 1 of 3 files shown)


pylint/config/argument.py
```diff
@@ -39,10 +39,18 @@ def _yn_transformer(value: str) -> bool:
     )
 
 
+def _non_empty_string_transformer(value: str) -> str:
```
Pierre-Sassoulas, line 42: Are we progressively discovering what we need to add based on the existing tests ? It's smart and efficient but I'm a little afraid that the tests will not be exhaustive. Maybe it would be safer to look at what optparse was actually doing and try to translate ?

DanielNoord: No we have this in `optparse` as well. This copied from `VALIDATORS`.

I just don't want to use anything in the old files because those are validators instead of transformers: ie., they get called every time a value is parsed or added instead of only when needed in the new setup.
Besides, I feel like it's easier to just cut-off everything instead of trying to build on top of it. Just copying what seems relevant :)

DanielNoord: Sorry, I misunderstood what you meant. 

Yeah, we're doing this progressively. But I think it's better to do it progressively now and do one final check at the end if we're missing something obvious. Copying stuff over from the `optparse` implementation is exactly what we want to avoid as I noticed a number of things are different in `argparse`. I'll do a big final check for this at the end.

Pierre-Sassoulas: > Copying stuff over from the optparse implementation is exactly what we want to avoid 

Yeah, fair enough !

Pierre-Sassoulas (approved):

PR #5511, opened 2021-12-13
Add a DeprecationWarning to set_config_directly

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

See https://github.com/PyCQA/pylint/pull/5509#issuecomment-992346711

Issue #5513, opened 2021-12-13
To-do list for 2.14.0


DanielNoord:
### Current problem

- [ ] Remove ``set_config_directly``, see https://github.com/PyCQA/pylint/pull/5511

### Desired solution

.

DanielNoord:
I'm adding to this that https://bugs.python.org/issue33053 seems to have been fixed in 3.7. We mention that issue in here:
https://github.com/PyCQA/pylint/blob/019794b808271d45f86a7014e9c91cb04458a47a/pylint/__init__.py#L56

We need to update the docstring to reflect whatever the situation in 3.7+ is.

cdce8p:
We could also consider if we want to require `3.7.2` instead of `3.7.0`. That would prevent any upcoming issues with `NoReturn`. Just to be clear though, the current code works with `3.7.0` just fine.
_For reference #5304._

DanielNoord:
> I'm adding to this that https://bugs.python.org/issue33053 seems to have been fixed in 3.7. We mention that issue in here:
> 
> https://github.com/PyCQA/pylint/blob/019794b808271d45f86a7014e9c91cb04458a47a/pylint/__init__.py#L56
> 
> We need to update the docstring to reflect whatever the situation in 3.7+ is.

I re-read that issue and there is no action to be taken here. We handle this correctly. Ticking off.

DanielNoord:
@cdce8p @Pierre-Sassoulas @jacobtylerwalls 

I have been trying to warp my head around https://github.com/PyCQA/pylint/issues/2862 as the issue has a fairly easily reproducible example and should allow narrowing down the `namespace` issue.

One thing I was wondering is whether this function is truly necessary:
https://github.com/PyCQA/pylint/blob/019794b808271d45f86a7014e9c91cb04458a47a/pylint/__init__.py#L49-L67

I'm asking this because with the following diff it shows that we do not recover ``path`` correctly:
```diff

diff --git a/pylint/lint/utils.py b/pylint/lint/utils.py
index 792c47f2d..8c6fc427d 100644
--- a/pylint/lint/utils.py
+++ b/pylint/lint/utils.py
@@ -117,6 +117,8 @@ def fix_import_path(args: Sequence[str]) -> Iterator[None]:
     `sys.path` is reset to its original value upon exiting this context.
     """
     original = _patch_sys_path(args)
+    for i in sys.path:
+        print(i)
     try:
         yield
     finally:
```

```console
[...]

PR #6123, opened 2022-04-02
Remove ``set_config_directly``

DanielNoord:

## Type of Changes

## Description

As discussed in #5511.

I still think that `2.14` is good enough. Especially since this is a test decorator and won't impact performance of a plugin directly.


Pierre-Sassoulas (approved):
Actually we might need to check the TODO before releasing because it's easier to forget about that. Maybe upgrade the #5513 type list when we add todos ?

DanielNoord:
> Actually we might need to check the TODO before releasing because it's easier to forget about that. Maybe upgrade the #5513 type list when we add todos ?

Because of `argparse` I'll need to do a full review of all TODOs anyway. But I'll add that to the list!

PR #6124, opened 2022-04-02
Use ``argparse`` config handler on three checkers

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/base/docstring_checker.py
+++ b/pylint/checkers/base/docstring_checker.py
@@ -115,12 +115,12 @@ class DocStringChecker(_BasicChecker):
 
     @utils.check_messages("missing-docstring", "empty-docstring")
     def visit_classdef(self, node: nodes.ClassDef) -> None:
-        if self.config.no_docstring_rgx.match(node.name) is None:
+        if self.linter.namespace.no_docstring_rgx.match(node.name) is None:
             self._check_docstring("class", node)
 
     @utils.check_messages("missing-docstring", "empty-docstring")
     def visit_functiondef(self, node: nodes.FunctionDef) -> None:
-        if self.config.no_docstring_rgx.match(node.name) is None:
+        if self.linter.namespace.no_docstring_rgx.match(node.name) is None:
             ftype = "method" if node.is_method() else "function"
             if (
                 is_property_setter(node)
@@ -176,7 +176,7 @@ class DocStringChecker(_BasicChecker):
                 # If the module does not have a body, there's no reason
                 # to require a docstring.
                 return
-            max_lines = self.config.docstring_min_length
+            max_lines = self.linter.namespace.docstring_min_length
 
             if node_type != "module" and max_lines > -1 and lines < max_lines:
                 return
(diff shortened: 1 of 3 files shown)


DanielNoord:
Blocked by #5921.

Pierre-Sassoulas (approved):
Let's do the migration from python 3.6 now then.

DanielNoord:
> Let's do the migration from python 3.6 now then.

We're still merging hot fixes. Still is a fairly trivial PR. I would be okay with waiting a bit. There is enough else for me to do!

Pierre-Sassoulas:
OK, after https://github.com/PyCQA/pylint/milestone/61 is closed then ?

DanielNoord:
> OK, after https://github.com/PyCQA/pylint/milestone/61 is closed then ?

I'd say, let's wait a minimal week since last hotfix release. That gives people ample time to report any new issues. I agree with Marc, the change to `3.7` is likely going to make releasing hotfixes much more difficult.

Pierre-Sassoulas:
> last hotfix release. 

We continuously release false positive fixes now so they won't be a last hotfix release. Do you mean the last one with crashes fixed inside it ?

>  releasing hotfixes much more difficult.

I agree too, but the question is when do we release 2.14 and not ever hot fix python 3.6 anymore :) Depends on how blocked we are. There are maybe 4 MR now and 2.14 is close to 100 issues already (lots of documentation but still). So, I'm already considering releasing 2.14.

DanielNoord:
> We continuously release false positive fixes now so they won't be a last hotfix release. Do you mean the last one with crashes fixed inside it ?

Yeah, or false positives introduced by `2.12`.

> I agree too, but the question is when do we release 2.14 and not ever hot fix python 3.6 anymore :) Depends on how blocked we are. There are maybe 4 MR now and 2.14 is close to 100 issues already (lots of documentation but still). So, I'm already considering releasing 2.14.

If possible I would wait with `2.14` until the `argparse` migration has been finished. That way we can do a "light" release with a move to `3.7` + a request for plugin maintainers on feedback on the new options handling. Looking at the speed at which it's going now I think that it is feasible to do so.
Only 227 references to `.config` remain in the codebase right now.

pylint/checkers/base/name_checker/checker.py
```diff
@@ -298,7 +300,9 @@ def _create_naming_rules(self) -> Tuple[Dict[str, Pattern[str]], Dict[str, str]]
                 regexps[name_type] = DEFAULT_PATTERNS[name_type]
 
             custom_regex_setting_name = f"{name_type}_rgx"
-            custom_regex = getattr(self.config, custom_regex_setting_name, None)
+            custom_regex = getattr(
+                self.linter.namespace, custom_regex_setting_name, None
+            )
```
Pierre-Sassoulas, line 305: Should we handle this in the option parsing directly and not later with getattr ? I think it's not ideal as it's a delayed default value that put a little of option parsing inside checkers.

Pierre-Sassoulas (changes requested):

pylint/checkers/base/name_checker/checker.py
DanielNoord: These options are created dynamically for each name type we have.

We can probably define them in `checker.options` but that's more a general refactor of this checker. By creating them dynamically we do remove a lot of duplication of the code though.

Pierre-Sassoulas (approved):

DanielNoord:
Missing coverage is also missing on `main` so I'm merging.

PR #6127, opened 2022-04-02
Move ``_parse_rich_type_value`` to ``utils``

DanielNoord:

## Type of Changes

## Description

Removes the need to instantiate another `_ConfigurationFileParser` 😄

### DIFF
--- a/pylint/config/config_file_parser.py
+++ b/pylint/config/config_file_parser.py
@@ -8,7 +8,9 @@ import configparser
 import os
 import sys
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Depending on how much of those we have we might want to create a RichTypeParser class.

DanielNoord:
> Depending on how much of those we have we might want to create a RichTypeParser class.

Based on current knowledge I don't think it is necessary. The amount of types is minimal and there is only one place where we call it in our normal codebase. The other use is in `testutils`.

PR #6130, opened 2022-04-02
Use ``argparse`` config handler in ``similar.py``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/similar.py
+++ b/pylint/checkers/similar.py
@@ -781,14 +781,14 @@ class SimilarChecker(BaseChecker, Similar, MapReduceMixin):
     reports = (("RP0801", "Duplication", report_similarities),)
 
     def __init__(self, linter=None) -> None:
-        BaseChecker.__init__(self, linter)
+        BaseChecker.__init__(self, linter, future_option_parsing=True)
         Similar.__init__(
             self,
-            min_lines=self.config.min_similarity_lines,
-            ignore_comments=self.config.ignore_comments,
-            ignore_docstrings=self.config.ignore_docstrings,
-            ignore_imports=self.config.ignore_imports,
-            ignore_signatures=self.config.ignore_signatures,
+            min_lines=self.linter.namespace.min_similarity_lines,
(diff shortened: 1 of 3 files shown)


pylint/checkers/similar.py
```diff
@@ -861,15 +878,7 @@ def reduce_map_data(self, linter, data):
...
-        recombined = SimilarChecker(linter)
-        recombined.min_lines = self.min_lines
-        recombined.ignore_comments = self.ignore_comments
-        recombined.ignore_docstrings = self.ignore_docstrings
-        recombined.ignore_imports = self.ignore_imports
-        recombined.ignore_signatures = self.ignore_signatures
-        recombined.open()
-        Similar.combine_mapreduce_data(recombined, linesets_collection=data)
-        recombined.close()
+        Similar.combine_mapreduce_data(self, linesets_collection=data)
```
Pierre-Sassoulas, line 881: Is is safe to change that ? It touch the multiprocessing and the tests are automatically skipped if we don't have the number of core available. I'm very afraid 😄

Pierre-Sassoulas (commented):
Look good overall but change to multiprocessing makes me nervous.

pylint/checkers/similar.py
DanielNoord: I tested this locally: all passed.
The only thing `combine_mapreduce_data` does is:
```python
self.linesets = [line for lineset in linesets_collection for line in lineset]
```

So instead, of setting `self.linesets` on `recombined` we do it on `self` now. I changed some tests to also use `-j 4` locally and none failed.

I'm not 100% sure, but the current approach doesn't work anyway. Because we create a new `SimilarChecker` with an existing `linter` we get a duplicate argument error (which is correct).
We can't import `PyLinter` and instantiate a new one here because of cyclic imports (I do do this in the tests). So if this regresses we'll need to find a better solution than the current one anyway.

Pierre-Sassoulas (approved):

PR #6131, opened 2022-04-02
Enable usage of custom pylintrc file for message documentation tests

DudeNr33:

## Type of Changes

## Description

In order to test some messages (e.g. for messages of optional checkers), we need a way to use a custom config.
With this modification it is possible to place a ``pylintrc`` file next to ``good.py`` and ``bad.py``.
To prove it works I added a documentation example for one of the optional checkers.

Ref: https://github.com/PyCQA/pylint/issues/5953#issuecomment-1085587049

### DIFF
--- a/doc/data/messages/c/confusing-consecutive-elif/details.rst
+++ b/doc/data/messages/c/confusing-consecutive-elif/details.rst
@@ -1,2 +1,1 @@
-Following an indented ``if`` block directly with an ``elif`` of lower indentation can lead to confusion whether the unindent is deliberate or accidental.
-Adding an explicit ``else`` in the indented ``if`` statement, even if it only contains a ``pass`` statement, can help clarify the code.
+Creating a function for the nested conditional, or adding an explicit ``else`` in the indented ``if`` statement, even if it only contains a ``pass`` statement, can help clarify the code.
--- a/doc/data/messages/c/confusing-consecutive-elif/details.rst
+++ b/doc/data/messages/c/confusing-consecutive-elif/details.rst
@@ -1,2 +1,2 @@
-Following an indentend ``if`` block directly with an ``elif`` of lower indentation can lead to confusion whether the unindent is deliberate or accidental.
+Following an indented ``if`` block directly with an ``elif`` of lower indentation can lead to confusion whether the unindent is deliberate or accidental.
 Adding an explicit ``else`` in the indented ``if`` statement, even if it only contains a ``pass`` statement, can help clarify the code.
--- a/doc/data/messages/c/confusing-consecutive-elif/details.rst
+++ b/doc/data/messages/c/confusing-consecutive-elif/details.rst
@@ -1,0 +1,2 @@
+Following an indentend ``if`` block directly with an ``elif`` of lower indentation can lead to confusion whether the unindent is deliberate or accidental.
+Adding an explicit ``else`` in the indented ``if`` statement, even if it only contains a ``pass`` statement, can help clarify the code.
--- a/doc/data/messages/c/confusing-consecutive-elif/good.py
+++ b/doc/data/messages/c/confusing-consecutive-elif/good.py
@@ -1,0 +1,8 @@
+def myfunc(shall_continue: bool, shall_exit: bool):
+    if shall_continue:
+        if input("Are you sure?") == "y":
+            print("Moving on.")
+        else:
+            pass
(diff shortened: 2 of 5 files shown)


doc/data/messages/c/confusing-consecutive-elif/good.py
```diff
@@ -0,0 +1,8 @@
+def myfunc(shall_continue: bool, shall_exit: bool):
+    if shall_continue:
+        if input("Are you sure?") == "y":
+            print("Moving on.")
+        else:
+            pass
+    elif shall_exit:
+        print("Exiting.")
```
Pierre-Sassoulas, line 8:
```suggestion
def user_confirmation():
    if input("Are you sure?") == "y":
        print("Moving on.")

def myfunc(shall_continue: bool, shall_exit: bool):
    if shall_continue:
        user_confirmation()
    elif shall_exit:
        print("Exiting.")
```
Another possibility imo is to create a function for the nested if/elif.

doc/data/messages/c/confusing-consecutive-elif/details.rst
```diff
@@ -0,0 +1,2 @@
+Following an indented ``if`` block directly with an ``elif`` of lower indentation can lead to confusion whether the unindent is deliberate or accidental.
+Adding an explicit ``else`` in the indented ``if`` statement, even if it only contains a ``pass`` statement, can help clarify the code.
```
Pierre-Sassoulas, line 2:
```suggestion
Creating a function for the nested conditional, or adding an explicit ``else`` in the indented ``if`` statement, even if it only contains a ``pass`` statement, can help clarify the code.
```

doc/test_messages_documentation.py
```diff
@@ -53,7 +53,13 @@ def __init__(self, test_file: Tuple[str, Path]) -> None:
         self._linter.config.persistent = 0
         checkers.initialize(self._linter)
 
-        config_file = next(config.find_default_config_files(), None)
+        # Check if this message has a custom configuration file (e.g. for enabling optional checkers).
+        # If not, use the default configuration.
+        config_file: Optional[Path]
```
Pierre-Sassoulas, line 58: @DanielNoord I think we talked about this already before but what do you think about using [``FunctionalTestFile``](https://github.com/PyCQA/pylint/blob/main/pylint/testutils/functional/test_file.py#L46) instead here ? At some point we will encounter the same issue than for functional tests (i.e. handling min_version, max_version, some options...). We could do ``test_file.option_file`` instead of recreating the logic a little differently.

Pierre-Sassoulas (changes requested):
Thanks you @DudeNr33 !

doc/test_messages_documentation.py
DudeNr33: That was actually my initial approach, but the problem is that ``FunctionalTestFile`` assumes the config file to have the same name as the ``.py`` file.
There is quite a good portion of code duplication also in the ``LintModuleTest`` class here. 

Besides the different folder structure, the tight coupling between the "original" ``LintModuleTest`` and ``test_functional.py`` is another problem.

doc/data/messages/c/confusing-consecutive-elif/good.py
DudeNr33: Thanks, I will include both possibilities. 👍

doc/test_messages_documentation.py
DanielNoord: Yeah, I advise against trying to use `LintModuleTest` here. The idea is similar but because of the different folder structures it doesn't really work.

It's similar to your `Duck` vs. `PlasticDuck` example @Pierre-Sassoulas  😄

Pierre-Sassoulas (approved):

doc/test_messages_documentation.py
DanielNoord, line 58: Minor nitpick but I believe this isn't necessary. `mypy` should allow to assign `None` to a previously assigned variable (L60 vs L62).

Pierre-Sassoulas: Sorry I merged too fast.

DudeNr33: I tried it first without this annotation, but ``mypy`` complained.

DanielNoord: Ah, never mind then!

PR #6135, opened 2022-04-02
Add last missing transformers

DanielNoord:

## Type of Changes

## Description

See:
https://github.com/PyCQA/pylint/blob/5e1926e77fb5956f3776f7a75b18e809bce56a62/pylint/config/option.py#L99-L114

@Pierre-Sassoulas  Turning on `argparse` for the `PyLinter` class is a mammoth task so I'm splitting some of the jobs into separate PRs that can be merged separately. They will have low coverage, but they are needed for the final `PyLinter` PR to be somewhat reviewable.

### DIFF
--- a/pylint/config/argument.py
+++ b/pylint/config/argument.py
@@ -2,20 +2,28 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
-"""Definition of an Argument class and validators for various argument types.
+"""Definition of an Argument class and transformers for various argument types.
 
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
Maybe it would be beneficial to add small unit tests for transformers ? That way we could catch problems even without a functional/end to end tests.

DanielNoord:
> Maybe it would be beneficial to add small unit tests for transformers ? That way we could catch problems even without a functional/end to end tests.

Not sure that's worth our time. These are such basic operations (and as you said we already test them e2e). Perhaps we should add them if somebody finds an issue with them and expand like this? Writing tests for things we already tests seems a bit wasteful considering the work that's still left 😅

Pierre-Sassoulas (approved):

PR #6138, opened 2022-04-02
Add a _StoreTrueArgument class

DanielNoord:

## Type of Changes

## Description

**Need to rebase**.

`store_true` arguments can't have a `type`. I thought about subclassing this, but then I would need to change the order of the `__init__` parameters to provide a default value for `type`. I'll probably do that at the end but I want to wait for any other new action types I haven't thought about yet.

### DIFF
--- a/pylint/config/argument.py
+++ b/pylint/config/argument.py
@@ -121,3 +121,47 @@ class _Argument:
         See:
         https://docs.python.org/3/library/argparse.html#metavar
         """
+
+
+class _StoreTrueArgument:
+    """Class representing a 'store_true' argument to be passed by an argparse.ArgumentsParser.
+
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (commented):
Does'nt ``store_true`` already exists in argparse as an argument ? Why do we need the ``_StoreTrueArgument`` class ?

DanielNoord:
> Does'nt `store_true` already exists in argparse as an argument ? Why do we need the `_StoreTrueArgument` class ?

Because `_Argument` requires a `type` parameter and `_StoreTrueArgument` doesn't. They are two different types of arguments that have different requirements in terms of attributes.

Pierre-Sassoulas (approved):

PR #6141, opened 2022-04-02
Suppress ``useless-super-delegation`` if return type changed (#5822)

### LINKED ISSUE #5822: useless-super-delegation when an override changes the return type annotation
### Current problem

Similar to #1923, but for the return type annotation.
It is sometimes useful to override a method and change the return type annotation - whether to help a static type checker out or to help the IDE make better autocomplete suggestions.

Here's an example - it's a bit contrived, I actually ran into this when overriding to add typing to a method from a third-party library (ANTLR).

```python
"""Quick test to demonstrate useless-super-declaration raised for overrides
that change return type annotations."""
import random
from dataclasses import dataclass
from typing import Any

class _AGenerator:
    """Abstract base class for generators."""
    _valid_values: list[Any]

    def get_random(self) -> Any:
        """Generates a random value from the valid values that this generator
        can return."""
        return random.choice(self._valid_values)

@dataclass
class StrGenerator(_AGenerator):
    """Generator that accepts and returns string values."""
    _valid_values: list[str]

    # override to add typing
    def get_random(self) -> str:
        return super().get_random()

s = StrGenerator(['foo', 'bar', 'qux'])
print(s.get_random())
print(s.get_random())
print(s.get_random())
```

Running pylint with default config on it:
```
$ python -m pylint test.py 
************* Module test
test.py:24:4: W0235: Useless super delegation in method 'get_random' (useless-super-delegation)

Your code has been rated at 9.33/10 (previous run: 9.33/10, +0.00)
```

### Desired solution

pylint should not emit a warning in this case as the override has legitimate uses.

### Additional context

_No response_

timmartin:
## Type of Changes

## Description

In the `useless-super-delegation` checker, check whether the return type annotation has been specialized in the derived class. If so, then the override is not useless and we should not issue the warning.

Closes #5822

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1299,8 +1299,8 @@ a metaclass class method.",
                     return
 
             if (
-                function.returns is not None and
-                meth_node.returns is None
+                function.returns is not None
+                and meth_node.returns is None
                 or meth_node.returns.as_string() != function.returns.as_string()
             ):
                 # Override adds typing information to the return type
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1299,9 +1299,9 @@ a metaclass class method.",
                     return
 
             if (
-                function.returns is not None and
-                meth_node.returns is not None and
-                meth_node.returns.as_string() != function.returns.as_string()
+                function.returns is not None
+                and meth_node.returns is not None
+                and meth_node.returns.as_string() != function.returns.as_string()
             ):
                 # Override adds typing information to the return type
                 return
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1299,9 +1299,9 @@ a metaclass class method.",
                     return
 
             if (
-                function.returns is not None
-                and meth_node.returns is None
-                or meth_node.returns.as_string() != function.returns.as_string()
+                function.returns is not None and
+                meth_node.returns is not None and
+                meth_node.returns.as_string() != function.returns.as_string()
             ):
                 # Override adds typing information to the return type
                 return
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1298,6 +1298,14 @@ a metaclass class method.",
                 if called_annotations != overridden_annotations:
                     return
 
+            if (
+                function.returns is not None and
+                meth_node.returns is None
+                or meth_node.returns.as_string() != function.returns.as_string()
+            ):
+                # Override adds typing information to the return type
+                return
+
         if _definition_equivalent_to_call(params, args):
             self.add_message(
                 "useless-super-delegation", node=function, args=(function.name,)
--- a/tests/functional/u/useless/useless_super_delegation_py35.py
+++ b/tests/functional/u/useless/useless_super_delegation_py35.py
@@ -36,7 +36,7 @@ class NoReturnType:
 class ReturnTypeSpecified(NoReturnType):
     choices = ['a', 'b']
 
-    def draw(self) -> str:
+    def draw(self) -> str: # [useless-super-delegation]
         return super().draw()
 
 class ReturnTypeSame(ReturnTypeAny):
--- a/tests/functional/u/useless/useless_super_delegation_py35.py
+++ b/tests/functional/u/useless/useless_super_delegation_py35.py
@@ -10,3 +10,37 @@ class UselessSuper(object):
 
     def useless(self, first, *, second=None, **kwargs): # [useless-super-delegation]
         return super().useless(first, second=second, **kwargs)
+
+# pylint: disable=wrong-import-position
+import random
+from typing import Any
+
+class ReturnTypeAny:
+    choices = ['a', 1, (2, 3)]
+
+    def draw(self) -> Any:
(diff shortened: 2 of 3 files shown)


tests/functional/u/useless/useless_super_delegation_py3.py
```diff
@@ -41,3 +41,30 @@ def __init__(self, thing: object) -> None: # [useless-super-delegation]
...
     def __init__(self, _arg: List[int]) -> None:
         super().__init__()
+
+
+import random
+
+class ReturnTypeAny:
+    choices = ['a', 1, (2, 3)]
+
+    def draw(self) -> None:
```
DanielNoord, line 51: This annotation is incorrect, is that intentional?

timmartin: Well spotted, that wasn't intentional.

While I was fixing that, I realized there's another case we need to think about:

    import typing
    from typing import Any

    class Base:
        def foo(self) -> Any:
            pass

    class Derived:
        def foo(self) -> typing.Any:
            return super().foo()

This will be harder to fix, and it looks like the existing rules around type hints on parameters don't cope with this case either, so I suggest this can be handled as a later fix to both parameter type hints and return type hint.

Pierre-Sassoulas (approved):
LGTM.

DanielNoord:
@cdce8p I'd like to get your input here. I'm not sure if this change is reasonable. We're not a type checker, but should we support this? You're probably the best person to judge this.

tushar-deepsource:
I'd say this is a good change. It's a change that will lead to less false positives, and it's something that we can infer, so why not take advantage of it?

Pierre-Sassoulas:
@cdce8p could you check this, if you have some time ? It's blocking the release of 2.13.6 :)

pylint/checkers/classes/class_checker.py
```diff
@@ -1263,6 +1263,13 @@ def _check_useless_super_delegation(self, function):
 
         if meth_node is not None:
 
+            if function.returns is not None and (
+                meth_node.returns is None
+                or meth_node.returns.as_string() != function.returns.as_string()
+            ):
```
cdce8p, line 1269:
```py
            if (
                function.returns is not None
                and meth_node.returns is not None
                and meth_node.returns.as_string() != function.returns.as_string()
            ):
```

cdce8p (commented):
I'm not fully convinced it is all that useful, but I guess we can add it. Especially since we do something similar for the parameter annotations already.

Just two suggestions:
* Do `not` return if one function doesn't have a return type (similar to parameter annotations).
* Move it below the `if called_annotations and overridden_annotations` block.

timmartin:
I'm not sure what you mean by "Do not return if one function doesn't have a return type (similar to parameter annotations)." Currently my reasoning is that if the derived method adds a type missing from the base method we should not warn `useless-super-delegation`, but if the base method has the annotation and the derived method doesn't then we should warn. Is that what you're meaning?

cdce8p:
> I'm not sure what you mean by "Do not return if one function doesn't have a return type (similar to parameter annotations)." Currently my reasoning is that if the derived method adds a type missing from the base method we should not warn `useless-super-delegation`, but if the base method has the annotation and the derived method doesn't then we should warn. Is that what you're meaning?

My idea was to be careful not to block too many errors. I.e. my suggestion would be to only ignore `useless-super-delegation` if **both** the parent and child method have function annotations and they differ.

pylint/checkers/classes/class_checker.py
```diff
@@ -1298,6 +1298,14 @@ def form_annotations(arguments):
                 if called_annotations != overridden_annotations:
                     return
 
+            if (
+                function.returns is not None
+                and meth_node.returns is None
+                or meth_node.returns.as_string() != function.returns.as_string()
+            ):
```
cdce8p, line 1305:
```suggestion
                if function.returns is not None and (
                    meth_node.returns is None
                    or meth_node.returns.as_string() != function.returns.as_string()
                ):
```
To fix the tests, you would need to use the old condition.

For my https://github.com/PyCQA/pylint/pull/6141#issuecomment-1099627809, I suggested something a little bit different.
```suggestion
            if (
                function.returns is not None
                and meth_node.returns is not None
                and meth_node.returns.as_string() != function.returns.as_string()
            ):
```

cdce8p (commented):
Please don't force push changes new here. That will make it difficult / impossible to see and compare old revisions.

timmartin:
My thinking was that the most likely reason I could think to override a method just to change the return type annotation is that I was extending some library code that doesn't have type declarations and I wanted my library to have type declarations, in which case issuing a diagnostic isn't helpful. I'm not sure I care enough about this to prolong the discussion, though.

cdce8p:
> My thinking was that the most likely reason I could think to override a method just to change the return type annotation is that I was extending some library code that doesn't have type declarations and I wanted my library to have type declarations, in which case issuing a diagnostic isn't helpful. I'm not sure I care enough about this to prolong the discussion, though.

Usually, I think it would be better if the original library added type hints itself. However, I do agree that it's not worth the discussion. If you think it would be helpful, I guess we could do that, too. Will leave that up to you.

timmartin:
No, I'm done with this change. Merge it or close the PR as you see fit.

cdce8p (approved):

Pierre-Sassoulas:
Thank you for the fix @timmartin !

PR #6142, opened 2022-04-02
Exclude ``unnecessary-dunder-call`` calls on uninstantiated classes and within dunder funcdefs

### LINKED ISSUE #6074: False positive `unnecessary-dunder-call` when overriding dunder method
### Bug description

```python
class A(list):
    def __contains__(self, item):
        print("do some special checks")
        return super().__contains__(item)  # raises unnecessary-dunder-call
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:4: [C2801(unnecessary-dunder-call), A.__contains__] Unnecessarily calls dunder method __contains__. Use in keyword.
```

### Expected behavior

No message, since the example cannot be rewritten as `return item in super()`:
```
TypeError: argument of type 'super' is not iterable
```

This wasn't happening for `__init__` because it was allow-listed, but I think all dunder methods could be overridden in subclasses, so they all need to be made compatible with `super().__...`

### Pylint version

```shell
pylint 2.14.0-dev0
astroid 2.12.0-dev0
Python 3.10.2 (v3.10.2:a58ebcc701, Jan 13 2022, 14:50:16) [Clang 13.0.0 (clang-1300.0.29.30)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jpy-git:
## Type of Changes

## Description

As described in this [comment](https://github.com/PyCQA/pylint/issues/6074#issuecomment-1086664202) it's possible (if not stylistically odd) to call dunder method on uninstantiated classes so we exclude these

Closes #6074


jpy-git:
Marking this as draft whilst I implement this [comment](https://github.com/PyCQA/pylint/issues/6074#issuecomment-1086741459)

pylint/checkers/dunder_methods.py
```diff
@@ -17,20 +18,22 @@ class DunderCallChecker(BaseChecker):
...
+    Additionally we exclude uninstantiated classes since these
+    might be used to access the dunder methods of a base class of an instance.
     """
 
     __implements__ = IAstroidChecker
 
     includedict = {
+        "__init__": "Instatiate class directly",
+        "__del__": "Use del keyword",
         "__repr__": "Use repr built-in function",
```
jpy-git, line 37: We were only excluding these before since these were being used within dunder method defs. Now that we exclude dunder calls within dunder method defs in general, we can flag these bad uses as well 👍

pylint/checkers/dunder_methods.py
```diff
@@ -136,23 +139,41 @@ class DunderCallChecker(BaseChecker):
...
+        parent = node.parent
+        while parent is not None:
+            if (
+                isinstance(parent, nodes.FunctionDef)
+                and parent.name.startswith("__")
+                and parent.name.endswith("__")
+            ):
+                return True
+            parent = parent.parent
+        return False
```
jpy-git, line 154: Logic is super simple, look up parent stack and check if we are within some sort of dunder method def.

N.B. I've not put any constraints on it having to be the same dunder method, since we can do things like
```python
class A:
    def __getattr__(self, name):
        ...
        return self.__getattribute__(name)
```
Think this approach is sensible

pylint/checkers/base_checker.py
```diff
@@ -61,7 +61,7 @@ def __init__(
 
     def __gt__(self, other):
         """Permit to sort a list of Checker by name."""
-        return f"{self.name}{self.msgs}" > f"{other.name}{other.msgs}"
+        return f"{self.name}{self.msgs}".__gt__(f"{other.name}{other.msgs}")
```
jpy-git, line 64: reverting this from the original PR to match our improved approach 👍

jpy-git:
@cdce8p I've updated this PR to allow dunder method calls within dunder method funcdefs 😄

ChangeLog
```diff
@@ -10,6 +10,10 @@ Release date: TBA
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
 
+* Exclude ``unnecessary-dunder-call`` on uninstantiated classes and within dunder funcdefs.
+
+  Closes #6074
+
 * Exclude dunder calls on super() from raising ``unnecessary-dunder-call``.
 
   Closes #6074
```
Pierre-Sassoulas, line 19:
```suggestion
```
I think we can remove fixes we do before releasing from the changelog and add the "Closes #6074" after Closes #5936 line 56.

Pierre-Sassoulas (commented):
Not sure about authorizing dunder method with a different name if they are used in dunder method definition. ...But the implementation could get complicated fast if we need to define what dunder methods are compatibles together so it's probably not worth it to forbid it.

cdce8p:
> @cdce8p I've updated this PR to allow dunder method calls within dunder method funcdefs 😄

Thanks for the heads up! I've tested your changes against Home Assistant. Seems to work just as intended.

One question though, did you exclude `object.__setattr__`? With the last changes, pylint doesn't emit a warning for this example anymore. _I'm honestly not sure if it should have emitted one to begin with._
```py
class Area:
    def func(self) -> None:
        object.__setattr__(self, "id", 42)
```

Additionally, another example. It's an edge case and tbh I'm not sure anyone would do something like that in a dunder method, but it might be worth adding it as test case never the less.
```py
class X:
    def __init__(self):
        self.x = object()

    def __setattr__(self, name, value):
        def nested_function():
            self.x.__setattr__(name, value)  # Should be fine (probably)

        nested_function()
```

--
Thanks again @jpy-git! Just a heads up, unfortunately I don't have time for a full review at the moment. Maybe someone else can do that.

jpy-git:
> One question though, did you exclude `object.__setattr__`? With the last changes, pylint doesn't emit a warning for this example anymore. _I'm honestly not sure if it should have emitted one to begin with._
> 
> ```python
> class Area:
>     def func(self) -> None:
>         object.__setattr__(self, "id", 42)
> ```

Yes this is intended, it's the ignore dunder calls on uninstantiated classes exception I added in this PR. There are some unit tests that show this behaviour.

i.e. don't flag this
```python
object.__setattr__(self, "id", 42)
```
but flag this
```python
object().__setattr__("id", 42)
```
the reason being that if someone is calling the dunder method of a class then they are likely calling one of the base classes of self which is different behaviour from calling the same dunder method of self.

e.g.
```python
from collections import OrderedDict
odict = OrderedDict()

# These two `__setitem__` calls are not the same
dict.__setitem__(odict, "key", "value")
odict.__setitem__("key", "value")

# Only the second one can be rewritten as
odict["key"] = "value"
```

jpy-git:
> Additionally, another example. It's an edge case and tbh I'm not sure anyone would do something like that in a dunder method, but it might be worth adding it as test case never the less.
> 
> ```python
> class X:
>     def __init__(self):
>         self.x = object()
> 
>     def __setattr__(self, name, value):
>         def nested_function():
>             self.x.__setattr__(name, value)  # Should be fine (probably)
> 
>         nested_function()
> ```

This one is already covered since I work up the parent stack to check if we are at any depth within a dunder method. I've added this as an extra unit test 😄

jpy-git:
> Not sure about authorizing dunder method with a different name if they are used in dunder method definition. ...But the implementation could get complicated fast if we need to define what dunder methods are compatibles together so it's probably not worth it to forbid it.

I think it's a fairly safe bet that someone defining dunder methods within dunder method funcdefs probably knows what they're doing so agree it's not worth adding extra complexity of all possible combinations.

Pierre-Sassoulas (approved):
LGTM, nice change @jpy-git !

cdce8p:
> Yes this is intended, it's the ignore dunder calls on uninstantiated classes exception I added in this PR. There are some unit tests that show this behaviour.
> 
> i.e. don't flag this
> 
> ```python
> object.__setattr__(self, "id", 42)
> ```

👍🏻

> This one is already covered since I work up the parent stack to check if we are at any depth within a dunder method. I've added this as an extra unit test 😄

Saw your initial comment about it. While thinking about ways it can break, I came up with the example.

> > Not sure about authorizing dunder method with a different name if they are used in dunder method definition. ...But the implementation could get complicated fast if we need to define what dunder methods are compatibles together so it's probably not worth it to forbid it.
> 
> I think it's a fairly safe bet that someone defining dunder methods within dunder method funcdefs probably knows what they're doing so agree it's not worth adding extra complexity of all possible combinations.

👍🏻

pylint/checkers/dunder_methods.py
```diff
@@ -17,20 +18,24 @@ class DunderCallChecker(BaseChecker):
...
-    Additionally we exclude dunder method calls on super() since
+    Additionally we exclude uninstantiated classes since these
+    might be used to access the dunder methods of a base class of an instance.
+    We also exclude dunder method calls on super() since
     these can't be written in an alternative manner.
     """
 
     __implements__ = IAstroidChecker
 
     includedict = {
```
cdce8p, line 36: `includedict` isn't modified. There is no need for it to be an instance attribute.

pylint/checkers/dunder_methods.py
```diff
@@ -136,23 +141,46 @@ class DunderCallChecker(BaseChecker):
...
             and not (
                 isinstance(node.func.expr, nodes.Call)
                 and isinstance(node.func.expr.func, nodes.Name)
                 and node.func.expr.func.name == "super"
             )
         ):
+            inf_expr = safe_infer(node.func.expr)
+            if inf_expr not in (None, Uninferable) and not isinstance(
+                inf_expr, Instance
+            ):
```
cdce8p, line 173: This maybe?
```suggestion
            if not (
                inf_expr is None
                or inf_expr is Uninferable
                or isinstance(inf_expr, Instance)
            ):
```

tests/functional/u/unnecessary/unnecessary_dunder_call.py
```diff
@@ -27,12 +42,39 @@ class Bar2(object):
...
-            return subklass
-        return object
+class CustomRegistry(dict):
+    def __init__(self) -> None:
+        super().__init__()
+        self._entry_ids = {}
+
+    def __setitem__(self, key, entry) -> None:
+        super().__setitem__(key, entry)
+        self._entry_ids.__setitem__(entry.id, entry)
```
cdce8p, line 52: Could you add a test case where the call function name is different to the parent function name? I.e. `__delitem__` inside `__setitem__`.

cdce8p (commented):
Looks good overall. Just some small comments

cdce8p (approved):
I've pushed some last changes. Looks good now.

PR #6143, opened 2022-04-02
Work towards turning on ``future_option_parsing`` in ``PyLinter``

DanielNoord:

## Type of Changes

## Description

This is best viewed from the commit view.

Tried to keep it as short as possible, but it's much easier to do this by flipping the switch once. I'll annotate a little bit.

Essentially after this PR all callback options such as ``--generate-rc-file`` are now handled by `argparse`.

### DIFF
--- a/pylint/checkers/unicode.py
+++ b/pylint/checkers/unicode.py
@@ -100,7 +100,7 @@ BAD_CHARS = [
         "\\x1A",
         "E2512",
         (
-            "Ctrl+Z “End of text” on Windows. Some programs (such as type) ignore "
+            'Ctrl+Z "End of text" on Windows. Some programs (such as type) ignore '
             "the rest of the file after it."
         ),
     ),
--- a/pylint/config/argument.py
+++ b/pylint/config/argument.py
@@ -11,10 +11,22 @@ An Argument instance represents a pylint option to be handled by an argparse.Arg
 import argparse
 import pathlib
 import re
-from typing import Callable, Dict, List, Optional, Pattern, Sequence, Tuple, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Pattern,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 
 from pylint import interfaces
 from pylint import utils as pylint_utils
+from pylint.config.callback_actions import _CallbackAction
 
 _ArgumentTypes = Union[
     str,
@@ -199,3 +211,32 @@ class _StoreTrueArgument:
         # argparse uses % formatting on help strings, so a % needs to be escaped
         self.help = arg_help.replace("%", "%%")
         """The description of the argument."""
+
+
+class _CallableArgument:
+    """Class representing an callable argument to be parsed by an argparse.ArgumentsParser.
+
+    This is based on the parameters passed to argparse.ArgumentsParser.add_message.
+    See:
+    https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument
+    """
+
+    def __init__(
+        self,
+        flags: List[str],
+        action: Type[_CallbackAction],
+        arg_help: str,
+        kwargs: Dict[str, Any],
+    ) -> None:
+        self.flags = flags
+        """The name of the argument."""
+
+        self.action = action
+        """The action to perform with the argument."""
+
+        # argparse uses % formatting on help strings, so a % needs to be escaped
+        self.help = arg_help.replace("%", "%%")
+        """The description of the argument."""
+
+        self.kwargs = kwargs
+        """Any additional arguments passed to the action."""
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -130,12 +130,8 @@ class _MessageHelpAction(_CallbackAction):
         values: Union[str, Sequence[str], None],
         option_string: Optional[str] = "--help-msg",
     ) -> None:
-        if isinstance(values, (list, tuple)):
-            self.run.linter.msgs_store.help_message(values)
-        elif isinstance(values, str):
-            self.run.linter.msgs_store.help_message(utils._splitstrip(values))
-        else:
-            raise argparse.ArgumentTypeError(f"{values} should be a string.")
+        assert isinstance(values, (list, tuple))
+        self.run.linter.msgs_store.help_message(values)
         sys.exit(0)
 
 
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -2,11 +2,20 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
+# pylint: disable=too-many-arguments
+
 """Callback actions for various options."""
 
 import abc
 import argparse
-from typing import Any, Optional, Sequence, Union
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Optional, Sequence, Union
+
+from pylint import extensions, interfaces, utils
+
+if TYPE_CHECKING:
+    from pylint.lint.run import Run
 
 
 class _CallbackAction(argparse.Action):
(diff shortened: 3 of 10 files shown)


pylint/config/callback_actions.py
```diff
@@ -38,3 +47,261 @@ def __call__(
...
+        const: None = None,
+        default: None = None,
+        arg_type: None = None,
+        choices: None = None,
+        required: bool = False,
+        help_msg: str = "",
+        metavar: str = "",
+        **kwargs: "Run",
+    ) -> None:
+        self.run = kwargs["Run"]
```
DanielNoord, line 69: This is a trick to be able to pass the `Run` object to these callbacks. At some point I might investigate whether alle these arguments are actually needed.

pylint/config/callback_actions.py
```diff
@@ -38,3 +47,261 @@ def __call__(
...
+
+        super().__init__(
+            option_strings,
+            dest,
+            0,
```
DanielNoord, line 74: Since these are callbacks they _should_ never require any arguments.

pylint/config/callback_actions.py
```diff
@@ -38,3 +47,261 @@ def __call__(
...
+        required: bool = False,
+        help_msg: str = "",
+        metavar: str = "",
+        **kwargs: "Run",
+    ) -> None:
+        self.run = kwargs["Run"]
+        super().__init__(
+            option_strings,
+            dest,
+            "+",
```
DanielNoord, line 116: `--help-msg` was the only callback that uses arguments.

However, to be able to work with `argparse` we do need to change one thing. We can no longer allow:
``pylint --help-msg=missing-docstring``
But only:
``pylint --help-msg missing-docstring``.

This is because callable actions cannot also store a value like the `=` operator does. I have changed the tests accordingly.

pylint/config/option.py
```diff
@@ -194,6 +194,8 @@ def _check_choice(self):
     optparse.Option.CHECK_METHODS[2] = _check_choice  # type: ignore[index]
 
     def process(self, opt, value, values, parser):
+        if self.callback and self.callback.__module__ == "pylint.lint.run":
```
DanielNoord, line 197: This makes it so that whenever we encounter a callback function defined in our own `Run` file we don't execute. Other callback defines by plugins are still being executed.

This is not necessarily a problem: their callbacks are simply "executed" by `optparse`. I thought it would be a good milestone for us to let `argparse` handle this.

pylint/config/options_provider_mixin.py
```diff
@@ -72,10 +73,12 @@ def set_option(self, optname, value, action=None, optdict=None):
                 setattr(self.config, optname, _list + (value,))
             else:
                 _list.append(value)
-        elif action == "callback":
-            optdict["callback"](None, optname, value, None)
-        elif not isinstance(action, str):
-            optdict["callback"](None, optname, value, None)
+        elif (
```
DanielNoord, line 76: Similar as above. This makes it so that `optparse` no longer fires prematurely.

I don't think these options should fire anyway. If you using `set_option` to call a callback function you're doing something wrong anyway.

pylint/config/utils.py
```diff
@@ -7,38 +7,54 @@
...
     # pylint: disable-next=fixme
-    # TODO: This should be removed once the migration to argparse is finished
-    for key, value in optdict.items():
-        if key not in IMPLEMENTED_OPTDICT_KEYS:
-            print("Unhandled key found in Argument creation:", key)  # pragma: no cover
-            print("It's value is:", value)  # pragma: no cover
+    # TODO: Do something with the 'group', 'level' and 'hide' keys of optdicts
+
+    # pylint: disable-next=fixme
+    # TODO: Do something with the 'dest' key and deprecation of options
```
DanielNoord, line 22: This is the last step (I think) before we can start using `namespace` in `PyLinter` as well.

pylint/config/utils.py
```diff
@@ -7,38 +7,54 @@
...
+    # Get the long and short flags
+    flags = [f"--{opt}"]
+    if "short" in optdict:
+        flags += [f"-{optdict['short']}"]
+
+    # Get the action type
     action = optdict.get("action", "store")
+
+    # pylint: disable-next=fixme
+    # TODO: Remove this handling after we have deprecated multiple-choice arguments
```
DanielNoord, line 33: `multiple-choice` values like `confidence` are not allowed in `argparse`. You can only have a `choice` or nothing.

Our transformer does the same as `multiple-choice` does, but `argparse` trips up as soon as there is a `choices` attribute. I think we should deprecate `multiple-choice`, as it is unmaintainable.

pylint/lint/pylinter.py
```diff
@@ -620,7 +620,7 @@ def __init__(
 
         reporters.ReportsHandlerMixIn.__init__(self)
         config.OptionsManagerMixIn.__init__(self, usage=__doc__)
-        checkers.BaseTokenChecker.__init__(self)
+        checkers.BaseTokenChecker.__init__(self, self, future_option_parsing=True)
```
DanielNoord, line 623: 🎉

tests/test_self.py
```diff
@@ -475,7 +475,9 @@ def test_pylintrc_plugin_duplicate_options(self) -> None:
         self._test_output(
             [
                 f"--rcfile={config_path}",
-                "--help-msg=dummy-message-01,dummy-message-02",
+                "--help-msg",
```
DanielNoord, line 478: This is the change I mentioned.

pylint/config/utils.py
Pierre-Sassoulas: Yeah also it makes more sense imo to use "this confidence and higher confidences" from a strictly usability point of view. Who want to get UNDEFINED and HIGH confidences, with no in between ?

Pierre-Sassoulas: I don't even know if the feature is used at all because as you know we did not specify the confidence until very recentely.

DanielNoord: That would be a different, although I agreed sensible, change!

DanielNoord: > I don't even know if the feature is used at all because as you know we did not specify the confidence until very recentely.

I'm scared plugin writers might use it... 😢

pylint/config/argument.py
```diff
@@ -199,3 +211,32 @@ def __init__(
         # argparse uses % formatting on help strings, so a % needs to be escaped
         self.help = arg_help.replace("%", "%%")
         """The description of the argument."""
+
+
+class _CallableArgument:
```
Pierre-Sassoulas, line 216: Regarding the callback (is it Callable or Callback ?), could we move the "callback" to "before we instantiate a pylinter ? I wouldn't mind just having a long list of
```
if args.help:
    print_help()
if args.version:
    print_version()
...
PyLinter(...).run(args)
```
I remember that it was really messy and unnecessary, except maybe for init-hook.

pylint/config/callback_actions.py
Pierre-Sassoulas: That's a small breaking change, we might want to put this in the changelog but I guess no one is using the help message in a script, so...

pylint/config/callback_actions.py
```diff
@@ -38,3 +47,261 @@ def __call__(
...
+
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Union[str, Sequence[Any], None],
+        option_string: Optional[str] = "--long-help",
+    ) -> None:
+        print(self.run.linter.help(1))
+        sys.exit(0)
```
Pierre-Sassoulas, line 307: Nevermind my last remark about callbacks, you did a lot of work here.

pylint/config/utils.py
Pierre-Sassoulas: > I'm scared plugin writers might use it...

Yeah if it wasn't such a low level API that is probably used somewhere, we could change it to an ordered enum and get a lot of simplification. I think it would create a bigger mess than just working with what currently exists though.

pylint/config/argument.py
DanielNoord: No, we need to the `PyLinter` for stuff like ``list-messages`` and ``--help-msg``.

I have called the `Argument` `Callable` since they are callable arguments. I have named the actions `Callback` as that was the terminology previously used.

pylint/config/callback_actions.py
DanielNoord: Yeah I think this would be fine without one. But let me know if you want me to add one!

Pierre-Sassoulas (changes requested):
Some functions in ``run`` can be removed I think (lost coverage so probably unused now). I think it's private API and it should not be used by downstream library.

DanielNoord:
> Some functions in `run` can be removed I think (lost coverage so probably unused now). I think it's private API and it should not be used by downstream library.

I'm a bit worried they can't still get called by some weird `optparse` trick. As long as they are in `callback` they can be accessed by `optparse`. I thought about removing them and the `callback` `keys` in the `optdicts` but then we don't have any code that checks whether our code is still compatible with "old" optdicts...

Not sure what the best approach is here.

Pierre-Sassoulas:
I would be in favor of covering everything that is reasonably coverable, because we don't have a lot of tests for calling pylint with various options right now, so I'm a little worried of the result of the release if we don't. primers won't help for this, and there is a lot of existing scripts calling pylint that will potentially break at once at release.

DanielNoord:
> I would be in favor of covering everything that is reasonably coverable, because we don't have a lot of tests for calling pylint with various options right now, so I'm a little worried of the result of the release if we don't. primers won't help for this, and there is a lot of existing scripts calling pylint that will potentially break at once at release.

I have added tests for all callbacks and made it so that we raise `UnimplementedError` when any of the old callbacks gets called.

tests/test_self.py
```diff
@@ -1321,11 +1281,190 @@ def test_regression_recursive_current_dir(self):
...
+            run = Run(["--errors-only"])
+            assert run.linter._error_mode
+
+    @staticmethod
+    def test_verbose() -> None:
+        """Test the --verbose flag."""
+        with pytest.raises(SystemExit):
+            run = Run(["--verbose"])
+            assert run.verbose
+
```
Pierre-Sassoulas, line 1477: I don't know why I can't make suggestion here, but we can parametrize most of the test using:

```python
    @pytest.mark.parametrize("options,expecteds", [
        [["--full-documentation"], ["Pylint global options and switches"]]
        # etc.
    ])
    def test_pylint_option(options,expecteds) -> None:
        process = subprocess.run(
            ["pylint"] + options,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            check=False,
        )
       output =  process.stdout.decode("utf-8")
        for expected in expecteds:
            assert expected in output
```

DanielNoord: Does that make the tests more readable in your opinion? I think the current tests are a bit clearer. They also allow easy expanding such as `test_help_msg` does.

Pierre-Sassoulas: I'm not advocating to put all the options here, only the easy one where we check the output and that it's not outright crashing. We do need to copy paste for test that need to be extended.

carlijoy made some good looking parametrize here https://github.com/PyCQA/pylint/blob/main/tests/checkers/unittest_unicode/unittest_invalid_encoding.py#L91 that could make the list look better.

DanielNoord:
@Pierre-Sassoulas Do we have someone on Windows who can help debug this? I really don't know what is causing issues here..

Pierre-Sassoulas:
> Do we have someone on Windows who can help debug this? I really don't know what is causing issues here..

Maybe I'm not remembering well but I think areveny does.

Before pinging her, I'm thinking it might be due to ``process.stdout.decode(sys.getdefaultencoding())``,  I guess ``sys.getdefaultencoding()`` might not return ``utf8`` on Windows but cp850 (cp something). We could try to change the default encoding ourselves with ``locale.setlocale`` for Windows tests ?

DanielNoord:
> > Do we have someone on Windows who can help debug this? I really don't know what is causing issues here..
> 
> 
> 
> Maybe I'm not remembering well but I think areveny does.
> 
> 
> 
> Before pinging her, I'm thinking it might be due to ``process.stdout.decode(sys.getdefaultencoding())``,  I guess ``sys.getdefaultencoding()`` might not return ``utf8`` on Windows but cp850 (cp something). We could try to change the default encoding ourselves with ``locale.setlocale`` for Windows tests ? 

I thought by using the get method we would whatever it currently is? Looking at the test output it still seems to be utf though

tests/test_self.py
```diff
@@ -1319,11 +1281,220 @@ def test_regression_recursive_current_dir(self):
...
+    def test_list_msgs() -> None:
+        """Test the --list-msgs flag."""
+        process = subprocess.run(
+            [
+                sys.executable,
+                "-m",
+                "pylint",
+                "--list-msgs",
+            ],
+            stdout=subprocess.PIPE,
```
Pierre-Sassoulas, line 1329: Not sure about this but worth a try I guess ? 
```suggestion
            stdout=subprocess.PIPE,
            encoding="utf8",
```
https://docs.python.org/3/library/subprocess.html#subprocess.run

> If encoding or errors are specified, or text is true, file objects for stdin, stdout and stderr are opened in text mode using the specified encoding and errors or the [io.TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper) default. The universal_newlines argument is equivalent to text and is provided for backwards compatibility. By default, file objects are opened in binary mode.

Pierre-Sassoulas:
The error we got in mypy seems to indicate that it "worked" and we have a string instead of the bytes we needed to decode.

tests/test_self.py
DanielNoord: Still crashes :(

I can't really figure out what the character is we're failing on...

Pierre-Sassoulas: Can we print the output ?

tests/test_self.py
DanielNoord: Done!

tests/test_self.py
DanielNoord: No since the error is now happening inside `run()`

DanielNoord: Got access to a windows laptop. Turns out we were using an unsupported character.

Pierre-Sassoulas: Which one, the suspense is killing me ? 😄

pylint/checkers/unicode.py
```diff
@@ -100,7 +100,7 @@ def human_code(self) -> str:
         "\\x1A",
         "E2512",
         (
-            "Ctrl+Z “End of text” on Windows. Some programs (such as type) ignore "
+            'Ctrl+Z "End of text" on Windows. Some programs (such as type) ignore '
             "the rest of the file after it."
```
Pierre-Sassoulas, line 104: ``“`` ?

DanielNoord: Yeah! I don't know why and how because some times they could print but I think they can't print to the console of Windows?
When I ran the test in the "test explorer" of VS Code on Windows the test even passed, but in the terminal it crashed. This should fix it though 😄

pylint/config/callback_actions.py
```diff
@@ -38,3 +47,261 @@ def __call__(
...
+        namespace: argparse.Namespace,
+        values: Union[str, Sequence[str], None],
+        option_string: Optional[str] = "--help-msg",
+    ) -> None:
+        if isinstance(values, (list, tuple)):
+            self.run.linter.msgs_store.help_message(values)
+        elif isinstance(values, str):
+            self.run.linter.msgs_store.help_message(utils._splitstrip(values))
+        else:
+            raise argparse.ArgumentTypeError(f"{values} should be a string.")
```
Pierre-Sassoulas, line 138:
```suggestion
```
Isn't all arguments from argparse string or list ? We could add a test for this otherwise.

pylint/config/callback_actions.py
```diff
@@ -38,3 +47,261 @@ def __call__(
...
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Union[str, Sequence[str], None],
+        option_string: Optional[str] = "--help-msg",
+    ) -> None:
+        if isinstance(values, (list, tuple)):
+            self.run.linter.msgs_store.help_message(values)
+        elif isinstance(values, str):
+            self.run.linter.msgs_store.help_message(utils._splitstrip(values))
```
Pierre-Sassoulas, line 136: This part is untested, maybe we could add a test for it.

pylint/config/callback_actions.py
```diff
@@ -38,3 +47,261 @@ def __call__(
...
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Union[str, Sequence[Any], None],
+        option_string: Optional[str] = "--verbose",
+    ) -> None:
+        self.run.verbose = True
+
+
+class _EnableAllExtensionsAction(_AccessRunObjectAction):
```
Pierre-Sassoulas, line 279: We could add a tests for this Action as it's untested right now.

Pierre-Sassoulas (changes requested):
I think it's really nice that we caught an undetected windows bug, it's also very nice that the coverage went up 0.2% (probably the reason why we caught the bug in the first place).The new design is also a lot cleaner imo.

pylint/config/callback_actions.py
DanielNoord: Removed and changed tests to "catch" this case. I doesn't go here as `argparse` errors out earlier when the `values` is `None`.

pylint/config/callback_actions.py
DanielNoord: Changed implementation. Couldn't get this to trigger from command line and also not from `subprocess`. So should be fine!

pylint/config/callback_actions.py
DanielNoord: We can't since this is done via the `preprocessing` thing in `Run`. We should probably do that via these callbacks as well, but let's do that in another PR.

DanielNoord:
@Pierre-Sassoulas Sorry for all the force-pushes. Want to keep the history clean for a rebase.

pylint/config/callback_actions.py
DanielNoord: Same goes for `--verbose` by the way.

Pierre-Sassoulas (approved):
Thank you that is a significant step toward using argparse and deprecating optparse. I'll let you rebase merge when you think it's ready, I do think that it's important to have separate commits here.

PR #6144, opened 2022-04-02
Refactor all ``_Argument`` classes

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Blocked by #6143.

Now that I have done all options I now which `Arguments` we support. The last two commits in this PR fix up their inheritance cycle.

PR #6146, opened 2022-04-02
Refactor the help message

DanielNoord:

## Type of Changes

## Description

Blocked by #6143 and #6144

The second to last commit is a short refactor. The other is a refactor of how we generate the help message. I have chosen to deprecate the `level` option as it makes everything way more complicated than it needs to be. Some of the options with `level==1` should arguably be displayed always and it is pretty difficult to retrieve their original `help` message after they have been parsed as `SUPRESS` previously.
I think this makes it a lot simpler and doesn't change that much for the end-user.

### DIFF
--- a/pylint/config/argument.py
+++ b/pylint/config/argument.py
@@ -151,13 +151,19 @@ class _Argument:
     https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument
     """
 
-    def __init__(self, *, flags: List[str], arg_help: str) -> None:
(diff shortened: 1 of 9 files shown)


Pierre-Sassoulas (commented):
I did not do a full review because there is a lot of change but I think the coverage decreasing is an indication that we can remove some code that became dead. For example ``_expand_default`` in the OptionManagerMixin but I did not check everything.

Pierre-Sassoulas (approved):
It's nice that we're decoupling the help generation from the run class.

Issue #6087, opened 2022-04-01
Re-enable ``bad-docstring-quotes`` for Python 3.7


DanielNoord:
### Current problem

With changes to `doc_node` in astroid we can now check docstring quotes in `3.7` as well. This was turned off in #5526.

### Desired solution

Turn the checker on again.

### Additional context

@kasium Do you still want to do this? Should I assign you?

kasium:
@DanielNoord thanks for checking. Seems I was too slow 😄

PR #5526, opened 2021-12-14
Disable bad-docstring-quotes for <= 3.7

kasium:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

In python 3.7 and below, the ast parser might return wrong line numbers.

Closes #3077

PR #6150, opened 2022-04-03
Add example for `bad-docstring-quotes` message documentation

gunungpw:
## Type of Changes

## Description

- add code example for `bad-docstring-quotes`
- add link for PEP 257 and PEP 8

Ref #5953 #6087 #5526

### DIFF
--- a/doc/data/messages/b/bad-docstring-quotes/bad.py
+++ b/doc/data/messages/b/bad-docstring-quotes/bad.py
@@ -1,0 +1,5 @@
+# fmt: off
+def square_root():
+    '''Return the square root of input.'''  # [bad-docstring-quotes]
+    pass
+# fmt: on
--- a/doc/data/messages/b/bad-docstring-quotes/bad.py
+++ b/doc/data/messages/b/bad-docstring-quotes/bad.py
@@ -1,5 +1,4 @@
 # fmt: off
 def square_root():
     '''Return the square root of input.'''  # [bad-docstring-quotes]
-    pass
 # fmt: on
--- a/doc/data/messages/b/bad-docstring-quotes/good.py
+++ b/doc/data/messages/b/bad-docstring-quotes/good.py
@@ -1,0 +1,3 @@
+def square_root():
+    """Return the square root of input."""
+    pass
--- a/doc/data/messages/b/bad-docstring-quotes/good.py
+++ b/doc/data/messages/b/bad-docstring-quotes/good.py
@@ -1,3 +1,2 @@
 def square_root():
     """Return the square root of input."""
-    pass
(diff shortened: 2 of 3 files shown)


gunungpw:
pre-commit don't want black `# fmt: off` so I can give `''' ... '''` example. Help, any advice for this problem?

doc/data/messages/b/bad-docstring-quotes/bad.py
```diff
@@ -0,0 +1,4 @@
+# fmt: off
+def square_root():
+    '''Return the square root of input.'''  # [bad-docstring-quotes]
+# fmt: on
```
Pierre-Sassoulas, line 4:
```suggestion
def square_root():
    "Return the square root of input."  # [bad-docstring-quotes]
```

gunungpw: 😄 Wow I don't know that will work, thanks

Pierre-Sassoulas: Well that works but it's not really a docstring, just a piece of code that does nothing thus the message 😄. It's like 
```python
1 # Could be 'a = 1' and be useful
```
Just a declaration that does nothing.

jpy-git: FYI both of these examples are actually valid as docstrings.
```python
>>> def foo():
...     "I am a docstring"
...
>>> foo.__doc__
'I am a docstring'
```
```python
>>> def foo():
...     '''I am a docstring'''
...
>>> foo.__doc__
'I am a docstring'
```
This more of a style check than functionality 😄

jpy-git: > Just a declaration that does nothing.

This is actually true if someone tries to use an f-string as a docstring though 😄 

```python
>>> def foo():
...     f"""f-strings can't be docstrings"""
...
>>> foo.__doc__
>>>
```

We have a check in `flake8-bugbear` that warns about this if the first statement is a pointless f-string statement since you can assume that the user probably wanted to use it as a docstring. Potentially could add that as another future pylint enhancement

Pierre-Sassoulas: Hmm, interesting, I did not know. The three quotes standard is so prevalent that I did not even try !

We do have a ``W0104: Statement seems to have no effect (pointless-statement)``  for 
```
def f():
    f'''' ''''
````
 and  nothing for 
```
def f():
    " "
````
So a contributor knew at some point 😄

jpy-git: yeah I had the same reaction when I found out, I only know about it because of that f-string check 😄 

> So a contributor knew at some point 😄

It's because the quotes in that example go in the `doc_node` attribute in the AST so likely the writer of W0104 had no idea haha

PR #6073, opened 2022-03-31
Account for more node types in handling of except block homonyms with comprehensions

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Several nodes types didn't meet the allow list for homynym handling b/w except blocks and comprehensions.

Closes #6069

PR #6154, opened 2022-04-03
Fix handling of "for x in x" homonyms

### LINKED ISSUE #6136: False positive `unused-variable` for homonyms between comprehension variable and outer container
### Bug description

```python
def f():
    project_id = []  # this ought to be called project_ids, but...
    assert (project_id > 0 for project_id in project_id)
```

### Configuration

_No response_

### Command used

```shell
pylint z.py
```

### Pylint output

```shell
************* Module z
z.py:2:4: W0612: Unused variable 'project_id' (unused-variable)

Your code has been rated at 6.67/10 (previous run: 10.00/10, -3.33)
```

### Expected behavior

No msg

### Pylint version

Regression in #6073. Found it by diffing primer package messages.

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #6136

Extends the fix from #6073. (There, I had to be careful to not emit `unused-argument`, since that's all the test suite alerted me to, but with diffing primer output locally, I noticed the same deal with `unused-variable`. So I handle that the same way.)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -66,6 +66,11 @@ Release date: TBA
 
   Closes #6069
 
+* Fix false positive for ``unused-variable`` (in unreleased development) for
+  patterns like ``[_ for same_name in same_name]``.
+
+  Closes #6136
+
 * Narrow the scope of the ``unnecessary-ellipsis`` checker to:
   * functions & classes which contain both a docstring and an ellipsis.
   * A body which contains an ellipsis ``nodes.Expr`` node & at least one other statement.
--- a/ChangeLog
+++ b/ChangeLog
@@ -65,12 +65,7 @@ Release date: TBA
   subscripts in comprehensions.
 
   Closes #6069
-
-* Fix false positive for ``unused-variable`` (in unreleased development) for
-  patterns like ``[_ for same_name in same_name]``.
-
   Closes #6136
-
 * Narrow the scope of the ``unnecessary-ellipsis`` checker to:
   * functions & classes which contain both a docstring and an ellipsis.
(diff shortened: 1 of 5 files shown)


tests/functional/u/undefined/undefined_variable_py38.py
```diff
@@ -123,7 +123,7 @@ def type_annotation_used_after_comprehension():
 
 def type_annotation_unused_after_comprehension():
     """https://github.com/PyCQA/pylint/issues/5326"""
-    my_int: int  # [unused-variable]
+    my_int: int
```
jacobtylerwalls, line 126: This message was added in #5711 without much conviction. It probably shouldn't be here at all. Removing this is a better fix for the issue in #5326, which is a ticket I opened to say _hey we started emitting a message here!_

ChangeLog
```diff
@@ -66,6 +66,11 @@ Release date: TBA
 
   Closes #6069
 
+* Fix false positive for ``unused-variable`` (in unreleased development) for
+  patterns like ``[_ for same_name in same_name]``.
+
+  Closes #6136
+
```
Pierre-Sassoulas, line 73:
```suggestion
  Closes #6136
```
I suppose we can reduce the changelog if it's all fixed before we release.

Pierre-Sassoulas (approved):
👍

ChangeLog
```diff
@@ -65,7 +65,7 @@ Release date: TBA
   subscripts in comprehensions.
 
   Closes #6069
-
+  Closes #6136
```
Pierre-Sassoulas, line 68:
```suggestion
  Closes #6136

```

PR #6157, opened 2022-04-03
Added comparison-with-itself message example

matusvalo:
## Type of Changes

## Description

Related #5953

### DIFF
--- a/doc/data/messages/c/comparison-with-itself/bad.py
+++ b/doc/data/messages/c/comparison-with-itself/bad.py
@@ -1,3 +1,3 @@
-value = 5
-if value == value:  # [comparison-with-itself]
-    pass
+def is_an_orange(fruit):
+    an_orange = "orange"
+    return fruit == fruit  # [comparison-with-itself]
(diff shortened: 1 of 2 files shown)


doc/data/messages/c/comparison-with-itself/bad.py
```diff
@@ -0,0 +1,3 @@
+value = 5
+if value == value:  # [comparison-with-itself]
+    pass
```
Pierre-Sassoulas, line 3:
```suggestion
def is_an_orange(value):
    an_orange = "orange"
    return value == value  # [comparison-with-itself]
```

doc/data/messages/c/comparison-with-itself/good.py
```diff
@@ -0,0 +1,3 @@
+value = 5
+if value:
+    pass
```
Pierre-Sassoulas, line 3:
```suggestion
def is_an_orange(value):
    an_orange = "orange"
    return an_orange == value
```

Pierre-Sassoulas (commented):
I think the check where the problem is obliviously a typo/a mistake are hard to document minimally because you have to have sufficient context to convey that there was an error. I think it was hard to understand why ``if value`` was better than ``value == value`` when value is always equal to 5.

Pierre-Sassoulas (approved):

PR #6159, opened 2022-04-03
Added literal-comparison message example

matusvalo:

## Type of Changes

## Description

Related #5953

### DIFF
--- a/doc/data/messages/l/literal-comparison/bad.py
+++ b/doc/data/messages/l/literal-comparison/bad.py
@@ -1,3 +1,2 @@
-value = 25
-if value is 25:  # [literal-comparison]
-    pass
+def is_an_orange(fruit):
+    return fruit is "orange" # [literal-comparison]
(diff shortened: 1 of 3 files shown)


doc/data/messages/l/literal-comparison/bad.py
```diff
@@ -0,0 +1,3 @@
+value = 25
+if value is 25:  # [literal-comparison]
+    pass
```
Pierre-Sassoulas, line 3:
```suggestion
def is_an_orange(fruit):
    return fruit is "orange" # [literal-comparison]
```

doc/data/messages/l/literal-comparison/good.py
```diff
@@ -0,0 +1,3 @@
+value = 25
+if value == 25:
+    pass
```
Pierre-Sassoulas, line 3:
```suggestion
def is_an_orange(fruit):
    return fruit == "orange"
```

Pierre-Sassoulas (changes requested):
I think a function is better sometime so we have an "unknown" value and meaningful code.

Pierre-Sassoulas:
We should probably link to : https://docs.python.org/3/library/stdtypes.html#comparisons

matusvalo:
> We should probably link to : https://docs.python.org/3/library/stdtypes.html#comparisons

done

Pierre-Sassoulas (approved):

Issue #5838, opened 2022-02-25
False negative `unused-variable` on reassignment depending on old value


He3lixxx:
### Bug description

When reassigning a variable with a value that depends on its old value, pylint will not issue an `unused-variable` / W0612.

```python3
#! /usr/bin/env python3
# pylint: disable=missing-docstring

def some_func():
    var = 3
    var = max(1, var)
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

### Expected behavior

```
************* Module test
test.py:6:4: W0612: Unused variable 'var' (unused-variable)

-------------------------------------------------------------------
Your code has been rated at 6.67/10 (previous run: 10.00/10, -3.33)
```

Note that this happens when assigning some value independent from `var`, eg. with `var = max(1, 3)`.

### Pylint version

```shell
pylint 2.13.0-dev0
astroid 2.9.3
Python 3.9.7 (default, Sep 10 2021, 14:59:43) 
[GCC 11.2.0]

(installed from main at 6622d101ab09a23a8529ce01aea285d636235a76)
```

### OS / Environment

Ubuntu 21.10

Pierre-Sassoulas:
Thank you for opening the issue after checking on the latest main branch :)

jacobtylerwalls:
Interesting case to think about. What definition would we rely on for "used", then? `var` was used in the course of getting the max of itself and 1. What about that is not "use"? 🤔

Pierre-Sassoulas:
Hmm, I marked as false negative too fast, I think @jacobtylerwalls is right.

He3lixxx:
I was thinking of "unused" as "the value you assign here is never used afterwards", inspired by how clang-tidy would [lint it as a dead store](https://clang.llvm.org/docs/analyzer/checkers.html#deadcode-deadstores) in c++ code.

I thought pylint would lint such cases in general, but just learned that it doesn't. In this case, this is not really a bug report. Would it make sense to open a feature request for this?

DanielNoord:
I think this could be a feature request for the `CodeStyle` checker. Although obviously impacting performance I think such intermediate assignment are often used for clarity and readability and I don't think we should warn against them outright.

He3lixxx:
[...]

PR #6162, opened 2022-04-03
Remove TODOs in `unused-variable` tests

jacobtylerwalls:
## Type of Changes

## Description
Per discussion on #4391, which we closed, I think we can stick to the one-warning-per-variable-per-scope contract, so these TODOs are out of date.

If we want to look into adding warnings, we can do it more broadly in #5838.

### DIFF
--- a/tests/functional/u/unused/unused_variable.py
+++ b/tests/functional/u/unused/unused_variable.py
@@ -66,7 +66,7 @@ def hello(arg):
         return True
     raise Exception
 
-# pylint: disable=redefined-outer-name, wrong-import-position,misplaced-future
+# pylint: disable=wrong-import-position,misplaced-future
 from __future__ import print_function
 PATH = OS = collections = deque = None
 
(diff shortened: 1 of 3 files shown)


tests/functional/u/unused/unused_variable.py
```diff
@@ -66,7 +66,7 @@ def hello(arg):
         return True
     raise Exception
 
-# pylint: disable=redefined-outer-name, wrong-import-position,misplaced-future
+# pylint: disable=wrong-import-position,misplaced-future
```
jacobtylerwalls, line 69: I thought it was worth surfacing the `redefined-outer-name` in these tests so you can more easily see that _some_ warning is raised for the fishiness.

Pierre-Sassoulas (approved):
Nice cleanup !

jacobtylerwalls:
I'm looking into the behavior change -- see pylint run. Whether I have to restore the special case or not I'll add a functional test.

jacobtylerwalls:
Hmm. Whatever it is belongs in another PR. Haven't narrowed it down yet. I'll restore the code I thought was useless, and then this can keep moving.

Pierre-Sassoulas (approved):

Issue #4795, opened 2021-08-03
Discussion around a proper implementation of control-flow in pylint


Pierre-Sassoulas:
### Current problem

The main limitation regarding control flow is the fact that the variable checkers rely on ``NamesConsumer``'s to_consume/consumed states, and these are updated "all-or-nothing" per variable, with additional special-case checks here and there. To give a simpler example, pylint doesn't report any error on:

```python
def function():
    x = 10
    del x
    print(x)

function()
```

### Desired solution

I don't know the history of the Variables checker, but it seems to me like pylint will keep running into issues around control flow as long as this basic approach is used. I wouldn't say that this is a limitation of the AST visitor pattern though. I think improvements could be made by doing any of (in order of increasing code complexity and running time, but also correctness IMO):

    - Modifying ``NamesConsumer`` so that variables can be processed more than once.
    - Using astroid's variable lookup functionality, which handles control flow better. [1]
    - Building a control flow graph and then running ``VariablesChecker`` checks on the graph rather than the AST itself.

But all of these would be really significant changes so I'm not exactly advocating for them, just thought I'd mention. Personally I'm still very happy using pylint with the current functionality!

### Additional context

This issue is taken entirely from a comment by @david-yz-liu [here](https://github.com/PyCQA/pylint/pull/4791#issuecomment-892076763) and will serve as the reference for discussion regarding control-flow implementation.

jacobtylerwalls:
Just noting for anyone following the discussion I've proposed baby steps in #5402 (and eventually #5384) following approach 1, above:

> Modifying NamesConsumer so that variables can be processed more than once.

PR #6165, opened 2022-04-03
Add test coverage for sibling except handlers

jacobtylerwalls:
## Type of Changes

## Description

This adds coverage in case this block is removed without a replacement. (Marking as draft for now: I don't understand why it's needed, so I might try to replace it.)

https://github.com/PyCQA/pylint/blob/a93602d92edbf11885645866216d194324578acf/pylint/checkers/variables.py#L2362-L2366

### DIFF
--- a/tests/functional/u/unused/unused_variable.py
+++ b/tests/functional/u/unused/unused_variable.py
@@ -179,3 +179,14 @@ def func5():
     x = []
     # Test case requires homonym between "for x" and "in x"
     assert [True for x in x]
+
+
+def sibling_except_handlers():
+    try:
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
I looked at this again: I think we could make things more consistent by removing the special case (see end of discussion on #4391), but that would make pylint a bit worse, since it's already kind of a problem that pylint is all-or-nothing per name in the variables checker (see #4795).

I figure just merging this test would suffice for now.

Pierre-Sassoulas (approved):
Let's merge this then :) !

PR #6168, opened 2022-04-04
New check: TypeVar name mismatch

### LINKED ISSUE #5224: Checker for `TypeVar` name and first argument

jpy-git:
## Type of Changes

## Description

Catches when typevar name doesn't match the assigned variable name.
Follows on from work in #6015.

Closes #5224

### DIFF
--- a/pylint/checkers/base/name_checker/checker.py
+++ b/pylint/checkers/base/name_checker/checker.py
@@ -164,6 +164,12 @@ class NameChecker(_BasicChecker):
             'Emitted when both the "covariant" and "contravariant" '
             'keyword arguments are set to "True" in a TypeVar.',
         ),
+        "C0131": (
+            'TypeVar name "%s" does not match assigned variable name "%s"',
+            "typevar-name-mismatch",
+            "Emitted when a TypeVar is assigned to a variable "
+            "that does not match its name argument.",
+        ),
         "W0111": (
             "Name %s will become a keyword in Python %s",
             "assign-to-new-keyword",
@@ -376,6 +382,7 @@ class NameChecker(_BasicChecker):
         "assign-to-new-keyword",
         "typevar-name-incorrect-variance",
         "typevar-double-variance",
+        "typevar-name-mismatch",
     )
     def visit_assignname(self, node: nodes.AssignName) -> None:
         """Check module level assigned names."""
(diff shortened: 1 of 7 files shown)


Pierre-Sassoulas (approved):
LGTM, the tests are on point.

pylint/checkers/base/name_checker/checker.py
```diff
@@ -164,6 +164,12 @@ class NameChecker(_BasicChecker):
             'Emitted when both the "covariant" and "contravariant" '
             'keyword arguments are set to "True" in a TypeVar.',
         ),
+        "C0131": (
```
DanielNoord, line 167: Shit, I didn't pay enough attention. We already emit W0130.

C0131 is fine though.

Could you perhaps renumber these? Giving `C0130` to `typevar-double-variance`? We haven't released this yet so we should be good to do so!

tests/functional/t/typevar_name_mismatch.py
```diff
@@ -0,0 +1,47 @@
...
+    contravariant=True
+)
+X_co = TypeVar("T_co")  # [typevar-name-mismatch,typevar-name-incorrect-variance]
+X_contra = TypeVar(  # [typevar-name-mismatch,typevar-name-incorrect-variance]
+    "T_contra",
+    covariant=True
+)
+
+# name can be specified as a keyword argument as well.
+T = TypeVar(name="T")
```
DanielNoord, line 37: Thanks for catching this. This was likely to generate many false positives.

DanielNoord (changes requested):

pylint/checkers/base/name_checker/checker.py
jpy-git: Ahh it's because W130 (`duplicate-values`) was added since I made the original PR

Will rename `typevar-double-variance` to `C0132` 👍

tests/functional/t/typevar_name_mismatch.py
jpy-git: Yeah looks like MS didn't consider these valid test cases for Pylance haha
image

pylint/checkers/base/name_checker/checker.py
jpy-git: I ended up changing `typevar-double-variance` to `C0131` and `typevar-name-mismatch` to `C0132`

DanielNoord: Thanks a lot!

PR #6171, opened 2022-04-04
Move preprocessing of arguments into ``config`` directory

DanielNoord:

## Type of Changes

## Description

Think this might be the final refactor of `Run` for now.

Cleans up the remaining bits and pieces of the options handling here.

### DIFF
--- a/pylint/lint/__init__.py
+++ b/pylint/lint/__init__.py
@@ -16,6 +16,7 @@
 """
 import sys
 
+from pylint.config.exceptions import ArgumentPreprocessingError
 from pylint.lint.parallel import check_parallel
 from pylint.lint.pylinter import PyLinter
 from pylint.lint.report_functions import (
@@ -24,12 +25,7 @@ from pylint.lint.report_functions import (
     report_total_messages_stats,
 )
 from pylint.lint.run import Run
-from pylint.lint.utils import (
-    ArgumentPreprocessingError,
-    _patch_sys_path,
-    fix_import_path,
-    preprocess_options,
-)
+from pylint.lint.utils import _patch_sys_path, fix_import_path
 
 __all__ = [
     "check_parallel",
@@ -41,7 +37,6 @@ __all__ = [
     "ArgumentPreprocessingError",
     "_patch_sys_path",
     "fix_import_path",
-    "preprocess_options",
 ]
 
 if __name__ == "__main__":
--- a/tests/lint/unittest_lint.py
+++ b/tests/lint/unittest_lint.py
@@ -14,7 +14,7 @@ from io import StringIO
 from os import chdir, getcwd
 from os.path import abspath, dirname, join, sep
 from shutil import rmtree
-from typing import Iterable, Iterator, List, Optional, Tuple
+from typing import Iterable, Iterator, List
 
 import platformdirs
 import pytest
@@ -29,7 +29,7 @@ from pylint.constants import (
     OLD_DEFAULT_PYLINT_HOME,
 )
 from pylint.exceptions import InvalidMessageError
-from pylint.lint import ArgumentPreprocessingError, PyLinter, Run, preprocess_options
+from pylint.lint import PyLinter, Run
 from pylint.message import Message
 from pylint.reporters import text
 from pylint.testutils import create_files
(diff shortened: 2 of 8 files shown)


pylint/lint/__init__.py
```diff
@@ -41,7 +37,6 @@
     "ArgumentPreprocessingError",
     "_patch_sys_path",
     "fix_import_path",
-    "preprocess_options",
 ]
```
DanielNoord, line 40: Or should I add it back? Why is this in the public API?

tests/lint/unittest_lint.py
```diff
@@ -706,38 +706,6 @@ def test_pylintrc_parentdir_no_package() -> None:
                     assert config.find_pylintrc() == expected
```
DanielNoord, line 708: I'm waiting on test coverage and will tests if needed, but these tests don't really work.

We need a `Run` object for `preprocess_options` now, which basically means all other integrated tests where the preprocessed arguments are tested directly do exactly the same (ie., set up a `Run` and parse args).
Felt unnecessary to keep this in, but will add tests if necessary to keep coverage.

DanielNoord: Added two tests that should cover the initial coverage drop :)

pylint/lint/__init__.py
Pierre-Sassoulas: It's because ``pylint/lint.py`` was a gigantic file that we burst into a module so everything that was in it originally became an API in order to not make any assumption and break existing code.

DanielNoord: Do you want me to add it back? I did change the signature... So we would need to do something with `DeprecationWarnings`.

Pierre-Sassoulas: It wasn't clear from the above message but I think we can start to make some assumptions in order to not have to maintain useless API forever.

Pierre-Sassoulas: > we would need to do something with DeprecationWarnings.

Yeah no, I think it's something that is really pylint internal. I don't think it's used. It could but I don't see how.

Pierre-Sassoulas:
Look like format_option_help is not used at all or not covered during tests : https://coveralls.io/builds/47963363/source?filename=pylint%2Fconfig%2Foption_parser.py

Also we don't test invalid choices while option parsing: https://coveralls.io/builds/47963363/source?filename=pylint%2Fconfig%2Foption.py

Nor some cases in the OptionManagerMixin the "append" option in particular https://coveralls.io/builds/47963363/source?filename=pylint%2Fconfig%2Foptions_provider_mixin.py 

The ``pylint/config/__init__.py`` is not covered either but I have a plan to create function and to add tests, I just do not have the time right now, please don't move anything here :)

DanielNoord:
> Look like format_option_help is not used at all or not covered during tests : https://coveralls.io/builds/47963363/source?filename=pylint%2Fconfig%2Foption_parser.py

This is `optparse`.

> Also we don't test invalid choices while option parsing: https://coveralls.io/builds/47963363/source?filename=pylint%2Fconfig%2Foption.py

This is `optparse`.

> Nor some cases in the OptionManagerMixin the "append" option in particular https://coveralls.io/builds/47963363/source?filename=pylint%2Fconfig%2Foptions_provider_mixin.py
> 
> The `pylint/config/__init__.py` is not covered either but I have a plan to create function and to add tests, I just do not have the time right now, please don't move anything here :)

This is `optparse` as well.

Should I wait with this PR for your tests?

DanielNoord:
> We're still using optparse alongside argparse right now if I understand correctly ?
> 
> > Should I wait with this PR for your tests?
> 
> No, it's independent :)

Yeah, but I'm not sure if it's worth the effort to write tests for code we're going to deprecate. Especially since some of the code is getting removed as we switch to `argparse` partially (see these options for example).

Pierre-Sassoulas (approved):
Sure, I was just asking because it was unclear for me if it's dead code yet. Let's not add tests for code we're going to remove before releasing 2.14.

DanielNoord:
> Sure, I was just asking because it was unclear for me if it's dead code yet. Let's not add tests for code we're going to remove before releasing 2.14.

We're currently parsing both with `optparse` and `argparse`. However, after that blocked PR has landed we won't be using the `config` object created by `optparse` internally anymore. Still, we should deprecate that so we'll need to keep parsing with `optparse` to make sure that both `linter.config` and `linter.namespace` work.

That said, now that basic functionality is almost there I have been looking into whether I can make `argparse` parse to `linter.config` instead. I think there might actually be a chance that if my code is good enough and handles all the quirks of our previous system we can simply drop in replace it. That would reduce the number of deprecation warnings 😄

PR #6173, opened 2022-04-04
Deprecate ignore-mixin-members + add ignored-checks-for-mixins

### LINKED ISSUE #5205: Add new option ``ignore-checks-for-mixin`` and deprecate ``ignore-mixin-members``
### Current problem

What about removing `ignore-members` and adding a new option `ignore-checks-for-mixin` which would be a comma separated list of checks that exclude Mixin? We could add the supported checks to the description of that option.
I think I have already thought of good way to deprecate an option.

_Originally posted by @DanielNoord in https://github.com/PyCQA/pylint/pull/5203#discussion_r735300500_

The current ``ignore-mixin-members`` does not really convey what it is doing: disabling certain checks for mixin classes. Most notably `not-async-context-manager` refers to the option, which is not really clear from the option name.

### Desired solution

We would like to deprecate the option and create a new option that contains a list of options that should be disabled for mixin classes.

However, deprecation of options is quite difficult with our current set-up and since we need to move to `argparse` anyway it might be better to wait for that change to be done first. This is thus blocked by our move to `argparse`.

### Additional context

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #5205.

Finally, the first real new option that comes from this migration. With the use of `new_names` and `old_names` we can now deprecate options and warn about them accordingly.

I have applied it to two options already.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -228,6 +228,7 @@ class PyLinter(
                     "type": "csv",
                     "metavar": "<file>[,<file>...]",
                     "dest": "black_list",
+                    "kwargs": {"old_names": ["black_list"]},
                     "default": ("CVS",),
                     "help": "Files or directories to be skipped. "
                     "They should be base names, not paths.",
--- a/ChangeLog
+++ b/ChangeLog
@@ -42,6 +42,11 @@ Release date: TBA
 
   Closes #578
 
+* The ``ignore-mixin-members`` option has been deprecated. You should now use the new
+  ``ignored-checks-for-mixins`` option.
+
+  Closes #5205
+
 * Added new checker ``unnecessary-list-index-lookup`` for indexing into a list while
   iterating over ``enumerate()``.
 
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -257,10 +257,9 @@ methods is doing nothing but raising NotImplementedError.
 5.2 How do I avoid "access to undefined member" messages in my mixin classes?
 -------------------------------------------------------------------------------
 
-To do so you have to set the ignore-mixin-members option to
-"yes" (this is the default value) and name your mixin class with
-a name which ends with "Mixin" or "mixin" (default) or change the
-default value by changing the mixin-class-rgx option.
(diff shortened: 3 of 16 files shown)


pylint/lint/pylinter.py
```diff
@@ -228,6 +228,7 @@ def make_options() -> Tuple[Tuple[str, OptionDict], ...]:
                     "type": "csv",
                     "metavar": "<file>[,<file>...]",
                     "dest": "black_list",
+                    "kwargs": {"old_names": ["black_list"]},
```
Pierre-Sassoulas, line 231: Does it means that the old parameter will be ignored ?

Pierre-Sassoulas (commented):
Looks good ! I feel that there is a lot of argparse related code being added, is this in order to keep the same behavior or close than optparse (i.e. could we simplify if we change the behavior for 3.0 ?) or is it the way you would do it from scratch ? I'm used to args parsing being relatively small and decoupled.

pylint/lint/pylinter.py
DanielNoord: No I have set it up so that both `dest` and `old_names` in `kwargs` will also get the value of the option. So both `namespace.ignore` as well as `namespace.black_list` will work.
I have added both `dest` and `old_names` here for test coverage as the order of the `ifs` makes it that this together with the `black_list_re` option tests all combinations 😄

DanielNoord:
> Looks good ! I feel that there is a lot of argparse related code being added, is this in order to keep the same behavior or close than optparse (i.e. could we simplify if we change the behavior for 3.0 ?) or is it the way you would do it from scratch ? I'm used to args parsing being relatively small and decoupled.

A bit of both. I just try to enable `argparse` and see what issues come up and what I need to implement. It's not pure whack-a-mole, but a little 😄 

That said, this functionality of deprecating options is not supported by `argparse` out of the box. They only allow a single `dest` for each option. So I came up with this myself. We had an open issue about it already and it seems like something useful for an old project like ours.

Pierre-Sassoulas (approved):
The pypy 3.7 job is not working for some time please feel free to rebase if only this one fail. I've been trying to fix it with no success so far.

PR #6182, opened 2022-04-04
Rename ``cache-max-size-none`` and check ``functools.cache``

### LINKED ISSUE #5670: Investigate use of ``lru_cache`` on methods

DanielNoord:

## Type of Changes

## Description

Closes #5670.


pylint/checkers/stdlib.py
```diff
@@ -565,6 +573,7 @@ def visit_boolop(self, node: nodes.BoolOp) -> None:
     def visit_functiondef(self, node: nodes.FunctionDef) -> None:
         if node.decorators and isinstance(node.parent, nodes.ClassDef):
             self._check_lru_cache_decorators(node.decorators)
+            self._check_cache_decorators(node.decorators)
```
Pierre-Sassoulas, line 576: The check message just above need to be completed too or we'll have trouble for  "functools-cache-decorating-method" when "lru-cache-decorating-method" is disabled.

Pierre-Sassoulas (changes requested):
Needs a documentation with good.py and bad.py too :)

jacobtylerwalls:
Haven't looked in detail yet, but why wouldn't this just be a case for the existing message for lru_cache(maxsize=None)? It seems a bit of a usability flaw to have to enable/disable both messages when they do/check the same thing.

DanielNoord:
> Haven't looked in detail yet, but why wouldn't this just be a case for the existing message for lru_cache(maxsize=None)? It seems a bit of a usability flaw to have to enable/disable both messages when they do/check the same thing.

The other message has `lru...` in it while the decorator doesn't. I can imagine people might wonder why we're complaining about `lru_cache` when they are using `cache`.

I'm fine with adding to the other message as well though

jacobtylerwalls:
Yeah. It's hard because we've already released it. I feel like it should be renamed to something like `cache-max-size-none`. But on the other hand we're addressing it really quickly in a patch release. I don't want to push too hard -- let's see what others think!

DanielNoord:
> Yeah. It's hard because we've already released it. I feel like it should be renamed to something like `cache-max-size-none`. But on the other hand we're addressing it really quickly in a patch release. I don't want to push too hard -- let's see what others think!

With `old-names` it is fairly trivial to rename it.

doc/data/messages/f/functools-cache-decorating-method/bad.py
```diff
@@ -0,0 +1,7 @@
+import functools
+
+
+class MyClassWithMethods:
+    @functools.cache  # [functools-cache-decorating-method]
+    def my_func(self, param):
+        return param + 1
```
Pierre-Sassoulas, line 7:
```suggestion
import functools


class Fibonnaci:
    @functools.cache  # [functools-cache-decorating-method]
    def fibonacci(self, n):
        if n in {0, 1}:
            return n
        return self.fibonacci(n - 1) + self.fibonacci(n - 2)
```

doc/data/messages/f/functools-cache-decorating-method/good.py
```diff
@@ -0,0 +1,3 @@
+class MyClassWithMethods:
+    def my_func(self, param):
+        return param + 1
```
Pierre-Sassoulas, line 3:
```suggestion
import functools


@functools.cache
def cached_fibonacci(n):
    if n in {0, 1}:
        return n
    return cached_fibonacci(n - 1) + cached_fibonacci(n - 2)

class Fibonnaci:
    def fibonacci(self, n):
        return cached_fibonacci(n)
```

Pierre-Sassoulas (commented):
Also agree that we should rename and merge the two message.

doc/data/messages/c/cache-max-size-none/good.py
```diff
@@ -0,0 +1,13 @@
+import functools
+
+
+@functools.cache
```
Pierre-Sassoulas, line 4: Is the default value for functools.cache None ? But then why do we raise on method call only ? And if we only raise on method call because 'self' prevents garbage collection then why is the message name about maxsize being None and not about garbage collection ?

DanielNoord: `functools.cache` is an alias for `lru_cache(maxsize=None`. See: https://docs.python.org/3/library/functools.html#functools.cache

Pierre-Sassoulas: Sorry for committing directly it was impossible to make a suggestion the file was not modified.

DanielNoord:
@jacobtylerwalls Gentle ping 😄

ChangeLog
```diff
@@ -74,6 +74,10 @@ Release date: TBA
 
   Closes #5936
 
+* * Added ``cache-max-size-none`` checker will now also check ``functools.cache``.
```
Pierre-Sassoulas, line 77:
```suggestion
* The ``cache-max-size-none`` checker will now also check ``functools.cache``.
```

Pierre-Sassoulas (approved):

doc/data/messages/c/cache-max-size-none/good.py
jacobtylerwalls: I also don't understand why this check is limited to methods.

jacobtylerwalls (changes requested):
Thanks for the ping! Following Pierre's earlier comment, I'm also surprised this doesn't emit on pure functions. Was there a discussion elsewhere that clarified the reasoning?

doc/data/messages/c/cache-max-size-none/good.py
DanielNoord: The original rationale for this checker was https://www.youtube.com/watch?v=sVjtp6tGo0g

It is mostly because people are probably unaware that the instances gets retained until the cache entry gets evicted. This might not be what they expect. Note that `flake8 bugbear`  has also recently added this https://github.com/PyCQA/flake8-bugbear/pull/218. Probably inspired by the same video.

jacobtylerwalls: Sorry for being laconic! :-) I understood the original motivation, but since we're iterating on this a bit I wanted to ask if others felt the name of the check is out of sync with its scope. I would be in favor of either enlarging the scope of the check or retitling the message `cache-maxsize-none-instance-method`, but maybe this ⛵ has sailed and it shouldn't be revisited now (which I'm comfortable with.)

In any case, I think we need a good `details.rst`, because this is an "are-you-really-sure-about-that?" type of message, not "this is a bad pattern". It's also a YMMV message like #6231, in that following the advice could lead to worse results, if your instance is lightweight and it's fine to cache it. (We're about to do that in astroid!) That's all fine and how pylint is supposed to work--always use good judgment--but I think somewhere (maybe in `details.rst` we need to explain there's nothing special about instance methods--we're simply trying to surface possible user confusion. What do you think?

jacobtylerwalls: (I can take a stab wording some brief documentation if you agree it's worth it.)

DanielNoord: Yeah I think the ship might have sailed. Although we could probably improve the message warning to be more descriptive. I must say I struggled with finding good examples for this problem as I'm used to short-lived CLI programs for Python. I think the value of this warning comes for programs such as `pygame` or an (eventual) `pylintd`. In those case unexpectedly keeping an instance in memory might not be desired.

> (I can take a stab wording some brief documentation if you agree it's worth it.)

This would (obviously) be much appreciated!

Pierre-Sassoulas: > Yeah I think the ship might have sailed

We did rename it in 2.13.5 and it was published, but it's still possible to rename: 'old_names' are very cheap, confusing name and misunderstanding are very costly for everyone.

jacobtylerwalls: > In those case unexpectedly keeping an instance in memory might not be desired.

Django apps definitely don't want database objects hanging around in memory. :D

jacobtylerwalls: > confusing name and misunderstanding are very costly

What about `self-cache-max-size-none`? As in, "did you realize you were self-caching (unbounded)?"

jacobtylerwalls: Or even `self-cache-unbounded`?

Pierre-Sassoulas: I think the basics of what the check should do is unclear for me. From what I understood there is two checks and we should raises ``cache-used-on-method`` (?) for cached methods, even when maxsize is defined, because even one cache entry can prevent the garbage collection of an instance later. But we should also raise ``unbounded-cache-size`` (?) (currently ``cache-max-size-none``) when max-size is ``None`` for both method and function, because then the cache can grow without limit ? Did I misunderstand something ?

jacobtylerwalls: yep, we have two possible things to check, and currently the checker is just checking the intersection. 

> we should raises cache-used-on-method (?) for cached methods, even when maxsize is defined, because even one cache entry can prevent the garbage collection of an instance later.

That strikes me as too opinionated. If I've set a maxsize explicitly, I've exercised a choice. Plus, why are we assuming the instances any more memory-intensive than the other arguments to these hypothetical functions (we shouldn't imagine primitives, we should imagine `self` and `other`: `other` is just as expensive as `self`!) Speed vs memory use is a very different tradeoff for CLI vs a web app. `pylint` is not in a position to judge. maybe `pylint-django` would want to add a check for that. or `pylint` could make it optional.

>  But we should also raise unbounded-cache-size (?) (currently cache-max-size-none) when max-size is None for both method and function, because then the cache can grow without limit ?

Yeah, so that was essentially what I was arguing for a few comments above with my wondering if the ⛵ sailed. The _least effort_ approach would be to leave as is and call it `self-cache-unbounded`. I'm still warm to expanding the checker to just become `unbounded-cache`.

jacobtylerwalls: One argument for doing just the intersection is that even `unbounded-cache` is a little opinionated. (But seems less so than expanding to all methods).

DanielNoord: I think we should stick with the intersection. What about `method-cache-max-size-none` instead of `self-cache-max-size-none`?

jacobtylerwalls: Great!

jacobtylerwalls (approved):

PR #6183, opened 2022-04-04
Add ``Options`` and ``OptionsDict`` to typing utils

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/pyreverse/main.py
+++ b/pylint/pyreverse/main.py
@@ -19,6 +19,7 @@ from pylint.pyreverse.utils import (
     check_if_graphviz_supports_format,
     insert_default_options,
 )
+from pylint.typing import Options
 
 DIRECTLY_SUPPORTED_FORMATS = (
     "dot",
@@ -29,7 +30,7 @@ DIRECTLY_SUPPORTED_FORMATS = (
     "html",
 )
 
-OPTIONS = (
+OPTIONS: Options = (
     (
         "filter-mode",
         dict(
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -15,6 +15,7 @@ from pylint.constants import _MSG_ORDER, WarningScope
 from pylint.exceptions import InvalidMessageError
 from pylint.interfaces import Confidence, IRawChecker, ITokenChecker, implements
 from pylint.message.message_definition import MessageDefinition
+from pylint.typing import Options
 from pylint.utils import get_rst_section, get_rst_title
 
 if sys.version_info >= (3, 8):
@@ -31,7 +32,7 @@ class BaseChecker(OptionsProviderMixIn):
     # options level (0 will be displaying in --help, 1 in --long-help)
     level = 1
     # ordered list of options to control the checker behaviour
-    options: Any = ()
+    options: Options = ()
     # messages issued by this checker
     msgs: Any = {}
     # reports issued by this checker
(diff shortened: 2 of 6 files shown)


pylint/pyreverse/main.py
```diff
@@ -76,7 +77,7 @@
         "all-ancestors",
         dict(
             short="A",
-            default=None,
+            default="",
```
DanielNoord, line 80: @DudeNr33 Could you tell me if these default values make sense? Tests do pass but I'm not sure how extensive our `pyreverse` tests are..

Pierre-Sassoulas (approved):
Nice cleanup ! (I don't know about the default value for pyreverse so I'll let Andreas review this part)

pylint/pyreverse/main.py
DanielNoord: @DudeNr33 Gentle ping as this is blocking some of my other PRs for `argparse`. Never mind if you just didn't have the time but my recent work has probably flooded most of our inboxes and this might have gone unnoticed 😄

DudeNr33: Just saw it like a minute before you pinged me! 😁 I probably won't have time to look at it thoroughly today, but will try to tomorrow!

pylint/pyreverse/main.py
```diff
@@ -111,7 +112,7 @@
         "module-names",
         dict(
             short="m",
-            default=None,
+            default="no",
```
DudeNr33, line 115: Most common default value I found across the code base for ``type="yn"`` options was using booleans directly, but looking at the the ``_yn_transformer`` this should be fine as well.

DanielNoord: That's actually more performant. Going to change to booleans!

pylint/pyreverse/main.py
DudeNr33: ``"all-ancestors"`` and ``"all-associated"`` are used as simple flags. 
Looking at ``OptionsProviderMixin.load_defaults()`` we should either leave this at ``None`` or remove the ``default`` key: ``load_defaults`` calls ``set_option`` with ``optdict.get("default")`` for the ``value`` parameter. 

We would then end up with ``""`` as value for this option, but the ``pyreverse`` explicitly checks for ``is None`` (in ``DiaDefGenerator._set_option``).

pylint/pyreverse/main.py
DudeNr33: Upon a closer look, we can't change that, it actually alters the behavior (see above: the ``pyreverse`` code explicitly checks for ``is None`` in some cases).

DudeNr33 (changes requested):
I fear that we can not change the default values here without also changing the logic in ``pyreverse``.

We have a few tests which use ``pylint.testutils.PyreverseConfig``. But this class represents already parsed options, so we would have to update the values there based on how the defaults would be parsed.
Applying any of the changed defaults here leads to failed tests. 😞

DanielNoord:
> I fear that we can not change the default values here without also changing the logic in ``pyreverse``.
> 
> 
> 
> We have a few tests which use ``pylint.testutils.PyreverseConfig``. But this class represents already parsed options, so we would have to update the values there based on how the defaults would be parsed.
> 
> Applying any of the changed defaults here leads to failed tests. 😞 

Hmm... I'll do a refactor to fix this then. Default can't be None in argparse as it needs to have a type that is similar to the other values it can take.

Thanks for spotting this. I'll ping you on the next refactor!

DanielNoord:
Since @Pierre-Sassoulas already approved the refactor of the typing into `Options` and I have moved out the changes to default values within `pyreverse` I'm going to merge this :)

PR #6188, opened 2022-04-05
Make ``linter`` non-optional for ``BaseChecker``

DanielNoord:

## Type of Changes

## Description

**I'd like input on this.**

For those unaware: we're trying to move the configuration parsing of `pylint` from the deprecated `optparse` module to the newer `argparse` module. This involves major refactoring of everything related to option handling and the current change is prompted by this.

The benefit of making `linter` non-optional is that I think I can make it so that after all the work has been completed I can replace `self.linter.namespace` with `self.config`. That would mean that we can do the move to `argparse` without bumping to `3.0` as all plugins can still access `self.config` to get the config object like they can right now.
_I do think I would make a deprecation warning to say they should access `self.linter.config` but that's a discussion for another time._

However, we really do need each `BaseChecker` to have a `linter` as without it this won't be possible.
Obviously we should normally deprecate and warn about a significant function signature change like this, but I think there are two main reasons why the proposed change might still be feasible.
The first is that a checker without a `linter` can't do much. The main function of a checker `add_message` calls `self.linter.add_message` and thus needs a `linter`.
Secondly, the docstring of `BaseChecker` explicitly says that `BaseChecker` `should have the linter as argument`.

I have not written many (zero... 😅) plugins, but I think this change might be possible without breaking too much downstream plugins. I personally don't know if there is any benefit in writing a plugin that uses `BaseChecker` without using the `add_message` method but I can't immediately come up with on.

Just gently pinging some people because I'd like full consensus on this change. If you just want to express your opinion without writing a full comment please use a emoji on this opening post 😄 

_Blocked by the discussion._


pylint/checkers/base_checker.py
```diff
@@ -40,13 +43,9 @@ class BaseChecker(OptionsProviderMixIn):
     enabled: bool = True
 
     def __init__(
-        self, linter=None, *, future_option_parsing: Literal[None, True] = None
+        self, linter: "PyLinter", *, future_option_parsing: Literal[None, True] = None
     ):
-        """Checker instances should have the linter as argument.
-
-        :param ILinter linter: is an object implementing ILinter.
-        :raises MissingArgumentManager: If no linter object is passed.
```
Pierre-Sassoulas, line 48: See thing, apparently the intent was that a linter was necessary.

DanielNoord:
Almost a week has passed and all people who have responded to this issue have agreed with this change or brought additional arguments in favour of it.
Therefore, I'm going ahead and merge this. Thanks to all who took the time to review this!

PR #6187, opened 2022-04-05
Create ``_ArgumentsProvider``

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Blocked by #6183.

Turns out we do need `ArgumentsProvider` for our configuration stuff to work with `pyreverse`. I am planning to cut out some of the additional boilerplate though 😄

PR #6189, opened 2022-04-05
Finalize ``_parse_command_line_configuration``

DanielNoord:

## Type of Changes

## Description

Splitting up #6187.

### DIFF
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -5,7 +5,8 @@
 """Arguments manager class used to handle command-line arguments and options."""
 
 import argparse
-from typing import TYPE_CHECKING, Dict, List
+import sys
+from typing import TYPE_CHECKING, Dict, List, Optional
 
 from pylint.config.argument import (
     _Argument,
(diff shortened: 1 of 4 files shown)


DanielNoord:
Love it when behaviours is different across python versions 😅

pylint/config/arguments_manager.py
```diff
@@ -145,15 +146,17 @@ def _parse_configuration_file(self, config_data: Dict[str, str]) -> None:
...
 
-    def _parse_command_line_configuration(self, arguments: List[str]) -> None:
+    def _parse_command_line_configuration(
+        self, arguments: Optional[List[str]] = None
+    ) -> List[str]:
         """Parse the arguments found on the command line into the namespace."""
-        # pylint: disable-next=fixme
-        # TODO: This should parse_args instead of parse_known_args
-        self.namespace = self._arg_parser.parse_known_args(arguments, self.namespace)[0]
+        arguments = sys.argv[1:] if arguments is None else arguments
```
Pierre-Sassoulas, line 153: I think we should force this function to have arguments (``def _parse_command_line_configuration(self, arguments: List[str]) ->List[str]``). Imo the ``sys.argv[1:] if arguments is None else arguments`` should be done in only one place: for the default value of the API entrypoint and then we have list of strings with no default value allowed everywhere else. This way we make sure that we have a clearly defined API for our internal functions, no special case to handle for None value and we have only one place where we define the arguments.

pylint/config/deprecation_actions.py
```diff
@@ -33,7 +33,7 @@ def __init__(
         super().__init__(
             option_strings,
             dest,
-            "+",
+            1,
```
DanielNoord, line 36: https://docs.python.org/3/library/optparse.html#optparse.Option.nargs

Turns out `1` is the default in `optparse`. Should probably have looked that up when I made this 😅 

This should work 😄 _It did locally..._

Pierre-Sassoulas (changes requested):

pylint/config/arguments_manager.py
DanielNoord: https://github.com/PyCQA/pylint/blob/48e658551308daddf52fb62483356e39a3de097e/pylint/config/option_manager_mixin.py#L317-L332

Also does this, so in order not to break compatibility with it I thought I would copy the approach there...
I'm not sure what's best here. I have seen that in `pyreverse` we actually do this as well but before we call `load_command_line_configuration`. For `pylint` I think there are 3 places where we do this 😅 

Perhaps we could add a deprecation warning after everything else has been done?

pylint/config/arguments_manager.py
```diff
@@ -33,7 +34,6 @@ def __init__(self) -> None:
         self._arg_parser = argparse.ArgumentParser(
             prog="pylint",
             usage="%(prog)s [options]",
```
DanielNoord, line 36: Behaviour changed here after `3.7`. Couldn't see this in the docs but by comparing `argparse.py` I finally figured it out.

Since we register most checkers in `argparse` now there is no ambiguity anymore, or at least it is figured out by `argparse`, so this should be okay.

I have tested with both `3.7` and `3.10` locally.

pylint/config/arguments_manager.py
Pierre-Sassoulas: >  we could add a deprecation warning after everything else has been done?

Seems sensible.

Pierre-Sassoulas (approved):

PR #6191, opened 2022-04-05
Partially use ``namespace`` in ``PyLinter``

DanielNoord:

## Type of Changes

## Description

These are the easy options without complicated test that need updating.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -785,7 +785,7 @@ class PyLinter(
         Convert values in config.fail_on (which might be msg category, msg id,
         or symbol) to specific msgs, then enable and flag them for later.
         """
-        fail_on_vals = self.config.fail_on
+        fail_on_vals = self.namespace.fail_on
         if not fail_on_vals:
             return
 
@@ -1060,7 +1060,7 @@ class PyLinter(
                 DeprecationWarning,
             )
             files_or_modules = (files_or_modules,)  # type: ignore[assignment]
-        if self.config.recursive:
+        if self.namespace.recursive:
             files_or_modules = tuple(self._discover_files(files_or_modules))
         if self.config.from_stdin:
             if len(files_or_modules) != 1:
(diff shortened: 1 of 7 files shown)


pylint/lint/pylinter.py
```diff
@@ -1442,7 +1444,10 @@ def _get_message_state_scope(
         """Returns the scope at which a message was enabled/disabled."""
         if confidence is None:
             confidence = interfaces.UNDEFINED
-        if self.config.confidence and confidence.name not in self.config.confidence:
+        if (
+            self.namespace.confidence
+            and confidence.name not in self.namespace.confidence
+        ):
             return MSG_STATE_CONFIDENCE  # type: ignore[return-value] # mypy does not infer Literal correctly
```
Pierre-Sassoulas, line 1451: Look like we handling the case where ``self.config.confidence`` is falsey, but it should be None or a confidence, at this point, right ? Is that for mypy ?

tests/lint/unittest_lint.py
```diff
@@ -239,7 +240,7 @@ def test_enable_message_category(initialized_linter: PyLinter) -> None:
 
 
 def test_message_state_scope(initialized_linter: PyLinter) -> None:
-    class FakeConfig:
+    class FakeConfig(argparse.Namespace):
```
Pierre-Sassoulas, line 243: 👍

pylint/lint/pylinter.py
DanielNoord: That's interesting actually. I'll try and see if this could just be an `in` check.

DanielNoord:
Drafting while I figure out why this decreases coverage. Man this morning is rough...

DanielNoord:
@Pierre-Sassoulas PR became a bit different than original title but this doesn't decrease coverage :) Uncovered lines are already uncovered.

Pierre-Sassoulas (approved):
👍

Issue #6163, opened 2022-04-03
``bad-string-format-type``: only works on old-style string formatting and can't infer values


DudeNr33:
### Bug description

Two issues with ``bad-string-format-type``:

1. Just like ``bad-format-character`` the ``bad-string-format-type`` message is currently only raised for old-style string formatting.
2. If only a single value needs to be formatted, the check does not work if the value to format is passed in as a variable. Using variables in tuples however is fine.

Given a file ``a.py``:

```python
# pylint: disable=consider-using-f-string,missing-module-docstring
WORD = "abc"
print("%d" % "abc")  # works
print("%d %d" % (WORD, 1))  # works
print("%d", WORD)  # doesn't work
print("{:d}".format("abc"))  # doesn't work
print(f"{'abc':d}")  #  doesn't work
```

### Configuration

```ini
``pylintrc`` from ``pylint`` git repo.
```

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
.notes/a.py:3:6: E1307: Argument 'builtins.str' does not match format type 'd' (bad-string-format-type)
.notes/a.py:4:6: E1307: Argument 'builtins.str' does not match format type 'd' (bad-string-format-type)
```

### Expected behavior

All of the examples should raise the message.

### Pylint version

```shell
pylint 2.14.0-dev0
astroid 2.11.2
Python 3.10.2 (main, Apr  3 2022, 14:46:07) [Clang 12.0.5 (clang-1205.0.22.9)]
```

### OS / Environment

macOS BigSur 11.6

DudeNr33:
When this is resolved, the ``doc/data/messages/b/bad-string-format-type/details.rst`` must be updated.

Two hints so far:
* The message is not raised for new-style formatting and f-strings because the logic for the check is only contained in ``visit_binop`` of ``StringFormatChecker``
* When using a variable as format value (third example ``print("%d" % WORD)``, we end up in this ``else`` block: https://github.com/PyCQA/pylint/blob/30302f2555008a83db77aa0d61696bf5da4b3c51/pylint/checkers/strings.py#L360
It should be enough to add another ``elif`` just above:
```python
            elif isinstance(args, nodes.Name):
                args_elts = [args]
                num_args = 1
```

clavedeluna:
@DudeNr33 
> print("%d", WORD)  # doesn't work

is that third example which you also mention 

> third example print("%d" % WORD), we end up in this else block

supposed to be

[...]

PR #6194, opened 2022-04-05
Fix false negative for ``bad-string-format-type`` when using a variable as input

DudeNr33:

## Type of Changes

## Description

First and easy part of #6163.

After I started to work on this I realized that making ``bad-string-format-type`` and ``bad-format-character`` work on new-style formatting and f-strings will require a change to ``pylint.checkers.utils.parse_format_method_string``.
As this also leads to small adjustments in the callers of this function, I would rather do this in a separate PR, and then continue working on ``bas-string-format-type`` and ``bad-format-character``.

### DIFF
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -244,7 +244,6 @@ class StringFormatChecker(BaseChecker):
         "format-needs-mapping",
         "too-many-format-args",
         "too-few-format-args",
-        "bad-string-format-type",
         "format-string-without-interpolation",
     )
     def visit_binop(self, node: nodes.BinOp) -> None:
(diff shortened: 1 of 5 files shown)


pylint/checkers/strings.py
```diff
@@ -244,7 +244,6 @@ class StringFormatChecker(BaseChecker):
         "format-needs-mapping",
         "too-many-format-args",
         "too-few-format-args",
-        "bad-string-format-type",
```
DudeNr33, line 247: To make up for the times a message is forgotten in ``check_messages`` we had ``bad-string-format-type`` twice here. 😀

DanielNoord (approved):

PR #6195, opened 2022-04-05
Add examples of config file content to FAQ

bittner:
## Type of Changes

## Description

The documentation doesn't seem to explain what needs or can be added to Pylint configuration files. TOML support seems to be mentioned in the "What's New" section only.

This change updates the FAQ question that gives the impression that there was no TOML support, and gives actionable hints on how to write the configuration, effectively.

We may want to add those details to the [Messages Control](https://pylint.pycqa.org/en/latest/user_guide/message-control.html) or [Configuration](https://pylint.pycqa.org/en/latest/user_guide/options.html) section in the docs, one day. I have [seen in the code](https://github.com/PyCQA/pylint/blob/main/pylint/lint/pylinter.py#L547-L550) that there also seems to be a `[Reports]` section being evaluated from the config file, which also may warrant mentioning.


DanielNoord:
@bittner Thanks for taking the time to write documentation, that's often not the most fun task!

I'm marking this as blocked as we are actually in the process of completely overhauling the configuration parsing for the next release. A lot of things will become easier both for us and for end-users. For example, the use of sections in the `toml` file (expect for `tool.pylint` of course) will be unnecessary, although still supported.

I didn't know we had a section about this in the FAQ so it is good to know this, but I'd rather revisit this after that work has been done. That way we don't suggest stuff in the meantime that might change or become redundant in a couple of weeks.

Pierre-Sassoulas:
Thank you for your contribution! (especially on a pinned issue) Infortunately as Daniel said we're modifying this part of the code right now. The good news is that it should be easier to document after that.

DanielNoord:
@bittner I finished the transition and actually added some auto-generated documentation about our configuration options. You can see it in action here: https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html

If you want you can update this PR and perhaps link to that page as well? I have removed the `blocked` label from the PR 😄

Pierre-Sassoulas (approved):
LGTM, thank you @bittner !

doc/faq.rst
```diff
@@ -186,26 +187,53 @@ No, you can use symbolic names for messages::
...
-        method-hidden,
+    [Messages Control]
+    disable =
+        method-hidden
         too-many-lines
+        wildcard-import
+
+Alternatively, if you use ``pyproject.toml``, e.g.
+
+.. code-block:: toml
```
bittner, line 222: Looks like "toml" highlighting is not active for Sphinx, which results in a warning that [makes the build job fail](https://github.com/PyCQA/pylint/runs/6099963981?check_suite_focus=true). Pygments [does support TOML](https://pygments.org/languages/) in general, though, [since version 2.10.0](https://github.com/pygments/pygments/blob/master/CHANGES#L160-L189) (we have [2.11.2 installed](https://github.com/PyCQA/pylint/runs/6099963981?check_suite_focus=true#step:7:17)). What could be the problem? :thinking:

doc/faq.rst
```diff
@@ -186,26 +187,53 @@ No, you can use symbolic names for messages::
...
+    [tool.pylint.messages_control]
+    disable = [
+        "method-hidden",
+        "too-many-lines",
+        "wildcard-import",
+    ]
+
+    See also the `exhaustive list of possible options`_
 
+.. _`exhaustive list of possible options`: https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html
```
bittner, line 236: I'd better make this a local link. :worried:

Pierre-Sassoulas: It's a generated file I think, do you know how we could make a local link ?

doc/faq.rst
DanielNoord: I think this is because of the `See also`. It shouldn't be indented as that is not valid `toml` and thus can't be "lexed".

doc/faq.rst
DanielNoord: Since links are checked after extensions are loaded I think this should be able to directly link to the generated file.

bittner: @Pierre-Sassoulas I'll try to make it work [as described in the Sphinx docs](https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#automatically-label-sections).

bittner:
All ticks are green. The PR is ready for merging. :checkered_flag:

PR #6197, opened 2022-04-05
Use namespace in PyLinter and tests

DanielNoord:

## Type of Changes

## Description

Almost there...

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -839,6 +839,12 @@ class PyLinter(
 
         self.disable_noerror_messages()
         self.disable("miscellaneous")
+        self._arg_parser.parse_args(
+            ["--reports", "no", "--score", "no", "--persistent", "no"], self.namespace
+        )
+
+        # pylint: disable-next=fixme
+        # TODO: Potentially remove after 'set_option' has been refactored
         self.set_option("reports", False)
         self.set_option("persistent", False)
         self.set_option("score", False)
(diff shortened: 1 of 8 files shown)


pylint/lint/pylinter.py
```diff
@@ -839,6 +839,12 @@ def _parse_error_mode(self) -> None:
 
         self.disable_noerror_messages()
         self.disable("miscellaneous")
+        self._arg_parser.parse_args(
```
Pierre-Sassoulas, line 842: Hmm that's strange, why do we need that ? It's overriding actual values, right ?

DanielNoord: I haven't implemented a `set_option` that alters `namespace` yet. The previous one only does `config`. However, the current `set_option` is defined on `BaseChecker`. That's strange as it should be on `OptionsManager`.

I'm planning to implement `set_option` on `ArgumentsManager` but I need #6187 for it. I thought in the mean time this was good enough as the TODO makes sure I'll revisit and I'll probably do a search for `_arg_parser.parse_args` at the end anyway.
This keeps the ball rolling so to speak 😄

Pierre-Sassoulas: But to be clear it means users won't be able to change the score or report value themself ?

DanielNoord: No, that's what `errors-only` mode does. See the 3 lines just under the fix I added here. They do essentially the same as this `parse_args` only they affect `linter.config` and this changes `linter.namespace`.

Pierre-Sassoulas (approved):

PR #6199, opened 2022-04-05
Work towards a command to generate toml config file

### LINKED ISSUE #5462: Provide an auto-upgrade option / migration tool for pylint configurations

DanielNoord:

## Type of Changes

## Description

Far from done, but just to show what I have been working on. The test file won't be part of the final commit.

Could potentially close #5462. Just use `pylint --generate-toml-config`.


Pierre-Sassoulas (commented):
Looks great. I think we could do subcommand to separate normal pylint and utility. I think ``pylint autoupgrade`` would make a lot of sense. We load the current configuration and dump a up to date toml (without section, without "blacklist" or "master"). I think being able to add option to the command to change the output is a nice to have but would also make things more complicated for not much value added. This is the kind of command you launch once per project.

DanielNoord:
> Looks great. I think we could do subcommand to separate normal pylint and utility. I think `pylint autoupgrade` would make a lot of sense. We load the current configuration and dump a up to date toml 

This already happens. The configuration is loaded before we dump the `toml`. So when you don't have a `pylintrc` file you get a `.toml` file with the default values (as those get loaded into the `namespace` object), but if you do you can simply run this command to get a `.toml` file that represents your current `pylintrc` file.

We could probably document that a little better, but adding a second option only increases maintaining burden while it does essentially the same.

> (without section, without "blacklist" or "master"). 

`blacklist` is not in it currently as old names are excluded.

I chose to use section names as spitting out all options (120+) without any sections is probably a bit too much for new users? I think it is better if we support no-sections, but have the standard output be with-sections. In the end we won't get any bug reports about misspelled section names, but it looks a bit cleaner. 

> I think being able to add option to the command to change the output is a nice to have but would also make things more complicated for not much value added. This is the kind of command you launch once per project.

What extra options are you thinking of?

Pierre-Sassoulas:
> We could probably document that a little better, but adding a second option only increases maintaining burden while it does essentially the same.

No I understood what it was doing, but I think ``pylint --generate-toml-config`` would be drowned in the current long list of pylint's option. Separating makes it stand out in our documentation because we can document it as a standalone migration tool. Also it decreases the number of options in the main pylint. 

> blacklist is not in it currently as old names are excluded.

I thought they were already renamed in the result ?

> all options (120+) without any sections is probably a bit too much

Makes sense and the new ``--help`` with sections looks great btw. We'd have consistency if an option is used in multiple sections though.

> What extra options are you thinking of?

``pylint --disable=bare-except,invalid-name --class-rgx='[A-Z][a-z]+' --generate-toml-config`` if ``pylint autoupgrade --disable=bare-except,invalid-name --class-rgx='[A-Z][a-z]+'`` does not work and you can only do ``pylint autoupgrade`` it's not really a problem imo.

DanielNoord:
> No I understood what it was doing, but I think `pylint --generate-toml-config` would be drowned in the current long list of pylint's option. Separating makes it stand out in our documentation because we can document it as a standalone migration tool. Also it decreases the number of options in the main pylint.

I don't think I understand you 😅 If we have a `--generate-toml-config` and a `--autoupgrade` we have two instead of one options right?

> I thought they were already renamed in the result ?

Yeah, that's why I meant.

> Makes sense and the new `--help` with sections looks great btw. We'd have consistency if an option is used in multiple sections though.

🎉 

> `pylint --disable=bare-except,invalid-name --class-rgx='[A-Z][a-z]+' --generate-toml-config` if `pylint autoupgrade --disable=bare-except,invalid-name --class-rgx='[A-Z][a-z]+'` does not work and you can only do `pylint autoupgrade` it's not really a problem imo.

Ah I should a test then. This is supposed to work. As those options should get parsed first and then `--generate-toml-config` should be triggered. Will look into this!

Pierre-Sassoulas:
> I don't think I understand you 😅 If we have a --generate-toml-config and a --autoupgrade we have two instead of one options right?

I was thinking of a subcommand. So ``pylint autoupgrade`` not as an option, but as the equivalent of ``git add`` (pylint autoupgrade) vs ``git --help`` (pylint . --enable=all)

DanielNoord:
> I was thinking of a subcommand. So `pylint autoupgrade` not as an option, but as the equivalent of `git add` (pylint autoupgrade) vs `git --help` (pylint . --enable=all)

Ah I could look into that certainly. But probably better in a follow-up PR?

pylint/config/arguments_manager.py
```diff
@@ -43,6 +47,7 @@
 else:
     from typing import OrderedDict
 
+
```
DanielNoord, line 50:
```suggestion
```

DanielNoord:
I have removed the closing of the issue from this PR. Let's use this PR to create the general flag and use the issue to create a sub-parser or sub-command that uses this flag and some other parsing to create an auto-update function.

Something like `pylint config-update --old=pylintrc --new-format=toml`

doc/tutorial.rst
```diff
@@ -273,7 +273,8 @@ example but go ahead and `read up`_ on them if you want.
  It would really be a pain to specify that regex on the command line all the time, particularly if we're using many other options.
  That's what a configuration file is for. We can configure our Pylint to
  store our options for us so we don't have to declare them on the command line.  Using a configuration file is a nice way of formalizing your rules and
- quickly sharing them with others. Invoking ``pylint --generate-rcfile`` will create a sample rcfile with all the options set and explained in comments.
+ quickly sharing them with others. Invoking ``pylint --generate-toml-config`` will create a sample ``.toml`` section with all the options set and explained in comments.
+ This can then be added to your ``pyproject.toml`` file.
```
Pierre-Sassoulas, line 277: (or any toml configuration file that you're linking with the ``--rc-file`` option)

pylint/lint/run.py
```diff
@@ -243,6 +270,7 @@ def __init__(
                         "callback": Run._not_implemented_callback,
                         "help": "Load and enable all available extensions. "
                         "Use --list-extensions to see a list all available extensions.",
+                        "hide_from_config_file": True,
```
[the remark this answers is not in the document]
DanielNoord: I'm not sure if we actually support this. I'll open a follow-up issue to investigate.

Either way, I don't think we should add it in the default output of this flag as it can only confuse users.

doc/tutorial.rst
DanielNoord: Want me to add that? I think following the (slow) trend towards unification in configuration files is good and a contribution we can make as project to the general python community.

pylint/lint/run.py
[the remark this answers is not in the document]
Pierre-Sassoulas: What I was sawing is we should probably hide rc-file from the config file too ? Or did I misunderstand the purpose of "hide_from_config_file" ?

pylint/lint/run.py
```diff
@@ -104,6 +105,7 @@ def __init__(
                         "callback": Run._not_implemented_callback,
                         "group": "Commands",
                         "help": "Specify a configuration file to load.",
+                        "hide_from_config_file": True,
```
DanielNoord, line 108: @Pierre-Sassoulas It is already excluded!

doc/tutorial.rst
Pierre-Sassoulas: I think it's better to use a toml file that can be copy paster in pyproject.toml in the future than to stick to an ini pylintrc so if someone is recluctant to use pyproject.toml right now he or she can still use a toml with another name and we won't have to maintain the tool to generate ini file.

DanielNoord: Added a new line!

tests/test_self.py
```diff
@@ -34,6 +34,12 @@
 from pylint.reporters.text import BaseReporter, ColorizedTextReporter, TextReporter
 from pylint.utils import utils
 
+if sys.version_info >= (3, 11):
```
Pierre-Sassoulas, line 37: Should we create a PY311_PLUS constant ?

Pierre-Sassoulas (approved):

tests/test_self.py
DanielNoord: I believe we don't use those on import guards because `mypy` doesn't like them.

Pierre-Sassoulas: Right, I forgot

DanielNoord:
Blocked by #6290.

PR #6200, opened 2022-04-06
add example for `function-redefined` message documentation

gunungpw:
## Type of Changes

## Description
- add code example for `function-redefined`

Related #5953

### DIFF
--- a/doc/data/messages/f/function-redefined/good.py
+++ b/doc/data/messages/f/function-redefined/good.py
@@ -1,0 +1,6 @@
+def get_email():
+    pass
+
+
+def get_address():
+    pass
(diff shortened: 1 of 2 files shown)


doc/data/messages/f/function-redefined/good.py
```diff
@@ -0,0 +1,6 @@
+def get_email():
+    pass
+
+
+def get_address():
+    pass
```
Pierre-Sassoulas, line 6:
```suggestion
```
Imo it's often due to a copy paste so the solution is often to remove completely. It's also clearer to not add another function as we don't have any context in the example.

Pierre-Sassoulas (approved):

PR #6209, opened 2022-04-06
Update optdicts in ``pyreverse``

DanielNoord:

## Type of Changes

## Description

After @DudeNr33's review of #6183 I took another look at if we can handle `None` as `default` and it should be possible. However, these optdicts were still missing some required keys. This PR adds these in preparation for #6183 and an eventual move of `pyreverse` to `argparse`.


pylint/pyreverse/main.py
```diff
@@ -77,6 +78,8 @@
         dict(
             short="A",
             default=None,
+            type="yn",
```
DudeNr33, line 81: Sorry if I maybe was not clear enough on the other review.
``A`` and ``S`` are flags, that are used like ``pyreverse -AS <mypackage>``. Changing the type to "yn" is incompatible with the current usage.

pylint/pyreverse/main.py
```diff
@@ -95,6 +99,8 @@
         dict(
             short="S",
             default=None,
+            type="yn",
```
DudeNr33, line 102: Same as for ``A``.

pylint/pyreverse/main.py
```diff
@@ -146,6 +155,7 @@
             dest="colorized",
             action="store_true",
             default=False,
+            metavar="<y or n>",
```
DudeNr33, line 158: This is also a flag and just used as ``--colorized``, so the metavar does not really fit here.

pylint/pyreverse/main.py
```diff
@@ -123,6 +130,7 @@
             short="k",
             action="store_true",
             default=False,
+            metavar="<y or no>",
```
DudeNr33, line 133: Also a flag.

pylint/pyreverse/main.py
```diff
@@ -104,6 +110,7 @@
             short="b",
             action="store_true",
             default=False,
+            metavar="<y or n>",
```
DudeNr33, line 113: Also a flag.

DudeNr33 (changes requested):
I fear that most of the changes are incompatible with the current usage of ``pyreverse``.
There are a lot of options that are just flags and do not require additional input. 
I haven't checked yet how these are handled in other checkers, for example.

pylint/pyreverse/main.py
DanielNoord: Should they be changed to action `store_true` then?

DudeNr33: This would definitely make sense, but we have to change the code as well. If you look at the help message for the ``class`` option you can see that the ``class`` option has a coupling with ``A``, ``S`` and ``m`` and the default value ``None`` (see ``DiaDefGenerator._set_default_options`` and the calls to ``_set_option``).

DanielNoord: Yeah I saw that, but I think `store_true` should work. As long as `default` is not a string the `type` (this time the type of `store_true`) won't be called so they would still be `None`.

DudeNr33 (changes requested):
Sorry for the late reply, I was busy the last few days.

I created a short test to check if the default values remain the same after the change (feel free to copy it and commit it in this PR if you want).

The new action ``store_true`` seems to be problematic. Funnily enough, it not only ignores the specified ``default``, but it also sets the value not to ``True`` or ``False`` but to ``0``:

```shell
FAILED tests/pyreverse/test_main.py::test_command_line_arguments_defaults[all_ancestors-None] - AssertionError: assert 0 == None
FAILED tests/pyreverse/test_main.py::test_command_line_arguments_defaults[all_associated-None] - AssertionError: assert 0 == None
FAILED tests/pyreverse/test_main.py::test_command_line_arguments_defaults[module_names-None] - AssertionError: assert 0 == None
```
Unfortunately, those are exactly those three options that must default to ``None``.

Here is the test case (I would put it into ``tests/pyreverse/test_main.py``:

```python
@pytest.mark.parametrize(
    ("arg", "expected_default"), 
    [
        ("mode", "PUB_ONLY"), 
        ("classes", []), 
        ("show_ancestors", None), 
        ("all_ancestors", None),
        ("show_associated", None),
        ("all_associated", None),
        ("show_builtin", 0),
        ("module_names", None),
        ("output_format", "dot"),
        ("colorized", 0),
        ("max_color_depth", 2),
        ("ignore_list", ("CVS",)),
        ("project", ""),
        ("output_directory", ""),
    ]
)
@mock.patch("pylint.pyreverse.main.Run.run", new=mock.MagicMock())
@mock.patch("pylint.pyreverse.main.sys.exit", new=mock.MagicMock())
def test_command_line_arguments_defaults(arg, expected_default):
    run = main.Run([TEST_DATA_DIR])
    assert getattr(run.config, arg) == expected_default
```

DanielNoord:
@DudeNr33 Turns out there was actually a bug in our `optparse` implementation. `optparse` was handling `None` correctly, but we were setting it to `0` or `1` ourselves. Let's see if this fix breaks anything else.

Thanks for writing the test that should also help with porting `pyreverse` to `argparse`.

pylint/pyreverse/main.py
```diff
@@ -112,8 +117,7 @@
         dict(
             short="m",
             default=None,
-            type="yn",
-            metavar="<y or n>",
+            action="store_true",
```
Pierre-Sassoulas, line 120: Ain't it a breaking change ?

pylint/pyreverse/main.py
```diff
@@ -77,6 +79,7 @@
         dict(
             short="A",
             default=None,
```
Pierre-Sassoulas, line 81: If this a boolean if could be 
```suggestion
            default=False,
```
, right ?

tests/pyreverse/test_main.py
```diff
@@ -118,3 +118,30 @@ def test_graphviz_unsupported_image_format(capsys):
     assert "Format somethingElse is not supported by Graphviz. It supports:" in stdout
     # Check that we exited with the expected error code
     assert wrapped_sysexit.value.code == 32
+
+
+@pytest.mark.parametrize(
+    ("arg", "expected_default"),
```
Pierre-Sassoulas, line 124: 👌

pylint/pyreverse/main.py
DanielNoord: See previous discussion. We actually use the `None` value here to see if the option was explicitly set to `True` or `False`. If it was not set explicitly we determine the value based on another option's value.
It's a bit convoluted, but changing it would be a breaking change as default behaviour would change.

pylint/pyreverse/main.py
Pierre-Sassoulas: Imo it means we goes from having to do ``-m y`` or  ``-m=n``   to simply ``-m`` with a parsing error for the former.

DanielNoord: Ah that might be true. I'll investigate and add a test this afternoon.

Pierre-Sassoulas: But I do agree that not having y/n options is better. Not sure if asking user to upgrade their scripts is worth it even in 3.0 or another breaking change. Maybe allowing both behavior would be preferable, but ti's not going to make our maintenance burden lighter.

DanielNoord: Added a regression test!

Pierre-Sassoulas (approved):

DudeNr33 (approved):
👍 Looks good to me now!

PR #6210, opened 2022-04-06
Another round of small refactors in ``argparse``

DanielNoord:

## Type of Changes

## Description

Extracting some of the changes in #6199 as well as fixing one easy `TODO`.

### DIFF
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -47,11 +47,7 @@ class _ArgumentsManager:
             argument = _convert_option_to_argument(opt, optdict)
             section = argument.section or provider.name.capitalize()
 
-            # pylint: disable-next=fixme
-            # TODO: Optparse: Always define option_groups_descs on ArgumentsProvider
-            section_desc = None
-            if hasattr(provider, "option_groups_descs"):
-                section_desc = provider.option_groups_descs.get(section, None)  # type: ignore[attr-defined]
+            section_desc = provider.option_groups_descs.get(section, None)
             self._add_arguments_to_parser(section, section_desc, argument)
 
         self._load_default_argument_values()
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -19,6 +19,7 @@ from pylint.config.argument import (
 from pylint.config.exceptions import UnrecognizedArgumentAction
 from pylint.config.help_formatter import _HelpFormatter
 from pylint.config.utils import _convert_option_to_argument
+from pylint.typing import OptionDict
 
 if TYPE_CHECKING:
     from pylint.config.arguments_provider import _ArgumentsProvider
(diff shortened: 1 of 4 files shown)


pylint/config/arguments_manager.py
```diff
@@ -41,17 +42,17 @@ def __init__(self, prog: str, usage: Optional[str] = None) -> None:
...
+            self._option_dicts[opt] = optdict
             argument = _convert_option_to_argument(opt, optdict)
             section = argument.section or provider.name.capitalize()
 
-            # pylint: disable-next=fixme
-            # TODO: Optparse: Always define option_groups_descs on ArgumentsProvider
-            section_desc = None
-            if hasattr(provider, "option_groups_descs"):
-                section_desc = provider.option_groups_descs.get(section, None)  # type: ignore[attr-defined]
+            section_desc = provider.option_groups_descs.get(section, None)
```
Pierre-Sassoulas, line 55: get default value is always None for a dict. I should have said that there were other instances of this in the last MR but it's not that important.

tests/test_self.py
```diff
@@ -1363,6 +1344,29 @@ def test_generate_rcfile() -> None:
         )
         assert process.stdout == process_two.stdout
 
+    @staticmethod
```
Pierre-Sassoulas, line 1347: Why move this ? It could create problem if we need to rebase or merge change for it at some point later on.

Pierre-Sassoulas (changes requested):

pylint/config/arguments_manager.py
DanielNoord: Ah yeah, force of habit. Want me to remove it here?

tests/test_self.py
DanielNoord: Because it fits in the `TestCallbackOptions` class that it will now be part of. I should have moved it when I made tests for all callback options but I didn't want to change any of the pre-existing tests. In #6199 I do move it as I thought it would be better to have all tests for `--generate-rcfile` and `--generate-toml-config` in one place.
I moved it to this PR to reduce the diff there.

Pierre-Sassoulas: Make sense !

pylint/config/arguments_manager.py
Pierre-Sassoulas: Let's keep it that way, being explicit can be considered better too and we have other more pressing issues 😄

Pierre-Sassoulas (approved):

tests/test_self.py
Pierre-Sassoulas: I just thought : maybe bursting test_self would be a good thing at this point. The file is humongous.

DanielNoord: Makes sense! We could move the callback tests to the config folder.

PR #6212, opened 2022-04-07
Handle asterisks better in Sphinx and Google style docstrings

### LINKED ISSUE #5406: [v2.12 Regression] Asterisk argument `mising-param-doc` vs `anomalous-backslash-in-string`

### LINKED ISSUE #5815: Escaped '*' in docstring leads to missing-param-doc
### Bug description

If a docstring contains the arguments for `*args` and `**kwargs` and RST is used, `*` must be escaped.

```python
def myfunc(value: int, *args: Any) -> None:
    """This is myfunc.

    Args:
        \\*args: this is args
        value: this is value
    """
    print(*args, value)
```

However, the pylint docparams check doesn't like the escape

### Configuration

_No response_

### Command used

```shell
pylint --load-plugins pylint.extensions.docparams foo.py --enable all
```

### Pylint output

```shell
************* Module foo
foo.py:4:0: W9015: "*args" missing in parameter documentation (missing-param-doc)
```

### Expected behavior

Pylint should ignore escaping

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.7.1 (default, Oct 22 2018, 13:16:18) [GCC]
```

### OS / Environment

Linux 

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #5815, closes #5406.

See the tests for what is allowed now. Asterisks are no longer required and will be parsed correctly in Google style. I have chosen to fix this in the extension itself. Internally we still see the argument names including the asterisk, but when we finally check if we should emit the warning we also check whether the name without asterisks exists. If so, we don't emit.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -85,7 +85,11 @@ What's New in Pylint 2.13.6?
 ============================
 Release date: TBA
 
+* Asterisks are no longer required in Sphinx and Google style parameter documentation
+  for ``missing-param-doc`` and are parsed correctly.
 
+  Closes #5815
+  Closes #5406
 
 
 What's New in Pylint 2.13.5?
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -199,6 +199,12 @@ Other Changes
 
   Closes #5614
 
+* Asterisks are no longer required in Sphinx and Google style parameter documentation
+  for ``missing-param-doc`` and are parsed correctly.
+
+  Closes #5815
+  Closes #5406
+
 * Use the ``tomli`` package instead of ``toml`` to parse ``.toml`` files.
 
  Closes #5885
--- a/pylint/extensions/_check_docs_utils.py
+++ b/pylint/extensions/_check_docs_utils.py
@@ -269,7 +269,7 @@ class SphinxDocstring(Docstring):
         \s+
         )?
 
-        ((\\\*{{1,2}}\w+)|(\w+))  # Parameter name with potential asterisks
+        ((\\\*{{0,2}}\w+)|(\w+))  # Parameter name with potential asterisks
         \s*                       # whitespace
         :                         # final colon
         """
@@ -469,7 +469,7 @@ class GoogleDocstring(Docstring):
 
     re_param_line = re.compile(
         rf"""
-        \s*  (\*{{0,2}}\w+)             # identifier potentially with asterisks
+        \s*  ((?:\\?\*{{0,2}})?\w+)     # identifier potentially with asterisks
         \s*  ( [(]
             {re_multiple_type}
             (?:,\s+optional)?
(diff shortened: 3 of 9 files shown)


Pierre-Sassoulas (approved):
Great job adding/updating the typing at the same time !

DanielNoord:
I think this warrants one final patch release on `3.6` considering the amount of feedback we got about this 😅

Pierre-Sassoulas:
I'm waiting for #5822 to be merged to release the last patch. I burst between 2.13.5 and 2.13.6 to release sooner because there was a crash fix that was waiting for a long time already.

PR #6213, opened 2022-04-07
Assert that links in related.rst are a list

DanielNoord:

## Type of Changes

### DIFF
--- a/doc/test_messages_documentation.py
+++ b/doc/test_messages_documentation.py
@@ -33,7 +33,10 @@ MessageCounter = CounterType[Tuple[int, str]]
 
 
 def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str, Path]]:
-    """Get all functional tests in the input_dir."""
+    """Get all functional tests in the input_dir.
+
+    This also checks the formatting of related.rst files.
+    """
     suite: List[Tuple[str, Path]] = []
 
     for subdirectory in input_dir.iterdir():
@@ -46,6 +49,12 @@ def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str,
                 suite.append(
                     (message_dir.stem, message_dir / "bad.py"),
                 )
+            if (message_dir / "related.rst").exists():
+                with open(message_dir / "related.rst", encoding="utf-8") as file:
+                    text = file.read()
+                    assert text.startswith(
+                        "-"
+                    ), f"{message_dir / 'related.rst'} should be a list using '-'."
     return suite
 
 
--- a/doc/data/messages/c/cell-var-from-loop/related.rst
+++ b/doc/data/messages/c/cell-var-from-loop/related.rst
@@ -1,1 +1,1 @@
-* `stackoverflow discussion <https://stackoverflow.com/questions/25314547/cell-var-from-loop-warning-from-pylint>`_
+- `Stackoverflow discussion <https://stackoverflow.com/questions/25314547/cell-var-from-loop-warning-from-pylint>`_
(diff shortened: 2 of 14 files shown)


doc/test_messages_documentation.py
```diff
@@ -46,6 +49,12 @@ def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str,
                 suite.append(
                     (message_dir.stem, message_dir / "bad.py"),
                 )
+            if (message_dir / "related.rst").exists():
+                with open(message_dir / "related.rst", encoding="utf-8") as file:
+                    text = file.read()
+                    assert text.startswith(
+                        "-"
```
Pierre-Sassoulas, line 56:
```suggestion
                        "*"
```

doc/test_messages_documentation.py
```diff
@@ -46,6 +49,12 @@ def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str,
...
+                    ), f"{message_dir / 'related.rst'} should be a list using '-'."
```
Pierre-Sassoulas, line 57:
```suggestion
                    ), f"{message_dir / 'related.rst'} should be a list using '*'."
```

Pierre-Sassoulas (commented):
Are you sure that rst lists are done with ``-`` and not ``*`` ? https://gdal.org/contributing/rst_style.html#lists

DanielNoord:
> Are you sure that rst lists are done with `-` and not `*` ? https://gdal.org/contributing/rst_style.html#lists

See:
https://pylint--6213.org.readthedocs.build/en/6213/messages/warning/cell-var-from-loop.html

Seems to work? I tested there were more cases with `-` currently so this seemed easier 😄

Pierre-Sassoulas:
See for example the current state of cell-var-from-loop : https://pylint.pycqa.org/en/latest/messages/warning/cell-var-from-loop.html
It seems to also work using ``-`` https://pylint--6213.org.readthedocs.build/en/6213/messages/warning/cell-var-from-loop.html

DanielNoord:
Btw, I really like the new theme 😄 Think it was a good discussion to go ahead with it!

doc/test_messages_documentation.py
Pierre-Sassoulas, line 57:
```suggestion
                    assert text.startswith(                        "-"                    ) or text.startswith(                        "*"                    ), f"{message_dir / 'related.rst'} should be a list using '-'."
```

DanielNoord: Why not require consistency now? We're not too far in to change it now

Pierre-Sassoulas (changes requested):
I don't think we should force the use of one or the other if it's valid rst. (Especially as it's not automatically formatted).

DanielNoord:
> I don't think we should force the use of one or the other if it's valid rst. (Especially as it's not automatically formatted).

If we do force consistency we can more easily parse this info programmatically. I have been thinking of making a PR for `pylint-errors` to get the info they use form our documentation instead of their internal database to make sure the effort of writing these examples is not duplicated. Removing inconsistencies like these would help with doing so 😄

Pierre-Sassoulas:
> we do force consistency we can more easily parse this info programmatically.

Hmm right.

>  I have been thinking of making a PR for pylint-errors to get the info they use form our documentation instead of their internal database

I think vald-phoenix is currently busy "fighting for the bright future of 🇺🇦" so it make take some time to merge it but also isn't the base idea to integrate the work and then to not have to maintain pylint-error anymore ?

Pierre-Sassoulas (approved):
Let's normalize then.

DanielNoord:
> I think vald-phoenix is currently busy "fighting for the bright future of 🇺🇦" so it make take some time to merge it but also isn't the base idea to integrate the work and then to not have to maintain pylint-error anymore ?

We could also do that. However, they also have a CLI tool which we would need to integrate. Not sure if we have the bandwidth to maintain that.
Making `pylint-errors` a CLI tool that depends on our documentation as a `git submodule` might be a good compromise, but that's definitely something for "future us". Right now we're not even on par with their data yet.

DanielNoord:
> > Making pylint-errors a CLI tool that depends on our documentation as a git submodule might be a good compromise, but that's definitely something for "future us". Right now we're not even on par with their data yet.
> 
> I think we have a --msg-help option already, right ?

Ah of course. We could probably integrate the code examples there, although that would inflate the size of the `pylint` package. Something to think about...

PR #6214, opened 2022-04-07
Copy all of ``OptionsManagerMixIn`` methods into ``_ArgumentsManager``

DanielNoord:

## Type of Changes

## Description

Sorry for the size of this PR. Basically these are all the methods defined on `OptionsManagerMixIn`. Sadly they are all public so they will need to be deprecated.
I thought it would be better to move them on `_ArgumentsManager` and deprecate from there. This would allow us to deprecate instantiating `OptionsManagerMixIn` and completely disregard that file until `3.0` gets released.

### DIFF
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -392,7 +392,9 @@ class _ArgumentsManager:
         """
         if not config_file:
             if verbose:
-                print("No config file found, using default configuration", file=sys.stderr)
+                print(
+                    "No config file found, using default configuration", file=sys.stderr
+                )
             return
         config_file = Path(os.path.expandvars(config_file)).expanduser()
         if not config_file.exists():
@@ -415,7 +417,7 @@ class _ArgumentsManager:
                     parser._sections[sect.upper()] = values
 
         if verbose:
-             print(f"Using config file '{config_file}'", file=sys.stderr)     
+            print(f"Using config file '{config_file}'", file=sys.stderr)
 
     def _parse_toml(self, config_file: Path, parser: configparser.ConfigParser) -> None:
         """Parse and handle errors of a toml configuration file."""
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -390,35 +390,32 @@ class _ArgumentsManager:
         """Read the configuration file but do not load it (i.e. dispatching
         values to each option's provider)
         """
-        if config_file:
-            config_file = Path(os.path.expandvars(config_file)).expanduser()
-            if not config_file.exists():
-                raise OSError(f"The config file {str(config_file)} doesn't exist!")
-
-            parser = self.cfgfile_parser
-            if config_file.suffix == ".toml":
-                try:
-                    self._parse_toml(config_file, parser)
-                except tomllib.TOMLDecodeError:
-                    pass
-            else:
-                # Use this encoding in order to strip the BOM marker, if any.
-                with open(config_file, encoding="utf_8_sig") as fp:
-                    parser.read_file(fp)
-                # normalize each section's title
-                for sect, values in list(parser._sections.items()):
-                    if sect.startswith("pylint."):
-                        sect = sect[len("pylint.") :]
-                    if not sect.isupper() and values:
-                        parser._sections[sect.upper()] = values
-
-        if not verbose:
+        if not config_file:
+            if verbose:
+                print("No config file found, using default configuration", file=sys.stderr)
             return
-        if config_file and config_file.exists():
-            msg = f"Using config file '{config_file}'"
+        config_file = Path(os.path.expandvars(config_file)).expanduser()
+        if not config_file.exists():
+            raise OSError(f"The config file {str(config_file)} doesn't exist!")
+        parser = self.cfgfile_parser
+        if config_file.suffix == ".toml":
+            try:
+                self._parse_toml(config_file, parser)
+            except tomllib.TOMLDecodeError:
+                pass
         else:
-            msg = "No config file found, using default configuration"
-        print(msg, file=sys.stderr)
+            # Use this encoding in order to strip the BOM marker, if any.
+            with open(config_file, encoding="utf_8_sig") as fp:
+                parser.read_file(fp)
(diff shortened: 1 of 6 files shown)


tests/lint/unittest_lint.py
```diff
@@ -710,7 +710,7 @@ def test_pylintrc_parentdir_no_package() -> None:
                     assert config.find_pylintrc() == expected
 
 
-class _CustomPyLinter(PyLinter):  # pylint: disable=too-many-ancestors
+class _CustomPyLinter(PyLinter):
```
Pierre-Sassoulas, line 713: 🙏

DanielNoord:
The coverage is the same as for `OptionsManagerMixIn`. Will probably be fixed when I start working on deprecation warnings here.

pylint/config/arguments_manager.py
```diff
@@ -161,3 +225,274 @@ def _parse_command_line_configuration(
...
+                        sect = sect[len("pylint.") :]
+                    if not sect.isupper() and values:
+                        parser._sections[sect.upper()] = values
+
+        if not verbose:
+            return
+        if config_file and config_file.exists():
+            msg = f"Using config file '{config_file}'"
+        else:
+            msg = "No config file found, using default configuration"
```
Pierre-Sassoulas, line 420: This will never be executed because of line 396.

Pierre-Sassoulas (commented):
There's a a whole untested function "_expand_default" I think we should cover it or remove it unless it's optparse specific code. Also I think we could replace OptionManagerMixin by ArgumentManger like this:

```python

class OptionManagerMixin:

   def __init__(...)
        warnings.warn(...)
       self.argument_manager = ...
       
    def old_option_manager_function(...)
        return self.argument_manager.new_function(...)
```
To avoid duplication.

pylint/config/arguments_manager.py
DanielNoord: I think it can be executed if L393 fails?

DanielNoord:
> There's a a whole untested function "_expand_default" I think we should cover it or remove it unless it's optparse specific code. 

It's `optparse` specific. It's used in `_patch_optparse`.

> Also I think we could replace OptionManagerMixin by ArgumentManger like this:
> 
> ```python
> class OptionManagerMixin:
> 
>    def __init__(...)
>         warnings.warn(...)
>        self.argument_manager = ...
>        
>     def old_option_manager_function(...)
>         return self.argument_manager.new_function(...)
> ```
> 
> To avoid duplication.

I'd rather have the duplication for a little bit. Imo, it's much easier to just deprecate `OptionManagerMixin` and the complete file and work within the `ArgumentsManager` file. We need to keep instantiating `OptionManagerMixin` while `pyreverse` and `ConfigurationMixIn` are not yet on `argparse`. With the method proposed in this PR that work doesn't slow us down while working on refactoring these methods. We can start refactoring them already without impacting `pyreverse`.

_In general I prefer cutting ties with anything in the `optparse` files as strictly as possible. I think it makes it easier to see how far along in the process we are._

Obviously if you have a strong preference for doing it within the `OptionManagerMixin` class I will change this around!

pylint/config/arguments_manager.py
```diff
@@ -161,3 +225,274 @@ def _parse_command_line_configuration(
...
+        """Initialize configuration using default values."""
+        for provider in self.options_providers:
+            provider.load_defaults()
+
+    def read_config_file(
+        self, config_file: Optional[Path] = None, verbose: bool = False
+    ) -> None:
+        """Read the configuration file but do not load it (i.e. dispatching
+        values to each option's provider)
+        """
```
Pierre-Sassoulas, line 392:
```suggestion
        """
        if not config_file:
            if verbose:
                print("No config file found, using default configuration", file=sys.stderr)
            return
```

pylint/config/arguments_manager.py
```diff
@@ -161,3 +225,274 @@ def _parse_command_line_configuration(
...
+                    if not sect.isupper() and values:
+                        parser._sections[sect.upper()] = values
+
+        if not verbose:
+            return
+        if config_file and config_file.exists():
+            msg = f"Using config file '{config_file}'"
+        else:
+            msg = "No config file found, using default configuration"
+        print(msg, file=sys.stderr)
```
Pierre-Sassoulas, line 421:
```suggestion
        if verbose:
             print(f"Using config file '{config_file}'", file=sys.stderr)     
```

pylint/config/arguments_manager.py
Pierre-Sassoulas: You're right.

Pierre-Sassoulas (changes requested):

pylint/config/arguments_manager.py
```diff
@@ -161,3 +225,274 @@ def _parse_command_line_configuration(
...
+            else:
+                # Use this encoding in order to strip the BOM marker, if any.
+                with open(config_file, encoding="utf_8_sig") as fp:
+                    parser.read_file(fp)
+                # normalize each section's title
+                for sect, values in list(parser._sections.items()):
+                    if sect.startswith("pylint."):
+                        sect = sect[len("pylint.") :]
+                    if not sect.isupper() and values:
+                        parser._sections[sect.upper()] = values
```
Pierre-Sassoulas, line 413:
```suggestion
        config_file = Path(os.path.expandvars(config_file)).expanduser()
        if not config_file.exists():
            raise OSError(f"The config file {str(config_file)} doesn't exist!")
        parser = self.cfgfile_parser
        if config_file.suffix == ".toml":
            try:
                self._parse_toml(config_file, parser)
            except tomllib.TOMLDecodeError:
                pass
        else:
            # Use this encoding in order to strip the BOM marker, if any.
            with open(config_file, encoding="utf_8_sig") as fp:
                parser.read_file(fp)
            # normalize each section's title
            for sect, values in list(parser._sections.items()):
                if sect.startswith("pylint."):
                    sect = sect[len("pylint.") :]
                if not sect.isupper() and values:
                    parser._sections[sect.upper()] = values
```

Pierre-Sassoulas (approved):

PR #6218, opened 2022-04-07
Move utility function in testutil

Pierre-Sassoulas:
## Type of Changes

## Description

Done to avoid duplication when bursting test_self.py during the migration to argparse.

### DIFF
--- a/pylint/testutils/utils.py
+++ b/pylint/testutils/utils.py
@@ -4,7 +4,22 @@
 
 from __future__ import annotations
 
+import contextlib
 import os
+import sys
+from collections.abc import Iterator
+from typing import TextIO
+
+
+@contextlib.contextmanager
+def _patch_streams(out: TextIO) -> Iterator[None]:
+    """Patch and subsequently reset a text stream."""
+    sys.stderr = sys.stdout = out
+    try:
+        yield
+    finally:
+        sys.stderr = sys.__stderr__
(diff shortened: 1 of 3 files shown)


pylint/testutils/utils.py
```diff
@@ -2,8 +2,20 @@
...
 
+import contextlib
 import os
-from typing import List
+import sys
+from typing import Iterator, List, TextIO
+
+
+@contextlib.contextmanager
+def _patch_streams(out: TextIO) -> Iterator:
```
DanielNoord, line 12:
```suggestion
def _patch_streams(out: TextIO) -> Iterator[None]:
    """Patch and subsequently reset a text stream."""
```
Do we want to keep this private?

Pierre-Sassoulas: I think so, beside here's probably a way to use patch like I did for sys.path, I'm just not used to it.

DanielNoord (approved):

Pierre-Sassoulas:
Sorry for force pushing, there was a small formatting change I wanted to do and still be able to separate the two changes.

DanielNoord (approved):

pylint/testutils/utils.py
Pierre-Sassoulas: Ha shit I push forced without pulling the change.

PR #6223, opened 2022-04-07
Decrease number of runners used in CI

DanielNoord:

## Type of Changes

## Description

This should improve the speed of our CI. Since all tests jobs now depend on `Linux` passing there is really no need to separate these set-ups steps. They just take extra time setting up Python, checking out code and restoring the venv. That can all be done in the same runner: freeing up more runners and decreasing the overall runtime of the CI.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,11 +14,12 @@ env:
   CACHE_VERSION: 6
 
 jobs:
-  prepare-tests-linux:
-    name: prepare / ${{ matrix.python-version }} / Linux
+  tests-linux:
+    name: run / ${{ matrix.python-version }} / Linux
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 25
     strategy:
+      fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
     outputs:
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
But isn't having multiple step better from a caching point of view ?

DanielNoord:
> But isn't having multiple step better from a caching point of view ?

I don't think so in this instance. If you look at which steps gets removed those exact steps are take in the prepare step as well. The only difference is `Create Python virtual environment` and `Fail job if Python cache restore failed`. In the preparation job we create a new `venv` if we fail to restore it and in the test job we fail.

Pierre-Sassoulas:
I also realized (because the number of jobs is more manageable surely) that we're running the benchmark for 3.8 and 3.9. Should we also run it just for one interpreter ? (Especially seeing we never did anything with the result.)

DanielNoord:
> I also realized (because the number of jobs is more manageable surely) that we're running the benchmark for 3.8 and 3.9. Should we also run it just for one interpreter ? (Especially seeing we never did anything with the result.)

Might as well? Want me to add a commit to this PR to change it?

Pierre-Sassoulas:
I'm close to removing it completely, but I think we could do a github action on the model you did for pylint false positive we just do the benchmark on top of it. And we display new message, removed message and the modification of the time compared to the previous commit.

cdce8p:
Saw this PR today. Thanks @DanielNoord for doing it! We should probably do that change for astroid, too.

A bit more background information: The original structure (separate jobs for prepare and test) was similar to the one Home Assistant uses. For HA it has serval advantages that don't fully apply here.
* We only have a few dependencies, so creating the venv is fairly quick. In comparison HA has multiple hundreds deps.
* `prepare-tests` was only really used to run the `tests` itself. HA runs 4 different checks afterwards.
* I believe the "new" cache is only stored if the job itself completes successfully. That isn't a problem when venv creation is quick. For HA, it would be a noticeable slowdown.

PR #6232, opened 2022-04-08
Add example for `use-maxsplit-arg` message documentation

### LINKED ISSUE #6231: What is the rationale behind use-maxsplit-arg?
### Question

I have a python script that includes the following line:
```
"/my/endpoint".split("/")[-1]
```

When I run my script through `pylint`, it hits me with a `use-maxsplit-arg` warning and suggests I do the following instead:
```
"/my/endpoint".rsplit("/", maxsplit=1)[-1]
```

What I'm wondering is, **what is the rationale behind that rule?** It seems unnecessarily complex, and when I time things with `timeit` it looks like the recommended change actually takes slightly longer to execute than the original:
```
python -m timeit '"/my/endpoint".split("/")[-1]'
2000000 loops, best of 5: 124 nsec per loop

python -m timeit '"/my/endpoint".rsplit("/", maxsplit=1)[-1]'
2000000 loops, best of 5: 154 nsec per loop
```

Thanks!

This might already exist, but it'd be nice if there was an easy to access "rationale guide" behind some of these rules. Like I said, maybe that already exists but I wasn't able to find it quickly. I don't want to just make pylint ignore a rule because I don't understand it, but this one comes up often and I've never been able to figure out the rationale.

### Additional context

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #6231

### DIFF
--- a/doc/data/messages/u/use-maxsplit-arg/bad.py
+++ b/doc/data/messages/u/use-maxsplit-arg/bad.py
@@ -1,0 +1,2 @@
+url = "www.example.com"
+suffix = url.split(".")[-1]  # [use-maxsplit-arg]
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
👌

PR #6238, opened 2022-04-08
Added ungrouped-imports message example

matusvalo:
## Type of Changes

## Description

Refs #5953


matusvalo:
isort pre-commit hook "fixed" bad.py file causing it is not showing bad case anymore :-(

Pierre-Sassoulas:
> isort pre-commit hook "fixed" bad.py file causing it is not showing bad case anymore :-(

Right now you can do something similar to what was done for "reimported": https://github.com/PyCQA/pylint/blob/main/.pre-commit-config.yaml#L40. If it become too much we might exclude all the bad.py from autofixing.

PR #6239, opened 2022-04-08
Added wrong-import-order message example

matusvalo:
## Type of Changes

## Description

Refs #5953


matusvalo:
isort pre-commit hook "fixed" bad.py file causing it is not showing bad case anymore :-(

matusvalo:
Do we want to exclude isort per file

https://github.com/PyCQA/pylint/blob/0afa5bc5ea8641cce751b8f995fe8ce682f397fa/.pre-commit-config.yaml#L40

or we will exclude whole  `doc/data/messages` directory?

Pierre-Sassoulas:
> we will exclude whole doc/data/messages directory?

We could exclude all the ``bad.py`` first is the 'exclude' become hard to manage

.pre-commit-config.yaml
```diff
@@ -37,7 +37,7 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
-        exclude: doc/data/messages/r/reimported/bad.py
+        exclude: doc/data/messages/r/reimported/bad.py|doc/data/messages/w/wrong-import-order/bad.py
```
Pierre-Sassoulas, line 40:
```suggestion
        exclude: doc/data/messages/(r/reimported|w/wrong-import-order)/bad.py
```

Pierre-Sassoulas (changes requested):
Not sure about the regex, I'll revert if it breaks the existing code :)

Pierre-Sassoulas (approved):
Thank you @matusvalo !

PR #6263, opened 2022-04-11
Added no-else-continue message example

matusvalo:
## Type of Changes

## Description

Related #5953

### DIFF
--- a/doc/data/messages/n/no-else-continue/bad.py
+++ b/doc/data/messages/n/no-else-continue/bad.py
@@ -1,0 +1,6 @@
+def iterator_suffix(iterator, stop: int):
+    for i, item in enumerate(iterator):
+        if i < stop:  # [no-else-continue]
+            continue
+        else:
+            yield item
--- a/doc/data/messages/n/no-else-continue/good.py
+++ b/doc/data/messages/n/no-else-continue/good.py
@@ -1,0 +1,5 @@
+def iterator_suffix(iterator, stop: int):
+    for i, item in enumerate(iterator):
+        if i < stop:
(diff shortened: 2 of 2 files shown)


doc/data/messages/n/no-else-continue/bad.py
```diff
@@ -0,0 +1,6 @@
+def iterator_suffix(iterator, stop: int):
+    for i, item in enumerate(iterator):
+        if i < stop:  # [no-else-continue]
+            continue
+        else:
+            yield item
```
Pierre-Sassoulas, line 6:
```suggestion
def even_number_under(n: int):
    for i in range(n):
        if i%2 == 1:  # [no-else-continue]
            continue
        else:
            yield i
```

doc/data/messages/n/no-else-continue/good.py
```diff
@@ -0,0 +1,5 @@
+def iterator_suffix(iterator, stop: int):
+    for i, item in enumerate(iterator):
+        if i < stop:
+            continue
+        yield item
```
Pierre-Sassoulas, line 5:
```suggestion
def even_number_under(n: int):
    for i in range(n):
        if i%2 == 1:  # [no-else-continue]
            continue
        yield i
```

Pierre-Sassoulas (commented):
I feel like this example is easier to understand, let me know what you think :)

matusvalo:
I personally consider them as equal. I don't to use any of them.

Pierre-Sassoulas (approved):
Thanks for fixing my suggestion 😄

PR #6267, opened 2022-04-11
Create ``_OutputFormatAction``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -17,6 +17,7 @@ from pylint import extensions, interfaces, utils
 
 if TYPE_CHECKING:
     from pylint.config.help_formatter import _HelpFormatter
+    from pylint.lint import PyLinter
     from pylint.lint.run import Run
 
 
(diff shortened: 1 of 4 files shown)


pylint/lint/pylinter.py
```diff
@@ -282,13 +281,18 @@ def make_options() -> Options:
                 {
                     "default": "text",
                     "type": "string",
+                    # pylint: disable-next=fixme
+                    # TODO: Optparse: Uncomment these lines.
```
DanielNoord, line 285: Uncommenting this will fix the coverage drop in this PR but that requires completely fixing the `PyLinter` class for `argparse`. I have all PRs for that ready to go but I'm breaking them up to keep them reviewable.

pylint/config/callback_actions.py
```diff
@@ -283,3 +284,63 @@ def __call__(
...
+        dest: str,
+        nargs: None = None,
+        const: None = None,
+        default: None = None,
+        type: None = None,
+        choices: None = None,
+        required: bool = False,
+        help: str = "",
+        metavar: str = "",
+        **kwargs: "PyLinter",
```
Pierre-Sassoulas, line 304: I don't understand this one, we want to be able to handle any additional keyword argument and dump them ?

DanielNoord: No, it is similar to the `Run` actions. In the option dictionary I create the following key: `kwargs: {"linter": self}`. This means that by accessing `kwargs["linter"]` and setting it on the `Action` as instance attribute I can access the linter in `__call__`.
By adding this typing here we add typing to `self.linter` for this action.

Pierre-Sassoulas (approved):

PR #6272, opened 2022-04-11
Add ``music21`` and remove ``graph-explorer`` from primer

DanielNoord:

## Type of Changes

## Description

See https://github.com/PyCQA/pylint/issues/6049#issuecomment-1086108440.

The other package got its last commit in 2015 and hasn't updated its `pylintrc`. This will become problematic and we should probably exclude it. Open to any other suggestions for now packages!


Pierre-Sassoulas (approved):
Fair enough, removing dead project is a good thing !

PR #6273, opened 2022-04-11
Add typing to BaseChecker.process_token

Pierre-Sassoulas:
## Type of Changes

## Description

Add some typing in the checker. Simplify a piece of code that was hard to type otherwise.

### DIFF
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -428,7 +428,7 @@ class FormatChecker(BaseTokenChecker):
                 dispatch[token] = handler
         return dispatch
 
-    def process_tokens(self, tokens):
+    def process_tokens(self, tokens: list[tokenize.TokenInfo]) -> None:
         """Process tokens and search for :
 
         _ too long lines (i.e. longer than <max_chars>)
@@ -507,14 +507,14 @@ class FormatChecker(BaseTokenChecker):
                 "too-many-lines"
             )[0]
             names = (message_definition.msgid, "too-many-lines")
-            line = next(
+            lineno = next(
                 filter(None, (self.linter._pragma_lineno.get(name) for name in names)),
                 1,
             )
             self.add_message(
                 "too-many-lines",
                 args=(line_num, self.linter.config.max_module_lines),
-                line=line,
+                line=lineno,
             )
 
         # See if there are any trailing lines.  Do not complain about empty
--- a/pylint/checkers/raw_metrics.py
+++ b/pylint/checkers/raw_metrics.py
@@ -111,9 +111,7 @@ def get_type(
         line_type = "empty"
     elif i < len(tokens) and tokens[i][0] == tokenize.NEWLINE:
         i += 1
-    # We could do:
-    # line_type = cast(Literal["code", "docstring", "comment", "empty"], line_type)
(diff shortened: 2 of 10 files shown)


pylint/checkers/format.py
```diff
@@ -499,20 +499,15 @@ def process_tokens(self, tokens):
...
-            message_definition = self.linter.msgs_store.get_message_definitions(
-                "too-many-lines"
-            )[0]
-            names = (message_definition.msgid, "too-many-lines")
-            line = next(
-                filter(None, (self.linter._pragma_lineno.get(name) for name in names)),
-                1,
-            )
+            # Get the line where the too-many-lines (or its message id) was disabled
+            lineno = self.linter._pragma_lineno.get("C0302")
```
DanielNoord, line 503: I think using `get_message_definitions` here makes this more robust for any future changes to message id (going from 4 to 5 digits, old names, split-up checkers and new id's).

pylint/checkers/raw_metrics.py
```diff
@@ -107,6 +109,7 @@ def get_type(tokens, start_index):
         line_type = "empty"
     elif i < len(tokens) and tokens[i][0] == tokenize.NEWLINE:
         i += 1
+    line_type = cast(Literal["code", "docstring", "comment", "empty"], line_type)
     return i, pos[0] - start[0] + 1, line_type
```
DanielNoord, line 113: Does `pyright` do this correctly? I think adding an ignore here might be better: this seems like `mypy` failing to understand difficult work flow. By using a cast we never know when it does figure this out and when the cast is unnecessary.

pylint/checkers/format.py
Pierre-Sassoulas: It would also be vastly less performant: there's 4/5 calls to the message store to get the value, in process_token for each pylint run, for each user to get a value we already have that we'll maybe change once or twice in the future. (Also it did not handle old names).

DanielNoord: It's cached with a `maxsize=None` so performance should be okay after a first call right?

Pierre-Sassoulas: Right, let me do some benchmark / handling of old_names then.

pylint/checkers/raw_metrics.py
Pierre-Sassoulas: Fixed in latest commit

pylint/checkers/raw_metrics.py
```diff
@@ -109,7 +111,10 @@ def get_type(tokens, start_index):
         line_type = "empty"
     elif i < len(tokens) and tokens[i][0] == tokenize.NEWLINE:
         i += 1
-    return i, pos[0] - start[0] + 1, line_type
+    # We could do:
+    # line_type = cast(Literal["code", "docstring", "comment", "empty"], line_type)
+    # See https://github.com/PyCQA/pylint/pull/6273#discussion_r847783759
+    return i, pos[0] - start[0] + 1, line_type  # type: ignore[return-value]
```
DanielNoord, line 117:
```suggestion
    # Mypy fails to infer the literal of line_type
    return i, pos[0] - start[0] + 1, line_type  # type: ignore[return-value]
```

DanielNoord (approved):
One minor nitpick.

PR #6276, opened 2022-04-12
Deprecate most methods of ``_ArgumentsProvider``

DanielNoord:

## Type of Changes

## Description

One method remains. Since these methods are only used in `optparse` code I thought I would do most of them at once. We're nearing the end 😄

### DIFF
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -79,6 +79,12 @@ Other Changes
 
   Ref #5392
 
+* ``OptionsProviderMixIn`` has been replaced with ``ArgumentsProvider``. ``ArgumentsProvider`` is considered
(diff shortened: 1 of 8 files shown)


doc/whatsnew/2.14.rst
```diff
@@ -79,6 +79,12 @@ Other Changes
 
   Ref #5392
 
+* ``OptionsProviderMixIn`` has been replaced with ``ArgumentsProvider``. ``ArgumentsProvider`` is considered
+  private API and most methods that were public on ``OptionsProviderMixIn`` have now been deprecated and will
```
Pierre-Sassoulas, line 83: Pretty much agree that we should clearly state that things are private API, we want to be able to change our internal freely.

Pierre-Sassoulas (approved):
👍

PR #6280, opened 2022-04-12
Detect certain TypeErrors with incorrect usage of some built-ins

### LINKED ISSUE #6252: Some TypeErrors are not caught
### Current problem

It turns out that the built-in `tuple` takes exactly one argument. That is, to instantiate a tuple, it's `tuple([a, b])`, not `tuple(a, b)`. And maybe better still would be `(a, b)`. This is for Python 2 and 3. 

But pylint does not warn me of my mistake, which was `tuple(a, b)`. This statement throws a TypeError in both 2.7 and 3.9. 

### Desired solution

It would be great if this kind of brainfart could be caught by the linter. Consider also the other builtins, `len`, `list`, etc etc.

### Additional context

I'm not that good at Python (my "native languages" are C, Rust, assembler, which invest more heavily in static analysis and compiler checks and whatnot). This is why I'm using Pylint, and especially its `-E` mode to detect problems in my code before running it. I'm not sure if this means I'm using pylint outside of its intended arena, but I figured I could maybe file this in the hope its generally useful anyhow.

omarandlorraine:
## Type of Changes

## Description

I left both the bug and the glitter in that table up there because I'm not sure whether this counts as a bugfix or a new feature, or both!

But I ran into an issue where pylint didn't find a TypeError in my code. That's because the `visit_call` in the BaseChecker does not look at built-ins. So I just made some simple checks for some of the built-ins I know and added this check to pylint.

Closes #6252


omarandlorraine:
There is an error on lines 26 and 27 of `tests/functional/c/consider/consider_merging_isinstance.py`. My new lint is catching it, meaning that pylint's output is different from the expected output and the tests fail.

What shall we do?

jacobtylerwalls:
> There is an error on lines 26 and 27 of tests/functional/c/consider/consider_merging_isinstance.py. My new lint is catching it, meaning that pylint's output is different from the expected output and the tests fail.

Let's disable the new lint in those two lines: I think it's worth keeping the tests as written to make sure they don't double-warn with "consider-merging-isinstance".

jacobtylerwalls:
OTOH, the failing test for `        return lambda: tuple(1, 2)` I would just rewrite to `        return lambda: tuple([1, 2])`.

pylint/checkers/base/basic_checker.py
```diff
@@ -590,8 +595,72 @@ def _check_misplaced_format_function(self, call_node):
...
+            "divmod",
+            "filter",
+            "hasattr",
+            "help",
+            "isinstance",
+        ):
+            # These built-ins take 2 arguments
+            if len(node.args) != 2:
+                self.add_message(
+                    "incorrect-number-of-parameters", node=node, args=(node.func.name,)
```
jacobtylerwalls, line 655: Please add confidence=HIGH

pylint/checkers/base/basic_checker.py
```diff
@@ -590,8 +595,72 @@ def _check_misplaced_format_function(self, call_node):
...
+            "oct",
+            "ord",
+            "repr",
+            "reversed",
+            "set",
+        ):
+            # These built-ins take 0 or 1 argument
+            if len(node.args) > 1:
+                self.add_message(
+                    "incorrect-number-of-parameters", node=node, args=(node.func.name,)
```
jacobtylerwalls, line 642: confidence=HIGH

pylint/checkers/base/basic_checker.py
```diff
@@ -590,8 +595,72 @@ def _check_misplaced_format_function(self, call_node):
...
+                )
+        elif node.func.name in (
+            "delattr",
+            "divmod",
+            "filter",
+            "hasattr",
+            "help",
```
jacobtylerwalls, line 649: Doesn't `help` take 0-1 args?

jacobtylerwalls (changes requested):
Thanks for the PR! The next steps include:
 - a functional test in `tests/functional/i`
 - good.py and bad.py examples in `doc/data/messages/i`
 - a PR title that could serve as a commit message when squashed

DanielNoord:
I haven't looked at the code yet, but shouldn't this be taken care of in the `astroid` `stdlib` `brain`? What's the difference between `incorrect-number-of-parameters` and `too-many-function-args`?

omarandlorraine:
> Thanks, I wondered if we had a message already but didn't find one. Good spot!
> 
> Relevant:
> 
> https://github.com/PyCQA/pylint/blob/cd83ef4e5e4ae9576190e11b02bd0bc83a070ca5/pylint/checkers/typecheck.py#L1335-L1336

Oh, does my new code belong in `pylint/checkers/typecheck.py` then?

pylint/checkers/base/basic_checker.py
omarandlorraine: You are correct!

pylint/checkers/base/basic_checker.py
omarandlorraine: added in commit [c7f3753](https://github.com/PyCQA/pylint/pull/6280/commits/c7f3753b78313726a9b3ac7784534ff3ad1dd92a)

omarandlorraine:
> Thanks for the PR! The next steps include:
> 
>     * a PR title that could serve as a commit message when squashed

Is this title better?

omarandlorraine:
I'm trying to figure out how to make my `tests/functional/i/incorrect_number_of_parameters.txt` to match; this file is not formatted the same as "normal pylint". Is there some easy way to generate this file, or?

jacobtylerwalls:
Thanks for the follow-up. I think it might be wise to hear more from @DanielNoord about any hints for implementing https://github.com/PyCQA/pylint/pull/6280#issuecomment-1096706193 before going further. It's starting to look likely that the fix would be in the `astroid` library.

omarandlorraine:
> I'm trying to figure out how to make my `tests/functional/i/incorrect_number_of_parameters.txt` to match; this file is not formatted the same as "normal pylint". Is there some easy way to generate this file, or?

For my own reference, basically `python test_functional.py --update-functional-output -k "test_functional[incorrect_number_of_parameters]"`, as the output of a failed test says.

>Thanks for the follow-up. I think it might be wise to hear more from @DanielNoord about any hints for implementing https://github.com/PyCQA/pylint/pull/6280#issuecomment-1096706193 before going further. It's starting to look likely that the fix would be in the astroid library.

Alright, I'll keep an eye out to see what happens.

DanielNoord:
> Thanks for the follow-up. I think it might be wise to hear more from @DanielNoord about any hints for implementing [#6280 (comment)](https://github.com/PyCQA/pylint/pull/6280#issuecomment-1096706193) before going further. It's starting to look likely that the fix would be in the `astroid` library.

I haven't completely looked into it, but I think the fix should be here:
https://github.com/PyCQA/astroid/blob/6578a7d04dc28a539402dc27ef50ac756bc9a6d7/astroid/brain/brain_builtin_inference.py#L217

This is used by `infer_tuple` to create a `nodes.Tuple` I think. In any case looking at `register_builtin_transform(infer_tuple, "tuple")` on L902 there should be the starting point. I'm sorry @omarandlorraine this would have been a nice contribution, but I think that improving the `astroid` library (which helps us understand python code) is the better option here.
That might be a difficult task for a new contributor though, but I think we would all be happy to help and review a potential attempt at it!

omarandlorraine:
> I'm sorry @omarandlorraine this would have been a nice contribution, but I think that improving the astroid library (which helps us understand python code) is the better option here.

Yeah it's no problem, of course. And it wasn't a complete waste for me: I have a custom pylint I can use until something gets put in properly 😉. Plus it's very likely I'll be adding more lints soon, and now I know how to do it. But how will I know whether it should go in Pylint proper or in Astroid?

That astroid project looks interesting. The name clues me to think it has something to do with ASTs, so it seems like a richer alternative to [this std library](https://docs.python.org/3/library/ast.html).

But is there any real syntactic difference between builtins and other functions in Python? A few tests show that `tuple` et al. are not keywords, but more like names that magically don't need to be imported. So they can be shadowed, referenced, assigned and in other ways treated just like another variable. So by trying to make the correct inferences on the AST level, we need to watch out for horrorshows like the below:

```
a = tuple
a([1, 2]) # evaluates to a tuple, okay.

tuple = lambda x: x # pylint warns us about this
tuple([1, 2]) # evaluates to a list, not a tuple, bad times
a(1) # Should throw E0120 but doesn't
```

I'm not sure what to do about these details, if anything. And of course, the solution proposed by this PR does not produce the error it should either, because it looks at the name of the function call instead of the actual thing it references.

DanielNoord:
> Yeah it's no problem, of course. And it wasn't a complete waste for me: I have a custom pylint I can use until something gets put in properly 😉. Plus it's very likely I'll be adding more lints soon, and now I know how to do it. But how will I know whether it should go in Pylint proper or in Astroid?

In general, whenever we're trying to "understand" python code it is likely that it should be in `astroid`. In this case the hint was/is that there is already a warning for other type of functions just not for the `stdlib`. That's likely telling us that it is not our linter that is failing here, but our understanding of `stdlib` functions in `astroid`.

> That astroid project looks interesting. The name clues me to think it has something to do with ASTs, so it seems like a richer alternative to [this std library](https://docs.python.org/3/library/ast.html).

That's exactly what it is.

> But is there any real syntactic difference between builtins and other functions in Python? A few tests show that `tuple` et al. are not keywords, but more like names that magically don't need to be imported. So they can be shadowed, referenced, assigned and in other ways treated just like another variable. So by trying to make the correct inferences on the AST level, we need to watch out for horrorshows like the below:

The problem with `stdlib` functions is that most of them are implemented in C I think. So we can't use the standard inference to infer their signature. The `brain` I was referring to is our way of telling `astroid` "we know some stuff about these functions even though we can't actually read their signature". However, the brain is incomplete and doesn't tell `astroid` about the arguments apparently.

> I'm not sure what to do about these details, if anything. And of course, the solution proposed by this PR does not produce the error it should either, because it looks at the name of the function call instead of the actual thing it references.

Those details should probably be captured in the `astroid` `brain`.

I'm going to close this PR for now!

PR #6282, opened 2022-04-12
Use ``namespace`` in ``--generate-rcfile`` and deprecate it

DanielNoord:

## Type of Changes

## Description

I forgot about the fact that we do use `linter.config` in this function 😅 

Potentially blocked by #6199 but can be merged without it.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -151,13 +151,3 @@ features.rst: $(shell find ../pylint/checkers -type f -regex '.*\.py')
 	PYTHONPATH=$(PYTHONPATH) $(PYTHON) ./exts/pylint_features.py
 
 messages: $(PYTHONPATH) $(PYTHON) ./exts/pylint_messages.py
-
-gen-examples:
-	chmod u+w ../examples/pylintrc
-	pylint --rcfile=/dev/null --generate-rcfile > ../examples/pylintrc
-
-gen-man:
-	chmod u+w ../man/pylint.1
-	pylint --rcfile=/dev/null --generate-man > ../man/pylint.1
-
-all: html gen-man gen-examples
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -237,10 +237,8 @@ class _GenerateRCFileAction(_AccessRunObjectAction):
         values: Union[str, Sequence[Any], None],
         option_string: Optional[str] = "--generate-rcfile",
     ) -> None:
-        print(
-            "The '--generate-rcfile' option has been deprecated. "
-            "Please use the --generate-toml-config flag."
-        )
+        # pylint: disable-next=fixme
+        # TODO: Optparse: Deprecate this after discussion about this removal has been completed.
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", category=DeprecationWarning)
             self.run.linter.generate_config(skipsections=("Commands",))
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -237,12 +237,13 @@ class _GenerateRCFileAction(_AccessRunObjectAction):
         values: Union[str, Sequence[Any], None],
         option_string: Optional[str] = "--generate-rcfile",
     ) -> None:
-        # pylint: disable-next=fixme
-        # TODO: Optparse: Deprecate this function and raise a warning.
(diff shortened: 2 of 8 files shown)


Pierre-Sassoulas (changes requested):
I reviewed [just before you requested the review](https://github.com/PyCQA/pylint/pull/6282#pullrequestreview-939707353) 😄

DanielNoord:
> I reviewed [just before you requested the review](https://github.com/PyCQA/pylint/pull/6282#pullrequestreview-939707353) 😄

I already removed the deprecation after it 😄 I was requesting a re-review!

doc/Makefile
```diff
@@ -151,13 +151,3 @@ features.rst: $(shell find ../pylint/checkers -type f -regex '.*\.py')
 	PYTHONPATH=$(PYTHONPATH) $(PYTHON) ./exts/pylint_features.py
 
 messages: $(PYTHONPATH) $(PYTHON) ./exts/pylint_messages.py
-
-gen-examples:
```
Pierre-Sassoulas, line 155: We should generate the example as toml now, right ?

doc/Makefile
```diff
@@ -151,13 +151,3 @@ features.rst: $(shell find ../pylint/checkers -type f -regex '.*\.py')
...
-	chmod u+w ../examples/pylintrc
-	pylint --rcfile=/dev/null --generate-rcfile > ../examples/pylintrc
-
-gen-man:
```
Pierre-Sassoulas, line 159: Nice cleanup 👍

pylint/config/callback_actions.py
```diff
@@ -238,11 +238,10 @@ def __call__(
         option_string: Optional[str] = "--generate-rcfile",
     ) -> None:
         # pylint: disable-next=fixme
-        # TODO: Optparse: Deprecate this function and raise a warning.
-        # This is (obviously) dependent on a --generate-toml-config flag.
+        # TODO: Optparse: Deprecate this after discussion about this removal has been completed.
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", category=DeprecationWarning)
-            self.run.linter.generate_config(skipsections=("COMMANDS",))
+            self.run.linter.generate_config(skipsections=("Commands",))
```
Pierre-Sassoulas, line 244: Is the section title case insensitive ?

Pierre-Sassoulas (changes requested):

doc/Makefile
DanielNoord: We don't really use this.

Should probably take another look at this with that PR/issue about config documentation. For now this is unused.

pylint/config/callback_actions.py
DanielNoord: No, but it now follows the actual section name as defined in `options_groups`. I'm getting the option dicts from another dictionary attribute of `PyLinter` in which we use the actual names of the groups instead of these capitalised ones.

doc/Makefile
Pierre-Sassoulas: Yeah I think we have an outdated example of pylintrc because of this. I think it was supposed to be updated periodically. I would remove this script and add the equivalent inside the tbump.toml

DanielNoord: I think it can be easily done with an extension for Sphinx like we have for messages. I'll take a stab at this after `argparse` has been done to fix all of our documentation about this!

Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/6282/commits/c284444c3f18dd2c761233a9f2ee6bac88e8b172

tbump.toml
```diff
@@ -32,6 +32,10 @@ cmd = "python3 script/bump_changelog.py {new_version}"
 name = "Upgrade and check doc"
 cmd = "tox -e docs||echo 'Hack so this command does not fail'"
 
+[[before_commit]]
+name = "Upgrade the pylintrc examples"
+cmd = "pylint --rcfile=/dev/null --generate-rcfile > ../examples/pylintrc"
```
Pierre-Sassoulas, line 37:
```suggestion
cmd = "pylint --rcfile=/dev/null --generate-rcfile > examples/pylintrc"
```

Pierre-Sassoulas (approved):

PR #6286, opened 2022-04-12
Update links in documentation

### LINKED ISSUE #5830: Links for spyder in documentation are broken
### Question

The links for spyder, in the documentation, seem to be broken.

For example, in [Editor and IDE Integration](https://pylint.pycqa.org/en/latest/user_guide/ide-integration.html#ide-integration) it lists a good link as [this](https://docs.spyder-ide.org/pylint.html), which gives you a 404 error.  

Also the [Pylint home page](https://www.pylint.org/) lists [this](https://pythonhosted.org/spyder/pylint.html) which can't be reached.

I'm not sure, but there are probably other places you've got the wrong spyder link...

I think the better link is [this](https://docs.spyder-ide.org/current/panes/pylint.html)

### Additional context

_No response_

DanielNoord:

## Type of Changes

## Description

A lot of stuff in that make file is unnecessary I believe. I only kept the bare minimum we need to remove some clutter.

I have also turned on `linkcheck`. Turns out a lot of our links were broken. I have fixed all dead links (or removed them), while also changing a lot of the links that were automatically redirected.

Closes #5830


DanielNoord:
@Pierre-Sassoulas Would you consider making me a maintainer on RTD? I think we need to activate `linkcheck` there as well.

doc/Makefile
```diff
@@ -11,153 +11,29 @@ PYTHON        = /usr/bin/env python
...
-	-rm -rf $(BUILDDIR)/*
-	-rm -f reference_guide/features.rst
-	-rm -f extensions.rst
-
-html: features.rst extensions.rst
+html:
 	@echo
 	@echo "Install dependencies"
-	$(PIP) install python-docs-theme
+	$(PIP) install furo
```
Pierre-Sassoulas, line 26: Need to install from the requirement file instead.

doc/Makefile
```diff
@@ -11,153 +11,29 @@ PYTHON        = /usr/bin/env python
...
-
-gen-examples:
-	chmod u+w ../examples/pylintrc
-	pylint --rcfile=/dev/null --generate-rcfile > ../examples/pylintrc
-
-gen-man:
-	chmod u+w ../man/pylint.1
-	pylint --rcfile=/dev/null --generate-man > ../man/pylint.1
-
-all: html gen-man gen-examples
```
Pierre-Sassoulas, line 163: I don't think we should remove automatically generated option from the sphinx makefile, it cost nothing to maintain and if someone want to generate says an epub or json, it costs a lot to reinstate.

doc/make.bat
```diff
@@ -186,13 +40,4 @@ or in %BUILDDIR%/linkcheck/output.txt.
...
 
-if "%1" == "doctest" (
-	%SPHINXBUILD% -b doctest %ALLSPHINXOPTS% %BUILDDIR%/doctest
-	if errorlevel 1 exit /b 1
-	echo.
-	echo.Testing of doctests in the sources finished, look at the ^
-results in %BUILDDIR%/doctest/output.txt.
-	goto end
-)
-
```
Pierre-Sassoulas, line 197: Ditto.

doc/whatsnew/1.7.rst
```diff
@@ -444,9 +444,9 @@ New checkers
   One of the major undertakings with Python 3 was a reorganization of the standard library to
   remove old or supplanted modules and reorganize some of the existing modules.  As a result,
   roughly 100 modules that exist in Python 2 no longer exist in Python 3.  See
-  https://www.python.org/dev/peps/pep-3108/ and https://www.python.org/dev/peps/pep-0004/ for more
+  https://peps.python.org/pep-3108/ and https://peps.python.org/pep-0004/ for more
   information.  For suggestions on how to handle this, see
```
Pierre-Sassoulas, line 448:
```suggestion
  information.  There were suggestions on how to handle this, at
```

pylint/checkers/stdlib.py
```diff
@@ -420,7 +420,7 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             "unspecified-encoding",
             "It is better to specify an encoding when opening documents. "
             "Using the system default implicitly can create problems on other operating systems. "
-            "See https://www.python.org/dev/peps/pep-0597/",
+            "See hthttps://peps.python.org/pep-0597/",
```
Pierre-Sassoulas, line 423:
```suggestion
            "See https://peps.python.org/pep-0597/",
```

Pierre-Sassoulas (changes requested):

doc/Makefile
DanielNoord: Yes, but that doesn't really work here. This should be run from within the `doc` directory so the `-e .` line in the requirements file doesn't really work.
On the other hand RTD seems to install the requirements from the root directory so for them `-e .` does work. Can't use the requirements files here.

Either way, a reference to the previous theme should probably be removed.

doc/Makefile
DanielNoord: Re-instated 👍

Pierre-Sassoulas:
@DanielNoord what is your read the doc username ?

DanielNoord:
> @DanielNoord what is your read the doc username ?

DanielNoord

doc/Makefile
Pierre-Sassoulas: Ok, not worth the headache to change the -e 😄

DanielNoord:
Going to merge the change to links and do the update to the configuration in another PR as I'm struggling to get that right... 😅

PR #6288, opened 2022-04-13
docparams extension considers type comments as type documentation.

### LINKED ISSUE #6287: docparams should support missing type documentation when type comments are used
### Current problem

When running pylint (`pylint --disable=all --enable=missing-type-doc --load-plugins=pylint.extensions.docparams code.py`) with the following code, no errors are displayed:

```python
def variant(node: str) -> list[str]:
    """Get the variants for the given `node`.

    Args:
        node: The node to get the variants for.

    Returns:
        The available variants.

    """
    return ["hi"]
```

However when type comments are used, a `missing-type-doc` warning is shown:

```python
def variant(node):
    # type: (str) -> list[str]
    """Get the variants for the given `node`.

    Args:
        node: The node to get the variants for.

    Returns:
        The available variants.

    """
    return ["hi"]
```

### Desired solution

Ideally docparams would allow type comments to count as type documentation, just like regular type annotations are allowed.
This will require changes on these lines: https://github.com/PyCQA/pylint/blob/a6cb836da56b051cdc0d584d726fb7579415cb54/pylint/extensions/docparams.py#L560-L563

### Additional context

_No response_

AWhetter:
## Type of Changes

## Description

The docparams extension won't raise a `missing-type-doc` when type annotations are given. This changes means that it also won't raise a  `missing-type-doc` when type comments are given.

Closes #6287

### DIFF
--- a/pylint/extensions/_check_docs_utils.py
+++ b/pylint/extensions/_check_docs_utils.py
@@ -6,7 +6,9 @@
 
 from __future__ import annotations
 
+import itertools
 import re
+from collections.abc import Iterable
 
 import astroid
 from astroid import nodes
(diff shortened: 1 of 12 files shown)


pylint/extensions/_check_docs_utils.py
```diff
@@ -156,6 +158,109 @@ def possible_exc_types(node: nodes.NodeNG) -> Set[nodes.ClassDef]:
...
 
 
+def _is_ellipsis(node: astroid.NodeNG):
+    if sys.version_info < (3, 8):
+        return isinstance(node, astroid.Ellipsis)
+
+    return isinstance(node, astroid.Const) and node.value == Ellipsis
+
+
+def _merge_annotations(annotations, comment_annotations):
```
Pierre-Sassoulas, line 168:
```suggestion
def _merge_annotations(annotations: List[nodes.?], comment_annotations: List[nodes.?]) -> Generator:
```
Not sure if we can be more specific than NodeNG here..

Pierre-Sassoulas:
What do you think of this astroid issue https://github.com/PyCQA/astroid/issues/1508 @AWhetter ?

pylint/extensions/_check_docs_utils.py
AWhetter: I don't think that we can. The astroid annotations of the attributes that we get values from to pass into this function list them only as NodeNG at the moment.

Pierre-Sassoulas (dismissed):
LGTM, thank you 👍 I'm going to wait for another review if possible but this is definitely going in 2.16.0

pylint/extensions/_check_docs_utils.py
```diff
@@ -159,6 +162,111 @@ def possible_exc_types(node: nodes.NodeNG) -> set[nodes.ClassDef]:
...
 
+def _is_ellipsis(node: astroid.NodeNG) -> bool:
+    if not PY38_PLUS:
+        return isinstance(node, astroid.Ellipsis)
+
+    return isinstance(node, astroid.Const) and node.value == Ellipsis
+
+
+def _merge_annotations(
+    annotations: Iterable[astroid.NodeNG], comment_annotations: Iterable[astroid.NodeNG]
```
DanielNoord, line 173: We should probably follow the style of importing from `nodes` instead of `astroid` here.

Pierre-Sassoulas (approved):
The CI fail on the doc is a temporary fail in reaching a twitter link.

Pierre-Sassoulas (approved):
Let's merge this 🎉

Pierre-Sassoulas:
@AWhetter the pypi 3.7 tests seems genuine, but also it's on an old interpreter, should we just ignore it ?

Pierre-Sassoulas:
I tried to add coverage ans force pushed some code but I was out of idea to get the last line @AWhetter :) we're very close to being able to merge

AWhetter:
@Pierre-Sassoulas Thanks for adding those tests! I've included them again and added a test that covers that final line.
The test failure seems to be happening in the main branch and is unrelated to these changes. So I think that this is ready to merge.

Pierre-Sassoulas (approved):
Amazing, thank you for covering the last missing part 👍

PR #6290, opened 2022-04-13
Make the enable and disable config attributes an actual list

DanielNoord:

## Type of Changes

## Description

I tried to find when this was actually introduced and went back 8+ years into the code but couldn't find the specific commit that did this.

I don't think the current attribute is how we want this to be: why would it be 
```python
self.namespace.disable = [["raw-checker-failed"], ["bad-inline-option"], ["locally-disabled"]]
```
instead of
```python
self.namespace.disable = ["raw-checker-failed", "bad-inline-option", "locally-disabled"]
```

Also note, we don't actually use this config attribute 😅 The only time this gets accessed is during `--generate-rcfile` and `--generate-toml-config`. The other times we simply call `self.linter.disable()` or `enable()` and only update this attribute in `_set_msg_status`. It's just used as reference, but `namespace.disable` or `namespace.enable` serves no other purposes beyond that as far as I can see.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -1820,9 +1820,9 @@ class PyLinter(  # type: ignore[misc]
         self.namespace.disable = []
         for mid, val in self._msgs_state.items():
             if val:
-                self.namespace.enable.append(self._message_symbol(mid))
+                self.namespace.enable += self._message_symbol(mid)
             else:
-                self.namespace.disable.append(self._message_symbol(mid))
+                self.namespace.disable += self._message_symbol(mid)
 
     def _register_by_id_managed_msg(
         self, msgid_or_symbol: str, line: Optional[int], is_disabled: bool = True
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -1722,17 +1722,6 @@ class PyLinter(  # type: ignore[misc]
 
     # Setting the state (disabled/enabled) of messages and registering them
 
-    def _message_symbol(self, msgid: str) -> List[str]:
-        """Get the message symbol of the given message id.
-
-        Return the original message id if the message does not
-        exist.
-        """
-        try:
-            return [md.symbol for md in self.msgs_store.get_message_definitions(msgid)]
-        except exceptions.UnknownMessageError:
-            return [msgid]
-
     def _set_one_msg_status(
         self, scope: str, msg: MessageDefinition, line: Optional[int], enable: bool
     ) -> None:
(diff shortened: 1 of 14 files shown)


pylint/lint/pylinter.py
```diff
@@ -1793,9 +1793,9 @@ def _set_msg_status(
         self.namespace.disable = []
         for mid, val in self._msgs_state.items():
             if val:
```
Pierre-Sassoulas, line 1795:
```suggestion
        for msgid_or_symbol, is_enabled in self._msgs_state.items():
            if is_enabled:
```

pylint/lint/pylinter.py
```diff
@@ -1793,9 +1793,9 @@ def _set_msg_status(
...
-                self.namespace.enable.append(self._message_symbol(mid))
+                self.namespace.enable += self._message_symbol(mid)
```
Pierre-Sassoulas, line 1796:
```suggestion
                self.namespace.enable += self._message_symbol(msgid_or_symbol)
```

pylint/lint/pylinter.py
```diff
@@ -1793,9 +1793,9 @@ def _set_msg_status(
...
             else:
-                self.namespace.disable.append(self._message_symbol(mid))
+                self.namespace.disable += self._message_symbol(mid)
```
Pierre-Sassoulas, line 1798:
```suggestion
                self.namespace.disable += self._message_symbol(msgid_or_symbol)
```

Pierre-Sassoulas (commented):
I started to make suggestion in the review but I think ``_message_symbol`` needs to be refactored. I think I4m the one who introduced this change when we had to go from a string to a list of message because missing-docstring has multiple child so we needed to handle a list of disabled messages for a single disable/enable.

Let me do that in an IDE :)

DanielNoord:
> I started to make suggestion in the review but I think `_message_symbol` needs to be refactored. I think I4m the one who introduced this change when we had to go from a string to a list of message because missing-docstring has multiple child so we needed to handle a list of disabled messages for a single disable/enable.
> 
> Let me do that in an IDE :)

I think if we want to return multiple names we will always have this issue though right? Appending a tuple or iterator or whatever will just add nested lists while we should add them up.

I like the name change!

tests/config/functional/toml/unknown_msgid/enable_unknown_msgid.result.json
```diff
@@ -0,0 +1,9 @@
+{
+  "functional_append": {
+    "disable": ["logging-format-interpolation"],
+    "enable": ["locally-disabled"]
```
Pierre-Sassoulas, line 4: I don't know if it's a good thing that the issue is silently ignored when we do not recognize a msgid or symbol but this is what happens right now.

Pierre-Sassoulas: Maybe something to handle in another PR.

DanielNoord: Yeah this what was I was talking about in that previous PR. There is an `TODO: Optparse` for it!

DanielNoord (commented):
LGTM!

I can't review my own PR

PR #6292, opened 2022-04-13
Update documentation related settings

DanielNoord:

## Type of Changes

## Description

Let's see if this works.


.github/workflows/checks.yaml
```diff
@@ -182,3 +183,8 @@ jobs:
         run: |
           . venv/bin/activate
           pytest doc/test_messages_documentation.py
+      - name: Check documentation build and links
+        run: |
+          . venv/bin/activate
+          cd doc
+          make html
```
DanielNoord, line 190: We can't add `linkcheck` to readthedocs as the only allow one build target (and they don't use `make html` but `sphinx-build -b html` directly).

Thus, we need our own CI to run `linkcheck`. Takes 2 additional minutes but I think that's okay considering the time most other jobs take.

.readthedocs.yaml
```diff
@@ -7,6 +7,10 @@ sphinx:
...
 
 python:
-  version: 3.8
   install:
     - requirements: doc/requirements.txt
+
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.8"
```
DanielNoord, line 16: `python.version` has been deprecated and should now be specified in the `build` tag

doc/Makefile
```diff
@@ -13,9 +13,9 @@ PYTHONPATH    =
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
+ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees -T -E -W --keep-going $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
```
DanielNoord, line 16: I added these flags as they are also used by readthedocs and we therefore mimic that behaviour locally now

doc/Makefile
```diff
@@ -34,21 +34,21 @@ help:
...
+html:
 	@echo
 	@echo "Install dependencies"
-	$(PIP) install python-docs-theme
+	$(PIP) install furo
 	@echo "Finished installing dependencies"
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	$(SPHINXBUILD) -b linkcheck $(ALLSPHINXOPTS) $(BUILDDIR)/linkcheck
 	@echo
-	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
```
DanielNoord, line 51: `-b html` already logs stuff like this to `stdout`. No reason to log twice.

Pierre-Sassoulas (approved):

Issue #5364, opened 2021-11-22
Better primer tests (False positives detection)


Pierre-Sassoulas:
### Current problem

Right now the primer tests are only checking for fatal messages and crashes. But anticipating false positive would be a huge plus in term of quality of releases.

### Desired solution

A solution to anticipate the false positives by running pylint on external repositories. 

### Additional context

Originally posted by @DanielNoord (https://github.com/PyCQA/pylint/pull/5173#issuecomment-968105390):

Some points I found while working on this:
1) The initial approach of asserting `ex.code == 0` doesn't work since many packages will return error messages. Even the `stdlib` packages. We should therefore only look for crashes (`ex.code == 32`) or fatal messages (`ex.code == 1`) I think.

2) I have created a new CI job to do the primer tests on Linux on every push or pull requests. We don't want to run them only when bumping a version (as we did with the previous acceptance tests), but I think it is good to separate them from the other tests. Especially since they will probably take much longer to complete (more on that later) and sometimes early fails on the tests are helpful in finalising a PR (especially after GitHub review comments break a test).

3) The lazy loading of repo's checks for the SHA1 hashes of the local commit and remote commit and reclones when it finds a difference.

4) `tests/primer/primer_external_packages.py` is used to store a list of packages to download. We currently include `numpy` but I would argue against including it. `numpy` has its tests included in the sources files. Normally we don't want to run `pylint` over tests as this can create problems when tests use frameworks that use non-normal python code. Besides, running over the tests also really inflates the time it takes to run the primer tests. I would think we could come up with 2/3 other projects that might be better. Note that these projects do not need to use `pylint`, we just need to be able to assume that their code is "normal" and therefore shouldn't crash `pylint` or `astroid`.

5) We might also want to improve the message that gets raised when the test fails. Currently it is not immediately clear where `pylint` crashed. Improving the message might help expedite the process of fixing it.
Perhaps we should add `--only-fatal`? To make the output only print fatal errors?

Originally posted by @cdce8p  (https://github.com/PyCQA/pylint/pull/5173#discussion_r753879354):

Maybe the --capture=tee-sys option works?
https://pytest.org/en/6.2.x/capture.html#setting-capturing-methods-or-disabling-capturing

Originally posted by @DanielNoord (https://github.com/PyCQA/pylint/pull/5173#issuecomment-968272755):

[...]

PR #6293, opened 2022-04-13
Do not use deprecated methods in ``PyLinter``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -670,9 +670,6 @@ class PyLinter(  # type: ignore[misc]
             ("RP0003", "Messages", report_messages_stats),
         )
         self.register_checker(self)
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", category=DeprecationWarning)
-            self.load_provider_defaults()
 
     def load_default_plugins(self):
         checkers.initialize(self)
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (changes requested):
The decrease in coverage makes me think that PyLinter might be used by Plugin lib and that it would be safer to remove that only after the deprecation period.

DanielNoord:
@Pierre-Sassoulas I have fixed the coverage be re-enabling another test (in a hacky way).

The coverage decrease is now due to some inner `if` statements of some functions not being reached or some validators not being used. This is expected as we decrease the number of calls to `optparse` functions. Note that we do cover all `DeprecationWarnings`.

Pierre-Sassoulas (approved):

cdce8p:
This change seems to cause a major regression. At least for Home Assistant the `disable` config values aren't loaded anymore. _HA uses `pyproject.toml` for the pylint config if that helps._

To reproduce, it's enough to install the basic HA requirements and run
```
pylint homeassistant/core.py
```

DanielNoord:
Investigating!

DanielNoord:
> This change seems to cause a major regression. At least for Home Assistant the `disable` config values aren't loaded anymore. _HA uses `pyproject.toml` for the pylint config if that helps._
> 
> To reproduce, it's enough to install the basic HA requirements and run
> 
> ```
> pylint homeassistant/core.py
> ```

Did you bisect to this commit? I'm trying to replicate but I'm seeing difference in behaviour in `pyenv` and `virtualenv`. Seems like configuration discovery might be broken. Don't immediately see why this commit would cause that, but I'll keep looking.

cdce8p:
> Did you bisect to this commit?

Yes.

> I'm trying to replicate but I'm seeing difference in behaviour in `pyenv` and `virtualenv`. Seems like configuration discovery might be broken. Don't immediately see why this commit would cause that, but I'll keep looking.

I'm using the stdlib `venv` module. I also did a `pip install -e <path>/pylint` beforehand.

DanielNoord:
Error found! PR incoming 😄

Pierre-Sassoulas:
The fact we merged without detecting the problem is worrying, maybe #5364 would have helped ?

DanielNoord:
> The fact we merged without detecting the problem is worrying, maybe #5364 would have helped ?

We actually had a test for this specific bug but our test was wrong 😅

PR #6299, opened 2022-04-13
Add ``_ArgumentsManager.set_option`` and deprecate old option setters

DanielNoord:

## Type of Changes

## Description

Bit of a larger one.

@Pierre-Sassoulas  Let me know what you think of my hack of deprecating unwanted arguments. It's not pretty, but it might work and is more efficient than doing it with `inspect` or something like that.

### DIFF
--- a/pylint/checkers/similar.py
+++ b/pylint/checkers/similar.py
@@ -798,33 +798,18 @@ class SimilarChecker(BaseChecker, Similar, MapReduceMixin):
         Overridden to report options setting to Similar
         """
         # pylint: disable-next=fixme
-        # TODO: Refactor after OptionProvider has been moved to argparse
-        BaseChecker.set_option(self, optname, value, action, optdict)
+        # TODO: Deprecate the non required arguments
+        self.linter.set_option(optname, value)
         if optname == "min-similarity-lines":
-            self.min_lines = (
-                getattr(self.linter.namespace, "min_similarity_lines", None)
-                or self.config.min_similarity_lines
-            )
+            self.min_lines = self.linter.namespace.min_similarity_lines
         elif optname == "ignore-comments":
-            self.ignore_comments = (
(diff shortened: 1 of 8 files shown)


DanielNoord:
Coverage drop is as expected.

pylint/checkers/variables.py
```diff
@@ -988,7 +988,7 @@ class VariablesChecker(BaseChecker):
         (
             "init-import",
             {
-                "default": 0,
+                "default": False,
```
Pierre-Sassoulas, line 991: 👍

pylint/checkers/similar.py
```diff
@@ -798,33 +798,18 @@ def set_option(self, optname, value, action=None, optdict=None):
...
-        BaseChecker.set_option(self, optname, value, action, optdict)
+        # TODO: Deprecate the non required arguments
+        self.linter.set_option(optname, value)
         if optname == "min-similarity-lines":
-            self.min_lines = (
-                getattr(self.linter.namespace, "min_similarity_lines", None)
-                or self.config.min_similarity_lines
-            )
+            self.min_lines = self.linter.namespace.min_similarity_lines
         elif optname == "ignore-comments":
```
Pierre-Sassoulas, line 805: Do we need the conditional and set_option here ? Can't we do 
```python
self.min_lines = self.linter.namespace.min_similarity_lines
...
self.ignore_signatures = self.linter.namespace.ignore_signatures
```
Directly inside the constructor of ``Similar`` ?

DanielNoord: We probably can, but I need to refactor this anyway. (See the TODO).

I'll take a look at these assignments at the same time. I'm likely going to deprecate this complete method and just use `set_option` from the `Manager`. No need to duplicate if both work fine.

Pierre-Sassoulas (approved):

PR #6300, opened 2022-04-14
Create ``_ExtendArgument``

DanielNoord:

## Type of Changes

## Description

Currently uncovered but necessary for `pyreverse` since it has an `append` option, which in `argparse` is the same as an `extend` option.

Added in a separate PR to keep the final `pyreverse` PR light-weight.

### DIFF
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -52,6 +52,27 @@ class _DoNothingAction(_CallbackAction):
         return None
 
 
+class _ExtendAction(argparse._AppendAction):
+    """Action that adds the value to a pre-existing list.
+
+    It is directly copied from the stdlib implementation which is only available
+    on 3.8+.
+    """
+
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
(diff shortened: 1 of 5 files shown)


pylint/config/callback_actions.py
```diff
@@ -52,6 +52,27 @@ def __call__(
         return None
 
 
+class _ExtendAction(argparse._AppendAction):
+    """Action that adds the value to a pre-existing list.
+
+    It is directly copied from the stdlib implementation which is only available
+    on 3.8+.
```
Pierre-Sassoulas, line 59: Maybe we could start a TODO list for when we drop 3.7 like #4683 to not forget to cleanup that kind of thing.

Pierre-Sassoulas (approved):
LGTM !

pylint/config/callback_actions.py
DanielNoord: I think whenever we drop support for a version we're likely to do a `grep` for `3.8` `3, 8` and other combinations. But feel free to open one!

Pierre-Sassoulas: https://github.com/PyCQA/pylint/issues/6306

PR #6302, opened 2022-04-14
Use ``python-typing-update`` on ``pylint`` directory

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/__init__.py
+++ b/pylint/__init__.py
@@ -2,16 +2,18 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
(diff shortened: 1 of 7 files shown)


cdce8p (approved):

DanielNoord:
@cdce8p Thanks for the review.

Not sure if this is something you want to look into but I noticed the tool has some trouble with our copyright notices and module docstrings. Removing most whitespaces around them. Might also be because of `reorder-imports`, but it could be something you want to look into.

cdce8p:
> Not sure if this is something you want to look into but I noticed the tool has some trouble with our copyright notices and module docstrings. Removing most whitespaces around them. Might also be because of `reorder-imports`, but it could be something you want to look into.

Saw that too. I think it's not that big of an issue. As you said, likely caused by `reorder-imports`.

PR #6304, opened 2022-04-14
Use ``python-typing-update`` on ``pylint/utils`` and ``message`` dirs

DanielNoord:

## Type of Changes

## Description

Ref #4683.

### DIFF
--- a/pylint/utils/file_state.py
+++ b/pylint/utils/file_state.py
@@ -2,18 +2,11 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
+from __future__ import annotations
+
 import collections
 import sys
-from typing import (
-    TYPE_CHECKING,
-    DefaultDict,
-    Dict,
-    Iterator,
-    Optional,
-    Set,
-    Tuple,
-    Union,
-)
+from typing import TYPE_CHECKING, DefaultDict, Dict, Iterator
 
 from astroid import nodes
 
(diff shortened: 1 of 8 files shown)


pylint/utils/file_state.py
```diff
@@ -2,18 +2,11 @@
...
-    TYPE_CHECKING,
-    DefaultDict,
-    Dict,
-    Iterator,
-    Optional,
-    Set,
-    Tuple,
-    Union,
-)
+from typing import TYPE_CHECKING, DefaultDict, Dict, Iterator
```
cdce8p, line 9: Just noticed `Iterator` and `DefaultDict` which could both be imported from `collections.abc`. For that to happen automatically, `python-typing-update` would need to be run with `--py39-plus` though. https://github.com/asottile/reorder_python_imports#rewriting-pep-585-typing-imports

Maybe that's worth doing after the initial round of changes, since it could also include incompatible changes. Otherwise pylint should also detect deprecated typing aliases once we update the py-version setting.

cdce8p (approved):

pylint/utils/file_state.py
DanielNoord: Hm, yeah I was wondering about `DefaultDict` anyway. This PR includes a line where a `defaultdict` is typed as `dict`. That might also be something we want to change? Manual check of all `defaultdicts` wouldn't hurt I guess.

cdce8p: Small update, for `python-typing-update`, both the `--py39-plus` and `--full-reorder` setting are necessary to rewrite all PEP 585 typing aliases. https://github.com/cdce8p/python-typing-update#configuration

However, this might also change runtime code (e.g. in type aliases). Might be best to just use the `consider-using-alias` check for it. (With `py-version = 3.7.2`)

DanielNoord: Yeah I saw one too just now. Fine with me, let's do these changes first and see what `pylint` reports for us!

PR #6307, opened 2022-04-14
Use ``python-typing-update`` on ``pylint/config`` directory

DanielNoord:

## Type of Changes

## Description

Ref #4683.

### DIFF
--- a/pylint/config/utils.py
+++ b/pylint/config/utils.py
@@ -29,7 +29,14 @@ if TYPE_CHECKING:
 
 def _convert_option_to_argument(
     opt: str, optdict: dict[str, Any]
-) -> _StoreArgument | _StoreTrueArgument | _CallableArgument | _StoreOldNamesArgument | _StoreNewNamesArgument | _ExtendArgument:
+) -> (
+    _StoreArgument
+    | _StoreTrueArgument
+    | _CallableArgument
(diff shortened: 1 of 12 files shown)


pylint/config/utils.py
```diff
@@ -27,15 +28,8 @@
...
-) -> Union[
-    _StoreArgument,
-    _StoreTrueArgument,
-    _CallableArgument,
-    _StoreOldNamesArgument,
-    _StoreNewNamesArgument,
-    _ExtendArgument,
-]:
+    opt: str, optdict: dict[str, Any]
+) -> _StoreArgument | _StoreTrueArgument | _CallableArgument | _StoreOldNamesArgument | _StoreNewNamesArgument | _ExtendArgument:
```
DanielNoord, line 32: @cdce8p For some reason `black` doesn't format this. Is this a known issue? And if so, how do we want to format this ourselves?

cdce8p (approved):

pylint/config/utils.py
cdce8p: Not sure. You could try adding brackets and see if that helps with black.
```py
    ...
) -> (
    _StoreArgument
    | _StoreTrueArgument
    ...
)
```

PR #6309, opened 2022-04-14
Use ``python-typing-update`` on ``pylint/pyreverse`` directory

DanielNoord:

## Type of Changes

## Description

Ref #4683.

@Pierre-Sassoulas @cdce8p Perhaps comment if you're going to review as not to duplicate work.

### DIFF
--- a/pylint/message/message.py
+++ b/pylint/message/message.py
@@ -2,8 +2,10 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
+from __future__ import annotations
+
 import collections
-from typing import Optional, Tuple, Union, overload
+from typing import overload
 from warnings import warn
 
 from pylint.constants import MSG_TYPES
@@ -41,8 +43,8 @@ class Message(_MsgBase):
         symbol: str,
         location: MessageLocationTuple,
         msg: str,
-        confidence: Optional[Confidence],
-    ) -> "Message":
+        confidence: Confidence | None,
+    ) -> Message:
         ...
 
     @overload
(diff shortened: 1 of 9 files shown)


pylint/message/message.py
```diff
@@ -41,33 +43,30 @@ def __new__(
...
 
     def __new__(
         cls,
         msg_id: str,
         symbol: str,
-        location: Union[
-            Tuple[str, str, str, str, int, int],
-            MessageLocationTuple,
-        ],
+        location: (tuple[str, str, str, str, int, int] | MessageLocationTuple),
```
Pierre-Sassoulas, line 66:
```suggestion
        location: MessageLocationTuple,
```
Could we do this ? It seems those should be the same thing.

DanielNoord: No, see the deprecation warning a couple of lines below. `location` needs to be initialised as `MessageLocationTuple` and not be a regular `tuple`.

pylint/message/message.py
```diff
@@ -41,33 +43,30 @@ def __new__(
...
+    ) -> Message:
         ...
 
     @overload
     def __new__(
         cls,
         msg_id: str,
         symbol: str,
-        location: Tuple[str, str, str, str, int, int],
+        location: tuple[str, str, str, str, int, int],
```
Pierre-Sassoulas, line 55:
```suggestion
        location: MessageLocationTuple,
```
(🤞 ?)

Pierre-Sassoulas (approved):

pylint/message/message.py
cdce8p, line 66: The brackets could have been removed here though.
```diff
-    location: (tuple[str, str, str, str, int, int] | MessageLocationTuple),
+    location: tuple[str, str, str, str, int, int] | MessageLocationTuple,
```

DanielNoord: Added in follow-up!

PR #6314, opened 2022-04-14
Set ``Namespace`` on ``config`` instead of ``namespace``

DanielNoord:

## Type of Changes

## Description

Ref #5392.

### DIFF
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -61,7 +61,7 @@ class _ArgumentsManager:
     """Arguments manager class used to handle command-line arguments and options."""
 
     def __init__(self, prog: str, usage: str | None = None) -> None:
-        self.namespace = argparse.Namespace()
+        self._config = argparse.Namespace()
         """Namespace for all options."""
 
         self._arg_parser = argparse.ArgumentParser(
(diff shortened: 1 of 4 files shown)


pylint/config/arguments_manager.py
```diff
@@ -95,7 +95,20 @@ def __init__(self, prog: str, usage: str | None = None) -> None:
...
+    def config(self, value: argparse.Namespace) -> None:
+        self._config = value
+
+    @property
+    def namespace(self) -> argparse.Namespace:
+        return self.config
+
+    @namespace.setter
+    def namespace(self, value: argparse.Namespace) -> None:
+        self.config = value
```
Pierre-Sassoulas, line 111:
```suggestion
```
Isn't it the moment where we change all our call to 'namespace' to 'config' ?

DanielNoord: Wanted to do that in a separate PR (due to the size of that change). Didn't want to let this important change go unnoticed!

tests/test_regr.py
```diff
@@ -118,6 +118,9 @@ def test_check_package___init__(finalize_linter: PyLinter) -> None:
     assert checked == ["__init__"]
 
 
+# pylint: disable-next=fixme
+# TODO: Optparse: Test are broken because of property shenanigans of config attribute
+@pytest.mark.xfail
```
Pierre-Sassoulas, line 123: Why not update the test ? We changed the Pylinter code, it's logical that the result of the test change ?

DanielNoord: Yeah, because of the three-step property chaining `astroid` can't infer `linter.config` anymore and returns `Uninferable`. When we simplify those attributes this will probably suddenly no longer fail 🎉

Pierre-Sassoulas: All right !

Pierre-Sassoulas (approved):

PR #6315, opened 2022-04-14
Fix parsing of unknown message IDs

### LINKED ISSUE #4324: Add warning if a message in disable or enable inside configuration does not exists
This was supposed to be a small change but this is taking longer than I thought and deserve an issue.

### Additional context

Detected here: https://github.com/PyCQA/pylint/issues/4265#issuecomment-810482361
First refactor leading to it here: #4323

DanielNoord:

## Type of Changes

## Description

The `try...expect` should have been placed differently. The test was wrong here as well, `"logging-format-interpolation"` should be in `disable` even after the previous message is unknown.

@cdce8p This was caused because in the `homeassistant` config there is a message that has been removed. This caused us to stop parsing the `disable` key when we reached it.

Ref #6293.

Closes #4324.

### DIFF
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -382,9 +382,8 @@ class _DisableAction(_AccessLinterObjectAction):
             try:
                 self.linter.disable(msgid)
             except exceptions.UnknownMessageError:
-                # pylint: disable-next=fixme
-                # TODO: Optparse: Raise an informational warning here
-                pass
+                msg = f"{option_string}. Don't recognize message {msgid}."
+                self.linter.add_message("bad-option-value", args=msg, line=0)
 
 
 class _EnableAction(_AccessLinterObjectAction):
@@ -403,9 +402,8 @@ class _EnableAction(_AccessLinterObjectAction):
             try:
                 self.linter.enable(msgid)
             except exceptions.UnknownMessageError:
-                # pylint: disable-next=fixme
-                # TODO: Optparse: Raise an informational warning here
-                pass
+                msg = f"{option_string}. Don't recognize message {msgid}."
+                self.linter.add_message("bad-option-value", args=msg, line=0)
 
 
 class _OutputFormatAction(_AccessLinterObjectAction):
--- a/pylint/config/callback_actions.py
+++ b/pylint/config/callback_actions.py
@@ -378,13 +378,13 @@ class _DisableAction(_AccessLinterObjectAction):
     ) -> None:
         assert isinstance(values, (tuple, list))
         msgids = utils._check_csv(values[0])
-        try:
-            for msgid in msgids:
+        for msgid in msgids:
+            try:
                 self.linter.disable(msgid)
-        except exceptions.UnknownMessageError:
-            # pylint: disable-next=fixme
-            # TODO: Optparse: Raise an informational warning here
-            pass
+            except exceptions.UnknownMessageError:
+                # pylint: disable-next=fixme
+                # TODO: Optparse: Raise an informational warning here
+                pass
 
 
 class _EnableAction(_AccessLinterObjectAction):
@@ -399,13 +399,13 @@ class _EnableAction(_AccessLinterObjectAction):
     ) -> None:
         assert isinstance(values, (tuple, list))
         msgids = utils._check_csv(values[0])
-        try:
-            for msgid in msgids:
+        for msgid in msgids:
+            try:
                 self.linter.enable(msgid)
-        except exceptions.UnknownMessageError:
-            # pylint: disable-next=fixme
-            # TODO: Optparse: Raise an informational warning here
-            pass
+            except exceptions.UnknownMessageError:
+                # pylint: disable-next=fixme
+                # TODO: Optparse: Raise an informational warning here
+                pass
 
 
 class _OutputFormatAction(_AccessLinterObjectAction):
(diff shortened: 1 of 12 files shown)


pylint/config/callback_actions.py
```diff
@@ -398,13 +398,13 @@ def __call__(
...
+            try:
                 self.linter.enable(msgid)
-        except exceptions.UnknownMessageError:
-            # pylint: disable-next=fixme
-            # TODO: Optparse: Raise an informational warning here
-            pass
+            except exceptions.UnknownMessageError:
+                # pylint: disable-next=fixme
+                # TODO: Optparse: Raise an informational warning here
+                pass
```
Pierre-Sassoulas, line 407: What about doing it right now ? I don't like this silent fail, in fact we had a more damaging silent fail before and we've been lucky that Marc signaled the issue.

DanielNoord: Sure! Drafting this then!

pylint/config/callback_actions.py
```diff
@@ -377,13 +377,13 @@ def __call__(
...
+        for msgid in msgids:
+            try:
                 self.linter.disable(msgid)
-        except exceptions.UnknownMessageError:
-            # pylint: disable-next=fixme
-            # TODO: Optparse: Raise an informational warning here
-            pass
+            except exceptions.UnknownMessageError:
+                # pylint: disable-next=fixme
+                # TODO: Optparse: Raise an informational warning here
```
cdce8p, line 385: > @cdce8p This was caused because in the homeassistant config there is a message that has been removed. This caused us to stop parsing the disable key when we reached it.

Didn't know that. I don't usually check all disable message ids to see if they are still valid.
A warning for it (not just informational) would be nice.

tests/config/functional/toml/unknown_msgid/enable_unknown_msgid.result.json
```diff
@@ -1,6 +1,7 @@
 {
   "functional_append": {
-    "enable": ["locally-disabled"]
+    "enable": ["locally-disabled"],
+    "disable": ["logging-format-interpolation"]
```
Pierre-Sassoulas, line 4: Oups, my bad 😅

cdce8p (commented):
The PR does resolve the issue!

pylint/config/config_initialization.py
```diff
@@ -58,10 +58,18 @@ def _config_initialization(
...
+    # to allow raising messages on it
+    linter.set_current_module("Command line")
+
     # Now we parse any options from the command line, so they can override
     # the configuration file
     parsed_args_list = linter._parse_command_line_configuration(args_list)
 
+    # Set the current module to configuration as we don't know where
+    # the --load-plugins key is coming from
+    linter.set_current_module("Configuration")
```
Pierre-Sassoulas, line 71:
```suggestion
    linter.set_current_module(f"Command line argument or {config_file}")
```

We can deduce the configuration file though.

tests/config/functional/toml/unknown_msgid/enable_unknown_msgid.2.out
```diff
@@ -0,0 +1,3 @@
+************* Module {abspath}
```
Pierre-Sassoulas, line 1: This kind of silent fail are a pain to deal with, glad it's fixed !

Pierre-Sassoulas (changes requested):
Looks pretty good !

pylint/config/config_initialization.py
DanielNoord: I think that might be a bit too long.

You would get 
```console
************* Module Command line argument or ~/Users/Daniel/Documents/MyHobbyProjects/MyProjectForNow/.pylintrc
Command line argument or ~/Users/Daniel/Documents/MyHobbyProjects/MyProjectForNow/.pylintrc:1:0: E0013: Plugin 'pylint_websockets' is impossible to load, is it installed ? ('No module named 'pylint_websockets'') (bad-plugin-value)
```

Do you have another, potentially shorter suggestion?

Pierre-Sassoulas: I think it might have been relpath before ? (probably the reason why we have abspath and relpath in the expected result)
```suggestion
   linter.set_current_module(f"Command line argument or {config_file.relative_to(base_dir)}")
```

DanielNoord: Where are you getting `base_dir` from?

Pierre-Sassoulas: That's the trick, nowhere 😄 I dont remember how I did it previously. Seeing https://github.com/PyCQA/pylint/blob/0fe866b8cc033334d4c438bcde92a324617ecd2b/pylint/testutils/configuration_test.py#L96 it was taken into account but it was probably handled in a pylint.config class on the code side.

DanielNoord: Yeah I know, but I don't really like that output now that I revisited this. Perhaps `Command line or configuration`?

Pierre-Sassoulas: Look like it was a constant in the package originally : https://github.com/PyCQA/pylint/commit/c62738b66bcbe7220624b04192e9cfe6820ed13a#diff-5c78c16e096b746bf02709ba80aa1f9133f7bcd7b7360f042b14766ce7aaf881R15
I wonder how well it works when not developing locally, i.e. if pylint is installed in a venv or somewhere far from the directory the user is in.

Pierre-Sassoulas:
```suggestion
    linter.set_current_module("Command line or configuration file")
```

Pierre-Sassoulas (approved):

DanielNoord:
Thanks for the clean up @Pierre-Sassoulas

PR #6316, opened 2022-04-14
Replace .namespace with .config

DanielNoord:

## Type of Changes

## Description

@Pierre-Sassoulas Sorry for the size. Just thought I would rip off the band-aid at once instead of in 5 PRs. Should be fairly easy to review. I did a first check already and everything passes. I'm only going to do one double check of all tests if we don't change:

```python
self.namespace.jobs = 1
self.config.jobs = 1
```
to
```python
self.config.jobs = 1
self.config.jobs = 1
```

### DIFF
--- a/pylint/checkers/base/name_checker/checker.py
+++ b/pylint/checkers/base/name_checker/checker.py
@@ -272,7 +272,7 @@ class NameChecker(_BasicChecker):
 
     def open(self):
         self.linter.stats.reset_bad_names()
-        for group in self.linter.namespace.name_group:
+        for group in self.linter.config.name_group:
             for name_type in group.split(":"):
                 self._name_group[name_type] = f"group_{group}"
 
@@ -280,10 +280,10 @@ class NameChecker(_BasicChecker):
         self._name_regexps = regexps
         self._name_hints = hints
         self._good_names_rgxs_compiled = [
-            re.compile(rgxp) for rgxp in self.linter.namespace.good_names_rgxs
+            re.compile(rgxp) for rgxp in self.linter.config.good_names_rgxs
         ]
         self._bad_names_rgxs_compiled = [
-            re.compile(rgxp) for rgxp in self.linter.namespace.bad_names_rgxs
+            re.compile(rgxp) for rgxp in self.linter.config.bad_names_rgxs
         ]
 
     def _create_naming_rules(self) -> tuple[dict[str, Pattern[str]], dict[str, str]]:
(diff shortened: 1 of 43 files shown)


tests/config/test_argparse_config.py
```diff
@@ -25,8 +25,8 @@ class TestArgumentsManager:
     def test_namespace_creation(self) -> None:
         """Test that the linter object has a namespace attribute and that it is not empty"""
 
-        assert self.base_run.linter.namespace
-        assert self.base_run.linter.namespace._get_kwargs()
+        assert self.base_run.linter.config
```
DanielNoord, line 28:
```suggestion
        assert self.base_run.linter.config == self.base_run.linter.namespace
```

pylint/checkers/base/name_checker/checker.py
```diff
@@ -303,9 +303,7 @@ def _create_naming_rules(self) -> tuple[dict[str, Pattern[str]], dict[str, str]]
                 regexps[name_type] = DEFAULT_PATTERNS[name_type]
 
             custom_regex_setting_name = f"{name_type}_rgx"
-            custom_regex = getattr(
-                self.linter.namespace, custom_regex_setting_name, None
-            )
+            custom_regex = getattr(self.linter.config, custom_regex_setting_name, None)
```
Pierre-Sassoulas, line 306: For this one I'm wondering if an hard coded value would not be better/clearer than using getattr(). It's not everyday that we add a new regex setting name. On the other hand it does permit to have a setting for a custom name pattern created by a plugin...

Pierre-Sassoulas (commented):
LGTM, I did not check that there was no more namespace used anywhere but I trust you to properly search and replace :) What about removing the namespace property  in the base class too ?

pylint/checkers/base/name_checker/checker.py
DanielNoord: Yeah, I think the current system is fine. We would just copy 10 relatively similar names.

DanielNoord:
> LGTM, I did not check that there was no more namespace used anywhere but I trust you to properly search and replace :) What about removing the namespace property in the base class too ?

Yeah I used VS Code to do it and went over it twice manually.

I'll do a follow-up in which I'll also check the documentation and see if I missed any reference that I already started updating previously.

Pierre-Sassoulas:
What do you think about removing the property for ``namespace`` altogether and only keeping ``config`` ?

DanielNoord:
> What do you think about removing the property for `namespace` altogether and only keeping `config` ?

THat's the plan! Just trying to separate tasks in PRs somewhat to reduce their size (although I miserably failed here...)

Pierre-Sassoulas:
No it's ok I think as long as it's on a single topic, it does not matter that there is a lot of code in the diff. Problem start to arise when two issues are fixed in the same MR and then we don't know which one brought a problem and can't revert one without also reverting the other.

Pierre-Sassoulas (approved):

PR #6318, opened 2022-04-14
Don't incorrectly suggest ``re.Pattern`` for ``consider-using-alias``

DanielNoord:

## Type of Changes

## Description

This is causing issues for ourselves and blocking us from upgrading our ``py-version``.

@cdce8p, I'm basing myself on https://docs.python.org/3/library/typing.html#typing.Match. I believe generic forms are only supported from 3.9+. Am I understanding this correctly?


Pierre-Sassoulas (approved):

cdce8p (changes requested):

DanielNoord:
I was indeed mistake. I forgot an `__future__` import and then misinterpreted the documentation. Sorry for the churn!

PR #6321, opened 2022-04-14
docs: make it clearer that only additional __slots__ should be specified

robin-wayve:
## Type of Changes

## Description

I found the "good" example for `redefined-slots-in-subclass` confusing since it added a new slot `"c"` instead of merely removing `"a"` which is the problem with the "bad" example. As per Python docs, subclasses "should only contain names of any _additional_ slots" https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots

### DIFF
--- a/doc/data/messages/r/redefined-slots-in-subclass/good.py
+++ b/doc/data/messages/r/redefined-slots-in-subclass/good.py
@@ -3,4 +3,4 @@ class Base:
 
 
 class Subclass(Base):
-    __slots__ = ("c", "d")
+    __slots__ = ("d",)


Pierre-Sassoulas (approved):
Thank you for making the doc better ! I tagged that 2.14 for issues triaging purposes but as soon as it's merged it's going to be live.

DanielNoord:
> pypy.org has a temporary outage and we can't merge, maybe we could make the check a warning @DanielNoord ?

Not sure whether that is possible. I'll need to check that!

DanielNoord:
> pypy.org has a temporary outage and we can't merge, maybe we could make the check a warning @DanielNoord ?

The issue with making it a warning is that we'll likely never check it again. I'd say let's keep it an error and just accept that some times a check fails. The rest of our CI isn't error-prone as well 😄

PR #6325, opened 2022-04-14
Add some manual typing changes

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Ref https://github.com/PyCQA/pylint/issues/4683.

PR #6326, opened 2022-04-14
Bump py-version and add python-typing-update to CI

DanielNoord:

## Type of Changes

## Description

We can tick off two things in #4683 😄

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -121,6 +121,7 @@ jobs:
           . venv/bin/activate
           pip install -e .
           pre-commit run pylint --all-files
+          pre-commit run --hook-stage manual python-typing-update --all-files
 
   spelling:
     name: spelling
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -121,7 +121,6 @@ jobs:
           . venv/bin/activate
           pip install -e .
           pre-commit run pylint --all-files
-          pre-commit run --hook-stage manual python-typing-update --all-files
 
   spelling:
     name: spelling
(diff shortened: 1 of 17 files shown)


.pre-commit-config.yaml
```diff
@@ -101,3 +101,16 @@ repos:
...
+      - id: python-typing-update
+        stages: [manual]
+        additional_dependencies:
+          - "black==22.3.0"
+        args:
+          - --py37-plus
+          - --force
+          - --keep-updates
+          - --black
+        files: ^(pylint)/.+\.py$
```
cdce8p, line 116: I don't think we should add this. The existing pylint checks are much better.

.github/workflows/checks.yaml
```diff
@@ -121,6 +121,7 @@ jobs:
           . venv/bin/activate
           pip install -e .
           pre-commit run pylint --all-files
+          pre-commit run --hook-stage manual python-typing-update --all-files
```
cdce8p, line 124: Same here

cdce8p (changes requested):

.pre-commit-config.yaml
Pierre-Sassoulas, line 116: I agree with what Marc said earlier, that pylint's typing extension is vastly more efficient. I would rather add the typing extension to our pylintrc. I think we should only handle the bump of the py-version in this and discuss this in another one.

cdce8p: > I would rather add the typing extension to our pylintrc.

It's already enabled. We just need to update `py-version`.

> I think we should only handle the bump of the py-version in this MR and discuss the pre-commit hook in another one.

AFAICT the typing changes here are from #6325.
With the typing extension, we don't need to add any other pre-commit hook 😄

DanielNoord, line 116:
```suggestion
```

.github/workflows/checks.yaml
DanielNoord, line 124:
```suggestion
```

DanielNoord:
Changed to only the `py-version` change 👍

cdce8p (approved):

PR #6333, opened 2022-04-15
Deprecate ``get_global_option``

DanielNoord:

## Type of Changes

### DIFF
--- a/pylint/utils/utils.py
+++ b/pylint/utils/utils.py
@@ -247,36 +247,21 @@ def get_global_option(
 def get_global_option(
     checker: BaseChecker,
     option: GLOBAL_OPTION_NAMES,
-    default: T_GlobalOptionReturnTypes | None = None,
+    default: T_GlobalOptionReturnTypes | None = None,  # pylint: disable=unused-argument
 ) -> T_GlobalOptionReturnTypes | None:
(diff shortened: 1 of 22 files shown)


pylint/utils/utils.py
```diff
@@ -247,36 +247,21 @@ def get_global_option(
 def get_global_option(
     checker: BaseChecker,
     option: GLOBAL_OPTION_NAMES,
-    default: T_GlobalOptionReturnTypes | None = None,
+    default: T_GlobalOptionReturnTypes | None = None,  # pylint: disable=unused-argument
 ) -> T_GlobalOptionReturnTypes | None:
```
Pierre-Sassoulas, line 251:
```suggestion
) -> T_GlobalOptionReturnTypes | None:  # pragma: no cover
```
Should we add a ``pragma: no cover`` ?

Pierre-Sassoulas (approved):
That's great, it's a lot more instinctive this way !

pylint/utils/utils.py
DanielNoord: It's covered by the deprecation tests. I added it to all the other methods because the deprecation tests only test the first few lines of those methods and all the different `ifs` within them didn't get covered. Seemed easier to just exclude them.

PR #6335, opened 2022-04-15
Regression tests for setup.cfg's options using the wrong section failing silently

### LINKED ISSUE #4272: load-plugins not working if configured in setup.cfg [pylint.MASTER]
load_plugins option does not work if defined in [pylint.MASTER] section of setup.cfg

### Steps to reproduce

Given a config file `setup.cfg`
```
[pylint.MASTER]
load_plugins=pylint.extensions.bad_builtin
[pylint.DEPRECATED_BUILTINS]
bad-functions=print
```
and a python file `a.py`
```python
print("Hello World!")
```

### Current behavior
Result of `pylint --rcfile=setup.cfg a.py
```
************* Module a

Your code has been rated at 0.00/10
```

### Expected behavior
```
************* Module a
a.py:1:0: W0141: Used builtin function 'print' (bad-builtin)

Your code has been rated at -10.00/10
```
### pylint --version output

```
pylint 2.7.4
astroid 2.5.2
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0]
```

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #4272, ~~work in progress as it's not raising anything right now~~. Maybe we don't care about setup.cfg ? Maybe it's easy to use the new mechanism from toml ?


tests/config/functional/setup_cfg/issue_4272/option_in_wrong_section.cfg
```diff
@@ -0,0 +1,6 @@
+# Test for the behavior of https://github.com/PyCQA/pylint/issues/4272
+[pylint.MASTER]
+load_plugins=pylint.extensions.bad_builtin
+persistent=no
+[pylint.DEPRECATED_BUILTINS]
+bad-functions=print,map
```
DanielNoord, line 6: Why is bad-functions unrecognised? It's defined in `deprecated_builtins`?

Pierre-Sassoulas: Hmm, I think you detected a problem in our own pylintrc, because it's in this section for our pylintrc but it's an option of the ``BadBuiltinChecker``. I wonder why it's not raising an error then.

DanielNoord: Shouldn't it be in the `.json` output?

Pierre-Sassoulas: Yes, but the value for ``bad-functions`` is not loaded and the test fails. I'm investigating right now.

Pierre-Sassoulas: The baseline for the default configuration without change do not load any plugins, so no extensions and no extensions' options. There's an easy fix (loading all plugin when creating the default conf) but it prevents testing the load-plugin option. I think I'm going to permit to define a baseline configuration alongside existing files.

DanielNoord: Hmm, what if we iterate over the updated config keys instead of over the default config? Wouldn't we see any new keys introduced?

Pierre-Sassoulas: We create the expected values from the default config we create in the tests modified with result.json. The issue we have is when creating the expected result for configuration where some non-default plugin should be loaded. If we iterate over the updated config how would we know what is the expected value ?

DanielNoord: Currently we do `default config - final config == expected json` right? Can't we do `final config - default config == expected json`? If there is a key in `final` that is not in `default` it should be completely equal to the value in the `json`. That way we can handle unexpected keys right?

Pierre-Sassoulas: Right, I think it could work, let me try.

DanielNoord:
@Pierre-Sassoulas  Since `argparse` we nog longer care about the section. In your original attempt you had some `-` and `_` switched around. This new test passes.

With https://github.com/PyCQA/pylint/pull/6356 we should probably revisit how we parse `setup.cfg` for sections as `[MASTER]` instead of `[pylint.MASTER]`, but I'll do that after!

Pierre-Sassoulas (commented):
The argparse migration really did simplify everything !

PR #6337, opened 2022-04-15
Several ``argparse`` related refactors

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,6 +37,11 @@ Release date: TBA
 
   Ref #5392
 
+* The ``option_groups`` attribute of ``PyLinter`` has been deprecated.
+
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (approved):
👍 There's so much pending deprecation that it makes me thing that we should start thinking about what should be or not be in 3.0 so we end up releasing it at some point 😄

DanielNoord:
> 👍 There's so much pending deprecation that it makes me thing that we should start thinking about what should be or not be in 3.0 so we end up releasing it at some point 😄

Hopefully we don't need to touch most of this code in the coming months so we don't get annoyed by it too much.

PR #6338, opened 2022-04-15
Deprecate all old classes from ``optparse``

DanielNoord:

## Type of Changes

## Description

🎉 

Bit premature because `pyreverse` is still using these, but we might as well deprecate them.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -47,6 +47,30 @@ Release date: TBA
 
   Ref #5392
 
+* ``UnsupportedAction`` has been deprecated.
+
+  Ref #5392
+
+* ``OptionsManagerMixIn`` has been deprecated.
+
+  Ref #5392
+
+* ``OptionParser`` has been deprecated.
+
+  Ref #5392
+
+* ``Option`` has been deprecated.
+
+  Ref #5392
+
+* ``OptionsProviderMixIn`` has been deprecated.
+
+  Ref #5392
+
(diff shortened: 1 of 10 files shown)


ChangeLog
```diff
@@ -47,6 +47,30 @@ Release date: TBA
 
   Ref #5392
 
+* ``UnsupportedAction`` has been deprecated.
```
Pierre-Sassoulas, line 50: Should we specify that it's because of the optparse migration and what the replacement is ?

doc/whatsnew/2.14.rst
```diff
@@ -97,6 +97,30 @@ Other Changes
 
   Ref #5392
 
+* ``UnsupportedAction`` has been deprecated.
+
+  Ref #5392
+
+* ``OptionsManagerMixIn`` has been deprecated.
```
Pierre-Sassoulas, line 104: This is highly satisfactory. I'm not going to comment on each mixin but I really hated those classes 😄

DanielNoord: Me too! `mypy` despises them too!

ChangeLog
DanielNoord: Uhm, there is no replacement for most of these. We mention `ArgumentsProvider` and `ArgumentsManager` in another line already. So I think this is fine as is.

Pierre-Sassoulas (approved):

PR #6342, opened 2022-04-15
Added missing-return-doc message example

matusvalo:
## Type of Changes

## Description

Related #5953

### DIFF
--- a/doc/data/messages/m/missing-return-doc/good.py
+++ b/doc/data/messages/m/missing-return-doc/good.py
@@ -1,10 +1,7 @@
-def integer_sum(a, b):
+def integer_sum(a: int, b: int) -> int:
     """Returns sum of two integers
     :param a: first integer
-    :type a: int
     :param b: second integer
-    :type b: int
     :return: sum of parameters a and b
-    :rtype: int
     """
     return a + b
--- a/doc/data/messages/m/missing-return-doc/good.py
+++ b/doc/data/messages/m/missing-return-doc/good.py
@@ -2,6 +2,8 @@ def integer_sum(a, b):
     """Returns sum of two integers
     :param a: first integer
     :type a: int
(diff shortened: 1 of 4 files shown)


doc/data/messages/m/missing-return-doc/good.py
```diff
@@ -0,0 +1,10 @@
+def integer_sum(a, b):
+    """Returns sum of two integers
+    :param a: first integer
+    :type a: int
+    :param b: second integer
+    :type b: int
+    :return: sum of parameters a and b
+    :rtype: int
```
Pierre-Sassoulas, line 8:
```suggestion
def integer_sum(a, b) -> int:
    """Returns sum of two integers
    :param a: first integer
    :type a: int
    :param b: second integer
    :type b: int
    :return: sum of parameters a and b
```
I'm wondering if this does not make more sense to use the builtin typing directly.

matusvalo: > I'm wondering if this does not make more sense to use the builtin typing directly.

I stick closely to checker documentation https://pylint.pycqa.org/en/latest/technical_reference/extensions.html#pylint-extensions-docparams so I used comments. But I can rework it to annotation types...

matusvalo: I suppose than also other examples should be reworked to use python typing right?

Pierre-Sassoulas: I think this checker doc was done a long time ago (when 'real' typing was not available yet) possibly with python 2 compatibility in mind. The other example would need to be modified too but let's hear what @DanielNoord think before changing everything :)

matusvalo: I will update them no worries

PR #1337, opened 2017-02-21
Optimize effect of disabling import warnings

erikwright:
Fixes https://github.com/PyCQA/pylint/issues/1336 on `master`.

PR #6347, opened 2022-04-15
Avoid reporting useless-suppression on wrong-import-position (#5219)

### LINKED ISSUE #1366: Create entry point console scripts pylint2 and pylint3
### Steps to reproduce

Install both pylint2 and pylint3.

```
sudo pip2 install pylint
sudo pip3 install pylint
```

Notice that they overlap or share the same entry point script which may be /usr/bin/pylint, /usr/local/bin/pylint, or even one of each.

This sole entry point is now ambiguous and we can't which version of pylint to run. Often we are working on a large project that uses both Python 2 and Python 3, or in several projects that use different versions of Python.

The difficulty is apparent here:
http://askubuntu.com/questions/340940/installing-pylint-for-python3-on-ubuntu/891037#891037

### Current behavior

```
pylint p2file.py
(many errors trying to treat p2file.py as Python3)
```

One unsatisfactory hack to get around this is run the console script in a peculiar way:

```
python2 /usr/local/bin/pylint p2file.py
python3 /usr/local/bin/pylint p3file.py
```

### Expected behavior

When pylint is installed using pip2, setup should create an entry_point console_script
/usr/local/bin/pylint2

and when pylint is installed using pip3, setup should create an entry_point console_script
/usr/local/bin/pylint3

This parallels the way one picks which version of Python to use:
```
/usr/bin/python2
/usr/bin/python3
```

Then picking a specific version of pylint to run is intuitive:

```
pylint2 /usr/local/bin/pylint p2file.py
pylint3 /usr/local/bin/pylint p3file.py
```

### LINKED ISSUE #5219: useless-suppression false positive with wrong-import-position
### Bug description

```python
# pylint: disable=wrong-import-position
"""Test."""
import time

satrt = time.time()

import os

_ = os
```

### Configuration

_No response_

### Command used

```shell
pylint a.py --enable=useless-suppression
```

### Pylint output

This message is issued:
```shell
$ pylint a.py --enable=useless-suppression
************* Module a
a.py:1:0: I0021: Useless suppression of 'wrong-import-position' (useless-suppression)

```

However, if the comment is removed from line 1, this message is issued:
```shell
$ pylint a.py --enable=useless-suppression
************* Module a
a.py:6:0: C0413: Import "import os" should be placed at the top of the module (wrong-import-position)

Your code has been rated at 7.50/10 (previous run: 10.00/10, -2.50)
```

### Expected behavior

expect get no useless-suppression message

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.4
Python 3.7.9 (default, Aug 31 2020, 07:22:35)
[Clang 10.0.0 ]
```

### OS / Environment

macOS

### Additional dependencies

_No response_

timmartin:
## Type of Changes

## Description

With a suppressed `wrong-import-position` warning and `useless-suppressions` enabled, Pylint was issuing a `useless-suppression` warning even though the suppression was not useless. This was happening because of a change [here](https://github.com/PyCQA/pylint/commit/4f5d2279ca3694ea363ac5a1df54ecbdba2b612d) that fixed #1366; It was skipping the entire calculation of the checker when the warning was disabled, which prevents it from recording that the checker would have been hit. There's [a similar fix here](https://github.com/PyCQA/pylint/pull/5063/files) for the same case with `wrong-import-order`.

Closes #5219


jacobtylerwalls:
Thanks for the PR. What do you think about this diff? A bit smaller.

```diff
diff --git a/pylint/checkers/imports.py b/pylint/checkers/imports.py
index d90173c0..4fa01e02 100644
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -503,6 +503,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
         self._imports_stack = []
         self._first_non_import_node = None
 
+    @check_messages("wrong-import-position")
     def compute_first_non_import_node(self, node):
         # if the node does not contain an import instruction, and if it is the
         # first node of the module, keep a track of it (all the import positions
@@ -543,6 +544,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
         visit_ifexp
     ) = visit_comprehension = visit_expr = visit_if = compute_first_non_import_node
 
+    @check_messages("wrong-import-position")
     def visit_functiondef(self, node: nodes.FunctionDef) -> None:
         # If it is the first non import instruction of the module, record it.
         if self._first_non_import_node:
@@ -594,16 +596,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
         # if a first non-import instruction has already been encountered,
         # it means the import comes after it and therefore is not well placed
         if self._first_non_import_node:
-            if self.linter.is_message_enabled(
-                "wrong-import-position", self._first_non_import_node.fromlineno
-            ):
-                self.add_message(
-                    "wrong-import-position", node=node, args=node.as_string()
-                )
-            else:
-                self.linter.add_ignored_message(
-                    "wrong-import-position", node.fromlineno, node
-                )
+            self.add_message("wrong-import-position", node=node, args=node.as_string())
 
     def _record_import(self, node, importedmodnode):
         """Record the package `node` imports from."""
```

Pierre-Sassoulas:
``@check_messages(*MSGS)`` has the same effect than no decorator at all so I think we could also remove this, see #6353. What I wonder is what is the purpose of ``self.linter.add_ignored_message`` and `` self.linter.is_message_enabled`` in general.

jacobtylerwalls (approved):
By the way, the test that failed with my smaller diff was the test in #1337. (Yes, 1337! 🙄 )

So I'm good with this approach, meaning we can still support that somewhat special case.

DanielNoord:
> By the way, the test that failed with my smaller diff was the test in #1337. (Yes, 1337! 🙄 )
> 
> So I'm good with this approach, meaning we can still support that somewhat special case.

Do we understand why that test fails? It seems like it should work right?

jacobtylerwalls:
> Do we understand why that test fails? It seems like it should work right?

My understanding is that it's a really special case to extend the disable functionality. `wrong-import-position` comes in pairs (it's always wrong on two lines). So these methods have special functionality to work all that out. Do I understand exactly how it works? Nope.

timmartin:
I don't have much to add, except to say that I was trying to change the logic of the check as little as possible while avoiding the `useless-suppression` warning. It seemed plausible that these checks would have to work slightly differently because they are about a relation between two lines and not an error on a single line. But I didn't dig any deeper than that.

If anyone thinks this needs a different approach I can take another run at it.

PR #6348, opened 2022-04-15
Fix type annotation for numbersion tuple

tusharsadhwani:
## Type of Change


pylint/__pkginfo__.py
```diff
@@ -7,7 +7,7 @@
 __version__ = "2.14.0-dev0"
 
 
-def get_numversion_from_version(v: str) -> tuple:
+def get_numversion_from_version(v: str) -> tuple[int, ...]:
```
Pierre-Sassoulas, line 10:
```suggestion
def get_numversion_from_version(v: str) -> tuple[int, int, int]:
```
Because of line 18 and line 32 / 33 we're guaranteed to have a 3 int tuple.

Pierre-Sassoulas (changes requested):

pylint/__pkginfo__.py
tusharsadhwani: A type checker can't infer that. We'll either have to add a `typing.cast` or leave it as a `tuple[int, ...]`.

I think the way it is is good enough. Does knowing that it'll always be 3 values in the tuple particularly useful?

Pierre-Sassoulas: We're maintaining this function for legacy reason, this information can be accessed with importlib. But yeah if mypy can't infer let's revert my suggestion

tusharsadhwani:
```console
$ mypy --strict pylint/__pkginfo__.py
pylint/__pkginfo__.py:34: error: Incompatible return value type (got "Tuple[int, ...]", expected "Tuple[int, int, int]")
Found 1 error in 1 file (checked 1 source file)
```

pylint/__pkginfo__.py
```diff
@@ -7,7 +7,7 @@
 __version__ = "2.14.0-dev0"
 
 
-def get_numversion_from_version(v: str) -> tuple:
+def get_numversion_from_version(v: str) -> tuple[int, int, int]:
```
Pierre-Sassoulas, line 10:
```suggestion
def get_numversion_from_version(v: str) -> Tuple[int, ...]:
```

pylint/__pkginfo__.py
```diff
@@ -7,7 +7,7 @@
 __version__ = "2.14.0-dev0"
 
 
-def get_numversion_from_version(v: str) -> tuple:
+def get_numversion_from_version(v: str) -> Tuple[int, ...]:
```
Pierre-Sassoulas, line 10:
```suggestion
def get_numversion_from_version(v: str) -> tuple[int, ...]:
```

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Thank you for opening the MR @tusharsadhwani, the typing will get better a we'll be able to use ``mypy --strict`` only with a lot of small step like this.

pylint/__pkginfo__.py
DanielNoord: Sorry to comment after merging, but I think we should do the suggested typing and a disable for `mypy`.
`while len(version) != 3:` makes it very clear that it should always be `len() == 3` as we would never exit as it was over 3.
This is an issue with `mypy` (which likely will never get fixed) but shouldn't be avoided by adding incomplete typing imo.

PR #6353, opened 2022-04-16
Remove pointless ``check_messages`` decorators

DudeNr33:

## Type of Changes

## Description

As discussed in #6060, ``@check_messages`` has no effect in the following cases:

1. Decorating a method that does not start with ``visit_*`` or ``leave_*``: https://github.com/PyCQA/pylint/blob/ebf7bd6793dc484513455d0c47acd6a285a8d1a5/pylint/utils/ast_walker.py#L35-L46)
2. Passing in all messages this checker defines with ``@check_messages(*MSGS)``. This has the same effect as not decorating it at all:
https://github.com/PyCQA/pylint/blob/ebf7bd6793dc484513455d0c47acd6a285a8d1a5/pylint/utils/ast_walker.py#L20-L23
If none of the checker's messages is enabled, we won't load it anyway: 
https://github.com/PyCQA/pylint/blob/ebf7bd6793dc484513455d0c47acd6a285a8d1a5/pylint/lint/pylinter.py#L987-L997

### DIFF
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1294,7 +1294,6 @@ accessed. Python regular expressions are accepted.",
             self.add_message("isinstance-second-argument-not-valid-type", node=node)
 
     # pylint: disable=too-many-branches,too-many-locals
-    @check_messages(*(list(MSGS.keys())))
     def visit_call(self, node: nodes.Call) -> None:
         """Check that called functions/methods are inferred to callable objects,
         and that passed arguments match the parameters in the inferred function.
@@ -1831,12 +1830,10 @@ accessed. Python regular expressions are accepted.",
                 self.add_message("unsupported-binary-operation", args=msg, node=node)
                 break
 
-    @check_messages("unsupported-binary-operation")
     def _visit_binop(self, node: nodes.BinOp) -> None:
         """Detect TypeErrors for binary arithmetic operands."""
         self._check_binop_errors(node)
 
-    @check_messages("unsupported-binary-operation")
     def _visit_augassign(self, node: nodes.AugAssign) -> None:
         """Detect TypeErrors for augmented binary arithmetic operands."""
         self._check_binop_errors(node)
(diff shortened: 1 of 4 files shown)


pylint/checkers/typecheck.py
```diff
@@ -1294,7 +1294,6 @@ def _check_isinstance_args(self, node):
             self.add_message("isinstance-second-argument-not-valid-type", node=node)
 
     # pylint: disable=too-many-branches,too-many-locals
-    @check_messages(*(list(MSGS.keys())))
```
DanielNoord, line 1297: 😅

pylint/checkers/typecheck.py
```diff
@@ -1831,12 +1830,10 @@ def _check_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> None
                 self.add_message("unsupported-binary-operation", args=msg, node=node)
                 break
 
-    @check_messages("unsupported-binary-operation")
     def _visit_binop(self, node: nodes.BinOp) -> None:
```
DanielNoord, line 1833: Do you understand why these are underscored? They don't seem to be called in this file..

Pierre-Sassoulas (approved):

pylint/checkers/typecheck.py
DudeNr33: I saw that the functional test for this message (``unsupported-binary-operation``) is always skipped, with the following comment:
```
Disable for unsupported-binary-operation

Unfortunately, this warning exhibits currently way too many false
positives, practically rendering this error useless. I can't fix
all of them until I release Pylint 1.5, so the most reasonable
choice for now is to disable it and reenable it as soon as we fix
those problems.
```

DanielNoord: Oh wow, we should probably follow up on that..

But then the decorator is "correct" we only need to de-underscore this function right? Perhaps we should keep the decorator and add a TODO to fix this?

DudeNr33: You are right - if we plan to re-enable this check, the decorator should stay. Will fix this and add a TODO.

PR #6356, opened 2022-04-16
Fix the loading of unrelated tools options

### LINKED ISSUE #4371: Every option in setup.cfg is loaded, even when the section is unknown to pylint
This is a slightly different issue from what @mteiste has seen with #4272 . My issue is with all sections and options defined in `setup.cfg` are processed/loaded, even if those sections are for other tools and unknown to `pylint`.

This has the effect of setting pylint's ignore to `D107,D400,D401`.

Going through the code, it feels like the problem is somewhere in [option_manager_mixin.py](https://github.com/PyCQA/pylint/blob/c7ac7c0ec339e9a8d105ccac0e59e65cacb79272/pylint/config/option_manager_mixin.py), either at `read_config_line` line [298:299](https://github.com/PyCQA/pylint/blob/c7ac7c0ec339e9a8d105ccac0e59e65cacb79272/pylint/config/option_manager_mixin.py#L298:L299) or at `load_config_file` line [312](https://github.com/PyCQA/pylint/blob/c7ac7c0ec339e9a8d105ccac0e59e65cacb79272/pylint/config/option_manager_mixin.py#L312).

Apparently, Toml files are not affected because the logic is set to read from `content["tool"]["pylint"]`.  But for `setup.cfg` there's no `try / except` block, iterating through every section:

```
if sect.startswith("pylint."):
	sect = sect[len("pylint.") :]
```

I think `pylint` should filtered / excluded any section that's not part of any pylint group -> `self._mygroups`.

### Steps to reproduce

Given a config file `setup.cfg`

```
[pylint.master]
ignore = migrations

[flake8]
ignore = D107,D400,D401
```

### Current behavior

Result of `pylint --generate-rcfile | grep -A2 'ignore='`:

```
ignore=D107,
       D400,
       D401
```
-->

### Expected behavior

Result of `pylint --generate-rcfile | grep -A2 'ignore='`:

```
ignore=migrations
```

### pylint --version output

```
pylint 2.7.4
astroid 2.5.3
Python 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
```

### Workaround

I had to monkey patch https://github.com/PyCQA/pylint/blob/master/pylint/config/option_manager_mixin.py#L308 with
```
        if section not in self._mygroups:
            continue
```

Resulting in:

```
def load_config_file(self):
    """Dispatch values previously read from a configuration file to each
    options provider)"""
    parser = self.cfgfile_parser
    for section in parser.sections():
        if section not in self._mygroups:
            continue
        for option, value in parser.items(section):
            try:
                self.global_set_option(option, value)
            except (KeyError, optparse.OptionError):
                continue
```

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #4371


DanielNoord:
@Pierre-Sassoulas @jacobtylerwalls 

I chose to start deprecating the use of `[MASTER]` in `setup.cfg`. I don't think we should claim that as it might interfere with other tools and doesn't really follow the standard as described by https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#compatibility-with-other-tools.

However, if we raise a `DeprecationWarning` nobody will ever see this... That's why I opted to print to `stdout` instead. I'm fine with reverting this and making it a warning instead, but I think it will be difficult to get the word out any other way. Let me know what you think!

pylint/config/config_file_parser.py
```diff
@@ -42,6 +42,15 @@ def _parse_ini_file(file_path: Path) -> tuple[dict[str, str], list[str]]:
         config_content: dict[str, str] = {}
         options: list[str] = []
         for section in parser.sections():
+            if "setup.cfg" in str(file_path) and not section.startswith("pylint"):
```
Pierre-Sassoulas, line 45: What if we use the ``--rcfile`` option and the file is not named "setup.cfg" ?

Pierre-Sassoulas (commented):
Thank you for handling this @DanielNoord ! I think that if we use the term "deprecate" for "master" and we replace it by "pylint.master" it's going to bring some controversy (we went to the trouble of deprecating something and we did not replace by pylint.main). Better have some message saying "you need to use pylint.master instead of master to disstinguish pylint from other tool" imo. Not sure how easy it is to have an alias for 'main' , do you have an idea ?

DanielNoord:
> Thank you for handling this @DanielNoord ! I think that if we use the term "deprecate" for "master" and we replace it by "pylint.master" it's going to bring some controversy (we went to the trouble of deprecating something and we did not replace by pylint.main). Better have some message saying "you need to use pylint.master instead of master to disstinguish pylint from other tool" imo. Not sure how easy it is to have an alias for 'main' , do you have an idea ?

We don't care about section names anymore 😄 

As long as they start with `pylint.` we are fine. We could even suggest `pylint.TheMostImportantSection`? 

I'll update the code to say `pylint.main`.

pylint/config/config_file_parser.py
DanielNoord: Then we load every section like we do now. I chose to do this because if people are going to the trouble of defining an `ini` file and pointing to it it is very likely that this is indeed a `pylintrc` or other `pylint` specific file. I don't think we want to enforce `pylint.main` in `pylintrc` itself.

Pierre-Sassoulas: > I don't think we want to enforce pylint.main in pylintrc itself.

Yeah not unless we have a migration tool ready.

Pierre-Sassoulas: Hmm wait, how do we know that this section was supposed to be related to pylint and not flake8 ?

DanielNoord: It is not going to raise a warning for an empty `setup.cfg` but it is for any `[MASTER]` or `[master]` section in a file called `setup.cfg`. I think that is the only way to even raise a warning.

Pierre-Sassoulas: That would be nice, but the test is ``and not section.startswith("pylint"):`` not ``"master" not in section.lower()`` ?

DanielNoord: We `continue` for `else`? We emit the warning for `master` and then continue parsing it. For all other sections that don't start with `pylint` we simply `continue`.

So `[MESSAGES CONTROL]` and `[flake8.Master]` won't be read any more.

Pierre-Sassoulas: Ha right I missed that, my bad.

pylint/config/config_file_parser.py
```diff
@@ -42,6 +42,15 @@ def _parse_ini_file(file_path: Path) -> tuple[dict[str, str], list[str]]:
...
+                if section.lower() == "master":
+                    # TODO: 3.0: Remove deprecated handling of master, only allow 'pylint.' sections
+                    print(
+                        "WARNING: The use of 'MASTER' or 'master' as configuration section for pylint "
+                        "has been deprecated. Please use 'pylint.main' instead.",
```
Pierre-Sassoulas, line 50:
```suggestion
                        "WARNING: The use of 'MASTER' or 'master' as configuration section for pylint "
                        "has been deprecated, as it's bad practice to not start sections titles with the "
                        "tool name. Please use 'pylint.main' instead.",
```

tests/config/functional/setup_cfg/deprecate_master/setup.0.out
```diff
@@ -0,0 +1 @@
+WARNING: The use of 'MASTER' or 'master' as configuration section for pylint has been deprecated. Please use 'pylint.main' instead.
```
Pierre-Sassoulas, line 1:
```suggestion
WARNING: The use of 'MASTER' or 'master' as configuration section for pylint has been deprecated, as it's bad practice to not start sections titles with the tool name. Please use 'pylint.main' instead.
```

Pierre-Sassoulas (commented):
I can't approve my own MR, thank you for finishing this, much appreciated.

DanielNoord:
Let's wait on @jacobtylerwalls to hear what they think of printing instead of warning.

jacobtylerwalls:
`UserWarning` is an option, right? To let folks filter the warnings idiomatically but also to surface them by default first.

DanielNoord:
> `UserWarning` is an option, right? To let folks filter the warnings idiomatically but also to surface them by default first.

TIL... Changing 😄

tests/config/test_functional_config_loading.py
```diff
@@ -78,9 +79,13 @@ def test_functional_config_loading(
     expected_loaded_configuration = get_expected_configuration(
         configuration_path, default_configuration
     )
-    mock_exit, _, runner = run_using_a_configuration_file(
-        configuration_path, file_to_lint_path
-    )
+    with warnings.catch_warnings():
```
jacobtylerwalls, line 82: does `pytest.warns` both do the ignoring and the asserting? This looks like it just does the ignoring?

jacobtylerwalls: Yeah let's not sink a lot of time into this but was just honestly wondering if it's easy as adding `pytest.warns`, but if not, then not!

DanielNoord: No since `pytest` will complain about all functional tests that do not have `MASTER`. I just checked again manually and your suggestion works as intended. Thanks!

jacobtylerwalls (approved):

PR #4678, opened 2021-07-06
Use absolute import qualified module name for deprecated module check

matusvalo:
## Steps

- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [ ] If it's a new feature or an important bug fix, add a What's New entry in
      `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Description

This PR adjust deprecated module check of from imports to use absolute import name. It helps avoid detecting deprecated
modules in relative imports.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |
|     | :sparkles: New feature |
| ✓   | :hammer: Refactoring   |
|     | :scroll: Docs          |

## Related Issue

Closes #4629

PR #4225, opened 2021-03-10
Move deprecated module checker to DeprecatedMixin

matusvalo:
## Steps

- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [x] Write a good description on what the PR does.

## Description
This PR logic of checking the use of deprecated modules to DeprecatedMixin class.
Hence, it can be easilly reused this way.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :hammer: Refactoring  |

## Related Issue

Closes #4086

PR #4051, opened 2021-01-29
Implement reusable Method Deprecation checker.

matusvalo:
This PR is moving deprecated check logic to separate mixin class `DeprecatedMixin`. This enables plugins to implement their own deprecation checks. The most visible use case can be pylint-django. Here is the example of simple plugin:
```python
"""
Simple DeprecatedChecker marking test.deprecated_func as deprecated.
"""
import astroid

from pylint.checkers import BaseChecker, DeprecatedMixin, utils
from pylint.interfaces import IAstroidChecker


class DeprecatedChecker(BaseChecker, DeprecatedMixin):
    __implements__ = (IAstroidChecker,)
    name = "deprecated"

    msgs = {
        "W1505": (
            "Using deprecated method %s()",
            "deprecated-method",
            "The method is marked as deprecated and will be removed in "
            "a future version of Python. Consider looking for an "
            "alternative in the documentation.",
        )
      }

    @utils.check_messages(
        "deprecated-method",
    )
    def visit_call(self, node):
        """Visit a Call node."""
        try:
            for inferred in node.func.infer():
                self.check_deprecated_method(node, inferred)
        except astroid.InferenceError:
            return

    def deprecated_methods(self):
        return {'test.deprecated_func'}


def register(linter):
    linter.register_checker(DeprecatedChecker(linter))
```

Current desing has multiple design choices which can be discussed:
1. Implementing Deprecated checker requires writing new checker instead of configuring existing one.
2. Using `DeprecatedMixin` requires to define `deprecated-method` msg in Checker class. This can be cumbersome, but it is backwards compatible and it conforms rule that each checker has it's own isolated set of IDs.

Current state of PR is mainly a starting point for final implementation. The following parts needs to be done:
- [x] Implementing/adjusting unittests
- [ ] Documentation

## Steps

- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [ ] If it's a new feature or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Type of Changes

|   | Type |
| ------------- | ------------- |
|    | :bug: Bug fix  |
| ✓  | :sparkles: New feature |
|    | :hammer: Refactoring  |
| ✓  | :scroll: Docs |

## Related Issue
Closes #4049

PR #6360, opened 2022-04-16
Run deprecated checks with `--enable=deprecated*` after `disable=all`

jacobtylerwalls:
## Type of Changes

## Description

Before, the `deprecated-module` etc. checks did not run if they were enabled with `--enable` after a `--disable=all`.

Follow-up to [fabc645](https://github.com/PyCQA/pylint/pull/6246/commits/fabc6453cb8c8ef9b5622677ab554af157cf42bd).

### DIFF
--- a/pylint/checkers/deprecated.py
+++ b/pylint/checkers/deprecated.py
@@ -59,6 +59,11 @@ class DeprecatedMixin(BaseChecker):
         ),
     }
 
+    def check_consistency(self):
+        """This is a mixin, so there is no expectation that the messages should be
+        mutually exclusive. They will overlap between checkers by design."""
+        pass
+
     @utils.check_messages(
         "deprecated-method",
         "deprecated-argument",
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -441,6 +441,7 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             {"old_names": [("W1516", "lru-cache-decorating-method")]},
         ),
     }
+    msgs.update(DeprecatedMixin.msgs)
 
     def __init__(self, linter: PyLinter) -> None:
         BaseChecker.__init__(self, linter)
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,8 @@ Release date: TBA
 
   Closes #5224
 
+* Fixed failure to run deprecated checks with ``--enable=deprecated*`` after a ``disable=all``.
+
 * The ``config`` attribute of ``BaseChecker`` has been deprecated. You can use ``checker.linter.config``
   to access the global configuration object instead of a checker-specific object.
 
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -25,6 +25,8 @@ New checkers
 
   Closes #5895
 
+* Fixed failure to run deprecated checks with ``--enable=deprecated*`` after a ``disable=all``.
+
 * Add new check ``unnecessary-dunder-call`` for unnecessary dunder method calls.
 
   Closes #5936
--- a/tests/functional/d/deprecated/deprecated_module_py3.rc
+++ b/tests/functional/d/deprecated/deprecated_module_py3.rc
@@ -1,5 +1,2 @@
 [typing]
 py-version=3.6
-
-[testoptions]
-exclude_from_minimal_messages_config=true
--- a/tests/functional/d/deprecated/deprecated_module_py310.rc
+++ b/tests/functional/d/deprecated/deprecated_module_py310.rc
@@ -1,3 +1,2 @@
 [testoptions]
 min_pyver = 3.10
-exclude_from_minimal_messages_config=true
--- a/tests/functional/d/deprecated/deprecated_module_py33.rc
+++ b/tests/functional/d/deprecated/deprecated_module_py33.rc
@@ -1,2 +1,0 @@
-[testoptions]
-exclude_from_minimal_messages_config=true
--- a/tests/functional/d/deprecated/deprecated_module_py36.rc
+++ b/tests/functional/d/deprecated/deprecated_module_py36.rc
@@ -1,2 +1,0 @@
-[testoptions]
-exclude_from_minimal_messages_config=true
--- a/tests/functional/d/deprecated/deprecated_module_py39_earlier_pyversion.rc
+++ b/tests/functional/d/deprecated/deprecated_module_py39_earlier_pyversion.rc
@@ -3,4 +3,3 @@ py-version=3.8
 
 [testoptions]
 min_pyver=3.9
-exclude_from_minimal_messages_config=true
--- a/tests/functional/d/deprecated/deprecated_relative_import/dot_relative_import.rc
+++ b/tests/functional/d/deprecated/deprecated_relative_import/dot_relative_import.rc
@@ -1,2 +1,0 @@
-[testoptions]
-exclude_from_minimal_messages_config=true
--- a/tests/functional/d/deprecated/deprecated_relative_import/subpackage/dot_dot_relative_import.rc
+++ b/tests/functional/d/deprecated/deprecated_relative_import/subpackage/dot_dot_relative_import.rc
@@ -1,2 +1,0 @@
-[testoptions]
-exclude_from_minimal_messages_config=true
--- a/tests/message/unittest_message_id_store.py
+++ b/tests/message/unittest_message_id_store.py
@@ -10,6 +10,7 @@ from pathlib import Path
 import pytest
 
 from pylint import lint
+from pylint.checkers.deprecated import DeprecatedMixin
 from pylint.exceptions import InvalidMessageError, UnknownMessageError
 from pylint.message.message_definition import MessageDefinition
 from pylint.message.message_id_store import MessageIdStore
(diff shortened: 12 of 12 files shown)


pylint/checkers/stdlib.py
```diff
@@ -441,6 +441,7 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             {"old_names": [("W1516", "lru-cache-decorating-method")]},
         ),
     }
+    msgs.update(DeprecatedMixin.msgs)
```
Pierre-Sassoulas, line 444: 👍

pylint/checkers/deprecated.py
```diff
@@ -59,6 +59,11 @@ class DeprecatedMixin(BaseChecker):
         ),
     }
 
+    def check_consistency(self):
+        """This is a mixin, so there is no expectation that the messages should be
+        mutually exclusive. They will overlap between checkers by design.
+        """
```
Pierre-Sassoulas, line 65: What's the problem here ? I think the msgid could still be consistent i.e. they should all start by "15" and be valid msgid with a letter and 4 number ?

Pierre-Sassoulas (commented):
Looks great. This is the kind of hard to debug issue that create obscure bug, good job tracking that one.

pylint/checkers/deprecated.py
jacobtylerwalls: Both `StdlibChecker` and `ImportsChecker` emit `W0402`, `W0404`... etc for `deprecated-module` and friends. This raises `InvalidMessageError`:

```
Inconsistent checker part in message id 'W0402' (expected 'x15xx' ...
```
Are they supposed to become W1502 and W1504 in the stdlib checker? That would defeat the point of the Mixin, I guess? I don't know what the backward-compatibility issues would be.

jacobtylerwalls: Well, we have more uncomfortable problems because now we have the docs complaining about two messages. I guess the recent infrastructure assumed no checkers shared messages.
```
/home/runner/work/pylint/pylint/doc/messages/messages_list.rst:185: WARNING: duplicated entry found in toctree: messages/warning/deprecated-module
```

jacobtylerwalls: Oh. The only overlap is W0402 `deprecated-module`. They differ in this way: The ImportsChecker looks at modules given in the config. The StdlibChecker looks at the hard-coded dictionary we maintain.

jacobtylerwalls: Maybe the answer is to just let the `ImportsChecker` check this message and check both sources. Currently if `optparse` is given at the command line, we emit twice, once by each checker:
```
pylint/config/option_manager_mixin.py:11:0: W0402: Uses of a deprecated module 'optparse' (deprecated-module)
pylint/config/option_manager_mixin.py:11:0: W0402: Uses of a deprecated module 'optparse' (deprecated-module)
```

jacobtylerwalls: Decided to go that route in [1e6bdcf]

pylint/checkers/imports.py
```diff
@@ -446,15 +466,15 @@ def visit_import(self, node: nodes.Import) -> None:
...
     def visit_importfrom(self, node: nodes.ImportFrom) -> None:
         """Triggered when a from statement is seen."""
         basename = node.modname
         imported_module = self._get_imported_module(node, basename)
+        absolute_name = get_import_name(node, basename)
 
         self._check_import_as_rename(node)
         self._check_misplaced_future(node)
-        self.check_deprecated_module(node, basename)
+        self.check_deprecated_module(node, absolute_name)
```
jacobtylerwalls, line 477: This change is to avoid these failures from tests in #4678:
```
___________________________ test_functional[dot_relative_import] ____________________________

self = <pylint.testutils.lint_module_test.LintModuleTest object at 0x113693850>

    def runTest(self) -> None:
>       self._runTest()
E       AssertionError: Wrong results for file "dot_relative_import":
E       
E          7: deprecated-module

pylint/testutils/lint_module_test.py:107: AssertionError
_________________________ test_functional[dot_dot_relative_import] __________________________

self = <pylint.testutils.lint_module_test.LintModuleTest object at 0x1138d3cd0>

    def runTest(self) -> None:
>       self._runTest()
E       AssertionError: Wrong results for file "dot_dot_relative_import":
E       
E          7: deprecated-module

pylint/testutils/lint_module_test.py:107: AssertionError
```

pylint/checkers/stdlib.py
```diff
@@ -750,7 +739,7 @@ def _check_invalid_envvar_value(self, node, infer, message, call_arg, allow_none
             self.add_message(message, node=node, args=(name, call_arg.pytype()))
 
     def deprecated_modules(self):
-        """Callback returning the deprecated modules."""
+        """See ImportsChecker: this one is empty."""
```
DanielNoord, line 742: Should this method be deprecated/removed?

pylint/checkers/imports.py
```diff
@@ -418,7 +432,13 @@ def close(self):
 
     def deprecated_modules(self):
```
DanielNoord, line 433:
```suggestion
    def deprecated_modules(self) -> set[str]:
```

DanielNoord (changes requested):

tests/functional/d/deprecated/deprecated_module_enabled.rc
```diff
@@ -0,0 +1,3 @@
+[Messages Control]
+disable=all
+enable=deprecated-module
```
DudeNr33, line 3: This disable/enable should be done automatically when running with ``--minimal-messages-config``. Is there a reason we include this explicitly here? (Also applies to tests/functional/d/deprecated/deprecated_module_redundant.rc)

DudeNr33 (commented):
Thanks for taking this on!
Message sharing across checkers is yet another thing that does not fit in with the logic for disabling / enabling specific checks. I was not sure how it would behave in this case where a Mixin class is used, though.

I was wondering if it would maybe be better to make the ``DeprecatedMixin`` a standalone checker? It implements quite a few ``visit_*`` methods, but looking at ``StdlibChecker`` and ``ImportsChecker`` I am not sure if they are really used as originally intended.

For example:
``DepcrecatedMixin.visit_call(self, node)`` is overwritten in ``StdlibChecker`` without a ``super()`` call. This means that the original implementation is only executed through ``ImportsChecker``, but I don't see why it should be executed at all in this context.

Similarly, ``ImportsChecker`` reimplements the logic for ``visit_import`` and ``visit_importfrom``.

This is not really what I would expect from a Mixin class.

tests/functional/d/deprecated/deprecated_module_enabled.rc
jacobtylerwalls: Ah, it was because we don't run `--minimal-messages-config` automatically. But if we're on the verge of turning that on I can remove these.

jacobtylerwalls:
> I was wondering if it would maybe be better to make the DeprecatedMixin a standalone checker? It implements quite a few visit_* methods, but looking at StdlibChecker and ImportsChecker I am not sure if they are really used as originally intended.

I think that's a good opportunity for a follow-up optimization PR. I would probably leave the mixin alone, since it sounds like plugins are using it as a mixin, see https://github.com/PyCQA/pylint/pull/4051#issue-797144367.

> For example:
DepcrecatedMixin.visit_call(self, node) is overwritten in StdlibChecker without a super() call. This means that the original implementation is only executed through ImportsChecker, but I don't see why it should be executed at all in this context.

Right, and I'm not even sure it's implemented there because of the `@check_messages` decorator, although I don't know the intricacies of how that works with mixins.

pylint/checkers/deprecated.py
Pierre-Sassoulas: Sorry I was not able to answer sooner. I think one of the possibility offered by the mixin was to be able to extends it easily see https://github.com/PyCQA/pylint/pull/4225. I'm not sure this will still be possible if we move it to the import checker ?

jacobtylerwalls: No worries. I think it's not a problem, since I didn't end up touching `DeprecatedMixin` after all. The only change is for `StdlibChecker` to stop checking for deprecated modules, and for `ImportsChecker` to add the stdlib modules to the user-given modules. A plugin seeing a behavior change would be a plugin inheriting from one of the two checkers, not the mixin, so it's just like any behavior change.

jacobtylerwalls:
> I think that's a good opportunity for a follow-up optimization PR. I would probably leave the mixin alone, since it sounds like plugins are using it as a mixin, see https://github.com/PyCQA/pylint/pull/4051#issue-797144367.

I think @Pierre-Sassoulas is having this feeling also (leave the mixin be), but we can still make sure we're using it correctly ourselves!

DudeNr33:
Yes, shifting the general discussion about ``DeprecatedMixin`` into a separate PR or issue is reasonable. I think we should also involve @matusvalo then, as he initially introduced the Mixin class and I don't want to destroy any of his work. 

Regarding the ``@check_messages`` decorator and the use of it in the Mixin class:
It really doesn't do anything special. It just takes the bound method it decorates and adds a ``checks_msgs`` attribute to it -  a list with all the messages passed into the decorator.
Which means:
if the implementing class does not override the method, the ``ASTWalker`` will see the method as it is defined by the Mixin. If it is overwritten, it will see the method as it is defined by the implementing class. The effect the decorator had in the Mixin class will be lost, and the implementing class must decorate the method correctly - implying that you also have to include all messages possibly emitted by the Mixin class if you use a ``super()`` call.

tests/functional/d/deprecated/deprecated_module_enabled.rc
DudeNr33: Yes, that would be great! I will prepare a PR to include it in the CI.

DudeNr33 (changes requested):
From my side this one is ready to merge after ``tests/functional/d/deprecated/deprecated_module_enabled.rc`` and ``tests/functional/d/deprecated/deprecated_module_redundant.rc`` are deleted / the enable/disable sections removed. 👍

Pierre-Sassoulas (approved):
👌

DudeNr33 (approved):

PR #6361, opened 2022-04-16
Add regression tests for cell var from loop (#4574)

### LINKED ISSUE #4574: Regression: --enable config is being ignored

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #4574

### DIFF
--- a/tests/functional/c/cell_var_from_loop_enabled_regression.py
+++ b/tests/functional/c/cell_var_from_loop_enabled_regression.py
@@ -1,0 +1,3 @@
+l = []
+for x in range(10):
+    l.append(lambda: x + 10)  # [cell-var-from-loop]
--- a/tests/functional/c/cell_var_from_loop_enabled_regression.rc
+++ b/tests/functional/c/cell_var_from_loop_enabled_regression.rc
@@ -1,0 +1,3 @@
+[Messages Control]
+disable=all
+enable=cell-var-from-loop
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas:
Thank you for detecting tat the issue could be closed @jacobtylerwalls !

jacobtylerwalls (approved):

jacobtylerwalls:
Thanks @itmarst for the test case!

PR #6362, opened 2022-04-17
Fix a failure to respect inline disables for `fixme`

jacobtylerwalls:
## Type of Changes

| ✓   | :bug: Bug   |

## Description

**Before**: `# pylint: disable=fixme` didn't work on the last line of a module if pylint was invoked with `--enable=fixme`.

Followup to 0281388, identifying and fixing the other functional bug revealed there.

### DIFF
--- a/pylint/checkers/misc.py
+++ b/pylint/checkers/misc.py
@@ -166,8 +166,8 @@ class EncodingChecker(BaseChecker):
                         line=comment.start[0],
                     )
                     continue
-                else:
-                    continue
+                self.linter.add_ignored_message("fixme", line=comment.start[0])
+                continue
 
             # emit warnings if necessary
             match = self._fixme_pattern.search("#" + comment_text.lower())
--- a/pylint/checkers/misc.py
+++ b/pylint/checkers/misc.py
@@ -159,15 +159,14 @@ class EncodingChecker(BaseChecker):
                     except PragmaParserError:
                         # Printing useful information dealing with this error is done in the lint package
                         pass
-                    if set(values) & set(self.linter.config.notes):
-                        continue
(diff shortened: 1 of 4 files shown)


DanielNoord (approved):

Pierre-Sassoulas (commented):
The fix is elegant but also very specific and to be honest I don't understand it. (I checked the code around and it's strange to have the fixme warning being emitted by a class called EncodingChecker but anyway...) Why is there a disable only for fixme and not for syntax-error or bad-inline-option that are also raised by this checker ? I also don't understand why the disable is required in the first place. Would you mind explaining the thinking behind it ?

jacobtylerwalls:
Ah hold on, there's a better solution. (You're right!)

pylint/checkers/misc.py
```diff
@@ -159,16 +159,15 @@ def process_tokens(self, tokens):
                     except PragmaParserError:
                         # Printing useful information dealing with this error is done in the lint package
                         pass
-                    if set(values) & set(self.linter.config.notes):
-                        continue
-                    self.linter.disable("fixme", scope="module", line=comment.start[0])
```
jacobtylerwalls: This test was always false, so we never got a `continue` that would prevent warnings. It's a bit of a silly test: `values` will always be `fixme`, the msgid. We're testing if the msgid in the disable, i.e. the `pylint: disable=fixme` is in `--notes`, which defaults to `FIXME, XXX, TODO`. Not only should we not be checking that (?!), because you should be able to disable the fixme msg no matter what comments you are searching for, we certainly shouldn't do a case-sensitive comparison.

Pierre-Sassoulas (approved):
Thanks for clarifying ! This small cleanup will save some kwh in years to come 😉

PR #6363, opened 2022-04-17
Move ``Run`` and ``PyLinter`` options to separate file

DanielNoord:

## Type of Changes

## Description


pylint/lint/pylinter.py
```diff
@@ -622,8 +261,7 @@ def __init__(
         self.stats = LinterStats()
 
         # Attributes related to (command-line) options and their parsing
-        self._external_opts = options
```
Pierre-Sassoulas, line 625: This was not used anywhere?

Pierre-Sassoulas (commented):
I think it's a pretty nice step in the right direction of moving things away from the PyLinter. I'm only wondering if ``base_option.py`` should be in ``pylint.config`` instead.

pylint/lint/pylinter.py
DanielNoord: No! I double-checked. It's private so we can remove it without any hassle, the first time I think 😄

DanielNoord:
> I think it's a pretty nice step in the right direction of moving things away from the PyLinter. I'm only wondering if `base_option.py` should be in `pylint.config` instead.

I'll let you decide. I was doubting this as well. I thought "options of checkers are also in the same file so it is probably better to keep the options together with the object that provides them".

Pierre-Sassoulas (approved):
> keep the options together with the object that provides them".

Make sense, let's keep it that way than !

PR #6365, opened 2022-04-17
Prevent false negatives for fatal errors with `--minimal-messages-config`

jacobtylerwalls:
## Type of Changes

## Description

`--minimal-messages-config` is an awesome testing tool. But while working on #6362, I found it surprising that crashes were silenced. There were a couple moments where I thought I had a solution that would pass the tests, only to find later that fatal errors were being silenced. (Unbeknownst to me, I was generating a bunch of crash_reports.txt, but pytest was green.)

By design, this tool allows false negatives for unmentioned errors. But I think `astroid-error` and `fatal`, which are the two errors `pylint` emits when any internal failure occurs, should be included with the "minimal" config, because we don't really need to test the isolation of `astroid-error` and `fatal`. It's more important to not have false negatives for crashes IMO.

### DIFF
--- a/pylint/testutils/lint_module_test.py
+++ b/pylint/testutils/lint_module_test.py
@@ -68,6 +68,9 @@ class LintModuleTest:
         if config and config.getoption("minimal_messages_config"):
             with self._open_source_file() as f:
                 messages_to_enable = {msg[1] for msg in self.get_expected_messages(f)}
+                # Always enable fatal errors
+                messages_to_enable.add("astroid-error")
+                messages_to_enable.add("fatal")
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):
Nice catch !

DudeNr33 (approved):
The change is fine for me. Enabling those two messages does not interfere with the goal of these tests.

But just out of curiosity: did those crashes only occur when you ran with ``--minimal-messages-config``, or also without?

jacobtylerwalls:
Only with `--minimal`. So I would have seen it eventually/CI would have caught it.

PR #6367, opened 2022-04-17
Fix a crash in the ``unsupported-membership-test`` checker involving tuple unpacking

### LINKED ISSUE #6366: AttributeError: 'ClassDef' object has no attribute 'value'
Running the latest version of pylint (2.6, and also with 2.4.4), I get this error:

astroid/bases.py:113: AttributeError: 'ClassDef' object has no attribute 'value'

Any suggestions on what could be causing it would be greatly appreciated.

jacobtylerwalls:
## Type of Changes

## Description
An assumption of direct parentage between `AssignName` and `Assign` was violated by tuple unpacking.

Closes #6366

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1066,8 +1066,15 @@ def _supports_protocol_method(value: nodes.NodeNG, attr: str) -> bool:
         return False
 
     first = attributes[0]
+
+    # Return False if a constant is assigned
     if isinstance(first, nodes.AssignName):
-        if isinstance(first.parent.value, nodes.Const):
+        this_assign_parent = get_node_first_ancestor_of_type(first, nodes.Assign)
+        assert this_assign_parent is not None
+        if isinstance(this_assign_parent.value, nodes.BaseContainer):
+            if all(isinstance(n, nodes.Const) for n in this_assign_parent.value.elts):
+                return False
+        if isinstance(this_assign_parent.value, nodes.Const):
             return False
     return True
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -186,6 +186,11 @@ What's New in Pylint 2.13.6?
 ============================
 Release date: TBA
 
+* Fix a crash in the ``unsupported-membership-test`` checker when assigning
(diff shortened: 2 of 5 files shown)


pylint/checkers/utils.py
```diff
@@ -1066,8 +1066,15 @@ def _supports_protocol_method(value: nodes.NodeNG, attr: str) -> bool:
         return False
 
     first = attributes[0]
+
+    # Return False if a constant is assigned
     if isinstance(first, nodes.AssignName):
-        if isinstance(first.parent.value, nodes.Const):
+        this_assign_parent = get_node_first_ancestor_of_type(first, nodes.Assign)
+        assert this_assign_parent is not None
```
Pierre-Sassoulas, line 1073: Are we sure that ``this_assign_parent `` is never None ? What are the downsides of handling the None case ?

jacobtylerwalls: Thanks, I did miss the sneaky walrus operator. (Turned into a hilarious test case.) So it's worth adding a pragma for the rest. I looked into testing `mixins.AssignTypeMixin` instead, but that pulled in too many classes, e.g. `AssignName` itself.

tests/functional/n/none_dunder_protocols_named_expr.py
```diff
@@ -0,0 +1,13 @@
+"""Although unlikely to happen, this test case was almost missed in development."""
```
jacobtylerwalls, line 1: I'm torn on whether this test is truly worth it 🤔

Pierre-Sassoulas: I think if it prevents a crash from going undetected it's worth it 😄 Some codebases have truly horrific code but pylint need to be all terrains. Our typing is not yet capable of warning of everything but I think we should not silence it with assert, even if admittedly this kind of code construct won't happen very often in the wild, and won't happen with sane lib.

DanielNoord (approved):

Pierre-Sassoulas (approved):

PR #6381, opened 2022-04-18
Make ``Message`` a ``dataclass``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/message/message.py
+++ b/pylint/message/message.py
@@ -70,7 +70,6 @@ class Message:  # pylint: disable=too-many-instance-attributes
         self.column = location.column
         self.end_line = location.end_line
         self.end_column = location.end_column
-        self.location = location
 
     def format(self, template: str) -> str:
         """Format the message according to the given template.
--- a/pylint/message/message.py
+++ b/pylint/message/message.py
@@ -30,7 +30,6 @@ class Message:  # pylint: disable=too-many-instance-attributes
     column: int
     end_line: int | None
     end_column: int | None
-    location: MessageLocationTuple
 
     def __init__(
         self,
--- a/pylint/message/message.py
+++ b/pylint/message/message.py
@@ -8,7 +8,7 @@ from dataclasses import asdict, dataclass
 from warnings import warn
 
 from pylint.constants import MSG_TYPES
-from pylint.interfaces import Confidence
+from pylint.interfaces import UNDEFINED, Confidence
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):
👌  (This is a lot faster than creating a class using an 'exec', with old style named tuple too )

pylint/message/message.py
```diff
@@ -4,90 +4,78 @@
...
+                None,
+            )
+
+        self.msg_id = msg_id
+        self.symbol = symbol
+        self.msg = msg
+        self.C = msg_id[0]
+        self.category = MSG_TYPES[msg_id[0]]
+        self.confidence = confidence or UNDEFINED
+        self.abspath = location.abspath
```
Pierre-Sassoulas, line 65: We could use the location as an attribute and create property to not break the API.

DanielNoord: Not sure what you mean, see L73. I added location to be able to access it directly and with the new `if` we can now assume it is `MessageLocationTuple` here.

Pierre-Sassoulas: Doesn't it mean that we could modify "line" without modifying "location.line" ? Maybe we need to use a frozen dataclass to be sure it won't be incoherent.

DanielNoord: Ah, yeah I didn't consider that!

I'll remove the `location` attribute. We don't even need it ourselves. That makes it easier. If we want to expand the API later we can think about it then.

pylint/message/message.py
```diff
@@ -4,90 +4,77 @@
...
+    confidence: Confidence
+    abspath: str
+    path: str
+    module: str
+    obj: str
+    line: int
+    column: int
+    end_line: int | None
+    end_column: int | None
+    location: MessageLocationTuple
```
DanielNoord, line 33:
```suggestion
```

Pierre-Sassoulas (approved):
Great ! I kinda want to know how much faster pylint's run right now but I'm too lazy to do it 😄 If I had to guess, something like 0.10% to 0.20%

DanielNoord:
> Great ! I kinda want to know how much faster pylint's run right now but I'm too lazy to do it 😄 If I had to guess, something like 0.10% to 0.20%

Anyway, my own contributing is 10% faster by not having to think about the horrible construct of `_MsgBase`... 😄

PR #6383, opened 2022-04-18
Make ``BaseChecker`` a ``MapReduceMixin``

DanielNoord:

## Type of Changes

## Description

Kinda high priority since `main` is broken due to two `typing` PRs apparently depending on each other but passing on their own.

`main` complains about `checker.get_map_data()` and `checker.reduce_map_data` in `parallel.py`. Instead of adding ignores there I think `mypy` is actually right and warns of a design flaw. There is no need to separate these `MixIns` (ideally we would deprecate the `MixIn` as there is little overhead of making `BaseChecker` a `MapReduceMixin` and just having the methods do nothing.
That's why I propose this solution.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -66,7 +66,7 @@ Release date: TBA
 * ``MapReduceMixin`` has been deprecated. ``BaseChecker`` now implements ``get_map_data`` and
   ``reduce_map_data``. If a checker actually needs to reduce data it should define ``get_map_data``
   as returning something different than ``None`` and let its ``reduce_map_data`` handle a list
-  of the type returned by ``get_map_data``.
+  of the types returned by ``get_map_data``.
   An example can be seen by looking at ``pylint/checkers/similar.py``.
 
 * ``UnsupportedAction`` has been deprecated.
--- a/ChangeLog
+++ b/ChangeLog
@@ -63,6 +63,12 @@ Release date: TBA
 
   Ref #5392
 
+* ``MapReduceMixin`` has been deprecated. ``BaseChecker`` now implements ``get_map_data`` and
+  ``reduce_map_data``. If a checker actually needs to reduce data it should define ``get_map_data``
+  as returning something different than ``None`` and let its ``reduce_map_data`` handle a list
+  of the type returned by ``get_map_data``.
+  An example can be seen by looking at ``pylint/checkers/similar.py``.
+
 * ``UnsupportedAction`` has been deprecated.
 
   Ref #5392
--- a/doc/how_tos/custom_checkers.rst
+++ b/doc/how_tos/custom_checkers.rst
@@ -246,11 +246,11 @@ Now we can debug our checker!
 Parallelize a Checker
 ---------------------
 
-``BaseChecker`` have a ``get_map_data`` and a ``reduce_map_data`` function that
-permit to parallelize the checks when used with the ``-j`` otion. If a checker
+``BaseChecker`` has two methods ``get_map_data`` and ``reduce_map_data`` that
+permit to parallelize the checks when used with the ``-j`` option. If a checker
 actually needs to reduce data it should define ``get_map_data`` as returning
 something different than ``None`` and let its ``reduce_map_data`` handle a list
-of the type returned by ``get_map_data``.
+of the types returned by ``get_map_data``.
 
 An example can be seen by looking at ``pylint/checkers/similar.py``.
 
--- a/doc/how_tos/custom_checkers.rst
+++ b/doc/how_tos/custom_checkers.rst
@@ -243,6 +243,17 @@ Now we can debug our checker!
     environment variable or by adding the ``my_plugin.py``
     file to the ``pylint/checkers`` directory if running from source.
 
+Parallelize a Checker
+---------------------
+
+``BaseChecker`` have a ``get_map_data`` and a ``reduce_map_data`` function that
+permit to parallelize the checks when used with the ``-j`` otion. If a checker
+actually needs to reduce data it should define ``get_map_data`` as returning
+something different than ``None`` and let its ``reduce_map_data`` handle a list
+of the type returned by ``get_map_data``.
+
+An example can be seen by looking at ``pylint/checkers/similar.py``.
+
 Testing a Checker
 -----------------
 Pylint is very well suited to test driven development.
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -28,7 +28,7 @@ New checkers
 * ``MapReduceMixin`` has been deprecated. ``BaseChecker`` now implements ``get_map_data`` and
   ``reduce_map_data``. If a checker actually needs to reduce data it should define ``get_map_data``
   as returning something different than ``None`` and let its ``reduce_map_data`` handle a list
-  of the type returned by ``get_map_data``.
+  of the types returned by ``get_map_data``.
   An example can be seen by looking at ``pylint/checkers/similar.py``.
 
 * Add new check ``unnecessary-dunder-call`` for unnecessary dunder method calls.
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -25,6 +25,12 @@ New checkers
 
   Closes #5895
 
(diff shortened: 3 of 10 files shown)


Pierre-Sassoulas (commented):
Won't this imply that BaseChecker is able to perform map reduce when it's not ?

DanielNoord:
> Won't this imply that BaseChecker is able to perform map reduce when it's not ?

Yes and no I guess. It is able to perform it in so far as that it has a `reduce_map_data` method. However, that simply does nothing. The way the `MixIn` works is that `get_map_data` should return something (`Any`) which then gets appended to a `list` specific to that checker. When reducing all data the respective `reduce_map_data` method is then called to reduce the specific from of data that `get_map_data` returns.
So, you will never be able to get this to work with only implementing `get_map_data` and I think you will always catch this when writing a test for this.


----------

The issue with the current implementation is that there is a dictionary mapping checker to data. Only checkers that implement `get_map_data` are added to this dict. However, when we interact with this dict later on `mypy` has no way of knowing that implementing `get_map_data` is a requirement for having an entry in this dictionary and just considers all entries to be `BaseChecker`, and then complains.

pylint/lint/parallel.py
```diff
@@ -71,11 +71,9 @@ def _worker_check_single_file(
...
     mapreduce_data = defaultdict(list)
     for checker in _worker_linter.get_checkers():
-        try:
-            data = checker.get_map_data()
-        except AttributeError:
-            continue
-        mapreduce_data[checker.name].append(data)
+        data = checker.get_map_data()
+        if data is not None:
+            mapreduce_data[checker.name].append(data)
```
jacobtylerwalls, line 76: I think this is a sign we would need a changelog. If someone's subclass implemented `get_map_data()` and returned None, they need to know there won't be a key in `mapreduce_data` now. Obviously we can spare all this detail, but just a short note to say we promoted the map/reduce functionality, and then at that point go ahead and deprecate MapReduceMixin, and then... I'm starting to worry this is solving a problem that's not really a problem.

I would just do this and be done with it. WDYT?
```diff
diff --git a/pylint/lint/parallel.py b/pylint/lint/parallel.py
index a1ecb7ac..6b927173 100644
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, Any
 import dill
 
 from pylint import reporters
+from pylint.checkers.mapreduce_checker import MapReduceMixin
 from pylint.lint.utils import _patch_sys_path
 from pylint.message import Message
 from pylint.typing import FileItem
@@ -71,11 +72,9 @@ def _worker_check_single_file(
     _worker_linter.check_single_file_item(file_item)
     mapreduce_data = defaultdict(list)
     for checker in _worker_linter.get_checkers():
-        try:
+        if isinstance(checker, MapReduceMixin):
             data = checker.get_map_data()
-        except AttributeError:
-            continue
-        mapreduce_data[checker.name].append(data)
+            mapreduce_data[checker.name].append(data)
     msgs = _worker_linter.reporter.messages
     assert isinstance(_worker_linter.reporter, reporters.CollectingReporter)
     _worker_linter.reporter.reset()
@@ -120,6 +119,7 @@ def _merge_mapreduce_data(
         if checker.name in collated_map_reduce_data:
             # Assume that if the check has returned map/reduce data that it has the
             # reducer function
+            assert isinstance(checker, MapReduceMixin)
             checker.reduce_map_data(linter, collated_map_reduce_data[checker.name])
 ```

Comment (DanielNoord) [reply pylint/lint/parallel.py:76]: I can certainly add a changelog. Note that if you currently return `None` it won't do much since then `reduce_map_data` will just return  `list[None]` which is difficult to reduce then.

That would also work, but for me the underlying issue is the `Mixin` to begin with. It seems to overcomplicate stuff which could be relatively simple. We'll never need `MapReduceMixin` with any other class than `BaseChecker` and I can't really see anything that makes it bad to have to have dummy methods on the base class for these methods.

Comment (jacobtylerwalls) [reply pylint/lint/parallel.py:76]: I guess I don't feel this is less complicated, given that we're not deprecating the mixin in this PR. I don't feel too strongly about it either way, though.

Comment (DanielNoord) [reply pylint/lint/parallel.py:76]: Would you want me to add the deprecation in this PR?

Comment (jacobtylerwalls) [reply pylint/lint/parallel.py:76]: Let's see what other reviewers think. If we really need to unblock main we can always do that separately with just fixing mypy.

Comment (DanielNoord) [reply pylint/lint/parallel.py:76]: I'll make a quick PR with some disables. Not being able to run `pre-commit` locally makes it impossible to commit in VSCode without ``--no-verify`` which has been bugging me for 24h now.

Comment (DanielNoord) [reply pylint/lint/parallel.py]: Added a last commit to show potential deprecation.

@Pierre-Sassoulas What do you think about deprecating this class?

**Review by Pierre-Sassoulas — APPROVED** [review]

**Review by Pierre-Sassoulas — COMMENTED** [review]
LGTM the final result is pretty nice and reasonable 👍

File: ChangeLog [hunk 2]
```diff
@@ -63,6 +63,12 @@ Release date: TBA
 
   Ref #5392
 
+* ``MapReduceMixin`` has been deprecated. ``BaseChecker`` now implements ``get_map_data`` and
+  ``reduce_map_data``. If a checker actually needs to reduce data it should define ``get_map_data``
+  as returning something different than ``None`` and let its ``reduce_map_data`` handle a list
+  of the type returned by ``get_map_data``.
```
jacobtylerwalls, line 69:
```suggestion
  of the types returned by ``get_map_data``.
```

doc/how_tos/custom_checkers.rst
```diff
@@ -243,6 +243,17 @@ Now we can debug our checker!
     environment variable or by adding the ``my_plugin.py``
     file to the ``pylint/checkers`` directory if running from source.
 
+Parallelize a Checker
+---------------------
+
+``BaseChecker`` have a ``get_map_data`` and a ``reduce_map_data`` function that
```
jacobtylerwalls, line 249:
```suggestion
``BaseChecker`` has two methods ``get_map_data`` and ``reduce_map_data`` that
```

doc/how_tos/custom_checkers.rst
```diff
@@ -243,6 +243,17 @@ Now we can debug our checker!
...
+permit to parallelize the checks when used with the ``-j`` otion. If a checker
+actually needs to reduce data it should define ``get_map_data`` as returning
+something different than ``None`` and let its ``reduce_map_data`` handle a list
+of the type returned by ``get_map_data``.
```
jacobtylerwalls, line 253:
```suggestion
of the types returned by ``get_map_data``.
```

doc/whatsnew/2.14.rst
```diff
@@ -25,6 +25,12 @@ New checkers
 
   Closes #5895
 
+* ``MapReduceMixin`` has been deprecated. ``BaseChecker`` now implements ``get_map_data`` and
+  ``reduce_map_data``. If a checker actually needs to reduce data it should define ``get_map_data``
+  as returning something different than ``None`` and let its ``reduce_map_data`` handle a list
+  of the type returned by ``get_map_data``.
```
jacobtylerwalls, line 31:
```suggestion
  of the types returned by ``get_map_data``.
```

jacobtylerwalls (commented):
Just some cleanups.

DanielNoord:
Thanks for the thorough review guys! Agree with @Pierre-Sassoulas!

doublethefish:
I completely missed this change, but it's great to see. When I added the Mixin object the intent was to have as little impact over the core code as possible, it's always been clear that the right _intemediary_ place for it was in the base-checker - longer term the map-reduce functionality should completely go in favour of a fully parallelised astroid or moved into astroid 
 or something.

> I think mypy is actually right and warns of a design flaw

Mypy is definitely correct that there's an issue, but it's not a warning about a design flaw. Mypy is just showing that there's a lack of type checking and/or explicit error handling around whether the mix-in has been applied or not. After all python isn't C++ and is a _dynamic_ language which allows for that sort of thing more easily :) IIRC the design implicitly ensures that the mix-in is applied.

PR #6402, opened 2022-04-19
WIP: Add regression test for #6094

### LINKED ISSUE #6094: no-member raised on __getattribute__ for modules
### Bug description

In the following code snippet, pylint complains about line 3 with `E1101: Module 'sys' has no '__getattribute__' member (no-member)`:

```
import sys

print(sys.__getattribute__("argv"))
print(sys.__dict__)

s = "hello"
print(s.__getattribute__("capitalize")())
```

Note that this seems to be OK for other usages of `__getattribute__`.

### Configuration

_No response_

### Command used

```shell
`pylint test.py`
```

### Pylint output

```shell
************* Module caosadvancedtools.models.test
test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
test.py:3:6: E1101: Module 'sys' has no '__getattribute__' member (no-member)
test.py:6:0: C0103: Constant name "s" doesn't conform to UPPER_CASE naming style (invalid-name)

Your code has been rated at -4.00/10 (previous run: -4.00/10, +0.00)
```
```


### Expected behavior

Line 3 should be just as OK as line 7.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110]
```

### OS / Environment

Debian 11

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #6094 (requires PyCQA/astroid#1519)


Pierre-Sassoulas (approved):

jacobtylerwalls:
Waiting on Daniel's investigation in https://github.com/PyCQA/astroid/pull/1519

DanielNoord:
My bad. I should have written something about this.


Turns out this is quite difficult to solve. The ObjectModel is used to resolve some methods but for some reasons it gets priority over many other sources in the mro. This is problematic as very often stuff like `__init__` and `__eq__` get overwritten.

However, in other places for some reason we depend on going to ObjectModel directly instead of using the mro.

I haven't really found a solution but looking at the failed test in that PR and debugging why it gets incorrectly ingerrred shows pretty clear what is going wrong there.

Pierre-Sassoulas:
Wouldn't it make sense to merge with the known bug instead ?

DanielNoord:
@Pierre-Sassoulas As can be seen by the CI the test doesn't pass currently so I don't think we can/should merge this?

tests/functional/n/no/no_member.py
```diff
@@ -44,5 +44,10 @@ class Derived(Base):
 print(Derived.label)
 
 
+# Regression test for https://github.com/pylint-dev/pylint/issues/6094
+# pylint: disable-next=unnecessary-dunder-call
+print(parse.__getattribute__("urlparse"))
```
Pierre-Sassoulas, line 49:
```suggestion
print(parse.__getattribute__("urlparse"))  # [no-member] known false positive (#6094)
```

We'd also need to upgrade the txt and remove the fragment but you get the idea.

DanielNoord: I don't think that adds much benefit compared to just keeping the issue open in the issue tracker. Somebody who wants to add a test for it will very likely come up with the same test in a new PR.

PR #6405, opened 2022-04-20
Do not append namespace packages to sys.path

### LINKED ISSUE #2648: Invalid "Module import itself" error for namespace packages.

### LINKED ISSUE #5226: Support an Option to Disable sys.path patching

DanielNoord:

## Type of Changes

## Description

Counter to #5235.

This is not finished yet as for some reason running `pylint` in pre-commit does not allow this fix. However, I think we should explore if this is a possible solution to the problem. There should be a way to determine whether something is a namespace package, the `python` import statement can do so as well.

Closes #5226, Closes #2648


Pierre-Sassoulas:
I think both approach are not mutually exclusive, the option is a permanent fix to let the user tell pylint what to do. But we can still make namespace automated detection better so users don't have to use the option as often.

DanielNoord:
> I think both approach are not mutually exclusive, the option is a permanent fix to let the user tell pylint what to do. But we can still make namespace automated detection better so users don't have to use the option as often.

I think we should try to avoid adding the option. There is no inherent reason why we wouldn't be able to mimic the import mechanics of `python`. It's difficult (for sure...), but there is nothing that blocks us from 100% copying it eventually.
When we finally do so we are then stuck with supporting an option that is not necessarily correct and blocks some of the behaviour of `pylint`. For example, with this change you could (in theory I think) lint both a normal and a namespace package at the same time. With the option you wouldn't be able to do so. 
Rather than turning off core behaviour of `pylint` I'd like to fix that behaviour and not have to support turning it off down the line.

pylint/lint/utils.py
```diff
@@ -71,12 +73,32 @@ def get_fatal_error_message(filepath: str, issue_template_path: Path) -> str:
...
 
+def _is_part_of_namespace_package(filename: str) -> bool:
+    """Check if a file is part of a namespace package."""
+    filepath = Path(filename)
+    try:
+        modname = modutils.modpath_from_file(filename)
+    except ImportError:
+        modname = [filepath.stem]
+    if filepath.is_dir():
+        filepath = filepath / "__init__.py"
```
DanielNoord, line 84:
```suggestion
    try:
        modname = modutils.modpath_from_file(filename)
    except ImportError:
        modname = [Path(filename).stem]
```

ChangeLog
```diff
@@ -17,6 +17,13 @@ Release date: TBA
 
   Closes #4301
 
+* Improved recognition of namespace packages. They should no longer be added to ``sys.path``
+  which should prevent various issues with false positives.
+  Because of the difficulities with mirroring the python import mechanics we would gladly
```
jacobtylerwalls, line 22:
```suggestion
  Because of the difficulties with mirroring the python import mechanics we would gladly
```

doc/whatsnew/2.14.rst
```diff
@@ -81,6 +81,13 @@ Other Changes
 
   Closes #4301
 
+* Improved recognition of namespace packages. They should no longer be added to ``sys.path``
+  which should prevent various issues with false positives.
+  Because of the difficulities with mirroring the python import mechanics we would gladly
```
jacobtylerwalls, line 86:
```suggestion
  Because of the difficulties with mirroring the python import mechanics we would gladly
```

ChangeLog
DanielNoord: Your review always make me wonder whether I paid any attention in school..

Pierre-Sassoulas: We could activate our own spelling checker for that, let me try 😄

Pierre-Sassoulas (changes requested):
The change look nice but this is a tricky to tests and easy to break. I'm not super confident. Probably a good thing that we release the beta first here.

test_namespace_pkg/existing/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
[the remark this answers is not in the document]
DanielNoord: We can't put it in `tests/` as `tests/` is a normal package.. :(

Would you mind making the `namespace` repository?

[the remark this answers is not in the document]
Pierre-Sassoulas: Can we launch the tests from ``tests/data/`` instead of from the root so the test suite do not realize that it's inside a package ? Depending on another repo is not ideal I think you were right to want to have the test data here.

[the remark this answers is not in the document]
DanielNoord: No, for this test to work there shouldn't be a `__init__` in between the package and the `cwd`...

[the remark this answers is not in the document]
Pierre-Sassoulas: I searched for ``os.chdir`` in the code base and we seem to have a number of test fixture of context manager to handle this, in particular ``            with _test_cwd():`` in test_self.py. I think we could even unify this at some point because there seem to be quite a bit of duplication.  Wouldn't that approach work ?

```python
@contextlib.contextmanager
def _test_cwd(new_cwd: Path | str) -> Generator[None, None, None]:
    original_dir = os.getcwd()
    os.chdir(new_cwd)
    try:
        yield
    finally:
        os.chdir(original_dir)



with _test_cwd("tests/data/namespace/"):
    ...
```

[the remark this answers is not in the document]
DanielNoord: Just quickly touching in: just got some bad family news. Not sure if I'll be able to devote time to programming over the next week.

If somebody could take a look at this final comment and get this PR merged for the beta that would be much appreciated!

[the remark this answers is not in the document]
jacobtylerwalls: Of course, and sorry to hear that. Be well.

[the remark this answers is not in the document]
Pierre-Sassoulas: Sorry to hear that Daniel, take care of yourself.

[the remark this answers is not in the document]
DanielNoord: I checked, but this doesn't work... We pass both with and without changes with the current tests. I really think we can't do much else than create a new root directory or a new package.

I believe this is due to how `pytest` adds `tests` to `sys.path` or something. Make all attempts to mix `namespace` with `normal` packages in a test directory fail.

tests/lint/unittest_lint.py
```diff
@@ -170,6 +171,16 @@ def test_more_args(fake_path, case):
...
 
 
+def test_namespace_package_sys_path() -> None:
+    """Test that we do not append namespace packages to sys.path."""
+    result = subprocess.run(
+        [sys.executable, "-m", "pylint", "test_namespace_pkg/"],
+        check=False,
+        stdout=subprocess.PIPE,
+    )
+    assert "Module import itself" not in result.stdout.decode("utf-8")
```
jacobtylerwalls, line 181: I'd prefer to demo in the tests what our docs say to do. (`subprocess` seems like something we may as well avoid here.)
```suggestion
    pylint_output = StringIO()
    reporter = text.TextReporter(pylint_output)
    Run(["test_namespace_pkg/"], reporter=reporter, exit=False)
    assert "Module import itself" not in pylint_output.getvalue()
```

test_namespace_pkg/existing/__init__.py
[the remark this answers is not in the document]
jacobtylerwalls: A new test file under `tests/regrtest_data/` fixes this. Maybe preferable?

```diff
diff --git a/tests/regrtest_data/test_namespace_package.py b/tests/regrtest_data/test_namespace_package.py
new file mode 100644
index 00000000..aa9194ad
--- /dev/null
+++ b/tests/regrtest_data/test_namespace_package.py
@@ -0,0 +1,11 @@
+from io import StringIO
+from pylint.reporters.text import TextReporter
+from pylint.testutils._run import _Run as Run
+
+
+def test_namespace_package_sys_path() -> None:
+    """Test that we do not append namespace packages to sys.path."""
+    pylint_output = StringIO()
+    reporter = TextReporter(pylint_output)
+    Run(["tests/regrtest_data/test_namespace_pkg/"], reporter=reporter, exit=False)
+    assert "Module import itself" not in pylint_output.getvalue()
diff --git a/tests/regrtest_data/test_namespace_pkg/else/__init__.py b/tests/regrtest_data/test_namespace_pkg/else/__init__.py
new file mode 100644
index 00000000..e69de29b
diff --git a/tests/regrtest_data/test_namespace_pkg/existing/__init__.py b/tests/regrtest_data/test_namespace_pkg/existing/__init__.py
new file mode 100644
index 00000000..1b93e7d8
--- /dev/null
+++ b/tests/regrtest_data/test_namespace_pkg/existing/__init__.py
@@ -0,0 +1,3 @@
+from existing import my_func  # type: ignore[attr-defined]
+
+my_func()
```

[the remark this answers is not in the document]
DanielNoord: Hm, that's weird. I thought I checked this and that this wouldn't currently fail. I'll try this `diff` tomorrow. Thanks @jacobtylerwalls for the additional investigation! 😄

tests/lint/test_namespace_packages.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_namespace_package_sys_path() -> None:
+    """Test that we do not append namespace packages to sys.path.
+
+    The test package is based on https://github.com/PyCQA/pylint/issues/2648.
+    """
+    pylint_output = StringIO()
+    reporter = TextReporter(pylint_output)
+    Run(["tests/regrtest_data/namespace_import_self/"], reporter=reporter, exit=False)
+    assert "Module import itself" not in pylint_output.getvalue()
```
jacobtylerwalls, line 21: I'm just wondering if we can make this more robust against other fatal errors, as we've seen elsewhere recently. Can we drill into the `.linter.stats` of the `Run` object and assert no fatal errors?

DanielNoord: I'll have a look!

jacobtylerwalls: For instance, copying just this test over to `main` passes because of a `syntax-error`. (But the test looks good -- with all of the changes in this PR, minus the substantive changes in `if _is_part_of_namespace_pacakge`, we get the failure with `module-imports-self` as expected.)

DanielNoord: Let me know what you think of the current solution!

DanielNoord: Never mind. That test doesn't work 😓

jacobtylerwalls: The message assertion or the whole thing?

DanielNoord: At the moment, both 😄 Working on a fix, think I found something!

DanielNoord: This should be better!

jacobtylerwalls (approved):
Thanks for the updates, looks great and fails on main!

Pierre-Sassoulas (approved):
The test is clever ! Do we release the beta right now or should we first release the current patch versions of astroid and pylint ?

DanielNoord:
> The test is clever ! Do we release the beta right now or should we first release the current patch versions of astroid and pylint ?

I think we can release the beta. I don't know of any blockers in `astroid`. I leave the decision for the `pylint` patch version to you. It doesn't really block the beta, although the "monday dependency update" might trigger people to visit our repo and see the news post about the beta. That would be an argument to do the patch first and then the beta, so the beta is the last post in `Releases`.

PR #6415, opened 2022-04-20
Fix crash when open is called with an integer mode

### LINKED ISSUE #6414: Exception thrown when checking code that passes an integer mode= to open
### Bug description

```python
"""."""

# pylint: disable=consider-using-with


open('foo', mode=2)
```

### Command used

```shell
pylint t.py
```

### Pylint output

```shell
************* Module t
~/tmp/t.py:6:0: W1501: "2" is not a valid mode for open. (bad-open-mode)
Exception on node <Call l.6 at 0x7fcce79d8190> in file '~/tmp/t.py'
  File "~/.pyenv/versions/3.9.10/lib/python3.9/site-packages/pylint/utils/ast_walker.py", line 72, in walk
    callback(astroid)
  File "~/.pyenv/versions/3.9.10/lib/python3.9/site-packages/pylint/checkers/stdlib.py", line 534, in visit_call
    self._check_open_encoded(node, inferred.root().name)
  File "~/.pyenv/versions/3.9.10/lib/python3.9/site-packages/pylint/checkers/stdlib.py", line 633, in _check_open_encoded
    and "b" not in mode_arg.value
TypeError: argument of type 'int' is not iterable
~/tmp/t.py:1:0: F0001: Fatal error while checking '~/tmp/t.py'. Please open an issue in our bug tracker so we address this. There is a pre-filled template that you can use in '~/.cache/pylint/pylint-crash-2022-04-20-11.txt'. (fatal)

Your code has been rated at 0.00/10 (previous run: -10.00/10, +10.00)
```

### Expected behavior

No exceptions should be thrown

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.9.10 (main, Mar 16 2022, 06:54:08)
[GCC 11.2.0]
```

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #6414

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -225,6 +225,16 @@ Release date: TBA
   (Ie. not necessarily at the end)
 
 
+What's New in Pylint 2.13.8?
+============================
+Release date: TBA
+
+* Fix a crash when linting a file that passes an integer ``mode=`` to
+  ``open``
+
+  Closes #6414
+
+
 What's New in Pylint 2.13.7?
 ============================
 Release date: 2022-04-20
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -610,3 +610,8 @@ Other Changes
 * Only raise ``not-callable`` when all the inferred values of a property are not callable.
 
   Closes #5931
+
+* Fix a crash when linting a file that passes an integer ``mode=`` to
+  ``open``
+
+  Closes #6414
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -659,7 +659,7 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
         if (
             not mode_arg
             or isinstance(mode_arg, nodes.Const)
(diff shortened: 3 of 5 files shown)


DanielNoord:
Should we take this opportunity to introduce ``unsupported-file-mode``?

Pierre-Sassoulas:
What would be the difference with ``bad-open-mode`` ?

DanielNoord:
😅 

Other suggestion: Let's merge `_check_open_encoded` and `_check_open_mode`. In both we're inferring the `mode` argument and duplicating some code.
Can't we do ``_check_open_call``?

Pierre-Sassoulas:
👍  but let's do the refactor in another 2.14 only MR, I don't want to deal with cherry-pick conflicts on the maintenance branch 😄

DanielNoord:
Are you assigning yourself to that refactor? If not, perhaps open an issue for `2.14`?

DanielNoord (approved):

Pierre-Sassoulas:
Opened #6422 (work in progress right now)

PR #6421, opened 2022-04-21
Remove ``assign-to-new-keyword``

### LINKED ISSUE #4683: Cleanup after we drop support for Python 3.6

DanielNoord:

## Type of Changes

## Description

Closes #4683

Ref. coverage drop in https://github.com/PyCQA/pylint/pull/5921


Pierre-Sassoulas (commented):
I don't know about this one, theoretically we could have new keywords in the future, so this is useless right now but it could become useful again. What about adding pragmas no cover ?

DanielNoord:
> I don't know about this one, theoretically we could have new keywords in the future, so this is useless right now but it could become useful again. What about adding pragmas no cover ?

https://github.com/PyCQA/pylint/pull/5921#issuecomment-1078704680

😄 

I'm not sure. It's trivial to re-add this later on and I don't think new keywords are likely to be added as the language matures. I'd say remove for now. With `DeletedMessage` we make sure to save the message id if we want to re-add it later on.

Pierre-Sassoulas (approved):
Alright, if past me said so.. 😄 

But yeah it's very unlikely that new keyword are introduced now (they reused from for ``raise Exception() from e`` instead of creating a new one. Let's merge !

douardda:
It's been quite a while, but out of curiosity, how did @SoftwareHeritage  swh-core get included as a sumodule here? (I randomly fell on this by finding a relation between pylint and swh-core in our big graph I was not expecting at all  :-) )

DanielNoord:
I think I was debugging an issue somebody reported with that module and forget to clean my tree when switching to this branch.

PR #6422, opened 2022-04-21
[refactor] Create a generic '_check_open_call' function for the StdlibChecker

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to #6415


DanielNoord:
@Pierre-Sassoulas Took a couple of tries, but this should pass 😅 

Let me know if we can merge this!

Pierre-Sassoulas (commented):
Look good, thank you for rethinking this, I did not realize the problem was the constants.

PR #6424, opened 2022-04-21
Add typing to ``pylint/config/__init__.py``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/config/__init__.py
+++ b/pylint/config/__init__.py
@@ -2,6 +2,8 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
+from __future__ import annotations
+
 import os
 import pathlib
 import pickle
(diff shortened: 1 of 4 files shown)


pylint/config/__init__.py
```diff
@@ -77,12 +79,12 @@
             )
 
 
-def _get_pdata_path(base_name, recurs):
+def _get_pdata_path(base_name: str, recurs: int) -> pathlib.Path:
     base_name = base_name.replace(os.sep, "_")
-    return os.path.join(PYLINT_HOME, f"{base_name}{recurs}.stats")
+    return pathlib.Path(PYLINT_HOME) / f"{base_name}{recurs}.stats"
```
Pierre-Sassoulas, line 84: We changed the typing here, right ?

Pierre-Sassoulas (changes requested):

pylint/config/__init__.py
DanielNoord: Yes, but it is a private function only used twice in this file. I thought I would conform to your `Path-love` 😄

Pierre-Sassoulas (approved):

PR #6425, opened 2022-04-21
Add typing to methods of ``PyLinter`` and related functions

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/__init__.py
+++ b/pylint/checkers/__init__.py
@@ -44,6 +44,7 @@ messages nor reports. XXX not true, emit a 07 report !
 from __future__ import annotations
 
 import sys
+from typing import TYPE_CHECKING
 
 from pylint.checkers.base_checker import (
     BaseChecker,
@@ -58,6 +59,9 @@ if sys.version_info >= (3, 8):
     from typing import Literal
 else:
     from typing_extensions import Literal
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
 
 
 def table_lines_from_stats(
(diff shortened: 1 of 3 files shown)


pylint/lint/pylinter.py
```diff
@@ -460,10 +460,8 @@ def enable_fail_on_messages(self) -> None:
                         # message starts with a category value, flag (but do not enable) it
                         self.fail_on_symbols.append(msg.symbol)
 
-    def any_fail_on_issues(self):
-        return self.stats and any(
-            x in self.fail_on_symbols for x in self.stats.by_msg.keys()
-        )
+    def any_fail_on_issues(self) -> bool:
+        return any(x in self.fail_on_symbols for x in self.stats.by_msg.keys())
```
Pierre-Sassoulas, line 464: Is self.stats always initialized now (even for lib maintainer) ?

pylint/checkers/__init__.py
```diff
@@ -128,7 +132,7 @@ def table_lines_from_stats(
     return lines
 
 
-def initialize(linter):
+def initialize(linter: PyLinter) -> None:
     """Initialize linter with checkers in this package."""
     register_plugins(linter, __path__[0])
```
Pierre-Sassoulas, line 137: Not related to the current MR but it looks a little too magical, maybe we should use hard coded values ?

pylint/lint/pylinter.py
DanielNoord: Yes 😄 

Finally some benefits from may refactors 😄

pylint/checkers/__init__.py
DanielNoord: ?

It works kinda nicely as we don't need to add any new checkers when we add them.

Pierre-Sassoulas: Well, I'm saying that because I don't understand how it works 😄

Pierre-Sassoulas (approved):

PR #6428, opened 2022-04-21
Finish typing of ``pylint/lint/pylinter.py``

DanielNoord:

## Type of Changes

## Description

Blocked by #6427.

This finished typing of `PyLinter`. The whole file now passes with ``--strict`` 🎉

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -65,12 +65,19 @@ from pylint.utils.pragma_parser import (
 )
 
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import Literal, Protocol
 else:
-    from typing_extensions import Literal
+    from typing_extensions import Literal, Protocol
 
 
 MANAGER = astroid.MANAGER
+
+
+class GetAstProtocol(Protocol):
+    def __call__(
+        self, filepath: str, modname: str, data: str | None = None
+    ) -> nodes.Module:
+        ...
 
 
 def _read_stdin() -> str:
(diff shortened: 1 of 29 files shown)


pylint/lint/pylinter.py
```diff
@@ -92,7 +92,9 @@ def _load_reporter_by_class(reporter_class: str) -> type[BaseReporter]:
     module_part = astroid.modutils.get_module_part(qname)
     module = astroid.modutils.load_module_from_name(module_part)
     class_name = qname.split(".")[-1]
-    return getattr(module, class_name)
+    klass = getattr(module, class_name)
+    assert issubclass(klass, BaseReporter)
```
Pierre-Sassoulas, line 96: I think this is a useful assert, as in fail fast if something is wrong. But I think it could be reached by users with a bad value for reporter in the configuration ? I.e. with:

```
[REPORTS]

# Set the output format. Available formats are text, parseable, colorized, msvs
# (visual studio) and html. You can also give a reporter class, eg
# mypackage.mymodule.MyReporterClass.
output-format=mypackage.mymodule.MyReporterClass
```
I would expect no crash and a configuration error if we're after the parsing, or an exit 2 during argument parsing.

Pierre-Sassoulas (changes requested):

pylint/lint/pylinter.py
DanielNoord: See https://github.com/PyCQA/pylint/pull/6428/commits/aeacd5c5684d51f5635f76885c83e4502737aa4c

We already raise an exception for incorrect reporters apparently. Just needed to catch this exception as well.

Pierre-Sassoulas: Maybe we can add a functional config test for good measure 😄 ? One class that do not exists and one class that is not a Reporter ?

Pierre-Sassoulas (approved):
Nice, the new tests are pretty elegant.

PR #6430, opened 2022-04-21
Finish typing of ``pylint/testutils``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/testutils/checker_test_case.py
+++ b/pylint/testutils/checker_test_case.py
@@ -11,7 +11,6 @@ from typing import Any
 
 from astroid import nodes
 
-from pylint.checkers.base_checker import BaseChecker
 from pylint.constants import PY38_PLUS
 from pylint.testutils.global_test_linter import linter
 from pylint.testutils.output_line import MessageTest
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (approved):
👌

Pierre-Sassoulas:
The pre-commit mypy error is a real one though.

DanielNoord:
> The pre-commit mypy error is a real one though.

hmm, didn't get that locally. Let me check!

DanielNoord:
Ah, that's quite annoying (and might become problematic for non test code as well).
Because `CheckerTestCase.CHECKER_CLASS` is now correctly typed as `BaseChecker` and not as `None` we now get warnings about certain methods not being defined. `BaseChecker` does not have all `visit_...` methods.

Pierre-Sassoulas:
I think the base problem is hard to solve as we need to find a way to make the visitor pattern understood by type checker. I found this (mostly unrelated) issue when searching or clue on how to do that:  https://github.com/gvanrossum/patma/issues/137 (It warms my heart to see that G Van Rossum himself  and python core devs miss pylint and flake8 when implementing new syntaxes ❤️)

PR #6433, opened 2022-04-21
Small typing fixes

Pierre-Sassoulas:
## Type of Changes

## Description

Add some typing detected when working on something else.

### DIFF
--- a/pylint/extensions/private_import.py
+++ b/pylint/extensions/private_import.py
@@ -112,7 +112,7 @@ class PrivateImportChecker(BaseChecker):
         )
 
     def _get_type_annotation_names(
-        self, node: nodes.Import, names: list[str]
+        self, node: nodes.Import | nodes.ImportFrom, names: list[str]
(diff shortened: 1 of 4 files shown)


pylint/extensions/private_import.py
```diff
@@ -185,7 +185,7 @@ def _populate_type_annotations_function(
 
     def _populate_type_annotations_annotation(
         self,
-        node: nodes.Attribute | nodes.Subscript | nodes.Name,
+        node: nodes.Attribute | nodes.Subscript | nodes.Name | None,
```
DanielNoord, line 188: @Pierre-Sassoulas Are you sure this is correct? I think we get `None` because of incorrect typing in `astroid`. Looking at the code it should never be `None`.

Pierre-Sassoulas: None is handled correctly inside the function though. if we have None we return None.

PR #6436, opened 2022-04-22
Remove unused ci settings

cdce8p:
## Description
* `fetch-depth: 0` was added a while ago to support `setuptools_scm`. We have since moved away from it, thus the default of `1` is enough. It's no longer necessary to fetch the full repo history.
* `python-key` output for windows and pypy. Those became unnecessary when we merged the prepare and test jobs. Only the `python-key` fro `tests-linux` and `prepare-base` are still used in different jobs.

Astroid PR: https://github.com/PyCQA/astroid/pull/1524

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -28,8 +28,6 @@ jobs:
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.1
-        with:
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
Thank you, the fetch depth change is going to make the pipeline fast for sure !

PR #6439, opened 2022-04-22
Finish typing of ``./pylint`` directory

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/epylint.py
+++ b/pylint/epylint.py
@@ -45,10 +45,10 @@ import sys
 from collections.abc import Sequence
 from io import StringIO
 from subprocess import PIPE, Popen
-from typing import NoReturn
+from typing import NoReturn, TextIO
 
 
-def _get_env():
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
👌

pylint/epylint.py
```diff
@@ -117,7 +123,32 @@ def lint(filename, options=()):
         return process.returncode
 
 
-def py_run(command_options="", return_std=False, stdout=None, stderr=None):
+@overload
+def py_run(
+    command_options: str = "",
+    return_std: Literal[False] = False,
```
cdce8p, line 129: Usually overloads should **not** use the explicit default value. Instead use `str = ...` or `Literal[False] = ...`.

DanielNoord: I'll do a follow-up!

PR #6442, opened 2022-04-22
Fixes #5533

### LINKED ISSUE #5533: false positive  for numpy ufunc: bad-string-format-type
### Bug description

```
from numpy import radians, degrees #  any ufunc will do
  
a=radians(90.0)
print ( "%6.3f"%( degrees(a),))
print ( "%6.3f"% degrees(a))

# ==>  [E1307(bad-string-format-type)] Argument '.ndarray' does not match format type 'f'

# even though the ufunc returns a float-compat arg in this case

# Note: using f-strings this is accepted:

print ( f"{degrees(a):6.3f}" )
```

~                                       

### Configuration

_No response_

### Command used

```shell
pylint check_pylint.py
```

### Pylint output

```shell
************* Module check_pylint
check_pylint.py:4: [E1307(bad-string-format-type), ] Argument '.ndarray' does not match format type 'f'
```
```


### Expected behavior

No warning as the ufunc result is usable in this case

Note: As can be seen, this warning is only displayed if passing a list to %-formatting.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0]
```

### OS / Environment

ubuntu 20.04

### Additional dependencies

_No response_

mccaidan:
## Type of Changes

## Description

Closes #5533

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,8 @@ What's New in Pylint 2.14.0?
 ============================
 Release date: TBA
 
+
+* Fixed a false positive when a numpy ufunc returns a scalar in % string formatting.
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -11,6 +11,7 @@ Summary -- Release highlights
 
 New checkers
 ============
+* Fixed a false positive when a numpy ufunc returns a scalar in % string formatting.
 
 * Added new checker ``typevar-name-mismatch``: TypeVar must be assigned to a variable with the same name as its name argument.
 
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas:
The issue here is that we do not install numpy during our tests.

Pierre-Sassoulas:
Installing libs that are not pure python is a rabbit hole. I think those kind if libs need a specific pylint plugin long term. Right now we have brains in astroid but it's not perfect (especially on multiple versions with breaking change) and take a lot of maintener ressources. It would also take a lot of CI ressources to install them in pylint main CI.

clavedeluna:
Makes sense, so... what's the action item here?

Pierre-Sassoulas:
I'm closing for staleness. It seems next step were already discussed in the issue : https://github.com/PyCQA/pylint/issues/5533#issuecomment-1258601202

PR #6443, opened 2022-04-23
Raise ``implicit-str-concat`` on call

### LINKED ISSUE #6441: False negatives for implicit string concatenation
### Current problem

The following code:
```python
print("Lorem ipsum " "dolor sit amet")
```
Could be rewritten:
```python
print("Lorem ipsum dolor sit amet")
```

But 
```python
print(
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit,"
    " sed do eiusmod tempor incididunt ut labore et dolore "
    "magna aliqua. Ut enim ad minim veniam, quis nostrud "
    "exercitation ullamco laboris nisi ut aliquip ex ea "
)
```
Should not, because the implicit string concatenation is not useless as removing it would make the line go over the max line length.

### Desired solution

Raise ``implicit-str-concat`` in this situation.

### Additional context

https://github.com/PyCQA/astroid/pull/1525#discussion_r856578221

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #6441

### DIFF
--- a/tests/functional/i/implicit/implicit_str_concat_multiline.py
+++ b/tests/functional/i/implicit/implicit_str_concat_multiline.py
@@ -2,3 +2,10 @@
 
 TEST_TUPLE = ('a', 'b'  # [implicit-str-concat]
               'c')
+
+print(
+    "Lorem ipsum dolor sit amet, consectetur adipiscing elit," # [implicit-str-concat]
+    " sed do eiusmod tempor incididunt ut labore et dolore "
+    "magna aliqua. Ut enim ad minim veniam, quis nostrud "
+    "exercitation ullamco laboris nisi ut aliquip ex ea "
+)
--- a/ChangeLog
+++ b/ChangeLog
@@ -92,6 +92,10 @@ Release date: TBA
 
   Ref #5392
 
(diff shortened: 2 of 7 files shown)


tests/functional/i/implicit/implicit_str_concat_multiline.py
```diff
@@ -2,3 +2,10 @@
 
 TEST_TUPLE = ('a', 'b'  # [implicit-str-concat]
               'c')
+
+print(
+    "Lorem ipsum dolor sit amet, consectetur adipiscing elit," # [implicit-str-concat]
```
DanielNoord, line 7: Why is this bad?

Pierre-Sassoulas: In this file the [``check-str-concat-over-line-jump``](https://github.com/PyCQA/pylint/blob/main/pylintrc#L556) option is set to yes.

Pierre-Sassoulas: The default option is no, the other file shows it's not raised by default see https://github.com/PyCQA/pylint/pull/6443/files#diff-2248542e115bc4d2f482c95ab3aba32f207108b89d18abede5c62e18db5e553cR36

PR #6445, opened 2022-04-23
Fix `AstCallbackMethod` typing

cdce8p:
## Description
Fix typing for `AstCallbackMethod` used in `check_messages` and `only_required_for_messages`. Generic type aliases need a subscript. Otherwise the TypeVar is interpreted as `Any`. Furthermore, the second parameter type also needs to be a TypeVar. That's because callback parameter are contravariant. I.e. if it's defined as `nodes.NodeNG`, we wouldn't be able to assign `nodes.Dict` to it.

Also renamed `T_Node` to `_NodeT` which better matches the new naming style. _`invalid-name` messages are disabled for pylint, which is why it didn't complain previously._ Will do a followup to rename the remaining TypeVars.

/CC: @DanielNoord

### DIFF
--- a/pylint/typing.py
+++ b/pylint/typing.py
@@ -18,7 +18,6 @@ from typing import (
     Pattern,
     Tuple,
     Type,
-    TypeVar,
     Union,
 )
 
@@ -30,7 +29,6 @@ else:
 if TYPE_CHECKING:
     from astroid import nodes
 
-    from pylint.checkers import BaseChecker
     from pylint.config.callback_actions import _CallbackAction
     from pylint.reporters.ureports.nodes import Section
     from pylint.utils import LinterStats
@@ -116,10 +114,6 @@ Options = Tuple[Tuple[str, OptionDict], ...]
 
 AstCallback = Callable[["nodes.NodeNG"], None]
 """Callable representing a visit or leave function."""
-
-CheckerT_co = TypeVar("CheckerT_co", bound="BaseChecker", covariant=True)
-AstCallbackMethod = Callable[[CheckerT_co, "nodes.NodeNG"], None]
-"""Callable representing a visit or leave method."""
 
 ReportsCallable = Callable[["Section", "LinterStats", Optional["LinterStats"]], None]
 """Callable to create a report."""
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -12,10 +12,10 @@ import numbers
 import re
 import string
 import warnings
-from collections.abc import Callable, Iterable
+from collections.abc import Iterable
 from functools import lru_cache, partial
 from re import Match
-from typing import TypeVar
+from typing import TYPE_CHECKING, Callable, TypeVar
 
 import _string
 import astroid.objects
@@ -23,9 +23,13 @@ from astroid import TooManyLevelsError, nodes
 from astroid.context import InferenceContext
 
 from pylint.constants import TYPING_TYPE_CHECKS_GUARDS
-from pylint.typing import AstCallbackMethod
 
-T_Node = TypeVar("T_Node", bound=nodes.NodeNG)
+if TYPE_CHECKING:
+    from pylint.checkers import BaseChecker
+
+_NodeT = TypeVar("_NodeT", bound=nodes.NodeNG)
+_CheckerT = TypeVar("_CheckerT", bound="BaseChecker")
+AstCallbackMethod = Callable[[_CheckerT, _NodeT], None]
 
 COMP_NODE_TYPES = (
     nodes.ListComp,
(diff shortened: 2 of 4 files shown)


pylint/typing.py
```diff
@@ -113,7 +111,3 @@ class ManagedMessage(NamedTuple):
 
 AstCallback = Callable[["nodes.NodeNG"], None]
```
DanielNoord, line 112: Should this also be updated? Either way, I think it might be good to keep it in this file as we are likely to need to in other places.

DanielNoord (commented):
Thanks @cdce8p!

I thought I used these correctly, but perhaps should have asked for a review from you. Strange how this does pass with ``mypy --strict``.

Pierre-Sassoulas (approved):
LGTM.

pylint/typing.py
cdce8p: This also doesn't work, unfortunately. Just a short example
```py
from typing import Callable

from astroid import nodes
from pylint.extensions.typing import TypingChecker

AstCallback = Callable[[nodes.NodeNG], None]

var: list[AstCallback] = []

var.append(TypingChecker(None).visit_name)
```
At least pyright complains here since we would assign `Callable[[nodes.Name], None]` to `list[Callable[[nodes.NodeNG], None]]` which isn't possible.

Technically the parameter type for `AstCallback` would need to be a Union of all node subclasses. I don't think that is practical in this case though. Might be best to just use `Any` here.

DanielNoord: Hm, but normally if you type something to receive a parent class it should also be able to accept a subclass right? As in `BaseChecker` works for all methods/functions that accept any subclass of it. So why wouldn't that work here?

cdce8p: Callable parameter are contravariant, thus the subclass relationship is inverted. A good explainer
https://blog.daftcode.pl/covariance-contravariance-and-invariance-the-ultimate-python-guide-8fabc0c24278

cdce8p: > Ah, that explains it quite well. But wouldn't the same trick with a bounded `TypeVar` work?

TypeVars only really work if you use them at least twice. If not Pylance will even emit a warning
```
Type variable "..." has no meaning in this context
```

```py
from typing import Callable, TypeVar

class GrandParent: ...
class Parent(GrandParent): ...
class Child(Parent): ...

var: list[Callable[[Parent], None]] = []

def func_grandparent(x: GrandParent) -> None: ...
def func_parent(x: Parent) -> None: ...
def func_child(x: Child) -> None: ...

var.append(func_grandparent)
var.append(func_parent)

var.append(func_child)  # error


_ParentT = TypeVar("_ParentT", bound=Parent)
var2: list[Callable[[_ParentT], None]] = []
var2.append(func_child)  # error
```

DanielNoord: How do you invoke `pyright`. Is there a strictness flag I might be missing? I get:
```console
❯ pip install pyright
Requirement already satisfied: pyright in /Users/daniel/.pyenv/versions/3.10.0/envs/pylint-3.10.0/lib/python3.10/site-packages (1.1.244)
Requirement already satisfied: nodeenv>=1.6.0 in /Users/daniel/.pyenv/versions/3.10.0/envs/pylint-3.10.0/lib/python3.10/site-packages (from pyright) (1.6.0)
WARNING: You are using pip version 22.0.3; however, version 22.0.4 is available.
You should consider upgrading via the '/Users/daniel/.pyenv/versions/3.10.0/envs/pylint-3.10.0/bin/python3.10 -m pip install --upgrade pip' command.
❯ cat test.py
from typing import Callable

from astroid import nodes

from pylint.extensions.typing import TypingChecker

AstCallback = Callable[[nodes.NodeNG], None]

var: list[AstCallback] = []

var.append(TypingChecker(PyLinter()).visit_name)
❯ pyright test.py
No configuration file found.
No pyproject.toml file found.
stubPath /Users/daniel/DocumentenLaptop/Programming/Github/pylint/typings is not a valid directory.
Assuming Python platform Darwin
Searching for source files
Found 1 source file
0 errors, 0 warnings, 0 inform
```

cdce8p: > How do you invoke `pyright`. Is there a strictness flag I might be missing?

`astroid` doesn't have a `py.typed` file yet. Thus all imports are basically ignored / set to `Any`. Try adding a `pyrightconfig.json` with
```json
{
    "typeCheckingMode": "strict",
    "useLibraryCodeForTypes": true,
}
```

If you use VSCode, you could also set `"python.analysis.typeCheckingMode": "strict"`.

```
test.py:3:6 - error: Stub file not found for "astroid" (reportMissingTypeStubs)
test.py:12:12 - error: Argument of type "(Name) -> None" cannot be assigned to parameter "__object" of type "AstCallback" in function "append"
    Type "(Name) -> None" cannot be assigned to type "AstCallback"
      Parameter 1: type "NodeNG" cannot be assigned to type "Name"
        "NodeNG" is incompatible with "Name" (reportGeneralTypeIssues)
```

PR #6447, opened 2022-04-23
Remove explicit default values from overloads

DanielNoord:

## Type of Changes

## Description

See https://github.com/PyCQA/pylint/pull/6439#discussion_r856940003

/CC @cdce8p

### DIFF
--- a/pylint/epylint.py
+++ b/pylint/epylint.py
@@ -125,10 +125,10 @@ def lint(filename: str, options: Sequence[str] = ()) -> int:
 
 @overload
 def py_run(
-    command_options: str = "",
-    return_std: Literal[False] = False,
-    stdout: TextIO | int | None = None,
-    stderr: TextIO | int | None = None,
+    command_options: str = ...,
+    return_std: Literal[False] = ...,
+    stdout: TextIO | int | None = ...,
+    stderr: TextIO | int | None = ...,
 ) -> None:
     ...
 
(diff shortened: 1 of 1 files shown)


cdce8p (commented):
Thanks for the quick followup! Could you take a look at other overloads, too? Saw that a few had default values.

I didn't explain the reason earlier. For type checkers it's just important to know if an overload parameter has a default value. As long as it's of a valid type, it's fine. By removing the explicit default value from the overload, we can change the value much more easily later. We wouldn't need to go through all overloads for it, just change it in the function definition.

DanielNoord:
> Thanks for the quick followup! Could you take a look at other overloads, too? Saw that a few had default values.
> 
> I didn't explain the reason earlier. For type checkers it's just important to know if an overload parameter has a default value. As long as it's of a valid type, it's fine. By removing the explicit default value from the overload, we can change the value much more easily later. We wouldn't need to go through all overloads for it, just change it in the function definition.

I'll draft this and do a full look later.

Pierre-Sassoulas (approved):

pylint/reporters/text.py
```diff
@@ -252,16 +252,16 @@ class ColorizedTextReporter(TextReporter):
...
     @overload
     def __init__(
         self,
-        output: TextIO | None = None,
-        color_mapping: dict[str, tuple[str | None, str]] | None = None,
+        output: TextIO | None = ...,
+        color_mapping: dict[str, tuple[str | None, str]] | None = ...,
     ) -> None:
         # Remove for pylint 3.0
         ...
```
cdce8p, line 267: These overloads don't add any more information.

cdce8p (commented):
Thanks @DanielNoord! Just a small comment

cdce8p (approved):

PR #6448, opened 2022-04-23
Move `no-self-use` to optional extension

### LINKED ISSUE #5502: Move `no-self-use` to optional checker
Related #5496

I recently came across code like this
```py
class Parent:
    @staticmethod
    def get_email() -> str:
        return "test@test.com"

class Child(Parent):
    def __init__(self, email: str) -> None:
        self.email = email

    def get_email(self) -> str:
        return self.email


def func(x: Parent):
    print(x.get_email())
```

Although `get_email` is only ever called on an instance, the author choose to add the `staticmethod` decorator. I believe, just so pylint wouldn't complain about `no-self-use`.

With #5412 this will however now emit a `arguments-differ` warning on `Child.get_email` (with and without `staticmethod`).

**Proposal**
I would like to propose moving `no-self-use` to an optional checker.

**Reason**
Although this check might be helpful in some instances, usually not using `self` is no indication that a method should be `static`. As shown in the example, sometimes this leads to devs writing worse code just to satisfy pylint instead of adding `pylint: disable`.

cdce8p:
## Description
Create new optional extension for `no-self-use` and move check.

Closes #5502


Pierre-Sassoulas (approved):
Great 👍 The fact that we had so much disable for it in our own code, is proof that it should not be enabled by default. I like that you actually made a standalone checker, independent and fully typed.

Pierre-Sassoulas (changes requested):
It seems like there's no functional test specific to this check, and the coverage is low for some part. Would you mind recreating a covering functional test ?

doc/whatsnew/2.14.rst
```diff
@@ -77,6 +77,10 @@ Removed checkers
 
   Closes #4683
 
+* Moved ``no-self-use`` check to optional extension.
```
Pierre-Sassoulas, line 80:
```suggestion
* Moved ``no-self-use`` check to optional extension. You now need to explicitly enable this check using
    ``load-plugins=pylint.extensions.no_self_use``.
```

ChangeLog
```diff
@@ -287,6 +287,10 @@ Release date: TBA
 
   Relates to #6493
 
+* Created ``NoSelfUseChecker`` extension and moved the ``no-self-use`` check.
```
Pierre-Sassoulas, line 290:
```suggestion
* Created ``NoSelfUseChecker`` extension and moved the ``no-self-use`` check.
 You now need to explicitly enable this check using ``load-plugins=pylint.extensions.no_self_use``.
```

pylint/extensions/no_self_use.py
```diff
@@ -0,0 +1,110 @@
...
+
+class NoSelfUseChecker(BaseChecker):
+
+    name = "no_self_use"
+    msgs = {
+        "R6301": (
+            "Method could be a function",
+            "no-self-use",
+            "Used when a method doesn't use its bound instance, and so could "
+            "be written as a function.",
```
Pierre-Sassoulas, line 33: We need to add an ``old_names`` here.

tests/functional/ext/no_self_use/no_self_use.py
```diff
@@ -83,3 +84,76 @@ def count(self):
...
+
+class B(A):
+    def get_memo(self, obj):
+        return super().get(obj)
+
+
+class C:
+    def a(self, /):  # [no-self-use]
+        ...
+
```
Pierre-Sassoulas, line 104:
```suggestion

    def b(self, /):  # pylint: disable=R0201
        ...
 
```

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas (approved):
👍

cdce8p:
> For some reason `duplicate-code` gets triggered on this branch and on some other PRs but not on `main` 😓

Saw that one too. A bit strange tbh. I can't reproduce the warnings locally although they are valid. I've added a disabled comment for the one instance directly caused by this change. This should allow us to move forward with it. The rest can be addressed sometime later.

cdce8p:
> @cdce8p The CI job still fails. Do you want to do a follow-up to unblock `main`?

I'll open a PR shortly. Interestingly, now `main` also fails. Honestly not sure why it hasn't previously.
https://github.com/PyCQA/pylint/runs/6310323884?check_suite_focus=true#step:9:34

PR #6434, opened 2022-04-22
Add multiplicity association to pyreverse

titilambert:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |

## Description

This PR simply adds the support of multiplicity association in pyreverse.

Closes #XXX

PR #6449, opened 2022-04-24
Functional tests for ``pyreverse``

DudeNr33:

## Type of Changes

## Description

I would like to add a more flexible construct for functional tests of ``pyreverse``.
PR #6434 once again showed me that the current test suite is not ideal. 
In general I want to see how the new feature looks like in a final diagram, so I have to ask the contributor to include new, suitable code in `tests/data`. 
This in turn means she/he needs to update **all** generated output files for every possible format, which is cumbersome.
It also does not necessarily makes sense if a PR does not touch the ``Printer`` classes, as it would be enough to check the output with one format, and not for all. 

Moreover, the test data becomes really bloated, and if the current tests fail you have to dig deep to see what feature of ``pyreverse`` is actually broken. 

This PR is currently an early WIP, so you can give feedback on the general direction.
Here is what's on my mind how to approach this:

* Create a folder structure in the form of `tests/pyreverse/functional/<diagram_type>/<feature>`, where `feature` are things like different types of relationships between two nodes. 
* Use MermaidJS as default output format. The reason is that it can be rendered natively in Github, which should make reviewing functional tests a lot easier.
* ~(Still ToDo, maybe at a later point when we first see the necessity for it)~ allow for optional configuration files analogous to the normal functional tests, to be able to choose a different output format for example.

Let me know what you think.

### DIFF
--- a/tests/pyreverse/test_pyreverse_functional.py
+++ b/tests/pyreverse/test_pyreverse_functional.py
@@ -1,6 +1,7 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
 from pathlib import Path
 
 import pytest
--- a/tests/pyreverse/test_pyreverse_functional.py
+++ b/tests/pyreverse/test_pyreverse_functional.py
@@ -1,0 +1,51 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+from pathlib import Path
+from typing import NamedTuple
+
+import pytest
+from py._path.local import LocalPath  # type: ignore[import]
+
+from pylint.pyreverse.main import Run
+
+
+class FunctionalPyreverseTestfile(NamedTuple):
+    """Named tuple containing the test file and the expected output."""
+
+    source: Path
+    expectation: Path
+
+
+def get_functional_test_files(
+    root_directory: Path,
+) -> list[FunctionalPyreverseTestfile]:
+    """Get all functional test files from the given directory."""
+    return [
+        FunctionalPyreverseTestfile(
+            source=test_file, expectation=test_file.with_suffix(".mmd")
+        )
+        for test_file in root_directory.rglob("*.py")
+    ]
+
+
+FUNCTIONAL_DIR = Path(__file__).parent / "functional"
+CLASS_DIAGRAM_TESTS = get_functional_test_files(FUNCTIONAL_DIR / "class_diagrams")
+CLASS_DIAGRAM_TEST_IDS = [testfile.source.stem for testfile in CLASS_DIAGRAM_TESTS]
+
+
+@pytest.mark.parametrize(
+    "testfile",
+    CLASS_DIAGRAM_TESTS,
+    ids=CLASS_DIAGRAM_TEST_IDS,
+)
+def test_class_diagrams(
+    testfile: FunctionalPyreverseTestfile, tmpdir: LocalPath
+) -> None:
+    input_file = testfile.source
+    with pytest.raises(SystemExit) as sys_exit:
+        Run(["-o", "mmd", "-d", str(tmpdir), str(input_file)])
+    assert sys_exit.value.code == 0
+    assert testfile.expectation.read_text(encoding="utf8") == Path(
+        tmpdir / "classes.mmd"
+    ).read_text(encoding="utf8")
--- a/tests/pyreverse/test_pyreverse_functional.py
+++ b/tests/pyreverse/test_pyreverse_functional.py
@@ -1,6 +1,8 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+from __future__ import annotations
+
 from pathlib import Path
 from typing import NamedTuple
 
--- a/tests/pyreverse/test_pyreverse_functional.py
+++ b/tests/pyreverse/test_pyreverse_functional.py
@@ -1,35 +1,16 @@
 # Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
-from __future__ import annotations
-
 from pathlib import Path
-from typing import NamedTuple
 
 import pytest
 from py._path.local import LocalPath  # type: ignore[import]
 
 from pylint.pyreverse.main import Run
-
-
-class FunctionalPyreverseTestfile(NamedTuple):
-    """Named tuple containing the test file and the expected output."""
-
-    source: Path
-    expectation: Path
-
-
-def get_functional_test_files(
-    root_directory: Path,
-) -> list[FunctionalPyreverseTestfile]:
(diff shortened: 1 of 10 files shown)


Pierre-Sassoulas (commented):
I like this idea a lot. Functional test using file like for the functional test and config functional tests are easier to use and understand than python tests with less boilerplate. Maybe we could add both mermaidjs and plantuml in order to catch plantuml problems too, but maybe it's overkill.

DudeNr33:
Dang it, I always forgot to put the labels and milestones on my tickets and PRs. 😄  

I'd say just go for MermaidJS by default. We have a well-defined interface for the `Printer` classes, and they only do the actual translation of our ``*Diagram`` objects to the required syntax of the output format. It would also double the execution time of the tests if we generate two output formats by default.

If a change actually modifies the code of the `Printer` classes, we can still create both output formats for this specific test.

DudeNr33:
Seems like we have some weird caching error in Python 3.7? I can reproduce the failure of the CI when running the whole test suite, but not if I just run the failing tests themselves:

```parsing /Users/andreas/programming/forks/pylint/tests/pyreverse/functional/class_diagrams/attributes/instance_attributes.py...
________________________________________________________________________________ test_class_diagrams[simple_inheritance] _________________________________________________________________________________

testfile = FunctionalPyreverseTestfile(source=PosixPath('/Users/andreas/programming/forks/pylint/tests/pyreverse/functional/class.../Users/andreas/programming/forks/pylint/tests/pyreverse/functional/class_diagrams/inheritance/simple_inheritance.mmd'))
tmpdir = local('/private/var/folders/2d/6hz_h_xj70nd_1y0b8g7fr480000gn/T/pytest-of-andreas/pytest-172/test_class_diagrams_simple_inh0')

    @pytest.mark.parametrize(
        "testfile",
        CLASS_DIAGRAM_TESTS,
        ids=CLASS_DIAGRAM_TEST_IDS,
    )
    def test_class_diagrams(
        testfile: FunctionalPyreverseTestfile, tmpdir: LocalPath
    ) -> None:
        input_file = testfile.source
        with pytest.raises(SystemExit) as sys_exit:
>           Run(["-o", "mmd", "-d", str(tmpdir), str(input_file)])

tests/pyreverse/test_pyreverse_functional.py:49: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pylint/pyreverse/main.py:228: in __init__
    sys.exit(self.run(args))
pylint/pyreverse/main.py:243: in run
    diadefs = handler.get_diadefs(project, linker)
pylint/pyreverse/diadefslib.py:219: in get_diadefs
    diagrams.append(generator.class_diagram(project, klass))
pylint/pyreverse/diadefslib.py:187: in class_diagram
    klass = next(module.ilookup(klass))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

stmts = (), context = <astroid.context.InferenceContext object at 0x15e132be8>, frame = <Module.builtins l.0 at 0x106720208>

    def _infer_stmts(stmts, context, frame=None):
        """Return an iterator on statements inferred by each statement in *stmts*."""
        inferred = False
        if context is not None:
            name = context.lookupname
            context = context.clone()
        else:
            name = None
            context = InferenceContext()
    
        for stmt in stmts:
            if stmt is Uninferable:
                yield stmt
                inferred = True
                continue
            context.lookupname = stmt._infer_name(frame, name)
            try:
                for inf in stmt.infer(context=context):
                    yield inf
                    inferred = True
            except NameInferenceError:
                continue
            except InferenceError:
                yield Uninferable
                inferred = True
        if not inferred:
            raise InferenceError(
                "Inference failed for all members of {stmts!r}.",
                stmts=stmts,
                frame=frame,
>               context=context,
            )
E           astroid.exceptions.InferenceError: Inference failed for all members of ().

../../../.pyenv/versions/3.7.12/envs/pylint3712/lib/python3.7/site-packages/astroid/bases.py:147: InferenceError
```

Pierre-Sassoulas:
> Seems like we have some weird caching error in Python 3.7?

Maybe a generated file that persist between test runs ?

DudeNr33:
> > Seems like we have some weird caching error in Python 3.7?
> 
> Maybe a generated file that persist between test runs ?

I suspect a problem with the mocking in `tests/pyreverse/test_main.py` - skipping this file resolves the error.

DanielNoord:
> > > Seems like we have some weird caching error in Python 3.7?
> > 
> > 
> > Maybe a generated file that persist between test runs ?
> 
> I suspect a problem with the mocking in `tests/pyreverse/test_main.py` - skipping this file resolves the error.
> 
> Edit: could track it down to `tests/pyreverse/test_main.py::test_class_command` and the `--class` option. @DanielNoord this seems to be a regression due to the migration to `argparse`. I locally checked out the `maintenance/2.13.x` branch and copied the new functional tests from this PR and the contents of `test_main.py` from the current main, and there all tests succeed when running them together. ~I don't have a clue yet how this could be caused by the migration. Any idea?~ It looks like `self.config.classes` is somehow cached. When running together with the other tests, `self.config.classes` is still set to the values from `tests/pyreverse/test_main.py::test_class_command`.

Hmm this might be to due to how `argparse` implements the `extend` action (which I basically the their form of doing `append` from `optparse`). That being said, it shouldn't persist across runs. Do the test runs use the same `Run` class?

DudeNr33:
No, the `Run` class is freshly instantiated. I am currently debugging it. 
Somehow values passed in with `--class` end up in the `default` for this option.

DanielNoord:
> No, the `Run` class is freshly instantiated. I am currently debugging it.
> Somehow values passed in with `--class` end up in the `default` for this option.

Hm, might this be due to using `[]` as default value? Classic case of mutable default values?

Perhaps we should make the default `None` and let the action handle that case? As in, instantiate a new `list` when it encounters `None`?

DudeNr33:
> > No, the `Run` class is freshly instantiated. I am currently debugging it.
> > Somehow values passed in with `--class` end up in the `default` for this option.
> 
> Hm, might this be due to using `[]` as default value? Classic case of mutable default values?
> 
> Perhaps we should make the default `None` and let the action handle that case? As in, instantiate a new `list` when it encounters `None`?

Ha, I just tried that locally. 😄 👍  Tests for `pyreverse` are now passing, running the whole test suite now locally.
If this works I can prepare a patch MR.

DudeNr33:
I added logic for per-test configuration files.
`pyreverse` does not support reading configuration from a file, so I went with a minimalistic approach. 

The `output_formats` option can be used to specify multiple output formats at once, if some feature should be tested with more than one format. This way we do not have to copy-paste test files for each desired output format (e.g. if we add colorisation support for MermaidJS, we don't have to create a `colorized_puml.py` and `colorized_mmd.py`, but can use the existing `colorized.py`).

DudeNr33:
From my side this would be ready for review.
I'd leave out functional tests for package diagrams for now, as they require a bit different approach (as we can't use single files there but have to create test packages). 

As this is just an internal tool for testing, I don't think a ChangeLog or whatsnew entry are necessary.

tests/pyreverse/test_pyreverse_functional.py
```diff
@@ -0,0 +1,38 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+from pathlib import Path
```
DanielNoord, line 4:
```suggestion

from pathlib import Path
```

DanielNoord (approved):
Just one minor nitpick.

I don't have any knowledge about the output so I'll assume they are correct. Testutils themselves look good!

PR #6450, opened 2022-04-24
Fix handling of default value for extend action in Python 3.7

DudeNr33:

## Type of Changes

## Description

Using `default=[]` in Python 3.7 has the effect that values passed in for this parameter will be stored in the default list object. In subsequent runs (in the same process), the default value is wrongly prepopulated.

Python 3.8 and above handle this situation better, but as we have to support Python 3.7 this fix is necessary.

Ref: https://github.com/PyCQA/pylint/pull/6449#issuecomment-1107831297

This change should be covered e.g. by `tests/pyreverse/test_main.py::test_command_line_arguments_defaults[classes-expected_default1]`.

### DIFF
--- a/pylint/config/utils.py
+++ b/pylint/config/utils.py
@@ -87,7 +87,7 @@ def _convert_option_to_argument(
         return _ExtendArgument(
             flags=flags,
             action=action,
-            default=default,
+            default=[] if default is None else default,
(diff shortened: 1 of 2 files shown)


DanielNoord (approved):

DanielNoord:
Thanks @DudeNr33!

PR #6451, opened 2022-04-24
Avoid reporting superfluous-parens after an ``is not`` operator

### LINKED ISSUE #5930: `superfluous-parens` false positive for `is not` tests
### Bug description

I just got the lint `C0325: Unnecessary parens after 'not' keyword (superfluous-parens)`, so I removed the parens it referred to, but thankfully a unit test caught that these are not the same, and the parens are needed (removing them changes the logic!).

I have been able to create a minimal example that I believe reproduces this

```python
In [1]: (None is None) is not (None == 1)  # Original, pylint lints with C0325
Out[1]: True

In [1]: (None is None) is not None == 1  # Recommended by C0325
Out[1]: False
```

### Configuration

_No response_

### Command used

```shell
pylint path/to/pkg/ --max-line-length=120 --fail-under=9 --load-plugins pylint_django
```

### Pylint output

```shell
************* Module pkg.module
path/to/pkg/module.py:63:0: C0325: Unnecessary parens after 'not' keyword (superfluous-parens)
```

### Expected behavior

C0325 should not be output here.

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.6
Python 3.9.7 (default, Sep 16 2021, 13:09:58) 
[GCC 7.5.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

timmartin:
## Type of Changes

## Description

The `superfluous-parens` checker was incorrectly reporting that parentheses after an `is not` expression were superfluous. The checker seemed to be assuming that any expression prefixed with `not` was a unary expression and not part of a larger expression that might need parentheses for correctness.

Closes #5930

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -184,6 +184,10 @@ Release date: TBA
 
   Closes #4324
 
+* Avoid reporting ``superfluous-parens`` on expressions using the ``is not`` operator.
+
+  Closes #5930
+
 * Added the ``super-without-brackets`` checker, raised when a super call is missing its brackets.
 
   Closes #4008
--- a/doc/whatsnew/2.13.rst
+++ b/doc/whatsnew/2.13.rst
@@ -611,6 +611,10 @@ Other Changes
 
   Closes #5931
 
+* Avoid reporting ``superfluous-parens`` on expressions using the ``is not`` operator.
+
+  Closes #5930
+
 * Fix a crash when linting a file that passes an integer ``mode=`` to
   ``open``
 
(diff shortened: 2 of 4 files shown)


ChangeLog
```diff
@@ -184,6 +184,10 @@ Release date: TBA
 
   Closes #4324
 
+* Avoid reporting ``superfluous-parens`` on expressions using the ``is not`` operator.
+
+  Closes #5930
+
```
jacobtylerwalls, line 190: Even though the top of this doc says to port "new features" to what's new/*.rst, we've been doing so for anything user-facing. So a copy-paste to 2.13.rst would be great.

Pierre-Sassoulas (approved):
LGTM

jacobtylerwalls (approved):

PR #6452, opened 2022-04-24
Use list comprehension when possible for better performance

Pierre-Sassoulas:
## Type of Changes

## Description

Reminder following the closing of #6078

### DIFF
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -433,9 +433,7 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
         self._lines = {}
         self._visited_lines = {}
         token_handlers: dict[str, Callable[[list[tokenize.TokenInfo], int], None]] = {
-            token: handler
-            for tokens, handler in ((_KEYWORD_TOKENS, self._check_keyword_parentheses),)
-            for token in tokens
+            token: self._check_keyword_parentheses for token in _KEYWORD_TOKENS
         }
         self._last_line_ending = None
         last_blank_line_num = 0
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -14,6 +14,7 @@ Some parts of the process_token method is based from The Tab Nanny std module.
 from __future__ import annotations
 
 import tokenize
+from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


pylint/checkers/format.py
```diff
@@ -418,12 +418,12 @@ def _check_keyword_parentheses(
                         self._check_keyword_parentheses(tokens[i:], 0)
                     return
 
-    def _prepare_token_dispatcher(self):
-        dispatch = {}
-        for tokens, handler in ((_KEYWORD_TOKENS, self._check_keyword_parentheses),):
-            for token in tokens:
-                dispatch[token] = handler
-        return dispatch
+    def _prepare_token_dispatcher(self) -> dict[str, Any]:
```
DanielNoord, line 421: I'd say let's leave the typing out if we don't really add anything. Easier to see that we still need to do this that way.

DanielNoord (approved):

pylint/checkers/format.py
Pierre-Sassoulas: I agree, but we're adding the dict with key that are strings information though ?

DanielNoord: Hmm, yeah I guess. Can't we solve the `Any`?

Pierre-Sassoulas: I don't understand what ``self._check_keyword_parentheses`` returns. There's no return, so I'm under the impression handler is always None, but I tried refactoring this and it started to fail.

Pierre-Sassoulas: Wait, it's ``self._check_keyword_parentheses`` not the result of the call.

Pierre-Sassoulas: I added the typing and also refactored the comprehension so it's less convoluted in https://github.com/PyCQA/pylint/pull/6452/commits/8882806e46a92c3930e8f3106ad47b4e9522a177

Pierre-Sassoulas: I removed the function, the attribute it was creating and the typing entirely. Taking the time to really understand the code is worth it sometime...

DanielNoord (approved):

PR #6456, opened 2022-04-25
Add typing to ``BaseChecker`` ``msgs`` and ``reports``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -39,6 +39,7 @@ from pylint.checkers.utils import (
     uninferable_final_decorators,
 )
 from pylint.interfaces import HIGH, INFERENCE
+from pylint.typing import MessageDefinitionTuple
 
 if sys.version_info >= (3, 8):
     from functools import cached_property
@@ -428,7 +429,9 @@ def _has_same_layout_slots(slots, assigned_value):
     return False
 
 
-MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
+MSGS: dict[
+    str, MessageDefinitionTuple
+] = {  # pylint: disable=consider-using-namedtuple-or-dataclass
     "F0202": (
         "Unable to check methods signature (%s / %s)",
         "method-check-failed",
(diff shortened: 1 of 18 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -428,7 +429,9 @@ def _has_same_layout_slots(slots, assigned_value):
     return False
 
 
-MSGS = {  # pylint: disable=consider-using-namedtuple-or-dataclass
+MSGS: dict[
```
Pierre-Sassoulas, line 432: I'd rather put that directly in the checker as ``msgs`` so the typing is implicit, what do you think ?

pylint/typing.py
```diff
@@ -117,3 +121,14 @@ class ManagedMessage(NamedTuple):
 CheckerT_co = TypeVar("CheckerT_co", bound="BaseChecker", covariant=True)
 AstCallbackMethod = Callable[[CheckerT_co, "nodes.NodeNG"], None]
 """Callable representing a visit or leave method."""
+
+ReportsCallable = Callable[["Section", "LinterStats", Optional["LinterStats"]], None]
+"""Callable to create a report."""
+
```
Pierre-Sassoulas, line 127: Thinking out loud here, but what if we introduce a class that does all this and make it compatible with the legacy ? (I.e. we handle both for a while but encourage the use of the class)

```python
class RawMessageDefinition(NamedTuple):

   """ This class look a lot like a ``MessageDefinition`` but without having to give the checker. """

   msgid: str #Or another class taking into account that it's a special string with a letter and a number
   symbol: str #Or another class taking into account that it's a special string with only letters and '-'
   ...
```

pylint/checkers/classes/class_checker.py
DanielNoord: That doesn't work as dictionaries are invariant. See https://mypy.readthedocs.io/en/stable/common_issues.html#variance

pylint/typing.py
DanielNoord: Probably need to this to complete the typing of `BaseChecker` as it uses `**definition` a couple of times.

For now, I think this is fine though. Small steps towards a complete and fully typed version.

Pierre-Sassoulas (approved):

PR #6459, opened 2022-04-25
Use an explicitly set empty pylintrc inside tests

Pierre-Sassoulas:
## Type of Changes

## Description

We don't want to use the project's  pylintrc during tests, because it means that a change in our config could break tests.
But we want to see if the changes to the default break tests. Refactor "required" to be able to add spelling checks in the codebase. It broke one (1) test :)

Tests to check that it's done correctly, is adding the following conf in pylintrc:
```ini
[MASTER]

init-hook=raise Exception("implicitly using the project's pylintrc")
```

Refactor prior to #6440

### DIFF
--- a/pylint/__pkginfo__.py
+++ b/pylint/__pkginfo__.py
@@ -2,6 +2,11 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
+"""This module exists for compatibility reasons.
+
+It's updated via tbump, do not modify.
+"""
+
 from __future__ import annotations
 
 __version__ = "2.14.0-dev0"
@@ -13,26 +18,26 @@ def get_numversion_from_version(v: str) -> tuple[int, int, int]:
     See https://github.com/PyCQA/pylint/issues/4399
     https://github.com/PyCQA/pylint/issues/4420,
     """
-    v = v.replace("pylint-", "")
-    version = []
-    for n in v.split(".")[0:3]:
+    version = v.replace("pylint-", "")
+    result_version = []
+    for number in version.split(".")[0:3]:
         try:
-            version.append(int(n))
+            result_version.append(int(number))
         except ValueError:
-            num = ""
-            for c in n:
-                if c.isdigit():
-                    num += c
+            current_number = ""
+            for char in number:
+                if char.isdigit():
+                    current_number += char
                 else:
                     break
             try:
-                version.append(int(num))
+                result_version.append(int(current_number))
             except ValueError:
-                version.append(0)
-    while len(version) != 3:
-        version.append(0)
+                result_version.append(0)
+    while len(result_version) != 3:
+        result_version.append(0)
 
-    return tuple(version)  # type: ignore[return-value] # mypy can't infer the length
+    return tuple(result_version)  # type: ignore[return-value] # mypy can't infer the length
 
 
 numversion = get_numversion_from_version(__version__)
--- a/pylint/testutils/_run.py
+++ b/pylint/testutils/_run.py
@@ -1,0 +1,45 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""Classes and functions used to mimic normal pylint runs.
+
+This module is considered private and can change at any time.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Any
+
+from pylint.lint import Run as LintRun
+from pylint.lint.run import UNUSED_PARAM_SENTINEL
+from pylint.reporters.base_reporter import BaseReporter
+from pylint.testutils.lint_module_test import PYLINTRC
+
+
+def _add_rcfile_default_pylintrc(args: list[str]) -> list[str]:
+    """Add a default pylintrc with the rcfile option in a list of pylint args."""
+    if not any("--rcfile" in arg for arg in args):
+        args.insert(0, f"--rcfile={PYLINTRC}")
+    return args
+
+
+class _Run(LintRun):
+
+    """Like Run, but we're using an explicitly set empty pylintrc.
+
+    We don't want to use the project's pylintrc during tests, because
+    it means that a change in our config could break tests.
+    But we want to see if the changes to the default break tests.
+    """
+
+    def __init__(
+        self,
+        args: Sequence[str],
+        reporter: BaseReporter | None = None,
+        exit: bool = True,  # pylint: disable=redefined-builtin
+        do_exit: Any = UNUSED_PARAM_SENTINEL,
+    ) -> None:
+        args = _add_rcfile_default_pylintrc(list(args))
+        super().__init__(args, reporter, exit, do_exit)
--- a/tests/benchmark/test_baseline_benchmarks.py
+++ b/tests/benchmark/test_baseline_benchmarks.py
@@ -15,8 +15,9 @@ import pytest
 from astroid import nodes
 
 from pylint.checkers import BaseRawFileChecker
-from pylint.lint import PyLinter, Run, check_parallel
+from pylint.lint import PyLinter, check_parallel
 from pylint.testutils import GenericTestReporter as Reporter
+from pylint.testutils._run import _Run as Run
 from pylint.typing import FileItem
 from pylint.utils import register_plugins
 
@@ -302,14 +303,9 @@ class TestEstablishBaselineBenchmarks:
         ), f"Expected no errors to be thrown: {pprint.pformat(linter.reporter.messages)}"
 
     def test_baseline_benchmark_j1_all_checks_single_file(self, benchmark):
-        """Runs a single file, with -j1, against all plug-ins.
-
-        ... that's the intent at least.
-        """
-        # Just 1 file, but all Checkers/Extensions
-        fileinfos = [self.empty_filepath]
-
-        runner = benchmark(Run, fileinfos, reporter=Reporter(), exit=False)
+        """Runs a single file, with -j1, against all checkers/Extensions."""
+        args = [self.empty_filepath, "--enable=all", "--enable-all-extensions"]
+        runner = benchmark(Run, args, reporter=Reporter(), exit=False)
         assert runner.linter.config.jobs == 1
         print("len(runner.linter._checkers)", len(runner.linter._checkers))
         assert len(runner.linter._checkers) > 1, "Should have more than 'master'"
--- a/tests/checkers/unittest_refactoring.py
+++ b/tests/checkers/unittest_refactoring.py
@@ -6,8 +6,8 @@ import os
 
 import pytest
 
-from pylint.lint import Run
 from pylint.reporters.text import TextReporter
+from pylint.testutils._run import _Run as Run
 
 PARENT_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
 REGR_DATA = os.path.join(PARENT_DIR, "regrtest_data")
@@ -16,7 +16,10 @@ REGR_DATA = os.path.join(PARENT_DIR, "regrtest_data")
 @pytest.mark.timeout(8)
 def test_process_tokens() -> None:
     with pytest.raises(SystemExit) as cm:
-        Run([os.path.join(REGR_DATA, "very_long_line.py")], reporter=TextReporter())
+        Run(
+            [os.path.join(REGR_DATA, "very_long_line.py"), "--disable=C"],
+            reporter=TextReporter(),
+        )
     assert cm.value.code == 0
 
 
@@ -28,6 +31,7 @@ def test_issue_5724() -> None:
             [
                 os.path.join(REGR_DATA, "issue_5724.py"),
                 "--enable=missing-final-newline",
+                "--disable=C",
             ],
             reporter=TextReporter(),
         )
--- a/tests/config/test_argparse_config.py
+++ b/tests/config/test_argparse_config.py
@@ -11,7 +11,7 @@ import pytest
 
 from pylint.config.arguments_manager import _ArgumentsManager
 from pylint.config.exceptions import UnrecognizedArgumentAction
-from pylint.lint import Run
+from pylint.testutils._run import _Run as Run
 
 HERE = abspath(dirname(__file__))
 REGRTEST_DATA_DIR = join(HERE, "..", "regrtest_data")
@@ -64,7 +64,7 @@ class TestDeprecationOptions:
         run = Run([EMPTY_MODULE, "--ignore=test,test_two"], exit=False)
         assert run.linter.config.ignore == ["test", "test_two"]
         assert run.linter.config.ignore == run.linter.config.black_list
-        assert run.linter.config.ignore_patterns == [re.compile("^\\.#")]
+        assert run.linter.config.ignore_patterns == (re.compile("^\\.#"),)
         assert run.linter.config.ignore_patterns == run.linter.config.black_list_re
 
 
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -11,7 +11,8 @@ import pytest
 from pytest import CaptureFixture
 
 from pylint.config.exceptions import _UnrecognizedOptionError
-from pylint.lint import Run
+from pylint.lint import Run as LintRun
+from pylint.testutils._run import _Run as Run
 from pylint.testutils.configuration_test import run_using_a_configuration_file
 
 HERE = Path(__file__).parent.absolute()
@@ -20,7 +21,7 @@ EMPTY_MODULE = REGRTEST_DATA_DIR / "empty.py"
 
 
 def check_configuration_file_reader(
-    runner: Run,
+    runner: LintRun,
     expected_disabled: set[str] | None = None,
     expected_jobs: int = 10,
     expected_reports_truthey: bool = True,
--- a/tests/lint/test_utils.py
+++ b/tests/lint/test_utils.py
@@ -7,8 +7,8 @@ from pathlib import Path, PosixPath
 
 import pytest
 
-from pylint.lint import Run
 from pylint.lint.utils import get_fatal_error_message, prepare_crash_report
+from pylint.testutils._run import _Run as Run
 
 
 def test_prepare_crash_report(tmp_path: PosixPath) -> None:
--- a/tests/lint/unittest_lint.py
+++ b/tests/lint/unittest_lint.py
@@ -32,10 +32,11 @@ from pylint.constants import (
     OLD_DEFAULT_PYLINT_HOME,
 )
 from pylint.exceptions import InvalidMessageError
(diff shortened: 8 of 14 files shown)


pylint/__pkginfo__.py
```diff
@@ -2,37 +2,42 @@
...
+It's updated via tbump, do not modify.
+"""
+
 from __future__ import annotations
 
 __version__ = "2.14.0-dev0"
 
 
-def get_numversion_from_version(v: str) -> tuple[int, int, int]:
+def get_numversion_from_version(version: str) -> tuple[int, int, int]:
```
DanielNoord, line 15: I don't think we can change the signature like that. `get_numversion_from_version(v="test")` won't work anymore.

pylint/testutils/run.py
```diff
@@ -0,0 +1,46 @@
...
+        args.insert(0, f"--rcfile={pylintrc}")
+    return args
+
+
+def add_rcfile_default_pylintrc(args: Sequence[str]) -> list[str]:
+    """Add a default pylintrc with the rcfile option in a list of pylint args."""
+    return add_rcfile_pylintrc_to_args(PYLINTRC, list(args))
+
+
+class Run(LintRun):
```
DanielNoord, line 29: You have probably explored this, but can't we just pass an empty `rcfile` whenever we want to? Overwriting `Run` feels like an overcomplicated solution. We already support ``rcfile=``, why not use that?

_This just adds more stuff to maintain._

tests/benchmark/test_baseline_benchmarks.py
```diff
@@ -302,14 +303,9 @@ def test_baseline_benchmark_j2_single_working_checker(self, benchmark):
...
-        """Runs a single file, with -j1, against all plug-ins.
-
-        ... that's the intent at least.
-        """
-        # Just 1 file, but all Checkers/Extensions
-        fileinfos = [self.empty_filepath]
-
-        runner = benchmark(Run, fileinfos, reporter=Reporter(), exit=False)
+        """Runs a single file, with -j1, against all checkers/Extensions."""
+        args = [self.empty_filepath, "--enable=all", "--enable-all-extensions"]
```
Pierre-Sassoulas, line 307: I fixed the intent of the tests here while at it.

tests/config/test_argparse_config.py
```diff
@@ -64,7 +64,7 @@ def test_old_names() -> None:
         run = Run([EMPTY_MODULE, "--ignore=test,test_two"], exit=False)
         assert run.linter.config.ignore == ["test", "test_two"]
         assert run.linter.config.ignore == run.linter.config.black_list
-        assert run.linter.config.ignore_patterns == [re.compile("^\\.#")]
+        assert list(run.linter.config.ignore_patterns) == [re.compile("^\\.#")]
```
Pierre-Sassoulas, line 67: The default value is a tuple, we have a list in our pylintrc.

tests/test_self.py
```diff
@@ -1399,17 +1396,18 @@ def test_generate_toml_config() -> None:
...
                 with warnings.catch_warnings():
                     warnings.simplefilter("ignore")
                     Run(["--generate-toml-config"])
 
-        bytes_out = BytesIO(out.getvalue().encode("utf-8"))
+        out = output_stream.getvalue()
+        bytes_out = BytesIO(out.encode("utf-8"))
         content = tomllib.load(bytes_out)
         messages = content["tool"]["pylint"]["messages control"]["disable"]
-        assert "invalid-name" in messages, out.getvalue()
```
Pierre-Sassoulas, line 1412: invalid-name is not a message disabled by default

pylint/__pkginfo__.py
Pierre-Sassoulas: Sure, but I think it's internal, it was introduced in 2.8.2 and our version and tag are clean semver at this point.

DanielNoord: Let's be conservative. With all the deprecations and changes in `2.14` we are already at risk of breaking compatibility somewhere in `2.14`... 😅

pylint/testutils/run.py
Pierre-Sassoulas: Yeah I started by passing rcfile "when needed" but after adding an init-hook to make a crash if we were using the project's pylintrc there was hundreds of failing tests. May as well modify the class that is being called by the test (in this case only the import from pylint.testutils instead of pylint.lint). Adding a new argument as a string for each tests is a lot of work (litteral hundred of tests to fix). Note that sometime I use the internal function directly.

DanielNoord: Hmm shall we make it private then? We need to deprecate one of the parameters to `Run` in `3.0` and having to do so in another class makes that more complicated. Since we don't really have anybody asking for us to add this to `testutils` we don't need to make it public.

pylint/__pkginfo__.py
Pierre-Sassoulas: 87f2cf032

pylint/testutils/run.py
Pierre-Sassoulas: 194e84215

pylint/testutils/__init__.py
```diff
@@ -19,6 +19,8 @@
     "UPDATE_OPTION",
     "UnittestLinter",
     "create_files",
+    "_Run",
```
DanielNoord, line 22:
```suggestion
```

pylint/testutils/run.py
```diff
@@ -0,0 +1,46 @@
...
+def _add_rcfile_pylintrc_to_args(pylintrc: str | Path, args: list[str]) -> list[str]:
+    """Add the rcfile option to the pylintrc path to an args list."""
+    if not any("--rcfile" in arg for arg in args):
+        args.insert(0, f"--rcfile={pylintrc}")
+    return args
+
+
+def add_rcfile_default_pylintrc(args: Sequence[str]) -> list[str]:
+    """Add a default pylintrc with the rcfile option in a list of pylint args."""
+    return _add_rcfile_pylintrc_to_args(PYLINTRC, list(args))
```
DanielNoord, line 26: Why do we need the private function?

pylint/testutils/run.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class _Run(LintRun):
+
+    """Like Run, but we're using an explicitly set empty pylintrc.
+
+    We don't want to use the project's  pylintrc during tests, because
```
DanielNoord, line 33:
```suggestion
    We don't want to use the project's pylintrc during tests, because
```

tests/benchmark/test_baseline_benchmarks.py
```diff
@@ -15,8 +15,9 @@
 from astroid import nodes
 
 from pylint.checkers import BaseRawFileChecker
-from pylint.lint import PyLinter, Run, check_parallel
+from pylint.lint import PyLinter, check_parallel
 from pylint.testutils import GenericTestReporter as Reporter
+from pylint.testutils import _Run as Run
```
DanielNoord, line 20:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/checkers/unittest_refactoring.py
```diff
@@ -6,8 +6,8 @@
 
 import pytest
 
-from pylint.lint import Run
 from pylint.reporters.text import TextReporter
+from pylint.testutils import _Run as Run
```
DanielNoord, line 10:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/checkers/unittest_refactoring.py
```diff
@@ -16,7 +16,10 @@
 @pytest.mark.timeout(8)
 def test_process_tokens() -> None:
     with pytest.raises(SystemExit) as cm:
-        Run([os.path.join(REGR_DATA, "very_long_line.py")], reporter=TextReporter())
+        Run(
+            [os.path.join(REGR_DATA, "very_long_line.py"), "--disable=C"],
```
DanielNoord, line 20: Do we disable `C` in our own `pylintrc`?

tests/config/test_argparse_config.py
```diff
@@ -11,7 +11,7 @@
 
 from pylint.config.arguments_manager import _ArgumentsManager
 from pylint.config.exceptions import UnrecognizedArgumentAction
-from pylint.lint import Run
+from pylint.testutils import _Run as Run
```
DanielNoord, line 14:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/config/test_argparse_config.py
DanielNoord, line 67:
```suggestion
        assert run.linter.config.ignore_patterns == (re.compile("^\\.#"),)
```

Should be this then?

tests/config/test_config.py
```diff
@@ -11,7 +11,8 @@
 from pytest import CaptureFixture
 
 from pylint.config.exceptions import _UnrecognizedOptionError
-from pylint.lint import Run
+from pylint.lint import Run as LintRun
+from pylint.testutils import _Run as Run
```
DanielNoord, line 15:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/config/test_config.py
```diff
@@ -20,7 +21,7 @@
 
 
 def check_configuration_file_reader(
-    runner: Run,
+    runner: Run | LintRun,
```
DanielNoord, line 24:
```suggestion
    runner: Run,
```

Should be fine since `_Run` subclasses `Run`.

tests/config/test_config.py
```diff
@@ -11,7 +11,8 @@
 from pytest import CaptureFixture
 
 from pylint.config.exceptions import _UnrecognizedOptionError
-from pylint.lint import Run
+from pylint.lint import Run as LintRun
```
DanielNoord, line 14:
```suggestion
```

tests/lint/test_utils.py
```diff
@@ -7,8 +7,8 @@
 
 import pytest
 
-from pylint.lint import Run
 from pylint.lint.utils import get_fatal_error_message, prepare_crash_report
+from pylint.testutils import _Run as Run
```
DanielNoord, line 11:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/lint/unittest_lint.py
```diff
@@ -32,9 +32,10 @@
     OLD_DEFAULT_PYLINT_HOME,
 )
 from pylint.exceptions import InvalidMessageError
-from pylint.lint import PyLinter, Run
+from pylint.lint import PyLinter
 from pylint.message import Message
 from pylint.reporters import text
+from pylint.testutils import _Run as Run
```
DanielNoord, line 38:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/message/unittest_message_id_store.py
```diff
@@ -9,10 +9,10 @@
 
 import pytest
 
-from pylint import lint
 from pylint.exceptions import InvalidMessageError, UnknownMessageError
 from pylint.message.message_definition import MessageDefinition
 from pylint.message.message_id_store import MessageIdStore
+from pylint.testutils import _Run as Run
```
DanielNoord, line 15:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/primer/test_primer_stdlib.py
```diff
@@ -10,7 +10,7 @@
 import pytest
 from pytest import CaptureFixture
 
-import pylint.lint
+from pylint.testutils import _Run as Run
```
DanielNoord, line 13:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/profile/test_profile_against_externals.py
```diff
@@ -11,8 +11,8 @@
 
 import pytest
 
-from pylint.lint import Run
 from pylint.testutils import GenericTestReporter as Reporter
+from pylint.testutils import _Run as Run
```
DanielNoord, line 15:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/test_self.py
```diff
@@ -30,11 +30,12 @@
 
 from pylint import extensions, modify_sys_path
 from pylint.constants import MAIN_CHECKER_NAME, MSG_TYPES_STATUS
-from pylint.lint import Run
 from pylint.lint.pylinter import PyLinter
 from pylint.message import Message
 from pylint.reporters import JSONReporter
 from pylint.reporters.text import BaseReporter, ColorizedTextReporter, TextReporter
+from pylint.testutils import _Run as Run
```
DanielNoord, line 37:
```suggestion
from pylint.testutils.run import _Run as Run
```

tests/test_self.py
```diff
@@ -182,8 +186,9 @@ def _test_output_file(
         assert expected_output.strip() in file_output.strip()
 
     def test_pkginfo(self) -> None:
-        """Make pylint check itself."""
-        self._runtest(["pylint.__pkginfo__"], reporter=TextReporter(StringIO()), code=0)
+        """Make pylint check 'pylint.__pkginfo__.py'."""
+        args = ["pylint.__pkginfo__", "--disable=invalid-name"]
```
DanielNoord, line 190:
```suggestion
        # Disable invalid-name because of invalid argument names
        args = ["pylint.__pkginfo__", "--disable=invalid-name"]
```

tests/test_self.py
```diff
@@ -1090,13 +1095,13 @@ def test_fail_on_exit_code(self, args, expected):
         path = join(HERE, "regrtest_data", "fail_on.py")
         # We set fail-under to be something very low so that even with the warnings
         # and errors that are generated they don't affect the exit code.
-        self._runtest([path, "--fail-under=-10"] + args, code=expected)
+        self._runtest([path, "--fail-under=-10", "--disable=C"] + args, code=expected)
```
DanielNoord, line 1098: See question above.

tests/test_similar.py
```diff
@@ -13,7 +13,7 @@
 
 import pytest
 
-from pylint.lint import Run
+from pylint.testutils import _Run as Run
```
DanielNoord, line 16:
```suggestion
from pylint.testutils.run import _Run as Run
```

pylint/testutils/run.py
```diff
@@ -0,0 +1,46 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
DanielNoord, line 1: If we want to make all functions here private we could also consider renaming this file to `testutils/_run.py` and not adding anything from it to the `__all__` over `testutils`.

Pierre-Sassoulas: Making everything explicitly private might end up having the opposite effect: makes people thing that if something is not prefixed by an underscore then it's public. We might want to list explicitly what is public API instead ?

tests/checkers/unittest_refactoring.py
Pierre-Sassoulas: No but some like ``invalid-name`` and ``missing-docstring`` that happens very often.

pylint/testutils/run.py
DanielNoord: Isn't that we are assuming currently?

tests/config/test_config.py
Pierre-Sassoulas: We need a typevar for this to work, right ?

DanielNoord: I think it should be fine? But I might be mistaken. It is the same as using `_ArgumentsProvider` when we pass `BaseChecker`.

Pierre-Sassoulas: I started with Run only the error is:
```
tests/config/test_config.py:56: error: Argument 1 to "check_configuration_file_reader" has incompatible type "Run"; expected "_Run"  [arg-type]
Found 1 error in 1 file (checked 256 source files)
```

DanielNoord: Ah, it should be the other way around. It should be typed as `LintRun` (the `pylint.lint` one). That should accept both.

pylint/testutils/run.py
Pierre-Sassoulas: Well it depends on our wet finger estimation of something's external use. For example we did not assume that the constant of the ``FormatChecker`` where public API in #6455 because it's probably not used but it can be imported and there is no underscore. But if we modify the BaseChecker or something really central then we deprecate.

pylint/testutils/run.py
Pierre-Sassoulas: I supposed it could be used to add another pylintrc in args, but doing "--rcfile={pylnitrc}" directly is better. YAGNI.

pylint/testutils/run.py
DanielNoord: Hmm okay. I'd say let's try and do stuff as we would like them to be and worry less about potential implication for modules we haven't updated yet. I think underscoring files that are not public is good practice and also done by others such as `pytest`. If a file has mixed external and internal functions the internal ones should be underscored and the externals be added to the respective `__all__`. That way we document it both through the `__all_` and the use of underscores (which I think is often more visible to users).

Pierre-Sassoulas: 👍 Now that I think of it, the test ``Run`` is actually a nice utility if you need to run pylint tests with default values in a repository where there's a pylintrc, I wonder if we should make it public with the same interface than ``pylint.lint.Run`` ?

tests/config/test_config.py
Pierre-Sassoulas: Damn that was a stupid mistake, my bad.

Pierre-Sassoulas: 5000538894d9fbc113c04bf5e5f8152de72ba25d

pylint/testutils/run.py
DanielNoord: The issue I have with this is: how much trouble do we save plugin writers compared to the trouble we get when we want to change anything related to these functions.
The functional test framework is a good example of a very useful test utility that would take a plugin writer considerable time to write and for which it would be good to get updates if we make any.
On the other hand, this utility is so straightforward that I wouldn't be surprised if plugin writers already wrote their own. It just adds overhead, maintenance and size to the overall package with relatively little benefit, imo.

That's also why I'm so "involved" in this particular discussion. I don't think this is such a valuable function that we should make it public while knowing that it is using arguments that we are in the process of deprecating (`do_exit`).
If this was my personal project I would likely add to this to a private `internal_testutils` package which wouldn't be included in the package at all.

Obviously I'm fine with putting it in if you want to, but then we should perhaps consider already removing the `do_exit` argument from `_Run`.

Pierre-Sassoulas: No you're right let's make this file private.

pylint/testutils/_run.py
```diff
@@ -0,0 +1,40 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
```
DanielNoord, line 4:
```suggestion

"""Classes and functions used to mimic normal pylint runs.

This module is considered private and can change at any time.
"""

```

DanielNoord (approved):
One final nitpick.

Rest looks good!

DanielNoord:
This breaks `main`. `pyenchant` is not a dependency so when you clone `main` and do `pip install -e .` you can no longer run `pylint pylint`...

PR #6462, opened 2022-04-26
Fix the installation of pyenchant

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to https://github.com/PyCQA/pylint/pull/6440#issuecomment-1109437387

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,13 @@ Release date: TBA
 
   Closes #5224
 
+* Pylint can now be installed with an extra-require called ``spelling`` (``pip install pylint[spelling]``).
+  This will add ``pyenchant`` to pylint's dependencies. You will still need to install the
+  requirements for ``pyenchant`` (the ``enchant`` library and any dictionaries) yourself. You will also
+  need to set the ``spelling-dict`` option.
+
+  Refs #6462
+
 * Removed the ``assign-to-new-keyword`` message as there are no new keywords in the supported Python
   versions any longer.
 
--- a/README.rst
+++ b/README.rst
@@ -130,8 +130,8 @@ To run individual tests with ``tox``, you can do::
 
     tox -e py37 -- -k name_of_the_test
 
-If you're testing new change in astroid you need to clone astroid_ and install
-the last version::
+If you're testing new changes in astroid you need to clone astroid_ and install
+with an editable installation as follows::
 
     git clone https://github.com/PyCQA/astroid.git
     cd astroid
--- a/README.rst
+++ b/README.rst
@@ -106,45 +106,34 @@ Pylint is shipped with following additional commands:
 Testing
 -------
 
-We use tox_ and pytest-benchmark_ for running the test suite. You should be able to install it with::
+You should be able to install our tests dependencies with::
 
-    pip install tox pytest pytest-benchmark
+    pip install -r requirements_test.txt
 
-
-To run the test suite for a particular Python version, you can do::
-
-    tox -e py37
-
-
-To run individual tests with ``tox``, you can do::
-
-    tox -e py37 -- -k name_of_the_test
-
-
-We use pytest_ for testing ``pylint``, which you can use without using ``tox`` for a faster development cycle.
-
-If you want to run tests on a specific portion of the code with pytest_, (pytest-cov_) and your local python version::
+You can then use pytest_ directly. If you want to run tests on a specific portion of the
+code with pytest_ and your local python version::
 
     # ( pip install pytest-cov )
-    # Everything:
-    python3 -m pytest tests/
+    python3 -m pytest
     # Everything in tests/message with coverage for the relevant code:
     python3 -m pytest tests/message/ --cov=pylint.message
     coverage html
     # Only the functional test "missing_kwoa_py3":
     python3 -m pytest "tests/test_functional.py::test_functional[missing_kwoa_py3]"
 
+You can also *optionally* install tox_. To run the test suite for a particular
+Python version, with tox you can do::
 
-Do not forget to clone astroid_ and install the last version::
+    tox -e py39
 
+To run individual tests with ``tox``, you can do::
+
+    tox -e py37 -- -k name_of_the_test
+
+If you're testing new change in astroid you need to clone astroid_ and install
+the last version::
 
     git clone https://github.com/PyCQA/astroid.git
-
-    # From source
-    python3 astroid/setup.py build sdist
-    pip3 install astroid/dist/astroid*.tar.gz
-
-    # Using an editable installation
     cd astroid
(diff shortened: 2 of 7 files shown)


DanielNoord:
Don't really like this solution as you still need to install stuff outside of the python ecosystem. This still does not allow:
`git clone` + `pip install -e .`. If `C Enchant` was necessary for correct use of `pylint` this would be fine, but to require such dependencies for a simple spelling check feels a bit too much.

I have another solution where we only run the spelling check in CI. I'll create a PR shortly!

Pierre-Sassoulas:
You're very fast 😄 I changed the approach completely, could you check again. I think this is now the correct approach as this also fix a long standing issue of the spell checker being "hidden". You had to know it existed and you had to install pyenchant yourself and then use the proper option. This does not make a lot of sense.

DanielNoord:
> You're very fast 😄 I changed the approach completely, could you check again. I think this is now the correct approach as this also fix a long standing issue of the spell checker being "hidden". You had to know it existed and you had to install pyenchant yourself and then use the proper option. This does not make a lot of sense.

The issue remains that doing `pip install pyenchant` will never be enough for `pyenchant` to work. It uses the `C Enchant` library which you need to install manually, for example through home-brew. See https://pyenchant.github.io/pyenchant/install.html#installation.

Therefore, we can't fix this issue without having developers/users go through the extra effort of installing `C Enchant`. For Windows the `pyenchant` documentation even says: `The enchant C library depends on glib2, which poses some interesting challenges.`.
I don't think we should require developers/users to jump through those hoops for a simple spelling check. It is a bit annoying that spelling issues will only surface in CI, but that's kinda what CI is for and since the `Checks` job doesn't take 30 minutes to run PR-submitters won't have to wait too long.

Obviously, we could add something about installing `C Enchant` in our developer documentation and recommend to run with spelling check before submitting PRs, but I don't think it should be standard.

Pierre-Sassoulas:
I think there's two issues here:
- Adding spelling check in pre-commit
- Adding the extra require for spelling

For spelling check in pre-commit I think it makes sense to add it anyway as a lot of first time contributors do not even run pre-commit locally (As evidenced by the fact that the pre-commit-ci bot very often automatically fix commits). I suppose we don't want to bother the one that do want to install pre-commit. Does the pip install fail if C enchant is not installed ?

Then for the extra require option I think not adding was a real problem as you really have to go out of your way to discover this spell checker.

DanielNoord:
> I think there's two issues here:
> 
> * Adding spelling check in pre-commit
> * Adding the extra require for spelling
> 
> For spelling check in pre-commit I think it makes sense to add it anyway as a lot of first time contributors do not even run pre-commit locally (As evidenced by the fact that the pre-commit-ci bot very often automatically fix commits). I suppose we don't want to bother the one that do want to install pre-commit. Does the pip install fail if C enchant is not installed ?

Yes, as you would define the dictionary to use in the `rcfile` or command line but that dictionary won't be found because `C Enchant` is not installed. That will result in a crash in the configuration parser.

> Then for the extra require option I think not adding was a real problem as you really have to go out of your way to discover this spell checker.

Adding an extra require option is fine, but the same goes here: it doesn't really work unless you also install `C Enchant` so that should probably be documented there as well.

Pierre-Sassoulas:
> that dictionary won't be found because C Enchant is not installed. That will result in a crash in the configuration parser.

I noticed that too. I think there might be a better way to handle this in pylint. We could add a hard coded list of choices for the parsing, and then raise a pylint-error message if enchant is not installed later. I did not think a lot about this (and the way we handle the fact that enchant is a schrodinger installation in the SpellingChecker) because there were other issues to fix first. 


> it doesn't really work unless you also install C Enchant so that should probably be documented there as well.

Yeah nothing we can do about it except documenting, but this is an issue for users also imo, not only for pylint's dev.

DanielNoord:
> > that dictionary won't be found because C Enchant is not installed. That will result in a crash in the configuration parser.
> 
> I noticed that too. I think there might be a better way to handle this in pylint. We could add a hard coded list of choices for the parsing, and then raise a pylint-error message if enchant is not installed later. I did not think a lot about this (and the way we handle the fact that enchant is a schrodinger installation in the SpellingChecker) because there were other issues to fix first. The manual check in pre-commit seems the way to fix this short term 👍
> 
> > it doesn't really work unless you also install C Enchant so that should probably be documented there as well.
> 
> Yeah nothing we can do about it except documenting, but this is an issue for users also imo, not only for pylint's dev.

Another issue I just thought of it is that there is no good way to make sure dictionaries are the same. `Enchant` can get dictionaries from a number of providers so there is no guarantee that my local `en_GB` is the same as your `en_GB`...
That is an additional argument to only let the spelling checker run in CI where we can control the dictionary to be used.

DanielNoord (changes requested):

README.rst
```diff
@@ -106,45 +106,34 @@ Pylint is shipped with following additional commands:
...
-To run individual tests with ``tox``, you can do::
-
-    tox -e py37 -- -k name_of_the_test
-
-
-We use pytest_ for testing ``pylint``, which you can use without using ``tox`` for a faster development cycle.
-
-If you want to run tests on a specific portion of the code with pytest_, (pytest-cov_) and your local python version::
+You can then use pytest_ directly. If you want to run tests on a specific portion of the
+code with pytest_, (pytest-cov_) and your local python version::
```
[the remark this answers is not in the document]
Pierre-Sassoulas: I guess it means pytest-cov is optional (to get coverage)

ChangeLog
```diff
@@ -30,6 +30,13 @@ Release date: TBA
 
   Closes #5224
 
+* Pylint can now be installed with an extra-require called ``spelling`` (``pip install pylint[spelling]``).
+  This will requires a local installation of ``pyenchant``. You will still need to install the
```
DanielNoord, line 34:
```suggestion
  This will add ``pyenchant`` to pylint's dependencies. You will still need to install the
```

README.rst
DanielNoord, line 114:
```suggestion
code with pytest_ and your local python version::
```

I'd say let's remove it here. It is not totally clear what users should do.

doc/whatsnew/2.14.rst
```diff
@@ -79,6 +79,13 @@ Extensions
 Other Changes
 =============
 
+* Pylint can now be installed with an extra-require called ``spelling`` (``pip install pylint[spelling]``).
+  This will requires a local installation of ``pyenchant``. You will still need to install the
```
DanielNoord, line 83:
```suggestion
  This will add ``pyenchant`` to pylint's dependencies. You will still need to install the
```

DanielNoord (approved):

PR #6466, opened 2022-04-27
Add an assertion to check that the documentation structure is correct

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to an issue where the documentation being in the wrong directory result in the documentation not displaying the content silently in #6404

### DIFF
--- a/doc/test_messages_documentation.py
+++ b/doc/test_messages_documentation.py
@@ -38,11 +38,12 @@ def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str,
     This also checks the formatting of related.rst files.
     """
     suite: List[Tuple[str, Path]] = []
+
     for subdirectory in input_dir.iterdir():
         for message_dir in subdirectory.iterdir():
             assert_msg = (
                 f"{subdirectory}: '{message_dir.name}' is in the wrong "
-                f"directory: it does not start by '{subdirectory.name}'"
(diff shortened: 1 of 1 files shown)


doc/test_messages_documentation.py
```diff
@@ -38,9 +38,13 @@ def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str,
     This also checks the formatting of related.rst files.
     """
     suite: List[Tuple[str, Path]] = []
-
     for subdirectory in input_dir.iterdir():
         for message_dir in subdirectory.iterdir():
+            assert_msg = (
+                f"{subdirectory}: '{message_dir.name}' is in the wrong "
+                f"directory: it does not start by '{subdirectory.name}'"
```
DanielNoord, line 45:
```suggestion
                f"directory: it does not start with '{subdirectory.name}'"
```

doc/test_messages_documentation.py
```diff
@@ -38,9 +38,13 @@ def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str,
     This also checks the formatting of related.rst files.
     """
     suite: List[Tuple[str, Path]] = []
```
DanielNoord, line 40:
```suggestion
    suite: List[Tuple[str, Path]] = []

```

DanielNoord (approved):

PR #6469, opened 2022-04-28
Some formatting of docstrings with ``pydocstringformatter``

DanielNoord:

## Type of Changes

## Description

@Pierre-Sassoulas This is only a partial run, but should make updating to a future release much easier with a smaller diff 😄

### DIFF
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -342,7 +342,7 @@ class BasicChecker(_BasicChecker):
 
     def visit_classdef(self, _: nodes.ClassDef) -> None:
         """Check module name, docstring and redefinition
-        increment branch counter
+        increment branch counter.
         """
         self.linter.stats.node_count["klass"] += 1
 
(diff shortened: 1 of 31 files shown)


DanielNoord:
> Look like it's cutting line a little too short, we have 88 char lines with black, docstring should be the same length imo.

Which lines are too short?

Pierre-Sassoulas:
In
```
    """The LinesChunk object computes and stores the hash of some consecutive
    stripped lines of a lineset.
    """
```    
the first line is 78 char but if you had `` stripped`` it would be 87 chars

DanielNoord:
Remove the wrapping commit. I'll investigate in `pydocstringformatter`. Probably something to do with taking `len()` and zero-indexing or something... 😓

DanielNoord:
Found the issue: I subtract the quotes from the line-length whenever we have single line docstring, so
```python
def func():
    """A multi-line summary that can be on one line. Something that is just too longgg."""
```
Becomes
```python
# Since the ending quotes will be appended on the same line this
# exceeds the max length.
def func():
    """A multi-line summary that can be on one line. Something that is just too
    longgg.
    """
```

However, in the case you mentioned it should be re-added as the quotes will no longer be on the first line anyway.

Pierre-Sassoulas (approved):
👍

PR #6479, opened 2022-05-01
Fix false positive for `undefined-loop-variable` for lambda in first of two loops

### LINKED ISSUE #6419: W0631 undefined-loop-variable false positive for lambda statements inside first of two loops
### Bug description
```python
def test():
   l = []
   for thing in l:
      x = lambda: print( thing )
      x()

   for thing in l:
      print( thing )
```
  
If I initialize the list, l, to anything besides the empty list, then the W0631 warning goes away.  Or instead if I delete the second for-loop, the W0631 warning goes away as well.
    
Seems similar to this bug: https://github.com/PyCQA/pylint/issues/202

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
% pylint test.py
************* Module test
test.py:16:25: W0640: Cell variable thing defined in loop (cell-var-from-loop)
test.py:16:25: W0631: Using possibly undefined loop variable 'thing' (undefined-loop-variable)

Your code has been rated at 8.46/10
```

### Expected behavior

I don't expect to see W0631 (undefined-loop-variable).  The (cell-var-from-loop) warning is warranted.

### Pylint version

```shell
% pylint --version
pylint 2.11.1
astroid 2.8.5
Python 3.8.12 (default, Aug 31 2021, 12:46:19)
[GCC 8.4.0 20200304 (Red Hat 8.4.0-4)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #6419

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2199,7 +2199,8 @@ class VariablesChecker(BaseChecker):
         # the variable is not defined.
         scope = node.scope()
         # FunctionDef subclasses Lambda due to a curious ontology. Check both.
-        # See https://github.com/PyCQA/astroid/issues/291 (fix abandoned in a stale branch)
+        # See https://github.com/PyCQA/astroid/issues/291
+        # TODO: Revisit when astroid 3.0 includes the change
         if isinstance(scope, nodes.Lambda) and any(
             asmt.scope().parent_of(scope) for asmt in astmts
         ):
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2198,7 +2198,7 @@ class VariablesChecker(BaseChecker):
         # the usage is safe because the function will not be defined either if
         # the variable is not defined.
         scope = node.scope()
-        # FunctionDef subclasses Lambda due to suboptimal ontology. Check both.
+        # FunctionDef subclasses Lambda due to a curious ontology. Check both.
         # See https://github.com/PyCQA/astroid/issues/291 (fix abandoned in a stale branch)
         if isinstance(scope, nodes.Lambda) and any(
             asmt.scope().parent_of(scope) for asmt in astmts
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2198,7 +2198,9 @@ class VariablesChecker(BaseChecker):
         # the usage is safe because the function will not be defined either if
         # the variable is not defined.
         scope = node.scope()
-        if isinstance(scope, nodes.FunctionDef) and any(
+        # FunctionDef subclasses Lambda due to suboptimal ontology. Check both.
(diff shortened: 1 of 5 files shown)


pylint/checkers/variables.py
```diff
@@ -2198,7 +2198,9 @@ def _loopvar_name(self, node: astroid.Name) -> None:
         # the usage is safe because the function will not be defined either if
         # the variable is not defined.
         scope = node.scope()
-        if isinstance(scope, nodes.FunctionDef) and any(
+        # FunctionDef subclasses Lambda due to a curious ontology. Check both.
+        # See https://github.com/PyCQA/astroid/issues/291 (fix abandoned in a stale branch)
```
Pierre-Sassoulas, line 2202: Do you think we could merge this commit on main astroid ?

jacobtylerwalls: I sort of like the idea, but it would be a very big breaking change. Everywhere in `pylint` (and I guess `astroid` also) that does any `isinstance` for lambda or function will need to be audited and have additional tests written. At the very least we could reopen the issue and add it to a milestone.

Pierre-Sassoulas: Yeah it sounds like something that could go in astroid 3.0.

jacobtylerwalls: For instance, my changes here in this PR would break!

jacobtylerwalls: Thinking more, maybe the proposal to switch the parent/child classes makes no more sense than the current state of things. I don't have a good intuition for which one should be the parent. Maybe they should both inherit from `BaseFunction` (like `BaseContainer`) and break less code (just the code that checks for lambda and expects to also catch FunctionDef? but that seems rare ...)

jacobtylerwalls: _Wow, oops_, that commit you linked looks like it does basically what I just said. So yeah, that's starting to look pretty good.

Pierre-Sassoulas:
```suggestion
        # See https://github.com/PyCQA/astroid/issues/291
        # TODO:  Revisit when astroid 3.0 include the change
```

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Hmm I'm going to fix that by amending the commit on main.

DanielNoord:
Can't remember if I ever asked this: but why don't we just disable `fixme`? It seems like such a silly linting warning..

Pierre-Sassoulas:
> Can't remember if I ever asked this: but why don't we just disable fixme? It seems like such a silly linting warning..

We didn't, because #3512 is overwhelming, but it's supposed to happens 😄 I think I'll have to check what need to be done for this one because it's not hard to do, we only have to follow Ashley's tabular lead.

PR #6481, opened 2022-05-01
Fix false positive for ``unused-variable`` when specifying a metaclass via a call

### LINKED ISSUE #4020: unused-variable false positive
### Steps to reproduce
This seems to be the minimal reproducer, 
```
def f():
    def g():
        return type

    class C(metaclass=g()):
        pass

    return C
```

### Current behavior
```
test.py:5:4: W0612: Unused variable 'g' (unused-variable)
```

### Expected behavior
No error, since the function is used.

### pylint --version output
```
pylint 2.6.0
astroid 2.4.2
Python 3.8.7 (default, Dec 24 2020, 17:53:09) 
[GCC 10.2.0]
```

jacobtylerwalls:
## Type of Changes

## Description

Closes #4020

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -247,6 +247,11 @@ Release date: TBA
 
   Closes #6414
 
+* Fix false positive for ``unused-variable`` for classes inside functions
+  and where a metaclass is provided via a call.
+
+  Closes #4020
+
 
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):
👌

PR #6483, opened 2022-05-01
Add related and details.rst to 'redundant-unittest-assert' doc

Pierre-Sassoulas:
## Type of Changes

## Description

Follow up to a late review in #6475. I realized that two very close messages where on two different checkers it's minor and hard to fix (it could be just an old name for 'redundant-unittest-assert' to add to  ' assert-on-string-literal' otherwise ).

### DIFF
--- a/doc/data/messages/a/assert-on-string-literal/details.rst
+++ b/doc/data/messages/a/assert-on-string-literal/details.rst
@@ -1,4 +1,4 @@
 Directly asserting a string literal will always pass. The solution is to
 test something that could fail, or not assert at all.
 
-This message is similar to ``redundant-unittest-assert``.
+For ``unittest`` assertions there is the similar :ref:`redundant-unittest-assert` message.
--- a/doc/data/messages/a/assert-on-string-literal/details.rst
+++ b/doc/data/messages/a/assert-on-string-literal/details.rst
@@ -1,1 +1,4 @@
-Directly asserting a string literal will always pass.
+Directly asserting a string literal will always pass. The solution is to
+test something that could fail, or not assert at all.
+
+This message is similar to ``redundant-unittest-assert``.
--- a/doc/data/messages/a/assert-on-tuple/details.rst
+++ b/doc/data/messages/a/assert-on-tuple/details.rst
@@ -1,0 +1,1 @@
+Directly asserting on a tuple will never fail.
--- a/doc/data/messages/a/assert-on-tuple/details.rst
+++ b/doc/data/messages/a/assert-on-tuple/details.rst
@@ -1,1 +1,4 @@
-Directly asserting on a tuple will never fail.
+Directly asserting a non-empty tuple will always pass. The solution is to
+ test something that could fail, or not assert at all.
+
+ For ``unittest`` assertions there is the similar :ref:`redundant-unittest-assert` message.
--- a/doc/data/messages/r/redundant-unittest-assert/bad.py
+++ b/doc/data/messages/r/redundant-unittest-assert/bad.py
@@ -4,3 +4,8 @@ import unittest
 class DummyTestCase(unittest.TestCase):
     def test_dummy(self):
         self.assertTrue("foo")  # [redundant-unittest-assert]
+
+
+def test_division():
+    a = 9 / 3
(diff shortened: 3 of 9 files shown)


doc/data/messages/r/redundant-unittest-assert/bad.py
```diff
@@ -4,3 +4,8 @@
 class DummyTestCase(unittest.TestCase):
     def test_dummy(self):
         self.assertTrue("foo")  # [redundant-unittest-assert]
+
+
+def test_division():
+    a = 9 / 3
+    assert "No ZeroDivisionError were raised"  # assert-on-string-literal
```
DanielNoord, line 11: Why are we adding an example for another message here?

doc/data/messages/r/redundant-unittest-assert/good.py
```diff
@@ -3,5 +3,9 @@
 
 class DummyTestCase(unittest.TestCase):
     def test_dummy(self):
-        actual = "test_result"
-        self.assertEqual(actual, "expected")
+        # Nothing, as an assert of a string literal will always pass
```
DanielNoord, line 6: I actually kind of like the current example. I think the comment works nicely in `details.rst`.

doc/data/messages/r/redundant-unittest-assert/good.py
```diff
@@ -3,5 +3,9 @@
...
+        pass
+
+def test_division():
+    a = 9 / 3
+    assert a == 3
```
DanielNoord, line 11: If we do `# nothing` above we should also do so here? Note, I actually think we should keep the code instead of the comment.

DanielNoord (commented):
I don't really understand the intention of this PR I think. If we want people to know about a similar but different message we could probably add a link in the `details.rst` of both of these messages.

doc/data/messages/r/redundant-unittest-assert/bad.py
Pierre-Sassoulas: I added an example expecting this code to raise a ``[redundant-unittest-assert]``, but I think those two messages should be the same one. I think it can illustrate the two ways to fix the issue.

DanielNoord: But this example is about `unittest` assertions, not normal `assert`. I don't think we should merge those.

doc/data/messages/a/assert-on-string-literal/details.rst
```diff
@@ -1 +1,4 @@
-Directly asserting a string literal will always pass.
+Directly asserting a string literal will always pass. The solution is to
+test something that could fail, or not assert at all.
+
+This message is similar to ``redundant-unittest-assert``.
```
DanielNoord, line 4:
```suggestion
For ``unittest`` assertions there is the similar :ref:`redundant-unittest-assert` message.
```

doc/data/messages/r/redundant-unittest-assert/details.rst
```diff
@@ -1 +1,4 @@
-Directly asserting a string literal will always pass.
+Directly asserting a string literal will always pass. The solution is to
+test something that could fail, or not assert at all.
+
+This message is similar to ``assert-on-string-literal``.
```
DanielNoord, line 4:
```suggestion
For assertions using ``assert`` there are similar messages:  :ref:`assert-on-string-literal` and :ref:`assert-on-tuple`.

```

DanielNoord (changes requested):
`details` of ``assert-on-tuple`` should probably also be updated.

doc/data/messages/a/assert-on-tuple/details.rst
```diff
@@ -0,0 +1 @@
+Directly asserting on a tuple will never fail.
```
DanielNoord, line 1:
```suggestion
Directly asserting a non-empty tuple will always pass. The solution is to
 test something that could fail, or not assert at all.

 For ``unittest`` assertions there is the similar :ref:`redundant-unittest-assert` message.
```

Pierre-Sassoulas (commented):
Thank you for the review @DanielNoord this looks better indeed.

DanielNoord (approved):

PR #6492, opened 2022-05-03
Ignore underscore as a local variable

### LINKED ISSUE #6488: too-many-locals should ignore underscore variable names
### Current problem

Not sure how underscores are handled everywhere else, but it would be nice if they didn't count toward local variable counts since they aren't being used.

### Desired solution

When unpacking triple, I am ignoring the middle variable with an "_" and it shouldn't count towards the local variable count, especially as it doesn't count as an unused variable. Basically I could easily avoid this by not unpacking my lists, referencing them with `triple[0]` and `triple[2]`, but that would lead to very obscure code.

### Additional context

_No response_

ayoung19:
## Type of Changes

## Description

Closes #6488

### DIFF
--- a/pylint/checkers/design_analysis.py
+++ b/pylint/checkers/design_analysis.py
@@ -536,6 +536,11 @@ class MisdesignChecker(BaseChecker):
             ignored_args_num = 0
         # check number of local variables
         locnum = len(node.locals) - ignored_args_num
+
+        # decrement number of local variables if '_' is one of them
+        if "_" in node.locals:
+            locnum -= 1
+
         if locnum > self.linter.config.max_locals:
             self.add_message(
                 "too-many-locals",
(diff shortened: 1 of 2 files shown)


DanielNoord (changes requested):
Thanks @ayoung19 for picking this up!

The change itself looks good to me. However, could you perhaps add a changelog entry? This should probably go both in `doc/whatsnew/2.14` and `./Changelog`.

Pierre-Sassoulas (approved):
Pretty straightforward fix, and the test is on point, great first contribution !

DanielNoord (approved):

ayoung19:
Just added the entry, let me know if I need to be more descriptive!

DanielNoord:
Thanks @ayoung19, that is perfect! Congratulations on becoming a `pylint` contributor 😄

PR #6491, opened 2022-05-02
Small refactor in spelling checker

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Refactor prior to #6137

PR #6493, opened 2022-05-03
[spelling checker] Use a context manager to access the private dict

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to #6491, before #6137

### DIFF
--- a/pylint/checkers/spelling.py
+++ b/pylint/checkers/spelling.py
@@ -268,8 +268,6 @@ class SpellingChecker(BaseTokenChecker):
 
     def open(self) -> None:
         self.initialized = False
-        self.private_dict_file = None
-
         if enchant is None:
             return
         dict_name = self.linter.config.spelling_dict
@@ -292,9 +290,6 @@ class SpellingChecker(BaseTokenChecker):
             self.spelling_dict = enchant.DictWithPWL(
                 dict_name, self.linter.config.spelling_private_dict_file
             )
-            self.private_dict_file = open(  # pylint: disable=consider-using-with
-                self.linter.config.spelling_private_dict_file, "a", encoding="utf-8"
-            )
         else:
             self.spelling_dict = enchant.Dict(dict_name)
 
@@ -315,10 +310,6 @@ class SpellingChecker(BaseTokenChecker):
             ],
         )
         self.initialized = True
-
-    def close(self) -> None:
-        if self.private_dict_file:
-            self.private_dict_file.close()
 
     def _check_spelling(self, msgid: str, line: str, line_num: int) -> None:
         original_line = line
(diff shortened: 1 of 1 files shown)


pylint/checkers/spelling.py
```diff
@@ -267,10 +267,15 @@ class SpellingChecker(BaseTokenChecker):
         ),
     )
 
+    def __init__(self, linter: PyLinter):
+        super().__init__(linter)
+        # Expand tilde to allow e.g. spelling-private-dict-file = ~/.pylintdict
+        self.private_dict_path: str = os.path.expanduser(
+            self.linter.config.spelling_private_dict_file
+        )
```
Pierre-Sassoulas, line 275: Safe to do because:
```
>>> import os
>>> os.path.expanduser("")
''
```

pylint/checkers/spelling.py
```diff
@@ -267,10 +267,15 @@ class SpellingChecker(BaseTokenChecker):
         ),
     )
 
+    def __init__(self, linter: PyLinter):
```
DanielNoord, line 270:
```suggestion
    def __init__(self, linter: PyLinter) -> None:
```

DanielNoord (changes requested):
Is there a direct need for `self.private_dict_path`? It seems like we could use `self.linter.config.spelling_private_dict_file` directly. The only difficulty is that we would need to do `os.path.expanduser` when loading the argument. So we would need a new argument type (probably called `path`).

Pierre-Sassoulas:
>  The only difficulty is that we would need to do os.path.expanduser when loading the argument. So we would need a new argument type (probably called path).

Having a generic path option that would make ALL options using a path, expand the user path correctly would be a huge win for pylint. There's 7 expandusers call in the codebase, but probably a lot of option where we do not support this yet.

DanielNoord:
> Having a generic path option that would make ALL options using a path, expand the user path correctly would be a huge win for pylint. There's 7 expandusers call in the codebase, but probably a lot of option where we do not support this yet.

Shouldn't be too difficult to create a new transformer. Do you want to look into this?

Pierre-Sassoulas:
> Do you want to look into this?

Sure. 

Regarding the other comment, using the configuration make for rather long unreadable variable name ``self.linter.config.spelling_private_dict_file``, spelling is noise when we're in the spelling checker (but not when we're in the pylint configuration) so I supposed that creating a small var name would clarify the code.

DanielNoord:
> > Do you want to look into this?
> 
> Sure.
> 
> Regarding the other comment, using the configuration make for rather long unreadable variable name `self.linter.config.spelling_private_dict_file`, spelling is noise when we're in the spelling checker (but not when we're in the pylint configuration) so I supposed that creating a small var name specific to the checker would clarify the code.

Hm, on the other hand decoupling a setting from the actual underlying setting could create other problems down the line. What if we have the per directory configuration enabled and the underlying `Namespace` changes? I'd be in favour of the long name (although I do agree it is a bit annoying).

DanielNoord (approved):

PR #6498, opened 2022-05-04
Add regression test for #6497

### LINKED ISSUE #6497: Pylint cannot find `misc_util` module in `numpy.distutils`
### Bug description

Hi there,

No sure whether this is an issue in numpy or pylint.

Starting from `pylint` 2.13.0, `pylint` reports an error when importing `numpy.distutils.misc_util`
When I downgrade pylint to the previous version (2.12.2) there is no error reported

script `a.py`:
```
# pylint: disable=unused-import,missing-module-docstring

from numpy.distutils.misc_util import is_sequence


assert is_sequence("ABC") is False
assert is_sequence([1, 2, 3]) is True

```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:3:0: E0611: No name 'misc_util' in module 'numpy.distutils' (no-name-in-module)
a.py:3:0: E0401: Unable to import 'numpy.distutils.misc_util' (import-error)
```

### Expected behavior

The module exists so I would not expect pylint to report an error here 

### Pylint version

```shell
2.13.0
```

### OS / Environment

Mac OS 12.3.1

### Additional dependencies

numpy==1.22.3

jacobtylerwalls:
## Type of Changes

## Description

Closes #6497 (depends on https://github.com/PyCQA/astroid/pull/1544)

Backporting an astroid patch upgrade is not essential, but possible?


jacobtylerwalls:
Hmm, we have other functional tests using `numpy` but apparently aren't installing it on CI runs.

DanielNoord:
Think it makes sense to do this as not finding modules can have more effect than a simple bad-open-mode false positive.

Pierre-Sassoulas:
> Backporting an astroid patch upgrade is not essential, but possible?

I think it makes sense, we want to make sure that the fix works. We already got an issue with 2.13.7, and the 2.13 maintenance branch is even more different than 2.14.0-dev0 now. It could also make sense to wait for 2.14 though, it's getting closer.

jacobtylerwalls:
> I think it makes sense, we want to make sure that the fix works. We already got an issue with 2.13.7, and the 2.13 maintenance branch is even more different than 2.14.0-dev0 now. It could also make sense to wait for 2.14 though, it's getting closer.

I'm tempted to wait and just do 2.14 final. May I ask what is the issue with 2.13.8 you had in mind? (Just making sure it's not the one I closed as a dupe!)

Pierre-Sassoulas:
> May I ask what is the issue with 2.13.8 you had in mind?

We had an issue with the config that changed so cherry-picking went wrong https://github.com/PyCQA/pylint/milestone/64?closed=1, it's fixed now, but this kind of problem can happens again (well we'll fix it before releasing this time, but it's more work).

requirements_test.txt
```diff
@@ -10,3 +10,4 @@ pytest-profiling~=1.7
 pytest-xdist~=2.5
 # Type packages for mypy
 types-pkg_resources==0.1.3
+numpy~=1.22.3
```
Pierre-Sassoulas, line 13: I don't think we should add numpy to our standard test requirements, if we do that then we could have code that misbehave if numpy is not installed and we would not know. What we could do is create a CI job where we install additional libraries. It's been a recurring topic when dealing with pandas/numpy, etc. To be clear we need to be able to handle crashes in external lib and make sure that we do not regress but for false positives it's a little blurrier. For django and pytest we deferred to external plugins for false positives. Starting to do lib specific code directly in pylint is a slipery slope, we already have a lot on our plates as it is.

jacobtylerwalls: I bet I could just rewrite with a `fake_package/distutils`.

jacobtylerwalls: Nah, that didn't work.

jacobtylerwalls: I added a comment. Adding a new workflow is not a very appealing ROI right now. We tested this with `numpy` installed in `astroid`, and if it fails somehow here in `pylint` again, a contributor will notice.

Pierre-Sassoulas (commented):
I agree about not adding a new CI job right now. Would you mind adding a test for [pydantic's distutil](https://github.com/PyCQA/pylint/issues/6497#issuecomment-1121334754) ?

PR #6499, opened 2022-05-05
Allow not yet published extensions links in ``linkcheck``

### LINKED ISSUE #6470: PRs creating extensions will always fail the documentation check
#6058 generates links to GitHub pages for optional checkers. This page won't actually be live until a new checker is merged.

#6292 added a documentation check to CI.

The combination of these two facts means PRs adding extensions will always fail the documentation check. We can tolerate that, I guess, but are there better ideas?

See https://github.com/PyCQA/pylint/runs/6210620669?check_suite_focus=true

DanielNoord:

## Type of Changes

## Description

Fixes #6470.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -241,3 +241,5 @@ intersphinx_mapping = {
 # Prevent label issues due to colliding section names
 # through including multiple documents
 autosectionlabel_prefix_document = True
+
+linkcheck_ignore = ["https://github.com/PyCQA/pylint/blob/main/pylint/extensions/.*"]


doc/conf.py
```diff
@@ -241,3 +241,5 @@
 # Prevent label issues due to colliding section names
 # through including multiple documents
 autosectionlabel_prefix_document = True
+
+linkcheck_ignore = ["https://github.com/PyCQA/pylint/blob/main/pylint/extensions/.*"]
```
Pierre-Sassoulas, line 245: Could we check the extension on ``main`` (where the extension should be published already) but not on feature branch by creating a CI job specific for this test ?

DanielNoord: Yeah, but then we're likely to miss failures on the step right? I don't think these links will ever break, so it should be fine to just ignore these very specific links in normal CI runs?

Pierre-Sassoulas: Well, 2 weeks ago we did not check ANY links, I think we'll be fine 😄

Pierre-Sassoulas (approved):

doc/conf.py
jacobtylerwalls: All the choices are somewhat flawed :-)

PR #6504, opened 2022-05-05
Do not report useless-suppression on duplicate-code

DanielNoord:

## Type of Changes

## Description

Because we handle ``duplicate-code`` a little differently with disables the normal way of figuring out if the disable is useless doesn't work. Let's just not worry about it like we do for other difficult messages.


Pierre-Sassoulas (approved):
Wise decision !

PR #6516, opened 2022-05-05
Fix pylint.config's caching deprecation

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to #6468

### DIFF
--- a/tests/test_deprecation.py
+++ b/tests/test_deprecation.py
@@ -12,6 +12,7 @@ import pytest
 
 from pylint.checkers import BaseChecker
 from pylint.checkers.mapreduce_checker import MapReduceMixin
+from pylint.config import load_results, save_results
 from pylint.interfaces import (
     IAstroidChecker,
     IChecker,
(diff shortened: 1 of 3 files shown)


tests/test_deprecation.py
```diff
@@ -83,6 +82,10 @@ def test_interfaces() -> None:
 
 def test_load_and_save_results() -> None:
     """Test that load_results and save_results are deprecated."""
+    # TODO 3.0: Remove this test.
+    # pylint: disable=import-outside-toplevel
+    from pylint.config import load_results, save_results
```
DanielNoord, line 87: Is this necessary to pass? All of these tests will be removed in `3.0`, as per test file name 😄

DanielNoord (changes requested):
Oops... 😄

tests/test_deprecation.py
Pierre-Sassoulas: Not strictly but the CI pipeline is half done 😄

PR #6518, opened 2022-05-06
Add typing to `pylint.pyreverse.main` and `pylint.pyreverse.diadefslib`

DudeNr33:

## Type of Changes

## Description
Added typing to `main.py` and `diadefslib.py`.
`diagrams.py`, `inspector.py` and `utils.py` are next.
Let me know if you prefer smaller chunks or if I can update this PR with the remaining modules.

### DIFF
--- a/pylint/pyreverse/diadefslib.py
+++ b/pylint/pyreverse/diadefslib.py
@@ -6,13 +6,15 @@
 
 from __future__ import annotations
 
+import argparse
+from collections.abc import Generator
 from typing import Any
 
 import astroid
 from astroid import nodes
 
 from pylint.pyreverse.diagrams import ClassDiagram, PackageDiagram
-from pylint.pyreverse.inspector import Project
+from pylint.pyreverse.inspector import Linker, Project
 from pylint.pyreverse.utils import LocalsVisitor
 
 # diagram generators ##########################################################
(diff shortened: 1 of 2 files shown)


pylint/pyreverse/diadefslib.py
```diff
@@ -21,29 +23,30 @@
...
 
-    def __init__(self, linker, handler):
+    def __init__(self, linker: Linker, handler: DiadefsHandler) -> None:
         """Common Diagram Handler initialization."""
         self.config = handler.config
+        self.module_names: bool = False
         self._set_default_options()
         self.linker = linker
-        self.classdiagram = None  # defined by subclasses
+        self.classdiagram: ClassDiagram | None = None  # defined by subclasses
```
DanielNoord, line 32: In some instances of this pattern in `pylint` we have chosen to do:
```suggestion
        self.classdiagram: ClassDiagram  # defined by subclasses
```

To avoid having to do `if self.classdiagram` everywhere. Is that something we should consider here too?

DanielNoord (commented):
The size of this PR is perfect. Especially since none of the other reviewers are very proficient with this code base so we need some more time to review 😄

Pierre-Sassoulas (approved):
LGTM, The ``X | None`` indeed are harder to handle than simply ``X`` realizing where there are some of those is in part the reason why we need to have typing. That being said I think we could first add the typing, then refactor as required (unless the refactor is really simple to do).

DanielNoord:
> LGTM, The `X | None` indeed are harder to handle than simply `X` realizing where there are some of those is in part the reason why we need to have typing. That being said I think we could first add the typing, then refactor as required (unless the refactor is really simple to do).

We need to refactor because of the `| None` right now, see the added `ifs` in this PR. That's why I asked if they are avoidable.

pylint/pyreverse/diadefslib.py
Pierre-Sassoulas: A pattern I like is to have for something like this:
```python
from abc import ABC, abstractmethod

class Animal(ABC):

    @classmethod
    @property
    @abstractmethod
    def sound_maker_class(cls):
        raise NotImplementedError
    
    def __init__(self):
        self.sound_maker = sound_maker_class()

class Cat:
    sound_maker_class = Meower
```

DudeNr33:
Thank you for your input! Daniel's suggestion works just fine, so I took that approach. 
I will open separate PRs for the other modules to keep to PR size easier to handle and review.

Issue #6531, opened 2022-05-06
False Positive F0401: Unable to import when importing parent directory


orSolocate:
### Bug description

Bug opened from [SO issue](https://stackoverflow.com/questions/46089480/pytest-fixtures-redefining-name-from-outer-scope-pylint)

The case is where you run `pylint inner_module.py` where your imported module is located in the parent dir.

```
__init__.py
myapp.py
imported_module.py
subdir\
    __init__.py
   inner_module.py  
```

I think the **unable to import** messages can be pretty annoying and this should be solved or at least discussed.

### Command used

```shell
`pylint inner_module.py`
```

### Pylint output

```shell
`F0401: Unable to import 'imported_module'`
```

### Expected behavior

No error

### Pylint version

```shell
pylint 2.8.4
astroid 2.5.8
Python 3.7.5
```

### OS / Environment

Windows

Pierre-Sassoulas:
Well, apparently this code crash on 2.14.0-dev0 and 2.13.8.

```python
Exception on node <Decorators l.2 at 0x7f1be24aaa00> in file '/home/pierre/pylint/b.py'
Traceback (most recent call last):
  File "/home/pierre/pylint/pylint/utils/ast_walker.py", line 73, in walk
    callback(astroid)
  File "/home/pierre/pylint/pylint/checkers/unsupported_version.py", line 64, in visit_decorators
    self._check_typing_final(node)
  File "/home/pierre/pylint/pylint/checkers/unsupported_version.py", line 79, in _check_typing_final
    for decorator in decorators or uninferable_final_decorators(node):
  File "/home/pierre/pylint/pylint/checkers/utils.py", line 825, in uninferable_final_decorators
    import_node = decorator.expr.lookup(decorator.expr.name)[1][0]
IndexError: tuple index out of range
```

DanielNoord:
We should probably fix the crash, but the issue of handling `pytest` fixtures should probably be delegated to [pytest-pylint](https://github.com/carsongee/pytest-pylint).

mbyrnepr2:
> We should probably fix the crash

I can take a look.

mbyrnepr2:
I could be wrong but regarding the "Refined name from outer scope" message. Isn't this correct of Pylint in highlighting what is also mentioned in the SO (pytest docs):

[...]

PR #6532, opened 2022-05-06
Add an exception for `IndexError` inside `uninferable_final_decorator`

mbyrnepr2:
## Type of Changes

## Description

Refs #6531

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,10 +25,6 @@ Release date: TBA
 
   Closes #4301
 
-* Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
-
- Relates to #6531
-
 * By default the similarity checker will now ignore imports and ignore function signatures when computing
   duplication. If you want to keep the previous behaviour set ``ignore-imports`` and ``ignore-signatures`` to ``False``.
 
@@ -311,6 +307,9 @@ What's New in Pylint 2.13.9?
 ============================
 Release date: TBA
 
+* Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
+
+ Relates to #6531
 
 
 What's New in Pylint 2.13.8?
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,8 @@ Release date: TBA
 
   Closes #4301
 
+* Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
+
 * By default the similarity checker will now ignore imports and ignore function signatures when computing
   duplication. If you want to keep the previous behaviour set ``ignore-imports`` and ``ignore-signatures`` to ``False``.
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -27,6 +27,8 @@ Release date: TBA
 
 * Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
 
+ Relates to #6531
+
 * By default the similarity checker will now ignore imports and ignore function signatures when computing
   duplication. If you want to keep the previous behaviour set ``ignore-imports`` and ``ignore-signatures`` to ``False``.
 
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -307,6 +307,8 @@ Other Changes
 
 * Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
 
+ Relates to #6531
+
 * Pylint now expands the user path (i.e. ``~`` to ``home/yusef/``) and expands environment variables (i.e. ``home/$USER/$project``
   to ``home/yusef/pylint`` for ``USER=yusef`` and ``project=pylint``) for pyreverse's ``output-directory``,
   ``import-graph``, ``ext-import-graph``,  ``int-import-graph`` options, and the spell checker's ``spelling-private-dict-file``
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -305,6 +305,8 @@ Other Changes
 * By default the similarity checker will now ignore imports and ignore function signatures when computing
   duplication. If you want to keep the previous behaviour set ``ignore-imports`` and ``ignore-signatures`` to ``False``.
 
+* Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
+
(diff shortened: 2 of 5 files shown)


ChangeLog
```diff
@@ -25,6 +25,8 @@ Release date: TBA
 
   Closes #4301
 
+* Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
+
```
Pierre-Sassoulas, line 29:
```suggestion
* Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.

 Relates to #6531

```

doc/whatsnew/2.14.rst
```diff
@@ -305,6 +305,8 @@ Other Changes
 * By default the similarity checker will now ignore imports and ignore function signatures when computing
   duplication. If you want to keep the previous behaviour set ``ignore-imports`` and ``ignore-signatures`` to ``False``.
 
+* Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
+
```
Pierre-Sassoulas, line 309:
```suggestion

 Relates to #6531

```

pylint/checkers/utils.py
```diff
@@ -850,7 +850,7 @@ def uninferable_final_decorators(
         if isinstance(decorator, nodes.Attribute):
             try:
                 import_node = decorator.expr.lookup(decorator.expr.name)[1][0]
-            except AttributeError:
+            except (AttributeError, IndexError):
```
Pierre-Sassoulas, line 853: I'm wondering if we should distinguish the two errors and try to handle the case where we can't do ``[1][0]``. I guess Daniel will have an opinion :)

mbyrnepr2: The IndexError occurs when `pytest` is not imported in the module; the tuple is empty in that case:
```python
> /Users/markbyrne/programming/pylint/pylint/checkers/utils.py(853)uninferable_final_decorators()
-> import_node = decorator.expr.lookup(decorator.expr.name)[1][0]
(Pdb) decorator.expr.lookup(decorator.expr.name)
(<Module.builtins l.0 at 0x108d82940>, ())
```

mbyrnepr2: I believe the logic is simpler now with 38cbeebdca1242b8a539cc85239808d90b12a571 & it more closely matches the `nodes.Name` case beneath it in the same function.

mbyrnepr2: No, need to re-add the try-except since the primer test fails without it.

pylint/checkers/utils.py
```diff
@@ -849,9 +849,12 @@ def uninferable_final_decorators(
     for decorator in getattr(node, "nodes", []):
         if isinstance(decorator, nodes.Attribute):
             try:
-                import_node = decorator.expr.lookup(decorator.expr.name)[1][0]
+                _, import_nodes = decorator.expr.lookup(decorator.expr.name)
             except AttributeError:
```
DanielNoord, line 853: Looking at the failed test this should be fixable with an `isinstance` on `decorator.expr`. I think we assume `expr` to a `LocalsDictNodeNG` while apparently in those regression tests it is a `nodes.Attribute`. Should we try and fix that while we're at it?
In general I prefer explicitness over broader `try...excepts`.

pylint/checkers/utils.py
```diff
@@ -849,9 +849,12 @@ def uninferable_final_decorators(
...
                 continue
+            if not import_nodes:
```
DanielNoord, line 855: Could you add a comment about why this is the case? Just so we don't wonder about it later on.

tests/functional/r/regression/regression_6531_crash_index_error.py
```diff
@@ -0,0 +1,28 @@
+# pylint: disable=missing-docstring, redefined-outer-name
```
DanielNoord, line 1:
```suggestion
"""Regression test for https://github.com/PyCQA/pylint/issues/6531."""

# pylint: disable=missing-docstring, redefined-outer-name
```

In general I like to add links to regression issues. Only adding the number doesn't make it clear whether it is the issue from `GitHub`, `gitlab` or even the old `logilab` tracker. Especially for some of the lower numbers this is an issue. We don't know where `pylint` might be hosted in ten years time so better save than sorry!

DanielNoord (changes requested):

pylint/checkers/utils.py
mbyrnepr2: `decorator.expr` has no attribute `loookup`. I couldn't try adding a `hasattr` instead of an `isinstance` if I understood you.

pylint/checkers/utils.py
mbyrnepr2: Makes sense.

pylint/checkers/utils.py
DanielNoord: But normally it does right? My hunch is that this code works we for `@decorator.func()` but not for `@decorator.module.func()`. Apparently sometimes `expr` sometimes does have `lookup` so instead of using `hasattr` can't we reliable determine what instance expr has and whether that instance is supposed to have lookup?

tests/functional/r/regression/regression_6531_crash_index_error.py
mbyrnepr2: Fair point. Sometimes I'm unsure of the convention but in this case I could have been more clear.

DanielNoord: I think I'm a little too docstring-loving myself, so don't always trust me 😄 But in this case I think there is a real benefit!

mbyrnepr2: I completely agree with you. To be honest I rushed the changes a bit last night/today, so I'm not surprised :)

Pierre-Sassoulas (approved):
The new code is a lot easier to understand, that's great 👌

mbyrnepr2:
Thanks Pierre. I do want to address the suggestion from @DanielNoord later tonight. If we can wait that would be good. Otherwise I can follow up in another MR

DanielNoord (approved):
I'm on mobile so I can't really check if the coverage is correct. But the code LGTM.

Thanks @mbyrnepr2 for the extra effort you put in! I think this actually improves the status quo as well!

mbyrnepr2:
Thank you @DanielNoord & @Pierre-Sassoulas for your help 👍

PR #6536, opened 2022-05-07
Don't emit unsubscriptable-object for string annotations

### LINKED ISSUE #4369: False unsubscriptable-object positive with generic subprocess.Popen and __future__ annotations
A variant of #4034, just with annotations imported on python 3.8. Per that issue, outputting the error is arguably correct if the `from __future__ import annotations` import is not present, but when it is, this should not result in an error.

I'm guessing (not tested) just adding `subprocess.Popen` to `pylint.checkers.utils.SUBSCRIPTABLE_CLASSES_PEP585` would be a quick and dirty fix, but `subprocess.Popen` isn't actually mentioned in PEP 585 so it wouldn't feel quite right.

There might be some others similarly affected, mypy docs at https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime mention `os.PathLike` and `queue.Queue`.

### Steps to reproduce

```python
from subprocess import Popen, PIPE

def foo(popen: Popen[str]) -> None:
    pass

foo(Popen(["ls"], stdout=PIPE))
```

### Current behavior

```
t4.py:5:15: E1136: Value 'Popen' is unsubscriptable (unsubscriptable-object)
```

### Expected behavior

No unsubscriptable-object

### pylint --version output

```
pylint 2.7.4
astroid 2.5.3
Python 3.8.0 (default, Mar 30 2021, 18:43:10) 
[GCC 9.3.0]
 
```

### LINKED ISSUE #6523: from __future__ import annotations ignored
### Bug description

```python
from csv import DictWriter
from typing import IO


def f_is_for_fun(file: IO[str]) -> DictWriter[str]:
    return DictWriter(file, ["id", "name"])
```

Reference https://github.com/PyCQA/pylint/issues/6519

### Command used

```shell
pylint playground.py
```

### Pylint output

```shell
************* Module playground
…
playground.py:6:35: E1136: Value 'DictWriter' is unsubscriptable (unsubscriptable-object)
```

### Expected behavior

No error, as the python interpreter doesn't complain, nor does mypy, and https://github.com/PyCQA/pylint/issues/3320 is closed.

### Pylint version

```shell
pylint 2.13.8
astroid 2.11.4
Python 3.9.12 (tags/v3.9.12:b28265d, Mar 23 2022, 23:52:46) [MSC v.1929 64 bit (AMD64)]
```

cdce8p:
## Description
We previously checked if a class is explicitly mentioned in PEP 585 as being subscriptable. That frequently results in errors as some classes are only generic in type stubs. We could continue to update the set of subscriptable names, but that still wouldn't cover custom packages and is impractical.

With this PR, pylint will ignore all string annotations (explicit or via `from __future__ import annotations`). We might miss a few issues, however those aren't runtime errors and a type checker is likely used anyway witch would catch these.

Not emitting false-positives is more important than catching all errors here IMO.

Closes #4369
Closes #6523

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -303,6 +303,11 @@ Release date: TBA
 
   Closes #4369 and #6523
 
+* ``is_class_subscriptable_pep585_with_postponed_evaluation_enabled`` has been deprecated.
+  Use ``subscriptable_with_postponed_evaluation_enabled`` instead.
+
+  Ref #6536
+
 ..
   Insert your changelog randomly, it will reduce merge conflicts
   (Ie. not necessarily at the end)
--- a/ChangeLog
+++ b/ChangeLog
@@ -298,6 +298,11 @@ Release date: TBA
 
   Closes #6394
 
+* Don't emit ``unsubscriptable-object`` for string annotations.
+  Pylint doesn't check if class is only generic in type stubs only.
+
+  Closes #4369 and #6523
+
 ..
   Insert your changelog randomly, it will reduce merge conflicts
   (Ie. not necessarily at the end)
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -316,3 +316,8 @@ Other Changes
   Pylint doesn't check if class is only generic in type stubs only.
 
   Closes #4369 and #6523
+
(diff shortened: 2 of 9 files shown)


Pierre-Sassoulas (approved):
Agree with the sentiment about reducing false positives vs false negative.

PR #6537, opened 2022-05-07
Make the scope of ``PyLinter`` messages explicit

DanielNoord:

## Type of Changes

## Description

Ref. https://github.com/PyCQA/pylint/issues/3312#issuecomment-1120203912

In preparation of moving `PyLinter` to `BaseChecker` instead of `BaseTokenChecker` we need to make the scope of its messages explicit. For `TokenChecker` it is `LINE` by default.

Even if my proposed move to a `_MessageStateHandler` isn't accepted, this PR might be worth considering as it doesn't hurt making this explicit.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -31,6 +31,7 @@ from pylint.constants import (
     MSG_TYPES,
     MSG_TYPES_LONG,
     MSG_TYPES_STATUS,
+    WarningScope,
 )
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
LGTM !

jacobtylerwalls (approved):
Great, as I understand it, this avoids having to do `isinstance(PyLinter)` in `create_message_definition_from_tuple()` which would introduce a circular import. 👍🏻

Issue #5884, opened 2022-03-09
'try-except-raise' warnings on 3.9 but not 3.7


RemiCardona:
### Bug description

```python
import asyncio

async def coro():
    try:
        await asyncio.sleep(0.5)
    except asyncio.CancelledError:
        raise
    except ValueError:
        pass
```

No warning raised when running in a python 3.7 venv, `W0706(try-except-raise)` warning raised with an identical pylint call when running in a 3.9 venv.

`# pylint: disable=try-except-raise,useless-suppression` thus needed to make pylint happy on both 3.7 and 3.9.

### Command used

```shell
pylint demo.py
```

### Pylint output

```shell
************* Module bluelink.demo
demo.py:7: [W0706(try-except-raise), coro] The except handler raises immediately
```

### Expected behavior

No warning raised, as on 3.7. It's a false positive.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.7.12 (default, Mar  8 2022, 11:09:12) 
[GCC 11.2.1 20220115]

and

pylint 2.12.2
astroid 2.9.3
Python 3.9.10 (main, Mar  8 2022, 11:00:59) 
[GCC 11.2.1 20220115]
```

### OS / Environment

~amd64 Gentoo (outputs above) and Debian Buster (with different python patch-level versions)

Pierre-Sassoulas:
Thank you for opening the issue. Are you using the ``py-version`` option (set to 3.7) ?

RemiCardona:
Nope, I'm not using this option. And I can reproduce the bug with the provided sample and a plain call to `pylint demo.py`.

RemiCardona:
I tried setting `py-version` to both `3.7` and `3.9`, under both venvs and the results are the same as before.

jacobtylerwalls:
Thanks for writing. This is a false negative for `try-except-raise` in python 3.7. The warning is emitted correctly on Python 3.9.

Most likely, the failure to raise on Python 3.7 is because before Python 3.8 `asyncio.CancelledError` had a more complicated inheritance structure (it just aliased `concurrent.futures.CancelledError`). See also: https://github.com/PyCQA/pylint/issues/3083#issuecomment-841075906.

Ultimately, this is just a failure to infer `asyncio.CancelledError` in python 3.7:
```python
(Pdb) handler.type
<Attribute.CancelledError l.6 at 0x7fc4709ced90>
(Pdb) utils.safe_infer(handler.type)
[...]

PR #6540, opened 2022-05-07
Added try-except-raise message example

matusvalo:
## Type of Changes

## Description

Related #5953

### DIFF
--- a/doc/data/messages/t/try-except-raise/bad.py
+++ b/doc/data/messages/t/try-except-raise/bad.py
@@ -1,0 +1,4 @@
+try:
+    1 / 0
+except:  # [try-except-raise]
+    raise
--- a/doc/data/messages/t/try-except-raise/bad.py
+++ b/doc/data/messages/t/try-except-raise/bad.py
@@ -1,4 +1,4 @@
 try:
     1 / 0
-except:  # [try-except-raise]
+except ZeroDivisionError as e:  # [try-except-raise]
     raise
--- a/doc/data/messages/t/try-except-raise/good.py
+++ b/doc/data/messages/t/try-except-raise/good.py
@@ -1,4 +1,4 @@
-# The try except might be remove entirely:
+# The try except might be removed entirely:
 1 / 0
 
 # Or another more detailed exception can be raised:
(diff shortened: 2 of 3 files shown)


doc/data/messages/t/try-except-raise/bad.py
```diff
@@ -0,0 +1,4 @@
+try:
+    1 / 0
+except:  # [try-except-raise]
+    raise
```
Pierre-Sassoulas, line 4:
```suggestion
try:
    1 / 0
except ZeroDivisionError as e:  # [try-except-raise]
    raise
```

doc/data/messages/t/try-except-raise/good.py
```diff
@@ -0,0 +1,4 @@
+try:
+    1 / 0
+except ZeroDivisionError as e:
+    raise ValueError from e
```
Pierre-Sassoulas, line 4:
```suggestion
# The try except might be remove entirely:
1 / 0

# Or another more detailed exception can be raised:
try:
    1 / 0
except ZeroDivisionError as e:
    raise ValueError("The area of the rectangle cannot be zero") from e
```

jacobtylerwalls:
Thanks. Is `details.rst` a good place to summarize the discussion from https://github.com/PyCQA/pylint/issues/5884#issuecomment-1064310365 involving filtering exceptions?

Pierre-Sassoulas (changes requested):
I agree that we should add a detail.rst, to be clear that there's a legitimate use case for re raising immediately but that this check is smart enough to not have a false positive for it. 

I.e. with the following inheritance tree:
```
      +-- ArithmeticError
      |    +-- FloatingPointError
      |    +-- OverflowError
      |    +-- ZeroDivisionError
```

```python      
try:
    some_calculation(...)
    raise
except ArithmeticError:
   pass
```   
will not raise because ZeroDivisionError  inherit from ArithmeticError.

doc/data/messages/t/try-except-raise/good.py
```diff
@@ -0,0 +1,8 @@
+# The try except might be remove entirely:
```
jacobtylerwalls, line 1:
```suggestion
# The try except might be removed entirely:
```

matusvalo:
Added details.rst. Please review if now it is OK.

Pierre-Sassoulas (approved):
👍

Issue #6543, opened 2022-05-07
pyreverse Aggregation and Composition arrows are same


khashashin:
### Bug description

pyreverse generates same diagram for aggregation and composition. Here is the sample code:

```python
# Aggregation
class Student:
    def __init__(self, id):
        self._id = id

    def registration_number(self, department_id):
        return str(self._id) + '-' + department_id


class Department:
    def __init__(self, id, student):
        self._id = id
        self._student = student

    def student_registration(self):
        return self._student.registration_number(self._id)


if __name__ == '__main__':
    student = Student(50)
    department = Department('ENG', student)

    print(department.student_registration())
```
---
```python
# Composition
class Student:
    def __init__(self, id):
        self._id = id

    def registration_number(self, department_id):
        return str(self._id) + '-' + department_id


class Department:
    def __init__(self, department_id, student_id):
        self._id = department_id
        self._student = Student(student_id)

    def student_registration(self):
        return self._student.registration_number(self._id)


if __name__ == '__main__':
    department = Department('ENG', 50)

    print(department.student_registration())

```

### Command used

```shell
pyreverse -o png -p composition.py .
pyreverse -o png -p aggregation.py .
```

### Pylint output

```shell
Format png is not supported natively. Pyreverse will try to generate it using Graphviz...
parsing ./__init__.py...
parsing /mnt/d/GitHub/test/django_with_sass/experiments/composition.py...
parsing /mnt/d/GitHub/test/django_with_sass/experiments/__init__.py...

Format png is not supported natively. Pyreverse will try to generate it using Graphviz...
parsing ./__init__.py...
parsing /mnt/d/GitHub/test/django_with_sass/experiments/aggregation.py...
parsing /mnt/d/GitHub/test/django_with_sass/experiments/__init__.py...
```

### Expected behavior

Different Arrow type.

![image](https://user-images.githubusercontent.com/17837758/167272523-36270d0a-6c63-493e-bb68-52c282f4d8ea.png)
[Source](https://www.visual-paradigm.com/guide/uml-unified-modeling-language/uml-class-diagram-tutorial/)

### Pylint version

```shell
pylint --version
pylint 2.13.8
astroid 2.11.4
Python 3.10.0 (default, Dec  3 2021, 00:34:18) [GCC 9.3.0]
```

### OS / Environment

Windows 11
WSL2

### Additional dependencies

sudo apt install graphviz

dgutson:
CC @qequ

[...]

PR #6547, opened 2022-05-08
Add typing to `pylint.pyreverse.diagrams.py`

DudeNr33:

## Type of Changes

## Description

Next chunk for typing of `pyreverse`.
This one was a bit trickier. The class hierarchy in this module is not ideal - `PackageDiagram` is a subclass of `ClassDiagram` and reuses some of its methods, so being more specific than `DiagramEntity` was not possible in all cases.

This could probably be solved with a refactoring to have a hierarchy like this:
```mermaid
classDiagram
  class Diagram {
    <<ABC>>
  }
  class PackageDiagram 
  class ClassDiagram 
  Diagram <|-- PackageDiagram
  Diagram <|-- ClassDiagram
```
But first I have to figure out which methods need to be provided by the base class and which are specific to the specific diagram class.
That should go into a separate PR.

### DIFF
--- a/pylint/pyreverse/diagrams.py
+++ b/pylint/pyreverse/diagrams.py
@@ -3,6 +3,7 @@
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
 """Diagram objects."""
+
 from __future__ import annotations
 
 from collections.abc import Iterable
@@ -45,7 +46,7 @@ class DiagramEntity(Figure):
 
     default_shape = ""
 
-    def __init__(self, title: str = "No name", node: NodeNG = None) -> None:
+    def __init__(self, title: str = "No name", node: NodeNG | None = None) -> None:
         super().__init__()
         self.title = title
         self.node: NodeNG = node if node else NodeNG()
@@ -78,6 +79,7 @@ class ClassDiagram(Figure, FilterMixIn):
         FilterMixIn.__init__(self, mode)
         Figure.__init__(self)
         self.title = title
+        # TODO: Specify 'Any' after refactor of `DiagramEntity`
         self.objects: list[Any] = []
         self.relationships: dict[str, list[Relationship]] = {}
         self._nodes: dict[NodeNG, DiagramEntity] = {}
--- a/pylint/pyreverse/diagrams.py
+++ b/pylint/pyreverse/diagrams.py
@@ -11,7 +11,6 @@ from typing import Any
 
 import astroid
 from astroid import nodes
-from astroid.nodes import ClassDef, FunctionDef, ImportFrom, Module, NodeNG
 
 from pylint.checkers.utils import decorated_with_property
 from pylint.pyreverse.utils import FilterMixIn, is_interface
@@ -46,10 +45,12 @@ class DiagramEntity(Figure):
 
     default_shape = ""
 
-    def __init__(self, title: str = "No name", node: NodeNG | None = None) -> None:
+    def __init__(
+        self, title: str = "No name", node: nodes.NodeNG | None = None
+    ) -> None:
         super().__init__()
         self.title = title
-        self.node: NodeNG = node if node else NodeNG()
+        self.node: nodes.NodeNG = node if node else nodes.NodeNG()
         self.shape = self.default_shape
 
 
@@ -64,10 +65,10 @@ class ClassEntity(DiagramEntity):
 
     default_shape = "class"
 
-    def __init__(self, title: str, node: ClassDef) -> None:
+    def __init__(self, title: str, node: nodes.ClassDef) -> None:
         super().__init__(title=title, node=node)
         self.attrs: list[str] = []
-        self.methods: list[FunctionDef] = []
+        self.methods: list[nodes.FunctionDef] = []
 
 
 class ClassDiagram(Figure, FilterMixIn):
@@ -82,7 +83,7 @@ class ClassDiagram(Figure, FilterMixIn):
         # TODO: Specify 'Any' after refactor of `DiagramEntity`
         self.objects: list[Any] = []
         self.relationships: dict[str, list[Relationship]] = {}
-        self._nodes: dict[NodeNG, DiagramEntity] = {}
+        self._nodes: dict[nodes.NodeNG, DiagramEntity] = {}
 
     def get_relationships(self, role: str) -> Iterable[Relationship]:
         # sorted to get predictable (hence testable) results
@@ -111,7 +112,7 @@ class ClassDiagram(Figure, FilterMixIn):
                 return rel
         raise KeyError(relation_type)
 
-    def get_attrs(self, node: ClassDef) -> list[str]:
+    def get_attrs(self, node: nodes.ClassDef) -> list[str]:
         """Return visible attributes, possibly with class name."""
         attrs = []
         properties = [
@@ -132,7 +133,7 @@ class ClassDiagram(Figure, FilterMixIn):
             attrs.append(node_name)
         return sorted(attrs)
 
-    def get_methods(self, node: ClassDef) -> list[FunctionDef]:
+    def get_methods(self, node: nodes.ClassDef) -> list[nodes.FunctionDef]:
         """Return visible methods."""
         methods = [
             m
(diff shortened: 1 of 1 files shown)


pylint/pyreverse/diagrams.py
```diff
@@ -28,59 +43,73 @@ def __init__(self, from_object, to_object, relation_type, name=None):
...
-    def __init__(self, title, mode):
+    def __init__(self, title: str, mode: str) -> None:
         FilterMixIn.__init__(self, mode)
         Figure.__init__(self)
         self.title = title
-        self.objects = []
-        self.relationships = {}
-        self._nodes = {}
-        self.depends = []
+        self.objects: list[Any] = []
```
DudeNr33, line 81: Not putting `Any` here would make `mypy` complain in `writer.py`. This can be resolved after the refactoring is done.

pylint/pyreverse/diagrams.py
```diff
@@ -232,29 +268,29 @@ def get_module(self, name, node):
...
         """Extract relationships between nodes in the diagram."""
         super().extract_relationships()
-        for obj in self.classes():
+        for class_obj in self.classes():
             # ownership
             try:
-                mod = self.object_from_node(obj.node.root())
-                self.add_relationship(obj, mod, "ownership")
+                mod = self.object_from_node(class_obj.node.root())
+                self.add_relationship(class_obj, mod, "ownership")
```
DudeNr33, line 285: Those two lines were not covered before.
Funnily enough, this is (at least part) of what would be needed for #6543, but currently no `Printer` knows how to handle "ownership" relationships anyway.

pylint/pyreverse/diagrams.py
```diff
@@ -3,8 +3,14 @@
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
 """Diagram objects."""
+from __future__ import annotations
```
DanielNoord, line 6:
```suggestion

from __future__ import annotations
```

pylint/pyreverse/diagrams.py
```diff
@@ -3,8 +3,14 @@
...
+
+from collections.abc import Iterable
+from typing import Any
+
 import astroid
 from astroid import nodes
+from astroid.nodes import ClassDef, FunctionDef, ImportFrom, Module, NodeNG
```
DanielNoord, line 13: We already import `nodes` above. Did you explicitly want to use the shorthand versions?

pylint/pyreverse/diagrams.py
```diff
@@ -28,59 +43,73 @@ def __init__(self, from_object, to_object, relation_type, name=None):
 class DiagramEntity(Figure):
     """A diagram object, i.e. a label associated to an astroid node."""
 
-    def __init__(self, title="No name", node=None):
+    default_shape = ""
+
+    def __init__(self, title: str = "No name", node: NodeNG = None) -> None:
```
DanielNoord, line 48: `Pylance`/`pyright` warns about this because `None != NodeNG`. Perhaps:

```suggestion
    def __init__(self, title: str = "No name", node: NodeNG | None = None) -> None:
```

pylint/pyreverse/diagrams.py
DanielNoord: Let's put a `TODO` here so we don't forgot about it.

```suggestion
        # TODO: Specify 'Any' after refactor of `DiagramEntity`
        self.objects: list[Any] = []
```

pylint/pyreverse/diagrams.py
```diff
@@ -136,30 +165,30 @@ def class_names(self, nodes_lst):
...
 
-    def object_from_node(self, node):
+    def object_from_node(self, node: NodeNG) -> DiagramEntity:
         """Return the diagram object mapped to node."""
         return self._nodes[node]
 
-    def classes(self):
+    def classes(self) -> list[ClassEntity]:
         """Return all class nodes in the diagram."""
-        return [o for o in self.objects if isinstance(o.node, nodes.ClassDef)]
```
DanielNoord, line 153: 👍

DanielNoord (changes requested):
Small nitpicks.

I noticed ``mypy --strict`` still does report some issue on this file. But it's probably good to do this as a first step and go through all other modules and then see what ``mypy --strict pyreverse`` comes up with.

pylint/pyreverse/diagrams.py
DudeNr33: Right, the `fixme` check was disabled for our codebase some days ago, right? So we can now use it without an `pylint disabe` now. 👍

pylint/pyreverse/diagrams.py
DudeNr33: Well, more a "have to" instead of "want to": `ClassDiagram` defines a method named `nodes`, and after this method declaration you can't use the imported `nodes` from `astroid` any more in the scope of this class definition.
But I can also remove the `nodes` method - from what I've seen so far it is never used anyways.

pylint/pyreverse/diagrams.py
DudeNr33: Thanks for catching that. I thought `node: NodeNG = None` would be implicitly equivalent to `node: Optional[NodeNG] = None` (and therefore `NodeNG | None`), but I am also not a typing expert. So these PRs are a good practice for me. 😁

pylint/pyreverse/diagrams.py
DudeNr33: Removed the `nodes()` method in the latest commit and use `nodes` instead of shorthand versions.

DudeNr33:
> Small nitpicks.
> 
> I noticed `mypy --strict` still does report some issue on this file. But it's probably good to do this as a first step and go through all other modules and then see what `mypy --strict pyreverse` comes up with.

Thank you for the review!
With the latest commit, I only get the following output for `diagram.py` when running with `--strict`:
```
pylint/pyreverse/diagrams.py:80: error: Call to untyped function "__init__" in typed context  [no-untyped-call]
pylint/pyreverse/diagrams.py:128: error: Call to untyped function "show_attr" in typed context  [no-untyped-call]
pylint/pyreverse/diagrams.py:144: error: Call to untyped function "show_attr" in typed context  [no-untyped-call]
pylint/pyreverse/diagrams.py:146: error: Returning Any from function declared to return "Union[SupportsDunderLT, SupportsDunderGT]"  [no-any-return]
pylint/pyreverse/diagrams.py:197: error: Call to untyped function "is_interface" in typed context  [no-untyped-call]
```
The calls to untyped functions are valid, but as we said in the previous PR, I will add the typing module per module, and those errors are due to those still untyped modules.

I don't understand the `no-any-return`.  It is this method:
```python
    def get_methods(self, node: nodes.ClassDef) -> list[nodes.FunctionDef]:
        """Return visible methods."""
        methods = [
            m
            for m in node.values()
            if isinstance(m, nodes.FunctionDef)
            and not isinstance(m, astroid.objects.Property)
            and not decorated_with_property(m)
            and self.show_attr(m.name)
        ]
        return sorted(methods, key=lambda n: n.name)
```

pylint/pyreverse/diagrams.py
DanielNoord: Yes!

DanielNoord:
> > Small nitpicks.
> > I noticed `mypy --strict` still does report some issue on this file. But it's probably good to do this as a first step and go through all other modules and then see what `mypy --strict pyreverse` comes up with.
> 
> Thank you for the review! With the latest commit, I only get the following output for `diagram.py` when running with `--strict`:
> 
> ```
> pylint/pyreverse/diagrams.py:80: error: Call to untyped function "__init__" in typed context  [no-untyped-call]
> pylint/pyreverse/diagrams.py:128: error: Call to untyped function "show_attr" in typed context  [no-untyped-call]
> pylint/pyreverse/diagrams.py:144: error: Call to untyped function "show_attr" in typed context  [no-untyped-call]
> pylint/pyreverse/diagrams.py:146: error: Returning Any from function declared to return "Union[SupportsDunderLT, SupportsDunderGT]"  [no-any-return]
> pylint/pyreverse/diagrams.py:197: error: Call to untyped function "is_interface" in typed context  [no-untyped-call]
> ```
> 
> The calls to untyped functions are valid, but as we said in the previous PR, I will add the typing module per module, and those errors are due to those still untyped modules.
> 
> I don't understand the `no-any-return`. It is this method:
> 
> ```python
>     def get_methods(self, node: nodes.ClassDef) -> list[nodes.FunctionDef]:
>         """Return visible methods."""
>         methods = [
>             m
>             for m in node.values()
>             if isinstance(m, nodes.FunctionDef)
>             and not isinstance(m, astroid.objects.Property)
>             and not decorated_with_property(m)
>             and self.show_attr(m.name)
>         ]
>         return sorted(methods, key=lambda n: n.name)
> ```

Yeah I saw that other issue as well. I think it has something to do with the `sorted` call, but I don't know. The typing seems correct though, so I'd say let's hold of on fixing this for now.

pylint/pyreverse/diagrams.py
DanielNoord: Yeah, for some reason `mypy` doesn't complain about it but it doesn't hurt to be explicit.

DanielNoord (approved):

DudeNr33:
> Yeah I saw that other issue as well. I think it has something to do with the `sorted` call, but I don't know. The typing seems correct though, so I'd say let's hold of on fixing this for now.

👍  agreed. I played a bit with it, storing the result of the `sorted` call before returning it, and explicitly annotating that variable with the return type of the function, but `mypy` was still not satisfied.

Issue #6538, opened 2022-05-07
`import-error` when importing another module from same namespace


jacobtylerwalls:
### Bug description

After #6405, linting `b.py` below raises `import-error`, so long as the parent of `test` is on `sys.path`:

```
test
    a.py
    b.py
```
**a.py**
```python
"""This module is a"""
print("You are importing a")
```
**b.py**
```python
"""This module imports a"""
import a
```

### Command used

```shell
pylint test/b.py
--or (no difference)--
cd test
pylint b.py
```

### Pylint output

```shell
************* Module b
test/b.py:2:0: E0401: Unable to import 'a' (import-error)

--------------------------------------------------------------------
Your code has been rated at 0.00/10 (previous run: 10.00/10, -10.00)
```

### Expected behavior

No msg, since `b.py` executes fine:
```shell
$ python3 b.py
You are importing a
```

### Pylint version

```shell
pylint 2.14.0-dev0
astroid 2.12.0-dev0
Python 3.10.1 (v3.10.1:2cd268a3a9, Dec  6 2021, 14:28:59) [Clang 13.0.0 (clang-1300.0.29.3)]
```

DanielNoord:
Hm, we might want to revert #6405 so we can unblock the `beta`. I'd rather release `2.14` with the `argparse` changes and delay the `namespace` changes (again) than having to fiddle with this for another two weeks and block the release.

I saw we had over 400 items already on the milestone so it's probably best to release soon.

Pierre-Sassoulas:
Opened https://github.com/PyCQA/pylint/pull/6548 to revert.

jacobtylerwalls:
What makes this a little tricky to reproduce is that the parent of the namespace directory has to be on `sys.path`.

jacobtylerwalls:
The test isolation issue with the regression test doesn't need block the beta.

jacobtylerwalls:
Fixed in #6548 and tested in #6553

PR #6548, opened 2022-05-08
Revert "Do not append namespace packages to sys.path (#6405)"

Pierre-Sassoulas:
## Type of Changes

## Description

This reverts commit 603be8408a5d2d98283f41cfd9a94998fd7cfd3c. See https://github.com/PyCQA/pylint/issues/6538#issuecomment-1120373164

Should we reopen https://github.com/PyCQA/pylint/issues/5226, and https://github.com/PyCQA/pylint/issues/2648 ?


jacobtylerwalls:
I hope to write a regression test case. Maybe we should merge that before reverting this. I think in trying to distill a minimal example I simplified it too much 😓

jacobtylerwalls:
> I think in trying to distill a minimal example I simplified it too much 😓

Turned out it was just that generic names like "test", "a", and "b" were ambiguous on my system. Using unique names helps to reproduce. I'll try to get a PR going with a test.

jacobtylerwalls:
Er.... it was actually that the parent of the namespace directory needed to be on `sys.path`! See #6553 for the test.

jacobtylerwalls:
> Should we reopen https://github.com/PyCQA/pylint/issues/5226, and https://github.com/PyCQA/pylint/issues/2648 ?

👍🏻 And add them to 2.15

jacobtylerwalls (approved):

DanielNoord (approved):

Pierre-Sassoulas:
@DanielNoord should we merge #5235 then ? Permitting user to disable sys.path patching when it make sense could be a temporary solution ?

DanielNoord:
> @DanielNoord should we merge #5235 then ? Permitting user to disable sys.path patching when it make sense could be a temporary solution ?

Like I said previously, I don't think so until we have fully investigated the issue. We would only end up having to support an option that does not mimic actual `Python` behaviour. Until we are sure we can't fix our handling of namespace packages we shouldn't do so I think.

PR #6549, opened 2022-05-08
Finish typing of `pylint.pyreverse.utils`

DudeNr33:

## Type of Changes

## Description

This module already had some typing, added typing to the classes and functions that did not have any yet.

Running with `mypy --strict` four issues remain, which I can't wrap my head around why `mypy` complains about them:
```
pylint [typing-pyreverse-p3] % mypy --strict pylint/pyreverse/utils.py           
pylint/pyreverse/utils.py:72: error: Returning Any from function declared to return "bool"  [no-any-return]
pylint/pyreverse/utils.py:77: error: Returning Any from function declared to return "bool"  [no-any-return]
pylint/pyreverse/utils.py:213: error: Returning Any from function declared to return "str"  [no-any-return]
pylint/pyreverse/utils.py:215: error: Returning Any from function declared to return "str"  [no-any-return]
Found 4 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/pylint/pyreverse/utils.py
+++ b/pylint/pyreverse/utils.py
@@ -170,7 +170,7 @@ class ASTWalker:
             e_method, l_method = methods
         return e_method, l_method
 
-    def visit(self, node: nodes.NodeNG) -> None:
+    def visit(self, node: nodes.NodeNG) -> Any:
         """Walk on the tree from <node>, getting callbacks from handler."""
         method = self.get_callbacks(node)[0]
         if method is not None:
--- a/pylint/pyreverse/utils.py
+++ b/pylint/pyreverse/utils.py
@@ -11,9 +11,16 @@ import re
 import shutil
 import subprocess
 import sys
+from collections.abc import Callable
+from typing import TYPE_CHECKING, Any
 
 import astroid
 from astroid import nodes
+
+if TYPE_CHECKING:
+    from pylint.pyreverse.diadefslib import DiaDefGenerator
+    from pylint.pyreverse.inspector import Linker
+
 
 RCFILE = ".pyreverserc"
 
@@ -46,7 +53,7 @@ PRIVATE = re.compile(r"^__(_*[^\W_])+_?$")
 PROTECTED = re.compile(r"^_\w*$")
 
 
-def get_visibility(name):
+def get_visibility(name: str) -> str:
     """Return the visibility from a name: public, protected, private or special."""
     if SPECIAL.match(name):
         visibility = "special"
@@ -60,37 +67,18 @@ def get_visibility(name):
     return visibility
 
 
-ABSTRACT = re.compile(r"^.*Abstract.*")
-FINAL = re.compile(r"^[^\W\da-z]*$")
-
-
-def is_abstract(node):
-    """Return true if the given class node correspond to an abstract class
-    definition.
-    """
-    return ABSTRACT.match(node.name)
-
-
-def is_final(node):
-    """Return true if the given class/function node correspond to final
-    definition.
-    """
-    return FINAL.match(node.name)
-
-
-def is_interface(node):
+def is_interface(node: nodes.ClassDef) -> bool:
     # bw compatibility
     return node.type == "interface"
 
 
-def is_exception(node):
+def is_exception(node: nodes.ClassDef) -> bool:
     # bw compatibility
     return node.type == "exception"
 
 
 # Helpers #####################################################################
 
-_CONSTRUCTOR = 1
 _SPECIAL = 2
 _PROTECTED = 4
 _PRIVATE = 8
@@ -111,7 +99,7 @@ VIS_MOD = {
 class FilterMixIn:
     """Filter nodes according to a mode and nodes' visibility."""
 
-    def __init__(self, mode):
+    def __init__(self, mode: str) -> None:
         """Init filter modes."""
         __mode = 0
         for nummod in mode.split("+"):
@@ -121,7 +109,7 @@ class FilterMixIn:
                 print(f"Unknown filter mode {ex}", file=sys.stderr)
         self.__mode = __mode
 
-    def show_attr(self, node):
+    def show_attr(self, node: nodes.NodeNG | str) -> bool:
         """Return true if the node should be treated."""
         visibility = get_visibility(getattr(node, "name", node))
         return not self.__mode & VIS_MOD[visibility]
@@ -137,11 +125,17 @@ class ASTWalker:
     the node in lower case
     """
 
-    def __init__(self, handler):
+    def __init__(self, handler: DiaDefGenerator | Linker | LocalsVisitor) -> None:
         self.handler = handler
-        self._cache = {}
+        self._cache: dict[
+            type[nodes.NodeNG],
+            tuple[
+                Callable[[nodes.NodeNG], Any] | None,
+                Callable[[nodes.NodeNG], Any] | None,
+            ],
+        ] = {}
 
-    def walk(self, node, _done=None):
+    def walk(self, node: nodes.NodeNG, _done: set[nodes.NodeNG] | None = None) -> None:
         """Walk on the tree from <node>, getting callbacks from handler."""
         if _done is None:
             _done = set()
@@ -155,7 +149,11 @@ class ASTWalker:
         self.leave(node)
         assert node.parent is not node
 
-    def get_callbacks(self, node):
+    def get_callbacks(
+        self, node: nodes.NodeNG
+    ) -> tuple[
+        Callable[[nodes.NodeNG], Any] | None, Callable[[nodes.NodeNG], Any] | None
+    ]:
         """Get callbacks from handler for the visited node."""
         klass = node.__class__
         methods = self._cache.get(klass)
@@ -173,13 +171,13 @@ class ASTWalker:
             e_method, l_method = methods
         return e_method, l_method
 
-    def visit(self, node):
+    def visit(self, node: nodes.NodeNG) -> None:
         """Walk on the tree from <node>, getting callbacks from handler."""
         method = self.get_callbacks(node)[0]
         if method is not None:
             method(node)
 
-    def leave(self, node):
+    def leave(self, node: nodes.NodeNG) -> None:
         """Walk on the tree from <node>, getting callbacks from handler."""
         method = self.get_callbacks(node)[1]
         if method is not None:
@@ -189,11 +187,11 @@ class ASTWalker:
 class LocalsVisitor(ASTWalker):
     """Visit a project by traversing the locals dictionary."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(self)
-        self._visited = set()
+        self._visited: set[nodes.NodeNG] = set()
 
-    def visit(self, node):
+    def visit(self, node: nodes.NodeNG) -> Any:
         """Launch the visit starting from the given node."""
         if node in self._visited:
             return None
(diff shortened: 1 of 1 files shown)


pylint/pyreverse/utils.py
```diff
@@ -60,37 +67,18 @@ def get_visibility(name):
...
-    definition.
-    """
-    return ABSTRACT.match(node.name)
-
-
-def is_final(node):
-    """Return true if the given class/function node correspond to final
-    definition.
-    """
-    return FINAL.match(node.name)
```
DudeNr33, line 78: Dead code -> removed instead of adding typing

pylint/pyreverse/utils.py
```diff
@@ -60,37 +67,18 @@ def get_visibility(name):
...
 
-def is_exception(node):
+def is_exception(node: nodes.ClassDef) -> bool:
     # bw compatibility
     return node.type == "exception"
 
 
 # Helpers #####################################################################
 
-_CONSTRUCTOR = 1
```
DudeNr33, line 93: Dead code

pylint/pyreverse/utils.py
```diff
@@ -155,7 +149,11 @@ def walk(self, node, _done=None):
         self.leave(node)
         assert node.parent is not node
 
-    def get_callbacks(self, node):
+    def get_callbacks(
+        self, node: nodes.NodeNG
+    ) -> tuple[
+        Callable[[nodes.NodeNG], Any] | None, Callable[[nodes.NodeNG], Any] | None
+    ]:
```
DudeNr33, line 156: There must be a better way to make a shorthand notation for it (maybe by creating a `TypeVar`?), but I could not come up with it.
Maybe one of the typing pros @DanielNoord or @cdce8p can help?

pylint/pyreverse/utils.py
```diff
@@ -60,37 +67,18 @@ def get_visibility(name):
...
-    """Return true if the given class/function node correspond to final
-    definition.
-    """
-    return FINAL.match(node.name)
-
-
-def is_interface(node):
+def is_interface(node: nodes.ClassDef) -> bool:
     # bw compatibility
     return node.type == "interface"
```
Pierre-Sassoulas, line 72: So, here, mypy thinks we return "Any" ?!

DudeNr33: Yes, no clue why. Maybe because the `type` attribute is monkey-patched on the `node` class.
But even VS Code recognises `node.type` correctly as `str`.

Pierre-Sassoulas: https://github.com/python/mypy/issues/5697#issuecomment-425738430

DudeNr33: Thank you, that explains why this happens here.
On lines 213 and 215 I still don't know why `mypy` complains here:
```python
def get_annotation_label(ann: nodes.Name | nodes.NodeNG) -> str:
    if isinstance(ann, nodes.Name) and ann.name is not None:
        return ann.name. # <-- l.213
    if isinstance(ann, nodes.NodeNG):
        return ann.as_string()  # <-- l.215
    return ""
```
Using `typing.reveal_type()`, both `ann.name` and `ann.as_string()` are recognised as `str`, not `Any`.

pylint/pyreverse/utils.py
DanielNoord: You can use:
`MyRecurringType = Callable[[int, int, int, int, int, int, ...], float]` at the top of the file just below the imports and then re-use it throughout the file.
The only thing you need to look out for is that `Callable` from `collections.abc` then needs to be imported from `typing`. As this way `MyRecurringType` is defined at runtime and on <3.8 `Callable` isn't subscriptable at runtime.

pylint/pyreverse/utils.py
```diff
@@ -137,11 +125,17 @@ class ASTWalker:
...
-    def __init__(self, handler):
+    def __init__(self, handler: DiaDefGenerator | Linker | LocalsVisitor) -> None:
         self.handler = handler
-        self._cache = {}
-
-    def walk(self, node, _done=None):
+        self._cache: dict[
+            type[nodes.NodeNG],
+            tuple[
+                Callable[[nodes.NodeNG], Any] | None,
```
DanielNoord, line 133: Don't we know what these `Callable` return?

pylint/pyreverse/utils.py
```diff
@@ -189,11 +187,11 @@ def leave(self, node):
...
     """Visit a project by traversing the locals dictionary."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(self)
-        self._visited = set()
+        self._visited: set[nodes.NodeNG] = set()
 
-    def visit(self, node):
+    def visit(self, node: nodes.NodeNG) -> Any:
```
DanielNoord, line 194: Is this truly `Any`? Don't all `visit_` methods return `None`?

pylint/pyreverse/utils.py
cdce8p: For the callable parameter type, take a look at my comment here as it's not completely correct (but probably the most practical):
https://github.com/PyCQA/pylint/blob/8df920b341a872b85754c910598d27a8846988f9/pylint/utils/ast_walker.py#L18-L24

--
As a side note: Shouldn't the return type be `None` instead of `Any`?

pylint/pyreverse/utils.py
cdce8p: Although we have already added some annotations to astroid, mypy will not use them until we add a `py.typed` file (once they are done). Until then all astroid types are basically inferred as `Any`.

Pierre-Sassoulas: https://github.com/PyCQA/astroid/issues/1287

pylint/pyreverse/utils.py
DudeNr33: As mentioned above: most of the time `None`, but for the one `leave_project` method it is a tuple of `ClassDiagram` and optionally `PackageDiagram`.  So a more precise type would be:
`Callable[[nodes.NodeNG],  None | tuple[ClassDiagram] | tuple[PackageDiagram, ClassDiagram]]`

DanielNoord: I'd be in favour of using the narrowed down type if we can. We can always change it after a refactor later.

cdce8p:
> I'm trying to incorporate the last changes requested by @DanielNoord, but I'm running into strange problems.

Try `Tuple[..]` with a capital first letter. Support for `tuple[..]` was added in Python 3.9. It's possible to use it in string annotations since they aren't parsed and thus don't need to be valid Python code.

DudeNr33:
I think as those lines are in the `if TYPE_CHECKING` block we should not have any runtime problems anyway, right?
But thanks for the hint, I changed it accordingly. 👍 

I could solve my `pre-commit` problems by reinstalling the pre-commit hooks (`pre-commit uninstall`, then `pre-commit install`).

cdce8p:
> I think as those lines are in the `if TYPE_CHECKING` block we should not have any runtime problems anyway, right? But thanks for the hint, I changed it accordingly. 👍

Yeah, technically true. I'm not sure though mypy knows that the type alias is defined in a `TYPE_CHECKING` block. Probably not yet implemented.

> ~I could solve my `pre-commit` problems by reinstalling the pre-commit hooks (`pre-commit uninstall`, then `pre-commit install`).~ Ok I'm completely lost now. Now the errors show up in CI, but not on my own machine. `mypy` version in the pre-commit hooks are both v0.950 in CI and my laptop.

Similar issue with PEP 604, i.e. Union operator. Use `Union` and `Optional` instead that should fix it.

pylint/pyreverse/utils.py
```diff
@@ -11,10 +11,23 @@
...
 from astroid import nodes
 
+if TYPE_CHECKING:
+    from pylint.pyreverse.diadefslib import DiaDefGenerator
+    from pylint.pyreverse.diagrams import ClassDiagram, PackageDiagram
+    from pylint.pyreverse.inspector import Linker
+
+    _CallbackT = Callable[
+        [nodes.NodeNG],
+        Union[Tuple[ClassDiagram], Tuple[PackageDiagram, ClassDiagram], None],
```
cdce8p, line 26: Returning tuples of different lengths is probably not the best API design

pylint/pyreverse/utils.py
```diff
@@ -137,11 +131,14 @@ class ASTWalker:
...
     """
 
-    def __init__(self, handler):
+    def __init__(self, handler: DiaDefGenerator | Linker | LocalsVisitor) -> None:
         self.handler = handler
-        self._cache = {}
+        self._cache: dict[
+            type[nodes.NodeNG],
+            _CallbackTupleT,
+        ] = {}
```
cdce8p, line 139:
```suggestion
        self._cache: dict[type[nodes.NodeNG], _CallbackTupleT] = {}
```

pylint/pyreverse/utils.py
cdce8p: With the updated typing, should this be `Union[Tuple[ClassDiagram], Tuple[PackageDiagram, ClassDiagram], None]`?

If that's the case, the visit method is probably not type compatible with the base implementation in `ASTWalker.visit` with is annotated to only return `None`. _The type of the child implementation should be a subclass of the parent impl._

pylint/pyreverse/utils.py
DudeNr33: No doubt about that. The code where this happens is over 16 years old, and I haven't touched those parts of `pyreverse` yet. I can take a look at refactoring those parts that we found to be problematic due to the typing in a follow-up PR.

pylint/pyreverse/utils.py
DudeNr33: Good catch. I can update both methods to use `Union[Tuple[ClassDiagram], Tuple[PackageDiagram, ClassDiagram], None]`.

DudeNr33: On second thought: 
This also comes down to design problems.
`ASTWalker.visit` does in fact always return `None` (there is no `return` inside the method).
`LocalsVisitor` (a subclass of `ASTWalker`) overrides `visit` and returns whatever the `leave` method for this node type returns. That's why I put `Any` here in the first place.

Currently the `ASTWalker` class is never used on its own. We could simply remove it and move the relevant methods into `LocalsVisitor`.

cdce8p: That should work. Alternatively, `ASTWalker.visit` can be annotated with the return type from `LocalsVisitor.visit`. Doesn't matter that only `None` is actually returned there. But with that the subclass implementation is compatible.

pylint/pyreverse/utils.py
```diff
@@ -121,7 +115,7 @@ def __init__(self, mode):
                 print(f"Unknown filter mode {ex}", file=sys.stderr)
         self.__mode = __mode
 
-    def show_attr(self, node):
+    def show_attr(self, node: nodes.NodeNG | str) -> bool:
         """Return true if the node should be treated."""
         visibility = get_visibility(getattr(node, "name", node))
```
DanielNoord, line 120:
```suggestion
        visibility = get_visibility(getattr(node, "name", node))
```

Unrelated, but this is quite a funny hack 😄

DanielNoord (approved):
I'll let you decide how to handle the `Any`. Rest LGTM

DudeNr33:
> I'll let you decide how to handle the `Any`. Rest LGTM

I annotated both parent and child implementation with `-> Any` - I don't really see the need to put perfect typing here if we are going to refactor this anyway.
I created #6582, and would first collect all issues we find while completing the typing, and then work on the refactoring.

PR #6551, opened 2022-05-08
Use 3.11 for Github actions

### LINKED ISSUE #5920: Add Python 3.11-dev to Github actions
### Current problem

Seems like [tomllib](https://peps.python.org/pep-0680/) has been accepted and merged https://github.com/python/cpython/pull/31498 but not yet released to any of the images available on Github.

[Next release](https://peps.python.org/pep-0664/) is expected on `2022-04-05` so we'll need to wait a bit.

### Desired solution

Add 3.11-dev when github actions are up to date and released.

### Additional context

https://github.com/PyCQA/pylint/issues/5919#issuecomment-1067728097

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #5920

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -8,7 +8,7 @@ on:
   pull_request: ~
 
 env:
-  CACHE_VERSION: 28
+  CACHE_VERSION: 29
   DEFAULT_PYTHON: "3.10"
   PRE_COMMIT_CACHE: ~/.cache/pre-commit
 
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -8,7 +8,7 @@ on:
   pull_request: ~
 
 env:
-  CACHE_VERSION: 27
+  CACHE_VERSION: 28
   DEFAULT_PYTHON: "3.10"
   PRE_COMMIT_CACHE: ~/.cache/pre-commit
 
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -9,7 +9,7 @@ on:
 
 env:
   CACHE_VERSION: 29
-  DEFAULT_PYTHON: "3.10"
+  DEFAULT_PYTHON: "3.11-dev"
   PRE_COMMIT_CACHE: ~/.cache/pre-commit
 
 concurrency:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -54,6 +54,11 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
+      - if: "contains(matrix.python-version, '-dev')"
+        # TODO: 2.16 Remove this when dill release their next version
+        run: |
+          . venv/bin/activate
+          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv/bin/activate
@@ -209,6 +214,11 @@ jobs:
           . venv\\Scripts\\activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test_min.txt
+      - if: "contains(matrix.python-version, '-dev')"
+        # TODO: 2.16 Remove this when dill release their next version
+        run: |
+          . venv\\Scripts\\activate
+          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv\\Scripts\\activate
@@ -255,6 +265,11 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
+      - if: "contains(matrix.python-version, '-dev')"
+        # TODO: 2.16 Remove this when dill release their next version
+        run: |
+          . venv/bin/activate
+          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv/bin/activate
@@ -299,6 +314,11 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test_min.txt
+      - if: "contains(matrix.python-version, '-dev')"
+        # TODO: 2.16 Remove this when dill release their next version
+        run: |
+          . venv/bin/activate
+          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv/bin/activate
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -10,7 +10,7 @@ on:
       - doc/data/messages/**
 
 env:
-  CACHE_VERSION: 28
+  CACHE_VERSION: 29
 
 jobs:
   tests-linux:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -54,11 +54,6 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
-      - if: "contains(matrix.python-version, '-dev')"
-        # TODO: 2.16 Remove this when dill release their next version
-        run: |
-          . venv/bin/activate
-          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv/bin/activate
@@ -80,7 +75,7 @@ jobs:
     needs: tests-linux
     strategy:
       matrix:
-        python-version: ["3.10"]
+        python-version: ["3.11-dev"]
     env:
       COVERAGERC_FILE: .coveragerc
     steps:
@@ -126,7 +121,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10"]
+        python-version: ["3.11-dev"]
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.2
@@ -214,11 +209,6 @@ jobs:
           . venv\\Scripts\\activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test_min.txt
-      - if: "contains(matrix.python-version, '-dev')"
-        # TODO: 2.16 Remove this when dill release their next version
-        run: |
-          . venv\\Scripts\\activate
-          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv\\Scripts\\activate
@@ -265,11 +255,6 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
-      - if: "contains(matrix.python-version, '-dev')"
-        # TODO: 2.16 Remove this when dill release their next version
-        run: |
-          . venv/bin/activate
-          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv/bin/activate
@@ -314,11 +299,6 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test_min.txt
-      - if: "contains(matrix.python-version, '-dev')"
-        # TODO: 2.16 Remove this when dill release their next version
-        run: |
-          . venv/bin/activate
-          pip install git+https://github.com/uqfoundation/dill.git@25a7e450ed76c7a0820834a3a91134476b1b8253
       - name: Run pytest
         run: |
           . venv/bin/activate
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11-dev"]
     outputs:
       python-key: ${{ steps.generate-python-key.outputs.key }}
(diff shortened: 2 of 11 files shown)


DanielNoord:
This is still broken. `dill` isn't `3.11` compatible yet. You can test this locally, but for some reason in Github Actions we don't fail but instead hang.

DanielNoord:
`dill` has released a new version but we still need an `astroid` update for these tests to pass.

Pierre-Sassoulas:
dill is still broken it seems.

DanielNoord:
They are releasing at the end of June.

hugovk:
`py311` also needs adding to `tox.ini`

Pierre-Sassoulas:
Re-trying as dill 0.3.5.1 was released.

Pierre-Sassoulas:
We do need the next version of dill.

AdamWill:
There's an astroid 2.12.1 now, which passes all its own tests on Python 3.11.

AdamWill:
With a recent snapshot of dill, and astroid 2.12.1, I'm down to 43 failures in the tests when building pylint 2.14.4:

https://kojipkgs.fedoraproject.org/work/tasks/1561/89381561/build.log

DanielNoord:
We are working on getting support for `astroid` `2.12.x` in `pylint`, progress of which you can track in #7153.

`dill` seems to have missed its release date on the 27th on June. I thought about sending them a reminder, but I think they know we're waiting on their release so I don't want to bug them too much.

AdamWill:
OK, so with new astroid and dill and #7153 , here's some really-python3.11 issues:

* telnetlib [is deprecated in 3.11](https://docs.python.org/3.11/whatsnew/3.11.html#deprecated), that breaks `access_attr_before_def_false_positive.py`
* `asyncio.coroutine` was [removed in 3.11](https://docs.python.org/3.11/whatsnew/3.11.html#removed), that breaks `iterable_context_py3.py`
* I *think* due to changes in `Enum` upstream, we're now getting an unexpected `super-init-not-called` message for `enum_self_defined_member_5138`

AdamWill:
Hmm, this is a fun one. the pylint output for the `syntax_error.py` test has changed, which breaks both that test and `import_error` (which imports it). It changed from just "invalid syntax" to "expected '('". The new error seems if anything better? But not sure why it changed, or how to allow the output to differ across Python versions (if that's desired).

DanielNoord:
> Hmm, this is a fun one. the pylint output for the `syntax_error.py` test has changed, which breaks both that test and `import_error` (which imports it). It changed from just "invalid syntax" to "expected '('". The new error seems if anything better? But not sure why it changed, or how to allow the output to differ across Python versions (if that's desired).

This is an annoying issue which I encountered before. We should probably merge https://github.com/PyCQA/pylint/pull/7097 first which deals with a similar issue and then any PR that moves `syntax-error` functional tests to unittests is much appreciated!

AdamWill:
Thanks. I'm trying to work on fixes for the things I'm sure are python 3.11 (and not just general running-the-tests-downstream stuff), but I don't know the codebase very well so I may be doing it wrong. :P I'll send a PR when I'm done.

AdamWill:
Thanks! The fork is https://github.com/AdamWill/pylint/commits/py311 , with my attempts at fixes for the telnetlib and asyncio.coroutine issues so far. I'll send a PR when I'm done looking through the other few failures that are left.

AdamWill:
@DanielNoord could we say that `test_stdin_syntaxerror` in `tests/test_self.py` already covers the same ground as the syntaxerror functional test?

DanielNoord:
> @DanielNoord could we say that `test_stdin_syntaxerror` in `tests/test_self.py` already covers the same ground as the syntaxerror functional test?
> 
> edit: hmm, but that doesn't account for `import_error`'s use of it. mmf.

I think there are some more `syntax-error` tests currently. For example there is one for `tokenize` and one for imports (you can find them by grepping for `syntax-error`).
If they don't break let's leave them, otherwise those should probably also be ported to unittests.

AdamWill:
With the changes I've submitted so far, plus a downstream patch for those two specific cases where the syntax error text changed, the test suite fully passes. So I don't think any of the others are a problem.

I think the intent of `syntax-error` is just to test that passing through a very basic `SyntaxError` from the interpreter works, so we could potentially change it to *any* simple syntax error whose text is the same on every tested version of Python, but it seems hard to find one :| I tried for like half an hour to find a variation on the theme ("an obviously erroneous function definition") which has the same text representation in Python 3.6 and 3.11, and struck out.

The `unknown_encoding_jython` one is Jython-only, and the `tokenize_error` one is meant to test a case where Python itself is OK but tokenize throws the exception. So I don't think we can quite sub in one of those two and throw out `syntax-error`, sadly.

Oooh! But. Ironically - I think the one that's used in the *unit test* is actually good! The unit test uses a file that's just this:

    for
    

i.e. `for` and a newline. That gives `SyntaxError: invalid syntax` on both Python 3.6 and 3.11 for me (the widest delta I can handily test). Maybe we could change the `syntax-error` functional test to use that.

AdamWill:
My problem with moving it to the unittests is then how do we keep the bit of `import-error` that tries to import `syntax-error.py` and expects it to throw specific errors? That seems a bit harder to move to the unit tests.

DanielNoord:
https://github.com/PyCQA/pylint/blob/eb57dedb622909ca2ccf00516b595df2ae85b75e/tests/test_self.py#L260

This is a test that uses a data file and then tests for emitted messages. I think we should be able to use something like that?

AdamWill:
Was the `Enum.__init__` thing resolved somehow? My PR wasn't merged and the upstream comment hasn't had a reply yet..

DanielNoord:
I think the recent change of not raising `super-init-not-called` on "abstract" classes fixes this.
We consider methods that only have `pass` as being abstract, so that should have fixed it.

Sorry I should have tagged you.

cclauss:
Given that Python 3.11 release candidate 2 is still experimental, would it be possible _on Python 3.11 only_ in the GitHub Actions to `pip install https://github.com/uqfoundation/dill.git@master` just to make sure our tests will pass when `dill` makes its next release?

Pierre-Sassoulas:
> pip install https://github.com/uqfoundation/dill.git@master

Yeah, you're right, that would make a lot of sense. We could not release a 3.11 compatible version but at least we'll detect the issue with 3.11 before the new release of dill !

Pierre-Sassoulas:
Well sadly it's not doable: https://github.com/pypa/pip/issues/5566#issuecomment-402417467

cclauss:
That issue is from four years ago.

Pierre-Sassoulas:
Hmm, yeah but it seems the issue still exists because:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 567274153..d5cf0b5e5 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,8 @@ classifiers = [
 ]
 requires-python = ">=3.7.2"
 dependencies    = [
-    "dill>=0.2",
+    "dill>=0.2; python_version<'3.11'",
+    "dill@git+https://github.com/uqfoundation/dill/commit/25a7e450ed76c7a0820834a3a91134476b1b8253; python_version=='3.11'",
     "platformdirs>=2.2.0",
     # Also upgrade requirements_test_min.txt and all the CACHE_VERSION in
     # github actions if you are bumping astroid.
```
(Or "dill@git+https://github.com:uqfoundation/dill/commit/25a7e450ed76c7a0820834a3a91134476b1b8253) raise an error:
```
      GIVEN VALUE:

          "dill@git+https://github.comuqfoundation/dill/commit/25a7e450ed76c7a0820834a3a91134476b1b8253; python_version=='3.11'"
      OFFENDING RULE: 'format'
      DEFINITION:
          {
              "$id": "#/definitions/dependency",
              "title": "Dependency",
              "type": "string",
              "format": "pep508"
          }
```
Maybe you know the correct syntax ?

cclauss:
I would do it in the Action, not in pyproject.toml

.github/workflows/tests.yaml
```diff
@@ -54,6 +54,10 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
+      - if: "contains(matrix.python-version, '-dev')"
+        run: |
+          . venv/bin/activate
+          pip install https://github.com/uqfoundation/dill/commit/25a7e450ed76c7a0820834a3a91134476b1b8253
```
cclauss, line 61:
```suggestion
          pip install git+https://github.com/uqfoundation/dill/commit/25a7e450ed76c7a0820834a3a91134476b1b8253.git
```

cclauss:
Well done!

cclauss:
`Tests / run / 3.11-dev / Windows (pull_request)` would need the identical fix.

cclauss:
My mistake.  On Windoze, it is `. venv\\Scripts\\activate`

cclauss:
We need a `Primer / prepare / 3.8 / Linux (pull_request)` if we want `Primer / Run / Run / 3.7 (pull_request)` to pass.

I think that line 2 of `.github/workflows/primer_run_main.yaml can be removed when comparing the URL to the current action.

Pierre-Sassoulas:
I actually did not dare to create our own version 😄 Amazing !

DanielNoord:
@Pierre-Sassoulas I think this is ready for review 😄 

I did some other clean up too. Let's use `3.11` everywhere we can, might as well help the testing efforts by checking if other jobs in our CI still pass with it.

DanielNoord:
Failures are expected. Everything else passes, which is also expected. The failures for `3.11` were already fixed in `astroid`.

cclauss:
Failures expected: --> https://github.com/PyCQA/pylint/pull/6551#issuecomment-1257140955

DanielNoord:
> I think that line 2 of `.github/workflows/primer_run_main.yaml` can be removed when comparing the URL to the current action.

Which line are you referring to here?

> We need a `Primer / prepare / 3.7 / Linux (pull_request)` if we want `Primer / Run / Run / 3.7 (pull_request)` to pass.

This is not completely true. We need a run on `main`. This is because we want `main` to create the `venv` to make sure that any commits on `astroid` `main` are "taken" by the primer run on `main` and then "distributed" to PRs rather than having PRs use a new version of `astroid` compared to the `main` run they are compared against.
This means that any change to the primer cache, workflow, etc. on PRs will always fail to execute as those changes need to be on `pylint` `main` first. This is (highly) annoying but I haven't come up with a good solution yet that 1) only allows `main` to create `venvs` and 2) exits gracefully on PRs when there is a discrepancy. I don't think there is a good solution for it, sadly.

cclauss:
> Which line are you referring to here?

Line 2.

Should the failing test `Primer / Run / Run / 3.7 (pull_request)` be dropped?

DanielNoord:
This is L2 of the file you mentioned:

https://github.com/PyCQA/pylint/blob/aca8dd546e6ec03d169eb84d7421d4a1427920ce/.github/workflows/primer_run_main.yaml#L2

I don't understand what it has to do with the rest of your comment, sorry!

> Should the failing test `Primer / Run / Run / 3.7 (pull_request)` be dropped?

Why should it? See my explanation above, it works for all other PRs. See also:
https://github.com/PyCQA/pylint/blob/aca8dd546e6ec03d169eb84d7421d4a1427920ce/.github/workflows/primer_run_pr.yaml#L14-L18

We ignore it where possible, but these ignores are not exhaustive. So if there is any file in the PR that is not being ignored then the workflow will still run.

cclauss (dismissed):
LGTM

.github/workflows/checks.yaml
```diff
@@ -8,8 +8,8 @@ on:
   pull_request: ~
 
 env:
-  CACHE_VERSION: 27
-  DEFAULT_PYTHON: "3.10"
+  CACHE_VERSION: 29
```
DanielNoord, line 11: @Pierre-Sassoulas Btw, we could also make this a repository secret. It would mean that you need to update it in the settings, but it might be easier than having to do so in PRs every time.
Same goes for `DEFAULT_PYTHON` and `PRE_COMMIT_CACHE`.

doc/whatsnew/fragments/5920.other
```diff
@@ -0,0 +1,3 @@
+Pylint now support Python 3.11.
```
DanielNoord, line 1:
```suggestion
Pylint now provides basic support for Python 3.11.
```

Should we also update the classifier?

DanielNoord (changes requested):

.github/workflows/checks.yaml
Pierre-Sassoulas: I'd rather make it accessible for everyone for now because I don't want to freeze pylint development if I'm not available. It would help if PyCQA was more democratic and I could add you to the admin group.

doc/whatsnew/fragments/5920.other
Pierre-Sassoulas: Marc made an argument not to in astroid. When someone search for the meta on pypi they expect to have really compatible package. pip will find a version compatible on 3.11, I think right now this is the goal.

.github/workflows/checks.yaml
DanielNoord: Yeah, but for some reason I feel like for `pylint` this is different. We also struggle with `match ... case` statements but we still support that version according to our classifier. For `astroid` I feel like we don't really introspect the `ast` fully but for `pylint` we do offer our current functionality completely on 3.11.
The objectives are a bit different, if you understand what I mean. But I'll leave this to you to decide!

Pierre-Sassoulas: I think you're answering the other comment about package metadata ? Let's add the classifier then.

DanielNoord: Ah yeah, my bad!

DanielNoord (approved):

Pierre-Sassoulas:
Shall we backport too ? Maybe release 2.16 immediately ?

DanielNoord:
I have some free time today so I was thinking of tackling some regressions. Notably #7003  which is blocked by https://github.com/PyCQA/pylint/pull/7496.

Either way, let's not release on Friday. That screwed us a bit with `2.15`.

cclauss:
Python 3.11.0 should ship on Wednesday so why not wait until after the tests pass on that?

Pierre-Sassoulas:
> Python 3.11.0 should ship on Wednesday so why not wait until after the tests pass on that?

Because then pylint would not be ready when 3.11 is out 😄 Release before Wednesday would be nice, especially considering Daniel created our own version of dill in order to be ready on time.

cclauss:
My mistake -- official release is planned for Monday, [2022-10-24](https://www.python.org/downloads/release/python-3110rc2).

PR #6558, opened 2022-05-09
Pass ``msg_store`` and ``node`` to ``FileState``

DanielNoord:

## Type of Changes

## Description

Base change required for https://github.com/PyCQA/pylint/pull/6556.
Let's also make `modname` required. It's a bit strange that we create a `FileState` before even encountering a file in the `__init__`, but refactoring that is not part of my plan for now.

### DIFF
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -159,6 +159,7 @@ def check_parallel(
             mapreduce_data,
         ) in pool.imap_unordered(_worker_check_single_file, files):
             linter.file_state.base_name = base_name
+            linter.file_state._is_base_filestate = False
             linter.set_current_module(module, file_path)
             for msg in messages:
                 linter.reporter.handle_message(msg)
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (changes requested):
It looks like it broke the config functional tests (report not generated anymore ?). I agree with the refactoring and deprecation in general. 

(I think ``self.file_state = FileState("BaseState", self.msgs_store)`` might be defined before we needs it so it's not Optional and so it's defined in the ``__init__`` ? I agree it's strange to do it like this.)

DanielNoord:
> It looks like it broke the config functional tests (report not generated anymore ?). I agree with the refactoring and deprecation in general.
> 
> (I think `self.file_state = FileState("BaseState", self.msgs_store)` might be defined before we needs it so it's not Optional and so it's defined in the `__init__` ? I agree it's strange to do it like this.)

I don't really know why this breaks 😓 

I think it has something to do with the mocking we do in the tests. For some reason not passing `self.msgs_store` in `self.file_state = FileState("BaseState", self.msgs_store)` fixes it. That doesn't seem like it should actually break but is more indicative of inter-test dependency.

DanielNoord:
Finally pass all tests on all platforms 🎉

Pierre-Sassoulas (approved):
👍

PR #6563, opened 2022-05-09
Release branch 2.13.9

Pierre-Sassoulas:
Blocked by #6528

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -29,6 +29,10 @@ Release date: TBA
 
  Relates to #6531
 
+* Fix a crash in ``unnecessary-dict-index-lookup`` when subscripting an attribute.
+
+  Closes #6557
+
 * Fix a crash when accessing ``__code__`` and assigning it to a variable.
 
   Closes #6539
--- a/ChangeLog
+++ b/ChangeLog
@@ -20,6 +20,11 @@ What's New in Pylint 2.13.9?
 ============================
 Release date: TBA
 
+
+* Fix false positives for ``no-name-in-module`` and ``import-error`` for ``numpy.distutils`` and ``pydantic``.
+
+  Closes #6497
+
 * Fix ``IndexError`` crash in ``uninferable_final_decorators`` method.
 
  Relates to #6531
--- a/ChangeLog
+++ b/ChangeLog
@@ -41,6 +41,10 @@ Release date: TBA
 
   Closes #6539
 
+* Fix a false positive for ``undefined-loop-variable`` when using ``enumerate()``.
+
+  Closes #6593
+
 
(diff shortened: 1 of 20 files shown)


jacobtylerwalls:
I think this needs astroid 2.11.4 also.

ChangeLog
```diff
@@ -16,10 +16,28 @@ Release date: TBA
   (Ie. not necessarily at the end)
 
 
+What's New in Pylint 2.13.10?
+=============================
+Release date: TBA
+
+
+
```
jacobtylerwalls, line 24: Shall we not plan on a 2.13.10?

DanielNoord: Good catch! Let's not!

pylint/lint/pylinter.py
```diff
@@ -1028,6 +1027,16 @@ def _discover_files(files_or_modules: Sequence[str]) -> Iterator[str]:
                     if any(root.startswith(s) for s in skip_subtrees):
                         # Skip subtree of already discovered package.
                         continue
+
+                    if _is_ignored_file(
+                        root,
+                        self.config.ignore,
```
DanielNoord, line 1033: Ref. https://github.com/PyCQA/pylint/pull/6528#issuecomment-1125734077

This seems to work right @Pierre-Sassoulas?

On `2.13.x` all config is still stored on the `BaseChecker` itself. Since `PyLinter` is a checker you can access it's options with `self.config`. If you need options from another checker you can do `get_global_option` which will look up options in the `config` attribute of all other registered checkers.

So, theoretically this should work and the passing tests seems to confirm this!

jacobtylerwalls:
Possible to sneak in two more fixes? #6602 #6604

pylint/lint/pylinter.py
Pierre-Sassoulas: I've had to do suspicious fix to make it work, in particular use ``ignore-patternS`` instead of  ``ignore-pattern`` in the tests. Did we change an option ?

Pierre-Sassoulas:
Blocked by a discussion on #6604

pylint/lint/pylinter.py
Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/6563/commits/e211718024f1353f3846586280559386218422b8

ChangeLog
```diff
@@ -18,8 +18,32 @@ Release date: TBA
 
 What's New in Pylint 2.13.9?
 ============================
-Release date: TBA
+Release date: 2022-05-09
```
Pierre-Sassoulas, line 21:
```suggestion
Release date: 2022-05-13
```
Need to be done offline

PR #6569, opened 2022-05-10
Use main instead of master as main checker name

### LINKED ISSUE #5467: Updating [MASTER] terminology to [MAIN]
From the discussion on #3961

To change the references to `MASTER` in the code to support `MASTER` and `MAIN` requires:

- 4 references to `MASTER` in `run.py`
- 2 references to `MASTER` in `lint_module_test.py`
- a constant `MAIN_CHECKER_NAME` found in `constants.py`
- tests using a lower case `master`
- documentation using all forms of `master`

If we are able to support both `MAIN` and `MASTER`, then we can update all docs to `MAIN` and still be backwards compatible. Changing those six `MASTER` references should be quick, but having a flexible `MAIN_CHECKER_NAME` does not look as easy. Only 11 references to the variable itself, but the value is inherited into a web of use cases. From `constants.py`:

```
# You probably don't want to change the MAIN_CHECKER_NAME
# This would affect rcfile generation and retro-compatibility
# on all project using [MASTER] in their rcfile.
MAIN_CHECKER_NAME = "master"
```

_Originally posted by @Talador12 in https://github.com/PyCQA/pylint/issues/3961#issuecomment-985817814_

DanielNoord:

## Type of Changes

## Description

Closes #5467.

Sorry for the size of the PR, but I just went through all cases of the term `master`.

I think it would be good to do this in the `beta` as well as this could unexpectedly break stuff and is then very easy to revert.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -40,7 +40,7 @@ Release date: TBA
 
   Closes #6101
 
-* The main checker name is now ``main`` instead of ``master``.
+* The main checker name is now ``main`` instead of ``master``. The configuration does not need to be updated as sections' name are optional.
 
   Closes #5467
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -40,6 +40,10 @@ Release date: TBA
 
   Closes #6101
 
+* The main checker name is now ``main`` instead of ``master``.
+
+  Closes #5467
+
 * Added new checker ``typevar-name-mismatch``: TypeVar must be assigned to a variable with the same name as its name argument.
 
   Closes #5224
--- a/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
@@ -35,7 +35,7 @@ body:
       placeholder: |
         # Leave this blank if the configuration is not relevant!
 
-        [MASTER]
+        [MAIN]
         load-plugins=
             pylint.extensions.code_style
 
--- a/doc/data/messages/b/broken-collections-callable/pylintrc
+++ b/doc/data/messages/b/broken-collections-callable/pylintrc
@@ -1,3 +1,3 @@
-[master]
+[main]
 py-version=3.9
 load-plugins=pylint.extensions.typing
--- a/doc/data/messages/c/confusing-consecutive-elif/pylintrc
+++ b/doc/data/messages/c/confusing-consecutive-elif/pylintrc
@@ -1,2 +1,2 @@
-[MASTER]
+[MAIN]
 load-plugins=pylint.extensions.confusing_elif
--- a/doc/data/messages/m/missing-return-doc/pylintrc
+++ b/doc/data/messages/m/missing-return-doc/pylintrc
@@ -1,4 +1,4 @@
-[master]
+[main]
 load-plugins=pylint.extensions.docparams
 
 [Parameter_documentation]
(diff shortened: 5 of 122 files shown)


Pierre-Sassoulas (commented):
I'd feel safer if we handled both i.e. we keep the old MASTER in the tests, and add one test using MAIN.

ChangeLog
```diff
@@ -40,6 +40,10 @@ Release date: TBA
 
   Closes #6101
 
+* The main checker name is now ``main`` instead of ``master``.
```
Pierre-Sassoulas, line 43:
```suggestion
* The main checker name is now ``main`` instead of ``master``. The configuration does not need to be updated as sections' name are optional.
```

DanielNoord:
> I'd feel safer if we handled both i.e. we keep the old MASTER in the tests, and add one test using MAIN.

We have tests for `MASTER` already. See:
https://github.com/PyCQA/pylint/tree/main/tests/config/functional/setup_cfg/deprecate_master

That's why I am not too worried about changing this across our own code. We support both since our migration to `argparse` so there is no reason to keep a (for some people) loaded term in. Let's change while we're taking the time to look into this anyway.

Pierre-Sassoulas:
> Until now only bad-option-value gave real headaches

I'm a little worried about bad-option-value being raised if a check became an extension (no-self-use). What do you think about handling this use case ? We're going to get this a lot because I can see how a lot of pylintrc would disable it right now.

DanielNoord:
> > Until now only bad-option-value gave real headaches
> 
> I'm a little worried about bad-option-value being raised if a check became an extension (no-self-use). What do you think about handling this use case ? We're going to get this a lot because I can see how a lot of pylintrc would disable it right now.

I think that's a feature not a bug. It immediately tells people to either add the extension to their config if they want to keep using it or remove it from their disabled list of they no longer want to.

`bad-option-value` has some difficult trade-offs, but I think in the long run its effect on keeping `pylint` configurations up to date will have a positive effect on the `pylint` ecosystem.

Pierre-Sassoulas:
Say you already explicitly disabled no-self-use, what's the value brought by having to remove it because it became optional in pylint ? It just feel like double the work (and just imagine if we want to change our mind about including a check, they'd need to re-disable it). I would be very conservative in forcing users to clean up their config file manually. How about a warning for explicit enable telling the user what plugin need to be loaded now and nothing for explicit disable ?

DanielNoord:
In that first case there is no specific value other than a shorter config file.

The only change users who don't want to receive these messages ever again is add `bad-option-value` once to the top of their `disable` list in their config file if they upgrade to `2.14+`. That should prevent any warnings from popping up about this ever again. (That could be considered a breaking change, but...)

Adding anything else to this imo just unnecessarily complicates things. It's a 1 or 2 minute change after `pre-commit` makes a PR against your branch to update to the new version. I think such a change isn't too much to ask for when upgrading a dependency. The benefits and downsides of the added complexity just don't fully weigh up for me: the group of users that will be pissed off by this vs. the group that welcomes us more pro-actively warning about this + the increased maintainability of simpler code.

Pierre-Sassoulas:
Wouldn't a basic version of the auto-migration in #5462 (doable by just loading a conf and re-generating it immediately) fix this issue by making it automated though ?

DanielNoord:
You would still need a manually maintained mapping of moved messages and extensions as well finding a way to determine whether a `bad-option-value` is due to a typo, a moved message, a removed message, a missing `,`, etc.

It might not be as simple as removing any `bad-option-value` we encounter.

Issue #6570, opened 2022-05-10
Ubuntu main: 4 tests fail


edreamleo:
### Bug description

Four unit tests fail on main branch with Ubuntu.

### Configuration

_No response_

### Command used

```shell
pytest
```


### Pylint output

```shell
edreamleo@edreamleo-MS-7D09:~$ cd ekr-pylint/
edreamleo@edreamleo-MS-7D09:~/ekr-pylint$ pytest
============================= test session starts ==============================
platform linux -- Python 3.10.4, pytest-7.1.2, pluggy-1.0.0
benchmark: 3.4.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/edreamleo/ekr-pylint, configfile: setup.cfg, testpaths: tests
plugins: benchmark-3.4.1
collected 1972 items                                                           

tests/test_check_parallel.py .s.....sssssssssssssssssssssssssss          [  1%]
tests/test_deprecation.py .......                                        [  2%]
tests/test_epylint.py ..                                                 [  2%]
tests/test_func.py ............                                          [  2%]
tests/test_functional.py ............................................... [  5%]
........................................................................ [  8%]
.....s..........s.....................................................s. [ 12%]
........................................................................ [ 16%]
....................s.s.s.......................s..............s........ [ 19%]
...................................................................s...s [ 23%]
..F..........s...........ss...s.s....................................... [ 27%]
........F..F.F..................................................s...s... [ 30%]
.......................................................s................ [ 34%]
.....................................s.....s............................ [ 38%]
.................................................................                                                                                                                                       [ 41%]
tests/test_functional_directories.py .                                                                                                                                                                  [ 41%]
tests/test_import_graph.py ......                                                                                                                                                                       [ 41%]
tests/test_numversion.py ...........                                                                                                                                                                    [ 42%]
tests/test_pragma_parser.py ............                                                                                                                                                                [ 42%]
tests/test_pylint_runners.py ........                                                                                                                                                                   [ 43%]
tests/test_regr.py .................x.                                                                                                                                                                  [ 44%]
tests/test_self.py ..........x.......................................... [ 46%]
.............s.....................................................                                                                                                                                     [ 50%]
tests/test_similar.py ............                                                                                                                                                                      [ 50%]
tests/unittest_reporters_json.py ..                                                                                                                                                                     [ 51%]
[...]

PR #6572, opened 2022-05-10
Use the standard config parser instead a new one in functional tests

DanielNoord:

## Type of Changes

## Description

Fixes the spurious warnings identified in https://github.com/PyCQA/pylint/issues/6570.

Ideally we should deprecate the use of `FunctionalTestFile.options` as we should just parse options like we normally parse the options file. But it is really not a priority and I want to focus attention on any regressions in `2.14` rather than add new ones.


Pierre-Sassoulas (approved):
LGTM !

Issue #1984, opened 2018-03-30
False positive for simplifiable-if-statement when the boolean values are not the same


PCManticore:
### Steps to reproduce
1. Save the following in a file and run pylint over it
```
A = True
B = 1
C = [1]
D = False
E = False

if A and B in C:
    D = True
else:
    E = True

print(D, E)
```

### Current behavior
pylint emits a `simplifiable-if-statement`.

### Expected behavior
pylint shouldn't emit `simplifiable-if-statement` since we assign to two different values in the `if` statement.

### pylint --version output
master & 1.8.3

Issue #1684, opened 2017-09-29
pylint 1.7.3 still has 1.7.2 version number


kyrios123:
### Steps to reproduce
1. pylint --version
2. check Changelog

### Current behavior
1.
```
No config file found, using default configuration
pylint 1.7.2, 
astroid 1.5.3
Python 2.7.13 (default, Jan 12 2017, 13:55:14) 
[GCC 6.3.0]
```
2.
```
------------------
Pylint's ChangeLog
------------------

What's New in Pylint 1.7.2?
===========================

Release date: 2017-06-03
```
### Expected behavior
1. version 1.7.3
2. changelog for version 1.7.3

### pylint --version output
see above

PCManticore:
make sure you are using the recently installed binary and not something else. This is what I'm getting on my machine.

```
(env) claudiu@root$ pylint --version
pylint 1.7.3, 
astroid 1.5.3
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609]
(env) claudiu@root$ 
```

kyrios123:
You're right, shame on me. I still had an old terminal open.

PR #6574, opened 2022-05-11
Add a documentation for simplifiable if expression

### LINKED ISSUE #5882: The ``simplifiable-if-expression`` message description should not only mention ``bool(test)``
### Bug description

Here is a simple example

```test.py
"""Simplifiable if expression false positive"""

TEST_VAR1 = True
TEST_VAR2 = 'True'
A = True if TEST_VAR1 is True else False
B = True if TEST_VAR2 == 'True' else False
```

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test2
test2.py:5:4: R1719: The if expression can be replaced with 'test' (simplifiable-if-expression)
test2.py:6:4: R1719: The if expression can be replaced with 'test' (simplifiable-if-expression)
```

### Expected behavior

I would only expect the first `simplifiable-if-expression` warning for `A`.

For `B` simplifying it would not work since it's comparing against the string `True` and eventually the string `False` and `bool(string)` is always `True`

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.9.10 (main, Feb  3 2022, 16:04:33) 
[GCC 11.1.0]
```

### OS / Environment

Linux 5.16.12-arch1-1 #1 SMP PREEMPT Wed, 02 Mar 2022 12:22:51 +0000 x86_64 GNU/Linux

### Additional dependencies

_No response_

Pierre-Sassoulas:
## Type of Changes

## Description

I realized during implementation that simplifiable if expression became worst and that I disagree with https://github.com/PyCQA/pylint/issues/1984 basically. If there are two values but one if the opposite of the other we can pretty much simplify the if expression. It look like the example from the issue come directly from [this stackoverflow question](https://stackoverflow.com/questions/49546992/) but the top answer make a compelling point that this is simplifiable.

Refs #5953 
Closes #5882

### DIFF
--- a/doc/data/messages/s/simplifiable-if-expression/bad.py
+++ b/doc/data/messages/s/simplifiable-if-expression/bad.py
@@ -1,0 +1,9 @@
+FLYING_THINGS = ["bird", "plane", "superman", "this example"]
+
+
+def is_flying_thing(an_object):
+    return True if an_object in FLYING_THINGS else False  # [simplifiable-if-expression]
+
+
+def is_not_flying_thing(an_object):
+    return False if an_object in FLYING_THINGS else True  # [simplifiable-if-expression]
--- a/doc/data/messages/s/simplifiable-if-expression/good.py
+++ b/doc/data/messages/s/simplifiable-if-expression/good.py
@@ -1,1 +1,9 @@
-# This is a placeholder for correct code for this message.
+FLYING_THINGS = ["bird", "plane", "superman", "this example"]
+
+
+def is_flying_thing(an_object):
+    return an_object in FLYING_THINGS
+
+
+def is_not_flying_thing(an_object):
+    return an_object not in FLYING_THINGS
--- a/doc/data/messages/s/simplifiable-if-expression/details.rst
+++ b/doc/data/messages/s/simplifiable-if-expression/details.rst
@@ -1,1 +1,0 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
--- a/doc/data/messages/s/simplifiable-if-expression/related.rst
+++ b/doc/data/messages/s/simplifiable-if-expression/related.rst
@@ -1,0 +1,1 @@
+- `Simplifying an 'if' statement with bool() <https://stackoverflow.com/questions/49546992/>`_
--- a/doc/data/messages/s/simplifiable-if-statement/bad.py
+++ b/doc/data/messages/s/simplifiable-if-statement/bad.py
@@ -1,0 +1,9 @@
+FLYING_THINGS = ["bird", "plane", "superman", "this example"]
+
+
+def is_flying_animal(an_object):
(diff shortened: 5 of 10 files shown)


doc/data/messages/s/simplifiable-if-expression/good.py
```diff
@@ -0,0 +1,5 @@
+FLYING_THINGS = ["bird", "plane", "superman", "this example"]
+
+def is_flying_animal(an_object):
+    is_flying = isinstance(an_object, Animal) and an_object.name in FLYING_THINGS
+    return is_flying, not is_flying
```
DanielNoord, line 5:
```suggestion
    is_flying = isinstance(an_object, Animal) and an_object.name in FLYING_THINGS
    is_not_flying = not is_flying
    return is_flying, is_not_flying
```

Perhaps? To show how to solve this when assigning?

DanielNoord (approved):

doc/data/messages/s/simplifiable-if-expression/bad.py
```diff
@@ -0,0 +1,8 @@
+FLYING_THINGS = ["bird", "plane", "superman", "this example"]
+
+def is_flying_animal(an_object):
+    if isinstance(an_object, Animal) and an_object in FLYING_THINGS:
```
DanielNoord, line 4: We need to expect a message here.

Pierre-Sassoulas: Yes, it's not raising anything right now, so I did not expect it to work I think it's a problem and that's why this issue is labelled discussion:

> I realized during implementation that simplifiable if expression became worst and that I disagree with https://github.com/PyCQA/pylint/issues/1984 basically. If there are two values but one if the opposite of the other we can pretty much simplify the if expression. It look like the example from the issue come directly from [this stackoverflow question](https://stackoverflow.com/questions/49546992/) but the top answer make a compelling point that this is simplifiable.

DanielNoord: Ah okay.

I'm not sure whether I think we should raise here. I'll let others give their opinion as well.

jacobtylerwalls: I think #1684 was correct, because even if the two values are given opposing boolean values, we can't simplify, because they might have had state before.

E.g. this raises `TypeError` because  `b` had state:
```python
def x():
  a = []
  b = []
  if bool('special test'):
    a = True
  else:
    b = False
  return a + b
```

But if we refactor to `return bool('special test')`, then the behavior is different.

Pierre-Sassoulas: You're right, thank you for your input 👍 I think this is still simplifiable because clearly the code style is horrible (with type redefinition of the type and the conditional can still be simplified : clearly we don't have this type of construct in the primer 😄) but it's another MR's scope.

PR #6576, opened 2022-05-11
[raise-missing-from] Clearer message and example in the documentation

### LINKED ISSUE #3707: Message for raise-missing-from
I think that the message we added to the `raise-missing-from` rule isn't user friendly enough: https://github.com/PyCQA/pylint/blob/master/pylint/checkers/exceptions.py#L157

The goal is that a user who has never seen the syntax before will be able to use it. We might also want to link to PEP 3134, hopefully to the relevant segment.

Is anyone interested in writing the new message? Otherwise I'll give it a try.

Pierre-Sassoulas:
## Type of Changes

## Description

Co-authored-by: @cool-RR <ram@rachum.com>

Refs #5953 
Closes #3707

### DIFF
--- a/doc/data/messages/r/raise-missing-from/related.rst
+++ b/doc/data/messages/r/raise-missing-from/related.rst
@@ -1,0 +1,1 @@
+- `PEP 3132 <https://peps.python.org/pep-3132/>`_
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -15,6 +15,7 @@ from astroid import nodes, objects
 
 from pylint import checkers
 from pylint.checkers import utils
+from pylint.interfaces import HIGH
 from pylint.typing import MessageDefinitionTuple
 
 if TYPE_CHECKING:
@@ -131,13 +132,13 @@ MSGS: dict[
         "try-except-raise block!",
     ),
     "W0707": (
-        "Consider explicitly re-raising using the 'from' keyword",
+        "Consider explicitly re-raising using %s'%s from %s'",
         "raise-missing-from",
-        "Python 3's exception chaining means it shows the traceback of the "
-        "current exception, but also the original exception. Not using `raise "
-        "from` makes the traceback inaccurate, because the message implies "
-        "there is a bug in the exception-handling code itself, which is a "
-        "separate situation than wrapping an exception.",
+        "Python's exception chaining shows the traceback of the current exception, "
+        "but also of the original exception. When you raise a new exception after "
+        "another exception was caught it's likely that the second exception is a "
+        "friendly re-wrapping of the first exception. In such cases `raise from` "
+        "provides a better link between the two tracebacks in the final error.",
     ),
     "W0711": (
         'Exception to catch is the result of a binary "%s" operation',
@@ -334,16 +335,36 @@ class ExceptionsChecker(checkers.BaseChecker):
         if containing_except_node.name is None:
             # The `except` doesn't have an `as exception:` part, meaning there's no way that
             # the `raise` is raising the same exception.
-            self.add_message("raise-missing-from", node=node)
-        elif isinstance(node.exc, nodes.Call) and isinstance(node.exc.func, nodes.Name):
-            # We have a `raise SomeException(whatever)`.
-            self.add_message("raise-missing-from", node=node)
+            class_of_old_error = "Exception"
+            if isinstance(containing_except_node.type, nodes.Name):
+                class_of_old_error = containing_except_node.type.name
+            elif isinstance(containing_except_node.type, nodes.Tuple):
+                # I.e. 'except (ZeroDivisionError, ValueError, ArithmeticError)'
+                exceptions = ", ".join(n.name for n in containing_except_node.type.elts)
+                class_of_old_error = f"({exceptions})"
+            self.add_message(
+                "raise-missing-from",
+                node=node,
+                args=(
+                    f"'except {class_of_old_error} as exc' and ",
+                    node.as_string(),
+                    "exc",
+                ),
+                confidence=HIGH,
+            )
         elif (
-            isinstance(node.exc, nodes.Name)
+            isinstance(node.exc, nodes.Call)
+            and isinstance(node.exc.func, nodes.Name)
+            or isinstance(node.exc, nodes.Name)
             and node.exc.name != containing_except_node.name.name
         ):
-            # We have a `raise SomeException`.
-            self.add_message("raise-missing-from", node=node)
+            # We have a `raise SomeException(whatever)` or a `raise SomeException`
+            self.add_message(
+                "raise-missing-from",
+                node=node,
+                args=("", node.as_string(), containing_except_node.name.name),
+                confidence=HIGH,
+            )
 
     def _check_catching_non_exception(self, handler, exc, part):
         if isinstance(exc, nodes.Tuple):
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -336,12 +336,9 @@ class ExceptionsChecker(checkers.BaseChecker):
             # The `except` doesn't have an `as exception:` part, meaning there's no way that
             # the `raise` is raising the same exception.
             class_of_old_error = "Exception"
-            if isinstance(containing_except_node.type, nodes.Name):
-                class_of_old_error = containing_except_node.type.name
-            elif isinstance(containing_except_node.type, nodes.Tuple):
-                # I.e. 'except (ZeroDivisionError, ValueError, ArithmeticError)'
-                exceptions = ", ".join(n.name for n in containing_except_node.type.elts)
-                class_of_old_error = f"({exceptions})"
+            if isinstance(containing_except_node.type, (nodes.Name, nodes.Tuple)):
+                # 'except ZeroDivisionError' or 'except (ZeroDivisionError, ValueError)'
+                class_of_old_error = containing_except_node.type.as_string()
             self.add_message(
                 "raise-missing-from",
                 node=node,
--- a/doc/data/messages/r/raise-missing-from/bad.py
+++ b/doc/data/messages/r/raise-missing-from/bad.py
@@ -1,0 +1,4 @@
+try:
+    1 / 0
+except ZeroDivisionError as e:
+    raise ValueError("Rectangle Area cannot be zero")  # [raise-missing-from]
--- a/doc/data/messages/r/raise-missing-from/good.py
+++ b/doc/data/messages/r/raise-missing-from/good.py
@@ -1,0 +1,4 @@
+try:
+    1 / 0
+except ZeroDivisionError as e:
+    raise ValueError("Rectangle Area cannot be zero") from e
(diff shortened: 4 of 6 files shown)


doc/data/messages/r/raise-missing-from/details.rst
```diff
@@ -0,0 +1,4 @@
+You are raising the %s exception after the %s exception was caught.
+It's likely that the second exception is a friendly re-wrapping of the first exception,
+therefore you should use ``raise new_exception from old_exception`` to see an accurate message
+between the two tracebacks.
```
Pierre-Sassoulas, line 4:
```suggestion
You are raising a new exception after an old exception was caught.

It's likely that the second exception is a friendly re-wrapping of the first exception,
therefore you should use ``raise NewException(...) from old_exception`` to see an accurate message
between the two tracebacks.
```

pylint/checkers/exceptions.py
```diff
@@ -131,7 +131,7 @@ def _is_raising(body: list) -> bool:
         "try-except-raise block!",
     ),
     "W0707": (
-        "Consider explicitly re-raising using the 'from' keyword",
+        "Consider explicitly re-raising using 'raise NewError(...) from old_error'",
```
DanielNoord, line 134: It would probably be a lot more user-friendly if we passed `old_error` as an argument and showed the actual error that was caught. I think we have that information when we `add_message`, right?

doc/data/messages/r/raise-missing-from/related.rst
```diff
@@ -0,0 +1 @@
+- `PEP 3132 <https://www.python.org/dev/peps/pep-3132/>`_
```
DanielNoord, line 1:
```suggestion
- `PEP 3132 <https://peps.python.org/pep-3132/>`_
```

pylint/checkers/exceptions.py
```diff
@@ -131,7 +131,7 @@ def _is_raising(body: list) -> bool:
...
         "raise-missing-from",
         "Python 3's exception chaining means it shows the traceback of the "
```
DanielNoord, line 136:
```suggestion
        "Python's exception chaining shows the traceback of the "
        "current exception, but also of the original exception. When you raise a "
        "new exception after another exception was caught it's likely that the "
        "second exception is a friendly re-wrapping of the first exception. In such "
        "cases `raise from` provides a better link between the two tracebacks in the final error."
```

DanielNoord (changes requested):
Agree with the original issue that the message description can be a lot better. Then we can probably also remove the `details.rst`.

pylint/checkers/exceptions.py
DanielNoord:
```diff
diff --git a/pylint/checkers/exceptions.py b/pylint/checkers/exceptions.py
index 7e182c3dc..26024f8aa 100644
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -336,6 +336,12 @@ class ExceptionsChecker(checkers.BaseChecker):
         if containing_except_node.name is None:
             # The `except` doesn't have an `as exception:` part, meaning there's no way that
             # the `raise` is raising the same exception.
+            if isinstance(containing_except_node.type, nodes.Name):
+                name_of_old_error = containing_except_node.type.name
+            elif isinstance(containing_except_node.type, nodes.Tuple):
+                name_of_old_error = ", ".join(
+                    n.name for n in containing_except_node.type.elts
+                )
             self.add_message("raise-missing-from", node=node)
         elif isinstance(node.exc, nodes.Call) and isinstance(node.exc.func, nodes.Name):
             # We have a `raise SomeException(whatever)`.
```

? We have access to the wrapping except handler and can determine the node name from it.

There are three other places at which the message is raised but I think we should be able to get the names there as well.

Pierre-Sassoulas: Say we have this:
```python
try:
    1 / 0
except ZeroDivisionError:
    # +1: [raise-missing-from]
    raise KeyError
```
``name_of_old_error`` would be ``ZeroDivisionError`` when we would want it to be ``e`` or something so the end result is this:
```python
try:
    1 / 0
    raise KeyError from e
```
So the message should be something like `` "Consider explicitly re-raising using 'raise KeyError from old_error'",`` and if we have ``except ZeroDivisionError as exc:`` in the original code it should be  ``Consider explicitly re-raising using 'raise KeyError from exc"``, right ?

DanielNoord: Ah. I think we should potentially rephrase the message. What about:
``Consider explicitly re-raising using 'except %s as exc' and 'raise from exc'``

That should work for both situations right?

Pierre-Sassoulas: 👍 I don't know if we can do ``Consider explicitly re-raising using 'except %s as exc' and 'raise%s from exc'``  with the second args being ``KeyError`` or ``ValueError("my message"), but ``Consider explicitly re-raising using 'except %s as exc' and 'raise ... from exc'`` is probably good enough ?

DanielNoord: Yeah I think the `...` works. Messages shouldn't be too long I think so that seems like a nice middle ground.

Is `exc` the most common name to do this? Docs seems to use `err` and `exc`: https://docs.python.org/3/tutorial/errors.html

Pierre-Sassoulas: A small refactor permitted to have a nice little message. At this point we might as well autofix files 😄 Something to consider when we have less than 10 issues opened :trollface:

doc/data/messages/r/raise-missing-from/related.rst
Pierre-Sassoulas:
```suggestion
- `PEP 3132 <https://peps.python.org/pep-3132/>`_
```

pylint/checkers/exceptions.py
```diff
@@ -333,19 +334,39 @@ def _check_raise_missing_from(self, node: nodes.Raise) -> None:
         # like `exc.with_traceback(whatever)`. We won't analyze these, we'll just assume
         # there's a violation on two simple cases: `raise SomeException(whatever)` and `raise
         # SomeException`.
+
+        need_to_raise = False
```
DanielNoord, line 338: I would be okay with two `add_message` instead of this variable. Would also save some assignments as we can just pass `""` or the actual `except_as_exc`.

doc/data/messages/r/raise-missing-from/details.rst
```diff
@@ -0,0 +1,4 @@
+You are raising the %s exception after the %s exception was caught.
```
DanielNoord, line 1: Let's remove this file now that this is basically in the message description.

DanielNoord (changes requested):

pylint/checkers/exceptions.py
```diff
@@ -333,19 +334,39 @@ def _check_raise_missing_from(self, node: nodes.Raise) -> None:
         # like `exc.with_traceback(whatever)`. We won't analyze these, we'll just assume
         # there's a violation on two simple cases: `raise SomeException(whatever)` and `raise
         # SomeException`.
+
```
DanielNoord, line 337:
```suggestion
```

pylint/checkers/exceptions.py
```diff
@@ -333,19 +334,39 @@ def _check_raise_missing_from(self, node: nodes.Raise) -> None:
...
         if containing_except_node.name is None:
             # The `except` doesn't have an `as exception:` part, meaning there's no way that
             # the `raise` is raising the same exception.
-            self.add_message("raise-missing-from", node=node)
-        elif isinstance(node.exc, nodes.Call) and isinstance(node.exc.func, nodes.Name):
-            # We have a `raise SomeException(whatever)`.
-            self.add_message("raise-missing-from", node=node)
+            class_of_old_error = "Exception"
```
DanielNoord, line 341:
```suggestion
            class_of_old_error: str
```

If this crashes we have found a new pattern for which we should create an `elif`.

pylint/checkers/exceptions.py
```diff
@@ -333,19 +334,39 @@ def _check_raise_missing_from(self, node: nodes.Raise) -> None:
...
+                class_of_old_error = containing_except_node.type.name
+            elif isinstance(containing_except_node.type, nodes.Tuple):
+                class_of_old_error = ", ".join(
+                    n.name for n in containing_except_node.type.elts
+                )
+            except_as_exc = f"'except {class_of_old_error} as exc' and "
+            self.add_message(
+                "raise-missing-from",
+                node=node,
+                args=(except_as_exc, node.as_string(), "exc"),
```
DanielNoord, line 352:
```suggestion
                args=(f"'except {class_of_old_error} as exc' and ", node.as_string(), "exc"),
```

pylint/checkers/exceptions.py
```diff
@@ -333,19 +334,39 @@ def _check_raise_missing_from(self, node: nodes.Raise) -> None:
...
-            # We have a `raise SomeException(whatever)`.
-            self.add_message("raise-missing-from", node=node)
+            class_of_old_error = "Exception"
+            if isinstance(containing_except_node.type, nodes.Name):
+                class_of_old_error = containing_except_node.type.name
+            elif isinstance(containing_except_node.type, nodes.Tuple):
+                class_of_old_error = ", ".join(
+                    n.name for n in containing_except_node.type.elts
+                )
+            except_as_exc = f"'except {class_of_old_error} as exc' and "
```
DanielNoord, line 348:
```suggestion
```

pylint/checkers/exceptions.py
```diff
@@ -333,19 +334,39 @@ def _check_raise_missing_from(self, node: nodes.Raise) -> None:
...
+            )
         elif (
-            isinstance(node.exc, nodes.Name)
-            and node.exc.name != containing_except_node.name.name
+            isinstance(node.exc, nodes.Call)
+            and isinstance(node.exc.func, nodes.Name)
+            or (
+                isinstance(node.exc, nodes.Name)
+                and node.exc.name != containing_except_node.name.name
+            )
```
DanielNoord, line 361:
```suggestion
            or isinstance(node.exc, nodes.Name)
            and node.exc.name != containing_except_node.name.name
```

`And` should bind before `or` right?

DanielNoord (changes requested):
Final nitpicks

pylint/checkers/exceptions.py
Pierre-Sassoulas: It crashes for 
```python
try:
    int("9a")
except:
    raise ValueError('Invalid integer')
```
``Exception`` felt like a nice default value that handled broad-except and raise-missing-from with one message.

DanielNoord: Ah yeah. We could suggest `except ... as exc` as well but I guess `except Exception as exc` is fine.

DanielNoord (approved):

DanielNoord (approved):

Pierre-Sassoulas:
I fixed-up what was previously reviewed, the new code is in https://github.com/PyCQA/pylint/pull/6576/commits/642b18d25aa2ea68d579542048dc40dfb8e9c710

DanielNoord (approved):

PR #6577, opened 2022-05-11
Fix recognition of files named ``setup.cfg`` (#3630)

### LINKED ISSUE #3630: Some options are not parsed correctly from setup.cfg
With this [recent PR](https://github.com/carsongee/pytest-pylint/pull/107), pylint is supposed to support loading options through `setup.cfg`.

I am finding that while it loads some options fine, it fails to load many options (a little unclear on why, but it seems to come down to the section where they are defined). E.g, this **works**:

```
[tool.pylint.MASTER]
init-hook="from pylint.config import find_pylintrc; import os, sys; sys.path.append(os.path.dirname(find_pylintrc())); "
```

This does **not**:
```
[tool.pylint.MASTER]
load-plugins=pylint_flask,pylint_flask_sqlalchemy
```

Oddly, this **does**:
```
[MASTER]
load-plugins=pylint_flask,pylint_flask_sqlalchemy
```

But **neither** of these two works:
```
[tool.pylint.FORMAT]
max-line-length=220
ignore-long-lines=[a-z]*

[FORMAT]
max-line-length=220
ignore-long-lines=[a-z]*
```

and yet, this **works**:
```
[tool.pylint.FORMAT]
good-names-rgxs=[a-z]
```

All these options work fine if I add them to a `pylintrc` file, so it does not seem to be an issue of format.

I have looked through the code a bit (which is how I was able to figure out that `load-plugins` worked with the short section name), but I cannot figure out what it is doing with these missing `FORMAT` options…

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #3630. It seems there's still some oddity, init_hook is not treated the same as other option apparentely.


tests/config/functional/setup_cfg/issue_3630/1.result.json
```diff
@@ -0,0 +1 @@
+{}
```
Pierre-Sassoulas, line 1: Here's it's strange that we have no value set for init-hook

tests/config/functional/setup_cfg/issue_3630/1.cfg
```diff
@@ -0,0 +1,3 @@
+# Wrong header ?
+[tool.pylint.MASTER]
+init-hook="from pylint.config import find_pylintrc;"
```
DanielNoord, line 3: Shall we use something like `print()` so we can actually check the `init-hook` is executed?

tests/config/functional/setup_cfg/issue_3630/1.result.json
DanielNoord: Yeah, this should have a value I think. But I'm not 100% sure since this is one of the "pre-processed" options so it might be a bit broken.

tests/config/functional/setup_cfg/issue_3630/2.2.out
```diff
@@ -0,0 +1,2 @@
+************* Module Command line or configuration file
```
DanielNoord, line 1: What is this testing? This seems a little redundant with all the other tests that check for (incorrect) section headers.

DanielNoord (changes requested):
Other than the first test (which I agree seems like it is still broken) I don't think the other tests add much value? Those regressions are covered by other tests I think.

If anything, we should probably add comments or change the file names to reflect the issues.

Also note, with the migration to `argparse` there is now slightly different behaviour between `setup.cfg` and `myconfig.cfg`. As for `setup` we require `pylint.` section names (although we still support `[MASTER]`. It might be necessary to use a `setup.cfg` name instead.

tests/config/functional/setup_cfg/issue_3630/2.2.out
Pierre-Sassoulas: Just took all the example from the issue :)

DanielNoord:
@Pierre-Sassoulas This is ready for review!

Pierre-Sassoulas:
Can't approve the MR I opened myself but let's merge this 👍

Issue #971, opened 2016-07-01
Generate a Pylint configuration interactively


PCManticore:
Pylint's configurations tend to be extremely verbose, since we have plenty of options for users to choose from. At this point, I think it would be a good idea to have a command, something as `pylint --quickstart`, which can help generating a new configuration, interactively, by suggesting potential areas where an user can look into. I will come up with some more details later.

rogalski:
Similar idea was implemented in tox, it may be a good place to start, see [tox-quickstart source](https://github.com/tox-dev/tox/blob/master/tox/_quickstart.py).

CarliJoy:
Also pyscaffold implemented a simple `-interactive` mode:

https://github.com/pyscaffold/pyscaffold#quickstart

DanielNoord:
With the migration to `argparse` almost finished I think we can start thinking about this issue and #5462 as the final main objectives of the project.

I'm wondering if perhaps we should merge the two and provide a command:
``pylint --interactive-config-setup``

The first question could then be:
Do you want to start from scratch or take an existing config as the base?

Another question to tackle is the output of this command. I think we want it to be `.toml` so that we can deprecate ``--generate-rcfile`` after we have implemented this.

Pierre-Sassoulas:
> Do you want to start from scratch or take an existing config as the base?

We already have a command to generate the config file containing the default value from zero (both ini and toml). If we have #5403 and the user want to use this functionality, we can generate a configuration to use baseline with default value (i.e. all default messages, for new issues) and it's not much change from what we already have. The next step imo, is then https://github.com/PyCQA/pylint/issues/5604 to guide the change in an existing configuration where some message are disabled to guide the user on what makes the most sense to enable next.

Pierre-Sassoulas:
There's an official looking google pylint configuration out there if we want to propose template: https://google.github.io/styleguide/pyguide.html#21-lint

PR #6578, opened 2022-05-11
Add experimental ``pylint-config`` command

DanielNoord:

## Type of Changes

## Description

Ref. https://github.com/PyCQA/pylint/issues/971

Let's put this in `2.15`. But would this be the way to do this?

### DIFF
--- a/.pyenchant_pylint_custom_dict.txt
+++ b/.pyenchant_pylint_custom_dict.txt
@@ -186,6 +186,7 @@ misdesign
 misdesigns
 mixin
 modname
+monkeypatch
 mro
 # Used so much that we need the abbreviation
 msg
@@ -217,6 +218,7 @@ numpy
 ok
 optdict
 optik
+optionals
 optiondict
 optname
 optparse
@@ -229,6 +231,7 @@ params
 paren
 parens
 passthru
+positionals
 png
 pragma
 pragma's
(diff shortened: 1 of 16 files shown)


Pierre-Sassoulas (commented):
I think it would make sense to create a subcommand ``pylint generate-config``. The original issue implicitly want to do a lot more than simply generating the configuration, a simple way to do more would be to have templates like "django-web", "ML-tensorflow", "legacy-project" (everything disabled at first ?), "new-project (everything enabled), and to simply propose to choose one of the template.

Pierre-Sassoulas:
>  We would then also need to do pylint run

Hurgh. That could be a cleanup for 3.0 but there isn't any other command that make sense to create right now. I was thinking of "suggest" as in ``pylint suggest`` (#5604), maybe ``baseline`` ? It's probably never going to justify breaking all the script in the world.

DanielNoord:
> > We would then also need to do pylint run
> 
> Hurgh. That could be a cleanup for 3.0 but there isn't any other command that make sense to create right now. I was thinking of "suggest" as in `pylint suggest` (#5604), maybe `baseline` ? It's probably never going to justify breaking all the script in the world.

My idea was to do a ``pylint --generate-config`` and ``pylint --generate-config-dont-ask-me-questions-just-fix-everything-for-me``. With the latter being essentially the first with all questions taking the default value.
That seems to have the good bits of both?

_I also don't want to do ``pylint --generate-config --outputfile=CONFIG`` as then ``--outputfile`` would end up in the general options... So it's probably best to handle everything interactively._

DanielNoord:
Going to close this as I am now using a different approach compared to the original attempt and would like a clean PR.

PR #3309, opened 2019-12-17
Add new `consider-using-generator` checker Issue #3165

ikraduya:
## Steps

- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [x] If it's a new feature or an important bug fix, add a What's New entry in `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Description
[Quoting from Issue #3165]
There are many cases of comprehensions inside of `any` or `all` calls that are unnecessary and should be replaced by a generator. 
For example:

```
some_list = list(range(1000))
test_old = any([x % 7 == 0 for x in some_list])
```
Instead, a generator would be less code, and way faster:

```
test_new = any(x % 7 == 0 for x in some_list)
```
Speed comparisons:
```
64 µs ± 1.67 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each) # old
447 ns ± 3.48 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each) # new
```
`any` and `all` are functions that profit from that because the loop is breaking and can short cut; the remaining elements are not even looped over if the test returns a True in the `any` function or a False in the `all` function.

This PR creates new `consider-using-generator` checker written in pylint/checkers/refactoring.py and add unit test consider_using_generator.py and consider_using_generator.txt

## Type of Changes

|   | Type |
| ------------- | ------------- |
|   | :bug: Bug fix  |
| ✓  | :sparkles: New feature |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

## Related Issue

Closes #3165

PR #6595, opened 2022-05-12
Raise use_a_generator for `sum()`, `max()`, `min()`

matusvalo:
## Type of Changes

## Description

Based on https://peps.python.org/pep-0289/ also `sum()`, `max()`, `min()` should avoid using
lists. Using generator, the memory is not allocated for all items but the items are processed lazily.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,7 +10,7 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
-* Checker ``use_a_generator`` now checks also ``max()``, ``min()`` and ``sum()``.
+* The ``use_a_generator``checker now also checks ``max()``, ``min()`` and ``sum()``.
 
 * We have improved our recognition of inline disable and enable comments. It is
   now possible to disable ``bad-option-value`` inline  (as long as you disable it before
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,8 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
+* Checker ``use_a_generator`` now checks also ``max()``, ``min()`` and ``sum()``.
+
 * We have improved our recognition of inline disable and enable comments. It is
   now possible to disable ``bad-option-value`` inline  (as long as you disable it before
   the bad option value is raised, i.e. ``disable=bad-option-value,bad-message`` not ``disable=bad-message,bad-option-value`` ) as well as certain other previously unsupported messages.
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -400,7 +400,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
         "R1729": (
             "Use a generator instead '%s(%s)'",
             "use-a-generator",
-            "Comprehension inside of 'any' or 'all' is unnecessary. "
+            "Comprehension inside of 'any', 'all', 'max', 'min' and 'sum' is unnecessary. "
             "A generator would be sufficient and faster.",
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas:
The benchmark result could be counterintuitive for small sum/max/min, see https://github.com/PyCQA/pylint/pull/3309#discussion_r576683109. Depending on the results we'd need to use  'consider-using-generator' or 'use-a-generator'.

matusvalo:
> The benchmark result could be counterintuitive for small sum/max/min, see [#3309 (comment)](https://github.com/PyCQA/pylint/pull/3309#discussion_r576683109). Depending on the results of a benchmark we'd need to use 'consider-using-generator' or 'use-a-generator'.

I will do benchmarking. Just a comment about it -  I think it is not 100% truth in PEP using generator is recommended. Doing microbenchmarks can be misleading - e.g. in python 3.9 generator version can be slower but in 3.11 it can be opposite:

> Python 3.11 is up to 10-60% faster than Python 3.10. [1]

I think we should follow best practises instead of tightly following microbenchmarks.

[1] https://docs.python.org/3.11/whatsnew/3.11.html#summary-release-highlights

matusvalo:
```
python -m timeit "b = sum([x for x in range(10000)]);"
# 500 loops, best of 5: 503 usec per loop
python -m timeit "b = sum(x for x in range(10000));"
# 500 loops, best of 5: 466 usec per loop


python -m timeit "b = max([x for x in range(10000)]);"
# 500 loops, best of 5: 574 usec per loop
python -m timeit "b = max(x for x in range(10000));"
# 500 loops, best of 5: 546 usec per loop

python -m timeit "b = min([x for x in range(10000)]);"
# 500 loops, best of 5: 569 usec per loop
python -m timeit "b = min(x for x in range(10000));"
# 500 loops, best of 5: 535 usec per loop
```

ChangeLog
```diff
@@ -10,6 +10,8 @@ Release date: TBA
 ..
   Put new features here and also in 'doc/whatsnew/2.14.rst'
 
+* Checker ``use_a_generator`` now checks also ``max()``, ``min()`` and ``sum()``.
```
```suggestion
* The ``use_a_generator``checker now also checks ``max()``, ``min()`` and ``sum()``.
```

DanielNoord (approved):

Pierre-Sassoulas:
> I think we should follow best practises instead of tightly following microbenchmarks.

I agree with the sentiment of the benchmark not being perfect. In fact, that's my bad for suggesting it, the decision for any/all vs list/set/dict was not taken based on the benchmark but based on the theory  that a generator for all/any will be cutting the execution tree in most case so it will be faster, sometime by a lot.

I think ``max``, ``min`` and ``sum`` will have to evaluate all elements whatever happens, so the benefit of a generator in term of performance will happens for very long containers only. We do have a message for best practises (``consider-``...), the wording is just stronger (``use-...``) if we have absolute certainty that the performance will be better/ i.e. all or any will be better whatever the interpreter or the size of the container. If a benchmark depending on the interpreter, the environment or the problem being solved is required the user need to consider the issue because we don't know for sure if the user favor best practices or raw performance.

Pierre-Sassoulas:
I think it should be ``consider-using-a-generator`` instead, for small list the performance are worst:
```
python -m timeit "b = sum([x for x in range(5)]);" # 329 nsec per loop
python -m timeit "b = sum(x for x in range(5));"   #369 nsec per loop

python -m timeit "b = max([x for x in range(5)]);" # 378 nsec per loop
python -m timeit "b = max(x for x in range(5));" # 425 nsec per loop

python -m timeit "b = min([x for x in range(5)]);" # 379 nsec per loop
python -m timeit "b = min(x for x in range(5));" # 417 nsec per loop
```

PR #6601, opened 2022-05-13
Add ``macOS`` to the test environments

### LINKED ISSUE #4759: Add Mac OS tests to Github Action
### Current problem

The tests suite is not run on Mac so the tests are failing locally for mac developpers.

### Desired solution

Run Mac tests in Github Actions.

### Additional context

See discussion in https://github.com/PyCQA/pylint/pull/4753#issuecomment-887819516

DanielNoord:

## Type of Changes

## Description

Closes #4759.

Test issues discussed in that issue appear to be fixed. Still, it would be good to test `macOS` on one python version to catch potential regressions. I consider it overkill to test on all supported versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -58,7 +58,7 @@ jobs:
       - name: Run pytest
         run: |
           . venv/bin/activate
-          pytest -vv --durations=20 --benchmark-disable --cov --cov-report= tests/
+          pytest --benchmark-disable --cov --cov-report= tests/
       - name: Run functional tests with minimal messages config
         run: |
           . venv/bin/activate
@@ -213,7 +213,7 @@ jobs:
       - name: Run pytest
         run: |
           . venv\\Scripts\\activate
-          pytest -vv --durations=20 --benchmark-disable tests/
+          pytest --benchmark-disable tests/
 
   tests-macos:
     name: run / ${{ matrix.python-version }} / macOS
@@ -223,7 +223,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # We only run on the latest supported version on Mac
+        # We only run on the oldest supported version on Mac
         python-version: [3.7]
     steps:
       - name: Check out code from GitHub
@@ -259,7 +259,7 @@ jobs:
       - name: Run pytest
         run: |
           . venv/bin/activate
-          pytest -vv --durations=20 --benchmark-disable tests/
+          pytest --benchmark-disable tests/
 
   tests-pypy:
     name: run / ${{ matrix.python-version }} / Linux
@@ -303,4 +303,4 @@ jobs:
       - name: Run pytest
         run: |
           . venv/bin/activate
-          pytest -vv --durations=20 --benchmark-disable tests/
+          pytest --benchmark-disable tests/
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -215,6 +215,52 @@ jobs:
...
+  tests-macos:
+    name: run / ${{ matrix.python-version }} / macOS
+    runs-on: macos-latest
+    timeout-minutes: 25
+    needs: tests-linux
+    strategy:
+      fail-fast: false
+      matrix:
+        # We only run on the latest supported version on Mac
+        python-version: [3.7]
```
cdce8p, line 227:
```suggestion
        # We only run on the latest supported version on Mac
        python-version: [3.10]
```
If it's that latest version, shouldn't it be `3.10`?

I do agree though to only test one, max two Python version for MacOS.

.github/workflows/tests.yaml
```diff
@@ -215,6 +215,52 @@ jobs:
...
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          python -m venv venv
+          . venv/bin/activate
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r requirements_test.txt
+      - name: Run pytest
+        run: |
+          . venv/bin/activate
+          pytest -vv --durations=20 --benchmark-disable tests/
```
cdce8p, line 262:
```suggestion
          pytest --durations=20 --benchmark-disable tests/
```
Why use `-vv` for pytest? Just my opinion, but I find it rather annoying to have such a long test output. That's something which bugs me frequently when I look at the Linux test output tbh.

.github/workflows/tests.yaml
DanielNoord: Ah that should probably be `oldest`. I'm not sure I don't really mind and don't expect to find many actual regression so using 1 to free up runners is probably best.

Do you have a preference for either?

.github/workflows/tests.yaml
DanielNoord: We did this when we had the frequent issues with the CI failing to see what failed when since we couldn't reproduce the issues locally and had to get some form of consistent data.

Now that those issues have been fixed we can probably remove those flags again.

Same goes for ``--durations``.

Pierre-Sassoulas: I can confirm that we never have any timeout for the main branch but we still have issue on the maintenance 2.13 branch so the issue was definitely fixed in between

Pierre-Sassoulas (approved):
Having just one macos runner make sense imo. The issue we had were with symlink in particular installations so even if we wanted to be exhaustive we'd have to test for multiple MacOs operating system, not multiple python interpreters which is unreasonable (MacOs/Windows runners 'cost' more).

.github/workflows/tests.yaml
cdce8p: Maybe a slight preference towards `3.10` but there are good arguments for `3.7` too.

.github/workflows/tests.yaml
cdce8p: > Same goes for `--durations`.

Just listing the longest running tests can be helpful sometimes. Maybe a limit of 10 is enough?

.github/workflows/tests.yaml
DanielNoord: Benefit of using `3.7` is that we're likely to update it when we bump last supported version. Feel free to change it if you want to! But I kept it on `3.7` for now!

cdce8p (approved):

Issue #6603, opened 2022-05-13
Crash when parsing empty `enumerate()` call


jacobtylerwalls:
When parsing the following file:

```python
for i, num in enumerate():
  pass
```

pylint crashed with an ``IndexError`` and with the following stacktrace:
```
Traceback (most recent call last):
  File "/Users/.../pylint/pylint/lint/pylinter.py", line 661, in _check_files
    self._check_file(get_ast, check_astroid_module, file)
  File "/Users/.../pylint/pylint/lint/pylinter.py", line 701, in _check_file
    check_astroid_module(ast_node)
  File "/Users/.../pylint/pylint/lint/pylinter.py", line 896, in check_astroid_module
    retval = self._check_astroid_module(
  File "/Users/.../pylint/pylint/lint/pylinter.py", line 946, in _check_astroid_module
    walker.walk(node)
  File "/Users/.../pylint/pylint/utils/ast_walker.py", line 92, in walk
    self.walk(child)
  File "/Users/.../pylint/pylint/utils/ast_walker.py", line 89, in walk
    callback(astroid)
  File "/Users/.../pylint/pylint/checkers/refactoring/refactoring_checker.py", line 652, in visit_for
    self._check_unnecessary_list_index_lookup(node)
  File "/Users/.../pylint/pylint/checkers/refactoring/refactoring_checker.py", line 1996, in _check_unnecessary_list_index_lookup
    or not isinstance(node.iter.args[0], nodes.Name)
IndexError: list index out of range
```

PR #6605, opened 2022-05-13
Remove release note for unreleased crash fix

jacobtylerwalls:
Followup to #6603. Remove release note since this was actually an unreleased issue in 2.14.


Pierre-Sassoulas (approved):
👍

PR #6606, opened 2022-05-13
Update ranges for error messages

cdce8p:
## Description
Update ranges for `using-constant-test` and `missing-parentheses-for-call-in-test` error messages.
Both should only be emitted on the `test` node itself and not the whole `if` statement.

**Old**
Screen Shot 2022-05-13 at 17 48 01

**New**
Screen Shot 2022-05-13 at 17 47 30

Related: #5466


Pierre-Sassoulas (approved):
Nice catch !

PR #6607, opened 2022-05-13
Remove `utils.subscriptable_with_postponed_evaluation_enabled`

cdce8p:
## Description
Update documentation and remove `utils.subscriptable_with_postponed_evaluation_enabled` added in #6536.

The helper function is just one more function call and with that specific name can't really be used anywhere else. Much better to just call the two functions directly. Which is already done elsewhere anyway. The helper has only been included in `2.14.0b1`, so I think it's ok to remove without deprecation.
```py
if is_postponed_evaluation_enabled(node) and is_node_in_type_annotation_context(node):
    ...
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -314,7 +314,8 @@ Release date: TBA
   Closes #6597
 
 * ``is_class_subscriptable_pep585_with_postponed_evaluation_enabled`` has been deprecated.
-  Use ``subscriptable_with_postponed_evaluation_enabled`` instead.
+  Use ``is_postponed_evaluation_enabled(node) and is_node_in_type_annotation_context(node)``
+  instead.
 
   Ref #6536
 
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
My bad I missed this one !

cdce8p:
@Pierre-Sassoulas I've updated the PR to remove the new helper function completely.

> Update documentation and remove `utils.subscriptable_with_postponed_evaluation_enabled` added in #6536.
> 
> The helper function is just one more function call and with that specific name can't really be used anywhere else. Much better to just call the two functions directly. Which is already done elsewhere anyway. The helper has only been included in `2.14.0b1`, so I think it's ok to remove without deprecation.
> ```py
> if is_postponed_evaluation_enabled(node) and is_node_in_type_annotation_context(node):
>     ...
> ```

Pierre-Sassoulas (approved):
👍

PR #6608, opened 2022-05-13
Fix `no-member` in type annotations with future import

### LINKED ISSUE #6594: `no-member` error in annotations
### Bug description

First discovered in https://github.com/PyCQA/astroid/pull/1555

```py
# pylint: disable=missing-docstring,unused-argument

import ast

def func(node: ast.Match) -> None:  # no-member
    ...
```

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
test.py:6:15: E1101: Module 'ast' has no 'Match' member (no-member)
```

### Expected behavior

With `from __future__ import annotations`, the type annotation isn't evaluated at runtime and thus won't throw an error. Furthermore it's possible to use classes that way which are only defined in stub files.

Similarly to string annotations, pylint should just ignore these cases and let type checker handle it.

### Pylint version

```shell
pylint 2.14.0-b1
astroid 2.12.0-dev0
Python 3.9.12
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

cdce8p:
## Description
Don't emit ``no-member`` inside type annotations with ``from __future__ import annotations``.

With PEP 563, annotations are stored as strings and thus not executed. This can cause false-positives if attributes are used which don't exist in the version used to run pylint.
```py
# pylint: disable=missing-docstring,unused-argument

import ast
import sys

if sys.version_info >= (3, 10):
    def func(node: ast.Match) -> None:  # no-member
        ...
```

If Python 3.8 is used, pylint would currently emit a `no-member` error for `ast.Match` (which was added in `3.10`).
By excluding type annotations with `from __future__ import annotations`, the behavior will be similar to normal string annotations (which don't emit an error).
```py
def func(node: "ast.Match") -> None:
    ...
```

Closes #6594

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -321,6 +321,11 @@ Release date: TBA
 * Update ranges for ``using-constant-test`` and ``missing-parentheses-for-call-in-test``
   error messages.
 
+* Don't emit ``no-member`` inside type annotations with
+  ``from __future__ import annotations``.
+
+  Closes #6594
+
 ..
   Insert your changelog randomly, it will reduce merge conflicts
   (Ie. not necessarily at the end)
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -266,6 +266,11 @@ Other Changes
 * Update ranges for ``using-constant-test`` and ``missing-parentheses-for-call-in-test``
   error messages.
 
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas (approved):
LGTM

PR #6611, opened 2022-05-13
Suppress useless-super-delegation on __hash__ in classes with __eq__

### LINKED ISSUE #3934: useless-super-delegation incorrect warning
### Steps to reproduce
1. With pylint
### pylint --version output
```
pylint 2.6.0
astroid 2.4.2
Python 3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AMD64)]
```
on windows 10

2. Given the following file:
```
class Variable(Reference):
    def __eq__(self, other: object) -> bool:
        if not isinstance(other, Variable):
            return False

        return super().__eq__(other)

    def __hash__(self) -> int:
        return super().__hash__()
```

3. `pylint variable.py` results in:
```
variable.py:18:4: W0235: Useless super delegation in method '__hash__' (useless-super-delegation)
```
4. Yet removing this `__hash__` function and running python 3.9.0 results in
```
TypeError: unhashable type: 'Variable'
```
whenever it is used in a hashable collection (such as a set).

### Current behavior
I don't know how correct the python code is, but
either python or pylint is wrong.

### Expected behavior
Make python and pylint agree: either by removing, changing, or generating another warning.

timmartin:
## Type of Changes

## Description

If a class overrides `__eq__` then we should allow trivial overrides of `__hash__` without issuing a `useless-super-delegation` warning, because the override can be significant.

As explained in the Python docs:
> A class that overrides `__eq__` and does not define `__hash__` will have its `__hash__()` implicitly set to `None`. When the
> `__hash__` method of a class is None, instances of the class will raise an appropriate `TypeError` when a program attempts to retrieve their hash value, and will also be correctly identified as unhashable when checking `isinstance(obj, collections.abc.Hashable)`.

Closes #3934

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -133,7 +133,8 @@ def _is_trivial_super_delegation(function: nodes.FunctionDef) -> bool:
     """
     if (
         not function.is_method()
-        # With decorators is a change of use
+        # Adding decorators to a function changes behavior and
+        # constitutes a non-trivial change.
         or function.decorators
     ):
         return False
@@ -146,7 +147,7 @@ def _is_trivial_super_delegation(function: nodes.FunctionDef) -> bool:
 
     statement = body[0]
     if not isinstance(statement, (nodes.Expr, nodes.Return)):
-        # Doing something else than what we are interested into.
+        # Doing something else than what we are interested in.
         return False
 
     call = statement.value
@@ -1227,7 +1228,7 @@ a metaclass class method.",
 
     visit_asyncfunctiondef = visit_functiondef
 
-    def _check_useless_super_delegation(self, function: nodes.FunctionDef):
+    def _check_useless_super_delegation(self, function: nodes.FunctionDef) -> None:
         """Check if the given function node is an useless method override.
 
         We consider it *useless* if it uses the super() builtin, but having
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1222,6 +1222,13 @@ a metaclass class method.",
         if call.func.attrname != function.name:
             return
 
+        # Classes that override __eq__ should also override
+        # __hash__, even a trivial override is meaningful
+        if function.name == '__hash__':
+            for other_method in function.parent.mymethods():
+                if other_method.name == '__eq__':
+                    return
+
         # Should be a super call with the MRO pointer being the
         # current class and the type being the current instance.
         current_scope = function.parent.scope()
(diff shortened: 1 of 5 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -127,6 +127,62 @@ def _definition_equivalent_to_call(definition, call):
     return all(kw in call.args or kw in definition.kwonlyargs for kw in call.kws)
 
 
+def _is_trivial_super_delegation(function: nodes.FunctionDef) -> bool:
+    """Check whether a function definition is a method consisting only of a
+    call to the same function on the superclass.
+    """
+    if (
+        not function.is_method()
+        # With decorators is a change of use
```
DanielNoord, line 136: I'm not sure I understand what you mean with this comment. Could you be a little bit more specific?

pylint/checkers/classes/class_checker.py
```diff
@@ -127,6 +127,62 @@ def _definition_equivalent_to_call(definition, call):
...
+
+    body = function.body
+    if len(body) != 1:
+        # Multiple statements, which means this overridden method
+        # could do multiple things we are not aware of.
+        return False
+
+    statement = body[0]
+    if not isinstance(statement, (nodes.Expr, nodes.Return)):
+        # Doing something else than what we are interested into.
```
DanielNoord, line 149:
```suggestion
        # Doing something else than what we are interested in.
```

pylint/checkers/classes/class_checker.py
```diff
@@ -127,6 +127,62 @@ def _definition_equivalent_to_call(definition, call):
...
+    if (
+        not isinstance(call, nodes.Call)
+        # Not a super() attribute access.
+        or not isinstance(call.func, nodes.Attribute)
+    ):
+        return False
+
+    # Should be a super call.
+    try:
+        super_call = next(call.func.expr.infer())
```
DanielNoord, line 162: Can you use `safe_infer` here?

pylint/checkers/classes/class_checker.py
```diff
@@ -1171,7 +1227,7 @@ def visit_functiondef(self, node: nodes.FunctionDef) -> None:
 
     visit_asyncfunctiondef = visit_functiondef
 
-    def _check_useless_super_delegation(self, function):
+    def _check_useless_super_delegation(self, function: nodes.FunctionDef):
```
DanielNoord, line 1230:
```suggestion
    def _check_useless_super_delegation(self, function: nodes.FunctionDef) -> None:
```

DanielNoord (changes requested):
Thanks for working on this!

I made comments on the new method before I saw you just copied what was already there. Perhaps you can't make the comment better, but using `safe_infer` would be good. We would fix a long standing bug and improve our code coverage in one PR 🎉

DanielNoord (approved):

Pierre-Sassoulas (approved):
👍

PR #6589, opened 2022-05-12
Reorganize the documentation table of content

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :scroll: Docs          |

## Description

Create a user guide part and a developper guide part to distinguish the two. First step for #5728. If we want to do that properly (old links working)  we need to do quite a lot of redirection.

Link that need to still works after this:
https://pylint--6589.org.readthedocs.build/en/6589/user_guide/index.html
https://pylint--6589.org.readthedocs.build/en/6589/development_guide/index.html
https://pylint--6589.org.readthedocs.build/en/6589/development_guide/contribute.html
https://pylint--6589.org.readthedocs.build/en/6589/how_tos/custom_checkers.html
https://pylint--6589.org.readthedocs.build/en/6589/how_tos/index.html
https://pylint--6589.org.readthedocs.build/en/6589/how_tos/plugins.html
https://pylint--6589.org.readthedocs.build/en/6589/how_tos/transform_plugins.html
https://pylint--6589.org.readthedocs.build/en/6589/technical_reference/checkers.html
https://pylint--6589.org.readthedocs.build/en/6589/technical_reference/index.html
https://pylint--6589.org.readthedocs.build/en/6589/technical_reference/startup.html
https://pylint--6589.org.readthedocs.build/en/6589/technical_reference/features.html
https://pylint--6589.org.readthedocs.build/en/6589/user_guide/options.html
https://pylint--6589.org.readthedocs.build/en/6589/user_guide/ide_integration/ide-integration.html
https://pylint--6589.org.readthedocs.build/en/6589/development_guide/tests/index.html
https://pylint--6589.org.readthedocs.build/en/6589/development_guide/tests/install.html
https://pylint--6589.org.readthedocs.build/en/6589/development_guide/tests/launching_test.html
https://pylint--6589.org.readthedocs.build/en/6589/development_guide/tests/writting_test.html
https://pylint--6589.org.readthedocs.build/en/6589/development_guide/contributor_guide.html
https://pylint--6589.org.readthedocs.build/en/6589/user_guide/ide_integration/flymake-emacs.html
https://pylint--6589.org.readthedocs.build/en/6589/user_guide/ide_integration/textmate.html
https://pylint--6589.org.readthedocs.build/en/6589/user_guide/pre-commit-integration.html
https://pylint--6589.org.readthedocs.build/en/6589/user_guide/options.html
[...]

PR #6613, opened 2022-05-14
[Doc] Create a contact section and cleanup the introduction

Pierre-Sassoulas:
## Type of Changes

## Description

Small reviewable extract from #6589


doc/backlinks.rst
```diff
@@ -1,14 +0,0 @@
-
-Some projects using Pylint
```
Pierre-Sassoulas, line 2: Removed because it's super outdated and we're not some kind of startup trying to secure funding so it's noise.

doc/contact/index.rst
```diff
@@ -0,0 +1,8 @@
+Contact
```
DanielNoord, line 1: I'm not sure this page is needed. I don't really like how this works out:
https://pylint--6613.org.readthedocs.build/en/6613/contact/index.html

Would be better if clicking on `Contact` in the sidebar brings you to the page immediately.

doc/index.rst
```diff
@@ -1,28 +1,48 @@
+Pylint documentation
+====================
 
-Pylint User Manual
-==================
+Pylint is a `static code analyser`_ for python 2 or 3. Only python interpreter above 3.7.2 are supported
+in the latest version.
```
DanielNoord, line 5:
```suggestion
Pylint is a `static code analyser`_ for Python 3. The latest version supports Python 3.7.2 and newer.
```

doc/index.rst
```diff
@@ -1,28 +1,48 @@
...
 
-Pylint is a tool that checks for errors in Python code, tries to enforce a
-coding standard and looks for code smells. It can also look for certain type
-errors, it can recommend suggestions about how particular blocks can be
-refactored and can offer you details about the code's complexity.
+Pylint analyse your code without actually running it. It checks for errors, enforce a coding
+standard, looks for `code smells`_, and can make suggestions about how the code could be refactored.
```
DanielNoord, line 8:
```suggestion
Pylint analyses your code without actually running it. It checks for errors, enforces a coding
standard, looks for `code smells`_, and can make suggestions about how the code could be refactored.
```

doc/index.rst
```diff
@@ -1,28 +1,48 @@
...
+.. _mypy: https://github.com/python/mypy
+.. _pyright: https://github.com/microsoft/pyright
+.. _pyre: https://github.com/facebook/pyre-check
+.. _black: https://github.com/psf/black
+.. _isort: https://pycqa.github.io/isort/
+.. _`static code analyser`: https://en.wikipedia.org/wiki/Static_code_analysis
+.. _`code smells`: https://martinfowler.com/bliki/CodeSmell.html
+
+Pylint can infer actual values from your code using it's internal code representation (astroid).
+If your code is ``import logging as argparse``, Pylint can knows that ``argparse.error(...)``
```
DanielNoord, line 25:
```suggestion
If your code is ``import logging as argparse``, Pylint will know that ``argparse.error(...)``
```

doc/index.rst
```diff
@@ -1,28 +1,48 @@
...
+is in fact a logging call and not an argparse call.
+
+Pylint isn't smarter than you: it may warn you about things that you have
+conscientiously done or checks for some things that you don't care about.
+During adoption, especially in a legacy project where pylint was never enforced,
+it's best to start with the ``--errors-only`` flag, then disable
+convention and refactor message with ``--disable=C,R`` and progressively
+re-evaluate and re-enable messages as your priorities evolve.
 
 .. toctree::
```
DanielNoord, line 35: Let's remove this `toctree` as well. It doesn't look very pretty, see:
https://pylint--6613.org.readthedocs.build/en/6613/index.html

doc/intro.rst
```diff
@@ -1,58 +0,0 @@
-.. -*- coding: utf-8 -*-
-
-==============
```
DanielNoord, line 3: Is it worth breaking old links for this? I think we should add a redirect.

doc/support.rst
```diff
@@ -1,17 +0,0 @@
-Support
```
DanielNoord, line 1: Same here, let's add a redirect.

doc/faq.rst
```diff
@@ -303,13 +303,3 @@ Pylint, you can set ``evaluation`` to the above expression to get the new
 behavior. Likewise, since negative values are still technically supported,
 ``evaluation`` can be set to a version of the above expression that does not
 enforce a floor of zero.
-
-6.2 I think I found a bug in Pylint. What should I do?
```
DanielNoord, line 307: This doesn't hurt right?

doc/contact/contact.rst
```diff
@@ -0,0 +1,56 @@
+Bug reports, feedback
+---------------------
+.. _bug reports, feedback:
+
+You think you have found a bug in Pylint? Well, this may be the case
+since Pylint and python are under heavy development!
```
DanielNoord, line 6:
```suggestion
since Pylint and Python are under heavy development!
```

doc/contact/contact.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Please take the time to check if it is already in the issue tracker at
+https://github.com/PyCQA/pylint
+
+Note that if you don't find something you have expected in Pylint's issue
+tracker, it may be because it is an issue with one of its dependencies, namely
+astroid:
```
DanielNoord, line 13:
```suggestion
Note that the issue might also be reported in one of Pylint's major dependencies,
astroid:
```

DanielNoord (changes requested):

doc/index.rst
Pierre-Sassoulas: You can use pylint for python 2 : it's version 1.9.6 
```suggestion
Pylint is a `static code analyser`_ for python 2 or 3. Only python interpreter above 3.7.2 are still maintained currently.
```

DanielNoord: Not a super issue for me, but let's not mention Python 2 in the first line of our index. Small contribution to the general `Python` ecosystem to nudge everybody towards 3 😄

doc/faq.rst
Pierre-Sassoulas: Imo the FAQ is just something were we duplicate information because the overall organization of the doc is horrible. I'm aiming to removing it (almost entirely).

doc/index.rst
Pierre-Sassoulas: I mean I'm ok with that but how do we tell sphinx that it needs to display this toctree and include files ?

DanielNoord: Can't you hide a `toctree` with an attribute. `hidden: True` I believe? I think I use it in `pydocstringformatter`.

doc/index.rst
Pierre-Sassoulas: I looked at mypy's doc for an example, this is why I changed it : https://mypy.readthedocs.io/en/stable/index.html

I don't think Guido can be suspected of trying to incite people to keep/start using python 2 😄

doc/index.rst
Pierre-Sassoulas: Thanks for the tip 😄 This is indeed a lot better

doc/index.rst
DanielNoord: I think the difference is that `mypy` can actually check `2.7` code although you need to run it using `3.6`. I don't actually know this: but do we still support checking `2.7` code? It feels like that would violate a lot of assumptions in our checkers which wouldn't work.

Pierre-Sassoulas: We have old versions that do check python 2, some versions only python 2. If you pip install pylint you'll get a version compatible with your interpreter. I did not try to check if pylint 2.14 can check python 2 code, probably not as you say.

doc/intro.rst
Pierre-Sassoulas: I've searched a way to do it in the configuration file but I don't think it's possible. Do you happen to know how to do it ?

DanielNoord: No, I have never done this, but it should be possible. At least with a Sphinx extension.

doc/index.rst
DanielNoord:
```suggestion
Pylint is a `static code analyser`_ for Python 2 or 3. The latest version supports Python 3.7.2 and above.
```

doc/index.rst
```diff
@@ -1,28 +1,53 @@
...
+3.7.2 and above.
 
-Pylint is a tool that checks for errors in Python code, tries to enforce a
-coding standard and looks for code smells. It can also look for certain type
-errors, it can recommend suggestions about how particular blocks can be
-refactored and can offer you details about the code's complexity.
+Pylint analyses your code without actually running it. It checks for errors, enforces a coding
+standard, looks for `code smells`_, and can make suggestions about how the code could be refactored.
 
+Project that you might want to use alongside pylint include flake8_ (faster and simpler checks
```
DanielNoord, line 10:
```suggestion
Projects that you might want to use alongside pylint include flake8_ (faster and simpler checks
```

doc/index.rst
```diff
@@ -1,28 +1,53 @@
...
+with very few false positives), mypy_, pyright_ or pyre_ (typing checks), bandit_ (security
+oriented checks), black_ and isort_ (auto-formatting), autoflake_ (automated removal of
+unused import or variable), pyupgrade_ (automated upgrade to newer python syntax) and
+pydocstringformatter_ (automated pep257).
```
DanielNoord, line 14: 😄

doc/index.rst
```diff
@@ -1,28 +1,53 @@
...
+.. _pyre: https://github.com/facebook/pyre-check
+.. _black: https://github.com/psf/black
+.. _autoflake: https://github.com/myint/autoflake
+.. _pyupgrade: https://github.com/asottile/pyupgrade
+.. _pydocstringformatter: https://github.com/DanielNoord/pydocstringformatter
+.. _isort: https://pycqa.github.io/isort/
+.. _`static code analyser`: https://en.wikipedia.org/wiki/Static_code_analysis
+.. _`code smells`: https://martinfowler.com/bliki/CodeSmell.html
+
+Pylint can infer actual values from your code using it's internal code representation (astroid).
```
DanielNoord, line 29:
```suggestion
Pylint can infer actual values from your code using its internal code representation (astroid).
```

doc/faq.rst
```diff
@@ -6,50 +6,20 @@
...
-2.3 What are Pylint's dependencies?
------------------------------------
-
-Pylint depends on astroid_ and a couple of other packages.
-See the following section for details on what versions of Python are
-supported.
-
-.. _`astroid`: https://github.com/PyCQA/astroid
-
-2.4 What versions of Python is Pylint supporting?
```
DanielNoord, line 48: I think it wouldn't hurt to keep this in the FAQ. As I can easily see people wondering about this and the FAQ would be one the first places I would look for this information.

DanielNoord (changes requested):
Last comments about the actual content. Only the redirections left.

doc/intro.rst
Pierre-Sassoulas: It seems it should be possible to do it with readthedoc conf directly but after 60+mn I just took an extension to do it.

doc/index.rst
Pierre-Sassoulas: It's your last chance to give this tool a cool name, Daniel.

doc/faq.rst
Pierre-Sassoulas: I'm going to remove the faq or create two faqs, one for users and one for developer. We might as well document the pylint version vs python interpreter... or maybe pip is handling that and you don't need to know ? Do you mind if we do that in the following merge request ?

doc/index.rst
DanielNoord: Ship has sailed I think 😄 And I really don't know anything better.

doc/conf.py
```diff
@@ -43,8 +46,16 @@
     "pylint_options",
     "sphinx.ext.autosectionlabel",
     "sphinx.ext.intersphinx",
+    "sphinx_reredirects",
```
DanielNoord, line 49: This project doesn't seem very well maintained. After the initial release of `0.0.1` a lot of open issue were never dealt with.

Are we sure want to use it?

Pierre-Sassoulas: Yeah 0.0.1 does not inspire confidence. But the current version is working well enough, and it's just a dict of simple redirect so worst case scenario we can still switch later if required.

Pierre-Sassoulas: ( https://pylint--6613.org.readthedocs.build/en/6613/intro.html / https://pylint--6613.org.readthedocs.build/en/6613/support.html )

DanielNoord (approved):

PR #6614, opened 2022-05-14
Add typing to `pylint.pyreverse.inspector`

DudeNr33:
-->

## Description

Add remaining typing to `inspector.py`.
This was the last untyped module.
After this change I can take a look at the refactoring TODOs we found.

Especially the parts of `inspector.py` that deal with interfaces and implementation links can be removed in my opinion, as their behaviour is based on the `__implements__` attribute which is not a standard Python feature and the corresponding PEP 245 is rejected. Plus, `pylint`'s codebase itself does not use it any more.
Those are also the only places in this module where I used `Any`.

### DIFF
--- a/pylint/pyreverse/inspector.py
+++ b/pylint/pyreverse/inspector.py
@@ -6,12 +6,13 @@
 
 Try to resolve definitions (namespace) dictionary, relationship...
 """
+
 from __future__ import annotations
 
 import collections
 import os
 import traceback
-from collections.abc import Iterator
+from collections.abc import Generator
 from typing import Any, Callable, Optional
 
 import astroid
@@ -44,7 +45,7 @@ def interfaces(
     node: nodes.ClassDef,
     herited: bool = True,
     handler_func: Callable[[nodes.NodeNG | Any], bool] = _iface_hdlr,
-) -> Iterator[Any] | None:
+) -> Generator[Any, None, None]:
     """Return an iterator on interfaces implemented by the given class node."""
     try:
         implements = astroid.bases.Instance(node).getattr("__implements__")[0]
--- a/pylint/pyreverse/inspector.py
+++ b/pylint/pyreverse/inspector.py
@@ -6,22 +6,30 @@
 
 Try to resolve definitions (namespace) dictionary, relationship...
 """
+from __future__ import annotations
+
 import collections
 import os
 import traceback
+from collections.abc import Callable, Iterator
+from typing import Any, Optional
 
 import astroid
 from astroid import nodes
 
 from pylint.pyreverse import utils
 
+_WrapperFuncT = Callable[[Callable[[str], nodes.Module], str], Optional[nodes.Module]]
 
-def _iface_hdlr(_):
+
+def _iface_hdlr(_: nodes.NodeNG | Any) -> bool:
     """Handler used by interfaces to handle suspicious interface nodes."""
     return True
 
 
-def _astroid_wrapper(func, modname):
+def _astroid_wrapper(
+    func: Callable[[str], nodes.Module], modname: str
+) -> nodes.Module | None:
     print(f"parsing {modname}...")
     try:
         return func(modname)
@@ -32,7 +40,11 @@ def _astroid_wrapper(func, modname):
     return None
 
 
-def interfaces(node, herited=True, handler_func=_iface_hdlr):
+def interfaces(
+    node: nodes.ClassDef,
+    herited: bool = True,
+    handler_func: Callable[[nodes.NodeNG | Any], bool] = _iface_hdlr,
+) -> Iterator[Any] | None:
     """Return an iterator on interfaces implemented by the given class node."""
     try:
         implements = astroid.bases.Instance(node).getattr("__implements__")[0]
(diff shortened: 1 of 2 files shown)


pylint/pyreverse/inspector.py
```diff
@@ -6,22 +6,30 @@
 
 Try to resolve definitions (namespace) dictionary, relationship...
 """
+from __future__ import annotations
```
DanielNoord, line 9:
```suggestion

from __future__ import annotations
```

pylint/pyreverse/inspector.py
```diff
@@ -6,22 +6,30 @@
...
 import traceback
+from collections.abc import Callable, Iterator
+from typing import Any, Optional
 
 import astroid
 from astroid import nodes
 
 from pylint.pyreverse import utils
 
+_WrapperFuncT = Callable[[Callable[[str], nodes.Module], str], Optional[nodes.Module]]
```
DanielNoord, line 22: These need to be imported from `typing` since they are used at runtime.

pylint/pyreverse/inspector.py
```diff
@@ -32,7 +40,11 @@ def _astroid_wrapper(func, modname):
     return None
 
 
-def interfaces(node, herited=True, handler_func=_iface_hdlr):
+def interfaces(
+    node: nodes.ClassDef,
+    herited: bool = True,
+    handler_func: Callable[[nodes.NodeNG | Any], bool] = _iface_hdlr,
+) -> Iterator[Any] | None:
```
DanielNoord, line 47: I think this should be typed as a `Generator` using something like `Generator[Any, None, None]`.

pylint/pyreverse/inspector.py
DudeNr33: Some day I will learn this! 😄 
Maybe that could be a pre-commit check or pylint extension as well?

pylint/pyreverse/inspector.py
DudeNr33: Good catch, thank you.

pylint/pyreverse/inspector.py
DudeNr33: Fixed.

pylint/pyreverse/inspector.py
jacobtylerwalls: I'm curious, what is the source for this convention?

jacobtylerwalls: PEP8 seems to say the opposite (emphasis mine): Imports are always put at the top of the file, _just after_ any module comments and docstrings, and before module globals and constants.

DanielNoord: It's my interpretation of PEP8's: `You should put a blank line between each group of imports.` Since the `module docstring` is certainly not the same "group" as the future import it makes sense the separate them just like you separate the last import from the first definition. 

Note the example just below your quote:
```python
"""This is the example module.

This module does stuff.
"""

from __future__ import barry_as_FLUFL

__all__ = ['a', 'b', 'c']
__version__ = '0.1'
__author__ = 'Cardinal Biggles'

import os
import sys
```

https://peps.python.org/pep-0008/ for reference.


But, I agree this is somewhat arbitrary. There does seem to be consensus though as https://github.com/psf/black/pull/2996 seems to have general support and would help us with this as well.

pylint/pyreverse/inspector.py
```diff
@@ -301,11 +319,14 @@ def _imported_module(self, node, mod_path, relative):
 
 
 def project_from_files(
-    files, func_wrapper=_astroid_wrapper, project_name="no name", black_list=("CVS",)
-):
+    files: list[str],
+    func_wrapper: _WrapperFuncT = _astroid_wrapper,
+    project_name: str = "no name",
+    black_list: tuple[str, ...] = ("CVS",),
```
DanielNoord, line 325: We might want to make this a constant and keep it up to date with `pylint`'s default value for the similar option?

DanielNoord (approved):
LGTM!

pylint/pyreverse/inspector.py
jacobtylerwalls: Ah, yeah I wondered if that's what you were picking up on. I don't think the document is very clear. I think it's not clear that it's incompatible with "put a blank line before each subsequent group of imports". I'm not going to open a PR against PEP8 (😅 ) but maybe I'll chime in on the black issue, thanks for posting it!

DanielNoord: I do think a suggestion to edit the PEP and linking to `black` wouldn't hurt. Just to see if there is actual agreement on it.

If `black` is going to enforce it it will be pretty "standard" anyway.

jacobtylerwalls: I guess I'm feeling we shouldn't enforce this style in pylint or do comments on it. If `black` changes, we'll get it for free. That example is showing a blank line at the top, but it's not "between" two groups of imports, it's between a docstring, so it's just a random blank line. But no big deal either way obviously.

DanielNoord: Fine with me! Agree that increasing maintenance churn for something that is likely auto-fixed by `black` in a couple of weeks is not worth the effort!

Pierre-Sassoulas: isort is the auto-formatter that would enforce this I think.

pylint/pyreverse/inspector.py
DudeNr33: You mean the `black_list` / `ignore`? That would make sense imo.

DanielNoord: Yes!

PR #6612, opened 2022-05-14
Fix "dictionary changed size during iteration"

nyabkun:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

`pyreverse -k -m y -o puml -p manim.camera manim.camera`

This command produced the following error.

```
File "C:\ScoopG\apps\python\current\lib\site-packages\pylint\pyreverse\inspector.py", line 177, in visit_classdef
   for assignattrs in node.instance_attrs.values():
RuntimeError: dictionary changed size during iteration
```
---

ManimCE
https://github.com/ManimCommunity/manim

manim            0.15.2
pylint              2.13.8

---

I'm new here and I don't fully understand the structure of the library.
But at any rate, with this change, I could fix the problem.

PR #6617, opened 2022-05-15
Run linking and diadef handling with patched `sys.path`

DudeNr33:

## Type of Changes

## Description

Everything until (and including) `handler.get_diadefs` should happen in the `with fix_import_path(args)` block:
operations that include anything related to inference may try to import modules.

As my original goal was to enable testing the fix of #6612, I included the other necessary change for that:
Do not treat modules with a leading underscore in the functional tests folder as standalone test files.
And I noticed that I didn't write tests for the `pyreverse` testutils so far, so added some basic tests for that as well.


Pierre-Sassoulas (approved):
LGTM !

PR #6623, opened 2022-05-16
Faster makefile for doc development offline

Pierre-Sassoulas:
## Type of Changes

## Description

I did a lot of doc modification recently for #6589 and the makefile installing dependency each time is taking a lot of time unnecessarily. I'm also thinking about lazy generating message files etc.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -37,7 +37,7 @@ help:
 
 clean:
 	-rm -rf $(BUILDDIR)/* messages/convention messages/error messages/fatal messages/information \
-	  messages/refactor messages/warning
+	  messages/refactor messages/warning technical_reference/extensions.rst technical_reference/features.rst
 
 install-dependencies:
 	@echo "Install dependencies"
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -38,14 +38,14 @@ help:
 clean:
 	-rm -rf $(BUILDDIR)/*
 
-html:
-	@echo
-	@echo "Install dependencies"
-	$(PIP) install furo
-	@echo "Finished installing dependencies"
+build-html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	@echo "HTML build complete"
+
+html: build-html
+	@echo "Install dependencies"
+	$(PIP) install -r requirements.txt
 	$(SPHINXBUILD) -b linkcheck $(ALLSPHINXOPTS) $(BUILDDIR)/linkcheck
-	@echo
 	@echo "Link check complete; look for any errors in the above output " \
 	      "or in $(BUILDDIR)/linkcheck/output.txt."
 
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -13,7 +13,7 @@ PYTHONPATH    =
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees -T -E -W --keep-going $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
+ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees -T -W --keep-going $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
 .PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck
 
@@ -36,7 +36,9 @@ help:
 	@echo "  linkcheck  to check all external links for integrity"
 
 clean:
-	-rm -rf $(BUILDDIR)/*
+	-rm -rf $(BUILDDIR)/* messages/convention messages/error messages/fatal messages/information \
+	  messages/refactor messages/warning
+
 
 build-html:
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -39,14 +39,16 @@ clean:
 	-rm -rf $(BUILDDIR)/* messages/convention messages/error messages/fatal messages/information \
 	  messages/refactor messages/warning
 
+install-dependencies:
+	@echo "Install dependencies"
+	cd ../ && $(PIP) install -r doc/requirements.txt && cd doc/
 
 build-html:
+	@echo "If you did not run 'make install-dependencies', you need to."
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
 	@echo "HTML build complete"
 
-html: build-html
-	@echo "Install dependencies"
-	$(PIP) install -r requirements.txt
+html: install-dependencies build-html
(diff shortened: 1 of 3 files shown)


doc/Makefile
```diff
@@ -13,7 +13,7 @@ PYTHONPATH    =
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees -T -E -W --keep-going $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
+ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees -T -W --keep-going $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
```
Pierre-Sassoulas, line 16: >   -E                don't use a saved environment, always read all files

In the CI we have a fresh environment, so this option is just making repeated build slower locally

doc/exts/pylint_messages.py
```diff
@@ -158,32 +158,58 @@ def _get_all_messages(
...
         for message in messages:
-            checker_module_rel_path = os.path.relpath(
-                message.checker_module_path, PYLINT_BASE_PATH
-            )
-            messages_file = os.path.join(category_dir, f"{message.name}.rst")
-            with open(messages_file, "w", encoding="utf-8") as stream:
-                stream.write(
-                    f""".. _{message.name}:
+            if not message_needs_update(message, category):
+                print(f"{message.name} is up to date.", end="\r")
```
DanielNoord, line 200: Shall we log this? I haven't tested locally, but doesn't this feel spammy?

doc/exts/pylint_messages.py
```diff
@@ -158,32 +158,58 @@ def _get_all_messages(
...
+    message_path_stats = message_path.stat().st_mtime
+    checker_path_stats = Path(message_data.checker_module_path).stat().st_mtime
+    return checker_path_stats > message_path_stats
+
+
+def get_category_directory(category: str) -> Path:
+    return PYLINT_MESSAGES_PATH / category
+
+
+def get_message_path(category: str, message: MessageData) -> Path:
```
DanielNoord, line 187:
```suggestion
def _get_message_path(category: str, message: MessageData) -> Path:
```

doc/exts/pylint_messages.py
```diff
@@ -158,32 +158,58 @@ def _get_all_messages(
...
+    """Do we need to regenerate this message .rst ?"""
+    message_path = get_message_path(category, message_data)
+    if not message_path.exists():
+        return True
+    message_path_stats = message_path.stat().st_mtime
+    checker_path_stats = Path(message_data.checker_module_path).stat().st_mtime
+    return checker_path_stats > message_path_stats
+
+
+def get_category_directory(category: str) -> Path:
```
DanielNoord, line 183:
```suggestion
def _get_category_directory(category: str) -> Path:
```

Just to be safe 😄

doc/exts/pylint_messages.py
```diff
@@ -158,32 +158,58 @@ def _get_all_messages(
...
                 category = MSG_TYPES_DOC[old_name[0][0]]
                 old_messages[category][(old_name[1], old_name[0])].append(
-                    (message.symbol, MSG_TYPES_DOC[message.msgid[0]])
+                    (message.symbol, msg_type)
                 )
 
     return messages_dict, old_messages
 
 
+def message_needs_update(message_data: MessageData, category: str) -> bool:
```
DanielNoord, line 173:
```suggestion
def _message_needs_update(message_data: MessageData, category: str) -> bool:
```

I like this trick!

DanielNoord (changes requested):

doc/requirements.txt
```diff
@@ -1,4 +1,3 @@
 Sphinx==4.5.0
 sphinx-reredirects<1
 furo==2022.4.7
```
DanielNoord, line 3: We need the `-e .` for RTD as it needs access to `pylint`. We provide `doc/requirements.tx` while we're `cd`'ed into `root`.

I know I looked into this previously but couldn't figure out an elegant solution of allowing both RTD and `cd docs` + `pip install -r ...` to work.

doc/exts/pylint_messages.py
Pierre-Sassoulas: The ``, end="\r")`` always rewrite the same line. ``\r`` is a carriage return so it behave like a typing machine where you don't provide a line feed (it's the reason why windows end of line are/can be``\r\n`` and not ``\n``, really legacy stuff.)

doc/requirements.txt
Pierre-Sassoulas: How about we suggest to install from the root with ``doc/requirements.txt``, and we remove the requirement from the makefile ? (I.e. it's the job of the prepare-base job to install the proper dependencies.)

doc/exts/pylint_messages.py
DanielNoord: Smart, learned a lot from this PR 🎉

doc/requirements.txt
DanielNoord: Can we `cd` in a makefile? That way we could `cd` before and after installing the documentation dependencies.

I do think it would be nice to have one command you can run to completely make the documentation without any further trouble.

doc/exts/pylint_messages.py
Pierre-Sassoulas: Fun fact originally each OS had chosen a different line ending convention.
- Windows : carriage return then line feed (most historically correct regarding what you really do on a typing machine)
- Mac: carriage return (would rewrite the current line)
- Linux : line feed (would write on the next line but starting from the column of the last char)

doc/development_guide/contribute.rst
```diff
@@ -87,6 +87,10 @@ Tips for Getting Started with Pylint Development
 Building the documentation
 ----------------------------
 
-We use **tox** for building the documentation::
+You can use the makefile and build-html command for building the documentation during developement::
```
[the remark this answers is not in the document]
Pierre-Sassoulas: 👍

doc/requirements.txt
Pierre-Sassoulas: [21f099a](https://github.com/PyCQA/pylint/pull/6623/commits/21f099a0195fa165040b20c437704ac733b8ff7b)

doc/Makefile
```diff
@@ -39,14 +39,17 @@ clean:
 	-rm -rf $(BUILDDIR)/* messages/convention messages/error messages/fatal messages/information \
 	  messages/refactor messages/warning
 
+install-dependencies:
+	@echo "Install dependencies"
+	pwd
```
Pierre-Sassoulas, line 44: I'm going to remove this debug log and rebase later.

doc/development_guide/contribute.rst
```diff
@@ -87,6 +87,11 @@ Tips for Getting Started with Pylint Development
 Building the documentation
 ----------------------------
 
-We use **tox** for building the documentation::
+You can use the makefile in the doc directory with ``make html`` to build the
+documentation. For repeated use directly using ``build-html`` will be faster::
```
DanielNoord, line 91:
```suggestion
documentation. To test smaller changes you can consider ``build-html``, which skips some checks but will be faster::
```

DanielNoord (approved):
One final nitpick. If this passes `RTD` this LGTM!

PR #6626, opened 2022-05-16
Refactors in preparation of ``pylint-config``

DanielNoord:

## Type of Changes

## Description

Preparation for https://github.com/PyCQA/pylint/pull/6578.

### DIFF
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -71,6 +71,8 @@ class _ArgumentsManager:
             usage=usage or "%(prog)s [options]",
             description=description,
             formatter_class=_HelpFormatter,
+            # Needed to let 'pylint-config' overwrite the -h command
+            conflict_handler="resolve",
         )
         """The command line argument parser."""
 
--- a/.pyenchant_pylint_custom_dict.txt
+++ b/.pyenchant_pylint_custom_dict.txt
@@ -186,6 +186,7 @@ misdesign
 misdesigns
 mixin
 modname
+monkeypatch
 mro
 # Used so much that we need the abbreviation
 msg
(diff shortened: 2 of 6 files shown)


DanielNoord:
@Pierre-Sassoulas The coverage drop is expected as I don't use the new `Action` yet in this PR. But due to the size of the code I thought I would separate it.

pylint/config/arguments_manager.py
```diff
@@ -71,6 +71,8 @@ def __init__(
             usage=usage or "%(prog)s [options]",
             description=description,
             formatter_class=_HelpFormatter,
+            # Needed to let 'pylint-config' overwrite the -h command
```
Pierre-Sassoulas, line 74: Shoudl'nt ``pylint-config`` be enitrely independant from pylint ? In my mind it would remove code and complexity from ``pylint``. This seem to be a little too "smart" and potentially hard to understand in 6 months.

DanielNoord: I don't think so. `pylint-config` heavily depends on `pylint`: we need access to message definitions to check for `bad-option-value`, we need access to plugins to load them and see if they are correct, we need access to options definitions to get the default value, etc.

You can see my implementation in https://github.com/PyCQA/pylint/pull/6578. Basically, we set up as if we are in a normal run of `pylint` to make sure we have access to everything we would normally have access to, but there are 2 changes: 1) before doing the configuration parsing we add some `pylint-config` specific options and subparsers (this removes the need to do `pylint run` but allows `pylint-config generate`) and 2) we shortcircuit into the `pylint-config` handler just before we start checking files.

Pierre-Sassoulas (approved):

PR #6628, opened 2022-05-16
Move the messages documentation to the user guide

Pierre-Sassoulas:
## Type of Changes

## Description

Move the messages documentation to user guide and do the proper redirection.

Small reviewable part of #6589 and follow-up to #6613 and #6622


doc/exts/pylint_messages.py
```diff
@@ -294,9 +302,7 @@ def _write_messages_list_page(
    :maxdepth: 1
    :titlesonly:
 
-{old_messages_string}
-
-"""
+{old_messages_string}"""
```
Pierre-Sassoulas, line 305: Required so there's no change to do with pre-commit after we generate the file.

DanielNoord:
This will break any existing link to the documentation of individual messages, see for example:
https://pylint.pycqa.org/en/latest/messages/fatal/astroid-error.html
vs.
https://pylint--6628.org.readthedocs.build/en/latest/messages/fatal/astroid-error.html

I don't think we should do this unless we find a way to fix the redirects for individual message pages.

doc/conf.py
```diff
@@ -49,6 +57,32 @@
     "sphinx_reredirects",
 ]
 
+MSG_TYPES_DOC = {k: v if v != "info" else "information" for k, v in MSG_TYPES.items()}
```
DanielNoord, line 60: Can't we use https://documatt.gitlab.io/sphinx-reredirects/usage.html#source-wildcards here? I only just found this in the documentation.

Pierre-Sassoulas: It does not work, we need a file to exists.
```python
def get_wildcard_messages_redirect() -> dict[str, str]:
    result = {}
    for category in MSG_TYPES_DOC.values():
        before = f"messages/{category}/*"
        result[before] = f"user_guide/messages/{category}/$source.html"
    return result
```
Result:
```
WARNING: No documents match to 'messages/information/*' redirect.
WARNING: No documents match to 'messages/convention/*' redirect.
WARNING: No documents match to 'messages/refactor/*' redirect.
WARNING: No documents match to 'messages/warning/*' redirect.
WARNING: No documents match to 'messages/error/*' redirect.
WARNING: No documents match to 'messages/fatal/*' redirect.
```

Pierre-Sassoulas: I'm trying to make the redirect by specifying everything works. If this does not work I'm thinking about:
- Having the file on the repository just to be able to redirect ?
- Not redirecting at all, after all the message have not been available for very long how many link would really be broken ?

DanielNoord: I'm trying to figure out RTD redirects. It "should" work. I'll give an update if I give up 😄

Pierre-Sassoulas: Using ``'messages/error/invalid-length-returned': 'user_guide/messages/error/invalid-length-returned.html'``, ``messages/error/invalid-length-returned.html`` is redirected to ``messages/error/user_guide/messages/error/invalid-length-returned.html`` it's unexpected. Any idea for a hack ? Maybe I should fork the lib and "fix" it ?

DanielNoord: Okay, I got the RTD redirects working.

See:
https://pylint--6628.org.readthedocs.build/en/6628/messages/fatal/astroid-error.html

Pierre-Sassoulas: Hmm good job, as the code did not change I suppose you used RTD interface? We might as well do everything in interface and remove the dependency to avoid confusion ?

DanielNoord: Yeah, you can check the `Redirects` tab to see what I did.

I'm opening an issue at RTD as we speak because the process is quite difficult and doesn't allow us all the freedom we want. I would suggest doing all manual redirects with the extension and doing the global redirects for directories through RTD (there are some limitations to it).

Would you mind cleaning up the PR and removing the unnecessary code from `conf.py`?

Pierre-Sassoulas: Sure thing ! I did not want to use the interface at all (because only maintainer have access) but whatever work 😄 Also only maintainer should be confident enough to redirect entire directories in the doc imo. And I already took an unreasonable amount of time to make this 0.0.1 lib work

DanielNoord: See https://github.com/readthedocs/readthedocs.org/issues/9240.

Yeah, I don't really understand why this is not easier. It seems like something more projects would have run into already.
Anyway, for now I think this solution works. Although we'll never be sure until this gets merged and we're live.

Pierre-Sassoulas: > Although we'll never be sure until this gets merged and we're live.

I'm pretty confident it will work 😄

doc/exts/pylint_messages.py
```diff
@@ -230,7 +230,9 @@ def _write_single_message_page(category_dir: Path, message: MessageData) -> None
             stream.write(
                 f"""
 .. note::
-  This message is emitted by an optional checker which requires the ``{message.checker_module_name}`` plugin to be loaded. See: :ref:`{message.checker_module_name}`.
+  This message is emitted by :ref:`the optional '{message.checker}' checker
+  <{message.checker_module_name}>` which requires the ``{message.checker_module_name}``
+  plugin to be loaded.
```
Pierre-Sassoulas, line 235: Hmm this is an unrelated change, I don't remember if I opened a PR for this. Maybe slightly related.

doc/conf.py
```diff
@@ -45,12 +45,21 @@
     "sphinx_reredirects",
 ]
 
+
+# Single file redirect are handled in this file and can be done by a pylint
```
DanielNoord, line 49:
```suggestion
# Single file redirects are handled in this file and can be done by a pylint
```

doc/conf.py
```diff
@@ -45,12 +45,21 @@
...
+# contributor. See:
 # https://documatt.gitlab.io/sphinx-reredirects/usage.html
+# Directory redirect are handled in ReadTheDoc admin interface and can only be done
```
DanielNoord, line 52:
```suggestion
# Directory redirects are handled in ReadTheDoc admin interface and can only be done
```

doc/conf.py
```diff
@@ -45,12 +45,21 @@
...
+# by pylint maintainer at the following URL:
```
DanielNoord, line 53:
```suggestion
# by pylint maintainers at the following URL:
```

doc/conf.py
```diff
@@ -45,12 +45,21 @@
...
+# https://readthedocs.org/dashboard/pylint/redirects/
 redirects: dict[str, str] = {
     # "<source>": "<target>"
     "intro": "index.html",
     "support": "contact.html",
     "user_guide/ide-integration": "installation.html",
+    "messages/messages_introduction": "user_guide/messages/index.html",
+    "messages/messages_list": "user_guide/messages/messages_overview.html",
+    "user_guide/message-control": "user_guide/messages/message-control.html",
```
DanielNoord, line 62: This seems like a good time to standardise either `word-word` or `word_word`.

doc/exts/pylint_messages.py
```diff
@@ -230,7 +230,9 @@ def _write_single_message_page(category_dir: Path, message: MessageData) -> None
             stream.write(
                 f"""
 .. note::
-  This message is emitted by an optional checker which requires the ``{message.checker_module_name}`` plugin to be loaded. See: :ref:`{message.checker_module_name}`.
+  This message is emitted by :ref:`the optional '{message.checker}' checker
+  <{message.checker_module_name}>` which requires the ``{message.checker_module_name}``
```
DanielNoord, line 234:
```suggestion
  This message is emitted by the optional :ref:`'{message.checker}'<{message.checker_module_name}>` checker which requires the ``{message.checker_module_name}``
```

doc/user_guide/index.rst
```diff
@@ -9,7 +9,7 @@ User Guide
    installation
    run
    output
-   message-control
+   messages/index
```
DanielNoord, line 12: Don't we want `messages` here so that we get all sub-headers as links as well?

DanielNoord (changes requested):
Some small nitpicks.

doc/user_guide/index.rst
Pierre-Sassoulas: Maybe something I don't know about sphinx but when using ``messages/`` here I get : ``/home/pierre/pylint/doc/user_guide/index.rst:5: WARNING: toctree contains reference to nonexisting document 'user_guide/messages'
`` during build ?

DanielNoord: Ah, the solution is different: add a hidden `toctree` to `messages/index.rst`. See difference between certain headers on:
https://pylint--6628.org.readthedocs.build/en/6628/user_guide/index.html

Pierre-Sassoulas: Unless I'm mistaken there's already a hidden toctree in message/index ?

Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/6628/files#diff-02baef687b1c6a68cf7f524437fced9e405fe591b8dff8e7736b038aecced7dcR9

DanielNoord: I checked, for the headers in which the sub-headers do show up the `toctree` is actually visible and not hidden.

I'm not sure we want to make the `toctree` for messages visible so I guess this will have to make do? Or is there a `overwrite-subhearder-visibility` option for `toctree`?

doc/conf.py
Pierre-Sassoulas: Let's use word_word, but let's do the normalization after we move the other one in the next MR.

doc/user_guide/index.rst
Pierre-Sassoulas: I'm not sure about what we want to do, maybe we can delay this to the last doc reorganization MR ?

doc/conf.py
DanielNoord: I'd prefer doing it now. You never know when somebody might copy a link to our documentation (perhaps this afternoon) and we would need to add another redirect for `message-control` > `message_control`.

Pierre-Sassoulas: Also it means the temporary link to ``transform_feature`` must be maintained for all eternity after being live during the time to review the next MR.

doc/conf.py
```diff
@@ -45,12 +45,21 @@
...
+# by pylint maintainers at the following URL:
+# https://readthedocs.org/dashboard/pylint/redirects/
 redirects: dict[str, str] = {
     # "<source>": "<target>"
     "intro": "index.html",
     "support": "contact.html",
     "user_guide/ide-integration": "installation.html",
+    "messages/messages_introduction": "user_guide/messages/index.html",
+    "messages/messages_list": "user_guide/messages/messages_overview.html",
+    "user_guide/message-control": "user_guide/messages/message_control.html",
```
Pierre-Sassoulas, line 62: @DanielNoord here the updated ``word-word`` to ``word_word`` for this MR. (The view in the comment is not up to date).

doc/conf.py
```diff
@@ -45,12 +45,21 @@
...
+# Directory redirects are handled in ReadTheDoc admin interface and can only be done
+# by pylint maintainers at the following URL:
+# https://readthedocs.org/dashboard/pylint/redirects/
 redirects: dict[str, str] = {
     # "<source>": "<target>"
     "intro": "index.html",
     "support": "contact.html",
     "user_guide/ide-integration": "installation.html",
+    "messages/messages_introduction": "user_guide/messages/index.html",
+    "messages/messages_list": "user_guide/messages/messages_overview.html",
```
DanielNoord, line 61:
```suggestion
```

https://pylint--6628.org.readthedocs.build/en/6628/messages/messages_introduction

Doesn't work 😢 

Let's handle all the `messages/*` redirects in `RTD`. I'll fix them after this gets merged!

DanielNoord (approved):
One final change... 😅

doc/conf.py
Pierre-Sassoulas: Ha right, RTD seems to preempt the single file redirect. Let me add a comment.

Pierre-Sassoulas: [350bf76](https://github.com/PyCQA/pylint/pull/6628/commits/350bf76ceea668af36c45a707e615aee43b9c126)

DanielNoord (approved):

doc/conf.py
Pierre-Sassoulas: Doesn't it mean we should do 
```
    "user_guide/messages/messages_introduction": "user_guide/messages/index.html",
    "user_guide/messages/messages_list": "user_guide/messages/messages_overview.html",
```    
?

DanielNoord: Lol, yeah that's probably it. Because the above link still doesn't work.

Pierre-Sassoulas: Waiting for the pipeline to confirm it works but I think it make more sense to keep RTD for full dir redirect and our conf for single link redirect even if it means having two redirections instead of one.

Pierre-Sassoulas: Hmm no, not working, even when clearing the redirect cache. Should I revert the latest commit ?

Pierre-Sassoulas: The link we used to test is wrong I think. ``https://pylint--6628.org.readthedocs.build/en/6628/messages/messages_introduction`` should be ``https://pylint--6628.org.readthedocs.build/en/6628/messages/messages_introduction.html``

DanielNoord:
The redirects for messages seem to work. However:
https://pylint.pycqa.org/en/latest/messages/messages_list.html doesn't.

https://pylint.pycqa.org/en/latest/messages/message_control.html does.

DanielNoord:
@Pierre-Sassoulas Will you also fix the other two missing redirects in a follow-up?

Pierre-Sassoulas:
I tried, I don't understand why it's not working anymore (I tested in the middle of the MR and did not check again) I'm wondering if we should do everything in RTD interface ? I'm under the impression that they could interfere with each other. Another solution is to just let the 404 be, we're investing a colossal amount of time in this, if the doc is good and searchable the problem of 404 will not be a real one.

PR #6636, opened 2022-05-17
[consider-using-in] Suggest set instead of tuple for membership

Pierre-Sassoulas:
## Type of Changes

## Description

Set are more performant for membership tests, we have another checks for this, we may as well suggest the right test from the start. Follow-up to #6635

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1267,6 +1267,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "consider-using-in",
             node=node,
             args=(common_variable, maybe_not, values_string),
+            confidence=HIGH,
         )
 
     def _check_chained_comparison(self, node):
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -321,7 +321,8 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "and increases readability compared to for-loop iteration.",
         ),
         "R1714": (
-            'Consider merging these comparisons with "in" to %r',
+            "Consider merging these comparisons with 'in' by using '%s %sin (%s)'."
+            " Use a set instead if elements are hashable.",
             "consider-using-in",
             "To check if a variable is equal to one of many values,"
             'combine the values into a tuple and check if the variable is contained "in" it '
@@ -1258,13 +1259,15 @@ class RefactoringChecker(checkers.BaseTokenChecker):
 
         # Gather information for the suggestion
         common_variable = sorted(list(common_variables))[0]
-        comprehension = "in" if node.op == "or" else "not in"
         values = list(collections.OrderedDict.fromkeys(values))
         values.remove(common_variable)
         values_string = ", ".join(values) if len(values) != 1 else values[0] + ","
-        suggestion = f"{common_variable} {comprehension} ({values_string})"
-
-        self.add_message("consider-using-in", node=node, args=(suggestion,))
+        maybe_not = "" if node.op == "or" else "not "
+        self.add_message(
+            "consider-using-in",
+            node=node,
+            args=(common_variable, maybe_not, values_string),
+        )
 
     def _check_chained_comparison(self, node):
         """Check if there is any chained comparison in the expression.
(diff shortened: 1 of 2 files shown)


cdce8p (changes requested):
If I recall correctly, we decided against it because sets only work if **all** items are hashable. Checking for that is quite complicated, so to be safe we just default to tuples.

Pierre-Sassoulas:
Wondering if we should not suggest set by default: ``Consider merging these comparisons with "in" by using "value in {1, 2}". Use a tuple instead if elements are not hashable.``

tests/functional/c/consider/consider_using_in.txt
```diff
@@ -1,14 +1,14 @@
...
-consider-using-in:21:0:21:36::"Consider merging these comparisons with ""in"" to 'value1 in (value2,)'":UNDEFINED
-consider-using-in:22:0:22:35::"Consider merging these comparisons with ""in"" to 'a_list in ([1, 2, 3], [])'":UNDEFINED
-consider-using-in:53:0:53:28::"Consider merging these comparisons with ""in"" to 'A.value in (1, 2)'":UNDEFINED
+consider-using-in:10:0:10:24::"Consider merging these comparisons with ""in"" by using ""value in (1, )"". Use a set instead if elements are hashable.""":HIGH
+consider-using-in:11:0:11:24::"Consider merging these comparisons with ""in"" by using ""value in (1, 2, )"". Use a set instead if elements are hashable.""":HIGH
+consider-using-in:12:0:12:36::"Consider merging these comparisons with ""in"" by using ""value in ('value', )"". Use a set instead if elements are hashable.""":HIGH
+consider-using-in:13:0:13:34::"Consider merging these comparisons with ""in"" by using ""value in (1, undef_value, )"". Use a set instead if elements are hashable.""":HIGH
+consider-using-in:14:0:14:38::"Consider merging these comparisons with ""in"" by using ""value in (1, 2, 3, )"". Use a set instead if elements are hashable.""":HIGH
```
cdce8p, line 5:
```suggestion
consider-using-in:14:0:14:38::"Consider merging these comparisons with ""in"" by using ""value in (1, 2, 3)"". Use a set instead if elements are hashable.""":HIGH
```
It doesn't really make sense to suggest a trailing comma here.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1426,7 +1432,8 @@ def visit_assign(self, node: nodes.Assign) -> None:
         "consider-using-ternary",
         "consider-swap-variables",
     )
-    def visit_return(self, node: nodes.Return) -> None:
+    def visit_return(self, node: nodes.Return | nodes.Assign) -> None:
+        """This function can be called from 'visit_assign'."""
```
cdce8p, line 1436: The typing would be correct. However, I'm not really a fan of calling `visit_return` from another `visit_*` method. Maybe it would make sense to create a new method for it and call that one instead. Something like `_check_simplify_expression`? Anyway this should probably be done separately.

cdce8p (changes requested):
> Wondering if we should not suggest set by default: `Consider merging these comparisons with "in" by using "value in {1, 2}". Use a tuple instead if elements are not hashable.`

Suggesting tuples is the safer option. So I would stick with that. The note helpful on its own
`"Use a set instead if elements are hashable."`

pylint/checkers/refactoring/refactoring_checker.py
Pierre-Sassoulas: I agree, it's better to have a subfunction called from both visit functions and proper unsurprising typing on the visit function. There are other instance of that, in the codebase I could open another merge request, searching for ``| nodes.`` should work I guess.

Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/6892

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1446,7 +1450,8 @@ def visit_assign(self, node: nodes.Assign) -> None:
...
```
cdce8p, line 1454: Please move that to #6892

cdce8p (changes requested):
Looks good. One last comment

cdce8p (approved):

DanielNoord:
@Pierre-Sassoulas Can this be merged?

Pierre-Sassoulas:
This can be merged but I want to test the new primer comments on it first.

Pierre-Sassoulas:
I'm going to merge because it's ready and we can always test on another issue of the "better error messages" project.

mbyrnepr2:
The pytest one looks like a false positive? https://github.com/pytest-dev/pytest/blob/main/src/_pytest/junitxml.py#L623

PR #6641, opened 2022-05-18
Burst additional command in two files 

Pierre-Sassoulas:
## Type of Changes

## Description

Move the messages documentation to user guide and do the proper redirection.

Small reviewable part of #6589 and follow-up to #6613, #6622 and #6628


DanielNoord:
See https://github.com/PyCQA/pylint/pull/6589#issuecomment-1129746617.

I am -0.5 on this change. I think the `User guide` (which we might want to rename) should focus on `pylint` whereas these are different commands and probably deserve their own section in the index.

Pierre-Sassoulas:
> these are different commands and probably deserve their own section in the index.

You're right, I updated the MR accordingly.

> User guide (which we might want to rename)

What do you have in mind ?

DanielNoord:
What about `Using pylint`? Or `Usage and configuration`?  `black` uses that under their `User Guide`, see https://black.readthedocs.io/en/stable/. I like that term as I think it covers the topics better than "Guide".

DanielNoord:
> Edit: answered too fast, I'm ok with merging pyreverse/symilar in a usage and configuration section.

I would do:
```ini
Tutorial
Usage and configuration
Development
Additional included programs
Changelogs
Contact
```

Although I'm not 100% on `Additional included programs`. For me this is the clearest separation of the various topics we have.

Pierre-Sassoulas:
Regarding Changelogs my idea is to separate changelog for dev and changelog for users so in order to do that it would make sense to have two changelogs, one in ``Usage and configuration`` and one in ``Development``.

DanielNoord:
> Regarding Changelogs my idea is to separate changelog for dev and changelog for users so in order to do that it would make sense to have two changelogs, one in `Usage and configuration` and one in `Development`.

Can't we do two sections in the same changelog? Why would we force developers to look in two places? We could just make an `internal changes` section. There might also be some grey-area changes.

DanielNoord:
> > Why would we force developers to look in two places?
> 
> The plugin developers are maybe 0.5% of total users, and they're power users very knowledgable in pylint. We could separate the two changelog so the vast majority of user do not have to triage what they should read in the changelog.

Yeah, but still. I often look back at older changelogs to see when something changed and I tend to look at the changelogs of other projects as well. That can often give you inspiration about relevant changes.
Why not do:
```
Breaking changes or removed functionality: (if any)
- ...
- ...

New checkers or functionality:
- ...
- ...

Bug fixes:
- ...
- ...
- ...

Internal changes:
- ...
- ...
```

If the internal changes are separated from the user-facing changes in a separate section I really don't think any user will mind.

DanielNoord:
Just from the `2.14` changelog alone I can get these examples:
https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/ChangeLog#L287
https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/ChangeLog#L278
https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/ChangeLog#L166-L170
https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/ChangeLog#L120-L123
https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/ChangeLog#L40

I'm nog suggesting to move those around, but just to illustrate that we already struggle with deciding where to put an entry. I just think that many changes can't be easily categorised as either being `Developer` and `User` and I don't think it makes sense to separate those changelogs. I don't know of any other projects that splits its changelogs like that. 
If we're going through the effort of moving stuff aroundwhy not simplify?
_I am aware that most of those PRs are made by myself, so perhaps I am the only one who struggles to identify the difference, but I think you understand what I mean 😄_

There are multiple solutions to the problem of merging the current two locations. We could even do something where we keep the `whatsnew` and move those to `Changelogs`. Start adapting a new structure from now on and provide a link at the top of the page to the old format changelogs. For the old version that wouldn't change the status quo: 2 places to look. But for now versions there would be one place with clearly separated sections. That also has the benefit of making it easier to point contributors where they should put the entry instead of asking them to do it in two places.
If at some point anybody gets bored they can write a nice script that would merge the two older versions 😄

Pierre-Sassoulas:
> we already struggle with deciding where to put an entry. 

Yeah, right now we're duplicating everything everywhere 😄 Actually if you check the whatsnew for 2.5 before hippo91 and myself became release manager what is in whatsnew is really a small summary of changelog. We just had a lot of other things to take care of at the time and inadvertently created the habit of duplicating the changelog in whatsnew all the time.

That's why we need to define what we want to do now (before 2.15 so the 2.15 changelog is the new example to follow, and can be enforced by CI checks + label chosen by maintainer like in black). I think we should not document simple refactor or small internal change. I.e. for the label documentation and maintenance we do not add any changelog, at all.  For the API changes and deprecation should be in the dev facing changelog. For the dependency label we probably need to separate dev dependency from pylint's actual dependency  that will affect installation alongside other libs, and dev dependencies from pre-commit dependency.

I think that the user facing changelog should only be **changes that will modify something that will affect you when you install pylint or ``launch`` pylint with ``pylint [options] or use pylint integrated in an IDE **.  So with that definition in mind:

https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/doc/whatsnew/2.14.rst#L377

Dev only.

https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/doc/whatsnew/2.14.rst#L204

The only one where it's ambiguous imo. But there was no way to use this priority in any pylint, options, dev facing too.

https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/doc/whatsnew/2.14.rst#L216

Dev only.

https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/doc/whatsnew/2.14.rst#L329

Dev only.

https://github.com/PyCQA/pylint/blob/a6ae75a62020aba78eccd2988cc0b1b02f9aee05/doc/whatsnew/2.14.rst#L289

Dev only.

This is probably something we should have discussed in #5728 before I create 5 MRs 😄 The need to separate user guide and dev guide - while it make sense in itself, because right now everything is mixed up and impossible to follow- stem from the fact that I wanted to not **have to** merge changelog and whatsnew for older changelogs but I still wanted a clear rules for new changelog in the future.

DanielNoord:
> The need to separate user guide and dev guide - while it make sense in itself, because right now everything is mixed up and impossible to follow- stem from the fact that I wanted to not **have to** merge changelog and whatsnew for older changelogs but I still wanted a clear rules for new changelog in the future.

See my proposal in my last comment:

I don't think we should do something just because we have to deal with a legacy format of the changelog. Let's start writing the changelog in our preferred format and worry about how to store the old changelogs afterwards. I wouldn't mind having a sentence: "Pylint's changelog format changed after 2.x. Therefore, for older versions there are two related but different changelogs which can be found here > link to changelog archive". Or something like that...
I don't think it will be an issue if there is less consistency between newer and older changelogs. As a matter of fact, we already have that.

Pierre-Sassoulas:
> I don't think we should do something just because we have to deal with a legacy format of the changelog

Reorganizing the doc is a good thing in itself too. and it's better to have a clean doc for 2.14 because it's the reference in read the doc (but we could cherry-pick it on maintenance/2.14 too when the patch become the reference and only do the changelog change during the beta). 

But let's just apply https://github.com/PyCQA/pylint/pull/6641#issuecomment-1130147692 for this MR first :)

doc/conf.py
```diff
@@ -62,6 +62,7 @@
     "support": "contact.html",
     "user_guide/ide-integration": "installation.html",
     "user_guide/message-control": "user_guide/messages/message_control.html",
+    "additional_commands/index": "index.html",
```
DanielNoord, line 65: This redirect doesn't work 😓

DanielNoord (changes requested):

doc/conf.py
Pierre-Sassoulas: I don't understand why redirecting in sphinx is so freaking hard 😄

Pierre-Sassoulas: https://github.com/sphinx-contrib/redirects seems serious and 'simple' how about that ?

DanielNoord: Yeah that's the one I was originally looking at originally. It's not "official" but the organization gives me some hope.
It doesn't support pattern matching, which is something we'll need to keep using `RTD` for I guess. (You might want to subscribe to https://github.com/readthedocs/readthedocs.org/issues/4221 which makes it so contributors can also help make redirects)

On the other hand, the maintainer of `reredirects` is quite responsive (I have actually gotten two maintenance PRs merged already). So if we can specify what is currently missing we can also ask them to fix it?

DanielNoord: For future reference, we're trying to get https://pylint--6641.org.readthedocs.build/en/6641/additional_commands/index.html to work.

DanielNoord:
```suggestion
    "additional_commands/index": "../index.html",
```

This works! 🎉

Pierre-Sassoulas: Link for test: https://pylint--6641.org.readthedocs.build/en/6641/additional_commands/index.html (right now it's an infinite redirect)

DanielNoord: It is also good to know why. Apparently these redirects are relative. So it kept redirecting from `additional_commands/index.html` to ... `additional_commands/index.html` 😅

Pierre-Sassoulas: Yeah this is actually the bug that made it click for me.

doc/index.rst
```diff
@@ -37,6 +37,12 @@ it's best to start with the ``--errors-only`` flag, then disable
 convention and refactor message with ``--disable=C,R`` and progressively
 re-evaluate and re-enable messages as your priorities evolve.
 
+
+Pylint ships with two additional tools:
+
+- :ref:`pyreverse <pyreverse>` (standalone that generates package and class diagrams.)
```
DanielNoord, line 43:
```suggestion
- :ref:`pyreverse <pyreverse>` (standalone tool that generates package and class diagrams.)
```

doc/index.rst
```diff
@@ -37,6 +37,12 @@ it's best to start with the ``--errors-only`` flag, then disable
...
+- :ref:`symilar <symilar>`  (duplicate code finder also integrated in pylint)
```
DanielNoord, line 44:
```suggestion
- :ref:`symilar <symilar>`  (duplicate code finder that is also integrated in pylint)
```

DanielNoord (approved):

Pierre-Sassoulas:
Can't merge yet I did not apply https://github.com/PyCQA/pylint/pull/6641#issuecomment-1130147692 yet

PR #6642, opened 2022-05-18
[doc] Fix the link to block disable in useless-import-alias related

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Apparently the redirect in #6628 is not perfect.

PR #6647, opened 2022-05-19
Change names of documentation files

### LINKED ISSUE #6646: This page does not exist yet.
### Question

https://pylint.pycqa.org/en/latest/user_guide/user_guide/messages/message_control.html
https://pylint.pycqa.org/en/latest/user_guide/messages/messages_list.html

This page does not exist yet. 

hope page exitst

### Additional context

_No response_

DanielNoord:

## Type of Changes

## Description

Follow-up to https://github.com/PyCQA/pylint/pull/6628#issuecomment-1129999997 and closes #6646.


DanielNoord:
To test, these links should work now:
https://pylint--6647.org.readthedocs.build/en/6647/messages/messages_list.html
https://pylint--6647.org.readthedocs.build/en/6647/messages/messages_introduction.html

DanielNoord:
@Pierre-Sassoulas Failure is due to `block-disables`. See the comment above.

Rather than trying to fiddle around with redirects on three different levels, let's just keep the old names. Sphinx takes the top header as file name anyway so that's not an issue. And using `messages_introduction` instead of `index` isn't really an issue imo. Both indicate that it is the top-level page of the `messages` section.

See also the issue, which is my rationale for fixing this rather than keeping as is.

Pierre-Sassoulas:
Make sense. Regarding the block disable, should we merge https://github.com/PyCQA/pylint/pull/6642 ? I don't see why the check fails (it's very hard to see where there was an error in the output).

PR #6652, opened 2022-05-19
Emit `modified-iterating-{list|dict|set}` when iterating literals or using `del`

### LINKED ISSUE #6648: False negative for ``modified-iterating-list`` when using the ``del`` keyword
### Bug description

```python
letters=['a','b','c','d','e','f','g','h','i','j','k','l']
for i in letters:
    del i
print(letters)  # Prints ['b', 'd', 'f', 'h', 'j', 'l']
```

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
Nothing.
```

### Expected behavior

``modified-iterating-list`` raised. See https://github.com/PyCQA/pylint/issues/2471#issuecomment-1130962298

### Pylint version

```shell
2.14.0
```

jacobtylerwalls:
## Type of Changes

## Description
Closes #6648
Solves an additional false negative: the check wasn't emitted when iterating literals.

### DIFF
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -31,6 +31,11 @@ False positives fixed
 False negatives fixed
 =====================
 
+* Emit ``modified-iterating-list`` and analogous messages for dicts and sets when iterating
+  literals, or when using the ``del`` keyword.
+
+  Closes #6648
+
 
 Other bug fixes
 ===============
--- a/pylint/checkers/modified_iterating_checker.py
+++ b/pylint/checkers/modified_iterating_checker.py
@@ -56,9 +56,8 @@ class ModifiedIterationChecker(checkers.BaseChecker):
     )
     def visit_for(self, node: nodes.For) -> None:
         iter_obj = node.iter
-        if isinstance(iter_obj, nodes.Name):
-            for body_node in node.body:
-                self._modified_iterating_check_on_node_and_children(body_node, iter_obj)
+        for body_node in node.body:
+            self._modified_iterating_check_on_node_and_children(body_node, iter_obj)
 
     def _modified_iterating_check_on_node_and_children(
         self, body_node: nodes.NodeNG, iter_obj: nodes.NodeNG
@@ -72,7 +71,19 @@ class ModifiedIterationChecker(checkers.BaseChecker):
         self, node: nodes.NodeNG, iter_obj: nodes.NodeNG
     ) -> None:
         msg_id = None
-        if self._modified_iterating_list_cond(node, iter_obj):
+        if isinstance(node, nodes.Delete) and any(
+            self._deleted_iteration_target_cond(t, iter_obj) for t in node.targets
+        ):
+            inferred = utils.safe_infer(iter_obj)
+            if isinstance(inferred, nodes.List):
+                msg_id = "modified-iterating-list"
+            elif isinstance(inferred, nodes.Dict):
+                msg_id = "modified-iterating-dict"
+            elif isinstance(inferred, nodes.Set):
+                msg_id = "modified-iterating-set"
+        elif not isinstance(iter_obj, nodes.Name):
+            pass
+        elif self._modified_iterating_list_cond(node, iter_obj):
             msg_id = "modified-iterating-list"
         elif self._modified_iterating_dict_cond(node, iter_obj):
             msg_id = "modified-iterating-dict"
@@ -148,6 +159,22 @@ class ModifiedIterationChecker(checkers.BaseChecker):
         return (
             self._common_cond_list_set(node, iter_obj, infer_val)
             and node.value.func.attrname in _SET_MODIFIER_METHODS
+        )
+
+    def _deleted_iteration_target_cond(
+        self, node: nodes.DelName, iter_obj: nodes.NodeNG
+    ) -> bool:
+        if not isinstance(node, nodes.DelName):
+            return False
+        if not isinstance(iter_obj.parent, nodes.For):
+            return False
+        if not isinstance(
+            iter_obj.parent.target, (nodes.AssignName, nodes.BaseContainer)
+        ):
+            return False
+        return any(
+            t == node.name
+            for t in utils.find_assigned_names_recursive(iter_obj.parent.target)
         )
 
 
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -12,7 +12,7 @@ import numbers
 import re
 import string
 import warnings
-from collections.abc import Iterable
+from collections.abc import Iterable, Iterator
 from functools import lru_cache, partial
 from re import Match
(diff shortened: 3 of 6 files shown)


pylint/checkers/modified_iterating_checker.py
```diff
@@ -72,7 +72,15 @@ def _modified_iterating_check(
         self, node: nodes.NodeNG, iter_obj: nodes.NodeNG
     ) -> None:
         msg_id = None
-        if self._modified_iterating_list_cond(node, iter_obj):
+        if self._deleted_iteration_target_cond(node, iter_obj):
```
DanielNoord, line 75: Not sure if this should be done in the same PR, but we're calling `safe_infer` quite a lot here. We probably hit the cache on every call but just to be safe:
shouldn't we just do `safe_infer` once before all the `ifs` and `elif` in between L75-87 and call methods accordingly?

jacobtylerwalls: Good Q. I think the case where `safe_infer` is called multiple times is not tremendously realistic--just tuples and custom iterables that don't subclass list/dict/set, right? I should add a test case that relates to this PR, but the usefulness of a refactor is probably a separate PR.

jacobtylerwalls: Never mind, I follow you now. Yeah that should be refactored in another PR. 👍🏻

ChangeLog
```diff
@@ -3,6 +3,19 @@ Pylint's ChangeLog
...
 
+What's New in Pylint 2.15.0?
+============================
+Release date: TBA
+
+..
+  Put new features here and also in 'doc/whatsnew/2.15.rst'
+
+
+* Emit ``modified-iterating-list`` and analogous messages for dicts and sets when using the ``del`` keyword.
```
DanielNoord, line 14: I'd be fine with putting this in `2.14`. Doesn't seem like likely to regress. Knock on wood...

pylint/checkers/utils.py
```diff
@@ -1778,3 +1778,17 @@ def in_for_else_branch(parent: nodes.NodeNG, stmt: nodes.Statement) -> bool:
     return isinstance(parent, nodes.For) and any(
         else_stmt.parent_of(stmt) or else_stmt == stmt for else_stmt in parent.orelse
     )
+
+
+def find_assigned_names_recursive(
```
DanielNoord, line 1783: It feels a bit unpythonic to update in place, but that might be me. Looking at the other use-case with a small refactor using `+=` we could just make this return a list. But I'll leave that choice to you.

DanielNoord: We could even let this `yield` so that we exit early in `_deleted_iteration_target_cond`

ChangeLog
jacobtylerwalls: I guess I'd prefer not to add things to something we pre-released. And it's not ready anyway, since there's a failing test to look into.

ChangeLog
```diff
@@ -3,6 +3,19 @@ Pylint's ChangeLog
 ------------------
 
 
+What's New in Pylint 2.15.0?
```
Pierre-Sassoulas, line 6: There's going to be some change to the changelog/whatsnew in 2.15, I'm trying to have only one changelog entry to add in #5728.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -150,6 +158,23 @@ def _modified_iterating_set_cond(
...
+            and isinstance(
+                iter_obj.parent.target, (nodes.AssignName, nodes.BaseContainer)
+            )
+        ):
+            targeted_names: list[str] = []
+            utils.find_assigned_names_recursive(iter_obj.parent.target, targeted_names)
+            for target_name in targeted_names:
+                if target_name == node.name:
+                    return True
+        return False
```
Pierre-Sassoulas, line 176:
```suggestion
        if not (
            isinstance(node, nodes.DelName)
            and isinstance(iter_obj.parent, nodes.For)
            and isinstance(
                iter_obj.parent.target, (nodes.AssignName, nodes.BaseContainer)
            )
        ):
            return False
        targeted_names: list[str] = []
        utils.find_assigned_names_recursive(iter_obj.parent.target, targeted_names)
        return all(t==node.name for t in targeted_names)
```

pylint/checkers/modified_iterating_checker.py
jacobtylerwalls: Ah, hang on. Just took a look -- it's different items being inferred. Only the list and the set branches (in the existing code) share a `safe_infer` call, and nothing is shared with the call in this PR. I'm not sure it's worth optimizing further -- we might be adding calls to `safe_infer` rather than minimizing them.

DanielNoord (approved):

doc/whatsnew/2/2.15/full.rst
```diff
@@ -0,0 +1,12 @@
+Full changelog
+==============
+
+What's New in Pylint 2.14.0?
```
DanielNoord, line 4:
```suggestion
What's New in Pylint 2.15.0?
```

doc/whatsnew/2/2.15/index.rst
```diff
@@ -0,0 +1,12 @@
+***************************
+ What's New in Pylint 2.14
```
DanielNoord, line 2:
```suggestion
 What's New in Pylint 2.15
```

doc/whatsnew/2/2.15/summary.rst
```diff
@@ -0,0 +1,26 @@
+:Release: 2.14
```
DanielNoord, line 1:
```suggestion
:Release: 2.15
```

DanielNoord:
@jacobtylerwalls Sorry I messed up the merge of main. Got a little overexcited to see if the primer actually works 😄

DanielNoord:
> 🤖 According to the primer, this change has **no effect** on the checked open source code. 🤖🎉

🎉 

Would have been nice to actually find a change 😄 

_Github not showing the comment because a similar comment has been hidden is annoying though. I wonder if this is a common issue._

jacobtylerwalls:
I expect more with #6677

doc/whatsnew/2/2.15/summary.rst
Pierre-Sassoulas: Regarding the whatsnew for 2.15 I was thinking of discussing the new format here : https://github.com/PyCQA/pylint/pull/6688/files

Let me know what you think :)

Pierre-Sassoulas:
Regarding the whatsnew for 2.15 I was thinking of discussing the new format here : https://github.com/PyCQA/pylint/pull/6688/files (i.e. a single file, enforced in CI according tu label)

Let me know what you think :)

DanielNoord (approved):

Pierre-Sassoulas (approved):
Great !

PR #6653, opened 2022-05-19
Retry user input if it is not correct

DanielNoord:

## Type of Changes

## Description

Blocked by https://github.com/PyCQA/pylint/pull/6634.

### DIFF
--- a/pylint/config/_pylint_config/utils.py
+++ b/pylint/config/_pylint_config/utils.py
@@ -4,8 +4,10 @@
 
 """Utils for the 'pylint-config' command."""
 
+from __future__ import annotations
 
 import sys
+from collections.abc import Callable
 
 if sys.version_info >= (3, 8):
     from typing import Literal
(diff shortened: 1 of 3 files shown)


pylint/config/_pylint_config/utils.py
```diff
@@ -16,6 +19,39 @@
...
+        called_once = False
+        while True:
+            try:
+                func()
+            except InvalidUserInput as exc:
+                if called_once and exc.input == "exit()":
+                    print("Stopping 'pylint-config'.")
+                    sys.exit()
+                print(exc.message)
+                print("Type 'exit()' if you want to exit the program.")
```
Pierre-Sassoulas, line 46: I think the user can exit with ctrl + C maybe we don't have to handle this ?

DanielNoord: I didn't want to think of the various keybindings windows, Mac and Linux use. And I thought offering a clean exit instead of printing a stracktrace for `KeyboardInterrupt` would be nice.

I'm fine with removing it!

Pierre-Sassoulas (approved):
Changed the milestone to 2.14, I'll let you confirm that we can merge already but considering the decision on the previous MR I think it's consensual.

PR #6658, opened 2022-05-21
Burst run's documentation between user guide and dev guide, merge FAQ's run questions in the user guide

Pierre-Sassoulas:
## Type of Changes

## Description

Reviewable cleanup before #6589


doc/development_guide/run.rst
```diff
@@ -0,0 +1,83 @@
+===========================================
```
DanielNoord, line 1: Let's call this file `run_as_package` or something like that. `run.html` feels like we might want to use it for some more general pages in the future.

doc/development_guide/run.rst
```diff
@@ -0,0 +1,83 @@
...
+ Running Pylint from another python program
+===========================================
+
+You can call Pylint from another Python program thanks to the ``Run()``
+function in the ``pylint.lint`` module
```
DanielNoord, line 6:
```suggestion
class in the ``pylint.lint`` module
```

doc/development_guide/run.rst
```diff
@@ -0,0 +1,83 @@
...
+  pylint.run_pylint()
+  # Or:
+  pylint.run_pylint(argv=["your_file"])
+
+To silently run Pylint on a ``module_name.py`` module,
+and get its standard output and error:
+
+.. sourcecode:: python
+
+  from pylint import epylint as lint
```
DanielNoord, line 32: This example should be updated as `epylint` is not the same as `pylint`. We shouldn't suggest them interchangeably.

doc/development_guide/run.rst
```diff
@@ -0,0 +1,83 @@
...
+
+  (pylint_stdout, pylint_stderr) = lint.py_run('module_name.py', return_std=True)
+
+It is also possible to include additional Pylint options in the first argument to ``py_run``:
+
+.. sourcecode:: python
+
+  from pylint import epylint as lint
```
DanielNoord, line 40: Same here.

DanielNoord (changes requested):
I like how `user_guide/run.rst` got a lot simpler. Just some final thoughts!

Pierre-Sassoulas:
I made a somewhat big change by renaming "run" to "API" for the dev guide.

doc/development_guide/api/epylint.rst
```diff
@@ -0,0 +1,22 @@
+=======
+epylint
```
DanielNoord, line 2: This whole section of `API` should probably go into `User guide`? It's more usage related than development related right?

doc/development_guide/api/index.rst
```diff
@@ -0,0 +1,10 @@
+###
+API
+###
+
```
DanielNoord, line 4: We should probably write something here. https://pylint--6658.org.readthedocs.build/en/6658/development_guide/api/index.html feels a bit empty.

doc/development_guide/api/epylint.rst
Pierre-Sassoulas: For me everything that is not doable by using the pylint executable with options is development. (Beside creating a file to call epylint's API is development ?).

doc/development_guide/api/pylint.rst
```diff
@@ -0,0 +1,76 @@
...
+    run_pylint()
+
+
+Or supply arguments yourself:
+
+.. sourcecode:: python
+
+    from pylint import run_pylint
+
+    run_pylint(argv=["--disable=line-too-long", "myfile.py"])
```
Pierre-Sassoulas, line 25: Considering:
```suggestion
As you would launch the command line
------------------------------------

You can use the ``run_pylint`` function, which is the same function
called by the command line.

.. sourcecode:: python

    from pylint import run_pylint

    run_pylint(argv=["--disable=line-too-long", "myfile.py"])
```

> There should be one obvious way to do it

doc/development_guide/api/epylint.rst
DanielNoord: But `Development` is about developing `pylint` right? Since `pylint` is a developer's tool everything related to it will be "development", but I feel like this is more part of "using pylint" than "developing or contributing to pylint". We never use these APIs in our own development and you don't really need them when writing a checker.

doc/development_guide/api/pylint.rst
DanielNoord: Yeah I like the suggestion better although I would keep `Or supply arguments yourself:` at the end.

doc/development_guide/api/epylint.rst
Pierre-Sassoulas: Right, it's still unclear what is what. How about this:
- User guides : using pylint's result from the command line
- Developer guides: using pylint's API programmatically, understanding pylint plugin system, extension, checkers, etc.
- Contributor guides : developing for pylint itself launching tests, process, etc. (also maintaining pylint : we have a release.md document that is not included in the doc yet).

DanielNoord: I like that! Although I would perhaps put Developers and Contributing under a more general header of "Development" like black does. That would also allow showing some of the sub-headers of Usage in the sidebar.

Pierre-Sassoulas: I think we're almost ready to do the big table of content revamp then 😄 .

Pierre-Sassoulas: I'm planning on adding readme and release.md first.

doc/development_guide/api/pylint.rst
Pierre-Sassoulas: [e87b2ce](https://github.com/PyCQA/pylint/pull/6658/commits/e87b2ce0deab9adaea0fd493c03e0a4d5a327774) ? Not sure if it is what you meant :)

DanielNoord (approved):

doc/development_guide/api/epylint.rst
DanielNoord: I'll let you choose, but I'd prefer a PR that sets up the TOC first. It's quite hard to review content + redirects + new pages at the same time.
If we create one PR with all pages/sections we want to have and make sure redirects work we can then do the content afterwards. I would even be okay with adding some blank pages for now considering that you're working on this quite hard at the moment.

PR #6662, opened 2022-05-22
Allow specifying an output file when generating a configuration file

DanielNoord:

## Type of Changes

## Description

Blocked by https://github.com/PyCQA/pylint/pull/6661.

Adds a sequence of specifying an output file to the ``generate`` command. Also handles overwriting with a default to not overwrite.

### DIFF
--- a/pylint/config/_pylint_config/utils.py
+++ b/pylint/config/_pylint_config/utils.py
@@ -87,9 +87,7 @@ def validate_yes_no(question: str, default: Literal["yes", "no"] | None) -> bool
     if answer not in YES_NO_ANSWERS:
         raise InvalidUserInput(", ".join(sorted(YES_NO_ANSWERS)), answer)
 
-    if answer.startswith("y"):
-        return True
-    return False
+    return answer.startswith("y"):
 
 
 def get_and_validate_output_file() -> tuple[bool, Path]:
--- a/pylint/config/_pylint_config/utils.py
+++ b/pylint/config/_pylint_config/utils.py
@@ -87,7 +87,7 @@ def validate_yes_no(question: str, default: Literal["yes", "no"] | None) -> bool
     if answer not in YES_NO_ANSWERS:
         raise InvalidUserInput(", ".join(sorted(YES_NO_ANSWERS)), answer)
 
-    return answer.startswith("y"):
+    return answer.startswith("y")
 
 
 def get_and_validate_output_file() -> tuple[bool, Path]:
--- a/pylint/config/_pylint_config/utils.py
+++ b/pylint/config/_pylint_config/utils.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Callable
+from pathlib import Path
 
 if sys.version_info >= (3, 8):
     from typing import Literal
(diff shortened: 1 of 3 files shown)


pylint/config/_pylint_config/utils.py
```diff
@@ -66,3 +70,44 @@ def get_and_validate_format() -> Literal["toml", "ini"]:
...
+
+    if answer == "" and default:
+        answer = default
+
+    if answer not in YES_NO_ANSWERS:
+        raise InvalidUserInput(", ".join(sorted(YES_NO_ANSWERS)), answer)
+
+    if answer.startswith("y"):
+        return True
+    return False
```
Pierre-Sassoulas, line 92:
```suggestion
    return answer.startswith("y"):
```

tests/config/pylint_config/test_pylint_config_generate.py
```diff
@@ -33,6 +40,12 @@ def test_format_of_output(
     monkeypatch: MonkeyPatch, capsys: CaptureFixture[str]
 ) -> None:
     """Check that we output the correct format."""
+    # Monkeypatch everything we don't want to check in this test
+    monkeypatch.setattr(
+        "pylint.config._pylint_config.utils.get_and_validate_output_file",
+        lambda: (False, Path()),
+    )
```
Pierre-Sassoulas, line 47: Shouldn't we separate functions so they are easily testable ? I.e. we provide the result of ``get_and_validate_output_file`` to the function.

Pierre-Sassoulas (approved):
LGTM, apart from two nitpicks :)

tests/config/pylint_config/test_pylint_config_generate.py
DanielNoord: I'm not sure I understand you. What functions do you want to separate?

pylint/config/_pylint_config/utils.py
```diff
@@ -66,3 +70,42 @@ def get_and_validate_format() -> Literal["toml", "ini"]:
...
+    # pylint: disable-next=bad-builtin
+    answer = input(question).lower()
+
+    if answer == "" and default:
+        answer = default
+
+    if answer not in YES_NO_ANSWERS:
+        raise InvalidUserInput(", ".join(sorted(YES_NO_ANSWERS)), answer)
+
+    return answer.startswith("y"):
```
DanielNoord, line 90:
```suggestion
    return answer.startswith("y")
```

tests/config/pylint_config/test_pylint_config_generate.py
Pierre-Sassoulas: ``generate_interactive_config`` and ``get_and_validate_output_file `` so the former take the result of the latter as argument and it's easier to test it without mocking user outputs.

DanielNoord: Those are currently separated right? `generate_interactive_config` is the handler for the general `generate --interactive` command, while each subsequent question or set of questions has its own `util` function. So for every set of questions we can simply mock the function and continue.

We could test those `utils` directly instead of using `Run()`? Is that what you mean? I chose not to do this as this makes it easier to test combinations of response: just remove the mock and add the desired answers to `answers`.

Pierre-Sassoulas: > We could test those utils directly instead of using Run()? Is that what you mean?

Yes :) But never-mind if it was voluntary !

PR #6664, opened 2022-05-22
Fix ``unsupported-binary-operation`` on classes that overload or

### LINKED ISSUE #4951: DjangoRestFramework: E1131: unsupported operand type(s) for | (unsupported-binary-operation)
### Bug description

```python

from rest_framework.permissions import BasePermission, IsAuthenticated, SAFE_METHODS

class ReadOnly(BasePermission):
    def has_permission(self, request, view):
        return request.method in SAFE_METHODS

permission_classes = [IsAuthenticated|ReadOnly]
```

### Configuration

_No response_

### Command used

```shell
pylint binary.py
```

### Pylint output

```shell
************* Module binary
binary.py:9:22: E1131: unsupported operand type(s) for | (unsupported-binary-operation)

Your code has been rated at 0.00/10
```

### Expected behavior

For pylint to not throw this error. 

Django Rest Framework supports binary operands for permissions [1]
This example code is taken from the documentation [2]
[1] https://github.com/encode/django-rest-framework/blob/master/rest_framework/permissions.py
[2] https://www.django-rest-framework.org/api-guide/permissions/

### Pylint version

```shell
pylint 2.10.2
astroid 2.7.3
Python 3.9.4 (default, Apr 10 2021,
```

### OS / Environment

Debian GNU/Linux 10 (buster)

### Additional dependencies

djangorestframework==3.12.4

*edit*: Added link [2] to django rest frameworks documentation on permissions

timmartin:
## Type of Changes

## Description

We have a specific checker looking for people using `|` with two types, apparently on the assumption that anyone doing that is mistakenly trying to use union syntax where it's not supported. This seems to overlook the fact that it is possible for class definitions to use a metaclass that implements `__or__` or `__ror__` as appropriate, which is the case in the permission classes for Django REST Framework, for example.

Closes #4951


pylint/checkers/typecheck.py
```diff
@@ -1900,13 +1900,34 @@ def _detect_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> Non
...
+        right_obj = astroid.helpers.object_type(node.right)
+        left_is_type = False
+        right_is_type = False
+
+        if isinstance(left_obj, nodes.ClassDef) and is_classdef_type(left_obj):
+            try:
+                left_obj.getattr("__or__")
+                return
+            except astroid.NotFoundError:
+                left_is_type = True
```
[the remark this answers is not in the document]
timmartin: If you mean Python built-in `hasattr` then I don't think so, because we need to look up the attribute in Astroid's inferred knowledge of the real class, not in the `left_obj` runtime object. If there's a Pylint equivalent to `hasattr` on inferred representations then we could use that, but I couldn't find one.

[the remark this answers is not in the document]
Pierre-Sassoulas: Do you think it's worth it to add hasattr in astroid ?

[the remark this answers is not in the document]
timmartin: Sure, I'd be happy to give it a go.

timmartin:
I'm a bit confused about the functional tests. It looks like with my changes in place, we stop reporting this warning when run under Python 3.10 (because the `type` metaclass now defines `__or__`, and Astroid picks that up) but that's not how it's supposed to work.

If we're running on Python 3.10 but the `.rc` file sets `py-version=3.8` should we report errors according to the `py-version`? If that's the case, then we're in trouble as we'll have to differentiate between a case where `__or__` is coming from `type` (in which case we should only use it if we're configured for a version of Python that defines it) or whether it's in user-defined code, in which case we can use it on any version of Python.

Is that correct?

Pierre-Sassoulas:
> If we're running on Python 3.10 but the .rc file sets py-version=3.8 should we report errors according to the py-version?

Yes. I almost commented on this but did not because I thought the ``if not allowed_nested_syntax:`` condition might already be handling this. ``py-version`` is a valuable option but also relatively new in the codebase.

timmartin:
It feels like the right way to solve this is for Astroid not to report the existence of an `__or__` method on Python before v3.10, because it doesn't exist. If we solve this at the Astroid level, Pylint doesn't need any special rules and Astroid will be reporting slightly more correct data. But based on a quick search of the repo, it looks like Astroid doesn't support anything like a `py-version` setting.

Does this mean we need to have logic to deal with this in the Pylint checker? If so, I think it means that we can't benefit from implementing a `hasattr`, because we'll need to inspect the attribute to see what class it belongs to.

Pierre-Sassoulas:
> If we solve this at the Astroid level, Pylint doesn't need any special rules and Astroid will be reporting slightly more correct data. But based on a quick search of the repo, it looks like Astroid doesn't support anything like a py-version setting.

I think adding a ``py-version`` in astroid is very high effort because we'd have to know every specificity of each interpreter instead of just relying on the ast module for the interpreter (basically emulate a 3.8 parsing on a 3.10 interpreter). By contrast in pylint for a specific case we have to know what is the specificity of versions older than 3.x for this specific use case. What do you think @cdce8p ?

timmartin:
I have what I think is a fix for this. It's not elegant, but I can see that making Astroid infer all the cases correctly based on the configured version is not going to be easy.

doc/whatsnew/2/2.14/summary.rst
```diff
@@ -256,6 +256,11 @@ Other Changes
 
   Relates to #6612
 
+* Don't report ``unsupported-binary-operation`` on Python <= 3.9 when using the ``|`` operator
+  with types, if one has a metaclass that overloads ``__or__`` or ``__ror__`` as appropriate.
+
+  Closes #4951
+
```
Pierre-Sassoulas, line 263: Unfortunately we need to still add the changelog in full.rst too until we release 2.14.0.

Pierre-Sassoulas (approved):
There's a small changelog issue (sorry about that, we're trying to get rid of the duplication in the next versions), but other than that LGTM. I'll wait for a review from someone else.

tests/functional/a/alternative/alternative_union_syntax_error.py
```diff
@@ -87,3 +87,26 @@ class CustomDataClass3:
...
+
+class WithReverse(metaclass=ReverseMetaclass):
+    pass
+
+class DefaultMetaclass:
+    pass
+
+class_list = [WithForward | DefaultMetaclass]
+class_list_reversed_invalid = [WithReverse | DefaultMetaclass] # [unsupported-binary-operation]
+class_list_reversed_valid = [DefaultMetaclass | WithReverse]
```
jacobtylerwalls, line 112: Sorry to suggest annoying test cases, but I've seen people do the darndest things ⚡ 
```
  File "/Users/.../pylint/pylint/checkers/typecheck.py", line 1923, in <genexpr>
    attr.parent.qname() == "builtins.type" for attr in attrs
AttributeError: 'Assign' object has no attribute 'qname'
```
```suggestion
class_list_reversed_valid = [DefaultMetaclass | WithReverse]


# Pathological cases
class HorribleMetaclass(type):
    __or__ = lambda x: x

class WithHorrible(metaclass=HorribleMetaclass):
    pass

class_list = [WithHorrible | DefaultMetaclass]
```

pylint/checkers/typecheck.py
```diff
@@ -1902,13 +1902,48 @@ def _detect_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> Non
...
+        right_obj = astroid.helpers.object_type(node.right)
+        left_is_type = False
+        right_is_type = False
+
+        def _has_usable_overload(attrs):
+            # We have a candidate overload of the operator. However, if this
+            # is a Python 3.10 runtime but configured for pre-3.10 compatibility
+            # Astroid will have inferred the existence of __or__ / __ror__ on
+            # builtins.type, which is not suitable.
+            is_py310_builtin = all(
```
jacobtylerwalls, line 1922: You can just check the `pytype()` of the object. This might help you skirt my nasty test case.
```python
>>> import astroid
>>> l = astroid.extract_node("list")
>>> inferred = next(l.infer())
>>> inferred.pytype()
'builtins.type'
```

jacobtylerwalls (changes requested):

Pierre-Sassoulas:
Moving to 2.15 as we try to release asap :)

pylint/checkers/typecheck.py
timmartin: Thanks, it's good to know that exists. However, it doesn't seem to do what I need here. On a `nodes.ClassDef`, the `pytype()` method will always return `"builtins.type"` so I can't use it to differentiate between a user-defined class an a builtin.

timmartin:
I don't fully understand, but on Python 3.10 Astroid will infer two values for the attribute built-in on `type`, a function definition and a bound method. However, it seems simple enough to filter out both of these.

doc/whatsnew/2/2.14/summary.rst
timmartin: Now I've missed the 2.14 cut-off I guess I have to move this. I've put the note in `index.rst` in the 2.15 section because that's where all the section headings are, but I don't know if this is going to get moved into separate files later?

timmartin:
I think you're right in your expectation that it should not report a warning with the second-level metaclass. I was assuming Astroid `getattr` would do the right thing here, but I guess it's more complicated than that. I'll have a closer look at this.

jacobtylerwalls:
I think it might just be the assumption of direct parentage. Any time I see `.parent` I always go to try to violate that assumption somehow :D

tests/functional/a/alternative/alternative_union_syntax_error.py
```diff
@@ -87,3 +87,36 @@ class CustomDataClass3:
...
+    pass
+
+class WithReverse(metaclass=ReverseMetaclass):
+    pass
+
+class DefaultMetaclass:
+    pass
+
+class_list = [WithForward | DefaultMetaclass]
+class_list_reversed_invalid = [WithReverse | DefaultMetaclass] # [unsupported-binary-operation]
```
cdce8p, line 111: Can you add this testcase also to `alternative_union_syntax.py` which sets `py-version=3.10`?

pylint/checkers/typecheck.py
```diff
@@ -1902,13 +1903,50 @@ def _detect_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> Non
...
+            )
+            return (not is_py310_builtin) or self._py310_plus
+
+        if isinstance(left_obj, nodes.ClassDef) and is_classdef_type(left_obj):
+            try:
+                attrs = left_obj.getattr("__or__")
+                if _has_usable_overload(attrs):
+                    return
+                left_is_type = True
+            except astroid.NotFoundError:
```
cdce8p, line 1936:
```suggestion
            except astroid.exceptions.NotFoundError:
```

pylint/checkers/typecheck.py
```diff
@@ -1902,13 +1903,50 @@ def _detect_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> Non
...
+                left_is_type = True
+
+        if isinstance(right_obj, nodes.ClassDef) and is_classdef_type(right_obj):
+            try:
+                attrs = right_obj.getattr("__ror__")
+                if _has_usable_overload(attrs):
+                    return
+                right_is_type = True
+            except astroid.NotFoundError:
```
cdce8p, line 1945:
```suggestion
            except astroid.exceptions.NotFoundError:
```

cdce8p (changes requested):
> > If we solve this at the Astroid level, Pylint doesn't need any special rules and Astroid will be reporting slightly more correct data. But based on a quick search of the repo, it looks like Astroid doesn't support anything like a py-version setting.
> 
> I think adding a `py-version` in astroid is very high effort because we'd have to know every specificity of each interpreter instead of just relying on the ast module for the interpreter (basically emulate a 3.8 parsing on a 3.10 interpreter). By contrast in pylint for a specific case we have to know what is the specificity of versions older than 3.x for this specific use case. What do you think @cdce8p ?

It would be quite difficult to implement `py-version` in astroid unfortunately. True, we could do the AST parsing with a different version but that doesn't really change much. For instance, I believe the `__or__` and `__ror__` methods on the `type` metaclass are read directly from the interpreter. Thus we would need to manually update each and every change from Python at which point we're basically rebuilding CPython itself.

`py-version` was more designed for things which can easily be inferred by pylint, like recommendations for the use of new features (`f-strings`, ...).

The way you solved the issue here is quite clever! I left a few notes but after that I think it's ready.

pylint/checkers/typecheck.py
```diff
@@ -1901,14 +1902,55 @@ def _detect_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> Non
             if not allowed_nested_syntax:
                 self._check_unsupported_alternative_union_syntax(node)
 
+    def _includes_version_compatible_overload(self, attrs: list[nodes.NodeNG]):
+        """Check if a set of overloads of an operator includes one that
+        can be relied upon for our configured Python version.
+
+        If we are running under a Python 3.10+ runtime but configured for
+        pre-3.10 compatibility then Astroid will ahve inferred the
```
cdce8p, line 1910:
```suggestion
        pre-3.10 compatibility then Astroid will have inferred the
```

pylint/checkers/typecheck.py
```diff
@@ -1901,14 +1902,55 @@ def _detect_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> Non
...
+        existence of __or__ / __ror__ on builtins.type, but these aren't
+        available in the configured version of Python.
+        """
+        is_py310_builtin = all(
+            isinstance(attr, (nodes.FunctionDef, astroid.BoundMethod))
+            and attr.parent.qname() == "builtins.type"
+            for attr in attrs
+        )
+        return (not is_py310_builtin) or self._py310_plus
```
cdce8p, line 1919:
```suggestion
        return not is_py310_builtin or self._py310_plus
```
The brackets are redundant. `not` binds stronger than `or`.

pylint/checkers/typecheck.py
```diff
@@ -1901,14 +1902,55 @@ def _detect_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> Non
             if not allowed_nested_syntax:
                 self._check_unsupported_alternative_union_syntax(node)
 
+    def _includes_version_compatible_overload(self, attrs: list[nodes.NodeNG]):
```
cdce8p, line 1905:
```suggestion
    def _includes_version_compatible_overload(self, attrs: list) -> bool:
```
I'm not sure about the type for `attrs`. At the very least, it can also be `astroid.bases.BoundMethod` which doesn't inherit from `nodes.NodeNG`. Maybe lets leave it at just `list` for now.

pylint/checkers/typecheck.py
```diff
@@ -21,6 +21,7 @@
 from typing import TYPE_CHECKING, Any, Union
 
 import astroid.exceptions
```
cdce8p, line 23:
```suggestion

import astroid
import astroid.exceptions
```
Sorry for the confusion here. I missed that the bare `astroid` import is used as well.

tests/functional/a/alternative/alternative_union_syntax_error.py
```diff
@@ -87,3 +87,44 @@ class CustomDataClass3:
...
+
+class_list = [WithHorrible | DefaultMetaclass]
+
+class SecondLevelMetaclass(ForwardMetaclass):
+    pass
+
+class WithSecondLevel(metaclass=SecondLevelMetaclass):
+    pass
+
+class_list_with_second_level = [WithSecondLevel | DefaultMetaclass]
```
cdce8p, line 130: This doesn't seem to work at the moment. I only spend a few minutes investigating it, but couldn't  a quick fix. Maybe it's better to remove that test case here and open a followup issue for it to be addressed later. At least that way this PR could be merged.

pylint/checkers/typecheck.py
cdce8p: Please address the comment above

pylint/checkers/typecheck.py
cdce8p: Please address the comment above

cdce8p (changes requested):

tests/functional/a/alternative/alternative_union_syntax_error.py
timmartin: I was hoping to get to the bottom of it, because I didn't want to commit a half-working fix. But thinking about it, this failure case isn't a new false positive so I guess this fix will be an improvement overall. I'll pull this out and open another PR for it.

jacobtylerwalls:
Yep, thanks for checking! I was the one that asked for that test case we haven't solved, but I agree we can postpone. I think it's just a matter of doing a recursive search on ancestors instead of relying on direct `.parent`-age, but we can postpone.

jacobtylerwalls (approved):

PR #6665, opened 2022-05-22
Remove redundant options documentation and improve formatting

DanielNoord:
## Type of Changes

## Description

@Pierre-Sassoulas Potentially do this before https://github.com/PyCQA/pylint/pull/6589. I tried to improve the formatting of the options page a little bit. After this we can do one more PR that removes the `pylint features` page all together I think.

### DIFF
--- a/doc/exts/pylint_options.py
+++ b/doc/exts/pylint_options.py
@@ -66,7 +66,8 @@ def _get_all_options(linter: PyLinter) -> OptionsDataDict:
 def _create_checker_section(
     checker: str, options: list[OptionsData], linter: PyLinter
 ) -> str:
-    checker_string = get_rst_title(f"``{checker.capitalize()}`` **Checker**", "-")
+    checker_string = f".. _{checker}-options:\n\n"
+    checker_string += get_rst_title(f"``{checker.capitalize()}`` **Checker**", "-")
 
     toml_doc = tomlkit.document()
     pylint_tool_table = tomlkit.table(is_super_table=True)
--- a/doc/exts/pylint_options.py
+++ b/doc/exts/pylint_options.py
@@ -66,7 +66,8 @@ def _get_all_options(linter: PyLinter) -> OptionsDataDict:
 def _create_checker_section(
     checker: str, options: list[OptionsData], linter: PyLinter
 ) -> str:
-    checker_string = get_rst_title(f"``{checker.capitalize()}`` Checker", "^")
+    checker_string = get_rst_title(f"``{checker.capitalize()}`` **Checker**", "-")
+
     toml_doc = tomlkit.document()
     pylint_tool_table = tomlkit.table(is_super_table=True)
     toml_doc.add(tomlkit.key(["tool", "pylint"]), pylint_tool_table)
@@ -75,11 +76,11 @@ def _create_checker_section(
 
     for option in sorted(options, key=lambda x: x.name):
         checker_string += get_rst_title(f"--{option.name}", '"')
-        checker_string += f"\nDescription:\n  *{option.optdict.get('help')}*\n\n"
+        checker_string += f"*{option.optdict.get('help')}*\n\n"
         if option.optdict.get("default") == "":
-            checker_string += 'Default:\n  ``""``\n\n\n'
+            checker_string += '**Default:** ``""``\n\n\n'
         else:
-            checker_string += f"Default:\n  ``{option.optdict.get('default')}``\n\n\n"
+            checker_string += f"**Default:**  ``{option.optdict.get('default')}``\n\n\n"
 
         # Start adding the option to the toml example
         if option.optdict.get("hide_from_config_file"):
@@ -138,13 +139,13 @@ def _write_options_page(options: OptionsDataDict, linter: PyLinter) -> None:
         ".. This file is auto-generated. Make any changes to the associated\n"
         ".. docs extension in 'doc/exts/pylint_options.py'.\n\n"
         ".. _all-options:",
-        get_rst_title("Standard Checkers:", "^"),
+        get_rst_title("Standard Checkers", "^"),
     ]
     found_extensions = False
 
     for checker, checker_options in options.items():
         if not found_extensions and checker_options[0].extension:
-            sections.append(get_rst_title("Extensions:", "^"))
+            sections.append(get_rst_title("Extensions", "^"))
             found_extensions = True
         sections.append(_create_checker_section(checker, checker_options, linter))
 
--- a/doc/user_guide/configuration/all-options.rst
+++ b/doc/user_guide/configuration/all-options.rst
@@ -9,6 +9,8 @@ Standard Checkers
 ^^^^^^^^^^^^^^^^^
 
 
+.. _main-options:
+
 ``Main`` **Checker**
 --------------------
 --analyse-fallback-blocks
@@ -271,6 +273,8 @@ Standard Checkers
 
    </details>
 
+
+.. _basic-options:
 
 ``Basic`` **Checker**
 ---------------------
@@ -580,6 +584,8 @@ Standard Checkers
    </details>
 
 
+.. _classes-options:
+
 ``Classes`` **Checker**
 -----------------------
 --check-protected-access-in-special-methods
@@ -644,6 +650,8 @@ Standard Checkers
 
    </details>
 
+
+.. _design-options:
 
 ``Design`` **Checker**
 ----------------------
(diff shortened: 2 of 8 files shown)


Pierre-Sassoulas (commented):
I think it would be nice to add the generated file(s?) to the repository (as we did for the messages list) in order to see what changed because the review is really hardcore otherwise 😄 We could auto-update them for release with tbump too. What do you think ?

DanielNoord:
> I think it would be nice to add the generated file(s?) to the repository (as we did for the messages list) in order to see what changed because the review is really hardcore otherwise 😄 We could auto-update them for release with tbump too. What do you think ?

Agreed! But perhaps let's do that after the current rework is done? I was quite happy this PR was under 100 lines and I tend to look at the final output in RTD anyway when reviewing these PRs.

Pierre-Sassoulas:
> I was quite happy this PR was under 100 lines

Yeah sure this PR is quite elegant 😄 We don't have to ruin it though, we can add the generated file from another one and rebase on main when it's merged. It's just a lot harder to check everything when you need to compare two HTMLs side by side instead, of a a git diff on .rst code then check the modified HTML visually. Do you want me to add the generated file in another PR ?

DanielNoord:
> Do you want me to add the generated file in another PR ?

Yeah, but I would wait with adding the generated file right now. It's just more stuff we potentially need to move around during the renaming process. We shouldn't really need to touch these generate files at this point anyway 😄

Pierre-Sassoulas:
OK then let's do #6589 first.

Pierre-Sassoulas:
Opened #6678 regarding the generated rst added to the repository.

pylint/utils/docs.py
```diff
@@ -42,27 +42,29 @@ def _get_checkers_infos(linter: PyLinter) -> dict[str, dict[str, Any]]:
...
-                        result += get_rst_title(title, "~")
-                        assert isinstance(options, list)
-                        result += f"{get_rst_section(None, options)}\n"
+        for checker in linter.get_checkers():
+            name = checker.name
+            if name == MAIN_CHECKER_NAME:
+                if checker.options:
+                    with warnings.catch_warnings():
+                        warnings.filterwarnings("ignore", category=DeprecationWarning)
+                        for section, options in checker.options_by_section():
```
Pierre-Sassoulas: ``options_by_section`` is deprecated, it might be a good thing to find a solution without using it so the transition to 3.0 is easier. Is it a problem ?

pylint/utils/docs.py
```diff
@@ -42,27 +42,29 @@ def _get_checkers_infos(linter: PyLinter) -> dict[str, dict[str, Any]]:
     return by_checker
 
 
-def _get_checkers_documentation(linter: PyLinter) -> str:
+def _get_checkers_documentation(linter: PyLinter, show_options: bool = True) -> str:
     """Get documentation for individual checkers."""
-    result = get_rst_title("Pylint global options and switches", "-")
-    result += """
+    result = ""
+    if show_options:
```
Pierre-Sassoulas, line 48: I was going to suggest a function but it's easier if I do the commit directly.

pylint/utils/docs.py
DanielNoord: I don't see this as a priority. We could also make it a private function in `3.0` and keep using it ourselves. The important thing is that downstream packages don't depend on it.

Pierre-Sassoulas: Right, that make sense.

pylint/utils/docs.py
Pierre-Sassoulas: (Beside we can check that the refactor is not changing the behavior easily as we have the full generated output in the repo)

DanielNoord:
> Found an issue when the default value is the current interpreter in all-options.rst:
> 
> ```diff
> -**Default:**  ``(3, 10)``
> +**Default:**  ``(3, 8)``
> -   py-version = [3, 10]
> +   py-version = [3, 8]
> ```
> 
> My interpreter is 3.8, yours is probably 3.10 :)

That should probably be fixed in a follow-up PR. We need to special case it I guess... Annoying 😓

Pierre-Sassoulas:
On the plus side we detected the problem, adding the full output has its perks 😄

Pierre-Sassoulas:
I want to add link between section, I think we'll need to do a refactor first. There's issue with the doc generation because we have a string for the checker name instead of the checker so it's hard to do the right ref between extension and standard checker. I can push the work in progress if you're interested or simply want to understand what I'm talking about 😄

DanielNoord:
> I want to add link between section, I think we'll need to do a refactor first. There's issue with the doc generation because we have a string for the checker name instead of the checker so it's hard to do the right ref between extension and standard checker. I can push the work in progress if you're interested or simply want to understand what I'm talking about 😄

I understand what you are talking about, but wouldn't ```:ref:`checker_name Checker` ``` work?

Anyway, let's merge this PR first and do follow-ups. This already has 1500 changed lines 😅

doc/exts/pylint_options.py
```diff
@@ -63,10 +63,21 @@ def _get_all_options(linter: PyLinter) -> OptionsDataDict:
     return all_options
 
 
+def _get_checker_title(checker: str) -> str:
+    result = _get_checker_configuration_rst_anchor(checker)
+    result += get_rst_title(f"``{checker.capitalize()}`` **Checker**", "-")
+    result += f":ref:`See the description of the checker itself <pylint.extensions.{checker}>`.\n\n"
```
Pierre-Sassoulas, line 69:
```suggestion
    result += f":ref:`See the description of the checker itself <pylint.extensions.{checker}>`.\n\n"
```
Here with a string we can't know if the checker is a pylint extension or a standard checker.

doc/user_guide/configuration/all-options.rst
```diff
@@ -1192,7 +1240,7 @@ Standard Checkers
...
+*Spelling dictionary name. Available dictionaries: en_GB (aspell), en_US (hunspell), en_AU (aspell), en (aspell), en_CA (aspell).*
```
DanielNoord, line 1243: This is another problematic default value. I'm starting to doubt whether we should in fact include these generated files...

DanielNoord:
> > Anyway, let's merge this PR first and do follow-ups
> 
> We're removing the information from the page so we should probably add a link to where the information are at, right ? Otherwise it's really hard to find the option / message generated.
> 
> (I thought of a follow-up where I'd add the link the to the message overview in the message list of the extension page)

Hm, then we should probably split some of the work here. Perhaps by adding the links first? I'm losing track of the refactors, added TODOs and actual changes that are in this PR.

doc/user_guide/configuration/all-options.rst
Pierre-Sassoulas: It's not really problematic if the generated file change a little (In fact I could have avoided adding this by doing a partial add). The fact is it permits to see the doc depends on the generator's operating system which it should not, so imo it's a good think that it makes problem apparent. (We can fix it later, should be something like "depends on your operating system" or "default to your current interpreter").

Pierre-Sassoulas:
> Hm, then we should probably split some of the work here. Perhaps by adding the links first?

Let me open a pre-refactor MR :) !

DanielNoord:
Btw, perhaps we should draft this and focus on the changelog changes so we can release `2.14`?

Pierre-Sassoulas (changes requested):
I think we need to add a link to the option page of the checker, did you try ``:ref:`checker_name Checker`` like you suggested earlier ? I'm going to continue working on the doc for some time. Mainly links between section for easier navigability.

pylint/checkers/base_checker.py
```diff
@@ -126,8 +127,11 @@ def get_full_documentation(
         # options might be an empty generator and not be False when cast to boolean
         options_list = list(options)
         if options_list:
-            result += get_rst_title(f"{checker_title} Options", "^")
-            result += f"{get_rst_section(None, options_list)}\n"
+            if show_options:
+                result += get_rst_title(f"{checker_title} Options", "^")
+                result += f"{get_rst_section(None, options_list)}\n"
+            else:
+                result += f"You can find the options of this checker here: :ref:`{self.name}-options`\n\n"
```
Pierre-Sassoulas, line 134:
```suggestion
                result += f"See also :ref:`{self.name}'s options' documentation <{self.name}-options`>\n\n"
```

Pierre-Sassoulas (approved):
Look neat ! I have a suggestion for the text of the link, let me know what you think.

PR #6666, opened 2022-05-22
Add captions to indexbar in documentation

DanielNoord:
## Type of Changes

## Description

@Pierre-Sassoulas Just to show you what I was thinking of in terms of sections. You can also do this in https://github.com/PyCQA/pylint/pull/6589 if you want, but I think these headings in the index help visualize stuff. With this we can also think a little better about the index itself: do we want more levels? Do we want to elevate some stuff to a higher level?

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -6,12 +6,36 @@
    :hidden:
 
    tutorial
+
+.. toctree::
+   :caption: Using pylint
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
This looks great, I'm going to integrate that in #6589

PR #6677, opened 2022-05-23
Emit `used-before-assignment` for variables only defined under always false tests

### LINKED ISSUE #4913: False negative for `used-before-assignment` for variables only defined beneath always false conditions
### Bug description

The following code produces no warning:
```python
""" bla """

VAR1 = False
if VAR1:
    VAR2 = True
if VAR2:
    pass
```
even though `VAR2` is clearly uninitialized.

This is similar to the issue reported in https://github.com/PyCQA/pylint/issues/4045 , thouhg in a much simpler scenario, without the need for exceptions to be handled correctly.

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
```

### Expected behavior

Warning about the possibly uninitialized variable

### Pylint version

```shell
Reproduces on 2.5.0, 2.7.0, 2.8.0, 2.9.0, 3.0.0
```

### OS / Environment

Windows

jacobtylerwalls:
## Type of Changes

## Description

Closes #4913

Determine if a variable is defined only under a code path guarded by an IF statement that `astroid` can determine is always falsy. Emit `used-before-assignment` if that variable is later accessed.


DanielNoord:
@jacobtylerwalls This seems like it could benefit from https://github.com/PyCQA/astroid/pull/1189. Perhaps we should try and get that landed first?

jacobtylerwalls:
I agree that's a good use of effort to get that merged. Not sure I immediately see the utility here, but it's certainly possible after looking more closely! :D

DanielNoord:
> I agree that's a good use of effort to get that merged. Not sure I immediately see the utility here, but it's certainly possible after looking more closely! :D

Well, if you don't see the usability than that's a clear point of review. The system introduced there should be "easily" extendable to also include `is False`, `is sys.version` etc. checks.
I have eagerly waiting for the PR to get merged/reviewed to start exploring some of the other constraints we could start checking for.

jacobtylerwalls:
Leaving notes for when I come back: many of the messages in the primer differ refer to this pattern:
```python
if magic:
    magic_num = 1
...
if magic:
   print(magic_num)  # likely that nothing happened to `magic` in the meantime, so safe(ish)
```
Maybe a loose comparison of if-node tests or the pending astroid constraints PR could help.

Pierre-Sassoulas:
I'm still in shock about how great the primer bot is. It's like developing pylint in easy mode.

jacobtylerwalls:
Along the lines of #2835, this should probably be a new message `possibly-used-before-assignment` or `possibly-unbound`. The two patterns that the primer projects and pylint itself are failing on are:

**1**
```python
if magic:
    magic_num = 1
...
print(magic_num)  # likely but not certain that nothing happened to `magic` in the meantime
```

**2**
```python
# handle two possible values
if flag == "expected_val1":
    magic_num = 1
elif flag == "expected_val2":
    magic_num = 2
# no else ... so magic_num may not be defined ... but does this ever realistically happen?
print(magic_num)
```

jacobtylerwalls:
> Along the lines of https://github.com/PyCQA/pylint/issues/2835, this should probably be a new message possibly-used-before-assignment or possibly-unbound. The two patterns that the primer projects and pylint itself are failing on are:

Actually, these patterns turned out to be not an issue because I wasn't correctly checking that all the inferred values were falsy. So that problem goes away. At least pattern 1 could be worth thinking about when doing #2835, though.

pylint/checkers/variables.py
```diff
@@ -665,6 +673,146 @@ def get_next_to_consume(self, node: nodes.Name) -> list[nodes.NodeNG] | None:
...
+        ) or (
+            any(
+                NamesConsumer._exhaustively_define_name_raise_or_return(name, nested)
+                for nested in body
+                if isinstance(
+                    nested,
+                    (nodes.If, nodes.TryExcept, nodes.TryFinally, nodes.With),
+                )
+            )
+        )
```
Pierre-Sassoulas, line 732:
```suggestion
        return any(
            NamesConsumer._defines_name_raises_or_returns(name, if_body_stmt)
            or isinstance(
                if_body_stmt,
                (nodes.If, nodes.TryExcept, nodes.TryFinally, nodes.With),
            )
            and NamesConsumer._exhaustively_define_name_raise_or_return(
                name, if_body_stmt
            )
            for if_body_stmt in body
        )
```
Can't we do that and loop only once instead of twice ?

pylint/checkers/variables.py
```diff
@@ -665,6 +673,146 @@ def get_next_to_consume(self, node: nodes.Name) -> list[nodes.NodeNG] | None:
...
+                if closest_if.test.name in loop_names:
+                    continue
+                if for_ancestor is not None and isinstance(
+                    other_node, nodes.AssignName
+                ):
+                    if any(
+                        an.name == other_node.name
+                        for an in for_ancestor.nodes_of_class(nodes.AssignName)
+                    ):
+                        continue
```
Pierre-Sassoulas, line 810: I feel like creating a function here would make sense, but the ``continue`` design makes it a little hard. To be fair I find the ``_uncertain_nodes_in_false_tests`` function a little complicated, Wild proposition but maybe we could loop on a list of conditional instead of having a conditional then a continue ?

Pierre-Sassoulas (commented):
Thank you @jacobtylerwalls, clearly you invested a lot of work in this one and it shows. It become stale because it feel hard to review the variable checker, but I gave it a try.

pylint/checkers/variables.py
jacobtylerwalls: I know it's long! Thanks for reviewing. I think the issue with a list of conditionals is that without a walrus operator I'd be repeating lookups like `closest_if` and `other_node_frame` unnecessarily. Happy to discuss other approaches for factoring, though.

Pierre-Sassoulas: Let's revisit when python 3.7 has reached end of life then 😄

jacobtylerwalls (commented):
Let me know if you think I should follow the idea below to reduce the size of the PR.

pylint/checkers/variables.py
```diff
@@ -665,6 +673,149 @@ def get_next_to_consume(self, node: nodes.Name) -> list[nodes.NodeNG] | None:
...
+            uncertain_nodes.append(other_node)
+            self._if_nodes_deemed_uncertain.add(closest_if)
+
+        return uncertain_nodes
+
+    @staticmethod
+    def _name_defined_later_in_loop(
+        closest_if: nodes.If, other_node: nodes.NodeNG
+    ) -> bool:
+        """Identify if the name tested in `closest_if` is assigned later in a loop."""
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Let's use inference if it can simplify things. It's "costly" in term of performance but not if we have time to make it better in astroid, and it's a lot easier to maintain than something clever in pylint so it could be a virtuous cycle where it create time to make astroid faster.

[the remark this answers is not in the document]
Pierre-Sassoulas: (Unrelated but this is what I have in mind after reviewing a lot of issues recentely and half of them are control flow issue : It's the same thing for control flow, if we can have a clean generic way to do it in astroid, it could avoid dozen of little checks in pylint.)

Pierre-Sassoulas (commented):
The tests cases are on point, but the code is complicated (because the problem we're trying to solve is complicated). I feel like we need a generic solution in astroid, something like infer raising a different exception when the variable can be unassigned, maybe ? Or an assigned attribute ? There is an a lot of check that require control flow, implement control flow in each checkers is going to be a lot more work than doing that in astroid. What do you think ?

pylint/checkers/variables.py
```diff
@@ -668,6 +676,116 @@ def get_next_to_consume(self, node: nodes.Name) -> list[nodes.NodeNG] | None:
...
+                        name, handler
+                    )
+                    for handler in node.handlers
+                )
+            )
+        if isinstance(node, nodes.With):
+            return NamesConsumer._defines_name_raises_or_returns_recursive(name, node)
+
+        if not isinstance(node, nodes.If):
+            return False
```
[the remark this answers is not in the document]
jacobtylerwalls: I looked at this, but `pylint` doesn't understand `cast` and raises `no-member` for members like` orelse.` So I think altogether, I'd keep it the way it is.

jacobtylerwalls:
> I feel like we need a generic solution in astroid, something like infer raising a different exception when the variable can be unassigned, maybe ? Or an assigned boolean attribute in nodes ?

I just want to relate this to @DanielNoord's [comment](https://github.com/PyCQA/pylint/pull/6677#issuecomment-1135026433) about the astroid constraint PR.

I think we might be conflating two things. The current proposal in PyCQA/astroid#1189 is to add constraints on inferred values. Here, we're not using inference. We don't care what values `undefined_var` takes: if it's never defined, it's undefined! We don't need to infer it, and it would be a waste of time to do so.

> There is an a lot of check that require control flow, implementing control flow in each checkers is going to be a lot more work than doing that in astroid directly.

I'm sympathetic to that, but another way to look at it would be this is a "soft-launch" of control flow, using just one message, where the pylint primer results look promising, and then as we iron out kinks, we can implement a similar approach in astroid later, and extend it for `del` and the rest.

Pierre-Sassoulas:
The code in psycopg looks very suspicious, it might be a genuine bug. But in pandas it seems like a false positive.

Pierre-Sassoulas (approved):
LGTM 👌 Feel free to merge, I'm not super comfortable with the variables checker and I don't understand everything, But the tests and the primer make me pretty confident it won't create too much issues.

tests/functional/u/undefined/undefined_variable_py38.py
```diff
@@ -180,3 +180,8 @@ def expression_in_ternary_operator_inside_container_wrong_position():
 # Self-referencing
 if (z := z):  # [used-before-assignment]
     z = z + 1
+
+
+if (never_defined := False):
+    pass
+print(never_defined)  # [used-before-assignment]
```
cdce8p, line 187: This looks like a new false-positive. Assignment expressions are bound `before` the conditional check. Thus this executes just fine and will print `False`.

jacobtylerwalls: Good catch. Fixed in #7847.

Issue #6676, opened 2022-05-23
Public interface & versioning


MrMino:
### Question

Does Pylint follow any particular versioning scheme when it comes to adding new warning types?

#1650 would lead me to believe that the project has a notion of a public interface, which should not be broken within minor releases. So, in general:

What (at minimum) is considered to be "public-interface" of pylint?

### Documentation for future user

Somewhere in the documentation, but I don't see any intuitive place for it.

Searching for `pylint versioning` shows an entry in the [FAQ](https://pylint.pycqa.org/en/latest/faq.html#what-kind-of-versioning-system-does-pylint-use), so it might be good to place it somewhere near there.

Pierre-Sassoulas:
[This](https://github.com/PyCQA/pylint/issues/1650#issuecomment-327223970) is still relevant:

> Obviously we do not guarantee that linting output won't change. New checks are added. Old checks are improved to catch more violations (or to avoid false-positives). Sometimes avoiding false positives means that we get some new not-emitted false-negatives. You can also get different results with different version of astroid.

Regarding the second part of the thread, the json output is public API we won't change it in a minor, but we did remove pylint options like ``py3k``. I considered that it was a tool not included in pylint (it was also deactivated for a long time and used for a 'one of' check before migration to python 3).

>  any function, class or module that is underscored is considered private and can change at any time. All other stuff is "public" unless there is an explicit __all__ variable. In that case __all__ is determines what is and what isn't public.

That being said it's true for new code, but we have a lot of small legacy helper functions everywhere that aren't prefixed  and it's not wise to use them like a public API. I don't see myself increasing our maintenance burden threefold by declaring everything public when we document our API later on. The deprecation for 3.0 are already taking a lot of time for pylint's secondary purpose (being an extendable linting lib) to the detriment of the main purpose of pylint (linting code, fast, with few false positives or negatives).

Even if we take care of deprecation the best we can they could inadvertently be modified to fix things...  We also have a thousands little thing to do before being able to release 3.0 so if something make pylint a lot better at the cost of a little breaking change for maintainer... my vote will go in favor of a little breaking change instead of doing all the changes we need to do for 3.0. So if it's not clearly documented, like the custom checker, or ``pylint.lint.Run``... probably don't build anything important on it.

MrMino:
Thanks for fast response. I think I might've under-specified my second question. I wasn't really asking about API, but rather things like CLI, JSON output, config, output stability in general, etc.

[...]

Issue #6632, opened 2022-05-17
Todo/to-do list for 2.15.0


DanielNoord:
TODO:
- [x]  Fix any outstanding `TODO: 2.15:` comments
- [x] Integrate latest astroid patch
- [x] Revert #7244 
Opening this now to track discussion on our next major release.

DanielNoord:
One of the things I would like to discuss is to reserve the milestone for actual user-facing changes. I just had a look at the [2.14 milestone](https://github.com/PyCQA/pylint/milestone/55) and it's closed tab is actually pretty unusable. I have done this myself as well, but by adding all changes to documentation and our CI runs it's not actually clear what real changes were made to the code/functionality.

Obviously there will always be a balance somewhere, but I would like to propose to only add new functionality, bug fixes or refactors of actual code into the milestone. If people want to see all commits between `2.13.0` and `2.14.0` they can use `git` to tell then.

Pierre-Sassoulas:
Regarding user-facing or not in #5728, I'd like to separate clearly what is user facing from what is useful only for pylint's devs. I'm going to enforce a regex, maybe the current ``(\*\s[\S[\n  ]+?]*\n\n\s\s(Refs|Closes)) #\d{1,5}`` maybe something with only the PR in parenthesis like black ``* yadayada on a single line (#6632)`` (I think changing the style of the release notes makes sense as the PR will be linked to the issues anyway, and it's easier to check that the PR number is in the whats'new because we have access to it from the CI). It's still a work in progress but I can push the draft if you're interested. 

Regarding the milestone, I'm using it as a big todo list to see what still need to be done. I think it's a good thing to have every issue/PR from a version to be in the milestone. It's possible to check what you want to check by excluding label or requiring labels for example this indicate that we have currently 136 meaningful issue in 2.14 : https://github.com/PyCQA/pylint/issues?q=is%3Aclosed+milestone%3A2.14.0+-label%3A%22Documentation+%F0%9F%93%96%22+-label%3Adependency+-label%3AMaintenance 

The way to do that is at the bottom of the label list in the interface.

![filtering labels](https://user-images.githubusercontent.com/5493666/168763474-3ac7a160-7bad-4951-8fae-9ea41968dda5.png)

DanielNoord:
> I'd like to separate clearly what is user facing from what is useful only for pylint's devs.

👍 

> Regarding the milestone, I'm using it as a big todo list to see what still need to be done.

[...]

Issue #3604, opened 2020-05-07
Warning with underscore _ breaks pragma parsing


VladimirSlavik:
For a particular warning in disable pragma, pylint fails to parse the identifier string. The warning comes from a plugin (pocketlint) and is called `found-_-in-module-class` (note underscore). Instead of disabling, E0012 is emitted and the disable does not happen (obviously). Once the disable is changed to use the numeric identifier `W9902`, the issue goes away.

This regression appeared after updating pylint from 2.4.4. to 2.5.0.

### Steps to reproduce
1. pip3 install --user pocketlint
2. pip3 install --user pylint==2.5.0
3. git clone git@github.com:rhinstaller/anaconda.git
4. cd anaconda
5. git checkout anaconda-33.13-1
6. PYTHONPATH=\`pwd` python3 -m pylint --rcfile tests/pylint/pylintrc anaconda.py
7. Replace `found-_-in-module-class` in the indicated line with `W9902`
8. PYTHONPATH=\`pwd` python3 -m pylint --rcfile tests/pylint/pylintrc anaconda.py

### Current behavior
Buggy behavior:
```
E0012(bad-option-value):anaconda.py:379,0: : Bad option value 'found-'
E0012(bad-option-value):anaconda.py:379,0: : Bad option value '-in-module-class'
W9902(found-_-in-module-class):anaconda.py:381,57: : Found _ call at module/class level
W9902(found-_-in-module-class):anaconda.py:382,35: : Found _ call at module/class level
```
After replacing:
(no output)

See relevant files:
* Reported error location: https://github.com/rhinstaller/anaconda/blob/master/anaconda.py#L379
* Definition of the message: https://github.com/rhinstaller/pocketlint/blob/master/pocketlint/checkers/intl.py#L66

### Expected behavior

No output 

### pylint --version output

```
$ python3 -m pylint --version
pylint 2.5.0
astroid 2.4.1
Python 3.7.7 (default, Mar 13 2020, 21:39:43) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]
```

VladimirSlavik:
I have edited the report to pin the version of code for the "broken" one - obviously the line will be changed to the variant that does not trigger the bug (for now? forever?).

jkonecny12:
Another reproducer which does not depend on Anaconda project.

pylint-reproducer.py
```
# pylint: disable=raw_input-builtin
raw_input()
```

Then run
```
[...]

PR #6688, opened 2022-05-24
Reorganize the changelog and the way it's displayed in the doc

### LINKED ISSUE #5728: A single changelog file instead of two

Pierre-Sassoulas:
## Type of Changes

## Description

This is a work in progress to discuss the reorganisation of the changelog. TODO:

- [x] Burst the current changelog into multiple file in the documentation
- [x] Validate the reorganization in doc for changelog < 2.14.0 with summary / full changelog or both #6721 
- [ ] Discuss new changelog for 2.15.0 as proposed in https://github.com/PyCQA/pylint/pull/6641#issuecomment-1130257706
- [ ] Add a CI check for the new changelog as proposed in https://github.com/PyCQA/pylint/issues/6632#issuecomment-1128866849
- [x] Document what can change between major / minor / patch version as seen in https://github.com/PyCQA/pylint/issues/6676#issuecomment-1135423026
- [x] Upgrade the tbump script to release a version accordingly
- [ ] ~~Propagate the change to astroid (?!)~~
- [x] Upgrade the help around the changelog everywhere

Closes #5728

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,25 @@
+name: changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled, reopened]
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Changelog Entry Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Grep CHANGES.md for PR number
+        if:
+          contains(github.event.pull_request.labels.*.name, 'skip news :mute:') != true
+        run: |
+          grep -Pz "(\*\s[\S[\n
+]+?]*\n\n\s\s(Refs|Closes|Follow-up in|Fixes part of)) (PyCQA/astroid)?#${{ github.event.pull_request.number }}" doc/whatsnew/2/**/*.rst || \
+          (echo "Please add '(#${{ github.event.pull_request.number }})' change line to 'doc/whatsnew/2/2.15/index.rst' \
+          (or if appropriate, ask a maintainer to add the 'skip news' label)" && \
+          exit 1)
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -47,3 +47,5 @@ Other Changes
 
 Internal changes
 ================
+
+* Added a check in CI for changelogs entry (#6688)
--- a/script/check_changelog.py
+++ b/script/check_changelog.py
@@ -1,0 +1,150 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+"""Small script to check the changelog. Used by 'changelog.yml' and pre-commit.
+
+If no issue number is provided we only check that proper formatting is respected."""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from collections.abc import Iterator
+from pathlib import Path
+from re import Pattern
+
+VALID_ISSUES_KEYWORDS = [
+    "Refs",
+    "Closes",
+    "Follow-up in",
+    "Fixes part of",
+    "Closes BitBucket",
+    "Closes Logilab",
+]
+ISSUE_NUMBER_PATTERN = r"#\d{1,5}"
+VALID_ISSUE_NUMBER_PATTERN = r"\*[\S\s]*?" + ISSUE_NUMBER_PATTERN
+ISSUES_KEYWORDS = "|".join(VALID_ISSUES_KEYWORDS)
+PREFIX_CHANGELOG_PATTERN = (
+    rf"(\*\s[\S[\n ]+?]*\n\n\s\s({ISSUES_KEYWORDS})) (PyCQA/astroid)?"
+)
+VALID_CHANGELOG_PATTERN = PREFIX_CHANGELOG_PATTERN + ISSUE_NUMBER_PATTERN
+
+ISSUE_NUMBER_COMPILED_PATTERN = re.compile(ISSUE_NUMBER_PATTERN)
+VALID_CHANGELOG_COMPILED_PATTERN: Pattern[str] = re.compile(VALID_CHANGELOG_PATTERN)
+VALID_ISSUE_NUMBER_COMPILED_PATTERN: Pattern[str] = re.compile(
+    VALID_ISSUE_NUMBER_PATTERN
+)
+
+DOC_PATH = (Path(__file__).parent / "../doc/").resolve()
+PATH_TO_WHATSNEW = DOC_PATH / "whatsnew"
+UNCHECKED_VERSION = [
+    # Not checking version prior to 1.0.0 because the issues referenced are a mix
+    # between Logilab internal issue and Bitbucket. It's hard to tell, it's
+    # inaccessible for Logilab and often dead links for Bitbucket anyway.
+    # Not very useful generally, unless you're an open source historian.
+    "0.x",
+]
+
+NO_CHECK_REQUIRED_FILES = {
+    "index.rst",
+    "full_changelog_explanation.rst",
+    "summary_explanation.rst",
+}
+
+
+def sorted_whatsnew(verbose: bool) -> Iterator[Path]:
+    """Return the whats-new in the 'right' numerical order ('9' before '10')"""
+    numeric_whatsnew = {}
+    for file in PATH_TO_WHATSNEW.glob("**/*"):
+        relpath_file = file.relative_to(DOC_PATH)
+        if file.is_dir():
+            if verbose:
+                print(f"I don't care about '{relpath_file}', it's a directory : 🤖🤷")
+            continue
+        if file.name in NO_CHECK_REQUIRED_FILES:
+            if verbose:
+                print(
+                    f"I don't care about '{relpath_file}' it's in 'NO_CHECK_REQUIRED_FILES' : 🤖🤷"
+                )
+            continue
+        version = (
+            file.parents[0].name if file.stem in {"summary", "full"} else file.stem
+        )
+        if any(version == x for x in UNCHECKED_VERSION):
+            if verbose:
+                print(
+                    f"I don't care about '{relpath_file}' {version} is in UNCHECKED_VERSION : 🤖🤷"
+                )
+            continue
+        if verbose:
+            print(f"I'm going to check '{relpath_file}' 🤖")
+        num = tuple(int(x) for x in (version.split(".")))
+        numeric_whatsnew[num] = file
+    for num in sorted(numeric_whatsnew):
+        yield numeric_whatsnew[num]
+
+
+def main(argv: list[str] | None = None) -> int:
+    argv = argv or sys.argv[1:]
+    description = __doc__
+    description += " The regex we enforce is: " + VALID_CHANGELOG_PATTERN
+    parser = argparse.ArgumentParser(description=description)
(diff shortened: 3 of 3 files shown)


DanielNoord:
I like the separation of the different major version in the sidebar.

However, I don't really like how there are 3/4 different pages for one version. Taking `2.13` as example. There is:
https://pylint--6688.org.readthedocs.build/en/6688/whatsnew/2/2.13/index.html#
https://pylint--6688.org.readthedocs.build/en/6688/whatsnew/2/2.13/summary.html
https://pylint--6688.org.readthedocs.build/en/6688/whatsnew/2/2.13/full.html

The amount of pages also doesn't really correspond with the amount of links in the sidebar. Which is somewhat confusing.
Besides, what does the https://pylint--6688.org.readthedocs.build/en/6688/whatsnew/2/2.13/index.html page offer besides the sidebar? We can add the warning in a `.. note::` directive at the top of each page and then we can remove the extra click we need to do now.

My suggestion would be one page for every minor version which starts with the current `Changelog` (so the different sections) and then just all what's new under it in the same page. By using different arguments in `get_rst_title` we can create different header levels so that those headers aren't all so big.

Pierre-Sassoulas:
It's not generated I did everything manually, so I can't use 'get_rst_title' maybe we could transform the changelog to JSON so it's easier to generate and we have proper link to issues.. ? I'm not sure it's worth it though. 

We'd need a one off script to convert to JSON and a script to generate from JSON for a changelog that would never change. Maybe we could use the JSON for everything even the new changelogs ?

DanielNoord:
> It's not generated I did everything manually, so I can't use 'get_rst_title' maybe we could transform the changelog to JSON so it's easier to generate and we have proper link to issues.. ? I'm not sure it's worth it though.
> 
> We'd need a one off script to convert to JSON and a script to generate from JSON for a changelog that would never change. Maybe we could use the JSON for everything even the new changelogs ?

Ah, we could also just change the headings. Sphinx decides the headers based on the order they occur on.
So if we do:
```rst
2.13
###
2.13.0
-----
New checkers
^^^^^^^^^^^
```

We will get different `<h>` tags. That should be fairly easy to do with a replace all in the changelog files.

Pierre-Sassoulas:
Let's work the structure on 2.13 only then when we're satisfied let's do the rest.

Pierre-Sassoulas:
I pushed a new version with only 2.13 modified, not very satisfied to be honest, let me know what you think :).

DanielNoord:
I think this is definitely an improvement. Only thing I would change is to remove this page:
https://pylint--6688.org.readthedocs.build/en/6688/whatsnew/2/2.13/index.html
And make that go to:
https://pylint--6688.org.readthedocs.build/en/6688/whatsnew/2/2.13/content.html
immediately.

Pierre-Sassoulas:
Do you know how we can still have the dropdown for the link ? (We have a flat link and no navigation in the left toc bar if we do that.

DanielNoord:
@Pierre-Sassoulas I think a `:hidden:` `toctree` directive on the `content.html` page should display in the index. The issue is that `content.html` doesn't have `toctree` so Sphinx doesn't know what to put as links there I think.

Pierre-Sassoulas:
All right, I fixed 2.13 and modified 2.12 too, I prefer the 2.12 version because the full changelog and summary are separated like they were before. I think otherwise it's identical. Let me know what you think :)

Pierre-Sassoulas:
I think we should merge the two first point of the TODO so we don't have rebase conflict when something new end up in 2.14 changelog while we figure out the new changelog.

DanielNoord:
> All right, I fixed 2.13 and modified 2.12 too, I prefer the 2.12 version because the full changelog and summary are separated like they were before (It's handy when there is a duplicated link so we don't have to modify the current changelog to avoid an error, but also make sense as they were supposed to be separated). I think otherwise it's identical. Let me know what you think :)

I don't really see the difference between `2.13` and `2.12`?

I was thinking of removing the `index.html` page and only keep `content.html` is that a possibility you think?

Pierre-Sassoulas:
> I don't really see the difference between 2.13 and 2.12?

In 2.12 the full changelog and what's new's summary are each on their own page. (In 2.13 I had to merge the two and fix an issue with a link that was duplicated).

> I was thinking of removing the index.html page and only keep content.html is that a possibility you think?

It's making a flat link on the left TOC, I've tried for maybe 30mn before giving up 😄

DanielNoord:
> In 2.12 the full changelog and what's new's summary are each on their own page. (In 2.13 I had to merge the two and fix an issue with a link that was duplicated).

I like the single page better. Whenever you're opening old changelogs you're likely looking for something specific. So being Abel to search on one page instead of two is better imo.

> It's making a flat link on the left TOC, I've tried for maybe 30mn before giving up 😄

Hm, that's annoying. I might have a look as well.

Pierre-Sassoulas:
> So being Abel to search on one page instead of two is better imo.

But the full changelog contain everything, right ? So you can search in the full changelog only and get your result, (or on the full website with the global search)

DanielNoord:
> > So being Abel to search on one page instead of two is better imo.
> 
> But the full changelog contain everything, right ? So you can search in the full changelog only and get your result, (or on the full website with the global search)

Yeah, I just don't understand while we would want to separate related information across multiple pages/files. With this change you also lose the ability to search within the `CHANGELOG` file and would need to do that with your IDE or by grepping. I consider that a little less user-friendly.
But if you want to keep separating them let's do that then.

Pierre-Sassoulas:
> Yeah, I just don't understand while we would want to separate related information across multiple pages/files.

- Keeping thing separated means adding a section and changing the header level which is quite fast. Merging means reorganizing the files, copy pasting to merge the file, changind the header level, removing the old file (there will be conflict for 2.14 if something is merged meanwhile on main), then there can be issues when we merge two files together (and we have 18 files to merge). It's a lot of work
- It's possible to search on ReadtheDoc directly and dev probably have an IDE anyway
- Keeping thing separated make historic sense and the data would be more organized at the end.

I'm going to apply 2.12 example everywhere and open another MR for this so we can merge asap an stop having conflict on ``Changelog``/``2.14/full.rst``

DudeNr33:
Besides the discussion about the structure of the changelog, did you consider generating the content automatically?
I googled a bit and found for example a [Sphinx extension](https://sphinx-github-changelog.readthedocs.io/en/stable/) as well as a [Github Action](https://github.com/marketplace/actions/generate-changelog). 
From those two the Github Action sounds more promising as it is more configurable. It only outputs Markdown files, but those can be [easily converted to reStructuredText](https://github.com/github-changelog-generator/github-changelog-generator/issues/528).

And if that does not satisfy our needs we could write an own simple Sphinx extension which uses the Github REST Api directly (e.g. through [PyGithub](https://github.com/PyGithub/PyGithub)).

Benefits from generating the future changelogs automatically would be:

* No more merge conflicts in the changelog/whatsnew files
* Structure and filtering can be enforced more easily based on labels
* Structure can be changed later on without having to modify / move each entry by hand

If you are interested, I would be willing to give it a try.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -0,0 +1,50 @@
...
+- ...
+- ...
+
+False positives fixed
+=====================
+- ...
+
+False negatives fixed
+=====================
+- ...
```
jacobtylerwalls, line 41: I like this, but I would put Bug fixes third and call it "Other bug fixes" to help new contributors.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -0,0 +1,50 @@
+***************************
+ What's New in Pylint 2.15
+***************************
+
+.. toctree::
+   :maxdepth: 2
+
+:Release: 2.14
```
jacobtylerwalls, line 8:
```suggestion
:Release: 2.15
```

jacobtylerwalls:
+1 on separating the files, I found https://github.com/PyCQA/pylint/pull/6688#issuecomment-1140172495 convincing.

DanielNoord:
> +1 on separating the files, I found [#6688 (comment)](https://github.com/PyCQA/pylint/pull/6688#issuecomment-1140172495) convincing.

Also for the new changelogs? I'm fine with separating in #6688, but would prefer one file from now on. It's just a lot easier and more manageable.

Pierre-Sassoulas:
I was thinking of treating release notes according to their version:
- pre 1.0 (We do nothing, the issues number are logilab's internal)
- pre 1.6 (We only have the full changelog, we check correctness)
- pre 2.14 (Separated and, we check correctness)
- post 2.14:
  - all in a single file, in the right sections (sections and organisation of the file to be determined)
  - We check for correctness
  - We enforce by CI that a properly formatted line with the PR number (not the original issue number) exists
  - If the PR is labelled maintenance, primer or documentation in github then nothing is required

Pierre-Sassoulas:
Blocked by #6735

doc/whatsnew/2/2.15/index.rst
```diff
@@ -42,3 +42,5 @@ Other Changes
 
 Internal changes
 ================
+
+* Added a check in CI for changelogs entry (#6688)
```
Pierre-Sassoulas, line 46: The question here is should we keep the old format with:
```
* Fix a regression where messages with dash are not fully parsed

  Closes #3604
```
Or is ``* Fix a regression where messages with dash are not fully parsed (#3604)`` better ? Sometime we might want to be more detailed use multiple line ? Note that #3604 was an issue number but we would use the PR number now (It was just a commit at the time https://github.com/PyCQA/pylint/commit/ffb354aea057c25d9e48fa22da2840a450d99f3e but now we can't push to main so there's always a PR)

Pierre-Sassoulas:
The follow-up will be #6781 but adding the CI check asap is important imo.

DanielNoord:
@Pierre-Sassoulas I'll review this in a minute, but could you look into https://github.com/PyCQA/pylint/pull/6787#issuecomment-1143779977
`pre-commit` is broken locally and I can't commit anything without `no-verifying`.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,24 @@
+name: changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled, reopened]
```
DanielNoord, line 5: Is there anything we exclude here? Can't it just be `on: pull_request`?

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,24 @@
...
+
+permissions:
+  contents: read
```
DanielNoord, line 8: I believe this is automatic.

doc/whatsnew/2/2.15/index.rst
DanielNoord: Let's keep the old system. It brings some uniformity to all changelogs and is also easier to parse the data programatically.

.github/workflows/changelog.yml
Pierre-Sassoulas: It needs to have labeled and unlabeled because if we add or remove "skip news" the pipeline result is not the same.

Pierre-Sassoulas:
I'm closing this in favor of #6875. @DudeNr33 you could take some inspiration [from the changelog pipeline](https://github.com/PyCQA/pylint/pull/6688/files#diff-a0aba5a2bd30b2f80a3a18aa24ce43670b1320ca0a8b9c103bb3bc971ae30a85R1), I think instead of a regex to check on the changelog we need to check that the description exists in the issue but everything around it could be the same.

PR #6691, opened 2022-05-24
Allow disabling `bad-option-value` and `unrecognized-option` with `--disable=all`

### LINKED ISSUE #6565: Allow disabling `bad-option-value` and `unrecognized-option` with `--disable=all`
We should allow `--disable=all` to disable the messages emitted for a `pylintrc`, e.g. `bad-option-value` and `unrecognized-option`, as it's slightly unexpected to not be able to disable everything.

The `pylintrc` used in the following run is found [here](https://github.com/cuthbertLab/music21/blob/master/.pylintrc).

```python
% pylint music21 --disable=all 
************* Module /Users/.../music21/.pylintrc
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message bad-whitespace. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message bad-continuation. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message no-self-use. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message star-args. (bad-option-value)
************* Module /Users/.../music21/.pylintrc
.pylintrc:1:0: E0015: Unrecognized option found: profile, files-output, bad-functions, function-name-hint, variable-name-hint, const-name-hint, attr-name-hint, argument-name-hint, class-attribute-name-hint, inlinevar-name-hint, class-name-hint, module-name-hint, method-name-hint, no-space-check (unrecognized-option)
```

_Originally posted by @jacobtylerwalls in https://github.com/PyCQA/pylint/pull/6556#pullrequestreview-966841907_

jacobtylerwalls:
## Type of Changes

## Description
Previously, due to the order in which options are parsed (config file, then command line), any unrecognized or bad options in the config file would emit messages unconditionally, even if suppressed on the CLI.

Closes #6565

### DIFF
--- a/pylint/config/config_initialization.py
+++ b/pylint/config/config_initialization.py
@@ -58,11 +58,6 @@ def _config_initialization(
     except _UnrecognizedOptionError as exc:
         unrecognized_options_message = ", ".join(exc.options)
 
-    bad_option_value_messages_before_parsing_cli = (
-        linter.stashed_bad_option_value_messages[:]
-    )
-    linter.stashed_bad_option_value_messages = []
-
     # Then, if a custom reporter is provided as argument, it may be overridden
     # by file parameters, so we re-set it here. We do this before command line
     # parsing, so it's still overridable by command line options
@@ -96,11 +91,7 @@ def _config_initialization(
             "unrecognized-option", args=unrecognized_options_message, line=0
         )
 
-    linter._emit_bad_option_value(
-        messages_from_config_file=bad_option_value_messages_before_parsing_cli,
-        messages_from_cli=linter.stashed_bad_option_value_messages,
-    )
-    linter.stashed_bad_option_value_messages = []
+    linter._emit_bad_option_value()
 
     # Set the current module to include command line as we don't know where
     # the --load-plugins key is coming from
--- a/pylint/config/config_initialization.py
+++ b/pylint/config/config_initialization.py
@@ -58,6 +58,10 @@ def _config_initialization(
     except _UnrecognizedOptionError as exc:
         unrecognized_options_message = ", ".join(exc.options)
 
+    bad_option_value_messages_before_parsing_cli = set(
+        linter.stashed_bad_option_value_messages
+    )
+
     # Then, if a custom reporter is provided as argument, it may be overridden
     # by file parameters, so we re-set it here. We do this before command line
     # parsing, so it's still overridable by command line options
@@ -90,12 +94,19 @@ def _config_initialization(
         linter.add_message(
             "unrecognized-option", args=unrecognized_options_message, line=0
         )
+    bad_option_value_messages_from_cli = {
+        msg
+        for msg in linter.stashed_bad_option_value_messages
+        if msg not in bad_option_value_messages_before_parsing_cli
+    }
+    linter._emit_bad_option_value(
+        messages_from_config_file=bad_option_value_messages_before_parsing_cli,
+        messages_from_cli=bad_option_value_messages_from_cli,
+    )
 
     # Set the current module to include command line as we don't know where
     # the --load-plugins key is coming from nor the bad-option-value messages
     linter.set_current_module("Command line or configuration file")
-
-    linter._emit_bad_option_value()
 
     # We have loaded configuration from config file and command line. Now, we can
     # load plugin specific configuration.
--- a/pylint/config/config_initialization.py
+++ b/pylint/config/config_initialization.py
@@ -90,11 +90,12 @@ def _config_initialization(
         linter.add_message(
             "unrecognized-option", args=unrecognized_options_message, line=0
         )
-    linter._emit_bad_option_value()
 
-    # Set the current module to configuration as we don't know where
-    # the --load-plugins key is coming from
+    # Set the current module to include command line as we don't know where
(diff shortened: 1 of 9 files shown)


Pierre-Sassoulas (commented):
I think we could have a mechanism in place to signal that the configuration is parsed (boolean config_is_parsed ?) and use a stash directly in ``add_message`` for **any** message if the configuration is not parsed yet.

pylint/lint/pylinter.py
```diff
@@ -283,6 +283,14 @@ def __init__(
         self.msgs_store = MessageDefinitionStore()
         self.msg_status = 0
         self._by_id_managed_msgs: list[ManagedMessage] = []
+        self.stashed_bad_option_value_messages: list[str] = []
```
DanielNoord, line 286: This should probably go in `_MessageStateHandler`.

jacobtylerwalls:
> I think we could have a mechanism in place to signal that the configuration is parsed (boolean config_is_parsed ?) and use a stash directly in add_message for any message if the configuration is not parsed yet.

I guess I didn't want to generalize it because the use case is so niche and we don't want to encourage or suggest using it for other messages (these are the only messages that it makes sense to emit before configuration is finished). There's also the matter of needing to stashing the location if we generalize it, which increases complexity (and might needlessly change the signature of `add_message`.)

DanielNoord (commented):
Well done getting this to work @jacobtylerwalls 

Just one question:
Not sure if you thought about this but what about making `stashed_bad_option_value_messages` a dictionary with keys reflecting the current module. We could use `linter.current_name` to look up the correct name in the dictionary to put the message under and then just iterate over the keys in `emit_bad_option_value` (and reset `current_name` there.

At some point I want to be able to parse multiple configuration files at once to create different config namespaces for sub-directories. So, just `messages_from_config_file` might not be enough as we would need access to the different config file names.

Besides it would remove the need for L61-64 which is arguably a bit hacky. (The need for `set_module` before `add_message` is horrible anyway).


Another thing is that `stashed_bad_option_value_messages` should probably be `_stashed_bad_option_value_messages` to avoid any future deprecation issues.

jacobtylerwalls:
Nice. Yeah I went that direction and paused as I wasn't sure of the use case, so that's helpful to hear it would be used! Will do.

pylint/config/config_initialization.py
```diff
@@ -83,7 +83,17 @@ def _config_initialization(
         msg = ", ".join(unrecognized_options)
         linter._arg_parser.error(f"Unrecognized option found: {msg}")
 
-    # Set the current module to configuration as we don't know where
+    # Now that config file and command line options have been loaded
+    # with all disables, it is safe to emit messages
+    linter.set_current_module(str(config_file) if config_file else None)
```
DanielNoord, line 88: We have deprecated passing `None` to `set_current_module`. So it should probably be `""`.

pylint/config/config_initialization.py
```diff
@@ -83,7 +83,17 @@ def _config_initialization(
         msg = ", ".join(unrecognized_options)
         linter._arg_parser.error(f"Unrecognized option found: {msg}")
 
-    # Set the current module to configuration as we don't know where
+    # Now that config file and command line options have been loaded
+    # with all disables, it is safe to emit messages
+    linter.set_current_module(str(config_file) if config_file else "")
+    if unrecognized_options_message is not None:
```
DanielNoord, line 89:
```suggestion
    if unrecognized_options_message is not None:
        linter.set_current_module(str(config_file) if config_file else "")
```

A small optimisation 😄

pylint/lint/message_state_handler.py
```diff
@@ -53,6 +54,16 @@ def __init__(self, linter: PyLinter) -> None:
             "enable-msg": self._options_methods["enable"],
         }
         self._pragma_lineno: dict[str, int] = {}
+        self._stashed_bad_option_value_messages: defaultdict[
```
DanielNoord, line 57:
```suggestion
        TODO: 3.0: Update key type to str when current_name is always str
        self._stashed_bad_option_value_messages: defaultdict[
```

DanielNoord (approved):

pylint/config/config_initialization.py
jacobtylerwalls: Good catch, now that we set module locations in `_emit_bad_option_value()`

Pierre-Sassoulas (approved):
LGTM :)

pylint/config/config_initialization.py
```diff
@@ -83,7 +83,17 @@ def _config_initialization(
...
+    # with all disables, it is safe to emit messages
+    if unrecognized_options_message is not None:
+        linter.set_current_module(str(config_file) if config_file else "")
+        linter.add_message(
+            "unrecognized-option", args=unrecognized_options_message, line=0
+        )
+
+    linter._emit_bad_option_value()
+
+    # Set the current module to include command line as we don't know where
```
jacobtylerwalls, line 96: I can revert back to the original comment after the latest edits, as it now makes sense again.
```suggestion
    # Set the current module to configuration as we don't know where
```

PR #6693, opened 2022-05-24
Add support of sharing message in multiple checkers. Fix DeprecatedChecker example

### LINKED ISSUE #6656: deprecation_checker example is not working
### Bug description

Trying to implement an checker based on the `deprecation_checker` example gives the following error when loading the plugin:

```
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "<path to venv>/lib/python3.10/site-packages/pylint/__main__.py", line 10, in <module>
    pylint.run_pylint()
  File "<path to venv>/lib/python3.10/site-packages/pylint/__init__.py", line 22, in run_pylint
  File "<path to venv>/lib/python3.10/site-packages/pylint/lint/run.py", line 280, in __init__
    linter.load_plugin_modules(self._plugins)
  File "<path to venv>/lib/python3.10/site-packages/pylint/lint/pylinter.py", line 645, in load_plugin_modules
    module.register(self)
  File "<path to cwd>/pylint_deprecated_checker.py", line 36, in register
    linter.register_checker(DeprecationChecker(linter))
  File "<path to venv>/lib/python3.10/site-packages/pylint/lint/pylinter.py", line 773, in register_checker
    self.msgs_store.register_messages_from_checker(checker)
  File "<path to venv>/lib/python3.10/site-packages/pylint/message/message_definition_store.py", line 39, in register_messages_from_checker
    checker.check_consistency()
  File "<path to venv>/lib/python3.10/site-packages/pylint/checkers/base_checker.py", line 133, in check_consistency
    raise InvalidMessageError(error_msg)
pylint.exceptions.InvalidMessageError: Inconsistent checker part in message id 'W1505' (expected 'x04xx' because we already had ['W0402']).
```

Checker code (placed in `pylint_deprecated_checker.py`; I removed comments and some test prints, so the lines won't match the above traceback):
```py

from pylint.checkers import BaseChecker, DeprecatedMixin

if TYPE_CHECKING:


class DeprecationChecker(DeprecatedMixin, BaseChecker):
    name = "deprecated-function-call"

    def deprecated_methods(self) -> set[str]:
        return set()

    def deprecated_arguments(self, method: str) -> tuple[tuple[int | None, str], ...]:
        return ()

def register(linter: 'PyLinter') -> None:
    linter.register_checker(DeprecationChecker(linter))
```

### Configuration

```ini
[MAIN]
ignore-patterns=.*?_test.py

disable=all

enable=
    deprecated-method

jobs=0
suggestion-mode=yes

fail-under=9.0
fail-on=deprecated-method
```

### Command used

```shell
PYTHONPATH="." python -m pylint path_to_code --load-plugins=pylint_deprecated_checker --rcfile=pylintrc
```

### Pylint output

```shell
See bug description.
```

### Expected behavior

I expected the checker to initialize correctly.

### Pylint version

```shell
pylint 2.13.9
astroid 2.11.5
Python 3.10.4 (main, Mar 23 2022, 23:05:40) [GCC 11.2.0]
```

### OS / Environment

Arch Linux

### Additional dependencies

_No response_

matusvalo:
## Type of Changes

## Description

This PR move message definitions from `msgs` attribute of DeprecatecMixin to separate class attributes. `msgs` attribute cannot be used directly anyway because it mixes two different message classes: `W15XX` and `W04XX`.

Closes #6656

### DIFF
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -7,7 +7,7 @@
 import os
 from collections import defaultdict
 from inspect import getmodule
-from itertools import chain
+from itertools import chain, groupby
 from pathlib import Path
 from typing import DefaultDict, Dict, List, NamedTuple, Optional, Tuple
 
@@ -43,6 +43,7 @@ class MessageData(NamedTuple):
     related_links: str
     checker_module_name: str
     checker_module_path: str
+    shared: bool = False
 
 
 MessagesDict = Dict[str, List[MessageData]]
@@ -134,6 +135,12 @@ def _get_all_messages(
         ((checker, msg) for msg in checker.messages)
         for checker in linter.get_checkers()
     )
+    shared_msg_ids = set(
+        chain.from_iterable(
+            [checker.shared_message_ids for checker in linter.get_checkers()]
+        )
+    )
+
     for checker, message in checker_message_mapping:
         message_data_path = (
             PYLINT_MESSAGES_DATA_PATH / message.symbol[0] / message.symbol
@@ -157,15 +164,19 @@ def _get_all_messages(
             related,
             checker_module.__name__,
             checker_module.__file__,
+            message.msgid in shared_msg_ids,
         )
         msg_type = MSG_TYPES_DOC[message.msgid[0]]
         messages_dict[msg_type].append(message_data)
         if message.old_names:
             for old_name in message.old_names:
                 category = MSG_TYPES_DOC[old_name[0][0]]
-                old_messages[category][(old_name[1], old_name[0])].append(
-                    (message.symbol, msg_type)
-                )
+                if (message.symbol, msg_type) not in old_messages[category][
+                    (old_name[1], old_name[0])
+                ]:
+                    old_messages[category][(old_name[1], old_name[0])].append(
+                        (message.symbol, msg_type)
+                    )
 
     return messages_dict, old_messages
 
@@ -196,19 +207,26 @@ def _write_message_page(messages_dict: MessagesDict) -> None:
         if not category_dir.exists():
             category_dir.mkdir(parents=True, exist_ok=True)
         for message in messages:
+            if message.shared:
+                continue
             if not _message_needs_update(message, category):
                 continue
             _write_single_message_page(category_dir, message)
+        for _, shared_messages in groupby(
+            sorted(
+                (message for message in messages if message.shared), key=lambda m: m.id
+            ),
+            key=lambda m: m.id,
+        ):
+            shared_messages_list = list(shared_messages)
+            if len(shared_messages_list) > 1:
+                _write_single_shared_message_page(category_dir, shared_messages_list)
+            else:
+                _write_single_message_page(category_dir, shared_messages_list[0])
 
 
-def _write_single_message_page(category_dir: Path, message: MessageData) -> None:
-    checker_module_rel_path = os.path.relpath(
-        message.checker_module_path, PYLINT_BASE_PATH
-    )
-    messages_file = os.path.join(category_dir, f"{message.name}.rst")
-    with open(messages_file, "w", encoding="utf-8") as stream:
-        stream.write(
-            f""".. _{message.name}:
+def _generate_single_message_body(message: MessageData) -> str:
+    body = f""".. _{message.name}:
 
 {get_rst_title(f"{message.name} / {message.id}", "=")}
 **Message emitted:**
@@ -224,19 +242,44 @@ def _write_single_message_page(category_dir: Path, message: MessageData) -> None
 {message.details}
 {message.related_links}
 """
-        )
-        if message.checker_module_name.startswith("pylint.extensions."):
-            stream.write(
-                f"""
+    if message.checker_module_name.startswith("pylint.extensions."):
+        body += f"""
 .. note::
   This message is emitted by the optional :ref:`'{message.checker}'<{message.checker_module_name}>` checker which requires the ``{message.checker_module_name}``
   plugin to be loaded.
 
 """
-            )
-        checker_url = (
-            f"https://github.com/PyCQA/pylint/blob/main/{checker_module_rel_path}"
+    return body
+
+
+def _generate_checker_url(message: MessageData) -> str:
+    checker_module_rel_path = os.path.relpath(
+        message.checker_module_path, PYLINT_BASE_PATH
+    )
+    return f"https://github.com/PyCQA/pylint/blob/main/{checker_module_rel_path}"
+
+
+def _write_single_shared_message_page(
+    category_dir: Path, messages: List[MessageData]
+) -> None:
+    message = messages[0]
+    messages_file = os.path.join(category_dir, f"{message.name}.rst")
+    with open(messages_file, "w", encoding="utf-8") as stream:
+        stream.write(_generate_single_message_body(message))
+        checker_urls = ", ".join(
+            [
+                f"`{message.checker} <{_generate_checker_url(message)}>`__"
+                for message in messages
+            ]
         )
+        stream.write(f"Created by the {checker_urls} checkers.")
+
+
+def _write_single_message_page(category_dir: Path, message: MessageData) -> None:
+    messages_file = os.path.join(category_dir, f"{message.name}.rst")
+    with open(messages_file, "w", encoding="utf-8") as stream:
+        stream.write(_generate_single_message_body(message))
+        checker_url = _generate_checker_url(message)
         stream.write(f"Created by the `{message.checker} <{checker_url}>`__ checker.")
 
 
@@ -271,7 +314,11 @@ Pylint can emit the following messages:
             "refactor",
             "information",
         ):
-            messages = sorted(messages_dict[category], key=lambda item: item.name)
+            # We need remove all duplicated shared messages
+            messages = sorted(
+                {msg.id: msg for msg in messages_dict[category]}.values(),
+                key=lambda item: item.name,
+            )
             old_messages = sorted(old_messages_dict[category], key=lambda item: item[0])
             messages_string = "".join(
                 f"   {category}/{message.name}\n" for message in messages
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -260,8 +260,7 @@ def _write_single_shared_message_page(
     category_dir: Path, messages: List[MessageData]
 ) -> None:
     message = messages[0]
-    messages_file = os.path.join(category_dir, f"{message.name}.rst")
-    with open(messages_file, "w", encoding="utf-8") as stream:
+    with open(category_dir / f"{message.name}.rst", "w", encoding="utf-8") as stream:
         stream.write(_generate_single_message_body(message))
         checker_urls = ", ".join(
             [
@@ -273,8 +272,7 @@ def _write_single_shared_message_page(
 
 
 def _write_single_message_page(category_dir: Path, message: MessageData) -> None:
-    messages_file = os.path.join(category_dir, f"{message.name}.rst")
-    with open(messages_file, "w", encoding="utf-8") as stream:
+    with open(category_dir / f"{message.name}.rst", "w", encoding="utf-8") as stream:
         stream.write(_generate_single_message_body(message))
         checker_url = _generate_checker_url(message)
         stream.write(f"Created by the `{message.checker} <{checker_url}>`__ checker.")
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -135,11 +135,6 @@ def _get_all_messages(
         ((checker, msg) for msg in checker.messages)
         for checker in linter.get_checkers()
     )
-    shared_msg_ids = set(
-        chain.from_iterable(
-            [checker.shared_message_ids for checker in linter.get_checkers()]
-        )
-    )
 
     for checker, message in checker_message_mapping:
         message_data_path = (
@@ -164,7 +159,7 @@ def _get_all_messages(
             related,
             checker_module.__name__,
             checker_module.__file__,
-            message.msgid in shared_msg_ids,
+            message.shared,
         )
         msg_type = MSG_TYPES_DOC[message.msgid[0]]
         messages_dict[msg_type].append(message_data)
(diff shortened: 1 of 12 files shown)


Pierre-Sassoulas (commented):
Would it be possible to add a short example in the developer guide, please ? I think this one checker is very likely to be customized. (Maybe we could provide a way to just configure what is deprecated or not in the configuration but that's probably for another MR)

matusvalo:
> Would it be possible to add a short example in the developer guide, please ?

I can. Should be done in this MR or should I create new one?

Moreover I think, we should include in our tests execution of the examples to avoid having them outdated and not running...

Pierre-Sassoulas:
Related to your last point about checking the exemples of the doc we'd need some kind of structure like for bad.py / good.py, but this is a really neat idea !

pylint/checkers/deprecated.py
```diff
@@ -26,12 +26,16 @@
...
+    DEPRECATED_IMPORT_MSGS: dict[str, MessageDefinitionTuple] = {
+        "W0402": (
+            "Deprecated module %r",
+            "deprecated-module",
+            "A module marked as deprecated is imported.",
+        ),
+    }
 
-    msgs: dict[str, MessageDefinitionTuple] = {
+    DEPRECATED_MSGS: dict[str, MessageDefinitionTuple] = {
```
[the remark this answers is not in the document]
matusvalo: This is a question how DeprecatedMixin is supposed to be used:
* should external users reuse DeprecatedMixin?
  * if yes should they use MSGS which are used also in STDLIB checker (W15XX) and import checker (W0402)? If W15XX and W0402 are reused also for external users the **renaming is confusing then** because it is not connected to STDLIB only.
  * if no we should somehow parametrize the mixin allowing users to specify which message should be raised
* if external users should not reuse DeprecatedMixin we can do the renaming.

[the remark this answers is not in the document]
Pierre-Sassoulas: > should external users reuse DeprecatedMixin? 

I'm wondering why we do not have a configuration for deprecations checks ? Isn't it possible to have user modifiable ``deprecated_functions``, ``deprecated_classes`` list options ?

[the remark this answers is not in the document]
matusvalo: > I'm wondering why we do not have a configuration for deprecations checks ? Isn't it possible to have user modifiable deprecated_functions, deprecated_classes list options ?

We can provide to the user these options. Should it go with this PR?

In any case I think some users will need subclassing `DeprecatedMixin`. By subclassing mixin user can check the deprecated function also based on version - e.g. django plugin can check deprecated django functions/classes based on django version. This cannot be done by simple list of functions/modules in configuration file.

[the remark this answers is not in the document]
Pierre-Sassoulas: > We can provide to the user these options. Should it go with this PR?

Let's first handle this PR.

> By subclassing mixin user can check the deprecated function also based on version - e.g. django plugin can check deprecated django functions/classes based on django version.

Ha right, it make sense.

DanielNoord:
@matusvalo What about moving these messages to (for example) `49xx` and special casing `49xx` in the CI check to be mixed with other prefixes? Wouldn't that also solve the problem?

matusvalo:
> What about moving these messages to (for example) `49xx` and special casing `49xx` in the CI check to be mixed with other prefixes? Wouldn't that also solve the problem?

@DanielNoord unfortunately I don't understand. :-( What is CI?

DanielNoord:
> > What about moving these messages to (for example) `49xx` and special casing `49xx` in the CI check to be mixed with other prefixes? Wouldn't that also solve the problem?
> 
> @DanielNoord unfortunately I don't understand. :-( What is CI?

Continuous Integration. The checks we run with Github Actions.

The reason for the weird dictionary creation is because we have a check that makes sure that all messages in a checker have the same first two digits. This is what is causing issues for this `Mixin`. It's defining messages for different checkers that have different "prefixes" (those two first digits).
My idea is to create a prefix for the mixing (let's say `49`) and then allow that prefix to be mixed with other prefixes in checkers. That would remove the need for the double dictionaries and still pass all tests.

matusvalo:
>My idea is to create a prefix for the mixing (let's say 49) and then allow that prefix to be mixed with other prefixes in checkers. That would remove the need for the double dictionaries and still pass all tests.

I understand now. But I have a two questions:
1. stdlib checker and import checker should be migrated to the new codes of mixin? If yes it will break backward compatibility (e.g. msg ID `W0402` for `deprecated-module` will change to `W49XY`).
2. other posibility is to leave `deprecated-module`/`W0402` in stdlib checker and create new on `derecated-module2`/`W4902`. I don't like it because it is confusing for users (why we have multiple messages for `deprecated-module`? Again maybe we can have 2 checks with same name with different msg ID but this will be totally confusing

Hence, for now I did most safest change:
1. I have added shared_messages dictionary where you have messages which are not checked by consistency check.
2. I have left all messages from stdlib checker/import checker in `DeprecatedMixin`
The change is backward compatible and user friendly. The only issue is that the Message IDs are coming from different checkers (not ideal better to have special msg IDs for the mixin), but I consider it a lesser evil comparing to previous 2 points/questions.

matusvalo:
Hm CI is failing due inconsistent Messages. Now I see why we were talking about CI... :-/

DanielNoord:
> 1. stdlib checker and import checker should be migrated to the new codes of mixin? If yes it will break backward compatibility (e.g. msg ID `W0402` for `deprecated-module` will change to `W49XY`).

Let's do this with the `old_names` setup. You can look at `missing-class-docstring` to see how that works normally. That is a safe way to rename messages to new ids.

matusvalo:
OK documentation now should be able to build. Currently it is done that for shared messages, the list of checkers having this message is created in documentation. Now the unittest checking uniqueness of message is failing. I will fix it in next round of changes in this PR...

pylint/checkers/base_checker.py
```diff
@@ -45,6 +45,9 @@ class BaseChecker(_ArgumentsProvider):
     reports: tuple[tuple[str, str, ReportsCallable], ...] = ()
     # mark this checker as enabled or not.
     enabled: bool = True
+    # Set of messages allowed to be reused in multiple checkers.
+    # Hence, the Message ID does not need to match Checker ID.
+    shared_message_ids: set[str] = set()
```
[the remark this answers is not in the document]
matusvalo: `shared_message_ids` was replaced with `shared` option of the message.

matusvalo:
@DanielNoord @Pierre-Sassoulas  please have a look to PR:
1. The Deprecated messages can be now shared between Checkers using `shared` option.
2. The sphinx plugin now can detect shared messages and add the links to the documentation of the messages
3. The deprecated example was updated and now is working

doc/exts/pylint_messages.py
```diff
@@ -157,15 +159,19 @@ def _get_all_messages(
...
         )
         msg_type = MSG_TYPES_DOC[message.msgid[0]]
         messages_dict[msg_type].append(message_data)
         if message.old_names:
             for old_name in message.old_names:
                 category = MSG_TYPES_DOC[old_name[0][0]]
-                old_messages[category][(old_name[1], old_name[0])].append(
-                    (message.symbol, msg_type)
-                )
+                if (message.symbol, msg_type) not in old_messages[category][
```
DanielNoord, line 169:
```suggestion
                # We check if the message is already in old_messages so
                # we don't duplicate shared messages.
                if (message.symbol, msg_type) not in old_messages[category][
```

doc/exts/pylint_messages.py
```diff
@@ -271,7 +309,11 @@ def _write_messages_list_page(
             "refactor",
             "information",
         ):
-            messages = sorted(messages_dict[category], key=lambda item: item.name)
+            # We need remove all duplicated shared messages
```
DanielNoord, line 312:
```suggestion
            # We need to remove all duplicated shared messages
```

doc/exts/pylint_messages.py
```diff
@@ -7,7 +7,7 @@
 import os
```
DanielNoord, line 7: This all seems to work.

The only thing I would add is to add all checkers that can emit a message to the page. So for these messages both `stdlib` and `imports`. But that can also be done in a follow-up.

pylint/checkers/deprecated.py
```diff
@@ -24,41 +24,58 @@
     nodes.ClassDef,
 )
 
+DEPRECATED_MODULE_MESSAGE: dict[str, MessageDefinitionTuple] = {
```
DanielNoord, line 27: Perhaps we should make this class variables? So they don't need to be imported together with the mixin.

pylint/message/message_definition.py
```diff
@@ -25,23 +26,21 @@ def __init__(
         msg: str,
         description: str,
         symbol: str,
-        scope: str,
-        minversion: tuple[int, int] | None = None,
-        maxversion: tuple[int, int] | None = None,
-        old_names: list[tuple[str, str]] | None = None,
+        extra_options: ExtraMessageOptions,
```
DanielNoord, line 29: Is this public/private? I'm not sure we can change this signature without a changelog warning.

DanielNoord (changes requested):
Some early feedback!

Thanks for all the work you put in to this @matusvalo!

Pierre-Sassoulas (commented):
LGTM, amazing work ! It will be useful to make the checker more modular. I wonder if you could create a small explanation on how to do that in the contributor doc on checkers ?

doc/exts/pylint_messages.py
```diff
@@ -224,19 +239,44 @@ def _write_single_message_page(category_dir: Path, message: MessageData) -> None
...
+    )
+    return f"https://github.com/PyCQA/pylint/blob/main/{checker_module_rel_path}"
+
+
+def _write_single_shared_message_page(
+    category_dir: Path, messages: List[MessageData]
+) -> None:
+    message = messages[0]
+    messages_file = os.path.join(category_dir, f"{message.name}.rst")
+    with open(messages_file, "w", encoding="utf-8") as stream:
```
DanielNoord, line 264:
```suggestion
    with open(category_dir / f"{message.name}.rst", "w", encoding="utf-8") as stream:
```

doc/exts/pylint_messages.py
```diff
@@ -224,19 +239,44 @@ def _write_single_message_page(category_dir: Path, message: MessageData) -> None
...
+                f"`{message.checker} <{_generate_checker_url(message)}>`__"
+                for message in messages
+            ]
         )
+        stream.write(f"Created by the {checker_urls} checkers.")
+
+
+def _write_single_message_page(category_dir: Path, message: MessageData) -> None:
+    messages_file = os.path.join(category_dir, f"{message.name}.rst")
+    with open(messages_file, "w", encoding="utf-8") as stream:
```
DanielNoord, line 277:
```suggestion
    with open(category_dir / f"{message.name}.rst", "w", encoding="utf-8") as stream:
```

examples/deprecation_checker.py
```diff
@@ -51,14 +51,20 @@ def mymethod(self, arg0, arg1, deprecated1=None, arg2='foo', deprecated2='bar',
 class DeprecationChecker(DeprecatedMixin, BaseChecker):
     """Class implementing deprecation checker."""
 
-    # DeprecationMixin class is Mixin class implementing logic for searching deprecated methods and functions.
+    # DeprecatedMixin class is Mixin class implementing logic for searching deprecated methods and functions.
     # The list of deprecated methods/functions is defined by implementing class via deprecated_methods callback.
```
DanielNoord, line 55:
```suggestion
    # The list of deprecated methods/functions is defined by the implementing class via deprecated_methods callback.
```

pylint/checkers/base_checker.py
```diff
@@ -174,6 +174,8 @@ def check_consistency(self) -> None:
         checker_id = None
         existing_ids = []
         for message in self.messages:
+            if message.shared:
```
DanielNoord, line 177:
```suggestion
            # Id's for shared messages such as the 'deprecated-*' messages
            # can be inconsistent with their checker id.
            if message.shared:
```

pylint/message/message_definition.py
```diff
@@ -18,6 +18,8 @@
 
 
 class MessageDefinition:
+    # TODO: remove disable and move all extra option to single parameter.
```
DanielNoord, line 21:
```suggestion
```

Like I said in the issue. I don't think we should do this and just keep the extra disable.

tests/message/unittest_message_definition.py
```diff
@@ -64,15 +64,15 @@ def assert_with_fail_msg(msg: MessageDefinition, expected: bool = True) -> None:
...
         return MessageDefinition(
             FalseChecker(),
             "W1234",
             "message",
             "description",
             "msg-symbol",
             WarningScope.NODE,
-            **kwargs,
+            minversion=None,
+            maxversion=None,
```
DanielNoord, line 75:
```suggestion
```

Let's just make use of the default values.

DanielNoord (changes requested):
Final nitpicks. Thanks @matusvalo 👍

DanielNoord (changes requested):
https://github.com/PyCQA/pylint/pull/6693/files#r896124542
https://github.com/PyCQA/pylint/pull/6693/files#r896129061

@matusvalo I think you resolved these without answering the questions in them. Or am I overseeing your answers somewhere?

matusvalo:
> https://github.com/PyCQA/pylint/pull/6693/files#r896124542 https://github.com/PyCQA/pylint/pull/6693/files#r896129061
> 
> @matusvalo I think you resolved these without answering the questions in them. Or am I overseeing your answers somewhere?

@DanielNoord  My bad. I had pending review, for days and I missed that. I left you multiple messages waiting for publishing 🤦

DanielNoord:
> > https://github.com/PyCQA/pylint/pull/6693/files#r896124542 https://github.com/PyCQA/pylint/pull/6693/files#r896129061
> > @matusvalo I think you resolved these without answering the questions in them. Or am I overseeing your answers somewhere?
> 
> @DanielNoord My bad. I had pending review, for days and I missed that. I left you multiple messages waiting for publishing 🤦

Well it seems I was mistaken as well. The only thing I was worried about was https://github.com/PyCQA/pylint/pull/6693#discussion_r897567337 and you already fixed that before I even mentioned it.

Thanks for all the work here @matusvalo, LVGTM 👍

DanielNoord (approved):

matusvalo:
Thank you @DanielNoord  @Pierre-Sassoulas  for your review!

PR #6703, opened 2022-05-26
Cache and store primer packages via the default branch

DanielNoord:

## Type of Changes

## Description

Ref. https://github.com/PyCQA/pylint/issues/5364.

This is in preparation of a full fledged primer which can warn us about false positives and negatives.

This adds:
1) A workflow triggered on every push to `main` that caches and downloads all packages in `tests/primer/packages_to_prime.json`.
2) A workflow triggered on all relevant PRs that searches for the last successful run on `main` and gets its cache from there.

The cache is managed by what I call the `commitstring`, which is just a concatenation of all commits that the packages to prime are on.

Because this needs to be merged into `main` to work you can see a working example here:
Job on `main`:
https://github.com/DanielNoord/pylint/runs/6606171195?check_suite_focus=true
PR that gets triggered after that push:
https://github.com/DanielNoord/pylint/pull/145
Successful job on the PR:
https://github.com/DanielNoord/pylint/runs/6606228882?check_suite_focus=true

You can look at the restore cache jobs to see that this works.

In itself this PR doesn't really add much, but the extra run time is about 60 seconds for both jobs together and only about 20 seconds for a PR. Since the workflow for `primer` depends on stuff being already merged on `main` I'd like to do some preparation PRs to build on like this.

### DIFF
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -9,6 +9,8 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - doc/data/messages/**
 
 env:
   CACHE_VERSION: 1
@@ -20,7 +22,7 @@ jobs:
     timeout-minutes: 45
     strategy:
       matrix:
-        python-version: ["3.10"]
+        python-version: ["3.8"]
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.2
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -1,0 +1,82 @@
+# Most of this is inspired by the mypy primer
+# See: https://github.com/hauntsaninja/mypy_primer
+# This is the primer job that runs on the default 'main' branch
+# It is also responsible for caching the packages to prime on
+
+name: Primer / Main
+
+on:
+  push:
+    branches:
+      - main
+
+env:
+  CACHE_VERSION: 1
+
+jobs:
+  run-primer:
+    name: Run / ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    strategy:
+      matrix:
+        python-version: ["3.10"]
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.0.2
+      - name: Set up Python ${{ matrix.python-version }}
+        id: python
+        uses: actions/setup-python@v3.1.2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # Restore cached Python environment
+      - name: Generate partial Python venv restore key
+        id: generate-python-key
+        run: >-
+          echo "::set-output name=key::venv-${{ env.CACHE_VERSION }}-${{
+            hashFiles('setup.cfg', 'requirements_test.txt', 'requirements_test_min.txt')
+          }}"
+      - name: Restore Python virtual environment
+        id: cache-venv
+        uses: actions/cache@v3.0.2
+        with:
+          path: venv
+          key: >-
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{
+            steps.generate-python-key.outputs.key }}
+          restore-keys: |
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
+      - name: Create Python virtual environment
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          python -m venv venv
+          . venv/bin/activate
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r requirements_test.txt
+
+      # Cache primer packages
+      - name: Get commit string
+        id: commitstring
+        run: |
+          . venv/bin/activate
+          python tests/primer/primer_tool.py prepare --make-commit-string
+          output=$(python tests/primer/primer_tool.py prepare --read-commit-string)
+          echo "::set-output name=commitstring::$output"
+      - name: Restore projects cache
+        id: cache-projects
+        uses: actions/cache@v3
+        with:
+          path: .pylint_primer_tests/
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas:
Should we set the commit we're analyzing for each repositories ? I.e. we set the sha for each package to prime. This way we can't have a new message that appears simply because something was added on the primed repo. We could upgrade the primer repo to the latest version before release so we're still somewhat up to date.

DanielNoord:
> Should we set the commit we're analyzing for each repositories ? I.e. we set the sha for each package to prime. This way we can't have a new message that appears simply because something was added on the primed repo. We could upgrade the primer repo to the latest version before release so we're still somewhat up to date.

I don't think so.

The idea is to run `main` over a set of packages on specific commit. That commit is the latest commit of each package on their default branch at the time of the run. Then `main` records those commits as the `commitstring`.
A PR then gets the `commitstring` and retrieves the cache created by `main` and validates it. That way `main` and `PR` are both on the same commits of the packages (so even if the package adds new commits between the `main` and the `PR` run, those don't affect us).

Lastly, we run `PR` over the packages and compare its output with `main`. Notifying of any added messages or any removed messages. In that way it is similar to the current primer: we don't really care about which messages are emitted when with ``--enable=all, --enable-all-extensions``. We only care about the difference between `main` and `PR`.

.github/workflows/primer_run_main.yaml
```diff
@@ -0,0 +1,82 @@
...
+# See: https://github.com/hauntsaninja/mypy_primer
+# This is the primer job that runs on the default 'main' branch
+# It is also responsible for caching the packages to prime on
+
+name: Primer / Main
+
+on:
+  push:
+    branches:
+      - main
```
jacobtylerwalls, line 11: 🤔 
```suggestion
    branches:
      - main
    paths-ignore:
      - doc/data/messages/**
```

.github/workflows/primer_run_pr.yaml
```diff
@@ -0,0 +1,125 @@
...
+        id: commitstring
+        run: |
+          . venv/bin/activate
+          output=$(python tests/primer/primer_tool.py prepare --read-commit-string)
+          echo "::set-output name=commitstring::$output"
+      - name: Restore projects cache
+        id: cache-projects
+        uses: actions/cache@v3
+        with:
+          path: .pylint_primer_tests/
```
jacobtylerwalls, line 109: Should we add a .gitignore to this directory to ensure it always exists? I'm thinking for local testing etc.

jacobtylerwalls (approved):

tests/primer/primer_tool.py
```diff
@@ -0,0 +1,100 @@
...
+                )
+                print(f"'{package}' remote is at commit '{remote_sha1_commit}'.")
+                commit_string += remote_sha1_commit + "_"
+        elif self.config.read_commit_string:
+            with open("commit_string.txt", encoding="utf-8") as f:
+                print(f.read())
+
+        if commit_string:
+            with open("commit_string.txt", "w", encoding="utf-8") as f:
+                f.write(commit_string)
```
jacobtylerwalls, line 87: Probably should add to the project's .gitignore.

.github/workflows/primer_run_pr.yaml
Pierre-Sassoulas: Isn't this a ``.gitkeep`` ?

.github/workflows/primer_run_main.yaml
```diff
@@ -0,0 +1,82 @@
...
+  CACHE_VERSION: 1
+
+jobs:
+  run-primer:
+    name: Run / ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    strategy:
+      matrix:
+        python-version: ["3.10"]
```
Pierre-Sassoulas, line 23: I think we should use the most popular interpreter here. It look like right now it's 3.8. 

```suggestion
        python-version: ["3.8"]
```

popular_interpreter

.github/workflows/primer_run_pr.yaml
```diff
@@ -0,0 +1,125 @@
...
+  CACHE_VERSION: 1
+
+jobs:
+  run-primer:
+    name: Run / ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    strategy:
+      matrix:
+        python-version: ["3.10"]
```
Pierre-Sassoulas, line 25:
```suggestion
        python-version: ["3.8"]
```

tests/primer/primer_tool.py
```diff
@@ -0,0 +1,100 @@
...
+
+MAIN_DIR = Path(__file__).parent.parent.parent
+PRIMER_DIRECTORY = MAIN_DIR / ".pylint_primer_tests/"
+PACKAGES_TO_PRIME_PATH = Path(__file__).parent / "packages_to_prime.json"
+
+
+class Primer:
+    """Main class to handle priming of packages."""
+
+    def __init__(self, package_list: Path) -> None:
```
Pierre-Sassoulas, line 23:
```suggestion
    def __init__(self, json_path: Path) -> None:
```

Pierre-Sassoulas (changes requested):

DanielNoord:
Test run on my fork `main`:
https://github.com/DanielNoord/pylint/runs/6624159814
And subsequent run of the `PR` workflow on a PR:
https://github.com/DanielNoord/pylint/runs/6624174015

tests/primer/primer_tool.py
DanielNoord: I have changed this to go into the primer directory instead of writing to root.

.github/workflows/primer_run_pr.yaml
DanielNoord: Added!

Pierre-Sassoulas (approved):
This is pretty great ! We'll have even more confidence when releasing.

PR #6707, opened 2022-05-26
Added documentation examples for `use-set-for-membership`.

mbyrnepr2:
## Type of Changes

## Description

Part of https://github.com/PyCQA/pylint/issues/5953

### DIFF
--- a/doc/data/messages/u/use-set-for-membership/bad.py
+++ b/doc/data/messages/u/use-set-for-membership/bad.py
@@ -1,2 +1,3 @@
-def is_a_fruit_a_cat_could_eat(fruit: str) -> bool:
-    return fruit not in ["cherry", "grapes"]  # [use-set-for-membership]
+def fruit_is_dangerous_for_cat(fruit: str) -> bool:
+    """This list is only a silly example, don't make decision regarding your cat diet based on it."""
+    return fruit in ["cherry", "grapes"]  # [use-set-for-membership]
--- a/doc/data/messages/u/use-set-for-membership/bad.py
+++ b/doc/data/messages/u/use-set-for-membership/bad.py
@@ -1,2 +1,2 @@
-if "banana" in ["apple", "banana", "cantaloupe"]:  # [use-set-for-membership]
-    pass
+def is_a_fruit_a_cat_could_eat(fruit: str) -> bool:
+    return fruit not in ["cherry", "grapes"]  # [use-set-for-membership]
--- a/doc/data/messages/u/use-set-for-membership/bad.py
+++ b/doc/data/messages/u/use-set-for-membership/bad.py
@@ -1,0 +1,2 @@
+if "banana" in ["apple", "banana", "cantaloupe"]:  # [use-set-for-membership]
+    pass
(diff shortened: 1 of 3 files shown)


doc/data/messages/u/use-set-for-membership/bad.py
```diff
@@ -0,0 +1,2 @@
+if "banana" in ["apple", "banana", "cantaloupe"]:  # [use-set-for-membership]
+    pass
```
Pierre-Sassoulas, line 2:
```suggestion
def is_a_fruit_a_cat_could_eat(fruit: str) -> bool:
    return fruit not in ["cherry", "grapes"]  # [use-set-for-membership]
```

doc/data/messages/u/use-set-for-membership/good.py
```diff
@@ -1 +1,2 @@
-# This is a placeholder for correct code for this message.
+if "banana" in {"apple", "banana", "cantaloupe"}:
+    pass
```
Pierre-Sassoulas, line 2:
```suggestion
def is_a_fruit_a_cat_could_eat(fruit: str) -> bool:
    return fruit not in {"cherry", "grapes"}
```

Pierre-Sassoulas (commented):
I think some examples make more sense if it's a function and one of the value is unknown (especially code with boolean that would otherwise always have the same value).

mbyrnepr2:
> I think some examples make more sense if it's a function and one of the value is unknown (especially code with boolean that would otherwise always have the same value).

Yes I like it. Thanks @Pierre-Sassoulas.

doc/data/messages/u/use-set-for-membership/bad.py
```diff
@@ -0,0 +1,2 @@
+def is_a_fruit_a_cat_could_eat(fruit: str) -> bool:
+    return fruit not in ["cherry", "grapes"]  # [use-set-for-membership]
```
Pierre-Sassoulas, line 2:
```suggestion
def fruit_is_dangerous_for_cat(fruit: str) -> bool:
    """This list is only a silly example, don't make decision regarding your cat diet based on it."""
    return fruit in ["cherry", "grapes"]  # [use-set-for-membership]
```

doc/data/messages/u/use-set-for-membership/good.py
```diff
@@ -1 +1,2 @@
-# This is a placeholder for correct code for this message.
+def is_a_fruit_a_cat_could_eat(fruit: str) -> bool:
+    return fruit not in {"cherry", "grapes"}
```
Pierre-Sassoulas, line 2:
```suggestion
def fruit_is_dangerous_for_cat(fruit: str) -> bool:
    """This list is only a silly example, don't make decision regarding your cat diet based on it."""
    return fruit in {"cherry", "grapes"}
```

Pierre-Sassoulas (commented):
My bad regarding the example: I think if it's ``not in`` there's no benefit to using a set because you will need to check all the iterable anyway and pylint rightly do not complain about it.

Pierre-Sassoulas (approved):

Issue #6582, opened 2022-05-11
Refactoring ToDos for `pyreverse`


DudeNr33:
### Current problem

While adding typing to `pyreverse`, some parts of the code base were found that can be refactored.
This issue should primarily serve as a expandable ToDo list for the issues found that should be addressed after finishing the typing.

So far we have:

* ~In `pylint.pyreverse.utils`: The `ASTWalker` base class is not really necessary, as the only child is `LocalsVisitor`.~ --> #6712 
* In `pylint.pyreverse.diadefslib`: `DefaultDiadefGenerator.leave_project` returns a variable length tuple with the generated diagrams. 
A) It is unusual (for `pylint`'s standards) for a Visitor to return something from a `visit_*`/`leave_*` method.
B) Returning a tuple with variable length (and a non-defined order) is not user friendly
* ~In `pylint.pyreverse.inspector`: `Linker.inherited_interfaces` can be removed, together the `herited` parameter to `interfaces(node, herited, handler_func`. The value is always constant.~ --> #6713 
* ~In `pylint.pyreverse.inspector`: `Linker.link_*` are defined, but never used.~ --> #6713 
* ~In `pylint.pyreverse.inspector`: `_iface_hdlr(_)` returns a constant and is only used as the default value for `handler_func` in `interfaces`. There is no caller that uses this `handler_func` parameter, so we can just get rid of it.~ --> #6713 
* In `pylint.pyreverse.inspector`: we can probably get rid of everything related to `interfaces` - the implementation relies on the existence of the `__implements__` class attribute. The corresponding [PEP 245](https://peps.python.org/pep-0245/) is rejected, and `pylint` itself also got rid of the `__implements__` pattern. This would also mean we could get rid of all code related to handling "implementation" links in the diagrams themselves. --> `DeprecationWarning` issued after #6713 if a class with `__implements__` is found, complete removal planned in 3.0.
* In `pylint.pyreverse.diagrams` and `pylint.pyreverse.printer`: Type annotations for attributes are determined in `diagrams.py`, but for method signatures it is done in `printer.py`. This should be done in a uniform way.

### Desired solution

Refactor the problems found to get code that is easier to understand and maintain.

PR #6712, opened 2022-05-27
Remove unnecessary base class and dead code in `pylint.pyreverse.utils`

DudeNr33:

## Type of Changes

## Description
Ref #6582 
Closes the first bullet point from the list:
`ASTWalker` and `LocalsVisitor` were apparently moved from `astroid.utils` to `pylint.pyreverse.utils` some years ago.
Maybe originally there were also other subclasses of `ASTWalker`, but at least in `pyreverse` that is not the case.

Thus we can remove `ASTWalker` and some of the unused code in there and directly put the remaining stuff in `LocalsVisitor`.

Some remarks, as the diff is a bit hard to read imo:

* `handler` was always `self` -> removed `handler` and replaced references with `self`
* `walk` and `leave` were never called -> removed

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -177,6 +177,10 @@ Release date: TBA
 
   Ref #5392
 
+* ``pylint.pyreverse.ASTWalker`` has been removed, as it was only used internally by a single child class.
+
+  Ref #6712
+
 * ``interfaces.implements`` has been deprecated and will be removed in 3.0. Please use standard inheritance
   patterns instead of ``__implements__``.
 
--- a/doc/whatsnew/2.14.rst
+++ b/doc/whatsnew/2.14.rst
@@ -281,6 +281,10 @@ Other Changes
 
   Closes #6644
 
+* ``pylint.pyreverse.ASTWalker`` has been removed, as it was only used internally by a single child class.
+
+  Ref #6712
+
 
 Deprecations
 ============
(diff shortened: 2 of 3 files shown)


DudeNr33:
@Pierre-Sassoulas regarding your idea of user facing / dev related changes in https://github.com/PyCQA/pylint/issues/6632#issuecomment-1128565501:
Do you think I should add a Changelog entry somewhere?

Pierre-Sassoulas:
> @Pierre-Sassoulas regarding your idea of user facing / dev related changes in https://github.com/PyCQA/pylint/issues/6632#issuecomment-1128565501: Do you think I should add a Changelog entry somewhere?

We'll do the switch for 2.15, let's keep on doing it like we were for 2.14.

Pierre-Sassoulas (approved):
LGTM, I think it can go in 2.14 even without changelog, it's an internal refactor, right ?

DudeNr33:
Yes, this should be purely internal. At least I would not say the `ASTWalker` class is considered public API, and there is no plugin system for `pyreverse`. 

I did a GitHub search and only found one instance where it was actually used, but that was just part of a [demo talk in 2017](
https://github.com/vintasoftware/django-linters-talk-demo/blob/145e5a7fadf94b92d1ee261f79f756e01a919f73/ast_qs_inferred.py).

Pierre-Sassoulas (approved):

PR #6714, opened 2022-05-27
Run and upload primer over packages

DanielNoord:

## Type of Changes

## Description

Ref. https://github.com/PyCQA/pylint/issues/5364.

Next step for the primer.

Test run on local `main`:
https://github.com/DanielNoord/pylint/actions/runs/2396609324
Test run on a PR:
https://github.com/DanielNoord/pylint/actions/runs/2396610764

Basically, the PR run will now result in three files: the PR number, the output of the primer on the PR and the output of the `main` run against which the PR is being compared. 

This can be downloaded by a last and final step which takes all these files. Compares them and then comments on the relevant PR.

### DIFF
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -111,7 +111,7 @@ jobs:
                archive_format: "zip",
             });
             fs.writeFileSync("primer_output_main.zip", Buffer.from(downloadWorkflowTwo.data));
-            return lastRunMain.head_sha
+            return lastRunMain.head_sha;
       - name: Copy and unzip the commit string
         run: |
           unzip primer_commitstring.zip
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -33,9 +33,9 @@ jobs:
         uses: actions/setup-python@v3.1.2
         with:
           python-version: ${{ matrix.python-version }}
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v3
         with:
-          version: 12
+          node-version: 16
       - run: npm install @octokit/rest
 
       # Restore cached Python environment
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -26,6 +26,8 @@ jobs:
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.2
+        with:
+          fetch-depth: 0
       - name: Set up Python ${{ matrix.python-version }}
         id: python
         uses: actions/setup-python@v3.1.2
@@ -62,7 +64,8 @@ jobs:
           pip install -U -r requirements_test.txt
 
       # Cache primer packages
-      - name: Download diffs
+      - name: Download last 'main' run info
+        id: download-main-run
         uses: actions/github-script@v6
         with:
           script: |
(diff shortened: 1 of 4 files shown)


.github/workflows/primer_run_pr.yaml
```diff
@@ -95,10 +100,24 @@ jobs:
...
+            const [matchArtifactMainOutput] = artifacts_main.data.artifacts.filter((artifact) =>
+              artifact.name == "primer_output");
+            const downloadWorkflowTwo = await github.rest.actions.downloadArtifact({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               artifact_id: matchArtifactMainOutput.id,
+               archive_format: "zip",
+            });
+            fs.writeFileSync("primer_output_main.zip", Buffer.from(downloadWorkflowTwo.data));
+            return lastRunMain.head_sha
```
jacobtylerwalls, line 114:
```suggestion
            return lastRunMain.head_sha;
```

tests/primer/primer_tool.py
```diff
@@ -88,6 +100,35 @@ def _handle_prepare_command(self) -> None:
...
+        # We want to test all the code we can
+        enables = ["--enable-all-extensions", "--enable=all"]
+        # Duplicate code takes too long and is relatively safe
+        disables = ["--disable=duplicate-code"]
+        arguments = data.directories + data.pylint_args + enables + disables
+        if data.pylintrc_relpath:
+            arguments += [f"--rcfile={data.pylintrc_relpath}"]
+        output = StringIO()
+        reporter = JSONReporter(output)
+        Run(arguments, reporter=reporter, do_exit=False)
```
jacobtylerwalls, line 129: We don't want to lint every primer package twice, so I think we will need to turn off the python test that checks for crashes, yeah?

tests/primer/primer_tool.py
```diff
@@ -88,6 +100,35 @@ def _handle_prepare_command(self) -> None:
...
+
+        packages: dict[str, list[dict[str, str | int]]] = {}
+
+        for package, data in self.packages.items():
+            output = self._lint_package(data)
+            packages[package] = output
+            print(f"Successfully primed {package}.")
+
+        with open(
+            PRIMER_DIRECTORY / f"output_{self.config.type}.txt", "a", encoding="utf-8"
```
jacobtylerwalls, line 115: If multiple PRs are each running the primer step, does this mean they will append their results to each other?

tests/primer/primer_tool.py
DanielNoord: Yes! Although this still needs some work before we get there. We don't actually prime on `astroid` and `black` doesn't take too long. So for now I think this is fine.

tests/primer/primer_tool.py
DanielNoord: No, they don't share their environments.

tests/primer/primer_tool.py
```diff
@@ -88,6 +100,35 @@ def _handle_prepare_command(self) -> None:
             ) as f:
                 f.write(commit_string)
 
+    def _handle_run_command(self) -> None:
+        if Path(PRIMER_DIRECTORY / f"output_{self.config.type}.txt").exists():
+            os.remove(PRIMER_DIRECTORY / f"output_{self.config.type}.txt")
```
jacobtylerwalls, line 105: I guess this is what confused me -- if this isn't executing on the CI runners, is this for local testing? If so, should we just remove this and do "w" mode below?

jacobtylerwalls:
Should we bump `setup-node` to v3? See:
```
Warning: Input 'version' has been deprecated with message: The version property will not be supported after October 1, 2019. Use node-version instead
```

Pierre-Sassoulas (approved):
LGTM,, but I think @jacobtylerwalls is right that we should try to remove the package from the normal primer if they are in the false positive primer.

jacobtylerwalls (approved):
I think it's unlikely we'll forget to turn off the double-primer runs once we enable running this over the full set of primer repositories, so if you wish to merge now and then sort that out in a future PR, 👍🏻

PR #6719, opened 2022-05-27
Fix return type annotations in MermaidJS diagrams.

### LINKED ISSUE #6467: `pyreverse` does not display return types of methods in `mmd` format
### Bug description

`pyreverse` uses the wrong syntax for the return type of methods in MermaidJS format.
Current output for the following file:
```python
class Dummy:
    def foo(self) -> None:
        return
```
produces the following output:
```
classDiagram
  class Dummy {
    foo() -> None
  }
```
which looks like this, and the return type is missing.
```mermaid
classDiagram
  class Dummy {
    foo() -> None
  }
```

### Configuration

_No response_

### Command used

```shell
pyreverse a.py
```

### Pylint output

```shell
see above
```

### Expected behavior

The correct syntax for mermaid would be to just use `foo() None`:
```
classDiagram
  class Dummy {
    foo() None
  }
```

```mermaid
classDiagram
  class Dummy {
    foo() None
  }
```

### Pylint version

```shell
Current main
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

DudeNr33:

## Type of Changes

## Description
The ` ->` before the return type annotation is incorrect, and the return type is not shown in the final diagram.

Closes #6467

### DIFF
--- a/tests/pyreverse/functional/class_diagrams/annotations/method_annotation.py
+++ b/tests/pyreverse/functional/class_diagrams/annotations/method_annotation.py
@@ -1,0 +1,17 @@
+from __future__ import annotations
+
+# pylint: disable-next=import-error
+from lib import Banana, Coconut, Sound  # type: ignore[import]
+
+
+class Monkey:
+    def eat(self, food: Banana | Coconut) -> None:
+        print(f"Monkey eats {food}")
+
+    def jump(self, height: int | None = 10) -> None:
+        print(f"Monkey jumps {height}")
+
+    def scream(self, volume: int | None) -> Sound:
+        if volume is None:
+            volume = 0
+        return Sound(volume)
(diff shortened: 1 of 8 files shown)


tests/pyreverse/functional/class_diagrams/annotations/method_annotation.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import annotations
+
+# pylint: disable-next=import-error
+from lib import Banana, Coconut, Sound  # type: ignore[import]
+
+
+class Monkey:
+    def eat(self, food: Banana | Coconut) -> None:
```
Pierre-Sassoulas, line 8: Seems like you're handling ``|`` it made me think of this stackoverflow question : https://stackoverflow.com/questions/72388388/. Maybe we can add some test case with Union too ?

DudeNr33: Adding a corresponding test case is a good idea!

The alternative union syntax however gives me some headaches for the tests, as the output differs for different Python versions. I will have to take a closer look.

Pierre-Sassoulas (approved):
LGTM !

PR #6720, opened 2022-05-27
Fix `pyreverse` type annotations for attributes using alternative union syntax

DudeNr33:

## Type of Changes

## Description

Follow up to https://github.com/PyCQA/pylint/pull/6719#discussion_r883800036
The annotations were working correctly in method signatures, but not for attributes.

/CC @mbyrnepr2 would appreciate if you could do a review!


pylint/pyreverse/diagrams.py
```diff
@@ -159,7 +159,9 @@ def class_names(self, nodes_lst: Iterable[nodes.NodeNG]) -> list[str]:
             if isinstance(node, astroid.Instance):
                 node = node._proxied
             if (
-                isinstance(node, (nodes.ClassDef, nodes.Name, nodes.Subscript))
+                isinstance(
+                    node, (nodes.ClassDef, nodes.Name, nodes.Subscript, nodes.BinOp)
```
mbyrnepr2, line 163: I don't think you need to add this because it adds the BinOp node to the "class name" section in the diagram?

output

DudeNr33: Thank you for noticing this produces wrong output for `dot` files.

Regarding the code: it is necessary to add this, as this method is used to get the names for the names in the type annotation.
The problem was that the `|` character needs to be escaped when creating `dot` files, as it is otherwise rendered as the horizontal line we can see here between `str` and `int` for the `alternative_union_syntax` member.

It would probably be better to refactor and reuse your `get_annotation_label` here, too.
I noticed that currently the type annotations for attributes are determined in `diagrams.py`, whereas the type annotations for method signatures are determined in `printer.py`.
I added that to the list of refactoring todos (#6582).

mbyrnepr2: I think that sounds great @DudeNr33; I am sure there is room for improvements & I think your TODOs will make it a better tool for everyone. Let me know if I can help in any way & I'll be happy to pitch in when I can get time.

pylint/pyreverse/utils.py
```diff
@@ -218,7 +218,7 @@ def infer_node(node: nodes.AssignAttr | nodes.AssignName) -> set[Any]:
     ann = get_annotation(node)
     try:
         if ann:
-            if isinstance(ann, nodes.Subscript):
+            if isinstance(ann, (nodes.Subscript, nodes.BinOp)):
```
mbyrnepr2, line 221: Perhaps we could make it more strict as this would print something like `int + str` to the diagram if a `+` is used in the type annotation.

pylint/pyreverse/dot_printer.py
```diff
@@ -81,24 +81,26 @@ def emit_node(
...
-            label = "<<interface>>\\n" + label
-
         if properties.attrs is None and properties.methods is None:
             # return a "compact" form which only displays the class name in a box
             return label
 
         # Add class attributes
         attrs: list[str] = properties.attrs or []
-        label = "{" + label + "|" + r"\l".join(attrs) + r"\l|"
+        label = (
```
mbyrnepr2, line 95: It may be worth using an f-string here if it improves readability.

pylint/pyreverse/diagrams.py
mbyrnepr2: Perhaps one for the TODO list also is this [issue mentioned on SO](https://stackoverflow.com/questions/71059123/pyreverse-not-showing-composition-relationships-in-the-umls-when-using-absolute).

The thing here seems to be that relationships are broken when a type-annotation uses a class imported from another module. It works ok if both classes are in the same module.

DudeNr33: Thank you for your offer to contribute! 😊 👍  If you'd like to take a look at unifying the way type annotations for attributes and method signatures are determined, go ahead! 

Regarding the StackOverflow question: I haven't tried it yet, but my guess is that this will be fixed in 2.14 through #6617.

doc/whatsnew/2/2.14/summary.rst
```diff
@@ -260,6 +260,8 @@ Other Changes
 
   Closes #6467
 
+* Fix type annotations of class and instance attributes using the alternative union syntax.
+
```
DudeNr33, line 264: @Pierre-Sassoulas from what I've read in the new PR template this should be the right place for a user-facing bug fix.

mbyrnepr2 (approved):

doc/whatsnew/2/2.14/summary.rst
Pierre-Sassoulas: If it's in summary.py it should also be in the full changelog for 2.14 (we're going to release soon so it's not going to last). For 2.15 we're discussing what we're going to do in #6688, right now it's "all in a single file, in the right section (to be determined), with the PR number (not the original issue number) and enforced by the CI if the PR is not labelled maintenance or documentation in github". If you have an idea on how to do that better that would be welcome 😄

Pierre-Sassoulas (approved):

Issue #4171, opened 2021-03-02
No cyclic-import messages with jobs=0


udifuchs:
Running `pylint -jobs=0` does not report about cyclic-import. This can be reproduced with a trivial example.
`a.py`:
```
import b
```
`b.py`:
```
import a
```

This issue is not new to pylint 2.7. It is more noticeable only because running jobs in parallel became more useful in this version.
```
pylint --version
pylint 2.7.1
astroid 2.5
Python 3.8.6 (default, Sep 25 2020, 09:36:53) 
[GCC 10.2.0]
```

doublethefish:
This is issue is mentioned in #374 .

udifuchs:
Sorry, I missed this bug report.
I will close this issue.

doublethefish:
It might be good to keep this open as it is a simpler repro and the other bug contains a lot of stuff.

hippo91:
@udifuchs thanks for the report.

praiskup:
The problem with this trivial example is that it actually detects a non-issue.  It seems completely fine to cross-include the modules, as long as the "use" (accessing objects from the module) happens on runtime (not import time).

doublethefish:
> The problem with this trivial example is that it actually detects a non-issue. It seems completely fine to cross-include the modules, as long as the "use" (accessing objects from the module) happens on runtime (not import time).

Whilst I agree that, in practical terms, the trivial example should "work", I would state that any form of cyclic import is either problematic or will become problematic. That said, if it were possible to detect use-at-import that would be a better, more complete check - I suspect that is almost impossible to do thoroughly because of [aliasing](https://en.wikipedia.org/wiki/Aliasing_(computing)).

Simpler and more useful approaches (vs use-at-import detection) would be to either:
1. scope the import:
```py
def run_time_use():
   """Safer but uglier.

   Still suffers from the potential aliasing problem, but is explicit and scoped, so the problem is contained."""
   import A
   a.do_something()
```
2. explicitly disable the warning
```py
"""Dangerous as use-at-import can be introduced later, and deployed to prod"""
import A  # pylint: disable=cyclic-import
```

Pierre-Sassoulas:
Closing as duplicate of #374, I added the example in a comment there.

jacobtylerwalls:
> It might be good to keep this open as it is a simpler repro and the other bug contains a lot of stuff.

Indeed! Reopening because I'm about to mark #374 closed by a separate PR.

PR #6723, opened 2022-05-28
Compare difference between primer runs and post comment

DanielNoord:

## Type of Changes

## Description

Ref. #5364 

Example can be seen here:
https://github.com/DanielNoord/pylint/pull/145

We can't show the code in a code snippet easily as the normal functionality Github offers only works for permalinks within the same repository. So we would need to create a script that downloads the code ourselves or store it as an artefact in the CI runs. For now, I think this should be fine though.

I'm not 100% confident that this won't crash as it is quite hard to test this. The comment workflow needs to be on the default branch and I could only do so much testing on my local fork.
Therefore I have only set three packages up for priming with this system so that the others can be taken care of by the others.


tests/primer/primer_tool.py
```diff
@@ -112,11 +130,79 @@ def _handle_run_command(self) -> None:
...
+                comment += "The following messages are no longer emitted:\n"
+            count = 1
+            for message in missing_messages:
+                comment += (
+                    f"{count}) {message['symbol']} on line {message['line']} here:\n"
+                )
+                filepath = str(message["path"]).replace(
+                    str(package_data.clone_directory), ""
+                )
+                comment += f"{package_data.url}/blob/{package_data.branch}/{filepath}#L{message['line']}\n"
```
Pierre-Sassoulas, line 173:
```suggestion
                filepath = str(message["path"]).replace(
                    str(package_data.clone_directory), ""
                )
                comment += (
                    f"{count}) [{message['symbol']} on line {message['line']}]("
                     f"{package_data.url}/blob/{package_data.branch}/{filepath}#L{message['line']})"
                )
```
Not sure if I did that right but the Intent is to have 

1. [locally-disabled on line 91](https://github.com/PyCQA/astroid/blob/main//astroid/__init__.py#L91)

It seems there's a problem with some links that lack the L before the line

Pierre-Sassoulas (approved):
This is awesome 🤩 !

tests/primer/primer_tool.py
DanielNoord: I fixed the `L`. I would keep the two lines for now as (at some point) we want to change the second line to an actual code snippet instead of just the link.

Pierre-Sassoulas: Code snippet sure would be nice, but can't we do that by accessing the local version of the code we just linted ?

Pierre-Sassoulas: Something like this -5 / +5 lines:
```python
# It's impossible to import from astroid.nodes with a wildcard, because
# there is a cyclic import that prevent creating an __all__ in astroid/nodes
# and we need astroid/scoped_nodes and astroid/node_classes to work. So
# importing with a wildcard would clash with astroid/nodes/scoped_nodes
# and astroid/nodes/node_classes.
from astroid.nodes import (  # pylint: disable=redefined-builtin (Ellipsis)
    CONST_CLS,
    AnnAssign,
    Arguments,
    Assert,
    Assign,
```

DanielNoord: We don't load the packages when making the comment. As I said in OP: we might be able to add the code snippets in the artefacts or download them in the comment workflow. But let's do that in a follow up.

Pierre-Sassoulas: Shouldn't we do the comment at the same time we lint then ? I agree it's mergeable as is, but the decision to separate the check/message is taken now and would have to be refactored later.

DanielNoord: No we can't do that at the same time. See the comment at the top of this workflow file.

We need a different workflow trigger for the comment PR as it needs more permissions than you get when using `pull_request` as trigger. This is also why commenting and priming is separated on `mypy` and `pydocstringformatter`.

Pierre-Sassoulas: All right, thanks for explaining :)

tests/primer/primer_tool.py
```diff
@@ -112,11 +130,79 @@ def _handle_run_command(self) -> None:
...
+        with open("comment.txt", "w", encoding="utf-8") as f:
+            f.write(comment)
+
     def _lint_package(self, data: PackageToLint) -> list[dict[str, str | int]]:
         # We want to test all the code we can
         enables = ["--enable-all-extensions", "--enable=all"]
         # Duplicate code takes too long and is relatively safe
-        disables = ["--disable=duplicate-code"]
+        # TODO: Find a way to allow cyclic-import and compare output correctly
+        disables = ["--disable=duplicate-code,cyclic-import"]
```
jacobtylerwalls, line 205: This is a sign we shouldn't be using multiprocessing. As I think you might be aware, parallel mode is currently nondeterministic/unsafe in pylint, and it's not just `cyclic-import`, it's potentially anywhere given #4874 and other issues we haven't fully investigated. I'd like to look at that for 2.15, as it's major technical debt, but until then, I think we need to use `--jobs=1` for the primer runs. Sorry I didn't catch this on the other PR. Then we can revert this change here.

If we want to speed up primer runs we could use parallel CI jobs instead trying to get pylint to do the heavy lifting.

.github/workflows/primer_comment.yaml
```diff
@@ -0,0 +1,140 @@
...
+            steps.generate-python-key.outputs.key }}
+          restore-keys: |
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
+      - name: Create Python virtual environment
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          python -m venv venv
+          . venv/bin/activate
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r requirements_test.txt
```
jacobtylerwalls, line 62: This applies to a different workflow file, but just commenting here. 

Also something I should have mentioned two PRs ago: I wonder if we want to use bleeding-edge `astroid` for these linting runs. If there are pending changes in astroid that cause false positives/negatives/crashes with someone's pylint PR, the pylint PR is our chance to address them before the contributor loses interest.

jacobtylerwalls (changes requested):
Terrific! I have two comments that I think are important to discuss before we merge this and turn this feature "on", even though I should have brought them up earlier :D

tests/primer/primer_tool.py
jacobtylerwalls: Ah -- I might be wrong! Now that I look at the `Run` call, I suppose we're not using parallel mode. Do we have a GitHub issue that captures this problem with `cyclic-import` without using `multiprocessing`?

Pierre-Sassoulas: Possibly https://github.com/PyCQA/pylint/issues/4171 ?

jacobtylerwalls: Ah, but that's `--jobs=0`, not 1, which is partly what confused me, since that _does_ use parallel mode!

jacobtylerwalls: Maybe Daniel can speak to the failures he saw and if he was testing locally in parallel mode.

tests/primer/primer_tool.py
jacobtylerwalls: I don't think it's worth adding a code snippet; linking to the github view better anyway as it's interactive. (Next best thing to actually loading the project in an IDE).

.github/workflows/primer_comment.yaml
DanielNoord: As to `astroid` I'm not actually sure. There are also cases where bleeding edge astroid is incompatible with latest pylint. So both can cause issues.

Perhaps we should default to one case and then turn on the other when we need to? We could just add a flag to `run` mode. Not sure if that would be a blocker though. Perhaps we should just let this run as a non required workflow? There are many edge cases that we probably haven't considered yet.

_I also thought about only running this workflow when a PR has a certain label. That way we can control this a little better while the tool is still in it's beta phase._

As for cyclic import. I have no idea actually. It couldn't get it to reproduce locally easily and thought about all the problems we have with the message so just thought I would disable it.

jacobtylerwalls: > As for cyclic import. I have no idea actually. It couldn't get it to reproduce locally easily and thought about all the problems we have with the message so just thought I would disable it.

I got it just now locally on the first try, so maybe I can try to help debug. Looking at the JSON output of the runs I'm seeing a lot fatal errors ("no module named src/astroid") etc., so I'm wondering if we're not invoking the runs correctly.

jacobtylerwalls: > There are also cases where bleeding edge astroid is incompatible with latest pylint. So both can cause issues.

That could slow down astroid contributors dealing with pre-commit, but would that affect pylint contributors? It's true we don't have to block, so if you prefer I'll open a PR and we can discuss there. I think we should do it. If pending `astroid` changes broke `pylint` main, we want to know immediately and block all pylint dev anyway, IMO.

jacobtylerwalls: > I also thought about only running this workflow when a PR has a certain label. That way we can control this a little better while the tool is still in it's beta phase.

Neat! So for now, any functional change or nontrivial refactor gets "primer-diff-messages"?

tests/primer/primer_tool.py
```diff
@@ -112,11 +130,79 @@ def _handle_run_command(self) -> None:
...
+                count += 1
+
+        if comment == "":
+            comment = "🤖 According to the primer, this change has **no effect** on the checked open source code. 🤖🎉"
+        else:
+            comment = (
+                "🤖 **Effect of this PR on checked open source code:** 🤖\n\n" + comment
+            )
+
+        with open("comment.txt", "w", encoding="utf-8") as f:
```
jacobtylerwalls, line 197: Can we put this under `.pylint_primer_tests`?

tests/primer/primer_tool.py
```diff
@@ -112,11 +130,79 @@ def _handle_run_command(self) -> None:
...
+                continue
+
+            comment += f"\n\n**Effect on [{package}]({self.packages[package].url}):**\n"
+
+            if missing_messages:
+                comment += "The following messages are no longer emitted:\n"
+            count = 1
+            for message in missing_messages:
+                comment += (
+                    f"{count}) {message['symbol']} on line {message['line']} here:\n"
```
jacobtylerwalls, line 168: For simple messages where the surrounding context is irrelevant, or messages that actually know the relevant context, like `redefined-outer-name`, everything relevant is already in the message text.  So what do you think about using the message text?
```suggestion
                    f"{count}) {message['symbol']}: {message['message']}:\n"
```

.github/workflows/primer_comment.yaml
DanielNoord: Yeah that would be a possibility. At some point I think we want to turn this on based on paths so we don't need to keep adding the label, but it would be a possibility to start with "label-only".

Do you want me to change it do work like that?

tests/primer/primer_tool.py
DanielNoord: Yes, although some other stuff will need to move as well.

Btw, I was thinking of moving `. pylint_primer_tests` into `./tests`. It feels out of place in root. What do you think?

tests/primer/primer_tool.py
DanielNoord: Perhaps we should format it with italics, but I like the suggestion!

tests/primer/primer_tool.py
DanielNoord: See https://github.com/PyCQA/pylint/pull/6728

jacobtylerwalls:
I'm definitely in favor of punting the `cyclic-import` issues if they are hard to solve, but I'm wondering if we merge `main` especially after #6725 the problems just go away! A couple CI runs would be nice to see!

DanielNoord:
> I'm definitely in favor of punting the `cyclic-import` issues if they are hard to solve, but I'm wondering if we merge `main` especially after #6725 the problems just go away! A couple CI runs would be nice to see!

Well I jusst merged `main`. Let me remove and see if it works on my local branch.

.github/workflows/primer_comment.yaml
jacobtylerwalls: Let's not do it now. Maybe the paths can be done eventually and we won't have to spend time on labels.

DanielNoord:
@jacobtylerwalls That doesn't look good:
https://github.com/DanielNoord/pylint/pull/145

You said you could reproduce locally? Could you check what is up with those `cyclic-imports`? It seems like there are similar messages, but not similar enough.

DanielNoord:
@jacobtylerwalls It just seems as if `cylic-import` is not deterministic enough. Not sure if it is worth the time to investigate for this PR.

jacobtylerwalls:
I'm about to take a look. I agree we will punt if we're doing everything right. In the meantime, did you notice
```shell
git.exc.NoSuchPathError: /home/runner/work/pylint/pylint/tests/.pylint_primer_tests/pytest-dev/pytest
```
?

DanielNoord:
> I'm about to take a look. I agree we will punt if we're doing everything right. In the meantime, did you notice
> 
> ```shell
> git.exc.NoSuchPathError: /home/runner/work/pylint/pylint/tests/.pylint_primer_tests/pytest-dev/pytest
> ```
> 
> ?

Yeah that's another thing which I am not sure we can/should fix right now.

Whenever we add a new package to prime on the `main` run won't have cached that package and thus the run will fail. We could obviously recreate the cache, but that doesn't really help: `main` won't have output for that package so we will fail other stages in the process where we assume that `main` and `PR` have the same packages in their output dictionary (`output_main.txt` and `output_pr.txt`).

jacobtylerwalls:
Got it. Forgot that adding `pytest` was in this changeset. We're likely to do this anyway, but when we add more packages from now let's add em in PRs that only do just that.

DanielNoord:
> Got it. Forgot that adding `pytest` was in this changeset. We're likely to do this anyway, but when we add more packages from now let's add em in PRs that only do just that.

Completely agree. In fact, we might even consider not running `Primer / Run` whenever we change the `to_prime_packages` `JSON` file, `primer_main.yaml` and `primer_comment.yaml`. All of those changes will always need to be on `main` to work so it doesn't really make sense to run when any of those files gets changed.

jacobtylerwalls:
If you don't mind, could you just do that now: open and merge a PR that adds pytest? And then this PR should get past that failure. It's worth going a little slower on this first PR, it helps me understand what I'm reviewing!

jacobtylerwalls:
Do we know what the double slashes in the URLs are doing there?
```.../blob/main//astroid/...```

jacobtylerwalls:
What about wrapping the diffed items in a `<details>`? That's always nice to keep the pull request discussion from getting overwhelming. Some functional changes could be quite overwhelming, if a new feature causes 250 messages in Django, 40 in pytest, 40 in music21, etc.

DanielNoord:
> What about wrapping the diffed items in a `<details>`? That's always nice to keep the pull request discussion from getting overwhelming. Some functional changes could be quite overwhelming, if a new feature causes 250 messages in Django, 40 in pytest, 40 in music21, etc.

Good idea! That might interfere with code snippets at some point, but for now it is absolutely better!

jacobtylerwalls:
> @jacobtylerwalls It just seems as if cylic-import is not deterministic enough. Not sure if it is worth the time to investigate for this PR.

Agreed. I see this locally even after the changes we just did. Let's go back to ignoring them and add a comment that it's currently nondeterministic. We have various issues open, I don't know that we need another.

DanielNoord:
> Do we know what the double slashes in the URLs are doing there?
> `.../blob/main//astroid/...`

Should be fixed!


> What about wrapping the diffed items in a `<details>`? That's always nice to keep the pull request discussion from getting overwhelming. Some functional changes could be quite overwhelming, if a new feature causes 250 messages in Django, 40 in pytest, 40 in music21, etc.

Added!


> > @jacobtylerwalls It just seems as if cylic-import is not deterministic enough. Not sure if it is worth the time to investigate for this PR.
> 
> Agreed. I see this locally even after the changes we just did. Let's go back to ignoring them and add a comment that it's currently nondeterministic. We have various issues open, I don't know that we need another.

Re-added!

jacobtylerwalls (approved):
Wonderful!

DanielNoord:
@Pierre-Sassoulas Do you want to do one final review? We changes some stuff around based on @jacobtylerwalls's very useful suggestions!

Pierre-Sassoulas (approved):

DudeNr33:
Just wanted to say that this is awesome and will be very valuable in the future! Great work. 👍

PR #6724, opened 2022-05-28
Use bleeding-edge `astroid` for primer comparisons

jacobtylerwalls:
## Type of Changes

## Description

If there are changes in bleeding-edge `astroid` that might affect a pending `pylint` PR, I think it's important to surface the issue during review.

### DIFF
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -56,12 +56,12 @@ jobs:
             ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
       - name: Create Python virtual environment
         if: steps.cache-venv.outputs.cache-hit != 'true'
-        # Use bleeding-edge astroid
         run: |
           python -m venv venv
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
+          # Use bleeding-edge astroid
           pip install git+https://github.com/PyCQA/astroid.git
 
       # Cache primer packages
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -32,13 +32,18 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Get latest astroid commit
+        id: get-astroid-sha
+        run: |
+          curl https://api.github.com/repos/PyCQA/astroid/commits |
+          python -c "import json, sys; print(json.load(sys.stdin)[0]['sha'])" > astroid_sha.txt
       # Restore cached Python environment
       - name: Generate partial Python venv restore key
         id: generate-python-key
         run: >-
           echo "::set-output name=key::venv-${{ env.CACHE_VERSION }}-${{
-            hashFiles('setup.cfg', 'requirements_test.txt', 'requirements_test_min.txt')
-          }}"
+            hashFiles('setup.cfg', 'requirements_test.txt', 'requirements_test_min.txt',
+          'astroid_sha.txt') }}"
       - name: Restore Python virtual environment
         id: cache-venv
         uses: actions/cache@v3.0.2
@@ -51,11 +56,13 @@ jobs:
             ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
       - name: Create Python virtual environment
         if: steps.cache-venv.outputs.cache-hit != 'true'
+        # Use bleeding-edge astroid
         run: |
           python -m venv venv
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
+          pip install git+https://github.com/PyCQA/astroid.git
 
       # Cache primer packages
       - name: Get commit string
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -37,7 +37,8 @@ jobs:
         run: |
           curl https://api.github.com/repos/PyCQA/astroid/commits |
           python -c "import json, sys; print(json.load(sys.stdin)[0]['sha'])" > astroid_sha.txt
-      # Restore cached Python environment
+
+      # Restore cached Python environment      
       - name: Generate partial Python venv restore key
         id: generate-python-key
         run: >-
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -38,7 +38,7 @@ jobs:
           curl https://api.github.com/repos/PyCQA/astroid/commits |
           python -c "import json, sys; print(json.load(sys.stdin)[0]['sha'])" > astroid_sha.txt
 
-      # Restore cached Python environment      
+      # Restore cached Python environment
       - name: Generate partial Python venv restore key
         id: generate-python-key
         run: >-
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -38,13 +38,18 @@ jobs:
           node-version: 16
       - run: npm install @octokit/rest
 
+      - name: Get latest astroid commit
+        id: get-astroid-sha
+        run: |
+          curl https://api.github.com/repos/PyCQA/astroid/commits |
+          python -c "import json, sys; print(json.load(sys.stdin)[0]['sha'])" > astroid_sha.txt
       # Restore cached Python environment
       - name: Generate partial Python venv restore key
         id: generate-python-key
         run: >-
           echo "::set-output name=key::venv-${{ env.CACHE_VERSION }}-${{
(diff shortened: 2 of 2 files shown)


Pierre-Sassoulas (changes requested):
I'm -1 on this. Previously we were releasing astroid alongside pylint (2.5 with 2.5) and using bleeding edge astroid too. We've changed it at some point because if there's an issue in astroid and we're using astroid's bleeding edge it's breaking all the pylint pipelines at once. Then we have to tell every contributor that it's not them it's astroid, fix pylint's main because if we don't we're going to have to do more work because of the broken main branch anyway, then tell everyone to rebase on main after that. It's a headache, I'd rather upgrade astroid when we have the time for that and rebase the "require-astroid-update" MR and only them. It can be a warning so that we're aware of the problem sooner but if it's breaking pipeline overnight it's going to be more problem than it solves.

jacobtylerwalls:
I hear that, but I just want to make sure it's clear that the pipeline wouldn't fail, my proposal is only using this for the diff primer messages job which only emits a comment and doesn't fail any pipelines--I'm not proposing upgrading astroid to bleeding-edge version for pylint's CI in general.

.github/workflows/primer_run_main.yaml
```diff
@@ -51,11 +56,13 @@ jobs:
             ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
       - name: Create Python virtual environment
         if: steps.cache-venv.outputs.cache-hit != 'true'
+        # Use bleeding-edge astroid
         run: |
           python -m venv venv
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
+          pip install git+https://github.com/PyCQA/astroid.git
```
Pierre-Sassoulas, line 65:
```suggestion
          # Use bleeding-edge astroid
          pip install git+https://github.com/PyCQA/astroid.git
```

.github/workflows/primer_run_main.yaml
```diff
@@ -51,11 +56,13 @@ jobs:
             ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
       - name: Create Python virtual environment
         if: steps.cache-venv.outputs.cache-hit != 'true'
+        # Use bleeding-edge astroid
```
Pierre-Sassoulas, line 59:
```suggestion
```

.github/workflows/primer_run_pr.yaml
```diff
@@ -57,11 +62,13 @@ jobs:
             ${{ runner.os }}-${{ steps.python.outputs.python-version }}-venv-${{ env.CACHE_VERSION }}-
       - name: Create Python virtual environment
         if: steps.cache-venv.outputs.cache-hit != 'true'
+        # Use bleeding-edge astroid
         run: |
           python -m venv venv
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
+          pip install git+https://github.com/PyCQA/astroid.git
```
Pierre-Sassoulas, line 71:
```suggestion
        run: |
          python -m venv venv
          . venv/bin/activate
          python -m pip install -U pip setuptools wheel
          pip install -U -r requirements_test.txt
          # Use bleeding-edge astroid
          pip install git+https://github.com/PyCQA/astroid.git
```

Pierre-Sassoulas (approved):
All right if it's just a warning in the comment of the MR then it's actually a great improvement, 👍

jacobtylerwalls:
> if it's just a warning in the comment of the MR

Actually there won't be any warnings just yet, since I think any failures in astroid just are silenced by emitting `astroid-error` or other fatal messages, and we're not failing these primer runs on fatal messages.

So I think this still has value without a warning -- I'm thinking of subtle changes in astroid's brains or import system that sit for a couple months without being upgraded in pylint -- if a pylint contributor is about to cause false negatives/positives on a system we know we're eventually upgrading to anyway, I just wanted to make sure we captured those behavior differences.

DanielNoord:
@jacobtylerwalls Would you mind refactoring this to work with a label? As in, create a label that does `primer-astroid-bleeding-edge` which makes it so that the primer job runs with bleeding edge?

My rationale for this is that this is likely only relevant for PRs that will never get merged but where we want the info from the primer: if a PR needs `astroid` bleeding edge it won't be merged into `main` unless we release `astroid` itself. Thus, those situations should be contained to individual PRs and should be fine with only a label. That way we aren't running some part of our CI on bleeding edge and some of it not. Down the line I can see how differences like that can become confusing.

jacobtylerwalls:
I think we're possibly miscommunicating. :D

This isn't for PRs which "need" astroid bleeding-edge or PRs that will never be merged. I think you're imagining a scenario where we _know_ that pending astroid changes are relevant. Whereas what I'm imagining, and the entire point of the tool, is to _surface_ behavior changes we _were not expecting_. The idea is to find _unforeseen_ behavior changes between `pylint main + astroid bleeding edge` versus `pylint main + astroid bleeding edge + pylint PR`.

Example (inspired by real life :D):
- pylint contributor opens PR to fix false negative for `no-value-for-parameter`
- there is a merged PR already in astroid that the contributor doesn't know about, and the contributor isn't developing against bleeding-edge astroid
- the primer run doesn't pick up any differences because pylint is not keeping up with astroid
- 2 months later we upgrade astroid in pylint
- we release pylint
- 4 months later someone opens a bug, we bisect it to the changes in 1, but the primer missed it because astroid was lagging
- contributor is long gone

DanielNoord:
> I think we're possibly miscommunicating. :D
> 
> This isn't for PRs which "need" astroid bleeding-edge or PRs that will never be merged. I think you're imagining a scenario where we _know_ that pending astroid changes are relevant. Whereas what I'm imagining, and the entire point of the tool, is to _surface_ behavior changes we _were not expecting_. The idea is to find _unforeseen_ behavior changes between `pylint main + astroid bleeding edge` versus `pylint main + astroid bleeding edge + pylint PR`.
> 
> Example (inspired by real life :D):
> 
> * pylint contributor opens PR to fix false negative for `no-value-for-parameter`
> * there is a merged PR already in astroid that the contributor doesn't know about, and the contributor isn't developing against bleeding-edge astroid
> * the primer run doesn't pick up any differences because pylint is not keeping up with astroid
> * 2 months later we upgrade astroid in pylint
> * we release pylint
> * 4 months later someone opens a bug, we bisect it to the changes in 1, but the primer missed it because astroid was lagging
> * contributor is long gone

Thanks, that explanation helped clear things up.

I guess the only situation in which this would break is if `astroid` bleeding edge introduces a non-backwards compatible change that breaks all `pylint` runs (which is likely what was the original rationale for not running CI on it). Then the whole primer job might fail to produce any meaningful output.
However, that's probably a good thing was we wouldn't want that. Perhaps we should add a check for a change in major version and not use bleeding edge if it changes the major version? Or we fix that issue whenever it actually comes up 😄 
I would be fine with both, but perhaps link to the discussion with a comment. Simply disabling bleeding edge while `astroid` is on `3.0.0+dev` should work.

Pierre-Sassoulas:
>  Perhaps we should add a check for a change in major version and not use bleeding edge if it changes the major version?

Yeah, let's cross that bridge when we come to it 😄 we did not release a major in the last 5 years

.github/workflows/primer_run_main.yaml
```diff
@@ -32,13 +32,18 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Get latest astroid commit
+        id: get-astroid-sha
+        run: |
+          curl https://api.github.com/repos/PyCQA/astroid/commits |
+          python -c "import json, sys; print(json.load(sys.stdin)[0]['sha'])" > astroid_sha.txt
       # Restore cached Python environment
```
DanielNoord, line 40:
```suggestion

      # Restore cached Python environment      
```

.github/workflows/primer_run_pr.yaml
```diff
@@ -38,13 +38,18 @@ jobs:
           node-version: 16
       - run: npm install @octokit/rest
 
+      - name: Get latest astroid commit
+        id: get-astroid-sha
+        run: |
+          curl https://api.github.com/repos/PyCQA/astroid/commits |
+          python -c "import json, sys; print(json.load(sys.stdin)[0]['sha'])" > astroid_sha.txt
       # Restore cached Python environment
```
DanielNoord, line 46:
```suggestion

      # Restore cached Python environment
```

DanielNoord (approved):

jacobtylerwalls:
> I guess the only situation in which this would break is if astroid bleeding edge introduces a non-backwards compatible change that breaks all pylint runs (which is likely what was the original rationale for not running CI on it). Then the whole primer job might fail to produce any meaningful output.

I think as it is, we wrap any astroid failures in `astroid-error` and we don't fail the run. We would just have false-negative passing primer runs where n instances of `astroid-error` == n instances of `astroid-error`. So no change from status quo with this PR.

We ought to do something, like print a message in the comment: fatal errors found or something similar.

PR #6725, opened 2022-05-28
Primer tool: remove duplicated directory paths

jacobtylerwalls:
## Type of Changes

## Description

`pylint_args` reads `paths_to_lint`, which reads `directories`, so adding `directories` here is both duplicative and error-prone (because it doesn't account for the path to the cloned directory): we were getting fatal errors like "unable to import module src/blackd")

In a follow-up PR I'll probably add `--fail-on=fatal`.

### DIFF
--- a/tests/primer/primer_tool.py
+++ b/tests/primer/primer_tool.py
@@ -117,7 +117,7 @@ class Primer:
         enables = ["--enable-all-extensions", "--enable=all"]
         # Duplicate code takes too long and is relatively safe
         disables = ["--disable=duplicate-code"]
-        arguments = data.directories + data.pylint_args + enables + disables
+        arguments = data.pylint_args + enables + disables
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Nice catch !

DanielNoord:
> ## Type of Changes
> Type
> ✓	🐛 Bug fix
> ## Description
> `pylint_args` reads `paths_to_lint`, which reads `directories`, so adding `directories` here is both duplicative and error-prone (because it doesn't account for the path to the cloned directory): we were getting fatal errors like "unable to import module src/blackd")
> 
> In a follow-up PR I'll probably add `--fail-on=fatal`.

This also explains why I was getting double messages. I opted to ignore those, but this seems like it was the culprit. Stupid mistake, thanks for fixing!

Issue #6670, opened 2022-05-23
Deleted messages and checkers are still being searched on ReadTheDoc but they are not documented at all in latest version


Pierre-Sassoulas:
### Question

See [this message for context](https://github.com/PyCQA/pylint/issues/5953#issuecomment-1134172901)

### Documentation for future user

Document the deleted messages, maybe with the version of removal and the reasoning behind deletion.

DanielNoord:
I would be in favour of just closing this. It doesn't seem worth our effort, especially now that we warn about non-existing messages in people's configuration by default.

Pierre-Sassoulas:
I started implementing something I just prioritized  reviews and PyCQA drama :)

PR #6727, opened 2022-05-28
Refactor to be able to document deleted message

Pierre-Sassoulas:
## Type of Changes

## Description

Refactor before #6670

### DIFF
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -60,61 +60,72 @@ def _register_all_checkers_and_extensions(linter: PyLinter) -> None:
 
 def _get_message_data(data_path: Path) -> Tuple[str, str, str, str]:
     """Get the message data from the specified path."""
-    good_code, bad_code, details, related = "", "", "", ""
     good_py_path = data_path / "good.py"
+    bad_py_path = data_path / "bad.py"
     details_rst_path = data_path / "details.rst"
+    related_rst_path = data_path / "related.rst"
     if not data_path.exists():
-        data_path.mkdir(parents=True)
-        with open(good_py_path, "w", encoding="utf-8") as file:
-            file.write(
-                """\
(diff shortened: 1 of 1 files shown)


doc/exts/pylint_messages.py
```diff
@@ -60,61 +60,72 @@ def _register_all_checkers_and_extensions(linter: PyLinter) -> None:
...
-{file_content_string}"""
-    return good_code, bad_code, details, related
+def _create_placeholders(
+    data_path: Path, details_rst_path: Path, good_py_path: Path
+) -> None:
+    data_path.mkdir(parents=True)
+    with open(good_py_path, "w", encoding="utf-8") as file:
+        file.write(
+            """\
+# This is a placeholder for correct code for this message.
```
DanielNoord, line 120: Shall we remove those other files then?

Pierre-Sassoulas: What files are we talking about ?

DanielNoord: The files you added recently in the placeholder PR.

Pierre-Sassoulas: Ha, right. I was thinking I added the function at the same time than the files but it was not the case. Well I think it's a good thinq that we can generate easily but also that a new contributor can just use the already generated file when he's doing the doc addition, so we might keep both.

DanielNoord: Yeah the only thing I worried about is that we might forget to remove one of the placeholder files if we update the other, as the placeholder text is divided over two different files.

Pierre-Sassoulas: Yeah this is a little dangerous if the contributor forget to remove the details.rst and there's no detail the review seems fine at first glance but it's not.

Pierre-Sassoulas: I added an assert

DanielNoord (approved):

PR #6729, opened 2022-05-28
Add ``pytest`` to new primer and exclude some file changes

DanielNoord:

## Type of Changes

## Description

Ref. https://github.com/PyCQA/pylint/pull/6723#issuecomment-1140323738

### DIFF
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -11,6 +11,11 @@ on:
       - "tests/primer/**"
       - "requirements*"
       - ".github/workflows/**"
+      # We ignore these specific files because they need to be changed
+      # on 'main' and will always fail the PR run.
+      - "!.github/workflows/primer_run_main.yaml"
+      - "!.github/workflows/primer_comment.yaml"
+      - "!tests/primer/packages_to_prime.json"
 
 env:
   CACHE_VERSION: 1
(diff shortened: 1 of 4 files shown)


DanielNoord:
@jacobtylerwalls Correct me if I'm wrong, but if I understand the documentation correctly (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) whenever one of the files in a PR does not match any of the ignored patterns we will still run.

So because we remove `pytest` is removed from the old file we still get a failed run here.

.github/workflows/primer_run_pr.yaml
```diff
@@ -11,6 +11,11 @@ on:
       - "tests/primer/**"
       - "requirements*"
       - ".github/workflows/**"
+      # We ignore these specific files because they need to be changed
+      # on 'main' and will always fail the PR run.
+      - "!.github/workflows/primer_run_main.yaml"
+      - "!.github/workflows/primer_comment.yaml"
```
jacobtylerwalls, line 17: Why will these two always fail? #6724 didn't fail.

DanielNoord: It passed because there wasn't any difference between using bleeding edge or latest release. `run_main` didn't use bleeding edge yet because the workflow wasn't on `main` yet.

I'd say we don't want to give a false sense of security and just exclude any PRs that only change `main` files.

jacobtylerwalls: OK to merge -- not worth trying to anticipate more small targeted PRs like #6724 which should be rare.

jacobtylerwalls (approved):

PR #6730, opened 2022-05-28
Small refactors to primer

DanielNoord:

## Type of Changes

## Description

Ref. https://github.com/PyCQA/pylint/pull/6723#issuecomment-1140323738

### DIFF
--- a/tests/primer/primer_tool.py
+++ b/tests/primer/primer_tool.py
@@ -8,6 +8,7 @@ import argparse
 import json
 from io import StringIO
 from pathlib import Path
+from typing import Dict, List, Union
 
 import git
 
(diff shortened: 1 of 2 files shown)


tests/primer/primer_tool.py
```diff
@@ -8,6 +8,7 @@
 import json
 from io import StringIO
 from pathlib import Path
+from typing import Dict, List, Union
```
jacobtylerwalls, line 11: any reason not to do __future__ style types?

DanielNoord: `PackageMessages` is created at runtime so we need the `typing` imports.

`if TYPE_CHECKING` could also work, but so far our "style" has been to keep typing variables in the main scope and just use `from typing` imports.

jacobtylerwalls (approved):

PR #6731, opened 2022-05-28
Add concurrency to primer

DanielNoord:

## Type of Changes

## Description

Based on tips by original `mypy` primer author over at https://github.com/hauntsaninja/mypy_primer/issues/29.

I considered adding `pre-commit` as a non-triggering actor like they do, but that doesn't work with concurrency. The `pre-commit` commit would trigger the concurrency so then no run would succeed.


Pierre-Sassoulas (approved):
👍

PR #6735, opened 2022-05-29
[doc] Add a check for changelogs and fix the issues encountered

Pierre-Sassoulas:
## Type of Changes

## Description

Refs #6688

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,7 +78,6 @@ repos:
         alias: check-changelog
         name: check-changelog
         types: [text]
-        args: ["--verbose"]
         entry: python3 script/check_changelog.py
         pass_filenames: false
         language: system
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -74,6 +74,14 @@ repos:
         args: ["-rn", "-sn", "--rcfile=pylintrc", "--fail-on=I", "--spelling-dict=en"]
         exclude: tests/functional/|tests/input|tests(/\w*)*data/|doc/
         stages: [manual]
+      - id: check-changelog
+        alias: check-changelog
+        name: check-changelog
+        types: [text]
+        args: ["--verbose"]
+        entry: python3 script/check_changelog.py
+        pass_filenames: false
+        language: system
       - id: fix-documentation
         name: Fix documentation
         entry: python3 -m script.fix_documentation
--- a/doc/whatsnew/0/0.x.rst
+++ b/doc/whatsnew/0/0.x.rst
@@ -658,7 +658,7 @@ Release date: 2006-08-10
 * started a reference user manual
 
 * new W0212 message for access to protected member from client code
-  (close #14081)
+  (Closes #14081)
 
 * new W0105 and W0106 messages extracted from W0104 (statement seems
   to have no effect) respectively when the statement is actually string
@@ -668,7 +668,7 @@ Release date: 2006-08-10
 * reclassified W0302 to C0302
 
 * fix so that global messages are not anymore connected to the last
-  analyzed module (close #10106)
+  analyzed module (Closes #10106)
 
 * fix some bugs related to local disabling of messages
 
@@ -708,7 +708,7 @@ Release date: 2006-03-06
 
 * W0705 has been reclassified to E0701, and is now detecting more
   inheriting problem, and a false positive when empty except clause is
-  following an Exception catch has been fixed (close #10422)
+  following an Exception catch has been fixed (Closes #10422)
 
 * E0212 and E0214 (metaclass/class method should have mcs/cls as first
   argument have been reclassified to C0202 and C0203 since this not as
@@ -812,13 +812,13 @@ Release date: 2005-11-07
   module whose name starts with a deprecated module's name (close
   #10061)
 
-* fix "module has no name __dict__" false positive (close #10039)
+* fix "module has no name __dict__" false positive (Closes #10039)
 
 * fix "access to undefined variable __path__" false positive (close
   #10065)
 
 * fix "explicit return in __init__" false positive when return is
-  actually in an inner function (close #10075)
+  actually in an inner function (Closes #10075)
 
 
 What's New in Pylint 0.8.0?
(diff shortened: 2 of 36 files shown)


doc/whatsnew/0/0.x.rst
```diff
@@ -658,7 +658,7 @@ Release date: 2006-08-10
 * started a reference user manual
```
Pierre-Sassoulas, line 658: Not touching this file (unless it's a global sed) because it's a mix between Logilab's internal and bitbucket. Hard to tell, probably dead link anyway, and not very useful generally.

script/check_changelog.py
```diff
@@ -0,0 +1,105 @@
...
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from collections.abc import Iterator
+from pathlib import Path
+from re import Pattern
+
+ISSUE_NUMBER_PATTERN = re.compile(r"#\d{1,5}")
```
DanielNoord, line 18: Can't we reuse this pattern in the other patterns?

```suggestion
ISSUE_NUMBER_PATTERN = re.compile(r"#\d+")
```

script/check_changelog.py
```diff
@@ -0,0 +1,105 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
DanielNoord, line 1: Is this file used anywhere? Or have I missed something?

doc/user_guide/checkers/features.rst
```diff
@@ -1198,7 +1198,8 @@ Refactoring checker Messages
   This message is emitted when pylint encounters boolean operation like"a < b
   and b < c", suggesting instead to refactor it to "a < b < c"
 :simplifiable-if-expression (R1719): *The if expression can be replaced with %s*
-  Used when an if expression can be replaced with 'bool(test)'.
+  Used when an if expression can be replaced with 'bool(test)' or simply 'test'
```
DanielNoord, line 1201: This shouldn't really be in this PR. Not sure if we can make the documentation build as a pre-commit check in CI?

script/check_changelog.py
Pierre-Sassoulas: I launch it manually for now.

doc/user_guide/checkers/features.rst
Pierre-Sassoulas: We ruled against that at some point and decided to only upgrade with tbump for each release, but it does not make sense anymore as we added a lot of generated content and it would make sense to have to review the doc changes too for each code change. What do you think ?

DanielNoord: Yeah, I'm starting to wonder if adding the generated content was the right decision. It feels like it is giving us a lot of headaches at the moment. I would definitely be in favour of adding a `make html` step in the `pre-commit` CI run. But only in CI though, it would be very annoying to have to wait on the documentation build for every commit.

script/check_changelog.py
Pierre-Sassoulas: I see you're an optimist 😄

doc/user_guide/checkers/features.rst
Pierre-Sassoulas: Regarding the generated content I think it's the right decision, it's possible to modify a docstring on a checker that would break the doc if it's incompatible with RST, and if it's generated only at release we'll never catch it during review. Contributor can't know about every scripts that generates the documentation and the implication each change they do entice. Let me do something about this pre-commit step.

Pierre-Sassoulas: #6736

script/check_changelog.py
Pierre-Sassoulas: I added it as a pre-commit hook, it's fast enough.

doc/user_guide/checkers/features.rst
Pierre-Sassoulas: I just upgraded the generated content in #6775 because the generated content is unstable for some default value that depends on the OS or interpreter and #6736 is blocked.

.pre-commit-config.yaml
```diff
@@ -74,6 +74,14 @@ repos:
         args: ["-rn", "-sn", "--rcfile=pylintrc", "--fail-on=I", "--spelling-dict=en"]
         exclude: tests/functional/|tests/input|tests(/\w*)*data/|doc/
         stages: [manual]
+      - id: check-changelog
+        alias: check-changelog
+        name: check-changelog
+        types: [text]
+        args: ["--verbose"]
```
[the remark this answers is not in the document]
Pierre-Sassoulas:
```suggestion
```

script/check_changelog.py
DanielNoord: Still, let's reuse the pattern so if we ever need to change it that's just in one place.

script/check_changelog.py
```diff
@@ -0,0 +1,131 @@
...
+    # Not very useful generally, unless you're an open source historian.
+    "0.x",
+    # Too much Bitbucket issues in this one :
+    "1.0",
+    "1.1",
+    "1.2",
+]
+
+NO_CHECK_REQUIRED_FILES = {
+    "index.rst",
```
DanielNoord, line 38: Shouldn't these be excluded in pre-commit? Feels a bit weird to exclude them here.

DanielNoord (changes requested):

script/check_changelog.py
DanielNoord: Hm, doesn't it make more sense to let this accept a `files` argument? That's how most of the linting, formatting, etc. tools work

script/check_changelog.py
Pierre-Sassoulas: Done in [9a60d73](https://github.com/PyCQA/pylint/pull/6735/commits/9a60d73c49cc293a666f5ed32b3dc685e19f8872)  with some bonus modularity

script/check_changelog.py
Pierre-Sassoulas: It does the job, we won't use it that much (when we modify the old changelog), it's not perfect but I don't see the point in refactoring it to perfection right now as it's blocking 2.14. Let's refactor it if we're going to keep the 
```
* blabla

  Closes #5445345
```  
format in #6688. Right now I think ``* blabla (#123456789)`` with 123456789 being the PR number and not the issue number would be better, which means we would delete this.

Pierre-Sassoulas: Also see #6781, if we auto-generate the changelog we won't need it either

PR #6736, opened 2022-05-29
[doc] Add a CI check for the doc being properly generated

Pierre-Sassoulas:
## Type of Changes

## Description

Refs #5953 

Refs https://github.com/PyCQA/pylint/pull/6735#discussion_r884226339

Require #8949, #8950, #8951, #8956

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -182,4 +182,8 @@ jobs:
         run: |
           . venv/bin/activate
           cd doc
-          make html
+          pre-commit run --hook-stage push sphinx-generated-doc --all-files || {
+            git diff ; \
+            echo "Make sure that there are no modifications locally when launching 'make html'" ; \
+            exit 1; \
+          }
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -94,6 +94,13 @@ repos:
         args: ["-rn", "-sn", "--rcfile=pylintrc", "--fail-on=I", "--spelling-dict=en"]
         exclude: tests(/\w*)*/functional/|tests/input|tests(/\w*)*data/|doc/
         stages: [manual]
+      - id: sphinx-generated-doc
+        alias: sphinx-generated-doc
+        name: sphinx-generated-doc
+        entry: make -C doc/ html
+        pass_filenames: false
+        language: system
+        stages: [push]
       - id: check-newsfragments
         name: Check newsfragments
         entry: python3 -m script.check_newsfragments
--- a/doc/development_guide/contributor_guide/tests/install.rst
+++ b/doc/development_guide/contributor_guide/tests/install.rst
@@ -28,6 +28,10 @@ This ensures your testing environment is similar to Pylint's testing environment
 pre-commit hooks which should take care of the autoformatting for you before each
 commit. To enable it, run ``pre-commit install`` in the ``pylint`` root directory.
 
+**Even more optionally**: You can enable slow on push hooks with ``pre-commit install --install-hooks -t pre-push``.
+It will do slow checks like checking that the generated documentation is up to date
+before each push.
+
 Astroid installation
 --------------------
 


.pre-commit-config.yaml
```diff
@@ -74,6 +74,13 @@ repos:
         args: ["-rn", "-sn", "--rcfile=pylintrc", "--fail-on=I", "--spelling-dict=en"]
         exclude: tests/functional/|tests/input|tests(/\w*)*data/|doc/
         stages: [manual]
+      - id: sphinx-generated-doc
+        alias: sphinx-generated-doc
+        name: sphinx-generated-doc
+        entry: make -C doc/ build-html
```
Pierre-Sassoulas, line 80:
```suggestion
        entry: make -C doc/ html
```

.github/workflows/checks.yaml
```diff
@@ -190,3 +190,4 @@ jobs:
           . venv/bin/activate
           cd doc
           make html
```
Pierre-Sassoulas, line 192:
```suggestion
```

.github/workflows/checks.yaml
```diff
@@ -189,4 +189,4 @@ jobs:
         run: |
           . venv/bin/activate
           cd doc
-          make html
+          pre-commit run --hook-stage manual pylint-with-spelling --all-files
```
Pierre-Sassoulas, line 192:
```suggestion
          pre-commit run --hook-stage manual sphinx-generated-doc --all-files
```

Pierre-Sassoulas:
Task failed successfully. I think we need a better error message because it's very unclear that the issue is that the file were modified during the make. Ideally pre-commit could autofix, but I don't know how well that work with the "manual" option.

Pierre-Sassoulas:
I think that adding this hook as pre-push is reasonable. The alternative I think would be to add it all the time but only if some specific file are modified like checkers in pylint/checkers and python file in doc/exts/. 

Con/pro of the pre-push hook:
- No automatic fix
- Need to install pre-commit with command that is not well known imo
+ Not launched for every commit locally
 

Con/pro of the pre-commithook:
+ Automatic fix online
+ It's very slow, we're going to ``--no-verify`` a lot of commits  if we don't do something about it

.github/workflows/checks.yaml
```diff
@@ -189,4 +189,5 @@ jobs:
         run: |
           . venv/bin/activate
           cd doc
-          make html
+          pre-commit run --hook-stage push sphinx-generated-doc --all-files||echo "Make sure that there\
```
Pierre-Sassoulas, line 192:
```suggestion
          pre-commit run --hook-stage push sphinx-generated-doc --all-files||echo "Make sure that there \
```

doc/development_guide/contributor_guide/contribute.rst
```diff
@@ -27,12 +27,14 @@ of Pylint as it gives you access to the latest ``ast`` parser.
   For more information on how to use our test suite and write new tests see :ref:`testing`.
 
 - ``pylint`` uses black_ and isort_ among other Python auto-formatters.
-  We have a pre-commit hook which should take care of the autoformatting for
+
+  We have pre-commit hooks which should take care of the autoformatting for
   you. To enable it, do the following:
 
     * install ``pre-commit`` using ``pip install pre-commit``
```
Pierre-Sassoulas, line 35:
```suggestion
  you. To enable it, run ``pre-commit install --install-hooks -t pre-push -t pre-commit`` in the ``pylint`` root
  directory.
```

doc/development_guide/contributor_guide/contribute.rst
```diff
@@ -27,12 +27,14 @@ of Pylint as it gives you access to the latest ``ast`` parser.
...
 
-    * then run ``pre-commit install`` in the ``pylint`` root directory to enable the git hooks.
+    * then run ``pre-commit install --install-hooks -t pre-push -t pre-commit`` in the ``pylint`` root
+      directory to enable the git hooks.
```
Pierre-Sassoulas, line 37:
```suggestion
```

DanielNoord:
I'm still not sure about adding this as a local `pre-commit` hook. I worry that I would start using `--no-verify` quite a bit more. For me creating the documentation takes 30+ seconds while also spinning up all fans my laptop can find 😅 When I'm working on the `primer` for example this is something I would like to avoid.

On `pydocstringformatter` I created a workflow that builds the documentation and then checks if there is a `diff` after that build. If there is we fail the workflow. Would that be okay here as well?

doc/development_guide/contributor_guide/contribute.rst
```diff
@@ -27,12 +27,9 @@ of Pylint as it gives you access to the latest ``ast`` parser.
...
 
 - ``pylint`` uses black_ and isort_ among other Python auto-formatters.
-  We have a pre-commit hook which should take care of the autoformatting for
-  you. To enable it, do the following:
 
-    * install ``pre-commit`` using ``pip install pre-commit``
-
-    * then run ``pre-commit install`` in the ``pylint`` root directory to enable the git hooks.
+  We have pre-commit hooks which should take care of the autoformatting for
+  you. To enable it, run ``pre-commit install --install-hooks -t pre-push -t pre-commit`` in the ``pylint`` root directory.
```
jacobtylerwalls, line 32: If this is opt-in, then I guess it's alright. Daniel and I can opt-out. :D

Pierre-Sassoulas: It's launched just before pushing not for every commit. I could also limit what trigger the check.

Pierre-Sassoulas:
@DanielNoord @jacobtylerwalls sorry I changed the contributor doc a little more than strictly necessary but it was in such a state of disrepair that I could not help it.

Pierre-Sassoulas:
Let it be noted that the tests fail successfully before the generation of the doc on [cc2b93a](https://github.com/PyCQA/pylint/pull/6736/commits/cc2b93abebe3af36a167eb330e578a4edba5af2f)

DanielNoord:
@Pierre-Sassoulas I tried to understand the documentation and code for `pre-commit.ci`.

`ci` is triggered with: ``'python3', '-mpre_commit', 'run', '--all-files'``.
I'm not sure but does this also trigger hooks in the `manual` stage? If that's the case we can just add this as a manual hook and not add it to pre-push.

.github/workflows/checks.yaml
```diff
@@ -190,4 +190,5 @@ jobs:
         run: |
           . venv/bin/activate
           cd doc
-          make html
+          pre-commit run --hook-stage push sphinx-generated-doc --all-files || { echo "Make sure that there \
+          are no modification locally when launching 'make html'" ; exit 1; }
```
jacobtylerwalls, line 194:
```suggestion
          are no modifications locally when launching 'make html'" ; exit 1; }
```

.github/workflows/checks.yaml
```diff
@@ -190,4 +190,5 @@ jobs:
         run: |
           . venv/bin/activate
           cd doc
-          make html
+          pre-commit run --hook-stage push sphinx-generated-doc --all-files || { echo "Make sure that there \
```
DanielNoord, line 193: How does this work? This won't exit if there are any added files right? We need to do some kind of `git status` for that?

Pierre-Sassoulas: It's launched all the time, you can try ``pre-commit run --all-fies`` locally, there's just some additional option for manual hooks.

Pierre-Sassoulas:
This PR is still blocked because some default value are OS/interpreter dependant.

.github/workflows/checks.yaml
DanielNoord: No, I meant: how does this fail if the doc hasn't been updated properly? Won't this only fail of the `make` commands throws an error?

Pierre-Sassoulas: Pre-commit fail is a file was modified during the run (it's dealing with autoformatter often).

jacobtylerwalls (dismissed):
Tested, works when pre-push hook is enabled, off by default. Looks good!

I started on a PR for something like this last week, forgetting that this one existed. Good sign to merge it :D

jacobtylerwalls:
> This PR is still blocked because some default value are OS/interpreter dependant.

☝️ still relevant?

DanielNoord:
> > This PR is still blocked because some default value are OS/interpreter dependant.
> 
> 
> 
> ☝️ still relevant?

Sadly it is.. `py-version`'s generated value is based on the current interpreter..

jacobtylerwalls:
Oh. Can you walk me through that briefly? I couldn't find where it was discussed. What's the impact?

Pierre-Sassoulas:
https://github.com/pylint-dev/pylint/pull/6736#issuecomment-1142177076. There wasn't much discussed, there's only a few option that are sys env dépendant but something need to be designed so the doc understand that this is dynamic and not a constant value.

DanielNoord:
You can see it in this PR. There is a discrepancy between what we build on CI and what we currently have on `master`. I'm not sure we can ever resolve these, as it makes sense for `py-version` to be interpreter based as a default.

We could also just make sure the CI runner is on the latest version and keep that as expected values?

jacobtylerwalls:
> You can see it in this PR. 

Sorry, that's still what I need help with. How can I see a doc diff from the Sphinx logs?

jacobtylerwalls:
The docs on `main` are well out of date, three files need to be updated. But none of those have to do with `py-version`.

DanielNoord:
> > You can see it in this PR.
> 
> Sorry, that's still what I need help with. How can I see a doc diff from the Sphinx logs?

You can't 😓 

You can probably test it by running the `pre-commit` hook as called in `.github/workflows/checks.yaml` on `3.10`. There will be meaningful updates (as we haven't updated in a while) but also one to `py-version`.

Pierre-Sassoulas:
It's possible to fix this by adding a special handling to dynamic values so we have a description of the default for the Doc or for argparse help (as well as the actual value ?). If we don't show the actual value in Doc but only the description it's already fixed, but we probably want the actual value in argparse help.

We should show the diff in the script output so the message is clearer too and maybe it's possible to use the CI value and use it to make the generated Doc identical even if locally we don't have the same output (i.e. using the CI value in the Doc and upgrading manually). It will get old fast if you can never upgrade the Doc automatically, there would be a partial git add to do for each PR for everyone.

jacobtylerwalls:
> It's possible to fix this by adding a special handling to dynamic values

Yeah. I'd just think that's too much effort.

> It will get old fast if you can never upgrade the Doc automatically, there would be a partial git add to do for each PR for everyone.

But only if it's a pre-commit hook, correct? Maybe this is better not as a pre-commit hook but just a simple check like this, which is basically the PR I started to open last week:

> On pydocstringformatter I created a workflow that builds the documentation and then checks if there is a diff after that build. If there is we fail the workflow. Would that be okay here as well?

DanielNoord:
> > On pydocstringformatter I created a workflow that builds the documentation and then checks if there is a diff after that build. If there is we fail the workflow. Would that be okay here as well?

That's what we have here as well! :smile: 

The main issue is that it blocks all contributions to be made from the same interpreter as our CI is running. A contributor on `3.8` who changes something in the docs build will never be able to fix it.
If we don't see that as a big issue I think we can rebase this and merge the current approach.

jacobtylerwalls:
> That's what we have here as well! 😄

Sorry, I'm not understanding 😄 , don't we have a pre-commit check here that will check every push, causing false positives every time someone tries to push on 3.10? A CI check only (as opposed to a prepush hook) wouldn't cause that problem.

> A contributor on 3.8 who changes something in the docs build will never be able to fix it.

"Never" seems pretty strong for "use common sense and don't alter a setting that's clearly unrelated to your changes" 😄. Again, another reason for it not to be a pre-commit hook, so that you can still get your changes pushed. Or am I missing something?

jacobtylerwalls:
I guess we could merge the current approach, given how unlikely it is that people will opt in to the expensive pre-push hook, and if they do, they're a super-contributor who should get on the same interpreter as CI anyway. :D

DanielNoord:
> Sorry, I'm not understanding smile , don't we have a pre-commit check here that will check every push, causing false positives every time someone tries to push on 3.10? A CI check only (as opposed to a prepush hook) wouldn't cause that problem.

Oh I misread.. I think this should be a `manual` `pre-commit` hook indeed. Building CI really takes too long to do on every push I think.
 


> I guess we could merge the current approach, given how unlikely it is that people will opt in to the expensive pre-push hook, and if they do, they're a super-contributor who should get on the same interpreter as CI anyway. :D

No, I agree!

doc/development_guide/contributor_guide/tests/install.rst
```diff
@@ -27,6 +27,10 @@ This ensures your testing environment is similar to Pylint's testing environment
 pre-commit hooks which should take care of the autoformatting for you before each
 commit. To enable it, run ``pre-commit install`` in the ``pylint`` root directory.
 
+**Even more optionally**: You can enable slow on push hooks with ``pre-commit install --install-hooks -t pre-push``.
```
jacobtylerwalls, line 30: @DanielNoord here is where you have to opt-in. I think we're good. If a contributor pushes up the changes with py-version, sees it fail again, they'll look at the diff and figure it out.

Pierre-Sassoulas:
```suggestion
**Even more optionally**: You can enable slow on push hooks with ``pre-commit install --install-hooks -t pre-push``. You will need to use the same python interpreter and enchant dict installed than the CI for it to work.
```

Pierre-Sassoulas:
This is going to be cumbersome as is. Maybe we can add a flag on option to indicate that the doc need to render the string from the default value instead of the actual value ? (I.e. ``sys.version_info`` or ``_get_enchant_dict_choices(enchant_dicts)`` instead of ``3.8`` or ``{"en", "en_gb", "fr_ca")`` respectively)

doc/development_guide/contributor_guide/tests/install.rst
Pierre-Sassoulas: Not applying my own suggestion as I'm going to fix it in the script.

Pierre-Sassoulas:
I chose to hard code the value in a dict, the generic way to do it was not worth it as we have so little dynamically generated value.

Pierre-Sassoulas:
Sorry for the ping on this one, I'm going to create 2 new merge requests to not mix things up (#8949 and #8950).

DanielNoord (dismissed):

Pierre-Sassoulas (commented):
I'm thinking that it could be nice to launch the fast python generation directly in a pre-commit hook but I want to prioritize the conf upgrade tool atm

DanielNoord (approved):

PR #6741, opened 2022-05-29
Added documentation examples for `no-method-argument`.

mbyrnepr2:
## Type of Changes

## Description

### DIFF
--- a/doc/data/messages/n/no-method-argument/good.py
+++ b/doc/data/messages/n/no-method-argument/good.py
@@ -1,6 +1,5 @@
 class Person:
-    def __init__(self):
-        self.greeting = "hello"
+    greeting = "hello"
 
     def print_greeting(self):
         print(self.greeting)
--- a/doc/data/messages/n/no-method-argument/good.py
+++ b/doc/data/messages/n/no-method-argument/good.py
@@ -1,5 +1,3 @@
 class Person:
-    greeting = "hello"
-
     def print_greeting(self):
-        print(self.greeting)
+        print("hello")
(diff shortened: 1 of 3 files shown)


doc/data/messages/n/no-method-argument/good.py
```diff
@@ -1 +1,6 @@
-# This is a placeholder for correct code for this message.
+class Person:
+    def __init__(self):
+        self.greeting = "hello"
```
Pierre-Sassoulas, line 3: Is the init required here ?

mbyrnepr2: No good point!

doc/data/messages/n/no-method-argument/good.py
```diff
@@ -1 +1,5 @@
-# This is a placeholder for correct code for this message.
+class Person:
+    greeting = "hello"
+
+    def print_greeting(self):
+        print(self.greeting)
```
Pierre-Sassoulas, line 5:
```suggestion
    def print_greeting(self):
        print("hello")
```
Can't we do that ? The self is unused but we're illustrating the no-method-argument message so it's not a problem, right ?

mbyrnepr2: Ah ok like this. Sure. I thought that would break the tests if more than the expected message was emitted.

Pierre-Sassoulas: We disable everything except the message that is documented, this is why we can have faulty code for clarity ``isinstance(thing, Animal)`` when ``Animal`` is not defined for example).

PR #6746, opened 2022-05-30
Fail "new primer" runs on fatal errors, like "old primer"

jacobtylerwalls:
## Type of Changes

## Description
The old primer did one thing very well -- it failed CI jobs when encountering `fatal` errors during the linting of open source packages, whether caused in new pylint or astroid PRs, or when adding new packages to the primer, or when third-party packages pushed new changes.

The new primer needs some analogous functionality -- something beyond the message differ. If some fatal errors "sneak in" (we merge a PR without noticing the change, or we believe it to be related to a different PR, or incompatible PRs are merged, or we add a package, or third-party maintainers publish new commits), we don't want `n` fatal errors on main to silently match `n` fatal errors on a PR and for the comment-bot to stay quiet about it. We want a failure.

I persuaded (💪🏻 😅 🙏🏻 📈 )  us into running the primer on bleeding-edge astroid, and from the feedback that we want to be careful not to introduce a dependency in pylint's CI on bleeding-edge astroid with respect to failing jobs (because that is more significant than simply the contents of a bot's comments on a PR), I filter those out here and just issue a warning instead.

### DIFF
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -97,7 +97,7 @@ jobs:
           . venv/bin/activate
           pip install -e .
           python tests/primer/primer_tool.py run --type=main 2>warnings.txt
-          WARNINGS=$(<warnings.txt)
+          WARNINGS=$(head -c 65000 < warnings.txt)
       - name: Emit Actions warnings for python warnings
         run: |
           if [[ $WARNINGS ]]
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -98,8 +98,6 @@ jobs:
           pip install -e .
           python tests/primer/primer_tool.py run --type=main 2>warnings.txt
           WARNINGS=$(head -c 65000 < warnings.txt)
-      - name: Emit Actions warnings for python warnings
-        run: |
           if [[ $WARNINGS ]]
           then echo "::warning ::$WARNINGS"
           fi
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -96,7 +96,13 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          python tests/primer/primer_tool.py run --type=main
+          python tests/primer/primer_tool.py run --type=main 2>warnings.txt
+          WARNINGS=$(<warnings.txt)
+      - name: Emit Actions warnings for python warnings
+        run: |
+          if [[ $WARNINGS ]]
+          then echo "::warning ::$WARNINGS"
+          fi
       - name: Upload output
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -169,8 +169,6 @@ jobs:
           pip install -e .
           python tests/primer/primer_tool.py run --type=pr 2>warnings.txt
           WARNINGS=$(<warnings.txt)
-      - name: Emit Actions warnings for python warnings
-        run: |
           if [[ $WARNINGS ]]
           then echo "::warning ::$WARNINGS"
           fi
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas (approved):
I think it's the best of both world 👌

tests/primer/primer_tool.py
```diff
@@ -130,6 +132,22 @@ def _handle_run_command(self) -> None:
...
+        # for the purpose of diffing emitted messages and generating PR comments.
+        messages = list(chain.from_iterable(packages.values()))
+        astroid_errors = [msg for msg in messages if msg["symbol"] == "astroid-error"]
+        other_fatal_msgs = [
+            msg
+            for msg in messages
+            if msg["type"] == "fatal" and msg["symbol"] != "astroid-error"
+        ]
+        if astroid_errors:
+            warnings.warn(f"Fatal errors traced to astroid:\n{astroid_errors}")
```
DanielNoord, line 148: Does this automatically translate into an Actions warning? (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message)

If not, that would be nice I guess.

Pierre-Sassoulas: Maybe if we print something if would work like ``echo "::warning file=app.js,line=1,col=5,endColumn=7::Missing semicolon"`` from the example ?

.github/workflows/primer_run_main.yaml
```diff
@@ -96,7 +96,13 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          python tests/primer/primer_tool.py run --type=main
+          python tests/primer/primer_tool.py run --type=main 2>warnings.txt
+          WARNINGS=$(<warnings.txt)
+      - name: Emit Actions warnings for python warnings
```
DanielNoord, line 101: Since both of these steps are a normal `bash run`, can't they be merged?

.github/workflows/primer_run_pr.yaml
```diff
@@ -167,7 +167,13 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          python tests/primer/primer_tool.py run --type=pr
+          python tests/primer/primer_tool.py run --type=pr 2>warnings.txt
+          WARNINGS=$(<warnings.txt)
+      - name: Emit Actions warnings for python warnings
+        run: |
```
jacobtylerwalls, line 173:
```suggestion
```

jacobtylerwalls:
This will need a preparatory commit on main to make sure all astroid failures are re-raised as a flavor of `AstroidError`.

.github/workflows/primer_run_main.yaml
DanielNoord: Just because I find this very neat:
https://stackoverflow.com/a/962268

Change you just made is fine with me though: explicit over intricate or whatever is in the PEP.

jacobtylerwalls:
Here's a proof of concept showing a passing CI job on the primer and an Actions warning annotation with the traceback from astroid: [Run / 3.8](https://github.com/jacobtylerwalls/pylint/runs/6660281189?check_suite_focus=true#step:17:35668)

.github/workflows/primer_run_main.yaml
DanielNoord: Did you not merge these steps for a particular reason?

.github/workflows/primer_run_main.yaml
```diff
@@ -96,7 +96,13 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          python tests/primer/primer_tool.py run --type=main
+          python tests/primer/primer_tool.py run --type=main 2>warnings.txt
+          WARNINGS=$(head -c 65000 < warnings.txt)
+      - name: Emit Actions warnings for python warnings
+        run: |
```
jacobtylerwalls, line 102:
```suggestion
```

DanielNoord (approved):

PR #6753, opened 2022-05-30
Move more packages to the new primer

DanielNoord:

## Type of Changes

## Description

I'll remove the empty file when we can remove the old primer as removing now requires additional refactors.

### DIFF
--- a/tests/primer/packages_to_lint_batch_one.json
+++ b/tests/primer/packages_to_lint_batch_one.json
@@ -4,11 +4,6 @@
     "directories": ["django"],
     "url": "https://github.com/django/django.git"
   },
-  "home-assistant": {
-    "branch": "dev",
-    "directories": ["homeassistant"],
-    "url": "https://github.com/home-assistant/core.git"
-  },
   "keras": {
     "branch": "master",
     "directories": ["keras"],
--- a/tests/primer/packages_to_lint_batch_two.json
+++ b/tests/primer/packages_to_lint_batch_two.json
@@ -1,18 +1,1 @@
-{
-  "pandas": {
-    "branch": "main",
-    "directories": ["pandas"],
-    "pylint_additional_args": ["--ignore-patterns=\"test_"],
-    "url": "https://github.com/pandas-dev/pandas.git"
-  },
-  "psycopg": {
-    "branch": "master",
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas (approved):

DanielNoord:
@jacobtylerwalls This is running into CI time-outs. See `main`.

I don't really understand as this batch of packages passed on my local branch in CI and locally. Does this pass in reasonable time for you locally?

jacobtylerwalls:
Make sure to try with bleeding-edge astroid, it has good nutrients :D

DanielNoord:
> Make sure to try with bleeding-edge astroid, it has good nutrients :D

Does `bleeding-edge` fail with these packages? 😢

jacobtylerwalls:
Well, I'm waiting to find out. It takes a long time to lint `home-assistant`! How long does it take on your system?

DanielNoord:
I got 21m in CI here:
https://github.com/DanielNoord/pylint/runs/6657709454?check_suite_focus=true
Note that that is with:
https://github.com/DanielNoord/pylint/commit/e739427c52dce6f0b8e8a6275af23c660b42bc48

I removed `keras`, `Django` and `music21` as they all crashed due to the `importlib` issue.

jacobtylerwalls:
@DanielNoord the other thing I can think of is the message dict might be huge. Could we write the message dict and flush it as each package is linted, rather than writing it out after all packages? This will also help with debugging.

DanielNoord:
Now that I look at that workflow again I see that we never actually go beyond `keras`. So it was probably broken as well, but just failed prematurely. I didn't recognise that correctly... Sorry!

> @DanielNoord the other thing I can think of is the message dict might be huge. Could we write the message dict and flush it as each package is linted, rather than writing it out after all packages? This will also help with debugging.

Yeah that should be do-able.
The only thing we need to take care of is writing `{` before the first package `,` after each package and `}` after the last package to make the final output a valid JSON dictionary.

I was actually planning on shutting down for the night. If you want feel free to take a stab at this or just revert this PR so `main` gets unblocked. I think I won't get to this before tomorrow evening (+24h from now).

jacobtylerwalls:
Woo hoo! The issue with keras is fixed at PyCQA/astroid#1579.

PR #6755, opened 2022-05-30
Reraise failures in astroid as `AstroidError`

jacobtylerwalls:
## Type of Changes

## Description

Not 100% bullet-proof, but we're making an effort not to fail primer jobs if the failures stem from astroid and just warn about them instead. There could be little places all over the codebase that import little helpers, but these are the big ones.

Ref https://github.com/PyCQA/pylint/pull/6746#issuecomment-1141266169

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -771,7 +771,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
             message = f"Cannot import {modname!r} due to syntax error {str(exc.error)!r}"  # pylint: disable=no-member; false positive
             self.add_message("syntax-error", line=importnode.lineno, args=message)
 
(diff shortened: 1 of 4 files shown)


pylint/checkers/imports.py
```diff
@@ -771,7 +771,7 @@ def _get_imported_module(self, importnode, modname):
             message = f"Cannot import {modname!r} due to syntax error {str(exc.error)!r}"  # pylint: disable=no-member; false positive
             self.add_message("syntax-error", line=importnode.lineno, args=message)
 
-        except astroid.AstroidBuildingException:
+        except astroid.AstroidBuildingError:
```
jacobtylerwalls, line 774: this was just an old alias

DanielNoord (approved):

PR #6756, opened 2022-05-30
Run primer on ``3.7`` and ``3.10``

DanielNoord:

## Type of Changes

## Description

I think it makes most sense to run the primer on the last and least supported versions. Running it on any other doesn't really add much I think.

Besides, running on `3.7` has the additional benefit of testing the old parser (<3.8).

### DIFF
--- a/.github/workflows/primer_comment.yaml
+++ b/.github/workflows/primer_comment.yaml
@@ -30,11 +30,11 @@ jobs:
       - run: npm install @octokit/rest
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.2
-      - name: Set up Python 3.8
+      - name: Set up Python 3.10
         id: python
         uses: actions/setup-python@v3.1.2
         with:
-          python-version: 3.8
+          python-version: 3.10
 
       # Restore cached Python environment
       - name: Generate partial Python venv restore key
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
The comment will just be for 3.10, though?

DanielNoord:
> The comment will just be for 3.10, though?

Yeah. I didn't think the rare case where there is a difference between `3.7` and `3.10` in messages emitted would be worth the effort to try and merge output of different versions.
But by running on multiple versions we can (eventually) deprecate/remove the old primer as we now have almost feature-parity.

jacobtylerwalls (approved):
Super. I'd still like to let the current job on `main` finish first to see if the keras warning goes away. Then we should remove the old primer.

DanielNoord:
> Super. I'd still like to let the current job on `main` finish first to see if the keras warning goes away. Then we should remove the old primer.

Well, the old primer has the benefit of splitting up in different batches. `home-assistant` has caught multiple fatal errors previously so it's a real shame that we can't add it here.
Unless we find a clean way to have multiple batches in the new primer I think it make sense to leave the `batch_one` workflow of `old primer` running.

jacobtylerwalls:
A `--packages` argument to the run command?

DanielNoord:
> A `--packages` argument to the run command?

Probably should do ``--packages-json=``? And then make `Primer.packages` be based on that or something.

Haven't looked at it yet. Might be tricky to download and compare artefacts from different jobs in the same workflow.

DanielNoord:
Note that merging this will likely break all PRs until we get a new successful run as we need a `3.10` output for the commenter to work (and not the `3.8`).

PR #6758, opened 2022-05-30
Warn users who want to be warned about failing comment workflows

DanielNoord:

## Type of Changes

## Description

Since Comment workflows fail silently expect for PR creators we can easily miss whenever there are issues in that pipeline.
This makes it so that I will be tagged in any failing workflows which is quite handy.

See [this](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary) and [this](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/)

### DIFF
--- a/.github/workflows/primer_comment.yaml
+++ b/.github/workflows/primer_comment.yaml
@@ -24,6 +24,10 @@ jobs:
     name: Run
     runs-on: ubuntu-latest
     steps:
+      - name: Skip if PR run failed
+        if: ${{ github.event.workflow_run.conclusion == 'success' }}
+        run: exit 1
+      - run: echo 'The triggering workflow passed'
       - uses: actions/setup-node@v3
         with:
           node-version: 16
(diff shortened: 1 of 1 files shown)


DanielNoord:
@jacobtylerwalls You seem to be interested in the primer as well. Feel free to add your name to the to-be tagged list.

Not sure if I'll keep this around. But it's nice to get notified of this early.

.github/workflows/primer_comment.yaml
```diff
@@ -24,6 +24,10 @@ jobs:
     name: Run
     runs-on: ubuntu-latest
     steps:
+      - name: Skip if PR run failed
+        if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
jacobtylerwalls, line 28: is this the right way to check for failure?

DanielNoord: For some reason you can't see you only want to trigger this workflow if the dependent workflow succeeds. Only completed is supported.

Thus, even if Primer / PR fails this workflow gets triggered. This is a sort of quick stop so that you won't try and investigate why Primer / Comment failed but rather why Primer / PR fails.

Wait..... If we start failing Primer / Run for fatal warnings then we won't comment with this line anymore. Perhaps we should remove the top lines.

The lines added at the bottom are still valid though.

jacobtylerwalls: I just would have thought you have to check for != success. Otherwise on every success, this exits with a failure code?

And yeah, I'll remove the huge projects like home-assistant and/or pandas later tonight if need be. No worries. Thanks for everything!

.github/workflows/primer_comment.yaml
```diff
@@ -24,6 +24,10 @@ jobs:
...
+        run: exit 1
+      - run: echo 'The triggering workflow passed'
```
DanielNoord, line 30:
```suggestion
```

PR #6759, opened 2022-05-30
Allow specifying numbers of jobs to use in primer

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -96,7 +96,7 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          python tests/primer/primer_tool.py run --type=main 2>warnings.txt
+          python tests/primer/primer_tool.py run --type=main --jobs=0 2>warnings.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/primer_run_main.yaml
```diff
@@ -96,7 +96,7 @@ jobs:
         run: |
           . venv/bin/activate
           pip install -e .
-          python tests/primer/primer_tool.py run --type=main 2>warnings.txt
+          python tests/primer/primer_tool.py run --type=main --jobs=0 2>warnings.txt
```
jacobtylerwalls, line 99: I think this is really dangerous and will make things a lot worse. We discussed it at https://github.com/PyCQA/pylint/pull/6723#discussion_r884134628.

jacobtylerwalls (changes requested):
If the current job on main fails again, let's just isolate if things pass without `homeassistant`.

.github/workflows/primer_run_main.yaml
DanielNoord: Yeah, I guess you're right.

Let's wait on your investigation before adding this.

PR #6733, opened 2022-05-29
Refactors related to primer

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Closes https://github.com/PyCQA/pylint/issues/6732.

Added the verbose mode cause I'd like to see which packages take longest to lint in the old primer.

PR #6768, opened 2022-05-31
Fix broken links in primer comments by stripping ".git"

jacobtylerwalls:
Second time I noticed broken URLs with ".git" in them, so better to fix at the source! Ref #6733

### DIFF
--- a/tests/primer/primer_tool.py
+++ b/tests/primer/primer_tool.py
@@ -237,7 +237,8 @@ class Primer:
                 filepath = str(message["path"]).replace(
                     str(package_data.clone_directory), ""
                 )
-                comment += f"{package_data.url}/blob/{package_data.branch}{filepath}#L{message['line']}\n"
+                without_suffix = package_data.url.strip(".git")
+                comment += f"{without_suffix}/blob/{package_data.branch}{filepath}#L{message['line']}\n"
                 count += 1
                 print(message)
             if missing_messages:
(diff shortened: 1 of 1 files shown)


tests/primer/primer_tool.py
```diff
@@ -237,7 +237,8 @@ def _create_comment(
                 filepath = str(message["path"]).replace(
                     str(package_data.clone_directory), ""
                 )
-                comment += f"{package_data.url}/blob/{package_data.branch}{filepath}#L{message['line']}\n"
+                without_suffix = package_data.url.strip(".git")
```
Pierre-Sassoulas, line 240: This won't do what you want it to do if the repo name end with g i or t or a combination of them. See `"https://github/a/myrepotigg.git".strip(".git") `

Pierre-Sassoulas (changes requested):

tests/primer/primer_tool.py
Pierre-Sassoulas: I'd use `rsplit(".git") [0]` (sorry can't suggest I'm on mobile.

jacobtylerwalls: yikes, thanks.

jacobtylerwalls, line 240:
```suggestion
                without_suffix = package_data.url.rsplit(".git")[0]
```

Pierre-Sassoulas (approved):

DanielNoord (approved):

PR #6769, opened 2022-05-31
Add typing to ``basic_checker``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import collections
 import itertools
 import sys
+from collections.abc import Iterator
 from typing import TYPE_CHECKING, cast
 
 import astroid
@@ -21,7 +22,7 @@ from pylint.reporters.ureports import nodes as reporter_nodes
 from pylint.utils import LinterStats
 
 if TYPE_CHECKING:
-    pass
+    from pylint.lint.pylinter import PyLinter
 
 if sys.version_info >= (3, 8):
     from typing import Literal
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
Is the primer broken because we're on bleeding edge astroid ? I feel like it should be said somewhere if that's the case. Also we probably want to provide the crash template file instead of the pylint message ?

DanielNoord:
> Is the primer broken because we're on bleeding edge astroid ? 

@jacobtylerwalls Another `importlib` issue.. Do you recognise it?

> I feel like it should be said somewhere if that's the case. Also we probably want to provide the crash template file instead of the pylint message ?

I don't think we can easily extract the crash template from the environment. Perhaps we can redirect `stderr` to a file and print it so you at least get the stacktrace.

jacobtylerwalls:
It was fixed in the Astroid PR we merged this morning. The job did not fail, just extraneous comments. I'm a little surprised that it appeared in the content of the message. I would have thought that 5 Astroid errors == 5 Astroid errors leads to no message.

jacobtylerwalls:
> Also we probably want to provide the crash template file instead of the pylint message ?
All of it in the log, including a GitHub Actions warning to point folks directly to it. We did a lot of work around this :D

I agree it could still surprise a contributor. I'm realizing it's not just `astroid` -- we've solved the failing jobs issue from using bleeding edge. It could be any of the open source projects in the primer that push a new commit, we then have a drift in messages emitted between main and PR, no? I'll post on the other thread.

Pierre-Sassoulas:
> It could be any of the open source projects in the primer that push a new commit, we then have a drift in messages emitted between main and PR, no? I'll post on the other thread.

I think we discussed this with @DanielNoord when we added the original primer, there's a system in place for caching based on a concatenation of all repo hashes.

Pierre-Sassoulas:
>  I found the issue. It's a funny one (for us developers that is...)

That is some back-of-the-labtop-worthy quote. (Also, great analysis 👍 )

jacobtylerwalls:
> Because I'm dumb and ignore any changes to doc/data/messages/** in

I was the one that asked for this, sorry! 🧐

PR #6772, opened 2022-05-31
Run ``Primer / Main`` on all pushes to main

DanielNoord:

## Type of Changes

## Description

See https://github.com/PyCQA/pylint/pull/6769#issuecomment-1142052024

### DIFF
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -9,8 +9,6 @@ on:
   push:
     branches:
       - main
-    paths-ignore:
-      - doc/data/messages/**
 
 env:
   CACHE_VERSION: 1


DanielNoord:
Note that we are correctly skipping `Primer / Run` on this PR as we are only changing something in `Primer / Main`. So, it doesn't make sense to run the workflow as we will never be able to gauge its impact.

Nice to see that is working as intended :)

DanielNoord:
> Just an unrelated thought. Obviousely we don't want a changelog entry for this and it has the bug label (rightly so). So regarding the changelog check maybe it could be "only the primer/documentation/maintenance label" or a "no-changelog" label with any combination of other label.

Let's create a `no-news` label like other projects do!

PR #6773, opened 2022-05-31
Filter out astroid errors from the bot comment

jacobtylerwalls:
## Type of Changes

## Description
See potential for confusion like at https://github.com/PyCQA/pylint/pull/6769#issuecomment-1141953893. Just point people to the GitHub Actions warning if they wan't to look into it, but mostly just advising chilling out.
Refs #5364

### DIFF
--- a/tests/primer/primer_tool.py
+++ b/tests/primer/primer_tool.py
@@ -184,10 +184,8 @@ class Primer:
 
             # Create comment for new messages
             count = 1
-            fatal_count = 1
-            astroid_errors = False
-            new_non_fatal_messages = ""
-            new_fatal_messages = ""
+            astroid_errors = 0
+            new_non_astroid_messages = ""
             if new_messages:
                 print("Now emitted:")
             for message in new_messages:
@@ -197,17 +195,9 @@ class Primer:
                 # Existing astroid errors may still show up as "new" because the timestamp
                 # in the message is slightly different.
                 if message["symbol"] == "astroid-error":
-                    astroid_errors = True
-                elif message["type"] == "fatal":
-                    new_fatal_messages += (
-                        f"{fatal_count}) {message['symbol']}:\n*{message['message']}*\n"
-                        "**Please check your changes on the following file**:\n"
-                        f"{package_data.url}/blob/{package_data.branch}{filepath}#L{message['line']}\n"
-                    )
-                    print(message)
-                    fatal_count += 1
+                    astroid_errors += 1
                 else:
-                    new_non_fatal_messages += (
+                    new_non_astroid_messages += (
                         f"{count}) {message['symbol']}:\n*{message['message']}*\n"
                         f"{package_data.url}/blob/{package_data.branch}{filepath}#L{message['line']}\n"
                     )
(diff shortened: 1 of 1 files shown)


tests/primer/primer_tool.py
```diff
@@ -185,15 +185,18 @@ def _create_comment(
...
-                if message["type"] == "fatal":
-                    filepath = str(message["path"]).replace(
-                        str(package_data.clone_directory), ""
-                    )
+                filepath = str(message["path"]).replace(
+                    str(package_data.clone_directory), ""
+                )
+                if message["symbol"] == "astroid-error":
+                    new_astroid_errors = True
+                elif message["type"] == "fatal":
```
DanielNoord, line 199: Shouldn't this be `if`? I would still like these to be reported in the comment, just for us maintainers to notice them.

jacobtylerwalls: Either way. I thought the idea was to hide as much of this as possible from contributors. We could tag people in the comment. I'm curious what @Pierre-Sassoulas thinks.

jacobtylerwalls: There's no useful information in the message anyway. All of the useful information is in the GitHub Actions warning link to the crash report.

tests/primer/primer_tool.py
```diff
@@ -202,16 +205,20 @@ def _create_comment(
...
                         f"{package_data.url}/blob/{package_data.branch}{filepath}#L{message['line']}\n"
                     )
                     print(message)
                     count += 1
 
+            if new_astroid_errors:
+                comment += (
+                    "New error(s) were found stemming from the `astroid` library. "
+                    "This is unlikely to have been caused by your changes. "
+                    "(Pushing to this pull request after another commit has been pushed to main may resolve the issue.) "
```
Pierre-Sassoulas, line 219:
```suggestion
                    "(Pushing to this pull request after another commit has been pushed to main may resolve the issue.) "
```
I'm not sure I understand this. Do you mean rebasing on main after a fix is found and applied by a maintainer ?

Pierre-Sassoulas: Ho right, did not think of that.

tests/primer/primer_tool.py
jacobtylerwalls: I'm not sure what you're referring to: other fatal errors like `import-error`? We fail the job completely and never get to the comment step.

Pierre-Sassoulas: I misunderstood the code 😄 Anyway I think it would be nice to have a little message but to make sure it's clear for contributor that it's a message for maintainer that are going to review, not them.

jacobtylerwalls: Sure. I'm not sure what else to say, though, other than reporting the issue against astroid. I don't like sending the message that people aren't welcome to help if they want.

jacobtylerwalls: @DanielNoord Am I correct that after #6746 we won't reach the "fatal errors" block anyway? A few sentences pointing people to the actual info in the logs should be sufficient?

jacobtylerwalls: I'm talking about non-astroid fatal-errors, like `import-error`.

jacobtylerwalls (commented):
Remove fatal block

PR #6780, opened 2022-05-31
[ADD] missing-timeout: Used when a method calls an external request

moylop260:
## Description

Calling external request needs to use timeout in order to avoid waiting for a long time

You can even reproduce the case using deelay.me
e.g.

```python
import requests
response = requests.get("https://deelay.me/5000/http://localhost:80")  # It will spend 5s

response = requests.get("https://deelay.me/5000/http://localhost:80", timeout=2)  # timeout

response = requests.get("https://deelay.me/1000/http://localhost:80", timeout=2)  # fine

```

After 2s if the request doesn't have response it raises the following exception
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='deelay.me', port=443): Read timed out. (read timeout=2)

But if it responses <=1s it is fine

Now you can test the same but using a bigger delay

You can define your custom methods using the configuration parameter: --timeout-methods

The default methods are:
    - requests.delete
    - requests.get
    - requests.head
    - requests.options
    - requests.patch
    - requests.post
    - requests.put
    - requests.request
    - serial.Serial

You even could add the following extra methods manually if there is not a default defined:
    - http.client.HTTPConnection
    - http.client.HTTPSConnection
    - smtplib.SMTP

NOTE: The maintainers don't like the idea to add in the doc the following description, so adding it here to remember it myself

```rst
Notice the libraries needs to be installed and the name should be the output of pylint infered. "
            "from astroid import extract_node;"
            "node = extract_node('from requests import get;get()');"
            "print(next(node.func.infer()).qname())",
```

## Type of Changes


Pierre-Sassoulas:
Thank you for working on this. It seems the primer links are still broken (this is still very new 😄), it works if you remove the ``.git`` after pandas or sentry though.

CONTRIBUTORS.txt
```diff
@@ -88,6 +88,7 @@ contributors:
   * Add consider-merging-isinstance, superfluous-else-return
   * Fix consider-using-ternary for 'True and True and True or True' case
   * Add bad-docstring-quotes and docstring-first-line-empty
+  * Add missing-timeout
```
Pierre-Sassoulas, line 91: Welcome back :)

moylop260:
> Thank you for working on this. It seems the primer links are still broken (this is still very new 😄), it works if you remove the `.git` after pandas or sentry though.

Thank you !

What should I do with the result of pandas and sentry?

DanielNoord:
Haven't looked at the code, but I wondered what `requests` themselves say about this parameter.
See [their documentation](https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts).


_You can tell Requests to stop waiting for a response after a given number of seconds with the timeout parameter. Nearly all production code should use this parameter in nearly all requests. Failure to do so can cause your program to hang indefinitely:_

I think it makes sense to have this a default checker with some sane defaults (I haven't looked at the current lists).
It seems like it is similar to the `unspecified-encoding` warning I originally added: yeah, your code works without the argument, but it's much safer/robust if you include it.

Pierre-Sassoulas:
> I think it makes sense to have this a default checker with some sane defaults

For ``requests`` it make sense to make it the default but we have a lot of other function that can take a ``timeout``. For some, the default value might be something sensible like 5s and then it's not as clear cut.

DanielNoord:
> > I think it makes sense to have this a default checker with some sane defaults
> 
> For `requests` it make sense to make it the default but we have a lot of other function that can take a `timeout`. For some, the default value might be something sensible like 5s and then it's not as clear cut.

Yeah, so it's more about finding sane defaults than having this as an extension, right?

Pierre-Sassoulas:
> Yeah, so it's more about finding sane defaults than having this as an extension, right?

How did you populate [the list](https://github.com/PyCQA/pylint/pull/6780/files#diff-1c142afc02e8b8c4cf872f61ff8693a0b25646d6db50f16f8027dc130e7f719fR1065) @moylop260 ? Are all values there reasonable ?

moylop260:
> > Yeah, so it's more about finding sane defaults than having this as an extension, right?
> 
> How did you populate [the list](https://github.com/PyCQA/pylint/pull/6780/files#diff-1c142afc02e8b8c4cf872f61ff8693a0b25646d6db50f16f8027dc130e7f719fR1065) @moylop260 ? Are all values there reasonable ?

This lint was born only with `requests` then I have noticed that there are other methods with timeout parameter so I looked for methods using `timeout` parameter manually (e.g. `grep`)

Then confirming in the documentation:

# 3rd party libs

[serial.Serial](https://pyserial.readthedocs.io/en/latest/pyserial_api.html)
 - timeout = None: wait forever / until requested number of bytes are received

[requests](https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts)
 - You can tell Requests to stop waiting for a response after a given number of seconds with the timeout parameter. Nearly all production code should use this parameter in nearly all requests. Failure to do so can cause your program to hang indefinitely:

# built-in libs

[urllib.request](https://docs.python.org/3/library/urllib.request.html#module-urllib.request)
 - The optional timeout parameter specifies a timeout in seconds for blocking operations like the connection attempt (if not specified, the global default timeout setting will be used). This actually only works for HTTP, HTTPS and FTP connections.

[http.client](https://docs.python.org/3/library/http.client.html)
 - If the optional timeout parameter is given, blocking operations (like connection attempts) will timeout after that many seconds (if it is not given, the global default timeout setting is used)

[smtplib](https://docs.python.org//3/library/smtplib.html)
 - The optional timeout parameter specifies a timeout in seconds for blocking operations like the connection attempt (if not specified, the global default timeout setting will be used)

There is a default timeout setting but It is not easy to detect it if it is already defined in the code

```python
import socket
socket.getdefaulttimeout()  # by default is None
```

So, let me know if we should reduce the list for timeout-methods default in order to be added manually for each project or each project should remove manually these fields

Note: I missed ftp.lib too but I will wait for you reply in order to know if we should consider add more cases by default or less

moylop260:
IMHO I fixed all the pending feedback

Let me know what should we do here, please

pylint/checkers/variables.py
```diff
@@ -1062,6 +1070,30 @@ class VariablesChecker(BaseChecker):
...
+                    "requests.get",
+                    "requests.head",
+                    "requests.options",
+                    "requests.patch",
+                    "requests.post",
+                    "requests.put",
+                    "requests.request",
+                    "serial.Serial",
+                    "smtplib.SMTP",
+                    "urllib.request.urlopen",
```
DanielNoord, line 1089: Let's remove this. 
https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen
The documentation states that the default timeout will be used if it is not specified. So there is no risk of hanging I think. Or at least, we have no way to determine if the global timeout is indeed set so we shouldn't warn about it.

Edit: I only just saw that you already identified this issue. I agree, let's remove them.

pylint/checkers/variables.py
```diff
@@ -1062,6 +1070,30 @@ class VariablesChecker(BaseChecker):
                 "help": "List of names allowed to shadow builtins",
             },
         ),
+        (
+            "timeout-methods",
+            {
+                "default": (
+                    "http.client.HTTPConnection",
+                    "http.client.HTTPSConnection",
```
DanielNoord, line 1078: Let's remove this as well. See my comment about `urllib`.

pylint/checkers/variables.py
```diff
@@ -1062,6 +1070,30 @@ class VariablesChecker(BaseChecker):
...
+                    "requests.delete",
+                    "requests.get",
+                    "requests.head",
+                    "requests.options",
+                    "requests.patch",
+                    "requests.post",
+                    "requests.put",
+                    "requests.request",
+                    "serial.Serial",
+                    "smtplib.SMTP",
```
DanielNoord, line 1088: Let's remove this as well. See comment about `urllib`.

pylint/checkers/variables.py
```diff
@@ -1091,6 +1123,7 @@ def visit_module(self, node: nodes.Module) -> None:
         """Visit module : update consumption analysis variable
         checks globals doesn't overrides builtins.
         """
+        self._imports: dict = {}
```
DanielNoord, line 1126: Could you add further typing to this dict. Like `dict[..., ...]`.

pylint/checkers/variables.py
```diff
@@ -1174,6 +1209,45 @@ def leave_setcomp(self, _: nodes.SetComp) -> None:
         # do not check for not used locals here
         self._to_consume.pop()
 
+    @utils.only_required_for_messages(
+        "missing-timeout",
+    )
+    def visit_call(self, node: nodes.Call) -> None:
+        """Check if the call needs timeout parameter based on package.func_name
```
DanielNoord, line 1216:
```suggestion
        """Check if the call needs a timeout parameter based on package.func_name
```

pylint/checkers/variables.py
```diff
@@ -1174,6 +1209,45 @@ def leave_setcomp(self, _: nodes.SetComp) -> None:
...
+        configured in config.timeout_methods.
+
+        Package uses the self._imports dict variable
+        in order to know the original package name imported
+        """
+        lib_name = ""
+        if isinstance(node.func, nodes.Attribute) and isinstance(
```
DanielNoord, line 1223: I'll need to check this myself locally. But the way you determine the library name seems like it could just work with an inference of the `nodes.Name` and checking the `qname` of the import.

moylop260: Yes, it is valid for only built-in libraries or 3rd-party library already installed

But the CIs rarely install all dependencies to run the static checkers like pylint

Another matter is that using the infered qname the name of the method change a little bit and there is not an easy way to find these kind of names without using pylint/astroid

e.g. the infered qname for `from requests import delete` the result is `requests.api.delete`

Even if you are importing with `requests.delete` the name has an `*.api.*`

So, it could be hard to know that is the real name for the user trying to set values in the pylint.cfg for `timeout-methods` parameter

IMHO it is easier to know the name using the same way than it was imported

and even if it is not installed

DanielNoord: I think we should use the `qname`. It might be a little bit more difficult to set up, but it's arguably more correct. `pylint` depends on dependencies being installed for checkers like `no-member` anyway so not installing dependencies in the environment that `pylint` uses will likely create troubles anyway.

moylop260: I'm not sure about this

I mean, almost all of the current environments running static checkers are not installing extra dependencies
e.g. pre-commit, travis-CI, github actions

So, It could be virtually disabled for almost all of this environment already configured

What about doing work done without installing extra packages?

DanielNoord: > I mean, almost all of the current environments running static checkers are not installing extra dependencies e.g. pre-commit, travis-CI, github actions

All of those also allow specifying additional dependencies. For example, our own configuration for `mypy` had dependencies:
https://github.com/PyCQA/pylint/blob/0777a0ecb5c730acfa54186c56af508692bec821/.pre-commit-config.yaml#L95-L106
Similarly, in our own run of `pylint` we make sure to install a set of dependencies into the environment:
https://github.com/PyCQA/pylint/blob/0777a0ecb5c730acfa54186c56af508692bec821/.github/workflows/checks.yaml#L58-L59

I understand that creates some additional work for users, but I think that's an inherent limitation to static checkers.

> So, It could be virtually disabled for almost all of these environments already configured

I'd suggest to anybody to install dependencies in environments that use static checkers. Using `caches` this shouldn't even slow down your CI after the initial cache has been created.

> What about doing work done without installing extra packages?

If that was the general policy of `pylint` I'd be fine with that, but we have other checks that do depend on you installing the dependencies. I don't think we should mix that.

Pierre-Sassoulas: Most of pylint checks won't work without dependencies installed, it has some disadvantages but also some advantages that are a differentiating factor of pylint compared to other linters. This is a fundamental assumption of the software that can't be changed (unless we want to have the worst of both world).

moylop260: Yes, there are not another way to raise a `no-member` or 

So, it is not a static checker but dynamic one since that it is mutable based on the environment

I mean, it is not possible to define a static checker for `no-member`

Similar case for `import-error`

But it is not the case for `missing-timeout` since that already there is a way to define it as static checker without extra dependencies installed

Similar to `reimported`

You even don't need to `infer` to know if a library was `reimported`

Of course, you could use infer but there is a way to avoid it so use it

So, you will get false positives for reimported if you have:

```python
# file1.py
from os import os as os1

# file2.py
import os
from .file1 import os1  # re-imported
```

But IMHO it is better if you have "plug&play" static checkers

moylop260: I have created a separated PR with inferred:

https://github.com/PyCQA/pylint/pull/6885#issuecomment-1148055447

Notice that the detected cases were less than this PR

I mean,

PR using static checkers is detecting in:
- sentry
- pandas
- pytest

but new PR only in sentry

Pierre-Sassoulas: We're going to get issues opened for it if we don't handle this. We're talking about a fundamental assumption that define what pylint is supposed to do since 2003. The expectation from pylint users is that aliases and namespace aliases should be handled correctly, see https://github.com/PyCQA/pylint/pull/6787 for example. We're advertising that [in the very first paragraph of our doc introduction](https://pylint.pycqa.org/en/latest/). It's a differentiating factor of pylint and one of the reason pylint should be used alongside flake8. pylint needs dependencies installed anyway and already is relying on this assumption for other checks so there is no cost to pylint's user of doing it this way and pylint do not function without this assumption. 

There could be a "light mode" pylint with only static analysis but imo it's a lot of work to create a flake8 clone.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -15,6 +15,8 @@ Summary -- Release highlights
 New checkers
 ============
 
+Added new checker ``missing-timeout``
```
Pierre-Sassoulas, line 18:
```suggestion
Added new checker ``missing-timeout`` to warn of default timeout values that could cause 
a program to be hanging indefinitely.
```

Pierre-Sassoulas (changes requested):
I think this should be separated in a (default) checker of its own and not be in the variable checker.

pylint/checkers/variables.py
moylop260: Fixed in last commit, thank you for your feedback!

moylop260:
Feedback is done!

Thank you for your feedback!

Where should I install the dependencies `requests` and `pyserial` in order to get the tests running well for the inferred checks (I have added in `requirements_test_min.txt` but IDK if it is good)?

The doc pipeline is failing since that the file reference still doesn't exists in `main` branch:
 - `/home/runner/work/pylint/pylint/doc/user_guide/messages/warning/missing-timeout.rst:25: WARNING: broken link: https://github.com/PyCQA/pylint/blob/main/pylint/checkers/method_args.py (404 Client Error: Not Found for url: https://github.com/PyCQA/pylint/blob/main/pylint/checkers/method_args.py)`

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,94 @@
...
+    """BaseChecker for method_args.
+
+    Checks for
+    * missing-timeout
+    """
+
+    name = "method_args"
+    msgs = {
+        "W3101": (
+            "Missing timeout for method '%s' can cause your program to hang indefinitely",
```
DanielNoord, line 30:
```suggestion
            "Missing timeout argument for method '%s' can cause your program to hang indefinitely",
```

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,94 @@
...
+                ),
+                "type": "csv",
+                "metavar": "<comma separated list>",
+                "help": "List of library.method which require a timeout parameter "
+                "e.g. 'requests.api.get,requests.api.post' "
+                "Notice the libraries needs to be installed and the library.method should be "
+                " the output of pylint infered.\n"
+                "from astroid import extract_node;"
+                "node = extract_node('from requests import get;get()');"
+                "print(next(node.func.infer()).qname())",
```
DanielNoord, line 65:
```suggestion
```

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    @utils.only_required_for_messages(
+        "missing-timeout",
+    )
+    def visit_call(self, node: nodes.Call) -> None:
+        """Check if the call needs a timeout parameter based on package.func_name
+        configured in config.timeout_methods.
+
+        Package uses inferred node in order to know the package imported
```
DanielNoord, line 77:
```suggestion
        Package uses inferred node in order to know the package imported.
```

requirements_test_min.txt
```diff
@@ -5,3 +5,6 @@ typing-extensions~=4.2
 pytest~=7.1
 pytest-benchmark~=3.4
 pytest-timeout~=2.1
+# missing-timeout inferred check
+requests
+pyserial
```
DanielNoord, line 10:
```suggestion
requests
```

tests/functional/m/missing/missing_timeout.py
```diff
@@ -0,0 +1,92 @@
+# pylint: disable=consider-using-with,import-error,missing-module-docstring,no-member,no-name-in-module,reimported
+
+import requests
+import serial
```
DanielNoord, line 4: I would be okay with only checks for ``requests``. @Pierre-Sassoulas what do you think?

tests/functional/m/missing/missing_timeout.py
```diff
@@ -0,0 +1,92 @@
+# pylint: disable=consider-using-with,import-error,missing-module-docstring,no-member,no-name-in-module,reimported
```
DanielNoord, line 1:
```suggestion
"""Tests for missing-timeout."""

# pylint: disable=consider-using-with,import-error,no-member,no-name-in-module,reimported
```

tests/functional/m/missing/missing_timeout.py
```diff
@@ -0,0 +1,92 @@
...
+delete_r("http://localhost", timeout=10)
+get_r("http://localhost", timeout=10)
+head_r("http://localhost", timeout=10)
+options_r("http://localhost", timeout=10)
+patch_r("http://localhost", timeout=10)
+post_r("http://localhost", timeout=10)
+put_r("http://localhost", timeout=10)
+request_r("call", "http://localhost", timeout=10)
+
+delete("http://localhost", timeout=10)
```
DanielNoord, line 73: Could you also add a test where `timeout` is passed as a keyword variable by unpacking a dictionary?

DanielNoord (changes requested):
Thanks for working on this @moylop260.

I left some feedback. 

> Where should I install the dependencies `requests` and `pyserial` in order to get the tests running well for the inferred checks (I have added in `requirements_test_min.txt` but IDK if it is good)?

That's the right place, but I wonder if we should only do `requests`. See my other comment.

> The doc pipeline is failing since that the file reference still doesn't exists in `main` branch:

Could you open an issue for this. This should be fixed by ignoring those links in `linkcheck`. We already do so for other types of links.

Could you also add a documentation example like explained in https://github.com/PyCQA/pylint/issues/5953. I think https://github.com/PyCQA/pylint/pull/6874/ offers a nice example of the folder structure.

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,94 @@
...
+                    "requests.api.patch",
+                    "requests.api.post",
+                    "requests.api.put",
+                    "requests.api.request",
+                    "serial.serialcli.Serial",
+                ),
+                "type": "csv",
+                "metavar": "<comma separated list>",
+                "help": "List of library.method which require a timeout parameter "
+                "e.g. 'requests.api.get,requests.api.post' "
```
moylop260, line 60:
```suggestion
                "e.g. 'requests.api.get,requests.api.post' ",
```

tests/functional/m/missing/missing_timeout.py
moylop260: Good catch!

I just bypassed these kinds of cases since I didn't find a way to get the value of the dictionary in order to check if `timeout` was defined in the dict

moylop260:
@DanielNoord 

Thanks for your feedback

It is done!

requirements_test_min.txt
```diff
@@ -5,3 +5,4 @@ typing-extensions~=4.2
 pytest~=7.1
 pytest-benchmark~=3.4
 pytest-timeout~=2.1
+requests
```
Pierre-Sassoulas, line 8: I don't think we need request installed during our tests ?

```suggestion
```

Pierre-Sassoulas (changes requested):
This is looking pretty good already, thank you @moylop260 !

requirements_test_min.txt
moylop260: We are using a `infered` way to get the name of the method

If the library is not installed the unittest fails since that the check is not detected

My first approach was without infered (not library needed):
 - https://github.com/PyCQA/pylint/pull/6885

Check the following discussion about this:
 - https://github.com/PyCQA/pylint/pull/6780#discussion_r889506944

and last review where he confirms it is needed:
 - https://github.com/PyCQA/pylint/pull/6780#pullrequestreview-1001058586

doc/data/messages/m/missing-timeout/details.rst
```diff
@@ -0,0 +1,64 @@
...
+And the library needs to be installed in order to be detected
+
+You can add the following possible cases ``timeout-methods``:
+
+* ``http.client.HTTPConnection``
+* ``http.client.HTTPSConnection``
+* ``serial.serialcli.Serial``
+* ``smtplib.SMTP``
+* ``suds.client.Client``
+* ``urllib.request.urlopen``
```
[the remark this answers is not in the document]
moylop260: Check the Daniel review where he requested to add it here, please

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,88 @@
...
+            for keyword in node.keywords:
+                if isinstance(keyword, nodes.Keyword) or keyword.arg == "timeout":
+                    break
+            else:
+                self.add_message(
+                    "missing-timeout",
+                    node=node,
+                    args=(node.func.as_string(),),
+                    confidence=INFERENCE,
+                )
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Say we want to catch call of ``whatever.bar`` and ``sentry.something.foo`` that do not define a ``foobar`` argument, and then raise a ``missing-foobar`` message, we'd have to refactor the checker for it to work. The checker's current name imply this could be a goal. We could also rename the checker to ``MissingTimeoutChecker`` and not handle the generic case, I wouldn't mind, but I think it was nice to identify that it could be generic.

requirements_test_min.txt
Pierre-Sassoulas: Hmm strange, it worked for me locally without installing request. Would you mind applying the suggestion and see how the continuous integration react ?

tests/functional/m/missing/missing_timeout.py
Pierre-Sassoulas: Do you mean in the example of in the checker itself ? (Please do not resolve discussion where the discussion did not happen please @moylop260 I missed that one, as I have hundred of mails per day I rely on the web interface :) )

requirements_test_min.txt
moylop260: Try uninstalling it explicilty
`pip uninstall requests`

Even, try to use the code added in rst doc file

```python
from astroid import extract_node
node = extract_node('from requests import get;get()')
print(next(node.func.infer()).qname())
```

In fact, it is the way like infered works:
 - Screen Shot 2022-06-10 at 15 28 37

> Would you mind applying the suggestion and see how the continuous integration react ?

I did, the previous commit has that proof and the CI failed in the unittest of this check

pylint/checkers/method_args.py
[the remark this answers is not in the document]
moylop260: It is expected

A new checker needs to change a lot of things

One of them is create the new checker number, doc, unittest and change the logic of the method to support it

It can not be configured dynamically since that is not possible to add new checkers dynamically as far I know

doc/data/messages/m/missing-timeout/details.rst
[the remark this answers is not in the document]
moylop260: No, they don't see the previous discussion about:
 - https://github.com/PyCQA/pylint/pull/6780#discussion_r890401237
   - Screen Shot 2022-06-10 at 15 35 58

tests/functional/m/missing/missing_timeout.py
moylop260: I have already removed since that I agree with Daniel about it was changed in the code too

Since it could be not the same for external request than internal port

Even, we can add it in the configuration than the documentation says

So, IMHO nothing to discuss here that is reason it is resolved

doc/data/messages/m/missing-timeout/details.rst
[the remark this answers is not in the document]
moylop260: @DanielNoord 

Something to add here?

pylint/checkers/method_args.py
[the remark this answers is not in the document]
Pierre-Sassoulas: We could add values to check the argument x is defined in functions, f1, f2, f3, by modifying a data structure instead of adding new code. Say something like this:

```diff
        [
            {
                "functions": (
                    "requests.api.delete",
                    "requests.api.get",
                    "requests.api.head",
                    "requests.api.options",
                    "requests.api.patch",
                    "requests.api.post",
                    "requests.api.put",
                    "requests.api.request",
                ),
                "required_args": "timeout",
            },
+           {
+               "functions": (
+                   "whatever.bar",
+                   "sentry.something.foo",
+               ),
+               "required_args": "foobar",
+           }
        ]
```

But let's just rename the checker.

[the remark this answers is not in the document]
moylop260: I got it

Thank you for clarify

I even don't got you

How will you match that `timeout` is for requests and foobar is for `whatever.bar`?

Could you share me a checker using this way, please?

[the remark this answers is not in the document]
moylop260: Another question

> We could also rename the checker to MissingTimeoutChecker 

Do you mean, Using only one checker by file?

Is this a good idea?

I mean, if you have a guideline to define a checker by file you could have too many files in the future instead of re-use the same checker to add new ones in the future

Similar to imports.py, variables.py, strings.py... and so on

Using your guideline it should be:
 - bad_format_character.py

because it was the first check

Then add `truncated_format_string.py`

and so on

IMHO It is not a good idea if you can re-use a file to add many checkers

Even using this way the limit of 99 checkers prefix could be reached faster

Is not?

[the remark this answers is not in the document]
Pierre-Sassoulas: > How will you match that timeout is for requests and foobar is for whatever.bar?

They are in a different structure, we can iterate over the list of structure and get ``functions`` and  ``required_args``for each.  Please note that it's only an illustration the option and message name and description are not included there's some design to do before it works.

> Could you share me a checker using this way, please?

This would be a novel approach I guess. But the deprecated checker is pretty modular for example, you can define deprecated functions, and programmatically check that a particular function is deprecated by inheriting the checker (to check for example the version of django and have different result based on this).

>  It is not a good idea if you can re-use a file to add many checkers

There's a balance to have between mixing many messages in the same file and having strange bug when they interact strangely with each other, or having a lot of independent files. I'm not too afraid of having too many files and having to add a zero in the message id, but grouping is also nice if it's done in a way that is logical and easy to follow.

It could go both way, if you want to do a generic checker we can do that, but if you just want to only do a missing timeout checker it works too.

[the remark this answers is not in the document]
moylop260: Maybe using

library.method:parameter?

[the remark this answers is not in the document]
Pierre-Sassoulas: Hmm yeah that would work, but I don't think we want all the function/parameter to be mixed and the message to be a very generic "missing-argument". We want to explain why it's better to have timeout defined (hanging indefinitely otherwise) and we don't know yet why ``foobar`` in ``libary.method:parameter`` need to be defined.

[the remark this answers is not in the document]
DanielNoord: Sorry I was away yesterday so I haven't completely followed the discussion. I don't think this option should be generic: that will become unnecessarily confusing for users.

I do think that the checker class can be generic. If we find new arguments we want to check for we can just add a new option.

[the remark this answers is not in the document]
moylop260: Totally agree

So, Let me know if it requires a change, please

doc/data/messages/m/missing-timeout/details.rst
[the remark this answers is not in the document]
DanielNoord: Yeah, I think @Pierre-Sassoulas Is right this might be a bit too much for a simple `details.rst` file. I'm going to commit his changes.

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,88 @@
...
+            "Missing timeout argument for method '%s' can cause your program to hang indefinitely",
+            "missing-timeout",
+            "Used when a method needs a 'timeout' parameter "
+            "in order to avoid waiting for a long time. If no timeout "
+            "is specified explicitly the default value is used, for example "
+            "for 'requests' the program will never time out (i.e. hang indefinitely). "
+            "Notice the libraries needs to be installed and the name should be the output of pylint infered. "
+            "from astroid import extract_node;"
+            "node = extract_node('from requests import get;get()');"
+            "print(next(node.func.infer()).qname())",
```
DanielNoord, line 39:
```suggestion
```

requirements_test_min.txt
DanielNoord: I think adding `requests` to our test dependencies is not unreasonable. Many environments will already have it since the package is used so often in the Python ecosystem.

I'll leave final decision to @Pierre-Sassoulas but I would be okay with adding it here.

We could also add it as an additional test dependencies and only run the test when it is installed.

tests/functional/m/missing/missing_timeout.py
```diff
@@ -0,0 +1,85 @@
...
+get("http://localhost")  # [missing-timeout]
+head("http://localhost")  # [missing-timeout]
+options("http://localhost")  # [missing-timeout]
+patch("http://localhost")  # [missing-timeout]
+post("http://localhost")  # [missing-timeout]
+put("http://localhost")  # [missing-timeout]
+request("call", "http://localhost")  # [missing-timeout]
+
+kwargs_wo_timeout = {}
+post("http://localhost", **kwargs_wo_timeout)
```
DanielNoord, line 54: This should ideally raise, but let's leave as is for now. This PR adds enough as is 😄

DanielNoord (approved):

DanielNoord:
@moylop260 Could you give maintainers write acces to this branch? I'd like to merge some of the final changes but I don't have permission.

moylop260:
> @moylop260 Could you give maintainers write acces to this branch? I'd like to merge some of the final changes but I don't have permission.

I know that option but I even didn't find it in this PR:
 - Screen Shot 2022-06-21 at 11 02 54

I have added your gh user in our fork repository explicitly

Let me know if you need something else

and thanks for take care about this

DanielNoord:
@moylop260 Thanks! That worked. I think organisations can turn off that function globally, perhaps something to consider as it is quite useful for projects maintainers.

Pushed my changes! 😄

doc/data/messages/m/missing-timeout/details.rst
```diff
@@ -0,0 +1,10 @@
+You can add new methods that should have a defined ```timeout`` argument as qualified names
+in the ``timeout-methods`` options, for example:
```
DanielNoord, line 2:
```suggestion
in the ``timeout-methods`` option, for example:
```

tests/functional/m/missing/missing_timeout.py
moylop260: I totally agree

To be honest I didn't know how to inspect a dictionary from astroid/pylint

If you can share me a example I can do it in a new PR

Pierre-Sassoulas: I think some inspiration could be taken from the type checker (``missing-kwoa`` i.e. missing mandatory keyword argument in particular ) : https://github.com/PyCQA/pylint/blob/main/pylint/checkers/typecheck.py#L1379

pylint/checkers/method_args.py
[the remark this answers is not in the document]
Pierre-Sassoulas: > I do think that the checker class can be generic. 

Let's do this in a follow-up.

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,84 @@
...
+
+    name = "method_args"
+    msgs = {
+        "W3101": (
+            "Missing timeout argument for method '%s' can cause your program to hang indefinitely",
+            "missing-timeout",
+            "Used when a method needs a 'timeout' parameter "
+            "in order to avoid waiting for a long time. If no timeout "
+            "is specified explicitly the default value is used, for example "
+            "for 'requests' the program will never time out (i.e. hang indefinitely). ",
```
Pierre-Sassoulas, line 35:
```suggestion
            "Used when a method needs a 'timeout' parameter in order to avoid waiting "
            "for a long time. If no timeout is specified explicitly the default value "
            "is used. For example for 'requests' the program will never time out "
            "(i.e. hang indefinitely). ",
```

pylint/checkers/method_args.py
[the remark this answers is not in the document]
DanielNoord: My bad, I meant `checker` function. Let's not start adding abstract generic checker classes 😅

pylint/checkers/method_args.py
```diff
@@ -0,0 +1,84 @@
...
+                "type": "csv",
+                "metavar": "<comma separated list>",
+                "help": "List of qualified names (i.e., library.method) which require a timeout parameter "
+                "e.g. 'requests.api.get,requests.api.post'",
+            },
+        ),
+    )
+
+    @utils.only_required_for_messages(
+        "missing-timeout",
```
Pierre-Sassoulas, line 61:
```suggestion
        "missing-timeout"
```

tests/functional/m/missing/missing_timeout.py
moylop260: Thank you @Pierre-Sassoulas 

It worked like a charm

Also, It was committed

pylint/checkers/method_args.py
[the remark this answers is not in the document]
moylop260: Sorry, I even don't get what should I change here

Could you help me to understand the changes required, please?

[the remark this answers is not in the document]
Pierre-Sassoulas: The idea is to have a generic function so we could add data to a structure to check similar cases by modifying the data structure and not adding code (https://github.com/PyCQA/pylint/pull/6780#discussion_r894875851). But let's design it and refactor in another PR, this one is already doing a lot.

Pierre-Sassoulas (approved):
This is a great new checker, thank you for taking all the remark into account @moylop260 ! I think we can merge.

Pierre-Sassoulas:
The problem in readthedoc is not that easy to spot it's: ``/home/docs/checkouts/readthedocs.org/user_builds/pylint/checkouts/6780/doc/user_guide/messages/warning/missing-timeout.rst:12: WARNING: Inline emphasis start-string without end-string.``

PR #6789, opened 2022-06-01
Store namespaces respective to directories

DanielNoord:

## Type of Changes

## Description

This is a first step towards `per-directory-config` as I think they make the most sense:
1) Create a basic `Namespace`
2) Iterate over all `files_or_modules` and see if we can create additional `Namespaces` for subdirectories
3) Look up the correct `Namespace` before checking a file

This is a basic implementation of 3 which allows storing the information that should be gathered in step 2.

### DIFF
--- a/pylint/lint/utils.py
+++ b/pylint/lint/utils.py
@@ -99,3 +99,16 @@ def fix_import_path(args: Sequence[str]) -> Iterator[None]:
         yield
     finally:
         sys.path[:] = original
+
+
+def _is_relative_to(self: Path, *other: Path) -> bool:
+    """Checks if self is relative to other.
+
+    Backport of pathlib.Path.is_relative_to for Python <3.9
+    TODO: py39: Remove this backport and use stdlib function.
+    """
+    try:
+        self.relative_to(*other)
+        return True
+    except ValueError:
+        return False
--- a/tests/config/test_per_directory_config.py
+++ b/tests/config/test_per_directory_config.py
@@ -8,17 +8,17 @@ from pathlib import Path
 from pylint.lint import Run
 
 
-def test_fall_back_on_base_config() -> None:
+def test_fall_back_on_base_config(tmpdir: LocalPath) -> None:
     """Test that we correctly fall back on the base config."""
     # A file under the current dir should fall back to the highest level
     # For pylint this is ./pylintrc
+     test_file = tmpdir /  "test.py"
     runner = Run([__name__], exit=False)
     assert id(runner.linter.config) == id(runner.linter._base_config)
 
     # When the file is a directory that does not have any of its parents in
     # linter._directory_namespaces it should default to the base config
-    with tempfile.TemporaryDirectory() as tmpdir:
-        with open(Path(tmpdir) / "test.py", "w", encoding="utf-8") as f:
-            f.write("1")
-        Run([str(Path(tmpdir) / "test.py")], exit=False)
-        assert id(runner.linter.config) == id(runner.linter._base_config)
+    with open(test_file, "w", encoding="utf-8") as f:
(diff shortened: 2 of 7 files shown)


pylint/lint/utils.py
```diff
@@ -99,3 +99,16 @@ def fix_import_path(args: Sequence[str]) -> Iterator[None]:
         yield
     finally:
         sys.path[:] = original
+
+
+def _is_relative_to(self: Path, *other: Path) -> bool:
```
Pierre-Sassoulas, line 104:
```suggestion
def pathlib_relative_to_is_callable(self: Path, *other: Path) -> bool:
```
? I don't think I understand what this function does. Shouldn't the value be returned too ?

tests/config/test_per_directory_config.py
```diff
@@ -0,0 +1,24 @@
...
+    runner = Run([__name__], exit=False)
+    assert id(runner.linter.config) == id(runner.linter._base_config)
+
+    # When the file is a directory that does not have any of its parents in
+    # linter._directory_namespaces it should default to the base config
+    with tempfile.TemporaryDirectory() as tmpdir:
+        with open(Path(tmpdir) / "test.py", "w", encoding="utf-8") as f:
+            f.write("1")
+        Run([str(Path(tmpdir) / "test.py")], exit=False)
+        assert id(runner.linter.config) == id(runner.linter._base_config)
```
Pierre-Sassoulas, line 24:
```suggestion
    with open(test_file, "w", encoding="utf-8") as f:
        f.write("1")
    Run([str(test_file)], exit=False)
    assert id(runner.linter.config) == id(runner.linter._base_config)
```

tests/config/test_per_directory_config.py
```diff
@@ -0,0 +1,24 @@
...
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+import tempfile
+from pathlib import Path
+
+from pylint.lint import Run
+
+
+def test_fall_back_on_base_config() -> None:
```
Pierre-Sassoulas, line 11:
```suggestion
def test_fall_back_on_base_config(tmpdir: LocalPath) -> None:
```

There's a pytest fixture for that.

tests/config/test_per_directory_config.py
```diff
@@ -0,0 +1,24 @@
...
+    """Test that we correctly fall back on the base config."""
+    # A file under the current dir should fall back to the highest level
+    # For pylint this is ./pylintrc
```
Pierre-Sassoulas, line 14:
```suggestion
    # For pylint this is ./pylintrc
     test_file = tmpdir /  "test.py"
```

Pierre-Sassoulas (changes requested):

pylint/lint/utils.py
DanielNoord: It's just a line for line copy of the `stdlib` function. We don't really need to return anything else?

tests/config/test_per_directory_config.py
```diff
@@ -0,0 +1,24 @@
...
+from pathlib import Path
+
+from pylint.lint import Run
+
+
+def test_fall_back_on_base_config(tmpdir: LocalPath) -> None:
+    """Test that we correctly fall back on the base config."""
+    # A file under the current dir should fall back to the highest level
+    # For pylint this is ./pylintrc
+     test_file = tmpdir /  "test.py"
```
DanielNoord, line 15:
```suggestion
    test_file = tmpdir /  "test.py"
```

Pierre-Sassoulas (approved):

PR #6797, opened 2022-06-02
Rename clashing import name in one test case

kulikjak:
## Type of Changes

## Description

On Solaris (both Oracle and OpenIndiana), `import pkg` actually find something because pkg is our OS package manager. This results in pylint warnings in **regression_4680** test being very different. Would it be possible to rename it to something non clashing? It seems that foo package doesn't exist, but if it does, I am happy to change it to anything else ;).

### DIFF
--- a/tests/functional/r/regression/regression_4680.py
+++ b/tests/functional/r/regression/regression_4680.py
@@ -1,9 +1,9 @@
 # pylint: disable=missing-docstring,too-few-public-methods
 
-import pkg.sub  # [import-error]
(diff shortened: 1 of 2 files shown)


DanielNoord (approved):

Pierre-Sassoulas (approved):

kulikjak:
Thank you!

Pierre-Sassoulas:
Thank you for the contribution ! I think it's the first one for Solaris ! 😄

PR #6808, opened 2022-06-02
Prevent warnings on dict / list index lookup with destructuring assignment

### LINKED ISSUE #6788: unnecessary-list-index-lookup inconsistent when assigning to iterating object in iterable unpacking
### Bug description
```python
myList = [1, 2, 3]

for a, b in enumerate(myList):
    myList[a] = 2

for a, b in enumerate(myList):
    myList[a], _ = (2, 1)
```
### Configuration

_No response_

### Command used

```shell
pylint test2.py
```

### Pylint output

```shell
************* Module test2
test2.py:8:4: R1736: Unnecessary list index lookup, use 'b' instead (unnecessary-list-index-lookup)

Your code has been rated at 8.00/10 (previous run: 8.00/10, +0.00)
```

### Expected behavior

I would expect the output for both loops to be the same. The latter loop leads to an unnecessary-list-index-lookup though.

### Pylint version

```shell
pylint 2.14.0
astroid 2.11.5
Python 3.9.12 (main, Jun  1 2022, 11:38:51) 
[GCC 7.5.0]
```

### OS / Environment

Ubuntu 20.04

### Additional dependencies

_No response_

timmartin:
## Type of Changes

## Description

The checkers `unnecessary-list-index-lookup` and `unnecessary-dict-index-lookup` have a check that the usage of the index isn't appearing on the left hand side of an assignment. So for example:

    for key, val in my_dict:
        my_dict[key] = 'x'

should not be a warning. However, this checker was not catching the case where the assignment was part of a destructuring expression, so there was a false positive for:

    for key, val in my_dict:
        (my_dict[key], _) = ('x', 1)

This has been fixed.

Closes #6788


pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -160,6 +160,25 @@ def _will_be_released_automatically(node: nodes.Call) -> bool:
...
+    This requires recursive checking, because destructuring assignment can have
+    arbitrarily nested tuples and lists to unpack.
+    """
+    if isinstance(node.parent, nodes.Assign):
+        return node in node.parent.targets
+
+    if isinstance(node.parent, nodes.AugAssign):
+        return node == node.parent.target
+
+    if isinstance(node.parent, (nodes.Tuple, nodes.List)):
```
jacobtylerwalls, line 176: This will let you handle a set.
```suggestion
    if isinstance(node.parent, nodes.BaseContainer):
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -160,6 +160,25 @@ def _will_be_released_automatically(node: nodes.Call) -> bool:
...
+        return _is_part_of_assignment_target(node.parent)
+
+    return None
```
jacobtylerwalls, line 179:
```suggestion
    return False
```

doc/whatsnew/2/2.15/index.rst
```diff
@@ -27,6 +27,10 @@ Extensions
 False positives fixed
 =====================
 
+* Avoid reporting ``unnecessary-dict-index-lookup`` or ``unnecessary-list-index-lookup``
+  when the index lookup is part of a destructuring assignment.
+
+  Closes #6788
```
jacobtylerwalls, line 33: As a false positive fix it's eligible for a 2.14 backport. Do you mind moving this to 2.14.rst?

jacobtylerwalls (commented):
Thanks for the speedy fix! One tiny suggestion to handle sets.

pylint/checkers/refactoring/refactoring_checker.py
jacobtylerwalls: Never mind! You can't unpack into a set. I didn't look closely at my example. So this is purely up to you whether to change this.

doc/whatsnew/2/2.15/index.rst
timmartin: I've moved it, but I guess the code will also need to be rebased. I don't mind doing that, but I don't want to rewrite history in this and mess up the code review.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -27,7 +27,6 @@ Extensions
 False positives fixed
 =====================
 
-
 False negatives fixed
```
jacobtylerwalls, line 30:
```suggestion

False negatives fixed
```

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
Thank you @timmartin , it looks great, I also like the new named function that makes it easier to understand the conditions.

PR #6822, opened 2022-06-03
typecheck/Enum: fix crash in Enum attributes check

### LINKED ISSUE #6805: AttributeError: 'Call' object has no attribute 'name'
### Bug description

Happens with 2.14.

See stacktrace in the [build log](https://github.com/diofant/diofant/runs/6705072636?check_suite_focus=true).

### Configuration

```ini
See pyproject.toml of the project.
```

### Command used

```shell
pylint diofant
or (more minimal ;)) pylint -j1 diofant/stats/drv.py
```

### Pylint output

```shell
Crash
```

### Expected behavior

No crash, as before.

### Pylint version

```shell
pylint 2.14.0
astroid 2.11.5
Python 3.10.2+ (heads/3.10:75c5dbc27e, Feb 19 2022, 06:58:32) [GCC 10.2.1 20210110]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

huwcbjones:
## Type of Changes

## Description

Fixes #6805 by:
 - only checking for Enum attributes when metaclass is an Enum metaclass
 - coping with a name not being returned in `__new__`

### DIFF
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -451,10 +451,10 @@ def _emit_no_member(
             except astroid.MroError:
                 return False
             if metaclass:
-                if _enum_has_attribute(owner, node):
-                    return False
                 # Renamed in Python 3.10 to `EnumType`
-                return metaclass.qname() in {"enum.EnumMeta", "enum.EnumType"}
+                if metaclass.qname() in {"enum.EnumMeta", "enum.EnumType"}:
+                    return not _enum_has_attribute(owner, node)
+                return False
             return False
         if not has_known_bases(owner):
             return False
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -583,7 +583,7 @@ def _enum_has_attribute(
             (c.value for c in dunder_new.get_children() if isinstance(c, nodes.Return)),
             None,
         )
-        if returned_obj_name is not None:
+        if isinstance(returned_obj_name, nodes.Name):
             # Find all attribute assignments to the returned object
             enum_attributes |= _get_all_attribute_assignments(
                 dunder_new, returned_obj_name.name
--- a/doc/whatsnew/2/2.14/full.rst
+++ b/doc/whatsnew/2/2.14/full.rst
@@ -9,6 +9,10 @@ Release date: TBA
 
   Closes #6800
 
+* Fixed a crash when linting ``__new__()`` methods that return a call expression.
+
+  Closes #6805
+
 * Don't crash if we can't find the user's home directory.
 
   Closes #6802
--- a/tests/functional/e/enum_self_defined_member_6805.py
+++ b/tests/functional/e/enum_self_defined_member_6805.py
@@ -1,6 +1,7 @@
 """Tests for self-defined Enum members (https://github.com/PyCQA/pylint/issues/6805)"""
 # pylint: disable=missing-docstring
 # pylint: disable=too-few-public-methods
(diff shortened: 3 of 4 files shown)


pylint/checkers/typecheck.py
```diff
@@ -583,7 +583,7 @@ def _enum_has_attribute(
             (c.value for c in dunder_new.get_children() if isinstance(c, nodes.Return)),
             None,
         )
-        if returned_obj_name is not None:
+        if isinstance(returned_obj_name, nodes.Name):
```
DanielNoord, line 586: Wouldn't it be better if we actually infer the value here? Instead of handling only `nodes.Name`?

huwcbjones: Sorry Daniel, I'm trying to understand what you mean, but I'm struggling to figure out how we'd infer the value?

DanielNoord: With the function `safe_infer` (imported from `utils`) you can try to let `astroid` (an upgrade to the stdlib `ast` library) try to "infer" what the actual value of a node is.
So for ``x = 1`` `astroid` will be able to determine that `x` is a `nodes.Const` with `value = 1` instead of simply a `nodes.Name`. That way you get much more information about a particular node.

In fact, you can even infer a `nodes.Call` and try to determine what a call to a certain function actually returns, which seems to be what you're trying to do here. I should have caught this when I reviewed https://github.com/PyCQA/pylint/pull/5847/.

Looking at the code we should be able to infer what `dunder_new` returns and then do the subsequent checks for the message. You can look for `safe_infer` in the code base to see how other checkers user inference to know more about nodes.

huwcbjones: Ok, so I've given it a go and it doesn't seem very promising.
For all the test cases in `tests/functional/e/enum_self_defined_member_5138.py`, the result of `safe_infer` comes out as `Uninferable`.
For the new tests cases in `tests/functional/e/enum_self_defined_member_6805.py`, the result of `safe_infer` is `None`

jacobtylerwalls: Interesting suggestion, but it looks like there are changes in `astroid` needed so definitely not anything to look into in the context of a backportable pylint crash fix.

I'll open a followup issue in astroid, and then maybe we can revisit this later.

pylint/checkers/typecheck.py
```diff
@@ -451,10 +451,10 @@ def _emit_no_member(
             except astroid.MroError:
                 return False
             if metaclass:
-                if _enum_has_attribute(owner, node):
-                    return False
                 # Renamed in Python 3.10 to `EnumType`
-                return metaclass.qname() in {"enum.EnumMeta", "enum.EnumType"}
+                if metaclass.qname() in {"enum.EnumMeta", "enum.EnumType"}:
+                    return not _enum_has_attribute(owner, node)
```
jacobtylerwalls, line 456: Outside the scope of this PR, since status quo, but this looks fragile against importing `enum` under an alias, which @DanielNoord was just looking into at https://github.com/PyCQA/astroid/pull/1587. Probably worth a follow-up issue.

jacobtylerwalls (commented):
Thanks for the speedy work. Needs a short release note in 2.14/full.rst

pylint/checkers/typecheck.py
DanielNoord: Yeah, we should probably revisit this when we bump to `2.12`.

jacobtylerwalls (approved):

Issue #6794, opened 2022-06-02
pylint 2.14.0 seems to generate superfluous output


alex4200:
### Bug description

When using pylint 2.14.0 I get output messages like 

```
.pylintrc:1:0: E0015: Unrecognized option found: no-space-check (unrecognized-option)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message print-statement. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message parameter-unpacking. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message unpacking-in-except. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message old-raise-syntax. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message backtick. (bad-option-value)
...
```

With pylint 2.13.9 I do not get any of these error message (no other change)

### Command used

```shell
pylint <folder>
```

### Pylint output

```shell
************* Module /home/adietz/Local/1.25_create_doc/Repos/docs-create/.pylintrc
.pylintrc:1:0: E0015: Unrecognized option found: no-space-check (unrecognized-option)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message print-statement. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message parameter-unpacking. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message unpacking-in-except. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message old-raise-syntax. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message backtick. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message long-suffix. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message old-ne-operator. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message old-octal-literal. (bad-option-value)
.pylintrc:1:0: E0012: Bad option value for --disable. Don't recognize message import-star-module-level. (bad-option-value)
[...]

PR #6824, opened 2022-06-03
Distinguish between genuine typoes and old configuration that could be cleaned in configuration warning

Pierre-Sassoulas:
## Type of Changes

## Description

Deleted messages that do not exists anymore in pylint now raise ``useless-option-value`` instead of ``bad-option-value``
which permit to distinguish between genuine typoes and old configuration that could be cleaned.

Refs #6794 ... and 2 or 3 other issues opened recently. There's also question on stackoverflow about this.


pylint/constants.py
```diff
@@ -91,107 +91,114 @@ class DeletedMessage(NamedTuple):
...
+                ("C0322", "no-space-before-operator"),
+            ],
+        ),
+    ],
+    "https://github.com/PyCQA/pylint/pull/3571": [
+        DeletedMessage("C0330", "bad-continuation")
+    ],
+    "https://pylint.pycqa.org/en/latest/whatsnew/1/1.4.html#what-s-new-in-pylint-1-4-3": [
+        DeletedMessage("R0921", "abstract-class-not-used"),
+        DeletedMessage("R0922", "abstract-class-little-used"),
```
Pierre-Sassoulas, line 192: This was not documented before, I had to dig a little.

pylint/lint/message_state_handler.py
```diff
@@ -411,9 +412,14 @@ def process_tokens(self, tokens: list[tokenize.TokenInfo]) -> None:
                         try:
                             meth(msgid, "module", l_start)
                         except exceptions.UnknownMessageError:
```
DanielNoord, line 414: I just wonder, there are more place where we catch this exception. Should we also add this special handling there?

pylint/message/message_definition_store.py
```diff
@@ -32,6 +33,28 @@ def __init__(self) -> None:
...
+    @staticmethod
+    def is_deleted_msgid_or_symbol(msgid_or_symbol: str) -> str | None:
+        """Return the explanation for removal if the message was removed."""
+        for explanation, deleted_messages in DELETED_MESSAGES.items():
+            for deleted_message in deleted_messages:
+                if msgid_or_symbol in [deleted_message.msgid, deleted_message.symbol]:
+                    return explanation
+        return None
+
+    def kwargs_for_bad_option_value_add_message(
```
DanielNoord, line 45: Let's make both of these private while we haven't documented our public API specifically.

tests/config/functional/ini/pylintrc_with_missing_comma.2.out
```diff
@@ -1,3 +1,3 @@
 ************* Module {abspath}
-{relpath}:1:0: E0012: Bad option value for --disable. Don't recognize message logging-not-lazylogging-format-interpolation. (bad-option-value)
-{relpath}:1:0: E0012: Bad option value for --enable. Don't recognize message locally-disabledsuppressed-message. (bad-option-value)
+{relpath}:1:0: E0012: Bad option value for '--disable'. Don't recognize message 'logging-not-lazylogging-format-interpolation'. (bad-option-value)
```
DanielNoord, line 2: We should probably join the messages with a comma.

DanielNoord (commented):
Well done on finding a nice fix for this. This does indeed seem like a good middle-ground.

And well done on finding those old messages.

Some earlier comments before I do a full review/

pylint/message/message_definition_store.py
Pierre-Sassoulas: We're using it in PyLinter ourself so it's complicated to make them private. I'm also ok with poeple using the ``is_deleted_msgid_or_symbol`` as a public API. What about making ``kwargs_for_bad_option_value_add_message`` protected ?

tests/config/functional/ini/pylintrc_with_missing_comma.2.out
DanielNoord: @Pierre-Sassoulas My bad, I meant the message that we don't recognise 😅 

But the comma you added now is also a good changed imo 👍

DanielNoord: Oops, I was looking at another test just above it and thought they were connected. This is indeed correct.

Pierre-Sassoulas:
The primer actually find something interesting in astroid, we can't use parenthesis to justify a pragma. We had a discussion about this but I think we should just fix this in 2.14.1 and do better later.

pylint/lint/message_state_handler.py
Pierre-Sassoulas: Well, something really elegant would be to reraise a ``DeletedMessageError`` with the justification...

DanielNoord:
> The primer actually find something interesting in astroid, we can't use parenthesis to justify a pragma. We had a discussion about this but I think we should just fix this in 2.14.1 and do better later.

Should be fixed with an additional #. Other tools require this as well, so I don't think it's too big of an issue.

_I believe mypy doesn't even recognise its disables if they come after an pylint disable. They want priority 😄_

.pyenchant_pylint_custom_dict.txt
```diff
@@ -325,6 +325,7 @@ typechecking
 typeddict
 typehint
 typeshed
+typoes
```
jacobtylerwalls, line 328: Don't blame me, blame https://www.merriam-webster.com/dictionary/typos
```suggestion
```

doc/whatsnew/2/2.14/full.rst
```diff
@@ -31,6 +31,11 @@ Release date: TBA
 
   Closes #6624
 
+* Deleted messages that do not exists anymore in pylint now raise ``useless-option-value`` instead of ``bad-option-value``
+  which permit to distinguish between genuine typoes and old configuration that could be cleaned.
```
jacobtylerwalls, line 35:
```suggestion
* Deleted messages that do not exist anymore in pylint now raise ``useless-option-value`` instead of ``bad-option-value``,
  which permit to distinguish between genuine typos and old configuration that could be cleaned.
```

jacobtylerwalls (approved):

.pyenchant_pylint_custom_dict.txt
Pierre-Sassoulas: I've been doing this mistakes for years. Years 😄

Pierre-Sassoulas:
Sorry to make changes after your review @jacobtylerwalls, but I figured:

- ``bad-option-value`` is not an "error", in the sense that it's not going to break the code, it's only a potential problem in pylint. So it make sense to make it a warning and to not be warned for this in ``--error-only`` mode (only problem could be if a user think he enabled a message and it's not enabled, but it must have been disabled previously so I think it's unlikely to happen).
- We were adding a new message in a patch version, we need to use old_names so the previous disable still works. Otherwise people that already got bothered by 2.14.0 and disabled bad-option-value will still get ``useless-option-value``... that can be annoying.

So I'm going to add a new commit using ``old-names`` this time.

DanielNoord:
@Pierre-Sassoulas I'd advise against making this an `I` warning. I think `W` is better. Information warnings are turned off by default and need to be turned on individually. So nobody will likely every find this message.

doc/data/messages/u/useless-option-value/bad.py
```diff
@@ -0,0 +1,3 @@
+"""'bad-continuation' was removed from pylint in https://github.com/PyCQA/pylint/pull/3571"""
```
DanielNoord, line 1: The empty `good.py` looks a bit weird:
https://pylint--6824.org.readthedocs.build/en/6824/user_guide/messages/information/useless-option-value.html

Pierre-Sassoulas: Does https://pylint--6824.org.readthedocs.build/en/6824/user_guide/messages/refactor/useless-option-value.html looks better ?

DanielNoord: Yes I think so!

Pierre-Sassoulas:
Urgh, the number of change I need to do for changing the message type is gross 😄 Do you confirm a refactor level for useless-option-value is ok for you ? I think it's less than a warning, maybe a convention i.e there's a convention to have a clean pylintrc / disable inside the code.

DanielNoord:
> Urgh, the number of change I need to do for changing the message type is gross 😄 Do you confirm a refactor level for useless-option-value is ok for you ? I think it's less than a warning, maybe a convention i.e there's a convention to have a clean pylintrc / disable inside the code.

I'd be okay with refactoring but preference goes to warning:
1) It gives some uniformity to this category of messages. I think they are all pretty much related so they should probably be in the same category
2) There is still the risk if people thinking that they check their code for something while they are not actually. To me that's more of a warning than a refactoring.

Pierre-Sassoulas:
Ok I will keep refactor because I think using a deleted message i.e error is caused by us removing it and there was nothing wrong with it before vs having a typo and the message do not exists and the option is wrong and does nothing whatever the version should not be the same importance. Also I wanted something else than warning to keep the ``0012`` number from ``bad-option-value`` (``E0012``), ``unknown-option-value`` is ``W0012`` so we can't reuse that for ``useless-option-value`` unless the message type is something else than erorr or warning.

DanielNoord:
> Ok I will keep refactor because I think using a deleted message i.e error is caused by us removing it and there was nothing wrong with it before vs having a typo and the message do not exists and the option is wrong and does nothing whatever the version should not be the same importance. Also I wanted something else than warning to keep the `0012` number from `bad-option-value` (`E0012`), `unknown-option-value` is `W0012` so we can't reuse that for `useless-option-value` unless the message type is something else than erorr or warning.

Let's not reuse numbers!! We have been actively trying to avoid that in all other checkers but can't have a CI check because of legacy issues.
Reusing numbers makes it so that you can easily change the level of a message without also changing the number.

Pierre-Sassoulas:
> Let's not reuse numbers!! We have been actively trying to avoid that in all other checkers but can't have a CI check because of legacy issues.
Reusing numbers makes it so that you can easily change the level of a message without also changing the number.

Do you suggest I use a new number instead of reusing ``0012`` ? Isn't it dividing by 4 the msgid we can use ? Also isn't it a good thing that the new msgid is somewhat similar to the old one ? I think it's intuitive, only the message type changed so the character for the message type change but not the number. What's the downside of it ? (It's been done in the past, which is probably what you call "legacy issues" and this is why I "keep the tradition").

DanielNoord:
> Do you suggest I use a new number instead of reusing ``0012`` ? Isn't it dividing by 4 the msgid we can use ? Also isn't it a good thing that the new msgid is somewhat similar to the old one ? I think it's intuitive, only the message type changed so the character for the message type change but not the number. What's the downside of it ? (It's been done in the past, which is probably what you call "legacy issues" and this is why I "keep the tradition"). 

Yeah, but we're still not really running out and there is already an issue about extending the message id's to 5 digits.

The issue is that whenever you decide to change the level of a message you can only do so by also changing the numbers. What if we want to demote the message to informational at some point? Or actually increase its level? I think it's good practice to keep one number per message and have the letter as an indication of the level, not as a trick to increase the number of messages.

We can have 50 (prefixes) * 99 messages without running out. We're nowhere near there yet.

Pierre-Sassoulas:
> What if we want to demote the message to informational at some point? Or actually increase its level? I think it's good practice to keep one number per message and have the letter as an indication of the level, not as a trick to increase the number of messages.

Sure, but isn't it exactly what we're doing now ? ``useless-option-value`` is not a new message, it's the new name of bad-option-value (with its sister message ``unknown-option-value``) ? I agree we should not reuse number in new messages for precisely this scenario.

DanielNoord:
> Sure, but isn't it exactly what we're doing now ? `useless-option-value` is not a new message, it's the new name of bad-option-value (with its sister message `unknown-option-value`) ? I agree we should not reuse number in new messages for precisely this scenario.

Yeah, one of them should keep ``0012``, but the other should get its own number. We don't really have a concept of "sister messages" and I don't think we should start creating those.

Pierre-Sassoulas:
>  We don't really have a concept of "sister messages"

In fact the only other messages used multiple time as old_name is ``missing-docstring`` but we did not change the message type of the children so we did C0111 => [C0012 C0013 C0014]

DanielNoord:
> > We don't really have a concept of "sister messages"
> 
> In fact the only other messages used multiple time as old_name is `missing-docstring` but we did not change the message type of the children so we did C0111 => [C0012 C0013 C0014]

I was more referring to messages that are related but are on different levels. To take `docstring` as an example. Let's say we find `module` docstrings more important we could have `C0012` for `missing-function-docstring` and `E0012` for `missing-module-docstring`.
Coupling messages across different levels like that might only create issues down the line I think. We're changing the symbols of these messages anyway and we're not running out of IDs so I really don't see why would want to couple them like that.
It is also confusing for users as we would have `C0012`, `R0012` and `W0012` which are all different messages but with almost the same id.

Pierre-Sassoulas:
Sorry for rebasing but I think the commits need to be separated. I renamed R0012 to R0022 (first number not used elsewhere in main checker).

DanielNoord (commented):
Just some initial review. If you're able to perhaps split some changes from this PR. There is a lot going on, with some of it just being general refactoring. Makes it quite hard to review this.

After all the opened issues I'd like to get this right once and for all.

Pierre-Sassoulas:
>  There is a lot going on, with some of it just being general refactoring. Makes it quite hard to review this.

Right let's start by https://github.com/PyCQA/pylint/pull/6866

doc/test_messages_documentation.py
```diff
@@ -71,7 +73,6 @@ def get_functional_test_files_from_directory(input_dir: Path) -> List[Tuple[str,
 class LintModuleTest:
     def __init__(self, test_file: Tuple[str, Path]) -> None:
         self._test_file = test_file
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Haha no, it's personal preference, imo compactness is good for readability, I think creating smaller code unit with functions also make more sense if something need to be separated. Anyway I'm going to stop doing that if you don't like it as we don't need churn around this.

doc/test_messages_documentation.py
```diff
@@ -81,22 +82,36 @@ def __init__(self, test_file: Tuple[str, Path]) -> None:
         # Check if this message has a custom configuration file (e.g. for enabling optional checkers).
         # If not, use the default configuration.
         config_file: Optional[Path]
-        if (test_file[1].parent / "pylintrc").exists():
-            config_file = test_file[1].parent / "pylintrc"
+        msgid, full_path = test_file
```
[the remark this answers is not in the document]
Pierre-Sassoulas: The tests run on it though.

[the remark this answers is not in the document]
DanielNoord: They won't if we remove the files. Or at least, they shouldn't 😄 If they are, that's something I should fix!

Pierre-Sassoulas:
Then https://github.com/PyCQA/pylint/pull/6867. Even if we don't need to activate the old messages there was still some things to keep there.

DanielNoord (approved):

PR #6826, opened 2022-06-04
[deleted messages] Refactor the constant to be able to give a reason

Pierre-Sassoulas:
## Type of Changes

## Description

Refactor prior to #6824 in order to be able to give a reason for why the message was deleted, and better overall organization as deleted messages are not public and specific to ``pylint.messages``.

### DIFF
--- a/pylint/config/utils.py
+++ b/pylint/config/utils.py
@@ -203,9 +203,8 @@ def _enable_all_extensions(run: Run, value: str | None) -> None:
 PREPROCESSABLE_OPTIONS: dict[
     str, tuple[bool, Callable[[Run, str | None], None], int]
 ] = {  # pylint: disable=consider-using-namedtuple-or-dataclass
-    # pylint: disable=wrong-spelling-in-comment
     # Argparse by default allows abbreviations. It behaves differently
-    # if you turn this off, so we also turn it on. We mimick this
+    # if you turn this off, so we also turn it on. We mimic this
     # by allowing some abbreviations or incorrect spelling here.
     # The integer at the end of the tuple indicates how many letters
     # should match, include the '-'. 0 indicates a full match.
--- a/tests/functional/b/bad_option_value.txt
+++ b/tests/functional/b/bad_option_value.txt
@@ -1,1 +1,9 @@
-bad-option-value:1:0:None:None::Bad option value for enable. Don't recognize message W04044.:UNDEFINED
+bad-option-value:4:0:None:None::Bad option value for disable. Don't recognize message C05048.:UNDEFINED
+bad-option-value:6:0:None:None::Bad option value for disable. Don't recognize message execfile-builtin.:UNDEFINED
+bad-option-value:8:0:None:None::Bad option value for disable. Don't recognize message W1656.:UNDEFINED
+bad-option-value:10:0:None:None::Bad option value for disable-next. Don't recognize message R78948.:UNDEFINED
+bad-option-value:12:0:None:None::Bad option value for disable-next. Don't recognize message deprecated-types-field.:UNDEFINED
+bad-option-value:14:0:None:None::Bad option value for disable-next. Don't recognize message W1634.:UNDEFINED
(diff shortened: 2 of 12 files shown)


pylint/message/deleted_message_ids.py
```diff
@@ -0,0 +1,146 @@
...
+from typing import NamedTuple
+
+
+class DeletedMessage(NamedTuple):
+    msgid: str
+    symbol: str
+    old_names: list[tuple[str, str]] = []
+
+
+_DELETED_MSGID_PREFIXES: list[int] = []
```
DanielNoord, line 17: If we underscore this file we can then not underscore anything else in this file 😄

Pierre-Sassoulas: Clever 😄

pylint/config/utils.py
```diff
@@ -203,9 +203,8 @@ def _enable_all_extensions(run: Run, value: str | None) -> None:
 PREPROCESSABLE_OPTIONS: dict[
     str, tuple[bool, Callable[[Run, str | None], None], int]
 ] = {  # pylint: disable=consider-using-namedtuple-or-dataclass
-    # pylint: disable=wrong-spelling-in-comment
     # Argparse by default allows abbreviations. It behaves differently
-    # if you turn this off, so we also turn it on. We mimick this
+    # if you turn this off, so we also turn it on. We mimic this
```
Pierre-Sassoulas, line 207: Sorry about the unrelated change but I don't know why suddenly pre-commit pylint was not happy about this.

Pierre-Sassoulas: I must have a dictionary with both mimic and mimick.

Pierre-Sassoulas: We might want to add to our own dictionary instead of disabling because the ``useless-suppression`` it will get annoying fast.

pylint/message/deleted_message_ids.py
Pierre-Sassoulas: Done

pylint/config/utils.py
DanielNoord: This is because you removed a `disable` on L206.

tests/functional/b/bad_option_value.txt
```diff
@@ -1 +1,9 @@
-bad-option-value:1:0:None:None::Bad option value for enable. Don't recognize message W04044.:UNDEFINED
+bad-option-value:4:0:None:None::Bad option value for disable. Don't recognize message C05048.:UNDEFINED
```
DanielNoord, line 1: Do you mind doing a follow up PR just before the other PR which changes the confidence for `bad-option-value`?

Please don't do it in this PR as it already quite large.

DanielNoord (approved):

tests/functional/b/bad_option_value.txt
Pierre-Sassoulas: Yeah sure, I already have a lot of cherry-pick to do for 2.14.1 anyway 😄

DanielNoord: 😅 

All good fixes though 😄

pylint/config/utils.py
Pierre-Sassoulas: No, I had the useless suppression, thougth it was a spellcheck issue, "fixed" "mimick" (suggestion from IDE) then still had an useless-suppression, and removed the disable. I think it's the dict that has "mimick" in it on ubuntu.

DanielNoord:
@Pierre-Sassoulas This breaks `main`. The `pylint` check didn't pass.

pylint/config/utils.py
DanielNoord: Let's add `useless-suppression` to the disable there then. Because the other stuff in that comment still needs to be disabled.

Pierre-Sassoulas:
> This breaks main. The pylint check didn't pass.

Damn the spellcheck is fragile. It worked locally and in CI. I don't think we can do anything about this, ``enchant`` itself is fragile. Let me fix this.

PR #6836, opened 2022-06-05
Add `trailing-whitespace` example

harupy:

## Type of Changes

## Description

Refs #5953 

Closes #XXXX

### DIFF
--- a/doc/data/messages/t/trailing-whitespace/bad.py
+++ b/doc/data/messages/t/trailing-whitespace/bad.py
@@ -1,1 +1,1 @@
-print("Hello")  # [trailing-whitespace]
+print("Hello")  # [trailing-whitespace]  
--- a/doc/data/messages/t/trailing-whitespace/bad.py
+++ b/doc/data/messages/t/trailing-whitespace/bad.py
@@ -1,1 +1,2 @@
-print("Hello")  # [trailing-whitespace]  
+print("Hello")  # [trailing-whitespace]   
+#                                       ^^^ trailing whitespaces
--- a/doc/data/messages/t/trailing-whitespace/bad.py
+++ b/doc/data/messages/t/trailing-whitespace/bad.py
@@ -1,1 +1,1 @@
-print("Hello")  # [trailing-whitespace]  
+print("Hello")  # [trailing-whitespace]
--- a/doc/data/messages/t/trailing-whitespace/bad.py
+++ b/doc/data/messages/t/trailing-whitespace/bad.py
@@ -1,0 +1,1 @@
+print("Hello")  # [trailing-whitespace]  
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (commented):
You're going to have to add an exclude in the pre-commit configuration for this one. Check this : https://github.com/PyCQA/pylint/blob/main/.pre-commit-config.yaml#L40, but for this hook : https://github.com/PyCQA/pylint/blob/main/.pre-commit-config.yaml#L8

harupy:
@Pierre-Sassoulas Thanks for the comment. What should the `exclude` field for `black` look like?

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 833ba0fff..996fc2867 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     rev: v4.2.0
     hooks:
       - id: trailing-whitespace
-        exclude: tests(/\w*)*/functional/t/trailing_whitespaces.py|tests/pyreverse/data/.*.html
+        exclude: tests(/\w*)*/functional/t/trailing_whitespaces.py|tests/pyreverse/data/.*.html|doc/data/messages/t/trailing-whitespace/bad.py
       - id: end-of-file-fixer
         exclude: tests(/\w*)*/functional/m/missing/missing_final_newline.py|tests/functional/t/trailing_newlines.py
   - repo: https://github.com/myint/autoflake
@@ -43,7 +43,7 @@ repos:
     hooks:
       - id: black
         args: [--safe, --quiet]
-        exclude: *fixtures
+        exclude: ???
   - repo: https://github.com/Pierre-Sassoulas/black-disable-checker
     rev: v1.1.0
     hooks:
```

Pierre-Sassoulas:
I'm not sure we need to exclude in black can we try with only the trailing whitespace first ? If it does not work we'll need to change the fixture definition (``& fixture``)

doc/data/messages/t/trailing-whitespace/bad.py
```diff
@@ -0,0 +1 @@
+print("Hello")  # [trailing-whitespace]
```
harupy, line 1:
```suggestion
print("Hello")  # [trailing-whitespace]   
                #                      ^^^ whitespaces
```

Do we want to mark whitespaces like this?

Pierre-Sassoulas: Good idea, that could be helpful when reading the doc. Especially since I don't know if sphinx is not automatically removing trailing whitespaces.

Pierre-Sassoulas:
```suggestion
print("Hello")  # [trailing-whitespace]   
#                                                   ^^^ trailing whitespaces
```

Pierre-Sassoulas (approved):

PR #6837, opened 2022-06-05
Add `unused-variable` example

harupy:

## Type of Changes

## Description

Refs #5953

Closes #XXXX

### DIFF
--- a/doc/data/messages/u/unused-variable/bad.py
+++ b/doc/data/messages/u/unused-variable/bad.py
@@ -1,0 +1,4 @@
+def foo():
+    x = 1
+    y = None  # [unused-variable]
+    print(x)
--- a/doc/data/messages/u/unused-variable/bad.py
+++ b/doc/data/messages/u/unused-variable/bad.py
@@ -1,4 +1,4 @@
-def foo():
-    x = 1
-    y = None  # [unused-variable]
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (changes requested):
Looks good thank you ! Could you use an example without using ``foo`` ? Generally we use fruit, animal school or cultural reference in examples. It makes the doc less "dry" and beginners do not understand "foo" / " bar".

doc/data/messages/u/unused-variable/bad.py
```diff
@@ -0,0 +1,4 @@
+def foo():
+    x = 1
+    y = None  # [unused-variable]
+    print(x)
```
harupy, line 4:
```python
def print_fruits():
    orange = "orange"
    apple = "apple"  # [unused-variable]
    print(orange)
```

harupy, line 4:
```suggestion
def print_fruits():
    orange = "orange"
    apple = "apple"  # [unused-variable]
    print(orange)
```

Pierre-Sassoulas: 👌

Pierre-Sassoulas (approved):

PR #6844, opened 2022-06-05
Add `unnecessary-lambda` example

harupy:
Signed-off-by: harupy <hkawamura0130@gmail.com>

## Type of Changes

## Description

Refs #5953

Closes #XXXX

### DIFF
--- a/doc/data/messages/u/unnecessary-lambda/bad.py
+++ b/doc/data/messages/u/unnecessary-lambda/bad.py
@@ -1,1 +1,5 @@
-map(lambda x: hash(x), ["x", "y"])  # [unnecessary-lambda]
+function = lambda x: print(x)  # [unnecessary-lambda]
+
+function("Hello world !")
+
+df.apply(lambda x: str(x))  # [unnecessary-lambda]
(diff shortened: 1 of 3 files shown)


doc/data/messages/u/unnecessary-lambda/bad.py
```diff
@@ -0,0 +1 @@
+map(lambda x: hash(x), ["x", "y"])  # [unnecessary-lambda]
```
Pierre-Sassoulas, line 1:
```suggestion
function = lambda x: print(x)  # [unnecessary-lambda]

function("Hello world !")

df.apply(lambda x: str(x))  # [unnecessary-lambda]
```

doc/data/messages/u/unnecessary-lambda/good.py
```diff
@@ -1 +1 @@
-# This is a placeholder for correct code for this message.
+map(hash, ["x", "y"])
```
Pierre-Sassoulas, line 1:
```suggestion
print("Hello world !")

df.apply(str)
```

Pierre-Sassoulas (changes requested):
Let's use simpler and well known function here.

Pierre-Sassoulas (approved):

PR #6845, opened 2022-06-05
Suppress ``unnecessary-list-index-lookup`` for whole loop on write

### LINKED ISSUE #6818: unnecessary-list-index-lookup doesn't detect list mutation in a nested while loop
### Bug description

Simple example:
```python
$ cat -n a.py 
     1  eqs = [1, 2, 3, 11, 4]
     2  for i, e in enumerate(eqs):
     3      while eqs[i] < 10:
     4          eqs[i] += 1
```
Similar issue has unnecessary-dict-index-lookup

BTW, another simple failure is https://github.com/PyCQA/pylint/issues/6788.  These checkers seems to be too naive to be enabled per default (as in the 2.14).

### Configuration

```ini
None
```

### Command used

```shell
pylint -j1 a.py
```

### Pylint output

```shell
************* Module a
a.py:3:10: R1736: Unnecessary list index lookup, use 'e' instead (unnecessary-list-index-lookup)

```

### Expected behavior

No R1736.

### Pylint version

```shell
pylint 2.14.0
astroid 2.11.5
Python 3.10.2+ (heads/3.10:75c5dbc27e, Feb 19 2022, 06:58:32) [GCC 10.2.1 20210110]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

timmartin:
## Type of Changes

## Description

The ``unnecessary-list-index-lookup`` checker was assuming that if the
subscript was written to then it would only invalidate access on later
lines, but this is not necessarily the case if an inner loop is nested
inside the one being checked.

Closes #6818


timmartin:
I was going to apply the same change to `unnecessary-dict-index-lookup`, but I discovered that we already have a regression test that would be regressed by this change:

    for k, v in a_dict.items():
        print(a_dict[k])  # [unnecessary-dict-index-lookup]
        print(b_dict[k])  # Should not emit warning, accessing other dictionary
        a_dict[k] = 123  # Should not emit warning, key access necessary
        a_dict[k] += 123  # Should not emit warning, key access necessary
        print(a_dict[k])  # Should not emit warning, v != a_dict[k]****

We could just update the regression test, but I was going to have a go at changing the behavior to detect whether there is a loop of any kind nested within the outer `for` loop. I think if there is no inner loop we can stick with the old behavior and only suppress the warnings if there is a nested loop. This seems reasonable, but I'm worried that it will make the code hard to follow.

timmartin:
I'm not sure I understand the build errors here. It seems to be a problem reported from a Javascript process, so it's not obvious how my changes could have affected it.

Pierre-Sassoulas:
I think it's due to other fails before:

```
************* Module pylint.checkers.refactoring.refactoring_checker
pylint/checkers/refactoring/refactoring_checker.py:2136:0: C0401: Wrong spelling of a word 'fo' in a comment:
# end fo the loop without updating the collection
      ^^
Did you mean: ''few' or 'of''? (wrong-spelling-in-comment)
```

Primer are costly, so if other job fails we cut them short I think.

DanielNoord (commented):
One nit about typing. Change itself looks good to resolve the current issue!

Pierre-Sassoulas (approved):
LGTM, thank you @timmartin !

PR #6853, opened 2022-06-05
Add `repeated-keyword` example

harupy:
## Type of Changes

## Description

Refs #5953

Closes #XXXX

### DIFF
--- a/doc/data/messages/r/repeated-keyword/bad.py
+++ b/doc/data/messages/r/repeated-keyword/bad.py
@@ -1,2 +1,6 @@
-print("Hello ", "orld", sep="W", end=" !", **{"end": " :("})  # [repeated-keyword]
-print("Hello ", "orld", **{"sep": "W", "end": " !"}, **{"end": " :("})  # [repeated-keyword]
+def func(a, b, c):
+    return a, b, c
+
+
+func(1, 2, c=3, **{"c": 4})  # [repeated-keyword]
+func(1, 2, **{"c": 3}, **{"c": 4})  # [repeated-keyword]
--- a/doc/data/messages/r/repeated-keyword/bad.py
+++ b/doc/data/messages/r/repeated-keyword/bad.py
@@ -3,4 +3,4 @@ def func(a, b, c):
 
 
 func(1, 2, c=3, **{"c": 4})  # [repeated-keyword]
-func(1, 2, c=3, **{"c": 3}, **{"c": 4})  # [repeated-keyword]
+func(1, 2, **{"c": 3}, **{"c": 4})  # [repeated-keyword]
--- a/doc/data/messages/r/repeated-keyword/bad.py
+++ b/doc/data/messages/r/repeated-keyword/bad.py
@@ -1,0 +1,6 @@
+def func(a, b, c):
+    return a, b, c
+
+
+func(1, 2, c=3, **{"c": 4})  # [repeated-keyword]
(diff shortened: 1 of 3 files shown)


doc/data/messages/r/repeated-keyword/bad.py
```diff
@@ -0,0 +1,6 @@
+def func(a, b, c):
+    return a, b, c
+
+
+func(1, 2, c=3, **{"c": 4})  # [repeated-keyword]
+func(1, 2, **{"c": 3}, **{"c": 4})  # [repeated-keyword]
```
Pierre-Sassoulas, line 6:
```suggestion
print("Hello ", "orld", sep="W", end=" !", **{"end": " :("})  # [repeated-keyword]
print("Hello ", "orld", **{"sep": "W", "end": " !"}, **{"end": " :("})  # [repeated-keyword]
```

doc/data/messages/r/repeated-keyword/good.py
```diff
@@ -1 +1,5 @@
-# This is a placeholder for correct code for this message.
+def func(a, b, c):
+    return a, b, c
+
+
+func(1, 2, c=3)
```
Pierre-Sassoulas, line 5:
```suggestion
print("Hello ", "orld", sep="W", end=" !")
print("Hello ", "orld", **{"sep": "W", "end": " !"})
```

Pierre-Sassoulas (commented):
Let me know what you think :)

doc/data/messages/r/repeated-keyword/bad.py
```diff
@@ -0,0 +1,2 @@
+print("Hello ", "orld", sep="W", end=" !", **{"end": " :("})  # [repeated-keyword]
+print("Hello ", "orld", **{"sep": "W", "end": " !"}, **{"end": " :("})  # [repeated-keyword]
```
harupy, line 2: Interesting. These two lines don't raise an error.

harupy:
```python
def print(*args, **kwargs):
    pass

print("Hello ", "orld", sep="W", end=" !", **{"end": " :("})  # [repeated-keyword]
print("Hello ", "orld", **{"sep": "W", "end": " !"}, **{"end": " :("}) # [repeated-keyword]
```

worked.

harupy:
```python
def f(*args, **kwargs):
    pass

f("Hello ", "orld", sep="W", end=" !", **{"end": " :("})  # [repeated-keyword]
f("Hello ", "orld", **{"sep": "W", "end": " !"}, **{"end": " :("}) # [repeated-keyword]
```

also worked.

Pierre-Sassoulas: Hurgh, look like a bug in pylint. Maybe we can't determine the signature of print properly.

Pierre-Sassoulas: I opened #6856, but maybe there's another well known function that exists without the bug ?

harupy: Thanks! How about `open`?

Pierre-Sassoulas: Well this is not raising anything either:
```python
with open("file", "r", encoding="utf8", **{"encoding": "ascii"}) as f:
    f.read()
```
Let's keep your original example.

Pierre-Sassoulas (approved):

PR #6855, opened 2022-06-05
Add `raising-non-exception` example

harupy:
## Type of Changes

## Description

Refs #5953

Closes #XXXX

### DIFF
--- a/doc/data/messages/r/raising-non-exception/good.py
+++ b/doc/data/messages/r/raising-non-exception/good.py
@@ -1,1 +1,1 @@
-# This is a placeholder for correct code for this message.
+raise Exception
--- a/doc/data/messages/r/raising-non-exception/good.py
+++ b/doc/data/messages/r/raising-non-exception/good.py
@@ -1,1 +1,1 @@
-raise Exception
+raise Exception("Goodbye world !")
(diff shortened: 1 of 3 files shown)


doc/data/messages/r/raising-non-exception/good.py
```diff
@@ -1 +1 @@
-# This is a placeholder for correct code for this message.
+raise Exception
```
Pierre-Sassoulas, line 1:
```suggestion
raise Exception("Goodbye world !")
```

Pierre-Sassoulas:
~~I think we should have valid code in good.py~~ Nvm did not know it's possible to raise without a message. It's bad style though, what do you think ?

harupy:
> ~I think we should have valid code in good.py~ Nvm did not know it's possible to raise without a message. It's bad style though, what do you think ?

`raise Excpetion` is equivalent to `raise Exception`.

https://docs.python.org/3/tutorial/errors.html#raising-exceptions explains as follows:

> The sole argument to [raise](https://docs.python.org/3/reference/simple_stmts.html#raise) indicates the exception to be raised. This must be either an exception instance or an exception class (a class that derives from [Exception](https://docs.python.org/3/library/exceptions.html#Exception)). If an exception class is passed, it will be implicitly instantiated by calling its constructor with no arguments:

Pierre-Sassoulas (approved):
Thank you for teaching me something 😉

PR #6860, opened 2022-06-05
Release branch for 2.14.1

Pierre-Sassoulas:


Pierre-Sassoulas:
Ha I see. Maybe we can limit the new primer to pull request that are asking to merge in the main branch ? Merging maintainance 2.14 back on main after a release is a whole thing in itself, so pre-emptively making it mergeable is going to be painful. In the end we're going to get a comparison of main against the maintenance branch that won't tell us a lot apart that we did not cherry-pick everything and that there are new feature in 2.15, right ?

DanielNoord:
> Ha I see. Maybe we can limit the new primer to pull request that are asking to merge in the main branch ? Merging maintainance 2.14 back on main after a release is a whole thing in itself, so pre-emptively making it mergeable is going to be painful. In the end we're going to get a comparison of main against the maintenance branch that won't tell us a lot apart that we did not cherry-pick everything and that there are new features in 2.15, right ?

Yeah, although the check for fatal errors might still be useful? There might be PRs that miss the `needs backport` label which are necessary? Although that's probably rare.

Better to exclude all PRs not targeting `main`.

Pierre-Sassoulas:
If I remember correctly we removed some old style primer in favor of the new style primer. Maybe we should run old style primer for maintenance branch too (i.e. not targeting main) ? I'm pretty confident that most commit are atomic and will work well together if the "normal" tests are working correctly so it feel like being overtly cautious.

PR #6861, opened 2022-06-05
Add `compare-to-empty-string` example

harupy:
## Type of Changes

## Description

Refs #5953

### DIFF
--- a/doc/data/messages/c/compare-to-empty-string/bad.py
+++ b/doc/data/messages/c/compare-to-empty-string/bad.py
@@ -1,8 +1,8 @@
 x = ""
 y = "hello"
 
-if not x:
+if x == "":  # [compare-to-empty-string]
     print("x is an empty string")
 
-if y:
+if y != "":  # [compare-to-empty-string]
     print("y is not an empty string")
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (commented):
I did not remember we had this check. This is very, very similar to ``use-implicit-booleaness-not-x`` type check, it kinda make me want to rename this or move them all in their own checker to make it coherent. Do you have an opinion on this @harupy ?

harupy:
> move them all in their own checker to make it coherent.

Sounds good me :)

Pierre-Sassoulas:
Ok, I took a closer look. ``compare-to-zero`` and ``compare-to-empty-string``are quite more opinionated than  ``use-implicit-booleaness-not-len`` and  ``use-implicit-booleaness-not-copmparison`` which is probably why they are in their own optional checker. 

Doing ``x != 0`` or feel a lot safer and easier to understand than ``bool(x)`` for int (you need a background in C to understand that). Renaming them could be an option, I'm going to open an issue to discuss it but in the meantime I think we could merge this.

Pierre-Sassoulas (approved):
Created https://github.com/PyCQA/pylint/issues/6871 if you want to chime in :)

PR #6713, opened 2022-05-27
Refactor and deprecate resolving of interface implementations in `pyreverse`

DudeNr33:
- [x] Add a ChangeLog entry describing what your PR does.
- [x] If it's a new feature, or an important bug fix, add a What's New entry in
      `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.
- [x] If you used multiple emails or multiple names when contributing, add your mails
   and preferred name in ``script/.contributors_aliases.json``

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description
Ref #6582 
As we also deprecated the use of `__implements__` in our own codebase, deprecating the `pyreverse` functionality to resolve and display implemented interfaces is the right consequence.
To make this change transparent for users, a `DeprecationWarning` will be issued when such an `__implements__` attribute has been found.

Additionally, I removed parameters and checks that were always constant in production use of the code.

PR #6869, opened 2022-06-06
Better target DeprecationWarnings in pyreverse

jacobtylerwalls:
## Type of Changes

## Description
Reduces 26 instances of this `DeprecationWarning` in the suite to 1. (That 1 will need a substantive fix.)
Refs #6713

### DIFF
--- a/pylint/pyreverse/inspector.py
+++ b/pylint/pyreverse/inspector.py
@@ -187,13 +187,14 @@ class Linker(IdGeneratorMixIn, utils.LocalsVisitor):
             ifaces = interfaces(node)
             if ifaces is not None:
                 node.implements = list(ifaces)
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Make sense to me, but let's wait the review by DudeNr33 to merge :)

DudeNr33 (approved):
Thanks for catching that. 👍 
The way it was implemented by me was actually wrong, as the warning also triggered when no `__implements__` attribute was present. So you'd get the warning every time when running `pyreverse` with warnings enabled.

While I added a unit test to check if the `DeprecationWarning` is emitted if `__implements__` exists, I forgot to add a test case for the case of a "normal" class...

PR #6870, opened 2022-06-06
Merging of ``use-implicit-booleaness``-like checkers for consistency and performance

### LINKED ISSUE #5234: C1901 (compare-to-empty-string) message is not explicit about the problem and the solution
### Bug description

Consider

```python
   val_type, val_value = get_type(name)  # returns str, Any
   if val_type is Environment:
     if val_value is None or val_value is "":  # or ==
```

The last line produces a pylint error, with the common guidance (e.g https://vald-phoenix.github.io/pylint-errors/plerr/errors/compare-to-empty-string/C1901.html) suggesting that you should use:

```python
    if val_value is None or not len(val_value):
```

The user used `val_value is ""` to explicitly say: is it a string, and is it empty?

To avoid doing this `is` or `==` comparison, the user would need to write:

```python
    if val_value is None or (isinstance(val_value, str) and not len(val_value)):
```

example of following the guidance too literally:

```python
    def fn(v):
      if len(v):
          return "empty"
      return "other"

    fn(None)    # OK: returns "other"
    fn("a")     # OK: returns "other"
    fn([None])  # OK: returns "other"
    fn("")      # OK: returns "empty"
    fn([])      # WRONG: returns "other"
    fn(0)       # ERROR: int has no len
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
a.py.3.1: [C1901(compare-to-empty-string), a.py] Avoid comparisons to empty string [229,1]
```

### Expected behavior

This should either be disabled by default or it should provide better guidance. Why is `x is ""` bad?

### Pylint version

```shell
pylint 1.9.5,
astroid 1.6.6
Python 2.7.18 (v2.7.18:8d21aa21f2, Apr 20 2020, 13:25:05) [MSC v.1500 64 bit (AMD64)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

### LINKED ISSUE #6871: Uniformize ``compare-to-zero``, ``compare-to-empty-string`` with the ``use-implicit-booleaness-x`` checks
### Current problem

 ``compare-to-zero`` and ``compare-to-empty-string`` are the same check basically than ``use-implicit-booleaness-not-copmparison``.  

Doing ``x != 0`` feel safer and easier to understand than ``bool(x)`` for int (you need a background in C to understand that). So``compare-to-zero`` feel more opinionated  which is probably why they are in their own optional checker. 

``compare-to-empty-string`` though is really an ``use-implicit-booleaness-not-copmparison`` check.  

### Desired solution

I think it would make sense to merge ``compare-to-empty-string`` . Renaming them so they all follow the same pattern could be an option.

### Additional context

https://github.com/PyCQA/pylint/pull/6861

I've done the work in #6870 then realized that ``compare-to-zero`` was opinionated.

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to #8630 
Closes #6871
Closes #5234

### DIFF
--- a/pylint/checkers/refactoring/implicit_booleaness_checker.py
+++ b/pylint/checkers/refactoring/implicit_booleaness_checker.py
@@ -74,14 +74,14 @@ class ImplicitBooleanessChecker(checkers.BaseChecker):
             {"old_names": [("C1801", "len-as-condition")]},
         ),
         "C1803": (
-            '"%s" can be simplified to "%s" as an empty %s is falsey',
+            '"%s" can be simplified to "%s", if it is strictly a sequence, as an empty %s is falsey',
             "use-implicit-booleaness-not-comparison",
             "Used when Pylint detects that collection literal comparison is being "
             "used to check for emptiness; Use implicit booleaness instead "
             "of a collection classes; empty collections are considered as false",
         ),
         "C1804": (
-            '"%s" can be simplified to "%s" as an empty string is falsey',
+            '"%s" can be simplified to "%s", if it is striclty a string, as an empty string is falsey',
             "use-implicit-booleaness-not-comparison-to-string",
             "Used when Pylint detects comparison to an empty string constant.",
             {
@@ -90,7 +90,7 @@ class ImplicitBooleanessChecker(checkers.BaseChecker):
             },
         ),
         "C1805": (
-            '"%s" can be simplified to "%s" as 0 is falsey',
+            '"%s" can be simplified to "%s", if it is strictly an int, as 0 is falsey',
             "use-implicit-booleaness-not-comparison-to-zero",
             "Used when Pylint detects comparison to a 0 constant.",
             {"default_enabled": False, "old_names": [("C2001", "compare-to-zero")]},
--- a/pylint/checkers/refactoring/implicit_booleaness_checker.py
+++ b/pylint/checkers/refactoring/implicit_booleaness_checker.py
@@ -74,7 +74,7 @@ class ImplicitBooleanessChecker(checkers.BaseChecker):
             {"old_names": [("C1801", "len-as-condition")]},
         ),
         "C1803": (
-            "'%s' can be simplified to '%s' as an empty %s is falsey",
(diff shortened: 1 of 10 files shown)


pylint/checkers/refactoring/implicit_booleaness_checker.py
```diff
@@ -4,13 +4,21 @@
...
+
 import astroid
 from astroid import bases, nodes
 
 from pylint import checkers
 from pylint.checkers import utils
 
 
+def _is_constant_zero(node):
+    return isinstance(node, astroid.Const) and node.value == 0
```
Pierre-Sassoulas, line 19: This trigger the check which made me think it's opinionated.

pylint/checkers/refactoring/implicit_booleaness_checker.py
```diff
@@ -139,6 +152,80 @@ def visit_unaryop(self, node: nodes.UnaryOp) -> None:
     @utils.only_required_for_messages("use-implicit-booleaness-not-comparison")
     def visit_compare(self, node: nodes.Compare) -> None:
         self._check_use_implicit_booleaness_not_comparison(node)
+        self._check_compare_to_empty_string(node)
+        self._check_compare_to_zero(node)
+
+    def _check_compare_to_zero(self, node: nodes.Compare) -> None:
```
Pierre-Sassoulas, line 158: Those two functions are highly duplicated, they could be merged if we decide that everything can be merged in this checker.

Pierre-Sassoulas:
Following #7629 we can merge everything in the same checker but still make one check optional by default.

Pierre-Sassoulas:
There might be some optimization between ``"use-implicit-booleaness-not-comparison`` and the two other but they're not obvious, the logic of "to-string" and "to-zero" would need a serious refactor. I think it's good enough for now.

pylint/checkers/refactoring/implicit_booleaness_checker.py
Pierre-Sassoulas: I did that.

DanielNoord (approved):

PR #6872, opened 2022-06-06
Remove all placeholder files

DanielNoord:

## Type of Changes

### DIFF
--- a/doc/data/messages/a/astroid-error/details.rst
+++ b/doc/data/messages/a/astroid-error/details.rst
@@ -1,1 +1,0 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
--- a/doc/data/messages/a/astroid-error/good.py
+++ b/doc/data/messages/a/astroid-error/good.py
@@ -1,1 +1,0 @@
-# This is a placeholder for correct code for this message.
--- a/doc/data/messages/b/bad-configuration-section/details.rst
+++ b/doc/data/messages/b/bad-configuration-section/details.rst
@@ -1,1 +1,0 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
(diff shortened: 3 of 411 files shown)


DanielNoord:
@Pierre-Sassoulas I should have tested this but the assertion you added doesn't actually work. The placeholder text isn't in the string that is being asserted anyway.

Let's remove all those files just to be sure they don't stay around.

Pierre-Sassoulas:
It's very efficient to know what still need to be done (I'm so sick of updating #5953 all the time.) and thanks to it contributor do not need to create the directory and good.py which prevent errors and is nice. How about fixing the assert ? I also thougth about creating a script to generate #5953's updated description.

DanielNoord:
> It's very efficient to know what still need to be done (I'm so sick of updating #5953 all the time.) and thanks to it contributor do not need to create the directory and good.py which prevent errors and is nice. How about fixing the assert ? I also thougth about creating a script to generate #5953's updated description.

I was just going to update #5953. I'll do that once a month or something.

Fixing the assert is a bit annoying as the strings we return have the `literalinclude` directive instead of the actual text.

I'd prefer to remove these files as they can only be forgotten. I'll make sure to update #5953 a little more often.

Pierre-Sassoulas:
There's also the fact that it's advertisement in our documentation for the issue when someone search for it. What's wrong with having to check that they don't stay around in 6 months when most of the messages are done and there's very few of those file anyway ?

DanielNoord:
> There's also the fact that it's advertisement in our documentation for the issue when someone search for it. What's wrong with having to check that they don't stay around in 6 months when most of the messages are done and there's very few of those file anyway ?

Okay but then the assertion should be fixed. Could you look into that? As of now it doesn't do much.

Pierre-Sassoulas:
Let me check the assertion. (One reason to keep it is also to have something telling someone reading the doc that not having an explanation is not "normal" and that we're going to get better maybe with their help. Otherwise we just have an empty page, which feel "not helpful" and "the only thing I'm supposed to get").

Pierre-Sassoulas:
Did some investigation, the assert was broken yesterday in https://github.com/PyCQA/pylint/commit/beb8b6f921057aeea0631d9440bb7acd7f8603bc#diff-c451abc3b7961e84318d215e0b2f2e12860a4f416ff70df520674bbf83b28927R109

PR #6874, opened 2022-06-06
Add `implicit-str-concat` example

harupy:
## Type of Changes

## Description

Refs #5953

Closes #XXXX

### DIFF
--- a/doc/data/messages/i/implicit-str-concat/good.py
+++ b/doc/data/messages/i/implicit-str-concat/good.py
@@ -1,1 +1,1 @@
-# This is a placeholder for correct code for this message.
+x = ["a", "b"]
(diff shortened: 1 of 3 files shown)


doc/data/messages/i/implicit-str-concat/good.py
```diff
@@ -1 +1 @@
-# This is a placeholder for correct code for this message.
+x = ["a", "b"]
```
Pierre-Sassoulas, line 1:
```suggestion
x = ["a", "b"]

with open("hello.txt", "r") as f:
    print(f.read())
```

Pierre-Sassoulas (approved):
Thanks, I was thinking about adding an example of a call where it's immediately obvious something is wrong, what do you think ?

Pierre-Sassoulas (approved):

PR #6875, opened 2022-06-06
First draft for automatic generation of changelog entries

DudeNr33:
- [ ] Add an entry to the change log describing the change in

## Type of Changes

## Changelog

The changelog is now automatically generated from the PR content.
This makes it easier for contributors as merge conflicts of the changelog file are now a thing of the past.

(edited 12.06. to see if the CI check correctly determines a changelog is included in the description)

## Description

Refs #6688 

This is a first draft for automatically generating the changelog using a custom Sphinx extension, which provides a new directive `.. changelog::`.

I discarded the idea of using [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator). While it produced good results on one of my own, small repos, I had some problems when running on pylint's repo. 

The main idea is:

* The `changelog` directive allows the user to generate a list of issues or pull requests matching the query passed in through the `:query:` option. 
* The `:caption:` option is used as title in the final document.
* With `:hide_if_empty:` the whole section will be skipped if no pull requests matching the query were found.
* General configuration options like repository owner, project, token and labels that should always be included can be defined in `conf.py`.

This means that we have to apply the appropriate labels to each pull request. We will also have to come up with new labels for things like "new checker", "removed checker", etc.

I want to discuss some open issues and problems before moving forward:

Open issues:

* [ ] Define corresponding labels.
**Decision:** TBD
* [ ] Define labels that shall always be ignored.
**Decision:** TBD
* [ ] Define what to use as the changelog entry. Currently I take the PR title and add a link to the PR beneath. However, the PR title is not always in a form that fits in the changelog, or is too short to convey all relevant information. We could add a `# Changelog:` section in the PR template and extract this part from the PR body with a regular expression. A CI check could enforce that this section is not removed manually. PR title could still be used as a fallback to make the changelog generation more robust.
**Decision:** TBD

There are also some problems that need to be adressed:

* [ ] Accessing the GitHub API requires a token, as unauthenticated requests are subject to rate limiting. While this is not a problem for generating on readthedocs, our `checks.yml` also tries to build the documentation. We can include a secret to store the token to make it accessible in GitHub actions for our main branch, but secrets are not accessible for PRs that originate from forks. Should we simply skip the changelog generation for PR runs?
**Decision:** TBD
* [ ] I currently use [PyGithub](https://github.com/PyGithub/PyGithub) for accessing the GitHub REST API. However, this is a bit buggy. Each query returns one item less than what is found when running the same query on GitHub directly. Looking at the project page it looks abandoned, so I don't think fixing this in `PyGithub` is viable. However, there are [various other libraries](https://docs.github.com/en/rest/overview/libraries#python) I can take a look at. 
**Decision:** TBD 
* [ ] The query for "Other Changes" is not really maintainable, as I manually exclude all labels that are used in the other sections. I would be in favour of adding an explicit label for PRs that should go in there.
**Descision:** TBD

Let me know what you think, and if we want to move forward with this or stick to manual creation of the changelog entries!


DanielNoord:
@DudeNr33 This looks very promising! Before delving any deeper, have we considered using existing libraries such as [`towncrier`](https://github.com/twisted/towncrier)? It's used by `pip` and `pytest` for example.

It would require some changes in the way we store changelog entries from `2.15+` onwards, but it's a tried and tested library which we could perhaps benefit from?

doc/exts/pylint_changelog.py
```diff
@@ -0,0 +1,95 @@
...
+        text = f"{issue.title}\n\nPR: `#{issue.number} <https://github.com/PyCQA/pylint/pull/{issue.number}>`_"
+        return nodes.list_item("", *self._parse_as_rst(text).children)
+
+    def _parse_as_rst(self, text: str) -> nodes.document:
+        parser = Parser()
+        components = (Parser,)
+        settings = frontend.OptionParser(components=components).get_default_values()
+        document = new_document("<rst-doc>", settings=settings)
+        parser.parse(text, document)
+        return document
```
DudeNr33, line 86: Figuring out how to do this took hours and a bit of my sanity. 
For a library with "doc" in its name, the documentation of `docutils` is horrible. 😄  Not to mention the usability.

And to cite a YouTube video of a PyGotham talk:
"Sphinx is weird and crazy and old [...], and it sits on top of something that was old before the thing below it was old before Moses was old."

DudeNr33:
> @DudeNr33 This looks very promising! Before delving any deeper, have we considered using existing libraries such as [`towncrier`](https://github.com/twisted/towncrier)? It's used by `pip` and `pytest` for example.
> 
> It would require some changes in the way we store changelog entries from `2.15+` onwards, but it's a tried and tested library which we could perhaps benefit from?

I "know" `towncrier` (i.e. heard of it, but not used it myself yet). If this is an option for us, I will happily take a closer look.
We should get a clear consensus of what requirements and expectations we have for the changelog generation. 
From my point of view:

1. It must be possible to structure the generated changelog into different sections (not just for release versions, but for topics like false negative, false positive etc.).
2. It must be possible to exclude the "old" versions from the generation (pylint <= 2.15.0), as this would only be waste.
3. It should be possible to change the structure from one release to the other, without affecting the changelogs for already released versions.

doc/conf.py
```diff
@@ -297,3 +298,13 @@
...
 linkcheck_ignore = ["https://github.com/PyCQA/pylint/blob/main/pylint/extensions/.*"]
+
+
+# -- Options for pylint_changelog extension ------------------------------------
+
+pylint_changelog_user = "PyCQA"
+pylint_changelog_project = "pylint"
+pylint_changelog_token = os.getenv("GITHUB_TOKEN")
+pylint_changelog_exclude_labels = [
+    "Documentation 📖",
```
Pierre-Sassoulas, line 309: Regarding this: I said in the original issue "If the PR is labelled maintenance, primer or documentation in github then nothing is required". But I also added 'skip news :mute:', because there's some condition to take into account imo. documentation alone should be skipped but what about documentation and bug together or primer and something else ... ? Maybe just one label with clear expectation is better ? We're doing a lot of maintenance/documentation issue that will need to have one more label if we do that though.

I'm htinking about skip news or  maintenance, primer or documentation and only that label righ tnow.

Pierre-Sassoulas (commented):
>    It must be possible to structure the generated changelog into different sections (not just for release versions, but for topics like false negative, false positive etc.).

+1. maybe using the label from the pull request ?

> It must be possible to exclude the "old" versions from the generation (pylint <= 2.15.0), as this would only be waste.

👍 

> It should be possible to change the structure from one release to the other, without affecting the changelogs for already released versions.

👍 Separating the changlog for each minor version in particular

> It must be possible to build the documentation including changelog through CI and readthedocs.

Ideally. But where would we take the changelog entry ? In the commit message ? 

I think we also need to be able to create release notes for a patch version. Right now everything is mixed up in 2.15.0, but what will happen when we maintain it and want to release 2.15.1 ?

doc/conf.py
DudeNr33: I too think that an explicit "skip news" label would be the cleanest way to go.
Same for the "Other changes" section.

DudeNr33:
> We're already handling markdown for `release.md` (there's a dependency handling that), maybe's it's not going to be a problem to include markdown from the PR ?
> 

Yes, it should be possible by using the `myst-parser` that we already depend on. 👍  I did not think about that at first, good point! 😊 
As our custom directive only needs to return a list of `docutils` nodes, the source format is not really relevant.

DanielNoord:
Thanks for the extensive review @DudeNr33.

Do you have a preference yourself? To me it seems that `towncrier` has a higher initial cost but might be beneficial long term. But please correct me if I'm wrong. So that would mean we would need somebody to sponsor/push the initial change, but then be able to rely on `towncrier` to fix our issues?

DudeNr33:
If we can live without "preview" release notes for upcoming patch releases, I would vote for `towncrier`.
I don't think that the initial cost for implementing the `towncrier` approach is higher. With the Sphinx extension of this current PR there would be additional steps, like defining the necessary labels and apply them to all PRs for 2.15.0 and 2.14.2. Plus, there is always the potential for bugs with a homegrown solution. 

Let's vote for it:
* 🚀  `towncrier`
* 🎉  Sphinx extension of this PR (`pylint_changelog`)

Pierre-Sassoulas:
Thank you for the state of the art, very interesting ! 

> If we can live without "preview" release notes for upcoming patch releases, I would vote for towncrier.

It's a nice to have but it wasn't intended, we can live without.

> With the Sphinx extension of this current PR there would be additional steps, like defining the necessary labels and apply them to all PRs for 2.15.0 and 2.14.2.

Can't we use the milestone directly for this ? Or are we talking about 'skip-news 🔇' ? Wouldn't we need that for towncrier too ?

Pierre-Sassoulas:
> For those we would have to define the corresponding labels, and apply them to all PRs that need them.

It would definitely be nice to be able to choose the section using labels 👍 But towncrier do not need that if I understand correctly ?

DudeNr33:
Exactly. `towncrier` works with news fragments. Each fragment has a *type*, and the possible types can be configured through a configuration file. 
When using the `towncrier` CLI, it will also check that only one of the configured types is used. 
So for example, the news fragment for a PR fixing a false positive would be *1234.false_positive*.
I have to check if `towncrier` also already provides a way to prevent users from just manually creating a news fragment with wrong type. If not, it would be no problem to simply write a custom script to check that in the CI.

PR #6880, opened 2022-06-06
Added documentation examples for `nonlocal-and-global`.

mbyrnepr2:
## Type of Changes

## Description

### DIFF
--- a/doc/data/messages/n/nonlocal-and-global/bad.py
+++ b/doc/data/messages/n/nonlocal-and-global/bad.py
@@ -6,3 +6,6 @@ def update_number(number):  # [nonlocal-and-global]
     nonlocal NUMBER
     NUMBER = number
     print(f"New global number is: {NUMBER}")
+    
+
+update_number(24)
--- a/doc/data/messages/n/nonlocal-and-global/bad.py
+++ b/doc/data/messages/n/nonlocal-and-global/bad.py
@@ -6,6 +6,6 @@ def update_number(number):  # [nonlocal-and-global]
     nonlocal NUMBER
     NUMBER = number
     print(f"New global number is: {NUMBER}")
-    
+
 
 update_number(24)
--- a/doc/data/messages/n/nonlocal-and-global/bad.py
+++ b/doc/data/messages/n/nonlocal-and-global/bad.py
@@ -1,0 +1,8 @@
+NUMBER = 42
+
+
+def update_number(number):  # [nonlocal-and-global]
+    global NUMBER
+    nonlocal NUMBER
+    NUMBER = number
+    print(f"New global number is: {NUMBER}")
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (commented):
I'm not sure about the "good" example because I don't know what the problem being solved is, could you explain @mbyrnepr2 ? Maybe not using global nor non-local works ?

mbyrnepr2:
> I'm not sure about the "good" example because I don't know what the problem being solved is, could you explain @mbyrnepr2 ? Maybe not using global nor non-local works ?

The idea was to create a closure to remember the last value and update the global one.
However, since you need to ask then I think the example isn't clear. OK with me to close this

doc/data/messages/n/nonlocal-and-global/bad.py
```diff
@@ -0,0 +1,8 @@
+NUMBER = 42
+
+
+def update_number(number):  # [nonlocal-and-global]
+    global NUMBER
+    nonlocal NUMBER
+    NUMBER = number
+    print(f"New global number is: {NUMBER}")
```
Pierre-Sassoulas, line 8:
```suggestion
    print(f"New global number is: {NUMBER}")
    

update_number(24)
```

doc/data/messages/n/nonlocal-and-global/good.py
```diff
@@ -1 +1,19 @@
...
+        print(f"New global number is: {NUMBER}")
+
+        nonlocal previous
+        print(f"Previous global number was: {previous}")
+        previous = number
+    return update_number
+
+
+update_number = create_number_updater()
+update_number(24)
```
Pierre-Sassoulas, line 19:
```suggestion
NUMBER = 42
NUMBER = 24
```

Pierre-Sassoulas (commented):
No no, it's just that I don't understand the example/message and what the intent leading to the error was. Your example in bad.py is illustrating the problem well. Would this suggestion work for example ?

mbyrnepr2:
> No no, it's just that I don't understand the example/message and what the intent leading to the error was. Your example in bad.py is illustrating the problem well. Would this suggestion work for example ?

I see what you mean. My way of thinking of it was to somehow include usage of both global & nonlocal in the good example also. But that probably isn't necessary. Thanks for simplifying it!

doc/data/messages/n/nonlocal-and-global/good.py
Pierre-Sassoulas: @DanielNoord, do you mind giving your opinion on this ?

DanielNoord: Shouldn't the good example use either `global` or `nonlocal`? The warning is about using both and you needing to decide which it should not be, not remove both of them.

Pierre-Sassoulas (approved):
👍

PR #6885, opened 2022-06-06
[REF] missing-timeout: static without inferred

moylop260:
Backup from https://github.com/PyCQA/pylint/pull/6780 but using static without inferred so this version doesn't need to have installed the packages to work correctly

See @DanielNoord comment https://github.com/PyCQA/pylint/pull/6780#discussion_r890407456


moylop260:
Hi @DanielNoord 

Using inferred way, how are you installing requests and pyserial packages in order to be tested correctly?

I checked if you was installing `numpy` but I didn't find it

PR #6891, opened 2022-06-08
Makes functional tests auto-update works even if the current output is wrong

Pierre-Sassoulas:
## Type of Changes

## Description
Previously if an exception was raised during parsing we re-raised an explanation.
Most of the time I just want to regenerate the output from the current pylint's
output. It's what the exception suggested but it's impossible to do as the error
is also raised when updating, so you had to empty the file.

### DIFF
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -61,3 +61,9 @@ Internal changes
 * ``pylint.testutils.primer`` is now a private API.
 
   Refs #6905
+
+* Fixed an issue where it was impossible to update functional tests output when the existing
+  output was impossible to parse. Instead of raising an error we raise a warning message and
+  let the functional test fail with a default value.
+
+  Closes #6891
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -66,4 +66,4 @@ Internal changes
   output was impossible to parse. Instead of raising an error we raise a warning message and
   let the functional test fail with a default value.
 
-  Closes #6891
+  Refs #6891
--- a/pylint/testutils/output_line.py
+++ b/pylint/testutils/output_line.py
@@ -33,44 +33,6 @@ class MessageTest(NamedTuple):
     """
 
 
-class MalformedOutputLineException(Exception):
-    def __init__(
-        self,
-        row: Sequence[str] | str,
-        exception: Exception,
-    ) -> None:
-        example = "msg-symbolic-name:42:27:MyClass.my_function:The message"
-        other_example = "msg-symbolic-name:7:42::The message"
-        expected = [
-            "symbol",
-            "line",
-            "column",
-            "end_line",
-            "end_column",
-            "MyClass.myFunction, (or '')",
-            "Message",
-            "confidence",
-        ]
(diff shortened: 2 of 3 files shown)


DanielNoord (commented):
I don't really understand the use case.

If the test fails how can we know what the expected output is?

Pierre-Sassoulas:
>I don't really understand the use case.

You're rebasing a branch with change in the functional expected output. You don't  want to fix the merge conflict, you just want to regenerate with the update option. Right now the update option fail if the output contain ``>>>>>``, with a message telling you to use the update option. This adds a warning (with a 3.0 compatible format insteaf of the old ones) instead of an error so the update will work, and the non update will work as if there was nothing in the file instead of a bogus output.

DanielNoord:
Okay, but simply discarding the merge change also works right?

With the new code we only raise an `UserWarning` on actual malformed lines. I don't think that's a good change.

Pierre-Sassoulas:
> Okay, but simply discarding the merge change also works right?

Sure, but it's more work, and the message telling us to use update is wrong.

> With the new code we only raise an UserWarning on actual malformed lines. I don't think that's a good change.

The functional test will also fail because we return  ``cls("", 0, 0, None, None, "", "", "")`` which will never be valid.

DanielNoord (changes requested):
> The functional test will also fail and ask you to update because we return `cls("", 0, 0, None, None, "", "", "")` which will never be valid.

👍 

But let's add a changelog entry in the `Internal changes` section.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -61,3 +61,8 @@ Internal changes
 * ``pylint.testutils.primer`` is now a private API.
 
   Refs #6905
+
+* Fixed an issue where it was impossible to regenerate functional tests output when they existing
```
DanielNoord, line 65:
```suggestion
* Fixed an issue where it was impossible to regenerate functional tests output when the existing
```

DanielNoord (approved):

doc/whatsnew/2/2.15/index.rst
```diff
@@ -61,3 +61,9 @@ Internal changes
 * ``pylint.testutils.primer`` is now a private API.
 
   Refs #6905
+
+* Fixed an issue where it was impossible to update functional tests output when the existing
+  output was impossible to parse. Instead of raising an error we raise a warning message and
+  let the functional test fail with a default value.
+
+  Closes #6891
```
DanielNoord, line 69:
```suggestion
```

DanielNoord (changes requested):
We never put PR numbers in the changelog right? If there isn't an issue we just leave that line empty.

doc/whatsnew/2/2.15/index.rst
Pierre-Sassoulas: I think if there is no issue then we need to put the PR or there's no way to easily see the discussion around the PR otherwise. (You would have to git blame the changelog text...)

DanielNoord: Let's use `ref` then. Closes feels weird.

Pierre-Sassoulas:
```suggestion
  Refs #6891
```

PR #6761, opened 2022-05-30
Remove ``home-assistant`` from primer

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Let's see if this helps for now?

PR #6897, opened 2022-06-08
[primer] stop overwriting `--rcfile`

jacobtylerwalls:
## Type of Changes

## Description
~Move everything into the new primer and increase the timeout on the `main` job. (3.7 is about 2x as slow as 3.10)
Follow up to #6761 and #6753.~

Never mind, we will have to discuss that separately.

### DIFF
--- a/tests/primer/primer_tool.py
+++ b/tests/primer/primer_tool.py
@@ -292,8 +292,6 @@ class Primer:
         # TODO: Find a way to allow cyclic-import and compare output correctly
         disables = ["--disable=duplicate-code,cyclic-import"]
         arguments = data.pylint_args + enables + disables
-        if data.pylintrc_relpath:
-            arguments += [f"--rcfile={data.pylintrc_relpath}"]
         output = StringIO()
         reporter = JSONReporter(output)
         Run(arguments, reporter=reporter, exit=False)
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
The error of the CI is strange, it seems there's a mix up between django and black

DanielNoord:
@jacobtylerwalls Did you test this on your local branch to see the time differences?

jacobtylerwalls:
Good point, I hadn't. I just did. I removed `keras` to try to get the job under 45m. With just `django` and `music21`, added, the 3.10 job finishes in 30 minutes. 3.7 seems to take 1.5x as long, which just barely exceeds 45m (hard to tell, but that's my guess.)

Neither of the two projects I'm adding support 3.7 anymore. I'm wondering if we even need that job?

jacobtylerwalls:
The changes in `astroid` are to be expected because we're using the correct `--rcfile` now.

I plan to merge this tonight based on Pierre's approval earlier of the cleanups.

I'll move discussion about adding packages to #5359, since I think it's clear we need to look at parallelization instead of just adding packages to a single run. The 3.10 (the faster one!) errored out on CI with an out of memory error, probably trying to deal with the massive list of messages.

PR #6901, opened 2022-06-09
Add no-dict-direct-access checker

pmiddend:
## Type of Changes

## Description

I noticed that I rarely use direct dictionary access, as in `myfunction(foo["bar"])`, since `foo` might not contain `"bar"`. I usually go for `foo.get("bar")` (or some other default value) to be safe. I wanted a checker that checks my code for these direct dictionary subscripts and reports them. I found a few bugs that way already and thought this might fit into pylint in general.

The following will produce a message:

```python
mydict = { "foo": 3 }
print(mydict["foo"])
```

Whereas this will not:

```python
mydict = { "foo": 3 }
print(mydict.get("foo"))
```

This is also fine:

```python
mydict = {}
mydict["foo"] = 3
```

Since assignment is safe per se.

### DIFF
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -15,7 +15,6 @@ Summary -- Release highlights
 New checkers
 ============
 
-
 Removed checkers
 ================
 
@@ -23,6 +22,10 @@ Removed checkers
 Extensions
 ==========
 
+* ``NoDictSubscriptChecker``
+
+    * Added optional extension ``no-dict-subscript`` to emit messages when the ``[]`` operator is used
+      to access a dictionary value.
 
 False positives fixed
 =====================
--- a/pylint/checkers/misc.py
+++ b/pylint/checkers/misc.py
@@ -8,11 +8,11 @@ from __future__ import annotations
 
 import re
 import tokenize
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
-from astroid import nodes
+from astroid import nodes, Dict, Assign
 
-from pylint.checkers import BaseRawFileChecker, BaseTokenChecker
+from pylint.checkers import BaseRawFileChecker, BaseTokenChecker, BaseChecker, utils
 from pylint.typing import ManagedMessage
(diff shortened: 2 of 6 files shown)


doc/whatsnew/2/2.14/full.rst
```diff
@@ -5,6 +5,8 @@ What's New in Pylint 2.14.2?
 ----------------------------
 Release date: TBA
 
+* Added new checker ``no-dict-direct-access``.
+
```
jacobtylerwalls, line 9: New features go in the next minor version, so this should move to 2.15.rst

jacobtylerwalls (commented):
Thanks for your interest in contributing this to pylint. I think at the least this should be an optional extension. See the examples in `pylint/checkers/extensions`. So that will solve the bulk of the test failures.

You might wait to hear positive feedback from others before continuing, as this is the sort of thing that might be too strict even for an optional extension.

Pierre-Sassoulas:
I don't have time for a full review right now but as Jacob said it needs to be an extension. We have really opinionated extensions already so it's ok. (While used, for example).

pmiddend:
Thanks @Pierre-Sassoulas for the vote of confidence and @jacobtylerwalls for the general hints! I'll move the check to an extension and am absolutely open to more opinions on the matter. Worst case, I'll release it on pypi or something and it'll be more unofficial. :)

pmiddend:
I have fixed a few issues and put the checker into an extension.

Pierre-Sassoulas (changes requested):
I think this need some functional tests, in particular what happen when the code is:
```python
d = {}
try:
    d["a"] = 1
except KeyError:
    print("The 'a' key does not exists in d")
```

jacobtylerwalls:
Thanks for taking the time to prepare a pull request. I agree, though, with Pierre that this is probably better as a plugin maintained outside the pylint project. If we hear more requests, I'll point folks here and revisit the issue.

pmiddend:
@jacobtylerwalls I understand, and have begun migrating this to a pylint plugin which will be released on pypi. Thanks for the pleasent review process :)

PR #6905, opened 2022-06-10
[refactor] Create a primer package in 'pylint.testutils'

Pierre-Sassoulas:
## Type of Changes

## Description

First step to be able to add more testable things related to primer here. I'm wondering if an external package called ``pylint-testutils`` wouldn't be better at this point. We don't need to ship test helper in pylint, but pytest do not like import from test file contained in ``tests``. We might remove ``pylint/testutils`` in 3.0 and always use bleeding edge of the new package in pylint itself.


pylint/testutils/primer/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+__all__ = ["PackageToLint", "PRIMER_DIRECTORY_PATH"]
```
Pierre-Sassoulas, line 5: The path to the primer directory to not need to be modified because it's relative (``PRIMER_DIRECTORY_PATH = Path("tests") / ".pylint_primer_tests"``)

DanielNoord:
I have been thinking about this as well, but let's not do this as long as the `primer` isn't where we want this to be (parallelisation etc.). 

Just giving the list of packages we prime also doesn't make a lot of sense: if anybody would ever want to re-use our primer the first thing they'll likely change is the packages to prime.

Pierre-Sassoulas:
I think temporarily adding primer's helper in pylint.testutils to be able to test them properly is a decent compromise until they mature and we cleanup, what do you think ? I want to test and modify the primer comment next.

DanielNoord (approved):
To preserve backwards compatibility we might want to keep the old file and re-import there with a warning. However, I don't think there will be many packages that use this so let's not do that 😄

Pierre-Sassoulas:
> let's not do that

Yeah agree completely 👍 Package maintainers that discovered our primer's helper are big boys and girls, they can change their export. The cleanup we'll have to do for pylint 3.0 is already beyond epic as is 😄

PR #6910, opened 2022-06-10
Avoid `used-before-assignment` in except handlers employing type annotations

jacobtylerwalls:
## Type of Changes

## Description
Noticed this false positive while working on #6677.


Pierre-Sassoulas (approved):
That's a clean one 👍

PR #6911, opened 2022-06-10
Emit `using-constant-test` when testing truth value of a call returning a generator

### LINKED ISSUE #6909: Raise a warning ``using-constant-test`` when casting a generator to a boolean as they are always truthey
### Current problem

Suppose library code:

```python
SOME_DATA = {1:2, 3:4}

def some_function() -> Collection[str]:
    return SOME_DATA.items()
```

and code that use the function:

```python
if some_function():
    do_something()
```

At some point, authors decide to change signature in this way:

```python
def some_function() -> Iterable[str]:
    return ((key, value) for key, value in SOME_DATA.items())
```

Obviously (actually not) the end-user code breaks.

### Desired solution

I propose considering as error all cases where a generator implicitly interpreted as boolean. It will always return `True` even if a generator is exhausted, so this is useless. Error text should contain "to check if value is None use `is not None`".

jacobtylerwalls:
## Type of Changes

## Description
We already emitted a message for testing a generator directly, just not calls to functions that _only_ return generators.

Also update confidence level.

Closes #6909


pylint/checkers/base/basic_checker.py
```diff
@@ -317,6 +317,17 @@ def _check_using_constant_test(
...
         if not isinstance(test, except_nodes):
             inferred = utils.safe_infer(test)
+        # Emit if the call is to a function that only returns GeneratorExp (always truthy)
+        elif isinstance(test, nodes.Call):
+            inferred_call = utils.safe_infer(test.func)
+            if isinstance(inferred_call, nodes.FunctionDef):
+                return_nodes = list(inferred_call._get_return_nodes_skip_functions())
+                if return_nodes and all(
+                    isinstance(n.value, nodes.GeneratorExp) for n in return_nodes
+                ):
```
Pierre-Sassoulas, line 327:
```suggestion
                if all(
                    isinstance(n.value, nodes.GeneratorExp)
                    for n in inferred_call._get_return_nodes_skip_functions()
                ):
```
I guess you had a reason to generate the full list each time (it can be None instead of being simply empty ?), but wouldn't it be faster like this because we can shortcut when one of the value is not a generator expression ?

Pierre-Sassoulas (commented):
Looks pretty good already !

pylint/checkers/base/basic_checker.py
jacobtylerwalls: Yah I wanted to avoid two loops, but I need to return false if empty, and `all` returns True.

I could create some booleans and track them during a loop.

jacobtylerwalls: I always waffle on what's the clearest way to do this pattern.

Pierre-Sassoulas: I'm a little lost there. I'm under the impression that the design of ``_get_return_nodes_skip_functions`` prevents the use of a generator because we have to do a special case for when it's empty, is that right ?

pylint/checkers/base/basic_checker.py
```diff
@@ -317,6 +317,20 @@ def _check_using_constant_test(
...
+        # Emit if calling a function that only returns GeneratorExp (always tests True)
+        elif isinstance(test, nodes.Call):
+            inferred_call = utils.safe_infer(test.func)
+            if isinstance(inferred_call, nodes.FunctionDef):
+                all_returns_were_generator = None
+                for return_node in inferred_call._get_return_nodes_skip_functions():
+                    if not isinstance(return_node.value, nodes.GeneratorExp):
+                        all_returns_were_generator = False
+                        break
+                    all_returns_were_generator = True
```
Pierre-Sassoulas, line 329:
```suggestion
                if any(
                    not isinstance(return_node.value, nodes.GeneratorExp):
                    for return_node in inferred_call._get_return_nodes_skip_functions()
                 ):
                    break
```
Am I missing something ?

pylint/checkers/base/basic_checker.py
jacobtylerwalls: My bad for not taking the time to explain fully. `_get_return_nodes_skip_functions` returns a generator. I just can't do `all()` because that returns True for empty iterables, and I don't want to emit a message if the function doesn't have return statements.

pylint/checkers/base/basic_checker.py
jacobtylerwalls: Yes, I don't love this about Python.
```python
>>> all([])
True
```

jacobtylerwalls: oh hang on... let me see what you did...

jacobtylerwalls: I guess I need to see a suggestion in a fuller context to see if it would work. How does the suggestion track the case were the generator is empty?

Pierre-Sassoulas: (I'm learning of this at the same time as you but) I guess the default value for an empty list can be defined by using any or all (all will be true whatever the condition, any will be false whatever the condition):
```
>>> all(i == 0 for i in range(0))
True
>>> all(i != 0 for i in range(0))
True
>>> any(i == 0 for i in range(0))
False
>>> any(i != 0 for i in range(0))
False
```

jacobtylerwalls: I'd love for this to be true, but taking a step back:
1. first try: I want to use `all()`
2. i can't unless I use two iterators or use boolean flags, because `all` is True when empty, and I need False
3. so I rewrite `all` as `not any(not x)`
4. but I'm back to the same problem, since I've negated the expression, I need True for empty containers now
5. but `not any` gives me False for empty containers 🤕 
6. Plus "any" requires a complete scan, so we lose the short-circuiting 📉

Pierre-Sassoulas: Hmm, let me try that offline and report back 😄 

> Plus "any" requires a complete scan, so we lose the short-circuiting

In ``all(isinstance(v, nodes.GeneratorExp) for v in x)`` we shortcut the first time ``isinstance(v, nodes.GeneratorExp)`` is False because all values needs to be true. In ``any(not isinstance(v, nodes.GeneratorExp) for v in x)`` we shortcut the first time ``not isinstance(v, nodes.GeneratorExp) `` is True because any single true value will makes any True. So I think we shortcut for the same reason i.e. ``isinstance(v, nodes.GeneratorExp)`` is False or ``not isinstance(v, nodes.GeneratorExp)`` is True which are the same condition.

Pierre-Sassoulas: Ha of course you're right ``not any gives me False for empty containers``. I guess the python convention force us to handle the empty container ourselves in this case and using a for loop with break is the best we can do in term of performance.

pylint/checkers/base/basic_checker.py
```diff
@@ -317,6 +317,20 @@ def _check_using_constant_test(
         emit = isinstance(test, (nodes.Const,) + structs + const_nodes)
         if not isinstance(test, except_nodes):
             inferred = utils.safe_infer(test)
+        # Emit if calling a function that only returns GeneratorExp (always tests True)
+        elif isinstance(test, nodes.Call):
+            inferred_call = utils.safe_infer(test.func)
+            if isinstance(inferred_call, nodes.FunctionDef):
+                all_returns_were_generator = None
```
Pierre-Sassoulas, line 324:
```suggestion
                # Can't use any or all for the following condition as all([]) is True
                # which is not what we want here.
                all_returns_were_generator = None
```

Pierre-Sassoulas (approved):

DanielNoord (approved):

pylint/checkers/base/basic_checker.py
jacobtylerwalls: Ah yes good point about the short circuiting!

PR #6914, opened 2022-06-11
Fixed misinterpreted modulo sign for ``consider-using-f-string``

### LINKED ISSUE #6689: False positive for `consider-using-f-string` when using modulo operator on a string
### Bug description

```python
#!/usr/bin/python3

"""
Test unittest 1.
"""

import unittest

# pylint: disable=pointless-statement
class ModuloFailTestCase(unittest.TestCase):
    """Test modulo operator failure"""

    def test_int_str_modulo_fail(self):
        """Test int and str modulo operator failure"""
        with self.assertRaises(TypeError):
            1 % "str"
```

### Command used

```shell
pylint unittestprog1.py
```

### Pylint output

```shell
************* Module unittestprog1
unittestprog1.py:16:16: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
```

### Expected behavior

Do not flag a modulo (`%`) operation on a string with the `consider-using-f-string` lint.

(I discovered this issue while working on converting the justbytes module [ https://github.com/mulkieran/justbytes ] to formatted strings:)

```python
class ModTestCase(unittest.TestCase):
    """Test mod."""

    def test_exceptions(self):
        """Test that exceptions are thrown."""
        # pylint: disable=expression-not-assigned
        with self.assertRaises(RangeNonsensicalBinOpError):
            2048 % Range(12, B)
        with self.assertRaises(RangeNonsensicalBinOpError):
            Range(12) % "str"
        with self.assertRaises(RangeNonsensicalBinOpValueError):
            Range(12) % Range(0)
        with self.assertRaises(RangeNonsensicalBinOpValueError):
            Range(12) % 0
        with self.assertRaises(RangeNonsensicalBinOpError):
            Range(12) % Decimal("NaN")
```

### Pylint version

```shell
pylint 2.13.9
astroid 2.11.5
Python 3.10.4 (main, Mar 25 2022, 00:00:00) [GCC 12.0.1 20220308 (Red Hat 12.0.1-0)]

(This also occurs in the Fedora 36 package "pylint-2.13.7-1.fc36.noarch" .)
```

### OS / Environment

Fedora 36

d1gl3:
## Type of Changes

## Description

* Fixed a false positive in ``consider-using-f-string`` if the left side of a ``%`` is not a string.

  Closes #6689


DanielNoord (approved):

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -394,6 +394,10 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
             if "\\" in node.parent.right.as_string():
                 return
 
+            # If % applied to another type than str, it's modulo and can't be replaced by formatting
+            if not isinstance(node.parent.left.value, str):
```
jacobtylerwalls, line 398: It's not safe to assume node.parent.left has `value`, it could be a tuple.

jacobtylerwalls (changes requested):
Thanks for the fix! It needs just a bit of hardening against some possible crashes:

```python
(1, 2) % 'garbage'
```

pylint crashed with a ``AstroidError`` and with the following stacktrace:
```
  File "/Users/.../pylint/pylint/checkers/refactoring/recommendation_checker.py", line 398, in _detect_replacable_format_call
    if not isinstance(node.parent.left.value, str):
  File "/Users/.../astroid/astroid/bases.py", line 119, in __getattr__
    return getattr(self._proxied, name)
AttributeError: 'ClassDef' object has no attribute 'value'
```

pylint/checkers/refactoring/recommendation_checker.py
d1gl3: Ok, I check now first if node.parent.left has a value attribute.

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -394,6 +394,12 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
             if "\\" in node.parent.right.as_string():
                 return
 
+            # If % applied to another type than str, it's modulo and can't be replaced by formatting
+            if not hasattr(node.parent.left, "value") or not isinstance(
```
jacobtylerwalls, line 398: Should we infer the left-hand side (e.g. like the right is inferred just below) to avoid creating the false negatives reported by the primer?

Pierre-Sassoulas: Is this really false negative ? It's really hard to know if a string is supposed to be a template or not. 

```python
templates = {
    "config": "blabla %s blabla"
}

if some_condition:
    print(templates["config"} % "a string")
else:
    print(templates["config"} % "another string")
```
Is often seen and I don't think we should raise here. It's the reason why we do not have a check for missing leading f in supposed f-string (too much false positive).

jacobtylerwalls: Good point!

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
Thank you @d1gl3, great fix !

pylint/checkers/refactoring/recommendation_checker.py
DanielNoord, line 398: In the future I think it is better to test for the class of `left` with an `isinstance`. That's a little more robust and explicit about what we are checking for here.

PR #6925, opened 2022-06-12
Avoid raising `undefined-loop-variable` twice on same line

jacobtylerwalls:
## Type of Changes

## Description

See https://github.com/PyCQA/pylint/pull/6923#discussion_r895134495 for report.

This call has been here since the initial commit, but nothing fails when I remove it. Let's see what the primer says.

### DIFF
--- a/doc/whatsnew/2/2.14/full.rst
+++ b/doc/whatsnew/2/2.14/full.rst
@@ -5,6 +5,8 @@ What's New in Pylint 2.14.2?
 ----------------------------
 Release date: TBA
 
+* Avoided raising an identical ``undefined-loop-variable`` message twice on the same line.
+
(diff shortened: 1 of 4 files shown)


jacobtylerwalls:
I spot-checked the messages from sentry, and found that they were duplicates that are now just emitted once.

DanielNoord (approved):
While checking if `self._loopvar_name(node)` is still being used I noticed that the other use is guarded by `if self._is_undefined_loop_variable_enabled`.

This breaks `useless-suppression`. We could fix that in this PR as well or do a follow-up.

Pierre-Sassoulas (approved):
👌

Pierre-Sassoulas:
Nice catch @DanielNoord. Let's do a follow-up as it's another issue.

PR #6929, opened 2022-06-12
Add `useless-else-on-loop` example

harupy:
## Type of Changes

## Description

Refs #5953

Closes #XXXX

### DIFF
--- a/doc/data/messages/u/useless-else-on-loop/bad.py
+++ b/doc/data/messages/u/useless-else-on-loop/bad.py
@@ -2,4 +2,5 @@ def find_even_number(lst):
     for x in lst:
         if x % 2 == 0:
             return x
-    print("Did not find an even number")
+    else:  # [useless-else-on-loop]
+        print("Did not find an even number")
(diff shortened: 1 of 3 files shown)


doc/data/messages/u/useless-else-on-loop/bad.py
```diff
@@ -0,0 +1,6 @@
+def find_even_number(lst):
+    for x in lst:
```
Pierre-Sassoulas, line 2:
```suggestion
def find_even_number(numbers):
    for x in numbers:
```

doc/data/messages/u/useless-else-on-loop/good.py
```diff
@@ -1 +1,5 @@
-# This is a placeholder for correct code for this message.
+def find_even_number(lst):
+    for x in lst:
```
Pierre-Sassoulas, line 2:
```suggestion
def find_even_number(numbers):
    for x in numbers:
```

Pierre-Sassoulas (approved):

PR #6930, opened 2022-06-12
Add `bad-inline-option` example

harupy:
## Type of Changes

## Description

Refs #5953

Closes #XXXX

### DIFF
--- a/doc/data/messages/b/bad-inline-option/bad.py
+++ b/doc/data/messages/b/bad-inline-option/bad.py
@@ -1,2 +1,2 @@
 # 2:[bad-inline-option]
-long_string = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"  # pylint: disable line-too-long
+# pylint: disable line-too-long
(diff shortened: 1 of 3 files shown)


doc/data/messages/b/bad-inline-option/bad.py
```diff
@@ -0,0 +1,2 @@
+# 2:[bad-inline-option]
+long_string = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"  # pylint: disable line-too-long
```
Pierre-Sassoulas, line 2:
```suggestion
# pylint: disable line-too-long
```

doc/data/messages/b/bad-inline-option/good.py
```diff
@@ -1 +1 @@
-# This is a placeholder for correct code for this message.
+long_string = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"  # pylint: disable=line-too-long
```
Pierre-Sassoulas, line 1:
```suggestion
# pylint: disable=line-too-long
```

Pierre-Sassoulas (commented):
I think we don't need to have an actual line too long in the example. But If you want to keep "offending code", we could find a message that is less "noisy", maybe invalid-name ?

Pierre-Sassoulas (approved):

Issue #6931, opened 2022-06-12
Incompatible config file options for naming hints in pylint 2.14.0


andy-maier:
### Bug description

Pylint 2.14.0 rejects config files that worked before.

The problem with that is that we test our projects with many different Python versions, usually 2.7 and 3.5 and higher. Some minimum version requirements for pylint we define plus the supported Python versions of pylint itself usually result in the following Pylint versions that are used:
* Python 2.7: pylint>=1.6.4,<2.0.0
* Python 3.5: pylint>=2.5.2,<2.7.0
* Python 3.6: pylint>=2.10.0,<2.14.0
* Python 3.7: pylint>=2.10.0
* Python 3.8: pylint>=2.10.0
* Python 3.9: pylint>=2.10.0

Now pylint 2.14.0 requires to change the config file in a way that is no longer supported by earlier pylint versions.

Besides pinning pylint to <2.14.0, the other solution would be to start maintaining different versions of pylint config files, one for pylint <2.14, and one for pylint >=2.14 (plus to increase the minimum version of pylint to >=2.14.0 for Python >=3.7).

We are not really happy about the second option.

More specifically about the errors we get with pylint 2.14.1 and the older config files:

* Several occurrences of `E0015: Unrecognized option found` on the config file:
  - optimize-ast, files-output, bad-functions, no-space-check - I could live with removing those in order to get to a compatible config file.
  - function-name-hint, variable-name-hint, const-name-hint, attr-name-hint, argument-name-hint, class-attribute-name-hint, inlinevar-name-hint, class-name-hint, module-name-hint, method-name-hint - This is where the incompatibility is: The old option names are rejected by pylint 2.14, and the new option names are rejected by older pylint versions.
* One occurrence of `W0012: Unknown option value for '--disable'` on the config file:
  - locally-enabled - I can live with removing that in order to get to a compatible config file.
* Several occurrences of `W0012: Unknown option value for 'disable'` on source code:
  - no-self-use - I can live with removing that in order to get to a compatible config file.

An additional issue is that I could not find these incompatibilities in the change log of 2.14.0. Specifically, I did not find anything in the change log about how to transform the old naming hint options to the new ones, or even what the names of the new ones are.

### Configuration

```ini
# .pylintrc file from https://github.com/zhmcclient/zhmc-prometheus-exporter project:

#
# PyLint configuration file for the zhmc-prometheus-exporter project
#
# Originally generated by pylint 1.5.4, now manually maintained.
#


[...]

PR #6932, opened 2022-06-12
Mark `no-self-use` as moved to extensions

jacobtylerwalls:
## Type of Changes

## Description

See https://github.com/PyCQA/pylint/issues/6931#issuecomment-1153226753

**Before**
Command line:1:0: W0012: Unknown option value for '--enable', expected a valid pylint message and got 'no-self-use' (unknown-option-value)
**After**
Command line:1:0: R0022: Useless option value for '--enable', 'no-self-use' was moved to an optional extension, see https://pylint.pycqa.org/en/latest/whatsnew/2/2.14/summary.html#removed-checkers. (useless-option-value)

### DIFF
--- a/pylint/message/_deleted_message_ids.py
+++ b/pylint/message/_deleted_message_ids.py
@@ -123,7 +123,9 @@ DELETED_MESSAGES_IDS = {
     "https://github.com/PyCQA/pylint/pull/6421": [
         DeletedMessage("W0111", "assign-to-new-keyword"),
     ],
-    "https://github.com/PyCQA/pylint/pull/6448. It can be enabled as an optional extension": [
+}
+MOVED_TO_EXTENSIONS = {
+    "https://pylint.pycqa.org/en/latest/whatsnew/2/2.14/summary.html#removed-checkers": [
         DeletedMessage("R0201", "no-self-use")
     ],
 }
--- a/pylint/message/_deleted_message_ids.py
+++ b/pylint/message/_deleted_message_ids.py
@@ -123,6 +123,9 @@ DELETED_MESSAGES_IDS = {
     "https://github.com/PyCQA/pylint/pull/6421": [
         DeletedMessage("W0111", "assign-to-new-keyword"),
     ],
+    "https://github.com/PyCQA/pylint/pull/6448. It can be enabled as an optional extension": [
+        DeletedMessage("R0201", "no-self-use")
+    ],
 }
 
 
--- a/pylint/message/_deleted_message_ids.py
+++ b/pylint/message/_deleted_message_ids.py
@@ -153,3 +153,27 @@ def is_deleted_msgid(msgid: str) -> str | None:
             ):
                 return explanation
     return None
+
+
+@lru_cache(maxsize=None)
+def is_moved_symbol(symbol: str) -> str | None:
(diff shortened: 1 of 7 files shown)


pylint/message/_deleted_message_ids.py
```diff
@@ -123,6 +123,9 @@ class DeletedMessage(NamedTuple):
     "https://github.com/PyCQA/pylint/pull/6421": [
         DeletedMessage("W0111", "assign-to-new-keyword"),
     ],
+    "https://github.com/PyCQA/pylint/pull/6448. It can be enabled as an optional extension": [
+        DeletedMessage("R0201", "no-self-use")
+    ],
 }
```
Pierre-Sassoulas, line 129:
```suggestion
}
MOVED_TO_EXTENSIONS = {
    "https://pylint.pycqa.org/en/latest/whatsnew/2/2.14/summary.html#removed-checkers": [
        DeletedMessage("R0201", "no-self-use")
    ],
}
```

Pierre-Sassoulas (changes requested):
Thank you for the quick reaction ! 

``no-self-use`` was not deleted, it's moved to an extension, I think we need to separate the two cases (I plan to document deleted messages for #6670, so there will be an issue if a "deleted messages" contains extensions message that are documented elsewhere). This can't be merged with only the suggestion I made, we also need to check the use so we can merge both dict when applicable or use one or the other for other cases (we'll need to think about it for each use I think).

pylint/message/message_id_store.py
```diff
@@ -128,20 +134,27 @@ def get_active_msgids(self, msgid_or_symbol: str) -> list[str]:
...
         deletion_reason = None
+        moved_reason = None
         if msgid_or_symbol[1:].isdigit():
             # Only msgid can have a digit as second letter
             msgid = msgid_or_symbol.upper()
             symbol = self.__msgid_to_symbol.get(msgid)
             if not symbol:
                 deletion_reason = is_deleted_msgid(msgid)
+            if deletion_reason is None:
+                moved_reason = is_moved_msgid(msgid)
```
Pierre-Sassoulas, line 145:
```suggestion
                if deletion_reason is None:
                    moved_reason = is_moved_msgid(msgid)
```
Otherwise we're going to run ``is_moved_msgid`` even if ``symbol``is defined (which is all the time).

pylint/message/message_id_store.py
```diff
@@ -128,20 +134,27 @@ def get_active_msgids(self, msgid_or_symbol: str) -> list[str]:
...
         else:
             symbol = msgid_or_symbol
             msgid = self.__symbol_to_msgid.get(msgid_or_symbol)
             if not msgid:
                 deletion_reason = is_deleted_symbol(symbol)
+            if deletion_reason is None:
+                moved_reason = is_moved_symbol(symbol)
```
Pierre-Sassoulas, line 152:
```suggestion
                if deletion_reason is None:
                    moved_reason = is_moved_symbol(symbol)
```
Ditto

Pierre-Sassoulas (commented):
👍 extending the MessageIdStore handling., the good thing is it work for plugin creator too without them having to know about this constant we just added.

Pierre-Sassoulas (approved):
Great 👌

PR #6944, opened 2022-06-13
Mix incorrect parsing of multi-line options in ``ini`` files

### LINKED ISSUE #6888: init-hook no longer supports multi-line hook
### Bug description

Up to version 2.13.9 multi-line init-hook configurations like the following were supported:

```
init-hook=
    try: import pylint_venv
    except ImportError: pass
    else: pylint_venv.inithook()
```

From version 2.14.0 on this no longer works. The error looks like the lines are joined into a single line:

```
❯ pylint test.py
  File "/Users/jgosmann/Library/Python/3.8/bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
  File "/Users/jgosmann/Library/Python/3.8/lib/python/site-packages/pylint/__init__.py", line 25, in run_pylint
  File "/Users/jgosmann/Library/Python/3.8/lib/python/site-packages/pylint/lint/run.py", line 151, in __init__
    args = _config_initialization(
  File "/Users/jgosmann/Library/Python/3.8/lib/python/site-packages/pylint/config/config_initialization.py", line 48, in _config_initialization
    exec(utils._unquote(config_data["init-hook"]))  # pylint: disable=exec-used
  File "<string>", line 1
    try: import pylint_venvexcept ImportError: passelse: pylint_venv.inithook()
                                  ^
SyntaxError: invalid syntax
```

### Configuration

```ini
init-hook=
    try: import pylint_venv
    except ImportError: pass
    else: pylint_venv.inithook()
```
```


### Command used

```shell
pylint test.py
```

### Pylint output

```shell
  File "/Users/jgosmann/Library/Python/3.8/bin/pylint", line 8, in <module>
    sys.exit(run_pylint())
  File "/Users/jgosmann/Library/Python/3.8/lib/python/site-packages/pylint/__init__.py", line 25, in run_pylint
  File "/Users/jgosmann/Library/Python/3.8/lib/python/site-packages/pylint/lint/run.py", line 151, in __init__
    args = _config_initialization(
  File "/Users/jgosmann/Library/Python/3.8/lib/python/site-packages/pylint/config/config_initialization.py", line 48, in _config_initialization
    exec(utils._unquote(config_data["init-hook"]))  # pylint: disable=exec-used
  File "<string>", line 1
    try: import pylint_venvexcept ImportError: passelse: pylint_venv.inithook()
                                  ^
SyntaxError: invalid syntax
```

### Expected behavior

The init-hook runs without a syntax error.

### Pylint version

```shell
pylint 2.14.0 (and 2.14.1)
astroid 2.11.5
Python 3.8.13 (default, May 10 2022, 11:26:18)
[Clang 13.0.0 (clang-1300.0.29.3)]
```

### OS / Environment

_No response_

### Additional dependencies

pylint-venv==2.2.0

DanielNoord:

## Type of Changes

## Description

Closes #6888.

We could also use an `if` statement to check whether the option is `init-hook` and and then not strip the newlines. But that introduce additional overhead, while this makes pylint faster. The fact that the rest of the test suite passes makes me think that this will be fine. We strip most options of whitespaces in other places anyway.

If this regresses after `2.14.2` we can simply add the `if` statement as desired.

### DIFF
--- a/tests/config/functional/ini/pylintrc_with_multi_line_init_hook.ini
+++ b/tests/config/functional/ini/pylintrc_with_multi_line_init_hook.ini
@@ -1,0 +1,6 @@
+# Reported in https://github.com/PyCQA/pylint/issues/6888
+[MASTER]
+init-hook=
+    try: import pylint_venv
+    except ImportError: pass
+    else: pylint_venv.inithook()
--- a/doc/whatsnew/2/2.14/full.rst
+++ b/doc/whatsnew/2/2.14/full.rst
@@ -21,6 +21,10 @@ Release date: TBA
 
   Closes #6818
 
+* Fixed an issue with multi-line ``init-hook`` options which did not record the line endings.
+
+  Closes #6888
+
 * Fixed a false positive for ``used-before-assignment`` when a try block returns
   but an except handler defines a name via type annotation.
 
--- a/pylint/config/config_file_parser.py
+++ b/pylint/config/config_file_parser.py
@@ -56,7 +56,6 @@ class _ConfigurationFileParser:
                 else:
                     continue
             for opt, value in parser[section].items():
-                value = value.replace("\n", "")
                 config_content[opt] = value
                 options += [f"--{opt}", value]
         return config_content, options
--- a/pylint/config/config_initialization.py
+++ b/pylint/config/config_initialization.py
@@ -45,7 +45,7 @@ def _config_initialization(
 
     # Run init hook, if present, before loading plugins
     if "init-hook" in config_data:
-        exec(utils._unquote(config_data["init-hook"]))  # pylint: disable=exec-used
+        exec(config_data["init-hook"])  # pylint: disable=exec-used
 
     # Load plugins if specified in the config file
     if "load-plugins" in config_data:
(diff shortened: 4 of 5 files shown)


tests/config/functional/ini/pylintrc_with_multi_line_init_hook.ini
```diff
@@ -0,0 +1,6 @@
+# Reported in https://github.com/PyCQA/pylint/issues/6888
```
Pierre-Sassoulas, line 1: Shouldn't the result.json change ? Probably not as init_hook are treated as a special case before the config is instanced, so we should probably create a unit test with multiple print so we can actually check that the code of each line was executed.

Pierre-Sassoulas (changes requested):
I think this king of option is impossible to test with the config functional test framework.

tests/config/functional/ini/pylintrc_with_multi_line_init_hook.ini
DanielNoord: This raises a `SyntaxError` and fails the test without the changes in the PR. So it is actually a test against the reported issue: multi-line init hooks became syntactically incorrect code.

Pierre-Sassoulas (approved):

PR #6949, opened 2022-06-14
Fix false positive for ``useless-super-delegation`` for variadics

### LINKED ISSUE #2270: False positive on useless-super-delegation with *args
### Steps to reproduce

```python
class Super:
    def __init__(self, *args):
        self.args = args

class Sub(Super):
    def __init__(self, a, b):
        super().__init__(a, b)
```

### Current behavior

The following message is issued:
```
6:4: W0235: Useless super delegation in method '__init__' (useless-super-delegation)
```

### Expected behavior

I think no message should be issued, since the delegation is not useless: it forces the subclass to always take exactly 2 arguments, while the superclass can take any number of arguments.

This pattern occurs when defining a custom exception class, since the `Exception` class takes a variable number of arguments, but your custom exception type may have mandatory positional arguments.

### pylint --version output

```
pylint 2.0.0.dev2
astroid 2.0.0.dev4
Python 3.4.6 (default, Mar 22 2017, 12:26:13) [GCC]
```

DanielNoord:

## Type of Changes

## Description

Closes #2270.

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1276,6 +1276,14 @@ a metaclass class method.",
         args = _signature_from_call(call)
 
         if meth_node is not None:
+            # Detect if the super method uses varargs or kwargs and the function doesn't
+            if (
+                meth_node.args.vararg
+                and not function.args.vararg
+                or meth_node.args.kwarg
+                and not function.args.kwarg
+            ):
+                return
 
             def form_annotations(arguments):
                 annotations = chain(
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1276,7 +1276,7 @@ a metaclass class method.",
         args = _signature_from_call(call)
 
         if meth_node is not None:
-            # Detect if the super method uses varargs and the function doesn't
+            # Detect if the super method uses varargs and the function doesn't or makes some of those explicit
             if meth_node.args.vararg and (
                 not function.args.vararg
                 or len(function.args.args) > len(meth_node.args.args)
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):
LGTM !

tests/functional/u/useless/useless_super_delegation.py
```diff
@@ -303,3 +303,13 @@ def __str__(self):
...
+
+
+# Reported in https://github.com/PyCQA/pylint/issues/2270
+class Super:
+    def __init__(self, *args):
+        self.args = args
+
+class Sub(Super):
+    def __init__(self, a, b):
+        super().__init__(a, b)
```
mbyrnepr2, line 315: How about when the first argument is required? I see the useless-super-delegation for this:

```python
class Super:
    def __init__(self, *args):
        self.args = args

class Sub(Super):
    def __init__(self, a, *args):
        super().__init__(a, *args)
```

DanielNoord: So you would want this to raise whenever there is only a variadic in `super()` and one or more positional in child?

mbyrnepr2: @DanielNoord that's exactly it. It's just a variation of the original issue which someone may encounter.

mbyrnepr2: I think that will work @DanielNoord. I haven't time and haven't played with other variations but I think it's good to go. Thanks!

pylint/checkers/classes/class_checker.py
```diff
@@ -1276,6 +1276,12 @@ def _check_useless_super_delegation(self, function: nodes.FunctionDef) -> None:
         args = _signature_from_call(call)
 
         if meth_node is not None:
+            # Detect if the super method uses varargs and the function doesn't
```
DanielNoord, line 1279:
```suggestion
            # Detect if the super method uses varargs and the function doesn't or makes some of those explicit
```

tests/functional/u/useless/useless_super_delegation.py
DanielNoord: Updated!

PR #6956, opened 2022-06-15
Fix ``bad-super-call`` for non-direct parents

### LINKED ISSUE #2903: False positive: calling super on bases base class is shown as error
### Steps to reproduce
1. Make an empty folder and create the file test.py with this content:
image
2. `pylint test.py`

### Current behavior
Executing the code works, but a false positive error is shown.

`E1003: Bad first argument 'B' given to super() (bad-super-call)`

### Expected behavior
Pylint shouldn't show this error.

### pylint --version output
```pylint 2.3.1
astroid 2.2.5
Python 3.7.3 (default, Mar 26 2019, 21:43:19) 
[GCC 8.2.1 20181127]
```

### LINKED ISSUE #4922: E 1003: bad-super-call with two-argument super() call
### Bug description

```python
"""Produces 'bad-super-call'."""


class ClassB:
    "Dummy B."
    
    def whoami(self):
        """Print Name"""
        print("I am B")


class ClassC:
    "Dummy C."

    def whoami(self):
        """Print Name"""
        print("hello? this is C.")


class ClassD(ClassB, ClassC):
    "Dummy D."

    def greet(self):
        """Print Name"""
        super(ClassB, self).whoami()

d = ClassD()
d.greet()
d.whoami()
```

### Configuration

```ini
`super(ClassB, self).whoami()`
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
...
test.py:25:8: E1003: Bad first argument 'ClassB' given to super() (bad-super-call)
...
```

### Expected behavior

Hi!

This is similar to #2903, but the major difference is multiple inheritance. I hope the example makes sense from a code design perspective (if not just imagine that classB and classC inherit from an ABC that has the `whoami` method as an abstractmethod).

As far as I can tell, the flagged call to super is valid and the only way to actually call the `whoami` function of `ClassC` on `ClassD` instances.

Of course, in this example, switching the order of inheritance would be a simple solution, but just assume that based on other (not included) methods, the inheritance order has to be this way.

Thanks

### Pylint version

```shell
pylint 2.10.3-dev0
astroid 2.7.2
Python 3.9.6 (default, Jun 30 2021, 10:22:16) 
[GCC 11.1.0]
```

### OS / Environment

Linux 5.13.12-arch1-1 #1 SMP PREEMPT Wed, 18 Aug 2021 20:49:03 +0000 x86_64 GNU/Linux

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #4922, Closes #2903

### DIFF
--- a/doc/data/messages/b/bad-super-call/bad.py
+++ b/doc/data/messages/b/bad-super-call/bad.py
@@ -2,7 +2,7 @@ class Animal:
     pass
 
 
-class Fish:
+class Tree:
     pass
 
 
--- a/doc/data/messages/b/bad-super-call/bad.py
+++ b/doc/data/messages/b/bad-super-call/bad.py
@@ -8,5 +8,5 @@ class Tree:
 
 class Cat(Animal):
     def __init__(self):
-        super(Fish, self).__init__()  # [bad-super-call]
+        super(Tree, self).__init__()  # [bad-super-call]
         super(Animal, self).__init__()
--- a/doc/data/messages/b/bad-super-call/bad.py
+++ b/doc/data/messages/b/bad-super-call/bad.py
@@ -2,6 +2,11 @@ class Animal:
     pass
 
 
+class Fish:
+    pass
+
+
 class Cat(Animal):
     def __init__(self):
-        super(Animal, self).__init__()  # [bad-super-call]
+        super(Fish, self).__init__()  # [bad-super-call]
+        super(Animal, self).__init__()
--- a/doc/data/messages/b/bad-super-call/good.py
+++ b/doc/data/messages/b/bad-super-call/good.py
@@ -2,7 +2,7 @@ class Animal:
     pass
 
 
-class Fish:
+class Tree:
     pass
 
 
--- a/doc/data/messages/b/bad-super-call/good.py
+++ b/doc/data/messages/b/bad-super-call/good.py
@@ -2,6 +2,10 @@ class Animal:
     pass
 
 
+class Fish:
+    pass
+
+
 class Cat(Animal):
     def __init__(self):
-        super().__init__()
+        super(Animal, self).__init__()
--- a/tests/functional/s/super/super_checks.py
+++ b/tests/functional/s/super/super_checks.py
@@ -29,7 +29,7 @@ class Py3kAaaa(NewAaaa):
 class Py3kWrongSuper(Py3kAaaa):
     """new style"""
     def __init__(self):
-        super(NewAaaa, self).__init__()  # [bad-super-call]
+        super(NewAaaa, self).__init__()
 
 class WrongNameRegression(Py3kAaaa):
     """ test a regression with the message """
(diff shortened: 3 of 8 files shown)


doc/data/messages/b/bad-super-call/bad.py
```diff
@@ -2,6 +2,11 @@ class Animal:
     pass
 
 
+class Fish:
```
Pierre-Sassoulas, line 5:
```suggestion
class Fish(Animal):
```
Shouldn't Fish inherit from Animal ?

doc/data/messages/b/bad-super-call/good.py
```diff
@@ -2,6 +2,10 @@ class Animal:
     pass
 
 
+class Fish:
```
Pierre-Sassoulas, line 5:
```suggestion
class Fish(Animal):
```

tests/functional/s/super/super_checks.py
```diff
@@ -123,3 +123,34 @@ class SuperWithSelfClass(object):
...
+class GrandChild(Child):
+    def method(self):
+        print("Grandchild")
+        super(GrandChild, self).method()
+        super(Child, self).method()
+        super(Niece, self).method()  # [bad-super-call]
+
+
+# Reported in https://github.com/PyCQA/pylint/issues/4922
+class AlabamaCousin(Child, Niece):
```
Pierre-Sassoulas, line 153: Well, that's very funny 😄 ``CarlosTwoOfSpain`` works too if we want to keep the goodwill of alabamians 😛

Pierre-Sassoulas (changes requested):

doc/data/messages/b/bad-super-call/bad.py
DanielNoord: Yeah, but then this example doesn't work.

I'll change!

```suggestion
class Tree:
```

doc/data/messages/b/bad-super-call/good.py
DanielNoord:
```suggestion
class Tree:
```

doc/data/messages/b/bad-super-call/bad.py
```diff
@@ -2,6 +2,11 @@ class Animal:
...
 
 
+class Tree:
+    pass
+
+
 class Cat(Animal):
     def __init__(self):
-        super(Animal, self).__init__()  # [bad-super-call]
+        super(Fish, self).__init__()  # [bad-super-call]
```
DanielNoord, line 11:
```suggestion
        super(Tree, self).__init__()  # [bad-super-call]
```

doc/data/messages/b/bad-super-call/bad.py
Pierre-Sassoulas: I did suggest tree at first too but the Niece/parent example made me think this should work with fish/animal ?

DanielNoord: Don't you think this is clearer?

DanielNoord: Let's wait for the implementation of #6961 before adding that.

Pierre-Sassoulas (approved):

PR #6958, opened 2022-06-15
Emit `used-before-assignment` for self-referencing assignments under if conditions

### LINKED ISSUE #6643: False negative used-before-assignment for self-referencing assignment in if-statement
### Bug description

The following code  should trigger a `used-before-assignment` error from pylint, but it does not:

```
if 1 + 1 == 2:
    x = x + 1
```

When the if-statement is removed like
```
x = x + 1
```
the error is produced as expected.

### Configuration

_No response_

### Command used

```shell
python -m pylint foobar.py
```

### Pylint output

```shell
```

### Expected behavior

There should be a `used-before-assignment` error.

### Pylint version

```shell
$ python -m pylint --version
pylint 2.13.9
astroid 2.11.5
Python 3.8.13 (default, Mar 31 2022, 14:08:37)
```

### OS / Environment

Python interpreter was built by pyenv, on Ubuntu Linux 20.04

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description
Self-referencing assignments under if statements were getting a free pass.

Closes #6643

### DIFF
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -46,6 +46,10 @@ False negatives fixed
 
   Closes #5653
 
+* Emit ``used-before-assignment`` for self-referencing assignments under if conditions.
+
+  Closes #6643
+
 
 Other bug fixes
 ===============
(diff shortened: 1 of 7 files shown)


Pierre-Sassoulas (approved):
LGTM, nice refactor with returning the node instead of a boolean it feel a lot clearer that way.

PR #6974, opened 2022-06-18
Use `towncrier` for changelog generation

DudeNr33:

## Type of Changes

## Description
Draft for using `towncrier` to generate the release notes.
Feedback is welcome.

Things that need to be discussed:
* ~In this comment @Pierre-Sassoulas had the idea to reference the PR number rather than the issue number. Is this still the way we want to go? As it is not possible to know the PR number before creating the PR itself, this would mean one needs to always create and push the news fragment after drafting the PR. I find that inconvenient. 
https://github.com/PyCQA/pylint/pull/6688#discussion_r887582505~

Things I still need to implement:
* [x] Automatically update `towncrier.toml` with a `tbump` script - I would like to wait for feedback first before doing that.

Things that have to be done before merging:
* [x] extract all current changelog entries into news fragments


requirements_test.txt
```diff
@@ -11,3 +11,4 @@ pytest-xdist~=2.5
 # Type packages for mypy
 types-pkg_resources==0.1.3
 tox>=3
+towncrier~=21.9
```
DudeNr33, line 14: I would include the dependency here so that contributors can actually use the `towncrier create` command.

towncrier.toml
```diff
@@ -0,0 +1,50 @@
+[tool.towncrier]
+version = "2.15.0-dev0"
+directory = "doc/whatsnew/fragments"
+filename = "doc/whatsnew/2/2.15/index.rst"
```
DudeNr33, line 4: The version and filename should be updated through a `tbump` script later on.

Pierre-Sassoulas: Search for `__pkginfo__.py` in the tbump conf, to see how it's done :)

Pierre-Sassoulas:
The check on the pr number is possible in GitHub actions because GitHub actions have access to it. That's why I wanted to change that. If towncrier is doing the check for the change description another way I'd prefer keeping the consistency with the previous changelog. I don't mind either solution as we'll be able to navigate from issues to PR anyway as they are linked. (Some reference are for PR when no issues were created already so we'll have to handle the disambiguation pr/issues to link to te right thing in the documentation whatever we choose)

towncrier.toml
DudeNr33: Thanks for the hint! Just replacing the `version` is the easy part. Changing the `filename` would require a simple helper script which restricts the version to MAJOR.MINOR.

Pierre-Sassoulas (commented):
This look very promising !

DudeNr33:
> The check on the pr number is possible in GitHub actions because GitHub actions have access to it. That's why I wanted to change that. If towncrier is doing the check for the change description another way I'd prefer keeping the consistency with the previous changelog. I don't mind either solution as we'll be able to navigate from issues to PR anyway as they are linked. (Some reference are for PR when no issues were created already so we'll have to handle the disambiguation pr/issues to link to te right thing in the documentation whatever we choose)

As far as I understand the integrated `towncrier check` command only checks if there is *any* new news fragment.
It does not include any logic to see if the stem of the news fragment is actually an issue number. 
However, we can of course enhance the changelog pipeline with a custom script that searches the PR body for refs/closes/etc., and checks if the added news fragment starts with the number given there.

I currently use the stem of the news fragment file to construct a link to the issue in the changelog:
https://pylint--6974.org.readthedocs.build/en/6974/whatsnew/2/2.15/index.html#false-negatives-fixed
However, this approach is currently missing the information if this is "closes", "refs", or whatever else it may be.
We could change that by including your original `check_changelog` script from #6688 in the changelog CI and enforce that a news fragment has to include a reference to an issue.

DudeNr33:
I think we can split the check for the news fragments in two parts:
1. The `changelog.yml` workflow ensures that every PR without the "skip news" label contains a news fragment. It does not check the content.
2. The `check_newsfragments.py` script is used in the pre-commit configuration. It aims to check the contents of the news fragment. 

This way contributors can see as early as possible if the created news fragment has the correct format. And if they forget to add one, the CI will notice and inform them.

script/bump_towncrier.py
```diff
@@ -0,0 +1,38 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
Pierre-Sassoulas, line 1: Probably something that should be done in ``"python3 script/bump_changelog.py {new_version}"`` There's quite a bit of cleanup to do there with the adoption of towncrier ! We still have the script on the maintenance branch for 2.14 so almost everything can be removed (I think we only need to create the new file for minor automatically ?)

doc/development_guide/contributor_guide/contribute.rst
```diff
@@ -13,9 +13,9 @@ your patch gets accepted:
   you access to the latest ``ast`` parser.
 - Install the dev dependencies, see :ref:`contributor_install`.
 - Use our test suite and write new tests, see :ref:`contributor_testing`.
-- Add an entry to the change log describing the change in `doc/whatsnew/2/2.15/index.rst`
-  (or ``doc/whatsnew/2/2.14/full.rst`` if the change needs backporting in 2.14).
-  If necessary you can write details or offer examples on how the new change is supposed to work.
+- Create a news fragment with `towncrier create <IssueNumber>.<type>` which will be
+  included in the changelog. If necessary you can write details of offer examples on how
+  the new change is supposed to work.
```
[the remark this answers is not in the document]
DudeNr33: I tried the following:
```rst
.. include:: ../../../.github/PULL_REQUEST_TEMPLATE.md
   :start-after: - [ ] Write a good description on what the PR does.
   :end-before: - [ ] If you used multiple emails or multiple names when contributing, add your mails
```

However, I did not find an option that would allow me to drop the `[ ]` checkbox we have in the PULL_REQUEST_TEAMPLATE.md.
Any idea how to do that?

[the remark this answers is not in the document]
Pierre-Sassoulas: Maybe we can include the newline and `- [ ]` in the start after ?

script/bump_towncrier.py
DudeNr33: I agree. I was a bit hesitant because of the "DO NOT MODIFY DIRECTLY" comment. 😄  That's probably not relevant any more / not relevant until we might decide to use `towncrier` in `android` as well?

Reading the `release.md` documentation, we have to handle two different invocations of `tbump`:

1. Invocation for a release version (without a `-devX`): 
1.1 Update the `version` in `towncrier.toml` 
1.2 build the final documentation 
2. Invocation for a dev version: 
2.1 Update the `version` in `towncrier.toml`
2.2 Update the `filename` in `towncrier.toml`
2.3 Create a new `doc/whatsnew/MAJOR/MAJOR.MINOR/index.rst` and link that in the parent `index.rst` files' toctrees

Maybe we should do the `tbump` related things in a separate PR, to keep this one reviewable.

doc/development_guide/contributor_guide/contribute.rst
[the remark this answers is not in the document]
DudeNr33: I tried that, but simply appending "\n - [ ]" did not work, and Sphinx complained that it could not find the string.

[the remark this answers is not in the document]
Pierre-Sassoulas: I think you need a litteral newline not `\n` . Let's simply add a comment to remember that it's a duplication if it does not work.

script/bump_towncrier.py
Pierre-Sassoulas: Hmm, right. Astroid is behind now and has a lot less activity than pylint. Last time I did astroid's script first, with the intention to use the same in both but pylint has different need. I don't mind keeping the old release and script routine in astroid as there is no whatsnew in astroid so no duplication for contributor.

DudeNr33:
Sorry guys for letting this open for so long. 
I think beside the necessary rework of the tbump scripts this should no be ready for review. 
I had some trouble getting the `towncrier check` command to work in the CI, but it does now!

Check out the current docs build [here](https://pylint--6974.org.readthedocs.build/en/6974/whatsnew/2/2.15/index.html).
I am thinking about removing the hyperlink to the ticket/PR, as it is duplicated (once from the content of the news fragment itself without hyperlink, and once the hyperlink added by towncrier itself).

I added newsfragments for all changes currently included in the 2.15 changelog until now, and can do a last update before we merge.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,94 @@
+name: changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled, reopened]
```
DanielNoord, line 5: Which types are we excluding here? Can't we just use `on: pull_request`?

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,94 @@
...
+
+permissions:
+  contents: read
```
DanielNoord, line 8: Is there a reason we don't use the default `read|write` here?

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,94 @@
...
+
+env:
+  # Also change CACHE_VERSION in the other workflows
+  CACHE_VERSION: 8
+  DEFAULT_PYTHON: 3.8
```
DanielNoord, line 13:
```suggestion
  DEFAULT_PYTHON: 3.10
```

I was thinking of doing this for other workflows as well, but since `3.10` is faster than `3.8` I think it makes sense to run jobs on the latest stable versions.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,94 @@
...
+
+jobs:
+  prepare-base:
```
DanielNoord, line 16: Let's merge this with the other job, that saves time 😄

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,94 @@
...
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{
+            needs.prepare-base.outputs.python-key }}
+      - name: Fail job if Python cache restore failed
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          echo "Failed to restore Python venv from cache"
+          exit 1
+      - name: Emit warning if news fragment is missing
+        if:
+          contains(github.event.pull_request.labels.*.name, 'skip news :mute:') != true
```
DanielNoord, line 86: Can we completely skip this workflow if this is true? Instead of only in the last step?

doc/conf.py
```diff
@@ -131,7 +131,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ["_build", "data/**"]
+exclude_patterns = ["_build", "data/**", "whatsnew/fragments"]
```
DanielNoord, line 134: I don't know `towncrier`, but does it make sense to store fragments in separate directories for different versions?

script/bump_towncrier.py
```diff
@@ -0,0 +1,38 @@
...
+import argparse
+import re
+from pathlib import Path
+
+NEWSFILE_PATTERN = re.compile(r"doc/whatsnew/\d/\d.\d+/index\.rst")
+NEWSFILE_PATH = "doc/whatsnew/{major}/{major}.{minor}/index.rst"
+TOWNCRIER_CONFIG_FILE = Path("towncrier.toml")
+
+
+def main():
```
DanielNoord, line 16:
```suggestion
def main() -> None:
```

``mypy --strict`` will complain otherwise.

towncrier.toml
```diff
@@ -0,0 +1,50 @@
...
+template = "doc/whatsnew/fragments/_template.rst"
+issue_format = "`#{issue} <https://github.com/PyCQA/pylint/issues/{issue}>`_"
```
DanielNoord, line 6: This is really useful, let's keep this in!

towncrier.toml
DanielNoord: How does this work with fragments of older versions then? Won't they get mixed up?

DanielNoord (changes requested):
Thanks a lot for all the work here @DudeNr33!

I left some initial comments while I'm trying to wrap my head around what `towncrier` is doing for us 😄

.github/workflows/changelog.yml
DudeNr33: I took this over from Pierre's initial PR for checking the existence of a changelog entry. The reasoning was that with the default configuration the pipeline is not re-triggered when a label is applied. But we need that, because adding/removing the "Skip news" label must trigger a new run of this check.

.github/workflows/changelog.yml
DudeNr33: I took this over from Pierre's PR as well. We don't need write permissions here, so I guess this would just be the principle of least privileges. However I would be fine with removing this as well.

DanielNoord: 👍 

Let's remove in the spirit of less is more then. I don't like how these simple workflow files can become quite long just to get their set up right.

doc/conf.py
DudeNr33: This is not necessary - once you rown `towncrier build`, the existing newsfragments are deleted. Thus you only keep the fragments of the current, upcoming release.

DudeNr33:
Thanks for the feedback, Daniel! 

As it is now already 3 weeks since we initially discussed `towncrier`, a short recap what it can do for us:

* Instead of writing changelog entries in one common `index.rst`, individual files (news fragments) are used --> No more merge conflicts because of the changelog
* `towncrier check` can be used in CI to check if a news fragment was added --> Contributors will be notified automatically if they forget to write a changelog entry
* The concept of news fragments makes implementing a pre-commit hook to check the format of the changelog entry straightforward

I will incorporate the rest of your feedback tomorrow. 😊

Pierre-Sassoulas:
>  No more merge conflicts because of the changelog

70dda48d5b21b9c82783b2408ae4a5b7

doc/conf.py
Pierre-Sassoulas: Where are the old fragments stored ? Do we only store the generated result  ?

DudeNr33: Exactly. `towncrier build` appends to the specified output file, and deletes the existing news fragments.

script/bump_towncrier.py
DudeNr33: Changed.

.github/workflows/changelog.yml
DudeNr33: Good catch, moved the `if` up to the job level.

.github/workflows/changelog.yml
DudeNr33: Done 👍

.github/workflows/changelog.yml
DudeNr33: I wholeheartedly agree. It's already quite verbose just for the caching of the venv. 😞

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,77 @@
...
+    types: [opened, synchronize, labeled, unlabeled, reopened]
+
+env:
+  # Also change CACHE_VERSION in the other workflows
+  CACHE_VERSION: 8
+  DEFAULT_PYTHON: 3.10
+
+jobs:
+  check-changelog:
+    if: contains(github.event.pull_request.labels.*.name, 'skip news :mute:') != true
```
DanielNoord, line 14: I wonder how this interacts with setting this as a required action. Are skipped actions considered as being successful? Otherwise we might need a first step that checks the if statement and then set the state to successful and exit.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,77 @@
...
+    name: Changelog Entry Check
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      python-key: ${{ steps.generate-python-key.outputs.key }}
```
DanielNoord, line 19:
```suggestion
```

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          path: venv
+          key:
+            ${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{
+            needs.prepare-base.outputs.python-key }}
+      - name: Fail job if Python cache restore failed
+        if: steps.cache-venv.outputs.cache-hit != 'true'
+        run: |
+          echo "Failed to restore Python venv from cache"
+          exit 1
```
DanielNoord, line 68:
```suggestion
```

This is unnecessary now that we use one job 😄  See:
https://github.com/PyCQA/pylint/blob/main/.github/workflows/primer_run_main.yaml

script/check_newsfragments.py
```diff
@@ -0,0 +1,68 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "filenames",
+        nargs="*",
+        metavar="FILES",
+        help="File names to check",
+    )
+    parser.add_argument("--verbose", "-v", action="count", default=0)
+    args = parser.parse_args(argv)
+    verbose = args.verbose
```
DanielNoord, line 40:
```suggestion
```

script/check_newsfragments.py
```diff
@@ -0,0 +1,68 @@
...
+    is_valid = True
+    for filename in args.filenames:
+        file = Path(filename)
+        is_valid &= check_file(file, verbose)
```
DanielNoord, line 44:
```suggestion
        is_valid &= check_file(Path(filename), args.verbose)
```

script/check_newsfragments.py
```diff
@@ -0,0 +1,68 @@
...
+        nargs="*",
+        metavar="FILES",
+        help="File names to check",
+    )
+    parser.add_argument("--verbose", "-v", action="count", default=0)
+    args = parser.parse_args(argv)
+    verbose = args.verbose
+    is_valid = True
+    for filename in args.filenames:
+        file = Path(filename)
```
DanielNoord, line 43:
```suggestion
```

script/check_newsfragments.py
```diff
@@ -0,0 +1,68 @@
...
+        is_valid &= check_file(file, verbose)
+    return 0 if is_valid else 1
```
DanielNoord, line 45: `sys.exit` works with booleans so we can just use booleans although `bool(True)` is `1`. So we would need to slightly modify to `check_if_incorrect` or something.

DanielNoord (changes requested):
Thanks for the clear answers @DudeNr33 

Some (final) comments, but this looks really promising!

.github/workflows/changelog.yml
DudeNr33: I don't really know, I'm no expert in Github actions - so I say we just set the "Skip news" label on this PR and see what happens. 😄

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,10 +4,9 @@ Thank you for submitting a PR to pylint!
 To ease the process of reviewing your PR, do make sure to complete the following boxes.
 
 - [ ] Write a good description on what the PR does.
-- [ ] Add an entry to the change log describing the change in
-  `doc/whatsnew/2/2.15/index.rst` (or ``doc/whatsnew/2/2.14/full.rst``
-   if the change needs backporting in 2.14). If necessary you can write
-   details or offer examples on how the new change is supposed to work.
+- [ ] Create a news fragment with `towncrier create <IssueNumber>.<type>` which will be
+  included in the changelog. If necessary you can write details of offer examples on how
+  the new change is supposed to work.
```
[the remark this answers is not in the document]
Pierre-Sassoulas: It seems the list can be generated from ``towncrier.toml``, hardcoding is enough imo, I'll update the suggestion

script/check_newsfragments.py
DudeNr33: I'm not quite sure if I understand what you mean with "slightly modify to `check_if_incorrect` or something"? But the hint with `sys.exit` is nice, that makes it simpler.

.github/workflows/changelog.yml
DanielNoord: Well I tried and that is inconclusive.
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks also doesn't make me that much wiser.

@Pierre-Sassoulas Could you see the changelog action as a required check and see if the message about missing required checks pops up right now?

script/check_newsfragments.py
DanielNoord: I meant that currently we return `is_valid` but that evaluates to `True` which is `1` which gives `sys.exit(1)` so we'll fail when `is_valid` is `True`. So we need to turn the conditional around to have `check_file` return `is_invalid`.

.github/workflows/changelog.yml
Pierre-Sassoulas: We don't have any required checks right now so maintainers can take executives decisions in case something is down (or simply merge before any tests ran after fixing a small typo instead of waiting 35mn again). Do we want to make this check required ?

Pierre-Sassoulas: (Making it required affects maintainers, not contributors, we're not going to merge PR that do not pass the checks).

DanielNoord: I think it makes sense to turn it on for this check.

Either way, I'd like to see if the way we currently skip the workflow works with required checks because if not I'd like to make it compatible from the start.

script/check_newsfragments.py
DudeNr33: Aaah, got you. In that case I'd rather leave it as it is right now. Imo the negation of the return value would be similar confusing as the current ternary if, whereas the latter has the benefit that it's a bit clearer we want to return the actual exit code of the program.

DanielNoord: Wouldn't switching `True` and `False` around and renaming to `check_missing_entry` already work? Oh well, never mind!

script/check_newsfragments.py
```diff
@@ -0,0 +1,66 @@
...
+        issue = match.group("issue")
+        if file.stem != issue:
+            print(
+                f"{file} must be named '{issue}.<fragmenttype>', after the issue it references."
+            )
+            return False
+        if verbose:
+            print(f"Checked '{file}': LGTM 🤖👍")
+        return True
+    print(f"{file}: does not respect the standard format 🤖👎")
```
DanielNoord, line 61: Could we add something about what the format is here?

DanielNoord (changes requested):
One final nit.

I'd like to see the interaction with required checks but rest LGTM! 😄

.github/workflows/changelog.yml
Pierre-Sassoulas: This is also going to prevent any other MR to be merged until we merge this one and the check really exists, maybe we could do this in a follow-up to avoid that ?

DanielNoord: If you ping me just before you do it I'll do a quick check of the result and you can disable it again :)

Pierre-Sassoulas: Should be required now :)

DanielNoord: By requiring being up to date with `main` I can't see if this passes required tests without making a merge commit 😅 

Could you turn that off?

Pierre-Sassoulas: I don't think I changed that. Maybe requiring a job require the branch to be up to date ?

DanielNoord: The button to allow merging `main` from within the Github interface is now also on. Is that also a requirement? That's annoying..

DanielNoord: Schermafbeelding 2022-07-10 om 23 31 31

It's the second option. That should be turned off.

Pierre-Sassoulas: Ha thanks @DanielNoord, it reactivate the option automatically when you tick "require status checks"  so you have to deactivate everytime !

DanielNoord: This seems to work.

Schermafbeelding 2022-07-11 om 08 26 39

So we can leave this as is.

script/check_newsfragments.py
DudeNr33: Sure, this would be helpful. Instead of just printing the cryptic regex I included a short text with placeholders.

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
DudeNr33: Updated 👍

DanielNoord (approved):

DudeNr33:
One last question from my side: 
Should the PULL_REQUEST_TEMPLATE.md still contain some words if we want to treat issues for 2.14.x the same way we did before? Is there any patch release for 2.14 planned?

Beside that I think all open issues are addressed now. I'd just make sure that I double check if I have transformed all changelog entries into news fragments just before we merge.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -12,90 +12,4 @@ Summary -- Release highlights
...
-
-* ``pylint.testutils.primer`` is now a private API.
-
-  Refs #6905
-
-* Fixed an issue where it was impossible to update functional tests output when the existing
-  output was impossible to parse. Instead of raising an error we raise a warning message and
-  let the functional test fail with a default value.
-
-  Refs #6891
```
Pierre-Sassoulas, line 101: I did not check everything but we probably added entries here since we started the MR, we should do a last check before merging. (And also merge relatively fast to avoid this issue of course).

script/bump_towncrier.py
Pierre-Sassoulas: Did you have time to look at this @DudeNr33 ?

towncrier.toml
Pierre-Sassoulas, line 6: Open question : This look amazing, do you have an idea how we could create link like that for old changelog ? I've been wanting to create the link for some time now. Might be a one time sed now that we don't touch it anymore.

Pierre-Sassoulas (changes requested):
Looks terrific already !

doc/whatsnew/2/2.15/index.rst
DudeNr33: Yes, let me know when all review comments are resolved from your point of view and I'll check for completeness again!
Sorry that this topic is dragging itself a bit, I did not find that much time for this in the last few weeks.

script/bump_towncrier.py
DudeNr33: Not yet unfortunately. I'll try to get to it in the next days.

script/check_newsfragments.py
```diff
@@ -0,0 +1,74 @@
...
+            print(f"Checked '{file}': LGTM 🤖👍")
+        return True
+    print(
+        f"{file}: does not respect the standard format 🤖👎\n\n"
+        "The standard format is:\n\n"
+        "<one or more line of text>\n"
+        "<one blank line>\n"
+        "<issue reference>\n\n"
+        "where <issue reference> can be one of:\n"
+        f"{', '.join(VALID_ISSUES_KEYWORDS)}"
```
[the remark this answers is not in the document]
DudeNr33: That's much better, thank you!

towncrier.toml
DudeNr33: Yes, I would probably go the same route - use `sed` or a simple helper script to replace all occurrences of issue numbers with a link.

DanielNoord:
@jacobtylerwalls @Pierre-Sassoulas Shall we wait with merging those regressions PRs until this PR has been merged?

jacobtylerwalls:
That's alright with me.

Pierre-Sassoulas:
I think this MR and astroid 2.12 / python 3.11 MR are top priority, the other can wait (especially since merging other MR create work on this one).

Pierre-Sassoulas (approved):
It look great ! I did not see any missed entries in the new fragments. 

> And probably double check the results of the new bump_changelog.py when bumping 2.15.0 / 2.15.1-dev0.

Yes, I think we should release 2.14.5 then 2.15.0 for two reasons:
- Seeing what we missed during review asap (Theory only goes so far)
- Not having to maintain the old changelog anymore in the maintenance branch and being able to upgrade the contribution doc entirely

What I'm a little worried about is patch releases if we mess the 2.15.0 release and the process is not ready for 2.15.1. Worst case scenario I could create a manual release notes.

We have a new checker and a bunch of false negatives fixed, this is not as much as 2.13 or 2.14 but more frequent releases is a good thing imo. What do you think @DanielNoord @jacobtylerwalls ? Any unreleased issue we should fix first ? Maybe with astroid 2.12 ?

jacobtylerwalls:
I know we don't want to do a beta/pre-release every single release, but I think the namespace package changes might warrant doing so for pylint 2.15, because folks might not know that they're depending on _not_ linting a namespace package somewhere in their setup (like we had to fiddle with `tests` being interpreted as a namespace package).

> Any unreleased issue we should fix first ? Maybe with astroid 2.12 ?
Yes, there's an open crash filed against astroid 2.12.2

DanielNoord:
@Pierre-Sassoulas can we postpone 2.14 until Friday/Saturday? I'll be gone for holidays from next week and I think I had some config regressions on my radar that should probably be fixed...

This pr should be good to go though.

Pierre-Sassoulas:
@jacobtylerwalls is right we need a beta(especially for namespace package), and this will permit to test the changelog  just as well as a real release !

DudeNr33:
Pre-releases is actually a very good point - right now we would run `towncrier build` for every invocation of `tbump`, which would "finalise" the release notes (write everything to `index.rst` and delete the fragments).

This is probably not what we want for pre-releases or when bumping to a dev version, so I should move that into `bump_changelog.py` and include some logic to skip the `towncrier build` step for dev versions and pre-releases.

DudeNr33:
I moved the `towncrier build` step to `bump_changelog.py`, so that it is only executed for a release version.

DudeNr33:
Does anybody have an idea what the problem with the failing pipelines could be? 

It looks it restores the cache from the key:
```console
Run actions/cache@v3.0.4
  with:
    path: venv
    key: Linux-3.7.13-venv-[2](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:2)-421ce8c8e8ca415125fbb808[3](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:3)9fa803fd3570[4](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:4)[5](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:5)e9[6](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:6)ca96b0445b[7](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:7)3b9[8](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:8)3e5b[9](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:9)63
    restore-keys: Linux-3.7.13-venv-2-
  
  env:
    CACHE_VERSION: 2
    pythonLocation: /opt/hostedtoolcache/Python/3.7.13/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.7.13/x64/lib/pkgconfig
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.7.13/x64/lib
Received 36461966 of 36461966 ([10](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:10)0.0%), 54.5 MBs/sec
Cache Size: ~35 MB (36461966 B)
/usr/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/6d4ab4dc-5300-4edf-8d00-1862af7e48df/cache.tzst -P -C /home/runner/work/pylint/pylint
Cache restored successfully
Cache restored from key: Linux-3.7.[13](https://github.com/PyCQA/pylint/runs/7376108148?check_suite_focus=true#step:5:14)-venv-2-b39b2027c5ef4bb4ce81910ccd6872c3b0bc405ba3e9294e72a66d57959ce3cb
```

But then still tries to install the dependencies afterwards, and fails because `astroid` is installed in V2.11.7 but is now required in V2.12.
```console
pip._vendor.pkg_resources.ContextualVersionConflict: (astroid 2.11.7 (/home/runner/work/pylint/pylint/venv/lib/python3.7/site-packages), Requirement.parse('astroid<=2.14.0-dev0,>=2.12.2'), {'pylint'})
```

I'm no expert regarding the caching mechanisms.
I don't understand
* how GH actions still wants to create the virtual environment even though the cache restore *seemingly* succeeded?
* Why it restores an outdated cache (I assume it is outdated because of the `astroid` version)?
* Why the `pip install -U -r requirements_test.txt`  command fails with the version conflict and does not simply update `astroid` to the required version?

If anybody has a solution or idea what the problem might be I'd be glad to hear it before I start digging myself. 😊

Pierre-Sassoulas:
I tried a different cache number and it's not working either. . It's reproducible with this MR and other MR don't work at first then work and can be merged once this MR is inactive for some time, So I suppose the problem is not the cache itself but what this MR is doing (?) I don't know exactly what could cause it. I was investigating the primers at first and got distracted into adding debugging information in the primer. Maybe the new job introduced here is using the same cache key than another job and shouldn't ?

DanielNoord:
Probably something to do with broken caches. Sadly Github doesn't allow removing caches through a nice interface and you can only do so by sending several `POST` requests.
I believe they are working on making it do-able from the site.

Bumped the cache to a number that we have never used and that seems to work. All tests are running now.

Pierre-Sassoulas:
> Bumped the cache to a number that we have never used and that seems to work. All tests are running now.

It did not work two days ago, unless we already used ``21031606416161651656510161616879`` before 😄

DanielNoord:
> > Bumped the cache to a number that we have never used and that seems to work. All tests are running now.
> 
> It did not work two days ago, unless we already used `21031606416161651656510161616879` before 😄

That might have been too long? CI passed 😄

Pierre-Sassoulas:
I regret rebasing on the 6 tests commits I did now 😄 Tried, 6, 7 , 8, 10, and a big number... I guess we'll never know it's too hard to check in https://github.com/PyCQA/pylint/actions?query=branch%3Atowncrier. Thanks for fixing this @DanielNoord !

DudeNr33:
Thanks Pierre and Daniel!
Is there anything left to do for this MR or can we merge this? 
Would this better be rebase and merge or squash and merge?

DanielNoord:
> Would this better be rebase and merge or squash and merge?

I'd prefer a `git reset --soft` and then some smaller commits that are a bit more explicit (for example separating the cache change from the new news entries, etc.) and then rebase the new clean commit history.

DudeNr33:
> > Would this better be rebase and merge or squash and merge?
> 
> I'd prefer a `git reset --soft` and then some smaller commits that are a bit more explicit (for example separating the cache change from the new news entries, etc.) and then rebase the new clean commit history.

Thanks for the hint, @DanielNoord!
To make sure I don't mess this up, this is what I would do now:

1. Make a backup branch from the current state of the `towncrier` branch (`git checkout -b towncrier-backup`) in case I *do* mess up, then switch back to this branch (towncrier)
2. Soft reset to the commit I originally based my towncrier branch on: `git reset --soft 9069eb48d`
3. Create new commits grouping together related files, like one for all news fragments, one for the pipelines, one for the updated docs
4. **force push** those new commits to rewrite the history on the remote
5. use GitHub web interface to select "Rebase and merge" to apply those newly created commits onto `main`

I tried that with a toy repo and it looks like that should do the trick, but step 4 makes me want to re-assure myself if this is really correct. 😄

Pierre-Sassoulas:
I would add a rebase on origin/main before 4 (that should be painless). To be safe in 4) you can ``--force-push-with-lease`` (it won't push if you did not have the last version of the server), but considering you're the only one working on your branch now it should be safe either way 👍

DanielNoord:
> > > Would this better be rebase and merge or squash and merge?
> > 
> > 
> > I'd prefer a `git reset --soft` and then some smaller commits that are a bit more explicit (for example separating the cache change from the new news entries, etc.) and then rebase the new clean commit history.
> 
> Thanks for the hint, @DanielNoord! To make sure I don't mess this up, this is what I would do now:
> 
> 1. Make a backup branch from the current state of the `towncrier` branch (`git checkout -b towncrier-backup`) in case I _do_ mess up, then switch back to this branch (towncrier)
> 2. Soft reset to the commit I originally based my towncrier branch on: `git reset --soft 9069eb48d`
> 3. Create new commits grouping together related files, like one for all news fragments, one for the pipelines, one for the updated docs
> 4. **force push** those new commits to rewrite the history on the remote
> 5. use GitHub web interface to select "Rebase and merge" to apply those newly created commits onto `main`
> 
> I tried that with a toy repo and it looks like that should do the trick, but step 4 makes me want to re-assure myself if this is really correct. 😄

Yes, this is what I was thinking of!

DudeNr33:
Thanks a ton you two! I always learn something new here. 👍 
Looks like it worked out as expected by looking at the number of commits and the diffs.
I'll do the rebase and merge once all the checks have passed.

DanielNoord:
@DudeNr33 Note that the first commit in this history can be avoided by a `git pull --rebase` instead of a merge but rebasing these commits is fine with me :)

Always happy to help, I tend to learn quite a lot from your PRs as well.

Pierre-Sassoulas:
> I tend to learn quite a lot from your PRs as well.

Yeap, confirming that, I don't comment on it but I often learn of new, more elegant way to do thing when reviewing.

Great job @DudeNr33 this is a huge new feature, I feel like I handled changelog conflict 10% of the time for the last 5 years, I'm glad it's over 😄 I'm going to try an alpha/beta release with it and see how it goes 👍

PR #6982, opened 2022-06-20
Partial typing of ``imports.py``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -16,6 +16,7 @@ from typing import TYPE_CHECKING, Any
 
 import astroid
 from astroid import nodes
+from astroid.nodes._base_nodes import ImportNode
 
 from pylint.checkers import BaseChecker, DeprecatedMixin
 from pylint.checkers.utils import (
@@ -30,6 +31,7 @@ from pylint.graph import DotBackend, get_cycles
 from pylint.reporters.ureports.nodes import Paragraph, Section, VerbatimText
 from pylint.typing import MessageDefinitionTuple
 from pylint.utils import IsortDriver
+from pylint.utils.linterstats import LinterStats
 
 if TYPE_CHECKING:
     from pylint.lint import PyLinter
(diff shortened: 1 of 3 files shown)


pylint/checkers/imports.py
```diff
@@ -536,7 +549,17 @@ def leave_module(self, node: nodes.Module) -> None:
...
+    def compute_first_non_import_node(
+        self,
+        node: nodes.If
+        | nodes.Expr
+        | nodes.Comprehension
+        | nodes.IfExp
+        | nodes.Assign
+        | nodes.AssignAttr
+        | nodes.TryExcept
+        | nodes.TryFinally,
```
Pierre-Sassoulas, line 561: If this can be everything except an import node it could also be a classDef or a lot of things, right ? Maybe ``NodeNG`` would make sense ?

Pierre-Sassoulas (commented):
Maybe we could create an alias for ``nodes.Import | nodes.ImportFrom`` ?

pylint/checkers/imports.py
DanielNoord: See L577. I thought I would be explicit just like we normally are for `visit_` methods.

That said L622 shows that L602 is incorrect.

I'm drafting as I'd like to fix that in this PR.

DanielNoord:
> Maybe we could create an alias for `nodes.Import | nodes.ImportFrom` ?

There is `mixins.ImportFrom` in `astroid` but since that doesn't inherit from `NodeNG` it doesn't really work.

Let's put https://github.com/PyCQA/astroid/issues/1633 in `2.12` and do the necessary refactoring as well. Marc just gave his thumbs up for the issue so I think he is also okay with my proposal to start inheriting from `NodeNG` for the `mixins`.

TLDR: Drafting as we need `astroid` `2.12` with additional changes + a fix for `visit_functiondef` in this module.

Pierre-Sassoulas (approved):
If mypy is happy, I'm happy too 😄

PR #6973, opened 2022-06-18
Refactor the primer command to be able to make the comment better later on

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Follow-up to #6969

PR #6984, opened 2022-06-20
[primer] Create a class for easier comparison of pylint results

Pierre-Sassoulas:
## Type of Changes

## Description

Follow up to #6973


pylint/testutils/_primer/comparator.py
```diff
@@ -0,0 +1,41 @@
...
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+from __future__ import annotations
+
+import json
+from collections.abc import Generator
+from pathlib import Path
+
+from pylint.testutils._primer.primer_command import Messages, PackageMessages
```
DanielNoord, line 11: Could you do a follow up PR that moves some of these to `typing.py` file?

pylint/testutils/_primer/primer_command.py
```diff
@@ -11,7 +11,8 @@
 
 from pylint.testutils._primer import PackageToLint
 
-PackageMessages = Dict[str, List[Dict[str, Union[str, int]]]]
+Messages = List[Dict[str, Union[str, int]]]
```
DanielNoord, line 14: Never mind my earlier comment. Perhaps you extract these now.

pylint/testutils/_primer/comparator.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Comparator:
```
DanielNoord, line 14: Could you rename this to `PylintComparator`? At some point I want to make an `AstroidComparator` 😄 

If you're up for it perhaps we could even create an abstract class?

DanielNoord (changes requested):

pylint/testutils/_primer/comparator.py
Pierre-Sassoulas: I'm starting to think we should use the existing class for this, I'll give it a try.

pylint/testutils/_primer/comparator.py
DanielNoord: No, using the `primer` on `astroid/main` to catch crashes there before we merge them. That would require a different `Comparator` I think.

DanielNoord: Yeah, but I thinking of maybe running the `pylint` test suite on some version on `astroid` `main` or something like that. The overall goal is to remove the need for `pylint-tested` label and have all PRs get that automatically.

Pierre-Sassoulas: Yeah sure I understood that :) It make sense to test astroid more in the astroid CI, because pylint's CI is already crowded so I think it's a really good idea too.

pylint/testutils/_primer/comparator.py
Pierre-Sassoulas: Resolving as we merged #7000

DanielNoord:
@Pierre-Sassoulas I would be okay with rebasing this and not doing the rename if that means we can merge this easily :)

Pierre-Sassoulas:
I think there was a concurrent change that forced to redo everything or at least a lot. (Which is why it's stuck not the rename). Might aknowledge the loss and redo from scratch later instead of trying to rebase then getting demoralised :)  Also there was a gazillion thing on my todo list before that (too-few-logging and always suggest for no-member) but we now have a cache problem to fix first :D...

pylint/testutils/_primer/comparator.py
Pierre-Sassoulas: Coming back to this in order to make the better diff in #10914, I rebased and I think it's mergeable now. Since Comparator already lives in pylint/testutils/_primer/, the Pylint prefix felt redundant in context. Regarding the abstract class, I think an astroid comparator would be better in astroid (compare AST rather than pylint messages). The dependency is not worth it especially considering that pylint depends on astroid not the other way around.

pylint/testutils/_primer/comparator.py
```diff
@@ -0,0 +1,63 @@
...
+            for idx in range(batches):
+                main_data.update(
+                    self._load_json(base_file.replace("BATCHIDX", "batch" + str(idx)))
+                )
+                self.pr_data.update(
+                    self._load_json(new_file.replace("BATCHIDX", "batch" + str(idx)))
+                )
+
+        self.missing_messages: PackageMessages = {}
+        for package, data in main_data.items():
```
DanielNoord, line 47: I'd normally probably change this to not do the reading and iteration in the `__init__` but in the `__iter__` to not have the `__init__` side effect too much.

If you can be bothered you can change it, otherwise also fine 😄

tests/testutils/_primer/test_comparator.py
```diff
@@ -0,0 +1,48 @@
...
+# Copyright (c) https://github.com/pylint-dev/pylint/blob/main/CONTRIBUTORS.txt
+
+import json
+from pathlib import Path
+
+import pytest
+
+from pylint.testutils._primer.comparator import Comparator
+
+FIXTURES_PATH = Path(__file__).parent / "fixtures"
```
DanielNoord:
```suggestion
CASES_PATH = Path(__file__).parent / "cases"
```

tests/testutils/_primer/test_comparator.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.mark.parametrize(
+    "directory",
+    [
+        pytest.param(p, id=p.name)
+        for p in FIXTURES_PATH.iterdir()
+        if p.is_dir() and p.name != "batched"  # tested separately
```
DanielNoord: Perhaps just move that case out of the directory then?

DanielNoord (commented):
Cool! Some small nits, but I'll stay on top of reviewing this so we can merge this and not let this get stale again.

pylint/testutils/_primer/comparator.py
Pierre-Sassoulas: [cdf7b3f](https://github.com/pylint-dev/pylint/pull/6984/commits/cdf7b3f8bafeccc075b0ece18422530b465a92d0)

DanielNoord:
@Pierre-Sassoulas Feels like a rebase went wrong here? There are some unrelated dependency updates in this PR.

DanielNoord (commented):
Except for the unrelated changes this LGTM!

@Pierre-Sassoulas Can you rebase again? I'll approve and merge afterwards!

Pierre-Sassoulas:
The duplicate-code change is required, I don't know why it suddenly fails, probably that the pre-commit grouping changed by adding new files.

DanielNoord (approved):

PR #6987, opened 2022-06-20
Add ``unguarded-next-without-default``

### LINKED ISSUE #4725: Checker for ``unguarded-next-called`` for ``next`` called without catching ``StopIteration``

StopIteration exceptions raised directly or indirectly in coroutines
and generators are transformed into RuntimeError exceptions for python
3.7+.

### Additional context

We missed one ``next`` in  astroid when doing https://github.com/PyCQA/astroid/pull/1081 and it resulted in  the avoidable #4723 in astroid ``2.6.3``

DanielNoord:

## Type of Changes

## Description

Closes #4725.


DanielNoord:
I checked some the primer changes and they seem fine.

One issue is the second example in `astroid`. If the `StopIteration` is caught in the function that is calling the function then you don't really need a default. But I guess we consider that a bad pattern and you should just re-raise? I feel like we have other checks that also rely on this logic.

Pierre-Sassoulas:
This one (using ``next()`` without default value in ``__next__``) seems like a false positive: https://github.com/pandas-dev/pandas/blob/main/pandas/io/parsers/python_parser.py#L1256

tests/functional/u/unspecified_default_for_next.py
```diff
@@ -0,0 +1,35 @@
...
+except (ValueError, StopIteration):
+    pass
+
+try:
+    next(i for i in (1, 2))
+except ValueError:
+    pass
+except StopIteration:
+    pass
+
```
Pierre-Sassoulas, line 28: We have a lot of example in astroid, but we can rely on the primer. Let's just handpick one:

```suggestion


def _unpack_keywords(self, keywords, context=None):
    for name, value in keywords:
        try:
            inferred = next(value.infer(context=context))
        except StopIteration:
            continue


```

Pierre-Sassoulas (changes requested):
Related PEP that I think we should talk about : https://peps.python.org/pep-0479/

~~I think we need to check the scope of the routine where the ``next`` is used to verify that the ``StopIteration`` is caught~~  (except if it's a function that is supposed to raise ``StopIteration`` like ``__next__``). We also need to check that ``py-version`` is > 3.7. Raìsing a ``RuntimeError`` implicitly is bad, doing anything else explicitly after catching the ``StopIteration``is okay.

DanielNoord:
> This one (using `next()` without default value in `__next__`) seems like a false positive: https://github.com/pandas-dev/pandas/blob/main/pandas/io/parsers/python_parser.py#L1256

I don't think `__next__` ever has a default value, see https://docs.python.org/3/library/stdtypes.html#iterator.__next__.

DanielNoord:
> Related PEP that I think we should talk about : https://peps.python.org/pep-0479/

👍 

> ~I think we need to check the scope of the routine where the `next` is used to verify that the `StopIteration` is caught~ (except if it's a function that is supposed to raise `StopIteration` like `__next__`). We also need to check that `py-version` is > 3.7. Raìsing a `RuntimeError` implicitly is bad, doing anything else explicitly after catching the `StopIteration`is okay.

Shouldn't that be a different check? As this is a Warning and `RuntimeError` probably warrants an Error level?

pylint/checkers/stdlib.py
```diff
@@ -720,6 +729,20 @@ def _check_env_function(self, node, infer):
...
             )
 
+    def _check_next_call(self, node: nodes.Call, name: str) -> None:
+        if name != "builtins.next":
+            return
+        if len(node.args) != 1:
+            # We don't care about this call if there are zero arguments
+            return
+        if utils.get_exception_handlers(node, StopIteration):
+            return
```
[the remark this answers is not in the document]
DanielNoord: Yeah, as

```python
def h():
    try:
        next(i for i in (1, 2) if isinstance(i, str))
    except StopIteration:
        pass
```
is fine.

But:
```python

def h():
    next(i for i in (1, 2) if isinstance(i, str))
    
def f():
    try:
        h()
    except StopIteration:
        pass
```

Isn't.

doc/data/messages/u/unguarded-next-without-default/bad.py
```diff
@@ -0,0 +1 @@
+next(i for i in (1, 2) if isinstance(i, str))  # [unspecified-default-for-next]
```
DanielNoord, line 1:
```suggestion
next(i for i in (1, 2) if isinstance(i, str))  # [unguarded-next-without-default]
```

pylint/checkers/stdlib.py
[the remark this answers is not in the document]
Pierre-Sassoulas: The way I understand it ``h`` is raising a runtime error ? (I did not test that 😅)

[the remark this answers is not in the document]
DanielNoord: First `h` doesn't raise, second `h` raises `StopIteration`, `f` raises nothing.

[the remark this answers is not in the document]
Pierre-Sassoulas: I just understood that you need to be in a generator for the RunTimeError to be raised. For example this:
```python
def h(iterator):
    for item in iterator:
        yield item, next(iterator)

def f(iterator):
    try:
        h(iterator)
    except StopIteration:
        pass

print([str(i) for i in h(i for i in (1, 2))])
print([str(i) for i in h(i for i in (1,))])
print([str(i) for i in h(i for i in ())])
```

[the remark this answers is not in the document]
DanielNoord: But that's a different issue right?

That is https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/stop-iteration-return.html: don't use `StopIteration` in generators as it creates `RuntimeError`.

This is about don't use `next()` in all cases where you don't guard for `StopIteration` or are in `__next__`. This shouldn't raise on `raise StopIteration` while the message above should.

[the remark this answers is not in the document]
Pierre-Sassoulas: Ha right, right, my bad.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -15,6 +15,10 @@ Summary -- Release highlights
 New checkers
 ============
 
+* Added ``unguarded-next-without-default`` checker which warns about calls to ``next()``
+  without a default value.
```
Pierre-Sassoulas, line 19:
```suggestion
  without a default value or handling of possible ``StopIteration`` in the same scope.
```

pylint/checkers/stdlib.py
```diff
@@ -434,6 +434,13 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
                 ]
             },
         ),
+        "W1519": (
+            "Using next without explicitly specifying a default value or catching the StopIteration",
+            "unguarded-next-without-default",
+            "Without a default value calls to next() can raise a StopIteration "
+            "exception. This exception should be caught or a default value can "
```
Pierre-Sassoulas, line 441:
```suggestion
            "exception. This exception should be caught or a default value should "
```

Pierre-Sassoulas (commented):
There's still a lot of false positive in most of the primer repository. I'm wondering if we should:
- infer the value of the generator so we do not raise if it's obvious it will always be possible to call next on it (Using next directly is reasonable to if the generator is defined in the same scope ?)
- Put this in an extension, currently it's a default check and seeing the number of warning in working libraries it's a little too naive or opinionated right now

DanielNoord:
> * infer the value of the generator so we do not raise if it's obvious it will always be possible to call next on it (Using next directly is reasonable to if the generator is defined in the same scope ?)

I'm going to think a bit about this and see if I can come up with something. On the other hand, isn't this also what a linter is for: catching code that might work now, but is not always safe in the future? Stuff like `unspecified-encoding` can also be redundant if you're only running your code on Linux.

Pierre-Sassoulas:
For some thing we also assume the user knows what he does, i.e. we rejected a check for dict access without using ``get`` (https://github.com/PyCQA/pylint/pull/6901). It can raise a ``KeyError`` if the key does not exists. To be fair, the implementer did not want to check that ``KeyError`` were caught, but to forbid direct access entirely, which admittedly is a lot more opinionated than what we're doing here with ``next``. But this feel like the same type of check and we did not create the ``unguarded-direct-dict-access`` check (maybe because it's not what the original implementer wanted and it would be a valuable check if someone want to implement it).

DanielNoord:
> For some thing we also assume the user knows what he does, i.e. we rejected a check for dict access without using `get` (#6901). It can raise a `KeyError` if the key does not exists. To be fair, the implementer did not want to check that `KeyError` were caught, but to forbid direct access entirely, which admittedly is a lot more opinionated than what we're doing here with `next`. But this feel like the same type of check and we did not create the `unguarded-direct-dict-access` check (maybe because it's not what the original implementer wanted and it would be a valuable check if someone want to implement it).

Yeah, I'm just wondering what options users have to signal that they know the risk of `StopIteration` and are willing to "take it".
1. One is obviously guarding with an `except`.
2. Another is guarding in all calling functions > We would raise here due to limitations in `astroid` and this not actually being safe
3. ...

Using a variable that's defined in the same scope seems like it should also be one, but I don't think `astroid` allows us to make sure that that variable won't be exhausted.

Pierre-Sassoulas:
> 3 ...

I'd say this one could be "not activating the extensions that check this" or "disabling the default check". I think if we want to minimize false positives against false negatives the former is better.

> Using a variable that's defined in the same scope seems like it should also be one, but I don't think astroid allows us to make sure that that variable won't be exhausted.

Can we infer the generator length and compare against the number of call to ``next`` ?

DanielNoord:
@jacobtylerwalls Do you have an opinion about this? I'm still not sure whether I think this should be turned on by default. Take for example the following hit on `sentry`:
https://github.com/getsentry/sentry/blob/master/src/sentry/db/postgres/exceptions.py#L29

This seems fine: there is an `__iter__` method and there is even an assertion that `tb_list` is an iterable. However, if that `tb_list` was an empty iterable `__iter__` would still raise `StopIteration`. Of course this is something that the developer probably knows about, but it is also a potential source of errors: thus cause for a linting warning.
I think the biggest issue with this message is that people will often struggle what to define as `default`, so that's why they don't do so. But just putting `None` there shows that you have at least thought about the potential cause of issues.

jacobtylerwalls:
Let's block this on an astroid Constraint PR that adds a not-empty constraint. That would deal with most of the false positives, right? Then, if there are still code paths that make an empty generator reachable by an unguarded next, that's fine to warn about, even if they're unlikely paths in working libraries. People will just disable the whole check if it's too picky, but I think they'll see the value for others and won't complain :D

pylint/checkers/stdlib.py
```diff
@@ -463,6 +463,13 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
                 ]
             },
         ),
+        "W1519": (
+            "Using next without explicitly specifying a default value or catching the StopIteration",
```
Pierre-Sassoulas, line 467:
```suggestion
            "Using next without specifying a default value or catching the StopIteration",
```
There's no implicit default, it's going to raise a StopIteration if you don't set the default right ?

pylint/checkers/stdlib.py
```diff
@@ -463,6 +463,13 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
...
+            "unguarded-next-without-default",
+            "Without a default value calls to next() can raise a StopIteration "
```
Pierre-Sassoulas, line 469:
```suggestion
            "Without a default value calls to next() will raise a ``StopIteration`` "
```

pylint/checkers/stdlib.py
```diff
@@ -463,6 +463,13 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
...
+            "exception. This exception should be caught or a default value should "
```
Pierre-Sassoulas, line 470:
```suggestion
            "exception when there's nothing to iterate anymore. It's generally not a trivial task to prove that it's possible to call next safely so this exception should be caught or a default value should "
```

tests/functional/u/unguarded_next_without_default.py
```diff
@@ -0,0 +1,46 @@
...
+except StopIteration:
+    pass
+
+redefined_next = next
+redefined_next(i for i in (1, 2))  # [unguarded-next-without-default]
+
+
+class MyClass:
+    def __next__(self):
+        return next(i for i in (1, 2))
```
Pierre-Sassoulas, line 37:
```suggestion
class MyClass:
    def __next__(self):
        return next(i for i in (1, 2))
        
    def expected_to_raise_stop_iteration(self):
        """Custom iterator function that will raise StopIteration by design.

        Raises:
            StopIteration: If the end of the sequence has been reached.
        """
        return next(i for i in (1, 2))        
```
Just a though. Maybe we don't need to check that it's a proper docstring in numpy style or whatever. If there's StopIteration in the docstring we do not raise.

Pierre-Sassoulas (changes requested):
Wow, this has been stale for a long time. 😄  Let's put this little wonder in 2.17 !

jacobtylerwalls:
I thought we identified a class of false positives we could address with a not-empty constraint in astroid?

Pierre-Sassoulas:
Astroid inferring capability with none check has been improved in 2.13. there's still a lot of warning in the primers, I did not check how many less compared to before. Do we want to check what is inside the iterator to raise less and more accurately ? (I.e. is this a errors message or a convention/warning message ?)

jacobtylerwalls:
Sorry for not fleshing out further. I was suggesting blocking this PR on a yet to be written astroid PR that would build on the last constraint PR (`is None`) by adding a new constraint (`is not empty`), and that way this would no longer be flagged in this PR:

```python
bunch = []
if bunch:
    next(bunch)  # just doing this once, and we just checked not-empty
```

***
Also, I think this PR should also special case the infinite iterator used here: https://github.com/nedbat/coveragepy/blob/b059a67fd1fe5d514f7c283f5ab99052e1cea15f/coverage/debug.py#L414

jacobtylerwalls:
Maybe it can be done inside pylint only, without another astroid PR, I don't know.

DanielNoord:
This would require some work in `astroid`. I'll draft this again!

DanielNoord:
Going to close this. I never had time to look into this further and the amount of false positives without changes to `astroid` is way too high.

Pierre-Sassoulas:
It's a pity, that would be a nice check :( But yeah, if you use next without default you probably thought about what you're doing thus a lot of false positives.

Issue #6993, opened 2022-06-21
no_self_use fails to catch methods with empty bodies


scop:
### Bug description

no_self_use does not catch methods that have empty bodies (just a docstring or a `pass`)

### Command used

With the test cases from https://github.com/PyCQA/pylint/pull/6994 added:

```shell
pytest tests/test_functional.py::test_functional[no_self_use]
```

### Pylint output

```shell
____________________________________________________ test_functional[no_self_use] ____________________________________________________

self = <pylint.testutils.lint_module_test.LintModuleTest object at 0x7f13d49a0910>

    def runTest(self) -> None:
>       self._runTest()
E       AssertionError: Wrong results for file "no_self_use":
E       
E       Expected in testdata:
E         29: no-self-use
E         34: no-self-use

pylint/testutils/lint_module_test.py:145: AssertionError
```

### Expected behavior

No errors.

### Pylint version

```shell
pylint 2.15.0-dev0
astroid 2.11.6
Python 3.10.0 (default, Oct  8 2021, 12:02:33) [GCC 9.3.0]
```

### Additional dependencies

https://github.com/PyCQA/pylint/pull/6994

Pierre-Sassoulas:
The discussion on this issue happened in #6994, if you care about the issue please vote [here](https://github.com/PyCQA/pylint/pull/6994#issuecomment-1321117560)

Pierre-Sassoulas:
Closing as won't do, but we'll reconsider later if a clear consensus emerge on what should denote abstractness in the python community. See discussion in linked PR for more details.

PR #6994, opened 2022-06-21
[no_self_use] Add false negative test cases involving empty method body

scop:
## Type of Changes

## Description

Refs #6993 (eventually hopefully closes it)

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1557,7 +1557,7 @@ class VariablesChecker(BaseChecker):
 
         # Don't check arguments of abstract methods or within an interface.
         is_method = node.is_method()
-        if is_method and node.is_abstract():
+        if is_method and node.is_abstract(pass_is_abstract=False):
             return
 
         global_names = _flattened_scope_names(node.nodes_of_class(nodes.Global))
--- a/pylint/extensions/docparams.py
+++ b/pylint/extensions/docparams.py
@@ -253,7 +253,9 @@ class DocstringParameterChecker(BaseChecker):
     def check_functiondef_returns(
         self, node: nodes.FunctionDef, node_doc: Docstring
     ) -> None:
-        if (not node_doc.supports_yields and node.is_generator()) or node.is_abstract():
+        if (not node_doc.supports_yields and node.is_generator()) or node.is_abstract(
+            pass_is_abstract=False
+        ):
             return
 
         return_nodes = node.nodes_of_class(astroid.Return)
@@ -265,7 +267,7 @@ class DocstringParameterChecker(BaseChecker):
     def check_functiondef_yields(
         self, node: nodes.FunctionDef, node_doc: Docstring
     ) -> None:
-        if not node_doc.supports_yields or node.is_abstract():
+        if not node_doc.supports_yields or node.is_abstract(pass_is_abstract=False):
             return
 
         if (
--- a/pylint/extensions/no_self_use.py
+++ b/pylint/extensions/no_self_use.py
@@ -40,9 +40,7 @@ class NoSelfUseChecker(BaseChecker):
         self._meth_could_be_func: bool | None = None
 
     def visit_name(self, node: nodes.Name) -> None:
-        """Check if the name handle an access to a class member
-        if so, register it.
-        """
+        """Check if the name handle access to a class member if so, register it."""
         if self._first_attrs and (
             node.name == self._first_attrs[-1] or not self._first_attrs[-1]
         ):
--- a/pylint/extensions/no_self_use.py
+++ b/pylint/extensions/no_self_use.py
@@ -84,7 +84,7 @@ class NoSelfUseChecker(BaseChecker):
                 and node.type == "method"
                 and node.name not in PYMETHODS
                 and not (
-                    node.is_abstract()
+                    node.is_abstract(pass_is_abstract=False)
                     or overrides_a_method(class_node, node.name)
                     or decorated_with_property(node)
                     or _has_bare_super_call(node)
--- a/tests/functional/ext/no_self_use/no_self_use.py
+++ b/tests/functional/ext/no_self_use/no_self_use.py
@@ -28,13 +28,20 @@ class Toto:
 
 class Base:
     """an abstract class"""
-
     def __init__(self):
         self.aaa = 2
 
     def check(self, arg):
         """an abstract method, could not be a function"""
         raise NotImplementedError
+
+    def pass_only(self):  # [no-self-use]
+        # non-docstring because having one affects the result here
+        # this method isn't a real method since it doesn't need self
+        pass
+
+    def docstring_only(self):  # [no-self-use]
+        """this method isn't a real method since it doesn't need self"""
 
 
 class Sub(Base):
--- a/tests/functional/ext/no_self_use/no_self_use.py
+++ b/tests/functional/ext/no_self_use/no_self_use.py
@@ -122,8 +122,8 @@ class Foo1(ABC):
     def b(self):
         raise NotImplementedError
 
-    def c(self):
-        pass  # pass counts as abstract
+    def c(self):  # [no-self-use]
+        pass  # pass NO LONGER counts as abstract, this is a very old pattern
 
 
 class Foo2(Protocol):
@@ -142,7 +142,7 @@ class Foo3:
     def a(self, var): ...
 
     def a(self, var):
-        pass
+        raise NotImplementedError
 
 
 class Foo4:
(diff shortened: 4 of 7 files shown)


scop:
I don't think an empty method body is an appropriate way to specify an abstract method. Raising `NotImplementedError`, or marking it up with one of the [`@abc.abstract*`](https://docs.python.org/3/library/abc.html#abc.abstractmethod) decorators would be.

DanielNoord:
```diff
diff --git a/doc/whatsnew/2/2.15/index.rst b/doc/whatsnew/2/2.15/index.rst
index 46fa82192..e6d7b433d 100644
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -62,6 +62,10 @@ Other bug fixes
 Other Changes
 =============
 
+* ``Pylint`` no longer considers function with a single ``pass`` statement to be abstract.
+  They should either be decorated with ``@abstractmethod`` or ``raise NotImplementedError``.
+
+  Closes #6993
 
 Internal changes
 ================
diff --git a/pylint/checkers/variables.py b/pylint/checkers/variables.py
index 7684d1ac2..572cb726c 100644
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1238,7 +1238,7 @@ class VariablesChecker(BaseChecker):
 
         # Don't check arguments of abstract methods or within an interface.
         is_method = node.is_method()
-        if is_method and node.is_abstract():
+        if is_method and node.is_abstract(pass_is_abstract=False):
             return
 
         global_names = _flattened_scope_names(node.nodes_of_class(nodes.Global))
diff --git a/pylint/extensions/docparams.py b/pylint/extensions/docparams.py
index 35c847e0b..a126a23bb 100644
--- a/pylint/extensions/docparams.py
+++ b/pylint/extensions/docparams.py
@@ -252,7 +252,7 @@ class DocstringParameterChecker(BaseChecker):
     def check_functiondef_returns(
         self, node: nodes.FunctionDef, node_doc: Docstring
     ) -> None:
-        if (not node_doc.supports_yields and node.is_generator()) or node.is_abstract():
+        if (not node_doc.supports_yields and node.is_generator()) or node.is_abstract(pass_is_abstract=False):
             return
 
         return_nodes = node.nodes_of_class(astroid.Return)
@@ -264,7 +264,7 @@ class DocstringParameterChecker(BaseChecker):
     def check_functiondef_yields(
         self, node: nodes.FunctionDef, node_doc: Docstring
     ) -> None:
-        if not node_doc.supports_yields or node.is_abstract():
+        if not node_doc.supports_yields or node.is_abstract(pass_is_abstract=False):
             return
 
         if (
@@ -640,7 +640,7 @@ class DocstringParameterChecker(BaseChecker):
 
         :param node: The node show the message on.
         """
-        if node.is_abstract():
+        if node.is_abstract(pass_is_abstract=False):
             try:
                 missing_excs.remove("NotImplementedError")
             except KeyError:
diff --git a/pylint/extensions/no_self_use.py b/pylint/extensions/no_self_use.py
index 4a20873c8..51604c74f 100644
--- a/pylint/extensions/no_self_use.py
+++ b/pylint/extensions/no_self_use.py
@@ -87,7 +87,7 @@ class NoSelfUseChecker(BaseChecker):
                 and node.type == "method"
                 and node.name not in PYMETHODS
                 and not (
-                    node.is_abstract()
+                    node.is_abstract(pass_is_abstract=False)
                     or overrides_a_method(class_node, node.name)
                     or decorated_with_property(node)
                     or _has_bare_super_call(node)
diff --git a/tests/functional/ext/no_self_use/no_self_use.py b/tests/functional/ext/no_self_use/no_self_use.py
index 577216568..c0dce9390 100644
--- a/tests/functional/ext/no_self_use/no_self_use.py
+++ b/tests/functional/ext/no_self_use/no_self_use.py
@@ -130,8 +130,8 @@ class Foo1(ABC):
     def b(self):
         raise NotImplementedError
 
-    def c(self):
-        pass  # pass counts as abstract
+    def c(self):  # [no-self-use]
+        pass  # pass NO LONGER counts as abstract, this is a very old pattern
 
 
 class Foo2(Protocol):
@@ -150,7 +150,7 @@ class Foo3:
     def a(self, var): ...
 
     def a(self, var):
-        pass
+        raise NotImplementedError
 
 
 class Foo4:
diff --git a/tests/functional/ext/no_self_use/no_self_use.txt b/tests/functional/ext/no_self_use/no_self_use.txt
index 9cf9a049b..c7aa96f41 100644
--- a/tests/functional/ext/no_self_use/no_self_use.txt
+++ b/tests/functional/ext/no_self_use/no_self_use.txt
@@ -1,3 +1,6 @@
 no-self-use:17:4:17:23:Toto.function_method:Method could be a function:INFERENCE
 no-self-use:25:4:25:35:Toto.async_function_method:Method could be a function:INFERENCE
-no-self-use:102:4:102:9:C.a:Method could be a function:INFERENCE
+no-self-use:29:4:29:17:Toto.pass_only:Method could be a function:INFERENCE
+no-self-use:34:4:34:22:Toto.docstring_only:Method could be a function:INFERENCE
+no-self-use:110:4:110:9:C.a:Method could be a function:INFERENCE
+no-self-use:133:4:133:9:Foo1.c:Method could be a function:INFERENCE
```

This would fix it.

However, this fails some of our tests as we start warning about missing docstrings and unused variables. I think it might make sense to keep this as is: `pass` is often used to quickly draw up a certain class and worry about implementation later. I think users might consider it annoying if `pylint` starts warning about this in the middel of their writing process.

scop:
I think I'm in disagreement; the same argument could be used to justify disabling pretty much every pylint check. If people are annoyed by pylint starting to warn in the middle of their writing process, an obvious solution would be to just not run it at a time when they don't want to hear what it has to say. If it runs automatically e.g. within an IDE, then great: the issues it flags serve as a TODO list.

DanielNoord:
> I think I'm in disagreement; the same argument could be used to justify disabling pretty much every pylint check. If people are annoyed by pylint starting to warn in the middle of their writing process, an obvious solution would be to just not run it at a time when they don't want to hear what it has to say. If it runs automatically e.g. within an IDE, then great: the issues it flags serve as a TODO list.

```python
class Fruit:
    def cook(temperature):
        raise NotImplementedError

class Apple(Fruit):
    def cook(temperature):
        print(f"I can be cooked at {temperature}!")

class Coconut(Fruit):
    def cook(temperature):
        pass
```

I really struggled to come up with a good example here, but aren't there cases where a method doesn't need to do anything for a subclass? I can't think of a good example right now, but this does seem like a pattern that could be written.
Although it probably calls for a refactor I don't think raising `unused-argument` on `Coconut.cook` makes sense.

scop:
I do agree such cases likely exist. But considering the mentioned "use `pass` to indicate worrying about implementation later" workflow, there's no way pylint can know if it's an instance of that, or if it's intentionally done like that "for good". In that sense it seems to make sense to me to flag the issue(s). Users can let pylint know they know better by adding the needed `# pylint: disable`s where appropriate.

Pierre-Sassoulas:
I'm +1 on this one as I think simply 'pass' is not appropriate for abstractness, as it's doing nothing and thus failing silently. If the goal was to signal abstractness there's two valid way. In Daniel example it would actually hide the fact that you can't cook a Coconut. Here's an example of pass I think make sense:

```python
class LivingThing:
    def make_sound(self):
        raise NotImplementedError

class Cow(LivingThing):
    def make_sound(self):
        print(f"Moooh. Moomoo.")

class Plant():
    def make_sound(self):
        pass
```
Because plant do not make any sound, so the implementation is actually "doing nothing".

I added the need decision label.

Pierre-Sassoulas:
Please vote on this comment if you have an opinion on this change:
- 👍 Let's raise no-self-use on empty method body
- 👎 Keep not raising on empty method body

Pierre-Sassoulas:
I rebased and applied Daniel's patch from above.

DanielNoord:
@Pierre-Sassoulas Is there any chance we revive this? Or should we close due to inactivity?

Pierre-Sassoulas:
Not sure about closing, maybe we stick a need take over label on it ?

DanielNoord:
Can we retriever CI?

Pierre-Sassoulas:
I think we can, but merging / rebasing on main seems warranted and would relaunch CI too ?

DanielNoord:
@Pierre-Sassoulas I guess we know why this was controversial again.. We can't really change the definition of "abstract" without making breaking changes somewhere.

DanielNoord:
I think the main issue is that the initial change makes sense but we would want to keep our definition of `abstract` consistent across all other checks, which then does create some issue..

Pierre-Sassoulas:
Let's close this as won't do and reconsider later if a clear consensus emerge on what should denote abstractness in the python community.

PR #6998, opened 2022-06-22
Fix homepage link

hugovk:
## Type of Changes

## Description

The "Homepage" link at https://pypi.org/project/pylint/ points to https://www.pylint.org/ which is down:

image

Instead, link to the docs page at https://pylint.pycqa.org/.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,14 +30,12 @@ classifiers =
     Topic :: Software Development :: Testing
 keywords = static code analysis linter python lint
 project_urls =
-    Homepage = https://pylint.pycqa.org/
+    Docs: User Guide = https://pylint.pycqa.org/en/latest/
     Source Code = https://github.com/PyCQA/pylint
-    What's New = https://pylint.pycqa.org/en/latest/whatsnew/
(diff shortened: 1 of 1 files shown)


DanielNoord:
@Pierre-Sassoulas Do you know something about this? I never asked but do you have any affiliation with `logilab`? I think the site was still maintained by them. Their own site seems to be up and running: https://www.logilab.fr.

Pierre-Sassoulas (approved):
> Do you know something about this?

No, I don't control this domain, I don't know who does.

> I never asked but do you have any affiliation with logilab?

No 😄 Apart from the fact that hippo91 and I are french like Logilab. PcManticore didn't either as far as I know.

> I think the site was still maintained by them. Their own site seems to be up and running.

I think we should rely on common open-source resources for everything at this point (Changing the link to readthedoc ✔️).

Pierre-Sassoulas (commented):
Some cleanup of existing links

Pierre-Sassoulas:
Thank you for your contribution @hugovk, and congratulation for becoming a pylint contributor :)

PR #7000, opened 2022-06-22
[primer] Refactor the primer to use 'pylint.message.Message'

Pierre-Sassoulas:
## Type of Changes

## Description

Refs #6984, blocked by #7077

### DIFF
--- a/pylint/testutils/_primer/primer_run_command.py
+++ b/pylint/testutils/_primer/primer_run_command.py
@@ -11,6 +11,7 @@ from io import StringIO
 from itertools import chain
 
 from pylint.lint import Run
+from pylint.message import Message
 from pylint.reporters import JSONReporter
 from pylint.testutils._primer.package_to_lint import PackageToLint
 from pylint.testutils._primer.primer_command import PackageMessages, PrimerCommand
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas:
@DanielNoord before I really investigate this, do you have any idea why the installation fail for this change ? (https://github.com/PyCQA/pylint/runs/7038428338?check_suite_focus=true) It seems strange that it fail at the install stage, but it's fairly consistent across runs.

DanielNoord:
I don't think it fails at the install stage. Not that that is also the run stage. Have you tried this locally?

pylint/testutils/_primer/primer_run_command.py
```diff
@@ -48,7 +47,13 @@ def run(self) -> None:
             "w",
             encoding="utf-8",
         ) as f:
-            json.dump(packages, f)
+            json.dump(
+                {
+                    p: [JSONReporter.serialize(m) for m in msgs]
```
DanielNoord, line 52: Is this serialisation even necessary? Aren't `TypedDict` dumpable and loadable by default?

Pierre-Sassoulas: It's not a typed dict at this point, it's the actual message because when we load we deserialize.

pylint/testutils/_primer/primer_run_command.py
```diff
@@ -24,20 +25,18 @@ def run(self) -> None:
...
         other_fatal_msgs = []
         for msg in chain.from_iterable(packages.values()):
-            if msg["type"] == "fatal":
-                # Remove the crash template location if we're running on GitHub.
-                # We were falsely getting "new" errors when the timestamp changed.
-                assert isinstance(msg["message"], str)
-                if GITHUB_CRASH_TEMPLATE_LOCATION in msg["message"]:
-                    msg["message"] = msg["message"].rsplit(CRASH_TEMPLATE_INTRO)[0]
-                if msg["symbol"] == "astroid-error":
+            if msg.category == "fatal":
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Yes, as per https://github.com/PyCQA/pylint/blob/dee2f1449139aa013906cec685c05a8c1d5e6360/pylint/reporters/json_reporter.py#L85

pylint/testutils/_primer/primer_run_command.py
Pierre-Sassoulas: We could use the TypedDict we introduced in #7077 yes, but ``Message`` is used everywhere in the code base.

Pierre-Sassoulas: And there an immutable ``location`` in it that I'd like to use later on.

DanielNoord: Ah, I missed the `JSONReporter` serialises all of its messages before printing anyway. That makes this necessary.

DanielNoord (approved):

PR #7007, opened 2022-06-23
Add `hidden-parent-method` message

### LINKED ISSUE #6992: Double inheritance bug not reported
### Bug description

class A:
   def close(self):
      print('A')

class B:
   def close(self):
      print('B')

class C(A, B):
   def close(self):
      print('C')
      super().close()

C().stop()

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
(nothing)
```

### Expected behavior

"B.close() is not called" message

### Pylint version

```shell
latest
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description
Adds new refactoring message `hidden-parent-method`.

When the behavior of a method happens to depend on the order of the direct parents of a class (left-to-right in `Class A(left, right)`), then the code can yield potentially surprising results by hiding the method on the subsequent parent: better to explicitly reimplement the behavior or to create a single ancestor with the desired behavior.

Fixes #6992


jacobtylerwalls:
This should probably be an extension!

doc/data/messages/o/order-dependent-resolution/bad.py
```diff
@@ -0,0 +1,14 @@
+class HarshParent:
+    @staticmethod
+    def scold():
+        return "Don't do that!"
```
Pierre-Sassoulas, line 4:
```suggestion
class EmotionalBlackmailingParent:
    @staticmethod
    def scold():
        return "If you do that, mommy won't love you anymore !"
```

This is not a real suggestion. Great example, the possibility are endless !! 😄

pylint/checkers/design_analysis.py
```diff
@@ -81,6 +86,18 @@
...
         "Used when an if statement contains too many boolean expressions.",
     ),
+    "R0917": (
+        "Method(s) defined on multiple direct parents without reimplementation: %s",
+        "order-dependent-resolution",
+        "Used when methods defined on multiple direct parents are not reimplemented, "
+        "leading to fragility where the order of parents dictates behavior.",
+    ),
+    "R0918": (
+        "super() called despite multiple direct parents defining %s",
```
Pierre-Sassoulas, line 96:
```suggestion
        "super() called despite %s defining %s",
```
...despite 'HarshParent', 'SomeClass' and  'KindParent' defining... The more info we add to the actual message the less confusing it is for someone who don't initially know what order dependent resolution is.

pylint/checkers/design_analysis.py
```diff
@@ -491,6 +513,72 @@ def leave_classdef(self, node: nodes.ClassDef) -> None:
...
+        ]
+        if missing_reimplementations:
+            self.add_message(
+                "order-dependent-resolution",
+                node=node,
+                args=(", ".join(sorted(f.name for f in missing_reimplementations)),),
+                confidence=INFERENCE,
+            )
+
+        for parent_meth in methods_defined_in_multiple_bases:
```
Pierre-Sassoulas, line 562: Now, we also have the meth user parent 😱 ?

```python
class MethParent:
    @staticmethod
    def scold():
        return "Don't do that, do meth."
```

tests/functional/s/super/super_checks.py
```diff
@@ -153,4 +153,4 @@ def method(self):
 class AlabamaCousin(Child, Niece):
     def method(self):
         print("AlabamaCousin")
-        super(Child, self).method()
+        super(Child, self).method()  # [order-dependent-super-resolution]
```
Pierre-Sassoulas, line 156: I'm not surprised the Alabama cousin would have issue.

pylint/checkers/design_analysis.py
```diff
@@ -491,6 +513,72 @@ def leave_classdef(self, node: nodes.ClassDef) -> None:
...
+            for m in methods_defined_in_multiple_bases
+            if m.name not in (f.name for f in node.mymethods())
+        ]
+        if missing_reimplementations:
+            self.add_message(
+                "order-dependent-resolution",
+                node=node,
+                args=(", ".join(sorted(f.name for f in missing_reimplementations)),),
+                confidence=INFERENCE,
+            )
```
Pierre-Sassoulas, line 560:
```suggestion
        if missing_reimplementations:
            functions = ", ".join(sorted(f.name for f in missing_reimplementations))
            classes = ", ".join(b.name for b in node.bases)
            self.add_message(
                "order-dependent-resolution",
                node=node,
                args=(classes, functions),
                confidence=INFERENCE,
            )
```
(Untested but it should work ?)

Pierre-Sassoulas (changes requested):
This look pretty good ! It's always safer to put a new check in extension, but I think this is a very valuable check so I think we could add it to the default checks, I'll add the need decision label regarding this.

jacobtylerwalls:
> This is unlikely to have been caused by your changes.

This was definitely caused by my changes 😄

Pierre-Sassoulas:
>This was definitely caused by my changes

Yeah I realized that and deleted the message 😄 

The primer seems to have a real problem now though.

jacobtylerwalls:
Well, we may be on the hook anyway for dealing with all-paths-delegating-to-super based on the last test case I thought of. Rats. Back to work in progress.

tests/functional/h/hidden_parent_method.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class Grandchild1(Child6, Child7):
+    def close(self):
+        """All paths delegate to super -- no warning"""
+        print("Grandchild")
+        super().close()
+
+
+class Granchild2(Child6, Child7):
```
jacobtylerwalls, line 84:
```suggestion
class Grandchild2(Child6, Child7):
```

DanielNoord:
@jacobtylerwalls Is this unfixable?

jacobtylerwalls:
Perhaps. Needs more analysis.

PR #7012, opened 2022-06-23
Prevent additional `unsupported-binary-operation` messages for metaclasses defining `__or__`

jacobtylerwalls:
## Type of Changes

## Description
Follow-up to #6664, see https://github.com/PyCQA/pylint/pull/6664#discussion_r904224083. A search over multiple ancestors was needed.

Could have been part of #6664 and fully described by the changelog there.


tests/functional/a/alternative/alternative_union_syntax_error.py
```diff
@@ -15,7 +15,7 @@
 lst = [typing.Dict[str, int] | None,]  # [unsupported-binary-operation]
 
 var1: typing.Dict[str, int | None]  # [unsupported-binary-operation]
-var2: int | str | None  # [unsupported-binary-operation]
+var2: int | str | None  # [unsupported-binary-operation, unsupported-binary-operation]
```
jacobtylerwalls, line 18: This seemed like a reasonable capture of a false negative.

pylint/checkers/typecheck.py
```diff
@@ -1934,23 +1934,35 @@ def _check_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> None
...
+                    left_obj = left_or_ancestor
+                try:
+                    attrs = left_obj.getattr("__or__")
+                    if self._includes_version_compatible_overload(attrs):
+                        return
+                    left_is_type = True
+                    break
+                except astroid.NotFoundError:
+                    left_is_type = True
+                    break
```
Pierre-Sassoulas, line 1949: We can create a function returning a boolean and taking ``left_node``, and ``"__or__"`` or ``right_node`` and ``"__ror__"`` as argument.

pylint/checkers/typecheck.py
```diff
@@ -1934,23 +1934,35 @@ def _check_unsupported_alternative_union_syntax(self, node: nodes.BinOp) -> None
...
+                    if self._includes_version_compatible_overload(attrs):
+                        return
+                    right_is_type = True
+                    break
+                except astroid.NotFoundError:
+                    right_is_type = True
+                    break
 
         if left_is_type or right_is_type:
             self.add_message("unsupported-binary-operation", args=msg, node=node)
```
Pierre-Sassoulas, line 1968:
```suggestion
            self.add_message("unsupported-binary-operation", args=msg, node=node, confidence=HIGH)
```

Pierre-Sassoulas (changes requested):

pylint/checkers/typecheck.py
jacobtylerwalls: It will need to be `INFERENCE` due to `object_type()`.

tests/functional/a/alternative/alternative_union_syntax_error.py
```diff
@@ -120,3 +120,11 @@ class WithHorrible(metaclass=HorribleMetaclass):
...
 
 class_list = [WithHorrible | DefaultMetaclass]
+
+class SecondLevelMetaclass(ForwardMetaclass):
+    pass
+
+class WithSecondLevel(metaclass=SecondLevelMetaclass):
+    pass
+
+class_list_with_second_level = [WithSecondLevel | DefaultMetaclass]
```
Pierre-Sassoulas, line 130:
```suggestion
class_list_with_second_level = [1 | "DefaultMetaclass"]
```
To cover the last line of ``_recursive_search_for_classdef_type`` ?

pylint/checkers/typecheck.py
```diff
@@ -1921,6 +1930,23 @@ def _includes_version_compatible_overload(self, attrs: list):
...
+            if is_classdef_type(node_or_ancestor):
+                node = node_or_ancestor
+            try:
+                attrs = node.getattr(operation)
+                if self._includes_version_compatible_overload(attrs):
+                    return VERSION_COMPATIBLE_OVERLOAD_SENTINEL
+                return True
+            except astroid.NotFoundError:
+                return True
+        return False
```
Pierre-Sassoulas, line 1948: It seems we never reach this line during the tests, but I don't know if the test case I suggested would work.

jacobtylerwalls: Mypy insisted on it, but it's unreachable, since the iterator cannot be empty and every path through the iterator has a return statement. I'll add a pragma.

pylint/checkers/typecheck.py
```diff
@@ -1921,6 +1930,23 @@ def _includes_version_compatible_overload(self, attrs: list):
         )
         return not is_py310_builtin or self._py310_plus
 
+    def _recursive_search_for_classdef_type(
+        self, node: nodes.ClassDef, operation: Literal["__or__", "__ror__"]
+    ) -> bool | VERSION_COMPATIBLE_OVERLOAD:
+        if not isinstance(node, nodes.ClassDef):
+            return False
+        for node_or_ancestor in itertools.chain((node,), node.ancestors()):
```
DanielNoord, line 1938: Maybe I'm not reading this correctly, but how can this `for` loop ever reach a second loop? Don't all paths end in a `return`?

pylint/checkers/typecheck.py
```diff
@@ -1921,6 +1929,19 @@ def _includes_version_compatible_overload(self, attrs: list):
...
+    ) -> bool | VERSION_COMPATIBLE_OVERLOAD:
+        if not isinstance(node, nodes.ClassDef):
+            return False
+        try:
+            attrs = node.getattr(operation)
+            if self._includes_version_compatible_overload(attrs):
+                return VERSION_COMPATIBLE_OVERLOAD_SENTINEL
+            return True
+        except astroid.NotFoundError:
+            return True
```
DanielNoord, line 1943:
```suggestion
        try:
            attrs = node.getattr(operation)
        except astroid.NotFoundError:
            return True
        if self._includes_version_compatible_overload(attrs):
            return VERSION_COMPATIBLE_OVERLOAD_SENTINEL
        return True
```

Pierre-Sassoulas (approved):

PR #7015, opened 2022-06-24
Revise documentation example for `bad-docstring-quotes`

julthep:
## Type of Changes

## Description
Revise documentation example for `bad-docstring-quotes`


DanielNoord (commented):
This is just a cosmetic change, which even worsens the page.
See:
https://pylint--7015.org.readthedocs.build/en/7015/user_guide/messages/convention/bad-docstring-quotes.html
vs.
https://pylint.pycqa.org/en/latest/user_guide/messages/convention/bad-docstring-quotes.html

The current indentation is probably better. Furthermore, the example shows that both 1 quotes and using `'` is considered incorrect.

Please focus on adding new examples or fixing current issues rather than changing stuff stylistically.

PR #7024, opened 2022-06-24
Add documentation examples for `invalid-envvar-default`

julthep:

## Type of Changes

## Description
Add documentation examples for `invalid-envvar-default`

Refs #5953

### DIFF
--- a/doc/data/messages/i/invalid-envvar-default/bad.py
+++ b/doc/data/messages/i/invalid-envvar-default/bad.py
@@ -1,0 +1,3 @@
+import os
+
+env = os.getenv('SECRET_KEY', 1)  # [invalid-envvar-default]
--- a/doc/data/messages/i/invalid-envvar-default/details.rst
+++ b/doc/data/messages/i/invalid-envvar-default/details.rst
@@ -1,1 +1,0 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
--- a/doc/data/messages/i/invalid-envvar-default/good.py
+++ b/doc/data/messages/i/invalid-envvar-default/good.py
@@ -1,1 +1,3 @@
-# This is a placeholder for correct code for this message.
+import os
+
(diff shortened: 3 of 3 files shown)


Pierre-Sassoulas:
I'm wondering if we're not doing mypy's job here (on a very very limited scope, by hardocding the expected type). Maybe something to remove entirely, what do you think @DanielNoord ?

DanielNoord:
> I'm wondering if we're not doing mypy's job here (on a very very limited scope, by hardocding the expected type). Maybe something to remove entirely, what do you think @DanielNoord ?

I think this is valid, see: https://docs.python.org/3/library/os.html#os.getenv.

The default should simply be a string as that is what `getenv` is supposed to return. It makes no sense to return anything else: `var = os.getenv("VAR",1)` would still requires `var = int(var)` in case the default isn't used.

Pierre-Sassoulas:
> The default should simply be a string as that is what getenv is supposed to return. It makes no sense to return anything else: var = os.getenv("VAR",1) would still requires var = int(var) in case the default isn't used.

Sure, but my point is that mypy does that generically for all functions (builtins or otherwise), why do we have a message for this particular argument of this particular function ? If we extend this reasoning we would create a new message for each argument of each builtin function.

DanielNoord:
> > The default should simply be a string as that is what getenv is supposed to return. It makes no sense to return anything else: var = os.getenv("VAR",1) would still requires var = int(var) in case the default isn't used.
> 
> Sure, but my point is that mypy does that generically for all functions (builtins or otherwise), why do we have a message for this particular argument of this particular function ? If we extend this reasoning we would create a new message for each argument of each builtin function.

Perhaps we should? 😄 In this case I don't think it is such a problem. There are no open issue about it and https://github.com/PyCQA/pylint/issues/3512 also indicates that there are no older complaints about it. It's an useful check which could warn people who are no using `mypy` about a potential "mistake".

Pierre-Sassoulas (approved):

PR #7029, opened 2022-06-24
Add documentation examples for `invalid-sequence-index`

julthep:

## Type of Changes

## Description
Add documentation examples for `invalid-sequence-index`

Refs #5953

### DIFF
--- a/doc/data/messages/i/invalid-sequence-index/bad.py
+++ b/doc/data/messages/i/invalid-sequence-index/bad.py
@@ -1,0 +1,4 @@
+fruit = ['apple', 'banana', 'orange']
+print(fruit[object])
+# but be careful with data[True], it will work since True is represnted as 1
+# and False as 0, so it will work
--- a/doc/data/messages/i/invalid-sequence-index/bad.py
+++ b/doc/data/messages/i/invalid-sequence-index/bad.py
@@ -1,4 +1,4 @@
 fruits = ['apple', 'banana', 'orange']
-print(fruits["apple"])  # [invalid-sequence-index]
+print(fruits['apple'])  # [invalid-sequence-index]
 # but be careful with data[True], it will work since True is represnted as 1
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (changes requested):

doc/data/messages/i/invalid-sequence-index/bad.py
```diff
@@ -0,0 +1,4 @@
+fruit = ['apple', 'banana', 'orange']
+print(fruit[object])  # [invalid-sequence-index]
```
Pierre-Sassoulas, line 2:
```suggestion
fruits = ['apple', 'banana', 'orange']
print(fruits["apple"])  # [invalid-sequence-index]
```
We should use an example without a catch imo.

julthep: Yes, agree with you.

doc/data/messages/i/invalid-sequence-index/bad.py
```diff
@@ -0,0 +1,4 @@
+fruits = ['apple', 'banana', 'orange']
+print(fruits['apple'])  # [invalid-sequence-index]
+# but be careful with data[True], it will work since True is represnted as 1
+# and False as 0, so it will work
```
Pierre-Sassoulas, line 4: You can create a details.rst to add this information is you want. If you think pylint should raise in this case (arguably yes, then it's a false negative that we could fix).

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas (approved):

doc/data/messages/i/invalid-sequence-index/bad.py
julthep: Okey, so let's move them to `details.rst`. Thank you very much.

PR #7033, opened 2022-06-24
Added documentation examples for `unhashable-dict-key`.

mbyrnepr2:
## Type of Changes

## Description

### DIFF
--- a/doc/data/messages/u/unhashable-dict-key/bad.py
+++ b/doc/data/messages/u/unhashable-dict-key/bad.py
@@ -1,1 +1,2 @@
-{1: 2}[[1]]  # [unhashable-dict-key]
+fruits_number = {"apple": 42}
+print(f"Number of apples : {fruits_number[['apple']]}")  # [unhashable-dict-key]
--- a/doc/data/messages/u/unhashable-dict-key/bad.py
+++ b/doc/data/messages/u/unhashable-dict-key/bad.py
@@ -1,2 +1,2 @@
-fruits_number = {"apple": 42}
-print(f"Number of apples : {fruits_number[['apple']]}")  # [unhashable-dict-key]
+# Print the number of apples:
+print({"apple": 42}[["apple"]])  # [unhashable-dict-key]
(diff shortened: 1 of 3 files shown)


mbyrnepr2:
Note perhaps this checker should be expanded to cover the following TypeError:
```python
{[1, 2, 3]: 4}
```

doc/data/messages/u/unhashable-dict-key/bad.py
```diff
@@ -0,0 +1 @@
+{1: 2}[[1]]  # [unhashable-dict-key]
```
Pierre-Sassoulas, line 1:
```suggestion
fruits_number = {"apple": 42}
print(f"Number of apples : {fruits_number[['apple']]}")  # [unhashable-dict-key]
```

doc/data/messages/u/unhashable-dict-key/good.py
```diff
@@ -1 +1 @@
-# This is a placeholder for correct code for this message.
+{1: 2}[1]
```
Pierre-Sassoulas, line 1:
```suggestion
fruits_number = {"apple": 42}
print(f"Number of apples : {fruits_number['apple']}")
```

Pierre-Sassoulas (changes requested):
I had a really hard time to understand the example with so much abstraction 😄

Pierre-Sassoulas:
>  perhaps this checker should be expanded to cover the following TypeError

I think this is a really neat proposal !

doc/data/messages/u/unhashable-dict-key/bad.py
```diff
@@ -0,0 +1,2 @@
+fruits_number = {"apple": 42}
+print(f"Number of apples : {fruits_number[['apple']]}")  # [unhashable-dict-key]
```
Pierre-Sassoulas, line 2:
```suggestion
print(f"Number of apples : {{'apple': 42}[['apple']]}")  # [unhashable-dict-key]
```
The message might only be raised when the dict is defined in place like this ?

mbyrnepr2: Looks like it yeah. I've updated the examples but no worries if you would like to change them again - I am not sure the more I think about them :D

Pierre-Sassoulas (approved):
Let's not overthink it 😄 I wonder how often this message is raised though, we're really careful by not using any inference on the content of the dict.

PR #7035, opened 2022-06-24
Avoid hangs on many-core Windows machines

### LINKED ISSUE #6965: pylint 2.7 hangs on many-core Windows machines
### Bug description

We use PyLint in Chromium and we have been switching to Python 3 and PyLint 2.7 recently. While testing on a many-core (96 logical processor) machine I found that PyLint reliably hangs. This is due to a known Python 3 bug with multi-processing on machines with more than 56 (possibly more than 63 or 64) logical processors. We first hit this bug last year (discussed in crbug.com/1190269) but only recently discovered that PyLint is vulnerable.

If we were running PyLint from the command line then we could just pass a -j value that is capped at 56 (or 63, or 64) but I have not been able to figure out how to cap the jobs value when invoking PyLint through lint.Run().

Internally PyLint should cap its parallelism until the Python 3 bug (not yet reported I believe, sorry) is fixed.

### Configuration

```ini
96 logical processor Windows machine
The version of Windows does not seem to matter. This only happens on Python 3 and it is believed to affect all versions of Python 3.
```

### Command used

```shell
I invoke pylint quite directly using "git cl presubmit --all --force" from Chromium. This ends up invoking many steps including pylint_main.py, like this:

  from pylint import lint  # pylint: disable=bad-option-value,import-outside-toplevel
  lint.Run(argv)

https://source.chromium.org/chromium/chromium/tools/depot_tools/+/main:pylint_main.py;l=17?q=pylint_main.py&sq=
```

### Pylint output

```shell
No output - it hangs.
```

### Expected behavior

I expect the PyLint run to complete.

### Pylint version

```shell
pylint 2.7
Python 3.8.10
```

### OS / Environment

Windows 10

### Additional dependencies

_No response_

randomascii:
Creating a multiprocessing Pool with too many processes can hit
ValueError exceptions or hangs or both. The number that counts as "too
many" depends on the Python version so this change uses 56 as a
guaranteed safe limit.

Note that this only avoids the issue if an explicit jobs count is not
passed in.

Closes #6965

### DIFF
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -97,3 +97,5 @@ Internal changes
 
 * Fixed an issue where many-core Windows machines (>~60 logical processors) would hang when
   using the default jobs count.
+
+  Closes #6965
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -94,8 +94,3 @@ Internal changes
   let the functional test fail with a default value.
 
   Refs #6891
-
-* Fixed an issue where many-core Windows machines (>~60 logical processors) would hang when
-  using the default jobs count.
-
-  Closes #6965
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -94,3 +94,6 @@ Internal changes
   let the functional test fail with a default value.
 
   Refs #6891
+
+* Fixed an issue where many-core Windows machines (>~60 logical processors) would hang when
+  using the default jobs count.
--- a/pylint/lint/run.py
+++ b/pylint/lint/run.py
@@ -82,7 +82,7 @@ def _cpu_count() -> int:
         cpu_count = multiprocessing.cpu_count()
     else:
         cpu_count = 1
-    if sys.platform == 'win32':
+    if sys.platform == "win32":
         # Using too many child processes in Python 3 hits either hangs or a
         # ValueError exception, and, has diminishing returns. Clamp to 56 to
         # give margin for error.
--- a/pylint/lint/run.py
+++ b/pylint/lint/run.py
@@ -190,11 +190,11 @@ group are mutually exclusive.",
                 linter.set_option("jobs", 1)
             elif linter.config.jobs == 0:
                 linter.config.jobs = _cpu_count()
-                if sys.platform == 'win32':
+                if sys.platform == "win32":
                     # Using too many child processes in Python 3 hits either hangs or a
                     # ValueError exception, and, has diminishing returns. Clamp to 56 to
                     # give margin for error.
-                    jobs = min(jobs, 56)
+                    min(jobs, 56)
 
         if self._output:
             try:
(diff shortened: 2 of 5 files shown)


doc/whatsnew/2/2.15/index.rst
```diff
@@ -94,3 +94,6 @@ Internal changes
   let the functional test fail with a default value.
 
   Refs #6891
+
+* Fixed an issue where many-core Windows machines (>~60 logical processors) would hang when
+  using the default jobs count.
```
Pierre-Sassoulas, line 99: We're going to backport that in 2.14.4 :)

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas:
Thank you for the research you've done @randomascii, much appreciated !

randomascii:
> Thank you for the research you've done @randomascii, much appreciated !

You're welcome. It was interesting, and the discussion on the bug was helpful for me.

pylint/lint/run.py
```diff
@@ -190,6 +190,11 @@ def __init__(
                 linter.set_option("jobs", 1)
             elif linter.config.jobs == 0:
                 linter.config.jobs = _cpu_count()
+                if sys.platform == "win32":
```
DanielNoord, line 193: Doesn't it make more sense to de-dent this one level and always running this? Perhaps even with a nice error message to `stderr` saying that we have limited the jobs count from ... to 56?

randomascii: It's a judgment call. The wpt project felt it made more sense to not override explicit settings, but doing so with a warning message also makes perfect sense. I don't have strong feelings either way.

DanielNoord: Hm, yeah perhaps that's indeed better. If we overwrite by default there would be no way for a user to overwrite it themselves.

But then I think it should be in the `_cpu_count()` function, as that is responsible for getting the "correct" number so this fix belongs to its responsibilities.

Could you also add a short note about this here:
https://github.com/PyCQA/pylint/blob/17c8f26f35fbf05d149da9719bab17a148f2e363/pylint/lint/base_options.py#L247

That will automatically make it show up in our settings documentation!

randomascii: Done.

pylint/lint/run.py
```diff
@@ -82,6 +82,11 @@ def _cpu_count() -> int:
         cpu_count = multiprocessing.cpu_count()
     else:
         cpu_count = 1
+    if sys.platform == "win32":
+        # Using too many child processes in Python 3 hits either hangs or a
+        # ValueError exception, and, has diminishing returns. Clamp to 56 to
+        # give margin for error.
+        cpu_count = min(cpu_count, 56)
```
DanielNoord, line 89:
```suggestion
        # See also https://github.com/python/cpython/issues/94242
        cpu_count = min(cpu_count, 56)
```

DanielNoord (approved):
Only need to move the changelog entry to `2.14.4`.

Thanks very much for all the work here @randomascii! Much appreciated!

pylint/lint/run.py
randomascii: Done.

randomascii:
I updated 2.15/index.rst and put the same content in 2.14/index.rst - I hope that is as desired.

I also made the requested comment change (I didn't see the button to do it from the pull request until I'd already done it on my machine).

DanielNoord (approved):
I move the changelog to the `2.14.4` section as that will be the release this will be available in. Should have been clearer about that!

Thanks a lot @randomascii 😄

PR #7037, opened 2022-06-25
Don't emit `modified-iterating-dict` when updating existing keys

### LINKED ISSUE #6179: Should E4702 `modified-iterating-dict` be triggered for updating existing keys?
### Question

With pylint 2.13, the following file:

```python
"Test module"


def test():
    "Test function"
    my_dict = {"a": 0, "b": 1}
    for key in my_dict:
        my_dict[key] = my_dict[key] + 1

    return my_dict
```

results in:

```
************* Module test
test.py:8:8: E4702: Iterated dict 'my_dict' is being modified inside for loop body, iterate through a copy of it instead. (modified-iterating-dict)
```

The error can be avoided by changing `for key in my_dict` to `for key in iter(my_dict)` (or `for key in list(my_dict)`; maybe the `iter()` version should still be an error).

@orSolocate I am using the "support question" issue template because I am a bit conflicted about if this is handled right and thought it would be good to have it at least captured in a closed issue if nothing else.

The example I gave above is a fairly common case, I think -- iterate over the keys of a dict and change the values associated with those keys. Should that case be flagged as an error? Is it an implementation quirk that it works fine in CPython (in principle reassigning to an existing key could result in a new entry internally and have similar problems to adding new keys)? Or should pylint be more careful to check for new items added to the dict?

The justification for the error is currently documented as "Emitted when items are added or removed to a dict being iterated through." If the existing behavior is okay, perhaps it should be "Emitted when items are added to, removed from, or reassigned to a dict being iterated through."

### Additional context

_No response_

jacobtylerwalls:
## Type of Changes

## Description

Closes #6179


Pierre-Sassoulas (approved):
Those two false positives disappearing in pandas warm my heart

PR #7041, opened 2022-06-25
Add documentation examples for `lost-exception`

julthep:

## Type of Changes

## Description
Add documentation examples for `lost-exception`

Refs #5953

### DIFF
--- a/doc/data/messages/l/lost-exception/bad.py
+++ b/doc/data/messages/l/lost-exception/bad.py
@@ -3,5 +3,5 @@ def fruit(value):
         return 2 / value
     except ZeroDivisionError:
         print('Connot divide zero.')
-    
-    return 0
+    finally:
+        return 0  # [lost-exception]
--- a/doc/data/messages/l/lost-exception/bad.py
+++ b/doc/data/messages/l/lost-exception/bad.py
@@ -1,7 +1,12 @@
-def fruit(value):
+class FasterThanTheSpeedOfLightError(ZeroDivisionError):
+    def __init__(self):
+        super().__init__("You can't go faster than the speed of light !")
+
+
+def calculate_speed(distance: float, time: float) -> float:
     try:
-        return 2 / value
-    except ZeroDivisionError:
-        print('Connot divide zero.')
+        return distance / time
+    except ZeroDivisionError as e:
+        raise FasterThanTheSpeedOfLightError() from e
     finally:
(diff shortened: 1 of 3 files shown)


doc/data/messages/l/lost-exception/bad.py
```diff
@@ -0,0 +1,7 @@
+def fruit(value):
+    try:
+        return 2 / value
+    except ZeroDivisionError:
+        print('Connot divide zero.')
+    finally:
+        return 0  # [lost-exception]
```
Pierre-Sassoulas, line 7:
```suggestion
def calculate_speed(distance: float, time: float) -> float:
    try:
        return distance / time
    except ZeroDivisionError:
        print("You can't go faster than the speed of light !")
    finally:
        return 0  # [lost-exception]
```

doc/data/messages/l/lost-exception/good.py
```diff
@@ -1 +1,7 @@
-# This is a placeholder for correct code for this message.
+def fruit(value):
+    try:
+        return 2 / value
+    except ZeroDivisionError:
+        print('Connot divide zero.')
+
+    return 0
```
Pierre-Sassoulas, line 7:
```suggestion
class FasterThanTheSpeedOfLight(ZeroDivisionError):
    def __init__(self):
        super(FasterThanTheSpeedOfLight, self).__init__(
            "You can't go faster than the speed of light !"
        )


def calculate_speed(distance: float, time: float) -> float:
    try:
        return distance / time
    except ZeroDivisionError as e:
        raise FasterThanTheSpeedOfLight() from e

```

Pierre-Sassoulas (changes requested):
I think this example make more sense semantically. Also the good.py example was a false negative imo. pylint does not see that the exception is lost, or does not warn (because it's very opinionated, sometime you want to actually lose the exception).

doc/data/messages/l/lost-exception/bad.py
```diff
@@ -0,0 +1,7 @@
+def fruit(value):
+    try:
+        return 2 / value
+    except ZeroDivisionError:
+        print('Connot divide zero.')
```
Pierre-Sassoulas, line 5: Here the issue is that the exception is not re-raised, so it's not swallowed and does not illustrate the issue.

Pierre-Sassoulas (approved):

PR #7042, opened 2022-06-25
Add documentation examples for `method-hidden`

julthep:

## Type of Changes

## Description
Add documentation examples for `method-hidden`

Refs #5953

### DIFF
--- a/doc/data/messages/m/method-hidden/bad.py
+++ b/doc/data/messages/m/method-hidden/bad.py
@@ -1,7 +1,6 @@
 class Fruit:
-    def __init__(self, worm):
-        if worm:
-            self.worm = worm
+    def __init__(self, vitamins):
+        self.vitamins = vitamins
 
-    def worm(self):  # [method-hidden]
+    def vitamins(self):  # [method-hidden]
         pass
--- a/doc/data/messages/m/method-hidden/bad.py
+++ b/doc/data/messages/m/method-hidden/bad.py
@@ -3,5 +3,5 @@ class Fruit:
         if worm:
             self.worm = worm
 
-    def bore(self):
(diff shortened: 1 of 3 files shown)


doc/data/messages/m/method-hidden/bad.py
```diff
@@ -0,0 +1,7 @@
+class Fruit:
+    def __init__(self, worm):
+        if worm:
+            self.worm = worm
+
+    def worm(self):  # [method-hidden]
+        pass
```
Pierre-Sassoulas, line 7:
```suggestion
    def __init__(self, vitamins):
        self.vitamins = vitamins

    def vitamins(self):  # [method-hidden]
        pass
```

doc/data/messages/m/method-hidden/good.py
```diff
@@ -1 +1,7 @@
-# This is a placeholder for correct code for this message.
+class Fruit:
+    def __init__(self, worm):
+        if worm:
+            self.worm = worm
+
+    def bore(self):
+        pass
```
Pierre-Sassoulas, line 7:
```suggestion
    def __init__(self, vitamins):
        self.vitamins = vitamins

    def antioxidants(self):
        pass
```

Pierre-Sassoulas (changes requested):
I'm wondering about the conditional (``if worms``), was it required ?

doc/data/messages/m/method-hidden/good.py
Pierre-Sassoulas: To give the reasoning on that one, ``bore`` should not be a function of the fruit, the insect is boring not the apple.

Pierre-Sassoulas (approved):

PR #7047, opened 2022-06-25
Add documentation examples for `non-parent-init-called`

julthep:

## Type of Changes

## Description
Add documentation examples for `non-parent-init-called`

Refs #5953

### DIFF
--- a/doc/data/messages/n/non-parent-init-called/bad.py
+++ b/doc/data/messages/n/non-parent-init-called/bad.py
@@ -1,13 +1,15 @@
-class Shape:
+class Animal:
     def __init__(self):
-        self.shape = True
+        self.is_multicellular = True
 
-class Rectangle(Shape):
+
+class Vertebrate(Animal):
     def __init__(self):
         super().__init__()
-        self.rectangle = True
+        self.has_vertebrae = True
 
-class Square(Rectangle):
+
+class Cat(Vertebrate):
     def __init__(self):
-        Shape.__init__(self)  # [non-parent-init-called]
-        self.square = True
+        Animal.__init__(self)  # [non-parent-init-called]
+        self.adorable = True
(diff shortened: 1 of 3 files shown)


doc/data/messages/n/non-parent-init-called/bad.py
```diff
@@ -0,0 +1,13 @@
...
+
+class Rectangle(Shape):
+    def __init__(self):
+        super().__init__()
+        self.rectangle = True
+
+class Square(Rectangle):
+    def __init__(self):
+        Shape.__init__(self)  # [non-parent-init-called]
+        self.square = True
```
Pierre-Sassoulas, line 13:
```suggestion
class Animal:
    def __init__(self):
        self.is_multicellular = True


class Vertebrate(Animal):
    def __init__(self):
        super().__init__()
        self.has_vertebrae = True


class Cat(Vertebrate):
    def __init__(self):
        Animal.__init__(self)  # [non-parent-init-called]
        self.adorable = True
```

doc/data/messages/n/non-parent-init-called/good.py
```diff
@@ -1 +1,13 @@
...
+
+class Rectangle(Shape):
+    def __init__(self):
+        super().__init__()
+        self.rectangle = True
+
+class Square(Rectangle):
+    def __init__(self):
+        super().__init__()
+        self.square = True
```
Pierre-Sassoulas, line 13:
```suggestion
class Animal:
    def __init__(self):
        self.is_multicellular = True


class Vertebrate(Animal):
    def __init__(self):
        super().__init__()
        self.has_vertebrae = True


class Cat(Vertebrate):
    def __init__(self):
        Vertebrate.__init__(self)  # [non-parent-init-called]
        self.is_adorable = True
```

Pierre-Sassoulas (changes requested):
A little nitpick because the base class can't be the more specialized one, Shape must have a lot of behavior that are specific to it and Square is "simple" with only one attribute.

Pierre-Sassoulas (approved):

Pierre-Sassoulas (commented):
Thank you for all the work you're doing right now to document pylint @julthep , impressive !

julthep:
> Thank you for all the work you're doing right now to document pylint @julthep , impressive !

You're welcome. Actually, not as impressive as all of you folks are doing wonderful work without breaks for this awesome project.

PR #7049, opened 2022-06-25
Add documentation examples for `not-async-context-manager`

julthep:

## Type of Changes

## Description
Add documentation examples for `not-async-context-manager`

Refs #5953


Pierre-Sassoulas (approved):
I think this is one of the example where foo does not hurt. It's going to complicate a lot of we want a lively example. Also beginners probably do not create a lot of async context manager.

julthep:
> I think this is one of the example where foo does not hurt. It's going to complicate a lot of we want a lively example. Also beginners probably do not create a lot of async context manager.

Yes, I think so. :)

PR #7050, opened 2022-06-25
[primer] Move `django` to new primer

jacobtylerwalls:
Tested on my fork. With adding `django`, the 3.7 primer job takes 32m, and the 3.10 job 28m.

### DIFF
--- a/tests/primer/packages_to_lint_batch_one.json
+++ b/tests/primer/packages_to_lint_batch_one.json
@@ -1,9 +1,4 @@
 {
-  "django": {
-    "branch": "main",
-    "directories": ["django"],
-    "url": "https://github.com/django/django.git"
-  },
   "keras": {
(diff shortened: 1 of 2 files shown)


DanielNoord (approved):

Pierre-Sassoulas (approved):

DanielNoord:
@jacobtylerwalls Recent job on `main` timed out. Shall we increase the timer to 60 minutes?

jacobtylerwalls:
Eventually, yes, so we can add more projects (assuming we don't fix pylint's parallel mode first). But now, I'm not sure. 2 days ago (and after this PR was merged) the 3.7 jobs were finishing in 28m. Now they are taking 44m. Isn't that a performance regression to look into instead? My guess is one of the commits on `astroid` main 2 days ago.

PR #7062, opened 2022-06-26
Add documentation examples for `raising-bad-type`

julthep:

## Type of Changes

## Description
Add documentation examples for `raising-bad-type`

Refs #5953

### DIFF
--- a/doc/data/messages/r/raising-bad-type/bad.py
+++ b/doc/data/messages/r/raising-bad-type/bad.py
@@ -1,0 +1,5 @@
+def div(x, y):
+    try:
+        return x / y
+    except ZeroDivisionError:
+        raise None  # [raising-bad-type]
--- a/doc/data/messages/r/raising-bad-type/bad.py
+++ b/doc/data/messages/r/raising-bad-type/bad.py
@@ -1,5 +1,10 @@
-def div(x, y):
+class FasterThanTheSpeedOfLightError(ZeroDivisionError):
+    def __init__(self):
+        super().__init__("You can't go faster than the speed of light !")
+
+
+def calculate_speed(distance: float, time: float) -> float:
     try:
-        return x / y
-    except ZeroDivisionError:
+        return distance / time
+    except ZeroDivisionError as e:
         raise None  # [raising-bad-type]
--- a/doc/data/messages/r/raising-bad-type/good.py
+++ b/doc/data/messages/r/raising-bad-type/good.py
@@ -1,1 +1,5 @@
-# This is a placeholder for correct code for this message.
+def div(x, y):
+    try:
+        return x / y
+    except ZeroDivisionError as e:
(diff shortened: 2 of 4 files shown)


pylint/checkers/exceptions.py
```diff
@@ -69,7 +69,7 @@ def _is_raising(body: list) -> bool:
     "E0702": (
         "Raising %s while only classes or instances are allowed",
         "raising-bad-type",
-        "Used when something which is neither a class nor an instance "
+        "Used when something which is neither a class, an instance nor a string "
```
Pierre-Sassoulas, line 72: I don't understand this change. I don't think you can raise a string ?
```
>>> raise "A string"
TypeError: exceptions must derive from BaseException
```

Pierre-Sassoulas (changes requested):

pylint/checkers/exceptions.py
julthep: This addition comes from [pylint-errors E0702](https://vald-phoenix.github.io/pylint-errors/plerr/errors/exceptions/E0702). Should it be reverted?

Pierre-Sassoulas: Yeap I think it's wrong.
```suggestion
        "Used when something which is neither a class nor an instance "
```

doc/data/messages/r/raising-bad-type/bad.py
```diff
@@ -0,0 +1,5 @@
+def div(x, y):
+    try:
+        return x / y
+    except ZeroDivisionError:
+        raise None  # [raising-bad-type]
```
Pierre-Sassoulas, line 5:
```suggestion
class FasterThanTheSpeedOfLightError(ZeroDivisionError):
    def __init__(self):
        super().__init__("You can't go faster than the speed of light !")


def calculate_speed(distance: float, time: float) -> float:
    try:
        return distance / time
    except ZeroDivisionError as e:
        raise None  # [raising-bad-type]
```

doc/data/messages/r/raising-bad-type/good.py
```diff
@@ -1 +1,5 @@
-# This is a placeholder for correct code for this message.
+def div(x, y):
+    try:
+        return x / y
+    except ZeroDivisionError as e:
+        raise ValueError from e
```
Pierre-Sassoulas, line 5:
```suggestion
class FasterThanTheSpeedOfLightError(ZeroDivisionError):
    def __init__(self):
        super().__init__("You can't go faster than the speed of light !")


def calculate_speed(distance: float, time: float) -> float:
    try:
        return distance / time
    except ZeroDivisionError as e:
        raise FasterThanTheSpeedOfLightError() from e
```

Pierre-Sassoulas (approved):

PR #7063, opened 2022-06-26
Add documentation examples for `redefined-outer-name`

julthep:

## Type of Changes

## Description
Add documentation examples for `redefined-outer-name`

Refs #5953

### DIFF
--- a/doc/data/messages/r/redefined-outer-name/bad.py
+++ b/doc/data/messages/r/redefined-outer-name/bad.py
@@ -1,0 +1,6 @@
+var = 0
+
+
+def add(x, y):
+    var = x + y  # [redefined-outer-name]
+    return var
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (changes requested):
I think we need to talk about pytest's fixture in the details.rst, this is a false positive that is really annoying and the message need to be disabled alongside pytest because not using fixtures in pytest makes the code a lot worst. There's probably a related issue too.

julthep:
> I think we need to talk about pytest's fixture in the details.rst, this is a false positive that is really annoying and the message need to be disabled alongside pytest because not using fixtures in pytest makes the code a lot worst. There's probably a related issue too.

I will revert `details.rst` file, where is the information should it refer to?

Pierre-Sassoulas:
> I will revert details.rst file, where is the information should it refer to?

I don't have anything particular in mind but a google search for ``pytest  redefined-outer-name pylint`` will probably return relevant discussions about it

Pierre-Sassoulas:
Closing as inactive and the conflict means it's been done in another MR.

PR #7074, opened 2022-06-26
Add documentation examples for `unneeded-not`

julthep:

## Type of Changes

## Description
Add documentation examples for `unneeded-not`

Refs #5953

### DIFF
--- a/doc/data/messages/u/unneeded-not/good.py
+++ b/doc/data/messages/u/unneeded-not/good.py
@@ -1,1 +1,2 @@
-# This is a placeholder for correct code for this message.
+if not input():
+    pass
--- a/doc/data/messages/u/unneeded-not/good.py
+++ b/doc/data/messages/u/unneeded-not/good.py
@@ -1,2 +1,2 @@
-if not input():
+if input():
     pass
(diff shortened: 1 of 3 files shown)


doc/data/messages/u/unneeded-not/good.py
```diff
@@ -1 +1,2 @@
-# This is a placeholder for correct code for this message.
+if not input():
```
Pierre-Sassoulas, line 1:
```suggestion
if input():
```

Pierre-Sassoulas (approved):

doc/data/messages/u/unneeded-not/good.py
julthep: I thought the good example intend to check for `not input()` but the bad example is mistakenly put duplicated `not'.

Pierre-Sassoulas: 👍 I think in this case doing the same thing in both example is possible and match the message name more.

julthep: 👍 Agree with you.

PR #7077, opened 2022-06-27
Add explicit serialization / deserialization to the ``JsonReporter``

Pierre-Sassoulas:
## Type of Changes

## Description

Necessary for Refs #7000


DanielNoord (commented):
@Pierre-Sassoulas What is the bug/issue we're trying to solve here?

Pierre-Sassoulas:
The goal is to be able to use the message object in the primer instead of a complicated to type dict from the JSON reporter. I identified the issue of confidence and abspath that were lost in the reporting thanks to this.

DanielNoord:
Should we consider adding those? Adding new keys to a JSON result shouldn't really break anything right?

Pierre-Sassoulas:
It's a breaking change depending on the implementation ? Say if someone loop on it and try to assign each value to something and it error on unknown value.

DanielNoord:
> It's a breaking change depending on the implementation ? Say if someone loop on it and try to assign each value to something and it error on unknown value.

Hmm yeah. We really should provide a `json-version-2`, but I haven't gotten to it yet.

Pierre-Sassoulas:
> We really should provide a json-version-2, but I haven't gotten to it yet.

Silly me, we just have to add a JsonReporter and deprecate the old one ?

DanielNoord:
> Silly me, we just have to add a JsonReporter and deprecate the old one ?

Yeah, but we also need to fix some of the issues that are currently in the beta of `3.0`. That's why I haven't gotten to it.

Pierre-Sassoulas:
Hmm ok. Seems we can deal with it later and merge this one then ?

DanielNoord:
> Hmm ok. Seems we can deal with it later and merge this one then ?

Well, this is a bit of a handicapped solution. I'd prefer tackling the actual problem, but if you don't have the time for that now I guess this should would be okay?

Pierre-Sassoulas:
I'm not trying to fix the missing confidence / abspath in the json right now, it's just an issue I realized existed while doing what's necessary to use ``pylint.message.Message`` instance in the primer.

pylint/reporters/json_reporter.py
```diff
@@ -24,24 +27,47 @@ class JSONReporter(BaseReporter):
...
-                "message-id": msg.msg_id,
-            }
-            for msg in self.messages
-        ]
+        json_dumpable = [self.serialize(message) for message in self.messages]
         print(json.dumps(json_dumpable, indent=4), file=self.out)
 
+    @staticmethod
+    def serialize(message: Message) -> dict[str, int | str | None]:
+        # TODO (3.0) add abs-path and confidence
```
DanielNoord, line 35:
```suggestion
        # TODO: 3.0: Add abs-path and confidence or a new JSONReporter
```

pylint/reporters/json_reporter.py
```diff
@@ -24,24 +27,47 @@ class JSONReporter(BaseReporter):
...
+        }
+
+    @staticmethod
+    def deserialize(message_as_json: dict) -> Message:
+        return Message(
+            msg_id=message_as_json["message-id"],
+            symbol=message_as_json["symbol"],
+            msg=message_as_json["message"],
+            location=MessageLocationTuple(
+                # TODO (3.0) abs-path is not available in json export
```
DanielNoord, line 57:
```suggestion
                # TODO: 3.0: Add abs-path and confidence or a new JSONReporter
```

pylint/reporters/json_reporter.py
```diff
@@ -24,24 +27,47 @@ class JSONReporter(BaseReporter):
...
-                "message": msg.msg or "",
-                "message-id": msg.msg_id,
-            }
-            for msg in self.messages
-        ]
+        json_dumpable = [self.serialize(message) for message in self.messages]
         print(json.dumps(json_dumpable, indent=4), file=self.out)
 
+    @staticmethod
+    def serialize(message: Message) -> dict[str, int | str | None]:
```
DanielNoord, line 34: Let's use a `TypedDict` here.

pylint/reporters/json_reporter.py
```diff
@@ -24,24 +27,47 @@ class JSONReporter(BaseReporter):
...
+            "endLine": message.end_line,
+            "endColumn": message.end_column,
+            "path": message.path,
+            "symbol": message.symbol,
+            "message": message.msg or "",
+            "message-id": message.msg_id,
+        }
+
+    @staticmethod
+    def deserialize(message_as_json: dict) -> Message:
```
DanielNoord, line 51: We can use the `TypedDict` here.

tests/reporters/unittest_json_reporter.py
```diff
@@ -98,3 +103,35 @@ def get_linter_result(score: bool, message: dict[str, Any]) -> list[dict[str, An
...
+                ),
+                msg="This is the actual message",
+                confidence=UNDEFINED,
+            ),
+            id="everything-defined",
+        )
+    ],
+)
+def test_serialize_deserialize(message):
+    # TODO (3.0): Add confidence handling, add path and abs path handling
```
DanielNoord, line 135:
```suggestion
    # TODO: 3.0: Add confidence handling, add path and abs path handling or a new JSONReporter
```

pylint/reporters/json_reporter.py
Pierre-Sassoulas: I think we should definitely have two JSON reporters in 2.x series,  and keep only the new one in 3.0.

pylint/reporters/json_reporter.py
Pierre-Sassoulas: Done in latest commit.

Pierre-Sassoulas:
@DanielNoord can you review this again please ? I added the typed dict :) I have multiple PR ready as follow-up but I don't want to put too much into this one.

pylint/reporters/json_reporter.py
```diff
@@ -48,6 +60,62 @@ def display_reports(self, layout: Section) -> None:
     def _display(self, layout: Section) -> None:
         """Do nothing."""
 
+    @staticmethod
+    def serialize(message: Message) -> Any:
```
DanielNoord, line 64:
```suggestion
    def serialize(message: Message) -> OldJsonExport:
```

pylint/reporters/json_reporter.py
```diff
@@ -48,6 +60,62 @@ def display_reports(self, layout: Section) -> None:
...
+        raise NotImplementedError
+
+    @staticmethod
+    def deserialize(message_as_json: Any) -> Message:
```
DanielNoord, line 68:
```suggestion
    def deserialize(message_as_json: OldJsonExport) -> Message:
```

Pierre-Sassoulas:
TypedDict is not available in some python version this is why i added it in the type checking guard.

DanielNoord (approved):
I pushed some changes which were a bit hard to suggest using the interface.

Let's make the types of the serialisation more explicit and add types of we need to.

DanielNoord:
> TypedDict is not available in some python version this is why i added it in the type checking guard.

Whenever typing stuff is unavailable on older versions we tend to put those under `sys.version` guards as then we'll remember to remove them when the guard is no longer needed.

pylint/reporters/json_reporter.py
Pierre-Sassoulas: It will break if we return something different for the New Json Reporter. Let's discuss this in the follow-up PR.

DanielNoord: Yeah, I understand. But that's why I think it should be added here now and then changed in the follow-up. To keep this PR self-contained this method should expect a `OldJsonExport`. In a follow-up we can then make it handle more different types.

PR #7088, opened 2022-06-29
Add partial typing to ``class_checker``

DanielNoord:

## Type of Changes

## Description


pylint/checkers/classes/class_checker.py
```diff
@@ -702,17 +711,19 @@ def _has_same_layout_slots(slots, assigned_value):
 }
 
 
-def _scope_default():
-    return collections.defaultdict(list)
+def _scope_default() -> defaultdict[str, list[_AccessNodes]]:
+    return defaultdict(list)
```
Pierre-Sassoulas, line 715: Should we remove this function ?

pylint/checkers/classes/class_checker.py
```diff
@@ -1930,7 +1943,14 @@ def _check_first_arg_for_type(self, node, metaclass=0):
...
             self.add_message("no-self-argument", node=node)
 
-    def _check_first_arg_config(self, first, config, node, message, method_name):
+    def _check_first_arg_config(
+        self,
+        first,
+        config: Sequence[str],
+        node: nodes.FunctionDef,
+        message: str,
+        method_name,
```
Pierre-Sassoulas, line 1952:
```suggestion
        method_name: str,
```

pylint/checkers/classes/class_checker.py
```diff
@@ -1930,7 +1943,14 @@ def _check_first_arg_for_type(self, node, metaclass=0):
...
+    ):
```
Pierre-Sassoulas, line 1953:
```suggestion
    ) -> None:
```

Pierre-Sassoulas (changes requested):

pylint/checkers/classes/class_checker.py
DanielNoord: We can't. Apparently you can't nest `defaultdicts`... Which is probably why this was done like this in the first place.

Pierre-Sassoulas: All right ! 

```suggestion
def _scope_default() -> defaultdict[str, list[_AccessNodes]]:
    # It's impossible to nest defaultdicts so we must use a function
    return defaultdict(list)
```

Pierre-Sassoulas (approved):

Issue #7003, opened 2022-06-22
Pylint no longer support separating options from positional arguments with --


phst:
### Bug description

With Pylint 2.12, `pylint -- file.py` works as expected, with Pylint 2.14 it raises an error like "pylint: error: Unrecognized option found:". I guess that's related to the argparse migration.

### Command used

```shell
pylint -- file.py
```

### Pylint output

```shell
pylint: error: Unrecognized option found:
```

### Expected behavior

The `--` should separate options from positional arguments. In this case the invocation should have identical results to `pylint file.py`.

### Pylint version

```shell
pylint 2.14.3
astroid 2.11.6
Python 3.9.12 (main, May  6 2022, 20:15:26) 
[GCC 11.2.0]
```

### OS / Environment

Debian GNU/Linux

Pierre-Sassoulas:
Thank you for opening the issue.

I added the discussion label because I don't know if we want to deviate from what argparse is doing and if it's worth it. In your example it's possible to use ``pylint file.py`` directly, right ? What is the standard we're talking about here, is it just something that optparse was doing or is it something well established for CLI ?

phst:
> Thank you for opening the issue.
> 
> I added the discussion label because I don't know if we want to deviate from what argparse is doing and if it's worth it.

Argpase supports this syntax. From https://docs.python.org/3/library/argparse.html#arguments-containing:

> you can insert the pseudo-argument `'--'` which tells [parse_args()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_args) that everything after that is a positional argument

So this should work out of the box, and I'm wondering why it doesn't.

> In your example it's possible to use `pylint file.py` directly, right ?

yes, the real use case is e.g. a script that wraps Pylint and receives arbitrary filenames, which might start with a dash.

> What is the standard we're talking about here, is it just something that optparse was doing or is it something well established for CLI ?

It's quite established, see e.g. the GNU Coding Standards (https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html) or POSIX (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02, guideline 10).

phst:
I don't think this is fixed.

```shell
$ pylint -- build.py 
************* Module --
--:1: [F0001(fatal), ] No module named --
```

That is, Pylint now thinks that it should lint a file named `--`. Instead, it should skip the `--` argument.

[...]

PR #7096, opened 2022-06-30
Fix handling of ``--`` as separator of positional arguments and flags

DanielNoord:

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/7003.

### DIFF
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -116,3 +116,13 @@ def test_short_verbose(capsys: CaptureFixture) -> None:
     Run([str(EMPTY_MODULE), "-v"], exit=False)
     output = capsys.readouterr()
     assert "Using config file" in output.err
+
+
+def test_argument_separator(capsys: CaptureFixture) -> None:
+    """Check that we support using '--' to separate argument types.
+
+    Reported in https://github.com/PyCQA/pylint/issues/7003.
+    """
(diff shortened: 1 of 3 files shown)


tests/config/test_config.py
```diff
@@ -116,3 +116,13 @@ def test_short_verbose(capsys: CaptureFixture) -> None:
...
     output = capsys.readouterr()
     assert "Using config file" in output.err
+
+
+def test_argument_separator(capsys: CaptureFixture) -> None:
+    """Check that we support using '--' to separate argument types.
+
+    Reported in https://github.com/PyCQA/pylint/issues/7003.
+    """
+    Run(["--", str(EMPTY_MODULE)], exit=False)
```
Pierre-Sassoulas, line 126: Should we test that option are taken into account for example here an init-hook that print ?

DanielNoord: We could, but I'm fairly confident this should work. The code that this is changing is a `pylint` hack on top of `argparse`. We parse arguments with `parse_known_args` so `argparse` doesn't complain about unknown options and leaves that to us.
I simply didn't know that this is allowed so this wasn't excluded from the list of unknown options. `argparse` probably does dit automatically in the normale `parse_args`.

Pierre-Sassoulas (approved):

PR #7099, opened 2022-06-30
Fix false negative for `nonlocal-without-binding`.

### LINKED ISSUE #6883: False negative | No binding for nonlocal found
````python
def outer():
  nonlocal a
  a = 1
outer()
````

- pylint gives a success
- but if you comment out `a = 1` pylint also finds the missing binding

_Originally posted by @da-dada in https://github.com/PyCQA/pylint/issues/6812#issuecomment-1146395637_

Further information:
```python
python example.py
  File "/Users/markbyrne/programming/example.py", line 2
    nonlocal a
    ^^^^^^^^^^
SyntaxError: no binding for nonlocal 'a' found
```

mbyrnepr2:
It was not emitted when the `nonlocal` name is assigned later on in the same scope.

Closes #6883

## Type of Changes

## Description

Refs #XXXX

Closes #XXXX

### DIFF
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -65,6 +65,10 @@ False negatives fixed
 
   Closes #6643
 
+* Emit ``nonlocal-without-binding`` when a nonlocal name has been assigned at a later point in the same scope. 
+
+  Closes #6883
+
 * Rename ``unhashable-dict-key`` to ``unhashable-member`` and emit when creating sets and dicts,
   not just when accessing dicts.
 
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -404,7 +404,10 @@ class BasicErrorChecker(_BasicChecker):
                 self.add_message("nonlocal-without-binding", args=(name,), node=node)
                 return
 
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas (approved):
LGTM, thank you !

Pierre-Sassoulas:
> This could go into a patch release by the way

I think it's a false negative and we only put false positive in patch release or we can get a vicious circle where the issues increase with a new patch instead of decreasing 😄

DanielNoord:
> > This could go into a patch release by the way
> 
> I think it's a false negative and we only put false positive in patch release or we can get a vicious circle where the issues increase with a new patch instead of decreasing 😄

Ah, my bad. Never mind!

mbyrnepr2:
> Sorry!
> 
> If you're up for it I would appreciate if we could add the `confidence` level to this message as well. But no worries if you don't have the time to do that right now!

I've added HIGH; is that the right one?

Pierre-Sassoulas:
It look like there's no inference or control flow involved so I'd say yes. (Maybe some control flow ?)

DanielNoord:
I double checked. There is no control flow (it's a very basic check) so this is indeed correct!

DanielNoord (approved):

PR #7102, opened 2022-06-30
doc(fail-under): improve description

scop:
## Type of Changes

## Description

To exceed a threshold here sonds to me as if the score would have to be _above_ the threshold, which is the opposite of what actually happens. Use "maintain" instead of "exceed" to "fix" the description. Another possibility would be to "meet" the threshold, or to rephrase the description altogether.

### DIFF
--- a/doc/user_guide/configuration/all-options.rst
+++ b/doc/user_guide/configuration/all-options.rst
@@ -78,7 +78,7 @@ Standard Checkers
 
 --fail-under
 """"""""""""
-*Specify a score threshold to be exceeded before program exits with error.*
+*Specify a score threshold under which the program will exit with error.*
 
 **Default:**  ``10``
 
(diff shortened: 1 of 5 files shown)


mbyrnepr2:
'Specify a score threshold above which Pylint will succeed'?

I think to exceed a threshold is a logical statement(the beginning of something), but in this case, 'error' should be replaced with 'succeed' ?

scop:
While both suggestions above might be technically correct, I think they're still kind of weird right next to the option which is named `fail-under`, as they describe the conditions under which we'd succeed, not fail as the option name says.

Pierre-Sassoulas:
How about ``Specify a score threshold under which the program will exits with error`` ?

scop:
Works for me, changed to that (with `s/will exits/will exit/`).

Pierre-Sassoulas (approved):
Thank you for the clarification, always useful 👍

Issue #7094, opened 2022-06-30
`--ignore-paths` and escaping of regex special characters


stanislavlevin:
### Bug description

It's impossible to escape special  regex characters in option `--ignore-paths` for unix paths.

The basic example is how to escape the dot (`.`).

```
(pylint) [builder@localhost test]$ python -m pylint --recursive=y --ignore-paths='^\.tox/.*' .
> /usr/src/pyproject_installer/.tox/pylint/lib/python3/site-packages/pylint/config/argument.py(115)_regexp_paths_csv_transfomer()
-> patterns.append(
(Pdb) pattern
'^\\.tox/.*'
(Pdb) l
110     def _regexp_paths_csv_transfomer(value: str) -> Sequence[Pattern[str]]:
111         """Transforms a comma separated list of regular expressions paths."""
112         patterns: list[Pattern[str]] = []
113         breakpoint()
114         for pattern in _csv_transformer(value):
115  ->         patterns.append(
116                 re.compile(
117                     str(pathlib.PureWindowsPath(pattern)).replace("\\", "\\\\")
118                     + "|"
119                     + pathlib.PureWindowsPath(pattern).as_posix()
120                 )
(Pdb) pathlib.PureWindowsPath(pattern).as_posix()
'^/.tox/.*'
(Pdb) pathlib.PureWindowsPath(pattern)
PureWindowsPath('^/.tox/.*')
```

Thereby, assumption `regex pattern => PureWindowsPath => regex pattern` is wrong and who knows where it will shoot.

### Command used

```shell
mkdir .tox && echo 'import bar' > .tox/foo.py
python -m pylint --recursive=y --ignore-paths='^\.tox/.*' .
```

### Pylint output

```shell
(pylint) [builder@localhost test]$ python -m pylint --recursive=y --ignore-paths='^\.tox/.*' .
************* Module foo
.tox/foo.py:1:0: C0114: Missing module docstring (missing-module-docstring)
.tox/foo.py:1:0: C0104: Disallowed name "foo" (disallowed-name)
.tox/foo.py:1:0: E0401: Unable to import 'bar' (import-error)
.tox/foo.py:1:0: W0611: Unused import bar (unused-import)

------------------------------------------------------------------
Your code has been rated at 0.00/10 (previous run: 0.00/10, +0.00)
```

### Expected behavior

According to documentation:
```
--ignore-paths
[...]

Issue #7089, opened 2022-06-29
regex for ignoring hidden directories doesn't work as expected


tbennett0:
### Bug description

In running `pylint --recursive y .` to ignore hidden directories using 2.14.4, I discovered the following pattern works:
```
[tool.pylint.MASTER]
ignore-paths = [
  "^.local/.*$"
]
```

but this one doesn't:
```
[tool.pylint.MASTER]
ignore-paths = [
  "^\\.local/.*$"
]
```

I would expect that the regex would see `.` as any character, so it matches the first, but explicitly using an escaped period that should also match doesn't cause that path to be ignored.

### Configuration

```ini
% cat .local/test.py
import blam
```

### Command used

```shell
% pylint --recursive=y .
```

### Pylint output

```shell
************* Module test
.local/test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
.local/test.py:1:0: E0401: Unable to import 'blam' (import-error)
.local/test.py:1:0: W0611: Unused import blam (unused-import)
```

### Expected behavior

No output on both examples.

### Pylint version

```shell
pylint 2.14.4
astroid 2.11.6
Python 3.9.12 (main, Jun 27 2022, 13:42:07)
[Clang 13.1.6 (clang-1316.0.21.2.5)]
```

matusvalo:
@tbennett0 could you verify the issue with the main branch whether it fixes the issue?

DanielNoord:
@matusvalo `2.14.4` was just released, which included your fix for #6964. So this seems to be a regression after that fix was applied.

matusvalo:
OK I have misread the issue. So I need to fix the fix of the fix :-/. OK I am sorry for that. Doing recursive linting seems to be not so trivial as looked in the first place :-/

DanielNoord:
Ah yeah. This is a duplicate of https://github.com/PyCQA/pylint/issues/5398.

matusvalo:
OK so I am closing it as duplicate.

Issue #5398, opened 2021-11-25
Seems like 2.12.1 ignoring "ignore-paths" in config


zeloras:
### Bug description

Good day/morning/evening!

Today i've update pylint to version 2.12.1 and for now that version ignoring ignore-paths=.*/migrations/.*\.py,
in my .pylintrc config. But in version 2.11.1 everything is working correct

### Configuration

```ini
[MASTER]
ignore-paths=.*/migrations/.*\.py,.*/urls.py
disable=R0903,C0103,E0110,W0511,R0801,E1101
max-line-length=120
max-module-lines=800

[DESIGN]
max-attributes=30
max-parents=30
max-public-methods=50
```

### Command used

```shell
pylint --rcfile=.pylintrc apps
```

### Pylint output

```shell
************* Module apps.accounts.migrations.0001_initial
apps/accounts/migrations/0001_initial.py:27:0: C0301: Line too long (196/120) (line-too-long)
apps/accounts/migrations/0001_initial.py:28:0: C0301: Line too long (329/120) (line-too-long)
apps/accounts/migrations/0001_initial.py:29:0: C0301: Line too long (165/120) (line-too-long)
apps/accounts/migrations/0001_initial.py:30:0: C0301: Line too long (203/120) (line-too-long)
apps/accounts/migrations/0001_initial.py:38:0: C0301: Line too long (255/120) (line-too-long)
apps/accounts/migrations/0001_initial.py:45:0: C0301: Line too long (266/120) (line-too-long)
apps/accounts/migrations/0001_initial.py:46:0: C0301: Line too long (229/120) (line-too-long)
apps/accounts/migrations/0001_initial.py:1:0: C0114: Missing module docstring (missing-module-docstring)
apps/accounts/migrations/0001_initial.py:12:0: C0115: Missing class docstring (missing-class-docstring)
apps/accounts/migrations/0001_initial.py:9:0: C0411: standard import "import uuid" should be placed before "from django.conf import settings" (wrong-import-order)
************* Module apps.accounts.migrations.0002_alter_pacmanuser_last_login
apps/accounts/migrations/0002_alter_pacmanuser_last_login.py:1:0: C0114: Missing module docstring (missing-module-docstring)
apps/accounts/migrations/0002_alter_pacmanuser_last_login.py:6:0: C0115: Missing class docstring (missing-class-docstring)
************* Module apps.accounts.migrations.0003_alter_pacmanuser_username
[...]

PR #7103, opened 2022-06-30
Update docs to explain escape character not valid

tbennett0:
## Type of Changes

## Description

This just adds the note from #5415 to docs to explain that `\` is not valid in ignore-paths per https://github.com/PyCQA/pylint/issues/5398#issuecomment-981156554 because it also currently is needed to support Windows path normalization. This will help avoid additional issues being created (like the one I created in #7089 and also someone else in #7094) to reduce the burden on pylint maintainers/developers.

### DIFF
--- a/doc/user_guide/configuration/all-options.rst
+++ b/doc/user_guide/configuration/all-options.rst
@@ -99,7 +99,7 @@ Standard Checkers
 
 --ignore-paths
 """"""""""""""
-*Add files or directories matching the regex patterns to the ignore-list. The regex matches against paths and can be in Posix or Windows format.*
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (commented):
Thank you for contributing to pylint ! Indeed we should have documented this better. I think the file you modified is autogenerated, could you search for the Python file with the same text and modify this one instead, please ?

Pierre-Sassoulas (approved):
Thanks a lot, and congratulation on becoming a pylint contributor :) !

tbennett0:
@Pierre-Sassoulas it looks the the auto-upgrading tools rst doc removed my escape of the `\` and now it doesn't render correctly in docs:

image

What's the best way around that? Do I need to escape it again?

Pierre-Sassoulas:
Does double escape work ? ``\\`` (sometime even ``\\\\``)

PR #7113, opened 2022-07-02
Add regression test for #1667

### LINKED ISSUE #1667: ``ImportError`` when using PEP 420 namespaces
### Steps to reproduce
1. git clone http://forge.softwareheritage.org/source/swh-core.git
2. export PYTHONPATH=$(pwd)/swh-core
3. pylint3 swh.core

### Current behavior

`ImportError: Unable to find module for /tmp/swh-core/swh/core/api.py in /tmp,`

(or equivalent)

### Expected behavior

Should be able to walk the module and pylint it.
Note that both "import swh.core" and manually pylinting the specific module that pylint claims is not loadable (`pylint3 swh.core.api`) work just fine.

I've given more details about this issue on the code-quality at https://mail.python.org/pipermail/code-quality/2017-September/000937.html (no answer yet at the time of filing this).

I've seen there are a number of PEP 420 related issues, but as far as I can tell they are either supposed to be already fixed in the version of pylint I'm using, or the fail in different ways (e.g., with a proper "cannot load module" message instead of a runaway ImportError exception like in my case).

### pylint --version output

pylint3 1.7.2, 
astroid 1.5.3
Python 3.5.4 (default, Aug 12 2017, 14:08:14) 
[GCC 7.1.0]

jacobtylerwalls:
## Type of Changes

## Description
Closes #1667 
Depends on astroid 2.12 (not one of the patch versions)


Pierre-Sassoulas (approved):
That test with ``tempdir`` / ``create_files`` is pretty straightforward. It's very easy to create "complicated" example without resorting to an input/data/regr_test_data directory. (I.e. PR is so clean it makes me want to refactor 30% of the tests 😄 )

PR #7115, opened 2022-07-03
Expose types

### LINKED ISSUE #2079: Annotate pylint code with type annotations
Since we're on Python 3, we can use annotations for pylint.

### LINKED ISSUE #5488: Expose types
### Current problem

Pylint support plugins which might want to type-check their code. Therefore it would be great, if pylint could expose it's types.

### Desired solution

Add missing types and add a `py.typed` file. Else add dedicated stubs files.

### Additional context

_No response_

jacobtylerwalls:
## Type of Changes

## Description
Closes #5488

Edit by @DanielNoord: This can now also close #2079 as the typing is now (almost) complete


Pierre-Sassoulas (approved):

DanielNoord:
I'm not sure if we should do this. As long as `astroid` can't pass `mypy` I'm not truly comfortable that all of our typing is correct. Might be better to wait until that has been finished?

jacobtylerwalls:
> I'm not sure if we should do this. As long as astroid can't pass mypy I'm not truly comfortable that all of our typing is correct. Might be better to wait until that has been finished?

I hear that, but it's correct for the current version, right? If future versions of `astroid` change, then future versions of `pylint` change, but that's not an issue for people developing against the current version of pylint.

DanielNoord:
> I hear that, but it's correct for the current version, right? If future versions of `astroid` change, then future versions of `pylint` change, but that's not an issue for people developing against the current version of pylint.

Well, we think the typing is correct. But, to give an example, we also run into the issue described in https://github.com/PyCQA/astroid/issues/1680 in `pylint` as we use it at least once. By adding `py.typed` we signal that we think that our current typing is at least correct and can only be expanded to include additional signatures, but I wouldn't be 100% that it is actually correct.

Pierre-Sassoulas:
Sorry I approved without giving my reasonning. I thought that astroid had to be typed before pylint too, but there's some level of separation between pylint and astroid, and most pylint API do not expose astroid's API, so it's probably correct enough and useful enough. Also we'll get issue for the part that are used, maybe even contributions.

jacobtylerwalls:
> By adding py.typed we signal that we think that our current typing is at least correct and can only be expanded to include additional signatures, but I wouldn't be 100% that it is actually correct.

Until you mentioned PyCQA/astroid#1680 I would have said we are confident that it is correct _given current state of `astroid`, which is all we can ever promise._ (Even if we know `astroid` will keep changing.) But PyCQA/astroid#1680 is a good point -- we know that's wrong. So I'm happy to postpone this.

KotlinIsland:
Don't forget to add the `py.typed` to the `pylint` entry for `tool.setuptools.package-data` in `pyproject.toml`

DanielNoord:
Superseded by https://github.com/PyCQA/pylint/pull/8402

PR #7116, opened 2022-07-03
Add ``invalid-name`` check for ``TypeAlias`` names

### LINKED ISSUE #7081: TypeAlias not conforming to PascalCase to throw invalid-name
### Current problem

Python `typing.TypeAlias` are an alias to a type (as we all know).  Thus, I think `TypeAlias` should have to follow type naming conventions (e.g. PascalCase for classes, unless a primitive).

In other words, we should see `TypeAlias` really as not much different from a novel `class` in regards to naming conventions.  I think pylint is the tool to help enforce this!

### Desired solution

Please see the below code snippet

```python

from threading import Thread  # Placeholder class
from typing import TypeAlias

ThreadAlias: TypeAlias = Thread  # Say nothing here
THREAD_ALIAS: TypeAlias = Thread  # Throw invalid-name
# C0103: Type alias "THREAD_ALIAS" doesn't conform to PascalCase naming style (invalid-name)
```

To restate, I think pylint should error on `CLASS_NAME_ALIAS` with `invalid-name`.

### Additional context

See the TypeAlias PEP: https://peps.python.org/pep-0613/#error-messaging

They use `ClassName`, not `CLASS_NAME`.

DanielNoord:

## Type of Changes

## Description

Closes #7081.

### DIFF
--- a/doc/data/messages/i/invalid-name/details.rst
+++ b/doc/data/messages/i/invalid-name/details.rst
@@ -29,6 +29,8 @@ name is found in, and not the type of object assigned.
 +--------------------+---------------------------------------------------------------------------------------------------+
 | ``typevar``        | Type variable declared with ``TypeVar``.                                                          |
 +--------------------+---------------------------------------------------------------------------------------------------+
+| ``typealias``      | Type alias declared with ``TypeAlias`` or assignments of ``Union``.                               |
++--------------------+---------------------------------------------------------------------------------------------------+
 
 Default behavior
 ~~~~~~~~~~~~~~~~
@@ -80,12 +82,18 @@ The pattern can be overwritten with the options discussed below.
 
 The following type of names are checked with a predefined pattern:
 
-+--------------------+---------------------------------------------------+------------------------------------------------------------+
-| Name type          | Good names                                        | Bad names                                                  |
-+====================+===================================================+============================================================+
-| ``typevar``        | ``T``, ``_CallableT``, ``_T_co``, ``AnyStr``,     | ``DICT_T``, ``CALLABLE_T``, ``ENUM_T``, ``DeviceType``,    |
-|                    | ``DeviceTypeT``, ``IPAddressT``                   | ``_StrType``, ``TAnyStr``                                  |
-+--------------------+---------------------------------------------------+------------------------------------------------------------+
++--------------------+-------------------------------------------------------+------------------------------------------------------------+
+| Name type          | Good names                                            | Bad names                                                  |
++====================+=======================================================+============================================================+
+| ``typevar``        | ``T``, ``_CallableT``, ``_T_co``, ``AnyStr``,         | ``DICT_T``, ``CALLABLE_T``, ``ENUM_T``, ``DeviceType``,    |
+|                    | ``DeviceTypeT``, ``IPAddressT``                       | ``_StrType``                                               |
++--------------------+-------------------------------------------------------+------------------------------------------------------------+
+| ``typealias``      | ``GoodName``, ``_GoodName``, ``IPAddressType`` and    | ``BadNameT``, ``badName``, ``TBadName``, ``TypeBadName``   |
+|                    |  other PascalCase variants that don't start with ``T``|                                                            |
+|                    |  or ``Type``. This is to distinguish them from        |                                                            |
+|                    |  ``typevars``. Note that ``TopName`` is allowed but   |                                                            |
+|                    |  ``TTopName`` isn't.                                  |                                                            |
++--------------------+-------------------------------------------------------+------------------------------------------------------------+
 
 Custom regular expressions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -123,6 +131,8 @@ expression will lead to an instance of ``invalid-name``.
 .. option:: --inlinevar-rgx=<regex>
 
 .. option:: --typevar-rgx=<regex>
+
+.. option:: --typealias-rgx=<regex>
 
 Multiple naming styles for custom regular expressions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/pylint/checkers/base/name_checker/checker.py
+++ b/pylint/checkers/base/name_checker/checker.py
@@ -40,7 +40,8 @@ _BadNamesTuple = Tuple[nodes.NodeNG, str, str, interfaces.Confidence]
 DEFAULT_PATTERNS = {
     "typevar": re.compile(
         r"^_{0,2}(?!T[A-Z])(?:[A-Z]+|(?:[A-Z]+[a-z]+)+T?(?<!Type))(?:_co(?:ntra)?)?$"
-    )
+    ),
+    "typealias": re.compile(r"^_{0,2}(?!T[A-Z]|Type)[A-Z]+[a-z]+(?:[A-Z][a-z]+)*$"),
 }
 
 BUILTIN_PROPERTY = "builtins.property"
@@ -400,7 +401,9 @@ class NameChecker(_BasicChecker):
         "typevar-double-variance",
         "typevar-name-mismatch",
     )
-    def visit_assignname(self, node: nodes.AssignName) -> None:
+    def visit_assignname(  # pylint: disable=too-many-branches
+        self, node: nodes.AssignName
+    ) -> None:
         """Check module level assigned names."""
         frame = node.frame(future=True)
         assign_type = node.assign_type()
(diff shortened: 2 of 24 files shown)


jamesbraza (dismissed):
I think you've done an excellent job @DanielNoord!  Cheers!

DanielNoord:
@cdce8p Just checking in. Do you want to review this before we merge this? Or can we merge without a review by you.

doc/data/messages/i/invalid-name/details.rst
```diff
@@ -29,6 +29,8 @@ name is found in, and not the type of object assigned.
 +--------------------+---------------------------------------------------------------------------------------------------+
 | ``typevar``        | Type variable declared with ``TypeVar``.                                                          |
 +--------------------+---------------------------------------------------------------------------------------------------+
+| ``typealias``      | Type alias declared with ``TypeAlias`` or assignments of ``Union`` and ``Callable``.              |
```
cdce8p, line 32: `Callable` isn't used

doc/whatsnew/fragments/7801.new_check
```diff
@@ -0,0 +1,5 @@
+Added ``invalid-name`` check for ``TypeAlias`` names. This also includes assignments
+using ``Union`` and ``Callable``.
```
cdce8p, line 2: `Callable`

pylint/checkers/base/name_checker/checker.py
```diff
@@ -410,25 +411,43 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
-                        assign_type.value.elts[node.parent.elts.index(node)]
-                    )
                 ):
-                    self._check_name(
-                        "typevar",
-                        assign_type.targets[0].elts[node.parent.elts.index(node)].name,
-                        node,
-                    )
+                    if self._assigns_typevar(
+                        assign_type.value.elts[node.parent.elts.index(node)]
```
cdce8p, line 430: `assign_type.value.elts[node.parent.elts.index(node)]` is used for both method calls (`_assigns_typevar`, `_assigns_typealias`). Maybe it would make sense to create a temp var for it.

Thinking one step further, the following is also reused.
```py
                            assign_type.targets[0]
                            .elts[node.parent.elts.index(node)]
                            .name
```
Would it make sense to combine both methods (`_assigns_typevar` and `_assigns_typealias`) and simply pass it `assign_type`, `node`, and a "mode" argument?

pylint/checkers/base/name_checker/checker.py
```diff
@@ -440,10 +459,12 @@ def visit_assignname(self, node: nodes.AssignName) -> None:
...
-            elif isinstance(
-                assign_type, nodes.AnnAssign
-            ) and utils.is_assign_name_annotated_with(node, "Final"):
-                self._check_name("const", node.name, node)
+            elif isinstance(assign_type, nodes.AnnAssign):
+                if utils.is_assign_name_annotated_with(node, "Final"):
+                    self._check_name("const", node.name, node)
+                elif self._assigns_typealias(assign_type.annotation):
+                    self._check_name("typealias", node.name, node)
+                return
```
cdce8p, line 467: I don't think the `return` is necessary here.

tests/functional/ext/typing/typing_broken_callable.py
```diff
@@ -4,7 +4,7 @@
 
 Use 'typing.Callable' instead.
 """
-# pylint: disable=missing-docstring,unsubscriptable-object
+# pylint: disable=missing-docstring,unsubscriptable-object, invalid-name
```
cdce8p, line 7:
```suggestion
# pylint: disable=missing-docstring,unsubscriptable-object,invalid-name
```
Might be just me, but I always find it irritating to see a space between error codes. Both for pylint and mypy. Similarly in the other test files.

cdce8p (changes requested):
> @cdce8p Just checking in. Do you want to review this before we merge this? Or can we merge without a review by you.

I had been away for a while working on other projects. If you want to reach me in these cases it's often best to DM me on Discord.

--
Regarding the PR, I left a few comments. Mostly left over references to `Callable` and an optimization suggestion. What we should think a little bit more about is the regex though. I checked the PR against Home Assistant and saw errors for `IPAddress` and `IPNetwork`. Those should be ok, IMO. Maybe just the `T` prefix should be disallowed, e.g. `TBadName`?

doc/data/messages/i/invalid-name/details.rst
DanielNoord: Oops, missed that.

tests/functional/ext/typing/typing_broken_callable.py
DanielNoord: Oh wow, I actually thought the preferred style was to separate them. I'm fine either way but thought that was the style we were going for.

pylint/checkers/base/name_checker/checker.py
DanielNoord: I made a new temp var.

Merging doesn't make too much sense I think. The inference is different `utils.safe_infer(node.func)` and `utils.safe_infer(node)` and we check for different `qnames`. Based on the `assign_type` we would simply follow completely different code paths.

tests/functional/ext/typing/typing_broken_callable.py
Pierre-Sassoulas: Yeah I  try to separate them myself :smile: (like arguments in a function call)

DanielNoord:
@cdce8p Changed the regex. Let me know what you think.

This doesn't allow `TBADname` but I don't think there is a good way to reliably distinguish between `TESTType` and `TTest`. As in, when a `T` in the first place is a prefix or part of a longer word.

DanielNoord:
After the `primer` comment I think it's better to not worry about the `T`. `TokenInfo` such as in `black` shouldn't really be flagged. Making changes.

pylint/checkers/base/name_checker/checker.py
```diff
@@ -40,7 +40,8 @@
 DEFAULT_PATTERNS = {
     "typevar": re.compile(
         r"^_{0,2}(?:[^\W\da-z_]+|(?:[^\W\da-z_]+[^\WA-Z_]+)+T?(?<!Type))(?:_co(?:ntra)?)?$"
-    )
+    ),
+    "typealias": re.compile(r"^_{0,2}[A-Z]+[a-z]+([A-Z][a-z]+)*$"),
```
cdce8p, line 44:
```suggestion
    "typealias": re.compile(r"^_{0,2}(?!T[A-Z]|Type)[A-Z]+[a-z]+(?:[A-Z][a-z]+)*$"),
```
Would this be a solution?
The negative lookahead would prevent cases like `TBadName` and `TypeTodo` but still allow `TodoType`. `TodoType` and `TypeTodo` could also be added as test cases.

tests/functional/ext/typing/typing_broken_callable.py
cdce8p: To be fair I've seen (and used) both 😄 For files I write myself, I do prefer without spaces. However if I'm just editing a random test file, I usually continue with the style that is present.

pylint/checkers/base/name_checker/checker.py
DanielNoord: Yeah, although it would still disallow `TTCircuit` as in the circuit of a [tt race](https://en.wikipedia.org/wiki/Isle_of_Man_TT).

That's really far fetched but I don't really like how this is such an arbitrary rule that is really difficult to figure out if you stumble upon it. I don't think `TTCircuit` will be used very often but there might be real uses of `TT` at the start and I think it would be very frustrating to figure out why `pylint` disallows them.

What would you think about using the current pattern as the default value and using your suggestion as our own overwrite? We could potentially even add our own overwrite in the documentation of the message?

cdce8p: Getting the regex 100% correct is pretty difficult. What should we do for the pandas cases then (`DatetimeNaTType`, `IntervalOrNA`, ...)?

IMO we should try to find something that works for most cases and can prevent common issues. The remaining ones should be handled with `pylint: disable[invalid-name]`. In this case I believe it's a much more prevalent error to use a `T` or `Type` prefix then something like `TTCircuit`, so I think it would be worth it.

I just opened #7322 to fix the same issue for the `TypeVar` name style as I've seen `_TRemote` and `_TCommand` in HomeAssistant lately.

DanielNoord: I agree that `T` and `Type` will be much more prevalent but I feel like flagging a `TTCircuit` would be considered a false positive. That's why I think we might want to be a little less strict by default and let users be more strict themselves. For me it falls within the false negatives vs. false positives preference, where I think the former is always more preferable.

cdce8p: > For me it falls within the false negatives vs. false positives preference, where I think the former is always more preferable.

Usually I agree here. However, I would consider `invalid-name` to be at least somewhat special. It's one of the few checks which creates false-positives in order to highlight certain errors and I would think users might already be aware of that. Just an example, in HomeAssistant about 7-8% of all `pylint: disable` comments are for `invalid-name` (53 of 697).

Pierre-Sassoulas:
Check fail should be fixed by https://github.com/PyCQA/pylint/pull/7321

pylint/checkers/base/name_checker/checker.py
DanielNoord: Hmm doesn't that indicate that `pylint` is being too strict then? The main issue I have is that it is pretty hard to know why `TTCircuit` failed based on the `doesn't conform to predefined naming style` message. If we can find a good way to communicate that T's aren't allowed without having to open the pylint documentation I would be more than fine with disallowing them. We have some other issues about `invalid-name` and it being unclear why stuff is being flagged as well, I'd like to avoid those.

Pierre-Sassoulas: Also, how would that work with the new design proposed in #7305 ? Explaining better why invalid-name was triggered is a big priority.

cdce8p: I don't think it's that big of deal in practice. At least in my experience the `pylint: disable=invalid-name` is added quickly and everybody is happy.

I think of this check more like a pointer
```
Hey here could be an error, please check that.
But if that's what you want, it's also fine.
```

Maybe we need a new error category for formatting / style issues 🤔

Pierre-Sassoulas: (I was thinking of #2018  is the second most up-voted issues in the repo, but the regex is hiding the fact that the name is too short, #7305 aims to separate the two. Explaining *everything* in the regex is lower priority indeed)

Pierre-Sassoulas: > Hey here could be an error, please check that. But if that's what you want, it's also fine.

Isn't that "convention" ? It's ok to not follow convention.

cdce8p: > > Hey here could be an error, please check that. But if that's what you want, it's also fine.
> 
> Isn't that "convention" ? It's ok to not follow convention.

True 😄
In that case, we might only need to better explain that it's just a convention and nothing more.

Pierre-Sassoulas: The issue might be that pep8 is wildly adopted (to a point that a lot of junior dev allow more time for PEP8 than to writing code that actually bring value). But PEP8 does not say anything about typevar/typealias yet so we're mixing a very popular convention with one that is less established.

DanielNoord: I have updated to use @cdce8p's suggestion and hope that the additional comments in `details.rst` will prevent any issues. I think Marc is right that we should try to enforce good behaviour here rather than prevent 2/3 opened issues about this.

jamesbraza:
Can I help forward this at all?  I hate to see all this work go stale for 4 months

Pierre-Sassoulas (dismissed):
LGTM, thank you for reminding us @jamesbraza. We have a lot going on and some MR that are hard/long to review get lost in the flow.

cdce8p:
> Can I help forward this at all? I hate to see all this work go stale for 4 months

I'll take another look at it soon.

DanielNoord:
> > Can I help forward this at all? I hate to see all this work go stale for 4 months
> 
> I'll take another look at it soon.

@cdce8p Do you think you'll be able to have a look at this still?

DanielNoord:
@Pierre-Sassoulas I would prefer to merge this even without the final review from Marc. If we get feedback on the regex we can always change it, but the underlying changes seem good to me. Are you okay with this?

Pierre-Sassoulas:
Yeah allright, let's merge.

Pierre-Sassoulas (approved):

DanielNoord:
Had to completely rebase as I was getting a lot of CI issues. Sorry for the loss of history, but this was the most sane way to do this.

DanielNoord:
Also note that the recent PR broke `main`. I included a commit to fix this.

PR #7117, opened 2022-07-03
Add regression test for #3651

### LINKED ISSUE #3651: Relative imports broken in pylint 2.5.2
Pylint fails with relative imports within a package. Having a package b in directory a with three files:

`a/b/__init__.py`
<pre>from .one import ONE
from .two import TWO</pre>

`a/b/one.py`
<pre>ONE = 1</pre>

`a/b/two.py`
<pre>from .one import ONE
TWO = ONE + ONE</pre>

pylint 2.4.4 + astroid 2.3.3 do not generate any warning or error, which is as expected.

pylint 2.5.2 + astroid 2.4.1 generates an import-error:

<pre>$ export PYTHONPATH=/full/path/to/a
$ cd /full/path/to
$ pylint a
************* Module a.b
a/b/__init__.py:2:0: E0401: Unable to import 'a.b.two' (import-error)</pre>

Versions:

$ pylint --version
pylint 2.5.2
astroid 2.4.1
Python 3.8.3 (default, May 17 2020, 18:15:42) 
[GCC 10.1.0]

jacobtylerwalls:
## Type of Changes

## Description
Closes #3651 with test

### DIFF
--- a/tests/lint/unittest_lint.py
+++ b/tests/lint/unittest_lint.py
@@ -39,6 +39,7 @@ from pylint.constants import (
 )
 from pylint.exceptions import InvalidMessageError
 from pylint.lint import PyLinter
+from pylint.lint.pylinter import MANAGER
 from pylint.lint.utils import fix_import_path
 from pylint.message import Message
 from pylint.reporters import text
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
Ah, ok, some test isolation issues to look into. It passes in isolation.

Pierre-Sassoulas:
> So the next step (eventually) is to find out if that's a real problem

Our custom context managers could be at fault, I'm thinking in particular of ``_test_sys_path``and ``_test_cwd`` that could be replaced by ``tmpdir.as_cwd()`` but I did not find the time yet (some are tricky to replace).

jacobtylerwalls:
Found the test pollution issue!

pylint/lint/pylinter.py
```diff
@@ -651,12 +651,14 @@ def check(self, files_or_modules: Sequence[str] | str) -> None:
                     self.get_ast, self._iterate_file_descrs(files_or_modules)
                 )
         else:
+            original_sys_path = sys.path[:]
```
Pierre-Sassoulas, line 654: I'm wondering if we should move the testutil context manager that permit to restore the sys.path at the end into the actual code ?

Pierre-Sassoulas (approved):
That's a really great catch !

jacobtylerwalls:
This test passes for me locally, but I don't have the capacity to look into why it's failing on CI.

DanielNoord:
@jacobtylerwalls Test does pass on `3.7` and raises `parse-error` on other interpreters. Is this using syntax that was changed in `3.8+`?

jacobtylerwalls:
It passes locally for me on all interpreters, so something to investigate with path/cwd vagaries on CI?

Pierre-Sassoulas:
I opened https://github.com/Pierre-Sassoulas/pylint/pull/157 to investigate, I don't have pypy locally so I'm going this with the CI and it's not really practical.

DanielNoord:
@jacobtylerwalls I did a little investigation and I could get this to fail with `pytest -k unittest_lint`. The fail is related to `x/__init__.py` not existing. By changing the test to use `linter.check(["x/y"])` the failure goes away but I'm not sure if that breaks the intention of the test. Would that be a suitable fix?

jacobtylerwalls:
Unfortunately yes, that would break the intention of the test.

Possibly relevant: https://github.com/PyCQA/pylint/issues/3651#issuecomment-705198488

DanielNoord:
Hm, but it seems as if the failed test is due to something else? Namespace packages not being listed correctly? If you lint `x/y` it does work since that is a real package. For directory `x` with only subdirectory `y` we don't recognise that it is a linkable package, isn't that expected behaviour?

jacobtylerwalls:
Ah, you're right!

DanielNoord:
@jacobtylerwalls So, should the test be changed to close the issue? Or do we need another type of test for this issue?

jacobtylerwalls:
Yes, I have just pushed an update and reopened the PR. Thanks for walking me through it. (There were a lot of astroid changes recently, and I got distracted based on the author's original expectation that you could lint `x`.)

DanielNoord (approved):

Issue #5465, opened 2021-12-03
Inclusive language update


Talador12:
### Current problem

In `.pylintrc` we have the following attribute

`extension-pkg-whitelist=`

For context, here is a description of that 
```
# A comma-separated list of package or module names from where C extensions may
# be loaded. Extensions are loading into the active Python interpreter and may
# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
# for backward compatibility.)
```

At what point do we end backwards compatibility on this term? It is worth deprecating that language at some point.

### Desired solution

We should deprecate `extension-pkg-whitelist` in favor of `extension-pkg-allow-list` for all references

### Additional context

Lots of different inclusivity guides out there, but here's one
https://developers.google.com/style/inclusive-documentation

Talador12:
If anyone can find other references of language that should be deprecated, please add examples to this ticket.

Pierre-Sassoulas:
This has been answered in detail in the [initial MR](https://github.com/PyCQA/pylint/pull/3961), in particular [here](https://github.com/PyCQA/pylint/pull/3961#issuecomment-735419252) and [here](https://github.com/PyCQA/pylint/pull/3961#pullrequestreview-594778776):

> We're never going to break compatibility for the old name. That would be forcing maintenance work on pylint users for political reasons. 

So to do that partially would requires #5462. You're welcome to contribute to it if you like.

Talador12:
I disagree that this would be considered `political reasons` as this is more about ethics than politics. I will add to #5462

Talador12:
@Pierre-Sassoulas One more case to consider

pylintrc uses `[MASTER]` at the top of the file. Could this be changed to `[MAIN]` ? Currently it appears to only allow `[MASTER]`.

Pierre-Sassoulas:
Please read the initial MR provided and the two comments linked in particular.

Talador12:
I had read the MR and two comments you had linked. 

Confirmed that an updated pylint version has [MAIN] compatibility.

However, I did notice that `pylint --generate-rcfile` still uses `[MASTER]` terminology by default.

Since `[MAIN]` is compatible, it should be the default for `pylint --generate-rcfile` going forward.

Pierre-Sassoulas:
[...]

PR #7118, opened 2022-07-03
Deprecate extension-pkg-whitelist

DanielNoord:

## Type of Changes

## Description

Ref. https://github.com/PyCQA/pylint/issues/5465

Doesn't close as I also need to deprecate the `black_list` options.

### DIFF
--- a/doc/exts/pylint_options.py
+++ b/doc/exts/pylint_options.py
@@ -87,6 +87,11 @@ def _create_checker_section(
         if option.optdict.get("hide_from_config_file"):
             continue
 
+        if "kwargs" in option.optdict:
+            assert isinstance(option.optdict["kwargs"], dict)
+            if option.optdict["kwargs"].get("new_names"):
+                continue
+
         # Get current value of option
         value = getattr(linter.config, option.name.replace("-", "_"))
 
--- a/doc/user_guide/configuration/all-options.rst
+++ b/doc/user_guide/configuration/all-options.rst
@@ -227,8 +227,6 @@ Standard Checkers
 
    extension-pkg-allow-list = []
 
-   extension-pkg-whitelist = []
-
    fail-on = []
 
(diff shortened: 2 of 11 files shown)


Pierre-Sassoulas (commented):
I don't think we should actually deprecate this one. If we have the capability to auto-upgrade the configuration we don't need to deprecate, and if we don't then it's forcing work on our user to fix something that is not broken.

DanielNoord:
I mean, I'm not personally offended by this in anyway but if some users are, why wouldn't we put in a little effort to alleviate that? We already have the functionality to do so, so why not use it? I get that the discussion is a bit political, but it doesn't really hurt those who aren't offended and might help those who are.

Pierre-Sassoulas:
> but if some users are, why wouldn't we put in a little effort to alleviate that?

It's not only us doing the effort but every users that will need to upgrade their conf for 3.0. I don't want users to have to upgrade their conf in general... but in particular I don't want them to have to upgrade their configuration for political reasons.  And I especially don't want to deal with bigots that had to upgrade their configuration and now have a strong desire to tell us about the etymology of blacklist. Let's not create political breaking change, and let's not trigger anyone. 

If we do https://github.com/PyCQA/pylint/issues/5462 then we'll be able to clean up configuration. The amount of work to do this issue is tiny in comparison of the work involved for all pylint users to upgrade all their conf and probably comparable to the bigot handling we'd have to do for this anyway.

DanielNoord:
Should we close the related issue then?

@jacobtylerwalls I noticed the `primer` diff here. Something is up with my recent `astroid` PR 😅

Pierre-Sassoulas:
Yeah, sorry about that, #5465 only has reference to a very lengthy discussion in another PR so it's not self evident what needed to be done. I upgraded the issue as blocked by #5462

PR #7125, opened 2022-07-04
Rename bad-exception-context to bad-exception-cause

### LINKED ISSUE #3694: Rename `bad-exception-context` to `bad-exception-cause`
While working on #3687 , I noticed bad naming. The `bad-exception-context` rule needs to be renamed to `bad-exception-cause`, because it's about the cause and not the context.

Note that the only way that the part after `from` could be related to the `__context__` of the exception, is if it's `None`, but:

1. Our rule is only relevant when it's not `None`.
2. Even when it's `None`, the `__context__` is still the same, it's just the `__suppress_context__` attribute that gets set to `True`.

We should also rename the `_check_bad_exception_context` method accordingly, and its docstring.

Fifi31:
## Type of Changes

## Description

Rename `bad-exception-context` to `bad-exception-cause` and `_check_bad_exception_context` to `_check_bad_exception_cause`, as well as docstrings related and error description.

Closes #3694


pylint/checkers/exceptions.py
```diff
@@ -73,10 +73,10 @@ def _is_raising(body: list) -> bool:
...
     ),
     "E0703": (
-        "Exception context set to something which is not an exception, nor None",
-        "bad-exception-context",
+        "Exception cause set to something which is not an exception, nor None",
+        "bad-exception-cause",
         'Used when using the syntax "raise ... from ...", '
-        "where the exception context is not an exception, "
+        "where the exception cause is not an exception, "
         "nor None.",
```
Pierre-Sassoulas, line 80: This need to define an ``"old_names"`` so the message control still work with ``bad-exception-context``. You can search for an example in the codebase.

pylint/checkers/exceptions.py
```diff
@@ -302,22 +302,22 @@ def _check_misplaced_bare_raise(self, node):
...
+        An exception cause can be only `None` or an exception.
         """
         cause = utils.safe_infer(node.cause)
         if cause in (astroid.Uninferable, None):
             return
 
         if isinstance(cause, nodes.Const):
             if cause.value is not None:
-                self.add_message("bad-exception-context", node=node)
+                self.add_message("bad-exception-cause", node=node)
```
Pierre-Sassoulas, line 316:
```suggestion
                self.add_message("bad-exception-cause", node=node, confidence=INFERENCE)
```
Unrelated to the current MR, but as we upgrade the whole functional output we might as well add the right confidence at the same time. This is going to require a new import ``from pylint.interfaces import INFERENCE`` that I can't suggest.

pylint/checkers/exceptions.py
```diff
@@ -302,22 +302,22 @@ def _check_misplaced_bare_raise(self, node):
...
         elif not isinstance(cause, nodes.ClassDef) and not utils.inherit_from_std_ex(
             cause
         ):
-            self.add_message("bad-exception-context", node=node)
+            self.add_message("bad-exception-cause", node=node)
```
Pierre-Sassoulas, line 320:
```suggestion
            self.add_message("bad-exception-cause", node=node, confidence=INFERENCE)
```

Pierre-Sassoulas (changes requested):
Thank you for tackling this, much appreciated ! Just some small changes, it's already looking good.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -14,4 +14,12 @@ Summary -- Release highlights
...
   linted without resorting to using the ``-recursive=y`` option.
 
+Other Changes
+=============
+
+
+* ``bad-exception-context`` has been renamed to ``bad-exception-cause`` as it is about the cause and not the context.
+
+  Closes #3694
+
```
Pierre-Sassoulas: Sorry you've been caught in a change of how we do thing, we added towncrier for the changelog, the instruction are in the pull request template but you already created yours so you can't see them:

- [ ] Create a news fragment with `towncrier create <IssueNumber>.<type>` which will be

pylint/checkers/exceptions.py
```diff
@@ -94,6 +87,14 @@ def _is_raising(body: list[nodes.NodeNG]) -> bool:
         "as an exception is raised inside the try block, but it is "
         "nevertheless a code smell that must not be relied upon.",
     ),
+    "E0705": (
+        "Exception cause set to something which is not an exception, nor None",
+        "bad-exception-cause",
+        'Used when using the syntax "raise ... from ...", '
+        "where the exception cause is not an exception, "
+        "nor None.",
+        {"old_names": [("E0703", "bad-exception-context")]},
```
Pierre-Sassoulas, line 96: 👍

Pierre-Sassoulas (changes requested):

doc/whatsnew/2/2.15/index.rst
Fifi31: Okay I didn't know that, so I created the file 3694.other manually after seeing the pipeline failed. Could there be a problem if the file was created manually and should I generate it with towncrier instead ?

Issue #6633, opened 2022-05-17
New check ``use-match-statement`` to suggest using the ``match`` statement


Pierre-Sassoulas:
### Current problem

This first example is a lot slower than the second one:
```python
def sequence_match_logical():
    """ Test matching the first element of a sequence is a frog. """
    seq = ["🐸", "🐛", "🦋", "🪲"]
    frogs = 0
    for _ in range(100_000):
        if isinstance(seq, Sequence) and len(seq) > 0 and seq[0] == "🐸": 
            frogs += 1
    
    assert frogs == 100_000
```    
```python
def sequence_match_statement():
    """ Test matching the first element of a sequence is a frog. """
    seq = ["🐸", "🐛", "🦋", "🪲"]
    frogs = 0
    for _ in range(100_000):
        match seq:
            case ["🐸", *_]: frogs += 1
    
    assert frogs == 100_000
```

Even without the performance aspect it's a "cool" new feature of 3.10 so that could be a convention message.

### Desired solution

New message ``use-match-statement`` to suggest using the second example (``consider-using-match-statement``when it's not clear cut).

### Additional context

https://twitter.com/anthonypjshaw/status/1526034155448184832?s=21&t=r9ADTOqs_C0DP_RraR97Pg

https://github.com/tonybaloney/anti-patterns/blob/master/bench_match.py#L3-L23

PR #7133, opened 2022-07-05
Research about replacing conditionals with structural pattern matching

gaborantal:
Hello there!

We are currently working on a research project at our university, we made a tool that is able to detect and transform conditionals where structural pattern matching could be used.  If you're not familiar with structural pattern matching, [PEP 634](https://peps.python.org/pep-0634/), [PEP 635](https://peps.python.org/pep-0635/) and [PEP 636](https://peps.python.org/pep-0636/) is a good point to start.

Our tool is working as it could (we are optimizing it ever since), but as AST transformation is happening, the comments are lost. To prevent that, we collects the comments and place it above the transformed code.

**In our research, we're at the point where we would like to know what is your opinion about the replaced conditionals? Do you think it makes the code more or less readable in general?**

Of course, if you see any particular refactoring which looks bad, we very much appreciate if you mark them so that we can make our tool better.

We know that this PR won't be merged, we gave this idea on Discord, so that more and more people could comment on it.

**Notice**:  What you're about to see is the raw output of our tool, nothing was edited, that's the reason why there might be several issues with the code (e.g. comments, line feeds somewhere). Of course, we hope that one day we can use our tool and those modifications will be merged here, but that time, we will manually check the refactorings and place the comments to their place, and so on.

If you have some time, please check out the PR, and give a feedback on whether structural pattern matching makes the code more or less readable in practice.

Thanks for your help,
Gábor

### DIFF
--- a/pylint/checkers/async.py
+++ b/pylint/checkers/async.py
@@ -1,4 +1,4 @@
-# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+﻿# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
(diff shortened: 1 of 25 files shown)


Pierre-Sassoulas:
See #6633 for discussion

PR #7136, opened 2022-07-06
[configuration doc] Add the '--generate-rcfile' option

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to this stackoverflow question:  https://stackoverflow.com/questions/72877687/pylint-where-is-the-pylintrc-documentation

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -23,13 +23,12 @@ See :ref:`upgrading pylint in the installation guide <upgrading_pylint>`.
 How do I find the name corresponding to a specific command line option?
 -----------------------------------------------------------------------
 
-You can generate a sample configuration file with ``--generate-toml-config``.
-Every option present on the command line before this will be included in
-the toml file
+See :ref:`the configuration documentation <all-configurations-options>`.
 
-For example::
+What is the format of the configuration file?
+---------------------------------------------
 
-    pylint --disable=bare-except,invalid-name --class-rgx='[A-Z][a-z]+' --generate-toml-config
+The configuration file can be an ``ini`` or ``toml`` file. See the :ref:`exhaustive list of possible options <all-options>`.
 
 How to disable a particular message?
 ------------------------------------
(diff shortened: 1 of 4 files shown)


DanielNoord:
We have deprecated the functions used for ``--generate-rcfile`` internally. I know we didn't reach a decision last time, but weren't we thinking of nudging users towards `.toml` files? They are more easily parseable anyway and are becoming the standard within the ecosystem.

DanielNoord:
> Hmm yeah, but maybe we should still add the command in the doc so if someone really want an ini file we do not get an issue/question for it ? 

Yeah, I'm not opposed to it. We can just issue a warning about it on `3.0` and make the internal function private. Something like:
``You can still use this command but it is no longer maintained``

> For example flake8's really hell bent on not using pyproject.toml right now.

This issue is almost fixed I think. `setuptools` has a beta in which most of the original issues have been solved 😄

Pierre-Sassoulas:
> Something like: You can still use this command but it is no longer maintained

Yeah it would make sense to deprecate some configuration options, right now there's a lot to keep backward compatible. I think there's a lot to discuss on this issue though. (I don't think we can deprecate ini before auto-upgrade of configuration, do we support something else than pyproject.toml or only that...).

DanielNoord:
> Yeah it would make sense to deprecate some configuration options, right now there's a lot to keep backward compatible. I think there's a lot to discuss on this issue though. (I don't think we can deprecate ini before auto-upgrade of configuration, do we support something else than pyproject.toml or only that...).

We currently support anything that looks like an `.ini` or `.toml` file that has some form of `pylint` configuration in it.
`pylint-config` can already be used to auto-upgrade as it takes your current configuration into account. Running it from within a directory with a `pylintrc` and choosing `toml` should create a `.toml` file based on that previous file.

DanielNoord:
Probably 😄 

But since `pylint-config` is very barebones right now I haven't put much effort in its documentation just yet.

Pierre-Sassoulas:
By the way as the generate toml config take the current configuration into account doesn't it do that too ? I suppose it also use the right terminology and not the deprecated argument ? (Isn't a big part of https://github.com/PyCQA/pylint/issues/5462 already done ?)

PR #7141, opened 2022-07-07
Bump or remove some ``TODOs``

DanielNoord:

## Type of Changes

## Description

@Pierre-Sassoulas only two TODOs remain. Could you take a look at those:
https://github.com/PyCQA/pylint/blob/7616e83e2541a5f3a66ea2caa751979c0f398d2b/script/bump_changelog.py#L19
https://github.com/PyCQA/pylint/blob/7616e83e2541a5f3a66ea2caa751979c0f398d2b/script/fix_documentation.py#L17

Refs. #6632.

### DIFF
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -219,8 +219,6 @@ class ExceptionRaiseLeafVisitor(BaseVisitor):
     visit_exceptioninstance = visit_instance
 
     def visit_classdef(self, node: nodes.ClassDef) -> None:
-        # TODO (Typing): This function can be called from visit_instance
-        #  with the return type of instance._proxied (not typed right now)
         if not utils.inherit_from_std_ex(node) and utils.has_known_bases(node):
             if node.newstyle:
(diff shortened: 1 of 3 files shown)


pylint/checkers/exceptions.py
```diff
@@ -219,8 +219,6 @@ def visit_instance(self, instance: objects.ExceptionInstance) -> None:
     visit_exceptioninstance = visit_instance
 
     def visit_classdef(self, node: nodes.ClassDef) -> None:
```
Pierre-Sassoulas, line 221: Something needs to be done here, unless ``instance._proxied `` always return a nodes.ClassDef. But I think at least ``Instance`` is a possibility.

Pierre-Sassoulas (changes requested):
The last two todos will be done in #6974 and related merge request.

pylint/checkers/exceptions.py
DanielNoord: No. I fixed the typing here. `_proxied` is always `ClassDef`.

Issue #6284, opened 2022-04-12
Investigate our support of defining ``rcfile=`` in an ``rcfile``


DanielNoord:
Regarding this option, I think there's this related question on stackoverflow : https://stackoverflow.com/questions/71799145/pylint-nested-pylintrc-specify-a-configuration-file. Do we want to allow the rc-file option in the configuration file itself ?

_Originally posted by @Pierre-Sassoulas in https://github.com/PyCQA/pylint/pull/6199#discussion_r848693260_

------------

We should probably investigate whether we actually fully support this.

PR #7146, opened 2022-07-07
Remove unsupported option from our own ``pylintrc``

DanielNoord:

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/6284.

We don't support this and it would unnecessarily complicate things. The rationale for this is also flimsy at best:
The original reporter wanted to put all `linter` config in a special subdirectory and then create a `.pylintrc` file which would point to a file in that directory. That's just... Use a symlink or just use one file, this unnecessarily complicates stuff. 
_We're a linter, we should not support such anti-patterns_ 😄

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,8 +1,5 @@
 [MAIN]
 
-# Specify a configuration file.
-#rcfile=
-
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
 #init-hook=


Pierre-Sassoulas (commented):
Agree, but isn't it some code that we need to remove / or add for this ? Maybe something to error out if it's in the configuration file (``unknown-option-value``, or ``argparse.Error`` ?). I think a functional configuration test to check that it's not silently failing is in order.

DanielNoord:
> Agree, but isn't it some code that we need to remove / or add for this ? Maybe something to error out if it's in the configuration file (`unknown-option-value`, or `argparse.Error` ?). I think a functional configuration test to check that it's not silently failing is in order.

That's a bit more complicated as we don't really have a system in place to do that. There are other options that are affected by this as well: ``--version`` and ``--errors-only`` will also interact unexpectedly in a configuration file.
That's why those options have the ``hide_from_config_file`` attribute in their `optdict`. They will never be printed by ``--generate-rcfile`` and don't show up in the documentation. There is no indication that they should work in a configuration file.

Pierre-Sassoulas:
But we warn of unknown option in config file, right ?

Pierre-Sassoulas:
> This prints the long help message and then exits the run 😅

😄  good one !

>  there is not really a good way to determine those and what the behaviour should be.

Yeah we'll have to think about it for each one. But we can make an exhaustive list the CLI-only option, rigth ? Typically long-help is a CLI option.

DanielNoord:
> I'm thinking about:
> 
> * Raising a pylint message like unknown-option-value
> * Raising an argparse.Error (but it does not make much sense as the option parsing could have already happened when we parse the configuration ?)
> * warning message ?

We could, but I'm hesitant to add that complexity to the configuration parsing. I'd rather make the lack of support explicit in the documentation (which imo it kind of already is) and be done with it.
The added complexity doesn't really weigh up to the small number of users who want to use this imo.

Pierre-Sassoulas:
Is it really that complex, it's only a ``if x in hide_from_config_file`` + ``add_message()``, right ? (And it's consistent with us warning about unknown options)

DanielNoord:
> Is it really that complex, it's only a `if x in hide_from_config_file` + `add_message()`, right ? (And it's consistent with us warning about unknown options)

We could probably add something, but I was more talking about the added code + extra runtime for such small benefit.
We already need 100 lines of code just to get the order of parsing files and `CLI` correctly (`_config_initialization()`). I'm wary of adding anything to the configuration parsing that does not offer actual benefit or fixes a big issue in order not to end up with a situation like `flake8` where basically nobody wants to touch configuration parsing because it was so troublesome previously.

Pierre-Sassoulas:
Hmm, right we might not want to waste 3 millisecond for each pylint run for that. Maybe something to do with the autoupgrader of configuration then.

DanielNoord:
> Hmm, right we might not want to waste 3 millisecond for each pylint run for that. Maybe something to do with the autoupgrader of configuration then.

Since `pylint-config` uses `hide_from_config` it should remove those automatically.

_See, I did make some good choices during all those refactors_ 😄

Issue #5230, opened 2021-10-29
How to ignore typing stub pyi files? ignore-options don't work/break


ewerybody:
### Question

Hello! I'm using latest pylint through VS Code `1.61.2` and often have a couple super long `.pyi` files open.
Most of these are not written by me and I'm not interested in any issues there.

But I AM interested in pylint judging my very own code. So: HOW do I make it so that it ignores these `pyi`s?
From [the features documentaion](https://pylint.pycqa.org/en/latest/technical_reference/features.html) there are a couple of ignore options:
* `--ignore=*.pyi` - no change at all as if it wasn't there
* `--ignore-paths=*.pyi` or `--ignore-patterns=*.pyi` causes empty pylint returns like:
    ```
    ...
    ##########Linting Output - pylint##########
    ##########Linting Output - pylint##########
    ##########Linting Output - pylint##########
    ##########Linting Output - pylint##########
    ##########Linting Output - pylint##########
    ```
    at least in VS Code.
    If I try this in the commandline it yields a huge traceback:
    ```
    Traceback (most recent call last):
    File "C:\Python39\lib\runpy.py", line 197, in _run_module_as_main
        return _run_code(code, main_globals, None,
    File "C:\Python39\lib\runpy.py", line 87, in _run_code
        exec(code, run_globals)
    File "C:\Python39\Scripts\pylint.exe\__main__.py", line 7, in <module>
    File "C:\Python39\lib\site-packages\pylint\__init__.py", line 24, in run_pylint
        PylintRun(sys.argv[1:])
    File "C:\Python39\lib\site-packages\pylint\lint\run.py", line 332, in __init__
        args = linter.load_command_line_configuration(args)
    File "C:\Python39\lib\site-packages\pylint\config\option_manager_mixin.py", line 342, in load_command_line_configuration
        (options, args) = self.cmdline_parser.parse_args(args=args)
    File "C:\Python39\lib\optparse.py", line 1387, in parse_args
        stop = self._process_args(largs, rargs, values)
    File "C:\Python39\lib\optparse.py", line 1427, in _process_args
        self._process_long_opt(rargs, values)
    File "C:\Python39\lib\optparse.py", line 1501, in _process_long_opt
        option.process(opt, value, values, self)
    File "C:\Python39\lib\site-packages\pylint\config\option.py", line 168, in process
        value = self.convert_value(opt, value)
    File "C:\Python39\lib\optparse.py", line 771, in convert_value
        return self.check_value(opt, value)
[...]

PR #7147, opened 2022-07-07
Use ``regular expression`` instead of ``regex``

DanielNoord:

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/5230


Pierre-Sassoulas (commented):
LGTM, but is this the only occurrences of regex for the whole pylint codebase in user facing doc ?

DanielNoord:
No! And I even kept `regex` on the second line. I think `regex` is a fairly standard term so it shouldn't be too bad to use it, but since we got specific feedback about this (and I had assigned this issue to myself a couple of months ago) I thought I would tackle it.

Pierre-Sassoulas (approved):

Issue #4035, opened 2021-01-18
missing-param-doc for numpy doc style: variable with list of legal values


juliangilbey:
According to the [NumPy documentation style guidelines](https://numpydoc.readthedocs.io/en/latest/format.html), section 4:
> When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first:
```
order : {'C', 'F', 'A'}
    Description of `order`.
```
However, using this syntax in a function docstring causes pylint to complain about `missing-param-doc` and `missing-type-doc`.

Here's a minimal example.  (With only one parameter of this form, it is completely ignored.)

```python
def myfunc(arg, option):
    """Do something boring.

    Parameters
    ----------
    arg : int
        The number of times to print it.
    option : {"y", "n"}
        Do I do it?

    """
    for i in range(arg):
        print(option)
```
and here is the `pylint` output:
```
************* Module pylint_test
pylint_test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
pylint_test.py:1:0: W9015: "option" missing in parameter documentation (missing-param-doc)
pylint_test.py:1:0: W9016: "option" missing in parameter type documentation (missing-type-doc)
pylint_test.py:12:8: W0612: Unused variable 'i' (unused-variable)

------------------------------------
Your code has been rated at -3.33/10
```

### pylint --version output
```
pylint 2.6.0
astroid 2.4.2
Python 3.9.1+ (default, Jan 10 2021, 15:42:50) 
[GCC 10.2.1 20201224]
```

hippo91:
@juliangilbey thanks for this report.

DanielNoord:
I would like to fix this after merging #5231

PR #7149, opened 2022-07-08
Allow lists of default values in parameter documentation for ``Numpy``

DanielNoord:

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/4035.

I really don't want to talk about the number of hours it took me to get this regex right. The fact that I assigned this to myself in October is a small indication...

### DIFF
--- a/doc/whatsnew/2/2.14/full.rst
+++ b/doc/whatsnew/2/2.14/full.rst
@@ -17,6 +17,10 @@ Release date: TBA
 
 * Fixed the disabling of ``fixme`` and its interaction with ``useless-suppression``.
 
+* Allow lists of default values in parameter documentation for ``Numpy`` style.
+
+  Closes #4035
+
 
 What's New in Pylint 2.14.4?
 ----------------------------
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
I think we're hitting the limit of what regex are good for, but as you spent 9 months to birth it let's put a diaper on the little ``re_param_line`` and try to keep it alive and clean now 😄

DanielNoord:
> I think we're hitting the limit of what regex are good for, but as you spent 9 months to birth it let's put a diaper on the little `re_param_line` and try to keep it alive and clean now 😄

It's also that some of these regexes are really quite bad. Using both f-strings and regular expressions from different classes and in different recursive patterns is awful to work with. You can't easily copy them to `regex101` and try to fix something.
That made this really difficult.

Pierre-Sassoulas:
Outch should have waited for the primer, before approving.

Pierre-Sassoulas:
Hmm right, we have most message with actually a more accurate message with one param that do not warn anymore. This is why we need the better primer diff, I need to do it from scratch again though 😄

PR #7150, opened 2022-07-09
Fix Pylint spelling in docs

sisp:
## Type of Changes

## Description

I've fixed all occurrences of "pylint" that I could find in the docs and replaced them by "Pylint" because that's the correct spelling from what I can tell. I excluded occurrences in `doc/whatsnew/**` because there the formatting is a bit inconsistent anyway; cleaning those files up might be a different task altogether IMO. :wink:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -55,7 +55,7 @@ ecosystem of existing plugins for popular frameworks such as `pylint-django`_ or
 
 Pylint isn't smarter than you: it may warn you about things that you have
 conscientiously done or check for some things that you don't care about.
-During adoption, especially in a legacy project where pylint was never enforced,
+During adoption, especially in a legacy project where Pylint was never enforced,
 it's best to start with the ``--errors-only`` flag, then disable
 convention and refactor message with ``--disable=C,R`` and progressively
 re-evaluate and re-enable messages as your priorities evolve.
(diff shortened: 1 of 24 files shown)


DanielNoord:
Do we also want to add "``" everywhere?

sisp:
I'd say no because Pylint is the name of the tool and ``pylint`` is the executable IMO.

jacobtylerwalls:
Hi @sisp just wondering, what was your thought process behind determining the preference for one capitalization style over the other:

> I've fixed all occurrences of "pylint" that I could find in the docs and replaced them by "Pylint" because that's the correct spelling from what I can tell.

Pierre-Sassoulas:
> Pylint is the name of the tool/project and pylint is the executable or the cloned source directory in some cases or the package name.

Make sense but there's a lot of fix that are not correct or ambiguous in this pull request according to this definition and thinking about it each time is a waste of brainpower 😄 I would be okay with always talking about the executable and never capitalize anywhere.

sisp:
> Hi @sisp just wondering, what was your thought process behind determining the preference for one capitalization style over the other:
> 
> > I've fixed all occurrences of "pylint" that I could find in the docs and replaced them by "Pylint" because that's the correct spelling from what I can tell.

It's spelled "Pylint" e.g. in the headings on the landing page of the docs. I'm not opinionated about whether it should be spelled with a capital or lowercase P, but I think it should be consistent.

sisp:
> > Pylint is the name of the tool/project and pylint is the executable or the cloned source directory in some cases or the package name.
> 
> Make sense but there's a lot of fix that are not correct or ambiguous in this pull request according to this definition and thinking about it each time is a waste of brainpower 😄 I would be okay with always talking about the executable and never capitalize anywhere.

I agree with the brainpower argument. But I think there is a fine distinction between the project and the other meanings. Also, if all occurrences were spelled lowercase and as inline code, wouldn't it also apply to Black, isort, Flake8, etc.?

sisp:
Lowercase + "``"?

jacobtylerwalls:
> Lowercase + "``"?

I don't think we should overuse the backticks. Let's use it when it's very clear we're referring to the executable or a module and not just speaking generally about the tool.

Pierre-Sassoulas:
I think backtick would be nice and the consistency would be nice... but I don't want to cut hair about it. In particular I don't want to pester every contributor in review with "please use 'pylint' not 'Pylint' and also please use backtick around it", instead of focusing on what's really important for the project code wise or even documentation wise.

We use autoformatter on the codebase to not have to deal with this kind of neat picking in review, but enforcing this automatically in the doc would be waaaaay over the top.

jacobtylerwalls:
Yeah, the improbability of keeping this consistency permanently is a good reason to not take any action here.

Pierre-Sassoulas:
Thank you for contributing in pylint @sisp but I'm going to close this as "a standard that we won't be able to maintain" 😄 I hope you find an issue that suit you, there's quite a lot : https://github.com/PyCQA/pylint/issues?q=is%3Aopen+is%3Aissue+label%3A%22Documentation+%3Agreen_book%3A%22+label%3A%22Needs+PR%22 !

sisp:
Sure, no hard feelings. I thought it would be nice to be consistent regarding the spelling, but I understand your arguments.

PR #7154, opened 2022-07-10
Prevent appending distinct crash reports together that occur in the same hour

jacobtylerwalls:
Previously, any crash reports generated in the same hour would get appended into the same file.

When debugging multiple failures (or repeating a single failure while trimming down a large file into a minimal example) it's not ideal to have distinct crash reports all appended into the same file IMO.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -259,7 +259,7 @@ class PyLinter(
     name = MAIN_CHECKER_NAME
     msgs = MSGS
     # Will be used like this : datetime.now().strftime(crash_file_path)
-    crash_file_path: str = "pylint-crash-%Y-%m-%d-%H.txt"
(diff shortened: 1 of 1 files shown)


DanielNoord (approved):
I can't approve this enough. Man this annoyed me when I was working on the `pandas` recursion stuff.

👍 👍 👍

jacobtylerwalls:
ha. Do you want me to add a second to the timestamp in case you're generating more than one per minute?

DanielNoord:
> ha. Do you want me to add a second to the timestamp in case you're generating more than one per minute?

Yeah, good call! Probably even better!

Pierre-Sassoulas (approved):
Sorry, that was on me 😄 Rational that we don't want to spam the user with crash report, but indeed (and hopefully) we're the one that interact with this more often.

Issue #7085, opened 2022-06-29
Request: Add optional Checker to `enforce explicit namespace packages` aka `there must be an __init__.py in each subdir`


doublethefish:
### Current problem

Often tools (like Sphinx and pytest) do strange things because of `importlib`'s various quirks (see [importlib](https://docs.python.org/3/library/importlib.html) and [PEP420](https://peps.python.org/pep-0420/)*), specifically associated with `implicit namespaces` (vs explicit).

One explicit (and important) example is `coverage.py` + `pytest` doesn't always include files in implicit namespace packages - so code coverage is reported incorrectly 🤬. Obviously that's not the fault of `pylint` but having a pylint Checker to detect such problems would be ace.

Those quirks are most often solved if one enforces `__init__.py` files to be in every package subdir 🎊.

\* Sometimes, I wonder if PEP420 was conceived on the 20th April, sigh 😉.

### Desired solution

In my most nebulous projects, I have a `pytest` test that checks various things, but I am not sure if it is appropriate as a Checker as-is.

One day I might get around to integrating it into `pylint` but it would probably be a good first feature for someone to implement.

Here's some code that "does the trick".
Note: that it contains `no cover` declaratives only because we would be testing the test code, but `pylint`'s test infrastructure is very well set up to test this sort of code as a checker.

```py
# tests that all the conditions to get coverage are met
import os
import typing
from pathlib import Path


def _dir_has_any_pyfile(
    start_dir: str,
) -> typing.Union[bool, typing.List[str]]:  # pragma: no cover
    for _dir_name, _subdirs, file_list in os.walk(start_dir):
        subdir_pyfiles = [
            filename for filename in file_list if filename.endswith(".py")
        ]
        if subdir_pyfiles:
            return subdir_pyfiles

    # not found
    return False


def test_all_dirs_have_inits() -> None:
    dirs_with_no_init = []
    dirs_with_no_init_but_subdir_pys = []
    dirs_with_only_init = []

    base_path = (Path(os.path.dirname(__file__)) / ".." / "..").resolve()
    paths_to_test = (base_path / "mdgm", base_path / "tests")

    for start_dir in paths_to_test:
        for dir_name, subdirs, file_list in os.walk(start_dir):
            # strip out known util dirs
            subdirs[:] = [
                dname
                for dname in subdirs
                if dname not in ("__pycache__", ".pytest_cache")
            ]

            if str(Path(dir_name).relative_to(base_path)) == "tests/data":
[...]

PR #7155, opened 2022-07-10
Refactor context manager for sys path and current working directory in tests

Pierre-Sassoulas:
## Type of Changes

## Description

There's still a lot to do there, but I'm going slowly. In particular I would have though that some pytest fixtures would have existed for these use cases, but it seem it's not the case, or is non-trivial.

### DIFF
--- a/pylint/testutils/utils.py
+++ b/pylint/testutils/utils.py
@@ -8,6 +8,8 @@ import contextlib
 import os
 import sys
 from collections.abc import Generator, Iterator
+from copy import copy
+from pathlib import Path
 from typing import TextIO
 
 
@@ -23,18 +25,26 @@ def _patch_streams(out: TextIO) -> Iterator[None]:
 
 
 @contextlib.contextmanager
-def _test_sys_path() -> Generator[None, None, None]:
+def _test_sys_path(
+    replacement_sys_path: list[str] | None = None,
+) -> Generator[None, None, None]:
     original_path = sys.path
     try:
+        if replacement_sys_path is not None:
+            sys.path = copy(replacement_sys_path)
         yield
     finally:
         sys.path = original_path
 
 
 @contextlib.contextmanager
-def _test_cwd() -> Generator[None, None, None]:
+def _test_cwd(
+    current_working_directory: str | Path | None = None,
+) -> Generator[None, None, None]:
     original_dir = os.getcwd()
     try:
+        if current_working_directory is not None:
+            os.chdir(current_working_directory)
         yield
     finally:
         os.chdir(original_dir)
--- a/pylint/testutils/utils.py
+++ b/pylint/testutils/utils.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import contextlib
 import os
 import sys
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -13,6 +13,7 @@ test = pytest
 
 [tool:pytest]
 testpaths = tests
+pythonpath = data input regrtest_data
```
Pierre-Sassoulas, line 16: This could be a standalone pull request, it's using pytest config to add fixture directory to the python path.

jacobtylerwalls: I'm testing this change right now to see if that's all we need to revert the test adjustments in the astroid pr.

jacobtylerwalls: Doesn't seem to make a difference -- which makes sense based on my limited understanding: the astroid pathways for defining modules are outside the control of pytest, and `tests` definitely is a package, we finally see it now as a namespace package.

jacobtylerwalls: But please test and see if you get the same result! 🙏🏻

Pierre-Sassoulas: (Slightly related we had a lot of issue with no``__init__.py`` in test directory, we could change that if it's causing more problem see https://github.com/PyCQA/pylint/issues/7085#issuecomment-1169863733)

jacobtylerwalls: I think that's fixed in #7114, which has nothing to do with astroid thankfully, I just figured it out by reading the code comments :D

Pierre-Sassoulas: I'm going to test on the 2.12 branch, I don't know why I did not do that earlier.

jacobtylerwalls: If it helps, I noticed there were subtle behavior differences in running the tests depending on whether you have an editable install. (e.g. whether you're picking up `pylint/tests` or `astroid/tests` as `tests` automatically before you get to namespace package resolution)

Pierre-Sassoulas: I removed the commit as I think you're right that astroid is not using the sys.path provided by pytest.

jacobtylerwalls (approved):

DanielNoord:
This merge broke `pre-commit` on `main`. I think the runner image was updated on `pre-commit` and now probably runs Python 3.10.

@Pierre-Sassoulas We should probably also update the version with which we run `pre-commit` in CI for `pylint` and remove the useless ignores that are now being flagged.

DanielNoord:
> Thanks for following that I missed it ! Well I happen to also have upgraded to python 3.10 last week so I think it's the right move. 

Officialy we support developing on `3.8` since that's supported by `pydocstringformatter` and necessary for the correct functional test output, but I think it makes sense to run our CI on `3.10` as it just frees up the runners sooner.

> The tests are SO MUCH FASTER locally, it's crazy (Maybe it was bloat that accumulated on my OS over the year or ubuntu 22.04).

🎉

Pierre-Sassoulas:
Yes sure, we're not removing python 3.7 support, but LTS ubuntu is on 3.10 since april, and 3.10 is faster so.. let's run everything that run only once on 3.10 :)

PR #7156, opened 2022-07-10
Increase new primer timeout to 60m

jacobtylerwalls:
See [last run](https://github.com/PyCQA/pylint/actions/runs/2644912539) on main. The 3.7 job sometimes takes just barely over 45 minutes. The old primer timeout was set to 60m.

Clearly we need to add some concurrency. Rather than parallelizing the primer message comparison infrastructure, my hope is that we can just (later this year or next) solve the underlying issues with `--jobs` (see label `topic: multiprocessing`) and use that.

### DIFF
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -21,7 +21,7 @@ jobs:
   run-primer:
     name: Run / ${{ matrix.python-version }}
     runs-on: ubuntu-latest
-    timeout-minutes: 45
+    timeout-minutes: 60
     strategy:
       matrix:
         python-version: ["3.7", "3.10"]


Pierre-Sassoulas:
We can't count on the github runner machine to have more than one core consistently, so we probably need both parallelization

Pierre-Sassoulas (approved):

jacobtylerwalls:
Oooh, that's a much larger ask then. I looked at that briefly a month ago and then realized there's no good solution for streaming JSON results to a file.

DanielNoord:
> We can't count on the github runner machine to have more than one core consistently, so we probably need both parallelization

I do think the "partial" cores the runner machines offer could be used if we created a more barebones way of parallelisation. Instead of creating a completely new `PyLinter` just pass a `FileItem` and some basic options and run `check_file`.

I really don't have time to look into this in the coming weeks due to deadlines and holidays, but I would certainly be interested in a (Discord/Zoom) call in August and September to discuss some possibilities. It seems this requires such large refactoring efforts that it would be good to both discuss the approach and perhaps even divide some tasks.

jacobtylerwalls:
My time is also limited these next few months but in general 👍🏻

DanielNoord:
@jacobtylerwalls https://github.com/PyCQA/pylint/runs/7271165374?check_suite_focus=true

Am I overseeing something or is `pip install -e .` in the `Run pylint primer` step overwriting the install of `astroid` on the latest commit on `main`? 😅

Pierre-Sassoulas:
Should we relax the requirement even more ? I don't think a lot could go wrong (as long as we set it correctly before release). We can use tbump for that.

PR #7158, opened 2022-07-11
Add documentation on using a minimal config file

### LINKED ISSUE #6968: Should `.pylintrc` contain all settings explicitly or just overrides
### Question

What's the best practice regarding Pylint's configuration settings?

1. All settings are explicitly specified, even when most or all settings use default values.
2. Only overrides are specified, i.e. default values are used implicitly unless explicitly overridden.

I've not been able to find an answer to this question anywhere. Personally, I find materializing all settings explicitly quite verbose and difficult to maintain given the large number of settings. Also, it would be hard to tell which settings just use default values and which settings use custom values. On the other hand, changes in default values with a new release of Pylint would not cause errors with explicit settings. But then again some releases introduce new settings whose values would not be set explicity either unless they are manually added to the config, so it's easy to end up with a mix of explicit and implicit (default) settings.

Would the best practice be to have a config only with settings overrides (so no explicit default values) and use a package version specifier that pins the major and minor version (i.e. the version specifier only contains bug fix releases, e.g. `>=2.14,<2.15`)?

I think the Pylint documentation should describe the best practices for using Pylint, including best practices for the config and version specifier.

### Additional context

_No response_

sisp:
## Type of Changes

## Description

I've added some documentation on best practices regarding the configuration file. Specifically, it recommends to use a minimal configuration file which only contains configuration overrides rather than a full configuration file where most options use the default values.

Closes #6968

### DIFF
--- a/doc/user_guide/configuration/index.rst
+++ b/doc/user_guide/configuration/index.rst
@@ -15,8 +15,8 @@ For example::
 
     pylint --disable=bare-except,invalid-name --class-rgx='[A-Z][a-z]+' --generate-toml-config
 
-In practice, you should create a minimal configuration file which contains only
-configuration overrides. For all other options, Pylint will use default values.
+In practice, it is often better to create a minimal configuration file which only contains
+configuration overrides. For all other options, Pylint will use its default values.
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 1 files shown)


doc/user_guide/configuration/index.rst
```diff
@@ -15,6 +15,9 @@ For example::
 
     pylint --disable=bare-except,invalid-name --class-rgx='[A-Z][a-z]+' --generate-toml-config
 
+In practice, you should create a minimal configuration file which contains only
+configuration overrides. For all other options, Pylint will use default values.
```
DanielNoord, line 19:
```suggestion
In practice, it is often better to create a minimal configuration file which only contains
configuration overrides. For all other options, Pylint will use its default values.
```

DanielNoord (approved):
👍

Issue #6799, opened 2022-06-02
Configuration parsing fails if options for an unused extension are present


Mr-Pepe:
### Bug description

I run pylint twice, once without `pylint.extensions.docparams` and once with. Both use the same `pyproject.toml` configuration file. This was not a problem until 2.14.0 but now the first call crashes because the docparams options are not recognized.

### Configuration

```ini
[tool.pylint.master]
accept-no-param-doc = "no"
accept-no-raise-doc = "no"
accept-no-return-doc = "no"
accept-no-yields-doc = "no"
```

### Command used

```shell
pylint src
```

### Pylint output

```shell
pyproject.toml:1:0: E0015: Unrecognized option found: accept-no-param-doc, accept-no-raise-doc, accept-no-return-doc, accept-no-yields-doc (unrecognized-option)
```

### Expected behavior

Unused options do not lead to a crash or section names in `pyproject.toml` are honored.

### Pylint version

```shell
pylint 2.14.0
astroid 2.11.5
Python 3.9.5 (default, Nov 18 2021, 16:00:48) 
[GCC 10.3.0]
```

### OS / Environment

Kubuntu 21.04

DanielNoord:
I think this sort of use-case might warrant an additional discussion.  Running two versions of `pylint` with a single configuration file was not one of the things we had in mind when we worked in the configuration refactor for `2.14`. I also don't know if we want to explicitly support it.

For now I'd suggest disabling `unrecognized-option`. You could also do ``pylint --accept-no-param-doc=no etc.`` in your second invocation. Then you could keep using the same file without disabling the message.

Mr-Pepe:
That's unfortunate but makes sense. I will disable `unrecognized-option` for the time being. Thank you for the prompt replies! :)

DanielNoord:
No worries! I'm going to put this on the milestone for our next minor release so we make sure to take a decision before releasing it (which will probably in a couple of months).

DanielNoord:
> I think it would make sense to hard-code every extensions checker's options, and every deleted messages msgid/symbol (we already have a deleted messages list) and treat them differently with a message that could be disabled without touching `bad-option-value` and `unrecognised-option`. We discussed that during implementation but I did not realize that the use case of using the same configuration for two calls of pylint existed.

[...]

PR #7163, opened 2022-07-11
Search for pyproject.toml config file in parent dirs

### LINKED ISSUE #3289: Looking for configuration files like ``pylintrc`` or ``pyproject.toml`` in parent directories like ``black`` or ``isort``
I use pylint in vi with a pyproject.toml -- pyproject.toml isn't always in my $CWD.

What do you think about also looking at ../../.. recursively like you do the .pylintrc?

### Additional context

You might want to exclude setup.cfg from this for backwards compatibility.

Felixoid:
## Type of Changes

## Description
Search for pyproject.toml in parent directories.

Closes #3289

### DIFF
--- a/doc/user_guide/usage/run.rst
+++ b/doc/user_guide/usage/run.rst
@@ -105,6 +105,9 @@ configuration file in the following order and uses the first one it finds:
    providing it has at least one ``pylint.`` section
 #. ``tox.ini`` in the current working directory,
    providing it has at least one ``pylint.`` section
+#. Pylint will search for the ``pyproject.toml`` file up the directories hierarchy
+   unless it's found, or ``.git``/``.hg`` directory is found, or the file system root
+   is approached.
 #. If the current working directory is in a Python package, Pylint searches \
    up the hierarchy of Python packages until it finds a ``pylintrc`` file. \
    This allows you to specify coding standards on a module-by-module \
--- a/doc/user_guide/usage/run.rst
+++ b/doc/user_guide/usage/run.rst
@@ -106,7 +106,7 @@ configuration file in the following order and uses the first one it finds:
 #. ``tox.ini`` in the current working directory,
    providing it has at least one ``pylint.`` section
 #. Pylint will search for the ``pyproject.toml`` file up the directories hierarchy
-   unless it's found, or ``.git``/``.hg`` directory is found, or the file system root
+   unless it's found, or a ``.git``/``.hg`` directory is found, or the file system root
    is approached.
 #. If the current working directory is in a Python package, Pylint searches \
    up the hierarchy of Python packages until it finds a ``pylintrc`` file. \
--- a/pylint/config/find_default_config_files.py
+++ b/pylint/config/find_default_config_files.py
@@ -16,7 +16,28 @@ else:
     import tomli as tomllib
 
 RC_NAMES = (Path("pylintrc"), Path(".pylintrc"))
-CONFIG_NAMES = (*RC_NAMES, Path("pyproject.toml"), Path("setup.cfg"))
+PYPROJECT_NAME = Path("pyproject.toml")
+CONFIG_NAMES = (*RC_NAMES, PYPROJECT_NAME, Path("setup.cfg"))
+
+
+def _find_pyproject() -> Path:
+    """Search for file pyproject.toml in the parent directories recursively.
+
+    It resolves symlinks, so if there is any symlink up in the tree, it does not respect them
+    """
+    current_dir = Path.cwd().resolve()
+    is_root = False
+    while not is_root:
+        if (current_dir / PYPROJECT_NAME).is_file():
+            return current_dir / PYPROJECT_NAME
+        is_root = (
+            current_dir == current_dir.parent
+            or (current_dir / ".git").is_dir()
+            or (current_dir / ".hg").is_dir()
+        )
+        current_dir = current_dir.parent
+
+    return current_dir
 
 
 def _toml_has_config(path: Path | str) -> bool:
@@ -96,6 +117,13 @@ def find_default_config_files() -> Iterator[Path]:
 
     try:
         yield from _find_project_config()
+    except OSError:
+        pass
+
+    try:
+        parent_pyproject = _find_pyproject()
+        if parent_pyproject.is_file() and _toml_has_config(parent_pyproject):
+            yield parent_pyproject.resolve()
     except OSError:
         pass
 
--- a/tests/config/test_find_default_config_files.py
+++ b/tests/config/test_find_default_config_files.py
@@ -130,6 +130,47 @@ def test_pylintrc_parentdir() -> None:
 
 
 @pytest.mark.usefixtures("pop_pylintrc")
+def test_pyproject_toml_parentdir() -> None:
+    """Test the search of pyproject.toml file in parent directories"""
+    with tempdir() as chroot:
+        with fake_home():
+            chroot_path = Path(chroot)
+            files = [
+                "pyproject.toml",
+                "git/pyproject.toml",
+                "git/a/pyproject.toml",
+                "git/a/.git",
+                "git/a/b/c/__init__.py",
+                "hg/pyproject.toml",
+                "hg/a/pyproject.toml",
+                "hg/a/.hg",
+                "hg/a/b/c/__init__.py",
+                "none/sub/__init__.py",
+            ]
+            testutils.create_files(files)
+            for config_file in files:
+                if config_file.endswith("pyproject.toml"):
+                    with open(config_file, "w", encoding="utf-8") as fd:
+                        fd.write('[tool.pylint."messages control"]\n')
+            results = {
+                "": chroot_path / "pyproject.toml",
+                "git": chroot_path / "git" / "pyproject.toml",
+                "git/a": chroot_path / "git" / "a" / "pyproject.toml",
+                "git/a/b": chroot_path / "git" / "a" / "pyproject.toml",
+                "git/a/b/c": chroot_path / "git" / "a" / "pyproject.toml",
+                "hg": chroot_path / "hg" / "pyproject.toml",
+                "hg/a": chroot_path / "hg" / "a" / "pyproject.toml",
+                "hg/a/b": chroot_path / "hg" / "a" / "pyproject.toml",
+                "hg/a/b/c": chroot_path / "hg" / "a" / "pyproject.toml",
+                "none": chroot_path / "pyproject.toml",
+                "none/sub": chroot_path / "pyproject.toml",
+            }
+            for basedir, expected in results.items():
+                os.chdir(chroot_path / basedir)
+                assert next(config.find_default_config_files(), None) == expected
+
+
+@pytest.mark.usefixtures("pop_pylintrc")
 def test_pylintrc_parentdir_no_package() -> None:
     """Test that we don't find a pylintrc in sub-packages."""
     with tempdir() as chroot:
--- a/.pyenchant_pylint_custom_dict.txt
+++ b/.pyenchant_pylint_custom_dict.txt
@@ -261,6 +261,7 @@ pylint
 pylintdict
 pylintrc
 pylint's
+pyproject
 pypy
 pyreverse
 pytest
@@ -318,6 +319,8 @@ subtree
 supcls
 superclass
 symilar
+symlink
+symlinks
 sys
 tbump
 tempfile
--- a/doc/whatsnew/fragments/7163.other
+++ b/doc/whatsnew/fragments/7163.other
@@ -1,0 +1,3 @@
+Search for ``pyproject.toml`` recursively in parent directories up to a project or file system root.
+
+Refs #7163, Closes #3289


Felixoid:
How would I add exceptions for false-positive spell checker?

DanielNoord:
Thanks for the interest in this issue @Felixoid. I didn't notice we left that issue open, but due to discussion in https://github.com/PyCQA/pylint/pull/3769 I am not sure if we can merge this PR.
Even if we do want to I think we should consider whether this is a breaking change and if it can be included in `2.x`.

Pierre-Sassoulas:
I think this should use the specification from the discussion in https://github.com/PyCQA/pylint/issues/3289

Pierre-Sassoulas:
> How would I add exceptions for false-positive spell checker?

There's a dictionary at the root of the project.

Felixoid:
Hi @DanielNoord, there's nothing relatively new in the code that wasn't there already. The snippet `curdir = Path(os.getcwd()).resolve()` is a bit below, so it should work fine.

Dear @Pierre-Sassoulas, you mean that `.git/.hg` as another marker for the `pyproject.toml` directory? Or the max depth? Or rather that it should search for any possible configuration up to a repository root directory?

Pierre-Sassoulas:
We handle multiple configuration file (``pylintrc``/``.pylintrc`` / ``pyproject.toml``, ``setup.cfg`` too) I think we should upgrade the current function to search for configuration files to make it search up the tree until it hits a configuration file or a ``.git`` / ``.hg`` directory. Unless I'm mistaken it's this function that need to be modified:

https://github.com/PyCQA/pylint/blob/2e2f8f8c12a2e54c18c0e4dd49e4850f5c913b1f/pylint/config/find_default_config_files.py#L42

Felixoid:
Thank you for your review, I've added processing of all supported configs, and updated the docs

Felixoid:
It definitely went out of control. I'll take care of it.

DanielNoord:
The failing tests show that this is not a trivial change. Again, I'm against changing this in a `2.x` update. Consider:
 
A directory without a `git repo` and no `pyproject.toml`. For example, a collection of (internal) scripts.
An user can now run `pylint` on this directory and get the default values and pass the checks.
With this change we would now start going up until the user's root, thereby potentially changing the config file that is being used for their checks.

I'm usually fine with having users change some configuration stuff between minor patches, but forcing them to restructure their directories or move files around doesn't seem like a good thing to do in a minor version update.

Felixoid:
Ok, I misinterpret your first message. I thought that you've meant python2.

You are right on one hand. On another, now nobody can set a project global pylintrc, of there is a mixed repo. That's why I started working on the issue, basically

DanielNoord:
> Ok, I misinterpret your first message. I thought that you've meant python2.
> 
> You are right on one hand. On another, now nobody can set a project global pylintrc, of there is a mixed repo. That's why I started working on the issue, basically

See L62-80 of the original code. We currently support an environment variable and a global config in the users home directory. This probably doesn't cover all cases (and it might still make sense to merge this PR at some point), but I'm not sure we can change this now.

I haven't tested this myself but can't you also use `pylint --rcfile=../../../pylintrc`?

DanielNoord:
> It's precisely the issue I am trying to solve. And to show that it works, we already have the best proof. Take a look on [the report](https://github.com/PyCQA/pylint/pull/7163#issuecomment-1180916413). Each tested project has less number of reports than it has. It works as expected.

This is an unrelated report and caused by a bug in our primer. Actually I would expect this PR to have no effect in the primer as all primer projects are linted with a specified `pylintrc`. For now we should ignore all primer comments in this PR until #7160 has been merged.

> What can be done, for now I'd roll back to the approximate state from [the commit](https://github.com/PyCQA/pylint/compare/d5f99f6560afff9d88857af3dcfa22f6807db691..486ab25477e04a88d85786cd81545d04afa9ee12#diff-a76d406b648f773c81008092021cb6142d2c4d8e3d3e5bf241e88944a2afb541L33) and add `.git`/`.hg` there only for `pyproject.toml` configuration. It's not such a significant change, and one is looking for a way to set a repo-wide config will have it.
> 
> The major change to search all configs in the directories tree may be then addressed in a major release.
> 
> Does it sound reasonable?

This is still a breaking change for those who want to lint subdirectories with default values. I can't stress enough the range of different `pylint` invocations. Take for example https://github.com/PyCQA/pylint/issues/6799, which shows that some users actually run `pylint` twice over the same repository with the same configuration file but with different command line arguments. Initially that seemed totally bizarre to me, but I guess people find a use case for it. Same goes for linting subdirectories with default values, we just don't know whether users rely on this.

I can see the need for repository wide configuration files, but I don't think this can be default value. Perhaps we can work with a flag? `--look-for-parent-configuration` or something?

Felixoid:
Sorry, I can't call it breaking changes. For me, it is still an essential function that wasn't implemented for historical reasons.

The point is that each tool I can recall, e.g., black and isort (see the issue), [flake8](https://github.com/PyCQA/flake8/blob/ca573a7ccf2d4a1c7df0b577bb6d3455c941e828/src/flake8/options/config.py#L17-L31) already supports it at least somehow. 

I hear your point. I can only agree to disagree. I see people need the same thing as I do, so this PR solves the issue that has been hanging around for three years.

What I can do more here is write tests to put to stone the proposed behavior with `pyproject.toml or .git or .hg in a directory above`. I wouldn't invest my time to invest then even more time into wrapping around workarounds.

DanielNoord:
This discussion has gotten stuck I think. I'm tagging some other active maintainers to gather more opinions about this.

@Pierre-Sassoulas @jacobtylerwalls

jacobtylerwalls:
Hey @Felixoid, thanks for these changes. I think everyone is positive on your changes, we're just trying to figure out what release to put it in.

The good news is that even if it goes in 3.0, we'll probably put out a pre-release, and you can start using it. As I said [here](https://github.com/PyCQA/pylint/issues/3512#issuecomment-1173085153), merging great patches to a 3.0 branch just increases the urgency of releasing it. :-)

I see the argument on both sides. Honestly, this is what I thought the configurations were:

> I can see the need for repository wide configuration files

But @DanielNoord has a point that people's CI workflows will break given that often they don't even know what their `cwd` is. (Even in developing pylint itself recently we've had several issues that hinged on cwd-sensitivity.) And people can get cross when we make changes that affect how they invoke pylint.

I don't feel too strongly either way, but I just want to reiterate that this part of the discussion should slow down reviewing the changes and getting them merged either to `main` or to `3.0`.

Pierre-Sassoulas:
> merging great patches to a 3.0 branch just increases the urgency of releasing it.

👍 Strong agree. We have a LOT to release in 3.0, we need to add things that sparkle joy or we'll never ship it 😄

But this is not the reason I'm slightly on the side of doing the feature as a breaking change here. Environment bugs are really nasty to tests, troubleshoot, release, and get right in general. We can't anticipate all the possible environments used (we had solaris and another really niche OS recently, there's probably a Mars Rover prototype with custom hardware somewhere too). This is probably simpler in this case, which is why I'm only slightly in favor, but there's a very high probability that we're going to break something somewhere and this is not good practice to take that risk in a minor (especially as pylint add new messages with new checkers and fixed false negatives, so the migration to a new minor is not painless already).

Felixoid:
Thank you for your feedback, I got your point. I agree to redo it as a major change.

Unfortunately, I got serious sick today, and wouldn't be able to do it this week. Maybe, the weekend sounds realistic.

Felixoid:
As far as I see, the PR currently has a base `main`, so I guess it will end up in the 3.0, right?

Pierre-Sassoulas:
> As far as I see, the PR currently has a base main, so I guess it will end up in the 3.0, right?

We're putting everything from main in 3.0 right now, yes.

Felixoid:
Hello again.

Is there anything preventing it from merging now?

Pierre-Sassoulas:
I added the need review label, but I don't have the availability right now :)

DanielNoord:
I think the main issue is that we don't really have branch to put this on. We recently removed the `3.x` branch as it took effort to keep it up to date while we are no where near releasing `3.x` currently.

jacobtylerwalls:
That shouldn't stop a review, though. I think availability of a review is the immediate blocker.

Pierre-Sassoulas:
Creating a 3.x branch from the current main branch should be easy once we need it (Ideally I would create this branch officially only when we really need it because then we'll have to merge main back in it for quite a long time)

Felixoid:
I updated the branch, so it's adjusted to recent fixes. Can it process further, please?

pylint/config/find_default_config_files.py
```diff
@@ -17,7 +17,28 @@
     import tomli as tomllib
 
 RC_NAMES = (Path("pylintrc"), Path(".pylintrc"))
-CONFIG_NAMES = RC_NAMES + (Path("pyproject.toml"), Path("setup.cfg"))
+PYPROJECT_NAME = Path("pyproject.toml")
```
Pierre-Sassoulas, line 20:
```suggestion
PYPROJECT_NAME = "pyproject.toml"
```

tests/config/test_find_default_config_files.py
```diff
@@ -129,6 +129,49 @@ def test_pylintrc_parentdir() -> None:
...
+                "git/pyproject.toml",
+                "git/a/pyproject.toml",
+                "git/a/.git",
+                "git/a/b/c/__init__.py",
+                "hg/pyproject.toml",
+                "hg/a/pyproject.toml",
+                "hg/a/.hg",
+                "hg/a/b/c/__init__.py",
+                "none/sub/__init__.py",
+            ]
```
Pierre-Sassoulas, line 149: Shouldn't 'files' and 'expected' be in a ``@pytest.mark.parametrize`` instead ?

Pierre-Sassoulas (changes requested):
Thank you for implementing this, I have a nitpick around the test implementation but LGTM.

tests/config/test_find_default_config_files.py
Felixoid: I don't know, sorry. It's the same was as `test_pylintrc_parentdir` is tested RN. Can we consider refactoring both tests later, maybe?

Pierre-Sassoulas: I did not realize that it's the exact same list than another test. We might want to refactor before the change and use the same constant in both, then ?

pylint/config/find_default_config_files.py
Pierre-Sassoulas: I missed that, it felt like we did not need a path in this particular case, but let's keep it that way.

tests/config/test_find_default_config_files.py
Felixoid: It's not used anywhere else. It's just the same way of testing that is used in other places in this file. I don't see any other example of `files` and `expected` being a `@pytest.mark.parametrize`

See `test_pylintrc_parentdir`, `test_pylintrc_parentdir_no_package`, `test_verbose_output_no_config`, and `test_verbose_abbreviation`

Pierre-Sassoulas: I'm going to resolve then.

Pierre-Sassoulas (approved):

DanielNoord:
Thanks for the review @Pierre-Sassoulas.

As far as I can see though there is still no real decision on whether this should go in `2.16.0`. Imo, this is a hard to troubleshoot bug that can potentially break CIs. I really don't think the benefit it offers outweigh putting a breaking change in a minor patch.

Pierre-Sassoulas:
Right I forgot about that and just added it automatically to the next release. We need to create the 3.0 branch and deal with the 3.0 deprecations then.

DanielNoord:
> Right I forgot about that and just added it automatically to the next release. We need to create the 3.0 branch and deal with the 3.0 deprecations then.

We should probably create a separate issue for this discussion but I think we should at least make sure `astroid` can run `mypy`. After that we can be much more assured that the typing in `pylint` is correct and that we have collected all associated deprecations.
After that we can start thinking about any other new features we want to put in there.

Felixoid:
`Needs review` label is redundant here now, right?

Felixoid:
I am a bit confused, what is blocking the merge now?

Pierre-Sassoulas:
Sorry for the delay @Felixoid. We need to create a 3.0 branch to merge and it's something that is going to create a lot of work if we need to cherry-pick everything on it. We have a lot to do with maintaining the 2.x branch as is, so it does not get done. We also have a ton of deprecation to handle before releasing 3.0.

Felixoid:
It looks like now `.pylintrc` is not parsed unless `__init__.py` is found in the version 2.15.5 after https://github.com/PyCQA/pylint/pull/7423, and it breaks behavior for ClickHouse/ClickHouse. So, in the recent versions neither https://github.com/ClickHouse/ClickHouse/blob/a3e3be5/.pylintrc works nor pyproject.toml could be set up in the root directory.

Not sure if it was intended, but the current state is a quite strange from the configuration perspective.

Pierre-Sassoulas:
Sorry for the delay again @Felixoid. The 3.0 release cycle is coming, are you still available to work on this or should we take over ?

Felixoid:
Looks green as a coming spring!

Felixoid:
Hello, are there any updates on it?..

doc/user_guide/usage/run.rst
```diff
@@ -105,6 +105,9 @@ configuration file in the following order and uses the first one it finds:
    providing it has at least one ``pylint.`` section
 #. ``tox.ini`` in the current working directory,
    providing it has at least one ``pylint.`` section
+#. Pylint will search for the ``pyproject.toml`` file up the directories hierarchy
+   unless it's found, or ``.git``/``.hg`` directory is found, or the file system root
```
DanielNoord, line 109:
```suggestion
   unless it's found, or a ``.git``/``.hg`` directory is found, or the file system root
```

DanielNoord (approved):
I think we can merge this, but defer to @Pierre-Sassoulas for doing so!

Pierre-Sassoulas (approved):
Sorry for taking a long time to review and merge, this is going to be in the first beta version ``3.0.0b1``

PR #7165, opened 2022-07-11
test_run_pylint_config: ignore pytest args

AdamWill:
## Type of Changes

## Description

This test fails if run through pytest with any args, because
`_run_pylint_config` passes through args from the command line
without checking if the args are actually intended for
pylint-config. We have to pass it some fake args that evaluate
truth-y but don't break anything, to avoid this.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/tests/config/pylint_config/test_run_pylint_config.py
+++ b/tests/config/pylint_config/test_run_pylint_config.py
@@ -18,7 +18,7 @@ def test_invocation_of_pylint_config(capsys: CaptureFixture[str]) -> None:
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", message="NOTE:.*", category=UserWarning)
         with pytest.raises(SystemExit) as ex:
-            _run_pylint_config()
+            _run_pylint_config([""])
         captured = capsys.readouterr()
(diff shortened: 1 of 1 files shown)


AdamWill:
Didn't add a changelog entry as the change is only to a test. To see the problem here, just try:

    pytest -v tests/config/pylint_config/test_run_pylint_config.py

and watch it fail because `pylint-config` tries to handle the `-v` arg and gives unexpected output. With this change, the test passes no matter what args you give pytest.

AdamWill:
I guess technically it might be better to change `_run_pylint_config` somehow, but as it's only used in two ways - in this test, and as the entry point for the `pylint-config` command - I don't think it's really necessary?

jacobtylerwalls (approved):
Brilliant, I saw this from time to time but never found the energy to look into it.

Pierre-Sassoulas (approved):
This bothered me for month literally. Test worked as a standalone but not when launching everything and I thought there was a lot of work to fix it. Great job!

DanielNoord:
Worked on my machine 😅 

Thanks for fixing!

PR #7168, opened 2022-07-12
Use trace not turtle in import_outside_toplevel test

AdamWill:
## Type of Changes

## Description

Downstream (e.g. in Fedora) tkinter and turtle aren't always
included in the core Python package in order to avoid pulling in
a lot of graphical dependencies on systems that do not need them.
It doesn't really matter what libs are used in this test, so
let's replace it with trace.

### DIFF
--- a/tests/functional/i/import_outside_toplevel.py
+++ b/tests/functional/i/import_outside_toplevel.py
@@ -27,7 +27,7 @@ class C:
     import tokenize  # [import-outside-toplevel]
 
     def j(self):
-        import turtle  # [import-outside-toplevel]
+        import trace  # [import-outside-toplevel]
 
 
 def k(flag):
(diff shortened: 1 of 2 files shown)


AdamWill:
The Fedora package had the test disabled because of this failure (I guess the maintainer hadn't realized why the test was failing). We could just pull in our python-tkinter package as a build dependency, but I figured this was just as easy and doesn't bloat our build dependencies...

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
Make sense ! We had a similar report with solaris where apt was a builtin package. Sometime finding an import that work everywhere is not easy :)

DanielNoord (approved):

PR #7171, opened 2022-07-12
Use the pylint-pytest-plugin

DanielNoord:

## Type of Changes

## Description

@Pierre-Sassoulas Is this something we would want to add? Just showing off that it is possible.

### DIFF
--- a/pylint/testutils/lint_module_test.py
+++ b/pylint/testutils/lint_module_test.py
@@ -246,7 +246,7 @@ class LintModuleTest:
         return received_msgs, received_output_lines
 
     def _runTest(self) -> None:
-        __tracebackhide__ = True  # pylint: disable=unused-variable
+        __tracebackhide__ = True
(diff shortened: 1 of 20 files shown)


Pierre-Sassoulas (commented):
Did you consider contributing to the existing ``pylint-pytest`` ? (This look good enough but we might as well not step on the top of the existing plugin)

jacobtylerwalls:
> Effect on [pandas](https://github.com/pandas-dev/pandas):
17 error(s) were found stemming from the astroid library. This is unlikely to have been caused by your changes. A GitHub Actions warning links directly to the crash report template. Please open an issue against astroid if one does not exist already.

Should we stop being so modest and stop saying "unlikely to have been caused by your changes"? 🤔

DanielNoord:
> Should we stop being so modest and stop saying "unlikely to have been caused by your changes"? 🤔

😅

PR #7175, opened 2022-07-13
More parametrization in ``test_self.py`` tests

Pierre-Sassoulas:
## Type of Changes

## Description

This is the consensual part of https://github.com/PyCQA/pylint/pull/7173. First step before a MR to fix the pythonpath handling were the pythonpath is not rested correctly.

### DIFF
--- a/tests/test_self.py
+++ b/tests/test_self.py
@@ -817,53 +817,24 @@ class TestRunTC:
                 modify_sys_path()
             assert sys.path == paths[1:]
 
-    @staticmethod
-    def test_do_not_import_files_from_local_directory(tmpdir: LocalPath) -> None:
-        p_astroid = tmpdir / "astroid.py"
-        p_astroid.write("'Docstring'\nimport completely_unknown\n")
-        p_hmac = tmpdir / "hmac.py"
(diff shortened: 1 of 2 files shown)


tests/test_self.py
```diff
@@ -991,25 +887,21 @@ def test_allow_import_of_files_found_in_modules_during_parallel_check(
...
-                    "test_directory",
-                ],
-                code=0,
-            )
+            args = [
+                "-j2",
+                "--disable=missing-docstring, missing-final-newline",
+                "test_directory",
+            ]
+            self._runtest(args, code=0)
```
DanielNoord, line 895: Why the additional assignment? Same for the following tests.

Pierre-Sassoulas: Better formatting and easier comprehension of whats happenning

DanielNoord: Hm, I didn't think these were too problematic. In the future I would avoid such refactors, the only thing does currently does is make the test run slightly slower 😅

DanielNoord (approved):

Pierre-Sassoulas:
The coverage will be better in #7176 (Also a bug will be fixed).

PR #7167, opened 2022-07-11
Fix tests for Python 3.11, don't emit `super-init-not-called` for `Enum` subclasses

AdamWill:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

This fixes some issues in the tests with Python 3.11, and makes one change to the actual checks that is exposed by Python 3.11 - not emitting `super-init-not-called` for subclasses of `Enum`. `Enum` has a `__init__` in Python 3.11, but it does nothing at all (just `pass`), and the examples in Python's own `Enum` docs include sample subclasses with their own `__init__` which do not call `Enum.__init__`, so clearly it's OK/intended for you to do that. I've [asked upstream why this useless `Enum.__init__` was added at all](https://github.com/python/cpython/pull/30582#issuecomment-1181055042).

PR #7181, opened 2022-07-13
Don't emit super-init-not-called for Enum subclasses

AdamWill:
## Type of Changes

## Description

For some reason, commit 83d544b to cpython added a `__init__`
to `Enum` which does nothing (it just says `pass`). The examples
in the Enum docs:
https://docs.python.org/3.11/howto/enum.html
still do not include calling super's `__init__` for Enum
subclasses, that define their own `__init__`, and obviously there
is no practical point to calling a method that does nothing, so
it seems best just to skip this warning for Enum.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

Blocked by https://github.com/python/cpython/pull/30582#issuecomment-1181055042

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -2059,6 +2059,11 @@ a metaclass class method.",
                 except astroid.InferenceError:
                     continue
 
+            # Skip if klass is Enum, Python's own docs and examples
+            # do not recommend Enum subclasses call Enum.__init__
+            if klass.name == "Enum":
+                continue
+
             if decorated_with(node, ["typing.overload"]):
                 continue
             cls = node_frame_class(method)
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -33,6 +33,8 @@ Extensions
 False positives fixed
 =====================
 
+* Don't report ``super-init-not-called`` for subclasses of ``Enum`` (on Python >= 3.11).
+
 * Don't report ``unsupported-binary-operation`` on Python <= 3.9 when using the ``|`` operator
   with types, if one has a metaclass that overloads ``__or__`` or ``__ror__`` as appropriate.
 


AdamWill:
This is split out from https://github.com/PyCQA/pylint/pull/7167 , as requested by @DanielNoord .

AdamWill:
Uh, I guess that line needs to be changed on main so the bot doesn't add it to every single PR?

Pierre-Sassoulas:
> Uh, I guess that line needs to be changed on main so the bot doesn't add it to every single PR?

Thanks for noticing that's my bad.

pylint/checkers/classes/class_checker.py
```diff
@@ -2059,6 +2059,11 @@ def _check_init(self, node: nodes.FunctionDef, klass_node: nodes.ClassDef) -> No
                 except astroid.InferenceError:
                     continue
 
+            # Skip if klass is Enum, Python's own docs and examples
+            # do not recommend Enum subclasses call Enum.__init__
+            if klass.name == "Enum":
```
jacobtylerwalls, line 2064: I think elsewhere we check the qname to avoid getting red herrings from other modules.
```suggestion
            if klass.qname() == "enum.Enum":
```

jacobtylerwalls (commented):
Thank you!

@Pierre-Sassoulas can we remove the blocked label here? This doesn't depend on 3.11 support, IMO.

Pierre-Sassoulas:
It was blocked by https://github.com/python/cpython/pull/30582#issuecomment-1181055042 I think

AdamWill:
@DanielNoord suggests that https://github.com/PyCQA/pylint/pull/7227 would likely avoid the need for this.

Pierre-Sassoulas:
A functional test to make sure of that would be nice :)

DanielNoord:
@AdamWill Since #7227 was merged, would you be open to creating the functional test Pierre mentioned? That would indeed still be valuable.

DanielNoord:
Going to close this as we actually have various regression tests against this throughout our code. Thanks for the initial investigation @AdamWill!

AdamWill:
No problem. Sorry I didn't get around to writing any tests, have had too many piles of things on fire to deal with :|

PR #7186, opened 2022-07-14
Fix config section name in doc

tushar-deepsource:
The change should also be reflected in rendered docs, like in `doc/user_guide/configuration/all-options.rst`, but I think those are generated from this doc.


Pierre-Sassoulas (approved):
👍

PR #7190, opened 2022-07-14
[typing] Add typing to the class checkers

Pierre-Sassoulas:
## Type of Changes

 ✓   | :sparkles: New feature |

## Description

Refs #2079


pylint/checkers/classes/class_checker.py
```diff
@@ -192,7 +195,9 @@ def _is_trivial_super_delegation(function: nodes.FunctionDef) -> bool:
 # Deal with parameters overriding in two methods.
 
 
-def _positional_parameters(method):
+def _positional_parameters(
+    method: bases.BoundMethod | bases.UnboundMethod | nodes.FunctionDef,
```
DanielNoord, line 199: Why are these also methods? It is currently typed as `FunctionDef` in all its callers.

pylint/checkers/classes/class_checker.py
```diff
@@ -192,7 +195,9 @@ def _is_trivial_super_delegation(function: nodes.FunctionDef) -> bool:
...
+) -> list[nodes.AssignName]:
```
DanielNoord, line 200: This can also be `None`. See the docstring of `args.args` in `astroid`.

pylint/checkers/classes/class_checker.py
```diff
@@ -379,11 +386,11 @@ def _different_parameters(
     return output_messages
 
 
-def _is_invalid_base_class(cls):
+def _is_invalid_base_class(cls: InferenceResult) -> bool:
```
DanielNoord, line 389:
```suggestion
def _is_invalid_base_class(cls: nodes.ClassDef) -> bool:
```

pylint/checkers/classes/class_checker.py
```diff
@@ -402,7 +409,11 @@ def _has_data_descriptor(cls, attr):
     return False
 
 
-def _called_in_methods(func, klass, methods):
+def _called_in_methods(
+    func: nodes.FunctionDef | nodes.Module,
```
DanielNoord, line 413: Shouldn't this be `LocalsDictNodeNG`?

pylint/checkers/classes/class_checker.py
```diff
@@ -402,7 +409,11 @@ def _has_data_descriptor(cls, attr):
...
+    klass: nodes.ClassDef,
+    methods: tuple[str, str, str, str],
```
DanielNoord, line 415:
```suggestion
    methods: Sequence[str],
```

pylint/checkers/classes/class_checker.py
```diff
@@ -431,7 +442,7 @@ def _called_in_methods(func, klass, methods):
     return False
 
 
-def _is_attribute_property(name, klass):
+def _is_attribute_property(name: str, klass: nodes.ClassDef) -> bool:
```
DanielNoord, line 445: Can't `name` be `None`?

pylint/checkers/classes/class_checker.py
```diff
@@ -467,7 +478,9 @@ def _is_attribute_property(name, klass):
     return False
 
 
-def _has_same_layout_slots(slots, assigned_value):
+def _has_same_layout_slots(
+    slots: list[nodes.Const], assigned_value: nodes.Name
```
DanielNoord, line 482: Is this correct? Some of the patterns in the calling function seem to indicate this can also be `list[None]` for example.

pylint/checkers/classes/class_checker.py
```diff
@@ -826,7 +839,7 @@ def open(self) -> None:
         self._py38_plus = py_version >= (3, 8)
 
     @cached_property
-    def _dummy_rgx(self):
+    def _dummy_rgx(self) -> Pattern:
```
DanielNoord, line 842:
```suggestion
    def _dummy_rgx(self) -> Pattern[str]:
```

pylint/checkers/classes/class_checker.py
```diff
@@ -1322,7 +1335,9 @@ def form_annotations(arguments):
                 confidence=INFERENCE,
             )
 
-    def _check_property_with_parameters(self, node):
+    def _check_property_with_parameters(
+        self, node: nodes.AsyncFunctionDef | nodes.FunctionDef
```
DanielNoord, line 1339:
```suggestion
        self, node: nodes.FunctionDef
```

pylint/checkers/classes/class_checker.py
```diff
@@ -1331,7 +1346,11 @@ def _check_property_with_parameters(self, node):
         ):
             self.add_message("property-with-parameters", node=node)
 
-    def _check_invalid_overridden_method(self, function_node, parent_function_node):
+    def _check_invalid_overridden_method(
+        self,
+        function_node: nodes.AsyncFunctionDef | nodes.FunctionDef,
```
DanielNoord, line 1351: Let's just use `FunctionDef` for all of these

pylint/checkers/classes/class_checker.py
```diff
@@ -1547,7 +1568,7 @@ def _check_invalid_class_object(self, node: nodes.AssignAttr) -> None:
             return
         self.add_message("invalid-class-object", node=node)
 
-    def _check_in_slots(self, node):
+    def _check_in_slots(self, node: nodes.AssignAttr) -> None:
```
DanielNoord, line 1571: This seems to be `SuccessfulInferenceResult`?

pylint/checkers/classes/class_checker.py
```diff
@@ -1664,7 +1685,7 @@ def _check_classmethod_declaration(self, node):
         if any(method_name == member.name for member in parent_class.mymethods()):
             self.add_message(msg, node=node.targets[0])
 
-    def _check_protected_attribute_access(self, node: nodes.Attribute):
+    def _check_protected_attribute_access(self, node: nodes.Attribute) -> None:
```
DanielNoord, line 1688: This can also be `AssignAttr`.

pylint/checkers/classes/class_checker.py
```diff
@@ -1775,15 +1796,15 @@ def _is_type_self_call(self, expr: nodes.NodeNG) -> bool:
         )
 
     @staticmethod
-    def _is_classmethod(func):
+    def _is_classmethod(func: nodes.FunctionDef) -> bool:
```
DanielNoord, line 1799: Shouldn't this be `LocalsDictNodeNG`? Otherwise the check on 1802 makes no sense.

pylint/checkers/classes/class_checker.py
```diff
@@ -1775,15 +1796,15 @@ def _is_type_self_call(self, expr: nodes.NodeNG) -> bool:
...
         """Check if the given *func* node is a class method."""
 
         return isinstance(func, nodes.FunctionDef) and (
             func.type == "classmethod" or func.name == "__class_getitem__"
         )
 
     @staticmethod
-    def _is_inferred_instance(expr, klass):
+    def _is_inferred_instance(expr: nodes.Name, klass: nodes.ClassDef) -> bool:
```
DanielNoord, line 1807: `expr` can be other things I think.

pylint/checkers/classes/special_methods_checker.py
```diff
@@ -16,11 +20,16 @@
...
     safe_infer,
 )
+from pylint.lint.pylinter import PyLinter
 
 NEXT_METHOD = "__next__"
 
 
-def _safe_infer_call_result(node, caller, context=None):
+def _safe_infer_call_result(
+    node: nodes.AsyncFunctionDef | nodes.FunctionDef,
```
DanielNoord, line 29: See above.

pylint/checkers/classes/special_methods_checker.py
```diff
@@ -16,11 +20,16 @@
...
+    caller: nodes.AsyncFunctionDef | nodes.FunctionDef,
+    context: Any | None = None,
```
DanielNoord, line 31:
```suggestion
    context: InferenceContext | None = None,
```

pylint/checkers/classes/special_methods_checker.py
```diff
@@ -16,11 +20,16 @@
...
+) -> Any:
```
DanielNoord, line 32: This should be some form of `InferenceResult`.

pylint/checkers/classes/special_methods_checker.py
```diff
@@ -208,7 +219,7 @@ def _check_unexpected_method_signature(self, node):
             # to take all of them in consideration.
             emit = mandatory not in expected_params
             # pylint: disable-next=consider-using-f-string
-            expected_params = "between %d or %d" % expected_params
+            expected_params = "between %d or %d" % expected_params  # type: ignore[assignment]
```
DanielNoord, line 222: ?

DanielNoord (changes requested):
This was quite a lot to review. I have made an initial pass and think I got all but I'll probably need to do another pass next time.

Let's merge this one before doing the other typing one.

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: There's a guard against that earlier

https://github.com/PyCQA/pylint/blob/6417e8cdee987afb773ad5ab9f5f0e82028c7cac/pylint/checkers/classes/class_checker.py#L2096

Then we call ``_different_parameters`` :

https://github.com/PyCQA/pylint/blob/6417e8cdee987afb773ad5ab9f5f0e82028c7cac/pylint/checkers/classes/class_checker.py#L2107

Then ``_positional_parameters``

https://github.com/PyCQA/pylint/blob/6417e8cdee987afb773ad5ab9f5f0e82028c7cac/pylint/checkers/classes/class_checker.py#L321

https://github.com/PyCQA/pylint/blob/6417e8cdee987afb773ad5ab9f5f0e82028c7cac/pylint/checkers/classes/class_checker.py#L322

So effectively it can only be a ``list[nodes.AssignName]``. Maybe we need to add an assert there to prevent the function being used in another context without the check ?

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: There's a check before the call making sure that it's not None:

https://github.com/PyCQA/pylint/blob/6417e8cdee987afb773ad5ab9f5f0e82028c7cac/pylint/checkers/classes/class_checker.py#L1739-L1743

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: Fixed by a local commit.

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: Done

pylint/checkers/classes/special_methods_checker.py
Pierre-Sassoulas: mypy has a hard time with the length of the expected_params tuple.

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: Here's what mypy raise if we add the ``None``:
```
pylint/checkers/classes/class_checker.py:338: error: Item "None" of "Optional[List[Any]]" has no attribute "__iter__" (not iterable)  [union-attr]
pylint/checkers/classes/class_checker.py:340: error: Item "None" of "Optional[List[Any]]" has no attribute "__iter__" (not iterable)  [union-attr]
pylint/checkers/classes/class_checker.py:350: error: Argument 1 to "_has_different_parameters" has incompatible type "Optional[List[Any]]"; expected "List[Any]"  [arg-type]
pylint/checkers/classes/class_checker.py:350: error: Argument 2 to "_has_different_parameters" has incompatible type "Optional[List[Any]]"; expected "List[Any]"  [arg-type]
```

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: You're right.

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: I think you're right, it would explain the ``if all(
            first_slot and second_slot and first_slot.value == second_slot.value
            for (first_slot, second_slot) in zip_longest(slots, other_slots)
        ):`` later

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: It's the ``expr`` attribute of a ``node: nodes.Attribute | nodes.AssignAttr`` so I think it's always a ``Name`` even though ``expr`` is typed ``NodeNg | None``. I'm going to be be safe and use ``NodeNg | None``

pylint/checkers/classes/special_methods_checker.py
Pierre-Sassoulas: ``-> Iterator[nodes.NodeNG | Uninferable | None] | None:`` seems to work, the typing in astroid is currently:

```
    def infer_call_result(self, caller=None, context=None):
        """Infer what the function returns when called.

        :returns: What the function returns.
        :rtype: iterable(NodeNG or Uninferable) or None
        """
```

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: Hmm, I think the function never return anything ? I see only empty ``return`` or do you mean ``node`` ?

pylint/checkers/classes/class_checker.py
DanielNoord: Ah yeah. Sorry, should have seen that. Then this typing is very useful 😄

pylint/checkers/classes/class_checker.py
DanielNoord: `AssignName.name: str | None`. I think it probably shouldn't be `None`, but currently it is typed as such.

Whenever I encounter such issues I tend to leave out the typing as we first need to fix this in `astroid`.

pylint/checkers/classes/class_checker.py
```diff
@@ -467,7 +476,9 @@ def _is_attribute_property(name, klass):
     return False
 
 
-def _has_same_layout_slots(slots, assigned_value):
+def _has_same_layout_slots(
+    slots: list[nodes.Const] | list[None], assigned_value: nodes.Name
```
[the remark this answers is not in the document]
DanielNoord: And how did you determine `assigned_value`?

pylint/checkers/classes/special_methods_checker.py
DanielNoord: That's incorrect (_Never trust an astroid docstring 😄_).

There is at least one call to `infer()` in `infer_call_result` without any assertions so that can yield an `InferenceResult`.
I think we should leave this signature until we have `infer_call_result` typed correctly.

pylint/checkers/classes/special_methods_checker.py
DanielNoord: If we use an f-string does the message go away?

pylint/checkers/classes/special_methods_checker.py
```diff
@@ -308,55 +319,67 @@ def _is_iterator(node):
                     pass
         return False
 
-    def _check_iter(self, node, inferred):
+    def _check_iter(self, node: nodes.FunctionDef, inferred: Any) -> None:
         if not self._is_iterator(inferred):
             self.add_message("non-iterator-returned", node=node)
 
-    def _check_len(self, node, inferred):
+    def _check_len(self, node: nodes.FunctionDef, inferred: nodes.Const) -> None:
```
DanielNoord, line 326: Just picking this one as it is the easiest to illustrate, I don't really understand where these typings are coming from? Don't they all come from `_protocol_map` and called with `Callable[[nodes.FunctionDef, Any], None]`?

The `Any` being the result of `infer_call_result`? Why can this only be `Const`?

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: Yeah, that's a 10mn expedition using a good IDE I don't think you're supposed to just 'see' it :smile:

pylint/checkers/classes/special_methods_checker.py
Pierre-Sassoulas: No, mypy do not catch the isinstance( ... tuple) just above. But at least we would not crash for longer tuples.

DanielNoord:
@Pierre-Sassoulas Please don't rebase or force push these larger PRs, pretty difficult to keep track of the changes I already reviewed that way 😅

pylint/checkers/classes/special_methods_checker.py
```diff
@@ -16,11 +22,16 @@
...
 
 NEXT_METHOD = "__next__"
 
 
-def _safe_infer_call_result(node, caller, context=None):
+def _safe_infer_call_result(
+    node: nodes.FunctionDef,
+    caller: nodes.FunctionDef,
+    context: InferenceContext | None = None,
+) -> Iterator[nodes.NodeNG | Uninferable | None] | None:
```
DanielNoord, line 34: See previous comments, why isn't this `InferenceResult`?

pylint/checkers/utils.py
```diff
@@ -1653,7 +1652,7 @@ def get_subscript_const_value(node: nodes.Subscript) -> nodes.Const:
     return inferred
 
 
-def get_import_name(importnode: ImportNode, modname: str | None) -> str | None:
+def get_import_name(importnode: nodes.ImportNode, modname: str | None) -> str | None:
```
DanielNoord, line 1655: This shouldn't work. `ImportNode` is not part of the `nodes` module.

pylint/checkers/classes/class_checker.py
```diff
@@ -1443,7 +1458,9 @@ def _check_redefined_slots(
                 node=slots_node,
             )
 
-    def _check_slots_elt(self, elt, node):
+    def _check_slots_elt(
+        self, elt: nodes.Attribute | nodes.Const, node: nodes.ClassDef
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Most of the values are based on what were the actual type of the node when launching the tests. There's often a lof of condition applied on things after the inference is done so it narrows the type of sub-functions quite a lot sometime.

pylint/checkers/classes/class_checker.py
```diff
@@ -1775,23 +1792,20 @@ def _is_type_self_call(self, expr: nodes.NodeNG) -> bool:
...
+    def _is_classmethod(func: LocalsDictNodeNG) -> bool:
         """Check if the given *func* node is a class method."""
-
         return isinstance(func, nodes.FunctionDef) and (
             func.type == "classmethod" or func.name == "__class_getitem__"
         )
 
     @staticmethod
-    def _is_inferred_instance(expr, klass):
+    def _is_inferred_instance(expr: nodes.NodeNG | None, klass: nodes.ClassDef) -> bool:
```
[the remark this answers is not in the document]
Pierre-Sassoulas: This would work but maybe we could add best effort values in pylint and see where mypy is unhappy once astroid is actually typed ? (The overload for ``infer`` depending on what node we're inferring and based on condition on the node can become quite complicated, and won't be done in a month).

pylint/checkers/classes/special_methods_checker.py
Pierre-Sassoulas: For:
```
    def __len__(self):  # [invalid-length-returned]
        return lambda: 3
```
It's not a const but a ``<class 'astroid.nodes.scoped_nodes.scoped_nodes.Lambda>'. pyannotate must have a bug.

pylint/checkers/utils.py
Pierre-Sassoulas: Hmm; it's true I don't see it in astroid.nodes's API. But the tests are green.

pylint/checkers/classes/special_methods_checker.py
Pierre-Sassoulas: I did not take most comment into account yet just the one regarding f-string
@Pierre-Sassoulas
[Use an fstring so tuples with more than 2 element do not crash](https://github.com/PyCQA/pylint/pull/7190/commits/1a066943f544e002287fc85c1ab0549350e84fe6)
[1a06694](https://github.com/PyCQA/pylint/pull/7190/commits/1a066943f544e002287fc85c1ab0549350e84fe6)
@Pierre-Sassoulas
[Use standard import for ImportNode](https://github.com/PyCQA/pylint/pull/7190/commits/e69f75e09e2bfafea3fb602d92d63e7c9e76e045)
[e69f75e](https://github.com/PyCQA/pylint/pull/7190/commits/e69f75e09e2bfafea3fb602d92d63e7c9e76e045)

pylint/checkers/classes/class_checker.py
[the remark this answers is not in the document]
DanielNoord: Yeah, but just based on the current typing and code this typing is incorrect. We should mostly rely on the code rather than on tests as the tests will never be cover every edge case.

pylint/checkers/classes/class_checker.py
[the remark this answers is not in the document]
DanielNoord: The issue is that we won't know which signatures to revisit when we finally do update `astroid`. For any typing for which we are not sure that `astroid` is correct I think it is often better to leave them be and try and fix `astroid` first.

pylint/checkers/classes/special_methods_checker.py
DanielNoord: I think this is not necessarily a bug with `pyannotate` but just shows that our test case doesn't cover all edge cases. Like I said, perhaps revert the changes here and keep this PR focused on those that are definetly correct?

pylint/checkers/utils.py
DanielNoord: That is because `mypy` doesn't actually check `astroid` since it has no `py.typed` file.

pylint/checkers/classes/special_methods_checker.py
Pierre-Sassoulas: > our test case doesn't cover all edge cases.

Although true in the general case, I added an assert in ``def _check_len``, the example come from ``invalid_length_returned.py`` that is supposed to be launched during functional tests so either pyannotate has a bug or the way we launch tests is not understood.

pylint/checkers/utils.py
Pierre-Sassoulas: Ha, I thought the import would have failed but it does not because it's using a namespace and not a direct import (``nodes.ImportNode``, with ``from astroid import nodes`` not ``from astroid.nodes import ImportNode``.

DanielNoord:
@Pierre-Sassoulas Would you be okay with me finishing this PR with some final changes?

Pierre-Sassoulas:
Sure, I would appreciate that !

DanielNoord (approved):

DanielNoord:
@Pierre-Sassoulas PR is now ready for review!

Pierre-Sassoulas (commented):
Lgtm, thank you @DanielNoord !

PR #7196, opened 2022-07-17
Release branch 2.15.0 alpha 0

Pierre-Sassoulas:


Pierre-Sassoulas:
Blocked by 2.14.5 release and by a new procedure for release following https://github.com/PyCQA/pylint/pull/6974

Pierre-Sassoulas:
I think we beed to remove the fragment from https://github.com/PyCQA/pylint/pull/7197/files#diff-b202d4527912ada36756d49a70ddddbb6708e268756660dfce4e0ace6194d73dR7. I suppose the news fragments will disappear when we use towncrier on the release branch the next time @DudeNr33 ?

DudeNr33:
Do you mean the three PRs that were backported?

That depends how we want to document backports.
Until now, they were only included in the changelog for the patch version.
Currently with the cherry-pick approach, they would be included in both the patch changelog and the minor changelog, unless you would actually remove the fragments by hand.
However, having them included in both would be the cleaner approach imho.

Even if up to now we never(?) had the situation where we release a patch release for an older version after releasing a new minor (for example: first release 2.16.0, and then a few days later a 2.15.1), it would be more accurate to include this change in the changelog for both versions.

What do you think?

Pierre-Sassoulas:
> However, having them included in both would be the cleaner approach imho.

Fair enough, and it's also less work.

Pierre-Sassoulas:
New release procedure in #7198

PR #7212, opened 2022-07-20
[missing-raises-doc] Take 'accept-no-raise-doc' option into account

### LINKED ISSUE #7208: accept-no-raise-doc is not checked for docstrings with at least one matching section
### Bug description

The docstyle extension sometimes ignores the accept-no-raise-doc option because the `_add_raise_message`method is invoked directly instead of `_handle_no_raise_doc` in `visit_raise`.

~~~Python
"""Minimal example where a W9006 message is displayed even if the
accept-no-raise-doc option is set to True.

Requires at least one matching section (`Docstring.matching_sections`)."""

def w9006issue(dummy: int):
    """Sample function.

    :param dummy: Unused
    """
    raise AssertionError()
~~~

### Configuration

```ini
[MAIN]

[pylint.extensions.docstyle]

accept-no-raise-doc=yes
```

### Command used

```shell
pylint w9006_issue.py
```

### Pylint output

```shell
************* Module w9006_issue
w9006_issue.py:6:0: W9006: "AssertionError" not documented as being raised (missing-raises-doc)

```

### Expected behavior

The expected behavior is to have pylint display no message.

### Pylint version

```shell
pylint 2.14.5
astroid 2.11.7
Python 3.8.5 (default, Sep  7 2020, 16:44:59)
```

### OS / Environment

Debian 9.13 (stretch)

### Additional dependencies

_No response_

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #7208

It's better for performance to simply cut short at the beginning of the function imo.

Co-authored-by: Brice Chardin <brice.chardin@gmail.com>

### DIFF
--- a/tests/functional/ext/docparams/raise/missing_raises_doc.py
+++ b/tests/functional/ext/docparams/raise/missing_raises_doc.py
@@ -11,7 +11,7 @@ def test_ignores_no_docstring(self):
     raise RuntimeError("hi")
 
 
-def test_ignores_unknown_style(self):
+def test_ignores_unknown_style(self):  # [missing-raises-doc]
     """This is a docstring."""
     raise RuntimeError("hi")
 
--- a/doc/whatsnew/fragments/7208.bugfix
+++ b/doc/whatsnew/fragments/7208.bugfix
@@ -1,0 +1,6 @@
+The ``accept-no-raise-doc`` option related to ``missing-raises-doc`` will now be correctly
+taken into account all the time. As the default value was false and was not taken into
+account because of a bug, it means you will have to add ``accept-no-raise-doc=no`` in
+your configuration to keep the same behavior.
+
+Closes #7208
--- a/pylint/extensions/docparams.py
+++ b/pylint/extensions/docparams.py
@@ -274,6 +274,8 @@ class DocstringParameterChecker(BaseChecker):
             self.add_message("redundant-yields-doc", node=node)
 
     def visit_raise(self, node: nodes.Raise) -> None:
+        if self.linter.config.accept_no_raise_doc:
+            return
         func_node = node.frame(future=True)
         if not isinstance(func_node, astroid.FunctionDef):
             return
@@ -296,7 +298,7 @@ class DocstringParameterChecker(BaseChecker):
         if not doc.matching_sections():
             if doc.doc:
                 missing = {exc.name for exc in expected_excs}
-                self._handle_no_raise_doc(missing, func_node)
+                self._add_raise_message(missing, func_node)
             return
 
         found_excs_full_names = doc.exceptions()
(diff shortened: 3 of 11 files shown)


tests/functional/ext/docparams/raise/missing_raises_doc.py
```diff
@@ -11,7 +11,7 @@ def test_ignores_no_docstring(self):
     raise RuntimeError("hi")
 
 
-def test_ignores_unknown_style(self):
+def test_ignores_unknown_style(self):  # [missing-raises-doc]
```
Pierre-Sassoulas, line 14: This is the "breaking change". When we don't recognize the docstring style and accept-no-raises-doc is False, we now raise a missing-raises-doc.

Pierre-Sassoulas:
Similarly than with the json that did not have the score, fixing the bug is creating a breaking change. I added the need decision label to decide if we put it in a minor (change the default value ?) or wait for 3.0 and keep the default value.

tests/functional/ext/docparams/raise/missing_raises_doc.py
DanielNoord: I'm not sure we should introduce a breaking change. Can we fix this without doing so?

bchardin: I believe the issue is that my suggestion was not correct, and always suppressed missing-raises-doc when accept-no-raises-doc was set.

A better way to handle this option would be to suppress missing-raises-doc when there are no matching sections (as was the case previously), or (and this would be new) when no exceptions are referenced in the docstring.

Below is a possible way to do both. It allows reverting all tests to their previous behavior with the default accept-no-raises-doc value.

~~~diff
diff --git a/pylint/extensions/docparams.py b/pylint/extensions/docparams.py
index 936cda683..c789a8472 100644
--- a/pylint/extensions/docparams.py
+++ b/pylint/extensions/docparams.py
@@ -274,8 +274,6 @@ class DocstringParameterChecker(BaseChecker):
             self.add_message("redundant-yields-doc", node=node)
 
     def visit_raise(self, node: nodes.Raise) -> None:
-        if self.linter.config.accept_no_raise_doc:
-            return
         func_node = node.frame(future=True)
         if not isinstance(func_node, astroid.FunctionDef):
             return
@@ -295,6 +293,10 @@ class DocstringParameterChecker(BaseChecker):
         doc = utils.docstringify(
             func_node.doc_node, self.linter.config.default_docstring_type
         )
+
+        if self.linter.config.accept_no_raise_doc and not doc.exceptions():
+            return
+
         if not doc.matching_sections():
             if doc.doc:
                 missing = {exc.name for exc in expected_excs}
~~~

Pierre-Sassoulas: @bchardin I tried your change but there's not difference in the output afaik ? I(m going to investigate the issue.

DanielNoord:
Sorry, it has been a long time. What was the conclusion here? This can't be fixed without a breaking change?

Pierre-Sassoulas:
I still need to work on it to find a solution without breaking changes. I feel like this could be seen as a consistency fix too but I did not came to a conclusion.

gesslerpd:
I have `accept_no_raise_doc = true` and am still seeing `missing-raises-doc` warnings. Made a similar fix to this but then found this existing PR. If any help is needed on this, let me know!

Pierre-Sassoulas:
Do you have an opinion on the review discussion regarding the breaking change or not @gesslerpd ? I've been quite busy but feel free to propose something to accelerate things :)

Pierre-Sassoulas:
Closing in favor of #7581

PR #7217, opened 2022-07-21
Ignore path doc: double escape to add escape character in rst

tbennett0:
Adding another escape to small doc PR submitted in #7103, looks like the autoupgrade removed the escape.

CC @Pierre-Sassoulas

### DIFF
--- a/pylint/lint/base_options.py
+++ b/pylint/lint/base_options.py
@@ -69,7 +69,7 @@ def _make_linter_options(linter: PyLinter) -> Options:
                 "default": [],
                 "help": "Add files or directories matching the regular expressions patterns to the "
                 "ignore-list. The regex matches against paths and can be in "
-                "Posix or Windows format. Because '\\' represents the directory delimiter "
+                "Posix or Windows format. Because '\\\\' represents the directory delimiter "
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (commented):
Thank you ! It took my waaaay longer than I'd like to admit to find the result in read the doc.

This fix the issue in the online documentation but it's very likely  that the test is also used on the help text of the command line tool. Could we add a test to check that the command line tool's help still make sense ?

tbennett0:
Sure, I'll look into it 👍

DanielNoord (approved):
Let's just merge, seems like a good improvement. If somebody wants to fix the CLI they can provide a fix afterwards, online documentation is probably used way more often.
Even I forget we have a CLI tool for help messages sometimes.. 😅

Pierre-Sassoulas (approved):
Fair enough.

tbennett0:
Ah. I apologize, I didn't get the chance to take this up. I'll take a deeper look soon.

Pierre-Sassoulas:
Don't worry @tbennett0, Daniel is right I was nitpicking. Don't worry there's a lot of pylint issues to handle if you feel like it 😉

PR #7220, opened 2022-07-22
Do not lint ignored file on stdin

### LINKED ISSUE #4354: Files not ignored when linting from standard input
The `ignore-patterns` option can be used to ignore certain files based on the specified pattern. For example:

```
echo "def somefunc():\n    pass" > file.py; pylint --ignore-patterns 'file\.py' file.py
```

Here there would be no output because the input is ignored. This does not work when reading the input from standard in instead of from a file. This is problematic when using pylint as an on-the-fly linter (e.g. in vim) because there is no good way to ignore certain inputs.

### Steps to reproduce

```
echo "def somefunc():\n    pass" | pylint --ignore-patterns 'file\.py' --from-stdin file.py
```

### Current behavior

Input is linted.

### Expected behavior

Input is ignored.

### pylint --version output

```
pylint 2.7.2
astroid 2.5.1
Python 3.9.3 (default, Apr  8 2021, 23:35:02)
[GCC 10.2.0]
```

cblessing24:
## Type of Changes

## Description

This PR adds a check to the see whether the filename passed with the `--from-stdin` option is ignored. The input will not be linted if it is. A corresponding test is also included.

Closes #4354


pylint/lint/pylinter.py
```diff
@@ -641,6 +641,13 @@ def check(self, files_or_modules: Sequence[str] | str) -> None:
 
             filepath = files_or_modules[0]
             with fix_import_path(files_or_modules):
+                if _is_ignored_file(
```
Pierre-Sassoulas, line 644: Shouldn't we also remove the previous place where we where ignoring some file ? If this is the right place then we don't need to check in the other places it was checked ?

Pierre-Sassoulas (commented):
Thank you, the test is simple and efficient !

I think we also need to cleanup the "old faulty way of doing it" if filtering just before check file make more sense. In particular for multiprocessing we probably want  to do the filtering before forking (?)

Pierre-Sassoulas:
Do not mind the pypy failure, main is broken.

pylint/lint/pylinter.py
cblessing24: I am not completely sure I understand. The check was not present previously in this if statement. Do you mean that we should put the check before the second if statement and remove it from all the subsequent if statements?

Pierre-Sassoulas: We're already filtering files somewhere when they are not from stdin, I think there's a kind of refactor to do there so we filter files at the right place and so it's efficient (done only once per run, possibly when using multiprocessing). I would start by checking where ``_is_ignored_file`` is used.

cblessing24: Okay, I looked into it a bit more and it seems we are filtering files while doing the following tasks:

cblessing24: That function is used in two places:
* [When disovering files recursively](https://github.com/PyCQA/pylint/blob/8ccb52650f78f0650081753b218375bfcf47ab64/pylint/lint/pylinter.py#L600)
* [When expanding modules](https://github.com/PyCQA/pylint/blob/8ccb52650f78f0650081753b218375bfcf47ab64/pylint/lint/expand_modules.py#L79)

It seems like the `expand_modules` function is the place where the filtering of ignored files is meant to take place and the use of `_is_ignored_file` while discovering is more of a "hack". Unfortunately `expand_modules` does not get called when linting from standard input. Furthermore `expand_modules` seems to do two things:
1. Expand modules
2. Filter ignored files

It might be a good idea to entangle these two concerns. In any case I don't see a way to only filter once. Before `expand_modules` you want to filter once to avoid unnecessarily processing any ignored files but you can not filter everything at this stage because you do not have filepaths for the modules yet. Therefore you want to filter again after you have those.

So I am not sure how to proceed from here.

Pierre-Sassoulas: Thank you for checking this.

> It might be a good idea to entangle these two concerns. In any case I don't see a way to only filter once. Before expand_modules you want to filter once to avoid unnecessarily processing any ignored files but you can not filter everything at this stage because you do not have filepaths for the modules yet. Therefore you want to filter again after you have those.

Hmm, it sounds like a bigger refactor than I thought. If we're going to extract the file discovering from PyLinter we could expand modules, reading from stdin and files discovery at the same time, right ? Removing concerns from the PyLinter would be a good thing especially for more efficient multiprocessing (easier to map files to threads), but I don't have a clear vision about the issue we'll face when doing that.

cblessing24: Yes, I think it should be possible to do the discovering, filtering and expanding beforehand and then passing the result to `Pylinter.check` but as you said that is going to be a larger refactor.

Pierre-Sassoulas: Would you be willing to give it a try ?

cblessing24: I would like to give it a try. That said I have no idea how long it will take because it seems like a quite big refactor and I do not have a lot of time to work on this.

Pierre-Sassoulas: Nice to hear, don't worry about deadline there is none :) ! You can open a draft pull request to discuss the refactor with maintainers, early feedback would probably help you.

DanielNoord:
@Pierre-Sassoulas @cblessing24 Would we be okay with merging this without the refactor? This is breaking the `pylint` plugin for VS Code, see https://github.com/microsoft/vscode-pylint/issues/169

I think it make sense to offer "first-class" support for the extension as a large part of our user base is likely using VS Code. We could potentially get this in `2.15.1`.

Pierre-Sassoulas:
Right, but this is technical debt and if this kind of treatment isn't DRY we're going to have a bad time in the long term. We already have a bad time now, this issue is clearly a problem of not treating all sources the same way. Often new contributors come, fix an issue they want to fix but are not expert in the codebase so they do not know that there's two places to modify, tests only the one they changed but not the other and we get inconsistencies (new issues are opened, new contributor come, etc.). I think this is a vicious circle. 

I'm not saying we should not create this technical debt but we should make sure we think about it.

DanielNoord:
@cblessing24 I'm quite familiar with this code. Would you be okay with me taking a stab at this refactor?

cblessing24:
Sure, I am totally fine with that.

Pierre-Sassoulas (approved):
👍

PR #7227, opened 2022-07-25
Don't emit ``super-init-not-called`` for abstract ``__init__`` methods

### LINKED ISSUE #3975: False positive ``__init__ method from base class is not called`` when the constructor raises ``NotImplementedError`` without using ``abc``

In the following example, Pylint flags the `Derived`'s `__init__` method because it doesn't call `Bases`'s `__init__`.

```python
class Base:
    def __init__(self, a: int, b: str):
        raise NotImplementedError()

class Derived(Base):
    def __init__(self, a: int, b: str):
        self.a = a + 1
        self.b = b[::-1]
```

This is discussed here: https://stackoverflow.com/questions/25072363/error-init-method-from-base-class-is-not-called-for-an-abstract-class

Pylint could (should?) detect this case and not report an error.

### Additional context

An easy fix is to remove the init in the base class, but it can be useful to have it there to define what signature `__init__` should have.

DanielNoord:

## Type of Changes

## Description

Closes #3975.

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -2034,7 +2034,7 @@ a metaclass class method.",
                         # Record that the class' init has been called
                         parents_with_called_inits.add(node_frame_class(method))
                     except KeyError:
-                        if klass not in to_call:
+                        if klass not in klass_node.ancestors(recurs=False):
                             self.add_message(
                                 "non-parent-init-called", node=expr, args=klass.name
                             )
@@ -2060,9 +2060,6 @@ a metaclass class method.",
                     continue
 
             if decorated_with(node, ["typing.overload"]):
-                continue
-            cls = node_frame_class(method)
-            if klass.name == "object" or (cls and cls.name == "object"):
                 continue
             self.add_message(
                 "super-init-not-called",
@@ -2196,7 +2193,10 @@ def _ancestors_to_call(
     to_call: dict[nodes.ClassDef, bases.UnboundMethod] = {}
     for base_node in klass_node.ancestors(recurs=False):
         try:
-            to_call[base_node] = next(base_node.igetattr(method))
+            init_node: bases.UnboundMethod = next(base_node.igetattr(method))
+            if init_node.is_abstract():
+                continue
+            to_call[base_node] = init_node
         except astroid.InferenceError:
             continue
     return to_call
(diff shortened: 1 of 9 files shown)


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
I have no idea why the coverage went down on the three lines that are not covered anymore. coverall issue ?

vivi90:
https://github.com/PyCQA/pylint/issues/3975#issuecomment-1194435463

DanielNoord:
I think this is because "living" objects don't have a body and thus are considered abstract.
This prevents them from raising inference errors later on.

See also change to the ctypes test

DanielNoord:
@Pierre-Sassoulas can this be merged with the decreased coverage?

Pierre-Sassoulas:
Maybe it's possible to cleanup or to add a test that cover this ? If it's too complicated I'm ok with merging as is, the primer indicate the change will remove a lot of false positives.

vivi90:
I think, it might be an very bad idea to change ``pylint`s current behaviour for this case, because of: https://github.com/PyCQA/pylint/issues/3975#issuecomment-1196806113

Pierre-Sassoulas:
Look like the fail is genuine and was hidden by the broken main before.

pylint/checkers/classes/class_checker.py
```diff
@@ -2061,9 +2061,6 @@ def _check_init(self, node: nodes.FunctionDef, klass_node: nodes.ClassDef) -> No
 
             if decorated_with(node, ["typing.overload"]):
                 continue
```
DanielNoord, line 2063: The lines below are no longer necessary now that we skip `object.__init__` because of its abstractness.

tests/functional/s/super/super_init_not_called.py
```diff
@@ -53,5 +54,45 @@ def __init__(self):  # [super-init-not-called]
...
+
+class DerivedFromAbstract(AbstractBase):
+    def __init__(self, param: int) -> None:  # [super-init-not-called]
+        print("Called")
+
+    def abstract_method(self) -> str:
+        return "Implemented"
+
+
+class DerivedFrom(UnknownParent):  # [undefined-variable]
```
DanielNoord, line 91: This is a regression test for the coverage that dropped. We try to infer bases which fails here.

PR #7236, opened 2022-07-27
[tests] Test no-member false positive on exception without str args

### LINKED ISSUE #7233: False positive `no-member` on args of an exception
### Bug description

The following sample script triggers false positive 'no-member' error, although the code runs okay. Pylint considers the first element of the .args tuple as a str instance.

```python
from dataclasses import dataclass


class MyHelloError(RuntimeError):
    @property
    def detail(self):
        return self.args[0]


@dataclass
class ErrorInfo:
    errorcode: int


try:
    raise MyHelloError(ErrorInfo(1234))
except MyHelloError as exc:
    print(exc.detail.errorcode)
```

### Configuration

_No response_

### Command used

```shell
pylint -d all -e E hello1.py
```

### Pylint output

```shell
************* Module hello1
hello1.py:18:10: E1101: Instance of 'str' has no 'errorcode' member (no-member)

Your code has been rated at 5.00/10
```

### Expected behavior

No error on correct code.

### Pylint version

```shell
pylint 2.14.5
astroid 2.11.7
```

### OS / Environment

Ubuntu 22.04 (Jammy Jellyfish)

### Additional dependencies

_No response_

Pierre-Sassoulas:
## Type of Changes

## Description

Regression test for #7233, the fix is in https://github.com/PyCQA/astroid/pull/1722, it will require an astroid update.

Closes #7233

### DIFF
--- a/doc/whatsnew/fragments/7233.false_positive
+++ b/doc/whatsnew/fragments/7233.false_positive
@@ -1,0 +1,3 @@
+Fixed a false positive 'no-member' for first exception argument that were not a string. It's usually a string but not always.
+
+Closes #7233
--- a/tests/functional/r/regression_02/regression_7233.py
+++ b/tests/functional/r/regression_02/regression_7233.py
@@ -1,0 +1,22 @@
+"""Regression test for https://github.com/PyCQA/pylint/issues/7233"""
+
+# pylint: disable=missing-docstring
+
+from dataclasses import dataclass
+
+
+class MyHelloError(RuntimeError):
+    @property
+    def detail(self):
+        return self.args[0]
+
+
+@dataclass
+class ErrorInfo:
+    errorcode: int
+
+
+try:
+    raise MyHelloError(ErrorInfo(1234))
+except MyHelloError as exc:
+    print(exc.detail.errorcode)


DanielNoord:
See https://github.com/PyCQA/astroid/pull/1722#pullrequestreview-1056292990.

DanielNoord:
Shall we close this and move the test to `astroid`?

Pierre-Sassoulas:
As #7233 is a pylint bug too, I think it makes sense to also have a test in pylint. Maybe we could also move this issue in astroid and ask for a lot more test in astroid. The problem with that is that astroid do not have efficients and easy to create functional tests like pylint.. What do you think?

DanielNoord:
> As #7233 is a pylint bug too, I think it makes sense to also have a test in pylint. Maybe we could also move this issue in astroid and ask for a lot more test in astroid. The problem with that is that astroid do not have efficients and easy to create functional tests like pylint.. What do you think?

This test in `astroid` should be fairly straightforward. Since this is really more an `astroid` issue I think it makes more sense to have the regression test there instead of in `pylint`.

Pierre-Sassoulas:
Fair enough. Should we move such issues to astroid too ?

DanielNoord:
> Fair enough. Should we move such issues to astroid too ?

Perhaps? Although I think for visibility towards users it's probably best to keep them here.

PR #7244, opened 2022-07-30
[test] Temporary xfail deprecated methods py36 on pypy

Pierre-Sassoulas:
## Type of Changes

## Description

Maybe astroid 2.12.3 will fix this because of https://github.com/PyCQA/astroid/pull/1714 but this is taking some time to be released.

### DIFF
--- a/tests/functional/d/deprecated/deprecated_methods_py36.rc
+++ b/tests/functional/d/deprecated/deprecated_methods_py36.rc
@@ -1,6 +1,3 @@
-[Imports]
-deprecated-modules=imp
-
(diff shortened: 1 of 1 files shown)


DanielNoord:
Can this be added to a TODO for `2.15`? I don't think we should release which such a clear regression.

Pierre-Sassoulas:
Sure thing. Should we wait for releasing the alpha too ?

DanielNoord:
Well, since the alpha is mainly to test the namespace change we should probably wait until we have a fix for a crash related to namespaces.

PR #7251, opened 2022-07-31
Add type annotations to `format.py` module.

mbyrnepr2:
## Type of Changes

## Description

Refs #2079

### DIFF
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -646,7 +646,7 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
     def check_lines(self, lines: str, lineno: int) -> None:
         """Check given lines for potential messages.
 
-        Check lines have:
+        Check if lines have:
         - a final newline
         - no trailing white-space
         - less than a maximum number of characters
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -255,7 +255,7 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
     def __init__(self, linter: PyLinter) -> None:
         super().__init__(linter)
         self._lines: dict[int, str] = {}
-        self._visited_lines: dict[str | None, int | None] = {}
+        self._visited_lines: dict[int, int] = {}
         self._bracket_stack: list[str | None] = [None]
 
     def new_line(self, tokens: TokenWrapper, line_end: int, line_start: int) -> None:
@@ -379,7 +379,7 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
                     return
 
     def process_tokens(self, tokens: list[tokenize.TokenInfo]) -> None:
-        """Process tokens and search for :
+        """Process tokens and search for:
 
         - too long lines (i.e. longer than <max_chars>)
         - optionally bad construct (if given, bad_construct must be a compiled
@@ -393,7 +393,7 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
         self._visited_lines = {}
         self._last_line_ending: str | None = None
         last_blank_line_num = 0
-        for idx, (tok_type, token, start, _, line) in enumerate(tokens):
+        for idx, (tok_type, string, start, _, line) in enumerate(tokens):
             if start[0] != line_num:
                 line_num = start[0]
                 # A tokenizer oddity: if an indented line contains a multi-line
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
LGTM, thank you 👍

pylint/checkers/format.py
```diff
@@ -251,13 +252,13 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
...
 
-    def __init__(self, linter=None):
+    def __init__(self, linter: PyLinter):
         super().__init__(linter)
-        self._lines = None
-        self._visited_lines = None
-        self._bracket_stack = [None]
+        self._lines: dict[int, str] = {}
+        self._visited_lines: dict[str | None, int | None] = {}
+        self._bracket_stack: list[str | None] = [None]
```
DanielNoord, line 259: Does this do anything? It seems we only assign `[None]` to it, or am I missing something obvious?

pylint/checkers/format.py
```diff
@@ -251,13 +252,13 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
         ),
     )
 
-    def __init__(self, linter=None):
+    def __init__(self, linter: PyLinter):
```
DanielNoord, line 255:
```suggestion
    def __init__(self, linter: PyLinter) -> None:
```

pylint/checkers/format.py
```diff
@@ -251,13 +252,13 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
...
         super().__init__(linter)
-        self._lines = None
-        self._visited_lines = None
-        self._bracket_stack = [None]
+        self._lines: dict[int, str] = {}
+        self._visited_lines: dict[str | None, int | None] = {}
```
DanielNoord, line 258: This seems incorrect? We use `.get()` with `ints` on this dictionary. Shouldn't it be `dict[int, int]`?

pylint/checkers/format.py
```diff
@@ -466,7 +467,7 @@ def process_tokens(self, tokens: list[tokenize.TokenInfo]) -> None:
         if line_num == last_blank_line_num and line_num > 0:
             self.add_message("trailing-newlines", line=line_num)
 
-    def _check_line_ending(self, line_ending, line_num):
+    def _check_line_ending(self, line_ending: str, line_num: int) -> None:
```
DanielNoord, line 470: While we're at it, would you mind renaming `token` on line L396 and L414 to `string`? It what's being used to call this function and originally threw me off. The second argument of `TokenInfo` is called a `string` and not a `token`.

pylint/checkers/format.py
```diff
@@ -524,15 +525,15 @@ def visit_default(self, node: nodes.NodeNG) -> None:
...
+        lines: list[str] = []
         for line in range(line, tolineno + 1):
             self._visited_lines[line] = 1
             try:
                 lines.append(self._lines[line].rstrip())
             except KeyError:
                 lines.append("")
 
-    def _check_multi_statement_line(self, node, line):
+    def _check_multi_statement_line(self, node: nodes.NG, line: str) -> None:
```
DanielNoord, line 536:
```suggestion
    def _check_multi_statement_line(self, node: nodes.NodeNG, line: str) -> None:
```

?

pylint/checkers/format.py
```diff
@@ -645,7 +646,7 @@ def specific_splitlines(lines: str) -> list[str]:
     def check_lines(self, lines: str, lineno: int) -> None:
         """Check given lines for potential messages.
 
-        Check lines have :
+        Check lines have:
```
DanielNoord, line 649:
```suggestion
        Check if lines have:
```

pylint/checkers/format.py
```diff
@@ -127,22 +128,22 @@ def _last_token_on_line_is(tokens, line_end, token):
 class TokenWrapper:
     """A wrapper for readable access to token information."""
 
-    def __init__(self, tokens):
+    def __init__(self, tokens: list[tokenize.TokenInfo]):
```
DanielNoord, line 131:
```suggestion
    def __init__(self, tokens: list[tokenize.TokenInfo]) -> None:
```

DanielNoord (changes requested):

pylint/checkers/format.py
Pierre-Sassoulas: Nice catch ! I always forget that as long as astroid do not expose type we can put anything here and mypy is happy with really wrong typing...

pylint/checkers/format.py
mbyrnepr2: Good observation @DanielNoord. If you do `git show 480e0ad898` and view the module, there are some push operations to add values to `self._bracket_stack`, but as you say, we only assign `[None]` now. I'll remove it.

DanielNoord (approved):

mbyrnepr2:
Sorry for late changes, just noticed one more improvement on the typing: using `Literal[1, 2]` for possible dict values.

PR #7257, opened 2022-08-02
 Don't treat `__slots__` as enum member

### LINKED ISSUE #6887: False positives `invalid-enum-extension`
### Bug description

That check doesn't seem to check if the parent class actually defines enumeration members. This happens only if the parent class is defined in a differen module. MWE:

`bar.py`

```
import Enum

class StringEnum(str, Enum):
    def __repr__(self) -> str:
        return f"<{self.__class__.__name__}.{self.name}>"
```

`foo.py`

```
from bar import StringEnum


class ChatType(StringEnum):

    SENDER = "sender"
    PRIVATE = "private"
```

### Configuration

_No response_

### Command used

```shell
`pylint foo.py`
```

### Pylint output

```shell
************* Module foo
foo.py:1:0: C0114: Missing module docstring (missing-module-docstring)
foo.py:1:0: C0104: Disallowed name "foo" (disallowed-name)
foo.py:4:0: E0244: Extending inherited Enum class "StringEnum" (invalid-enum-extension)

```

### Expected behavior

pylint either checks the parent class for enum members or does enable this warning by default

### Pylint version

```shell
pylint 2.14.1
astroid 2.11.5
Python 3.8.5
```

### OS / Environment

Windows

### Additional dependencies

_No response_

Bibo-Joshi:
## Type of Changes

## Description

Closes #6887


mbyrnepr2:
I think the issue applies to any dunder name; not just `__slots__`. For example:

```python
class E(Enum):
    __x__ = 'x'

class F(E):
    member = 3

pylint x.py:
x.py:7:0: E0244: Extending inherited Enum class "E" (invalid-enum-extension)
```

Bibo-Joshi:
`__x__` could be an actual enum member though, right? I guess the question is if non-standard dunder attributes should be treated as enum members or not. Personally I'd say yes.

Another thought that I head is that it might be better to have `astroid` handle this better directly such that we don't have to apply special casing here …

Pierre-Sassoulas:
+1 @Bibo-Joshi I did not follow everything here but if a fix can be done in astroid it's better to do it in astroid 👍

Bibo-Joshi:
All right. I'll try to have a look into it or at least open an issue over at astroid :)

mbyrnepr2:
> `__x__` could be an actual enum member though, right? I guess the question is if non-standard dunder attributes should be treated as enum members or not. Personally I'd say yes.
> 
> Another thought that I head is that it might be better to have `astroid` handle this better directly such that we don't have to apply special casing here …

It doesn't look like it. The dunder name does not appear in the __members__ container:
```python
(venv310) Marks-MacBook-Air-2:programming markbyrne$ cat y.py 
from enum import Enum


class E(Enum):
    __x__ = 'x'
    AAA = "aaa"

print(E.__members__)
(venv310) Marks-MacBook-Air-2:programming markbyrne$ python y.py 
{'AAA': <E.AAA: 'aaa'>}
```

So I think the issue is Astroid is inserting them into __members__ when it shouldn't be.

mbyrnepr2:
Just to add - `__slots__` is also not something which is used by an Enum. There was an issue about adding it here but I don't think that was added: https://bugs.python.org/issue46132

So, with that in mind, perhaps there is scope in this MR to add a new warning if __slots__ are used in an Enum.
Otherwise we can simply focus on the astroid change.

Bibo-Joshi:
> All right. I'll try to have a look into it or at least open an issue over at astroid :)

Here it is: https://github.com/PyCQA/astroid/issues/1730. In any case, I guess this PR is superfluous. Closing.

Pierre-Sassoulas:
Thank you @Bibo-Joshi

PR #7259, opened 2022-08-03
[open source insight] Set the github action dependencies by sha

Pierre-Sassoulas:
## Type of Changes

## Description

I think the version number is clearer but this is done in order to
have a better score in https://deps.dev/pypi/pylint. The reasoning
is that it mitigate the risk of compromised actions.

It seems it's not going to change anything for the dependency update by dependabot:
https://github.com/systemd/systemd/pull/22638

Full disclosure, I'm being paid to make pylint more secure by Tidelift
and this is part of this effort.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -17,14 +17,14 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Check out code from GitHub
-        uses: actions/checkout@v3.0.2
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
         with:
           # `towncrier check` runs `git diff --name-only origin/main...`, which
           # needs a non-shallow clone.
           fetch-depth: 0
       - name: Set up Python ${{ env.DEFAULT_PYTHON }}
         id: python
-        uses: actions/setup-python@v4.1.0
+        uses: actions/setup-python@c4e89fac7e8767b327bbad6cb4d859eda999cf08
         with:
           python-version: ${{ env.DEFAULT_PYTHON }}
       - name: Generate partial Python venv restore key
@@ -35,7 +35,7 @@ jobs:
           }}"
       - name: Restore Python virtual environment
         id: cache-venv
-        uses: actions/cache@v3.0.5
+        uses: actions/cache@0865c47f36e68161719c5b124609996bb5c40129
         with:
           path: venv
           key: >-
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -26,10 +26,10 @@ jobs:
       pre-commit-key: ${{ steps.generate-pre-commit-key.outputs.key }}
     steps:
       - name: Check out code from GitHub
-        uses: actions/checkout@v3.0.2
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
       - name: Set up Python ${{ env.DEFAULT_PYTHON }}
         id: python
-        uses: actions/setup-python@v4.1.0
+        uses: actions/setup-python@c4e89fac7e8767b327bbad6cb4d859eda999cf08
         with:
           python-version: ${{ env.DEFAULT_PYTHON }}
       - name: Generate partial Python venv restore key
@@ -40,7 +40,7 @@ jobs:
           }}"
       - name: Restore Python virtual environment
         id: cache-venv
-        uses: actions/cache@v3.0.5
+        uses: actions/cache@0865c47f36e68161719c5b124609996bb5c40129
         with:
           path: venv
           key: >-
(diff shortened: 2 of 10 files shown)


DanielNoord:
Is there any way to still record the version number somewhere?
This is quite unuserfriendly 😅

Pierre-Sassoulas:
> This is quite unuserfriendly

Yeah I agree. Maybe we don't need to know ? I don't have a solution. The best I can think of right now is creating a script to add comment automatically but... yeah it's not going to be cheap 😄

Pierre-Sassoulas:
We would know when we merge the dependabot update because the name of the PR is human readable.

DanielNoord:
Yeah it's probably a bit of effort.
IMO, this whole change is not worth it. It satisfies a somewhat arbitrary quality metric. I understand the risk, but what could be gained from attacking our CI? A read/write access to our repository? If we're concerned about that we should probably disallow any CI to be run without approval 😅
There isn't any "company critical" information in our CI, we don't use any tokens as far as I know? If so, we should probably focus on finding ways to do without those instead of this change.

_Note that the changes here only affect our own CI. Not the runtime dependencies of pylint._

Pierre-Sassoulas:
I guess it's in order for the CI to be resilient if a GitHub actions is compromised. It's higly unlikely for the one we're using right now but a retagging could happen theorically and we would execute a tag we would not expect to execute then. 

There's one crucially important token that permits to release pylint in the release job. If it's accessed it could be really problematic especially considering I can't yank releases right now. 

So I think it's reasonable to do in general even if our GitHub actions are fairly safe right now. 

(As I said in the description the original reason is that I'm paid to increase this score. I can give feed-back about the score and not do it if it's justified though)

DanielNoord:
I'm on mobile and the site doesn't work very well, so I can't check the site very easily but I think _REASONING
dependency not pinned by hash detected -- score normalized to 2_ is misleading.
Those are not dependencies, those are CI tools. Our dependencies are specified in `pyproject.toml`.
The risks indicated by these messages should fall under `Dangerous-Workflow` for which we score 10/10 😅🎉

Pierre-Sassoulas:
Yeah, originally I thought the bad score in dependencies was suggesting that we pin our runtime depencies, which would break pylint install every time it's installed after a common library (like when tensorflow pinned typing_extension). So I'm glad it's not that :smile:

DanielNoord:
In the full explanation of their scorecard they do explicitly refer to pinning CI dependencies. Of course I can only agree that this is a potential risk, but (as explained above) I think in our case this is minimal and I don't think the additional obfuscation is worth it.

If we really need to bump our score here I guess we should merge (although I'd like the opinion of others). 
My preferred option though would be to ask them to reconsider putting CI dependencies under the `Pinned-Dependencies` category and move them to `Dangerous-Workflow` which they already track. The current scoring unnecessarily creates distrust for users. Then I would be okay with a 9/10 for workflows and a 10/10 for dependencies 😅😬

Pierre-Sassoulas:
So I talked about this with the other tidelift maintainers and apparently there's no other concern than the possibility to untag / retag as long as we're using the patch number of a github actions. One thing that was pointed out is that using ``@v3`` make us use the latest github actions automatically but this is not happening for patch version. Julian Grubber also pointed out that a system like npm do not permit to change a release after the fact like git/github does.

I thought about it a little since, retag is not the most likely attack vector, simply releasing a new version will be run by dependabot automatically and the result of the run is public. So using sha is not better security wise... Unless we remove dependabot and actually check the diff of all the tools when upgrading... which is a LOT of work and I'm not doing that unless paid at market rate so we're probably never going to do that. 

IMO we **have to** trust the creator of the github actions enough to act in good faith when upgrading his actions. I'm adding the work in progress label pending a discussion about this with Tidelift.

cdce8p:
Just stumbled across this one.

IMO the change isn't worth the additional effort. There are better more practical security improvements we can implement. Haven't had the time to do so yet, though.
* Always pin specific patch version tags for github actions, i.e. don't use `@v3`. That could also improve CI stability.
* Only allow actions from a known source, e.g. only `actions/*` and `github/*`. https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#managing-github-actions-permissions-for-your-repository
* https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#preventing-github-actions-from-creating-or-approving-pull-requests
* Add read-only token permissions for jobs. https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
```yaml
permissions:
    contents: read
```
* Create a production environment for the release workflow and require at least one or two reviews before the workflow pushes a new version to PyPI. https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment

cclauss:
Not a fan.  This is NOT safer.

Pierre-Sassoulas:
I'm going to close, the pushback is very strong from everyone in the lifter program, this is probably not going to be a requirement in the future.

PR #7279, opened 2022-08-09
Clean up beyond_top_level tests

### LINKED ISSUE #7252: `epylint` runner confounding results of `test_relative_beyond...` test methods
The tests in `TestImportChecker` added in #5059 use the `epylint` pylint runner, not the regular one. `epylint` patches `cwd`, so it's not an optimal runner for these tests that are import- and namespace-sensitive.

Indeed, I'm proposing to `xfail` one test in #7114, but the tests should probably just be removed.

<details>
<summary>Diff indicating that they don't pass using the regular runner:</summary>

```diff
diff --git a/tests/checkers/unittest_imports.py b/tests/checkers/unittest_imports.py
index ac36f784f..11f6f19ab 100644
--- a/tests/checkers/unittest_imports.py
+++ b/tests/checkers/unittest_imports.py
@@ -8,10 +8,10 @@ import os
 
 import astroid
 
-from pylint import epylint as lint
 from pylint.checkers import imports
 from pylint.interfaces import UNDEFINED
-from pylint.testutils import CheckerTestCase, MessageTest
+from pylint.lint.run import Run
+from pylint.testutils import CheckerTestCase, MessageTest, GenericTestReporter
 
 REGR_DATA = os.path.join(os.path.dirname(__file__), "..", "regrtest_data", "")
 
@@ -41,38 +41,36 @@ class TestImportsChecker(CheckerTestCase):
 
     @staticmethod
     def test_relative_beyond_top_level_two() -> None:
-        output, errors = lint.py_run(
-            f"{os.path.join(REGR_DATA, 'beyond_top_two')} -d all -e relative-beyond-top-level",
-            return_std=True,
+        run1 = Run(
+            [os.path.join(REGR_DATA, 'beyond_top_two'), "--disable=all -e relative-beyond-top-level"],
+            exit=False,
         )
         top_level_function = os.path.join(
             REGR_DATA, "beyond_top_two/namespace_package/top_level_function.py"
         )
-        output2, errors2 = lint.py_run(
-            f"{top_level_function} -d all -e relative-beyond-top-level",
-            return_std=True,
+        run2 = Run(
+            [top_level_function, "--disable=all -e relative-beyond-top-level"],
+            exit=False,
         )
 
-        assert len(output.readlines()) == len(output2.readlines())
-        assert errors.readlines() == errors2.readlines()
+        assert len(run1.linter.reporter.out.readlines()) == len(run2.linter.reporter.out.readlines())
 
     @staticmethod
     def test_relative_beyond_top_level_three() -> None:
-        output, errors = lint.py_run(
-            f"{os.path.join(REGR_DATA, 'beyond_top_three/a.py')} -d all -e relative-beyond-top-level",
-            return_std=True,
+        run = Run(
+            [os.path.join(REGR_DATA, 'beyond_top_three/a.py'), "--disable=all -e relative-beyond-top-level"],
+            exit=False,
+            reporter=GenericTestReporter()
         )
-        assert len(output.readlines()) == 5
-        assert errors.readlines() == []
+        assert len(run.linter.reporter.out.readlines()) == 5
 
     @staticmethod
     def test_relative_beyond_top_level_four() -> None:
-        output, errors = lint.py_run(
-            f"{os.path.join(REGR_DATA, 'beyond_top_four/module')} -d missing-docstring,unused-import",
-            return_std=True,
+        run = Run(
+            [os.path.join(REGR_DATA, 'beyond_top_four/module'), "--disable=missing-docstring,unused-import"],
+            exit=False,
         )
-        assert len(output.readlines()) == 5
-        assert errors.readlines() == []
+        assert len(run.linter.reporter.out.readlines()) == 5
 
     def test_wildcard_import_init(self) -> None:
         module = astroid.MANAGER.ast_from_module_name("init_wildcard", REGR_DATA)
```

</details>

```
==================================================== short test summary info =====================================================
FAILED tests/checkers/unittest_imports.py::TestImportsChecker::test_relative_beyond_top_level_three - AssertionError: assert 0 ...
FAILED tests/checkers/unittest_imports.py::TestImportsChecker::test_relative_beyond_top_level_four - assert 0 == 5
```
The first test also fails in reality but unexpectedly passes because the comparison is to another incorrect result:
```shell
$ pylint tests/regrtest_data/beyond_top_two
************* Module beyond_top_two
tests/regrtest_data/beyond_top_two/__init__.py:1:0: F0010: error while code parsing: Unable to load file tests/regrtest_data/beyond_top_two/__init__.py:
[Errno 2] No such file or directory: 'tests/regrtest_data/beyond_top_two/__init__.py' (parse-error)
```

***
cc/ @DanielNoord

DanielNoord:

## Type of Changes

## Description

Closes #7252.

### DIFF
--- a/tests/checkers/unittest_imports.py
+++ b/tests/checkers/unittest_imports.py
@@ -7,11 +7,11 @@
 import os
 
 import astroid
-import pytest
+from pytest import CaptureFixture
 
-from pylint import epylint as lint
 from pylint.checkers import imports
 from pylint.interfaces import UNDEFINED
+from pylint.lint import Run
 from pylint.testutils import CheckerTestCase, MessageTest
 
 REGR_DATA = os.path.join(os.path.dirname(__file__), "..", "regrtest_data", "")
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Nice clean up, using pytest fixtures really  help with tests complexity.

tests/checkers/unittest_imports.py
```diff
@@ -41,42 +41,57 @@ def test_relative_beyond_top_level(self) -> None:
...
+        Run(
+            [top_level_function, "-d all", "-e relative-beyond-top-level"],
+            exit=False,
         )
+        output2, errors2 = capsys.readouterr()
 
-        assert len(output.readlines()) == len(output2.readlines())
-        assert errors.readlines() == errors2.readlines()
+        # The first package fails to lint
+        assert len(output.split("\n")) == 1
```
jacobtylerwalls, line 65: Should we leave this as 5 and `xfail` it? Tests are documentation, and I don't want to document the wrong result.

jacobtylerwalls: Well, maybe the result is correct if the expectation is that you have to `cd` to a namespace package you want to lint.

DanielNoord: Yeah I was meaning to ask this. Is the new behavior how we want to do this? Or is this a temporary result because of improved but not perfect namespace package handling?

If this is desired (ie., you can't lint a package like this) I think it makes sense to just have this as a test itself.

jacobtylerwalls: Thanks for giving me time to double-check. Yes, you need to `cd` or otherwise do something fancy (with a hook?) if you want to find namespace packages. This matches the behavior of python's `import` statement, so far as I can tell. So the result is correct.

To go back to the original issue linked in these tests, we're just trying to test for the absence of `no-name-in-module` or `relative-beyond-top-level`.

jacobtylerwalls (approved):

tests/checkers/unittest_imports.py
jacobtylerwalls: I was wrong about this after all. I guess the `beyond_top_two` _is_ on `sys.path` somewhere, so it should be linted. (I guess because `tests` is on sys.path?) See #7304.

Ultimately new tests should create their own isolated packages with code, to make this easier to reason about!

Issue #7263, opened 2022-08-03
Checker plugin parallelization breaks algorithm


smac89:
### Question

I am building a checker plugin which is supposed to use data from all the files checked to compose a report. However, I am seeing that whenever I run pylint, my plugin is ran multiple times, and each time it is using incomplete data to do its tests.

My plugin is already making use of `get_map_data` and `reduce_map_data` to merge data between processes. Now I want to know how to avoid doing the actual checks (which are done in the `close` method), until the last minute. 
How do I save all data between plugin runs until the last one so that only the last thread/last run for the checker actually does the work required?
Is this even possible? If not, what do you advice?

### Documentation for future user

https://pylint.pycqa.org/en/latest/development_guide/how_tos/plugins.html

smac89:
My bad. I think I see my mistake now. The checks should be done in the `reduce_map_data`.

So that brings me to my next question:

I have now moved the algorithm which does the checks into a separate method called `_run`, and it is called by `reduce_map_data`. What I now do is that in the `close` method, I check if the linter was run with just one job (`self.linter.config.jobs == 1`), and then I will call the `_run` method inside the `close` method.

Is this the correct way of doing that?

Pierre-Sassoulas:
Thank you for opening the issue. As far as I know you're navigating in uncharted territory here. Some pylint checkers should use this but are not right now or have issues (see the multiprocessing label) ! So I welcome you're insight on a more intuitive design. It seems that the close function should handle the reduce map generically if I understand right ? The base class could be extended if required. 

(Pinging @doublethefish as they designed and implemented this).

doublethefish:
The map/reduce mixin (`pylint.checkers.mapreduce_checker.MapReduceMixin`) itself isn't designed to distribute computation (pylint already is trying to do that via `multiprocessing` in `lint.parallel.check_parallel`), it just helps collate data about those runs. Really it's a workaround to some deeper underlying issues in pylint's multi-job processes.

[...]

PR #7286, opened 2022-08-10
Refactors related to separating ast creating and linting

DanielNoord:

## Type of Changes

## Description

Refs https://github.com/PyCQA/pylint/issues/7263.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -902,7 +902,7 @@ class PyLinter(
 
     def get_ast(
         self, filepath: str, modname: str, data: str | None = None
-    ) -> nodes.Module:
+    ) -> nodes.Module | None:
         """Return an ast(roid) representation of a module or a string.
 
         :param str filepath: path to checked file.
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -17,6 +17,7 @@ from collections import defaultdict
 from collections.abc import Callable, Iterable, Iterator, Sequence
 from io import TextIOWrapper
 from pathlib import Path
+from re import Pattern
 from typing import Any
 
 import astroid
@@ -307,6 +308,7 @@ class PyLinter(
         self.current_name: str | None = None
         self.current_file: str | None = None
         self._ignore_file = False
+        self._ignore_paths: list[Pattern[str]] = []
 
         # Attributes related to stats
         self.stats = LinterStats()
@@ -576,6 +578,7 @@ class PyLinter(
 
         This method is called before any linting is done.
         """
+        self._ignore_paths = self.config.ignore_paths
         # initialize msgs_state now that all messages have been registered into
         # the store
         for msg in self.msgs_store.messages:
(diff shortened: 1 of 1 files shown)


pylint/lint/pylinter.py
```diff
@@ -623,7 +626,10 @@ def check(self, files_or_modules: Sequence[str] | str) -> None:
 
         files_or_modules is either a string or list of strings presenting modules to check.
         """
+        # 1) Initialize
         self.initialize()
+
+        # 2) Gather all files
```
Pierre-Sassoulas, line 632: We could create a `_gather_all_files` fonction and remove comments.

pylint/lint/pylinter.py
```diff
@@ -899,7 +912,7 @@ def _astroid_module_checker(
 
     def get_ast(
         self, filepath: str, modname: str, data: str | None = None
-    ) -> nodes.Module:
+    ) -> nodes.Module | None:
         """Return an ast(roid) representation of a module or a string.
 
         :param str filepath: path to checked file.
```
Pierre-Sassoulas, line 918:
```suggestion
        :param filepath: path to checked file.
```

pylint/lint/pylinter.py
```diff
@@ -639,23 +645,30 @@ def check(self, files_or_modules: Sequence[str] | str) -> None:
                     "Missing filename required for --from-stdin"
                 )
 
-            filepath = files_or_modules[0]
+        # 3) Get all FileItems
```
Pierre-Sassoulas, line 648: Same with `_get_all_fileitems`.

Pierre-Sassoulas (commented):
This would require another refactor but wouldn't it be a good thing to create à 'FileGatherer' class to have something testable and decoupled from the configuration in order to decrease the responsabilities of the PyLinter class ?

DanielNoord:
> This would require another refactor but wouldn't it be a good thing to create à 'FileGatherer' class to have something testable and decoupled from the configuration in order to decrease the responsabilities of the PyLinter class ?

Wouldn't that be a bit overkill? I think it makes more sense to focus on creating a `FileChecker` which is a step somewhere in this refactoring process 😄

Pierre-Sassoulas:
> I think it makes more sense to focus on creating a FileChecker which is a step somewhere in this refactoring process 

I'm not opinionated about creating a particular class or about a particular design, you know this part of the code better than me so you'll know better than me what to do :)

But I think we need to be able to test those changes at some point, right now we need to test through the PyLinter which means it's hard and we don't. The core internal of pylint (like pylint.message, and some part of the PyLinter) should be unitary tested (well, unitary testable is a first step) .

DanielNoord:
> I'm not opinionated about creating a particular class or about a particular design, you know this part of the code better than me so you'll know better than me what to do :)

That's not necessarily true 😅 

> But I think we need to be able to test those changes at some point, right now we need to test through the PyLinter which means it's hard and we don't. The core internal of pylint (like pylint.message, and some part of the PyLinter) should be unitary tested (well, unitary testable is a first step) .

Yeah agreed. For now I'd say let's merge this and #7288. That would definelty set us up for the creation of `FileChecker` or something like that.

Pierre-Sassoulas:
> Yeah agreed. For now I'd say let's merge this and https://github.com/PyCQA/pylint/pull/7288. That would definelty set us up for the creation of FileChecker or something like that.

👍

Pierre-Sassoulas (approved):

PR #7288, opened 2022-08-10
Refactor check()

DanielNoord:

## Type of Changes

## Description

Refs https://github.com/PyCQA/pylint/issues/7263.

Blocked by #7286.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -14,14 +14,14 @@ import tokenize
 import traceback
 import warnings
 from collections import defaultdict
-from collections.abc import Callable, Iterable, Iterator, Sequence
+from collections.abc import Callable, Iterator, Sequence
 from io import TextIOWrapper
 from pathlib import Path
 from re import Pattern
 from typing import Any
 
 import astroid
-from astroid import AstroidError, nodes
+from astroid import nodes
 
 from pylint import checkers, exceptions, interfaces, reporters
 from pylint.checkers.base_checker import BaseChecker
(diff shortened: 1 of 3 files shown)


pylint/lint/pylinter.py
```diff
@@ -691,27 +721,58 @@ def check_single_file_item(self, file: FileItem) -> None:
...
+        self._ignore_file = False
+        self.file_state = FileState(file.modpath, self.msgs_store, module)
+        # fix the current file (if the source file was not available or
+        # if it's actually a c extension)
+        self.current_file = module.file
+
+        try:
+            check_astroid_module(module)
+        except Exception as e:
+            raise astroid.AstroidError from e
```
Pierre-Sassoulas, line 768: This is not covered, we might want to add a test mocking a fail.

pylint/lint/pylinter.py
```diff
@@ -691,27 +721,58 @@ def check_single_file_item(self, file: FileItem) -> None:
...
+            try:
+                self._lint_file(fileitem, module, check_astroid_module)
+            except astroid.AstroidError as ex:
+                template_path = prepare_crash_report(
+                    ex, fileitem.filepath, self.crash_file_path
+                )
+                msg = get_fatal_error_message(fileitem.filepath, template_path)
+                self.add_message(
+                    "astroid-error", args=(fileitem.filepath, msg), confidence=HIGH
+                )
```
Pierre-Sassoulas, line 742: Not sure here, but did we remove the distinction between pylint error and astroid error ? It seems the "fatal" message can't be raised anymore.

pylint/lint/pylinter.py
Pierre-Sassoulas, line 768:
```suggestion
            if not isinstance(e, AstroidError):
                raise PylintFatalError from e
            raise
```
(Don't apply directly as the code is invalid as the class proposed does not exist and should be added in https://github.com/PyCQA/pylint/blob/main/pylint/exceptions.py)

Pierre-Sassoulas (changes requested):
Looks good, I only have time for a quick partial review right now, sorry :)

pylint/lint/pylinter.py
DanielNoord: I think this still exists? It seems to be covered?

pylint/lint/pylinter.py
DanielNoord: Why would we want this error class?

pylint/lint/pylinter.py
Pierre-Sassoulas: I made a search for `"fatal"` and found nothing. I think we'd raise an 'astroid-error' now instead of 'fatal' ? Maybe we don't need the distinction and can simplify (the suggestion of new error was to keep that behaviour)

DanielNoord: `get_fatal_error_message` is just a helper message that returns a string that points to the template. I don't think we need to change anything here.

pylint/lint/pylinter.py
DanielNoord: Added a test!

pylint/lint/pylinter.py
Pierre-Sassoulas: I'm thinking of ``self.add_message("astroid-error", ``, previously there was a ``self.add_message("fatal", `` that is not raised anymore. I.e. The new code always raise an astroid-error, even when the error come from pylint or somewhere else like an OSError "no space in device", or something unexpected.

tests/lint/test_pylinter.py
```diff
@@ -41,3 +42,19 @@ def test_check_deprecation(linter: PyLinter, recwarn):
...
+        "pylint.lint.PyLinter.check_astroid_module", side_effect=RuntimeError
+    ):
+        linter.crash_file_path = str(tmpdir / "pylint-crash-%Y")
+        linter.check([__file__])
+        out, err = capsys.readouterr()
+        assert not out
+        assert not err
+        files = tmpdir.listdir()
+        assert len(files) == 1
+        assert "pylint-crash-20" in str(files[0])
```
Pierre-Sassoulas, line 60: Will python/pylint still be relevant in 2100 ? If so we're creating work for the next generation of maintainers 😄

DanielNoord: Copied from the test above 😄

pylint/lint/pylinter.py
DanielNoord: Oh, my bad. I didn't read this correctly. You're right, this is a mistake. Fix incoming.

tests/lint/test_pylinter.py
```diff
@@ -33,9 +32,7 @@ def test_crash_in_file(
     files = tmpdir.listdir()
     assert len(files) == 1
     assert "pylint-crash-20" in str(files[0])
-    with open(files[0], encoding="utf8") as f:
-        content = f.read()
-    assert "Failed to import module spam." in content
+    assert any(m.symbol == "fatal" for m in linter.reporter.messages)
```
DanielNoord, line 35: Previous test makes no sense as this test tries to lint the current file. Since the crash report includes the current file in the report whatever you put in an assertion will always end up in the crash report as well.

Since the test doesn't matter that much anyway I thought replacing was best.

DanielNoord:
@Pierre-Sassoulas Gentle ping as this is very prone to creating merge conflicts and I'd like to merge quickyl

Pierre-Sassoulas (approved):
LGTM ! Sorry for the delay, it was a little hard to review as it seems the code was moved around but the diff is not very helpful. And sometime it's more than moving around  (like the fatal/astroid-error change).

DanielNoord:
> LGTM ! Sorry for the delay, it was a little hard to review as it seems the code was moved around but the diff is not very helpful. And sometime it's more than moving around (like the fatal/astroid-error change).

No worries, just thought you might have missed this as you had reviewed more recent PRs. Thanks for the review 😄

PR #7301, opened 2022-08-13
Fix false positive for `no-self-argument`

### LINKED ISSUE #7300: False positive on decorators that return a `staticmethod`
### Bug description

In some usecases, a decorator is applied to a method of a python class, and it returns a staticmethod instead. Those usecases are supported by mypy with a `TYPE_CHECKING` guard: https://github.com/python/mypy/issues/6726#issuecomment-486907505.

```python
# pylint: disable=missing-docstring,invalid-name

if TYPE_CHECKING:
    returns_staticmethod = staticmethod
else:
    def returns_staticmethod(f): ...


class MyClass:
    @returns_staticmethod
    def my_staticmethod(a, b): # <======== pylint (E0213): Method should have "self" as first argument
        return a + b
```

However, testing with pylint, an error pops up indicating `self` is not the first argument of the method above. Additionally, even if we are not using the `TYPE_CHECKING` guard, pylint does not seem to  propagate `staticmethod` decorator:

```python
# pylint: disable=missing-docstring,invalid-name,too-few-public-methods
my_decorator = staticmethod

class MyClass:
    @my_decorator
    def my_decorated_func(a, b): # <======== pylint (E0213): Method should have "self" as first argument
        return a + b
```

Ideally, if pylint propagates `staticmethod` through `my_decorator`, the false positive would go away.

Thanks in advance!

### Configuration

Configuration is not relevant

### Command used

```shell
pylint main.py
```

### Pylint output

```shell
************* Module main
main.py:14:4: E0213: Method should have "self" as first argument (no-self-argument)
```

### Expected behavior

Expected that pylint propagates `staticmethod` via `my_decorator` so that the snippets below will work without warning:

```python
# pylint: disable=missing-docstring,invalid-name,too-few-public-methods
my_decorator = staticmethod

class MyClass:
    @my_decorator
    def my_decorated_func(a, b): # <======== pylint (E0213): Method should have "self" as first argument
        return a + b
```

### Pylint version

```shell
>>> pylint --version
pylint 2.12.2
astroid 2.9.0
Python 3.8.13 (default, Mar 28 2022, 11:38:47)
[GCC 7.5.0]
```

### OS / Environment

Ubuntu 20.04

### Additional dependencies

N/A

mbyrnepr2:
## Type of Changes

## Description
Fix false positive for `no-self-argument` when a staticmethod is applied to a function but uses a different name.

Closes #7300

### DIFF
--- a/doc/whatsnew/fragments/7300.false_positive
+++ b/doc/whatsnew/fragments/7300.false_positive
@@ -1,0 +1,3 @@
+Fix false positive for `no-self-argument` when a staticmethod is applied to a function but uses a different name.
+
+Closes #7300
--- a/doc/whatsnew/fragments/7300.false_positive
+++ b/doc/whatsnew/fragments/7300.false_positive
@@ -1,3 +1,3 @@
-Fix false positive for `no-self-argument` when a staticmethod is applied to a function but uses a different name.
+Fix false positive for `no-self-argument`/`no-method-argument` when a staticmethod is applied to a function but uses a different name.
 
 Closes #7300
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1926,7 +1926,7 @@ a metaclass class method.",
                 self.add_message("bad-staticmethod-argument", args=first, node=node)
                 return
             self._first_attrs[-1] = None
-        elif any(name == "builtins.staticmethod" for name in node.decoratornames()):
+        elif "builtins.staticmethod" in node.decoratornames():
             # Check if there is a decorator which is not named `staticmethod` but is assigned to one.
             return
         # class / regular method with no args
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1926,6 +1926,9 @@ a metaclass class method.",
                 self.add_message("bad-staticmethod-argument", args=first, node=node)
                 return
             self._first_attrs[-1] = None
+        elif any(name == "builtins.staticmethod" for name in node.decoratornames()):
+            # Check if there is a decorator which is not named `staticmethod` but is assigned to one.
+            return
         # class / regular method with no args
         elif not node.args.args and not node.args.posonlyargs:
             self.add_message("no-method-argument", node=node)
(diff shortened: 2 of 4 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -1926,6 +1926,9 @@ def _check_first_arg_for_type(
                 self.add_message("bad-staticmethod-argument", args=first, node=node)
                 return
             self._first_attrs[-1] = None
+        elif any(name == "builtins.staticmethod" for name in node.decoratornames()):
```
jacobtylerwalls, line 1929:
```suggestion
        elif "builtins.staticmethod" in node.decoratornames():
```

tests/functional/n/no/no_self_argument.py
```diff
@@ -16,3 +23,18 @@ def abdc(yoo):  # [no-self-argument]
...
+        """A standard staticmethod"""
+        print("hello!")
+
+    @MYSTATICMETHOD
+    def say_goodbye():
+        """A staticmethod but using a different name"""
+        print("goodbye!")
+
+    @returns_staticmethod
+    def sum_strings(string1, string2):
```
jacobtylerwalls, line 38:
```suggestion
    def concatenate_strings(string1, string2):
```

doc/whatsnew/fragments/7300.false_positive
```diff
@@ -0,0 +1,3 @@
+Fix false positive for `no-self-argument` when a staticmethod is applied to a function but uses a different name.
```
jacobtylerwalls, line 1:
```suggestion
Fix false positive for `no-self-argument`/`no-method-argument` when a staticmethod is applied to a function but uses a different name.
```

tests/functional/n/no/no_self_argument.py
jacobtylerwalls: Just noting for other reviewers that this test case already passed. (It's the MYSTATICMETHOD that newly passes.)

jacobtylerwalls (approved):
Thanks! Looks good, just some nitpicks I noticed.

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: Wow I did not see that, this is clever 😄

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):

junrushao1994:
Thank you for fixing the issue @mbyrnepr2!!

PR #7320, opened 2022-08-18
Remove Python 2 code from the tests & refactor:

mbyrnepr2:
## Type of Changes

## Description

Inspired by the comment https://github.com/PyCQA/pylint/pull/7301/files#r945127634

Refactoring the test suite to update the code a bit with regard to:
- Importing print_function from __future__
- Classes which explicitly inherit `object`
- Remove unused assignments to `__revision__`


tests/functional/a/abstract/abstract_abc_methods.py
```diff
@@ -1,9 +1,9 @@
 """ This should not warn about `prop` being abstract in Child """
-# pylint: disable=too-few-public-methods, useless-object-inheritance
+# pylint: disable=too-few-public-methods
 
 import abc
 
-class Parent(object):
```
Pierre-Sassoulas, line 6: We need to keep **some** tests for ``useless-object-inheritance`` this is a python 3 checks : habits die hard and what was necessary in python 2 is not anymore but people keep doing it.

tests/functional/ext/private_import/private_import.py
```diff
@@ -18,8 +18,6 @@
 # Ignore dunders
 import __asd__
 import __future__
-from __future__ import print_function
-from __future__ import __print_function__
```
Pierre-Sassoulas, line 22: We should probably replace by something similar here, it feel like it's part of what is tested.

tests/functional/i/inherit_non_class.py
```diff
@@ -1,8 +1,8 @@
 """Test that inheriting from something which is not
 a class emits a warning. """
 
-# pylint: disable=import-error, invalid-name, using-constant-test, useless-object-inheritance
-# pylint: disable=missing-docstring, too-few-public-methods
+# pylint: disable=import-error, invalid-name, using-constant-test
+# pylint: disable=missing-docstring, too-few-public-methods, useless-object-inheritance
```
Pierre-Sassoulas, line 5:
```suggestion
# pylint: disable=missing-docstring, too-few-public-methods

```

tests/functional/i/iterable_context.py
```diff
@@ -2,8 +2,7 @@
 Checks that primitive values are not used in an
 iterating/mapping context.
 """
-# pylint: disable=missing-docstring,invalid-name,too-few-public-methods,import-error,unused-argument,bad-mcs-method-argument,wrong-import-position,no-else-return, useless-object-inheritance, unnecessary-comprehension,redundant-u-string-prefix
-from __future__ import print_function
+# pylint: disable=missing-docstring,invalid-name,too-few-public-methods,import-error,unused-argument,bad-mcs-method-argument,wrong-import-position,no-else-return, unnecessary-comprehension,redundant-u-string-prefix
```
Pierre-Sassoulas, line 6:
```suggestion
# pylint: disable=missing-docstring,invalid-name,too-few-public-methods,import-error,unused-argument,bad-mcs-method-argument,
# pylint: disable=wrong-import-position,no-else-return, unnecessary-comprehension,redundant-u-string-prefix
```
This permit to reduce the diff too as the line do not change if we keep using two lines.

tests/functional/n/name/name_styles.py
```diff
@@ -1,6 +1,5 @@
 """Test for the invalid-name warning."""
-# pylint: disable=useless-object-inheritance, unnecessary-pass, unnecessary-comprehension, unused-private-member, unnecessary-lambda-assignment
-from __future__ import print_function
+# pylint: disable=unnecessary-pass, unnecessary-comprehension, unused-private-member, unnecessary-lambda-assignment
```
Pierre-Sassoulas, line 2:
```suggestion
# pylint: disable=unnecessary-pass, unnecessary-comprehension, unused-private-member
# pylint: disable=unnecessary-lambda-assignment
```

tests/functional/n/namedtuple_member_inference.py
```diff
@@ -3,7 +3,6 @@
 Regression test for:
 https://bitbucket.org/logilab/pylint/issue/93/pylint-crashes-on-namedtuple-attribute
 """
-from __future__ import absolute_import, print_function
 from collections import namedtuple
 
 __revision__ = None
```
Pierre-Sassoulas, line 8: This is in a lot of file and I think it's useless too. Would you mind also removing them in this cleanup MR ?
```suggestion
```

Pierre-Sassoulas (changes requested):
Amazing cleanup ! Clearly a lot of work went into this. As I said in my first comment I think we should add a functional test specifically for ``useless-object-inheritance``, it was tested everywhere before, now it can live in a small clean three line file. That's pretty neat 🎉

tests/functional/ext/private_import/private_import.py
mbyrnepr2: Perhaps best to revert this one; it does indeed look like it is useful for the private import checking.

tests/functional/a/abstract/abstract_abc_methods.py
mbyrnepr2: Does [useless-object-inheritance.py ](https://github.com/PyCQA/pylint/blob/main/tests/functional/u/useless/useless_object_inheritance.py)cover it already?

tests/functional/i/inherit_non_class.py
mbyrnepr2: I think this one is necessary because of what [this test] is trying (https://github.com/PyCQA/pylint/blob/main/tests/functional/i/inherit_non_class.py#L27)

tests/functional/a/abstract/abstract_abc_methods.py
Pierre-Sassoulas: Sure ! I did not check, sorry 😄

tests/functional/ext/private_import/private_import.py
Pierre-Sassoulas: Yeah sure, we could update the import to something more modern but there's no real reason to do that.

tests/functional/n/namedtuple_member_inference.py
mbyrnepr2: Sure! There's a good few of these also

tests/functional/c/class_attributes.py
```diff
@@ -1,7 +1,7 @@
 """Test that valid class attribute doesn't trigger errors"""
-__revision__ = 'sponge bob'
```
Pierre-Sassoulas, line 2: All right, someone had fun here 😄

Pierre-Sassoulas (approved):
Very nice thank you 👍 I'm going to merge as the fail is unrelated (a very old link that no one use is now broken in the doc). not worth fixing before rebasing.merging.

tests/functional/c/class_attributes.py
mbyrnepr2: It feels a real shame to remove such a humorous line :D

PR #7322, opened 2022-08-19
Improve TypeVar name regex

cdce8p:
## Description
Small improvements for the `TypeVar` name regex.

* Simplify `[^\W\da-z_]` to `[A-Z]`
* Add negative lookahead to prevent `T` or `Type` prefix. E.g. `TAnyStr` should be marked as `invalid-name`.

### DIFF
--- a/doc/whatsnew/fragments/7322.false-positive
+++ b/doc/whatsnew/fragments/7322.false-positive
@@ -1,2 +1,4 @@
 Improve default TypeVar name regex. Disallow names prefixed with ``T``.
 E.g. use ``AnyStrT`` instead of ``TAnyStr``.
+
+Refs #7322
--- a/doc/whatsnew/fragments/7322.false-positive
+++ b/doc/whatsnew/fragments/7322.false-positive
@@ -1,0 +1,2 @@
+Improve default TypeVar name regex. Disallow names prefixed with ``T`` or ``Type``.
+E.g. use ``AnyStrT`` instead of ``TAnyStr``.
--- a/doc/whatsnew/fragments/7322.false-positive
+++ b/doc/whatsnew/fragments/7322.false-positive
@@ -1,2 +1,2 @@
-Improve default TypeVar name regex. Disallow names prefixed with ``T`` or ``Type``.
+Improve default TypeVar name regex. Disallow names prefixed with ``T``.
 E.g. use ``AnyStrT`` instead of ``TAnyStr``.
--- a/doc/data/messages/i/invalid-name/details.rst
+++ b/doc/data/messages/i/invalid-name/details.rst
@@ -84,7 +84,7 @@ The following type of names are checked with a predefined pattern:
 | Name type          | Good names                                        | Bad names                                                  |
 +====================+===================================================+============================================================+
 | ``typevar``        | ``T``, ``_CallableT``, ``_T_co``, ``AnyStr``,     | ``DICT_T``, ``CALLABLE_T``, ``ENUM_T``, ``DeviceType``,    |
-|                    | ``DeviceTypeT``, ``IPAddressT``                   | ``_StrType``                                               |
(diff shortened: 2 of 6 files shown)


DanielNoord (dismissed):

cdce8p:
@DanielNoord Do you know what I need to do for the checks to pass? Should I change the fragment suffix to `.other` or do I always need a `Ref`?

doc/whatsnew/fragments/7322.false-positive
```diff
@@ -0,0 +1,2 @@
+Improve default TypeVar name regex. Disallow names prefixed with ``T``.
+E.g. use ``AnyStrT`` instead of ``TAnyStr``.
```
Pierre-Sassoulas, line 2:
```suggestion
E.g. use ``AnyStrT`` instead of ``TAnyStr``.

Refs #7322
```

Pierre-Sassoulas (approved):

DanielNoord:
@Pierre-Sassoulas We still can't seem to merge without updating branches 😢

doc/whatsnew/fragments/7322.false-positive
Pierre-Sassoulas: We had a discussion recently about this, the other solution is to create an issue especially for that.

Pierre-Sassoulas:
> We still can't seem to merge without updating branches

I'm going to look into it, I'm not on mobile anymore. I hope you'll get admin-ship soon, be it the easy way or the hard way 😄

Pierre-Sassoulas (approved):
Seems like requiring a new review if the code change also auto-checked the "up to date code" option.

mattclay:
@cdce8p Where does the recommendation to avoid a `T` prefix come from?

DanielNoord:
@mattclay I think this follows from what we have "seen in the wild". Note that you can allow them again by setting the `typevar-rgx` pattern config in your configuration file or on the command line.

yilei:
The new regex also disallows things like `_T1` `_T2` which is a quite common practice

mattclay:
@DanielNoord I don't know how common the `T` prefix is, but it's definitely used "in the wild". For example, it's used in the ansible project, where I counted at least 24 uses. Here's one example:

https://github.com/ansible/ansible/blob/f8e83264372d583a8f819f7bfe03a0f53f1abc0a/test/lib/ansible_test/_internal/config.py#L39

I realize the pattern can be customized, but it seems like an unnecessarily arbitrary stylistic restriction for a default.

DanielNoord:
> The new regex also disallows things like `_T1` `_T2` which is a quite common practice

This is problematic indeed.

> I realize the pattern can be customized, but it seems like an unnecessarily arbitrary stylistic restriction for a default.

@cdce8p Do you have on opinion on this? If not I think we might want to revert this change. `_T1` seems like it should be allowed and perhaps allowing other examples with a prefix of `T` is also better?

cdce8p:
> @cdce8p Where does the recommendation to avoid a `T` prefix come from?

I mainly looked at a number of typing PEPs as well as the TypeVar names used within [typeshed](https://github.com/python/typeshed) as reference. Those do strongly prefer a `T` suffix over prefix which is why I advocated for that pattern.

> The new regex also disallows things like `_T1` `_T2` which is a quite common practice

It's not as common as you may think. I would even argue that it's an antipattern which should be avoided except for very specific cases. It's much better to use different / more verbose names to avoid confusion and improve readability. Some common once include `S`, `T`, `U`, `V`, `KT`, `VT`. Just as an example take a look at the TypeVars defined in [`typeshed/stdlib/typing.pyi`](https://github.com/python/typeshed/blob/d1375f600cc4a7e7d81dcd9229c6abb933efda2a/stdlib/typing.pyi#L243-L252).

PR #7323, opened 2022-08-20
Add enum regression test

cdce8p:
## Description
Add regression test for crash added in https://github.com/PyCQA/astroid/pull/1734
Which was fixed in https://github.com/PyCQA/astroid/pull/1743


mbyrnepr2 (approved):

mbyrnepr2:
Thanks @cdce8p!

PR #7324, opened 2022-08-20
Fix `undefined-loop-variable` from walrus in comprehension test

### LINKED ISSUE #7222: undefined-loop-variable in walrus in comprehension test
### Bug description

I think I'm getting a spurious undefined-loop-variable warning when:
- walrus expression in 'if' block of list comprehension
- loop variable in list comprehension (`x`, here) is the same as a previous loop

```python
# pylint: disable=C0114,C0116,C0103

def example(arr):
    for x in arr:
        pass
    print([y for x in arr if (y := x)]) # this warns undefined-loop-variable
    print([y for x in arr if x]) # this is fine

def example2(arr):
    print([y for x in arr if (y := x)]) # this is fine
```

### Configuration

_No response_

### Command used

```shell
pylint example.py
```

### Pylint output

```shell
************* Module example
example.py:6:35: W0631: Using possibly undefined loop variable 'x' (undefined-loop-variable)

Your code has been rated at 8.57/10 (previous run: 8.00/10, +0.57)
```

### Expected behavior

Shouldn't warn.

### Pylint version

```shell
pylint 2.14.5
astroid 2.11.7
Python 3.9.7 (default, Jun 22 2022, 20:11:26) 
[GCC 11.2.0]
```

### OS / Environment

ubuntu 21.10

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description
Closes #7222

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2249,6 +2249,22 @@ class VariablesChecker(BaseChecker):
         ):
             return
 
+        maybe_walrus = utils.get_node_first_ancestor_of_type(node, nodes.NamedExpr)
+        if maybe_walrus:
+            maybe_comprehension = utils.get_node_first_ancestor_of_type(
+                maybe_walrus, nodes.Comprehension
+            )
+            if maybe_comprehension:
+                comprehension_scope = utils.get_node_first_ancestor_of_type(
+                    maybe_comprehension, nodes.ComprehensionScope
+                )
+                assert comprehension_scope is not None
+                if (
+                    comprehension_scope.parent.scope() is scope
+                    and node.name in comprehension_scope.locals
+                ):
+                    return
+
         # For functions we can do more by inferring the length of the itered object
         try:
             inferred = next(assign.iter.infer())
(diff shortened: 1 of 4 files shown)


pylint/checkers/variables.py
```diff
@@ -2249,6 +2249,22 @@ def _loopvar_name(self, node: astroid.Name) -> None:
...
+        maybe_walrus = utils.get_node_first_ancestor_of_type(node, nodes.NamedExpr)
+        if maybe_walrus:
+            maybe_comprehension = utils.get_node_first_ancestor_of_type(
+                maybe_walrus, nodes.Comprehension
+            )
+            if maybe_comprehension:
+                comprehension_scope = utils.get_node_first_ancestor_of_type(
+                    maybe_comprehension, nodes.ComprehensionScope
+                )
+                assert comprehension_scope is not None
```
Pierre-Sassoulas, line 2261: This is making me nervous, maybe we can use a safer way to appease mypy ?

Pierre-Sassoulas (commented):
Looks pretty good already !

pylint/checkers/variables.py
jacobtylerwalls: Not a problem. I wish we had a utility that would alert us if/when these logic assertions fail, so that we could be safe for users and still enlighten ourselves :-)

Pierre-Sassoulas (approved):
👌

PR #7330, opened 2022-08-21
Some fixes to the primer

### LINKED ISSUE #7328: Primer test does not fail a CI correctly when it finds a crash
### Bug description

When a PR creates a new crash in one of the projects the CI should fail. Currently this isn't the case.

See:
https://github.com/PyCQA/pylint/pull/7327#issuecomment-1221539878
And the associated run:
https://github.com/PyCQA/pylint/runs/7938783923?check_suite_focus=true

DanielNoord:

## Type of Changes

## Description

Closes #7328

### DIFF
--- a/.github/workflows/primer_comment.yaml
+++ b/.github/workflows/primer_comment.yaml
@@ -134,10 +134,3 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           leave_visible: 1
           issue_number: ${{ steps.post-comment.outputs.result }}
-      - name: Warn about failure
-        if: ${{ failure() }}
-        run: |
-          echo "🤖 **Comment workflow failed**. 🤖" >> $GITHUB_STEP_SUMMARY
-          echo "" >> $GITHUB_STEP_SUMMARY
-          echo "Please investigate:" >> $GITHUB_STEP_SUMMARY
-          echo "@DanielNoord" >> $GITHUB_STEP_SUMMARY
--- a/pylint/testutils/_primer/primer_run_command.py
+++ b/pylint/testutils/_primer/primer_run_command.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 
 import json
 import sys
-import warnings
 from io import StringIO
 
 from pylint.lint import Run
@@ -23,20 +22,11 @@ CRASH_TEMPLATE_INTRO = "There is a pre-filled template"
 class RunCommand(PrimerCommand):
     def run(self) -> None:
         packages: dict[str, list[OldJsonExport]] = {}
-        astroid_errors: list[Message] = []
-        other_fatal_msgs: list[Message] = []
+        fatal_msgs: list[Message] = []
         for package, data in self.packages.items():
-            messages, p_astroid_errors, p_other_fatal_msgs = self._lint_package(
-                package, data
-            )
-            astroid_errors += p_astroid_errors
-            other_fatal_msgs += p_other_fatal_msgs
+            messages, p_fatal_msgs = self._lint_package(package, data)
+            fatal_msgs += p_fatal_msgs
             packages[package] = messages
-        plural = "s" if len(other_fatal_msgs) > 1 else ""
-        assert not other_fatal_msgs, (
-            f"We encountered {len(other_fatal_msgs)} fatal error message{plural}"
-            " that can't be attributed to bleeding edge astroid alone (see log)."
-        )
         path = (
             self.primer_directory
             / f"output_{'.'.join(str(i) for i in sys.version_info[:3])}_{self.config.type}.txt"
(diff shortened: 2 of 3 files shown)


.github/workflows/primer_comment.yaml
```diff
@@ -134,10 +134,3 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           leave_visible: 1
           issue_number: ${{ steps.post-comment.outputs.result }}
-      - name: Warn about failure
-        if: ${{ failure() }}
-        run: |
-          echo "🤖 **Comment workflow failed**. 🤖" >> $GITHUB_STEP_SUMMARY
-          echo "" >> $GITHUB_STEP_SUMMARY
-          echo "Please investigate:" >> $GITHUB_STEP_SUMMARY
-          echo "@DanielNoord" >> $GITHUB_STEP_SUMMARY
```
DanielNoord, line 143: This didn't work sadly.. :(

pylint/testutils/_primer/primer_run_command.py
```diff
@@ -23,57 +22,48 @@
...
+        # Assert that a PR run does not introduce new fatal errors
+        if self.config.type == "pr":
+            plural = "s" if len(fatal_msgs) > 1 else ""
+            assert (
+                not fatal_msgs
+            ), f"We encountered {len(fatal_msgs)} fatal error message{plural} (see log)."
 
     @staticmethod
-    def _filter_astroid_errors(
+    def _filter_fatal_errors(
```
DanielNoord, line 45: The failure that prompted me to open the issue clearly shows that `astroid-error` does not mean the error actually comes from `astroid` 😄 

Let's remove the distinction.

pylint/testutils/_primer/primer_run_command.py
```diff
@@ -23,57 +22,48 @@
...
-            astroid_errors += p_astroid_errors
-            other_fatal_msgs += p_other_fatal_msgs
+            messages, p_fatal_msgs = self._lint_package(package, data)
+            fatal_msgs += p_fatal_msgs
             packages[package] = messages
-        plural = "s" if len(other_fatal_msgs) > 1 else ""
-        assert not other_fatal_msgs, (
-            f"We encountered {len(other_fatal_msgs)} fatal error message{plural}"
-            " that can't be attributed to bleeding edge astroid alone (see log)."
-        )
```
jacobtylerwalls, line 39: Just checking my understanding: if a fatal error happens on a main run that has nothing to do with `astroid-error`, we no longer fail the job (we just warn in a comment)? I wonder if we should still fail for non-astroid errors.

pylint/testutils/_primer/primer_run_command.py
```diff
@@ -90,21 +80,12 @@ def _lint_package(
...
-            if astroid_errors:
-                warnings.warn(
-                    f"Fatal messages that could be related to bleeding edge astroid:\n"
-                    f"{self._print_msgs(astroid_errors)}\n\n"
-                )
-        return messages, astroid_errors, other_fatal_msgs
+            fatal_msgs = self._filter_fatal_errors(messages)
+            if fatal_msgs:
+                print(f"Encountered fatal errors while priming {package_name} !\n")
+                print(f"{self._print_msgs(fatal_msgs)}\n\n")
```
jacobtylerwalls, line 90: If we `print` instead of `warnings.warn` then it doesn't summarize as nicely in Github Actions (it will get lost).

pylint/testutils/_primer/primer_run_command.py
DanielNoord: Yeah. See my comment below. Relying on `astroid-error` clearly doesn't distinguish between "astroid bleeding edge" and "faulty PR code" enough as my PR caused an `astroid-error` without an astroid bump.
Thus, this gives a false sense of security which should hopefully be removed by now failing CIs on PR completely.

jacobtylerwalls: Yeah, I get that. I have no problem with the PR case. I'm wondering about the `main` case. Incompatible PRs are merged, or a new commit is pushed to a primer project, or some indeterminacy somewhere causes something to fail on main that we never noticed in a PR. Now, we can have failures on main with green CI. No?

DanielNoord: Yeah, but I think we "sort of" already could. The only `fatal` messages that are not `astroid-error` are:
`method-check-failed`, `parse-error`, `config-parse-error` and `fatal`.

1. I know `fatal` can only be raised in very specific circumstances and does not normally occur while linting a fail as there we raise `astroid-error` instead of `fatal` then.
2. `method-check-failed` doesn't have any tests so might me moot?
3. `parse-error` is raised when `astroid` can't import a module, so that's essentially a different `astroid-error`.
4. `config-parse-error` is related to configuration parsing and not very likely to trigger.

The problem is that `astroid-error` is our main error for anything causing a crash during linting. The other four are only raised in special circumstances. We weren't really catching fatal errors on `main` anyway 😅 
I think it makes more sense to find a good way to distinguish `astroid-errors` caused by bleeding edge or by a merged PR and act according to it.

jacobtylerwalls: Or a way to stop reraising any failures stemming from pure pylint as `AstroidError`, but that's backwards incompatible.

jacobtylerwalls (approved):

jacobtylerwalls:
Not a big deal, but I would have squashed the fixup commits so that we don't have "Use warnings" as a commit.

DanielNoord:
Yeah I was hesitating, but agreed, shouldn't have been lazy here.

PR #7331, opened 2022-08-21
Fix ``mypy`` issues in primer

DanielNoord:

## Type of Changes

## Description

Import directly from the `git.*` modules as `gitpython` uses some import guards that `mypy` doesn't really understand. Functionally this should be the same.


DanielNoord:
@jacobtylerwalls I have one primer PR after this which actually fixes/improves something I should have fixed a long time ago 😄

jacobtylerwalls (approved):
No problem with these changes, but could you add a sentence in the PR body explaining what the issue was?

DanielNoord:
Added!

PR #7334, opened 2022-08-22
Skip ``unnecessary-list-index-lookup`` if we encounter if statements

### LINKED ISSUE #6896: `unnecessary-list-index-lookup` false-positive when iterable is modified in loop
### Bug description

With some code like this:

```python

parts = ["a", "b", "c", "d"]
for i, part in enumerate(parts):
    if i == 3:  # more complex condition actually
        parts.insert(i, "X")
    print(part, parts[i])
```

([real-world example](https://github.com/qutebrowser/qutebrowser/blob/1996c4d7e72bb3764cd4ea316f8cf84d4fc32dca/qutebrowser/completion/completer.py#L154-L165)), pylint claims that `parts[i]` could be replaced by `part` - but as running the file shows, that isn't actually the case:

```
a a
b b
c c
d X
d d
```

You might think that this is a bit unorthodox (and I might agree, should probably rewrite this...), but if I'm reading the [Python docs](https://docs.python.org/3.7/reference/compound_stmts.html#the-for-statement) right, the behavior is well-defined (emphasis mine):

> There is a subtlety when the sequence is being modified by the loop (this can only occur for mutable sequences, e.g. lists). An internal counter is used to keep track of which item is used next, and this is incremented on each iteration. When this counter has reached the length of the sequence the loop terminates. This means that if the suite deletes the current (or a previous) item from the sequence, the next item will be skipped (since it gets the index of the current item which has already been treated). Likewise, **if the suite inserts an item in the sequence before the current item, the current item will be treated again the next time through the loop**.

cc @timmartin who added the checker in #5834.

### Configuration

_No response_

### Command used

```shell
pylint x.py
```

### Pylint output

```shell
************* Module x
x.py:7:16: R1736: Unnecessary list index lookup, use 'part' instead (unnecessary-list-index-lookup)
```

### Expected behavior

No error

### Pylint version

```shell
pylint 2.14.1
astroid 2.11.5
Python 3.10.5 (main, Jun  6 2022, 18:49:26) [GCC 12.1.0]
```

### OS / Environment

Archlinux

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #6896.

/CC @timmartin This makes the check very conservative but our current test suite still passes and thus the check still offers some benefit. I'd like to merge this and have us improve the checker if somebody wants to: avoiding false positives is more important than false negatives imo.

### DIFF
--- a/doc/whatsnew/fragments/6896.false_positive
+++ b/doc/whatsnew/fragments/6896.false_positive
@@ -1,0 +1,3 @@
+``unnecessary-list-index-lookup`` is now more conservative to avoid potential false positives.
+
+Closes #6896
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -2127,7 +2127,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
         )
         has_nested_loops = next(nested_loops, None) is not None
 
-        # Check if there are any if statemetns within the loop in question;
+        # Check if there are any if statements within the loop in question;
         # If so, we will be more conservative about reporting errors as we
         # can't yet do proper control flow analysis to be sure when
         # reassignment will affect us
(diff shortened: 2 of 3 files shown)


Pierre-Sassoulas (approved):
👌

PR #7342, opened 2022-08-23
Don't report ``trailing-whitespaces`` within strings

### LINKED ISSUE #3822: Inconsistent treatment of trailing whitespace in multiline strings
### Steps to reproduce
```python
def example(*args):
    print(*args)


example(
    "bob", """
    foobar 
    more text
""",
)

example(
    "bob",
    """
    foobar2 
    more text
""",
)
```

### Current behavior
pylint incorrectly warns about the trailing whitespace on the line with `foobar2` and correctly ignores the trailing whitespace on the line with `foobar`.

### Expected behavior
pylint should produce no trailing whitespace warnings, as all the occurrences are in literal strings.

### pylint --version output
pylint 2.6.0
astroid 2.4.2
Python 3.7.8 (default, Jul 30 2020, 09:48:41) 
[GCC 9.3.0]

DanielNoord:

## Type of Changes

## Description

Closes #3822.

### DIFF
--- a/doc/whatsnew/fragments/3822.false_positive
+++ b/doc/whatsnew/fragments/3822.false_positive
@@ -1,0 +1,3 @@
+``trailing-whitespaces`` is not longer reported within strings.
+
+Closes #3822
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -404,11 +404,10 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
                 # the full line; therefore we check the next token on the line.
                 if tok_type == tokenize.INDENT:
                     self.new_line(TokenWrapper(tokens), idx - 1, idx + 1)
-                # A tokenizer oddity: if a line contains a multi-line string,
-                # the NEWLINE also gets its own token which we already checked in
-                # the multi-line docstring case.
+                # We don't test for trailing whitespaces in strings
                 # See https://github.com/PyCQA/pylint/issues/6936
-                elif tok_type == tokenize.NEWLINE:
+                # and https://github.com/PyCQA/pylint/issues/3822
+                elif tok_type == tokenize.STRING:
                     pass
                 else:
                     self.new_line(TokenWrapper(tokens), idx - 1, idx)
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -273,7 +273,7 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
         line = tokens.line(line_start)
         if tokens.type(line_start) not in _JUNK_TOKENS:
             self._lines[line_num] = line.split("\n")[0]
-        self.check_lines(line, line_num)
+        self.check_lines(tokens, line_start, line, line_num)
 
     def process_module(self, node: nodes.Module) -> None:
         pass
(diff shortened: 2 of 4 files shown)


doc/whatsnew/fragments/3822.false_positive
```diff
@@ -0,0 +1,3 @@
+``trailing-whitespaces`` is not longer reported within strings.
```
Pierre-Sassoulas, line 1:
```suggestion
``trailing-whitespaces`` is no longer reported within strings.
```

DanielNoord:
Ah too bad. That function is also responsible for other messages. Back to the drawing board.

pylint/checkers/format.py
```diff
@@ -676,7 +667,14 @@ def check_lines(self, lines: str, lineno: int) -> None:
         split_lines = self.specific_splitlines(lines)
 
         for offset, line in enumerate(split_lines):
-            self.check_line_ending(line, lineno + offset)
+            if not line.endswith("\n"):
```
Pierre-Sassoulas, line 670: It seems like the check is done on every line, but it could be done only on the last line ? Also what about cross platform line ending (``\r\n```and ```\r``) for Windows and Mac ?

pylint/checkers/format.py
```diff
@@ -676,7 +667,14 @@ def check_lines(self, lines: str, lineno: int) -> None:
         split_lines = self.specific_splitlines(lines)
```
Pierre-Sassoulas, line 668:
```suggestion
last_line =  split_lines[-1][1]
offset =  split_lines[-1][0]
if not last_line.endswith("\n"):
    self.add_message("missing-final-newline", line=lineno + offset)
```

Pierre-Sassoulas (changes requested):

pylint/checkers/format.py
DanielNoord: I'm not sure. I just had to extract this out of the method because there was no simple way to not call the check of `trailing-whitespace`. I'll try out your suggestion but since we don't have any issues about this I don't think we should also change the line endings that are checked in this PR.

DanielNoord: Oh never mind.

I can't extract this. We need to be able to skip the `trailing-whitespace` check if we added the `missing-final-newline`. We can do something with a `emitted_final_newline` variable but I think checking `if emitted_final_newline` is as quick as `if line.endswith("\n")` and the current code is a bit clearer.

Pierre-Sassoulas: > We need to be able to skip the trailing-whitespace check if we added the missing-final-newline.

How about we loop on ``split_lines[:-1]`` and handle the last line separately ? I don't like knowingly making a check O(n) when it could be O(1)

DanielNoord: I don't think there is a good way without duplicating a big part of code + the comment? There is also the option to do something like `offset == len(split_lines)` but then again, I'm not sure an `endswith` is that much slower.

Pierre-Sassoulas: What about something like this ?
```python
for offset, line in enumerate(split_lines[-1]):
    self.check_trailing_whitespace_ending(last_line, lineno + offset, line_start):
last_line =  split_lines[-1][1]
offset =  split_lines[-1][0]
if not last_line.endswith("\n"):
    self.add_message("missing-final-newline", line=lineno + offset)
else:
    self.check_trailing_whitespace_ending(last_line, lineno + offset, line_start):

def check_trailing_whitespace_ending(...)
    # We don't test for trailing whitespaces in strings
    # See https://github.com/PyCQA/pylint/issues/6936
    # and https://github.com/PyCQA/pylint/issues/3822
    if tokens.type(line_start) == tokenize.STRING:
        return
    ...
```

DanielNoord: This also doesn't work since `split_lines` isn't always full, so `split_lines[-1]` can return an `IndexError` 😠 

I don't think there is anyway to do this without adding additional if statements and assignments which kind of defeat the purpose of calling `endswith` once.

Pierre-Sassoulas (approved):
I took some time to check the code around this and while I still think something that should be 0(1) is O(n) changing if would require changing the strange regex that I don't understand and seems to match trailing white spaces by also matching ``\n`` ...  (?!) So it might be possible but it's not going to be easy. Also currently this code is "not broken" so I'm going "not touch it" and trust you on this one 😄

PR #7354, opened 2022-08-25
[literal-comparison] Make the message more explicit with the solution

### LINKED ISSUE #5237: R0123 (literal-comparison) message is not explicit about the problem and the solution
### Current problem

R0123: Comparison to literal (literal-comparison), which I think could be made more explicit. Maybe something like a.py:3:25: R0123: 'val_value is ""' could be replaced by 'val_value' (literal-comparison) (or bool(val_value) if it's not a context where we have an auto-cast to boolean ?)

### Desired solution

Flake8 says : F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple). we can also piggy back on the documentation for this flake8 check when pylint users search for the error messages.

> Maybe we should include the words "Truth Value Testing" as this is the official wording in the Python docs, and people who want to know more can therefore easily find the page linked above.
> 'if val_value is ""' could be replaced by 'if val_value' (Truth Value Testing) (literal-comparison)

### Additional context

https://github.com/PyCQA/pylint/issues/5234#issuecomment-955671502

Pierre-Sassoulas:
## Type of Changes

## Description

Also update the confidence

Closes #5237

### DIFF
--- a/tests/functional/c/comparison_of_constants.py
+++ b/tests/functional/c/comparison_of_constants.py
@@ -1,9 +1,5 @@
 # pylint: disable=missing-docstring, comparison-with-itself, invalid-name
 
-
-if 2 is 2:  # [literal-comparison, comparison-of-constants]
-    pass
-
 while 2 == 2:  # [comparison-of-constants]
     pass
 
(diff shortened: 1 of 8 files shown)


tests/functional/c/comparison_of_constants.py
```diff
@@ -22,14 +18,6 @@
 
 CONST = 24
 
-if CONST is 0:  # [literal-comparison]
```
DanielNoord, line 25: Why did you remove these?

DanielNoord (commented):
Really like the new message, nice work!

Just one question!

tests/functional/c/comparison_of_constants.py
Pierre-Sassoulas: It's duplicated from the literal comparison functional tests.

DanielNoord (approved):

PR #7357, opened 2022-08-25
Patch ``sys.path`` on a per-file basis

DanielNoord:

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/7339

I agree with @gcbirzan-plutoflume that the updated test outputs seem like they are actually beter. The change itself also makes sense imo, as Python also fixes `sys.path` once. See the issue for some further discussion.

### DIFF
--- a/pylint/lint/utils.py
+++ b/pylint/lint/utils.py
@@ -10,13 +10,11 @@ import sys
 import traceback
 from collections.abc import Iterator, Sequence
 from datetime import datetime
-from functools import lru_cache
 from pathlib import Path
 
 from pylint.config import PYLINT_HOME
 
 
-@lru_cache(maxsize=None)
 def get_python_path(filepath: str) -> str:
     """TODO This get the python path with the (bad) assumption that there is always
     an __init__.py.
--- a/pylint/lint/utils.py
+++ b/pylint/lint/utils.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import contextlib
+import os
 import sys
 import traceback
 from collections.abc import Iterator, Sequence
@@ -12,7 +13,24 @@ from datetime import datetime
 from pathlib import Path
 
 from pylint.config import PYLINT_HOME
-from pylint.lint.expand_modules import get_python_path
+
+
+def get_python_path(filepath: str) -> str:
+    """TODO This get the python path with the (bad) assumption that there is always
+    an __init__.py.
+
+    This is not true since python 3.3 and is causing problem.
+    """
+    dirname = os.path.realpath(os.path.expanduser(filepath))
+    if not os.path.isdir(dirname):
+        dirname = os.path.dirname(dirname)
+    while True:
+        if not os.path.exists(os.path.join(dirname, "__init__.py")):
+            return dirname
+        old_dirname = dirname
+        dirname = os.path.dirname(dirname)
+        if old_dirname == dirname:
+            return os.getcwd()
 
 
 def prepare_crash_report(ex: Exception, filepath: str, crash_file_path: str) -> Path:
--- a/pylint/lint/utils.py
+++ b/pylint/lint/utils.py
@@ -10,11 +10,13 @@ import sys
 import traceback
 from collections.abc import Iterator, Sequence
 from datetime import datetime
+from functools import lru_cache
 from pathlib import Path
 
 from pylint.config import PYLINT_HOME
 
 
+@lru_cache(maxsize=None)
 def get_python_path(filepath: str) -> str:
     """TODO This get the python path with the (bad) assumption that there is always
     an __init__.py.
--- a/tests/lint/unittest_lint.py
+++ b/tests/lint/unittest_lint.py
@@ -12,6 +12,7 @@ import os
 import re
 import sys
 import tempfile
+import textwrap
 from collections.abc import Iterable, Iterator
 from contextlib import contextmanager
 from importlib import reload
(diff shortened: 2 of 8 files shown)


tests/test_functional.py
```diff
@@ -50,7 +51,11 @@ def test_functional(
     else:
         lint_test = testutils.LintModuleTest(test_file, pytestconfig)
     lint_test.setUp()
-    lint_test.runTest()
+    # Pytest changes 'sys.path' depending on the files your running over.
+    # To fix this, we simply patch sys.path like we do in a normal run, but
+    # as if we're running against the tes file.
```
Pierre-Sassoulas, line 56:
```suggestion
    # as if we're running against the test file.
```

pylint/lint/utils.py
```diff
@@ -5,14 +5,32 @@
...
+    dirname = os.path.realpath(os.path.expanduser(filepath))
+    if not os.path.isdir(dirname):
+        dirname = os.path.dirname(dirname)
+    while True:
+        if not os.path.exists(os.path.join(dirname, "__init__.py")):
+            return dirname
+        old_dirname = dirname
+        dirname = os.path.dirname(dirname)
+        if old_dirname == dirname:
+            return os.getcwd()
```
Pierre-Sassoulas, line 33: This condition is never true during the tests, could we add a test ?

DanielNoord: This is currently uncovered right? I'd like the investigate the TODO in the comment at some point, but for now I think focusing on the issue at the hand is big enough. I'm kind of worried about the effect of now warning on the correct module for users who rely on the false module right now.

Pierre-Sassoulas: Fair enough !

pylint/lint/utils.py
```diff
@@ -5,14 +5,32 @@
...
 import traceback
 from collections.abc import Iterator, Sequence
 from datetime import datetime
 from pathlib import Path
 
 from pylint.config import PYLINT_HOME
-from pylint.lint.expand_modules import get_python_path
+
+
+def get_python_path(filepath: str) -> str:
```
Pierre-Sassoulas, line 18: Should we cache this and/or ``fix_import_path`` now that we launch the function a lot ?

DanielNoord: `fix_import_path` is an iterator that changes `sys.path` so I don't think we can cache that.

I'll cache `get_python_path`.

DanielNoord: Oh never mind, I don't think `get_python_path` ever gets called with the same argument, as it is part of the exploration of modules and we don't "explore the same module twice".

Pierre-Sassoulas (dismissed):

tests/test_functional.py
```diff
@@ -68,6 +73,7 @@ def test_functional(
         UPDATE_FILE.touch()
         sys.argv.remove(UPDATE_OPTION)
     try:
+
```
jacobtylerwalls, line 76:
```suggestion
```

tests/test_functional.py
```diff
@@ -50,7 +51,11 @@ def test_functional(
     else:
         lint_test = testutils.LintModuleTest(test_file, pytestconfig)
     lint_test.setUp()
-    lint_test.runTest()
+    # Pytest changes 'sys.path' depending on the files your running over.
```
jacobtylerwalls, line 54:
```suggestion
    # Pytest changes 'sys.path' depending on the files you're running over.
```

tests/lint/unittest_lint.py
```diff
@@ -986,3 +987,32 @@ def test_lint_namespace_package_under_dir_on_path(initialized_linter: PyLinter)
         with fix_import_path([tmpdir]):
             linter.check(["namespace_on_path"])
     assert linter.file_state.base_name == "namespace_on_path"
+
+
+def test_sys_path_patching_for_namespace_package(initialized_linter: PyLinter) -> None:
+    """Test that a normal package in the arguments doesn't pollute sys.path.
+
+    Originally reported in https://github.com/PyCQA/pylint/issues/7339.
+    If pylint is ran over a file within a namespace package which
```
jacobtylerwalls, line 996:
```suggestion
    If pylint is run over a file within a namespace package which
```

jacobtylerwalls (commented):
I agree this looks promising. Some trivial cosmetic things.

jacobtylerwalls:
@DanielNoord Have you tested this with namespace packages? Will it be a problem that `is_namespace()` is cached in `astroid`?

DanielNoord:
> @DanielNoord Have you tested this with namespace packages? Will it be a problem that `is_namespace()` is cached in `astroid`?

Hmm, I'm not sure. Do you think it should?

The `primer` failures show that this has significant performance regressions so I'm going to investigate (and fix) those before merging.

tests/test_functional.py
DanielNoord: Ouch...

jacobtylerwalls:
I guess I was wondering if we should be doing less caching: if a module name that was cached by astroid's` is_namespace` may or may not be a namespace depending on whether sys.path was patched in pylint.

DanielNoord:
> I guess I was wondering if we should be doing less caching: if a module name that was cached by astroid's` is_namespace` may or may not be a namespace depending on whether sys.path was patched in pylint.

If anything I think this should reduce the occurrence of this issue (the original issue describes a similar issue actually).

My thinking of this is as follows.
Take file A1, A2 and B1. A en B represent two packages which are both supplied on the command line, and after `expand_modules` A turns out to have two files.
Currently we would patch `sys.path` with both A and B, which is what is creating the issue as described in the original issue. As far as I can see (for now) whether or not A is a namespace package only depends on the file structure of package A. The patching of B onto `sys.path` should never correctly make A namespace and can only do so incorrectly as in the issue.
Since we know patch `sys.path` with A or A1/A2 if we can be even more specific `(something,)` vs. `(file.filepath,)` we never lint A1/A2 while patching B. This should avoid the previous issues.

The only thing I can think of that would be problematic is if package A imports package B before we lint package B and we determine during that import that B is namespace. However, I think that if we determine that it is namespace then it is likely correct: if you execute package A normally the interpreter also wouldn't magically patch package B onto your `sys.path` I think. So even if we cache some package as namespace before actually linting it is likely correct?

Not sure if I have explained myself enough here.

Also, `lru_cache` doesn't seem to work sadly 😢

DanielNoord:
Not really sure what to do with this PR. It's clear that creating the contextmanager for every file has significant performance impact, but I also can't think of a solution without doing so...

The contextmanager itself also isn't that large so there isn't really anything to refactor to make it quicker to set up..

DanielNoord:
Going to close this as I have never found a good solution to the clear performance impact this has.

PR #7359, opened 2022-08-25
Trying out codeowners configuration

Pierre-Sassoulas:
## Type of Changes

## Description

This permits to automatically ask for review by the person most knowledgeable about the code. I wanted to document this but in fact there's a codified document proposed by github which is better. Except if the review is required and the contributor is not available of course.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,8 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners will be the default owners for everything in the repo.
-* @pierre-sassoulas @DanielNoord @cdce8p @jacobtylerwalls @hippo91
+# Right now there is not default owner to avoid spam
+# * @pierre-sassoulas @DanielNoord @cdce8p @jacobtylerwalls @hippo91
 
 # Order is important. The last matching pattern has the most precedence.
 
@@ -25,7 +26,7 @@ tests/functional/u/unused/unused_private_member.py @yushao2
 
 # @areveny
 pylint/extensions/for_any_all.py @areveny
-tests/functional/ext/for_any_all/for_any_all.py @areveny
+tests/functional/ext/for_any_all/* @areveny
 pylint/extensions/private_import.py @areveny
 tests/extensions/test_private_import.py @areveny
 tests/functional/ext/private_import/private_import.py @areveny
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,31 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+* @pierre-sassoulas @DanielNoord @cdce8p @jacobtylerwalls @hippo91
+
+# Order is important. The last matching pattern has the most precedence.
+
+# @DudeNr33
+pylint/pyreverse/*    @DudeNr33
+tests/pyreverse/*    @DudeNr33
+
+# @pierre-sassoulas
+pylint/messages/*    @pierre-sassoulas
+
+# typing
+pylint/typing.py     @DanielNoord @cdce8p
+
+# multiprocessing
+pylint/lint/parallel.py @doublethefish
+tests/test_check_parallel.py @doublethefish
+
+# @yushao2
+tests/functional/u/unused/unused_private_member.py @yushao2
+
+# @areveny
+pylint/extensions/for_any_all.py @areveny
+tests/functional/ext/for_any_all/for_any_all.py @areveny
+pylint/extensions/private_import.py @areveny
+tests/extensions/test_private_import.py @areveny
+tests/functional/ext/private_import/private_import.py @areveny
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,26 +7,30 @@
 
 # Order is important. The last matching pattern has the most precedence.
 
-# @DudeNr33
-pylint/pyreverse/*    @DudeNr33
-tests/pyreverse/*    @DudeNr33
+### Core components
 
-# @pierre-sassoulas
-pylint/messages/*    @pierre-sassoulas
+# internal message handling
+pylint/message/*    @pierre-sassoulas
+tests/message/*    @pierre-sassoulas
 
 # typing
 pylint/typing.py     @DanielNoord @cdce8p
 
-# multiprocessing
+# multiprocessing (doublethefish is not yet a contributor with write access)
 # pylint/lint/parallel.py @doublethefish
 # tests/test_check_parallel.py @doublethefish
 
-# @yushao2
-tests/functional/u/unused/unused_private_member.py @yushao2
+### Pyreverse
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,31 @@
...
+
+# @pierre-sassoulas
+pylint/messages/*    @pierre-sassoulas
+
+# typing
+pylint/typing.py     @DanielNoord @cdce8p
+
+# multiprocessing
+pylint/lint/parallel.py @doublethefish
+tests/test_check_parallel.py @doublethefish
```
Pierre-Sassoulas, line 21:
```suggestion
# pylint/lint/parallel.py @doublethefish
# tests/test_check_parallel.py @doublethefish
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,31 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+* @pierre-sassoulas @DanielNoord @cdce8p @jacobtylerwalls @hippo91
```
cdce8p, line 5: I would suggest to remove the default. It isn't really helpful and will just lead to notification spam.

Codeowners are most useful if they are different from the ones already looking at all PRs anyway.

.github/CODEOWNERS
```diff
@@ -0,0 +1,31 @@
...
+
+# @DudeNr33
+pylint/pyreverse/*    @DudeNr33
+tests/pyreverse/*    @DudeNr33
+
+# @pierre-sassoulas
+pylint/messages/*    @pierre-sassoulas
+
+# typing
+pylint/typing.py     @DanielNoord @cdce8p
```
cdce8p, line 17: Instead of ordering by owner, it's better to do by category (core files, checkers, extensions). If a specific checker has tests associated with it (mostly extensions) they should also be grouped together.

.github/CODEOWNERS
```diff
@@ -0,0 +1,31 @@
...
+
+# multiprocessing
+# pylint/lint/parallel.py @doublethefish
+# tests/test_check_parallel.py @doublethefish
+
+# @yushao2
+tests/functional/u/unused/unused_private_member.py @yushao2
```
cdce8p, line 24: I don't think we should add codeowners for tests alone. That's quite useless IMO.

.github/CODEOWNERS
```diff
@@ -0,0 +1,31 @@
...
+
+# @areveny
+pylint/extensions/for_any_all.py @areveny
+tests/functional/ext/for_any_all/for_any_all.py @areveny
```
cdce8p, line 28:
```suggestion
pylint/extensions/for_any_all.py @areveny
tests/functional/ext/for_any_all/* @areveny
```
I like that pattern, especially for extensions. We should just use a wildcard for the tests directory.

cdce8p (commented):
Left a few comments. Overall I'm +0 on the idea. It certainly has some advantages, but can also add unnecessary notification spam especially for refactorings.

If you like, you can add me as codeowner for the `typing` and `code_style` extensions.
```
pylint/extensions/code_style.*  @cdce8p
tests/functional/ext/code_style/*  @cdce8p
pylint/extensions/typing.*  @cdce8p
tests/functional/ext/typing/*  @cdce8p
```

.github/CODEOWNERS
Pierre-Sassoulas:
```suggestion
# These owners will be the default owners for everything in the repo.
# Right now there is not default owner to avoid spam
# * @pierre-sassoulas @DanielNoord @cdce8p @jacobtylerwalls @hippo91
```

.github/CODEOWNERS
Pierre-Sassoulas: The issue is that the "checker" itself is really mixed inside a "stdlib" checker. I think we need to make checker more independent and hide the checkers from the users but I did not have this hindsight when we added the unused-private-member check.

.github/CODEOWNERS
```diff
@@ -0,0 +1,36 @@
...
+### Extensions
+
+# For any all
+pylint/extensions/for_any_all.py @areveny
+tests/functional/ext/for_any_all/* @areveny
+
+# Private import
+pylint/extensions/private_import.py @areveny
+tests/extensions/test_private_import.py @areveny
+tests/functional/ext/private_import/private_import.py @areveny
```
Pierre-Sassoulas, line 36:
```suggestion
tests/functional/ext/private_import/* @areveny
```

.github/CODEOWNERS
```diff
@@ -0,0 +1,36 @@
...
+# tests/test_check_parallel.py @doublethefish
+
+### Pyreverse
+pylint/pyreverse/*    @DudeNr33
+tests/pyreverse/*    @DudeNr33
+
+### Extensions
+
+# For any all
+pylint/extensions/for_any_all.py @areveny
```
DanielNoord, line 30: Are these globs or regexes? Can we use `*for_any_all*`?

Pierre-Sassoulas: It's the same format than a .gitignore with a few exceptions : https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#syntax-exceptions

DanielNoord: Yeah, so then I think **/for_any_all*/** should work!

Pierre-Sassoulas:
```suggestion
pylint/extensions/for_any_all* @areveny
```
Do you mind if we only do this right now? I'm not sure about the syntax.

DanielNoord: Why not try the desired syntax? If we can't use a full glob pattern than I think this file might quickly become unmanageable..

Pierre-Sassoulas: In ``**/for_any_all*/**`` Isn't the last ``/`` preventing matching the file ? Also how many files do we expect to have to match for one extension ? Maybe multiple functional tests and a directory instead of one file for the actual code, but do we really need more ?

DanielNoord: > In `**/for_any_all*/**` Isn't the last `/` preventing matching the file ? 

Oh yeah. Hm, do we need 2 patterns then? Or does `glob` allow for this?

> Also how many files do we expect to have to match for one extension ? Maybe multiple functional tests and a directory instead of one file for the actual code, but do we really need more ?

Probably not, but I think we should explore whether this can be created in a somewhat maintainable state. If it just becomes a list of all files we have then I don't think we should go to the effort of creating and maintaining it.

Pierre-Sassoulas: > If it just becomes a list of all files we have

I don't think we should have a code owner for all files. It's rather rare that a file is complex enough to need a code owner (and it also need to be done almost entirely by a single person, that also have write right on pylint)

DanielNoord: Yeah I mean you're likely the one who has to do most of the work on the file, so if you're okay with it I guess we could go ahead?

DanielNoord (approved):

Issue #6211, opened 2022-04-06
False negative: missing numpy param doc when "default ..." in same line of "param : type"


vepain:
### Bug description

False negative : 
Warning "missing param doc" should'nt be raised when these type of docstring format:
```python
def func(x: int = 0):
  """This is func.

  Parameter
  ---------
  x : int, default 0  # this raises the warning
    The x parameter
  """
  # ...
```

### Configuration

```ini
[MASTER]
load-plugins =
    pylint.extensions.docstyle,
    pylint.extensions.docparams

[MESSAGES CONTROL]
enable = W

[BASIC]
accept-no-param-doc = no
```

### Command used

```shell
pylint --rcfile=tox.ini a.py
```

### Pylint output

```shell
"orientation" missing in parameter documentation
```

### Expected behavior

No warning

### Pylint version

```shell
pylint 2.13.5
astroid 2.11.2
Python 3.9.12 (main, Mar 24 2022, 16:20:11) 
[GCC 9.4.0]
```

### OS / Environment

Ubuntu 20.04

lkev:
+1 having the same issue.

Just to add - the following does not raise any error, so maybe just allow items in parenthesis and/or the "default" word after the comma?

```python
def func(x: int = 0):
  """This is func.

  Parameter
  ---------
  x : int, optional
    The x parameter, by default 0
  """
  # ...
```

DanielNoord:
See https://numpydoc.readthedocs.io/en/latest/format.html#parameters and the failures in https://github.com/PyCQA/pylint/pull/7149.

adam-grant-hendry:
I am seeing this as well with

```
python 3.8.10
pylint 2.14.5
```

using

```pylint.extensions.docparams```

@DanielNoord Has this not been fixed yet?

DanielNoord:
No, I don't think so. The linked PR fixed something different.

adam-grant-hendry:
@DanielNoord Can you point me to the section of code to review? I can submit a PR if you like.

adam-grant-hendry:
> @DanielNoord Can you point me to the section of code to review? I can submit a PR if you like.

Actually, I'm assuming this is [pylint/extensions/docparams.py](https://github.com/PyCQA/pylint/blob/main/pylint/extensions/docparams.py). I'll submit a PR.

Pierre-Sassoulas:
[...]

PR #7360, opened 2022-08-26
fix(numpy-param-doc): allow "default" in params

adam-grant-hendry:

## Type of Changes

## Description

NumPy doc style guide permits using `default` in parameter signatures.

Closes ##6211


adam-grant-hendry:
@DanielNoord @Pierre-Sassoulas Would you please kindly help review? Thank you.

tests/functional/ext/docparams/docparams.py
```diff
@@ -92,3 +93,40 @@ def params_are_documented(par1: int, *, par2: int) -> int:
...
+
+    See https://github.com/PyCQA/pylint/issues/6211
+
+    Parameter
+    ---------
+    x : int, default 0
+        The x parameter
+    """
+
+    print(x)
```
Pierre-Sassoulas, line 132:
```suggestion
def regression_6211(number: int = 0) -> None:
    """This is a regression test for issue #6211.

    False negative of "missing param doc" was being issued when "default" used in
    NumPy-style docs. This test should return no errors.

    See https://github.com/PyCQA/pylint/issues/6211

    Parameter
    ---------
    number : int, default 0
        The number parameter
    """

    print(number)
```

Pierre-Sassoulas (changes requested):
Could you add a fragment for the changelog please ? :)

DanielNoord:
@adam-grant-hendry The primer tests shows that we would be emitting a lot more messages which I don't really expect. Perhaps we can take a test from there?

tests/functional/ext/docparams/docparams.py
adam-grant-hendry: Good catch. This was causing an `invalid-name` error. I will fix.

adam-grant-hendry:
> @adam-grant-hendry The primer tests shows that we would be emitting a lot more messages which I don't really expect. Perhaps we can take a test from there?

One difficulty I have is that `pytest` isn't outputting what the regular expressions are matching and not matching. Is there a way I can do this with the `pylint` source, or do I have to write a throw-away script using `re` and check for matches?

tests/functional/ext/docparams/docparams.py
Pierre-Sassoulas: You can commit the suggestion directly in interface :)

adam-grant-hendry:
> Could you add a fragment for the changelog please ? :)

Yes, I added a fragment with my last push.

Pierre-Sassoulas (commented):
Would you mind rebasing/merging the main branch ? (there's modification from the recent release mixed in the diff.)

Without looking into it, I think something is wrong with the modified regex as the primer raise a lot more message now.

adam-grant-hendry:
> Would you mind rebasing/merging the main branch ? (there's modification from the recent release mixed in the diff.)

Just did, but couldn't figure out how to do that without force pushing. I used `--force-with-lease` to try not to overwrite work you or @DanielNoord may have done.
 
> Without looking into it, I think something is wrong with the modified regex as the primer raise a lot more message now.

Yes, the regex is not working, but I don't know why.

adam-grant-hendry:
@DanielNoord @Pierre-Sassoulas This should be working now. I wrote a test script to help me out:

<details>
<summary><b>test_script.py</b></summary>

```python

import re
import typing

from pylint.extensions._check_docs_utils import NumpyDocstring

if typing.TYPE_CHECKING:
    from pylint.extensions._check_docs_utils import Docstring


def space_indentation(s: str) -> int:
    """The number of leading spaces in a string.

    :param str s: input string

    :rtype: int
    :return: number of leading spaces
    """
    return len(s) - len(s.lstrip(" "))


def _is_section_header(line: str) -> bool:
    return bool(re.match(r"\s*-+$", line))


def min_section_indent(section_match: re.Match[str]) -> int:
    return len(section_match.group(1))


def _parse_section(
    section_re: re.Pattern[str],
    doc: str,
) -> list[str]:
    section_match = section_re.search(doc)
    if section_match is None:
        return []

    min_indentation = min_section_indent(section_match)

    entries: list[str] = []
    entry: list[str] = []
    is_first = True
    for line in section_match.group(2).splitlines():
        if not line.strip():
            continue
        indentation = space_indentation(line)
        if indentation < min_indentation:
            break

        # The first line after the header defines the minimum
        # indentation.
        if is_first:
            min_indentation = indentation
            is_first = False

        if indentation == min_indentation:
            if _is_section_header(line):
                break
            # Lines with minimum indentation must contain the beginning
            # of a new parameter documentation.
            if entry:
                entries.append("\n".join(entry))
                entry = []

        entry.append(line)

    if entry:
        entries.append("\n".join(entry))

    return entries


def match_param_docs(
    checker: NumpyDocstring,
    doc: str,
) -> tuple[set[str], set[str]]:
    """Matches parameter documentation section to parameter documentation rules."""
    params_with_doc = set()
    params_with_type = set()

    entries = _parse_section(checker.re_param_section, doc)
    entries.extend(_parse_section(checker.re_keyword_param_section, doc))
    for entry in entries:
        match = checker.re_param_line.match(entry)
        if not match:
            continue

        print(match)
        print()
        # check if parameter has description only
        re_only_desc = re.match(r"\s*(\*{0,2}\w+)\s*:?\n\s*\w*$", entry)
        print(f're_only_desc: {re_only_desc}')
        if re_only_desc:
            param_name = match.group('param_name')
            param_desc = match.group('param_type')
            param_type = None
        else:
            param_name = match.group('param_name')
            param_type = match.group('param_type')
            param_desc = match.group('param_desc')

        print(f"Param Name: '{param_name}'")
        print(f"Type: '{param_type}'")
        print(f"Description: '{param_desc}'")

        if param_type:
            params_with_type.add(param_name)

        if param_desc:
            params_with_doc.add(param_name)

    return params_with_doc, params_with_type


if __name__ == '__main__':
    docstring_1 = """This is a regression test for issue `#6211`_.

False negative of "missing param doc" was being issued when "default" used in
NumPy-style docs. This test should return no errors.

Parameters
----------
number : int, default 0
    The number parameter

.. _`#6211`:
    https://github.com/PyCQA/pylint/issues/6211
"""

    docstring_2 = """No errors raised when pipe symbol used for or.

`PEP 604`_ allows writing Union types as X | Y. Can be enabled in Python <3.10
using `from __future__ import annotations`.

Parameters
----------
arg1
    The first arg
arg2 : str | None, default=None
    The second arg

.. _`PEP 604`:
    https://peps.python.org/pep-0604/
"""

    docstring_3 = """Reported in https://github.com/PyCQA/pylint/issues/4035.

Parameters
----------
arg : int
    The number of times to print it.
option : {"y", "n"}
    Do I do it?
option2 : {"y", None, "n"}
    Do I do it?

"""

    docstring_4 = """function foobar ...

Parameters
----------
arg1 : int
arg2 : int
arg3 : str
"""

    docstring_5 = """function foobar ...
Parameters
----------
arg1:
    description
arg2: str
"""

    np_checker = NumpyDocstring(None)

    docstrings = [
        docstring_1,
        docstring_2,
        docstring_3,
        docstring_4,
        docstring_5,
    ]

    for ndx, docstr in enumerate(docstrings):
        print(f'========= Docstring {ndx+1} =========')
        params_with_description_only, params_with_a_type = match_param_docs(np_checker, docstr)

        print()
        print(f'Parameters with a description only: {params_with_description_only}')
        print(f'Parameters with a type: {params_with_a_type}')
        print()

```
</details>

A couple notes:

1. The OP to Issue #6211 used "Parameter" (singular) instead of the intended "Parameters" (plural), and as a result this regex fails to find the parameters section:

```python
re_param_section = re.compile(
        _re_section_template.format(r"(?:Args|Arguments|Parameters)"),
        re.X | re.S | re.M,
    )
```

2. The hyperlink I put in the body of the `regression_6211` docstring was causing errors (I suspect because of the colon after the "https"). I turned this into a directive.

3. Adding the `from __future__ import annotations` line to the top of `docparams.py` (so I can use python `3.10` style type annotations) shifted each function down by one line, so I needed to update the `docparams.txt` file.

4. The regular expression for default values only accounts for strings, booleans, and `None`, but not:

 - numbers (ints (`2`), floats (`2.35`), binary (`0b010`), oct (`0o007`), or hex (`0xDEADBEEF`)
 - those who use `_` as separators between sets of 3 digits (e.g. `10_000_000`)
 -  byte strings (i.e. `b'this will come to byte you')
 - callables (i.e. `my_funct=jerry`)

There may be other literal/default types to add in the future, but for now I accounted just for ints because that solves the issue at hand:

```python
re_default_value = r"""((['"]\w+\s*['"])|(\d+)|(True)|(False)|(None))"""
```

adam-grant-hendry:
> @adam-grant-hendry The primer tests shows that we would be emitting a lot more messages which I don't really expect. Perhaps we can take a test from there?

@DanielNoord @Pierre-Sassoulas After using my script, I found the original code was matching colon (`:`) for `type` instead of the
actual type. Most of these errors are accurate:

    - 1: `pandas/compat/pickle_compat.py`#L256
	- 6: `pandas/tests/extension/conftest.py`#L47

have the form

```
param : description instead of type
```

This is incorrect per the [NumPy Style Guide](https://numpydoc.readthedocs.io/en/latest/format.html#parameters). Descriptions must go
on the next line indented. Only types can occur after the colon on the same line.

    - 7, 8: `pandas/tests/io/test_parquet.py#L165`
	
is in fact missing documentation for the listed parameters.

    - 9, 13: `pandas/tests/io/formats/test_to_html.py#L30` (there are many others)

has incorrect types. `"pytest fixture"` and `matplotlib Axes object` are not types; these should be a class name (dotted attribute), type alias, or builtin (e.g. `_pytest.monkeypatch.MonkeyPatch` or `matplotlib.axes`)


If `pandas` wishes to use a style different from `numpy`, they should have their own style guide. It is confusing for those wanting to follow
the `numpy` guide to get unexpected allowed behavior (and it might cause `sphinx` not to work properly). For this, I recommend the `pylint`
team create a separate class in `_check_docs_utils.py` and call it `PandasDocstring` and to have the `pandas` team confirm it.

Numbers 2-# (from `pandas/tests/arithmetic/common.py`) have method formal parameters that are not type hinted, and hence should technically be
`Any`.

DanielNoord:
@adam-grant-hendry That does indeed seem like we were incorrectly matching previously.

Would you be willing to open a separate PR to fix that and add some regressions tests against that?

adam-grant-hendry:
> @adam-grant-hendry That does indeed seem like we were incorrectly matching previously.
> 
> Would you be willing to open a separate PR to fix that and add some regressions tests against that?

I would, but I'm not sure I have time at this stage. I doubt I would be able to finish this PR even. The amount of fixes required is more than I originally envisioned, I'm sorry.

At this stage, I'm simply getting by by:

1. Not using `default` in the type, but in the description (e.g. `Defaults to ...`)
2. Using `or` and `of` instead of `|` and `[]`, respectively (though this is actually per the `numpydoc style guide` and the latter is from the newer python typing annotations syntax)

adam-grant-hendry:
@DanielNoord @Pierre-Sassoulas Two more drive-by comments:

1. Because `:` is searched for in the regex, this also causes false-negatives when directives are used in the docstring body. Any colons in the body is going to be an issue.

e.g. This gives a `missing-param-doc` error

```python
def func(self, a: int, b: bool) -> None:
    """Does something funky.

    This is where you explain the funk.

    ..warning:: `a` Explodes!
        Don't use `a`

    Parameters
    ------------
    a : int
        An integer
    b : bool
        A bool
```

2. Your regex's don't match when a union of types ends with `, or type`. It matches `type of type` and `type or type`, but the following does not work:

```python
def funkify(self, a: int | bool | str) -> str:
    """Get with the funk.

    What the funk?!

    Parameters
    ------------
    a : int, bool, or str
        This is an ``int``, ``bool``, or ``str``. How funky!

    Returns
    --------
    b : str
        Because why not return a ``str``
```

DanielNoord:
> @DanielNoord @Pierre-Sassoulas Two more drive-by comments:

Both of these should probably be their own issue, right?

I have updated this PR after the merge of #7398. Let's see what the primer thinks!

DanielNoord:
@Pierre-Sassoulas We have a similar issue as discussed previously here.

Should we just accept anything that comes on the `type` line in a `Numpy` docstring? It feels better to do so than to try and determine whether `type or None` or `a description of a type` is a "real type".

Pierre-Sassoulas (commented):
Sorry for the delay I did not realize this was waiting for my input. I think we should accept anything from the type field, because being able to tell valid type from text is hard, and doing something like "None or MyClass" is frequent enough that the primers message was truncated because there's so much new message. (Removing from the 3.0 milestone because this is the kind of thing that can be merged later)

DanielNoord (approved):
Let's merge this :)

Issue #3976, opened 2020-12-12
unexpected-keyword-arg for typing_extensions.get_type_hints


sk-:
### Steps to reproduce
```py
import typing_extensions

def function():
    pass

typing_extensions.get_type_hints(function, include_extras=True)
```

### Current behavior
```
test.py:6:0: E1123: Unexpected keyword argument 'include_extras' in function call (unexpected-keyword-arg)
```
### Expected behavior
No error is reported as the code executes just fine.

### pylint --version output
````
pylint 2.6.0
astroid 2.4.2
Python 3.7.7 (default, May 10 2020, 19:54:23)
[Clang 11.0.3 (clang-1103.0.32.59)]
````

**Comment by brettplarson** [comment]
I am seeing the exact same issue but with an internal module we are using.
```
pylint 2.6.0
astroid 2.4.2
Python 3.8.6 (default, Sep 25 2020, 09:36:53) 
[GCC 10.2.0]
```

mbyrnepr2:
It looks like this one has been fixed somewhere along the way.
I can reproduce on 2.6.0 but not on the latest Pylint version (2.15.0).

Happy to re-open if any issues.

PR #7364, opened 2022-08-26
Add a functional regression test.

mbyrnepr2:
## Description
Add a functional regression test. The original issue is no longer reproducible on the latest Pylint (2.15.0).

Refs #3976


Pierre-Sassoulas (approved):
👍

Issue #7362, opened 2022-08-26
Better release notes generation and automated release process


Pierre-Sassoulas:
### Current problem

- contributor-txt pollutes the contributor file with a header because it does not detect the existing header
- The release notes are generated as rst but github want markdown
- Only the first link is generated when there is two or more issues referenced in a fragment

### Desired solution

Markdown creation so it can be used in both context, other issue automatically fixed

### Additional context

Noticed during the release of 2.15.0.

DudeNr33:
Markdown is not really directly supported by `towncrier`. 
But people have found [hacks](https://github.com/twisted/towncrier/issues/128) for a workaround.
Alternatively, we could use `pandoc` or a similar tool to convert it. 

For the link generation: This will also require some scripting. Currently we use the `issue_format` option:
https://github.com/PyCQA/pylint/blob/24eaec26f701cf8adddf27ffa8859e0bb5074a89/towncrier.toml#L6
The `{issue}` placeholder is simply the first part of the message fragment.
Maybe some pre-commit hook to convert simple `#1234` into actual GitHub links would be better?

Pierre-Sassoulas:
For github we could do with only  "#{issue}". I was thinking of using .md in order to not have to generate another document but this is done very unfrequentely so launching a script to generate github's text is reasonable.

DudeNr33:
Sorry for the delay. I looked into the issue with the duplicated labels in the release notes.

The problem comes from the `sphinx.ext.autosectionlabel` extension. Comparing our generated releasenotes with other projects like [`pytest`](https://github.com/pytest-dev/pytest/blob/main/doc/en/changelog.rst) and [`twisted`](https://github.com/twisted/twisted/blob/trunk/NEWS.rst) that use `towncrier`, it is normal to have duplicated titles in the resulting document. 

The options we have to resolve this are:

* Add `autosectionlabel_maxdepth = 2` to `doc/conf.py`, which limits the autogeneration of labels to the first two levels (see [docs](https://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html#confval-autosectionlabel_maxdepth))

OR

* Use `single_file=false` in `towncrier.toml`, which would generate separate files for each release (see [docs](https://github.com/twisted/towncrier#further-options)).

If the first option is OK for you, I would prefer that one.

Pierre-Sassoulas:
[...]

PR #7369, opened 2022-08-27
Fix 2.15 changelog

cdce8p:
## Description
Just checked the `2.15` release notes and noticed that my change wasn't included. I guess it's important to use an underscore in the suffix 🤷🏻‍♂️

Refs #7322


Pierre-Sassoulas (approved):
Nice catch. Something to add to #7362

PR #7370, opened 2022-08-27
Fix `used-before-assignment` for functions/classes defined in type checking guard

### LINKED ISSUE #7368: `E0601: used-before-assignment` false positive for import under `TYPE_CHECKING` check
### Bug description

```python
# pylint: disable=missing-module-docstring, missing-function-docstring

if TYPE_CHECKING:
    from typing_extensions import assert_type
else:

    def assert_type(val, _):
        return val

ANSWER = 42
assert_type(ANSWER, int)
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:13:0: E0601: Using variable 'assert_type' before assignment (used-before-assignment)

Your code has been rated at 2.86/10 (previous run: 2.86/10, +0.00)
```

### Expected behavior

This is a false positive, as `assert_type` is defined in both branches of the `if`.

### Pylint version

```shell
pylint 2.15.0
astroid 2.12.4
Python 3.10.6 (main, Aug  3 2022, 17:39:45) [GCC 12.1.1 20220730]
```

### OS / Environment

Archlinux

### Additional dependencies

astroid==2.12.4
dill==0.3.5.1
isort==5.10.1
lazy-object-proxy==1.7.1
mccabe==0.7.0
platformdirs==2.5.2
pylint==2.15.0
tomli==2.0.1
tomlkit==0.11.4
wrapt==1.14.1

Rogdham:
## Type of Changes

## Description

Closes #7368

This PR fixes the following false-positive:

```python

if TYPE_CHECKING:
    from typing_extensions import assert_type
else:

    def assert_type(val, _):
        return val

ANSWER = 42
assert_type(ANSWER, int)  # E0601: Using variable 'assert_type' before assignment (used-before-assignment)
```

The issue comes from definitions in the `else` block of a `if` type-guard.

In 3159b1744caee14068b4d875bb5adaa5ffbebd44 the case was already handled for assignments, I had only added cases for function and class that are defined in the `else` part.

### DIFF
--- a/doc/whatsnew/fragments/7368.false_positive
+++ b/doc/whatsnew/fragments/7368.false_positive
@@ -1,0 +1,3 @@
+Fix ``used-before-assignment`` for functions/classes defined in type checking guard.
+
+Closes #7368
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1841,7 +1841,6 @@ class VariablesChecker(BaseChecker):
         base_scope_type,
         is_recursive_klass,
     ) -> tuple[bool, bool, bool]:
-        # pylint: disable=too-many-nested-blocks
         maybe_before_assign = True
         annotation_return = False
         use_outer_definition = False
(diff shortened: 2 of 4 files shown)


jacobtylerwalls (approved):
Looks great. Congrats on becoming a pylint contributor!

PR #7373, opened 2022-08-28
Added documentation examples for `bad-format-string-key`

DudeNr33:
## Type of Changes

## Description
Documentation examples for `bad-format-string-key`.

Refs #5953

### DIFF
--- a/doc/data/messages/b/bad-format-string-key/bad.py
+++ b/doc/data/messages/b/bad-format-string-key/bad.py
@@ -1,0 +1,1 @@
+print("%(one)d" % {"one": 1, 2: 2})  # [bad-format-string-key]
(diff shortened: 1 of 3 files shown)


doc/data/messages/b/bad-format-string-key/bad.py
```diff
@@ -0,0 +1 @@
+print("%(one)d" % {"one": 1, 2: 2})  # [bad-format-string-key]
```
DudeNr33, line 1: `print("%(one)d %(two)d" % {"one": 1, 2: 2})` would additionally trigger `missing-format-string-key`.

Pierre-Sassoulas (approved):
👌

PR #7378, opened 2022-08-29
Fix following "Bump astroid from 2.12.4 to 2.12.5"

Pierre-Sassoulas:
Reverts PyCQA/pylint#7377

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   # Also change CACHE_VERSION in the other workflows
-  CACHE_VERSION: 22
+  CACHE_VERSION: 23
   DEFAULT_PYTHON: "3.10"
 
 jobs:
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -8,7 +8,7 @@ on:
   pull_request: ~
 
 env:
-  CACHE_VERSION: 22
+  CACHE_VERSION: 23
   DEFAULT_PYTHON: "3.10"
   PRE_COMMIT_CACHE: ~/.cache/pre-commit
 
--- a/.github/workflows/primer-test.yaml
+++ b/.github/workflows/primer-test.yaml
@@ -13,7 +13,7 @@ on:
       - ".github/workflows/primer-test.yaml"
 
 env:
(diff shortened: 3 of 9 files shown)


DanielNoord:
Why don't we simply fix the issue instead of reverting it? We only need to bump the upper bound in `requirements`, there is no need to bump anything else.

Pierre-Sassoulas:
> there is no need to bump anything else.

I think we still need to bump de cache, right ? It's not really a revert I've opened the issue with a revert (for speed), but I'm intending to keep the astroid upgrade, and fix the issue with the documentation that was highlighted by the MR that broke main (ie. implement this comment: https://github.com/PyCQA/pylint/pull/7377#issuecomment-1230332382).

DanielNoord:
> > there is no need to bump anything else.
> 
> I think we still need to bump de cache, right ?

Don't think so. If the venv creation failed it will auto create a new venv.

> It's not really a revert I've opened the issue with a revert (for speed), but I'm intending to keep the astroid upgrade

The `astroid` upgrade is not correct though. It doesn't download the correct version.

> and fix the issue with the documentation that was highlighted by the MR that broke main (ie. implement this comment: [#7377 (comment)](https://github.com/PyCQA/pylint/pull/7377#issuecomment-1230332382)).

👍

Pierre-Sassoulas:
Don't we want this version: https://pypi.org/project/astroid/2.12.5/ ? I suppose the primer are expected to fail, right ? 
The error we get on main is:
```
pip._vendor.pkg_resources.ContextualVersionConflict: (astroid 2.12.4 (/home/runner/work/pylint
/pylint/venv/lib/python3.10/site-packages), Requirement.parse('astroid<=2.14.0-dev0,>=2.12.5'), {'pylint'})
```
Which makes me think that astroid 2.12.4 is pre-installed in the cache instead ?

DanielNoord (approved):

Pierre-Sassoulas:
Yeah sorry I don't have a lot of time right now, still planning on fixing the issue.

DanielNoord:
No worries, just thought I would unblock this.

Oh and sorry, I completely missed the `blocked` label. Should have pinged you before merging.

Pierre-Sassoulas:
No worries simply fixing main makes a lot more sense !

PR #7386, opened 2022-08-31
Fix crash while iteraring over a class attribute

### LINKED ISSUE #7380: AttributeError: 'Attribute' object has no attribute 'name'
### Bug description

Linting:
```python
lst = []
for _var in lst:
    del _var
```

pylint crashed with a ``AstroidError`` and with the following stacktrace:
```
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/lint/pylinter.py", line 769, in _lint_file
    check_astroid_module(module)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/lint/pylinter.py", line 1029, in check_astroid_module
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/lint/pylinter.py", line 1079, in _check_astroid_module
    walker.walk(node)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/utils/ast_walker.py", line 93, in walk
    self.walk(child)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/utils/ast_walker.py", line 93, in walk
    self.walk(child)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/utils/ast_walker.py", line 93, in walk
    self.walk(child)
  [Previous line repeated 1 more time]
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/utils/ast_walker.py", line 90, in walk
    callback(astroid)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/checkers/modified_iterating_checker.py", line 60, in visit_for
    self._modified_iterating_check_on_node_and_children(body_node, iter_obj)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/checkers/modified_iterating_checker.py", line 66, in _modified_iterating_check_on_node_and_children
    self._modified_iterating_check(body_node, iter_obj)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/checkers/modified_iterating_checker.py", line 96, in _modified_iterating_check
    args=(iter_obj.name,),
AttributeError: 'Attribute' object has no attribute 'name'

The above exception was the direct cause of the following exception:

  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/lint/pylinter.py", line 734, in _lint_files
    self._lint_file(fileitem, module, check_astroid_module)
  File "/opt/hostedtoolcache/Python/3.8.13/x64/lib/python3.8/site-packages/pylint/lint/pylinter.py", line 771, in _lint_file
    raise astroid.AstroidError from e
astroid.exceptions.AstroidError
```

### Expected behavior

No crash.

### Pylint version

```shell
pylint-2.15.0
astroid-2.12.5
CPython 3.8.13
```

### OS / Environment

ubuntu-latest

DanielNoord:

## Type of Changes

## Description

Closes #7380

@orSolocate I haven't created an issue for this but:

```python
class MyClass:
    def __init__(self) -> None:
        self.attribute = [1, 2, 3]

   def my_other_method(self):
        """This should raise as we are appending."""
        for var in self.attribute:
            self.attribute.append(var)  # [modified-iterating-list]
```

Currently doesn't raise like it should.

This would require quite a bit of refactoring in the current checker as it isn't really suited for checking class attributes. I'm not sure and there is no obligation, but perhaps you find this interesting and want to look into it?

### DIFF
--- a/doc/whatsnew/fragments/7380.bugfix
+++ b/doc/whatsnew/fragments/7380.bugfix
@@ -1,3 +1,3 @@
-Fixed a crash with ``for`` loops that iterate over class attributes.
+Update ``modified_iterating`` checker to fix a crash with ``for`` loops on empty list
 
 Closes #7380
--- a/doc/whatsnew/fragments/7380.bugfix
+++ b/doc/whatsnew/fragments/7380.bugfix
@@ -1,3 +1,3 @@
-Update ``modified_iterating`` checker to fix a crash with ``for`` loops on empty list
+Update ``modified_iterating`` checker to fix a crash with ``for`` loops on empty list.
 
 Closes #7380
--- a/doc/whatsnew/fragments/7380.bugfix
+++ b/doc/whatsnew/fragments/7380.bugfix
@@ -1,0 +1,3 @@
+Fixed a crash with ``for`` loops that iterate over class attributes.
+
+Closes #7380
--- a/pylint/checkers/modified_iterating_checker.py
+++ b/pylint/checkers/modified_iterating_checker.py
@@ -81,7 +81,7 @@ class ModifiedIterationChecker(checkers.BaseChecker):
                 msg_id = "modified-iterating-dict"
             elif isinstance(inferred, nodes.Set):
                 msg_id = "modified-iterating-set"
-        elif not isinstance(iter_obj, nodes.Name):
+        elif not isinstance(iter_obj, (nodes.Name, nodes.Attribute)):
             pass
         elif self._modified_iterating_list_cond(node, iter_obj):
             msg_id = "modified-iterating-list"
@@ -90,10 +90,14 @@ class ModifiedIterationChecker(checkers.BaseChecker):
         elif self._modified_iterating_set_cond(node, iter_obj):
             msg_id = "modified-iterating-set"
         if msg_id:
+            if isinstance(iter_obj, nodes.Attribute):
+                obj_name = iter_obj.attrname
+            else:
+                obj_name = iter_obj.name
             self.add_message(
                 msg_id,
                 node=node,
-                args=(iter_obj.name,),
+                args=(obj_name,),
                 confidence=interfaces.INFERENCE,
             )
 
--- a/tests/functional/m/modified_iterating.py
+++ b/tests/functional/m/modified_iterating.py
@@ -26,7 +26,7 @@ my_dict = {"1": 1, "2": 2, "3": 3}
 i = 1
 for item in my_dict:
     item_list[0] = i  # for coverage, see reference at /pull/5628#discussion_r792181642
-    my_dict[i] = 1      # [modified-iterating-dict]
+    my_dict[i] = 1  # [modified-iterating-dict]
     i += 1
 
 i = 1
@@ -93,3 +93,15 @@ def update_existing_key():
     for key in my_dict:
         new_key = key.lower()
         my_dict[new_key] = 1  # [modified-iterating-dict]
+
+
+class MyClass:  # pylint: disable=too-few-public-methods
+    """Regression test for https://github.com/PyCQA/pylint/issues/7380"""
+
+    def __init__(self) -> None:
+        self.attribute = [1, 2, 3]
+
+    def my_method(self):
+        """This should raise as we are deleting."""
+        for var in self.attribute:
+            del var  # [modified-iterating-list]
--- a/tests/functional/m/modified_iterating.py
+++ b/tests/functional/m/modified_iterating.py
@@ -1,5 +1,5 @@
 """Tests for iterating-modified messages"""
-# pylint: disable=not-callable,unnecessary-comprehension
+# pylint: disable=not-callable,unnecessary-comprehension,too-few-public-methods
 
 import copy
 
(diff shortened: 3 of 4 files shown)


tests/functional/m/modified_iterating.py
```diff
@@ -93,3 +93,15 @@ def update_existing_key():
     for key in my_dict:
         new_key = key.lower()
         my_dict[new_key] = 1  # [modified-iterating-dict]
+
+
+class MyClass:  # pylint: disable=too-few-public-methods
```
orSolocate, line 98: This is the only class in the test file. you can add the disable flag at the top of the file

tests/functional/m/modified_iterating.py
```diff
@@ -93,3 +93,15 @@ def update_existing_key():
...
+    """Regression test for https://github.com/PyCQA/pylint/issues/7380"""
+
+    def __init__(self) -> None:
+        self.attribute = [1, 2, 3]
```
orSolocate, line 102:
```suggestion
        self.iterated_attributes = [1, 2, 3]
```

tests/functional/m/modified_iterating.py
```diff
@@ -93,3 +93,15 @@ def update_existing_key():
...
+
+    def my_method(self):
+        """This should raise as we are deleting."""
+        for var in self.attribute:
+            del var  # [modified-iterating-list]
```
orSolocate, line 107:
```suggestion
    def modifying_list_(self):
    # should be fine to modify a copy
            for var in self.iterated_attributes.copy():
            del var  # [modified-iterating-list]
     # should raise a message when modifying
        for var in self.iterated_attributes:
            del var  # [modified-iterating-list]

```

Also check the false positive case. Fix the `.txt` file accordingly.

pylint/checkers/modified_iterating_checker.py
```diff
@@ -81,7 +81,7 @@ def _modified_iterating_check(
                 msg_id = "modified-iterating-dict"
             elif isinstance(inferred, nodes.Set):
                 msg_id = "modified-iterating-set"
-        elif not isinstance(iter_obj, nodes.Name):
+        elif not isinstance(iter_obj, (nodes.Name, nodes.Attribute)):
```
orSolocate, line 84: should be worth to think more about the `Attribute` case, should it raise a message in all cases?
maybe there is a false positive here?

doc/whatsnew/fragments/7380.bugfix
```diff
@@ -0,0 +1,3 @@
+Fixed a crash with ``for`` loops that iterate over class attributes.
```
orSolocate, line 1:
```suggestion
Update ``modified_iterating`` checker to fix a crash with ``for`` loops that iterate over class attributes (dicts/lists/sets)
```

orSolocate (changes requested):
@DanielNoord  hey man consider these little improvements for this little addition to my checker.
Are you sure an `Attribute` will always be something to check here and not a false-positive? I trust your Python knowledge here.

Pierre-Sassoulas (approved):
LGTM, let's merge once the discussion with @orSolocate is resolved :)

tests/functional/m/modified_iterating.py
DanielNoord: This is personal preference I think. I like disabling at the occurrence as it allows more freedom in checking for some message further down the file.

But I'll change it here 😄

tests/functional/m/modified_iterating.py
orSolocate: @DanielNoord  yeah i re-read it and understood that the case which is not working is the class case. 
I don't fully understand the `append` example, how can you put `list.append` in a for loop?

DanielNoord:
```python
for x in a_list:
    a_list.append(x)
```

The issue here is that `append` is an `Attribute` node on `a_list` just as `attribute` is an `Attribute` node in the following example:
```python
class MyClass:
    def __init__():
        self.attribute = [1,2,3]

    def method(self):
        for x in self.attribute:
            del x
```

`Astroid` only recognises that they are nodes that come after a period, not that they are attributes in the traditional sense of the word.
This complicates the logic we're currently using in the checker. I tried to fix it, but it turned out to be a pretty large refactor which probably should be taken in multiple steps for ease of reviewing.
To make sure we no longer crash I wanted to push this fix at least.

orSolocate:
@Pierre-Sassoulas  you guys can merge it. You can open a new issue to add support for class attributes with my checker and assign me :)

Pierre-Sassoulas (approved):

tests/functional/m/modified_iterating.py
orSolocate: You mean like when iterating through a dictionary `d` :  iterating on  `d.keys` is an  `Attribute` which is actually a fix the our error of iterating through `d`? something like that? so your current solution will throw false-postive now?

DanielNoord: I think you'll understand what you mean when you add the example I gave in the opening post to the functional tests. If you run `pytest` then you'll see that some of the `isinstance(node, nodes.Attribute)` test you are using to pick up `d.append` will now also pick up `MyClass.attribute`. This causes some issues and methods such as `_modified_iterating_list_cond` will need to be changed to allow checking both type of nodes.

PR #7393, opened 2022-09-01
Release branch 2.15.1

Pierre-Sassoulas:


towncrier.toml
```diff
@@ -1,7 +1,7 @@
 [tool.towncrier]
-version = "2.16.0"
+version = "2.15.1"
 directory = "doc/whatsnew/fragments"
-filename = "doc/whatsnew/2/2.16/index.rst"
+filename = "doc/whatsnew/2/2.15/index.rst"
```
Pierre-Sassoulas, line 4: This is going to NOT be a merge conflict and we need to be careful when merging in main.

DanielNoord: I was wondering about this. Do we need to bump to the dev version on `main`? What's the benefit? Can't we just bump when we need to for a release?

Pierre-Sassoulas: Well it's useful to know the pylint version we're talking about (2.16.0-dev0 vs 2.15.1). scm_setuptool was nice for this but created a dependencie we did not want...

DanielNoord: Yeah I mean I don't really mind, but referring to `main` would also work for me.
It reduces the amount of work necessary for a release so I thought it might be something to consider.

DanielNoord (changes requested):
Release notes seem to be broken...

Pierre-Sassoulas:
> Release notes seem to be broken...

Yes, I was wondering what to do. Maybe use "False positive for 2.15.1" as title ? Maybe remove the titles altogether... ?

Pierre-Sassoulas:
We have titles that are duplicated when we generate the patch release note:
```
/home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst:63: WARNING: duplicate label whatsnew/2/2.15/index:false positives fixed, other instance in /home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst
/home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst:132: WARNING: duplicate label whatsnew/2/2.15/index:other bug fixes, other instance in /home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst
```

@DudeNr33 do you have an opinion about the easy way to fix this ?

towncrier.toml
Pierre-Sassoulas: referring to main ?

DanielNoord: As in, I never use `2.16.0-dev0` directly in any of my messages. I often just say `main` or something like that. There is not much benefit for me to having the bumped version, but that might just be me.

DanielNoord:
> We have titles that are duplicated when we generate the patch release note:
> 
> ```
> /home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst:63: WARNING: duplicate label whatsnew/2/2.15/index:false positives fixed, other instance in /home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst
> /home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst:132: WARNING: duplicate label whatsnew/2/2.15/index:other bug fixes, other instance in /home/runner/work/pylint/pylint/doc/whatsnew/2/2.15/index.rst
> ```
> 
> @DudeNr33 do you have an opinion about the easy way to fix this ?

Can we fix this by decreasing the title level? I haven't tested this, but can we have duplicate titles if they fall within different top level titles?

DudeNr33:
I'm currently on vacation and probably won't have time to try around myself until next Monday, sorry!
The underlines that are used by `towncrier` can be configured via the `underlines` option (see the [readme](https://github.com/twisted/towncrier#further-options)).

I am a bit surprised that this is an issue, as I would expect others to experience this as well, but I did not find any issue about this in their issue tracker.
Is this maybe just because we include a `:toctree:`?
In this case decreasing the title level could work, but then again the toctree would be useless anyway and we could also just remove that instead of changing the title level.

Pierre-Sassoulas:
I'm going to fix the release note manually to be able to release until we discuss what we should do. Not sure what to do with https://github.com/PyCQA/pylint/pull/7398, it's in 2.16.0 but it's labelled "need backport", what do you think @DanielNoord ?

DanielNoord:
I thought it would be good to backport as there were other PRs that target that code.

If there are regression it would be helpful to have the refactor available on the maintenance branch as well.

Pierre-Sassoulas:
Alright, I moved it to 2.15.1 and cherry-picked it :) I think we can release now

DanielNoord (dismissed):

doc/whatsnew/fragments/7368.false_positive
```diff
@@ -0,0 +1,3 @@
+Fix ``used-before-assignment`` for functions/classes defined in type checking guard.
```
DanielNoord, line 1: @Pierre-Sassoulas 

These are incorrect and won't and up in the changelog I think

DanielNoord:
btw, for future reviews I think we should split up the back port and release PRs. Stuff like the broken `towncrier` files are difficult to spot in larger PRs like this.

For now it is fine though!  😄

Pierre-Sassoulas:
Forgot to regenerate the towncrier file after adding 3 more cherry-pick, my bad.

doc/whatsnew/2/2.15/index.rst
```diff
@@ -29,6 +29,50 @@ Marc Byrne became a maintainer, welcome to the team !
...
+- Fix false positive for ``too-many-function-args`` when a function call is assigned to a class attribute inside the class where the function is defined.
+
+  Closes #6592 (`#6592 <https://github.com/PyCQA/pylint/issues/6592>`_)
+- Fix ``used-before-assignment`` for functions/classes defined in type checking guard.
+
+  Closes #7368 (`#7368 <https://github.com/PyCQA/pylint/issues/7368>`_)
+- Fix ignored files being linted when passed on stdin.
+
+  Closes #4354 (`#4354 <https://github.com/PyCQA/pylint/issues/4354>`_)
+- ``missing-return-doc``, ``missing-raises-doc`` and ``missing-yields-doc`` now respect
```
DanielNoord, line 59: Doesn't `towncrier` know about white spaces?  😅

Pierre-Sassoulas: I manually removed the duplicated titles, which is why there might be some problems.

Pierre-Sassoulas: Let's fix in #7426

PR #7395, opened 2022-09-01
False positive for `too-many-function-args`

### LINKED ISSUE #6592: Incorrect too-many-function-args for a function defined and called within a class
### Bug description

```python
class Dyn:
    def _make_method(name):     # pylint: disable=no-self-argument
        def _the_method(self):
            print(name)
        return _the_method

    print_nedbat = _make_method("nedbat")

Dyn().print_nedbat()
```
This code works. But pylint doesn't like it:
```
% pylint -E weirdmethod.py
************* Module weirdmethod
weirdmethod.py:7:19: E1121: Too many positional arguments for method call (too-many-function-args)
```

### Configuration

_No response_

### Command used

```shell
pylint -E weirdmethod.py
```

### Pylint output

```shell
************* Module weirdmethod
weirdmethod.py:7:19: E1121: Too many positional arguments for method call (too-many-function-args)
```

### Expected behavior

Since the code works, this is not an error. Pylint shouldn't complain.

### Pylint version

```shell
pylint 2.13.8
astroid 2.11.5
Python 3.7.13 (default, Mar 16 2022, 20:45:04)

(btw, this didn't happen with pylint 2.13.7 and astroid 2.11.3)
```

### OS / Environment

Mac OS 10.15.7

### Additional dependencies

_No response_

mbyrnepr2:
## Type of Changes

## Description
False positive `too-many-function-args`

Closes #6592

### DIFF
--- a/tests/functional/a/arguments.py
+++ b/tests/functional/a/arguments.py
@@ -1,6 +1,6 @@
 # pylint: disable=too-few-public-methods, missing-docstring,import-error,wrong-import-position
 # pylint: disable=wrong-import-order, unnecessary-lambda, consider-using-f-string
-# pylint: disable=unnecessary-lambda-assignment
+# pylint: disable=unnecessary-lambda-assignment, no-self-argument, unused-argument
 
 def decorator(fun):
     """Decorator"""
@@ -261,3 +261,20 @@ def func(one, two, three):
 
 
 CALL = lambda *args: func(*args)
+
+
+# Ensure `too-many-function-args` is not emitted when a function call is assigned
+# to a class attribute inside the class where the function is defined.
+# Reference: https://github.com/PyCQA/pylint/issues/6592
+class FruitPicker:
+    def _pick_fruit(fruit):
+        def _print_selection(self):
+            print(f"Selected: {fruit}!")
+        return _print_selection
+
+    pick_apple = _pick_fruit("apple")
+    pick_pear = _pick_fruit("pear")
+
+picker = FruitPicker()
+picker.pick_apple()
+picker.pick_pear()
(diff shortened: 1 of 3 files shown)


tests/functional/a/arguments.py
```diff
@@ -261,3 +261,20 @@ def func(one, two, three):
 
 
 CALL = lambda *args: func(*args)
+
+
+# Ensure `too-many-function-args` is not emitted when a function call is assigned
+# to a class attribute inside the class where the function is defined.
+# Reference: https://github.com/PyCQA/pylint/issues/6592
+class FruitPicker:
```
Pierre-Sassoulas, line 269: Would this raise before ? I suppose we need more (4 if we look above) arguments to trigger the message anyway ?

Pierre-Sassoulas (commented):
LGTM, I only have a question about the functional test.

DanielNoord (approved):

tests/functional/a/arguments.py
mbyrnepr2: Sorry if I misunderstand your question but let's say if I checkout `main` & copy the class as-is & run Pylint on it in an example.py, we get the error for each of these two lines in the test:
```python
pick_apple = _pick_fruit("apple")
pick_pear = _pick_fruit("pear")
```

Before the fix, these calls were seen as having two arguments each (one extra for an implicit `self`), and the code was comparing these arguments to the single `fruit` parameter from the definition of `_pick_fruit`.

Pierre-Sassoulas: Ho, yes, my bad. I've mistaken this check with ``too-many-arguments``

Pierre-Sassoulas (approved):

PR #7398, opened 2022-09-01
Fix and refactors for ``docparams`` extension

DanielNoord:

## Type of Changes

## Description

Some refactors based on the work in https://github.com/PyCQA/pylint/pull/7360

I'd like to see how the primer reacts to separate some of the large changes we were seeing in that PR.

### DIFF
--- a/doc/whatsnew/fragments/7398.other
+++ b/doc/whatsnew/fragments/7398.other
@@ -1,0 +1,4 @@
+The ``docparams`` extension now considers typing in Numpy style docstrings
+as "documentation" for the ``missing-param-doc`` message.
+
+Refs #7398
--- a/pylint/extensions/_check_docs_utils.py
+++ b/pylint/extensions/_check_docs_utils.py
@@ -806,6 +806,13 @@ class NumpyDocstring(GoogleDocstring):
                 param_name = match.group("param_name")
                 param_type = match.group("param_type")
                 param_desc = match.group("param_desc")
+                # The re_param_line pattern needs to match multi-line which removes the ability
+                # to match a single line description like 'arg : a number type.'
+                # We are not trying to determine whether 'a number type' is correct typing
+                # but we do accept it as typing as it is in the place where typing
+                # should be
+                if param_type is None and re.match(r"\s*(\*{0,2}\w+)\s*:.+$", entry):
+                    param_type = param_desc
 
             if param_type:
                 params_with_type.add(param_name)
--- a/pylint/extensions/_check_docs_utils.py
+++ b/pylint/extensions/_check_docs_utils.py
@@ -247,7 +247,7 @@ class SphinxDocstring(Docstring):
 
     re_multiple_simple_type = r"""
         (?:{container_type}|{type})
-        (?:(?:\s+(?:of|or)\s+|\s*,\s*)(?:{container_type}|{type}))*
+        (?:(?:\s+(?:of|or)\s+|\s*,\s*|\s+\|\s+)(?:{container_type}|{type}))*
     """.format(
         type=re_type, container_type=re_simple_container_type
     )
@@ -449,7 +449,7 @@ class GoogleDocstring(Docstring):
 
     re_multiple_type = r"""
         (?:{container_type}|{type}|{xref})
-        (?:(?:\s+(?:of|or)\s+|\s*,\s*)(?:{container_type}|{type}|{xref}))*
+        (?:(?:\s+(?:of|or)\s+|\s*,\s*|\s+\|\s+)(?:{container_type}|{type}|{xref}))*
     """.format(
         type=re_type, xref=re_xref, container_type=re_container_type
     )
@@ -728,13 +728,13 @@ class NumpyDocstring(GoogleDocstring):
         re.X | re.S | re.M,
     )
 
-    re_default_value = r"""((['"]\w+\s*['"])|(True)|(False)|(None))"""
+    re_default_value = r"""((['"]\w+\s*['"])|(\d+)|(True)|(False)|(None))"""
 
     re_param_line = re.compile(
         rf"""
-        \s*  (\*{{0,2}}\w+)(\s?(:|\n))              # identifier with potential asterisks
+        \s*  (?P<param_name>\*{{0,2}}\w+)(\s?(:|\n)) # identifier with potential asterisks
         \s*
-        (
+        (?P<param_type>
          (
           ({GoogleDocstring.re_multiple_type})      # default type declaration
           (,\s+optional)?                           # optional 'optional' indication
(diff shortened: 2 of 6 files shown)


DanielNoord:
@Pierre-Sassoulas This allows the PR of @adam-grant-hendry to only focus on the `default` aspect. I think we can merge this?

Pierre-Sassoulas (commented):
I like the new PR to separate the two modifications.

Are we sure the style of the docstring is detected properly ([Pandas seems to be using numpy](https://pandas.pydata.org/docs/dev/development/contributing_docstring.html#docstring, based on [this](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)), so it's strange that there is so much violations. On the other hand it seems they suggest to check with pylint so it's entirely possible that if pylint was not detecting it previously then some of their docstring are wrong...

Also shouldn't we add explicit functional tests for the new test cases ? Regexes are pretty easy to get wrong so regression tests would really help in case of future changes.

DanielNoord:
https://github.com/PyCQA/pylint/runs/8144779659?check_suite_focus=true

If you search for "fh" in the logs it shows that we are not actually recognizing these parameters to begin with. With this change we are recognising them, but do not consider them correctly typed.
I'm not sure what we should do there. 

Take:
https://github.com/pandas-dev/pandas/blob/bfdb524f5807892adde0e5d3d754351459cd1614/pandas/compat/pickle_compat.py#L256
https://github.com/pandas-dev/pandas/blob/bfdb524f5807892adde0e5d3d754351459cd1614/pandas/compat/numpy/function.py#L416

It's clear that they are trying to put types on the first line and not descriptions. But in the first example there are also descriptions and not only types. Should we raise `missing-type-doc` here? Or should we just accept it?

Pierre-Sassoulas:
> Should we raise missing-type-doc here? Or should we just accept it?

I think we should not raise ``missing-type-doc`` as we're not in the business of checking if a bunch of English words are an attempt to describe the type of a parameter 😄 But the function signature is typed, are we already raising ``missing-type-doc`` in this case ? I would say we should not but I want to avoid churn if we are.

DanielNoord:
> > Should we raise missing-type-doc here? Or should we just accept it?
> 
> I think we should not raise `missing-type-doc` as we're not in the business of checking if a bunch of English words are an attempt to describe the type of a parameter 😄 But the function signature is typed, are we already raising `missing-type-doc` in this case ? I would say we should not but I want to avoid churn if we are.

We are. I think we don't check signatures for this check.

I'll look into accepting these docstrings in `pandas`. Drafting this.

Pierre-Sassoulas:
> We are. I think we don't check signatures for this check.

Let's keep raising it then, we can't know that the description is trying to define the typing in plain English.

> I'll look into accepting these docstrings in pandas. Drafting this.

I think this could be merged if we add some functional tests to formalize and check what we want to prevent future regression.

DanielNoord:
> > We are. I think we don't check signatures for this check.
> 
> 
> 
> Let's keep raising it then, we can't know that the description is trying to define the typing in plain English.
> 
> 

Wouldn't it make more sense then to just accept anything that is being written on the line that is intended for typing? And not bother about determining whether it is actually typing.

Pierre-Sassoulas:
I thought it was supposed to be like ``:param int variable_name: description``, my numpy's style knowledge is... not rusty, I never had one :) So yeah, if it's supposed to be the typing then let's accept it.

DanielNoord:
> I thought it was supposed to be like `:param int variable_name: description`, my numpy's style knowledge is... not rusty, I never had one :) So yeah, if it's supposed to be the typing then let's accept it.

I think that's Sphinx

DanielNoord:
@Pierre-Sassoulas Let's get final confirmation from the primer, but I think I fixed the issue here 😄

Pierre-Sassoulas (approved):
👌

DanielNoord:
@Pierre-Sassoulas The last commit is debatable. Do we want to raise `missing-param-doc` if there is a type documentation? I'd say no, but our tests say otherwise 😓 

I think that if you're documenting the type in a docstring you'll probably add a description if you really want to.

Pierre-Sassoulas:
>  Do we want to raise missing-param-doc if there is a type documentation? I'd say no, but our tests say otherwise 

I'm personally rather lax with param documentation in docstring because if you use proper variable name then the param description is VERY often just the variable name with a more natural english capitalization... So if we have type information I would say it's voluntary that the param is not documented further... Now, I'm not using the docparam checker myself. Is my opinion the same as the one from persons that would actually use this checker... ? Probably not.

DanielNoord:
> > Do we want to raise missing-param-doc if there is a type documentation? I'd say no, but our tests say otherwise
> 
> I'm personally rather lax with param documentation in docstring because if you use proper variable name then the param description is VERY often just the variable name with a more natural english capitalization... So if we have type information I would say it's voluntary that the param is not documented further... Now, I'm not using the docparam checker myself. Is my opinion the same as the one from persons that would actually use this checker... ? Probably not.

Yeah I guess it depends on what "documentation" is. Is it any form of documentation or only a specific description?

adam-grant-hendry:
> Yeah I guess it depends on what "documentation" is. Is it any form of documentation or only a specific description?

@DanielNoord @Pierre-Sassoulas I'd be careful: documentation is code. The end user may add an examples section, and if they use incorrect names, that should fail. In these days, that documentation may even also include `mypy` type hints.

DanielNoord:
> > Yeah I guess it depends on what "documentation" is. Is it any form of documentation or only a specific description?
> 
> @DanielNoord @Pierre-Sassoulas I'd be careful: documentation is code. The end user may add an examples section, and if they use incorrect names, that should fail. In these days, that documentation may even also include `mypy` type hints.

Yeah agreed, but take:
```python
def func(arg):
     """     
     Parameters
     ----------
     arg : int or None
     """     
```

Should we really raise `missing-param-doc` here for arg? I would consider it to be documented..

Pierre-Sassoulas:
Agree with the last point, I say let's not raise but maybe also let's wait for additional opinion, 2.16 is not out yet we can wait a little before merging.

adam-grant-hendry:
> Should we really raise `missing-param-doc` here for arg?

No. I would raise `missing-param-type` though or a type mismatch error. `arg` is typed as `typing.Any`, but declared to be `int or None` here.

DanielNoord:
> > Should we really raise `missing-param-doc` here for arg?
> 
> No, but I would raise `missing-param-type` though or a type mismatch error. `arg` is typed as `typing.Any`, but declared to be `int or None` here.

Yeah, but that should be its own message probably. Something like `docstring-signature-type-mismatch`. Although I don't know if we want to get into that business. Resolving types in signatures is hard..

Pierre-Sassoulas:
> Although I don't know if we want to get into that business. Resolving types in signatures is hard..

Strong agree here, this is full project in itself. Let's raise one of the two existing messages and let the user sort it out.

Pierre-Sassoulas (approved):

adam-grant-hendry:
> Yeah, but that should be its own message probably. Something like docstring-signature-type-mismatch. Although I don't know if we want to get into that business. Resolving types in signatures is hard..

That's completely fair. I understand 👍

PR #7400, opened 2022-09-02
Fix #3299 false positives with names in string literal type annotations

### LINKED ISSUE #3299: Pylint incorrectly flags as `unused-import` things that are used in string literal type annotations
### Steps to reproduce

pylint_example.py:

```
class Class:
    pass
```

pylint_example2.py:

```
from pylint_example import Class


class Class2:
    cls: 'Class'
```

pylint pylint_example2.py

```
************* Module pylint_example2
pylint_example2.py:4:0: R0903: Too few public methods (0/2) (too-few-public-methods)
pylint_example2.py:1:0: W0611: Unused Class imported from pylint_example (unused-import)

Your code has been rated at 3.33/10 (previous run: 3.33/10, +0.00)
```

### Current behavior

String literal in type annotation is considered "unused" and warning is emitted.

### Expected behavior

There should not be any warning about unused import.
 
### pylint --version output

pylint 2.4.4
astroid 2.3.3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]

lggruspe:
## Type of Changes

## Description

Closes #3299

Changes:

- `VariablesChecker` now checks for possible names in `Const` strings that appear in a typing annotation context. This fixes some false positives for `unused-import` and `unused-variable`.
- string `Const`s used as args to `typing.Literal` aren't treated as names.
- String annotations like `"os.PathLike[str]"` or `"graphlib.TopologicalSorter[T]"` now work

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1891,6 +1891,25 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
     return False
 
 
+def is_typing_literal(node: nodes.NodeNG) -> bool:
+    """Check if a node refers to typing.Literal."""
+    if isinstance(node, nodes.Name):
+        import_from = node.lookup(node.name)[1][0]
+        if isinstance(import_from, nodes.ImportFrom):
+            return (
+                import_from.modname == "typing"
+                and import_from.real_name(node.name) == "Literal"
+            )
+    elif isinstance(node, nodes.Attribute):
+        inferred_module = safe_infer(node.expr)
+        return (
+            isinstance(inferred_module, nodes.Module)
+            and inferred_module.name == "typing"
+            and node.attrname == "Literal"
+        )
+    return False
+
+
 @lru_cache()
 def in_for_else_branch(parent: nodes.NodeNG, stmt: nodes.Statement) -> bool:
     """Returns True if stmt is inside the else branch for a parent For stmt."""
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1894,7 +1894,10 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
 def is_typing_literal(node: nodes.NodeNG) -> bool:
     """Check if a node refers to typing.Literal."""
     if isinstance(node, nodes.Name):
-        import_from = node.lookup(node.name)[1][0]
+        try:
+            import_from = node.lookup(node.name)[1][0]
+        except IndexError:
+            return False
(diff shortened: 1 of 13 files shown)


tests/functional/u/unused/unused_name_in_string_literal_type_annotation_py39.py
```diff
@@ -0,0 +1,10 @@
+# pylint: disable=missing-docstring
```
Pierre-Sassoulas, line 1: Why is this test python 3.9 only ?

Pierre-Sassoulas (commented):
Thank you, LGTM, I have only a small question.

tests/functional/u/unused/unused_name_in_string_literal_type_annotation_py39.py
lggruspe: graphlib was introduced in python 3.9.

I added an explicit test for it, because mypy expects TopologicalSorter to have a subscript even though it's not subscriptable at runtime. The only way to properly annotate it at the moment is to use a string.

pylint/checkers/utils.py
```diff
@@ -1891,6 +1891,25 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
     return False
 
 
+def is_typing_literal(node: nodes.NodeNG) -> bool:
+    """Check if a node refers to typing.Literal."""
+    if isinstance(node, nodes.Name):
+        import_from = node.lookup(node.name)[1][0]
```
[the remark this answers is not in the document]
lggruspe: I made the function return False when name lookup fails.

pylint/checkers/utils.py
```diff
@@ -1891,6 +1891,28 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
     return False
 
 
+def is_typing_literal(node: nodes.NodeNG) -> bool:
+    """Check if a node refers to typing.Literal."""
+    if isinstance(node, nodes.Name):
+        try:
+            import_from = node.lookup(node.name)[1][0]
+        except IndexError:
+            return False
```
DanielNoord, line 1900: The problem is that know this is probably uncovered. Should we make a unittest for it?

DanielNoord (approved):
For some reason the coverage is going down, but I think that is unrelated to the changes here. Probably coveralls screwing something up.

Thanks @lggruspe! 😄

Pierre-Sassoulas (approved):
Thank you @lggruspe. The changes look good, and the primer results warm my heart.

PR #7407, opened 2022-09-04
Turn on ``check-untyped-defs`` in ``mypy``

DanielNoord:

## Type of Changes

## Description

Once again sorry...

Without turning on the flag we get a lot of `unused ignore` warnings, but with the flag we need a PR of this size. Chicken and egg..
I think this is best reviewed in commit view, as I have tried to separate all actions.

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -160,7 +160,7 @@ def _make_tree_defs(mod_files_list: ItemsView[str, set[str]]) -> _ImportTree:
         node: list[_ImportTree | list[str]] = [tree_defs, []]
         for prefix in mod.split("."):
             assert isinstance(node[0], dict)
-            node = node[0].setdefault(prefix, ({}, []))  # type: ignore[arg-type]
+            node = node[0].setdefault(prefix, ({}, []))  # type: ignore[arg-type,assignment]
         assert isinstance(node[1], list)
         node[1] += files
     return tree_defs
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -39,7 +39,7 @@ if TYPE_CHECKING:
 
 # The dictionary with Any should actually be a _ImportTree again
 # but mypy doesn't support recursive types yet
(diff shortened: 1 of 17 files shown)


Pierre-Sassoulas:
Wow, huge step 🎉  ! I'll have more time to review next week, sorry. It seems like it require an uninterrupted time slot

DanielNoord:
> Wow, huge step 🎉 ! I'll have more time to review next week, sorry. It seems like it require an uninterrupted time slot

😄 

The commits itself are quite small so it shouldn't require too much brainpower, but indeed this is a little more difficult than a standard typing PR.

pylint/checkers/imports.py
```diff
@@ -420,7 +425,7 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
     def __init__(self, linter: PyLinter) -> None:
         BaseChecker.__init__(self, linter)
         self.import_graph: defaultdict[str, set[str]] = defaultdict(set)
-        self._imports_stack: list[tuple[Any, Any]] = []
+        self._imports_stack: list[tuple[ImportNode, str]] = []
```
Pierre-Sassoulas, line 428: 👍

pylint/checkers/imports.py
```diff
@@ -39,7 +39,7 @@
 
 # The dictionary with Any should actually be a _ImportTree again
 # but mypy doesn't support recursive types yet
-_ImportTree = Dict[str, List[Dict[str, Any] | List[str]]]
+_ImportTree = Dict[str, Union[List[Dict[str, Any], List[str]]]]
```
Pierre-Sassoulas, line 43: I don't understand this change, is'nt the ``|`` operator better ?

Pierre-Sassoulas (commented):
Look good ! I have a small question before merging.

pylint/checkers/imports.py
DanielNoord: The operator isn't supported for runtime executed statements. The future import allows to use them in typing, but aliases are executed so we need to use the typing imports directly rather than the more modern notation.

Pierre-Sassoulas (approved):

PR #7409, opened 2022-09-04
Add regression test for #3973

### LINKED ISSUE #3973: Can't disable "invalid-name" error for module name
### Steps to reproduce
1. Create an empty Python file called `run-tests.py`
2. At the top of the file, add `# pylint: disable=invalid-name` so that pylint won't complain about the module name
3. Pylint complains anyway

### Current behavior
Explained above

### Expected behavior
Pylint should not complain about the module name. Yes I realize I can disable the rule entirely in the config file, but 99% of the time I want the rule enforced. The exception is when I'm writing Python *scripts* (rather than modules), in which case I prefer the dashed naming style, as in `run-tests.py`.

### pylint --version output
```
pylint 2.6.0
astroid 2.4.2
Python 3.8.6 (default, Oct 29 2020, 08:08:02) 
[GCC 9.3.0]
```

DanielNoord:

## Type of Changes

## Description

Closes #3973

### DIFF
--- a/tests/functional/i/invalid/invalid_name/invalid_name-module-disable.py
+++ b/tests/functional/i/invalid/invalid_name/invalid_name-module-disable.py
@@ -1,0 +1,6 @@
+# pylint: disable=invalid-name
+
+"""Regression test for disabling of invalid-name for module names.
+
+See https://github.com/PyCQA/pylint/issues/3973.
+"""


Pierre-Sassoulas (approved):
We can close the issue I guess. There's no use for empty file with invalid name afaik

PR #7411, opened 2022-09-05
Make ``disable-next`` only consider the succeeding line

### LINKED ISSUE #7401: `disable-next` leaks into the whole block if it's followed by an empty line
### Bug description

1. Create a virtual environment (e.g. run `virtualenv pylint-repro`)
2. Activate the virtual environment (e.g. run `. pylint-repro/bin/activate`
3. Install pylint (e.g. `pip install pylint`)
4. Put these contents into an `example.py` file:
```py


def func():
    # pylint: disable-next=unused-variable

    test = 43
    blah = 123
```
5. Run `pylint example.py`
6. See the issue.

### Configuration

_No response_

### Command used

```shell
pylint example.py
```

### Pylint output

```shell
```

### Expected behavior

I expected the second unused variable to be reported, as is the case when not putting an empty line between `disable-next` and the line I'm suppressing errors for:
```py


def func():
    # pylint: disable-next=unused-variable
    test = 43
    blah = 123
```
Example output:
```
************* Module example
example.py:7:4: W0612: Unused variable 'blah' (unused-variable)

Your code has been rated at 6.67/10 (previous run: 6.67/10, +0.00)
```

It probably would also be fine if an empty line between the comment and line to suppress simply didn't work. My issue is that this *appears* to do what I've intended - it suppresses the unused variable for next line - but it also does things it wasn't supposed to.

### Pylint version

```shell
pylint 2.15.0
astroid 2.12.5
```

### OS / Environment

Ubuntu 22.04.1, Python 3.10 shipped with the system

### Additional dependencies

_No response_

DanielNoord:

## Type of Changes

## Description

Closes #7401

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2210,7 +2210,7 @@ class VariablesChecker(BaseChecker):
         # assignment (which does not exist in code per se) as well as any later
         # modifications.
         if (
-            not astmts # pylint: disable=too-many-boolean-expressions
+            not astmts  # pylint: disable=too-many-boolean-expressions
             or (
                 astmts[0].parent == astmts[0].root()
                 and astmts[0].parent.parent_of(node)
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2209,9 +2209,8 @@ class VariablesChecker(BaseChecker):
         # scope lookup rules would need to be changed to return the initial
         # assignment (which does not exist in code per se) as well as any later
         # modifications.
-        # pylint: disable-next=too-many-boolean-expressions
         if (
-            not astmts
+            not astmts # pylint: disable=too-many-boolean-expressions
             or (
                 astmts[0].parent == astmts[0].root()
                 and astmts[0].parent.parent_of(node)
--- a/tests/functional/d/disable_msg_next_line.py
+++ b/tests/functional/d/disable_msg_next_line.py
@@ -18,3 +18,10 @@ def function_C():
 
 def function_D(arg1, arg2):  # [unused-argument, invalid-name]
     return arg1
+
+
+def function_E():  # [invalid-name]
+    # pylint: disable-next=unused-variable
+
+    test = 43  # [unused-variable]
+    blah = 123  # [unused-variable]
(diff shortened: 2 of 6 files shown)


tests/functional/d/disable_msg_next_line.py
```diff
@@ -18,3 +18,10 @@ def function_C():
 
 def function_D(arg1, arg2):  # [unused-argument, invalid-name]
     return arg1
+
+
+def function_E():  # [invalid-name]
+    # pylint: disable-next=unused-variable
+
```
Pierre-Sassoulas, line 25:
```suggestion
bleh = 42
```

Pierre-Sassoulas (dismissed):
👌

tests/functional/d/disable_msg_next_line.py
DanielNoord: ?

Pierre-Sassoulas: (To check that the disable next do work on the next line)

Pierre-Sassoulas: Nvm, I did not read the issue properly.

DanielNoord: 😄 

Rest of the file tests that as well!

pylint/checkers/variables.py
```diff
@@ -2209,9 +2209,8 @@ def _loopvar_name(self, node: astroid.Name) -> None:
         # scope lookup rules would need to be changed to return the initial
         # assignment (which does not exist in code per se) as well as any later
         # modifications.
-        # pylint: disable-next=too-many-boolean-expressions
         if (
-            not astmts
+            not astmts # pylint: disable=too-many-boolean-expressions
```
DanielNoord, line 2213: Is this an acceptable change in behaviour? I think it is. This seems to be an unintended side effect of checking for block lines that we didn't consider previously.

Pierre-Sassoulas: Is there a way to make disable-next work on the whole if like before ? I think it's going to be reported as a regression, having to change disables when upgrading pylint does not feel great.

DanielNoord: Not if we also don't want to have the issue reported in the OP. I think it makes sense for `disable-next` only to work for the next line and I think this is a relatively minor change to have to fix in an auto-update. Obviously it is annoying but let's fix this now while the impact is still relatively minor (compared to a 10 year long supported behaviour).

Pierre-Sassoulas: Fair enough !

Pierre-Sassoulas (approved):

PR #7414, opened 2022-09-05
Disambiguate between str and enum member args to typing.Literal

lggruspe:
## Type of Changes

## Description

This is a continuation of #7400. This PR fixes some errors in the handling of typing.Literal.

From https://peps.python.org/pep-0586/#literals-enums-and-forward-references:

> One potential ambiguity is between literal strings and forward references to literal enum members. For example, suppose we have the type Literal["Color.RED"]. Does this literal type contain a string literal or a forward reference to some Color.RED enum member?

> In cases like these, we always assume the user meant to construct a literal string. If the user wants a forward reference, they must wrap the entire literal type in a string – e.g. "Literal[Color.RED]".

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2924,15 +2924,10 @@ class VariablesChecker(BaseChecker):
         parent = node.parent
         if isinstance(parent, nodes.Tuple):
             parent = parent.parent
-
         if isinstance(parent, nodes.Subscript):
             origin = next(parent.get_children(), None)
             if origin is not None and utils.is_typing_literal(origin):
                 return
-
-        if node.value.isidentifier():
-            self._type_annotation_names.append(node.value)
-            return
 
         try:
             annotation = extract_node(node.value)
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2919,16 +2919,6 @@ class VariablesChecker(BaseChecker):
             return
         if not utils.is_node_in_type_annotation_context(node):
             return
-        if not node.value.isidentifier():
-            try:
-                annotation = extract_node(node.value)
-                self._store_type_annotation_node(annotation)
-            except ValueError:
-                # e.g. node.value is white space
-                return
-            except astroid.AstroidSyntaxError:
-                # e.g. "?" or ":" in typing.Literal["?", ":"]
-                return
 
         # Check if parent's or grandparent's first child is typing.Literal
         parent = node.parent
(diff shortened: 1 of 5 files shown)


tests/functional/u/unused/unused_name_in_string_literal_type_annotation.py
```diff
@@ -21,6 +21,11 @@ def example3(_: "os.PathLike[str]") -> None:
 def example4(_: "PathLike[str]") -> None:
     """unused-import shouldn't be emitted for PathLike."""
 
+# pylint shouldn't crash with the following strings in a type annotation context
+example5: Set[""]
```
lggruspe, line 25: This was added to increase the test coverage. Should pylint emit a message when it finds an invalid type annotation?

DanielNoord: Let's leave that to `mypy` and the like 😄

tests/functional/u/unused/unused_name_in_string_literal_type_annotation_py38.py
```diff
@@ -2,8 +2,10 @@
 
 from argparse import ArgumentParser # [unused-import]
 from argparse import Namespace  # [unused-import]
-from typing import Literal as Lit
+import http #[unused-import]
```
DanielNoord, line 5:
```suggestion
import http  # [unused-import]
```

pylint/checkers/variables.py
```diff
@@ -2940,7 +2930,19 @@ def visit_const(self, node: nodes.Const) -> None:
             if origin is not None and utils.is_typing_literal(origin):
                 return
 
-        self._type_annotation_names.append(node.value)
+        if node.value.isidentifier():
```
DanielNoord, line 2933: Could you explain why you moved this?

lggruspe: I moved the is_typing_literal part up to stop checking for names if the string node is an arg to `Literal`. As for the `isidentifier` check, it's probably not needed anymore. I'll push some more changes.

DanielNoord (dismissed):
This does need a news fragment still.

DanielNoord (approved):

DanielNoord:
Thanks you @lggruspe!

DanielNoord:
@Pierre-Sassoulas Should this be included as well? Milestone is missing (my bad probably) but it has the label.

Pierre-Sassoulas:
Let's add it too.

PR #7426, opened 2022-09-06
Merge maintenance 2.15.x in main

Pierre-Sassoulas:
There's an issue in the fragment that are not removed after the merge that I added in #7362. Hopefully we can delete all false positives and bug fixes fragments.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   # Also change CACHE_VERSION in the other workflows
-  CACHE_VERSION: 24
+  CACHE_VERSION: 25
   DEFAULT_PYTHON: "3.10"
 
 jobs:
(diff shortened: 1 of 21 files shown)


DanielNoord:
> There's an issue in the fragment that are not removed after the merge that I added in #7362. Hopefully we can delete all false positives and bug fixes fragments.

What do you mean? Are some fragments still there? Isn't that what we want for the `2.16` release?

Pierre-Sassoulas:
> What do you mean? Are some fragments still there? Isn't that what we want for the 2.16 release?

Duh, you're right I forgot that.

Pierre-Sassoulas:
Did not manage to find the issue with the discussion (Even if I searched quite extensively), but if I remember correctly @DudeNr33 explained that it was intended to be like this a little prior to the discussion above in another issue and I forgot about it in this MR which is why the discussion is so short here. Would you mind sharing your view on the matter in https://github.com/pylint-dev/pylint/issues/9302 Andreas ?

PR #7430, opened 2022-09-07
Bump pylint to 2.15.2, update changelog

Pierre-Sassoulas:
Too much manual changes to not make mistakes sadly :-(

### DIFF
--- a/doc/whatsnew/2/2.15/index.rst
+++ b/doc/whatsnew/2/2.15/index.rst
@@ -30,10 +30,18 @@ Marc Byrne became a maintainer, welcome to the team !
 .. towncrier release notes start
 
 
-What's new in Pylint 2.15.1?
+What's new in Pylint 2.15.2?
 ----------------------------
-Release date: 2022-09-06
+Release date: 2022-09-07
 
+- Fixed a case where custom plugins specified by command line could silently fail.
+
+  Specifically, if a plugin relies on the ``init-hook`` option changing ``sys.path`` before
+  it can be imported, this will now emit a ``bad-plugin-value`` message. Before this
+  change, it would silently fail to register the plugin for use, but would load
+  any configuration, which could have unintended effects.
+
+  Fixes part of #7264. (`#7264 <https://github.com/PyCQA/pylint/issues/7264>`_)
 - Fix ``used-before-assignment`` for functions/classes defined in type checking guard.
 
   Closes #7368 (`#7368 <https://github.com/PyCQA/pylint/issues/7368>`_)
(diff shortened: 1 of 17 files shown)


DanielNoord (changes requested):
There are two additional crashes/regressions reported today.

`astroid` `2.12.10`? 😅

Pierre-Sassoulas:
It would be 2.12.9 I think ? I think we need to release, there's a lot of fix waiting in the pipeline and I botched 2.15.1 :(

I also need to fix the automated release tool because it's not as fast as I'd like right now but maybe when it's calmer.

DanielNoord:
> It would be 2.12.9 I think ? I think we need to release, there's a lot of fix waiting in the pipeline and I botched 2.15.1 :(
> 
> I also need to fix the automated release tool because it's not as fast as I'd like right now but maybe when it's calmer.

I have a fix for the first crash regression almost ready. Just adding some tests.

I though of a solution for the other issue while in the shower this morning. Should be fairly straightforward as well.

If it isn't too much trouble I think it would be good to put them in as well.

Pierre-Sassoulas:
All right then, let's release tonight.

DanielNoord:
Both fixes have been pushed to `astroid`. One is already merged 😄

doc/whatsnew/2/2.15/index.rst
```diff
@@ -30,10 +30,18 @@ Marc Byrne became a maintainer, welcome to the team !
...
+Release date: 2022-09-07
+
+- Fixed a case where custom plugins specified by command line could silently fail.
 
+  Specifically, if a plugin relies on the ``init-hook`` option changing ``sys.path`` before
+  it can be imported, this will now emit a ``bad-plugin-value`` message. Before this
+  change, it would silently fail to register the plugin for use, but would load
+  any configuration, which could have unintended effects.
+
+  Fixes part of #7264. (`#7264 <https://github.com/PyCQA/pylint/issues/7264>`_)
```
DanielNoord, line 44: Can we teach towncrier about whitespaces?

DanielNoord (changes requested):

doc/whatsnew/2/2.15/index.rst
Pierre-Sassoulas: What would you expect ?

DanielNoord: https://github.com/twisted/towncrier/issues/80

Setting `wrap` to `False` should work?

DanielNoord (approved):

PR #7431, opened 2022-09-07
[towncrier] Add whitespaces between fragment in towncrier

Pierre-Sassoulas:
## Type of Changes

## Description

Turn out I was checking the wrong file when generating the release notes.

### DIFF
--- a/doc/whatsnew/fragments/_template.rst
+++ b/doc/whatsnew/fragments/_template.rst
@@ -18,7 +18,6 @@ Release date: {{ versiondata.date }}
 {% for text, values in sections[section][category].items() %}
 - {{ text }} ({{ values|join(', ') }})
 
-
 {% endfor %}
 
 {% else %}
--- a/doc/whatsnew/fragments/_template.rst
+++ b/doc/whatsnew/fragments/_template.rst
@@ -17,6 +17,8 @@ Release date: {{ versiondata.date }}
 {% if definitions[category]['showcontent'] %}
 {% for text, values in sections[section][category].items() %}
 - {{ text }} ({{ values|join(', ') }})
+
+
 {% endfor %}
 
 {% else %}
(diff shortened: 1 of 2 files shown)


doc/whatsnew/fragments/_template.rst
```diff
@@ -17,6 +17,8 @@ Release date: {{ versiondata.date }}
 {% if definitions[category]['showcontent'] %}
 {% for text, values in sections[section][category].items() %}
 - {{ text }} ({{ values|join(', ') }})
+
```
DanielNoord, line 20: should this be 2?

Pierre-Sassoulas: If it's one there's the same spacing between two fragments and the space inside fragments ie. 
```rst
- Fixed a case where custom plugins specified by command line could silently fail.

  Specifically, if a plugin relies on the ``init-hook`` option changing ``sys.path`` before
  it can be imported, this will now emit a ``bad-plugin-value`` message. Before this
  change, it would silently fail to register the plugin for use, but would load
  any configuration, which could have unintended effects.

  Fixes part of #7264. (`#7264 <https://github.com/PyCQA/pylint/issues/7264>`_)

- Fix ``used-before-assignment`` for functions/classes defined in type checking guard.

  Closes #7368 (`#7368 <https://github.com/PyCQA/pylint/issues/7368>`_)
```  
vs 
```
```rst
- Fixed a case where custom plugins specified by command line could silently fail.

  Specifically, if a plugin relies on the ``init-hook`` option changing ``sys.path`` before
  it can be imported, this will now emit a ``bad-plugin-value`` message. Before this
  change, it would silently fail to register the plugin for use, but would load
  any configuration, which could have unintended effects.

  Fixes part of #7264. (`#7264 <https://github.com/PyCQA/pylint/issues/7264>`_)


- Fix ``used-before-assignment`` for functions/classes defined in type checking guard.

  Closes #7368 (`#7368 <https://github.com/PyCQA/pylint/issues/7368>`_)
```

DanielNoord: The first style is also the pre-towncrier style right? Both are fine with me, but perhaps following the old style is best?

DanielNoord (approved):

PR #7432, opened 2022-09-07
Merge maintenance 2.15.x in main

Pierre-Sassoulas:


Pierre-Sassoulas:
Pushing the merge commit directly into main was a lot easier, I have no idea how to proceed right now.

Pierre-Sassoulas:
I think the primer are expected to fail here. We need to be careful about how we integrate this in main, I think it's "create a merge commit"

.github/workflows/changelog.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 env:
   # Also change CACHE_VERSION in the other workflows
-  CACHE_VERSION: 24
+  CACHE_VERSION: 26
```
DanielNoord, line 9: Why are we bumping this again?

Pierre-Sassoulas: Because of astroid 2.12.8 and 2.12.9. (``25`` is already taken by ``2.12.8``)

DanielNoord: Oh huh, I didn't notice `main` isn't on `2.12.9` yet. How did that happen?

DanielNoord (approved):

.github/workflows/changelog.yml
Pierre-Sassoulas: I upgraded astroid on the maintenance branch directly instead of doing it on main and then backporting (as we were going go merge the maintenance branch in main anyway).

PR #7436, opened 2022-09-08
Feature italics abstract methods

### LINKED ISSUE #7346: Let pyreverse show abstract methods differently
### Current problem

Currently, there's no way to distinguish abstract methods.

### Desired solution

Abstract methods are shown in italic in UML diagrams.
I propose to get that idea and shows `@abstractmethod` also in italic.
CC @qequ

### Additional context

_No response_

qequ:

## Type of Changes

|      | Type                   |

## Description
Update Pyreverse's dot, plantuml and Mermaid.js printers to detect when class methods are abstract and show them with italic font. For the dot printer update the label to use html-like syntax.

Closes #7346

CC: @dgutson


pylint/pyreverse/dot_printer.py
```diff
@@ -92,17 +92,22 @@ def _build_label_for_node(self, properties: NodeProperties) -> str:
 
         # Add class attributes
         attrs: list[str] = properties.attrs or []
-        attrs_string = r"\l".join(attr.replace("|", r"\|") for attr in attrs)
-        label = rf"{{{label}|{attrs_string}\l|"
+        attrs_string = r'<br ALIGN="LEFT"/>'.join(
```
Pierre-Sassoulas, line 95: ``<br ALIGN="LEFT"/>`` could be a class constant instead (and used at line 110 too).

Pierre-Sassoulas (changes requested):
Thank you, it looks pretty good to me, but I'm not a pyreverse expert :)

DudeNr33:
Hi @qequ,
thank you for the contribution!
Sorry for the late response, but my internet at home is down for almost a week now.

Can you also add this feature for the `MermaidJSPrinter`?
For Mermaid, adding a `*` behind the parentheses should do the trick:

```
classDiagram
    class Duck{
      +String beakColor
      +swim()*
      +quack()
    }
```
renders as:
```mermaid
classDiagram
    class Duck{
      +String beakColor
      +swim()*
      +quack()
    }
```

pylint/pyreverse/dot_printer.py
qequ: Added class constant :+1:

qequ:
@DudeNr33 added italics for mermaidjs :+1:

pylint/pyreverse/dot_printer.py
```diff
@@ -10,13 +10,19 @@
...
 from pathlib import Path
 
 from astroid import nodes
 
 from pylint.pyreverse.printer import EdgeType, Layout, NodeProperties, NodeType, Printer
 from pylint.pyreverse.utils import get_annotation_label
 
+
+class HTMLLabels(Enum):
+    LINEBREAK_LEFT = '<br ALIGN="LEFT"/>'
```
Pierre-Sassoulas, line 23: It's only used in the dot printer so let's keep it in the dot printer as a nested class :)

qequ: @Pierre-Sassoulas do I have to change the `HTMLLabels` class?

Pierre-Sassoulas: You can make a single variable instead of an enum too, but both work.

DudeNr33 (approved):
Thank you for this addition. Looks good to me! 👍

pylint/pyreverse/dot_printer.py
qequ: @Pierre-Sassoulas  I think it's better to have it as a class in case more labels are added in the future

Pierre-Sassoulas (approved):
Thank you for implementing this feature and congratulation on your first contribution to pyreverse 🎉

PR #7444, opened 2022-09-09
Finish typing related to the ``config`` directory

DanielNoord:

## Type of Changes

## Description

@Pierre-Sassoulas Since these two modules are deprecated I thought it would be best to just `Any` a lot. We don't really need the typing anyway as most of it is uncovered currently and no longer used.

With this PR I got to `Found 227 errors in 35 files (checked 286 source files)` running in `strict` mode with one check disabled. It seems we're getting there 😄 🎉

### DIFF
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -646,7 +646,9 @@ class _ArgumentsManager:
             )
         return self._arg_parser.format_help()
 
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (approved):
Make sense, we'll have to prepare 3.0 and remove all the deprecated stuff at some point, we don't have a clear planning for this though.

PR #7446, opened 2022-09-09
Finish typing of ``variables``

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING, Any, NamedTuple
 
 import astroid
 from astroid import extract_node, nodes
+from astroid.nodes import _base_nodes
 from astroid.typing import InferenceResult
 
 from pylint.checkers import BaseChecker, utils
@@ -120,7 +121,7 @@ class VariableVisitConsumerAction(Enum):
     RETURN = 1
 
 
-def _is_from_future_import(stmt, name):
+def _is_from_future_import(stmt: nodes.ImportFrom, name: str) -> bool | None:
     """Check if the name is a future import from another module."""
     try:
         module = stmt.do_import_module(stmt.modname)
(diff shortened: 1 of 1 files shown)


pylint/checkers/variables.py
```diff
@@ -2402,21 +2412,28 @@ def _check_is_unused(
 
             self.add_message(message_name, args=name, node=stmt)
 
-    def _is_name_ignored(self, stmt, name):
+    def _is_name_ignored(
+        self, stmt: nodes.NodeNG, name: str
+    ) -> re.Pattern[str] | re.Match[str] | None:
```
Pierre-Sassoulas, line 2417: This is relying on arcane knowledge around how python ``and`` operator work (the like of which should be limited to python brain teaser imo). But thanks for typing it. Do you think it's fixable easily ?

DanielNoord: For now, not really. I actually mistyped it initially because I misremembered how the operator worked.

I think for now it is probably fine? I don't like touching this code without a specific need. There is so much going on in this file..

pylint/checkers/variables.py
```diff
@@ -2402,21 +2412,28 @@ def _check_is_unused(
...
         if (
             isinstance(stmt, nodes.AssignName)
             and isinstance(stmt.parent, nodes.Arguments)
             or isinstance(stmt, nodes.Arguments)
         ):
-            regex = self.linter.config.ignored_argument_names
+            regex: re.Pattern[str] = self.linter.config.ignored_argument_names
         else:
             regex = authorized_rgx
         return regex and regex.match(name)
```
Pierre-Sassoulas, line 2427:
```suggestion
        # See https://stackoverflow.com/a/47007761/2519059 to
        # understand what this function return. Please do NOT use
        # this elsewhere, this is confusing for no benefit
        return regex and regex.match(name)
```

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas (approved):

PR #7447, opened 2022-09-09
Finish typing across the codebase

DanielNoord:

## Type of Changes

## Description

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -697,21 +697,29 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
         self._imports_stack.append((node, importedname))
 
     @staticmethod
-    def _is_fallback_import(node, imports):
+    def _is_fallback_import(
+        node: ImportNode, imports: list[tuple[ImportNode, str]]
+    ) -> bool:
         imports = [import_node for (import_node, _) in imports]
         return any(astroid.are_exclusive(import_node, node) for import_node in imports)
 
-    def _check_imports_order(self, _module_node: nodes.Module):
+    def _check_imports_order(
+        self, _module_node: nodes.Module
+    ) -> tuple[
+        list[tuple[ImportNode, str]],
+        list[tuple[ImportNode, str]],
+        list[tuple[ImportNode, str]],
+    ]:
         """Checks imports of module `node` are grouped by category.
 
         Imports must follow this order: standard, 3rd party, local
         """
-        std_imports = []
-        third_party_imports = []
-        first_party_imports = []
+        std_imports: list[tuple[ImportNode, str]] = []
+        third_party_imports: list[tuple[ImportNode, str]] = []
+        first_party_imports: list[tuple[ImportNode, str]] = []
         # need of a list that holds third or first party ordered import
-        external_imports = []
-        local_imports = []
+        external_imports: list[tuple[ImportNode, str]] = []
+        local_imports: list[tuple[ImportNode, str]] = []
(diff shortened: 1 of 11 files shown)


pylint/checkers/imports.py
```diff
@@ -697,21 +697,29 @@ def _record_import(
...
+        node: ImportNode, imports: list[tuple[ImportNode, str]]
+    ) -> bool:
         imports = [import_node for (import_node, _) in imports]
         return any(astroid.are_exclusive(import_node, node) for import_node in imports)
 
-    def _check_imports_order(self, _module_node: nodes.Module):
+    def _check_imports_order(
+        self, _module_node: nodes.Module
+    ) -> tuple[
+        list[tuple[ImportNode, str]],
```
Pierre-Sassoulas, line 709: Maybe we could create an alias for that one ? It's used ~~a lot~~ (after full review: only in this file, but it's still arguable imo).

pylint/config/_pylint_config/__init__.py
```diff
@@ -7,5 +7,7 @@
 Everything in this module is private.
 """
 
-from pylint.config._pylint_config.main import _handle_pylint_config_commands  # noqa
-from pylint.config._pylint_config.setup import _register_generate_config_options  # noqa
+from pylint.config._pylint_config.main import _handle_pylint_config_commands
+from pylint.config._pylint_config.setup import _register_generate_config_options
+
+__all__ = ("_handle_pylint_config_commands", "_register_generate_config_options")
```
Pierre-Sassoulas, line 13: Weren't you against adding private api in ``__all__`` previously ? I think it's ok myself, it's also a tool to make our live easier and imports clearer even the private one. But we might want to do that elsewhere too in that case.

pylint/lint/pylinter.py
```diff
@@ -939,9 +939,7 @@ def _astroid_module_checker(
         walker = ASTWalker(self)
         _checkers = self.prepare_checkers()
         tokencheckers = [
-            c
-            for c in _checkers
-            if isinstance(c, checkers.BaseTokenChecker) and c is not self
+            c for c in _checkers if isinstance(c, checkers.BaseTokenChecker)
```
Pierre-Sassoulas, line 942: Nice simplification

tests/message/unittest_message_definition.py
```diff
@@ -21,7 +21,7 @@
         ("W12345", "Invalid message id 'W12345'"),
     ],
 )
-def test_create_invalid_message_type(msgid, expected):
+def test_create_invalid_message_type(msgid: str, expected: str) -> None:
```
Pierre-Sassoulas, line 24: As the code owner of ``tests/message``, I approve of this change ! (I think we bothered areveny and DudeNr33 with the typing changes because of codeowner but it should be a one time thing)

Pierre-Sassoulas (changes requested):

pylint/checkers/imports.py
DanielNoord: Yeah we could, but it is not too long imo. I like making typing explicit where possible because I don't need to look up what an alias refers to when I try and understand a piece of code.

pylint/config/_pylint_config/__init__.py
DanielNoord: Yeah I still am. But `mypy` complained about this and we have a docstring + module name that indicate that this is private. If anybody still uses it is their fault.

Pierre-Sassoulas (approved):

PR #7465, opened 2022-09-15
Use stacklevel=2 in warnings.warn calls to DeprecationWarning where makes sense.

yilei:
With this change, relevant `DeprecationWarnings` are now raised with `stacklevel=2`, so they have the callsite attached in the message.

## Type of Changes

I'm not sure this fits anything below?

|     | Type                   |
| --- | ---------------------- |
|    | :bug: Bug fix          |
|    | :sparkles: New feature |
|    | :hammer: Refactoring   |
|    | :scroll: Docs          |

Closes #7463

Issue #7463, opened 2022-09-14
DeprecationWarning raised by Pylint's own code should generally have stacklevel=2


yilei:
### Bug description

Looks like recently Pylint has added many `DeprecationWarning` in preparation for 3.0 (great!).

However, the `warnings.warn` call doesn't specify a `stacklevel=` thus default to `1`. Example: https://github.com/PyCQA/pylint/blob/27925396c85ffc443243589a0c03e430de9d7fa5/pylint/config/arguments_provider.py#L74

This makes the warning show up at the `warnings.warn` itself, like:

```
pylint/config/arguments_provider.py:74: DeprecationWarning: The checker-specific config attribute has been deprecated. Please use 'linter.config' to access the global configuration object.
```

It doesn't tell us where this deprecated method is called, making custom tooling that's built on top of Pylint difficult to migrate when 3.0 comes.

Most of the `warnings.warn` should use a `stacklevel=2` instead.

### Command used

```shell
N/A
```

### Pylint output

```shell
N/A
```

### Expected behavior

N/A

### Pylint version

```shell
Pylint at head
```

yilei:
FYI-- I'm working on a PR for this.

PR #7469, opened 2022-09-15
Provide more context in these two DeprecationWarning.

yilei:
Since this is called a few frame levels inside Pylint, providing the context of what optdict= can help with finding where the option originally comes from.

This is a follow-up to #7465

## Type of Changes

| ✓   | Maintenance          |

## Description

This improves the deprecation warnings similar to #7463.

### DIFF
--- a/pylint/config/utils.py
+++ b/pylint/config/utils.py
@@ -42,7 +42,8 @@ def _convert_option_to_argument(
     if "level" in optdict and "hide" not in optdict:
         warnings.warn(
             "The 'level' key in optdicts has been deprecated. "
-            "Use 'hide' with a boolean to hide an option from the help message.",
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):

DanielNoord (approved):
Makes sense!

Just out of curiosity, which project are you working on that you noticed these issues? We might make more deprecations before `3.0` and it might be worthwhile to discuss those with dependents 😄

yilei:
It's an internal tool at work we build on top of Pylint and we use that instead of Pylint itself. It includes some additional options & lint checkers, which was written with the old optparse based providers.

I'm currently upgrading the version from `v2.13.9` to `v2.15.2`, and addressing the deprecation warnings. Please do keep issuing `DeprecationWarning` as they are very valuable in preparation for 3.0 :)

PR #7276, opened 2022-08-08
Improve docs clarity on loading custom checkers

daogilvie:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :scroll: Docs          |

### Short
This PR just updates the documentation page on writing custom checkers. It adds
information about some restrictions between how `init-hook` changes the loading
path and when you can use that to import your custom module.
### Long (in the commit message)
Whilst the docs on testing custom checkers do touch on the fact the
module has to be in your load path, there are some restrictions that
were not made clear before.
Specifically, the init-hook configuration item (which is often used
explicitly to change the sys.path) is not automatically sufficient. If
the init hook is in a file, but the load-plugins argument is passed in
the CLI, then the plugins are loaded before the init-hook has run, and
so will not be imported. In this case, the failure can also be silent,
so the outcome is a lint that will simply not contain any information
from the checker, as opposed to a run that will error.
This corner case may merit fixing in a separate PR, that is to be
confirmed.

Closes #7264

Issue #7264, opened 2022-08-04
Cannot get pylint to register custom checker, is there a missing step in the docs?


daogilvie:
### Question

Hi! First: Thanks for supporting and maintaining such a great ecosystem!

My question: I'm trying to make a simple checker around some custom formatting in log strings. I have followed every tutorial I can find from within the last few years, and naturally looked at the guide [in the documentation](https://pylint.pycqa.org/en/v2.14.5/development_guide/how_tos/custom_checkers.html).

However, I am finding that the `register` function, described as the way to get your checker actually loaded into pylint, simply is not being called. The module is being loaded, for sure — I can see breakpoints in the top level of the module being hit, but not in the `register` function itself. I've tried setting the `--load-plugins` argument, I've tried using the `load_plugins` setting in pylintrc.

Is there an additional step or constraint missing in the implementation of custom checkers from the docs? Or am I simply making a mistake?

Below is the code for my absolute bare-bones first attempt.

Stored in `testchecker.py`:
```python
from astroid import nodes
from pylint.checkers import BaseChecker

breakpoint() # HIT!

class SimpleCheck(BaseChecker):
    """Simple"""
    name = "simple-test-check"
    msgs = {
            "W6900": ("Test", "test-check", "Just a test")
            }
    options = ( ("fake-option", {
                "default": False,
                "type": "yn", # EDIT: bool here was not allowed
                "metavar": "<y or n>",
                "help": "Some fake option"}
                ),
            )

    def visit_import(self, _: nodes.Import) -> None:
        breakpoint() # NOT HIT!

def register(linter):
    breakpoint() # NOT HIT!
    linter.register_checker(SimpleCheck(linter))
```

My file I am linting, `lintme.py`:
```python
"""MODULE DOCSTRING"""
import logging

logger = logging.getLogger(__name__)

logger.info("hello world")
```
Both files are in the same directory, in which I execute:
```bash
pylint --load-plugins=testchecker lintme.py
```

```
$ pylint --version
pylint 2.14.5
astroid 2.11.7
Python 3.10.4 (main, May 19 2022, 08:51:27) [Clang 13.1.6 (clang-1316.0.21.2.5)]
$ uname -a
Darwin OVO4941L.ovoenergy.local 21.6.0 Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:25 PDT 2022; root:xnu-8020.140.41~1/RELEASE_X86_64 x86_64
```

Thanks again!

### Documentation for future user

[...]

PR #7475, opened 2022-09-16
Add and fix tests for init-hook based plugin load

daogilvie:
## Type of Changes

## Description

This changes 4 existing tests, due to a misunderstanding of the author (past me) in a couple of the details of those tests. Specifically, we now copy a single python file to use as our plugin, and make sure to correctly check for the name as given in the checker classes. We also make sure not to accidentally load the old copy of the plugin, which apparently sits in a directory that is already on the system path.
There is also a single new test, which covers the cases of loading a plugin with ``init-hook`` magic, both specified in an rc file, but in different orders.

This and #7276 are the only outstanding PRs preventing closure of #7264.

Refs #7264.

### DIFF
--- a/tests/lint/unittest_lint.py
+++ b/tests/lint/unittest_lint.py
@@ -669,7 +669,7 @@ def test_load_plugin_path_manipulation_case_3() -> None:
 @pytest.mark.usefixtures("pop_pylintrc")
 def test_load_plugin_pylintrc_order_independent() -> None:
     """Test that the init-hook is called independent of the order in a config file.
-    
+
     We want to ensure that any path manipulation in init hook
     that means a plugin can load (as per GitHub Issue #7264 Cases 4+7)
     runs before the load call, regardless of the order of lines in the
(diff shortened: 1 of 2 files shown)


tests/lint/unittest_lint.py
```diff
@@ -662,49 +666,135 @@ def test_load_plugin_path_manipulation_case_3() -> None:
         sys.path.remove(home_path)
 
 
+@pytest.mark.usefixtures("pop_pylintrc")
+def test_load_plugin_pylintrc_order_independent() -> None:
+    """We want to ensure that any path manipulation in init hook
+    that means a plugin can load (as per GitHub Issue #7264 Cases 4+7)
+    runs before the load call, regardless of the order of lines in the
+    pylintrc file.
```
DanielNoord, line 674:
```suggestion
    """Test that the init-hook is called independent of the order in a config file.
    
    We want to ensure that any path manipulation in init hook
    that means a plugin can load (as per GitHub Issue #7264 Cases 4+7)
    runs before the load call, regardless of the order of lines in the
    pylintrc file.
```

DanielNoord (dismissed):

DanielNoord (approved):

Pierre-Sassoulas (approved):
Thank you @daogilvie !

PR #7481, opened 2022-09-17
Extend consider-using-join to detect non-empty separators

### LINKED ISSUE #6639: ``consider-using-join`` raised for code equivalent to ``", ".join("a", "b", "c")``
### Current problem

For

```python
result = ''
for number in ['1', '2', '3']:
    result += number  # [consider-using-join]
```

Where the final result would be ``"".join(['1', '2', '3'])`` we already raise a ``consider-using-join``. Sometime function that could be join instead are sneakier, for example:

```python
result = ''
for number in ['1', '2', '3']:
    result += f"{number}, "
result = result[:-2]
```
Could be  ``", ".join(['1', '2', '3'])``.

### Desired solution

``consider-using-join`` detecting possible ``join`` using a non-empty string.

lggruspe:
## Type of Changes

## Description

`consider-using-join` now detects more refactoring opportunities.

Example:

```python
result = ''
for number in ['1', '2', '3']:
    result += f"{number}, "
```

Closes #6639


lggruspe:
Should I add astroid to [`tests/primer/packages_to_lint_batch_one.json`](https://github.com/PyCQA/pylint/blob/main/tests/primer/packages_to_lint_batch_one.json)? I'm not sure how to fix the cache error.

Pierre-Sassoulas:
> I'm not sure how to fix the cache error.

I'm not sure either, it could be due to astroid 2.12.10 being released now and the dependencies not being exactly the same. It's going to calm down when we merge the maintenance branch with updated astroid. I relaunched the primer to see if it fix the problem.

doc/whatsnew/fragments/6639.other
```diff
@@ -0,0 +1,4 @@
+``consider-using-join`` enhancement.
+It can now be emitted in cases where the string separator is non-empty.
+
+Closes #6639
```
Pierre-Sassoulas, line 4:
```suggestion
``consider-using-join`` can now be emitted for non-empty string separators

Closes #6639
```

tests/functional/c/consider/consider_join.py
```diff
@@ -17,6 +17,24 @@
...
+    result += f'{number}c'  # [consider-using-join]
+assert result == 'a1c2c3c'
+assert result == 'a' + 'c'.join(['1', '2', '3']) + 'c'
+
+result = 'a'
+for number in ['1', '2', '3']:
+    result += f'b{number}c'  # [consider-using-join]
+assert result == 'ab1cb2cb3c'
+assert result == 'ab' + 'cb'.join(['1', '2', '3']) + 'c'
+
```
Pierre-Sassoulas, line 37:
```suggestion
result = ''
for number in ['1', '2', '3']:
    result += number  # [consider-using-join]


result = ''
for number in ['1', '2', '3']:
    result += f"{number}, "  # [consider-using-join]
result = result[:-2]
```
Code example from the original issue.

Pierre-Sassoulas (commented):
I'm not sure we should raise something when the possible solution is ``assert result == 'a' + 'c'.join(['1', '2', '3']) + 'c'``, because it's not that easier to understand than the for loop. But I agree that this simplify things a lot from our side (in particular for the example I gave in the original issue, that we won't have to handle). I added the need review label to gather opinions from other maintainers.

tests/functional/c/consider/consider_join.py
jacobtylerwalls: I think these test cases cover the same ground as the original ones, so I'm fine with the PR as is.

jacobtylerwalls (dismissed):
Pierre asks a question worth considering, but I think there's no way to programmatically decide what's readable, the variety is endless. I take the point of the check to be just to "consider" using join.

PR looks good!

Pierre-Sassoulas (dismissed):

jacobtylerwalls (approved):

PR #7482, opened 2022-09-17
Update W1402 message example to be an actually working code

jack1142:
## Type of Changes

## Description

Replaces erroneous usage of `.format()` on the `bytes` object with %-formatting in the message example for W1402 (anomalous-unicode-escape-in-string) so that the good example can actually be run.


jack1142:
Hmm, do I need to make an issue or can I just refer to this PR in the changelog fragment file?

mbyrnepr2:
Thank you @jack1142. No fragment required for these code-examples. I can skip the CI check for it.

jack1142:
Alright, I'll remove the file from the changeset then, thanks.

mbyrnepr2:
Ah I see Pierre already added the "so news" label

DanielNoord (approved):

PR #7497, opened 2022-09-19
Use qualified name when checking for overgeneral exceptions

### LINKED ISSUE #7495: overgeneral-exceptions doesn't work for exceptions outside builtins
### Bug description

```py

class GeneralException(Exception):
    pass


try:
    raise GeneralException('...')
except GeneralException:
    ...
# for comparison, pylint triggers on this
except Exception:
    ...
```

### Configuration

_No response_

### Command used

```shell
pylint repro.py --overgeneral-exceptions BaseException,Exception,GeneralException
```

### Pylint output

```shell
************* Module repro
repro.py:12:7: W0703: Catching too general exception Exception (broad-except)

Your code has been rated at 8.75/10 (previous run: 8.75/10, +0.00)
```

### Expected behavior

I expected to get `broad-except` message for line 9 (which is not happening regardless of whether `except Exception` is there or not).

This is also reproducible when making a whole package and referring to the exception class using `package.GeneralException` name (which I personally think would be the best way of specifying non-builtin exceptions).

This seems to vastly limit the usefulness of `--overgeneral-exceptions`.

### Pylint version

```shell
pylint 2.15.3
astroid 2.12.10
```

### OS / Environment

Kubuntu 22.04.1 LTS

### Additional dependencies

_No response_

jack1142:
## Type of Changes

## Description

Allow specifying non-builtin exceptions in overgeneral-exception option using exception's qualified name.

Additionally, tries to deprecate specifying built-in exception names without dots. Small problem - I don't know how to tell pylint to warn about deprecation as `warnings.warn(...)` does not seem to be a valid way to do that.

Closes #7495

### DIFF
--- a/doc/whatsnew/fragments/7495.bugfix
+++ b/doc/whatsnew/fragments/7495.bugfix
@@ -1,4 +1,4 @@
-Allow specifying non-builtin exceptions in overgeneral-exception option
-using exception's qualified name.
+Allow specifying non-builtin exceptions in the ``overgeneral-exception`` option
+using an exception's qualified name.
 
 Closes #7495
--- a/doc/whatsnew/fragments/7495.bugfix
+++ b/doc/whatsnew/fragments/7495.bugfix
@@ -1,0 +1,4 @@
+Allow specifying non-builtin exceptions in overgeneral-exception option
+using exception's qualified name.
+
+Closes #7495
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import builtins
 import inspect
+import warnings
 from collections.abc import Generator
 from typing import TYPE_CHECKING, Any
 
@@ -288,6 +289,15 @@ class ExceptionsChecker(checkers.BaseChecker):
 
     def open(self) -> None:
         self._builtin_exceptions = _builtin_exceptions()
+        if any(
+            "." in exc_name for exc_name in self.linter.config.overgeneral_exceptions
+        ):
+            warnings.warn(
+                "Specifying exception names in overgeneral-exceptions option without"
+                " module name has been deprecated."
+                " Use fully qualified name (e.g. builtins.Exception) instead.",
+                DeprecationWarning,
+            )
         super().open()
 
     @utils.only_required_for_messages(
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -293,7 +293,7 @@ class ExceptionsChecker(checkers.BaseChecker):
                     "Specifying exception names in overgeneral-exceptions option"
                     " without module name is deprecated and support for it"
                     " will be removed in pylint 3.0."
-                    f" Use fully qualified name (e.g. builtins.{exc_name}) instead.",
+                    f" Use fully qualified name (e.g. 'builtins.{exc_name}') instead.",
                     category=UserWarning,
                     filename="pylint: Command line or configuration file",
                     lineno=1,
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -287,15 +287,14 @@ class ExceptionsChecker(checkers.BaseChecker):
 
     def open(self) -> None:
         self._builtin_exceptions = _builtin_exceptions()
-        if any(
-            "." in exc_name for exc_name in self.linter.config.overgeneral_exceptions
-        ):
-            warnings.warn(
-                "Specifying exception names in overgeneral-exceptions option without"
-                " module name has been deprecated."
-                " Use fully qualified name (e.g. builtins.Exception) instead.",
-                DeprecationWarning,
-            )
+        for exc_name in self.linter.config.overgeneral_exceptions:
+            if "." not in exc_name:
+                warnings.warn(
+                    "Specifying exception names in overgeneral-exceptions option"
+                    " without module name has been deprecated."
+                    f" Use fully qualified name (e.g. builtins.{exc_name}) instead.",
+                    DeprecationWarning,
+                )
         super().open()
 
     @utils.only_required_for_messages(
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -59,7 +59,7 @@ def _is_raising(body: list[nodes.NodeNG]) -> bool:
     return any(isinstance(node, nodes.Raise) for node in body)
 
 
-OVERGENERAL_EXCEPTIONS = ("BaseException", "Exception")
+OVERGENERAL_EXCEPTIONS = ("builtins.BaseException", "builtins.Exception")
 
 MSGS: dict[str, MessageDefinitionTuple] = {
     "E0701": (
@@ -593,16 +593,19 @@ class ExceptionsChecker(checkers.BaseChecker):
                                 confidence=INFERENCE,
                             )
                     if (
-                        exception.name in self.linter.config.overgeneral_exceptions
+                        exception.qname() in self.linter.config.overgeneral_exceptions
+                        # undotted name, deprecated
+                        or "." not in exception.name
+                        and exception.name in self.linter.config.overgeneral_exceptions
                         and exception.root().name == utils.EXCEPTIONS_MODULE
-                        and not _is_raising(handler.body)
                     ):
-                        self.add_message(
-                            "broad-exception-caught",
-                            args=exception.name,
-                            node=handler.type,
-                            confidence=INFERENCE,
-                        )
+                        if not _is_raising(handler.body):
+                            self.add_message(
+                                "broad-exception-caught",
+                                args=exception.name,
+                                node=handler.type,
+                                confidence=INFERENCE,
+                            )
 
                     if exception in exceptions_classes:
                         self.add_message(
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -221,7 +221,7 @@ class ExceptionRaiseRefVisitor(BaseVisitor):
                     "broad-exception-raised",
                     args=exception.name,
                     node=self._node,
-                    confidence=HIGH,
+                    confidence=INFERENCE,
                 )
 
     def visit_call(self, node: nodes.Call) -> None:
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -208,15 +208,9 @@ class ExceptionRaiseRefVisitor(BaseVisitor):
             return
 
         for _, exception in exceptions:
-            if isinstance(exception, astroid.Instance) and utils.inherit_from_std_ex(
-                exception
-            ):
-                exception = exception._proxied
-
-            if not isinstance(exception, nodes.ClassDef):
-                continue
-
-            if self._checker._is_overgeneral_exception(exception):
+            if isinstance(
+                exception, nodes.ClassDef
+            ) and self._checker._is_overgeneral_exception(exception):
                 self._checker.add_message(
                     "broad-exception-raised",
                     args=exception.name,
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -602,7 +602,7 @@ class ExceptionsChecker(checkers.BaseChecker):
                             )
                     if (
                         exception.qname() in self.linter.config.overgeneral_exceptions
-                        # not a qualified name, deprecated
+                        # TODO: 3.0: not a qualified name, deprecated
                         or "." not in exception.name
                         and exception.name in self.linter.config.overgeneral_exceptions
                         and exception.root().name == utils.EXCEPTIONS_MODULE
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -200,7 +200,7 @@ class ExceptionRaiseRefVisitor(BaseVisitor):
             self._checker.add_message(
                 "notimplemented-raised", node=self._node, confidence=HIGH
             )
-        elif node.name in OVERGENERAL_EXCEPTIONS:
+        elif node.name in self._checker.linter.config.overgeneral_exceptions:
             self._checker.add_message(
                 "broad-exception-raised",
                 args=node.name,
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -300,7 +300,7 @@ class ExceptionsChecker(checkers.BaseChecker):
         for exc_name in self.linter.config.overgeneral_exceptions:
             if "." not in exc_name:
                 warnings.warn_explicit(
-                    "Specifying exception names in overgeneral-exceptions option"
+                    "Specifying exception names in the overgeneral-exceptions option"
                     " without module name is deprecated and support for it"
                     " will be removed in pylint 3.0."
                     f" Use fully qualified name (maybe 'builtins.{exc_name}' ?) instead.",
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -309,7 +309,7 @@ class ExceptionsChecker(checkers.BaseChecker):
                     "Specifying exception names in overgeneral-exceptions option"
                     " without module name is deprecated and support for it"
                     " will be removed in pylint 3.0."
-                    f" Use fully qualified name (e.g. 'builtins.{exc_name}') instead.",
+                    f" Use fully qualified name (maybe 'builtins.{exc_name}' ?) instead.",
                     category=UserWarning,
                     filename="pylint: Command line or configuration file",
                     lineno=1,
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -609,14 +609,13 @@ class ExceptionsChecker(checkers.BaseChecker):
                         or "." not in exception.name
                         and exception.name in self.linter.config.overgeneral_exceptions
                         and exception.root().name == utils.EXCEPTIONS_MODULE
-                    ):
-                        if not _is_raising(handler.body):
-                            self.add_message(
-                                "broad-exception-caught",
-                                args=exception.name,
-                                node=handler.type,
-                                confidence=INFERENCE,
-                            )
+                    ) and not _is_raising(handler.body):
+                        self.add_message(
+                            "broad-exception-caught",
+                            args=exception.name,
+                            node=handler.type,
+                            confidence=INFERENCE,
+                        )
 
                     if exception in exceptions_classes:
                         self.add_message(
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -200,13 +200,29 @@ class ExceptionRaiseRefVisitor(BaseVisitor):
             self._checker.add_message(
                 "notimplemented-raised", node=self._node, confidence=HIGH
             )
-        elif node.name in self._checker.linter.config.overgeneral_exceptions:
-            self._checker.add_message(
-                "broad-exception-raised",
-                args=node.name,
-                node=self._node,
-                confidence=HIGH,
-            )
+            return
+
+        try:
+            exceptions = list(_annotated_unpack_infer(node))
+        except astroid.InferenceError:
+            return
+
+        for _, exception in exceptions:
+            if isinstance(exception, astroid.Instance) and utils.inherit_from_std_ex(
+                exception
+            ):
+                exception = exception._proxied
+
+            if not isinstance(exception, nodes.ClassDef):
+                continue
+
+            if self._checker._is_overgeneral_exception(exception):
+                self._checker.add_message(
+                    "broad-exception-raised",
+                    args=exception.name,
+                    node=self._node,
+                    confidence=HIGH,
+                )
 
     def visit_call(self, node: nodes.Call) -> None:
         if isinstance(node.func, nodes.Name):
@@ -603,13 +619,9 @@ class ExceptionsChecker(checkers.BaseChecker):
                                 args=msg,
                                 confidence=INFERENCE,
                             )
-                    if (
-                        exception.qname() in self.linter.config.overgeneral_exceptions
-                        # TODO: 3.0: not a qualified name, deprecated
-                        or "." not in exception.name
-                        and exception.name in self.linter.config.overgeneral_exceptions
-                        and exception.root().name == utils.EXCEPTIONS_MODULE
-                    ) and not _is_raising(handler.body):
+                    if self._is_overgeneral_exception(exception) and not _is_raising(
+                        handler.body
+                    ):
                         self.add_message(
                             "broad-exception-caught",
                             args=exception.name,
@@ -626,6 +638,15 @@ class ExceptionsChecker(checkers.BaseChecker):
                         )
 
                 exceptions_classes += [exc for _, exc in exceptions]
+
+    def _is_overgeneral_exception(self, exception: nodes.ClassDef) -> bool:
+        return (
+            exception.qname() in self.linter.config.overgeneral_exceptions
+            # TODO: 3.0: not a qualified name, deprecated
+            or "." not in exception.name
+            and exception.name in self.linter.config.overgeneral_exceptions
+            and exception.root().name == utils.EXCEPTIONS_MODULE
+        )
 
 
 def register(linter: PyLinter) -> None:
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -60,8 +60,6 @@ def _is_raising(body: list[nodes.NodeNG]) -> bool:
     return any(isinstance(node, nodes.Raise) for node in body)
 
 
-OVERGENERAL_EXCEPTIONS = ("builtins.BaseException", "builtins.Exception")
-
 MSGS: dict[str, MessageDefinitionTuple] = {
     "E0701": (
         "Bad except clauses order (%s)",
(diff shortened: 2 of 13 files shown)


Pierre-Sassoulas (commented):
Look good thank you ! I think this is a breaking change that we need to put in 3.0 though

jack1142:
It keeps compatibility with the old way (names without dots are handled as they've been before and names with dots weren't supported in any way before) so I think it should be fine? The long-term goal is to remove the old way in 3.0.

The problem I've run into is that I don't really know how I should make pylint actually show the deprecation warning to the user (`warnings.warn` that's currently in here doesn't seem to work) which is why this PR is a draft.

doc/whatsnew/fragments/7495.bugfix
```diff
@@ -0,0 +1,4 @@
+Allow specifying non-builtin exceptions in overgeneral-exception option
+using exception's qualified name.
```
DanielNoord, line 2:
```suggestion
using an exception's qualified name.
```

doc/whatsnew/fragments/7495.bugfix
```diff
@@ -0,0 +1,4 @@
+Allow specifying non-builtin exceptions in overgeneral-exception option
```
DanielNoord, line 1:
```suggestion
Allow specifying non-builtin exceptions in the ``overgeneral-exception`` option
```

pylint/checkers/exceptions.py
```diff
@@ -59,7 +60,7 @@ def _is_raising(body: list[nodes.NodeNG]) -> bool:
     return any(isinstance(node, nodes.Raise) for node in body)
 
 
-OVERGENERAL_EXCEPTIONS = ("BaseException", "Exception")
+OVERGENERAL_EXCEPTIONS = ("builtins.BaseException", "builtins.Exception")
```
DanielNoord, line 63: Could you move this to L248? We don't instantiate this class more than once so this should just be a value in the `dict` like all other options do.

pylint/checkers/exceptions.py
```diff
@@ -254,6 +255,15 @@ class ExceptionsChecker(checkers.BaseChecker):
...
     def open(self) -> None:
         self._builtin_exceptions = _builtin_exceptions()
+        if any(
+            "." in exc_name for exc_name in self.linter.config.overgeneral_exceptions
+        ):
+            warnings.warn(
+                "Specifying exception names in overgeneral-exceptions option without"
+                " module name has been deprecated."
+                " Use fully qualified name (e.g. builtins.Exception) instead.",
+                DeprecationWarning,
```
DanielNoord, line 265: I have recently learned about `stacklevel`: https://github.com/PyCQA/pylint/pull/7465

We should probably use it here as well.

pylint/checkers/exceptions.py
```diff
@@ -254,6 +255,15 @@ class ExceptionsChecker(checkers.BaseChecker):
 
     def open(self) -> None:
         self._builtin_exceptions = _builtin_exceptions()
+        if any(
```
DanielNoord, line 258: Shouldn't this be the other way around? As in `not any`?

Either way, it might even be prettier to do `for exc_name in ...` as we can than create a warning with a very specific exception that name that needs to be changed. We could even suggest `builtins.{exc_name}`.

pylint/checkers/exceptions.py
```diff
@@ -254,6 +255,15 @@ class ExceptionsChecker(checkers.BaseChecker):
...
+            "." in exc_name for exc_name in self.linter.config.overgeneral_exceptions
+        ):
+            warnings.warn(
```
DanielNoord, line 261: I think the issue you refer to is that `DeprecationWarnings` are not normally shown. You could check with `pytest` if they do show. A test for this would be good.

pylint/checkers/exceptions.py
```diff
@@ -550,13 +560,16 @@ def visit_tryexcept(self, node: nodes.TryExcept) -> None:
                                 "bad-except-order", node=handler.type, args=msg
                             )
                     if (
-                        exception.name in self.linter.config.overgeneral_exceptions
+                        exception.qname() in self.linter.config.overgeneral_exceptions
+                        # not a qualified name, deprecated
```
DanielNoord, line 564: Could you add a `# TODO: 3.0: ...` about removal in `3.0`?

tests/functional/b/broad_except.rc
```diff
@@ -0,0 +1,4 @@
+[EXCEPTIONS]
+overgeneral-exceptions=builtins.BaseException,
+                       builtins.Exception,
+                       functional.b.broad_except.CustomBroadException
```
DanielNoord, line 4: 👍

pylint/checkers/exceptions.py
```diff
@@ -550,13 +560,16 @@ def visit_tryexcept(self, node: nodes.TryExcept) -> None:
...
+                        and exception.name in self.linter.config.overgeneral_exceptions
                         and exception.root().name == utils.EXCEPTIONS_MODULE
-                        and not _is_raising(handler.body)
                     ):
-                        self.add_message(
-                            "broad-except", args=exception.name, node=handler.type
-                        )
+                        if not _is_raising(handler.body):
+                            self.add_message(
+                                "broad-except", args=exception.name, node=handler.type
```
DanielNoord, line 571: While we're at it, would you mind adding `confidence=INFERENCE` here? We're updating the test output anyway so this is a good time to fix this as well 😄

DanielNoord (changes requested):

pylint/checkers/exceptions.py
jack1142: Using `python -Wdefault -m pylint ...` will make them show up but the problem is that absolutely nobody runs pylint like that (and they probably shouldn't, which I'll mention below).
People can see deprecation warnings when something they use in their own code is deprecated because their testing framework will enable deprecation warnings for them and so when the code is tested, they should see it. This is, however, not the case for running pylint because you don't run pylint under pytest but rather run it as its own program.

So IMO a proper solution here would be to either enable deprecation warnings for whole pylint (which has a very big problem of users seeing pylint's own deprecation warnings so I don't think it's actually a good solution) or have pylint show this warning by printing it or reporting it as a message, e.g. how something like `unrecognized-option` or `deprecated-pragma` works:
```
************* Module /home/ubuntu/work/pylint/pylintrc
pylintrc:1:0: E0015: Unrecognized option found: overgeneral-exceptions-typoed (unrecognized-option)
```
I think we would probably want it to show under `pylintrc` as `unrecognized-option` does *but* have it be under the `I` category as `deprecated-pragma`?

pylint/checkers/exceptions.py
jack1142: As explained in my other review comment about deprecation warnings, Python's warnings module is probably not the right tool here. But to address this, regardless of `stacklevel` we choose, it will look weird to the user as it will point at pylint's code (just at a different point in call stack):
```
/home/ubuntu/work/pylint/pylint/checkers/exceptions.py:259: DeprecationWarning: Specifying exception names in overgeneral-exceptions option without module name has been deprecated. Use fully qualified name (e.g. builtins.Exception) instead.
  warnings.warn(
```

pylint/checkers/exceptions.py
DanielNoord: Since this is something users can change themselves and they are not dependent on sub-packages fixing their reliance on deprecated functionality I think it makes sense to warn on each infraction.

pylint/checkers/exceptions.py
DanielNoord: Shall we use `UserWarning` instead? Or just print something to `stdout`? Using the `add_message` mechanism is a bit of a chore as we would to either 1) inject code specific to this option into the configuration parsing or 2) retrieve the place where the option was specified during a run (configuration file, and its name, vs. the CLI). We don't have a mechanism to do this currently.

jack1142: Maybe just print something to stdout? `UserWarning` can at best look like this:
```
Command line or configuration file:1: UserWarning: Specifying exception names in overgeneral-exceptions option without module name has been deprecated. Use fully qualified name (e.g. builtins.Exception) instead.
```
As the format of warnings is pre-defined based on `category`, `filename`, `lineno`, and `message` as passed to `warnings.warn_explicit()` (low-level version of `warnings.warn` which would allow us to at least specify custom `filename` and `lineno` and in the process also prevent code from showing up).

While with stdout we're free to format it however we like (which brings me to the next question, how do we want to format this?) and I'm actually not sure whether using `warnings` module would have any benefits over stdout in this case.

Also as I was typing this, I just thought of one more problem both of these run into... They break `--output-format=json` :grimacing: Maybe using `add_message` is inevitable?

DanielNoord: Does `UserWarning` also break `json`? That's something to look into for sure then!

But I don't really mind how the `UserWarning` looks. Isn't that what we are going for? Perhaps only add something about it being `pylint` that emits this?

jack1142: > But I don't really mind how the UserWarning looks. Isn't that what we are going for? Perhaps only add something about it being pylint that emits this?

We could have this:
```
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in overgeneral-exceptions option without module name has been deprecated. Use fully qualified name (e.g. builtins.Exception) instead.
```
Or subclass UserWarning:
```
Command line or configuration file:1: PylintUserWarning: Specifying exception names in overgeneral-exceptions option without module name has been deprecated. Use fully qualified name (e.g. builtins.Exception) instead.
```

> Does `UserWarning` also break `json`?

Yep. I feel like neither `UserWarning` nor `print` may end up working here.

DanielNoord: @Pierre-Sassoulas What do you think about this?

pylint/checkers/exceptions.py
jack1142: My point was that the calling code is going to be pylint regardless of how far up the stack we go (well ok, technically if you go high enough, you'll probably see stdlib's runpy module instead :)) as this deprecation warning is supposed to be shown when you run `pylint` cli - it's not about some plugin API in pylint.
I talked about it a bit more in the other review comment thread:
https://github.com/PyCQA/pylint/pull/7497#discussion_r975516077

Pierre-Sassoulas: Ha right. It's going to be fired when the old value without qualified name is used in a conf file or in CLI. Shouldn't it be a UserWarning instead ?

pylint/checkers/exceptions.py
Pierre-Sassoulas: Hmm in theory it should be a user warning, but we might need to do it as a breaking change for pylint 3.0 ?

DanielNoord: What is the normal process? UserWarning after the deprecation or before?

Pierre-Sassoulas: I was thinking of two possibilities: 
- UserWarning in 2.X and also 3.X if we modify this in 2.X
- Deprecation warning in 2.X then user warning in 3.X if we do this in 3.X

DanielNoord: We modify but in a backwards compatible way and we remove that functionality in `3.0`. So I think `UserWarning` makes sense then.

jack1142: What should be done about UserWarning breaking output of `--output-format=json`?
image

Pierre-Sassoulas: Is it really breaking or is it sent to stderr ?

jack1142: It's sent to stderr, I guess it's not breaking then?

jacobtylerwalls: > It's sent to stderr, I guess it's not breaking then?

That sounds right.

Pierre-Sassoulas:
Is this still a draft @jack1142 ?

pylint/checkers/exceptions.py
jack1142: Hopefully, this is resolved with the latest 2 commits.

jack1142:
> Is this still a draft @jack1142 ?

I was waiting for a resolution of the warning discussion before committing any changes. Should be ready for review now.

pylint/checkers/exceptions.py
```diff
@@ -550,13 +561,19 @@ def visit_tryexcept(self, node: nodes.TryExcept) -> None:
...
+                        # TODO: 3.0: not a qualified name, deprecated
+                        or "." not in exception.name
+                        and exception.name in self.linter.config.overgeneral_exceptions
                         and exception.root().name == utils.EXCEPTIONS_MODULE
-                        and not _is_raising(handler.body)
                     ):
-                        self.add_message(
-                            "broad-except", args=exception.name, node=handler.type
-                        )
+                        if not _is_raising(handler.body):
```
Pierre-Sassoulas, line 570: Why move this condition from the previous if ?

pylint/checkers/exceptions.py
```diff
@@ -550,13 +561,19 @@ def visit_tryexcept(self, node: nodes.TryExcept) -> None:
...
+                            self.add_message(
+                                "broad-except",
+                                args=exception.name,
+                                node=handler.type,
+                                confidence=INFERENCE,
```
Pierre-Sassoulas, line 575: 👍

pylint/checkers/exceptions.py
```diff
@@ -254,6 +253,18 @@ class ExceptionsChecker(checkers.BaseChecker):
 
     def open(self) -> None:
         self._builtin_exceptions = _builtin_exceptions()
+        for exc_name in self.linter.config.overgeneral_exceptions:
+            if "." not in exc_name:
+                warnings.warn_explicit(
+                    "Specifying exception names in overgeneral-exceptions option"
+                    " without module name is deprecated and support for it"
+                    " will be removed in pylint 3.0."
+                    f" Use fully qualified name (e.g. builtins.{exc_name}) instead.",
```
Pierre-Sassoulas, line 262:
```suggestion
                    f" Use fully qualified name (e.g. 'builtins.Exception' or 'mypackage.MyCustomException') instead.",
```
Maybe it's clearer as an example ?

Pierre-Sassoulas (commented):
Looks pretty refined already, good job !

pylint/checkers/exceptions.py
jack1142: This was suggested here:
https://github.com/PyCQA/pylint/pull/7497#discussion_r975003974

pylint/checkers/exceptions.py
jack1142: I think I was trying to avoid the line length limit but it doesn't seem like that was necessary, fixed.

pylint/checkers/exceptions.py
jack1142: For now, I added quotes to the exception name at least.

Pierre-Sassoulas (approved):
LGTM 👏

DanielNoord:
Shit.. This needs a rebase due to the conflicts. I'm sorry about that @jack1142.

I've been very busy. I promise that if you rebase this PR and tag me I'll make sure to merge it immediately after and not let this slip 😅

jack1142:
Damn, conflict with 4 files 😄 I will do it tomorrow as I'm a bit busy tonight, thanks for letting me know!

DanielNoord:
Yeah this is our bad probably. I have recently started a new job and have had less time for OS. This has messed a bit with out review/merge flow I think.

Like I said, make sure to ping my directly as I'll be more likely to see and follow up on it 😄

Pierre-Sassoulas:
Probably due to #7709, you're going to have to rename old broad-except and consider the new check too, sorry for letting this go stale.

pylint/checkers/exceptions.py
```diff
@@ -288,6 +287,18 @@ class ExceptionsChecker(checkers.BaseChecker):
 
     def open(self) -> None:
         self._builtin_exceptions = _builtin_exceptions()
+        for exc_name in self.linter.config.overgeneral_exceptions:
+            if "." not in exc_name:
+                warnings.warn_explicit(
+                    "Specifying exception names in overgeneral-exceptions option"
+                    " without module name is deprecated and support for it"
+                    " will be removed in pylint 3.0."
+                    f" Use fully qualified name (e.g. 'builtins.{exc_name}') instead.",
```
Pierre-Sassoulas, line 296:
```suggestion
                    f" Use fully qualified name (maybe 'builtins.{exc_name}' ?) instead.",
```

pylint/checkers/exceptions.py
```diff
@@ -201,7 +200,7 @@ def visit_name(self, node: nodes.Name) -> None:
             self._checker.add_message(
                 "notimplemented-raised", node=self._node, confidence=HIGH
             )
-        elif node.name in OVERGENERAL_EXCEPTIONS:
+        elif node.name in self._checker.linter.config.overgeneral_exceptions:
```
Pierre-Sassoulas, line 203:
```suggestion
        elif node.qname in self._checker.linter.config.overgeneral_exceptions:
```
Would this work ? If not we need to infer like in:
```python
                try:
                    exceptions = list(_annotated_unpack_infer(handler.type))
                    continue
```

Pierre-Sassoulas (changes requested):
Thank you for rebasing @jack1142 !

pylint/checkers/exceptions.py
jack1142: Nope because `node` is a `Name` node that doesn't have a `qname` attribute. Using the inference as you've shown seems to have worked, thanks.

jack1142:
@DanielNoord alright, tests are now passing and unless you have further feedback, this should be good to go. The failure from 3.10 primer seems to be unrelated.

tests/functional/b/broad_exception_caught.py
```diff
@@ -11,3 +19,21 @@
     __revision__ += 1
 except BaseException: # [broad-exception-caught]
     print('error')
+
+
+try:
+    __revision__ += 1
+except ValueError:
+    print('error')
```
Pierre-Sassoulas, line 27:
```suggestion
    print('error')
    raise TypeError()
```

tests/functional/b/broad_exception_caught.py
```diff
@@ -11,3 +19,21 @@
...
+
+
+try:
+    __revision__ += 1
+except CustomBroadException: # [broad-exception-caught]
+    print('error')
```
Pierre-Sassoulas, line 33:
```suggestion
except CustomBroadException as e: # [broad-exception-caught]
    print('error')
    raise CustomNarrowException() from e
```

tests/functional/b/broad_exception_caught.py
```diff
@@ -11,3 +19,21 @@
...
+
+
+try:
+    __revision__ += 1
+except CustomNarrowException:
+    print('error')
```
Pierre-Sassoulas, line 39:
```suggestion
except CustomNarrowException as e:
    print('error')
    raise CustomBroadException() from e  # [broad-exception-raised]
```

Pierre-Sassoulas (changes requested):
Suggested some more test for the new messages that appeared in the meantime, let me know what you think. The rest LGTM :)

Pierre-Sassoulas (approved):

Pierre-Sassoulas (commented):
It seems the ``CustomBroadException as e: # [broad-exception-caught]`` genuinely is not raised.

jack1142:
`broad-exception-caught` doesn't get triggered if the except block raises so these addition don't seem to make sense

jack1142:
> `broad-exception-caught` doesn't get triggered if the except block raises so these test additions don't seem to make sense

We could add few more try except blocks which test for that condition I suppose.

Pierre-Sassoulas (approved):
You're right, sorry for the noise !

pylint/checkers/exceptions.py
```diff
@@ -201,13 +200,29 @@ def visit_name(self, node: nodes.Name) -> None:
...
+
+            if not isinstance(exception, nodes.ClassDef):
+                continue
+
+            if self._checker._is_overgeneral_exception(exception):
+                self._checker.add_message(
+                    "broad-exception-raised",
+                    args=exception.name,
+                    node=self._node,
+                    confidence=HIGH,
```
DanielNoord, line 224:
```suggestion
                    confidence=INFERENCE,
```

One final nit.. This would require updating the test output as well.

DanielNoord (changes requested):

pylint/checkers/exceptions.py
```diff
@@ -288,6 +303,18 @@ class ExceptionsChecker(checkers.BaseChecker):
 
     def open(self) -> None:
         self._builtin_exceptions = _builtin_exceptions()
+        for exc_name in self.linter.config.overgeneral_exceptions:
+            if "." not in exc_name:
+                warnings.warn_explicit(
```
jacobtylerwalls, line 308: Can you also update `examples/pylintrc` with the new default value?

jacobtylerwalls (changes requested):
I'm concerned about the changes shown in the primer. Most of the patterns there look benign, where a broad exception is caught, and then some logging/cleanup is done, and then the original error is re-raised. I didn't think we had consensus on changing that (or at least, I don't see any discussion in the issue.) I think the status quo should be preserved where directly raising `MyBroadException` emits a message, but catching and reraising it doesn't.

jack1142:
> I'm concerned about the changes shown in the primer. Most of the patterns there look benign, where a broad exception is caught, and then some logging/cleanup is done, and then the original error is re-raised. I didn't think we had consensus on changing that (or at least, I don't see any discussion in the issue.)

Yeah, there was definitely no consensus on this change. The primer workflow didn't manage to run successfully on this PR 3 days ago when I added inference to broad-exception-raised so I wasn't really aware of these cases.

> I think the status quo should be preserved where directly raising `MyBroadException` emits a message, but catching and reraising it doesn't.

I agree, I've done some changes that should hopefully address this.

jacobtylerwalls (approved):
Thanks for the quick updates! (And sorry the primer was down.)

jack1142:
All good, I'm happy you/primer caught it before this PR was merged :)

pylint/checkers/exceptions.py
```diff
@@ -288,6 +297,18 @@ class ExceptionsChecker(checkers.BaseChecker):
...
+                    "Specifying exception names in overgeneral-exceptions option"
```
DanielNoord, line 303:
```suggestion
                    "Specifying exception names in the overgeneral-exceptions option"
```

DanielNoord (dismissed):
Thanks a lot @jack1142, especially for revisiting and fixing the conflicts after we forgot to merge this.

Much appreciated!

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
@DanielNoord please cleanup the commit message when squashing, it does not have to be super fancy and a perefect synthesis but this one was just a mess 😄 

```
Use qualified name when checking for overgeneral exceptions (https://github.com/PyCQA/pylint/pull/7497)

* Update tests

* Use qualified name when checking for overgeneral exceptions

* WIP: Add deprecation warning

* Add changelog fragment

* Use qualified name in test case

* spell check fix

* Update changelog fragment with suggested fixes

Co-authored-by: Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>

* Move OVERGENERAL_EXCEPTIONS directly to the default value in dict

* Mark as TODO for pylint 3.0

* Properly warn on each occurrence of name without dots

* Update the warning per the review

* Rephrase the warning to mention pylint 3.0

* Remove unnecessary nesting of the if condition

* Quote the exception name in deprecation warning

* Use config value for overgeneral exceptions in broad-exception-raised

* Infer qualified name of the exception in broad-exception-raised

* e.g. -> maybe?

Co-authored-by: Pierre Sassoulas <pierre.sassoulas@gmail.com>

* Suppress missing class docstrings

* Add few more tests for broad-exception-raised

Co-authored-by: Pierre Sassoulas <pierre.sassoulas@gmail.com>

* Fix unexpected missing-raise-from

* Revert "Fix unexpected missing-raise-from"

This reverts commit https://github.com/PyCQA/pylint/commit/d796e72035b7f7578b9e6bb1e45a30935e80b009.

* Revert "Add few more tests for broad-exception-raised"

This reverts commit https://github.com/PyCQA/pylint/commit/e5a193ee136f8566d43450fbb9fbf28cc717d307.

* Change confidence of broad-exception-raised from HIGH to INFERENCE

* Only trigger broad-exception-raised for raise with new exc instance

* Update overgeneral-exceptions definition in example pylintrc file

* Update pylint/checkers/exceptions.py

Co-authored-by: Daniël van Noord <13665637+DanielNoord@users.noreply.github.com>
Co-authored-by: Pierre Sassoulas <pierre.sassoulas@gmail.com>
```

DanielNoord:
Shit, I thought I did. Sorry about that!

pylint/checkers/exceptions.py
```diff
@@ -288,6 +297,18 @@ class ExceptionsChecker(checkers.BaseChecker):
...
+                warnings.warn_explicit(
+                    "Specifying exception names in the overgeneral-exceptions option"
+                    " without module name is deprecated and support for it"
+                    " will be removed in pylint 3.0."
+                    f" Use fully qualified name (maybe 'builtins.{exc_name}' ?) instead.",
+                    category=UserWarning,
+                    filename="pylint: Command line or configuration file",
+                    lineno=1,
+                    module="pylint",
+                )
```
cdce8p, line 311: With `--jobs 2`, the `open()` call is executed for each opened file. E.g. for HomeAssistant I got over 16.000 warnings just for 3 old config values. That should be fixed before we release `2.16`.
https://github.com/cdce8p/ha-core/actions/runs/3470026809/jobs/5797737991#step:8:25

PR #7502, opened 2022-09-20
Improve docs clarity on loading custom checkers - FIXED

### LINKED ISSUE #7264: Cannot get pylint to register custom checker, is there a missing step in the docs?

daogilvie:
NOTE: This PR replaces #7276 due to an admin error on my part.

## Type of Changes

### Short
This PR just updates the documentation page on writing custom checkers. It adds
information about some restrictions between how `init-hook` changes the loading
path and when you can use that to import your custom module.
### Long (in the commit message)
Whilst the docs on testing custom checkers do touch on the fact the
module has to be in your load path, there are some restrictions that
were not made clear before.
Specifically, the init-hook configuration item (which is often used
explicitly to change the sys.path) is not automatically sufficient. If
the init hook is in a file, but the load-plugins argument is passed in
the CLI, then the plugins are loaded before the init-hook has run, and
so will not be imported. In this case, the failure can also be silent,
so the outcome is a lint that will simply not contain any information
from the checker, as opposed to a run that will error.
This corner case may merit fixing in a separate PR, that is to be
confirmed.

Closes #7264

### DIFF
--- a/doc/development_guide/how_tos/custom_checkers.rst
+++ b/doc/development_guide/how_tos/custom_checkers.rst
@@ -223,8 +223,7 @@ Now we can debug our checker!
 
     Alternatively, this module can be made available to pylint by
     putting this module's parent directory in your ``PYTHONPATH``
-    environment variable or by adding the ``my_plugin.py``
-    file to the ``pylint/checkers`` directory if running from source.
+    environment variable.
 
     If your pylint config has an init-hook that modifies
     ``sys.path`` to include the module's parent directory, this
--- a/doc/development_guide/how_tos/custom_checkers.rst
+++ b/doc/development_guide/how_tos/custom_checkers.rst
@@ -220,20 +220,24 @@ Now we can debug our checker!
     ``my_plugin`` refers to a module called ``my_plugin.py``.
     The preferred way of making this plugin available to pylint is
     by installing it as a package. This can be done either from a packaging index like
-    ``PyPI`` or by installing it from a local source such as with ``pip install .``.
+    ``PyPI`` or by installing it from a local source such as with ``pip install``.
 
     Alternatively, the plugin module can be made available to pylint by
     putting this module's parent directory in your ``PYTHONPATH``
     environment variable.
 
-    If your pylint config has an init-hook that modifies
+    If your pylint config has an ``init-hook`` that modifies
     ``sys.path`` to include the module's parent directory, this
     will also work, but only if either:
 
     * the ``init-hook`` and the ``load-plugins`` list are both
-      defined in a pylintrc file, or...
+      defined in a configuration file, or...
     * the ``init-hook`` is passed as a command-line argument and
-      the ``load-plugins`` list is in the pylintrc file
+      the ``load-plugins`` list is in the configuration file
+
+    I.E You cannot load a custom plugin by modifying ``sys.path`` if you
+    supply the ``init-hook`` in a configuration file, but pass the module name
+    in via ``--load-plugins`` on the command line.
 
     This is because pylint loads plugins specified on command
     line before loading any configuration from other sources.
--- a/doc/development_guide/how_tos/custom_checkers.rst
+++ b/doc/development_guide/how_tos/custom_checkers.rst
@@ -219,7 +219,8 @@ Now we can debug our checker!
 
     ``my_plugin`` refers to a module called ``my_plugin.py``.
     The preferred way of making this plugin available to pylint is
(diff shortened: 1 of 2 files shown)


doc/development_guide/how_tos/custom_checkers.rst
```diff
@@ -218,10 +218,29 @@ Now we can debug our checker!
...
+    If your pylint config has an ``init-hook`` that modifies
+    ``sys.path`` to include the module's parent directory, this
+    will also work, but only if either:
+
+    * the ``init-hook`` and the ``load-plugins`` list are both
+      defined in a configuration file, or...
+    * the ``init-hook`` is passed as a command-line argument and
+      the ``load-plugins`` list is in the configuration file
+
+    I.E You cannot load a custom plugin by modifying ``sys.path`` if you
```
DanielNoord, line 238:
```suggestion
    So, you cannot load a custom plugin by modifying ``sys.path`` if you
```

doc/development_guide/how_tos/custom_checkers.rst
```diff
@@ -218,10 +218,29 @@ Now we can debug our checker!
...
+    supply the ``init-hook`` in a configuration file, but pass the module name
+    in via ``--load-plugins`` on the command line.
+
```
DanielNoord, line 241:
```suggestion
```

doc/whatsnew/fragments/7264.other
```diff
@@ -0,0 +1,3 @@
+Dev documentation is now clearer about custom checker loading.
```
DanielNoord, line 1: I think this can be removed.

DanielNoord (changes requested):
Final points!

DanielNoord (approved):

daogilvie:
You are very welcome, thanks for improving and maintaining a project that has saved me from myself more times than I can remember 😂

PR #7505, opened 2022-09-20
handle no-member false positive for generators

### LINKED ISSUE #2567: false positive on E1101 with contextmanager decorated generators
I have this little snippet:

```
import contextlib

@contextlib.contextmanager
def context_manager():
    try:
        yield
    finally:
        pass

cm = context_manager()
cm.__enter__()
cm.__exit__(None, None, None)
```

pylint signals the "direct" access to the __enter__ and __exit__ callables as an error:

```
contextmanaged.py:13:0: E1101: Generator 'generator' has no '__enter__' member (no-member)
contextmanaged.py:14:0: E1101: Generator 'generator' has no '__exit__' member (no-member)
```

while "uncommon" at least, I think it should not be marked as an error.

```
> pylint --version
pylint 2.1.1
astroid 2.0.4
Python 3.6.3 (v3.6.3:2c5fed8, Oct  3 2017, 18:11:49) [MSC v.1900 64 bit (AMD64)]
```

clavedeluna:
## Type of Changes

## Description

A false positive for when a generator called internal enter / exit methods was reported. This PR handles this case in the same way other outlier cases are handled, by adding code to detect this is a generator.

Closes #2567

### DIFF
--- a/tests/functional/n/no/no_member_generators.py
+++ b/tests/functional/n/no/no_member_generators.py
@@ -1,0 +1,15 @@
+"""Test no-member for generators
+"""
+# pylint: disable=missing-docstring, too-few-public-methods
+import contextlib
+
+@contextlib.contextmanager
+def context_manager():
+    try:
+        yield
+    finally:
+        pass
+
+cm = context_manager()
+cm.__enter__()
+cm.__exit__(None, None, None)
--- a/doc/whatsnew/fragments/2567.bugfix
+++ b/doc/whatsnew/fragments/2567.bugfix
@@ -1,0 +1,3 @@
+Fixed a false positive case when a generator object received ``no-member`` error.
+
+Closes #2567
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -510,6 +510,14 @@ def _emit_no_member(
             return False
         except astroid.NotFoundError:
             pass
+    if isinstance(owner, astroid.bases.Generator) and node.attrname in {
+        "__enter__",
+        "__exit__",
+    }:
+        # Avoid false positive on generators.
+        # See https://github.com/PyCQA/pylint/issues/2567
+        return False
+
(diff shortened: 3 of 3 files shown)


tests/functional/n/no/no_member_generators.py
```diff
@@ -0,0 +1,15 @@
...
+@contextlib.contextmanager
+def context_manager():
+    try:
+        yield
+    finally:
+        pass
+
+cm = context_manager()
+cm.__enter__()
+cm.__exit__(None, None, None)
```
clavedeluna, line 15: There is no corresponding .txt file because no pylint errors or warnings should be in the output.

Pierre-Sassoulas (approved):
LGTM, thank you for the fix. (Another possibility would be to make astroid understands decorator better but this is a lot harder to do). I'm going to wait for Daniel's opinion before merging.

DanielNoord (changes requested):
This should indeed be fixed in `astroid`.

Specifically by changing `GeneratorModel` which gets used by `bases.Generator` here:
https://github.com/PyCQA/astroid/blob/849d043f3e3cd63881ba919ccba4e6a726947d22/astroid/bases.py#L606

You can find `astroid.interpreter.objectmodel.GeneratorModel` here:
https://github.com/PyCQA/astroid/blob/849d043f3e3cd63881ba919ccba4e6a726947d22/astroid/interpreter/objectmodel.py#L601

We would need to add `attr___enter__` and `attr___exit__`. It should be relatively straightforward to do so. We only need to make sure that we get the typing and arguments that can be passed to these methods correctly.
The signature of these methods can be found here:
https://docs.python.org/3/reference/datamodel.html#object.__enter__

clavedeluna:
> This should indeed be fixed in `astroid`.

happy to do it. I've followed your instructions and started [this draft PR](https://github.com/PyCQA/astroid/pull/1803) but I'm hoping you can help me fill in the methods as I'm not too sure what they should be....

PR #7509, opened 2022-09-21
Fix false negative ``no-member`` when the code was a `manual` augmented assignments

### LINKED ISSUE #4562: False negative and false positive 'no-member' errors with `+=`
### Steps to reproduce

Given a file `a.py`:

```python
# pylint: disable=missing-class-docstring
# pylint: disable=invalid-name


class A:
    value: int


obj = A()
obj.value += 1
```

Given a file `b.py`:

```python
# pylint: disable=missing-class-docstring
# pylint: disable=invalid-name


class A:
    value: int


obj = A()
obj.value = 1 + obj.value
```

Given a file `c.py`:

```python
# pylint: disable=missing-class-docstring
# pylint: disable=invalid-name


class A:
    value: int


obj = A()
obj.value += 1
obj.value = 1 + obj.value
```

### Current behavior

Results:

```sh
$ pylint a.py b.py c.py
************* Module a
a.py:12:0: E1101: Instance of 'A' has no 'value' member (no-member)

Your code has been rated at 6.15/10
```

### Expected behavior

There should also be `no-member` issues in `b.py` and `c.py`.

### pylint --version output

```
pylint 2.8.2
astroid 2.5.6
Python 3.8.10 (default, May 13 2021, 15:41:32) 
[GCC 9.3.0]
```

clavedeluna:
## Type of Changes

## Description

Add code to detect `no-member` when an augmented assign is done "manually", that is, when a user does this

`obj.value = 1 + obj.value`
instead of 
`obj.value += 1`

If `no-member` would be raised in the second case, it should also be raised in the first case, which is why this PR is here!

Closes #4562 (altho I can remove this auto-close if we want to keep it open for the `pydantic` false positive case, though I Personally think it should be a separate issue)

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1965,6 +1965,7 @@ def is_hashable(node: nodes.NodeNG) -> bool:
         return True
 
 
+<<<<<<< HEAD
 def get_full_name_of_attribute(node: nodes.Attribute | nodes.AssignAttr) -> str:
     """Return the full name of an attribute and the classes it belongs to.
 
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1965,7 +1965,6 @@ def is_hashable(node: nodes.NodeNG) -> bool:
         return True
 
 
-<<<<<<< HEAD
 def get_full_name_of_attribute(node: nodes.Attribute | nodes.AssignAttr) -> str:
     """Return the full name of an attribute and the classes it belongs to.
 
--- a/doc/whatsnew/fragments/4562.bugfix
+++ b/doc/whatsnew/fragments/4562.bugfix
@@ -1,0 +1,3 @@
+Fix ``no-member`` false negative when augmented assign is done manually, without ``+=``.
+
+Closes #4562
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1160,7 +1160,10 @@ accessed. Python regular expressions are accepted.",
                     try:
                         if isinstance(
                             attr_node.statement(future=True), nodes.AugAssign
-                        ) or utils.is_augassign(attr_node, attr_parent):
+                        ) or (
+                            isinstance(attr_parent, nodes.Assign)
+                            and utils.is_augmented_assign(attr_parent)[0]
+                        ):
                             continue
                     except astroid.exceptions.StatementMissing:
                         break
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1160,7 +1160,7 @@ accessed. Python regular expressions are accepted.",
                     try:
                         if isinstance(
                             attr_node.statement(future=True), nodes.AugAssign
-                        ):
+                        ) or utils.is_augassign(attr_node, attr_parent):
                             continue
                     except astroid.exceptions.StatementMissing:
                         break
--- a/tests/functional/n/no/no_member_augassign.py
+++ b/tests/functional/n/no/no_member_augassign.py
@@ -1,0 +1,25 @@
+"""Tests for no-member in relation to AugAssign operations."""
+# pylint: disable=missing-module-docstring, too-few-public-methods, missing-class-docstring, invalid-name
+
+# Test for: https://github.com/PyCQA/pylint/issues/4562
+class A:
+    value: int
+
+obj_a = A()
+obj_a.value += 1  # [no-member]
+
+
+class B:
+    value: int
+
+obj_b = B()
+obj_b.value = 1 + obj_b.value  # [no-member]
+
+
+class C:
+    value: int
+
+
+obj_c = C()
+obj_c.value += 1  # [no-member]
+obj_c.value = 1 + obj_c.value  # [no-member]
--- a/tests/functional/n/no/no_member_augassign.py
+++ b/tests/functional/n/no/no_member_augassign.py
@@ -1,5 +1,5 @@
 """Tests for no-member in relation to AugAssign operations."""
-# pylint: disable=missing-module-docstring, too-few-public-methods, missing-class-docstring, invalid-name
+# pylint: disable=missing-module-docstring, too-few-public-methods, missing-class-docstring, invalid-name, consider-using-augmented-assign
 
 # Test for: https://github.com/PyCQA/pylint/issues/4562
 class A:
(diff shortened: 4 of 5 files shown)


pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,36 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+    attr = None
+    if isinstance(binop.right, nodes.Attribute):
+        attr = binop.right
+    elif isinstance(binop.left, nodes.Attribute):
+        attr = binop.left
+
+    if not attr:
+        return False
+
+    return binop.op == "+" and attr.attrname == node.attrname
```
clavedeluna, line 1998: this entire util func is pretty ugly ... all ears for improvements

pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,36 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+        obj.value = 1 + obj.value
+    is the same as
+        obj.value += 1
+    and this function is intended to detect the first case.
+    """
+    if not isinstance(node, nodes.AssignAttr):
+        return False
+
+    children = [x for x in parent.get_children()]
+    binops = [x for x in children if isinstance(x, nodes.BinOp)]
```
Pierre-Sassoulas, line 1983: As children is only used in the second list  comprehension, imbricating both list comprehension would be more efficient. 

I'm on mobile so I can't check get_children api but I wouldn't be surprised if you can get all the children of a certain type directly.

pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,36 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+    if not binops:
```
Pierre-Sassoulas, line 1984: We can avoid doing the following if in some cases like this:
```suggestion
    if not binops or binops[0] ! = "+":

File: pylint/checkers/utils.py (hunk 1 above)
Comment (Pierre-Sassoulas) [inline pylint/checkers/utils.py:1998]:
```suggestion
    return attr.attrname == node.attrname
```

pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,36 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+        return False
+
+    binop = binops[0]
+
+    attr = None
```
Pierre-Sassoulas, line 1989:
```suggestion

File: pylint/checkers/utils.py [hunk 5]
```diff
@@ -1963,3 +1963,36 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+    if isinstance(binop.right, nodes.Attribute):
+        attr = binop.right
+    elif isinstance(binop.left, nodes.Attribute):
+        attr = binop.left
+
+    if not attr:
```
Pierre-Sassoulas, line 1995:
```suggestion
    else:
```

Pierre-Sassoulas (changes requested):
Looks pretty good already, thank you. I made a small review, there's an issue with the pipeline to fix too.

DanielNoord:
@Pierre-Sassoulas I was thinking: does it make sense to make this a code style check? `consider-using-augmented-assign`. We would get a nice primer result to check for false positives on this util function.

Pierre-Sassoulas:
Sounds good !

DanielNoord:
@clavedeluna I'll have a look at converting this into a message tomorrow. You can do so as well if you want to? I'd be happy to guide you in the right direction?

pylint/checkers/utils.py
clavedeluna: I'm much more in favor of 
```
var = None
if ...
   var = 1
elif ...
  var = 2

if not var
```

but hey I"ll take your suggestion

Pierre-Sassoulas: I did not consider the case where var is none from the if, I'm not opinionated about this you asked about possible style fix so I tried :)

pylint/checkers/utils.py
clavedeluna: I don't see a way to get children of a certain type https://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.AugAssign.get_children
but will def combine the list comps

Pierre-Sassoulas: Maybe node_of_classes but I'm not sure it's only the children that are returned. Sorry for suggesting things that I can't test myself right now, I'm on mobile.

clavedeluna:
whoops... broke the tests

clavedeluna:
Alright so maybe I'm missing something obvious but how is
`pylint/checkers/utils.py:1996: error: Returning Any from function declared to return "bool`
an issue for `return attr.attrname == node.attrname`

is that not always a bool? do I wrap it in bool??

DanielNoord:
> Alright so maybe I'm missing something obvious but how is `pylint/checkers/utils.py:1996: error: Returning Any from function declared to return "bool` an issue for `return attr.attrname == node.attrname`
> 
> is that not always a bool? do I wrap it in bool??

This is because `astroid` is not typed so you're essentially comparing `Any with Any`. If you add a disable for this specific error with `type: ignore` we will remove it after `astroid` does have typing.

clavedeluna:
> we will remove it after `astroid` does have typing.

ah makes sense! I'm getting more and more familiar with the code base so thanks for the tip.

Pierre-Sassoulas (dismissed):
LGTM, thank you @clavedeluna !

DanielNoord:
I'm working a PR that incorporates these changes to get a primer result. Should be done in 30 minutes or something like that.

Pierre-Sassoulas:
I'll let you merge this one @DanielNoord

DanielNoord:
See https://github.com/PyCQA/pylint/pull/7514.

I had to change quite a bit. Preferably we merge that PR first and then use the utility introduced there in this PR. The robustness we get from checking the new checker on the primer is a huge benefit compared to the functional tests we can think of ourselves.

clavedeluna:
> See #7514.
> 
> I had to change quite a bit. Preferably we merge that PR first and then use the utility introduced there in this PR. The robustness we get from checking the new checker on the primer is a huge benefit compared to the functional tests we can think of ourselves.

I will rebase and re-do this PR once that one is merged

Pierre-Sassoulas (approved):
Thank you for inspiring the work on the augmented assignment checker, and thank you for fixing this !

Issue #3391, opened 2020-02-08
Simplifying assignment operators operation with same variable in left and right hand side of assignment


EugeneZelenko:
It'll be great to have check which will suggest to simplify assignment operators, like "value = value + 1" to "value += 1".

elfring:
:eyes: I am looking for further clarification also according to the general topic “[Increase the usage of augmented assignment statements](https://github.com/PyCQA/pylint/issues/5174 "Better support for augmented assignment statements")”.

EugeneZelenko:
@DanielNoord: Thank you!

PR #7514, opened 2022-09-22
Add ``consider-using-augmented-assign`` checker

DanielNoord:

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/3391

Based around the work by @clavedeluna in https://github.com/PyCQA/pylint/pull/7509. But to make this checker work I had to add some additional checking.

Let's see what the primer thinks of this! 😰


DanielNoord:
Something is up with the caches of the primer. Seems to have to do with a bump in the Python versions used by the runner that is happening in between the creation of the cache and the calling to the cache.

Let's see if all things return back to normal after the run on `main` has completed and we can start a new primer run on this branch.

jacobtylerwalls:
Noticed the first Django result is a false positive (string formatting, not modulo):
https://github.com/django/django/blob/37a13cc92d51d5a57b42240b9a5dd250648efa24/django/templatetags/i18n.py#L190

clavedeluna (dismissed):
Nice! And quite a bit more robust than the initial implementation 🚀

jacobtylerwalls (changes requested):
Requesting a test showing that `%=` does not raise a false positive.

DanielNoord:
> Noticed the first Django result is a false positive (string formatting, not modulo):
> https://github.com/django/django/blob/37a13cc92d51d5a57b42240b9a5dd250648efa24/django/templatetags/i18n.py#L190

Thanks for checking! PRs like these are where the primer shines 😄

DanielNoord:
> Noticed the first Django result is a false positive (string formatting, not modulo):
> https://github.com/django/django/blob/37a13cc92d51d5a57b42240b9a5dd250648efa24/django/templatetags/i18n.py#L190

Can't just simplify that as well though?

```console
❯ python
Python 3.10.0 (default, Oct 27 2021, 16:20:25) [Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> x = "%s"
>>> x % "my_string"
'my_string'
>>> x %= "my_string"
>>> x
'my_string'
```

jacobtylerwalls:
Ah, that might not be a false positive then. Still I'm worried there could be something lurking with incomplete string formatting (e.g. `TypeError: not all arguments converted...`). A couple tests would be nice just to show this isn't a worry.

DanielNoord:
> Ah, that might not be a false positive then. Still I'm worried there could be something lurking with incomplete string formatting (e.g. `TypeError: not all arguments converted...`). A couple tests would be nice just to show this isn't a worry.

I think those would be an issue without the augmented assign as well, but you're right: better safe than sorry!

tests/functional/c/consider/consider_using_augmented_assign.py
```diff
@@ -0,0 +1,62 @@
...
+x = x + 1  # [consider-using-augmented-assign]
+x = 1 + x  # [consider-using-augmented-assign]
+x, y = 1 + x, 2 + x
+# We don't warn on intricate expressions as we lack knowledge
+# of simplifying such expressions which is necessary to see
+# if they can become augmented
+x = 1 + x - 2
+x = 1 + x + 2
+
+# For anything other than a float or an int we only want to warn on
```
DanielNoord, line 17: Primer pointed this out. Appending list and changing strings is all order dependent. I think it makes sense to be a little stricter here to avoid false positives.

DanielNoord:
I don't understand `astroid` 1, but that seems to be caused by incorrect inference of `filled_args`.
I checked all other primer messages and they all seem to make sense to me.

I think this is ready for review!

tests/functional/c/consider/consider_using_augmented_assign.py
clavedeluna: If it's easy enough, could you share the primer link / output that helped you see this? I'm trying to understand how you maintainers are using primer as feedback

clavedeluna (approved):

tests/functional/c/consider/consider_using_augmented_assign.py
DanielNoord: It's really just manually going through the results. The previous comment by the primer (https://github.com/PyCQA/pylint/pull/7514#issuecomment-1255174121) has now been hidden but case 6 of `sentry` triggered me to think about the implications of list concatenation and how those are order dependent (in contrast to simple integer calculation).

It's a bit hands on and for larger new checks like this it is a bit much, but sometimes the `primer` shows only one or two changes for a PR for which we do no expect any changes. That can be a sign that the PR would introduce a new false positive or is actually a previously unknown false negative.

pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,71 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+            while isinstance(target_parent, (nodes.Attribute, nodes.Name)):
+                if isinstance(target_parent, nodes.Attribute):
+                    target_string = f"{target_parent.attrname}.{target_string}"
+                else:
+                    target_string = f"{target_parent.name}.{target_string}"
+                target_parent = target_parent.parent
+
+            side_parent = side.parent
+            side_string = side.attrname or ""
+            while isinstance(side_parent, (nodes.Attribute, nodes.Name)):
```
mbyrnepr2, line 1989: Perhaps these two`while` blocks could be extracted into a reusable function. They're both quite similar.

pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,71 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+                side_parent = side_parent.parent
+            return target_string == side_string
+        return False
+    return False
+
+
+def is_augmented_assign(node: nodes.Assign) -> tuple[bool, str]:
+    """Determine if the node is assigning itself (with modifications) to itself.
+
+    For example, x = 1 + x
```
mbyrnepr2, line 2003:
```suggestion
    For example: x = 1 + x
```

pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,71 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+
+def _is_target_name_in_binop_side(
+    target: nodes.AssignName | nodes.AssignAttr, side: nodes.NodeNG | None
+) -> bool:
+    """Determine whether the target name-like node is referenced in the side node."""
+    if isinstance(side, nodes.Name):
+        if isinstance(target, nodes.AssignName):
+            return target.name == side.name  # type: ignore[no-any-return]
+        return False
+    if isinstance(side, nodes.Attribute):
```
Pierre-Sassoulas, line 1976:
```suggestion
    if not isinstance(side, nodes.Attribute):
        return False
    if not isinstance(target, nodes.AssignAttr):
        return False
```
Then we can reduce everything 2 indentation level.

pylint/checkers/utils.py
```diff
@@ -1963,3 +1963,71 @@ def is_hashable(node: nodes.NodeNG) -> bool:
...
+                    side_string = f"{side_parent.attrname}.{side_string}"
+                else:
+                    side_string = f"{side_parent.name}.{side_string}"
+                side_parent = side_parent.parent
+            return target_string == side_string
+        return False
+    return False
+
+
+def is_augmented_assign(node: nodes.Assign) -> tuple[bool, str]:
```
Pierre-Sassoulas, line 2000: I like this design.

tests/functional/c/consider/consider_using_augmented_assign.py
```diff
@@ -0,0 +1,62 @@
...
+# Currently we disregard all calls
+my_str = return_str() + my_str
+
+my_str = my_str % return_str()
+my_str = my_str % 1  # [consider-using-augmented-assign]
+my_str = my_str % (1, 2)  # [consider-using-augmented-assign]
+my_str = "%s" % my_str
+my_str = return_str() % my_str
+my_str = Unknown % my_str
+my_str = my_str % Unknown  # [consider-using-augmented-assign]
```
Pierre-Sassoulas, line 62: A good example for ``%=`` could be:

```suggestion
my_str = my_str % Unknown  # [consider-using-augmented-assign]
obj.name = obj.name % {
    "app_label": cls._meta.app_label.lower(), "class": cls.__name__.lower(),
} 
```

Pierre-Sassoulas (changes requested):
There's a badly formatted message for black; ``Use '=' to do an augmented assign directly*`` (https://github.com/psf/black/blob/6ae8457a8628345c59fed22c58728ffa258a3e76/src/blib2to3/pgen2/literals.py#L51). It also happen for pandas here : https://github.com/pandas-dev/pandas/blob/71fc89cd515c3c19230fbab64e979118858b808a/pandas/tests/io/formats/test_format.py#L1535, so probably for all ``*``.

I added a suggestion for some functional test with operators I could think of.

tests/functional/c/consider/consider_using_augmented_assign.py
```diff
@@ -0,0 +1,62 @@
...
+# if they can become augmented
+x = 1 + x - 2
+x = 1 + x + 2
+
+# For anything other than a float or an int we only want to warn on
+# assignments where the 'itself' is on the left side of the assignment
+my_list = [2, 3, 4]
+my_list = [1] + my_list
+
+
```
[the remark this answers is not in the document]
DanielNoord: Do we really need this? We don't do any operator checking in the utility? The `in` and `not in` are `nodes.Comparse` so they don't feature in this utility.

tests/functional/c/consider/consider_using_augmented_assign.py
DanielNoord: Same comment as above, this doesn't really add too much value imo. If we handle `% tuple` we also handle `% dict`. Unless you have a specific regression of course, then I'm happy to add!

pylint/checkers/utils.py
DanielNoord: Done!

pylint/checkers/utils.py
DanielNoord: Code has been changed as I created a new utility.

tests/functional/c/consider/consider_using_augmented_assign.py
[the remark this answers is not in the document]
Pierre-Sassoulas: Well we did have ``>>=`` suggested in pygame, imo it does not hurt to check.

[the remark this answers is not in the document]
DanielNoord: > `>>=`

That's valid right?

```consol
Python 3.10.0 (default, Oct 27 2021, 16:20:25) [Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 2 << 6
128
>>> x = 2
>>> x <<= 6
>>> x
128
```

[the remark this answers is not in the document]
DanielNoord: I added the tests that make sense to me. Checking for `is not` uses a different node so I didn't include variants of those. I did do `x > 3` as it is visually more like "algebra" or `BinOps`.

pylint/checkers/utils.py
mbyrnepr2: 👍

tests/functional/c/consider/consider_using_augmented_assign.py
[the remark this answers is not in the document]
Pierre-Sassoulas: > That's valid right?

Sure ! I did not see the almost identical tests from line 64 below where you tested most of the thing I suggested. 

So the remaining one would be those:
```
x = Unknown
x = x not in 3
x = x not in [3]
x = x in 3
x = x in [3]
x = x is 3
x = x is not 3
x = x is 3
x = x is not 3
x = x and 3
x = x or 3
```
Kinda agree this might not be necessary to tests them.

tests/functional/c/consider/consider_using_augmented_assign.py
```diff
@@ -0,0 +1,77 @@
+"""Tests for consider-using-augmented-assign."""
+
+# pylint: disable=invalid-name,too-few-public-methods,import-error,consider-using-f-string
+
+from unknown import Unknown
+
+x = 1
+x = x + 1  # [consider-using-augmented-assign]
+x = 1 + x  # [consider-using-augmented-assign]
+x, y = 1 + x, 2 + x
```
Pierre-Sassoulas, line 10: Shouldn't the left part of the tuple raise ?

tests/functional/c/consider/consider_using_augmented_assign.py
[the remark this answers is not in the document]
DanielNoord: Yeah I mean that's just assignments of booleans. For those we have our whole test suite as a regressions 😄
We wouldn't be able to pass any run if we started flagging those 😅

tests/functional/c/consider/consider_using_augmented_assign.py
DanielNoord: To what?

```python
x += 1
y = 2 + x
```

I don't think that's nicer, or at least too opinionated for a check which is more subtle like the current one.

Pierre-Sassoulas: Right, I did not consider that.

Pierre-Sassoulas (approved):
Great idea for a checker, and nice execution 👌

jacobtylerwalls (approved):

cdce8p:
_I haven't followed the discussion around this check so it might have come up already._

Why was this not implemented as an optional extension?

From what I can tell there is no obvious benefit to using the augmented assignment over the regular one. IMO it just creates a lot of unnecessary noise. Similar checks like the one for assignment expressions are also optional.

Pierre-Sassoulas:
I think @cdce8p has a point here.

jacobtylerwalls:
Agreed.

DanielNoord:
Agreed. I'll create a PR when I get the time.

DanielNoord:
See opened issue for follow up question.

PR #7517, opened 2022-09-22
Fix false negative for `no-member`

### LINKED ISSUE #4048: False-negative E1101 no-member for function parameter
Is your issue fixed on the preview release?: `pip install pylint astroid --pre -U` --> no

### Steps to reproduce
```
"""False-negative no-member."""


def _func(param=True):
    print(param.name)
```

### pylint --version output
```
pylint 2.6.0
astroid 2.4.2
Python 3.8.6 (default, Sep 29 2020, 15:36:55) 
[GCC 7.5.0]
```

mbyrnepr2:
Fix false negative for ``no-member`` when accessing an attribute of a function argument.

Closes #4048

## Type of Changes

## Description

Closes #4048

### DIFF
--- a/tests/functional/c/confidence_filter.py
+++ b/tests/functional/c/confidence_filter.py
@@ -5,7 +5,7 @@ class Client:
     """use provider class"""
 
     def __init__(self):
-        self.set_later = 0
+        self.set_later = "tomato"
 
     def set_set_later(self, value):
         """set set_later attribute (introduce an inference ambiguity)"""
--- a/doc/whatsnew/fragments/4048.false_negative
+++ b/doc/whatsnew/fragments/4048.false_negative
@@ -1,0 +1,3 @@
+Fix false negative for ``no-member`` when accessing an attribute of a function argument.
+
+Closes #4048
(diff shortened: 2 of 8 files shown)


tests/functional/c/confidence_filter.py
```diff
@@ -5,7 +5,7 @@ class Client:
     """use provider class"""
 
     def __init__(self):
-        self.set_later = 0
+        self.set_later = "tomato"
```
mbyrnepr2, line 8: Reason: `E1101: Instance of 'int' has no 'lower' member (no-member)` is emitted with the changes in this PR.

DanielNoord: Isn't that a regression? We shouldn't emit `no-member` if we aren't certain about it's type.

mbyrnepr2: I don't think so? 
Before: `0.lower()` was not emitting a no-member. Now it is which is expected.

DanielNoord: But isn't `set_set_later` there so that we don't know the type of `set_later`? In this test we don't call `set_set_later` but if we were to add `self.set_set_later("")` on L9 do we still emit? If so this is a regression as then `set_later` is a `str`.

mbyrnepr2: I see your point. Thanks @DanielNoord.

PR #7518, opened 2022-09-23
Replace no-method-argument with no-self-argument when variadic args present

### LINKED ISSUE #7507: Method with variadic args/kwargs but no `self` reports `no-method-argument` instead of `no-self-argument`
### Bug description

```py

class MyClass:
    # correct warning: no-method-argument
    def no_args():
        """A method without a self argument or any other arguments."""

    # correct warning: no-self-argument
    def no_self_arg(arg):
        """A method without a self argument but with a standard argument."""

    # incorrect warning: no-method-argument
    def no_self_arg_but_varargs(*args):
        """A method without a self argument but with *args."""

    # incorrect warning: no-method-argument
    def no_self_arg_but_kwargs(**kwargs):
        """A method without a self argument but with **kwargs."""

    # incorrect warning: no-method-argument
    def no_self_arg_but_varargs_and_kwargs(*args, **kwargs):
        """A method without a self argument but with *args and **kwargs."""

    def correct(self):
        """Correct."""
        self.var = "correct"
```

### Configuration

_No response_

### Command used

```shell
pylint repro.py
```

### Pylint output

```shell
************* Module repro
repro.py:5:4: E0211: Method has no argument (no-method-argument)
repro.py:9:4: E0213: Method should have "self" as first argument (no-self-argument)
repro.py:13:4: E0211: Method has no argument (no-method-argument)
repro.py:17:4: E0211: Method has no argument (no-method-argument)
repro.py:21:4: E0211: Method has no argument (no-method-argument)

```

### Expected behavior

I expected to see `no-self-argument` when a method has `*args` or `**kwargs` rather than `no-method-argument` as the latter is quite confusing as the method does have arguments.

### Pylint version

```shell
pylint 2.15.3
astroid 2.12.10
Python 3.8.14 (default, Sep  8 2022, 00:02:07) 
[GCC 11.2.0]
```

### OS / Environment

Kubuntu 22.04.1 LTS

### Additional dependencies

_No response_

jack1142:
## Type of Changes

## Description

Report `no-self-argument` rather than `no-method-argument` for methods with variadic arguments.
Per the comment in #7507, I also added method name to the error messages of no-method-argument and no-self-argument

Closes #7507

### DIFF
--- a/doc/whatsnew/fragments/7507.bugfix
+++ b/doc/whatsnew/fragments/7507.bugfix
@@ -1,0 +1,4 @@
+Report `no-self-argument` rather than `no-method-argument` for methods
+with variadic arguments.
+
+Closes #7507
--- a/doc/whatsnew/fragments/7507.bugfix
+++ b/doc/whatsnew/fragments/7507.bugfix
@@ -1,4 +1,4 @@
-Report `no-self-argument` rather than `no-method-argument` for methods
+Report ``no-self-argument`` rather than ``no-method-argument`` for methods
 with variadic arguments.
 
 Closes #7507
--- a/doc/whatsnew/fragments/7507.other
+++ b/doc/whatsnew/fragments/7507.other
@@ -1,3 +1,3 @@
-Add method name to the error messages of no-method-argument and no-self-argument.
+Add method name to the error messages of ``no-method-argument`` and ``no-self-argument``.
 
 Closes #7507
--- a/doc/whatsnew/fragments/7507.other
+++ b/doc/whatsnew/fragments/7507.other
@@ -1,0 +1,3 @@
+Add method name to the error messages of no-method-argument and no-self-argument.
+
+Closes #7507
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1950,11 +1950,11 @@ a metaclass class method.",
             # Check if there is a decorator which is not named `staticmethod` but is assigned to one.
             return
         # class / regular method with no args
-        elif (
-            not node.args.args
-            and not node.args.posonlyargs
-            and not node.args.vararg
-            and not node.args.kwarg
+        elif not (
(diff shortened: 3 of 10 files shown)


doc/whatsnew/fragments/7507.bugfix
```diff
@@ -0,0 +1,4 @@
+Report `no-self-argument` rather than `no-method-argument` for methods
```
clavedeluna, line 1: You'll want double tick marks, ``...``` for both no-self-arg and no-method-arg so they show up as code

doc/whatsnew/fragments/7507.other
```diff
@@ -0,0 +1,3 @@
+Add method name to the error messages of no-method-argument and no-self-argument.
```
clavedeluna, line 1: may want to add the double tick marks ``..`` here for the msg names

pylint/checkers/classes/class_checker.py
```diff
@@ -1950,8 +1950,13 @@ def _check_first_arg_for_type(
             # Check if there is a decorator which is not named `staticmethod` but is assigned to one.
             return
         # class / regular method with no args
-        elif not node.args.args and not node.args.posonlyargs:
-            self.add_message("no-method-argument", node=node)
+        elif (
+            not node.args.args
+            and not node.args.posonlyargs
+            and not node.args.vararg
+            and not node.args.kwarg
```
clavedeluna, line 1957: could this be refactored to remove so many `not` to
```(
not (node.args.args and node.args.pos...... and ... and )
```

maybe not? worth checking

jack1142: Yeah, that makes sense, thanks.

clavedeluna (approved):

Pierre-Sassoulas (approved):
LGTM thank you. Better message in general, and a nice fix to a nonsensical message 👌

PR #7520, opened 2022-09-23
Detect unreachable code after sys.exit call

clavedeluna:
## Type of Changes

## Description

Pylint can now detect unreachable code for any code that follows a sys.exit call

Refs #519


mbyrnepr2:
Thanks for tackling this @clavedeluna!

I think the primer result shows that some control flow is needed since the message shouldn't emit in this case in the `else` block.

doc/whatsnew/fragments/519.other
```diff
@@ -0,0 +1,3 @@
+Code following call to ``sys.exit`` will be marked as `unreachable`
```
mbyrnepr2, line 1:
```suggestion
Code following call to ``sys.exit`` will be marked as `unreachable`. 
```

clavedeluna:
I'm still trying to figure out how to use the primer results, it's confusing because the run hasn't finished but there's an output above in the PR

Pierre-Sassoulas:
> it's confusing because the run hasn't finished but there's an output above in the PR

There's a new comment each time the primer is run, the new one should appear shortly.

pylint/checkers/base/basic_checker.py
```diff
@@ -658,12 +658,26 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
+        """Visit a Call node ->
+        1
+
+        check if this is not a disallowed builtin call
+        2. check for * or ** use.
+        3. check if this call is sys.exit and is followed by unreachable code
         """
+        if (
+            isinstance(node.func, nodes.Attribute)
+            and node.func.attrname == "exit"
```
jacobtylerwalls, line 677: Won't this catch `other_module.exit()`?

Pierre-Sassoulas: Yes, we need to use inference for this kind of thing.

tests/functional/u/unreachable.py
```diff
@@ -32,3 +33,15 @@ def func6():
     return
     yield
     print("unreachable")  # [unreachable]
+
+def func7():
+    sys.exit(1)
+    var = 2 + 2  # [unreachable]
+    print(var)
+
```
Pierre-Sassoulas, line 41:
```suggestion
class FalseExit:
    def exit():
        print("False positive not sys.exit")

def func_false_exit():
    sys  = FalseExit()
    sys.exit(1)
    var = 2 + 2
    print(var)
```
Following @jacobtylerwalls comment.

pylint/checkers/base/basic_checker.py
clavedeluna: Exactly the feedback I was looking for. Pierre, can you explain what you mean by use inference? Do you mean the INFERENCE confidence level?

Pierre-Sassoulas: Well yes, but when we use the inference confidence level it means we used inference to come to a conclusion. Here's an example of inference (notice the ``infer()`` call) :
```python
>>> import astroid
>>> a = astroid.extract_node('''
... import argparse
... 
... parser = argparse.ArgumentParser()
... parser.add_argument(
...     "config_folder", help="Path to the project configuration folder"
... )
... args = parser.parse_args()
... try:
...     int(args.config_folder) / 0
... except:
...     parser.error(
...         f"Your configuration folder can't be divided by zero, use something else !"
...     )
... ''')
>>> list(a.handlers[0].body[0].value.func.infer())[0]
<BoundMethod error of argparse.ArgumentParser at 0x140506382647696>
```

clavedeluna: AHA! You just answered one of my top questions about pylint / astroid. I couldn't figure out how the code would distinguish between a call to `method_a` if this method name could be defined in `module1.method_a` and `module2.method_a` looks like a call to `infer` is the way!

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,38 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
-        call and check for * or ** use.
+        """Visit a Call node ->
+        1
+
+        check if this is not a disallowed builtin call
+        2. check for * or ** use.
+        3. check if this call is sys.exit and is followed by unreachable code
         """
+        if self._is_sys_exit(node):
+            self._check_unreachable(node)
```
Pierre-Sassoulas, line 691: The confidence need to be INFERENCE in this case, but it should still be HIGH when we don't need inference.

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,38 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
+    @staticmethod
+    def _is_sys_exit(node: nodes.Call) -> bool:
+        """Detect call to sys.exit."""
+        if not isinstance(node.func, nodes.Attribute) or isinstance(
+            node.parent, nodes.Lambda
+        ):
+            return False
+
+        try:
+            inferred = next(node.func.infer())
```
Pierre-Sassoulas, line 669: What happen if the call to next raise a ``StopIteration`` ?

Pierre-Sassoulas (changes requested):
It's starting to look pretty good, thank you @clavedeluna :)

pylint/checkers/base/basic_checker.py
DanielNoord: Why not use `safe_infer`?

clavedeluna: > Why not use `safe_infer`?

because I'm just hearing about it 🤣 

Joking aside, I took the example from the many other uses of this pattern in the repo
```
        try:
            inferred = list(node....infer())
            return
```
so should these examples also be moved to use safe_infer which seems to do all the catching we need? (separate issue)

DanielNoord: Some... 

Sometimes we want to iterate over all inference results and just handle the exceptions. For example, for file wrappers the inference is flaky as the `io` module does things we don't understand. However, because we know this we don't want to focus on the flaky total result but focus on the first sub-result as we can determine whether we are looking at an `io` object from it. I know that that is not the best explanation, but I can't think of a better example right now.

That said, if you're looking to contribute checking all `infer()` calls and seeing whether their exception branched are covered would be a good way to start removing some of them. I think there is actually an issue in `astroid` about doing the same there.

_For a future project we could even look at something like `safe_iterating_infer` to catch some but not all exceptions and return an iterator, but that's not something we should look at right now_

doc/whatsnew/fragments/519.other
```diff
@@ -0,0 +1,3 @@
+Code following call to ``sys.exit`` will be marked as `unreachable`.
+
+Refs #1234
```
mbyrnepr2, line 3:
```suggestion
Refs #519
```

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,35 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
+        "exec-used",
+        "bad-reversed-sequence",
+        "misplaced-format-function",
+        "unreachable",
     )
     def visit_call(self, node: nodes.Call) -> None:
-        """Visit a Call node -> check if this is not a disallowed builtin
-        call and check for * or ** use.
+        """Visit a Call node ->
+        1
```
mbyrnepr2, line 681: Can remove empty line between 1. and 2.

pylint/checkers/base/basic_checker.py
```diff
@@ -731,7 +753,9 @@ def leave_tryfinally(self, _: nodes.TryFinally) -> None:
         self._tryfinallys.pop()
 
     def _check_unreachable(
-        self, node: nodes.Return | nodes.Continue | nodes.Break | nodes.Raise
+        self,
+        node: nodes.Return | nodes.Continue | nodes.Break | nodes.Raise,
```
mbyrnepr2, line 757:
```suggestion
        node: nodes.Return | nodes.Continue | nodes.Break | nodes.Raise | nodes.Call,
```

Since visit_call is also using this fn.

Pierre-Sassoulas (commented):
I think we should also handle ``quit()``, ``exit()``, and ``os._exit()``, what do you think ?

pylint/checkers/base/basic_checker.py
clavedeluna: This was actually done by the [pre-commit ci](https://github.com/PyCQA/pylint/pull/7520/commits/f085811b8a4a6250bd5b323b54eff1e49fa54652) and I'm not sure how to undo it without it doing it again

Pierre-Sassoulas: You need a summary for the docstring at the beggining for pep257, the following text won't be modified.

clavedeluna:
> I think we should also handle `quit()`, `exit()`, and `os._exit()`, what do you think ?

I was able to easily add `quit` and `exit` but `os._exit` is trickier bc it doesn't follow the same pattern. safe_infer doesn't return anything useful

```
(Pdb) [x for x in node.func.infer()]
[<FunctionDef._exit l.None at 0x1121f0eb0>, Uninferable]

(Pdb) [x for x in node.func.infer()][0].__dict__
{'name': '_exit', '_doc': 'Exit to the system with specified status, without normal exit processing.', 'doc_node': <Const.str l.None at 0x1121f0df0>, 'instance_attrs': {'__class__': [<ClassDef.builtin_function_or_method l.None at 0x1121f0e20>]}, 'locals': {}, 'body': [], 'lineno': None, 'col_offset': None, 'parent': <Module.posix l.0 at 0x11201e790>, 'end_lineno': None, 'end_col_offset': None, 'position': None, 'args': <Arguments l.0 at 0x1121f0dc0>, 'decorators': None, 'returns': None, 'type_comment_returns': None, 'type_comment_args': None, 'extra_decorators': [], 'type': 'function', 'fromlineno': None}
```

I left the functional test but I think we can leave it for now.

DanielNoord:
> > I think we should also handle `quit()`, `exit()`, and `os._exit()`, what do you think ?
> 
> I was able to easily add `quit` and `exit` but `os._exit` is trickier bc it doesn't follow the same pattern. safe_infer doesn't return anything useful
> 
> ```
> (Pdb) [x for x in node.func.infer()]
> [<FunctionDef._exit l.None at 0x1121f0eb0>, Uninferable]
> 
> (Pdb) [x for x in node.func.infer()][0].__dict__
> {'name': '_exit', '_doc': 'Exit to the system with specified status, without normal exit processing.', 'doc_node': <Const.str l.None at 0x1121f0df0>, 'instance_attrs': {'__class__': [<ClassDef.builtin_function_or_method l.None at 0x1121f0e20>]}, 'locals': {}, 'body': [], 'lineno': None, 'col_offset': None, 'parent': <Module.posix l.0 at 0x11201e790>, 'end_lineno': None, 'end_col_offset': None, 'position': None, 'args': <Arguments l.0 at 0x1121f0dc0>, 'decorators': None, 'returns': None, 'type_comment_returns': None, 'type_comment_args': None, 'extra_decorators': [], 'type': 'function', 'fromlineno': None}
> ```
> 
> I left the functional test but I think we can leave it for now.

This is a good example where using `infer()` can make sense. You could do `node.func.infer()[0].qname()` and see if that returns something usable.

clavedeluna:
> You could do `node.func.infer()[0].qname()` and see if that returns something usable.

Oh good catch! I initially did `.name` which only returned `_exit` but `posix._exit` returns `posix._exit` so that's way more useful.

DanielNoord:
> > You could do `node.func.infer()[0].qname()` and see if that returns something usable.
> 
> Oh good catch! I initially did `.name` which only returned `_exit` but `posix._exit` returns `posix._exit` so that's way more useful.

That seems like it would break on Windows... I guess we'll see what the CI thinks about it 😄

clavedeluna:
> That seems like it would break on Windows... I guess we'll see what the CI thinks about it

implemented, let's see 🤞

pylint/checkers/base/basic_checker.py
```diff
@@ -671,15 +671,23 @@ def _is_terminating_func(node: nodes.Call) -> bool:
 
         inferred = utils.safe_infer(node.func)
 
-        if not inferred and isinstance(node.func, nodes.Attribute) and node.func.attrname == "_exit":
+        if (
```
DanielNoord, line 674: @clavedeluna You might want to install `pre-commit` locally. Although it runs automatically in CI, all the CI jobs that start on the commit that is then rejected by `pre-commit` (in this case https://github.com/PyCQA/pylint/pull/7520/commits/59feeaeae8fdebacd878b197c8ae6a2c995e98b5) do take up a "runner" and some minutes of our CI. When there are multiple people working on various PRs this can block our CI a bit, especially since the Primer jobs block runners for 30+ min. Running `pre-commit` locally prevents taking up those initial runners.

clavedeluna: I was just looking into this bc the number of pre-commit commits is getting annoying. I can run `black` locally, but what's the procedure for using .pre-commit-comfig.yaml locally?

DanielNoord: Check out https://pre-commit.com 😄

clavedeluna:
> > That seems like it would break on Windows... I guess we'll see what the CI thinks about it
> 
> implemented, let's see 🤞

as predicted, windows isn't happy about the check for os._exit. Should I remove it for now? I'm not going to be able to reproduce it locally to debug.

doc/whatsnew/fragments/519.other
```diff
@@ -0,0 +1,3 @@
+Code following call to ``sys.exit`` will be marked as `unreachable`.
```
Pierre-Sassoulas, line 1:
```suggestion
Code following call to ``sys.exit``,  ``quit`` or  ``exit`` will be marked as `unreachable`.
```

Pierre-Sassoulas (changes requested):
Should we also advertise for os._exit even if we might have false positive ? It's supposed to only be used in fork so it's not that much of a problem imo.

pylint/checkers/base/basic_checker.py
```diff
@@ -731,7 +772,9 @@ def leave_tryfinally(self, _: nodes.TryFinally) -> None:
         self._tryfinallys.pop()
 
     def _check_unreachable(
-        self, node: nodes.Return | nodes.Continue | nodes.Break | nodes.Raise
+        self,
+        node: nodes.Return | nodes.Continue | nodes.Break | nodes.Raise | nodes.Call,
+        confidence: Confidence = HIGH,
```
DanielNoord, line 777: Why do we need a default value here? Can't we just make this be required?

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,54 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
     )
     def visit_call(self, node: nodes.Call) -> None:
-        """Visit a Call node -> check if this is not a disallowed builtin
-        call and check for * or ** use.
+        """Visit a Call node.
+
+        1. check if this is not a disallowed builtin call
+        2. check for * or ** use.
+        3. check if this call is a terminating function and is followed by unreachable code
         """
```
DanielNoord, line 705:
```suggestion
        """Visit a Call node."""
```

This was out of date, we don't do 2 anymore and we do more than 1 + 3. Let's just remove this as it has a high risk of becoming out of date again.

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,54 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
+
+        inferred = utils.safe_infer(node.func)
+
+        if (
+            not inferred
+            and isinstance(node.func, nodes.Attribute)
+            and node.func.attrname == "_exit"
+        ):
+            # edge case, try to detect os._exit
+            try:
```
DanielNoord, line 680: Let's just use the `try...except` for all of the inference. This is making two calls to `infer()` which are costly and I'm not sure this is cached correctly if `safe_infer` returns `Uninferable`.

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,54 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
-        """Visit a Call node -> check if this is not a disallowed builtin
-        call and check for * or ** use.
+        """Visit a Call node.
+
+        1. check if this is not a disallowed builtin call
+        2. check for * or ** use.
+        3. check if this call is a terminating function and is followed by unreachable code
         """
+        if self._is_terminating_func(node):
+            self._check_unreachable(node, confidence=INFERENCE)
```
DanielNoord, line 707: Do we need this call? Or can we just call `add_message` directly in `is_termaniting_func`?

tests/functional/u/unreachable.py
```diff
@@ -1,5 +1,7 @@
-# pylint: disable=missing-docstring
-
+# pylint: disable=missing-docstring, too-few-public-methods, redefined-outer-name, consider-using-sys-exit, protected-access
+import os
```
DanielNoord, line 2:
```suggestion

import os
```

DanielNoord (changes requested):

clavedeluna:
> > > That seems like it would break on Windows... I guess we'll see what the CI thinks about it
> > 
> > 
> > implemented, let's see 🤞
> 
> as predicted, windows isn't happy about the check for os._exit. Should I remove it for now? I'm not going to be able to reproduce it locally to debug.

@DanielNoord @Pierre-Sassoulas what decision should we take here to have the functional tests for Windows passing?

pylint/checkers/base/basic_checker.py
clavedeluna: Because in the terminating func case (what I've added in this PR), the confidence is INFERENCE, but in every other case, it's HIGH

DanielNoord: Let's make that explicit in all calling functions.

DanielNoord: Yeah, you're right. I should have checked. I thought the other calls to `_check_unreachable` included similar checks like `is_terminating` but they don't. They just directly call.

Never mind this comment, and good call on pushing back 😄

Pierre-Sassoulas:
> what decision should we take here to have the functional tests for Windows passing? 

I did not follow the full discussion but in last resort we can exclude windows from the functional test for ``os._exit`` by creating a test file specific to it with its own config file.

DanielNoord:
> > what decision should we take here to have the functional tests for Windows passing?

We should figure out what the `qname()` on Windows is and use that. You could probably find that online or use your own local fork on Github + it's CI with some print statements to do so.

pylint/checkers/base/basic_checker.py
clavedeluna: Just to make sure I understand your ask, you're saying to go from

```

        if (
            not inferred
            and isinstance(node.func, nodes.Attribute)
            and node.func.attrname == "_exit"
        ):
            # edge case, try to detect os._exit
            try:
                inferred = next(node.func.infer())
            except (StopIteration, astroid.InferenceError):
                return False
            return inferred.qname() == "posix._exit"
```

to

```
        try:
    
            if (
                not inferred
                and isinstance(node.func, nodes.Attribute)
                and node.func.attrname == "_exit"
            ):
                # edge case, try to detect os._exit
                inferred = next(node.func.infer())
                return inferred.qname() == "posix._exit"
        except Exception:  # pylint: disable=broad-except
            return False
```

pylint/checkers/base/basic_checker.py
clavedeluna: After carefully looking at this option, I don't think we should change the flow. `check_unreachable` does the work of getting the sibling and keeps that concern to itself. Separating that out seems to me it would break separation of concerns.

clavedeluna:
FYI, I was able to get a windows instance via AWS and figured out that edge case, here's some sample code
```
>>> import os
>>> import astroid
>>> node = astroid.extract_node("""import os;os._exit()""")
>>> inferred = next(node.func.infer())
>>> inferred.qname()
Uninferable
>>> inferred
Uninferable
>>> node
<Call l.1 at 0x252d479f070>
>>> node.func
<Attribute._exit l.1 at 0x252d479f0a0>
>>> node.func.infer()
<generator object NodeNG.infer at 0x00000252D46CE7B0>
>>> [x for x in node.func.infer()]
[Uninferable, <FunctionDef._exit l.None at 0x252d47b0a00>]
>>> [x for x in node.func.infer()][1]
<FunctionDef._exit l.None at 0x252d47b0a00>
>>> [x for x in node.func.infer()][1].qname()
'nt._exit'
```

doc/whatsnew/fragments/519.other
```diff
@@ -0,0 +1,3 @@
+Code following call to  ``quit``,  ``exit``, ``sys.exit`` or ``os._exit`` will be marked as `unreachable`.
```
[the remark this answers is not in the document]
DanielNoord: Yeah it's either a call or calls. The subject verb agreement isn't in play here I think because the subject is "calls somewhere in code".

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,49 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
             ):
                 self.add_message("misplaced-format-function", node=call_node)
 
+    @staticmethod
+    def _is_terminating_func(node: nodes.Call) -> bool:
+        """Detect call to exit(), quit(), os._exit(), or sys.exit()."""
+        if (
+            not isinstance(node.func, nodes.Attribute)
+            and not (
```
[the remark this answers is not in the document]
DanielNoord: Yeah but if you re assign `exit` to `another_exit` it won't catch anymore.

clavedeluna:
This relatively small PR has had quite a bit of back and forth. I think it's finally ready to be merged. Maintainers, please feel free to update any code yourselves and merge!

pylint/checkers/base/basic_checker.py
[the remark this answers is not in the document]
clavedeluna: I'm sorry, I don't understand your ask. Please give a specific code example or update it yourself. I think it's fine as is.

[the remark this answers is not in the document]
Pierre-Sassoulas: It's the same remark than in https://github.com/PyCQA/pylint/pull/7520#discussion_r979138466 afaiu

Pierre-Sassoulas:
I think there was a point made in https://github.com/PyCQA/pylint/issues/7563#issuecomment-1266281773 that could be applied here.

pylint/checkers/base/basic_checker.py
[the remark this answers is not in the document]
clavedeluna: I'm going through the added test cases and trying to figure out how to translate them to words, but it may make more sense for you to look at them and provide a test case that would fail the current code.  As written,  the current code in fact works great for `another_exit` function, which is what this added [test](https://github.com/PyCQA/pylint/pull/7520/files) handles. I'm trying to minimize back and forth here but I'm really not understanding how this would not work?

tests/functional/u/unreachable.py
```diff
@@ -32,3 +36,40 @@ def func6():
...
+
+def func10():
+    exit()
+    var = 2 + 2  # [unreachable]
+    print(var)
+
+def func11():
+    quit()
+    var = 2 + 2  # [unreachable]
+    print(var)
```
Pierre-Sassoulas, line 75:
```suggestion

incognito_function = sys.exit

def func12():
    incognito_function()
    var = 2 + 2  # [unreachable]
    print(var)
```

pylint/checkers/base/basic_checker.py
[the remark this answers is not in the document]
Pierre-Sassoulas: I think it's something like this : https://github.com/PyCQA/pylint/pull/7520/files#r987218501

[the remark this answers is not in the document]
clavedeluna: Aha! Thank you PIerre, that's 💯 the feedback I was looking for. That edge case hadn't occurred to me. Will add the test and fix it!

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,38 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
             ):
                 self.add_message("misplaced-format-function", node=call_node)
 
+    @staticmethod
+    def _is_terminating_func(node: nodes.Call) -> bool:
+        """Detect call to exit(), quit(), os._exit(), or sys.exit()."""
```
DanielNoord, line 662:
```suggestion
        """Detect a call to exit(), quit(), os._exit(), or sys.exit()."""
```

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,38 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
+        ):
+            return False
+
+        qnames = {"_sitebuiltins.Quitter", "sys.exit", "posix._exit", "nt._exit"}
+
+        try:
+            for inferred in node.func.infer():
+                if inferred.qname() in qnames:
+                    return True
+        except Exception:  # pylint: disable=broad-except
```
DanielNoord, line 676: This shouldn't be `Exception` but only catch stuff we actually want to catch here. The `pylint` warning exists for a reason 😄

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,38 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
+            and not (isinstance(node.func, nodes.Name))
+            or isinstance(node.parent, nodes.Lambda)
+        ):
+            return False
+
+        qnames = {"_sitebuiltins.Quitter", "sys.exit", "posix._exit", "nt._exit"}
+
+        try:
+            for inferred in node.func.infer():
+                if inferred.qname() in qnames:
```
DanielNoord, line 674: Not all types returned by `infer()` have a `qname()`. We should look for a specific type of `node` with `isinstance` and then use `qname`.

tests/functional/u/unreachable.py
```diff
@@ -32,3 +36,46 @@ def func6():
...
+
+incognito_function = sys.exit
```
DanielNoord, line 77: I can't find the previous discussion, but to reply to one of your earlier questions: `another_exit` worked because it ends with `exit` 😄 Since this ends with `function` it shows we really do need that inference here.

DanielNoord (changes requested):

clavedeluna:
@DanielNoord @Pierre-Sassoulas I've reached my capacity with this PR. The back and forth and amount of conflicting suggestions has been really frustrating. It's best if you can provide exact code and test cases that you're suggesting, instead of very general advice that you later disagree with. Don't expect us contributors to be able to read your mind.

Please feel free to take over this PR.

pylint/checkers/base/basic_checker.py
clavedeluna: > Let's just use the try...except for all of the inference. 

then the code base actually has some uses of `except Exception:  # pylint: disable=broad-except`.

Please beware of your conflicting reviews. It's really frustrating for contributors.

DanielNoord: I don't understand. We know which exceptions we want to catch here right? `StopIteration`, `InferenceError`? We shouldn't be using a broad except here.

DanielNoord:
> @DanielNoord @Pierre-Sassoulas I've reached my capacity with this PR. The back and forth and amount of conflicting suggestions has been really frustrating. It's best if you can provide exact code and test cases that you're suggesting, instead of very general advice that you later disagree with. Don't expect us contributors to be able to read your mind.

There is always a little risk of maintainers have different opinions about certain topics in open source projects.

I think in this case the different opinions where actually the result of the PR slowly growing in size and scope along the way. This causes different concerns to be raised. For example, the `infer()` call was "incorrect" when we didn't need the sub-results but when it then turned out that `sys.exit` is different on different platforms we had to use it again in order not to pay additional performance costs. I don't think maintainers can foresee such issues when reviewing.

That said I will take care to make sure I don't conflict other maintainers without discussion in future reviews.

> Please feel free to take over this PR.

That's too bad. I think we're nearing completion on this PR. I'm sorry you feel this way.

Pierre-Sassoulas:
Sorry about that @clavedeluna, we have a lot of code to review each week and mistake can happen when we do not have the time to read other maintainers review. Please do point us to the contradiction so we can discuss it between disagreeing maintainers.

clavedeluna:
Reviewers, I'm finishing this PR up with what I believe were the last comments.

pylint/checkers/base/basic_checker.py
```diff
@@ -657,13 +657,38 @@ def _check_misplaced_format_function(self, call_node: nodes.Call) -> None:
...
+            or isinstance(node.parent, nodes.Lambda)
+        ):
+            return False
+
+        qnames = {"_sitebuiltins.Quitter", "sys.exit", "posix._exit", "nt._exit"}
+
+        try:
+            for inferred in node.func.infer():
+                if hasattr(inferred, "qname") and inferred.qname() in qnames:
+                    return True
```
clavedeluna, line 675: I'll be the first to say using hasattr isn't ideal and I did attempt isinstance, however:

for the cases of node being built builtin exit or quit, these get an inferred instance if :

```
(Pdb) inferred
<Instance of _sitebuiltins.Quitter at 0x4428041360>
(Pdb) from _sitebuiltins import Quitter
(Pdb)  isinstance(inferred, Quitter)
False
```

so as much as I tried to do `if isinstance(inferred, (nodes.FunctionDef, Quitter)` or something similar, it never worked. so hasattr was the best I could get to work

doc/whatsnew/fragments/519.other
```diff
@@ -0,0 +1,3 @@
+Code following a call to  ``quit``,  ``exit``, ``sys.exit`` or ``os._exit`` will be marked as `unreachable`.
```
Pierre-Sassoulas, line 1: This should be a false négative fragment

Pierre-Sassoulas (changes requested):
LGTM, small nitpick and let's merge.

Pierre-Sassoulas (approved):

Pierre-Sassoulas (dismissed):
Thank you @clavedeluna 🎉

Pierre-Sassoulas:
Rebased on upstream/main so the pipeline can run properly.

Pierre-Sassoulas (approved):

pylint/checkers/base/basic_checker.py
DanielNoord: This seems to be a case of `astroid.Instance` right? We could just instance check for that?

If possible I'd like us to do a follow-up, since the `hasattr` doesn't provide actual type safety..

PR #7521, opened 2022-09-23
fix: test_logger_rcfile: replace only the last .py token in the string

sandrotosi:
## Type of Changes

## Description

Refs #XXXX

Closes #XXXX


sandrotosi:
i created this via the gh online editor, not even sure you like this approach (but there are none super-appealing anyway)

tests/config/test_argparse_config.py
```diff
@@ -43,7 +43,10 @@ def test_logger_commandline() -> None:
     def test_logger_rcfile() -> None:
         """Check that we parse the rcfile for the logging checker correctly."""
         with pytest.raises(SystemExit) as ex:
-            Run([LOGGING_TEST, f"--rcfile={LOGGING_TEST.replace('.py', '.rc')}"])
+            # replace only the last .py in the string with .rc
+            # we do so by inverting the string and replace the first occurence (of the inverted tokens!)
```
Pierre-Sassoulas, line 47:
```suggestion
            # we do so by inverting the string and replace the first occurrence (of the inverted tokens!)
```

tests/config/test_argparse_config.py
```diff
@@ -43,7 +43,10 @@ def test_logger_commandline() -> None:
     def test_logger_rcfile() -> None:
         """Check that we parse the rcfile for the logging checker correctly."""
         with pytest.raises(SystemExit) as ex:
-            Run([LOGGING_TEST, f"--rcfile={LOGGING_TEST.replace('.py', '.rc')}"])
+            # replace only the last .py in the string with .rc
+            # we do so by inverting the string and replace the first occurrence (of the inverted tokens!)
+            _rcfile = LOGGING_TEST[::-1].replace("yp.", "cr.", 1)[::-1]
```
Pierre-Sassoulas, line 48: I was searching for an elegant way to do that and saw : https://stackoverflow.com/a/2556156/2519059 😄 You've been there before me 😉 ?

Pierre-Sassoulas (approved):
Thank you ! How did you notice it was a problem if you don't mind ?

tests/config/test_argparse_config.py
sandrotosi: yep!

sandrotosi:
> Thank you ! How did you notice it was a problem if you don't mind ?

funny story, in debian we have a process called `pybuild` that takes care of building the package, and it does so by making all its operation in a directory called... drumroll... `.pybuild` :) so that was being replaced with `.rcbuild` and it was making the build fail

Pierre-Sassoulas:
We break your build in mysterious way ! 

The pipeline fail is not due to your changes, I'll take care of the cache issue, don't worry about it.

sandrotosi:
> The pipeline fail is not due to your changes, I'll take care of the cache issue, don't worry about it.

much appreciated :)

PR #6787, opened 2022-06-01
Prevent `unused-import` for `if t.TYPE_CHECKING`

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

The check for `typing.TYPE_CHECKING` was not robust against importing `typing` as `t`. Nor against importing `keyboarding` as `typing`. This could lead to false positives for `unused-import`.

Follow-up: we have the following plethora of utils. Would be good to standardize and deprecate some for 2.15 (not a backport task, though):
 - is_typing_guard()
 - is_node_in_typing_guarded_import_block()
 - in_type_checking_block()  <-- **this one** edited in this PR
 

Closes #3846

PR #7525, opened 2022-09-24
Prevent `redefined-outer-name` for `if t.TYPE_CHECKING`

### LINKED ISSUE #7524: TYPE_CHECKS_GUARDS fooled by import alias
### Bug description

```
# pylint: disable=missing-module-docstring,missing-class-docstring,too-few-public-methods                               
# pylint: disable=missing-function-docstring                                                                            
import typing as t

class Cls:
    def func(self, dd: defaultdict):
        # This import makes the definition work.                                                                        
        from collections import defaultdict

        obj = defaultdict()
        obj.update(dd)
        return obj

if t.TYPE_CHECKING:
    # This import makes the annotations work.                                                                           
    from collections import defaultdict
```

### Configuration

Default

### Command used

```shell
pylint redefined_outer.py
```

### Pylint output

```shell
************* Module redefined_outer
redefined_outer.py:10: [W0621(redefined-outer-name), Cls.func] Redefining name 'defaultdict' from outer scope (line 19)
```

### Expected behavior

```
```

Whether TYPE_CHECKING is imported from typing or called as typing.TYPE_CHECKING or imported as an alias should not matter to whether TYPE_CHECKING is called.

I looked at the code and realize that this is a difficult bug to fix -- it probably would involve putting a Mock around typing.TYPE_CHECKING before the module is imported so that the ast tree can report whether it's in a TYPE_CHECKING context or not.  This might be a bug left open for a while, but I do think it's an actual issue that might not be resolved until Python has a syntax similar to TypeScript's "from collections import type defaultdict"

I've noticed the same issue with @t.overload on mypy not properly recognizing it as a overload decorator, so the solution was something I could see.

Because it can be a pain to constantly update a `from typing import ....` statement at the top of a module when adding types, my project, `music21` uses a standard of importing typing as t at the top of each module so it's easy to know that `t.Union[t.Dict[t.Optional[str]]` will always be available.  (When our min Py version goes to 3.9/3.10, there'll be a lot less of this).  

Thanks for an amazing tool!

### Pylint version

```shell
pylint 2.15.3
astroid 2.12.10
Python 3.10.6 (v3.10.6:9c7b4bd164, Aug  1 2022, 17:13:48) [Clang 13.0.0 (clang-1300.0.29.30)]
```

### OS / Environment

MacOS 12.6

### Additional dependencies

None

jacobtylerwalls:
## Type of Changes

## Description

Closes #7524
Similar solution to #6787

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1230,9 +1230,7 @@ class VariablesChecker(BaseChecker):
                 # Do not take in account redefined names for the purpose
                 # of type checking.:
                 if any(
-                    isinstance(definition.parent, nodes.If)
-                    and definition.parent.test.as_string() in TYPING_TYPE_CHECKS_GUARDS
-                    for definition in globs[name]
+                    in_type_checking_block(definition) for definition in globs[name]
                 ):
                     continue
 
--- a/doc/whatsnew/fragments/7524.bugfix
+++ b/doc/whatsnew/fragments/7524.bugfix
@@ -1,0 +1,4 @@
+Fix false positive for ``redefined-outer-name`` when aliasing ``typing``
+e.g. as ``t`` and guarding imports under ``t.TYPE_CHECKING``.
+
+Closes #7524
--- a/tests/functional/r/redefined/redefined_outer_name_type_checking.py
+++ b/tests/functional/r/redefined/redefined_outer_name_type_checking.py
@@ -3,19 +3,27 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
+import typing as t
 
 
 class Cls:
-    def func(self, stuff: defaultdict):
-        # This import makes the definition work.
+    def func(self, stuff: defaultdict, my_deque: deque):
+        # These imports make the definition work.
         # pylint: disable=import-outside-toplevel
         from collections import defaultdict
+        from collections import deque
 
         obj = defaultdict()
+        obj2 = deque()
         obj.update(stuff)
+        obj2.append(my_deque)
         return obj
 
 
 if TYPE_CHECKING:
     # This import makes the annotations work.
     from collections import defaultdict
+
+if t.TYPE_CHECKING:
+    # This import makes the annotations work.
+    from collections import deque


pylint/checkers/variables.py
```diff
@@ -1230,9 +1230,7 @@ def visit_functiondef(self, node: nodes.FunctionDef) -> None:
                 # Do not take in account redefined names for the purpose
                 # of type checking.:
                 if any(
-                    isinstance(definition.parent, nodes.If)
-                    and definition.parent.test.as_string() in TYPING_TYPE_CHECKS_GUARDS
-                    for definition in globs[name]
+                    in_type_checking_block(definition) for definition in globs[name]
```
DanielNoord, line 1233: Not sure if this is in scope, but there is only one reference to `TYPING_TYPE_CHECKS_GUARDS` remaining. Should we remove it as well and use this function everywhere?

jacobtylerwalls: I think that should be a follow-up PR. At first glance I'm not sure there's a drop-in replacement.

DanielNoord:
```diff
diff --git a/pylint/checkers/variables.py b/pylint/checkers/variables.py
index 65a9e8c8b..59e9f7fba 100644
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2011,10 +2011,7 @@ class VariablesChecker(BaseChecker):
             if isinstance(defstmt, (nodes.Import, nodes.ImportFrom)):
                 defstmt_parent = defstmt.parent
 
-                if (
-                    isinstance(defstmt_parent, nodes.If)
-                    and defstmt_parent.test.as_string() in TYPING_TYPE_CHECKS_GUARDS
-                ):
+                if in_type_checking_block(defstmt):
                     # Exempt those definitions that are used inside the type checking
                     # guard or that are defined in both type checking guard branches.
                     used_in_branch = defstmt_parent.parent_of(node)

```

Passed all functional tests for me locally.

jacobtylerwalls: I'm concerned there aren't enough tests. This example depends on `defstmt_parent` being `nodes.If`. `in_type_checking_block` is more robust, and doesn't assume direct parentage. There's more control flow in this example; it needs more thought.

Pierre-Sassoulas (approved):
LGTM, I'll let you merge when the remaining discussion is resolved.

pylint/checkers/variables.py
DanielNoord: Ah yeah, my bad. What about:
```
if isinstance(defstmt_parent, nodes.If) and in_type_checking_block(defstmt):
```

The `in_type_checking_block` doesn't do any control flow analysis and only makes sure we infer instead of relying on the string representation. As far as I see it is a drop-in replacement for `defstmt_parent.test.as_string() in TYPING_TYPE_CHECKS_GUARDS`. (But not for `isinstance(defstmt_parent, nodes.If)` as you correctly spotted!)

jacobtylerwalls: I still don't think it's a drop-in replacement because `in_type_checking_block()` is recursive. We don't know which `if` ancestor met the condition, but the way this block is written, the assumption is that `defstmt_parent` is the one that met the condition, and doing a drop-in replacement will violate that condition. Leading me to think it needs more refactoring and a follow-up issue.

DanielNoord: Fair enough. Shall we open an issue/todo for this? I'm also fine with leaving as is.

jacobtylerwalls: No problem, I'll do it. 👍🏻

jacobtylerwalls: #7539

PR #7526, opened 2022-09-25
add `magic-number` checker for comparison with literals

### LINKED ISSUE #7281: Suggestion - checker that detects magic numbers/unnamed constants
### Current problem

I thought it would be nice to have a checker that detects magic numbers (aka. unnamed constants) in `Pylint`.

 As in all standard languages, `-1`, `0` and `1` are allowed by default. Other numbers should raise a message unless would be specified in the `pylintrc` file. 
 
As not all might want to use it and since it is just really a bad practice and never an error it should be considered as an extension instead of core checker.

According to my test there is nothing like that in `Pylint`, but I'd like to be proven wrong :)

Example:

```
seconds = 60

func(a, b, 50)

```
shows no error related to magic number

### Desired solution

```
1: seconds = 60

2: func(a, b, 50)

```

should show 2 messages of type `Refactor` (it is never an error just a bad practice in my opinion).

`1: "Unnamed constant '60' is used, consider using literal instead of magic numbers."`
`2: "Unnamed constant '50' is used, consider using literal instead of magic numbers."`

### Additional context

Similar issue was opened but it was a bit different and more complicated than what I suggest here, see #1379

Similar checkers exist in [ESlint](https://eslint.org/docs/latest/rules/no-magic-numbers) and [TSlint](https://palantir.github.io/tslint/rules/no-magic-numbers/).

orSolocate:
## Type of Changes

|   | :bug: Bug fix          |

## Description
An extension of `comparison-of-constants` to check one sided comparison of magic numbers that should be literals.
ignores:
1. -1,0,1, '', and  \__main__ cases
2.  both sides are constants. this is taken care by  `comparison-of-constants`
3. list/set/dict of constants (maybe a checker extension)

Closes #7281


orSolocate:
Will take forever to fix all Pylint source code magic number issues :(
https://github.com/PyCQA/pylint/actions/runs/3123350498/jobs/5065943521

DanielNoord:
@orSolocate As said in the in the original issue I think it makes more sense to have this as an optional extension. This is (probably) too opinionated to be part of the main checkers. Personally I would always disable it and consider it similar to the `while-checker`: I get the point, but sometimes it's just much easier to use a magic number. That shouldn't keep us from adding it as an extension though.

orSolocate:
@DanielNoord  - I will make a new commit of it as an extension, that will also solve the CI running pylint

orSolocate:
any idea how to fix the docs jobs?

pylint/checkers/base/comparison_checker.py
```diff
@@ -15,6 +15,7 @@
 COMPARISON_OPERATORS = frozenset(("==", "!=", "<", ">", "<=", ">="))
 TYPECHECK_COMPARISON_OPERATORS = frozenset(("is", "is not", "==", "!="))
 TYPE_QNAME = "builtins.type"
+SINGLETON_VALUES = {True, False, None}
```
Pierre-Sassoulas, line 18: I don't like this pattern, it adds coupling, and we never know if those constant can be removed even if the corresponding code do not exists anymore. Unless we use it twice or thrice it should not be a constant.

pylint/checkers/base/comparison_checker.py
```diff
@@ -89,16 +96,10 @@ def _check_singleton_comparison(
         checking_for_absence: bool = False,
     ) -> None:
         """Check if == or != is being used to compare a singleton value."""
-        singleton_values = (True, False, None)
-
-        def _is_singleton_const(node: nodes.NodeNG) -> bool:
```
Pierre-Sassoulas, line 94: If it's used only there it's good that it is an inner function, it's hard to know if you can remove a class function when it's no longer used.

pylint/extensions/magic_number.py
```diff
@@ -0,0 +1,82 @@
...
+from astroid import nodes
+
+from pylint.checkers import BaseChecker, utils
+from pylint.checkers.base.comparison_checker import ComparisonChecker
+from pylint.interfaces import HIGH
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+LEGAL_MAGIC_NUMBERS = {-1, 0, 1, "", "__main__"}
```
Pierre-Sassoulas, line 20: This should be a configurable option imo.

tests/functional/ext/magic_comparison/magic_comparison.py
```diff
@@ -0,0 +1,23 @@
+"""
+Checks that magic numbers are not used in comparisons
+"""
+# pylint: disable=missing-docstring,invalid-name,too-few-public-methods,import-error,wrong-import-position, comparison-of-constants, singleton-comparison
```
Pierre-Sassoulas, line 4:
```suggestion
# pylint: disable=missing-docstring,invalid-name,too-few-public-methods,import-error,wrong-import-position
```
This would help make pylint behavior clearer when reading the functional test, right ?

Pierre-Sassoulas (changes requested):
Looks like a nice new opinionated checker :) Could you revert the changes in the comparison checker please ? Also I think some more functional test would be a good thing, in particular dict or list with only some of the element that are "magic" and some named variables too (i.e. ``["apple", "orange", banana]``)

pylint/extensions/magic_number.py
```diff
@@ -0,0 +1,82 @@
...
+
+class MagicNumberChecker(BaseChecker):
+    """Checks for constants in comparisons."""
+
+    name = "magic-number"
+    msgs = {
+        "R2004": (
+            "Unnamed constant %s is used, consider using literal instead of magic numbers.",
+            "magic-comparison",
+            "Used when magic numbers are used in any side of a comparison ",
```
Pierre-Sassoulas, line 31: The "WARNING: Inline emphasis start-string without end-string." might comes from there, not sure. In any case this is the place where we argue about the usefulness of not using magic values, this could do with more details and reasonning.

doc/whatsnew/fragments/7281.new_check
```diff
@@ -0,0 +1,3 @@
+add ``magic-number`` plug-in checker for comparison with constants instead of literals.
```
Pierre-Sassoulas, line 1:
```suggestion
Add ``magic-number`` plug-in checker for comparison with constants instead of literals.
```

pylint/extensions/magic_number.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class MagicNumberChecker(BaseChecker):
+    """Checks for constants in comparisons."""
+
+    name = "magic-number"
+    msgs = {
+        "R2004": (
+            "Unnamed constant %s is used, consider using literal instead of magic numbers.",
+            "magic-comparison",
```
Pierre-Sassoulas, line 30:
```suggestion
            "magic-value-used-in-comparison",
```
I think the accepted software engineering term for unnamed constant is "magic values" or "magic numbers"

pylint/checkers/base/comparison_checker.py
orSolocate: I use it in my `magic_number,py` checker, I didn't won't to duplicate the function. what should I do then?

pylint/extensions/magic_number.py
orSolocate: good idea, do you mean from the pylintrc file? how do I do that?

tests/functional/ext/magic_comparison/magic_comparison.py
orSolocate: probably, and add the messages in the `.txt` file? even if it is not my checker?

pylint/checkers/base/comparison_checker.py
Pierre-Sassoulas: My bad, I missed that, how about we keep the function but inline the constant ?

pylint/extensions/magic_number.py
Pierre-Sassoulas: Yes or the CLI :) There's an example here : https://github.com/PyCQA/pylint/blob/main/pylint/checkers/classes/class_checker.py#L773

pylint/extensions/magic_number.py
orSolocate: I went with `magic-number-comparison`

orSolocate:
> Looks like a nice new opinionated checker :) Could you revert the changes in the comparison checker please ? Also I think some more functional test would be a good thing, in particular dict or list with only some of the element that are "magic" and some named variables too (i.e. `["apple", "orange", banana]`)

for 
`if item in ["apple", "orange", banana]`
would we like to show 2 errors for the same line? one for `apple` and one for `orange`?

orSolocate:
> Could you also add an example in the doc with bad/good.py, please

good idea :D

Pierre-Sassoulas:
> ``if item in ["apple", "orange", banana]`` would we like to show 2 errors for the same line? one for apple and one for orange?

Hmm I was thinking none initially but I'm probably not going to use this checker myself and you're right that if you follow the spirit of it, then there's two violations.

DanielNoord:
@orSolocate Which strategy did you use to merge main? With your current strategy you create new commits for each change. `git merge upstream main --no-ff` should avoid that. (With `upstream` being the remote for `PyCQA`).

orSolocate:
> @orSolocate Which strategy did you use to merge main? With your current strategy you create new commits for each change. `git merge upstream main --no-ff` should avoid that. (With `upstream` being the remote for `PyCQA`).

just fixed that, no worries

orSolocate:
@Pierre-Sassoulas  - I think we can keep the functionality simple for now without the support for lists/dicts/sets that includes constants in comparison.. correct me if you think it is important..

DanielNoord:
> > @orSolocate Which strategy did you use to merge main? With your current strategy you create new commits for each change. `git merge upstream main --no-ff` should avoid that. (With `upstream` being the remote for `PyCQA`).
> 
> just fixed that, no worries

For the future it's often better to merge instead of rebase. The GitHub UI doesn't always deal nicely with rebases so reviewers can lose track of which changes they have already reviewed. Merging `main` into your branch resolves this issue.

Pierre-Sassoulas:
>  I think we can keep the functionality simple for now without the support for lists/dicts/sets that includes constants in comparison.. correct me if you think it is important..

Sure, let's favor false negatives and never raise on complicated values then.

orSolocate:
@Pierre-Sassoulas  I have a problem registering the new 'option' for `pylintrc`, could you please tell me where should I add it? except for the checker class..

doc/data/messages/m/magic-value/good.py
```diff
@@ -0,0 +1,15 @@
+import random
+
+MAX_NUM_OF_ITERATIONS = 5
+THRESHOLD_VAL = 100
+MIN_MEASUREMNET_VAL = 0
+MAX_MEASUREMNET_VAL = 200
```
jacobtylerwalls, line 6:
```suggestion
MIN_MEASUREMENT_VAL = 0
MAX_MEASUREMENT_VAL = 200
```

doc/data/messages/m/magic-value/good.py
```diff
@@ -0,0 +1,15 @@
...
+
+measurement = random.randint(MIN_MEASUREMNET_VAL, MAX_MEASUREMNET_VAL)
```
jacobtylerwalls, line 8:
```suggestion
measurement = random.randint(MIN_MEASUREMENT_VAL, MAX_MEASUREMENT_VAL)
```

doc/data/messages/m/magic-value/good.py
```diff
@@ -0,0 +1,15 @@
...
+above_threshold = False
+i = 0
+while i < MAX_NUM_OF_ITERATIONS:
+    above_threshold = measurement > THRESHOLD_VAL
+    if above_threshold:
+        break
+    measurement = random.randint(MIN_MEASUREMNET_VAL, MAX_MEASUREMNET_VAL)
```
jacobtylerwalls, line 15:
```suggestion
    measurement = random.randint(MIN_MEASUREMENT_VAL, MAX_MEASUREMENT_VAL)
```

jacobtylerwalls (changes requested):
"Literal" doesn't seem like the best descriptor for the alternative. `42` is a literal. I think we should suggest "named constant or enum".

tests/functional/u/used/used_before_assignment_conditional.py
```diff
@@ -1,6 +1,6 @@
 """used-before-assignment cases involving IF conditions"""
 
-# pylint: disable=consider-using-augmented-assign
+# pylint: disable=consider-using-augmented-assign\
```
jacobtylerwalls, line 3:
```suggestion
# pylint: disable=consider-using-augmented-assign
```

jacobtylerwalls (commented):
In fact, some tests with enums would be good.

pylint/extensions/magic_value.py
```diff
@@ -0,0 +1,91 @@
...
+
+from pylint.checkers import BaseChecker, utils
+from pylint.checkers.base.comparison_checker import ComparisonChecker
+from pylint.interfaces import HIGH
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class MagicValueChecker(BaseChecker):
```
Pierre-Sassoulas, line 21:
```suggestion
class MagicValueChecker(BaseChecker, _ArgumentsProvider):
```

I thought this was required to auto register options. But it's not working and I had to read some code to find about that. @DanielNoord do you know how to register the options ? I'm going to document this in the contributor guide because I don't think there's anything yet.

DanielNoord: `BaseChecker` inherits from `_ArgumentsProvider` itself.

@orSolocate  I believe the issue is because of the use of `"` in the `pylintrc` file. Strings don't need to be quoted.

Pierre-Sassoulas: I expected to see the new option in the ``--help`` is it supposed to be printed ?

DanielNoord: Yes, isn't it?

Pierre-Sassoulas: Let me recheck that :)

Pierre-Sassoulas: My bad we need to load the plugin first of course.
```
(venv)  pierre@pylint  ~/pylint   orsolocate/magic_numbers  pylint --help|grep magic                                             
(venv)  ✘ pierre@pylint  ~/pylint   orsolocate/magic_numbers  
pylint --load-plugins=pylint.extensions.magic_value --help|grep magic
  --valid-magic-values <method names>
                        List of valid magic numbers that `magic-value-compare`
```

tests/functional/u/used/used_before_assignment_conditional.py
orSolocate: what is the suggestion?

orSolocate: I added a line with enum comparison could you please verify this is what you meant?

orSolocate:
@DanielNoord - this is what you meant by merge instead of rebase?

pylint/extensions/magic_value.py
orSolocate: I tried to remove the `""`, still not working.. also the docs test is failing and I don't see why, there was an issue with not adding `.rc` file in `good/bad.py`  but  I thought that was fixed.

orSolocate: > `BaseChecker` inherits from `_ArgumentsProvider` itself.
> 
> @orSolocate I believe the issue is because of the use of `"` in the `pylintrc` file. Strings don't need to be quoted.

How do you do an empty string then? maybe there was an error there?
`valid-magic-values=-1, 0, 1,, __main__`

DanielNoord:
> @DanielNoord - this is what you meant by merge instead of rebase?

Yes! Thank you. Just to show you, this is what the GitHub UI shows me now:

image

That makes reviewing much easier!

pylint/extensions/magic_value.py
DanielNoord: I'm not sure `ini` files support empty strings. I'll have a look at this somewhere this week. Probably around the weekend, bit busy currently.

tests/functional/u/used/used_before_assignment_conditional.py
orSolocate: No just a mistake probably because I added it as a core checker and then converted it to a plug-in checker. Deleting all the --disable in all tests, maybe I pressed `\` instead of `del` button in the keyboard..

doc/data/messages/m/magic-value/good.py
jacobtylerwalls: I'm still seeing the typos.

pylint/extensions/magic_value.py
orSolocate: I ran `tox -e docs` like in the good old days and found that the documentation issue was because there was a missing `.` in the end of line in checker description..

The `primer` test fails because when run it shows this error, look like parsing error:
`pylint: error: argument --valid-magic-values: expected one argument`

@DanielNoord  - Could you help me with it please?

P.S. why did we remove the `tox -e docs` to check docs locally from contribution page? Maybe it is a good idea to put it back..

jacobtylerwalls (dismissed):

doc/whatsnew/fragments/7281.new_check
```diff
@@ -0,0 +1,3 @@
+Add ``magic-number`` plug-in checker for comparison with constants instead of literals.
```
Pierre-Sassoulas, line 1:
```suggestion
Add ``magic-number`` plug-in checker for comparison with constants instead of literals.
You can use it with ``--load-plugins=pylint.extensions.magic_value``
```

pylint/checkers/base/comparison_checker.py
```diff
@@ -243,14 +252,14 @@ def _check_logical_tautology(self, node: nodes.Compare) -> None:
             suggestion = f"{left_operand} {operator} {right_operand}"
             self.add_message("comparison-with-itself", node=node, args=(suggestion,))
 
-    def _check_two_literals_being_compared(self, node: nodes.Compare) -> None:
-        """Check if two literals are being compared; this is always a logical tautology."""
+    def _check_constants_comparison(self, node: nodes.Compare) -> None:
+        """When two constants are being compared it is always a logical tautology."""
         left_operand = node.left
-        if not isinstance(left_operand, nodes.Const):
+        if not (isinstance(left_operand, nodes.Const)):
```
Pierre-Sassoulas, line 258: The parenthesis are not required right ?
```suggestion
        if not isinstance(left_operand, nodes.Const):
```

pylint/checkers/base/comparison_checker.py
```diff
@@ -243,14 +252,14 @@ def _check_logical_tautology(self, node: nodes.Compare) -> None:
...
             return
 
         right_operand = node.ops[0][1]
-        if not isinstance(right_operand, nodes.Const):
+        if not (isinstance(right_operand, nodes.Const)):
```
Pierre-Sassoulas, line 262:
```suggestion
        if not isinstance(right_operand, nodes.Const):
```

pylint/checkers/base/comparison_checker.py
```diff
@@ -308,7 +317,7 @@ def visit_compare(self, node: nodes.Compare) -> None:
 
         if operator in {"==", "!="}:
             self._check_singleton_comparison(
-                left, right, node, checking_for_absence=operator == "!="
+                left, right, node, checking_for_absence=(operator == "!=")
```
Pierre-Sassoulas, line 320:
```suggestion
                left, right, node, checking_for_absence=operator == "!="
```

pylint/extensions/magic_value.py
Pierre-Sassoulas: I don't think we removed tox -e docs ?

Pierre-Sassoulas (changes requested):

pylint/extensions/magic_value.py
DanielNoord: The primer issue is because of the default. By using `-1,0,...` there is a `-` at the beginning of the value and `argparse` chokes on this. Changing it around to have the `-1` at the second index fixes the issues and parses correctly.

orSolocate: > I don't think we removed tox -e docs ?

From [here ](https://pylint.pycqa.org/en/latest/development_guide/contributor_guide/tests/launching_test.html)i meant.

That is were it used to be.

Anyway I would assume it would be under the [testing chapter](https://pylint.pycqa.org/en/latest/development_guide/contributor_guide/tests/index.html) but I can't find it.. If it exist somewhere else it should be there as well I think..

Pierre-Sassoulas: Right, even if tox -e docs still exits for the CI, in the doc we documented something faster: https://pylint.pycqa.org/en/latest/development_guide/contributor_guide/contribute.html#building-the-documentation

orSolocate: > The primer issue is because of the default. By using `-1,0,...` there is a `-` at the beginning of the value and `argparse` chokes on this. Changing it around to have the `-1` at the second index fixes the issues and parses correctly.

Thanks for checking it for me, the `primer` tests pass now.
The `tests` section fails though, seems like for parsing error of the same line, this is the error message I saw [in CI](https://github.com/PyCQA/pylint/actions/runs/3218362280/jobs/5262443518):

```
usage: pylint-config [options]
pylint-config: error: argument config_subcommand: invalid choice: '0,-1,1,__main__' (choose from 'generate')
```

@DanielNoord  - Could you please help me out?

orSolocate:
I updated the checker and removed `""` from default to let the parsing complete.

Although I do think that `""`, `[]`  and `{}` should be ignored, they shouldn't be considered as magic values for this checker in my opinion.

orSolocate:
> > @DanielNoord - this is what you meant by merge instead of rebase?
> 
> Yes! Thank you. Just to show you, this is what the GitHub UI shows me now:
> 
> image
> 
> That makes reviewing much easier!

Do we still rebase before the PR gets merged to `main`?

Pierre-Sassoulas:
> Do we still rebase before the PR gets merged to main?

It depends on the MR but this one is probably going to be squashed as a single commit with a new checker is atomic and coherent.

Pierre-Sassoulas:
> and you could do the squash at the same time of merging the PR through GitHub, right?

Yes, no need to clean the history up, if the commit messages are clear enough so that we can synthesize / triage them in the squash commit message it's enough.

pylint/extensions/magic_value.py
DanielNoord: Tests pass if you remove the value from the `pylintrc` file. I'm not sure why and the error isn't really elegant but I can explain some part of it: your adding configuration for a plugin that is not being loaded in that `pylintrc` so `pylint` doesn't know what to do with those values.
It's very weird to throws this particular error and we should probably fix it, but removing would fix this.

orSolocate: That makes sense actually, because it is an extension checker, it is not registers by default including the `options` section of it.
I will add to my default also `[]`, `{}` and `""`, because otherwise using this checker is even more annoying. 

Do you want me to open a bug for unclear message with pylintrc parsing error?

DanielNoord: Yeah could you open an issue for it? This is really ugly.

orSolocate:
@Pierre-Sassoulas  - CI looks good now :)

doc/data/messages/m/magic-value/good.py
```diff
@@ -0,0 +1,15 @@
+import random
```
Pierre-Sassoulas, line 1: The example is good and almost made me reconsider using the check 😄

pylint/checkers/base/comparison_checker.py
Pierre-Sassoulas: The conversation is resolved but it does not seem the fix was applied ? (Same for https://github.com/PyCQA/pylint/pull/7526/files#r990597254 and https://github.com/PyCQA/pylint/pull/7526/files#r990597282)

pylint/extensions/magic_value.py
```diff
@@ -0,0 +1,91 @@
...
+    """Checks for constants in comparisons."""
+
+    name = "magic-value"
+    msgs = {
+        "R2004": (
+            "Constant %s is used in comparison. Consider using named constants or enums instead of magic values.",
```
Pierre-Sassoulas, line 27:
```suggestion
            "Consider using a named constant or an enum instead of '%s'."
```
The message feel a little long and the term magic values is already in ``magic-value-compare`` so I feel like it could be shortened.

pylint/extensions/magic_value.py
```diff
@@ -0,0 +1,91 @@
...
+                args=(left_operand.value),
+                confidence=HIGH,
+            )
+        elif const_operands[RIGHT_OPERAND] and self._is_magic_value(right_operand):
+            self.add_message(
+                "magic-value-compare",
+                node=node,
+                args=(right_operand.value),
+                confidence=HIGH,
+            )
```
Pierre-Sassoulas, line 78:
```suggestion
        operand_value = None
        if const_operands[LEFT_OPERAND] and self._is_magic_value(left_operand):
            operand_value  = left_operand.value
        elif const_operands[RIGHT_OPERAND] and self._is_magic_value(right_operand):
            operand_value  = right_operand.value
        if operand_value is not None:
            self.add_message(
                "magic-value-compare",
                node=node,
                args=(operand_value),
                confidence=HIGH
            )
```
Little refactor to appease my mind :)

Pierre-Sassoulas (changes requested):
Look pretty good already, thank you ! I have some nitpicks and the message change can be discussed before being applied because it will require to regenerate the functional tests result.

doc/data/messages/m/magic-value/good.py
orSolocate: Thanks mate!

pylint/extensions/magic_value.py
orSolocate: you are right about the parenthesis, my bad I changed it for the locally built file. now it will be resolved for real.

orSolocate:
> Look pretty good already, thank you ! I have some nitpicks and the message change can be discussed before being applied because it will require to regenerate the functional tests result.

Thanks for the review! 

No worries about regenerating the expected result for the test, it is done with a single command command:

`python tests/test_functional.py --update-functional-output -k "test_functional[magic_value_compare]"`

Pierre-Sassoulas (dismissed):

doc/whatsnew/fragments/7281.new_check
```diff
@@ -0,0 +1,4 @@
+Add ``magic-number`` plug-in checker for comparison with constants instead of named constants or enums.
```
DanielNoord, line 1: Name should be updated here :smile: I think `plug-in` can be written as `plugin`.

pylint/checkers/base/comparison_checker.py
```diff
@@ -81,6 +82,12 @@ class ComparisonChecker(_BasicChecker):
         ),
     }
 
+    @staticmethod
+    def is_singleton_const(node: nodes.NodeNG) -> bool:
+        return isinstance(node, nodes.Const) and any(
```
DanielNoord, line 87: `node.value is in SINGLETON_VALUES`, right?

pylint/checkers/base/comparison_checker.py
```diff
@@ -163,9 +164,10 @@ def _is_float_nan(node: nodes.NodeNG) -> bool:
                 return False
 
         def _is_numpy_nan(node: nodes.NodeNG) -> bool:
+            NUMPY_COMMON_LIB_NAMES = {"numpy", "nmp", "np"}
             if isinstance(node, nodes.Attribute) and node.attrname == "NaN":
                 if isinstance(node.expr, nodes.Name):
-                    return node.expr.name in {"numpy", "nmp", "np"}
+                    return node.expr.name in NUMPY_COMMON_LIB_NAMES
```
DanielNoord, line 170: Let's revert this, it doesn't add much.

pylint/extensions/magic_value.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class MagicValueChecker(BaseChecker):
+    """Checks for constants in comparisons."""
+
+    name = "magic-value"
+    msgs = {
+        "R2004": (
+            "Consider using a named constant or an enum instead of '%s'.",
+            "magic-value-compare",
```
DanielNoord, line 28: I think Pierre already suggested something here, but should we consider `magic-value-comparison`? It sounds a bit nicer imo.

pylint/extensions/magic_value.py
```diff
@@ -0,0 +1,89 @@
...
+"""Checks for magic values instead of literals."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from astroid import nodes
+
+from pylint.checkers import BaseChecker, utils
+from pylint.checkers.base.comparison_checker import ComparisonChecker
```
DanielNoord, line 14: Can't these be utils?

DanielNoord (changes requested):
I did not have time for a full review (please request one after you have gone through these comments).
I'd prefer if we could decouple this a little more from the current code, but I like that we got it working and appreciate that you have already put considerable time into this PR.

pylint/checkers/base/comparison_checker.py
orSolocate: I switched to:
`node.value in SINGLETON_VALUES`

pylint/extensions/magic_value.py
DanielNoord: Why are we importing a checker here? Can't the method we need from the checker be an utility function in `util.py`? Creating such coupling between checkers is generally something we avoid.

orSolocate: I understand now you are right :)

orSolocate:
@DanielNoord  I managed to clear the pipeline now after applying the fixes :)
Thanks for the review man!

Pierre-Sassoulas (approved):
👍

DanielNoord (approved):

cdce8p:
Just a note: I think it's the right call that this is an extension. Although I haven't tested it, I believe hard-coded constants are easier to optimize for and thus might be slightly faster. The check won't be able to differentiate between cases where it definitely makes sense and just nice to haves.

For pylint / astroid itself: I don't think we should use it.

orSolocate:
> Just a note: I think it's the right call that this is an extension. Although I haven't tested it, I believe hard-coded constants are easier to optimize for and thus might be slightly faster. The check won't be able to differentiate between cases where it definitely makes sense and just nice to haves.
> 
> For pylint / astroid itself: I don't think we should use it.

@cdce8p  Thanks for this note, it makes me think more about the purpose of this checker. I would say that using variables instead of magic values "makes sense" in cases where it avoids code duplication or improves readability. If adding a variable solves neither duplication nor readability than it shouldn't be used to keep up performance. 

On a different note, you could argue that the location of constants (in or outside of functions) for **immutable** types matter more than if it is a literal or not. See this [interesting answer in SO](https://stackoverflow.com/questions/46940622/python-compiler-and-constants-defined-inside-a-function
)

I agree with the use case, I would not use it for Pylint / astroid too. Might be useful for certain projects though, where they consider this issue as an important readability/quality matter.

cdce8p:
> On a different note, you could argue that the location of constants (in or outside of functions) for **immutable** types matter more than if it is a literal or not. See this [interesting answer in SO](https://stackoverflow.com/questions/46940622/python-compiler-and-constants-defined-inside-a-function)

I agree, but that's usually the case already. Most developers do that by intuition. Just an example from pylint itself
https://github.com/PyCQA/pylint/blob/dfd0e5aefe2e90b64099b97fe8dc78f813d5b149/pylint/checkers/design_analysis.py#L91-L101

So I don't know if a checker would be all that useful for it. Maybe an optional extension would be ok though.
In the end, I feel like this is something devs should decide on their own. It's really difficult for static analyzers to guess the intend behind a specific piece of code and we could and up breaking something with the suggestion.

PR #7535, opened 2022-09-27
exclude testutils from coverage

clavedeluna:
Maybe this is a silly or unnecessary change, but I noticed that coveralls included the `pylint/testutils` dir in its count for coverage. I"m not sure why this directory is within the `pylint` dir and not within the `test` dir, which already has a `testutils` dir, but I figured we could at least remove `testutils` from being counted in coverage so the coverage % is more accurate. 

changes
```
17109 of 17950 relevant lines covered (95.31%)
```

to 
```
16124 of 16859 relevant lines covered (95.64%)

```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,7 @@ include =
     pylint/*
 omit =
     */test/*
+    */testutils/*
 # TODO: 3.0: Remove these after these files have been removed
     pylint/config/configuration_mixin.py
     pylint/config/option.py


clavedeluna:
What's the procedure for adding a towncrier frag if this isn't tied to an issue? Create an issue? Add random number like `123.internal`?

Pierre-Sassoulas:
``pylint.testutil`` is (right now) part of pylint, I'm thinking of moving it in its own project but we need a new org and a deprecation period so lib developper can land on their feet. The test in tests/testutils test that that utility for tests work as expected because they're sometime quite complicated.

Pierre-Sassoulas:
> What's the procedure for adding a towncrier frag if this isn't tied to an issue? Create an issue? Add random number like 123.internal?

You can create an issue for it, or reference the pull request itself once you opened it :)

clavedeluna:
alright thanks for the clarification on this!

PR #7537, opened 2022-09-27
Fix bug in detecting ``unused-variable`` when iterating on variable.

### LINKED ISSUE #3044: Problem with nonlocal written only falsely reported as unused.
### Steps to reproduce

The above code in a Nuitka test gives a warning:

```python
def simpleFunction4():
    a = 1

    def nonlocal_writer():
        nonlocal a

        for a in range(10): 
            pass

    nonlocal_writer()

    assert a == 9, a
```

### Current behavior

I need to put # false alarm, pylint: disable=unused-variable in the loop.

### Expected behavior

Since the nonlocal value is being read, it shouldn't be considered unused.

### pylint --version output

$ /c/Python37/python -m pylint --version
__main__.py 2.3.1
astroid 2.2.5
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)]

clavedeluna:
Closes #3044

## Type of Changes

## Description

`unused-variable` should not be reported in the case when a previously defined variable is now being used to iterate with.

Closes #3044


Pierre-Sassoulas (approved):
Very elegant fix, it looks almost too simple. 😄 Let's wait for a review by @jacobtylerwalls before merging.

jacobtylerwalls (approved):
Thanks, that looks like the right fix. I wonder if we could eventually remove the `isinstance` entirely, but that would require more tests (in future). 👍🏻

Issue #7529, opened 2022-09-26
fales positive: unnecessary-dunder-call (__aiter__, __anext__)


belm0:
### Bug description

As of recent pylint versions, I'm seeing the errors below, even though it's suggesting the wrong builtin (`iter()` rather than `aiter()`), and even though I'm on Python 3.8 (which doesn't have `aiter()` and `anext()` builtins).

### Command used

```shell
pylint
```

### Pylint output

```shell
C2801: Unnecessarily calls dunder method __aiter__. Use iter built-in function. (unnecessary-dunder-call)
C2801: Unnecessarily calls dunder method __anext__. Use next built-in function. (unnecessary-dunder-call)
```

### Expected behavior

no errors

### Pylint version

```shell
pylint 2.15.3
astroid 2.12.10
Python 3.8.12 (default, Sep 23 2021, 08:42:37)
[Clang 12.0.0 (clang-1200.0.32.29)]
```

DanielNoord:
@jpy-git Do you happen to know anything about this? Or would you want to take a look at this?

belm0:
1. fix entries in `DUNDER_METHODS`.  "Use {iter,next}" --> "Use {aiter,anext}"

https://github.com/PyCQA/pylint/blob/ec85e2988679a28dbe11f46d60056b4dd569c44b/pylint/checkers/dunder_methods.py#L108-L109

2. this checker needs to have awareness of the Python version.  The brute force for this case would be to declare the set of builtins requiring >= 3.10, and add it to the checker condition.

```python
NEEDS_PYTHON_3_10 = {'__aiter__', '__anext__'}
```

DanielNoord:
> 1. fix entries in `DUNDER_METHODS`.  "Use {iter,next}" --> "Use {aiter,anext}"
> 
> https://github.com/PyCQA/pylint/blob/ec85e2988679a28dbe11f46d60056b4dd569c44b/pylint/checkers/dunder_methods.py#L108-L109
> 

This is an easy fix. Would you want to open a PR for this? You've done all the work already!

> 2. this checker needs to have awareness of the Python version.  The brute force for this case would be to declare the set of builtins requiring >= 3.10, and add it to the checker condition.
> 
> ```python
> NEEDS_PYTHON_3_10 = {'__aiter__', '__anext__'}
> ```
> 
> (If there is no time to do this for the next release, at least disable these entries in `DUNDER_METHODS`.)

[...]

PR #7549, opened 2022-10-01
Check `` py-version`` for async ``unnecessary-dunder-call``

DanielNoord:

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/7529.


pylint/checkers/dunder_methods.py
```diff
@@ -105,8 +105,6 @@
     "__floor__": "Use math.floor function",
     "__ceil__": "Use math.ceil function",
     "__enter__": "Invoke context manager directly",
```
Pierre-Sassoulas, line 107: I think we should handle this generically by changing this datastructure (DUNDER_METHODS) to include the min version of the dunder method. Maybe:

```python
DUNDER_METHODS = {
    "3.10": {
       "__aiter__": "Use iter built-in function",
       "__anext__": "Use next built-in function",
    }
    "default" : {
        "__index__": "Use index method",
        "__round__": "Use round built-in function",
        ...
    }
}

```

pylint/checkers/dunder_methods.py
```diff
@@ -16,99 +16,105 @@
...
-    "__trunc__": "Use math.trunc function",
-    "__floor__": "Use math.floor function",
-    "__ceil__": "Use math.ceil function",
-    "__enter__": "Invoke context manager directly",
-    "__aenter__": "Invoke context manager directly",
-    "__copy__": "Use copy.copy function",
-    "__deepcopy__": "Use copy.deepcopy function",
-    "__fspath__": "Use os.fspath function instead",
+DUNDER_METHODS: dict[tuple[int, int], dict[str, str]] = {
+    (0, 0): {
```
Pierre-Sassoulas, line 20: Clever 👍

pylint/checkers/dunder_methods.py
```diff
@@ -143,6 +147,12 @@ class DunderCallChecker(BaseChecker):
     }
     options = ()
 
+    def open(self) -> None:
+        self._dunder_methods: dict[str, str] = {}
+        for since_vers, dunder_methods in DUNDER_METHODS.items():
+            if since_vers <= self.linter.config.py_version:
+                self._dunder_methods.update(dunder_methods)
```
Pierre-Sassoulas, line 154:
```suggestion
        self._dunder_methods: dict[str, str] = {
            dunder_method: text
            for since_vers, dunder_methods in DUNDER_METHODS.items():
            for dunder_method, text in dunder_methods.items()
            if since_vers <= self.linter.config.py_version:
        }
```
I think this is better performance-wise ?

Pierre-Sassoulas (commented):
Thank you for the fix !

pylint/checkers/dunder_methods.py
DanielNoord: Are you sure? This seems a bit like pre-mature optimisation with much less readable code. I'd prefer the current code.

Pierre-Sassoulas: Well, actual benchmark with the current list show that the for loop is faster by an order of magnitude.
Code:
```python
tick = 10000
start = time.time()
for i in range(tick):
    dunder_methods: dict[str, str] = {}
    for since_vers, dunder_methods in DUNDER_METHODS.items():
        if since_vers <= (3, 10):
            dunder_methods.update(dunder_methods)
end = time.time()
print(end - start)
start = time.time()
for i in range(tick):
    dunder_methods: dict[str, str] = {
        dunder_method: text
        for since_vers, dunder_methods in DUNDER_METHODS.items()
        for dunder_method, text in dunder_methods.items()
        if since_vers <= (3, 10)
    }
end = time.time()
print(end - start)
```

Result:
```
0.014837980270385742
0.13480186462402344
```
dict comprehension would be faster but for 1000+ elements which is not going to happen.

Pierre-Sassoulas (approved):

PR #7550, opened 2022-10-01
Adding optional checker for nested min/max

### LINKED ISSUE #7546: Suggest >2 arguments for min/max functions
### Current problem

Some Python users may be unaware that the `min` and `max` functions allow for more than two arguments.

### Desired solution

Expressions of the form: `min(a, min(b, min(c, d))` can be replaced with the simpler `x = min(a, b, c, d)`.

### Additional context

_No response_

osherdp:

## Type of Changes

## Description

This adds a new checker (not active by default) which identifies usages similar to ``min(<arg1>, min(<arg2>, <arg3>))`` and suggests using a simplified form of ``min(<arg1>, <arg2>, <arg3>)``. Same goes for ``max`` usage.

The logic is as follows: it detects calls to either ``min`` or ``max`` functions, and whenever one of their arguments is that same function, it emits the message.

Closes #7546


pylint/extensions/nested_min_max.py
```diff
@@ -0,0 +1,54 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
DanielNoord, line 1: I think this can become part of `stdlib.py`. I don't think it needs to be an extension.

pylint/extensions/nested_min_max.py
```diff
@@ -0,0 +1,54 @@
...
+    from pylint.lint import PyLinter
+
+
+class NestedMinMaxChecker(BaseChecker):
+    FUNC_NAMES = ("min", "max")
+
+    name = "nested_min_max"
+    msgs = {
+        "W3201": (
+            "Using redundant nested calls for '%s'. This function may accept more than 2 arguments",
```
DanielNoord, line 27:
```suggestion
            "Using redundant nested calls for '%s'. This function accepts more than 2 arguments",
```

pylint/extensions/nested_min_max.py
```diff
@@ -0,0 +1,54 @@
...
+            "nested-min-max",
+            "Nested calls ``min(1, min(2, 3))`` can be rewritten as ``min(1, 2, 3)``.",
+        )
+    }
+
+    @only_required_for_messages("nested-min-max")
+    def visit_call(self, node: nodes.Call) -> None:
+        if (
+            isinstance(node.func, nodes.Name)
+            and node.func.name in self.FUNC_NAMES
```
DanielNoord, line 37: Should we use inference here? What if somebody overwrites `min` or `max`?

tests/extensions/test_nested_min_max.py
```diff
@@ -0,0 +1,93 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
```
DanielNoord, line 1: These tests are unnecessary if we use the functional test framework. Let's remove this file.

tests/functional/ext/nested_min_max/nested_min_max.py
```diff
@@ -0,0 +1,11 @@
+"""test detection of redundant nested calls to min/max functions"""
```
DanielNoord, line 1:
```suggestion
"""Test detection of redundant nested calls to min/max functions."""
```

DanielNoord (changes requested):

pylint/extensions/nested_min_max.py
Pierre-Sassoulas:
```suggestion
            "Do not use nested call of '%s'  it's possible to do '%s' instead",
```

tests/functional/ext/nested_min_max/nested_min_max.txt
```diff
@@ -0,0 +1,5 @@
+nested-min-max:4:0:4:17::Using redundant nested calls for 'min'. This function may accept more than 2 arguments:HIGH
+nested-min-max:5:0:5:17::Using redundant nested calls for 'max'. This function may accept more than 2 arguments:HIGH
+nested-min-max:6:0:6:17::Using redundant nested calls for 'min'. This function may accept more than 2 arguments:HIGH
+nested-min-max:9:0:9:25::Using redundant nested calls for 'min'. This function may accept more than 2 arguments:HIGH
+nested-min-max:10:0:10:40::Using redundant nested calls for 'min'. This function may accept more than 2 arguments:HIGH
```
Pierre-Sassoulas, line 5:
```suggestion
nested-min-max:10:0:10:40::Do not use nested call of 'min'  it's possible to do 'min(1, 2, 3)' instead",:HIGH
```

Pierre-Sassoulas (commented):
I would make the message clearer by stating the actual replacing code. (This could be useful if we want to automatically fix in the future too).

pylint/extensions/nested_min_max.py
osherdp: Does it mean it will be activated by default?

DanielNoord: Yes!

DanielNoord:
@osherdp Please don't force-push your branch. It makes reviewing more difficult. We will squash merge your changes into the `main` branch anyway so the commit history in this branch doesn't need to be perfect 😄

pylint/extensions/nested_min_max.py
Pierre-Sassoulas: I think it would be better if we keep the file as it is before adding to stdlib (I think the base checker is an example of a checker dispathed across multiple files, the checker name is the same in all files).

DanielNoord: Doesn't the splitting across multiple files make `pylint` slower? There are more functional calls, class initialisations etc. If the file isn't too large we can combine them right?

Pierre-Sassoulas: It make everything harder to maintain though. Clearly differentiating checks help tremendously when keeping clean code because it's possible to understand everything about a 50 lines checker, but it's clearly impossible to understand and thus cleanup a 2000 lines stdlib checker behemoth if a function, constant (or more interestingly for performances: condition or calculation) become useless.

DanielNoord: Hmm, okay. So how would you want to add this then?

Pierre-Sassoulas: If we want to make this a default check we can add a new checker (just move the file in pylint/checker) and make it a default checker (register it) with a new name or with an existing checker name.

pylint/extensions/nested_min_max.py
```diff
@@ -0,0 +1,79 @@
...
+                    node.args = node.args[:i] + arg.args + node.args[i + 1 :]
+                    break
+
+            redundant_calls = self.get_redundant_calls(node)
+
+        self.add_message(
+            "nested-min-max",
+            node=node,
+            args=(node.func.name, node.as_string()),
+            confidence=HIGH,
```
jacobtylerwalls, line 74:
```suggestion
            confidence=INFERENCE,
```

pylint/extensions/nested_min_max.py
osherdp: Changed it into a new checker!
Let me know which existing checker name I should use if you have any preference

pylint/extensions/nested_min_max.py
osherdp: fixed!

mbyrnepr2:
I think the doc CI is failing because of missing good/bad examples for this new checker. Existing docs live here: https://github.com/PyCQA/pylint/tree/main/doc/data/messages

pylint/checkers/nested_min_max.py
```diff
@@ -0,0 +1,79 @@
...
+    from pylint.lint import PyLinter
+
+
+class NestedMinMaxChecker(BaseChecker):
+    FUNC_NAMES = ("builtins.min", "builtins.max")
+
+    name = "nested_min_max"
+    msgs = {
+        "W3201": (
+            "Do not use nested call of '%s'  it's possible to do '%s' instead",
```
DanielNoord, line 27:
```suggestion
            "Do not use nested call of '%s' it's possible to do '%s' instead",
```

pylint/checkers/nested_min_max.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @classmethod
+    def is_min_max_call(cls, node: nodes.NodeNG) -> bool:
+        if not isinstance(node, nodes.Call):
+            return False
+
+        inferred = safe_infer(node.func)
+        return (
+            inferred is not None
+            and inferred.is_function
```
DanielNoord, line 41:
```suggestion
            and isinstance(inferred, nodes.FunctionDef)
```

pylint/checkers/nested_min_max.py
```diff
@@ -0,0 +1,79 @@
...
+            and inferred.qname() in cls.FUNC_NAMES
+        )
+
+    @classmethod
+    def get_redundant_calls(cls, node: nodes.Call) -> nodes.NodeNG:
```
DanielNoord, line 46:
```suggestion
    def get_redundant_calls(cls, node: nodes.Call) -> list[nodes.Call]:
```

pylint/checkers/nested_min_max.py
```diff
@@ -0,0 +1,79 @@
...
+            if cls.is_min_max_call(arg) and arg.func.name == node.func.name
+        ]
+
+    @only_required_for_messages("nested-min-max")
+    def visit_call(self, node: nodes.Call) -> None:
+        if not self.is_min_max_call(node):
+            return
+
+        redundant_calls = self.get_redundant_calls(node)
+        if len(redundant_calls) == 0:
```
DanielNoord, line 59:
```suggestion
        if not redundant_calls:
```

pylint/checkers/nested_min_max.py
```diff
@@ -0,0 +1,79 @@
...
+            return
+
+        while len(redundant_calls) > 0:
+            for i, arg in enumerate(node.args):
+                if arg in redundant_calls:
+                    node.args = node.args[:i] + arg.args + node.args[i + 1 :]
```
DanielNoord, line 65: Let's not reassign anything on `node` itself. That changes the `ast` representation and is a side effect that might not work well with future checkers.

tests/functional/n/nested_min_max.py
```diff
@@ -0,0 +1,17 @@
...
+
+min(1, min(2, 3))  # [nested-min-max]
+max(1, max(2, 3))  # [nested-min-max]
+min(min(1, 2), 3)  # [nested-min-max]
+min(min(min(1, 2), 3), 4)  # [nested-min-max]
+min(1, max(2, 3))
+min(1, 2, 3)
+min(min(1, 2), min(3, 4))  # [nested-min-max]
+min(len([]), min(len([1]), len([1, 2])))  # [nested-min-max]
+min(1, min(2, 3))  # pylint: disable=nested-min-max
```
DanielNoord, line 12:
```suggestion
```

DanielNoord (changes requested):
Thanks for moving the checker. Could you also add a `good.py` and `bad.py` example?

pylint/checkers/nested_min_max.py
mbyrnepr2: > 

```suggestion
            "Do not use nested call of '%s';  it's possible to do '%s' instead",
```

DanielNoord: In Dutch `it's possible` should be capitalised as the `;` is considered a period. I'm not sure about English in this case 😅

mbyrnepr2: Oh 😅. I'm not living in NL long enough to know that but I think it would be lowercase in English. Its OK with me either way!

DanielNoord:
Hi @osherdp are you still interested in working on this PR?

osherdp:
> Hi @osherdp are you still interested in working on this PR?

Unfortunately I cannot find time to invest in it :disappointed: 
If anyone like to take over this work I wouldn't mind

Pierre-Sassoulas:
Merged upstream/main to get the primer's result.

DanielNoord:
I think tests should now pass.

```python
min(min(min(1, 2), 3), 4) # [nested-min-max, nested-min-max]
```

Raises twice which could be avoided by doing some name and inference look up on the parent of a `nodes.Call`. However, that would slow down the checker and the message isn't incorrect: there are two simplify-able messages.

pylint/checkers/nested_min_max.py
```diff
@@ -0,0 +1,82 @@
...
+
+from pylint.checkers import BaseChecker
+from pylint.checkers.utils import only_required_for_messages, safe_infer
+from pylint.interfaces import INFERENCE
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class NestedMinMaxChecker(BaseChecker):
```
Pierre-Sassoulas, line 22:
```suggestion
class NestedMinMaxChecker(BaseChecker):
    """Multiple nested min on the same line will raise multiple message by design. The check is faster this way and the multiple messages aren't wrong."''
```

Pierre-Sassoulas (approved):
LGTM ! I'd suggest doing some documentation about the design choices. But feel free to disregard.

pylint/checkers/nested_min_max.py
DanielNoord:
```suggestion
class NestedMinMaxChecker(BaseChecker):
    """Multiple nested min/max calls on the same line will raise multiple messages.
    
    This behaviour is intended as it would slow down the checker to check
    for nested call with minimal benefits.
    """
```

Pierre-Sassoulas:
I forgot to wait for the primer's results.

Pierre-Sassoulas:
I'm wondering how to make ``max_len = max(max(self.adj.len(x) for x in fmt_values), header_colwidth)`` better ? ``max_len = max(self.adj.len(x) for x in fmt_values) + [header_colwidth])`` ? Is there another way ? I feel like it might be a false positive ?

DanielNoord:
So, that would necessitate stopping the warning when there is a `Comprehension` inside the `max` call right?

Pierre-Sassoulas:
Maybe, but I'm not even sure it's a false positive tbh.

DanielNoord:
I think the only code that works is:
`max_len = max([self.adj.len(x) for x in fmt_values] + [header_colwidth]) `

That doesn't seem very nice to read, but that might just be personal preference.

Pierre-Sassoulas:
I have the intuition that this is better performance-wise as there is only one max call. If that's the case it would not be a false positive. But the solution is not super obvious I had to think about for longer than what pylint user are ready to invest imo.

DanielNoord:
So, fix? Or keep it as a feature? 😄

Pierre-Sassoulas:
I was counting on you @jacobtylerwalls @mbyrnepr2  or @osherdp to take the decision for me 😄

Let's vote !

This is a false positive, we need to fix: : 👀 
Let's keep it as a feature: 🚀

nickdrozd:
The proposed fix uses one less `max` call, but adds a list allocation. That could be slower, especially in an inner loop.

mbyrnepr2:
I voted in favour of fixing the false positive; the broken code (it is broken, right?) suggested by the pandas primer result. So perhaps we are not on the same page. I don't necessarily think suggesting the list approach is the solution to go for. Perhaps we need to suppress instead when there is a comprehension like Daniël mentioned.

Pierre-Sassoulas:
@nickdrozd your comment seems to indicate that you should have voted for considering it a false positive instead ?

Pierre-Sassoulas:
The council has spoken unanimousely : it's a false positive.

DanielNoord:
For now, I'm going to change the PR to not warn on the pattern anymore. The more intricate suggestion would require more time than I or the original author seems to have for this PR.

I just want to get this in a minimal viable state and merge as I still think it would be useful in its minimal state!

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Thank you for the initial implementation @osherdp, and congratulation on becoming a pylint contributor. Thank you for taking over @DanielNoord !

cdce8p:
@Pierre-Sassoulas Please check the commit message before merging. I was just looking through the git log and saw that a new extension was added when in fact it's activated by default. I know these things are sometimes easy to miss 😅 https://github.com/PyCQA/pylint/commit/bb5781cd94eace021edf82954d151810770d241b

Pierre-Sassoulas:
Good catch, that's my bad. I was thinking that we always add new things as extension now. I don't think it's fixable at this point :(

PR #7552, opened 2022-10-01
Remove xfail from passing tests

DanielNoord:

## Type of Changes

## Description

@DudeNr33 Are these test passing correctly? Or should they be made to actually fail again?

### DIFF
--- a/tests/pyreverse/test_diadefs.py
+++ b/tests/pyreverse/test_diadefs.py
@@ -106,7 +106,6 @@ class TestDefaultDiadefGenerator:
         ("specialization", "Specialization", "Ancestor"),
     ]
 
-    @pytest.mark.xfail
     def test_extract_relations(self, HANDLER: DiadefsHandler, PROJECT: Project) -> None:
         """Test extract_relations between classes."""
         with pytest.warns(DeprecationWarning):
@@ -115,7 +114,6 @@ class TestDefaultDiadefGenerator:
         relations = _process_relations(cd.relationships)
         assert relations == self._should_rels
 
-    @pytest.mark.xfail
     def test_functional_relation_extraction(
         self, default_config: PyreverseConfig, get_project: GetProjectCallable
     ) -> None:


tests/pyreverse/test_diadefs.py
```diff
@@ -115,7 +114,6 @@ def test_extract_relations(self, HANDLER: DiadefsHandler, PROJECT: Project) -> N
         relations = _process_relations(cd.relationships)
         assert relations == self._should_rels
 
-    @pytest.mark.xfail
     def test_functional_relation_extraction(
         self, default_config: PyreverseConfig, get_project: GetProjectCallable
     ) -> None:
```
Pierre-Sassoulas, line 119: Line 123/124 would also needs to be updated here.

Pierre-Sassoulas (changes requested):

DudeNr33 (approved):
The `xfail` was added after the update to `astroid` 2.12, and AFAIR we suspected the handling of implicit namespace packages to be the reason the tests were failing.
As all other tests were still passing we went with just marking these two as `xfail`.

If they are now passing again, all the better!

Pierre already mentioned the comment that should be removed again, besides that I have nothing more to add. 👍

tests/pyreverse/test_diadefs.py
DanielNoord: @DudeNr33 Are we sure about that?
The comment was added 14 years ago in this commit https://github.com/PyCQA/pylint/commit/7af79e602152aa122328063d5d7d5a16c65e594a

I don't think it is related to the `xfail`?

DudeNr33: Oh, you are right. Sorry. To be honest, if the comment is this old I have no idea what "environment problem" is really meant here.

Pierre-Sassoulas: Yeah I also did not understand what the comment meant. Should we remove anyway ?

DanielNoord: I don't think so; https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence

Pierre-Sassoulas: This Chesterton guy must have had a lot of documentation stored in his attic 😄

Pierre-Sassoulas (approved):

jacobtylerwalls:
I agree with removing the xfails, as I remember adding them in the astroid 2.12 upgrade.

I agree with @DanielNoord and @Chesterton about leaving the comment as is. It's too bad it's vague, but it signals a lack of confidence about something in the test result even if the assertions are okay. It might be more evident when looking at the output visually.

jacobtylerwalls (approved):

PR #7573, opened 2022-10-04
Add docs for consider-using-assignment-expr and confusing-with-statement

clavedeluna:
## Type of Changes

## Description

Add docs for consider-using-assignment-expr and confusing-with-statement

Refs #5953


doc/data/messages/c/consider-using-assignment-expr/pylintrc
```diff
@@ -0,0 +1,2 @@
+[MAIN]
+py-version=3.8
```
clavedeluna, line 2: @Pierre-Sassoulas would this not have been sufficient to make the CI check `Checks / documentation` pass?

Pierre-Sassoulas: Does ``pylint /home/runner/work/pylint/pylint/doc/data/messages/c/consider-using-assignment-expr/bad.py --disable=all --enable=F,consider-using-assignment-expr,astroid-error,syntax-error`` work locally ?

Pierre-Sassoulas: There might be an issue with the decorator "only used for" on the check it happened often.

clavedeluna: It indeed does not work locally, No `consider-using-assignment-expr` is in the output. Is there something you suggest to do in this case to make this work?

Pierre-Sassoulas: I would start by checking the ``util.only_required_for_messages`` calls when ``consider-using-assignment-expr`` is involved in the decorated function. Oftentimes this is the reason a check do not work if other checks are disabled.

Pierre-Sassoulas: This is a function used to optimize execution time to not run code when not needed.

clavedeluna: I think the issue is in using this pylintrc file  inside the `.../c/consider-using-assignment-expr` dir. Take a look:

```
 pylint doc/data/messages/c/consider-using-assignment-expr/bad.py --disable=all --enable=F,consider-using-assignment-expr,astroid-error,syntax-error --py-version 3.8
************* Module bad
doc/data/messages/c/consider-using-assignment-expr/bad.py:3:3: R6103: Use 'if (apples := 2):' instead (consider-using-assignment-expr)
```

passing py-version as an arg works as expected! I assumed since other message dirs have a pylintrc file it should work?

Pierre-Sassoulas: It's possible that py-version is not handled in the functional test configuration. py-version was introduced after it and functional test helpers are not super well tested.

clavedeluna: I assumed loading pylintrc configs wasn't the responsibility of the functional tests in test_message_documentation.py, but instead is a core pylint responsibility. Am I mistaken?

clavedeluna: Interesting, unexpected results:

```
pylint doc/data/messages/c/consider-using-assignment-expr/bad.py --disable=all --enable=F,consider-using-assignment-expr,astroid-error,syntax-error --py-version 3.8
************* Module bad
doc/data/messages/c/consider-using-assignment-expr/bad.py:3:3: R6103: Use 'if (apples := 2):' instead (consider-using-assignment-expr)
```

with pylintrc contents as in this PR
```
[MAIN]
py-version=3.8
```

we get diff output
```
pylint doc/data/messages/c/consider-using-assignment-expr/bad.py --disable=all --enable=F,consider-using-assignment-expr,astroid-error,syntax-error --rcfile=doc/data/messages/c/consider-using-assignment-expr/pylintrc

```

This seems fishy to me...

Pierre-Sassoulas: Yes indeed, if ``py-version=3.8`` then the 3.8 checks should be activated not the other way around.

clavedeluna: It's worth noting `consider-using-assignment-exp` is not within the `extensions` dir... but I see it mentioned in extensions.rst 🤷‍♀️

doc/data/messages/c/confusing-with-statement/good.py
```diff
@@ -1 +1,5 @@
-# This is a placeholder for correct code for this message.
+from contextlib import ExitStack
+
+with ExitStack() as stack:
+    fh1 = stack.enter_context(open('file1.txt', 'w'))
+    fh2 = stack.enter_context(open('file2.txt', 'w'))
```
Pierre-Sassoulas, line 5: How about this:
```suggestion
with open('file.txt', 'w') as fh1:
    with open('file.txt', 'w') as fh2:
        pass
```
Using ExitStack is probably worse for performance because it has multiple unavoidable call, it add an import and deviate from the standard way of doing thing (you have to search for what ``ExitStack`` and ``enter_context`` do)

clavedeluna: 😆 
```
doc/data/messages/c/confusing-with-statement/good.py:1:5: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
doc/data/messages/c/confusing-with-statement/good.py:2:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
```

but I'll pick another example

Pierre-Sassoulas: Shouldn't this message be disabled ? We can specify the encoding too.

Pierre-Sassoulas (approved):

Issue #6917, opened 2022-06-11
Check For @singledispatch Decorator in classes without `@staticmethod`


cdreimer-thewriter:
### Current problem

The representative code implements single dispatching with the ``@singledispatch`` and ``@staticmethod`` decorators in a class.

```python
Class Board():

   @singledispatch
   @staticmethod
   def convert_position(position):
        [...]
   
    @convert_position.register(str)
    @staticmethod
    def _(position: str) -> tuple:
        [...]

    @convert_position.register(tuple)
    @staticmethod
    def _(position: tuple) -> str:
        [...]
```

Pylint reports nothing unusual about using ``@singledispatch`` inside a class.

### Desired solution

Pylint should check for the ``@singledispatch`` decorator inside a class. According to the [documentation](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod), the ``@singledispatchmethod`` decorator is for class methods.

### Additional context

The code works with the ``@singledispatch`` decorator as long as the single dispatch methods have the ``@staticmethod`` decorator (i.e., stateless methods). If the ``@staticmethod`` decorators were removed or replaced with the ``@classmethod`` decorator, the code stops working and the resulting error messages will indirectly point out the wrong decorator. A pylint check would prevent that from happening.

jacobtylerwalls:
Thanks for the clear report. This could be added to the `StdlibChecker`. No obligation, but if you have (or anyone reading has) an interest in preparing a patch yourself I'd be glad to answer questions as you go.

ramonsaraiva:
Hey @jacobtylerwalls - landing here from the `hacktoberfest` label and not familiar with pylint's codebase. Is this a similar implementation as `_check_lru_cache_decorators` in `StdlibChecker`?

DanielNoord:
@ramonsaraiva That should indeed be relatively similar. In fact, I think we can build upon some of the code that is already there. You can probably also look at https://github.com/PyCQA/pylint/blob/main/tests/functional/m/method_cache_max_size_none.py for some inspiration for functional tests.

ramonsaraiva:
Awesome, I'll be reading more about how everything works and attempt to come up with a draft patch for this. A few review interactions will most likely be needed. Thanks!

ramonsaraiva:
Created that draft PR and added in its description a bit of my thought process and things that I believe can be improved and simplified. I just want to have a better sense if I'm going in the right direction.

[...]

PR #7575, opened 2022-10-05
@singledispatch and @singledispatchmethod checks for methods and functions

ramonsaraiva:
## Type of Changes

## Description

Draft implementation for `singledispatch` and `singledispatchmethod` checks. ~~I just added one sample functional test file, still need to write the tests for `singledispatchmethod` decorators.~~

~~I don't love the way all those if statements are written~~, but I just want to have a sense if I'm going in the right direction so I can think about cleaning those up. This one function is also checking multiple states:~~
* `singledispatch` with `classmethod`
* `singledispatch` register call with `classmethod`
* `singledispatchmethod` with `staticmethod`
* `singledispatchmethod` register call with `staticmethod`
* `singledispatch` without any other decorator, but still in a class scope

~~I'm pretty sure there's a lot that can be simplified, like moving the lines where I check if the `register` call/attribute actually refers to a `singledispatch` or `singledispatchmethod` decorator to a utility function similar to `is_registered_in_singledispatch_function`. I also realized there is this `decorated_with` helper after I implemented the `decorators_names_set` and `decorators_name_node_map` approach for verifying the cases without `register` calls.~~

EDIT: I think https://github.com/PyCQA/pylint/pull/7575/commits/4d95d21ad41e1a219cc5917181b102e5de5832bd and https://github.com/PyCQA/pylint/pull/7575/commits/cc64ded7c6c5e5fd053e5a58fbbeebce0cff93c8 bring a much cleaner version that https://github.com/PyCQA/pylint/pull/7575/commits/3dc361624cd805e32972372257e3ee48ef1ac84a had - still open to suggestions.

I added a new util function for `singledispatchmethod` (similar to the `singledispatch` one) and moved some reusable logic to a separate one. There was a minor modification to the way we check for `register` calls, cause now with annotated types `register` can also be an attribute and not strictly a call. 

Refs #6917


pylint/checkers/stdlib.py
```diff
@@ -389,6 +389,22 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             "Env manipulation functions support only string type arguments. "
             "See https://docs.python.org/3/library/os.html#os.getenv.",
         ),
+        "E1510": (
```
DanielNoord, line 392: I'd prefer to use a number that is not being used yet. We have `W1510` which at some point we might want to escalate to `E1510`. I believe 1519 and 1520 are free.

pylint/checkers/stdlib.py
```diff
@@ -599,6 +620,35 @@ def _check_lru_cache_decorators(self, decorators: nodes.Decorators) -> None:
                 confidence=interfaces.INFERENCE,
             )
 
+    def _check_dispatch_decorators(self, node: nodes.FunctionDef) -> None:
+        decorators_names_set: set[str] = set()
+        decorators_name_node_map: dict[str, nodes.NodeNG] = {}
+
+        decorators = node.decorators.nodes if node.decorators else []
+        for decorator in decorators:
```
DanielNoord, line 628:
```suggestion
        for decorator in node.decorators:
```

We check for this in the calling code.

DanielNoord, line 628: We could use `inference` here as users might overwrite `classmethod`. But I think in this case we probably shouldn't. Overwriting `exit` or `print` I can sort of understand but if you're overwriting `classmethod` you probably also don't use `pylint` to make sure you have sane code :sweat_smile:

pylint/checkers/stdlib.py
```diff
@@ -599,6 +620,35 @@ def _check_lru_cache_decorators(self, decorators: nodes.Decorators) -> None:
...
+            elif utils.is_registered_in_singledispatch_function(node):
+                decorators_names_set.add("singledispatch")
+                decorators_name_node_map["singledispatch"] = decorator
+            elif utils.is_registered_in_singledispatchmethod_function(node):
+                decorators_names_set.add("singledispatchmethod")
+                decorators_name_node_map["singledispatchmethod"] = decorator
+
+        if {"singledispatch", "classmethod"}.issubset(decorators_names_set):
+            self.add_message(
+                "singledispatch-method", node=decorators_name_node_map["singledispatch"]
```
DanielNoord, line 641: Could you add `confidence` as well?

pylint/checkers/utils.py
```diff
@@ -1444,6 +1444,30 @@ def node_type(node: nodes.NodeNG) -> SuccessfulInferenceResult | None:
...
+    elif isinstance(node, nodes.Attribute):
+        func = node
+    else:
+        return None
+
+    if not isinstance(func, nodes.Attribute) or func.attrname != "register":
+        return None
+
+    try:
+        func_def = next(func.expr.infer())
```
DanielNoord, line 1461: Can we use `safe_inference` here?

pylint/checkers/utils.py
```diff
@@ -1457,21 +1481,27 @@ def is_registered_in_singledispatch_function(node: nodes.FunctionDef) -> bool:
...
+    return False
 
-        try:
-            func_def = next(func.expr.infer())
-        except astroid.InferenceError:
-            continue
 
-        if isinstance(func_def, nodes.FunctionDef):
-            return decorated_with(func_def, singledispatch_qnames)
+def is_registered_in_singledispatchmethod_function(node: nodes.FunctionDef) -> bool:
```
DanielNoord, line 1490: Perhaps put the refactoring in another PR. That would help reviewing this and keep the size of the PR smaller.

tests/functional/s/singledispatch_method.py
```diff
@@ -0,0 +1,41 @@
+"""Tests for singledispatch-method"""
+# pylint: disable=missing-class-docstring, missing-function-docstring,
+# pylint: disable=too-few-public-methods
```
DanielNoord, line 3:
```suggestion

# pylint: disable=missing-class-docstring,missing-function-docstring,too-few-public-methods
```

tests/functional/s/singledispatchmethod_function.py
```diff
@@ -0,0 +1,42 @@
+"""Tests for singledispatchmethod-function"""
+# pylint: disable=missing-class-docstring, missing-function-docstring,
+# pylint: disable=too-few-public-methods
```
DanielNoord, line 3:
```suggestion

# pylint: disable=missing-class-docstring,missing-function-docstring,disable=too-few-public-methods
```

DanielNoord (changes requested):
This already looks very good. Could you also add a `good.py` and `bad.py` example for both messages in the `docs/data` directory?

pylint/checkers/utils.py
ramonsaraiva: Just to make sure I understand, by refactoring you mean the common code i moved to `find_infered_fn_from_register`? or not touching `pylint/checkers/utils.py` at all and add the logic for resolving `singledispatch` and `singledispatchmethod` in the stdlib checker module?

DanielNoord: Yeah, the first thing. Let's focus this PR on everything that is strictly necessary for the new messages. Anything else can be moved to separate PRs as it can be easily approved and merged.

pylint/checkers/stdlib.py
ramonsaraiva: Yeah, this was me misunderstanding that numbers are shared between classifiers (W, E, ..)

pylint/checkers/utils.py
```diff
@@ -1457,11 +1457,15 @@ def is_registered_in_singledispatch_function(node: nodes.FunctionDef) -> bool:
...
     for decorator in decorators:
-        # func.register are function calls
-        if not isinstance(decorator, nodes.Call):
+        # func.register are function calls or register attributes
+        # when the function is annotated with types
+        if isinstance(decorator, nodes.Call):
+            func = decorator.func
+        elif isinstance(decorator, nodes.Attribute):
+            func = decorator
+        else:
```
ramonsaraiva, line 1466: Still needed to adjust this part cause we should consider, for..
```python
@singledispatch
def example(arg):
    pass
```

..that both below are classified as possible registers:
```python
@example.register(str)
def _(arg):
    pass

@example.register
def _(arg: str):
    pass
```

ramonsaraiva:
For some reason `decorated_with` in py37 is not infering decorator nodes properly so `utils.is_registered_in_singledispatch_function` and `utils.is_registered_in_singledispatchmethod_function` are both returning `False`, which makes this test case to fail:
```
E       Expected in testdata:
E         33: singledispatchmethod-function
E         39: singledispatchmethod-function
```
Only happens on py37, any idea?

DanielNoord:
> For some reason `decorated_with` in py37 is not infering decorator nodes properly so `utils.is_registered_in_singledispatch_function` and `utils.is_registered_in_singledispatchmethod_function` are both returning `False`, which makes this test case to fail:
> 
> ```
> E       Expected in testdata:
> E         33: singledispatchmethod-function
> E         39: singledispatchmethod-function
> ```
> 
> Only happens on py37, any idea?

`singledispatchmethod` was only added in `3.8`. I think that could explain it?

ramonsaraiva:
That does indeed explain.. Sorry for that 🤦🏻‍♂️

pylint/checkers/stdlib.py
```diff
@@ -599,6 +620,40 @@ def _check_lru_cache_decorators(self, decorators: nodes.Decorators) -> None:
                 confidence=interfaces.INFERENCE,
             )
 
+    def _check_dispatch_decorators(self, node: nodes.FunctionDef) -> None:
+        decorators_names_set: set[str] = set()
+        decorators_name_node_map: dict[str, nodes.NodeNG] = {}
```
DanielNoord, line 625: Can we making the mapping a tuple so we only have two temp vars?

I was also thinking of using `if "singledispatch" in decorators_map or "classmethod" in decorators` but I'm not sure that's more readable.

tests/functional/s/singledispatch_method.py
```diff
@@ -0,0 +1,41 @@
+"""Tests for singledispatch-method"""
+# pylint: disable=missing-class-docstring, missing-function-docstring,too-few-public-methods
+# pylint: disable=no-name-in-module
```
DanielNoord, line 3: Which module is reporting this?

ramonsaraiva: Oh, just realized this only happens for py37 cause `tests/functional/s/singledispatch_method.py` still imports `singledispatchmethod` for testing the right combination of `singledispatchmethod` and `classmethod`. I'll create a py37 and a py38+ versions of that test so I don't do that for py37.

pylint/checkers/stdlib.py
```diff
@@ -599,6 +620,38 @@ def _check_lru_cache_decorators(self, decorators: nodes.Decorators) -> None:
...
+                    decorator,
+                    interfaces.INFERENCE,
+                )
+
+        if "singledispatch" in decorators_map and "classmethod" in decorators_map:
+            decorator, confidence = decorators_map["singledispatch"]
+            self.add_message(
+                "singledispatch-method",
+                node=decorator,
+                confidence=confidence,
```
DanielNoord, line 642: Perhaps a bit pre-mature but let's skip the temporary assignment here. The code is short and clear enough that we can use `[0]` and `[1]`.

pylint/checkers/stdlib.py
```diff
@@ -599,6 +620,38 @@ def _check_lru_cache_decorators(self, decorators: nodes.Decorators) -> None:
...
+            )
+        elif (
+            "singledispatchmethod" in decorators_map
+            and "staticmethod" in decorators_map
+        ):
+            decorator, confidence = decorators_map["singledispatchmethod"]
```
DanielNoord, line 648: Same as above.

pylint/checkers/utils.py
DanielNoord: Do you need any help with that?

pylint/checkers/utils.py
```diff
@@ -1476,6 +1480,46 @@ def is_registered_in_singledispatch_function(node: nodes.FunctionDef) -> bool:
...
+
+def is_registered_in_singledispatchmethod_function(node: nodes.FunctionDef) -> bool:
+    """Check if the given function node is a 'singledispatchmethod' function."""
+
+    singledispatchmethod_qnames = (
+        "functools.singledispatchmethod",
+        "singledispatch.singledispatchmethod",
+    )
+
+    if not isinstance(node, nodes.FunctionDef):
```
DanielNoord, line 1511: This should be unreachable based on the typing, right?

DanielNoord (commented):
Looking good :smile:

pylint/checkers/utils.py
ramonsaraiva: It should, I think I copied this line from the other function that does something similar. I'll remove.

DanielNoord: I mean, I would also be okay with widening the typing and having this accept all types of `nodes`. Although I wouldn't know why you would ever call this with anything other than a `FunctionDef`. Up to you~

ramonsaraiva:
Do you know why that documentation check is failing?
```
sphinx-build -b linkcheck -q -d _build/doctrees -T -W --keep-going   -n . _build/linkcheck
/home/runner/work/pylint/pylint/doc/user_guide/checkers/features.rst:933: WARNING: broken link: https://docs.python.org/3/library/f (404 Client Error: Not Found for url: https://docs.python.org/3/library/f)
```
Here line 933 doesn't have a URL and both functools URLs seem to be complete.

DanielNoord:
@ramonsaraiva You can add the name here: https://github.com/PyCQA/pylint/blob/main/.pyenchant_pylint_custom_dict.txt

ramonsaraiva:
Cool. Besides the documentation check - which I validated that locally the links are not broken - I think we should be good, but please let me know if you think something else needs to be changed.

DanielNoord:
There does seem to be an issue with the documentation. See:

https://pylint--7575.org.readthedocs.build/en/7575/user_guide/checkers/features.html

There is a broken link there. Could you perhaps check if you understand what is causing this?

ramonsaraiva:
@DanielNoord what I found out this morning is that changing `https://docs.python.org/3/library/functools.html#functools.singledispatchmethod` to `https://docs.python.org/3/library/functools.html#functools.singledispatchmet` (<= 90 columns in the line), the doc build doesn't break the URL. Do you have any suggestions on how I can approach this? Maybe we can create an alias in rst for `https://docs.python.org` or something to have those URLs smaller for the auto doc generator?

To avoid increasing scope, I could also remove the reference to the docs and have a followup fixing this case. I also thought about possibly mentioning `See singledispatchmethod in https://docs.python.org/3/library/functools.html`, but not sure if that's as helpful as a direct link to the `singledispatchmethod` documentation.

DanielNoord:
> @DanielNoord what I found out this morning is that changing `https://docs.python.org/3/library/functools.html#functools.singledispatchmethod` to `https://docs.python.org/3/library/functools.html#functools.singledispatchmet` (<= 90 columns in the line), the doc build doesn't break the URL. Do you have any suggestions on how I can approach this? Maybe we can create an alias in rst for `https://docs.python.org` or something to have those URLs smaller for the auto doc generator?
> 
> 
> 
> To avoid increasing scope, I could also remove the reference to the docs and have a followup fixing this case. I also thought about possibly mentioning `See singledispatchmethod in https://docs.python.org/3/library/functools.html`, but not sure if that's as helpful as a direct link to the `singledispatchmethod` documentation.

Either of the two last solutions would be okay for me. I don't really have the bandwidth to follow this up and I don't think the feature should be blocked by something so trivial.

DanielNoord (dismissed):
Thanks @ramonsaraiva!

Reviewing this was a breeze! LGTM but I'll let another maintainer have a look before merging as this is a new checker.

ramonsaraiva:
Thanks for the great guidance @DanielNoord - it was pretty awesome to get at least to the surface of how pylint works. I hope I can help contributing with more things in the near future.

Pierre-Sassoulas (changes requested):
Looks great already, thank you ! Could you add a good/bad example for each messages in the documentation please ? You can take example on https://github.com/PyCQA/pylint/pull/7576

ramonsaraiva:
@Pierre-Sassoulas I tried adding a good and bad example earlier today, but for some reason the `bad.py` was not appearing in the checker page, only the `good.py`. It was setup exactly as the example you sent (https://github.com/PyCQA/pylint/pull/7576). Maybe it was something related to my doc build, but I couldn't get the `bad.py` examples to appear in the documentation page.

Pierre-Sassoulas:
Could you push the work in progress of good/bad please, so we can investigate, please ?

ramonsaraiva:
Will do that early tomorrow, thanks.

pylint/checkers/utils.py
```diff
@@ -1476,6 +1480,43 @@ def is_registered_in_singledispatch_function(node: nodes.FunctionDef) -> bool:
     return False
 
 
+def find_infered_fn_from_register(node: nodes.NodeNG) -> nodes.FunctionDef | None:
```
mbyrnepr2, line 1483:
```suggestion
def find_inferred_fn_from_register(node: nodes.NodeNG) -> nodes.FunctionDef | None:
```

pylint/checkers/utils.py
```diff
@@ -1476,6 +1480,43 @@ def is_registered_in_singledispatch_function(node: nodes.FunctionDef) -> bool:
...
+    """Check if the given function node is a singledispatchmethod function."""
+
+    singledispatchmethod_qnames = (
+        "functools.singledispatchmethod",
+        "singledispatch.singledispatchmethod",
+    )
+
+    decorators = node.decorators.nodes if node.decorators else []
+    for decorator in decorators:
+        func_def = find_infered_fn_from_register(decorator)
```
mbyrnepr2, line 1513:
```suggestion
        func_def = find_inferred_fn_from_register(decorator)
```

pylint/checkers/utils.py
```diff
@@ -1476,6 +1480,43 @@ def is_registered_in_singledispatch_function(node: nodes.FunctionDef) -> bool:
     return False
 
 
+def find_infered_fn_from_register(node: nodes.NodeNG) -> nodes.FunctionDef | None:
+    # func.register are function calls or register attributes
+    # when the function is annotated with types
```
mbyrnepr2, line 1485: Docstring instead of comment?

mbyrnepr2 (commented):
Overall looks great. A couple of small notes.

pylint/checkers/utils.py
ramonsaraiva: There's another similar function that we decided to not yet modify to simplify the review of this pull request, I'll be creating a follow-up pull request that will combine some common functionality between the function that finds `singledispatch` and `singledispatchmethod`. That function also has a similar comment for function calls.

Should I add a proper docstring as part of the followup PR and I'll combine common functionality between both functions?

mbyrnepr2: That sounds good to me 👌

ramonsaraiva:
@Pierre-Sassoulas @mbyrnepr2 https://github.com/PyCQA/pylint/pull/7575/commits/751ae3b95c114b439c3b920de325298b444d59bf adds the `good.py` and `bad.py` examples.. this is what I see in my documentation build after adding them:

image

image

I'm not sure why it is not showing the `bad.py` examples on the page,  I can see that other checkers have both good and bad. LMK if you have any idea of what it could be. Thanks!

DanielNoord:
https://pylint--7575.org.readthedocs.build/en/7575/user_guide/messages/error/singledispatch-method.html

Seems to be working on RTD :sweat_smile:

doc/data/messages/s/singledispatch-method/bad.py
```diff
@@ -0,0 +1,19 @@
+from functools import singledispatch
+
+
+class BadBoard:
```
Pierre-Sassoulas, line 4:
```suggestion
class Board:
```

doc/data/messages/s/singledispatch-method/good.py
```diff
@@ -0,0 +1,19 @@
+from functools import singledispatch
+
+
+class GoodBoard:
```
Pierre-Sassoulas, line 4:
```suggestion
class Board:
```
Nitpick : if the differences between good/bad are only problematic code construct vs good code construct it's easier to understand the example

doc/data/messages/s/singledispatchmethod-function/bad.py
```diff
@@ -0,0 +1,19 @@
+from functools import singledispatchmethod
+
+
+class BadBoard:
```
Pierre-Sassoulas, line 4:
```suggestion
class Board:
```

doc/data/messages/s/singledispatchmethod-function/good.py
```diff
@@ -0,0 +1,18 @@
+from functools import singledispatchmethod
+
+
+class GoodBoard:
```
Pierre-Sassoulas, line 4:
```suggestion
class Board:
```

Pierre-Sassoulas (dismissed):
Great first contribution @ramonsaraiva ! This should be available in 2.16.0.

mbyrnepr2 (approved):

ramonsaraiva:
Thank you all for the support. Really enjoyed contributing here.

Pierre-Sassoulas (approved):

PR #7577, opened 2022-10-06
Add message name is parameter and nonlocal

### LINKED ISSUE #6882: Add message `name-is-parameter-and-nonlocal`
### Current problem

The following produces a SyntaxError:
```python
def tomato(is_tasty: bool = True):
    nonlocal is_tasty
```

```
python example.py
  File "/Users/markbyrne/programming/example.py", line 2
    nonlocal is_tasty
    ^^^^^^^^^^^^^^^^^
SyntaxError: name 'is_tasty' is parameter and nonlocal
```

### Desired solution

Create a new message in Pylint and emit it in this scenario indicating the error.

### Additional context

_No response_

ramonsaraiva:
## Type of Changes

## Description

Adds a new message for when a `nonlocal` name is also a parameter in a function scope

Closes #6882

### DIFF
--- a/doc/user_guide/checkers/features.rst
+++ b/doc/user_guide/checkers/features.rst
@@ -65,7 +65,7 @@ Basic checker Messages
   Emitted when a name is used prior a global declaration, which results in an
   error since Python 3.6. This message can't be emitted when using Python <
   3.6.
-:name-is-parameter-and-nonlocal (E0134): *Name %r is parameter and nonlocal*
+:name-is-parameter-and-nonlocal (E0134): *Name %r is both parameter and nonlocal*
   Emitted when a nonlocal variable is also a parameter.
 :return-outside-function (E0104): *Return outside function*
   Used when a "return" statement is found outside a function or method.
--- a/doc/user_guide/checkers/features.rst
+++ b/doc/user_guide/checkers/features.rst
@@ -65,6 +65,8 @@ Basic checker Messages
   Emitted when a name is used prior a global declaration, which results in an
   error since Python 3.6. This message can't be emitted when using Python <
   3.6.
+:name-is-parameter-and-nonlocal (E0134): *Name %r is parameter and nonlocal*
+  Emitted when a nonlocal variable is also a parameter.
 :return-outside-function (E0104): *Return outside function*
   Used when a "return" statement is found outside a function or method.
 :return-arg-in-generator (E0106): *Return with argument inside generator*
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -204,6 +204,11 @@ class BasicErrorChecker(_BasicChecker):
             "which results in an error since Python 3.6.",
             {"minversion": (3, 6)},
         ),
+        "E0134": (
+            "Name %r is parameter and nonlocal",
+            "name-is-parameter-and-nonlocal",
+            "Emitted when a nonlocal variable is also a parameter.",
+        ),
     }
 
     def open(self) -> None:
@@ -267,6 +272,7 @@ class BasicErrorChecker(_BasicChecker):
     )
     def visit_functiondef(self, node: nodes.FunctionDef) -> None:
         self._check_nonlocal_and_global(node)
+        self._check_nonlocal_and_parameter(node)
         self._check_name_used_prior_global(node)
         if not redefined_by_decorator(
             node
@@ -326,6 +332,14 @@ class BasicErrorChecker(_BasicChecker):
                 self.add_message(
                     "used-prior-global-declaration", node=node_name, args=(name,)
                 )
+
+    def _check_nonlocal_and_parameter(self, node: nodes.FunctionDef) -> None:
+        """Check that a name is both parameter and nonlocal."""
+        node_arg_names = {arg.name for arg in node.args.args}
+        for body_node in node.nodes_of_class(nodes.Nonlocal):
+            for non_local_name in body_node.names:
+                if non_local_name in node_arg_names:
+                    self.add_message("name-is-parameter-and-nonlocal", node=body_node)
 
     def _check_nonlocal_and_global(self, node: nodes.FunctionDef) -> None:
         """Check that a name is both nonlocal and global."""
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -205,7 +205,7 @@ class BasicErrorChecker(_BasicChecker):
             {"minversion": (3, 6)},
         ),
         "E0134": (
-            "Name %r is parameter and nonlocal",
+            "Name %r is both parameter and nonlocal",
             "name-is-parameter-and-nonlocal",
             "Emitted when a nonlocal variable is also a parameter.",
         ),
@@ -269,6 +269,7 @@ class BasicErrorChecker(_BasicChecker):
         "duplicate-argument-name",
         "nonlocal-and-global",
         "used-prior-global-declaration",
+        "name-is-parameter-and-nonlocal",
     )
     def visit_functiondef(self, node: nodes.FunctionDef) -> None:
         self._check_nonlocal_and_global(node)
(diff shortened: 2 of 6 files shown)


mbyrnepr2:
At first glance since this is a new check, could you also please add a good.py/bad.py example please? Here is where they live: https://github.com/PyCQA/pylint/tree/main/doc/data/messages

These can be tested locally with `tox test_doc`.

tests/functional/n/name/name_is_parameter_and_nonlocal.py
```diff
@@ -0,0 +1,6 @@
+# pylint: disable=missing-docstring,invalid-name,nonlocal-without-binding
+
+
+def tomato(is_tasty: bool = True):
+    nonlocal color
+    nonlocal is_tasty # [name-is-parameter-and-nonlocal]
```
DanielNoord, line 6:
```suggestion
    nonlocal is_tasty # [name-is-parameter-and-nonlocal]


def tomato(*is_tasty):
    nonlocal is_tasty # [name-is-parameter-and-nonlocal]
```

I think this would also raise a syntax error, right?

pylint/checkers/base/basic_error_checker.py
```diff
@@ -204,6 +204,11 @@ class BasicErrorChecker(_BasicChecker):
             "which results in an error since Python 3.6.",
             {"minversion": (3, 6)},
         ),
+        "E0134": (
+            "Name %r is parameter and nonlocal",
+            "name-is-parameter-and-nonlocal",
+            "Emitted when a nonlocal variable is also a parameter.",
```
DanielNoord, line 210:
```suggestion
            "Emitted when a nonlocal variable is also a parameter within a function defintion.",
```

pylint/checkers/base/basic_error_checker.py
```diff
@@ -204,6 +204,11 @@ class BasicErrorChecker(_BasicChecker):
             "which results in an error since Python 3.6.",
             {"minversion": (3, 6)},
         ),
+        "E0134": (
+            "Name %r is parameter and nonlocal",
```
DanielNoord, line 208:
```suggestion
            "Name %r is both parameter and nonlocal",
```

DanielNoord (changes requested):
Could you also add some test with nested scopes and functions?

doc/user_guide/checkers/features.rst
```diff
@@ -65,6 +65,8 @@ Basic checker Messages
   Emitted when a name is used prior a global declaration, which results in an
   error since Python 3.6. This message can't be emitted when using Python <
   3.6.
+:name-is-parameter-and-nonlocal (E0134): *Name %r is parameter and nonlocal*
```
Pierre-Sassoulas, line 68: You're one of the very few contributors that actually generated the doc before opening a merge request, conratulation 👍  (Actually, you're probably the first one 😄 )

pylint/checkers/base/basic_error_checker.py
```diff
@@ -327,6 +333,19 @@ def _check_name_used_prior_global(self, node: nodes.FunctionDef) -> None:
                     "used-prior-global-declaration", node=node_name, args=(name,)
                 )
```
Pierre-Sassoulas, line 335: We can probably optimize by using ``util.only_required_for_messages`` here.

Pierre-Sassoulas (changes requested):

tests/functional/n/name/name_is_parameter_and_nonlocal.py
ramonsaraiva: I guess for that I can do something like
```python
        node_arg_names = {arg.name for arg in node.args.args}
        for local_name, local_nodes in node.locals.items():
            for local_node in local_nodes:
                if isinstance(local_node, nodes.Arguments):
                    node_arg_names.add(local_name)
```

Not sure if there's an easier way to retrieve locals that are considered arguments.

ramonsaraiva: For nested scopes and functions, are you looking for something like this?
```python
def grape(is_tasty: bool = True):
    nonlocal is_tasty # [name-is-parameter-and-nonlocal]

    def grape_juice(is_tasty: bool = True):
        nonlocal is_tasty  # [name-is-parameter-and-nonlocal]

    grape_juice()
```

DanielNoord: Doesn't `node.args.arguments()` work? According to the docstring that should be able to return all names that are used as parameters.

pylint/checkers/base/basic_error_checker.py
ramonsaraiva: I'm good to change to `.. is both parameter and nonlocal`, but python says `.. is parameter and nonlocal`:
```python
>>> def tomato(is_tasty):
...     nonlocal is_tasty
...
  File "<stdin>", line 2
SyntaxError: name 'is_tasty' is parameter and nonlocal
```
wdyt?

tests/functional/n/name/name_is_parameter_and_nonlocal.py
DanielNoord: Yeah, and variants of it. I haven't tested this out locally to see which combinations return a syntax error, but I think it is good to add some (regression tests) for possible combinations:
- outer scope with nonlocal inner scope with nonlocal
- outer scope without nonlocal inner scope with nonlocal
- outer scope with nonlocal inner scope without nonlocal
etc.

Scoping rules can create annoying false positives for messages like this.

pylint/checkers/base/basic_error_checker.py
DanielNoord: Hm, I'd say let's overwrite Python here and use more "readable" English. As pylint output is often read form IDEs I think we an be a little bit more verbose than the traditional error messages Python uses.

tests/functional/n/name/name_is_parameter_and_nonlocal.py
ramonsaraiva: > Doesn't `node.args.arguments()` work? According to the docstring that should be able to return all names that are used as parameters.

When I inspect `node.args` for a function that uses `(*is_tasty)`, it comes pretty much empty:
```python
(Pdb) vars(node.args)
{'lineno': None, 'col_offset': None, 'parent': <FunctionDef.watermelon l.9 at 0x7efe6e8078e0>, 'end_lineno': None, 'end_col_offset': None, 'position': None, 'vararg': 'is_tasty', 'kwarg': None, 'posonlyargs': [], 'posonlyargs_annotations': [], 'kwonlyargs_annotations': [], 'type_comment_args': [], 'type_comment_kwonlyargs': [], 'type_comment_posonlyargs': [], 'varargannotation': None, 'kwargannotation': None, 'args': [], 'defaults': [], 'kwonlyargs': [], 'kw_defaults': [], 'annotations': [], 'fromlineno': 9, 'tolineno': 9, 'arguments': []}
(Pdb) node.args.arguments
[]
```

pylint/checkers/base/basic_error_checker.py
ramonsaraiva: Totally agree.

tests/functional/n/name/name_is_parameter_and_nonlocal.py
ramonsaraiva: Sure! I can attempt that later today, this might also be related https://github.com/PyCQA/astroid/issues/761 - but not totally sure.

DanielNoord: Indeed let's not block this on too much refactoring and fix stuff as we run into them. I think this is a very clear use-case for a potential fix and I'd be happy to help release a patch version of `astroid` for it. I think there are other issues waiting on a patch as well, a regression in dataclasses for example.

DanielNoord:
This seems like an easy PR for a maintainer to pick up.

@ramonsaraiva are you still planning to return to this? If not, would you mind if I picked it up?

ramonsaraiva:
All good if you want to pick it up @DanielNoord - things got busier around here. I'll keep an eye on the solution!

DanielNoord:
The tests here show a genuine failing false positive.

PR #7579, opened 2022-10-06
Handle patterns in ignored-modules when checking for no-name-in-module

### LINKED ISSUE #7578: no-name-in-module does not respect patterns in ignored-modules
### Bug description

`ignored-modules` is supposed to support Unix-style pattern matching, but this is only the case for the `no-member` check whereas the `no-name-in-module` check only checks for exact matches. I've already fixed the issue and I'll submit a PR in a moment.

### Minimal example
```
.
├── a.py
└── foo
    ├── bar_pb2.py
    └── __init__.py
```
a.py:
```python
"""a.py"""
from foo.bar_pb2 import Bar
Bar()
```

### Command used

```shell
pylint a.py --ignored-modules='*_pb2'
```

### Pylint output

```shell
************* Module a
a.py:2:0: E0611: No name 'Bar' in module 'foo.bar_pb2' (no-name-in-module)
```

### Expected behavior

No error

### Pylint version

```shell
pylint 2.15.3
astroid 2.12.10
Python 3.8.10 (default, Jun 22 2022, 20:18:18)
[GCC 9.4.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

danmou:
## Type of Changes

## Description

Closes #7578


DanielNoord:
Could you add a test for this as well?

pylint/checkers/typecheck.py
```diff
@@ -124,32 +123,8 @@ def _is_owner_ignored(
...
-
-        while parts:
-            part = parts.popleft()
-            if not current_module:
-                current_module = part
-            else:
-                current_module += f".{part}"
-            if current_module in ignored_modules:
-                return True
+    if is_module_ignored(owner.root(), ignored_modules):
```
Pierre-Sassoulas, line 126: I like the new function, but could it be a refactor in another MR please ? So if we need to revert later we don't also revert the refactor and for easier review now :)

Pierre-Sassoulas (changes requested):

pylint/checkers/typecheck.py
Danmou: Not sure what you mean, do you want me to duplicate the code instead?

Danmou: Ah I get it now, yes I'll do that

Danmou: Done: #7604 (contains the first commit of this PR)

Pierre-Sassoulas: Thank you, would you mind rebasing this one so GitHub realize that one of the commit is now on main, please ?

Danmou:
I added some test cases and discovered that the import-error check also needed to be updated to use the new function. This is now done.

pylint/checkers/utils.py
```diff
@@ -2032,35 +2031,33 @@ def is_augmented_assign(node: nodes.Assign) -> tuple[bool, str]:
...
+    returns
+        ['pylint', 'pylint.checkers', 'pylint.checkers.ImportsChecker']
+    """
+    names = module_name.split(".")
+    return [".".join(names[0 : i + 1]) for i in range(len(names))]
+
+
 def is_module_ignored(
-    module: nodes.Module,
+    module_name: str,
```
Pierre-Sassoulas, line 2047:
```suggestion
    qualified_module_name: str,
```
```suggestion
    module: nodes.Module,
```
I think the refactor to give a string instead of a node module might not be the best in term of potential error being made later. It's possible to forget to give the qualified name by mistake (the name is still a string), but if you have to give a ``node.Module``, mypy will remind you that. Also it means the complexity of handling the module ignore is is kept inside the function that deals with it.

pylint/checkers/variables.py
```diff
@@ -2715,7 +2716,7 @@ def _check_module_attrs(
                 if module is astroid.Uninferable:
                     return None
             except astroid.NotFoundError:
-                if module.name in self._ignored_modules:
+                if is_module_ignored(f"{module.qname()}.{name}", self._ignored_modules):
```
Pierre-Sassoulas, line 2719: I don't understand what is done here, would you mind explaining ? Is there a problem with the ``module.qname()`` ? The fix probably needs to be in astroid directly if that's the case.

pylint/checkers/utils.py
Danmou: I agree, but the function also needs to be used by `_ignore_import_failure` (called by `_get_imported_module`), which only has access to an `ImportNode` and a string, which is the qualified module name. I don't think it's possible to create a `nodes.Module` without actually resolving the import, which we can't do in the case of `import-error`. Or do you have a different suggestion?

pylint/checkers/variables.py
Danmou: In this function we're checking for `no-name-in-module` errors, e.g. `from foo import bar` where `bar` is not found. In this case `module.qname()` is "foo" and `name` is "bar", and since `bar` may be a module we need to check if "foo.bar" matches the ignored modules.

Pierre-Sassoulas: All right, that definitely need to be added as a comment 😄 

```suggestion
                # We're checking for no-name-in-module errors, e.g. 'from foo import bar' where bar is not found. 
                # In this case module.qname() is "foo" and name is "bar", and since bar may be a module we need
                # to check if "foo.bar" matches the ignored modules.
                if is_module_ignored(f"{module.qname()}.{name}", self._ignored_modules):
```

pylint/checkers/utils.py
Pierre-Sassoulas: We should probably just rename the str argument to ``qualified_module_name`` like in the first suggestion then.

doc/whatsnew/fragments/7578.bugfix
```diff
@@ -0,0 +1,3 @@
+Handle patterns in ignored-modules when checking for no-name-in-module
```
Pierre-Sassoulas, line 1:
```suggestion
The ``ignored-modules`` option will now be correctly taken into account for ``no-name-in-module``.
```

Pierre-Sassoulas (changes requested):

doc/whatsnew/fragments/7578.bugfix
Pierre-Sassoulas: The "important" part is the dot, to make pre-commit pass, but I also suggested a clearer summary, let me know what you think.

Danmou: Thanks!

Pierre-Sassoulas (dismissed):
👍

Pierre-Sassoulas (commented):
An issue issue with mypy that seem genuine appeared, would you mind handling the null case, please ?

Pierre-Sassoulas:
@Danmou will you have time to look at the mypy issue in the near future ? Would you mind if we finish this MR ourselves ?

jproffitt:
@danmou @Pierre-Sassoulas Any update on this? This fix would be extremely helpful for me

danmou:
I won't have time to look into this in the near future, @Pierre-Sassoulas feel free to take it over.

pylint/checkers/utils.py
```diff
@@ -2032,35 +2031,33 @@ def is_augmented_assign(node: nodes.Assign) -> tuple[bool, str]:
...
+    """
+    names = qualified_module_name.split(".")
+    return [".".join(names[0 : i + 1]) for i in range(len(names))]
+
+
 def is_module_ignored(
-    module: nodes.Module,
+    qualified_module_name: str,
     ignored_modules: Iterable[str],
 ) -> bool:
```
jproffitt, line 2049:
```suggestion
    qualified_module_name: str | None,
    ignored_modules: Iterable[str],
) -> bool:
    if qualified_module_name is None:
        return False
```

Pierre-Sassoulas: Let's try :)

Pierre-Sassoulas: Look like it worked, thank you !

Pierre-Sassoulas:
Unsure if we put that in 2.17.2, depending on what no-name-in-module was supposed to do it could be a bug fix or a new feature.

DanielNoord (approved):

jproffitt:
@Pierre-Sassoulas Seems to me like it was a bug because the documentation said pattern matching was supported: https://pylint.readthedocs.io/en/v2.17.2/user_guide/configuration/all-options.html#ignored-modules

PR #7580, opened 2022-10-06
py_ver should be used when accounting for max/min version

### LINKED ISSUE #7569: py-version should use maxversion and minversion
### Current problem

According to pylint

```
:boolean-datetime (W1502): *Using datetime.time in a boolean context.*
  Using datetime.time in a boolean context can hide subtle bugs when the time
  they represent matches midnight UTC. This behaviour was fixed in Python 3.5.
  See https://bugs.python.org/issue13936 for reference. This message can't be
  emitted when using Python >= 3.5.
```

Since only py3.7+ versions are supported, I think we can safely delete this checker.

### Desired solution

Removal of all `boolean-datetime`  checker code.

### Additional context

_No response_

clavedeluna:
## Type of Changes

## Description

Refs #XXXX

Closes #7569

### DIFF
--- a/pylint/message/message_definition.py
+++ b/pylint/message/message_definition.py
@@ -70,11 +70,11 @@ class MessageDefinition:
     def __str__(self) -> str:
         return f"{repr(self)}:\n{self.msg} {self.description}"
 
-    def may_be_emitted(self) -> bool:
-        """Return True if message may be emitted using the current interpreter."""
-        if self.minversion is not None and self.minversion > sys.version_info:
+    def may_be_emitted(self, py_version: tuple = sys.version_info) -> bool:
+        """Return True if message may be emitted using the configured py_version"""
+        if self.minversion is not None and self.minversion > py_version:
             return False
-        if self.maxversion is not None and self.maxversion <= sys.version_info:
+        if self.maxversion is not None and self.maxversion <= py_version:
             return False
         return True
 
--- a/tests/functional/b/boolean_datetime.py
+++ b/tests/functional/b/boolean_datetime.py
@@ -1,0 +1,15 @@
+"""Test boolean-datetime
+
+'py-version' needs to be set to < '3.5'.
+"""
+import datetime
+
+if datetime.time(0, 0, 0):  # [boolean-datetime]
+    print("datetime.time(0,0,0) is not a bug!")
+else:
+    print("datetime.time(0,0,0) is a bug!")
+
+if datetime.time(0, 0, 1):  # [boolean-datetime]
+    print("datetime.time(0,0,1) is not a bug!")
+else:
+    print("datetime.time(0,0,1) is a bug!")
--- a/tests/functional/b/boolean_datetime.rc
+++ b/tests/functional/b/boolean_datetime.rc
@@ -1,0 +1,3 @@
+[testoptions]
+max_pyver=3.5
+
--- a/tests/message/unittest_message_definition.py
+++ b/tests/message/unittest_message_definition.py
@@ -51,16 +51,16 @@ class FalseChecker(BaseChecker):
 
 class TestMessagesDefinition:
     @staticmethod
-    def assert_with_fail_msg(msg: MessageDefinition, expected: bool = True) -> None:
+    def assert_with_fail_msg(msg: MessageDefinition, expected: bool = True, py_version: tuple = sys.version_info) -> None:
         fail_msg = (
             f"With minversion='{msg.minversion}' and maxversion='{msg.maxversion}',"
-            f" and the python interpreter being {sys.version_info} "
+            f" and the python interpreter being {py_version} "
             "the message should{}be emitable"
         )
         if expected:
-            assert msg.may_be_emitted(), fail_msg.format(" ")
+            assert msg.may_be_emitted(py_version), fail_msg.format(" ")
         else:
-            assert not msg.may_be_emitted(), fail_msg.format(" not ")
+            assert not msg.may_be_emitted(py_version), fail_msg.format(" not ")
 
     @staticmethod
     def get_message_definition() -> MessageDefinition:
@@ -73,7 +73,7 @@ class TestMessagesDefinition:
             WarningScope.NODE,
         )
 
-    def test_may_be_emitted(self) -> None:
+    def test_may_be_emitted_default(self) -> None:
         major = sys.version_info.major
         minor = sys.version_info.minor
         msg = self.get_message_definition()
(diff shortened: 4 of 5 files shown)


pylint/message/message_definition.py
```diff
@@ -70,11 +70,11 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return f"{repr(self)}:\n{self.msg} {self.description}"
 
-    def may_be_emitted(self) -> bool:
-        """Return True if message may be emitted using the current interpreter."""
-        if self.minversion is not None and self.minversion > sys.version_info:
+    def may_be_emitted(self, py_version: tuple = sys.version_info) -> bool:
```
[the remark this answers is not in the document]
clavedeluna: Please refer to the last paragraph of the PR description. How would I pass in py_version in the case of message_definition_store.py:find_emittable_messages in code `if message.may_be_emitted():` ? This is why I kept version_info as default, to remain consistent with previous code.

clavedeluna:
Maintainers, it's worth pointing out that this PR was still in draft mode.... I hadn't finished reviewing CI. Just FYI so you don't spend time reviewing code that will change.

pylint/message/message_definition.py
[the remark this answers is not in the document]
clavedeluna: > Isn't `sys.version_info`a nice enough default ? Basically it's the default value we had before implicitly.

That's my belief too :) It doesn't break any code. This PR isn't meant to magically fix everything, it's incremental.

Pierre-Sassoulas (dismissed):
👍

pylint/message/message_definition.py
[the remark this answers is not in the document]
DanielNoord: Yeah, that's also my concern: is this considered public API.

Imo, this is actually making the default implicit. The default of the option is explicit and this is what gets propagated to the function call. By adding the default here we add another source of a potential default: compared to the situation pre-PR there is a new source that could impact the `py-version` behaviour. In a way adding the default actually complicates the situation. I'm not sure that is an actual concern but I did want to raise a point about it as I think this can be easily overlooked.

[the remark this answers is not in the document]
Pierre-Sassoulas: One solution to this could be to deprecate the default value like this:
```python
def may_be_emitted(self, py_version: tuple | None = None) -> bool:
    if py_version is None:
        py_version =  sys.version_info
        warning.warn("Deprecation yadayada 'py_version' will be a required parameter in 3.0 yada yada...")
```
Or a more global refactor where we handle py_version in the message store and do not instantiate message definitions that will never be emitted.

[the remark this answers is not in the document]
DanielNoord: I think that would be good indeed!

doc/whatsnew/fragments/7569.bugfix
```diff
@@ -0,0 +1,3 @@
+Use `py-version` to determine if a message should be emitted for messages defined with `max-version` or `min-version`.
```
jacobtylerwalls, line 1: Single backticks just create italics, so for code highlighting, use two.

Pierre-Sassoulas:
```suggestion
Use ``py-version`` to determine if a message should be emitted for messages defined with ``max-version`` or ``min-version``.
```

pylint/message/message_definition.py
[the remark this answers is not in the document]
Pierre-Sassoulas: How about we take a decision about it in another issue and merge this as is considering there is no change in behavior ? The fast "fix" is deprecation but I'm not sure that forcing every checker to give the ``py_version`` from their config should be the final solution in 3.0, so I would favor the refactor long term. Let's open an issue to discuss this ?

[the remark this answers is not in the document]
DanielNoord: Let's do the deprecation here. `may_be_emitted` shouldn't really be used anyway: checkers should use `is_message_enabled` which itself doesn't even call `may_be_emitted`. So this is really only a method that we use internally but which isn't prefixed with `_` correctly. It's not too much effort to apply your suggestion (which I think is quite elegant).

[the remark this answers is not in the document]
Pierre-Sassoulas: [b4f9022](https://github.com/PyCQA/pylint/pull/7580/commits/b4f9022e3cf2ee9580ca53f90f572d1a5729d838) :)

Pierre-Sassoulas (dismissed):

pylint/message/message_definition.py
```diff
@@ -70,11 +71,24 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return f"{repr(self)}:\n{self.msg} {self.description}"
 
-    def may_be_emitted(self) -> bool:
-        """Return True if message may be emitted using the current interpreter."""
-        if self.minversion is not None and self.minversion > sys.version_info:
+    def may_be_emitted(
+        self,
+        py_version: tuple[int, ...] | sys._version_info | None = None,
```
DanielNoord, line 76:
```suggestion
        py_version: tuple[int, ...] || None = None,
```

`sys._version_info` is equal to `tuple[int, ...]` right?

tests/functional/b/boolean_datetime.py
```diff
@@ -0,0 +1,15 @@
+"""Test boolean-datetime
+
+'py-version' needs to be set to < '3.5'.
```
DanielNoord, line 3:
```suggestion
'py-version' needs to be set to <= '3.5'.
```

tests/message/unittest_message_definition.py
```diff
@@ -51,16 +53,20 @@ def __init__(self) -> None:
...
-    def assert_with_fail_msg(msg: MessageDefinition, expected: bool = True) -> None:
+    def assert_with_fail_msg(
+        msg: MessageDefinition,
+        expected: bool = True,
+        py_version: tuple[int, ...] | sys._version_info = sys.version_info,
+    ) -> None:
         fail_msg = (
             f"With minversion='{msg.minversion}' and maxversion='{msg.maxversion}',"
-            f" and the python interpreter being {sys.version_info} "
+            f" and the python interpreter being {py_version} "
```
DanielNoord, line 63:
```suggestion
            f" and the py-version option being {py_version} "
```

DanielNoord (changes requested):

pylint/message/message_definition.py
Pierre-Sassoulas: I would have said so until very recently 😄  https://github.com/PyCQA/pylint/pull/7587#issuecomment-1272480072

DanielNoord: 😄 

Never mind!

pylint/message/message_definition_store.py
```diff
@@ -9,6 +9,8 @@
 from collections.abc import Sequence, ValuesView
 from typing import TYPE_CHECKING
 
+import pytest
```
DanielNoord, line 12: You noticed this right? 😄

tests/functional/b/boolean_datetime.rc
```diff
@@ -0,0 +1,2 @@
+[testoptions]
+max_pyver=3.5
```
DanielNoord, line 2:
```suggestion
py-version=3.5
```

DanielNoord (changes requested):

pylint/message/message_definition_store.py
Pierre-Sassoulas: Wow. I thought it was inside a test.

Pierre-Sassoulas: I suggest doing this instead:

```python
            if message.may_be_emitted(sys.version_info):
                emittable.append(message)
            else:
                non_emittable.append(message)
```     

I think we need to make a follow-up MR to make things right (give py_version instead the whole way) and I think this is a looooot of effort just for the ``            "list-msgs-enabled",`` command.

DanielNoord: Can't we use the filter from `warnings` itself here? We use it in other places as well.

Pierre-Sassoulas: How about keeping the deprecation warning until we fix it properly in a follow-up MR. (If we just filter the warning then when we make the parameter mandatory in 3.0 we're going to have to fix it anyway)

DanielNoord: Fine with me! This is only called for the CLI command right?

Pierre-Sassoulas: Only in list-msg-enabled, yes. I don't think it's used that often.

Pierre-Sassoulas (dismissed):

clavedeluna:
Feel free to merge when you believe this is ready.

Pierre-Sassoulas (approved):

DanielNoord (approved):

PR #7581, opened 2022-10-06
Take 'accept-no-raise-doc' option into account

### LINKED ISSUE #7208: accept-no-raise-doc is not checked for docstrings with at least one matching section

bchardin:
## Type of Changes

## Description

Previously, Pylint correctly raised `missing-raises-doc` (W9006) when exceptions were raised by a function and:

- some exceptions were correctly documented and others missing, with `accept-no-raise-doc` being true or false
- the docstring had no matching sections (param, raise, returns, yields or property) and `accept-no-raise-doc` was false
- the docstring had at least one matching section (param, returns, yields or property) but no exceptions were referenced and `accept-no-raise-doc` was false

It also incorrectly raised `missing-raises-doc` when:

- the docstring had at least one matching section (param, returns, yields or property) but no exceptions were referenced and `accept-no-raise-doc` was true

This PR fixes this last case, for which I believe `missing-raises-doc` should not be raised.

Closes #7208


doc/whatsnew/fragments/7208.bugfix
```diff
@@ -0,0 +1,10 @@
+The ``accept-no-raise-doc`` option related to ``missing-raises-doc`` will now
```
DanielNoord, line 1: @DudeNr33 @Pierre-Sassoulas Just thinking out loud: can we create a `breaking_change` category in `towncrier` so this fragments gets automatically placed at the top?
We don't need to call it breaking change as I don't think this actually is, so perhaps `user_action_required` or something. I think that would help avoid confusion.

DanielNoord (commented):
Change itself LGTM, just wondering about how to communicate this as clearly as possible. I agree with putting this in a minor release (not a patch release), but we should communicate very clearly as I could imagine it being annoying if you forget this change in 6 months time.

doc/whatsnew/fragments/7208.bugfix
Pierre-Sassoulas: Yeah I think it would be very valuable to have a synthesis of change that user could need to do when upgrading. (for thing like ``no-self-use`` becoming optional and suddenly raising conf warning message, or a bug being fixed that means the behavior will change unexpectedly)

DanielNoord: Does `towncrier` support something like that out of the box?

DudeNr33: I checked the `towncrier` issues and it looks like it was supported in the past, but the latest version seems to have issues (see https://github.com/twisted/towncrier/issues/437).
Will look into it if it works with the version we currently use.

Pierre-Sassoulas (changes requested):
Sorry for the delay in review @bchardin. the changes themselves LGTM. Could you simply rename the fragment to ``user_action`` instead of bugfix now that it's possible, please ?

bchardin:
I just changed the filename from `bugfix` to `user_action`. Is that enough?

Pierre-Sassoulas (approved):
Yes it should be :)

Pierre-Sassoulas:
I'm moving this to 2.16.0 as it's going to raise a lot of new messages seeing the primer and patches. (Nothing to do on your end @bchardin it's just a maintenance decision).

Pierre-Sassoulas:
Congratulation on becoming a pylint contributor @bchardin 🎉 !

Issue #3941, opened 2020-11-10
False positive no-member for enum.Enum's _value2member_map_.


jamesbraza:
### Steps to reproduce

Here's a minimal repro:

```python
"""Show that pylint doesn't see enum's _value2member_map_."""

from enum import Enum


class Foo(Enum):
    """Enum class whose values we care about."""

    BAR = 0
    BAZ = 1

    @classmethod
    def has_value(cls, val: int) -> bool:
        """Get if a value is in this enum.

        SEE: https://stackoverflow.com/a/43634746/11163122

        """
        return val in cls._value2member_map_  # Location of the false positive
```

### Current behavior

Running `pylint` has the output:

```none
has_value_enum.py:19:22: E1101: Class 'Foo' has no '_value2member_map_' member (no-member)
```

### Expected behavior

As the `_value2member_map_` attribute is present, I think `pylint` should not warn that it's not present.

### pylint --version output

```none
pylint 2.6.0
astroid 2.4.2
Python 3.8.6 (default, Oct 27 2020, 11:18:42)
[Clang 12.0.0 (clang-1200.0.32.27)]
```

hippo91:
@jamesbraza thanks for the report. It is probably a lack in the `astroid`'s `brain_namedtuple_enum` brain.

mbyrnepr2:
> Can be closed once https://github.com/PyCQA/astroid/milestone/58 is released.

Closing!

cdce8p:
Unfortunately, this wasn't actually included in astroid `2.12.11`. I've pinned it for `2.12.13` now.

jacobtylerwalls:
> Unfortunately, this wasn't actually included in astroid 2.12.11.

@cdce8p can you help me understand what you're referring to? When I [compare the tags for 2.11 and 2.10](https://github.com/PyCQA/astroid/compare/v2.12.10...v2.12.11), I see this change included. I also downloaded the source for 2.11 from PyPI to be absolutely sure, and I see it there also.

We might need to undo part of https://github.com/PyCQA/astroid/pull/1863.

cdce8p:
> > Unfortunately, this wasn't actually included in astroid 2.12.11.
> 
> @cdce8p can you help me understand what you're referring to? When I [compare the tags for 2.11 and 2.10](https://github.com/PyCQA/astroid/compare/v2.12.10...v2.12.11), I see this change included. I also downloaded the source for 2.11 from PyPI to be absolutely sure, and I see it there also.

[...]

PR #7586, opened 2022-10-07
Add a functional regression test.

mbyrnepr2:
## Description

Refs #3941

### DIFF
--- a/tests/functional/r/regression_02/regression_enum_3941.py
+++ b/tests/functional/r/regression_02/regression_enum_3941.py
@@ -1,0 +1,16 @@
+"""
+Regression test for https://github.com/PyCQA/pylint/issues/3941
+
+E1101: Class 'Veg' has no '_value2member_map_' member (no-member)
+"""
+
+# pylint: disable=protected-access
+
+from enum import Enum
+
+
+class Veg(Enum):
+    ...
+
+
+print(Veg._value2member_map_)


DanielNoord:
I recently changed opinion about these regressions tests: imo they unnecessarily prolong the `pytest` test suite. As long as there is a regression test in `astroid` the only thing this is testing is whether we are using `astroid` above a certain version number. I don't think we need tests for that as I can't see a reason where that would regress unexpectedly.

I'm obviously open to counter arguments as to why we should add such tests!

Pierre-Sassoulas:
Maybe we should think about it on a case by case basis. I think that depending on the complexity of the pylint code that surround the astroid parsing it can make sense or not to add a test. For example if we were previously patching astroid's deficiency or if the checker is very complex we need the regression tests to be sure that we can remove the patching or refactor a checker without regression. If the check is very simple and is only based on astroid's result that is faulty then yes the regression tests is only a waste of time.

mbyrnepr2:
So perhaps the solution would be to move this test to where existing Enum tests live in the Pylint test suite instead of excluding the case entirely. Maybe someday a change in Astroid will reintroduce this issue to Pylint.

DanielNoord:
The test in https://github.com/PyCQA/astroid/pull/1820/ specifically checks for the name in `locals`. Imo, that's good enough for this particular example.

mbyrnepr2:
> The test in [PyCQA/astroid#1820](https://github.com/PyCQA/astroid/pull/1820) specifically checks for the name in `locals`. Imo, that's good enough for this particular example.

OK I'm happy with that!

PR #7587, opened 2022-10-08
Make primer use ``3.10``

DanielNoord:

## Type of Changes

## Description

Primer has been broken since we use `sys.version` to get the correct version for some attributes but that returns something different than the version that GA returns when determining the versions.
I don't really want to start investigating as this is fixed in a month or so. Let's just return to `3.10` while `3.11` is not officially released.

### DIFF
--- a/.github/workflows/primer_comment.yaml
+++ b/.github/workflows/primer_comment.yaml
@@ -37,7 +37,7 @@ jobs:
         id: python
         uses: actions/setup-python@v4.2.0
         with:
-          python-version: "3.11-dev"
+          python-version: "3.10"
 
       # Restore cached Python environment
       - name: Restore Python virtual environment
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -25,7 +25,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        python-version: ["3.7", "3.11-dev"]
+        python-version: ["3.7", "3.10"]
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.2
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -34,7 +34,7 @@ jobs:
     timeout-minutes: 120
     strategy:
       matrix:
-        python-version: ["3.7", "3.11-dev"]
+        python-version: ["3.7", "3.10"]
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.0.2


Pierre-Sassoulas:
Is the primer broken ? I don't see the latest commit on main as being broken ?

DanielNoord:
The commit run fails. There are no comments anymore.

Pierre-Sassoulas:
Hmm, I did not even notice. Shouldn't this fail more noisily ? (Also is the current fail unavoidable ?)

DanielNoord:
I tried to, but there is no good way to fail the comment run as far as I can see.
We just need to check that the comment is there.

jacobtylerwalls (approved):
> (Also is the current fail unavoidable ?)

I looked into it briefly. As Daniël explained, the python-versions are slightly different. The output file produced by the PrimerCommand uses the first three elements of the version info tuple:
```
>>> sys.version_info
sys.version_info(major=3, minor=11, micro=0, releaselevel='candidate', serial=2)
```

But the `python-version` emitted by `setup-python` is 3.11.0.rc-2 because 3.11.0 isn't out yet. We would have to change one or the other: change the PrimerCommand to create files names having "rc-2", which seems like overengineering for something that happens once a fall, or change the Actions script to overwrite the output of `python-version` or strip it to the first three elements everywhere used, which seems error-prone.

I agree we can run 3.10 for a couple weeks. Next year let's sanity check the primer once on a fork (maybe?) and then not officially use it until 3.12 is released.

Pierre-Sassoulas:
I meant the fail in the pipeline of this MR :) 

The base issue itself seems fixable though, shouldn't we simply handle the full sys.version tuple ?

jacobtylerwalls:
Hmm, probably not without knowing how to map from `releaselevel=candidate, serial=2` to `rc2` (couldn't there be other variants?)

Pierre-Sassoulas:
> couldn't there be other variants?)

There would also be alpha and beta (if dill is ready by then). Also apparently there's a "final" value most of the time  (https://docs.python.org/3/library/sys.html#sys.version_info)

> how to map from releaselevel=candidate, serial=2 to rc2

So I would say:
```python
SYS_VERSION_TO_SEMVER : dict[str, str] = {
    "final": "",
    "alpha": "a",
    "beta": "b",
    "candidate": "rc",
}
```

jacobtylerwalls:
Sounds good, would you like to open an alternative? I'm hearing Daniël say he doesn't have the inclination to work on it.

DanielNoord:
Yeah, there are definitely work-arounds but I just wanted to fix this quickly as there are some active PRs that introduce new messages that should ideally be tested with the primer.
I don't really have the time to look into this and it would only be beneficial 1 out of 12 months. I would be okay with merging this for now and hoping that I or somebody else has some more time to deal with this in 3.12.

Pierre-Sassoulas:
I'm going to open an alternative, but let's merge first, not having primer comment is really detrimental and my time is really limited right now.

PR #7588, opened 2022-10-08
Add "breaking change" and "user action" fragment types

DudeNr33:

## Type of Changes

## Description

See https://github.com/PyCQA/pylint/pull/7581#discussion_r990612131

With `towncrier` V22.8.0 and the array-style definition of fragment types in `towncrier.toml` the order of definition is preserved when generating the changelog file. I updated the comment in `towncrier.toml` so we are not tempted to change that in the future. 😁

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,7 +5,7 @@ To ease the process of reviewing your PR, do make sure to complete the following
 
 - [ ] Write a good description on what the PR does.
 - [ ] Create a news fragment with `towncrier create <IssueNumber>.<type>` which will be
-  included in the changelog. `<type>` can be one of: new_check, removed_check, extension,
+  included in the changelog. `<type>` can be one of: breaking, new_check, removed_check, extension,
   false_positive, false_negative, bugfix, other, internal. If necessary you can write
   details or offer examples on how the new change is supposed to work.
 - [ ] If you used multiple emails or multiple names when contributing, add your mails
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,7 +5,7 @@ To ease the process of reviewing your PR, do make sure to complete the following
 
 - [ ] Write a good description on what the PR does.
 - [ ] Create a news fragment with `towncrier create <IssueNumber>.<type>` which will be
-  included in the changelog. `<type>` can be one of: breaking, new_check, removed_check, extension,
+  included in the changelog. `<type>` can be one of: breaking, user_action, new_check, removed_check, extension,
   false_positive, false_negative, bugfix, other, internal. If necessary you can write
   details or offer examples on how the new change is supposed to work.
 - [ ] If you used multiple emails or multiple names when contributing, add your mails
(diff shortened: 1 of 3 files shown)


DudeNr33:
@DanielNoord I named it 'Breaking Changes' for now as I think it is more universal, but we can change that of course.

towncrier.toml
```diff
@@ -7,9 +7,15 @@ issue_format = "`#{issue} <https://github.com/PyCQA/pylint/issues/{issue}>`_"
...
-# custom types as toml tables:
-# https://github.com/twisted/towncrier#further-options
+# We want the changelog to show in the same order as the fragment types
+# are defined here. Therefore we have to use the array-style fragment definition.
+# The table-style definition, although more concise, would be sorted alphabetically.
+# https://github.com/twisted/towncrier/issues/437
+[[tool.towncrier.type]]
+directory = "breaking"
+name = "Breaking Changes"
+showcontent = true
```
Pierre-Sassoulas, line 17:
```suggestion
[[tool.towncrier.type]]
directory = "breaking"
name = "Breaking Changes"
showcontent = true

[[tool.towncrier.type]]
directory = "user_action"
name = "Changes requiring user actions"
showcontent = true
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,7 +5,7 @@ To ease the process of reviewing your PR, do make sure to complete the following
 
 - [ ] Write a good description on what the PR does.
 - [ ] Create a news fragment with `towncrier create <IssueNumber>.<type>` which will be
-  included in the changelog. `<type>` can be one of: new_check, removed_check, extension,
+  included in the changelog. `<type>` can be one of: breaking, new_check, removed_check, extension,
```
Pierre-Sassoulas, line 8:
```suggestion
  included in the changelog. `<type>` can be one of: breaking, user_action, new_check, removed_check, extension,
```

Pierre-Sassoulas (commented):
It's true that the difference between breaking change and thing that might require user actions is tenuous but I think there's a slight difference between fixing a bug or making a checker optional and a real breaking chance that we'd do for 3.0. We could also release major more often and call everything breaking change, but we advertised a lot of breaking changes for pylint 3.0 in the deprecation warning so we don't release a lot of major.  Maybe we should change the deprecation warning and just release majors ?

doc/development_guide/contributor_guide/contribute.rst
```diff
@@ -67,7 +67,7 @@ your patch gets accepted:
 .. keep this in sync with the description of PULL_REQUEST_TEMPLATE.md!
 
 - Create a news fragment with `towncrier create <IssueNumber>.<type>` which will be
-  included in the changelog. `<type>` can be one of: new_check, removed_check, extension,
+  included in the changelog. `<type>` can be one of: breaking, new_check, removed_check, extension,
```
Pierre-Sassoulas, line 70:
```suggestion
  included in the changelog. `<type>` can be one of: breaking, user_action, new_check, removed_check, extension,
```

DudeNr33:
I'm fine with both ways regarding major releases. I would lean a little bit more to the "keep things the way they are", though.
Pylint is a linter, nothing used in production. Being a bit more lax about the semantic versioning is not that tragic in this case.
Plus, we would not force people that have dependencies like "pylint=^2.15" in their requirements to update their dependency declarations that often.

jacobtylerwalls (approved):
No concerns here 👍🏻

DanielNoord (approved):
Thanks for this @DudeNr33 

I agree, current release process works well. With these new fragments we can communicate more clearly, but I don't think we need to release majors more often.

Issue #7464, opened 2022-09-14
pylint is confusing ExitStack() and AsyncExitStack() when both are used


kriek:
### Bug description

pylint is confusing ExitStack() and AsyncExitStack() when analyzing two different modules using each.

Consider as a minimized example, the following two modules.

sync.py
```
from contextlib import ExitStack
def sync_stack_use():
    stack = ExitStack()
    stack.pop_all().close()
```

async.py
```
from contextlib import AsyncExitStack
def async_stack_use():
    stack = AsyncExitStack()
    stack.pop_all().aclose()
```

### Command used

```shell
1. pylint --disable=all --enable=no-member sync.py
2. pylint --disable=all --enable=no-member async.py
3. pylint --disable=all --enable=no-member sync.py async.py
4. pylint --disable=all --enable=no-member async.py sync.py
```

### Pylint output

```shell
1. no finding
2. no finding
3. async.py:4:4: E1101: Instance of 'ExitStack' has no 'aclose' member; maybe 'close'? (no-member)
4. sync.py:4:4: E1101: Instance of 'AsyncExitStack' has no 'close' member; maybe 'aclose'? (no-member)
```

### Expected behavior

1. no finding
2. no finding
3. no finding
4. no finding

### Pylint version

```shell
pylint 2.15.2
astroid 2.12.9
Python 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)]
```

### OS / Environment

Windows (reproduced also on Ubuntu 20.04 LTS).

Pierre-Sassoulas:
Thank you for opening the issue. It look like it could be somewhat related to https://github.com/PyCQA/pylint/issues/4053

kriek:
I don't see the relation: here, the modules are fully different (path and content).

kriek:
The problem is around pop_all() method. In contextlib.py, both ExitStack and AsyncExitStack are inheriting their pop_all() method from _BaseExitStack whose implementation looks like:
```
class _BaseExitStack:
    def pop_all(self):
        # simplified
        return type(self)()
```
It seems pylint is inferring the type of pop_all() return type once and then re-using the same type again whatever the derived class is.
My understanding is incomplete because I failed to reproduce the issue by creating a file reproducing the inheritance diagram and type(self)() usage. Doing so, pylint always infers the base class type, unlike with (Async)ExitStack.

[...]

PR #7602, opened 2022-10-10
Make the order of functional tests deterministic

kriek:
## Type of Changes

| ✓   | Tests                  |

## Description

The pull request fixes the non deterministic ordering of functional tests depending on the underlying system.

Refs #7464

### DIFF
--- a/pylint/testutils/functional/find_functional_tests.py
+++ b/pylint/testutils/functional/find_functional_tests.py
@@ -38,9 +38,11 @@ def get_functional_test_files_from_directory(
 
     _check_functional_tests_structure(Path(input_dir))
 
-    for dirpath, _, filenames in os.walk(input_dir):
+    for dirpath, dirnames, filenames in os.walk(input_dir):
         if dirpath.endswith("__pycache__"):
             continue
+        dirnames.sort()
+        filenames.sort()
         for filename in filenames:
             if filename != "__init__.py" and filename.endswith(".py"):
                 suite.append(FunctionalTestFile(dirpath, filename))


kriek:
@Pierre-Sassoulas I've analyzed the failure with an interesting finding: we miss determinism in functional tests order. On Windows, I have no_member_type_introspection before no_member_type_introspection_base while on Linux, the order seems the other way around. Because we are dealing with a cache bug, the order affects the test outcome.

I propose to remove that non-determinism in get_functional_test_files_from_directory.

Pierre-Sassoulas:
I'm wondering if it's possible to fix the cache issue itself instead ?

kriek:
The cache issue is on the Astroid side. My recommandation would be to do it in 3 steps:
1. Add the test (this PR)
2. Fix the bug (Astroid side, thru dedicated issue)
3. Update the test when Astroid requirement is updated in pytest repo to a version shipping the fix

Does it look good to you?

DanielNoord:
Are we sure that `pylint` is the right place for tests? The `astroid` test suite is much smaller currently so I'm always in favour of putting tests there if they actually belong there.

kriek:
I agree the test is more relevant on the Astroid side. But do we have to choose? Why not both?

DanielNoord:
> I agree the test is more relevant on the Astroid side. But do we have to choose? Why not both?

CI takes about 6 minutes now, about 4 on my own machine. That's quite long if you just want to fix something and see if tests pass. Why test something that is not part of the library you're working on? Should we start testing the `stdlib` as well? It also gives a false sense of security as even if we regress there is nothing to do in this library to fix that regression.

kriek:
Hum, I thought those new tests had a bit of value because they are functional pylint tests about something with unclear outcome at first (i.e. is pylint able to follow type(self)?).

But I see your point and when I think more about it, it is indeed fully out of `pylint` control: if `astroid` is able to correctly infer type(self), the relevant no-member findings will follow and any existing no-member functional test should cover the part under `pylint` responsibility already. I agree to abandon the proposed new pylint tests.

Should we keep the commit to make the order of functional tests deterministic whatever the system?

DanielNoord:
> Should we keep the commit to make the order of functional tests deterministic whatever the system?

Yes, that's a nice fix!

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -38,9 +38,11 @@ def get_functional_test_files_from_directory(
 
     _check_functional_tests_structure(Path(input_dir))
 
-    for dirpath, _, filenames in os.walk(input_dir):
+    for dirpath, dirnames, filenames in os.walk(input_dir):
         if dirpath.endswith("__pycache__"):
             continue
+        dirnames.sort()
+        filenames.sort()
         for filename in filenames:
```
Pierre-Sassoulas, line 46:
```suggestion
        for filename in sorted(filenames):
```

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -38,9 +38,11 @@ def get_functional_test_files_from_directory(
 
     _check_functional_tests_structure(Path(input_dir))
 
-    for dirpath, _, filenames in os.walk(input_dir):
+    for dirpath, dirnames, filenames in os.walk(input_dir):
         if dirpath.endswith("__pycache__"):
             continue
+        dirnames.sort()
+        filenames.sort()
```
Pierre-Sassoulas, line 45:
```suggestion
```
Why does ``dirnames`` need to be sorted ? We're not using it afaik ?

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -38,9 +38,11 @@ def get_functional_test_files_from_directory(
 
     _check_functional_tests_structure(Path(input_dir))
 
-    for dirpath, _, filenames in os.walk(input_dir):
+    for dirpath, dirnames, filenames in os.walk(input_dir):
```
Pierre-Sassoulas, line 41:
```suggestion
    for dirpath, _, filenames in os.walk(input_dir):
```

pylint/testutils/functional/find_functional_tests.py
kriek: This is the [recommended](https://docs.python.org/3/library/os.html#:~:text=the%20caller%20can%20modify%20the%20dirnames%20list%20in%2Dplace) way to make the directory iteration order deterministic using `os.walk()`.

pylint/testutils/functional/find_functional_tests.py
kriek: Given that we have to use in place sort for `dirnames`, I would prefer to do the same for filenames so that the two `sort()` calls involved in traversal order determinism are grouped.

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: Relevant part of the doc:

> When topdown is True, the caller can modify the dirnames list in-place (perhaps using [del](https://docs.python.org/3/reference/simple_stmts.html#del) or slice assignment), and [walk()](https://docs.python.org/3/library/os.html#os.walk) will only recurse into the subdirectories whose names remain in dirnames; this can be used to prune the search, impose a specific order of visiting, or even to inform [walk()](https://docs.python.org/3/library/os.html#os.walk) about directories the caller creates or renames before it resumes [walk()](https://docs.python.org/3/library/os.html#os.walk) again. Modifying dirnames when topdown is False has no effect on the behavior of the walk, because in bottom-up mode the directories in dirnames are generated before dirpath itself is generated.

Pierre-Sassoulas (approved):

PR #7606, opened 2022-10-11
Add docs for various `invalid-*` messages

clavedeluna:
## Type of Changes

## Description

Add docs examples for various `invalid-*` messages

Refs #5953

### DIFF
--- a/doc/data/messages/i/invalid-class-object/bad.py
+++ b/doc/data/messages/i/invalid-class-object/bad.py
@@ -1,0 +1,5 @@
+class A:
+    pass
+
+
+A.__class__ = 1  # [invalid-class-object]
--- a/doc/data/messages/i/invalid-class-object/bad.py
+++ b/doc/data/messages/i/invalid-class-object/bad.py
@@ -1,5 +1,5 @@
-class A:
+class Apple:
     pass
 
 
-A.__class__ = 1  # [invalid-class-object]
+Apple.__class__ = 1  # [invalid-class-object]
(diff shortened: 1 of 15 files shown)


doc/data/messages/i/invalid-class-object/bad.py
```diff
@@ -0,0 +1,5 @@
+class A:
+    pass
+
+
+A.__class__ = 1  # [invalid-class-object]
```
Pierre-Sassoulas, line 5:
```suggestion
class Apple:
    pass


Apple.__class__ = 1  # [invalid-class-object]
```

doc/data/messages/i/invalid-class-object/good.py
```diff
@@ -1 +1,9 @@
-# This is a placeholder for correct code for this message.
+class A:
+    pass
+
+
+class B:
+    pass
+
+
+A.__class__ = B
```
Pierre-Sassoulas, line 9:
```suggestion
class Apple:
    pass


class RedDelicious:
    pass


Apple.__class__ = RedDelicious
```

Pierre-Sassoulas (changes requested):
Thank you !  I'm jst trying to avoid Foo/Bar/FooBar and boring one letter name. I often try to have fruit and animals in example so it's more "lively". Other example could work too, let me know if it make no sense. Admittedly I did not find the use case for changing ``__class__`` live so it might be off..

clavedeluna:
Github's batch commit feature must be down temporarily 🤔

Pierre-Sassoulas (approved):
👌

PR #7611, opened 2022-10-12
Implement chained comparison improvements and related checks

### LINKED ISSUE #5814: Check for circular comparisons and other comparison improvements
### Current problem

```python
a = 1
b = 2

if a > b and b > a:
    pass
```

`compare.py:4:3: R1716: Simplify chained comparison between the operands (chained-comparison)`

While it's nice some message that points to the error, it's not quite correct.

### Desired solution

Detect circular comparisons which simplify to `False`.

This is a graph pattern is solved with a topological sort. The main risks are the performance overhead for building the graph, which should be small because the number of statements in any given `if` is usually low. There is also the complexity risk of having a graph and associated algorithms implementation embedded in the code.

Having the graph could open up the following:

Could see if we can suggest simplifying `if a >= b and b <= a:` to `if a == b:` but this could be more difficult. There might also be something we can do with `or` statements.

Building the graph would also allow us to pick up situations like 
```
    elif a > 1 and a > 10:
        pass
    elif a < 100 and a < 10:
        pass
```
which currently do not emit.

This issue came up while thinking about #5800. It I'm not immediately sure how to suggest the refactor with the current implementation. Building this graph might more readily suggests the refactor.

We could modify `chained-comparison` to detect comparison cycles and do #5800. But unless there are objections I want to try the graph implementation because of the benefits described above.

areveny:
Hi.

This PR adds improvements to `chained-comparison` by processing the chain of comparisons into a graph. It's still a bit unpolished but I would like some higher-level feedback on the direction.

It adds two new checks:

1) `impossible-comparison` for cases when a comparison would always be false, e.g. `a > b and b > a` or `a == 5 and a > b and b == 1`.

2) `comparison-all-equal` for a special cycle case where every comparison allows equality. `a >= b >= c and c >= a` simplifies to `a == b == c`.

It modifies `chained-comparison` to be more accurate, specifically the case `a < b < c and a < c` should be simplified to just `a < b < c`. It currently is not.

It also prints what simplified comparison is.

Unfortunately, this is at the cost of complexity. The algorithm finds the minimum number of comparisons by finding the longest path through the graph of comparisons repeatedly, recalculating all the path lengths after each path is emitted. (I could not get it working by emitting paths in arbitrary order until all nodes are accounted for because of the case above, `a < b < c and a < c`.) The overhead of doing so should be limited by the reality that most comparisons tend to have just a few statements.

Please let me know what you think. I'm open to big suggestions, but want to share my code for feedback and to keep this moving.

## Type of Changes

## Description

Closes #5814


Pierre-Sassoulas (commented):
Thank you for the work @areveny. There's a lot to review but from what I can see after a surface review it's possible that making the gragh check's code independent from the RefactoringChecker might be a good idea, because 1) There's a lot of code and maintenance would benefit from it being separated 2) It's going to be easier if we want to extends ``usiing-constant-tests``directly where it's at instead of refactoring how the checker work to be able to raise any message anywhere.

jacobtylerwalls:
Hi @areveny, just wondering: are the false negatives in the primer result one of the items you were hoping to polish after getting an initial review?

pylint/graph.py
```diff
@@ -210,3 +210,175 @@ def _get_cycles(
...
+        max_nodes_path = max(adj_maxes[1] for adj_maxes in adj_maximums) + 1
+
+    # Update the graph and return
+    symbols_in_longest_path[node] = max_symbols_path
+    nodes_in_longest_path[node] = max_nodes_path
+    return (max_symbols_path, max_nodes_path)
+
+
+def strip_path(path: list[node_types]) -> list[node_types]:
+    """Removes redundant constant comparisons at the ends of a path, e.g. simplies {a,
```
Pierre-Sassoulas, line 363:
```suggestion
    """Removes redundant constant comparisons at the ends of a path, e.g. simplifies {a,
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1337,61 +1349,154 @@ def _check_consider_using_in(self, node: nodes.BoolOp) -> None:
...
+                # Update maps
+                graph_dict[left].add(right)
+                if not graph_dict[right]:
+                    graph_dict[right] = set()  # Ensure the node exists in graph
+                symbol_dict[(left, right)] = operator
+                indegree_dict[left] += 0  # Make sure every node has an entry
+                indegree_dict[right] += 1
+                frequency_dict[(left, right)] += 1
+
+                # advance onto the next comprison if it exists
```
Pierre-Sassoulas, line 1442:
```suggestion
                # advance onto the next comparison if it exists
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1337,61 +1349,154 @@ def _check_consider_using_in(self, node: nodes.BoolOp) -> None:
...
+                    for adj in graph_dict[smaller]:
+                        if isinstance(adj, str):
+                            graph_dict[largest].discard(adj)
+
+        return (graph_dict, symbol_dict, indegree_dict, frequency_dict)
+
+    def _get_compare_operand_value(
+        self, node: nodes.Compare, const_values: list[int | float]
+    ) -> str | int | float | None:
+        value = None
```
Pierre-Sassoulas, line 1475:
```suggestion
        value: str | int | float | None = None
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1337,61 +1349,154 @@ def _check_consider_using_in(self, node: nodes.BoolOp) -> None:
...
+                    # Remove paths from the larger number to the smaller number's adjacent nodes
+                    # This prevents duplicated paths in the output
+                    for adj in graph_dict[smaller]:
+                        if isinstance(adj, str):
+                            graph_dict[largest].discard(adj)
+
+        return (graph_dict, symbol_dict, indegree_dict, frequency_dict)
+
+    def _get_compare_operand_value(
+        self, node: nodes.Compare, const_values: list[int | float]
```
Pierre-Sassoulas, line 1473:
```suggestion
        self, node: nodes.Compare, const_values: list[int | float |None]
```

Pierre-Sassoulas (approved):

PR #7612, opened 2022-10-12
`check_newsfragments`: Improve output on Windows machines

### LINKED ISSUE #7593: `check-newsfragments` - pre-commit -  unclear error message when fragment doesn't end with `.`
### Bug description

When `check-newsfragments` checker fails in `pre-commit` because of no `.` in the end of description the error message is not clear. should be fixed.``

Steps to reproduce:

1. Checkout to main branch.
2. run in terminal `towncreier create 7777.new_check`
3. open `7777.new_check` file and write the following:

```
Bla bla.
bla bla

Closes #7777
```

4. Stage `7777.new_check`
5. run `pre-commit` in terminal

Adding a `.` to the second at the end solves the pre-commit error, but the error message is unclear.

P.S the error message when running the `newsfragments` job in CI actually has a clearer error message that helps you understand what to do to fix the error.

### Command used

run `pre-commit` in terminal

### Pylint output

result in log:
```
Check newsfragments......................................................Failed
- hook id: check-newsfragments
- exit code: 1

  File "C:\Program Files\Python38\lib\runpy.py", line 194, in _run_module_as_main
  File "C:\Program Files\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Or\Documents\Studies\TAU\Term2\Project2\Source_code\pylint_Forked\script\check_newsfragments.py", line 86, in <module>
    sys.exit(main())
  File "C:\Users\Or\Documents\Studies\TAU\Term2\Project2\Source_code\pylint_Forked\script\check_newsfragments.py", line 42, in main
    is_valid &= check_file(Path(filename), args.verbose)
  File "C:\Users\Or\Documents\Studies\TAU\Term2\Project2\Source_code\pylint_Forked\script\check_newsfragments.py", line 61, in check_file
    print(
  File "C:\Program Files\Python38\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 76-77: character maps to <undefined>

```

### Expected behavior

Error message that ideally states which file is the problematic file, and at least say that the error is that there is no `.` in the end of fragment description, or that it does not match the template, etc..

### Pylint version

```shell
pylint 2.16.0-dev
astroid 2.12.10
Python 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64 bit (AMD64)]
```

### OS / Environment

Windows 10 Pro 64 bit

DudeNr33:
## Type of Changes

## Description

Use the way `pydocstringformatter` handles console output in `script/check_newsfragments.py` to avoid `UnicodeEncodeError` on some Windows machines.

Closes #7593

### DIFF
--- a/script/check_newsfragments.py
+++ b/script/check_newsfragments.py
@@ -51,14 +51,14 @@ def check_file(file: Path, verbose: bool) -> bool:
     if match:
         issue = match.group("issue")
         if file.stem != issue:
-            print(
+            log(
                 f"{file} must be named '{issue}.<fragmenttype>', after the issue it references."
             )
             return False
         if verbose:
-            print(f"Checked '{file}': LGTM 🤖👍")
+            log(f"Checked '{file}': LGTM 🤖👍")
         return True
-    print(
+    log(
         f"""\
 {file}: does not respect the standard format 🤖👎
 
@@ -80,6 +80,12 @@ Refs #1234
 """
     )
     return False
+
+
+def log(msg: str) -> None:
+    # To support non-UTF-8 environments like Windows, we need
+    # to explicitly encode the message instead of using plain print()
+    sys.stdout.buffer.write(msg.encode("utf-8"))
 
 
 if __name__ == "__main__":
--- a/script/check_newsfragments.py
+++ b/script/check_newsfragments.py
@@ -51,14 +51,14 @@ def check_file(file: Path, verbose: bool) -> bool:
     if match:
         issue = match.group("issue")
         if file.stem != issue:
-            log(
+            echo(
                 f"{file} must be named '{issue}.<fragmenttype>', after the issue it references."
             )
             return False
         if verbose:
-            log(f"Checked '{file}': LGTM 🤖👍")
+            echo(f"Checked '{file}': LGTM 🤖👍")
         return True
-    log(
+    echo(
         f"""\
 {file}: does not respect the standard format 🤖👎
 
@@ -82,7 +82,7 @@ Refs #1234
     return False
 
 
-def log(msg: str) -> None:
+def echo(msg: str) -> None:
     # To support non-UTF-8 environments like Windows, we need
     # to explicitly encode the message instead of using plain print()
     sys.stdout.buffer.write(msg.encode("utf-8"))
--- a/script/check_newsfragments.py
+++ b/script/check_newsfragments.py
@@ -85,7 +85,7 @@ Refs #1234
 def echo(msg: str) -> None:
     # To support non-UTF-8 environments like Windows, we need
     # to explicitly encode the message instead of using plain print()
-    sys.stdout.buffer.write(f"{msg}\n".encode("utf-8"))
+    sys.stdout.buffer.write(f"{msg}\n".encode())
 
 
 if __name__ == "__main__":
--- a/script/check_newsfragments.py
+++ b/script/check_newsfragments.py
@@ -85,7 +85,7 @@ Refs #1234
 def echo(msg: str) -> None:
     # To support non-UTF-8 environments like Windows, we need
     # to explicitly encode the message instead of using plain print()
-    sys.stdout.buffer.write(msg.encode("utf-8"))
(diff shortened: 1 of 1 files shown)


DudeNr33:
@orSolocate as I could not reproduce it on my own Windows machine, could you give this a go and check if this solves the issue for you?

script/check_newsfragments.py
```diff
@@ -82,5 +82,11 @@ def check_file(file: Path, verbose: bool) -> bool:
     return False
 
 
+def log(msg: str) -> None:
```
Pierre-Sassoulas, line 85:
```suggestion
def echo(msg: str) -> None:
```
Nitpick: I would use the same terminology than click, because logging is something else.

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Blocked by a local test on Windows to confirm it's fixed.

script/check_newsfragments.py
DudeNr33: Makes sense, changed!

Pierre-Sassoulas (dismissed):

Pierre-Sassoulas (commented):
I had an unsent review here 😓

DudeNr33:
Thanks for the catch @Pierre-Sassoulas 

@orSolocate did you find the time to try it out on your machine?

Pierre-Sassoulas (approved):

orSolocate:
> Thanks for the catch @Pierre-Sassoulas
> 
> @orSolocate did you find the time to try it out on your machine?

When I create the falsy `7777.new_check` file I get the actual appropriate error message:


```
doc\whatsnew\fragments\1111.new_check: does not respect the standard format ��

The standard format is:

<one or more line of text ending with a '.'>
<one blank line>
<issue reference> #<issuenumber>

Where <issue reference> can be one of: Refs, Closes, Follow-up in, Fixes part of

The regex used is 're.compile('(?P<description>(.*\\n)*(.*\\.\\n))\\n(?P<ref>(Refs|Closes|Follow-up in|Fixes part of) (PyCQA/astroid)?#(?P<issue>\\d+))', re.MULTILINE)'.

For example:

``pylint.x.y`` is now a private API.

Refs #1234
```

That solves the issue for me. Nice commit.

Pierre-Sassoulas:
Nice ! Let's fix the issue with emoji not being displayed correctly in another MR or never unless you know exactly what to do @DanielNoord ? 😄

orSolocate:
> Nice ! Let's fix the issue with emoji not being displayed correctly in another MR or never unless you know exactly what to do @DanielNoord ? 😄

The emoji not displayed is not import in my opinion as this is only a visual issue, and only Pylint developers see it.
What mattered is that the error message is clear for the user to fix it :) It will help our developers commit quicker, also because integration with `towncrier` is fairly new and not everyone knows about the "ending `.`" restriction

DudeNr33:
What terminal are you using? It seems like the normal CMD is not very good at handling emojis. 
I am also no Encoding expert, but I could also imagine that not every encoding supports all emojis.

DudeNr33:
@Pierre-Sassoulas any idea how to get the two pending tests for Python 3.11 to run? 
I think those were added after opening the PR, and I haven't found a way to trigger them manually.

Pierre-Sassoulas:
Yes you need to rebase on main or merge the latest main to get the latest github actions :)

Pierre-Sassoulas:
Or ignore the error and merge anyway it's pretty safe here

orSolocate:
> What terminal are you using? It seems like the normal CMD is [not very good at handling emojis](https://stackoverflow.com/questions/71205271/python-rich-emojis-not-showing). I am also no Encoding expert, but I could also imagine that not every encoding supports all emojis.

I get those Question marks both in PyCharm terminal  and in GitExtensions Console..

PR #7615, opened 2022-10-13
add new ``invalid-bases-assignment`` check

### LINKED ISSUE #584: [easy] New rules for assignment to __bases__
Originally reported by: **Claudiu Popa (BitBucket: [PCManticore](http://bitbucket.org/PCManticore), GitHub: @PCManticore)**

---
- can assign only tuples to `__bases__`

```
#!python

class A: pass
A.__bases__ = "lala"
```
- can assign only non empty tuples to `__bases__`

```
#!python

class A: pass
A.__bases__ = ()
```
- `__bases__` must be tuple of classes

```
#!python

class A: pass
A.__bases__ = (1, )
```

---
- Bitbucket: https://bitbucket.org/logilab/pylint/issue/584

clavedeluna:
## Type of Changes

## Description

Pylint should warn a user if a class's `__bases__` dunder method is being assigned to anything other than a populated tuple. Python interpreter would raise a `TyperError` in these cases, for example

```
>>> class Apple:
...     pass
...
>>> Apple.__bases__ = "green"  
TypeError: can only assign tuple to Apple.__bases__, not str

>>> Apple.__bases__ = ()
TypeError: can only assign non-empty tuple to Apple.__bases__, not ()

```

So this check would warn a pylint user about this before the interpreter would.

Closes #584

### DIFF
--- a/doc/data/messages/i/invalid-bases-assignment/good.py
+++ b/doc/data/messages/i/invalid-bases-assignment/good.py
@@ -1,4 +1,4 @@
 class Apple:
     pass
 
-Apple.__bases__ = ("green", "red", )  # [invalid-bases-assignment]
+Apple.__bases__ = ("green", "red", )
--- a/doc/data/messages/i/invalid-bases-assignment/good.py
+++ b/doc/data/messages/i/invalid-bases-assignment/good.py
@@ -1,0 +1,4 @@
+class Apple:
+    pass
+
+Apple.__bases__ = ("green", "red", )  # [invalid-bases-assignment]
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -48,7 +48,7 @@ from pylint.checkers.utils import (
     supports_membership_test,
     supports_setitem,
 )
-from pylint.interfaces import INFERENCE, HIGH
+from pylint.interfaces import HIGH, INFERENCE
 from pylint.typing import MessageDefinitionTuple
 
 if sys.version_info >= (3, 8):
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -48,7 +48,7 @@ from pylint.checkers.utils import (
     supports_membership_test,
     supports_setitem,
 )
-from pylint.interfaces import INFERENCE
+from pylint.interfaces import INFERENCE, HIGH
 from pylint.typing import MessageDefinitionTuple
 
 if sys.version_info >= (3, 8):
@@ -397,6 +397,11 @@ MSGS: dict[str, MessageDefinitionTuple] = {
         "Second argument of isinstance is not a type",
         "isinstance-second-argument-not-valid-type",
         "Emitted when the second argument of an isinstance call is not a type.",
+    ),
+    "W1117": (
+        "__bases__ should be assigned to a populated tuple",
+        "invalid-bases-assignment",
+        "Emitted when a non-tuple or empty tuple is assigned to __bases__.",
     ),
 }
 
@@ -1200,12 +1205,12 @@ accessed. Python regular expressions are accepted.",
         "assignment-from-no-return",
         "assignment-from-none",
         "non-str-assignment-to-dunder-name",
+        "invalid-bases-assignment",
     )
     def visit_assign(self, node: nodes.Assign) -> None:
         """Process assignments in the AST."""
-
         self._check_assignment_from_function_call(node)
-        self._check_dundername_is_string(node)
+        self._check_dunder_methods(node)
 
     def _check_assignment_from_function_call(self, node: nodes.Assign) -> None:
         """When assigning to a function call, check that the function returns a valid
@@ -1277,17 +1282,22 @@ accessed. Python regular expressions are accepted.",
             and isinstance(utils.safe_infer(node.func.expr), nodes.List)
         )
 
-    def _check_dundername_is_string(self, node: nodes.Assign) -> None:
-        """Check a string is assigned to self.__name__."""
+    def _check_dunder_methods(self, node: nodes.Assign) -> None:
+        """Check that dunder methods are assigned to expected values."""
 
-        # Check the left-hand side of the assignment is <something>.__name__
         lhs = node.targets[0]
         if not isinstance(lhs, nodes.AssignAttr):
             return
+
+        self._check_dunder_name(node)
+        self._check_dunder_bases(node)
+
+    def _check_dunder_name(self, node: nodes.Assign) -> None:
+        lhs = node.targets[0]
         if not lhs.attrname == "__name__":
             return
 
-        # If the right-hand side is not a string
+        # Check if the right-hand side is not a string
         rhs = node.value
         if isinstance(rhs, nodes.Const) and isinstance(rhs.value, str):
             return
@@ -1295,8 +1305,21 @@ accessed. Python regular expressions are accepted.",
         if not inferred:
             return
         if not (isinstance(inferred, nodes.Const) and isinstance(inferred.value, str)):
-            # Add the message
-            self.add_message("non-str-assignment-to-dunder-name", node=node)
+            self.add_message(
+                "non-str-assignment-to-dunder-name", node=node, confidence=INFERENCE
+            )
+
+    def _check_dunder_bases(self, node: nodes.Assign) -> None:
+        lhs = node.targets[0]
+        if not lhs.attrname == "__bases__":
+            return
+
+        # Check if the right-hand side is not a populated tuple
+        rhs = node.value
+        if isinstance(rhs, nodes.Tuple) and len(rhs.elts) > 0:
+            return
+
+        self.add_message("invalid-bases-assignment", node=node, confidence=HIGH)
 
     def _check_uninferable_call(self, node: nodes.Call) -> None:
         """Check that the given uninferable Call node does not
--- a/tests/functional/i/invalid/invalid_bases_assignment.py
+++ b/tests/functional/i/invalid/invalid_bases_assignment.py
@@ -1,0 +1,13 @@
+# pylint: disable=missing-module-docstring, missing-class-docstring
+# pylint: disable=too-few-public-methods
+
+
+class Apple:
+    pass
+
+
+Apple.__bases__ = ()  # [invalid-bases-assignment]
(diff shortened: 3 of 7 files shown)


pylint/checkers/typecheck.py
```diff
@@ -1277,26 +1282,44 @@ def _is_list_sort_method(node: nodes.Call) -> bool:
...
+        lhs = node.targets[0]
+        if not lhs.attrname == "__bases__":
+            return
+
+        # Check if the right-hand side is not a populated tuple
+        rhs = node.value
+        if isinstance(rhs, nodes.Tuple) and len(rhs.elts) > 0:
+            return
+
+        self.add_message("invalid-bases-assignment", node=node, confidence=HIGH)
```
clavedeluna, line 1322: I also assumed HIGH is the right confidence here, lmk if that's not right.

clavedeluna:
Maintainers:  are functional tests sufficient for new checkers, or should we also always add unit tests?

doc/data/messages/i/invalid-bases-assignment/good.py
```diff
@@ -0,0 +1,4 @@
+class Apple:
+    pass
+
+Apple.__bases__ = ("green", "red", )  # [invalid-bases-assignment]
```
Pierre-Sassoulas, line 4:
```suggestion
Apple.__bases__ = ("green", "red", )
```

DanielNoord (changes requested):

tests/functional/i/invalid/invalid_bases_assignment.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+Apple.__bases__ = ()  # [invalid-bases-assignment]
+Apple.__bases__ = "lala"  # [invalid-bases-assignment]
+Apple.__bases__ = True  # [invalid-bases-assignment]
+Apple.__bases__ = (1,)
+Apple.__bases__ = ("red", "green")
+
+bases = ("orange", "yellow")
+Apple.__bases__ = bases  # [invalid-bases-assignment]
```
DanielNoord, line 16: Why is this invalid?

tests/functional/i/invalid/invalid_bases_assignment.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Apple:
+    pass
+
+
+Apple.__bases__ = ()  # [invalid-bases-assignment]
+Apple.__bases__ = "lala"  # [invalid-bases-assignment]
+Apple.__bases__ = True  # [invalid-bases-assignment]
+Apple.__bases__ = (1,)
```
mbyrnepr2, line 12: This one is a `TypeError`; perhaps we should make the [current definition](https://github.com/PyCQA/pylint/pull/7615/files#diff-9b76776694be02a09ca25a56adee0b3087ad5b86dfafd318923c4bf008c131c3R404) stricter.

tests/functional/i/invalid/invalid_bases_assignment.py
clavedeluna: HA as I wrote this I inverted what's allowed for this check in my mind! Of course this should've passed. Will take a look.

tests/functional/i/invalid/invalid_bases_assignment.py
mbyrnepr2: [Python docs](https://docs.python.org/3/c-api/object.html?highlight=__bases__) have it as _a tuple of base classes_

clavedeluna: good catch!

tests/functional/i/invalid/invalid_bases_assignment.py
clavedeluna: ah so turns out this is correct, both for the assigning to `x` and straight without assignment. it's because `Apple.__bases__ = ("red", "green")` gets a `TypeError: Apple.__bases__ must be tuple of classes, not 'str'` in the REPL. Also mentioned in a comment above  for ints. I'll update the definition for this checker to be more strict.

clavedeluna:
This is trickier than we thought!

```
>>> class Fruit:
...     pass
...
>>> class Apple:
...     pass
...
>>> Apple.__bases__ = (Fruit, )
TypeError: __bases__ assignment: 'Fruit' deallocator differs from 'object'
```

Reading about, including https://stackoverflow.com/questions/3193158/bases-doesnt-work-whats-next

tests/functional/i/invalid/invalid_bases_assignment.py
DanielNoord: Oke, but:
```python
bases = (int, str)
Apple__bases__ = bases
```

That works right?

Not to discourage you, but this is shaping up to be a lot more difficult than we might have originally envisioned.

clavedeluna: Huh yeah that does work, didn't expect it. Let me summarize the findings and see how we can limit the scope here, if at all.

clavedeluna:
@DanielNoord @Pierre-Sassoulas , maintainers et all:

Turns out this check is a bit more complicated than just "check if __bases__ is assigned to anything but a tuple". There are cases which also fail the interpreter that have to do way more with inheritance which I'm not sure is going to be easy to detect.

This is an example that will fail the interpreter
```
>>> class Tree:
...     pass
...
>>> class Shrub:
...     pass
...
>>> Shrub.__bases__
(<class 'object'>,)
>>> Shrub.__bases__ = (Tree,)
TypeError: __bases__ assignment: 'Tree' deallocator differs from 'object'
```

However, these example will pass:

```
class Tree:
    pass

class Apple(Tree):
    pass

class Fruit(Tree):
    pass

Apple.__bases__ = (Fruit,)
Apple.__bases__ = (Tree, )
```

So in the Shrub first case, while we're assigning to a `tuple of classes` as the python doc states, these is an object allocation and inheritance issue.

To limit scope for this first iteration of this check, is it ok to implement the "naive" check of as long as __bases__ is being assigned to a tuple of classes, then good? This means pylint would for now miss the Shrub example.

DanielNoord:
I'm not sure a "naive" check works here. I think it's very likely to create false positives here. The primer won't be able to give us much security as I don't think this is a very common pattern, which also doesn't help.

Personally I would prefer not to introduce such a difficult check if we haven't covered everything. We should probably investigate and come up with a good set of tests and work with those. `astroid` should have everything we need to check `mro's` etc.

clavedeluna:
> I'm not sure a "naive" check works here. I think it's very likely to create false positives here. The primer won't be able to give us much security as I don't think this is a very common pattern, which also doesn't help.
> 
> Personally I would prefer not to introduce such a difficult check if we haven't covered everything. We should probably investigate and come up with a good set of tests and work with those. `astroid` should have everything we need to check `mro's` etc.

I agree with the sentiment that maybe adding a naive checker may not make sense, but only if we have examples of false positives. Given this is a very rare (hopefully!) pattern, I can't think of any false positives. If anything, like I mentioned above with the MRO issue, we would have _false negatives_.

Pierre-Sassoulas:
Never used this pattern myself. How well does mypy or pyright handle this ? The easiest move might be not to play if they already do it well. I'm more and more wary of adding typing checks when other wildly adopted tools already do the jobs generically and easily.

clavedeluna:
Alright, if you run this code

```
# pylint: disable=missing-module-docstring, missing-class-docstring


class Tree:
    pass

class Shrub:
    pass

class Apple(Tree):
    pass

class Fruit(Tree):
    pass

Apple.__bases__ = (Fruit,)
Apple.__bases__ = (Tree, )

Tree.__bases__ = (Apple, )  # [invalid-bases-assignment]

Apple.__bases__ = ()  # [invalid-bases-assignment]
Apple.__bases__ = "lala"  # [invalid-bases-assignment]
Apple.__bases__ = True  # [invalid-bases-assignment]
Apple.__bases__ = (1,)  # [invalid-bases-assignment]
Apple.__bases__ = ("red", "green")  # [invalid-bases-assignment]

bases = ("orange", "yellow")
Apple.__bases__ = bases  # [invalid-bases-assignment]

Apple.__bases__ = Apple  # [invalid-bases-assignment]
Apple.__bases__ = (Apple, )  # [invalid-bases-assignment]


class EverGreen:
    pass

bases = (int, str)
EverGreen__bases__ = bases
```

against `mypy`, the results are interesting

```
23: error: Incompatible types in assignment (expression has type "str", variable has type "Tuple[type, ...]")  [assignment]
24: error: Incompatible types in assignment (expression has type "bool", variable has type "Tuple[type, ...]")  [assignment]
25: error: Incompatible types in assignment (expression has type "Tuple[int]", variable has type "Tuple[type, ...]")  [assignment]
26: error: Incompatible types in assignment (expression has type "Tuple[str, str]", variable has type "Tuple[type, ...]")  [assignment]
29: error: Incompatible types in assignment (expression has type "Tuple[str, str]", variable has type "Tuple[type, ...]")  [assignment]
31: error: Incompatible types in assignment (expression has type "Type[Apple]", variable has type "Tuple[type, ...]")  [assignment]
38: error: Incompatible types in assignment (expression has type "Tuple[Type[int], Type[str]]", variable has type "Tuple[str, str]")  [assignment]
```

this is telling me that mypy is actually doing a "naive" check as well. If it were a smart check against MRO, it should've detected
`Tree.__bases__ = (Apple, )  # [invalid-bases-assignment]`
or even easier
`Apple.__bases__ = (Apple, )  # [invalid-bases-assignment]`

because while this is assigning to a tuple of classes, it breaks inheritance.

@Pierre-Sassoulas @DanielNoord as maintainers could you decide what to do here?
1. do a naive approach as appears to be done in mypy (only check against tuple of classes, nothing else regarding inheritance)
2. close this PR, update issue (close it?) explaining why we won't do it

Pierre-Sassoulas:
I'm mixed, as mypy is also doing it naively, it might be valuable to do it in pylint (maybe better later?). Maybe mypy is doing it really easily and we're going to have a hard time maintaining our version later. On the other a hand it's a shame to throw a check from an issue that has been opened a long time. In any case we should salvage the unrelated confidence change and refactor. Let's gather other feedback from maintainers.

DanielNoord:
If we can't figure out how to do the non-naive check we don't provide much else compared to `mypy`. So for me I'd say either we implement the full check or leave this to `mypy`.

PR #7617, opened 2022-10-13
Improve bad-threading-instantiation check

### LINKED ISSUE #7570: False negative for `bad-thread-instantiation`
### Bug description

```python
# module test.py

import threading


def thread_target(n):
    print(n ** 2)


thread = threading.Thread(thread_target, args=(10,))
thread.start()
```

I actually saw this example in [pylint-errors](https://github.com/vald-phoenix/pylint-errors/blob/c5a1b2c994f63eba19d4eccf2741827b281dd6ee/plerr/errors/stdlib/W1506.md) and expected it to work.

### Configuration

```ini
None
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
```

### Expected behavior

```
test.py:8:9: W1506: threading.Thread needs the target function (bad-thread-instantiation)
```

### Pylint version

```shell
pylint 2.16.0-dev
astroid 2.12.10
Python 3.8.9 (default, Apr 13 2022, 08:48:07)
```

### OS / Environment

Mac

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

kinda both?

## Description

`bad-threading-instantiation` relied on the fact that if two args were passed, the user knew the second was the `target` func, but this is naive to assume. Instead, this PR improves this check by making it more strict - users should always identify the target func with the `target=` kwarg.

Closes #7570

### DIFF
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -378,10 +378,10 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             "should be emitted.",
         ),
         "W1506": (
-            "threading.Thread needs the target function passed as a kwarg",
+            "threading.Thread needs the target function",
             "bad-thread-instantiation",
             "The warning is emitted when a threading.Thread class "
-            "is instantiated without the target function being passed as a kwarg. "
+            "is instantiated without the target function being passed as a kwarg or as a second argument. "
             "By default, the first parameter is the group param, not the target param.",
         ),
         "W1507": (
@@ -488,7 +488,11 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
         # synced with the config argument deprecated-modules
 
     def _check_bad_thread_instantiation(self, node: nodes.Call) -> None:
-        if not node.kwargs and "target" not in {key.arg for key in node.keywords}:
+        if (
+            len(node.args) < 2
+            and not node.kwargs
+            and "target" not in {key.arg for key in node.keywords}
+        ):
             self.add_message(
                 "bad-thread-instantiation", node=node, confidence=interfaces.HIGH
             )
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -378,10 +378,10 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
             "should be emitted.",
         ),
         "W1506": (
-            "threading.Thread needs the target function",
+            "threading.Thread needs the target function passed as a kwarg",
             "bad-thread-instantiation",
             "The warning is emitted when a threading.Thread class "
-            "is instantiated without the target function being passed. "
+            "is instantiated without the target function being passed as a kwarg. "
             "By default, the first parameter is the group param, not the target param.",
         ),
         "W1507": (
@@ -488,8 +488,10 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
         # synced with the config argument deprecated-modules
 
     def _check_bad_thread_instantiation(self, node: nodes.Call) -> None:
-        if not node.kwargs and not node.keywords and len(node.args) <= 1:
-            self.add_message("bad-thread-instantiation", node=node)
+        if not node.kwargs and "target" not in {key.arg for key in node.keywords}:
+            self.add_message(
+                "bad-thread-instantiation", node=node, confidence=interfaces.HIGH
+            )
 
     def _check_for_preexec_fn_in_popen(self, node: nodes.Call) -> None:
         if node.keywords:
--- a/pylint/checkers/stdlib.py
+++ b/pylint/checkers/stdlib.py
@@ -488,11 +488,11 @@ class StdlibChecker(DeprecatedMixin, BaseChecker):
         # synced with the config argument deprecated-modules
 
     def _check_bad_thread_instantiation(self, node: nodes.Call) -> None:
-        if (
-            len(node.args) < 2
-            and not node.kwargs
-            and "target" not in {key.arg for key in node.keywords}
-        ):
+        func_kwargs = {key.arg for key in node.keywords}
+        if "target" in func_kwargs:
+            return
+
+        if len(node.args) < 2 and (not node.kwargs or "target" not in func_kwargs):
             self.add_message(
                 "bad-thread-instantiation", node=node, confidence=interfaces.HIGH
             )
--- a/tests/functional/b/bad_thread_instantiation.py
+++ b/tests/functional/b/bad_thread_instantiation.py
@@ -1,4 +1,4 @@
-# pylint: disable=missing-docstring, redundant-keyword-arg, invalid-name
+# pylint: disable=missing-docstring, redundant-keyword-arg, invalid-name, line-too-long
 import threading
 
 
@@ -16,3 +16,9 @@ def thread_target(n):
 
 
 thread = threading.Thread(thread_target, args=(10,))  # [bad-thread-instantiation]
+
+
+kw = {'target_typo': lambda x: x}
+threading.Thread(None, **kw)  # [unexpected-keyword-arg, bad-thread-instantiation]
+
+threading.Thread(None, target_typo=lambda x: x)  # [unexpected-keyword-arg, bad-thread-instantiation]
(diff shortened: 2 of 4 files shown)


pylint/checkers/stdlib.py
```diff
@@ -466,8 +466,10 @@ def __init__(self, linter: PyLinter) -> None:
         # synced with the config argument deprecated-modules
 
     def _check_bad_thread_instantiation(self, node: nodes.Call) -> None:
-        if not node.kwargs and not node.keywords and len(node.args) <= 1:
-            self.add_message("bad-thread-instantiation", node=node)
+        if not node.kwargs and "target" not in {key.arg for key in node.keywords}:
+            self.add_message(
+                "bad-thread-instantiation", node=node, confidence=interfaces.HIGH
+            )
```
clavedeluna, line 472: seems HIGH to be the right confidence here, no?

clavedeluna:
> 🤖 **Effect of this PR on checked open source code:** 🤖
> 
> **Effect on [django](https://github.com/django/django):** The following messages are now emitted:
> 
> **Effect on [pandas](https://github.com/pandas-dev/pandas):** The following messages are now emitted:
> 
> The following messages are no longer emitted:
> 
> **Effect on [sentry](https://github.com/getsentry/sentry):** The following messages are no longer emitted:
> 
> _This comment was generated for commit [4bf6d8f](https://github.com/PyCQA/pylint/commit/4bf6d8f78adef1557a967e2daee3903d0853b0fb)_

this output has to be a fluke right? doesn't seem related to this PR at all.

clavedeluna:
@Pierre-Sassoulas not sure why it didn't auto-add you as a reviewer

Pierre-Sassoulas:
> this output has to be a fluke right? doesn't seem related to this PR at all.

You probably need to rebase on main to fix it, but it's hard to know for sure.

pylint/checkers/stdlib.py
Pierre-Sassoulas: I don't see any inference so I'd say yes.

tests/functional/b/bad_thread_instantiation.py
```diff
@@ -1,8 +1,10 @@
-# pylint: disable=missing-docstring
+# pylint: disable=missing-docstring, redundant-keyword-arg
 import threading
 
 
 threading.Thread(lambda: None).run()  # [bad-thread-instantiation]
-threading.Thread(None, lambda: None)
+threading.Thread(None, lambda: None)  # [bad-thread-instantiation]
```
Pierre-Sassoulas, line 6: This wasn't raising previously, probably voluntarily as the test existed and there's no comment around it, did you check the reasoning in the original implementation?

clavedeluna: The reason this now raises is 💯 to do with this new PR. If any call to thread.Thread does NOT use `target=` as a kwarg, it should raise the warning, so this change makes perfect sense.  Previous impl didn't desire to do that, so it doesn't mention anything

Pierre-Sassoulas (commented):
Sorry I don't know about threading to be able to review correctly in a reasonable time for me at this time. Maybe later, maybe someone else :)

clavedeluna:
I can't update the list of reviewers so adding @DanielNoord in case you want to review.

tests/functional/b/bad_thread_instantiation.py
DanielNoord: I don't think this is a good change and might annoy users. Arguments are parsed in order so supplying `target` as the second argument is perfectly valid. Why would we change that?

_This is similar as requiring `maxsplit` in `str.split()` to always be passed as a kwarg. It seems unnecessarily pedantic._

clavedeluna: This change definitely dances the line of false positive/negatives. The difference in `threading.Thread` v. something like `split` is how likely we think users are to confuse the first and second arg. As mentioned in the current docs `By default, the first parameter is the group param, not the target param.`. So users tend to think that the first argument is the target func, but it isn't.

Happy to hear counter test cases which would solve the original false negative posed in the issue
`thread = threading.Thread(thread_target, args=(10,))`

DanielNoord: Why can't we catch the example from the issue but also accept `target` without they keyword?

clavedeluna: I'm not sure how we would catch the example from the issue otherwise??

Also! Just re-read the [docs](https://docs.python.org/3/library/threading.html#threading.Thread) check this out:
`This constructor should always be called with keyword arguments. Arguments are:`

seems to me that actually gives pylint a strong reason to warn users if they don't use kwargs

DanielNoord: hm, that is indeed interesting. But then shouldn't we warn for other arguments as well?

clavedeluna: Sure! Reading that now, it would seem to be appropriate. Want me to add it to this PR? Separate issue to discuss?

DanielNoord: Yeah I'm not sure. Changing the meaning of the warning seems to be suboptimal.

DanielNoord: @Pierre-Sassoulas do you have an opinion about this?

Pierre-Sassoulas: Sorry I don't have an opinion right now, I'd need to educate myself about threading first :)

jacobtylerwalls: The comment in the python docs is very old and seems to be applicable to all kinds of methods, not just the case at hand. So like @DanielNoord I'm reluctant to change the meaning of this message to start enforcing passing arguments by keyword, even though it's good practice.

What we should still be able to do in this PR is just check that there is either a `target` keyword arg or a second positional argument.

> I'm not sure how we would catch the example from the issue otherwise??

Hopefully it's trivial, that example had one positional arg, not two, and a kwarg other than `target`.

clavedeluna: Thanks for the helpful and specific feedback @jacobtylerwalls ! I think I can implement exactly that.

pylint/checkers/stdlib.py
```diff
@@ -488,8 +488,14 @@ def __init__(self, linter: PyLinter) -> None:
         # synced with the config argument deprecated-modules
 
     def _check_bad_thread_instantiation(self, node: nodes.Call) -> None:
-        if not node.kwargs and not node.keywords and len(node.args) <= 1:
-            self.add_message("bad-thread-instantiation", node=node)
+        if (
+            len(node.args) < 2
+            and not node.kwargs
```
jacobtylerwalls, line 493: Looks like a false negative here if `target` is omitted from the kwargs, like this:
```python
import threading

kw = {'target_typo': lambda x: x}
threading.Thread(None, **kw)
```
Would be good to add this as a test case.

jacobtylerwalls (changes requested):

jacobtylerwalls:
Thanks. You know, I just remembered there's a util that's supposed to deal with this complexity for you (`get_argument_from_call()`), but it has the same flaw that it doesn't check **kwargs. I'll open an issue. That would be an fun/straightforward PR fixing false negatives all over.

tests/functional/b/bad_thread_instantiation.py
```diff
@@ -1,8 +1,24 @@
...
 import threading
 
 
 threading.Thread(lambda: None).run()  # [bad-thread-instantiation]
 threading.Thread(None, lambda: None)
+threading.Thread(lambda: None, group=None)  # [bad-thread-instantiation]
+threading.Thread()  # [bad-thread-instantiation]
+
 threading.Thread(group=None, target=lambda: None).run()
-threading.Thread() # [bad-thread-instantiation]
```
jacobtylerwalls, line 8: Why was this case removed?

clavedeluna: Not removed, just moved up a few lines.

jacobtylerwalls (approved):
Thanks!

PR #7618, opened 2022-10-13
Replace deprecated set-output commands [ci]

cdce8p:
## Description
The Github actions `::set-output` command was deprecated and replaced with environment files.

Refs:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter


Pierre-Sassoulas (approved):
LGTM, thank you Marc.

PR #7620, opened 2022-10-13
fix: reporters receive copy of message

### LINKED ISSUE #7214: Colorized output should not modify `msg` objects
### Bug description

I noticed this problem because I use colorized text output alongside with the [pylint-gitlab](https://pypi.org/project/pylint-gitlab/) plugin. I filed [an issue there](https://gitlab.com/smueller18/pylint-gitlab/-/issues/11) as well, but I agree with [the author](https://gitlab.com/smueller18) of the plugin that the bug should be fixed within PyLint.

I use the pylint-gitlab reporter and colorized text output in the same run of pylint (see below). The plugin receives the messages modified by the `ColorizedTextReporter`. This causes an exception (`KeyError`) because the plugin receives the ANSI-color-coded category `\x1b[1mconvention\x1b[0m` instead of simply `convention`.

The modification happens inside the [`handle_message` method](https://github.com/PyCQA/pylint/blob/815d417e4776a7bbbdc87fff7d4501d2cca931be/pylint/reporters/text.py#L304-L307):
```
        msg.msg = colorize_ansi(msg.msg, msg_style)
        msg.symbol = colorize_ansi(msg.symbol, msg_style)
        msg.category = colorize_ansi(msg.category, msg_style)
        msg.C = colorize_ansi(msg.C, msg_style)
```

Suggested solution:
---------------------
As I haven't found a way of controlling the order in which the reporters are processed (the behavior would not be a problem if one could ensure that the `ColorizedTextReporter` is called last), I would suggest that the reporter should be changed to work in a re-entrant manner: The `ColorizedTextReporter` should not modify the original fields. It should rather add some extra fields (e.g. `colorized_category`) or keep a list of copies of the original `msg` objects.

### Configuration

_No response_

### Command used

```shell
python -m pylint --exit-zero --output-format=pylint_gitlab.GitlabCodeClimateReporter:pylint-codeclimate.json,pylint_gitlab.GitlabPagesHtmlReporter:index.html,text:pylint.txt,colorized src
```

### Pylint output

```shell
File "xxx\python-3.8.9\lib\runpy.py", line 194, in _run_module_as_main
  File "xxx\python-3.8.9\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "xxx\.nox\lint_pylint\lib\site-packages\pylint\__main__.py", line 10, in <module>
    pylint.run_pylint()
  File "xxx\.nox\lint_pylint\lib\site-packages\pylint\__init__.py", line 25, in run_pylint
  File "xxx\.nox\lint_pylint\lib\site-packages\pylint\lint\run.py", line 205, in __init__
    score_value = linter.generate_reports()
  File "xxx\.nox\lint_pylint\lib\site-packages\pylint\lint\pylinter.py", line 998, in generate_reports
    self.reporter.display_messages(report_nodes.Section())
  File "xxx\.nox\lint_pylint\lib\site-packages\pylint\reporters\multi_reporter.py", line 95, in display_messages
    rep.display_messages(layout)
  File "xxx\.nox\lint_pylint\lib\site-packages\pylint_gitlab\reporter.py", line 69, in display_messages
    "class": self._MSG_TYPES[message.category],
KeyError: '\x1b[1mconvention\x1b[0m'
```

### Expected behavior

ANSI-colored output to the screen plus some generated files (`pylint-codeclimate.json`, `index.html`, `pylint.txt`)

### Pylint version

```shell
pylint-2.14.x (the problem did not occur with pylint<=3.13.x)
```

### OS / Environment

Tested on Linux and Windows

### Additional dependencies

pylint-gitlab=1.0.0

Smixi:
## Type of Changes

## Description
MultiReporter was sending message by reference, which then could be changed by reporter. The colorized reporter was one of them, leading to changing any other reporters. This was broken since 2.14.0 and affected mostly people using colorized with a custom reporter.

Closes #7214

### DIFF
--- a/doc/whatsnew/fragments/7214.bugfix
+++ b/doc/whatsnew/fragments/7214.bugfix
@@ -1,0 +1,3 @@
+Message send to reporter are now copied so a reporter cannot modify the message sent to other reporters.
+
+Refs #7214
--- a/doc/whatsnew/fragments/7214.bugfix
+++ b/doc/whatsnew/fragments/7214.bugfix
@@ -1,3 +1,3 @@
 Message send to reporter are now copied so a reporter cannot modify the message sent to other reporters.
 
-Refs #7214
+Closes #7214
--- a/pylint/reporters/multi_reporter.py
+++ b/pylint/reporters/multi_reporter.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import os
 from collections.abc import Callable
+from copy import copy
 from typing import TYPE_CHECKING, TextIO
 
 from pylint.message import Message
@@ -77,7 +78,10 @@ class MultiReporter:
     def handle_message(self, msg: Message) -> None:
         """Handle a new message triggered on the current file."""
         for rep in self._sub_reporters:
-            rep.handle_message(msg)
+            msg_copy = copy(
+                msg
+            )  # Provide a copy so reporters can't modify message for others.
+            rep.handle_message(msg_copy)
 
     def writeln(self, string: str = "") -> None:
         """Write a line in the output buffer."""
--- a/tests/reporters/unittest_reporting.py
+++ b/tests/reporters/unittest_reporting.py
@@ -353,7 +353,9 @@ def test_multi_reporter_independant_messages() -> None:
 
     class ReporterCheck(BaseReporter):
         def handle_message(self, msg: Message) -> None:
-            assert msg.msg == check_message
+            assert (
+                msg.msg == check_message
+            ), "Message object should not be changed by other reporters."
 
         def writeln(self, string: str = "") -> None:
             pass
@@ -372,6 +374,10 @@ def test_multi_reporter_independant_messages() -> None:
     )
 
     multi_reporter.handle_message(message)
+
+    assert (
+        message.msg == check_message
+    ), "Message object should not be changed by reporters."
 
 
 def test_display_results_is_renamed() -> None:
--- a/tests/reporters/unittest_reporting.py
+++ b/tests/reporters/unittest_reporting.py
@@ -18,10 +18,12 @@ import pytest
 from _pytest.recwarn import WarningsRecorder
 
 from pylint import checkers
+from pylint.interfaces import HIGH
 from pylint.lint import PyLinter
+from pylint.message.message import Message
 from pylint.reporters import BaseReporter, MultiReporter
 from pylint.reporters.text import ParseableTextReporter, TextReporter
-from pylint.typing import FileItem
+from pylint.typing import FileItem, MessageLocationTuple
 
 if TYPE_CHECKING:
     from pylint.reporters.ureports.nodes import Section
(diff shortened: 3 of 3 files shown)


doc/whatsnew/fragments/7214.bugfix
```diff
@@ -0,0 +1,3 @@
+Message send to reporter are now copied so a reporter cannot modify the message sent to other reporters.
+
+Refs #7214
```
clavedeluna, line 3: should this be `Closes`?

Smixi: I wasn't sure of semantics of this, but yes ! Let me fix that.

pylint/reporters/multi_reporter.py
```diff
@@ -77,7 +78,10 @@ def linter(self, value: PyLinter) -> None:
     def handle_message(self, msg: Message) -> None:
         """Handle a new message triggered on the current file."""
         for rep in self._sub_reporters:
-            rep.handle_message(msg)
+            msg_copy = copy(
+                msg
+            )  # Provide a copy so reporters can't modify message for others.
+            rep.handle_message(msg_copy)
```
Pierre-Sassoulas, line 84:
```suggestion
            # We provide a copy so reporters can't modify message for others.
            rep.handle_message(copy(msg))
```

Pierre-Sassoulas (commented):
Thank you for the investigation. I wonder what are the performance implication and also if the side effect of being able to modify other reporter's message wasn't seen as a feature by some users. We'd need to do it as a breaking change if that's the case.

tests/reporters/unittest_reporting.py
```diff
@@ -334,6 +336,44 @@ def test_multi_format_output(tmp_path: Path) -> None:
...
+    message = Message(
+        symbol="missing-docstring",
+        msg_id="C0123",
+        location=MessageLocationTuple("abspath", "path", "module", "obj", 1, 2, 1, 3),
+        msg=check_message,
+        confidence=HIGH,
+    )
+
+    multi_reporter.handle_message(message)
+
```
Pierre-Sassoulas, line 375: It seems the assert is missing ? Something like this maybe:
```suggestion

assert multi_reporter.msg = check_message 
```

Pierre-Sassoulas (changes requested):

tests/reporters/unittest_reporting.py
Smixi: I putted the assert in the  `CheckReporter` to ensure he was not receiving a modified msg. Maybe it is more clear if the check is outside?

Pierre-Sassoulas: Ha I missed that. Yeah it's better outside 👍

Smixi: @Pierre-Sassoulas, I added the assert in the test but also keep the one with the reporter. This ensure each reporter is independant, but also that the handle_message in multi reporter does not change the message itself (if in the future this function is modified).

Pierre-Sassoulas (approved):

tests/reporters/unittest_reporting.py
Pierre-Sassoulas: Thank you !

DanielNoord:
We could also change the `handle_message` in the colour reporter to create a copy. That way reporters that depend on being able to modify for others aren't broken?

Smixi:
@

> We could also change the `handle_message` in the colour reporter to create a copy. That way reporters that depend on being able to modify for others aren't broken?

Do you have an usecase example ? Maybe that would be another option ? One issue is that we would need solve is to specify ordering.  From 2.13.9 to 2.14.0, coloriezd/text reporter is now always the first reporter. If two reporters are necessary to achieve a specific output, maybe it should be only one reporter with options ?

DanielNoord:
Yeah you're right. If anybody complains about this and has a usecase we can always update to allow it again.

PR #7621, opened 2022-10-14
Fix a false negative for ``unused-import``

### LINKED ISSUE #4150: Type hints persist module usage between independent modules (unused-import)
When a module is used in a type hint in a certain way it will count as using the module in all source files analyzed afterwards. This leads to a false negative for `unused-import` depending on the order that sources were scanned.

### Steps to reproduce

Given two files, `a.py`:

```python
"x"
import uuid
from typing import Optional
ID = None  # type: Optional[uuid.UUID]
```

And `b.py`:

```python
"x"
import uuid
```

### Current behavior

When analyzing `a.py` then `b.py` it misses the `unused-import`:

```
$ pylint a.py b.py

----------------------------------------------------------------------
```

But swap the order and `unused-import` appears:

```
$ pylint b.py a.py
************* Module b
b.py:2:0: W0611: Unused import uuid (unused-import)

Your code has been rated at 7.50/10
```

### Expected behavior

The outputs should be identical and both report `unused-import`.

### pylint --version output

```
pylint 2.7.1
astroid 2.5
Python 3.7.3 (default, Jul 25 2020, 13:03:44) 
[GCC 8.3.0]
```

mbyrnepr2:
Fix a false negative for ``unused-import`` when linting multiple modules & one module uses an import in a type annotation & the same import is unused in a subsequent module.

## Type of Changes

## Description

Closes #4150


doc/whatsnew/fragments/4150.false_negative
```diff
@@ -0,0 +1,3 @@
+Fix a false negative for ``unused-import`` when linting multiple modules & one module uses an import in a type annotation & the same import is unused in a subsequent module.
```
Pierre-Sassoulas, line 1:
```suggestion
Fix a false negative for ``unused-import`` when one module used an import in a type annotation that was also used in another module.
```

pylint/checkers/variables.py
```diff
@@ -1147,6 +1147,7 @@ def leave_module(self, node: nodes.Module) -> None:
             return
 
         self._check_imports(not_consumed)
+        self._type_annotation_names = []
```
Pierre-Sassoulas, line 1150: The fix itself is really simple 😄

Pierre-Sassoulas (approved):
Great fix !

Pierre-Sassoulas:
Not sure about what to do regarding the release we're aiming. It's a bug but at the same time it was a false negative. Maybe 2.16.0 so we do not raise new messages in a patch. If there's an oversight on our side we don't want cascading patches backport ?

mbyrnepr2:
> Not sure about what to do regarding the release we're aiming. It's a bug but at the same time it was a false negative. Maybe 2.16.0 so we do not raise new messages in a patch. If there's an oversight on our side we don't want cascading patches backport ?

Agreed. Happy to wait until 2.16.0.

mbyrnepr2:
About the primer [comment](https://github.com/PyCQA/pylint/pull/7621#issuecomment-1278650827). 

I notice that the output for Django, Pandas & Sentry is similar to some other open MRs; it doesn't look like it is caused by this MR.
The black output looks relevant to this MR but, at the same time, it is saying that a new message is emitted while I also see that message emitted on main.

Pierre-Sassoulas:
In black I think ``os`` is used here : https://github.com/psf/black/blob/f16333e78ba77ab29ab0b75e10891e6c65c6da7e/src/blib2to3/pgen2/pgen.py#L25. Feel like a false positive.

mbyrnepr2:
> In black I think `os` is used here : https://github.com/psf/black/blob/f16333e78ba77ab29ab0b75e10891e6c65c6da7e/src/blib2to3/pgen2/pgen.py#L25. Feel like a false positive.

I agree with you & I'm happy to dig into that one; but this false positive isn't created by the current MR.

The sample from the `black` primer will fail on `main` also:
```python
import os
from typing import Union
Path = Union[str, "os.PathLike[str]"]
```

Pierre-Sassoulas:
> I agree with you & I'm happy to dig into that one; but this false positive isn't created by the current MR.

Ha, something is wrong with the primer then, sorry.

jacobtylerwalls (approved):

Pierre-Sassoulas (changes requested):
@mbyrnepr2 can you merge or rebase on main so we launch the new python 3.11 tests please ?

Pierre-Sassoulas (approved):

DanielNoord:
@mbyrnepr2 The message in `black` seems like a false positive?

mbyrnepr2:
It does seem that way, I agree @DanielNoord but we can also see the same happening on the main branch (before it was merged). Perhaps an issue with the primer?

DanielNoord:
Probably! Went through my notifications of last week and noticed this message, but `pytest` is also broken in that comment so could very well be a fluke.

PR #5227, opened 2021-10-28
Fix incorrect ``consider-using-ternary`` when condition is inferable as False

areveny:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #5200

PR #7627, opened 2022-10-16
Fix a false positive for `simplify-boolean-expression` when multiple values inferred

### LINKED ISSUE #7626: False positive ``simplify-boolean-expression`` when boolean variables are modified after their intialization in a for loop
### Bug description

For the following code:

```python

def some_function() -> bool:
    '''a function

    Returns:
        True
    '''
    flag_a = False
    flag_b = False
    for word in ['alfa', 'bravo']:
        if 'b' not in word:
            flag_a = True
        else:
            flag_b = True
    multiple = (flag_a and flag_b) or returns_false()  # <- error here
    return multiple

def returns_false() -> bool:
    '''a different function

    Returns:
        False, but pylint probably doesn't know that
    '''
    return False


def main() -> None:
    '''entry point
    '''
    print(repr(some_function()))  # True

if __name__ == '__main__':
    main()
```

...pylint emits:
```
R1709: Boolean expression may be simplified to returns_false() (simplify-boolean-expression)
```

This statement is incorrect: `some_function()` always returns `True`.

Note: removing ...`or returns_false()` causes the error to disappear.

### Configuration

_No response_

### Command used

```shell
pylint --rcfile=/dev/null thecodeabove.py
```

### Pylint output

```shell
************* Module thecodeabove
thecodeabove.py:17:4: R1709: Boolean expression may be simplified to returns_false() (simplify-boolean-expression)

Your code has been rated at 9.38/10 (previous run: 9.38/10, +0.00)
```

### Expected behavior

Pylint should not emit R1709/simplify-boolean-expression, since it is mistaken.

### Pylint version

```shell
pylint 2.15.4
astroid 2.12.11
Python 3.8.12 (default, Jan 17 2022, 13:29:00)
[Clang 10.0.1 (clang-1001.0.46.4)]
```

### OS / Environment

macOS 10.14

### Additional dependencies

_No response_

jacobtylerwalls:
## Type of Changes

## Description
If multiple values are possible for a constant, we should silence the refactoring messages that claim to offer simplifications, since the inferred value is not reliable.

Adds a keyword-only `compare_constants` argument to `safe_infer` to accomplish this in a reusable way.

Closes #7626

I'm wondering if @areveny you might wish to have a look, as I saw #5227? No pressure.

### DIFF
--- a/doc/whatsnew/fragments/7626.false_positive
+++ b/doc/whatsnew/fragments/7626.false_positive
@@ -1,0 +1,6 @@
+Fix a false positive for ``simplify-boolean-expression`` when multiple values
+are inferred for a constant.
+
+Adds a keyword-only ``compare_constants`` argument to ``safe_infer``.
+
+Closes #7626
--- a/doc/whatsnew/fragments/7626.false_positive
+++ b/doc/whatsnew/fragments/7626.false_positive
@@ -1,6 +1,4 @@
 Fix a false positive for ``simplify-boolean-expression`` when multiple values
 are inferred for a constant.
 
-Adds a keyword-only ``compare_constants`` argument to ``safe_infer``.
-
 Closes #7626
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1368,10 +1368,6 @@ def safe_infer(
     if value is not astroid.Uninferable:
         inferred_types.add(_get_python_type_of_node(value))
 
-    first_constant_value = None
-    if compare_constants and isinstance(value, nodes.Const):
-        first_constant_value = value.value
-
     try:
         for inferred in infer_gen:
             inferred_type = _get_python_type_of_node(inferred)
@@ -1380,7 +1376,8 @@ def safe_infer(
             if (
                 compare_constants
                 and isinstance(inferred, nodes.Const)
-                and inferred.value != first_constant_value
+                and isinstance(value, nodes.Const)
+                and inferred.value != value.value
             ):
                 return None
             if (
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1343,12 +1343,18 @@ def _get_python_type_of_node(node: nodes.NodeNG) -> str | None:
 
 @lru_cache(maxsize=1024)
 def safe_infer(
-    node: nodes.NodeNG, context: InferenceContext | None = None
+    node: nodes.NodeNG,
+    context: InferenceContext | None = None,
+    *,
+    compare_constants: bool = False,
 ) -> InferenceResult | None:
     """Return the inferred value for the given node.
 
     Return None if inference failed or if there is some ambiguity (more than
     one node has been inferred of different types).
+
+    If compare_constants is True and if multiple constants are inferred,
+    unequal inferred values are also considered ambiguous and return None.
     """
     inferred_types: set[str | None] = set()
     try:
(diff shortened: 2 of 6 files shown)


pylint/checkers/utils.py
```diff
@@ -1362,11 +1368,21 @@ def safe_infer(
...
         for inferred in infer_gen:
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
+            if (
+                compare_constants
+                and isinstance(inferred, nodes.Const)
+                and inferred.value != first_constant_value
+            ):
+                return None
```
areveny, line 1385:
```suggestion
            if (
                compare_constants
                and isinstance(inferred, nodes.Const)
                and isinstance(value, nodes.Const)
                and inferred.value != value.value
            ):
                return None
```

I think you could write it like this and remove the logic above since `value` isn't mutated.

areveny (approved):
Looks good to me! Left one suggestion. Requesting another approver.

I wonder if there's anywhere else `compare_constants=True` could be used but after perusing usages of `safe_infer` no other applications are immediately clear.

doc/whatsnew/fragments/7626.false_positive
```diff
@@ -0,0 +1,6 @@
+Fix a false positive for ``simplify-boolean-expression`` when multiple values
+are inferred for a constant.
+
+Adds a keyword-only ``compare_constants`` argument to ``safe_infer``.
```
Pierre-Sassoulas, line 4: This one is an internal change, but we might want to add two fragments as it's indeed valuable to know for plugin developers.
```suggestion
```

tests/functional/t/ternary.py
```diff
@@ -1,18 +1,21 @@
 """Test for old ternary constructs"""
-from UNINFERABLE import condition, true_value, false_value, some_callable  # pylint: disable=import-error
+from UNINFERABLE import condition, some_callable  # pylint: disable=import-error
```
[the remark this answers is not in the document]
jacobtylerwalls: The previous tests were wrong IMO, they caused the same false positive as reported in the issue. If something is uninferable, we should not assume it's truthy!

I agree I can make this more explicit with a test.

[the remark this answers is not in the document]
Pierre-Sassoulas: > If something is uninferable, we should not assume it's truthy!

Agree. Maybe we can keep only one test for this scenario instead of what we had before ?

Pierre-Sassoulas (dismissed):
We could re-add the old test for uninferable value and fix it, but I'll let you decide :)

jacobtylerwalls:
Ah, I think I follow you now. I've got versions of the original cases, plus two new cases.

Pierre-Sassoulas (approved):
👌

jacobtylerwalls:
Thanks for the reviews!

Pierre-Sassoulas:
This is going to be released in pylint 2.15.5 :)

pylint/checkers/utils.py
```diff
@@ -1367,6 +1373,13 @@ def safe_infer(
             inferred_type = _get_python_type_of_node(inferred)
             if inferred_type not in inferred_types:
                 return None  # If there is ambiguity on the inferred node.
+            if (
```
DanielNoord, line 1376: I'm not sure how the compiler handle this but since this is called so often we might want to "optimize" the if's here. If we use `if isinstance(inferred, ...)` and then go down a level we can use `elifs` to make sure we aren't doing `if` statement that we know will never be true.

Not big enough to create a follow-up but I guess something to consider for future PRs that touch this code.

Pierre-Sassoulas: Good point. I'm not sure we'll be able to keep that in mind if we don't do the follow-up immediately though.

jacobtylerwalls: > I'm not sure how the compiler handle this

`and` statements don't evaluate the right hand side if the left side tests false:
```python
>>> not 1 and (1 / 0)
False
```

Do you have a suggested optimization that would short-circuit more?

DanielNoord: Since it's merged I can't create an actual suggestion, but:
```python
isinstance(inferred, nodes.Const):
    if compare_constants ...
        ...
elif isinstance(inferred, nodes.FunctionDef):
    if inferred.args.args is not None:
      ...
```

jacobtylerwalls: Ah. Perhaps. But we would need a microbenchmark for how many times we call `compare_constants`, as it's probably way faster to check the boolean, right?

DanielNoord: Yeah, you're right!

PR #7629, opened 2022-10-16
Add `default_enabled` option to optional message dict

cdce8p:
## Description
Add a new `default_enabled` option to make it easier to disable a specific message by default.
This is especially useful for extensions like the `CodeStyle` one. I'm not so sure if we should use it for regular checks though. The expectation I believe is that all of them are enabled by default. I.e. to disable one, we should move it to an extension first.

### DIFF
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -481,7 +481,7 @@ class PyLinter(
         if hasattr(checker, "msgs"):
             self.msgs_store.register_messages_from_checker(checker)
             for message in checker.messages:
-                if message.default_disabled:
+                if not message.default_enabled:
                     self.disable(message.msgid)
         # Register the checker, but disable all of its messages.
         if not getattr(checker, "enabled", True):
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -480,6 +480,9 @@ class PyLinter(
             self.register_report(r_id, r_title, r_cb, checker)
         if hasattr(checker, "msgs"):
             self.msgs_store.register_messages_from_checker(checker)
+            for message in checker.messages:
+                if message.default_disabled:
+                    self.disable(message.msgid)
         # Register the checker, but disable all of its messages.
         if not getattr(checker, "enabled", True):
             self.disable(checker.name)
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -44,7 +44,7 @@ class MessageData(NamedTuple):
     checker_module_name: str
     checker_module_path: str
     shared: bool = False
-    default_disabled: bool = False
+    default_enabled: bool = True
 
 
 MessagesDict = Dict[str, List[MessageData]]
@@ -195,7 +195,7 @@ def _get_all_messages(
             checker_module.__name__,
             checker_module.__file__,
             message.shared,
-            message.default_disabled,
+            message.default_enabled,
         )
         msg_type = MSG_TYPES_DOC[message.msgid[0]]
         messages_dict[msg_type].append(message_data)
@@ -274,7 +274,7 @@ def _generate_single_message_body(message: MessageData) -> str:
 
 *{message.definition.description}*
 """
-    if message.default_disabled:
+    if not message.default_enabled:
         body += f"""
 .. caution::
   This message is disabled by default. To enable it, add ``{message.name}`` to the ``enable`` option.
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -44,6 +44,7 @@ class MessageData(NamedTuple):
     checker_module_name: str
     checker_module_path: str
     shared: bool = False
+    default_disabled: bool = False
 
 
 MessagesDict = Dict[str, List[MessageData]]
@@ -194,6 +195,7 @@ def _get_all_messages(
             checker_module.__name__,
             checker_module.__file__,
             message.shared,
+            message.default_disabled,
         )
         msg_type = MSG_TYPES_DOC[message.msgid[0]]
         messages_dict[msg_type].append(message_data)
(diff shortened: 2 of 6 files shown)


DanielNoord:
This feels like it fixes part of https://github.com/PyCQA/pylint/issues/3512 but not fully.

I'm -0,5 for adding any more complexity to enabling and disabling that doesn't immediately solve that issue. It just creates more to consider before we implement the system discussed there.

pylint/lint/pylinter.py
```diff
@@ -480,6 +480,9 @@ def register_checker(self, checker: checkers.BaseChecker) -> None:
             self.register_report(r_id, r_title, r_cb, checker)
         if hasattr(checker, "msgs"):
             self.msgs_store.register_messages_from_checker(checker)
+            for message in checker.messages:
+                if message.default_disabled:
+                    self.disable(message.msgid)
```
Pierre-Sassoulas, line 485: It feel like it's the responsibility of the Message Store to not register disabled messages in ``register_messages_from_checker``.

Pierre-Sassoulas (commented):
I think it's a good idea, we definitely need an easy way to make a check optional (beside moving code to another directory and adding or removing code during registration, that's barbaric 😄). We need a better way but plugin maintainer also need a better way imo.

I would use a positive formulation for the option name to avoid a double negative (disabled = False => enabled = True), maybe ``default_enabled`` or ``enabled_by_default`` ?

pylint/lint/pylinter.py
cdce8p: We want to register all messages. Otherwise the user isn't able to enable it via the config setting. It's the `MessageStateHandler` which deals with `enable`, `disable`, and `disable-next`.

cdce8p (commented):
> I think it's a good idea, we definitely need an easy way to make a check optional (beside moving code to another directory and adding or removing code during registration, that's barbaric 😄). We need a better way but plugin maintainer also need a better way imo.

I think we should be careful when considering if a check should be disabled by default. It's much easier for new ones and extensions. However, I do believe there is at least the expectation that all default checks are also enabled.

That doesn't negate that having an easy option to do so would be useful IMO.

> I would use a positive formulation for the option name to avoid a double negative (disabled = False => enabled = True), maybe `default_enabled` or `enabled_by_default` ?

Done

Pierre-Sassoulas:
> I think we should be careful when considering if a check should be disabled by default.

Sure, definitely something that will help for #3512 but should not change much after that.

pylint/lint/pylinter.py
cdce8p: > Hmm, right. It feels wrong that this code is in `PyLinter`,

I don't think so. A few lines below we also disable messages if the complete checker is disabled.
https://github.com/PyCQA/pylint/blob/77e8ae4f3902697babb186482f577a2c8b1a3725/pylint/lint/pylinter.py#L483-L485

> ... it should probably happen somewhere in `pylint.message`. Do you think moving that to MessageStateHandler is possible ?

Not without some bigger refactoring. `MessageStateHandler` is part of the `pylint.lint` package and not related to `pylint.messages`. That's why `PyLinter` is the best place IMO. It's the connection between both.

Pierre-Sassoulas: > PyLinter is the best place IMO. It's the connection between both.

It's the connection between everything and is still way too huge even after Daniel most welcome refactors to separate the configuration parsing. But yeah, if it's a huge refactor then let's not do it right now.

Pierre-Sassoulas (approved):

DanielNoord:
But don't want we want to plan out that PR and then work based on that design? We're not sure how this would interact with the extending/disabling that is proposed in that issue.

Pierre-Sassoulas:
I'm pretty sure a boolean flag on each message is the simplest / most extensible / scalable way to disable / enable messages. Actually I've wanted to do that for a long time, (thank you @cdce8p). But I'm open to another design if you have one Daniel

Pierre-Sassoulas:
Could you explain what you meant and the advantages of the template approach @DanielNoord ? I'm not sure about what it mean.

DanielNoord:
Well just the whole discussion in that issue:
Do we want to create an `extend-disable` option?
Do we want to create shareable configuration templates? If so, how should these two interact.

The implementation of those questions might be more complicated if we add more stuff right now.

Pierre-Sassoulas:
Ha thank you I get it know. I think this MR won't impact the shareable template configuration, this is only a way for pylint to internally change our default values easily. I think we should not expose it in configuration templates.

cdce8p:
> Ha thank you I get it know. I think this MR won't impact the shareable template configuration, this is only a way for pylint to internally change our default values easily. I think we should not expose the `default_enabled` option in configuration templates.

I agree. The way I think about it, there are / will be multiple layers:
1. The default configuration (`default_enabled` being part of it)
2. Template / profiles which can overwrite the default configuration in a predefined manner. Similar to [isort profiles](https://pycqa.github.io/isort/docs/configuration/profiles.html).
3. Custom overwrites by the user.

Since `default_enabled` would be part of the default config, it shouldn't impact profiles at all. Similar to the way profiles won't change message ids or descriptions.

DanielNoord:
Yeah, but it might interact with how we implement `extend-enable` which in turn might depend on templating. If you both think this is fine please go ahead and merge, but I just wanted to point this might have unforeseen consequences for future implementations,

Pierre-Sassoulas:
I think the separation of default configuration / templated configuration /  user overwrite will prevent most issue with this, I'll let you merge if you agree @cdce8p.

cdce8p:
> I think the separation of default configuration / templated configuration / user overwrite will prevent most issue with this, I'll let you merge if you agree @cdce8p.

I agree. It's good to be careful but I don't see an issue with `default_enabled` going forward. It just represents the lowest config level, i.e. the default configuration. Anything else will be on top of that.

PR #7632, opened 2022-10-16
Add regression test for no-member with empty AnnAssign

### LINKED ISSUE #7631: Regression `no-member` with inherited AnnAssign nodes
Regression from #7509
The false-positive is actually caused by an issue in astroid. In particular with https://github.com/PyCQA/astroid/blob/v2.12.11/astroid/nodes/scoped_nodes/scoped_nodes.py#L2570-L2577

Opened this so I don't forget to add the example as regression test once the issue is fixed.

```py

class Base:
    attr: int | None = None

class Parent(Base):
    attr: int

class Child(Parent):
    attr = 2

    def __init__(self):
        self.attr = self.attr | 4  # false-positive
```
```
test.py:13:20: E1101: Instance of 'Child' has no 'attr' member (no-member)
```

cdce8p:
## Description
No changelog entry necessary since regression is only present in main and not released yet.

Requires https://github.com/PyCQA/astroid/pull/1834
Closes #7631

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -191,7 +191,7 @@ jobs:
         run: >-
           echo "key=venv-${{ env.CACHE_VERSION }}-${{
             hashFiles('setup.cfg', 'requirements_test_min.txt')
-          }}" >> $GITHUB_OUTPUT
+          }}" >> $env:GITHUB_OUTPUT
       - name: Restore Python virtual environment
         id: cache-venv
         uses: actions/cache@v3.0.11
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
The primer failing is expected when astroid change, but the tests should not fail under windows.

cdce8p:
> The primer failing is expected when astroid change, but the tests should not fail under windows.

The cache restore on windows doesn't work correctly at the moment. Thus astroid `2.12.11` is reused and the workflow fails. Fix #7652

Pierre-Sassoulas (approved):

PR #7642, opened 2022-10-17
Add `bad-dunder-name` checker

### LINKED ISSUE #3038: Add an option to check for misspellings in special functions
Currently, the errors in naming of special functions aren't checked. Normal functions act as interfaces and they are readily checked. But for special functions, it becomes hard to confirm that the code written will work and many times, the eyes gloss over things like number of underscores (too many or too few, hard to distinguish), spelling (should that be invert or inv, eq or equal or equals).
```py
class abc:
    __init_(self):
        pass
    ___neg__(self):
        pass
    __inv__(self):
        pass
```

Since the special functions are limited, testing for spellings (and offering the suggestions as a plugin) would be ideal.

clavedeluna:
## Type of Changes

## Description

The new `bad-dunder-name` checker flags any potentially misspelled known dunder name method and flags any method defined with `_..._` that's not one of the pre-defined ones.

If you can think of more test cases, happy to add them.

Closes #3038


pylint/checkers/dunder_methods.py
```diff
@@ -167,6 +173,29 @@ def within_dunder_def(node: nodes.NodeNG) -> bool:
...
+        of the pre-defined names.
+        """
+        # ignore module-level functions
+        if not node.is_method():
+            return
+
+        if (
+            # Detect something that could be a dunder method
+            node.name.startswith("_")
+            or node.name.endswith("_")
```
Pierre-Sassoulas, line 188:
```suggestion
            node.name.startswith("_")
            and node.name.endswith("_")
```
A function starting with ``_`` is protected, a function starting with ``__`` is private those should not raise.

tests/functional/b/bad_dunder_name.py
```diff
@@ -0,0 +1,30 @@
...
+# pylint: disable=missing-function-docstring, unused-private-member
+
+
+class Apples:
+    def __hello__(self):  # [bad-dunder-name]
+        print("hello")
+
+    def hello(self):
+        print("hello")
+
```
Pierre-Sassoulas, line 11:
```suggestion

   def _protected_method(self):
        print("Protected")

   def __private_method(self):
        print("Private")

**Review by Pierre-Sassoulas — CHANGES_REQUESTED** [review]

File: pylint/checkers/dunder_methods.py [hunk 3]
```diff
@@ -193,4 +254,4 @@ def visit_call(self, node: nodes.Call) -> None:
 
 
 def register(linter: PyLinter) -> None:
-    linter.register_checker(DunderCallChecker(linter))
+    linter.register_checker(DunderChecker(linter))
```
clavedeluna, line 257: could also split the two checks into their own checker classes if that makes which checker uses which names more clear

tests/functional/a/arguments_differ.py
```diff
@@ -1,5 +1,6 @@
 """Test that we are emitting arguments-differ when the arguments are different."""
-# pylint: disable=missing-docstring, too-few-public-methods, unused-argument,useless-super-delegation, unused-private-member
+# pylint: disable=missing-docstring, too-few-public-methods, unused-argument,useless-super-delegation
+# pylint: disable=unused-private-member, bad-dunder-name
 
 class Parent:
```
Pierre-Sassoulas, line 6:
```suggestion

```
To avoid the full result of the functional test being changed because of the line change.

tests/functional/b/bad_dunder_name.py
```diff
@@ -0,0 +1,50 @@
...
+    def init(self):
+        # valid name even though someone could accidentally mean __init__
+        pass
+
+    def __init_(self):  # [bad-dunder-name]
+        # author likely unintentionally misspelled the correct init dunder.
+        pass
+
+    def _init_(self):
```
Pierre-Sassoulas, line 26: This would raise with the change suggested earlier, I think it make sense.

tests/functional/b/bad_dunder_name.py
```diff
@@ -0,0 +1,50 @@
...
+        pass
+
+    def ___neg__(self):  # [bad-dunder-name]
+        # author likely accidentally added an additional `_`
+        pass
+
+    def __inv__(self):  # [bad-dunder-name]
+        # author likely meant to call the invert dunder method
+        pass
+
```
Pierre-Sassoulas, line 37:
```suggestion


    def __user_defined_good_dunder_name__(self):
        print("Whatever the people want they get.")
```

Will need a rc file with:
```ini
good_dunder_name = __user_defined_good_dunder_name__, __something_else__
```

pylint/checkers/dunder_methods.py
```diff
@@ -118,24 +118,53 @@
...
-    We exclude __new__, __subclasses__, __init_subclass__, __set_name__,
-    __class_getitem__, __missing__, __exit__, __await__,
-    __aexit__, __getnewargs_ex__, __getnewargs__, __getstate__,
-    __setstate__, __reduce__, __reduce_ex__
-    since these either have no alternative method of being called or
-    have a genuine use case for being called manually.
+
+    For `unnecessary-dunder-call` we exclude dunder names in list
+    `EXTRA_DUNDER_METHODS` since these either have no alternative
+    method of being called or have a genuine use case for being called manually.
```
Pierre-Sassoulas, line 159: How was it done previously ? It seems the refactor should include removing the old way of doing it to use the new list introduced here.

Pierre-Sassoulas (changes requested):
It starts to look pretty good 👍

pylint/checkers/dunder_methods.py
```diff
@@ -167,6 +202,32 @@ def within_dunder_def(node: nodes.NodeNG) -> bool:
...
+            return
+
+        if (
+            # Detect something that could be a dunder method
+            # It must start with `__` to be private, but could end
+            # with `_` if there is a misspelling.
+            node.name.startswith("__")
+            and node.name.endswith("_")
+            and node.name not in self._dunder_methods
+            and node.name not in EXTRA_DUNDER_METHODS + DUNDER_PROPERTIES
```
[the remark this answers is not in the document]
clavedeluna: This feels premature to me. Should be added when that config is added, no? I can make an issue to add it as a next iteration

tests/functional/b/bad_dunder_name.py
clavedeluna: see comment above. If you don't think it's premature, I"ll add it.

pylint/checkers/dunder_methods.py
clavedeluna: The methods listed in this docstring were never listed in any object. Instead, because the only checker that existed was `unnecessary-dunder-call`, these methods just needed to NOT be included in `DUNDER_METHODS`. BUT with the addition of the `bad-dunder-name` check ,if we don't specifically add these EXTRA methods and check against them, then we would think they are bad-dunder-names when they are not. Does that make sense?

pylint/checkers/dunder_methods.py
```diff
@@ -167,6 +202,30 @@ def within_dunder_def(node: nodes.NodeNG) -> bool:
...
+        """Check if known dunder method is misspelled or dunder name is not one
+        of the pre-defined names.
+        """
+        # ignore module-level functions
+        if not node.is_method():
+            return
+
+        # Detect something that could be a bad dunder method
+        if (
+            node.name.startswith(("_", "__"))
```
Pierre-Sassoulas, line 216:
```suggestion
            node.name.startswith("_")
```
After thinking about it, this is enough as if this is true the other one will be too.

pylint/checkers/dunder_methods.py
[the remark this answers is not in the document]
Pierre-Sassoulas: Well fair point, let's wait for someone to ask for it then.

DanielNoord:
I haven't followed discussion or review but was just triggered by the primer. Is this something for an extension? It seems many plugins and packages define non standard dunders (and that's technically allowed by the Python spec just without guarantee that it will always work).

pylint/checkers/dunder_methods.py
Pierre-Sassoulas: Make sense thank you for explaining.

pylint/checkers/dunder_methods.py
clavedeluna: lol 🤦‍♀️ of course, I missed this as well. I put that initially :)

clavedeluna:
> I haven't followed discussion or review but was just triggered by the primer. Is this something for an extension? It seems many plugins and packages define non standard dunders (and that's technically allowed by the Python spec just without guarantee that it will always work).

Ok cool, I can make it an extension. @Pierre-Sassoulas does that make the config for good dunder names moot, or do we want both?

Pierre-Sassoulas:
I think that if we want this check to be used at all, it needs an option to define what good dunder name are (django for example would probably add ``__conform__``, ``__html__``, etc.). We probably means to define it better somewhere but extensions is something we need to maintain too, so they should at least provide some value.

Pierre-Sassoulas:
> ? What's your decision?

We need the option for good dunder name.

Pierre-Sassoulas (approved):
Pretty nice checker @clavedeluna, I think the refactor to put in pylint.constants also make a lot of sense. I'll wait for another reviewer before merging, new checker are always what bring the most issue in a release.

pylint/extensions/dunder.py
```diff
@@ -0,0 +1,77 @@
...
+    def visit_functiondef(self, node: nodes.FunctionDef) -> None:
+        """Check if known dunder method is misspelled or dunder name is not one
+        of the pre-defined names.
+        """
+        # ignore module-level functions
+        if not node.is_method():
+            return
+
+        # Detect something that could be a bad dunder method
+        if (
```
DanielNoord, line 63: I missed the discussion here, but is there a reason why we don't check for `__`? Could `_test_` be something private?

Pierre-Sassoulas: A function starting with ``_`` is protected, a function starting with ``__`` is private those should not raise. But ``_init__`` is a bad dunder name that we should detect so a single underscore is enough.

DanielNoord (dismissed):

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Thank you for your work on this checker @clavedeluna, I think this is a nice addition to pylint ! Let's see how it goes, but I'm pretty confident we won't have too much issues opened for this one.

PR #7650, opened 2022-10-19
Remove __index__ from unnecessary-dunder-call check

### LINKED ISSUE #6795: False positive Unnecessarily calls dunder method __index__
### Bug description

A little background. The `str` justify methods (`rjust()`, `ljust()`, `center()`) take `SupportsIndex` types for `width` rather than just straight integers, allowing you to use custom types like so.

```python
>>> class Foo:
...     def __index__(self):
...         return 10
... 
>>> 'abcd'.rjust(Foo())
'      abcd'
```

If you subclass `str` and re-implement these methods in a compatible manner, you use `width.__index__()` instead of using `width` directly as an integer. With Pylint 2.14.0 this now raises `unnecessary-dunder-call` which seems intended for sequences, which have an `index()` method, rather than `SupportsIndex` types which don't.

A simple example:

```python
"""Example of false positive for Unnecessarily calls dunder method __index__"""

import typing

class MyString(str):
    """Custom str implementation"""

    def rjust(self, width: typing.SupportsIndex, fillchar: str = ' '):
        """Override of str.rjust"""
        width = width.__index__()
```

### Configuration

_No response_

### Command used

```shell
pylint index_test.py
```

### Pylint output

```shell
************* Module index_test
Desktop/index_test.py:10:16: C2801: Unnecessarily calls dunder method __index__. Use index method. (unnecessary-dunder-call)

Your code has been rated at 7.50/10
```

### Expected behavior

Should be a clean run. That said, I understand this is a corner case and may be difficult to detect.

### Pylint version

```shell
pylint 2.14.0
astroid 2.11.5
Python 3.10.4 (main, Mar 25 2022, 00:00:00) [GCC 12.0.1 20220308 (Red Hat 12.0.1-0)]
```

### OS / Environment

Fedora 36

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description

Discussion in #6795 led to the conclusion that there are legitimate reasons to call `__index__`, so it should not be flagged.

Closes #6795


Pierre-Sassoulas:
Thank you @clavedeluna !

PR #7655, opened 2022-10-21
Sort `--generate-rcfile` ouput

cdce8p:
## Description
Sort output of `--generate-rcfile` similar to  the  one from `--generate-toml-config` and  update  example pylintrc.
Noticed this while trying the release workflow.

Note: To make backporting easier, don't squash merge this PR.

https://github.com/PyCQA/pylint/blob/15dd079792ce8e6b37856450f9e4a928f11b3549/pylint/config/arguments_manager.py#L704-L707

### DIFF
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -447,7 +447,10 @@ class _ArgumentsManager:
         )
         options_by_section = {}
         sections = []
-        for group in self._arg_parser._action_groups:
+        for group in sorted(
+            self._arg_parser._action_groups,
+            key=lambda x: (x.title != "Main", x.title),
+        ):
             group_name = group.title
             assert group_name
             if group_name in skipsections:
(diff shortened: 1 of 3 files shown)


pylint/config/arguments_manager.py
```diff
@@ -447,7 +447,10 @@ def generate_config(
         )
         options_by_section = {}
         sections = []
-        for group in self._arg_parser._action_groups:
+        for group in sorted(
+            self._arg_parser._action_groups,
+            key=lambda x: (x.title != "Main", x.title),
+        ):
```
Pierre-Sassoulas, line 453: Shouldn't the options also be sorted line 460 to 464 ?

Pierre-Sassoulas (commented):
Makes perfect sense, thank you.

pylint/config/arguments_manager.py
cdce8p: 👍🏻  Done, although it didn't change much. I force pushed the changes to keep the history clean for the merge commit.

Pierre-Sassoulas (approved):
👍

PR #7665, opened 2022-10-23
Fix showing DeprecationWarnings for functional tests

cdce8p:
## Description
The `recwarn` fixture for `test_functional` prevents all `DeprecationWarnings` from showing up which aren't triggered at import. Noticed while working on a DeprecationWarning for https://github.com/PyCQA/astroid/pull/1389.

The previous code was added in https://github.com/PyCQA/pylint/commit/e37bf8c5d64ef5ce6becf631663b7880d8ee5a67.

Refs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html#warns

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -11,7 +11,6 @@ from pathlib import Path
 
 import pytest
 from _pytest.config import Config
-from _pytest.recwarn import WarningsRecorder
 
 from pylint import testutils
 from pylint.testutils import UPDATE_FILE, UPDATE_OPTION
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
Thank you for the fix, that was my bad !

cdce8p:
@Pierre-Sassoulas Seems like you accidentally dismissed the approval.

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
(I'm planning on creating a standalone pylint testutil package that would be properly tested, I don't dare to tell when 😄 I need adminship in pycka or a pylint organisation first)

cdce8p:
> (I'm planning on creating a standalone pylint testutil package that would be properly tested, I don't dare to tell when 😄 I need adminship in pycka or a pylint organisation first)

It seems I'm able to create a new repo in the PyCQA org. You should also be able to already.

Just to clarify, this is just a helper package for running functional tests? The tests themself will stay here?
_(That's what I would suggest, at least.)_

Pierre-Sassoulas:
It would be the current content of ``pylint.testutil``, we'd make it an optional install later on (so pylint is lighter for everyone, but lib developer can still use it without too much change in their requirements). The test we'd move would be the one to test the testutil themselves (tests.testutils https://github.com/PyCQA/pylint/tree/main/tests/testutils).

PR #7666, opened 2022-10-23
Confirm changes before tbump commit

cdce8p:
## Description
Add an additional `tbump` step to confirm changes before the commit is created.
Found this to be quite helpful when I was doing the patch release.

Refs: https://stackoverflow.com/questions/1885525/how-do-i-prompt-a-user-for-confirmation-in-bash-script


Pierre-Sassoulas (approved):
Amazing !

PR #7669, opened 2022-10-24
Add docs for modified-* messages and a couple others

clavedeluna:
## Type of Changes

## Description

Refs #5953


Pierre-Sassoulas (commented):
LGTM, there's a test that is failing, possibly because of failing optimization (?)

Pierre-Sassoulas:
Bad optimization with ``utils.only_required_for_messages`` decorator on some functions when it's not really the case are a prime suspect in these case. I would check the call stack of ``misplaced-comparison-constant`` and verify this decorator.

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Thank you for the investigation @clavedeluna 👍

PR #7671, opened 2022-10-25
Run tests with Python 3.11.0

cdce8p:
## Description
Python 3.11 was released yesterday 🎉

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -7,7 +7,7 @@ on:
 env:
   # Also change CACHE_VERSION in the other workflows
   CACHE_VERSION: 31
-  DEFAULT_PYTHON: "3.11-dev"
+  DEFAULT_PYTHON: "3.11"
 
 jobs:
   check-changelog:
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas:
Thank you @cdce8p 👍 It seems like the windows 3.11 fail is genuine, I don't see any significant change to the ``io`` module in the release note though.

cdce8p:
> Thank you @cdce8p 👍 It seems like the windows 3.11 fail is genuine, I don't see any significant change to the `io` module in the release note though.

Seems like an issue with the new pytest release (7.2.0) and pytest-benchmark. https://github.com/ionelmc/pytest-benchmark/issues/226
We could pin `pytest==7.1.3` until a new version for `pytest-benchmark` is released.

cdce8p:
> > Seems like an issue with the new pytest release (7.2.0) and pytest-benchmark.
> 
> Hmm, right, it's not related to this MR, 3.11-dev also fail on main right now. Do you think we should open another MR or fix in this one ?

Let's do it in a separate one in case we won't to backport it to the `2.15.x` branch.

Pierre-Sassoulas (approved):

PR #7672, opened 2022-10-25
Add docs for non-* and other messages

clavedeluna:
## Type of Changes

## Description

Refs #5953

### DIFF
--- a/doc/data/messages/n/nan-comparison/bad.py
+++ b/doc/data/messages/n/nan-comparison/bad.py
@@ -1,0 +1,5 @@
+import numpy
+
+x = 42
+if x == numpy.NaN:  # [nan-comparison]
+    print("Bad!")
--- a/doc/data/messages/n/nan-comparison/bad.py
+++ b/doc/data/messages/n/nan-comparison/bad.py
@@ -1,5 +1,5 @@
-import numpy
+import numpy as np
 
-x = 42
-if x == numpy.NaN:  # [nan-comparison]
-    print("Bad!")
+
+def both_nan(x, y) -> bool:
+    return x == np.NaN and y == float("nan")  # [nan-comparison, nan-comparison]
--- a/doc/data/messages/n/nan-comparison/good.py
+++ b/doc/data/messages/n/nan-comparison/good.py
@@ -1,1 +1,3 @@
-# This is a placeholder for correct code for this message.
+x = 42
+if x == 42:
+    print("Bad!")
--- a/doc/data/messages/n/nan-comparison/good.py
+++ b/doc/data/messages/n/nan-comparison/good.py
@@ -1,3 +1,5 @@
-x = 42
-if x == 42:
-    print("Bad!")
+import numpy as np
+
+
+def both_nan(x, y) -> bool:
+    return np.isnan(x) and np.isnan(y)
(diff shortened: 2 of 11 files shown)


doc/data/messages/n/nan-comparison/bad.py
```diff
@@ -0,0 +1,5 @@
+import numpy
+
+x = 42
+if x == numpy.NaN:  # [nan-comparison]
+    print("Bad!")
```
Pierre-Sassoulas, line 5:
```suggestion
def both_nan(x, y) -> bool:
   return x == np.NaN and y == float("nan")  # [nan-comparison]
```

doc/data/messages/n/nan-comparison/good.py
```diff
@@ -1 +1,3 @@
-# This is a placeholder for correct code for this message.
+x = 42
+if x == 42:  # [nan-comparison]
+    print("Bad!")
```
Pierre-Sassoulas, line 3:
```suggestion
import numpy as np

def both_nan(x, y) -> bool:
   return np.isnan(x) and np.isnan(y)
```

doc/data/messages/n/nan-comparison/bad.py
```diff
@@ -0,0 +1,5 @@
+import numpy
```
Pierre-Sassoulas, line 1:
```suggestion
import numpy as np
```

clavedeluna: I know `as np` is customary but it's not necessary to demonstrate nan-comparison...

doc/data/messages/n/nan-comparison/bad.py
clavedeluna: I think this suggestion adds unnecessary complexity and is harder to differentiate what the "problem" is

doc/data/messages/n/nan-comparison/bad.py
Pierre-Sassoulas: Sure, but if we have numpy example we might as well have the same style than the numpy doc.

doc/data/messages/n/nan-comparison/bad.py
Pierre-Sassoulas: if x = 42 then there's no need to compare it, so a function with unknown value is better imo, Then if we just return np.isnan(x) it's not a very good example, right, we could use isnan directly. Also the other variable permit to illustrate the two majors way to create a nan comparison.

doc/data/messages/n/nan-comparison/good.py
clavedeluna: functional tests fail if there's 2 linting failures on the same line so I just did one

Pierre-Sassoulas: It's possible to have multiple messages raised on the same line in functional tests with ``# [message-1, message-2]``. This would also permit to illustrate that both math.isnan or np.isnan are okay in good.py 😄

clavedeluna: didn't know that! Thanks! will update

Pierre-Sassoulas (approved):
Thank you !

PR #7673, opened 2022-10-25
[github action] Follow the principle of least privileges

Pierre-Sassoulas:
## Type of Changes

## Description

This set the least amount of privilege for each github actions when it wasn't already set. 

Full disclosure, I'm being paid to make pylint more secure by Tidelift and this is part of this effort.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -9,6 +9,9 @@ env:
   CACHE_VERSION: 31
   DEFAULT_PYTHON: "3.11"
 
+permissions:
+  contents: read
+
 jobs:
   check-changelog:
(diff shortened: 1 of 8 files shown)


Pierre-Sassoulas:
@cdce8p Let's use this MR to check that #7674 #7671 worked correctly.

cdce8p (approved):
Looks good, but I don't think it's necessary anymore.

The same can be archived by modifying the `Action permissions` -> `Workflow permissions` in the repo settings. I just toggled `Read repository contents permission`, so `contents: read` is the default for all workflows.

--
FYI I also modified a few other `Action permissions` settings to increase security.
* Disabled `Allow Github Actions to create and approve pull requests`
* Only allow specific / predefined actions and reusable workflows.

PR #7674, opened 2022-10-25
Pin pytest to resolve issue with pytest-benchmark

cdce8p:
## Description
Fix issue with pytest `v7.2.0` and `pytest-benchmark` discovered in #7671
Ref: https://github.com/ionelmc/pytest-benchmark/issues/226

### DIFF
--- a/requirements_test_min.txt
+++ b/requirements_test_min.txt
@@ -3,7 +3,9 @@
 # You need to increment the CACHE_VERSION in github actions too
 astroid==2.12.12  # Pinned to a specific version for tests
 typing-extensions~=4.4
-pytest~=7.1
+# Pin pytest until issue with pytest-benchmark and v7.2.0 is resolved
+# https://github.com/ionelmc/pytest-benchmark/issues/226
+pytest==7.1.3
 pytest-benchmark~=3.4
 pytest-timeout~=2.1
 towncrier~=22.8


Pierre-Sassoulas (approved):

cdce8p:
Seems like we shouldn't use auto-merge 🤔
It merged the PR before the Windows tests were run.

cdce8p:
> > Seems like we shouldn't use auto-merge
> 
> Right, I forgot. We could add more required tests but it can be annoying when we want to merge fast.

I've added `run / 3.11 / Linux` and `run / 3.11 / Windows`. That way we know that at least `some` things are working before an auto-merge.

Pierre-Sassoulas:
>  run / 3.11 / Windows

The problem was that this is one of the slower tests (only primer is longer). Ho well, we're just going to wait a little longer before merging typo-MR 🤷

Pierre-Sassoulas:
Beside we'll be able to enable auto-merge which was the point. Now that I think about it we could add primers and always be able to use auto-merge ?

cdce8p:
> Beside we'll be able to enable auto-merge which was the point. Now that I think about it we could add primers and always be able to use auto-merge ?

The primer tests will / should never fail. I don't think that will work as intended. Even if it's a bit more work, doing the merge manual is the best option for us I think.

Pierre-Sassoulas:
The issue with required job like this is that for documentation merge request where we don't want to run all the tests we can't merge them if tests are required.

cdce8p:
We could always modify the branch protection rules and remove the `Do not allow bypassing the above settings` option. Something I would recommend anyway as it's limited to admins and even we would need to click a checkbox before we can merge something without all checks.

Pierre-Sassoulas:
Agreed and done !

PR #7678, opened 2022-10-26
Fix a false negative for ``unused-import``

### LINKED ISSUE #7547: unused-import false negative when typing.Annotated has str parameters
### Bug description

example.py:

```python
from pathlib import Path
import typing as t

example: t.Annotated[str, "Path"] = "/foo/bar"
```

### Configuration

_No response_

### Command used

```shell
pylint example.py
```

### Pylint output

```shell
Success: no issues found in 1 source file

```

### Expected behavior

It should report `unused-import`, since the parameter to `Annotated` does not refer to `pathlib.Path`.

### Pylint version

```shell
pylint 2.15.3
astroid 2.12.10
Python 3.10.7 (main, Sep  7 2022, 00:00:00) [GCC 12.2.1 20220819 (Red Hat 12.2.1-1)]
```

### OS / Environment

fedora 36

### Additional dependencies

_No response_

mbyrnepr2:
## Type of Changes

## Description
Fix a false negative for unused-import when a constant parameter of typing.Annotated was treated as a reference to an import.

Closes #7547


Pierre-Sassoulas (approved):
LGTM !

DanielNoord (commented):
Is `typing_member` a term I just don't know? I don't really understand what you mean with it..

pylint/checkers/utils.py
```diff
@@ -1919,24 +1919,29 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
     return False
 
 
-def is_typing_literal(node: nodes.NodeNG) -> bool:
-    """Check if a node refers to typing.Literal."""
+def is_typing_member(node: nodes.NodeNG, typing_members: str | tuple[str, ...]) -> bool:
```
DanielNoord, line 1922:
```suggestion
def is_typing_member(node: nodes.NodeNG, names_to_check: tuple[str, ...]) -> bool:
```

I now understand the function name, let's just makes the argument a little clearer.

I would also suggest to do the tuple conversion outside of this helper: we're changing calls anyway, might as well make this handle `tuples` only.

pylint/checkers/utils.py
```diff
@@ -1919,24 +1919,29 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
...
             return False
+
+        if isinstance(typing_members, str):
+            typing_members = (typing_members,)
         if isinstance(import_from, nodes.ImportFrom):
-            return (  # type: ignore[no-any-return]
+            return (
                 import_from.modname == "typing"
-                and import_from.real_name(node.name) == "Literal"
+                and import_from.real_name(node.name) in typing_members
```
DanielNoord, line 1937: Any reason not to use `qname()` here? We could do something with a for loop or comprehension and an f-string like `f'typing.{n}'`

DanielNoord (changes requested):

pylint/checkers/utils.py
mbyrnepr2: The existing function logic just worked the way I needed it to so I didn't feel it needed changing.
Do you mean use `qname` instead of `real_name`? It looks like [`real_name`](https://github.com/PyCQA/astroid/blob/main/astroid/nodes/_base_nodes.py#L153) is intended for the current use case?

DanielNoord: Never mind, that's indeed the correct method. Got my methods confused here, sorry!

mbyrnepr2: No worries at all; I like your analytical reviews!

DanielNoord (approved):

Pierre-Sassoulas (approved):

DanielNoord:
The primer message above nudges towards my suspicion that something is off with the merging of `main` in the PR primer run.

PR #7685, opened 2022-10-27
false positive `unnecessary-list-index-lookup` for enumerate

### LINKED ISSUE #7682: unnecessary-list-index-lookup is falsely being reported when there is a start index
### Bug description

The code is:
```python
series = [1, 2, 3, 4, 5]
output_list = [
    (item, series[index])
    for index, item in enumerate(series, start=1)
    if index < len(series)
]
print(output_list)
```
Which produces the output:
`[(1, 2), (2, 3), (3, 4), (4, 5)]`

When there is a start parameter on the enumeration, "item" and "series[index]" are not the same object
but pylint suggests:
`R1736: Unnecessary list index lookup, use 'item' instead (unnecessary-list-index-lookup)`
If we follow the advice we would get a different result set 
`[(1, 1), (2, 2), (3, 3), (4, 4)]`

I think the error should be suppressed when there's a start index on the enumeration

### Configuration

_No response_

### Command used

```shell
pylint .\pylint.py
```

### Pylint output

```shell
************* Module pylint
pylint.py:4:11: R1736: Unnecessary list index lookup, use 'item' instead (unnecessary-list-index-lookup)
```

### Expected behavior

No error at all

### Pylint version

```shell
pylint 2.14.1
astroid 2.11.5
Python 3.9.6 (tags/v3.9.6:db3ff76, Jun 28 2021, 15:26:21) [MSC v.1929 64 bit (AMD64)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description

Pylint will now not emit a `unnecessary-list-index-lookup` warning if the loop / comprehension is iterating over `enumerate` called with a `start` arg/kwarg.

Closes #7682


pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -2199,3 +2204,16 @@ def _check_unnecessary_list_index_lookup(
                 args=(node.target.elts[1].name,),
                 confidence=HIGH,
             )
+
+    def _enumerate_with_start(self, node: nodes.For | nodes.Comprehension) -> bool:
```
jacobtylerwalls, line 2208: This works, but what if you just replace `or not node.iter.args` above with `or len(node.iter.args) != 1`? 🤔

doc/whatsnew/fragments/7682.bugfix
```diff
@@ -0,0 +1,3 @@
+``unnecessary-list-index-lookup`` will not be emitted if ``enumerate`` is called with a `start` arg or kwarg.
```
Pierre-Sassoulas, line 1:
```suggestion
``unnecessary-list-index-lookup`` will not be wrongly emitted if ``enumerate`` is called with ``start``.
```

tests/functional/u/unnecessary/unnecessary_list_index_lookup.py
```diff
@@ -81,3 +81,20 @@ def process_list_again(data):
...
+]
+
+output_list = [
+    (item, series[index])
+    for index, item in enumerate(series, 1)
+    if index < len(series)
+]
+
+for idx, val in enumerate(series, start=2):
+    print(my_list[idx])
```
Pierre-Sassoulas, line 100:
```suggestion
for idx, val in enumerate(series, start=2):
    print(my_list[idx])
    
for idx, val in enumerate(series, 2):
    print(my_list[idx])    
    
for idx, val in enumerate(series, start=0):
    print(my_list[idx])  # [unnecessary-list-index-lookup]
    
for idx, val in enumerate(series, 0):
    print(my_list[idx])  # [unnecessary-list-index-lookup]    
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -2199,3 +2204,16 @@ def _check_unnecessary_list_index_lookup(
...
+        """Check presence of `start` kwarg or second argument to enumerate
+        For example.
+
+        `enumerate([1,2,3], start=1)`
+        `enumerate([1,2,3], 1)`
+        """
+        if len(node.iter.args) > 1:
+            # We assume the second argument to `enumerate` is the `start` arg.
```
Pierre-Sassoulas, line 2216:
```suggestion
            # We assume the second argument to `enumerate` is the `start` arg.
            # It's a reasonable assumption for now as it's the only possible argument:
            # https://docs.python.org/3/library/functions.html#enumerate
```

Pierre-Sassoulas (changes requested):
Thank you for the fix @clavedeluna. I think we can handle the false negative with start=0 too.

pylint/checkers/refactoring/refactoring_checker.py
clavedeluna: given the more complicated logic, this isn't applicable

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -2199,3 +2204,39 @@ def _check_unnecessary_list_index_lookup(
...
+            if keyword.arg == "start":
+                start_val = (
+                    keyword.value.operand.value
+                    if isinstance(keyword.value, nodes.UnaryOp)
+                    else keyword.value.value
+                )
+
+                return not start_val == 0
+
+        return False
```
clavedeluna, line 2242: I actually tried to refactor this method because it looks like a bunch of duplicate code, but:
1. refactoring wouldn't allow for some short-circuting which seems best for performance
2. args v. kwargs has sliiightly different needs, `.value.operand` v. `.operand` etc

clavedeluna:
> Thank you for the fix @clavedeluna. I think we can handle the false negative with start=0 too.

good call, I also went ahead and added calls for negative start value which made the logic more robust albeit a bit more complicated

pylint/checkers/refactoring/refactoring_checker.py
Pierre-Sassoulas, line 2242:
```suggestion
         else:
            for keyword in node.iter.keywords:
                if keyword.arg == "start":
                    start_val = (
                    keyword.value.operand.value
                    if isinstance(keyword.value, nodes.UnaryOp)
                    else keyword.value.value
                )

        return not start_val == 0
```
How about this ?

clavedeluna: Hm I actually don't like it. The additional indent isn't great, and the nice thing about treating `for keyword in node.iter.keywords:` as an else is that if there are no keywords, it just won't enter the loop.

Pierre-Sassoulas (dismissed):
Great ! 👍

Pierre-Sassoulas:
The failing test on python 3.11 seems genuine

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -2199,3 +2204,43 @@ def _check_unnecessary_list_index_lookup(
...
+                start_val = self._get_start_value(keyword.value)
+                if start_val is None:
+                    return False
+                return not start_val == 0
+
+        return False
+
+    def _get_start_value(self, node: nodes.NodeNG) -> Any:
+        if isinstance(node, nodes.Name):
+            inferred = utils.safe_infer(node)
```
Pierre-Sassoulas, line 2240: It seems like the confidence is not HIGH anymore if we're in this branch. Maybe we should return the confidence in this function ?

Pierre-Sassoulas (changes requested):

pylint/checkers/refactoring/refactoring_checker.py
clavedeluna: oh good call! So the rule is, anytime inference is used to determine whether a msg should be emitted, then use INFER confidence, correct?

Pierre-Sassoulas: >  So the rule is, anytime inference is used to determine whether a msg should be emitted, then use INFER confidence

Yeap !

Pierre-Sassoulas (dismissed):
LGTM, thank you !

clavedeluna:
primer doesn't wanna prime 😞

Pierre-Sassoulas:
The primer exit with code 1 and no other information (apparently after priming every repositories correctly), do you have any idea what could be wrong @DanielNoord ? I did not have time to work on the primer refactor I started yet but I'm still planning to handle this and add tests and log in particular.

DanielNoord:
I haven't looked into it, but it could be that there is a fatal error? I think the primer crashes then.

Pierre-Sassoulas:
👍 I added the blocked label (need an investigation and enhancement on the primers).

DanielNoord:
@clavedeluna Have you ran the primer locally? And seen what its result was?

Pierre-Sassoulas (dismissed):
The code itself LGTM, I think we can merge as soon as we understand what's wrong with the primer 👍

clavedeluna:
> @clavedeluna Have you ran the primer locally? And seen what its result was?

I haven't run the primer locally before. I rely on CI for that since it's long. I did follow [these docs](https://pylint.pycqa.org/en/v2.12.2/development_guide/testing.html#primer-tests) as much as I can and ran 

` pytest -m primer_stdlib --primer-stdlib` > everything passed but running `pytest -m primer_external_batch_one --primer-external # Runs batch one` totally hung and never moved.

If you have suggestions, I can retry.

DanielNoord:
You can just follow what the `Primer / Main` workflow does locally. This should give you an output and might include fatal errors.

clavedeluna:
> You can just follow what the `Primer / Main` workflow does locally.

Easier said than done, unfortunately. Quite a few CI specific things are getting configured. I'm just not able to take the time to run it locally, after multiple attempts.

DanielNoord:
As expected this causes a fatal error. The primer should really be more focal about it but it fails on `Exception on node <For l.2871 at 0x1406429b0> in file '/Users/daniel/DocumentenLaptop/Programming/Github/pylint/tests/.pylint_primer_tests/pandas-dev/pandas/pandas/core/indexes/multi.py'`

If you know what to look for you can actually see it in the Actions output. The line `succesfully primed pandas` is missing, even though it did start priming on `pandas`.

clavedeluna:
@DanielNoord would you be able to provide a bit more info on the exception? was there a stack trace? Or maybe something I could minimally reproduce and add a test for? Great catch! No wonder my other PRs have no issues with the primer

DanielNoord:
> @DanielNoord would you be able to provide a bit more info on the exception? was there a stack trace? Or maybe something I could minimally reproduce and add a test for? Great catch! No wonder my other PRs have no issues with the primer

It's actually quite easy to reproduce this yourself locally. Just run:
```console
python tests/primer/__main__.py prepare --clone
python tests/primer/__main__.py run --type=pr
```

It will take some time to get to `pandas` but the exception will be there.
You can also look at the `pandas-dev/pandas/pandas/core/indexes/multi.py` file as that is where the issue seems to be.

clavedeluna:
glad I opened #7714

clavedeluna:
Thanks @DanielNoord for getting me to finally run the primer locally. I must've done too many of the CI steps and got stuck initially. The error was easy to spot and fix, so I hope everything passes now!

clavedeluna:
😢 primer, I don't think it has anything to do with this PR

Pierre-Sassoulas:
A case study for #7738

doc/whatsnew/fragments/7682.bugfix
```diff
@@ -0,0 +1,3 @@
+``unnecessary-list-index-lookup`` will not be wrongly emitted if ``enumerate`` is called with ``start``.
```
Pierre-Sassoulas, line 1: Shouldn't this be ``7682.false_positive`` ?

Pierre-Sassoulas (changes requested):
Small nitpick, it's ready to go otherwise, nice fix @clavedeluna 🎉

Pierre-Sassoulas (approved):

PR #7691, opened 2022-10-30
Extend R1735 use-dict-literal to highlight all usages of dict that can be replaced with a literal

### LINKED ISSUE #7690: Prefer dict literal over `dict` constructor
### Current problem

Constructing `dict`s using the constructor should be avoided. It has no advantages and may be a bit slower.

## Invalid code

```py
# construct dict using keywords
dict(foo="bar")
# construct dict using a dict
dict(**boo)
```

## Valid code 

```py
# dict literal
{"foo": "bar"}

# casting values to a dict
dict(some_method())

# methods that are not the builtin `dict`
def dict(baz):
    ...

dict(1)
```

### Desired solution

Create a message that finds and highlights invalid code from above. Maybe as an extension of `R1735 use-dict-literal`.

### Additional context

_No response_

hofrob:
## Type of Changes

## Description

Refs #7690 

Closes #7690


hofrob:
1. This could very well be added to the existing `R1735` check. I don't know what the general consensus is on extending existing messages.
2. The naming is kind of confusing now because `use-dict-literal` *only* covers empty dicts (`dict()`). So I decided to use `use-dict-literal-without-kwargs` for now though I think it adds to the confusion.
3. I'm not using a neat pylint feature right now: show how the code would look like if the suggestion is applied. I'd probably need some pointers to accomplish that.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -476,6 +476,12 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "value by index lookup. "
             "The value can be accessed directly instead.",
         ),
+        "R1737": (
+            "Consider using a dict literal {'foo': 'bar'} instead of dict(foo='bar')",
```
Pierre-Sassoulas, line 480:
```suggestion
            "Consider using a dict literal '%s' instead.",
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1594,6 +1602,15 @@ def _check_use_list_or_dict_literal(self, node: nodes.Call) -> None:
...
                     self.add_message("use-dict-literal", node=node)
 
+    def _check_use_dict_literal(self, node: nodes.Call) -> None:
+        """Check if dict is created by using the literal {}."""
+        if not isinstance(node.func, astroid.Name):
+            return
+        if node.func.name == "dict" and not node.args and node.keywords:
+            inferred = utils.safe_infer(node.func)
+            if inferred.qname() == "builtins.dict":
+                self.add_message("use-dict-literal-without-kwargs", node=node)
```
Pierre-Sassoulas, line 1612:
```suggestion
                self.add_message("use-dict-literal-without-kwargs", args=(code_suggestion), node=node)
```
Content of ``code_suggestion`` not included in this suggestion but it migth contain an ``.as_string()`` somewhere 😄

Pierre-Sassoulas (commented):
Thank you for bringing a proof of concept with the issue 👌 

I was also wondering if we should just extend what the existing message is doing or create another one. The former prevent user that don't want the new thing to disable it, but the second one add confusion. We could also rename the existing message to make things clearer and have the best of both world (?)

Regarding code suggestion, we should be careful to not create a behemoth message if the input dict is big we don't want a 400 lines message.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1594,6 +1602,25 @@ def _check_use_list_or_dict_literal(self, node: nodes.Call) -> None:
...
+        if (
+            isinstance(inferred, nodes.ClassDef)
+            and inferred.qname() == "builtins.dict"
+            and not node.args
+            and node.keywords
+        ):
+            code_suggestion = next(node.infer()).as_string()
+            short_suggestion = (
+                f"{code_suggestion[:100]}{'...' if len(code_suggestion) > 100 else ''}"
+            )
```
hofrob, line 1619: I'm limiting the dict output to 100 characters here. What do you think, @Pierre-Sassoulas?

I couldn't find any other examples in the code.

hofrob:
> I was also wondering if we should just extend what the existing message is doing or create another one. The former prevent user that don't want the new thing to disable it, but the second one add confusion. We could also rename the existing message to make things clearer and have the best of both world (?)

My personal opinion would be: Keep the name and code. Extend the functionality of the existing message. The name fits for both situations anyway.

People should expect new lint messages when upgrading the pylint package. But it's easy for me to say :grinning: I don't have to work with huge code bases.

hofrob:
Regarding the failing "readthedocs" build:

```
/home/docs/checkouts/readthedocs.org/user_builds/pylint/checkouts/7691/doc/whatsnew/2/2.16/index.rst:49: WARNING: Bullet list ends without a blank line; unexpected unindent.
```

No idea what to do here. My branch is up-to-date. Do I need to add anything else to the docs? Or is this related to the changes in `pylint/pyreverse/`?

script/.contributors_aliases.json
```diff
@@ -602,5 +602,9 @@
   "yileiyang@google.com": {
     "mails": ["yileiyang@google.com"],
     "name": "Yilei \"Dolee\" Yang"
+  },
+  "hofrob@protonmail.com": {
+    "mails": ["hofrob@protonmail.com"],
+    "name": "Robert Hofer"
```
Pierre-Sassoulas, line 608: No need to do that unless you have another email or an aliased name you want to use. (Just for other people information, do not remove it now that it's done :)

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1594,6 +1602,25 @@ def _check_use_list_or_dict_literal(self, node: nodes.Call) -> None:
...
+            self.add_message(
+                "use-dict-literal-without-kwargs", args=(short_suggestion,), node=node
```
Pierre-Sassoulas, line 1621:
```suggestion
                "use-dict-literal-without-kwargs", args=(short_suggestion,), node=node, confidence=INFERENCE
```
This will require to regenerate the expected output of functional tests and an import from pylint.interface.

Pierre-Sassoulas (commented):
Made some comments, I'm waiting for other maintainer opinion regarding the new message or the extension of the old one :)

DanielNoord:
I like the suggestion of extending the previous message!

hofrob:
> There is surprisingly few violation on the primer which is a very good sign. The primer for sentry:
> 
> ```
>     use-dict-literal-without-kwargs:
>     *Consider using a dict literal '
> 
> class dict(object):
> """dict() -> new empty dictionary
> dict(mapping) -> new dictionary initiali...' instead.*
> ```
> 
> https://github.com/getsentry/sentry/blob/4be98deb27375dac541740b9c8d7ce871ff2bc99/src/sentry/api/endpoints/organization_sdk_updates.py#L44 Seems like something we need to fix though.

I fixed those and added test cases. The code suggestion is now constructed "manually" to be able to handle `**kwargs` separate from actual `keyword=value` arguments.

hofrob:
> I like the suggestion of extending the previous message!

If you have the final say, I will go ahead. Or will others want to chime in?

DanielNoord:
> > I like the suggestion of extending the previous message!
> 
> If you have the final say, I will go ahead. Or will others want to chime in?

They obviously can (and are encouraged to do so of course), but I think two maintainers thinking positively about this is a good sign to go ahead 😄

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1594,6 +1602,37 @@ def _check_use_list_or_dict_literal(self, node: nodes.Call) -> None:
...
+                    rebuilt_dict.append(f'"{keyword.arg}": {keyword.value.as_string()}')
+
+            for keyword in node.kwargs:
+                if isinstance(keyword.value, nodes.Name):
+                    rebuilt_dict.append(f"**{keyword.value.name}")
+
+            code_suggestion = f"{{{', '.join(rebuilt_dict)}}}"
+            short_suggestion = (
+                f"{code_suggestion[:100]}{'...' if len(code_suggestion) > 100 else ''}"
+            )
```
Pierre-Sassoulas: Let's create a private function for rebuilding the dict, if we're going to truncate there's no need to create the whole ``rebuild_dict``. Also we could make the suggestion smaller (maybe 64 chars max) and cut in the middle of the string.

Something like this but in a private function:
```suggestion
            short_suggestion  = "{"
            for keyword in node.keywords:
                if keyword not in node.kwargs:
                    if len(short_suggestion) <= 64:
                        short_suggestion += f'"{keyword.arg}": {keyword.value.as_string()}'
            for keyword in node.kwargs:
                if isinstance(keyword.value, nodes.Name):
                    if len(short_suggestion) <= 64:
                        short_suggestion += f"**{keyword.value.name}"
            short_suggestion += "}"
            if len(short_suggestion) > 64                        
                short_suggestion = f"{short_suggestion[30]}...{short_suggestion[:-30]}
```

Pierre-Sassoulas:
It happened in the past that 2 maintainers agreed on something and another one made an argument so good we changed our mind but it's pretty rare 😄

hofrob:
Ok, so I did the following:

* one method that checks all calls to `dict` (meaning empty dicts are not separate anymore)
* constructing the suggestion happens in a new private static method
* I still use a list to construct the suggestion because handling commas would be a PITA otherwise :grin: (and `set()` won't work because it changes the order on every test run)
* all tests, examples, docs merged into the existing `use-dict-literal` files/directories
* cleaning up code in pylint that was highlighted by the new message resulted in another pylint message showing up (`consider-using-namedtuple-or-dataclass`), I will disable those

# **Important!** The confidence level of an existing message changes

While it was still a separate message, we set the confidence to `INFERENCE`. This will change the confidence for existing messages for empty calls to `dict()`. Just so you're aware.

doc/data/messages/u/use-dict-literal/details.rst
```diff
@@ -0,0 +1 @@
+Use a dict literal instead of keywords and `dict()`. The literal is faster as it avoids an additional function call.
```
Pierre-Sassoulas, line 1: Look like this is already in the description of the message (https://github.com/PyCQA/pylint/pull/7691/files#diff-af1d215b95c8b3ac59218ba588186da4aefd936eef4732518196ea309c3cc788R469) so we could remove the details. We could point to a benchmark of the perf different however.

doc/data/messages/u/use-dict-literal/good.py
```diff
@@ -1 +1,7 @@
 empty_dict = {}
+
+# create using a literal dict
+new_dict = {"foo": "bar"}
+
+# cast the result of a method
+new_dict = dict(some_method())
```
Pierre-Sassoulas, line 7: Is this better than ``dict(**another_dict)`` or ``another_dict.copy()`` ? We should probably justify this in details imo.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -464,9 +464,9 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "The literal is faster as it avoids an additional function call.",
         ),
         "R1735": (
-            "Consider using {} instead of dict()",
+            "Consider using a dict literal '%s' instead.",
```
Pierre-Sassoulas, line 467:
```suggestion
            "Consider using '%s' instead.",
```
Wondering if this would be better or not (I"m aware I suggested the initial message 😄).

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -464,9 +464,9 @@ class RefactoringChecker(checkers.BaseTokenChecker):
...
             "use-dict-literal",
-            "Emitted when using dict() to create an empty dictionary instead of the literal {}. "
+            "Emitted when using dict() to create a dictionary instead of the literal {}. "
```
Pierre-Sassoulas, line 469:
```suggestion
            "Emitted when using dict() to create a dictionary instead of a literal '{ ... }'. "
```

Pierre-Sassoulas (changes requested):
This looks pretty good already, thank you !

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1584,15 +1585,48 @@ def _check_consider_using_with(self, node: nodes.Call) -> None:
...
+                elements.append(f'"{keyword.arg}": {keyword.value.as_string()}')
+            if len(", ".join(elements)) >= 64:
+                break
+        for keyword in node.kwargs:
+            elements.append(f"**{keyword.value.as_string()}")
+            if len(", ".join(elements)) >= 64:
+                break
+        suggestion = f'{{{", ".join(elements)}}}'
+        if len(suggestion) > 64:
+            suggestion = f"{suggestion[:30]}…{suggestion[-30:]}"
```
Pierre-Sassoulas, line 1627: As the suggestion is close to 64 already (because we cut short previously after the one element that would make it more than 64) this might be better:
```suggestion
            suggestion = f"{suggestion}, … "
```
We won't cut any key/value this way and it's ~= 64. Let me know what you think

pylint/checkers/refactoring/refactoring_checker.py
hofrob: I'm always a fan of short and concise :+1: :slightly_smiling_face:

pylint/checkers/refactoring/refactoring_checker.py
hofrob: Yes, makes sense. There's gonna be odd cases one way or the other.

E.g.: a dict with only one key/value pair that has combined 500 characters or something. Those will look silly, no matter what we do.

doc/data/messages/u/use-dict-literal/details.rst
hofrob: https://gist.github.com/hofrob/ad143aaa84c096f42489c2520a3875f9
I tried this simple benchmark script on my machine. Not sure if this counts as a source :grin:.

doc/data/messages/u/use-dict-literal/good.py
hofrob: Well I'd say `{…}` is always faster than `dict(…)`. So `dict(**another_dict)` should be refactored to `{**another_dict}` and that's what this message would suggest.

Which of those is better when copying a dict is probably out of scope here. `copy`, `deepcopy` and `{**another_dict}` are all accepted to be valid.

My goal with this example `new_dict = dict(some_method())` was to show that casting a value to a dict are considered ok in the `dict` constructor. Finding out if one of those could be refactored requires quite a bit more work.

In short: all `dict` calls that contain args will be ignored. `**kwargs` and `keyword=value` arguments are fine since we can make a suggestion to rewrite them as literals.

I hope this makes everything even more confusing! :stuck_out_tongue_winking_eye:

doc/data/messages/u/use-dict-literal/details.rst
Pierre-Sassoulas: > Not sure if this counts as a source

Yes 👍

doc/data/messages/u/use-dict-literal/good.py
Pierre-Sassoulas: Shouldn't we suggest this then ?
```suggestion
new_dict = {**another_dict}
```

Pierre-Sassoulas: > Are you saying I should just list all examples that are valid for this specific pylint message?

No but the examples in good should reflect the example in bad (i.e. what is in good is "the official fix"). 

For ``dict(some_method()) `` we don't have a matching bad example because we don't know, so we could remove it.

doc/whatsnew/fragments/7690.new_check
```diff
@@ -0,0 +1,3 @@
+* Extended ``use-dict-literal`` to cover calls to ``dict()`` that could be replaced with a literal dict.
```
mbyrnepr2, line 1:
```suggestion
Extended ``use-dict-literal`` to cover calls to ``dict()`` that could be replaced with a literal dict.
```

Could be the issue with the readthedocs failure since the output of that gives:
`WARNING: Bullet list ends without a blank line; unexpected unindent.`

doc/data/messages/u/use-dict-literal/details.rst
```diff
@@ -0,0 +1,4 @@
+https://gist.github.com/hofrob/ad143aaa84c096f42489c2520a3875f9
+
+This example script shows an 18% increase in performance when using a literal over the
```
DudeNr33, line 3: I really like that you went the extra mile to provide some hard facts to provide evidence! 👍 
As the last Python versions broguht some major performance improvements, it might be good to note down the Python version with which this result was obtained.

doc/data/messages/u/use-dict-literal/good.py
DudeNr33: +1 for Pierre's last comment, I was also a bit confused how the third example in bad/good are connected (or not).
We should definitely add a 'good' example for copying a dict.

doc/whatsnew/fragments/7690.new_check
DudeNr33: Yes, the news fragments should not contain bullet points, as `towncrier` creates the bullet points for each entry itself.

doc/user_guide/checkers/features.rst
```diff
@@ -779,8 +779,8 @@ Refactoring checker Messages
 :consider-swap-variables (R1712): *Consider using tuple unpacking for swapping variables*
   You do not have to use a temporary variable in order to swap variables. Using
   "tuple unpacking" to directly swap variables makes the intention more clear.
-:use-dict-literal (R1735): *Consider using {} instead of dict()*
-  Emitted when using dict() to create an empty dictionary instead of the
+:use-dict-literal (R1735): *Consider using a dict literal '%s' instead.*
+  Emitted when dict() is used to create a dictionary instead of the
```
DudeNr33, line 783: This file is generated automatically when building the docs (with `doc/exts/pylint_features.py`).
The current description is not completely identical to the one in `refactoring_checker.py`, which means that every time someone builds the documentation locally the `features.rst` will show up as a modified file in `git status` afterwards.

@Pierre-Sassoulas that's a bit of a general problem also with other auto-generated files. Do you have a preference how those should be treated in a PR?

DudeNr33 (changes requested):
Just some minor comments from my side. Looks good to me overall!
I'm also +1 for extending the existing message.

doc/user_guide/checkers/features.rst
Pierre-Sassoulas: There's https://github.com/PyCQA/pylint/pull/6736 to fix this issue, but we need to make the doc generated the same on each interpreter / operating system first (some default value need to be changed to "system's interpreter"n or "system's installed enchant's dict".

doc/data/messages/u/use-dict-literal/details.rst
hofrob: As discussed, I added the python version and removed the exact percentage.

doc/data/messages/u/use-dict-literal/good.py
hofrob: Ah, now I get it, thanks!

I replace the `dict(some_method())` example with:

```py
# shallow copy a dict
new_dict = {**another_dict}
```

doc/user_guide/checkers/features.rst
hofrob: > The current description is not completely identical to the one in `refactoring_checker.py`

True, fixed!

hofrob:
I'm still unsure about the wording in `doc/data/messages/u/use-dict-literal/details.rst`. I'll gladly change this if you have something else in mind.

Btw. because I was curious I did a similar performance test with list joining compared to string concatenation: https://gist.github.com/hofrob/95b0d31ee4e168b876ecf82c92db2df4

Since we're using this in when creating the suggestion, I will leave it is.

doc/whatsnew/fragments/7690.new_check
```diff
@@ -0,0 +1,3 @@
+Extended ``use-dict-literal`` to cover calls to ``dict()`` that could be replaced with a literal dict.
```
Pierre-Sassoulas, line 1:
```suggestion
Extended ``use-dict-literal`` to also warn about call to ``dict()`` when passing arguments or keyword arguments.
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1584,15 +1585,48 @@ def _check_consider_using_with(self, node: nodes.Call) -> None:
...
-        if node.as_string() in {"list()", "dict()"}:
+    def _check_use_list_literal(self, node: nodes.Call) -> None:
+        """Check if empty list is created by using the literal []."""
+        if node.as_string() == "list()":
             inferred = utils.safe_infer(node.func)
             if isinstance(inferred, nodes.ClassDef) and not node.args:
                 if inferred.qname() == "builtins.list":
                     self.add_message("use-list-literal", node=node)
-                elif inferred.qname() == "builtins.dict" and not node.keywords:
-                    self.add_message("use-dict-literal", node=node)
```
Pierre-Sassoulas, line 1595: Creating a "dict specific" function here would be better performance wise while keeping the separation between list/dict clear for reader.

pylint/pyreverse/vcg_printer.py
```diff
@@ -154,20 +154,29 @@
     NodeType.CLASS: "box",
     NodeType.INTERFACE: "ellipse",
 }
+# pylint: disable-next=consider-using-namedtuple-or-dataclass
```
Pierre-Sassoulas, line 157: Should we use namedtuple or dataclass @DudeNr33 ? 😄 It seems like it, but I don't know how hard it would be to do that in pyreverse. Should we create another issue to fix that later ?

tests/functional/u/use/use_literal_dict.txt
```diff
@@ -1 +1,7 @@
-use-dict-literal:3:4:3:10::Consider using {} instead of dict():UNDEFINED
+use-dict-literal:3:4:3:10::Consider using '{}' instead.:INFERENCE
+use-dict-literal:4:4:4:28::"Consider using '{""a"": '1', ""b"": None, ""c"": 3}' instead.":INFERENCE
+use-dict-literal:8:4:8:13::Consider using '{**x}' instead.:INFERENCE
+use-dict-literal:13:4:13:19::"Consider using '{""foo"": bar()}' instead.":INFERENCE
+use-dict-literal:17:0:24:1::"Consider using '{""suggestions"": list(bar(boo=True)), **baz}' instead.":INFERENCE
+use-dict-literal:31:0:34:1::"Consider using '{""url"": '/foo', **inst.prop}' instead.":INFERENCE
+use-dict-literal:36:0:46:1::"Consider using '{""Lorem"": 'ipsum', ""dolor"": 'sit', ""amet"": 'consectetur', ""adipiscing"": 'elit', …' instead.":INFERENCE
```
Pierre-Sassoulas, line 7: There's an issue here, the final '}' is missing.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1584,15 +1585,48 @@ def _check_consider_using_with(self, node: nodes.Call) -> None:
...
+                elements.append(f'"{keyword.arg}": {keyword.value.as_string()}')
+        for keyword in node.kwargs:
+            if len(", ".join(elements)) >= 64:
+                break
+            elements.append(f"**{keyword.value.as_string()}")
+        suggestion = f'{{{", ".join(elements)}}}'
+        if len(suggestion) > 64:
+            suggestion = f"{suggestion[:-1]}, …"
+
+        return suggestion
```
Pierre-Sassoulas, line 1629:
```suggestion
        suggestion = f'{", ".join(elements)}'
        return f"{{{suggestion}{', … '  if len(suggestion) > 64 else ''}}}"
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1074,7 +1074,8 @@ def visit_call(self, node: nodes.Call) -> None:
         self._check_super_with_arguments(node)
         self._check_consider_using_generator(node)
         self._check_consider_using_with(node)
-        self._check_use_list_or_dict_literal(node)
+        self._check_use_list_literal(node)
+        self._check_use_dict_literal(node)
```
Pierre-Sassoulas, line 1078: The call to ``inferred = utils.safe_infer(node.func)`` is pretty costly, I think we should keep list/dict together for better performance.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -464,9 +464,9 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "The literal is faster as it avoids an additional function call.",
         ),
         "R1735": (
-            "Consider using {} instead of dict()",
+            "Consider using '%s' instead.",
```
Pierre-Sassoulas, line 467: Yet another change of mind for clarity, let's hear what other maintainer think before applying :) 

```suggestion
            "Consider using '%s' instead of a call to 'dict'.",
```

Pierre-Sassoulas (changes requested):
This is looking better and better 🔥

pylint/pyreverse/vcg_printer.py
DudeNr33: I will create a separate issue for this. 👍

DudeNr33:
> I'm still unsure about the wording in `doc/data/messages/u/use-dict-literal/details.rst`. I'll gladly change this if you have something else in mind.
> 
> Btw. because I was curious I did a similar performance test with list joining compared to string concatenation: https://gist.github.com/hofrob/95b0d31ee4e168b876ecf82c92db2df4
> 
> Since we're using this in when creating the suggestion, I will leave it is.

Sorry for the confusion - I actually liked that concrete number here, as it showed that we are not just talking about a 0.x % increase. I simply meant to add the Python version *additionally*, so we know in the future to what version this benchmark referred to.  
Apart from that no more comments from my side. 😊 👍

DudeNr33 (dismissed):

hofrob:
Sorry for the delay! I'm dealing with separate things for the moment.

Thanks for the feedback for now! I hope I'll be able to take another look in the course of the next week.

pylint/checkers/refactoring/refactoring_checker.py
hofrob: I don't think we will prevent any calls to `utils.safe_infer` in this case. The list-check `_check_use_list_literal` will _only_ do it if `node.as_string() == "list()"` and `_check_use_dict_literal` will _always_ do it if `isinstance(node.func, astroid.Name)`¹.

The only thing we could improve are the checks `_check_use_dict_literal`¹, right? Doing both in one method will not have any effect. Or will it?

¹: I added a check there now: `node.as_string().startswith("dict(")`

hofrob: > ¹: I added a check there now: node.as_string().startswith("dict(")

That "solution" sucks. I will do `node.func.name == 'dict'`.

hofrob: Of course, this won't work for the following code:

```py
foo = dict
new_dict = foo(asdf=1)
```

But I think we can live with it for performance reasons?

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1584,15 +1585,45 @@ def _check_consider_using_with(self, node: nodes.Call) -> None:
...
+            self.add_message(
+                "use-dict-literal",
+                args=(self._dict_literal_suggestion(node),),
+                node=node,
+                confidence=INFERENCE,
+            )
+
+    @staticmethod
+    def _dict_literal_suggestion(node: nodes.Call) -> str:
+        elements: list[str] = []
```
Pierre-Sassoulas, line 1615:
```suggestion
        """Return a suggestion of reasonable length."""
        elements: list[str] = []
```

pylint/checkers/refactoring/refactoring_checker.py
Pierre-Sassoulas: > I don't think we will prevent any calls to utils.safe_infer in this case. The list-check _check_use_list_literal will only do it if node.as_string() == "list()"

Fair point, I missed that.

> Of course, this won't work for the following code:
> 
> foo = dict
> new_dict = foo(asdf=1)
> 
> But I think we can live with it for performance reasons?


Yes definitely. Let's not punish everyone with slower linter to be accurate for those who redefine dict. They don't deserve a linter anyway 😄

Pierre-Sassoulas (approved):
The primer looks pretty good now, there's a case when the last argument under a certain length is very long that does not get truncated:
```
Consider using '{"by": dict(((key, reverse_resolve_tag_value(self._use_case_id, self._organization_id, value, weak=True)) if groupby_alias_to_groupby_column.get(key) not in NON_RESOLVABLE_TAG_VALUES else (key, value) for (key, value) in tags)), … }' instead of a call to 'dict'.
```

But I think it's acceptable.

Great first contribution @hofrob, this is definitely going in 2.16.0 :)

Pierre-Sassoulas:
This test fail under Windows seems genuine.

hofrob:
> This test fail under Windows seems genuine.

Ah yes. This may have failed because of the ellipsis character `…` in `features.rst` since it happened during this command: `command = ['--rcfile=d:\\a\\pylint\\pylint\\pylint\\testutils\\testing_pylintrc', '--full-documentation']`

I replaced it with dots `...` now. Can't really run the test locally, so let's see what the pipeline is going to say.

hofrob:
FYI: I replaced all ellipsis characters with dots since there seems to be an issue with windows.

Details: When running the test `test_self` with the `--full-documentation` arg on windows, the following happens and no docs are returned: `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x85 in position 47771: invalid start byte`

Pierre-Sassoulas:
> Details: When running the test test_self with the --full-documentation arg on windows, the following happens and no docs are returned: 

Ha, right, probably the same encoding issue than with emojis in #7612.

DanielNoord (approved):
Love this change! Thanks very much!

PR #7697, opened 2022-11-01
Add flake8 bugbear to pre-commit

Pierre-Sassoulas:
## Type of Changes

## Description

I realized that flake8-bugbear was configured in astroid but not pylint when upgrading pre-commit. Generally it's the other way around. There were some fixes to do, some of them valuable.

### DIFF
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -529,7 +529,7 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
             tolineno = node.tolineno
         assert tolineno, node
         lines: list[str] = []
-        for line in range(line, tolineno + 1):
+        for line in range(line, tolineno + 1):  # noqa: B020
             self._visited_lines[line] = 1
             try:
                 lines.append(self._lines[line].rstrip())
--- a/pylint/checkers/format.py
+++ b/pylint/checkers/format.py
@@ -676,15 +676,11 @@ class FormatChecker(BaseTokenChecker, BaseRawFileChecker):
             if tokens.type(line_start) != tokenize.STRING:
                 self.check_trailing_whitespace_ending(line, lineno + offset)
 
-        # hold onto the initial lineno for later
-        potential_line_length_warning = False
-        for offset, line in enumerate(split_lines):
-            # this check is purposefully simple and doesn't rstrip
-            # since this is running on every line you're checking it's
-            # advantageous to avoid doing a lot of work
-            if len(line) > max_chars:
-                potential_line_length_warning = True
-                break
+        # This check is purposefully simple and doesn't rstrip since this is running
+        # on every line you're checking it's advantageous to avoid doing a lot of work
+        potential_line_length_warning = any(
+            len(line) > max_chars for line in split_lines
+        )
 
         # if there were no lines passing the max_chars config, we don't bother
         # running the full line check (as we've met an even more strict condition)
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -835,7 +835,7 @@ class StringConstantChecker(BaseTokenChecker, BaseRawFileChecker):
     def process_string_token(self, token: str, start_row: int, start_col: int) -> None:
         quote_char = None
         index = None
-        for index, char in enumerate(token):
+        for index, char in enumerate(token):  # noqa: B007
             if char in "'\"":
                 quote_char = char
                 break
--- a/pylint/checkers/strings.py
+++ b/pylint/checkers/strings.py
@@ -834,16 +834,16 @@ class StringConstantChecker(BaseTokenChecker, BaseRawFileChecker):
 
     def process_string_token(self, token: str, start_row: int, start_col: int) -> None:
         quote_char = None
-        index = None
-        for index, char in enumerate(token):  # noqa: B007
+        for _index, char in enumerate(token):
             if char in "'\"":
                 quote_char = char
                 break
         if quote_char is None:
             return
-
-        prefix = token[:index].lower()  # markers like u, b, r.
-        after_prefix = token[index:]
+        # pylint: disable=undefined-loop-variable
+        prefix = token[:_index].lower()  # markers like u, b, r.
+        after_prefix = token[_index:]
+        # pylint: enable=undefined-loop-variable
         # Chop off quotes
         quote_length = (
             3 if after_prefix[:3] == after_prefix[-3:] == 3 * quote_char else 1
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1584,7 +1584,10 @@ accessed. Python regular expressions are accepted.",
 
         # 3. Match the **kwargs, if any.
         if node.kwargs:
-            for i, [(name, defval), assigned] in enumerate(parameters):
+            # TODO: It's possible to remove this disable by using dummy-variables-rgx
+            #  see https://github.com/PyCQA/pylint/pull/7697#discussion_r1010832518
+            # pylint: disable-next=unused-variable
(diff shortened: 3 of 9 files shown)


pylint/checkers/format.py
```diff
@@ -676,15 +676,11 @@ def check_lines(
...
-            # since this is running on every line you're checking it's
-            # advantageous to avoid doing a lot of work
-            if len(line) > max_chars:
-                potential_line_length_warning = True
-                break
+        # This check is purposefully simple and doesn't rstrip since this is running
+        # on every line you're checking it's advantageous to avoid doing a lot of work
+        potential_line_length_warning = any(
+            len(line) > max_chars for line in split_lines
+        )
```
Pierre-Sassoulas, line 683: This is a false negative for the for_any_all checker imo.

pylint/checkers/strings.py
```diff
@@ -834,16 +834,16 @@ def check_for_concatenated_strings(
 
     def process_string_token(self, token: str, start_row: int, start_col: int) -> None:
         quote_char = None
-        index = None
-        for index, char in enumerate(token):
+        _index = None
```
DanielNoord, line 837: Why are we even initialising `_index` here?

pylint/checkers/format.py
```diff
@@ -529,7 +529,7 @@ def visit_default(self, node: nodes.NodeNG) -> None:
             tolineno = node.tolineno
         assert tolineno, node
         lines: list[str] = []
-        for line in range(line, tolineno + 1):
+        for line in range(line, tolineno + 1):  # noqa: B020
```
DanielNoord, line 532: What's the warning here?

pylint/checkers/typecheck.py
```diff
@@ -1584,7 +1584,8 @@ def visit_call(self, node: nodes.Call) -> None:
 
         # 3. Match the **kwargs, if any.
         if node.kwargs:
-            for i, [(name, defval), assigned] in enumerate(parameters):
+            # pylint: disable-next=unused-variable
+            for i, [(name, _defval), _assigned] in enumerate(parameters):
```
DanielNoord, line 1588:
```suggestion
            for i, [(name, _), _] in enumerate(parameters):
```

? Or add `^_.*$` to the list of allowed unused variables in our config?

pylint/checkers/strings.py
Pierre-Sassoulas: We're implicitly using the enumerate to have the latest index value later. We could refactor to something clearer, not sure about the perf implication though.

pylint/checkers/format.py
Pierre-Sassoulas: 'Loop control variable overrides iterable it iterates' (It's by design)

DanielNoord: Are you sure? Can't `for line` be `for index`?
The iterable value just seems to be a integer that's called `line`.

pylint/checkers/strings.py
DanielNoord: You but we check `quote_char` so `_index` is always a value on L845 and L846. We don't need the `_index = None` as far as I can see.

pylint/checkers/typecheck.py
Pierre-Sassoulas: Yeah, we should accept this kind of variable (like flake8-bugbear) but this does not work. In the code we have ``IGNORED_ARGUMENT_NAMES = re.compile("_.*|^ignored_|^unused_")``  as default value for ``"ignored-argument-names"``. Look like something we should fix in another MR 😄

Pierre-Sassoulas: (Function that do not work seem to be ``_is_name_ignored``)

DanielNoord: I recently added `dummy-variables-rgx=^_.*$|dummy` to our company config and it worked like a charm 😄 

I think we're using the incorrect configuration then! Guess we need to improve the docs... 😅

pylint/checkers/strings.py
Pierre-Sassoulas: This would be a syntax error in python as _index do not leave the for loop scope if it's not defined before.

DanielNoord:
```console
❯ python
Python 3.11.0b4 (main, Jul 12 2022, 11:43:04) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> for index in range(1, 2):
...     print(index)
...
1
>>> print(index)
1
```

There is actually a `pylint` issue about this behaviour as it is both dangerous and useful 😄

pylint/checkers/typecheck.py
Pierre-Sassoulas: I had no idea this option existed and I did not find it when searching for it. No idea what ignored-argument-names is supposed to do either now 🤷 The default value need to be able to handle the standard way to signal that a variable is known to be unused..

DanielNoord: `Argument names that match this expression will be ignored.` 😅 

I guess it's `dummy-variables` but only for arguments?

pylint/checkers/strings.py
Pierre-Sassoulas:
```python
>>> for i, x in enumerate([]):
...     print(i, x)
... print(i)
  File "<stdin>", line 3
    print(i)
    ^^^^^
SyntaxError: invalid syntax
```

DanielNoord: But this can't happen due to L842 right?

Pierre-Sassoulas:
I don't have the time to fix everything, could we merge with the current disable so flake8 bugbear is applied on new code and the already working code stay as is ? In particular fixing dummy-variables-rgx seems 4 hours of work or more. Polishing working code we don't understand well means we can introduce bugs that we'll have to fix later.

pylint/checkers/typecheck.py
Pierre-Sassoulas: ``dummy-variables-rgx`` affect other checks, we get a false positive for ``pylint/extensions/docparams.py:474:4: W9019: "ignored_argument_names" useless ignored parameter documentation (useless-param-doc)`` when using the default value.

DanielNoord: Let's add a comment about this to the disable then, so we don't lose this discussion.

I think my other two comments can be easily fixed and then we can merge this 😄

pylint/checkers/strings.py
Pierre-Sassoulas: Right, I'll fix that.

Pierre-Sassoulas: We get 
```
************* Module pylint.checkers.strings

pylint/checkers/strings.py:844:24: W0631: Using possibly undefined loop variable '_index' (undefined-loop-variable)

pylint/checkers/strings.py:845:29: W0631: Using possibly undefined loop variable '_index' (undefined-loop-variable)

```
Which is probably the reason why we had the index definition in the first place.

DanielNoord: Ah, let's disable `pylint` there then. I prefer disabling tools that don't understand Python instead of adding more Python to make tools understand it.

Pierre-Sassoulas: 266c0f545

pylint/checkers/format.py
Pierre-Sassoulas: Handled in #7707

pylint/checkers/format.py
Pierre-Sassoulas: I'd rather not touch this, my goal is to add flake8 bugbear not "fix" all legacy issues (especially when the fix is not trivial).

DanielNoord (approved):

PR #7707, opened 2022-11-03
Fix false negative for-any-all

### LINKED ISSUE #7699: False negative 'for-any-all' when there is a break or a pre-initialized boolean
### Current problem

```python
def optimized_any_with_break(split_lines: list[str], max_chars: int) -> bool:
    """False negative found in https://github.com/PyCQA/pylint/pull/7697"""
    potential_line_length_warning = False
    for line in split_lines:  # [consider-using-any-or-all]
        if len(line) > max_chars:
            potential_line_length_warning = True
            break
    return potential_line_length_warning
```    
Can be simplified to 
```python
def optimized_any_with_break(split_lines: list[str], max_chars: int) -> bool:
    return any(len(line) > max_chars for line in split_lines)
```

### Desired solution

``for-any-all`` raised in this case.

### Additional context

I did not think very hard about potential false positive, but I think it make sense to not raise if ``potential_line_length_warning`` is not initialized directly above the for loop.

clavedeluna:
## Type of Changes

## Description

Closes #7699

This PR adds support for a very specific false negative case for `for-any-all`


pylint/extensions/for_any_all.py
```diff
@@ -36,20 +41,53 @@ def visit_for(self, node: nodes.For) -> None:
...
+                args=suggested_string,
+                confidence=HIGH,
             )
 
+    @staticmethod
+    def _assigned_reassigned_returned(
+        node: nodes.For, if_children: list[nodes.NodeNG], node_after_loop: nodes.NodeNG
+    ) -> bool:
+        node_before_loop = node.previous_sibling()
+        if assigned_bool(node_before_loop):
```
Pierre-Sassoulas, line 78: The indentation is pretty deep here, I would use premature return to simplify it.
```suggestion
        if not assigned_bool(node_before_loop):
            return False
```

Pierre-Sassoulas (commented):
It looks promising but I'd really want to see the primer, I don't know what went wrong.

Pierre-Sassoulas:
On second though maybe we need to fix the changelog, the primer are aborted if another job fail.

pylint/extensions/for_any_all.py
clavedeluna: oh for sure, this is gnarly code just to get it to pass the tests.  Just wanted to see what you all thought before I tried to refactor.

I'll add changelog and see primer

clavedeluna:
No primer results yet... as I mentioned in another PR I'm not able to run it locally by following the CI workflow setup. When do we predict the CI primer will be fixed?

Pierre-Sassoulas:
> When do we predict the CI primer will be fixed?

There's no planning for that, but the stdlib primer have a pretty actionable/understandable error that is specific to this merge request : https://github.com/PyCQA/pylint/actions/runs/3400193560/jobs/5654478930

Pierre-Sassoulas:
This needs to be rebased (not merged) on upstream/main for the primer to work :)

clavedeluna:
The primer results look really good, but I'm curious if anyone else thinks the black case could be wrong?

consider-using-any-or-all:
for loop could be not any(comment.value in FMT_ON for comment in list_comments(container.prefix, is_endmarker=False, preview=preview))
https://github.com/psf/black/blob/d97b7898b34b67eb3c6839998920e17ac8c77908/src/black/comments.py#L303

function is_fmt_on has an elif statment. From the docstring it also suggests to me that the author wants for full looping and to see what the last value is. So we shouldn't suggest replacing it with `any...`

pylint/extensions/for_any_all.py
```diff
@@ -36,19 +41,104 @@ def visit_for(self, node: nodes.For) -> None:
             return
 
         if_children = list(node.body[0].get_children())
-        if not len(if_children) == 2:  # The If node has only a comparison and return
-            return
-        if not returns_bool(if_children[1]):
+        if any(isinstance(child, nodes.If) for child in if_children):
+            # an if node within the if-children indicates an elif clause,
+            # suggesting complex logic.
             return
```
Pierre-Sassoulas, line 47: I don't have a deep comprehension of the problem but intuitively I suppose there could be other children than "if node" that would also be problematic ? Especially since we do not have a number of child's limit anymore. The primer result indicate I'm probably wrong but I don't understand why.

Pierre-Sassoulas (commented):
Thank you for handling the elif false positive 👍 I think we just need some coverage completion at this point.

pylint/extensions/for_any_all.py
```diff
@@ -36,19 +41,104 @@ def visit_for(self, node: nodes.For) -> None:
...
+        if not len(if_children) == 2:
+            # The If node has only a comparison and return
+            return False
+        if not returns_bool(if_children[1]):
+            return False
+
+        # Check for terminating boolean return right after the loop
+        if returns_bool(node_after_loop):
+            return True
+        return False
```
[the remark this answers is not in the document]
nickdrozd:
```suggestion
        return returns_bool(node_after_loop)
```

pylint/extensions/for_any_all.py
```diff
@@ -36,19 +41,104 @@ def visit_for(self, node: nodes.For) -> None:
...
+
+        node_before_loop_name = node_before_loop.targets[0].name
+        if (
+            first_target.name == node_before_loop_name
+            and isinstance(node_after_loop, nodes.Return)
+            and isinstance(node_after_loop.value, nodes.Name)
+            and node_after_loop.value.name == node_before_loop_name
+        ):
+            return True
+        return False
```
nickdrozd, line 141:
```suggestion
        return (
            first_target.name == node_before_loop_name
            and isinstance(node_after_loop, nodes.Return)
            and isinstance(node_after_loop.value, nodes.Name)
            and node_after_loop.value.name == node_before_loop_name
        )
```

pylint/extensions/for_any_all.py
clavedeluna: The number of child nodes is handled now within `if_statement_returns_bool` which was the original check before this PR. The reason we can't now limit to only two children as we did before is because there are legitimate cases (in tests too) where there are more than 2 children.

The entire logic in this check is very edge-case-y.

Pierre-Sassoulas (approved):
👌

Pierre-Sassoulas:
This is one of my favorite check and it's getting more powerful, which is pretty nice !

PR #7708, opened 2022-11-04
Update Primer venv caching [ci]

cdce8p:
## Description
Followup to https://github.com/PyCQA/pylint/pull/7651#issuecomment-1303397691

To prevent a `Primer / Run` failure if the hash changes, add separate step to create venv in that workflow too. With Github actions cache [scoping](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache), the cache entry can only be used for the specific PR.

Additionally, remove install of bleeding-edge astroid. That only works as long as the `main` astroid branch doesn't contain any breaking changes.

With the new cache key, it's no longer necessary to update the `CACHE_VERSION` when doing an astroid update.

/CC: @DanielNoord, @Pierre-Sassoulas

### DIFF
--- a/.github/workflows/primer_comment.yaml
+++ b/.github/workflows/primer_comment.yaml
@@ -50,7 +50,8 @@ jobs:
           path: venv
           key:
             ${{ runner.os }}-${{ steps.python.outputs.python-version }}-${{
-            env.KEY_PREFIX }}-${{ env.CACHE_VERSION }}
+            env.KEY_PREFIX }}-${{ env.CACHE_VERSION }}-${{ hashFiles('pyproject.toml',
+            'requirements_test.txt', 'requirements_test_min.txt') }}
       - name: Fail job if Python cache restore failed
         if: steps.cache-venv.outputs.cache-hit != 'true'
         run: |
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -40,12 +40,6 @@ jobs:
           python-version: ${{ matrix.python-version }}
           check-latest: true
 
-      - name: Get latest astroid commit
-        id: get-astroid-sha
-        run: |
-          curl https://api.github.com/repos/PyCQA/astroid/commits |
-          python -c "import json, sys; print(json.load(sys.stdin)[0]['sha'])" > astroid_sha.txt
-
       # Create a re-usable virtual environment
       - name: Create Python virtual environment cache
         id: cache-venv
@@ -61,8 +55,6 @@ jobs:
           . venv/bin/activate
           python -m pip install -U pip setuptools wheel
           pip install -U -r requirements_test.txt
-          # Use bleeding-edge astroid
-          pip install git+https://github.com/PyCQA/astroid.git
 
       # Cache primer packages
       - name: Get commit string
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas (approved):
It make sense to remove astroid bleeding edge imo, as we're not running on astroid bleeding edge anywhere else so it's confusing. Maybe we could run the pylint primer on astroid bleeding edge directly in astroid's CI. I don't know on with which pylint, as long as there are no breaking change in astorid the latest published pylint is the best one but it's hard to know what to use during the time where astroid has breaking change and bleeding edge pylint is not ready for it yet.

Let's wait for primer experts review (@jacobtylerwalls , @DanielNoord ) before merging :)

jacobtylerwalls:
My intent behind installing astroid bleeding edge in the primer was that it provides the greatest opportunity for finding new false positives caused in pylint PRs. It would be difficult to look through the haystack of closed PRs for examples, but I thought I saw potential regressions on PRs that were only evident when using astroid bleeding edge. I also wasn't too worried about breaking changes, because breaking-changes-on-main-causing-astroid-error == breaking-changes-on-pr-causing-astroid-error, but I guess that has the potential to hide some false positives, outweighing the original motivation. I guess it comes down to how often we forecast having breaking changes in astroid versus having regressions on pylint PRs we could otherwise catch.

cdce8p:
> Maybe we could run the pylint primer on astroid bleeding edge directly in astroid's CI.

That would be awesome. At the moment, I manually check bigger astroid PRs against Home Assistant just to have a first sanity check.

> My intent behind installing astroid bleeding edge in the primer was that it provides the greatest opportunity for finding new false positives caused in pylint PRs. It would be difficult to look through the haystack of closed PRs for examples, but I thought I saw potential regressions on PRs that were only evident when using astroid bleeding edge.

Not so sure about this one but maybe I'm missing something. Isn't the only case this would be useful if both an unreleased astroid change and some potential pylint change that uses the astroid one come together? Even then, we'll need to bisect astroid either way. Idk tbh, this seems like a fairly remote case. Without testing bleeding edge, it would show up in the `astroid` bump PR which I think I would prefer.

Either way the best workaround is probably to do frequent astroid releases if there are any new features. We have gotten a lot better with those lately. The new release process has definitely helped a lot.

jacobtylerwalls:
> Without testing bleeding edge, it would show up in the astroid bump PR which I think I would prefer.

That's fine, it's just the case that the original contributor would be long gone. Unless...

> Either way the best workaround is probably to do frequent astroid releases if there are any new features. We have gotten a lot better with those lately. The new release process has definitely helped a lot.

👍🏻 

> Isn't the only case this would be useful if both an unreleased astroid change and some potential pylint change that uses the astroid one come together?

Yes, but without the "use" of the astroid feature being intended. As astroid gets more powerful, more `Uninferables` turn into real nodes, and this increases the surface area of the primer. So someone's pylint PR for a small positive in something completely unrelated crashes when something obscure in `flask` is newly inferred correctly as a lambda and doesn't have a `.__qname__` that the pylint contributor was relying on, and we get the benefit of knowing without waiting for the astroid upgrade. The idea was to have the largest possible surface area for the primer at the moment we have the contributor's attention.

Ultimately I think you have a good point--the faster release cadence is a good reason to remove astroid bleeding edge in the primer. Was just adding some context.

DanielNoord (approved):

PR #7712, opened 2022-11-05
[doc] Add examples for wrong-spelling-in-x

Pierre-Sassoulas:
## Type of Changes

## Description

We can't document ``"invalid-characters-in-docstring"`` in docstring right now because this is raised in case of enchant error and it's hard to trigger. This should be treated like a fatal error for documentation purpose.


mbyrnepr2 (approved):
Looks good!

PR #7713, opened 2022-11-05
[doc] Document some messages where the code is not the problem

Pierre-Sassoulas:
## Type of Changes

## Description

Some of the time there is no good or bad example, so we need to document this in a peculiar way.


mbyrnepr2 (approved):
Looks 👌

PR #7716, opened 2022-11-05
add confidence levels to all exception checkers

### LINKED ISSUE #7715: Add confidence levels to exceptions checkers
### Current problem

Messages in [exceptions.py](https://github.com/PyCQA/pylint/blob/main/pylint/checkers/exceptions.py) can use confidence level.

Of course so many other locations need it too, but doing so iteratively makes the most sense.

### Desired solution

Update code. I have a commit for this so just creating a ticket to link to.

### Additional context

_No response_

clavedeluna:
## Type of Changes

## Description

While working on adding the new `broad-exception-raised` checker, I was able to work on adding confidence levels to other checkers in the same file

Don't think this needs changelog but lmk if you disagree.

Closes #7715

### DIFF
--- a/pylint/checkers/exceptions.py
+++ b/pylint/checkers/exceptions.py
@@ -198,7 +198,9 @@ class ExceptionRaiseRefVisitor(BaseVisitor):
 
     def visit_name(self, node: nodes.Name) -> None:
         if node.name == "NotImplemented":
-            self._checker.add_message("notimplemented-raised", node=self._node)
+            self._checker.add_message(
+                "notimplemented-raised", node=self._node, confidence=HIGH
+            )
         elif node.name in OVERGENERAL_EXCEPTIONS:
             self._checker.add_message(
                 "broad-exception-raised",
@@ -217,7 +219,9 @@ class ExceptionRaiseRefVisitor(BaseVisitor):
         ):
             msg = node.args[0].value
             if "%" in msg or ("{" in msg and "}" in msg):
-                self._checker.add_message("raising-format-tuple", node=self._node)
+                self._checker.add_message(
+                    "raising-format-tuple", node=self._node, confidence=HIGH
+                )
 
 
 class ExceptionRaiseLeafVisitor(BaseVisitor):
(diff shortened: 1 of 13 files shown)


pylint/checkers/exceptions.py
```diff
@@ -217,15 +219,20 @@ def visit_call(self, node: nodes.Call) -> None:
...
     """Visitor for handling leaf kinds of a raise value."""
 
     def visit_const(self, node: nodes.Const) -> None:
         self._checker.add_message(
-            "raising-bad-type", node=self._node, args=node.value.__class__.__name__
+            "raising-bad-type",
+            node=self._node,
+            args=node.value.__class__.__name__,
+            confidence=INFERENCE,
         )
```
clavedeluna, line 236: I believe it's inference because this class gets called with
`ExceptionRaiseLeafVisitor(self, node).visit(inferred)`

Pierre-Sassoulas (changes requested):
If all the message really have the confidence set now there's some action we can take:

- [ ] Add a changelog entry telling user that confidence can be used reliably now
- [ ] Enforce the confidence with a CI check (no more ``UNDEFINED`` in functional tests ?)
- [ ] (Deprecate the default value for confidence in ``add_message`` and enforce it in 3.0 or 4.0 ?) 

Only the first step would need to be done in this MR. What do you think @DanielNoord ?

clavedeluna:
> * [ ]  Enforce the confidence with a CI check (no more `UNDEFINED` in functional tests ?)

I assumed the changes in functional tests are sufficient?

> * [ ]  (Deprecate the default value for confidence in `add_message` and enforce it in 3.0 or 4.0 ?)

interesting asks here, I've definitely added confidence levels in some other PRs and we didn't decide to do this.

just bouncing ideas.

Pierre-Sassoulas:
Sorry if it wasn't clear @clavedeluna I'm just thinking of the next step after this MR. I should have done that in the issue you opened.  I did not think we would already have all confidences set so early :)

DanielNoord:
I think it makes sense to enforce no `UNDEFINED` for our own messages. Not sure if we should enforce it in `add_message` and therefore in plugins.

Pierre-Sassoulas:
I created the automated test locally, it seems we're still far from having set the confidence for all the messages.

Pierre-Sassoulas (approved):

clavedeluna:
Having a CI check to make sure there's always a confidence seems like a huge win, tho yes there's still quite a few calls to send_message without it.

Pierre-Sassoulas:
I misunderstood the MR and the issue I thought you added confidence everywhere, not only in the exceptions checker 😅 I pushed the branch with the enforced test here: https://github.com/PyCQA/pylint/tree/enforce-the-confidence-in-add-message

PR #7723, opened 2022-11-06
[compare-to-zero] More actionnable and understandable message

Pierre-Sassoulas:
## Type of Changes

## Description

Originally done in order to normalize the use-implicit-booleaness check. This make the message more understandable and permit to make #6870 more reviewable. Also fix a false positive with ``x is False`` being falsely detected as a comparison to 0.


Pierre-Sassoulas:
Drafting as the primer detects false positives where ``False`` is mistaken for a ``0`` here.

mbyrnepr2 (approved):
👍

PR #7725, opened 2022-11-07
More descriptive message for ``use-implicit-booleaness-not-comparison``

Pierre-Sassoulas:
## Type of Changes

## Description

Follow-up to #7723 in order to make #6870 more reviewable.

### DIFF
--- a/pylint/checkers/refactoring/implicit_booleaness_checker.py
+++ b/pylint/checkers/refactoring/implicit_booleaness_checker.py
@@ -51,7 +51,6 @@ class ImplicitBooleanessChecker(checkers.BaseChecker):
     * comparison such as variable != empty_literal:
     """
 
-    # configuration section name
     name = "refactoring"
     msgs = {
         "C1802": (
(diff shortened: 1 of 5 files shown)


mbyrnepr2:
Are you concerned about the first two Sentry primer messages @Pierre-Sassoulas or can we ignore/they are unrelated?

pylint/checkers/refactoring/implicit_booleaness_checker.py
```diff
@@ -197,23 +196,32 @@ def _check_use_implicit_booleaness_not_comparison(
...
-        if isinstance(literal_node, nodes.List):
-            collection_literal = "[]"
-        if isinstance(literal_node, nodes.Tuple):
-            collection_literal = "()"
+        description = self._get_node_description(literal_node)
+        collection_literal = {
+            "list": "[]",
+            "tuple": "()",
+            "dict": "{}",
+        }.get(description, "iterable")
```
mbyrnepr2, line 216: I wonder would it be best to factor this logic out and into the `_get_node_description` method? It is only used in this one place.

Pierre-Sassoulas: You're right, it does not make a lot of sense right now but I'm going to use it use-implicit-booleaness-not-len later :)

mbyrnepr2 (approved):
You are on 🔥🔥 @Pierre-Sassoulas

PR #7726, opened 2022-11-07
[compare-to-empty-string] More actionnable and understandable message

### LINKED ISSUE #5234: C1901 (compare-to-empty-string) message is not explicit about the problem and the solution

Pierre-Sassoulas:
## Type of Changes

## Description

Originally done in order to normalize the use-implicit-booleaness check. This make the message more understandable and permit to make https://github.com/PyCQA/pylint/pull/6870 more reviewable. We have to use " as string delimiter because node as string have ' as string delimiter.

Closes #5234 (the changelog will be added in #6870)

### DIFF
--- a/pylint/extensions/emptystring.py
+++ b/pylint/extensions/emptystring.py
@@ -20,18 +20,10 @@ if TYPE_CHECKING:
 
 
 class CompareToEmptyStringChecker(checkers.BaseChecker):
-    """Checks for comparisons to empty string.
-
-    Most of the time you should use the fact that empty strings are false.
-    An exception to this rule is when an empty string value is allowed in the program
-    and has a different meaning than None!
-    """
-
-    # configuration section name
     name = "compare-to-empty-string"
     msgs = {
         "C1901": (
-            "Avoid comparisons to empty string",
+            '"%s" can be simplified to "%s" as an empty string is falsey',
             "compare-to-empty-string",
             "Used when Pylint detects comparison to an empty string constant.",
         )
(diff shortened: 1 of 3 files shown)


pylint/extensions/emptystring.py
```diff
@@ -41,31 +33,45 @@ class CompareToEmptyStringChecker(checkers.BaseChecker):
...
             op_3: nodes.NodeNG | None = ops[ops_idx + 2]
             error_detected = False
-
+            if op_1 is None or op_3 is None or op_2 not in _operators:
+                continue
+            node_name = ""
             # x ?? ""
-            if utils.is_empty_str_literal(op_1) and op_2 in _operators:
+            if utils.is_empty_str_literal(op_1):
                 error_detected = True
```
mbyrnepr2, line 61: Are these `error_detected` values necessary; could we instead rely on `if node_name` further down?

pylint/extensions/emptystring.py
```diff
@@ -41,31 +33,45 @@ class CompareToEmptyStringChecker(checkers.BaseChecker):
...
         iter_ops = iter(ops)
         ops = list(itertools.chain(*iter_ops))  # type: ignore[arg-type]
-
         for ops_idx in range(len(ops) - 2):
             op_1: nodes.NodeNG | None = ops[ops_idx]
             op_2: str = ops[ops_idx + 1]  # type: ignore[assignment]
             op_3: nodes.NodeNG | None = ops[ops_idx + 2]
             error_detected = False
-
+            if op_1 is None or op_3 is None or op_2 not in _operators:
```
mbyrnepr2, line 56:
```suggestion
            if None in [op_1, op_3] or op_2 not in _operators:
```

pylint/extensions/emptystring.py
mbyrnepr2: Of course the empty string case :D

mbyrnepr2 (dismissed):
I like it. One small comment about readability which may make it a bit simpler.

pylint/extensions/emptystring.py
Pierre-Sassoulas: mypy does not understand this I'm going to revert

mbyrnepr2: [mypy has an issue for it.](https://github.com/python/mypy/issues/2980)

mbyrnepr2 (approved):

pylint/extensions/emptystring.py
Pierre-Sassoulas: Interesting consideration from the issue you linked: https://github.com/python/mypy/issues/2980#issuecomment-285185515

PR #7731, opened 2022-11-08
Docs for some refactor msgs

clavedeluna:
## Type of Changes

## Description

Refs #5953

### DIFF
--- a/doc/data/messages/s/simplifiable-condition/bad.py
+++ b/doc/data/messages/s/simplifiable-condition/bad.py
@@ -1,0 +1,3 @@
+APPLES = 100
+
+bool(APPLES or False)  # [simplifiable-condition]
--- a/doc/data/messages/s/simplifiable-condition/bad.py
+++ b/doc/data/messages/s/simplifiable-condition/bad.py
@@ -1,3 +1,2 @@
-APPLES = 100
-
-bool(APPLES or False)  # [simplifiable-condition]
+def has_apples(apples) -> bool:
+    return bool(apples or False)  # [simplifiable-condition]
--- a/doc/data/messages/s/simplifiable-condition/good.py
+++ b/doc/data/messages/s/simplifiable-condition/good.py
@@ -1,3 +1,2 @@
-APPLES = 100
-
-bool(APPLES)
+def has_apples(apples) -> bool:
+    return bool(apples)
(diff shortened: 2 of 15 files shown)


doc/data/messages/s/simplifiable-condition/bad.py
```diff
@@ -0,0 +1,3 @@
+APPLES = 100
+
+bool(APPLES or False)  # [simplifiable-condition]
```
Pierre-Sassoulas, line 3:
```suggestion
def has_apples(apples) -> bool:
    return bool(apples or False)  # [simplifiable-condition]
```

When there is a boolean message I think it's better if there is not too much information about the non problematic portion of the conditional, because if we have a constant it means we could further simplify (100 is truethey, we could just return True)

doc/data/messages/s/simplifiable-condition/good.py
```diff
@@ -1 +1,3 @@
-# This is a placeholder for correct code for this message.
+APPLES = 100
+
+bool(APPLES)
```
Pierre-Sassoulas, line 3:
```suggestion
def has_apples(apples) -> bool:
    return bool(apples)
```

doc/data/messages/s/simplify-boolean-expression/bad.py
```diff
@@ -0,0 +1,4 @@
+HAVE_APPLES = True
+HAVE_ORANGES = False
+
+result = HAVE_APPLES and False or HAVE_ORANGES  # [simplify-boolean-expression]
```
Pierre-Sassoulas, line 4:
```suggestion
def has_oranges(oranges, apples=None) -> bool:
    return apples and False or oranges # [simplify-boolean-expression]
```

doc/data/messages/s/simplify-boolean-expression/good.py
```diff
@@ -1 +1,4 @@
-# This is a placeholder for correct code for this message.
+HAVE_APPLES = True
+HAVE_ORANGES = False
+
+result = HAVE_ORANGES
```
Pierre-Sassoulas, line 4:
```suggestion
def has_oranges(oranges) -> bool:
    return bool(oranges)
```

Pierre-Sassoulas (changes requested):
Thank you !

Pierre-Sassoulas (approved):

PR #7732, opened 2022-11-08
[github action] Add python 3.12 to the workflow

Pierre-Sassoulas:
## Type of Changes

## Description

Work in progress to handle python 3.12, [the alpha-1 release was publied by github today](https://github.com/actions/python-versions/releases/tag/3.12.0-alpha.1-3420599852).

### DIFF
--- a/.github/workflows/primer-test.yaml
+++ b/.github/workflows/primer-test.yaml
@@ -30,7 +30,7 @@ jobs:
     timeout-minutes: 5
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


cdce8p:
Let's at least wait until the beta starts. During alpha almost everything can still change. A CI run wouldn't tell us much.
In contrast it might even hurt. We discover a new failure, start fixing it but then it gets changed or reverted in CPython and we can do it all over again.

Pierre-Sassoulas:
Let's close, there isn't that much done here, no reason to clutter the PR list. I created #7741

PR #7743, opened 2022-11-10
Do not report `consider_iterating_dictionary` if bitwise operations are used

### LINKED ISSUE #5478: False-positive for consider-iterating-dictionary
### Bug description

In the following example, line 4 is reported as consider-iterating-dictionary. However, without the .keys() the expression raises a TypeError. Please note that line 6 is _not_ reported, this seems to only be triggered with an `in`.

```python
a_dict = {"a": 1, "b": 2, "c": 3}
a_set = {"c", "d"}

if "a" in a_dict.keys() | a_set:  # reports consider-iterating-dictionary
	pass
print(a_dict.keys() | a_set)      # not reported
print(a_dict | a_set)             # TypeError```
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
test.py:4:10: C0201: Consider iterating the dictionary directly instead of calling .keys() (consider-iterating-dictionary)
```

### Expected behavior

`if "a" in a_dict.keys() | a_set` should not be reported, as the proposed fix is a TypeError.

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.9.9 (main, Nov 17 2021, 16:30:15) 
[GCC 10.2.1 20210110]
```

clavedeluna:
## Type of Changes

## Description

Closes #5478

Issue provided one test case when using `.keys()` is valid and pylint should not warn for `consider-iterating-dictionary`. I then realized this is likely a special case for bitwise operations so I added a couple more test cases.

The fix is simple, and I _don't think_ there will be any unexpected outcomes, given this is well tested and all other tests passed.


clavedeluna:
oh I guess it's worth adding confidence too...

doc/whatsnew/fragments/5478.bugfix
```diff
@@ -0,0 +1,3 @@
+``consider-iterating-dictionary`` will no longer be raised if used bitwise operations are used.
```
Pierre-Sassoulas, line 1:
```suggestion
``consider-iterating-dictionary`` will no longer be raised if bitwise operations are used.
```

Pierre-Sassoulas (changes requested):
LGTM

Pierre-Sassoulas (approved):

PR #7745, opened 2022-11-10
add ``redundant-union-assign-typehint`` checker for duplicated assign typehints

### LINKED ISSUE #7636: Check for duplication in union type annotations
### Current problem

Currently we emit nothing about the following pattern, which can be simplified:

```python
ANSWER: int | int = 42
```

### Desired solution

Emit a warning when this pattern is detected. A union typehint containing the same type multiple times can be simplified. Potential name: `typehint-duplicate-union-type`.

### Additional context

This is the same issue as reported in https://github.com/python/mypy/issues/13448 although it doesn't look like it will be implemented by the type-checker.

orSolocate:
## Type of Changes

## Description

A new checker for the assign case of duplicated union type hint.

Takes care of:
1. Legacy Union e.g. `Union[int, int]`
2. Legacy Optional with None case i.e. `Optional[None]`
3. New BinOp operation e.g. `int | int`
4. Multiple type hints e.g. `Union[int | int | int`, `int | int | int`, in this case only 1 message is emitted
5. Support subscripts e.g. `list[int] | list[int]`

Closes #7636


pylint/extensions/typing.py
```diff
@@ -124,6 +124,11 @@ class TypingChecker(BaseChecker):
             "Python 3.9.0 and 3.9.1. Use ``typing.Callable`` for these cases instead. "
             "https://bugs.python.org/issue42965",
         ),
+        "R6006": (
+            "Type-hint `%s` is used more than once in assignment. Consider removing redundant typehints. ",
+            "redundant-union-assign-typehint",
```
Pierre-Sassoulas, line 129:
```suggestion
            "duplicated-type-in-union",
```
(This is a suggestion, I'm trying to make the name clearer)

tests/functional/ext/typing/redundant_union_assign_typehint.rc
```diff
@@ -0,0 +1,6 @@
+[main]
+py-version=3.10
```
Pierre-Sassoulas, line 2: In my opinion we should separate code that can run in python 3.7 (``Union[int, int]``) from code that need python 3.8 with ``__future__``annotation and python 3.10 specific code. 3.7 is going to be used for a long time. In order to do that we can have multiple file with multiple py-version conf.

Pierre-Sassoulas (commented):
Look good. I skimmed the code and checked description / tests for a start :)

pylint/extensions/typing.py
orSolocate: maybe `duplicated-union_typehint` ?

tests/functional/ext/typing/redundant_union_assign_typehint.rc
orSolocate: you mean like having 2 different test names each with its own `rc` file?

Pierre-Sassoulas: Yes with one suffixed ``_py310`` with only python 3.10 specific checks, one suffixed ``_py38``... etc.

pylint/extensions/typing.py
Pierre-Sassoulas: ``duplicated-typehint-in-union`` ?

mbyrnepr2: How about we make it generic and similar to what the docs say:
`redundant-typehint-argument`

[_Redundant arguments are skipped_](https://docs.python.org/3/library/typing.html#typing.Union)

orSolocate: @mbyrnepr2  - nice idea to check it out mate.  Do you suggest it both for assigns and functional arguments? 
I will explain myself - This checker only covers assignments for now. After this is closed I would extend it for function arguments, do you think they both should have the same message?

tests/functional/ext/typing/redundant_union_assign_typehint.rc
orSolocate: Done. I like the new version of tests, I removed the `py_version` from the `rc` file because I saw that the fake version made python 3.9 tests pass even though they should have failed (before the tests were fixed). I figured all the `#pylint disable` is not needed anymore too :)

pylint/extensions/typing.py
mbyrnepr2: Yeah I think the name of the message should be the same in all cases. So, if the duplication occurs in an assignment or function parameter etc. it is the same issue in all cases.

orSolocate:
@Pierre-Sassoulas  - could you tell why the `Primer / Run / Run / 3.10 (pull_request)` test is failing? does it just need a re-run?

DanielNoord:
Nobody has found the time to investigate this, but there seems to be something going wrong with the caching of projects in the primer.
A rerun won't fix it sadly

Pierre-Sassoulas:
The primer were fixed on main thank to @jacobtylerwalls, you can rebase/merge to fix the pipeline :)

jacobtylerwalls:
> The primer were fixed on main thank to @jacobtylerwalls, you can rebase/merge to fix the pipeline :)

The primer was also broken thanks to me, so I'm back to the starting line :P

orSolocate:
@jacobtylerwalls  thank you so much brother! 

If anyone could review I would be grateful. I am not happy with the message description.. I would like to have the same message for both assign and function arguments case.

cdce8p:
> @jacobtylerwalls thank you so much brother!
> 
> If anyone could review I would be grateful. I am not happy with the message description.. I would like to have the same message for both assign and function arguments case.

I'm working on a full review. For the message, I would suggest `duplicate-union-item`. That will work for all cases and is easy enough to understand.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
         if self._should_check_callable and node.attrname == "Callable":
             self._check_broken_callable(node)
 
+    @only_required_for_messages("redundant-union-assign-typehint")
+    def visit_annassign(self, node: nodes.AnnAssign) -> None:
```
cdce8p, line 228: At the moment, almost all logic is inside this function. That will make it difficult to reuse for function arguments. Consider moving the logic to a new _private_ helper function.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        type_annotations = list(node.annotation.get_children())
```
cdce8p, line 229: I don't like the use of `get_children()`. It's easy to miss a case when accessing children by index and much more difficult to understand compared to normal attribute access.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        if not type_annotations:
+            return
```
cdce8p, line 231: This should probably be a check if `node.annotation` is a `nodes.Subscript`.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        if (
+            isinstance(type_annotations[0], nodes.Name)
+            and type_annotations[0].name in UNION_NAMES
+        ):
```
cdce8p, line 235: What I mentioned earlier, it's difficult to follow what `type_annotations[0]` is referring to. `node.annotation.value` would be better.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+            if isinstance(type_annotations[-1], nodes.Tuple):
+                types = list(type_annotations[-1].get_children())
```
cdce8p, line 237: Similarly here with `node.annotation.slice` instead of `type_annotations[-1]`.

For `types`, just use the `node.elts` attribute instead.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+            elif type_annotations[0].name == "Optional":
+                if (
+                    isinstance(type_annotations[-1], nodes.Const)
+                    and type_annotations[-1].pytype() == "builtins.NoneType"
```
cdce8p, line 241: `Const` nodes have a `value` attribute -> `node.value == None` is enough.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                        args="None",
+                        confidence=HIGH,
+                    )
+                return
+        elif self._should_check_alternative_union_syntax and isinstance(
+            node.annotation, nodes.BinOp
+        ):
+            types = self.parse_binops_typehints(
+                node.annotation,
+            )
```
cdce8p, line 255:
```suggestion
            types = self.parse_binops_typehints(node.annotation)
```
Formatting

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+            if isinstance(type_annotations[-1], nodes.Tuple):
+                types = list(type_annotations[-1].get_children())
+            elif type_annotations[0].name == "Optional":
+                if (
+                    isinstance(type_annotations[-1], nodes.Const)
+                    and type_annotations[-1].pytype() == "builtins.NoneType"
+                ):
+                    self.add_message(
+                        "redundant-union-assign-typehint",
+                        node=node,
```
cdce8p, line 245: The error-node shouldn't be the `AnnAssign`. We can be more specific. For `Optional`, it should probably be `node.annotation`.

For Union, we could even highlight the exact duplicate item (node) 🤔

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                return
+        elif self._should_check_alternative_union_syntax and isinstance(
+            node.annotation, nodes.BinOp
+        ):
+            types = self.parse_binops_typehints(
+                node.annotation,
+            )
+        else:
+            return
+        redundant_typehints: list[str] = []
```
cdce8p, line 258:
```suggestion
            return

        redundant_typehints: list[str] = []
```
Add an empty line

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        for idx, typehint in enumerate(types):
+            typehint_str = typehint.as_string()
+            if (
+                (idx < (len(types) - 1))
+                and (any(typehint_str == x.as_string() for x in types[idx + 1 :]))
+                and typehint_str not in redundant_typehints
+            ):
+                redundant_typehints.append(typehint_str)
```
cdce8p, line 266: This is quite inefficient. `x.as_string()` is basically called `n * n` times while not being cached. Additionally, the `typehint` on which the error would be emitted is the first occurrence and not the second, third, ...

A better idea, try using a dict with the `typehint_str` as key and the position as value. (A set might also work if we omit the position.)
1. Compute `typehint_str`
2. Check if the str is in the dict / set
3. If exist emit error
4. Else add item to dict / set

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                self.add_message(
+                    "redundant-union-assign-typehint",
+                    node=node,
```
cdce8p, line 269: This should again be more specific. I.e. the specific `typehint` node.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                    args=(typehint_str),
```
cdce8p, line 270: If you use a `dict`, you could include the position of the first occurrence in the message. Although then you might need different ones for `Union` and `Optional` or clever string formatting.

Maybe you have an idea. That's more a nice to have then really necessary so we might as well skip it if it's too complicated.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+                    confidence=HIGH,
+                )
+
+    def parse_binops_typehints(
+        self, binop_node: nodes.BinOp, typehints_list: list[nodes] | None = None
+    ) -> list[nodes]:
```
cdce8p, line 276:
```suggestion
        self, binop_node: nodes.BinOp, typehints_list: list[nodes.NodeNG] | None = None
    ) -> list[nodes.NodeNG]:
```

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        if typehints_list is None:
+            typehints_list = []
```
cdce8p, line 278:
```suggestion
       typehints_list = typehints_list or []
```

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        ops = list(binop_node.get_children())
```
cdce8p, line 279: Again, use `node.left` / `node.right` respectively. That will make it easier to read.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+        typehints_list.append(ops[1])
+        if isinstance(ops[0], nodes.BinOp):
+            self.parse_binops_typehints(ops[0], typehints_list)
+        else:
+            typehints_list.append(ops[0])
```
cdce8p, line 284: This should be switched. I.e. nodes should be added in the correct order. `left` first then `right`.

tests/functional/ext/typing/redundant_union_assign_typehint_py38.py
```diff
@@ -0,0 +1,16 @@
+""""Checks for redundant Union typehints in assignments - python 3.8"""
```
cdce8p, line 1: `from __future__ import annotations` was added in Python `3.7`.

pylint/extensions/typing.py
```diff
@@ -219,6 +224,66 @@ def visit_attribute(self, node: nodes.Attribute) -> None:
...
+            ):
+                redundant_typehints.append(typehint_str)
+                self.add_message(
+                    "redundant-union-assign-typehint",
+                    node=node,
+                    args=(typehint_str),
+                    confidence=HIGH,
+                )
+
+    def parse_binops_typehints(
```
cdce8p, line 274:
```suggestion
    def _parse_binops_typehints(
```
This should also be a private helper method.

cdce8p (commented):
Already left a comment earlier. I think `duplicate-union-item` could be a good name for this check.

Besides the annotation in `AnnAssigns`, it would make sense to check other common locations for type hits as well. You mentioned one already: function arguments, but there are also function return types and lastly, `TypeAliases`. Those could be a bit more difficult. It might be a performance hit if we start checking each assignment value. A better option might be to filter `AnnAssign` for the explicit `TypeAlias` type hint and only then check the rhs.

tests/functional/ext/typing/redundant_union_assign_typehint_py38.py
orSolocate: so what should be the difference between python 3.7 and 3.8 tests then? should I just merge them?

pylint/extensions/typing.py
orSolocate: You mean by replacing it to double underscore instead of a single?

cdce8p: > You mean by replacing it to double underscore instead of a single?

Just a single one

tests/functional/ext/typing/redundant_union_assign_typehint_py38.py
cdce8p: > so what should be the difference between python 3.7 and 3.8 tests then? should I just merge them?

👍🏻 At the moment, the only difference is the future import. However, for the cases here it doesn't actually matter.

_You could even combine all tests into one file, if you keep the future import and add this:_
```ini
[TYPING]
runtime-typing=no
```

orSolocate: I merged py37 and py38 it indeed worked. 

The `runtime-typing=no` option in the `rc` file works but with a little issue:
pylint with python 3.7 does not know 'deprecated-typing-alias', 'consider-alternative-union-syntax' and cannot parse them in the `# pylint: disable=` line.
pylint with python 3.10 knows all messages and was fine.

So I decided to keep 2 test versions for now. If you got any simple way to solve this (e.g. ignore certain message for certain python version) I could merge them into one, if it is important..

pylint/extensions/typing.py
orSolocate: I think `node.annotation`  is good. the specific `typehint` node  (as string) is sent as an argument and is shown in the message itself. For example in case you have 2 different type duplications you will get 2 messages, for the same node but with different `typehint` in the message itself.

pylint/extensions/typing.py
orSolocate: @cdce8p  - beautiful algorithm mate! using a set is very clever. Thanks!

orSolocate:
Pushed a fix which improves readability of the main `visit_annassign` method.
regarding function arguments and function return types better do it in separate commit.

regarding this:
> It might be a performance hit if we start checking each assignment value



I am not sure I completely understand. The checks are done only on annotated assigns, so they wouldn't be checked on **any** assignment line. What would the `TypeAliases` solve here exactly? or in other words, what is an example of type annotation that shouldn't be checked? if it is only a single typehint it goes to the `return` statement pretty fast..

doc/whatsnew/fragments/7636.new_check
```diff
@@ -0,0 +1,4 @@
+Add ``redundant-union-assign-typehint`` message for `typing` plugin for duplicate assign typehints.
```
orSolocate, line 1: Will change the name

pylint/extensions/typing.py
orSolocate: I don't see why position matter, If you get the annotation node and the name of the duplicated type..

tests/functional/ext/typing/redundant_union_assign_typehint_py38.py
Pierre-Sassoulas: > python 3.7 does not know 'deprecated-typing-alias', 'consider-alternative-union-syntax' and cannot parse them in the # pylint: disable= line.

It might be a bug but also it's possible that you need to set the py-version to 3.7 in the rc file.

orSolocate: solved now

tests/functional/ext/typing/redundant_typehint_argument.py
```diff
@@ -0,0 +1,21 @@
+""""Checks for redundant Union typehints in assignments"""
+# pylint: disable=deprecated-typing-alias,consider-alternative-union-syntax,consider-using-alias
+
+from __future__ import annotations
+from typing import Union, Optional, Sequence
+
+#  +1: [redundant-typehint-argument, redundant-typehint-argument]
+ANSWER_0: Union[int, int, str, bool, float, str] = 0  #
```
Pierre-Sassoulas, line 8:
```suggestion
ANSWER_0: Union[int, int, str, bool, float, str] = 0
```

tests/functional/ext/typing/redundant_typehint_argument.py
```diff
@@ -0,0 +1,21 @@
...
+ANSWER_1: Optional[int] = 1
+ANSWER_2: Sequence[int] = [2]
+ANSWER_3: Union[list[int], str, int, bool, list[int]] = 3  # [redundant-typehint-argument]
+ANSWER_4: Optional[None] = None   # [redundant-typehint-argument]
+ANSWER_5: Optional[list[int]] = None
+ANSWER_6: Union[None, None] = None   # [redundant-typehint-argument]
+#  +1: [redundant-typehint-argument]
+ANSWER_7: Union[list[int], dict[int], dict[list[int]], list[str], list[str]] = [7]  #
```
Pierre-Sassoulas, line 16:
```suggestion
ANSWER_7: Union[list[int], dict[int], dict[list[int]], list[str], list[str]] = [7]
```

tests/functional/ext/typing/redundant_typehint_argument.py
```diff
@@ -0,0 +1,21 @@
...
+ANSWER_8: int | int = 8  # [redundant-typehint-argument]
+ANSWER_9: str | int | None | int | bool = 9   # [redundant-typehint-argument]
+ANSWER_10: dict | list[int] | float | str | int | bool = 10
+#  +1: [redundant-typehint-argument]
+ANSWER_11: list[int] | dict[int] | dict[list[int]] | list[str] | list[str] = ['string']  #
```
Pierre-Sassoulas, line 21:
```suggestion
ANSWER_11: list[int] | dict[int] | dict[list[int]] | list[str] | list[str] = ['string']
```

Pierre-Sassoulas (changes requested):
Some nitpicks :) It looks pretty good already 👍  !

tests/functional/ext/typing/redundant_typehint_argument.py
orSolocate: @Pierre-Sassoulas 
I thought the extra `#` was needed..
Maybe the docs need a refactor, or I just didn't get it quite right...

from:
https://pylint.pycqa.org/en/latest/development_guide/contributor_guide/tests/writing_test.html

```
You can also use # +n: [ with n an integer if the above syntax would make the line too long or other reasons:

# +1: [empty-comment]
#
```

Pierre-Sassoulas: Ha I understand. The additional ``#`` is just to signal that there is a line under it I think. Maybe we should add a comment "There error is actually here" or something ?

orSolocate: Oh my god! now I get it.  The second comment is the `empty-comment` error. Therefore the second line `#` is not a part of the syntax it is just the line itself that has the error.
It is very unclear... maybe we should change the example to something clearer like:


```
A = 5
# +1: [singleton-comparison]
B = A == None
```

orSolocate: opened a separate PR to make this mystery clearer for future contributors.
#7870

Pierre-Sassoulas: > The second comment is the empty-comment error. 

Now I remember this 😄 . There's clearly no other way to do a functional test for ``[empty-comment]`` 😄 Thank you for making the doc clearer 👍

doc/whatsnew/fragments/7636.new_check
```diff
@@ -0,0 +1,4 @@
+Add ``edundant-typehint-argument`` message for `typing` plugin for duplicate assign typehints.
```
Pierre-Sassoulas, line 1:
```suggestion
Add ``redundant-typehint-argument`` message for `typing` plugin for duplicate assign typehints.
```

Pierre-Sassoulas (approved):
👌

PR #7746, opened 2022-11-11
Avoid deprecation warning in internal use of `may_be_emitted`

jacobtylerwalls:
## Type of Changes

## Description

#7580 added a deprecation warning but didn't silence it in the test suite. Just a cleanup.

### DIFF
--- a/pylint/message/message_definition_store.py
+++ b/pylint/message/message_definition_store.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import collections
 import functools
+import warnings
 from collections.abc import Sequence, ValuesView
 from typing import TYPE_CHECKING
 
@@ -110,8 +111,10 @@ class MessageDefinitionStore:
         emittable = []
         non_emittable = []
         for message in messages:
-            if message.may_be_emitted():
-                emittable.append(message)
-            else:
-                non_emittable.append(message)
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", category=DeprecationWarning)
+                if message.may_be_emitted():
+                    emittable.append(message)
+                else:
+                    non_emittable.append(message)
         return emittable, non_emittable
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas:
Here's some context as to why we did not filter the warning: 

https://github.com/PyCQA/pylint/pull/7580#discussion_r991612938

>  we need to make a follow-up MR to make things right (give py_version instead the whole way) and I think this is a looooot of effort just for the "list-msgs-enabled", command.

We should do the refactor instead of filtering (or remove the list-msgs-enabled command ? It feels like a debug tool that no one should use if the conf is properly parsed.).

Pierre-Sassoulas:
sys.version was what we had before, the issue is that ``list-msgs-enabled``  is wrong without the refactor (the list of message is not equal to the messages enabled if py_version != sys.version). We can remove the option entirely if fixing it is not worth it but I don't think we should knowingly keep it broken.

pylint/message/message_definition_store.py
```diff
@@ -21,17 +21,20 @@
 class MessageDefinitionStore:
 
     """The messages store knows information about every possible message definition but
-    has no particular state during analysis.
+    has no particular state during analysis other than py_version.
```
jacobtylerwalls, line 24: I'm not convinced of the need to add this, actually.

Pierre-Sassoulas: Hmm, yeah the intent of the design might be broken here, maybe we do need to give the py_version all the way to list the enabled message ? On the other hand, what's the worst that could happen? Having py_version here is convenient.

jacobtylerwalls: `py_version` is not really a state that changes during analysis, so next time I push I'll just revert this docs change.

Pierre-Sassoulas: 👍

Pierre-Sassoulas:
```suggestion
    has no particular state during analysis.
```

Pierre-Sassoulas (approved):

PR #7750, opened 2022-11-11
New checker `unbalanced dict unpacking`

### LINKED ISSUE #5797: New checker - `unbalanced-dictionary-unpacking`
### Bug description

Also mentioned in discussion in https://github.com/PyCQA/pylint/issues/5113.

The following situation does not issue a message from Pylint:
a.py:
```python
for a, b, c, d, e, f, g in {1: 2}.items():
    pass
```

On the other hand, Python gives:
```
  File "a.py", line 1, in <module>
    for a, b, c, d, e, f, g in {1: 2}.items():
ValueError: not enough values to unpack (expected 7, got 2)
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
No output
```

### Expected behavior

Emit a message indicating the unpacking error.

### Pylint version

```shell
pylint 2.13.0-dev0
astroid 2.9.3
Python 3.10.0b2 (v3.10.0b2:317314165a, May 31 2021, 10:02:22) [Clang 12.0.5 (clang-1205.0.22.9)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description

This is a new checker very similar to unbalanced-tuple-unpacking but which handle dict-specific behavior. This handles when dicts are unpacked via assignment `=` or via for loops.

While I tried to do more refactoring of functionality, the code in `visit_assign` and `visit_for` is slightly and subtly different enough that it wasn't worth it or read easily.

Closes #5797


pylint/checkers/variables.py
```diff
@@ -2666,17 +2737,19 @@ def _check_unpacking(
         values = self._nodes_to_unpack(inferred)
         if values is not None:
             if len(targets) != len(values):
-                # Check if we have starred nodes.
-                if any(isinstance(target, nodes.Starred) for target in targets):
-                    return
```
clavedeluna, line 2671: I recognized we could move this check up higher which should help performance, and it had no negative impacts

pylint/checkers/variables.py
```diff
@@ -1103,6 +1117,55 @@ def __init__(self, linter: PyLinter) -> None:
...
+
+            # Nodes named `_` or Starred nodes indicate ambiguous unpacking
+            # if `dict.items()` is used so we won't flag them.
+            if any(
+                isinstance(target, nodes.Starred)
+                or isinstance(target, nodes.AssignName)
+                and target.name == "_"
+                for target in targets
+            ):
+                return
```
Pierre-Sassoulas, line 1154: It seems this line is never tested. We should find a test case where the condition is True.

pylint/checkers/variables.py
```diff
@@ -1103,6 +1117,55 @@ def __init__(self, linter: PyLinter) -> None:
...
+        targets = node.target.elts
+
+        inferred = utils.safe_infer(node.iter)
+        if inferred is None or not isinstance(inferred, DICT_TYPES):
+            return
+
+        values = self._nodes_to_unpack(inferred)
+        if values is None or len(values) == 0:
+            # no dict items returned
+            return
```
Pierre-Sassoulas, line 1136:
```suggestion
        if not values:
            # no dict items returned
            return
```

pylint/checkers/variables.py
```diff
@@ -1103,6 +1117,55 @@ def __init__(self, linter: PyLinter) -> None:
...
+
+        if isinstance(inferred, astroid.objects.DictItems):
+            # dict.items() is a bit special because values will be a tuple
+            # So as long as there are always 2 targets and values each are
+            # a tuple with two items, this will unpack correctly.
+            # Example: `for key, val in {1: 2, 3: 4}.items()`
+            if len(targets) == 2 and all(len(x.elts) == 2 for x in values):
+                return
```
Pierre-Sassoulas, line 1144: Why wouldn't ``all(len(x.elts) == len_targets for x in values)`` with ``len_target = len(targets)`` work in the general case ?

pylint/checkers/variables.py
```diff
@@ -2666,17 +2737,19 @@ def _check_unpacking(
         values = self._nodes_to_unpack(inferred)
         if values is not None:
             if len(targets) != len(values):
-                # Check if we have starred nodes.
-                if any(isinstance(target, nodes.Starred) for target in targets):
-                    return
+                if isinstance(inferred, DICT_TYPES):
+                    msg_id = "unbalanced-dict-unpacking"
```
Pierre-Sassoulas, line 2741:
```suggestion
                    symbol = "unbalanced-dict-unpacking"
```
Technically, msgid is the C1234 and symbol is "msg-name". At least that's the standard in ``pylint.messages``

pylint/checkers/variables.py
```diff
@@ -2666,17 +2737,19 @@ def _check_unpacking(
...
+                else:
+                    msg_id = "unbalanced-tuple-unpacking"
```
Pierre-Sassoulas, line 2743:
```suggestion
                    symbol = "unbalanced-tuple-unpacking"
```

pylint/checkers/variables.py
```diff
@@ -2666,17 +2737,19 @@ def _check_unpacking(
...
                 self.add_message(
-                    "unbalanced-tuple-unpacking",
+                    msg_id,
```
Pierre-Sassoulas, line 2745:
```suggestion
                    symbol,
```

Pierre-Sassoulas (changes requested):
Look good already 👍 I have some questions and suggestions.

pylint/checkers/variables.py
```diff
@@ -520,6 +527,13 @@ def _has_locals_call_after_node(stmt: nodes.NodeNG, scope: nodes.FunctionDef) ->
         "Emitted when an index used on an iterable goes beyond the length of that "
         "iterable.",
     ),
+    "W0644": (
+        "Possible unbalanced dict unpacking with "
+        "sequence%s: "
```
DanielNoord, line 532:
```suggestion
        "sequence %s: "
```

pylint/checkers/variables.py
```diff
@@ -520,6 +527,13 @@ def _has_locals_call_after_node(stmt: nodes.NodeNG, scope: nodes.FunctionDef) ->
...
+        "left side has %d label(s), right side has %d value(s)",
+        "unbalanced-dict-unpacking",
+        "Used when there is an unbalanced dict unpacking in assignment",
```
DanielNoord, line 535:
```suggestion
        "Used when there is an unbalanced dict unpacking in assignments or for loops.",
```

pylint/checkers/variables.py
```diff
@@ -1103,6 +1117,55 @@ def __init__(self, linter: PyLinter) -> None:
...
+        "unbalanced-dict-unpacking",
+    )
+    def visit_for(self, node: nodes.For) -> None:
+        if not isinstance(node.target, nodes.Tuple):
+            return
+
+        targets = node.target.elts
+
+        inferred = utils.safe_infer(node.iter)
+        if inferred is None or not isinstance(inferred, DICT_TYPES):
```
DanielNoord, line 1130:
```suggestion
        if not isinstance(inferred, DICT_TYPES):
```

pylint/checkers/variables.py
```diff
@@ -1103,6 +1117,55 @@ def __init__(self, linter: PyLinter) -> None:
...
+
+        if isinstance(inferred, astroid.objects.DictItems):
+            # dict.items() is a bit special because values will be a tuple
+            # So as long as there are always 2 targets and values each are
+            # a tuple with two items, this will unpack correctly.
+            # Example: `for key, val in {1: 2, 3: 4}.items()`
+            if len(targets) == 2 and all(len(x.elts) == 2 for x in values):
+                return
+
+            # Nodes named `_` or Starred nodes indicate ambiguous unpacking
```
DanielNoord, line 1146: Does `_` have a special syntactic meaning?

pylint/checkers/variables.py
```diff
@@ -2666,17 +2737,19 @@ def _check_unpacking(
...
+                if isinstance(inferred, DICT_TYPES):
+                    msg_id = "unbalanced-dict-unpacking"
+                else:
+                    msg_id = "unbalanced-tuple-unpacking"
                 self.add_message(
-                    "unbalanced-tuple-unpacking",
+                    msg_id,
                     node=node,
                     args=(
                         _get_unpacking_extra_info(node, inferred),
```
DanielNoord, line 2748: This is a bit of a refactor but I would prefer if we could not put a `' '` at the beginning of this return value. As you can see in the generated documentation `*Possible unbalanced tuple unpacking with sequence%s:` looks very weird.

tests/functional/u/unbalanced_dict_unpacking.py
```diff
@@ -0,0 +1,90 @@
+# pylint: disable=missing-module-docstring, missing-function-docstring, invalid-name
```
DanielNoord, line 1:
```suggestion
# pylint: disable=missing-function-docstring, invalid-name
```

Might as well just a docstring, never hurts 😄

DanielNoord (changes requested):

pylint/checkers/variables.py
clavedeluna: I assume this goes along with your comment

> ... *Possible unbalanced tuple unpacking with sequence%s: looks very weird.

1. this is the same as above unbalanced-tuple... does
2. current output is

`Possible unbalanced dict unpacking with sequence: left side`
while if I make the suggested change it would be
`Possible unbalanced dict unpacking with sequence : left side`

is that your intention? Idk why we would add a space

pylint/checkers/variables.py
clavedeluna: oh good point.... None is never going to be an instance of DICT_TYPES :D

pylint/checkers/variables.py
DanielNoord: I just noticed that the help message looks off. Imo, the help message should read:
`Possible unbalanced dict unpacking with %s: ...`

pylint/checkers/variables.py
clavedeluna: Oh good catch. I followed test driven dev too much and wrote code to fit the tests. This isn't necessary!

pylint/checkers/variables.py
clavedeluna: I think this will now be resolved, let's see coverage 🤞

pylint/checkers/variables.py
clavedeluna: Good question, this looks trickier than it is. The answer is: because x.elts is always 2 because dicts always unpack to a tuple of len 2. There are no dicts with anything other than (key:value). so

```
for key, val in {1: 2, 3: 4}.items():
	print(key)
```
works but this
```
for key, val, other in {1: 2, 3: 4}.items():
	print(key)
```
does not work because it unpacks to `dict_items([(1, 2), (3, 4)])`

Does that clear this up?

pylint/checkers/variables.py
```diff
@@ -520,6 +527,12 @@ def _has_locals_call_after_node(stmt: nodes.NodeNG, scope: nodes.FunctionDef) ->
         "Emitted when an index used on an iterable goes beyond the length of that "
         "iterable.",
     ),
+    "W0644": (
+        "Possible unbalanced dict unpacking with %s: "
+        "left side has %d label(s), right side has %d value(s)",
```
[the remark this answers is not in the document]
clavedeluna: Sure, seems clean though not hugely important. Should I open an issue to add that functionality to unbalanced-tuple-unpacking and any other msg that could use it? If not this one doesn't seem worth it either.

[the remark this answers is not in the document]
Pierre-Sassoulas: Why not just apply the suggestion as it's alrerady done ?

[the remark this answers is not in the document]
clavedeluna: Because I have to update docs. I can't just hit add suggestion and expect the commits to run.

So issue for other msgs or nay?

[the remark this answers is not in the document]
Pierre-Sassoulas: I think I misunderstood what you said as the suggestion are applied in [41591f0](https://github.com/PyCQA/pylint/pull/7750/commits/41591f0948b16bef5781f60ae49119841f6b8308) ? Do you mean to do the same pluralization on other existing messages ?

doc/user_guide/checkers/features.rst
```diff
@@ -1309,6 +1309,8 @@ Variables checker Messages
   variable is not defined in the module scope.
 :self-cls-assignment (W0642): *Invalid assignment to %s in method*
   Invalid assignment to self or cls in instance or class method respectively.
+:unbalanced-dict-unpacking (W0644): *Possible unbalanced dict unpacking with %s: left side has %d label%s, right side has %d value%s*
+  Used when there is an unbalanced dict unpacking in assignment
```
DanielNoord, line 1313: This seems out of date.

DanielNoord (approved):
One final nit, rest LGTM but somebody should probably take another good look at the tests and see if we're missing anything.

doc/user_guide/checkers/features.rst
```diff
@@ -990,8 +990,8 @@ Stdlib checker Messages
   https://docs.python.org/3/library/subprocess.html#subprocess.run
 :bad-thread-instantiation (W1506): *threading.Thread needs the target function*
   The warning is emitted when a threading.Thread class is instantiated without
-  the target function being passed. By default, the first parameter is the
-  group param, not the target param.
+  the target function being passed as a kwarg or as a second argument. By
+  default, the first parameter is the group param, not the target param.
```
clavedeluna, line 995: we forgot to update docs in a previous PR

pylint/checkers/variables.py
```diff
@@ -2666,17 +2733,27 @@ def _check_unpacking(
...
+                    args = (
+                        node.value.as_string(),
+                        len(targets),
+                        "s" if len(targets) > 1 else "",
+                        len(values),
+                        "s" if len(values) > 1 else "",
+                    )
+                else:
+                    symbol = "unbalanced-tuple-unpacking"
+                    args = (
```
Pierre-Sassoulas, line 2747: There's a mypy error that we need to fix here, probably because we added the pluralization..

Pierre-Sassoulas (commented):
Looks very good, this will be a fine addition in 2.16 👍

DanielNoord (dismissed):

Pierre-Sassoulas:
Created https://github.com/PyCQA/pylint/pull/7778 so we can rebase on it and not bother with mypy.

pylint/checkers/variables.py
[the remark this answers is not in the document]
clavedeluna: Yep that's what I meant. Whether we should create a new issue to add the programmatic pluralization to unbalanced-tuple-unpacking, at the very least, which is the most similar msg.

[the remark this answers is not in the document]
Pierre-Sassoulas: Opened #7778 because we had a mypy error that could not be fixed without it

pylint/checkers/variables.py
```diff
@@ -1103,6 +1116,52 @@ def __init__(self, linter: PyLinter) -> None:
...
+
+        if len(targets) != len(values):
+            symbol = "unbalanced-dict-unpacking"
+            self.add_message(
+                symbol,
+                node=node,
+                args=(
+                    node.iter.as_string(),
+                    len(targets),
+                    "s" if len(targets) > 1 else "",
```
Pierre-Sassoulas, line 1158:
```suggestion
                    "s" if len(targets) == 1 else "",
```

pylint/checkers/variables.py
```diff
@@ -1103,6 +1116,52 @@ def __init__(self, linter: PyLinter) -> None:
...
+                    len(values),
+                    "s" if len(values) > 1 else "",
```
Pierre-Sassoulas, line 1160:
```suggestion
                    "s" if len(values) == 1 else "",
```

pylint/checkers/variables.py
```diff
@@ -2666,17 +2733,27 @@ def _check_unpacking(
...
-                self.add_message(
-                    "unbalanced-tuple-unpacking",
-                    node=node,
-                    args=(
+                if isinstance(inferred, DICT_TYPES):
+                    symbol = "unbalanced-dict-unpacking"
+                    args = (
+                        node.value.as_string(),
+                        len(targets),
+                        "s" if len(targets) > 1 else "",
```
Pierre-Sassoulas, line 2741:
```suggestion
                        "s" if len(targets) == 1 else "",
```

pylint/checkers/variables.py
```diff
@@ -2666,17 +2733,27 @@ def _check_unpacking(
...
+                        len(values),
+                        "s" if len(values) > 1 else "",
```
Pierre-Sassoulas, line 2743:
```suggestion
                        "s" if len(values) == 1 else "",
```

Pierre-Sassoulas (changes requested):
This can be rebased on the main branch following #7778. I've also been schooled about grammar since the last review : https://github.com/PyCQA/pylint/pull/7778#pullrequestreview-1187118042

pylint/checkers/variables.py
```diff
@@ -2670,39 +2733,55 @@ def _check_unpacking(
...
-                    node=node,
-                    args=(
-                        details,
-                        len(targets),
-                        "" if len(targets) == 1 else "s",
-                        len(values),
-                        "" if len(values) == 1 else "s",
-                    ),
+                self._report_unbalanced_unpacking(
+                    node, inferred, targets, values, details
```
Pierre-Sassoulas, line 2740:
```suggestion
                    "unbalanced-tuple-unpacking", node, inferred, targets, values, details
```

pylint/checkers/variables.py
```diff
@@ -1108,6 +1128,41 @@ def __init__(self, linter: PyLinter) -> None:
...
+                return
+
+            # Starred nodes indicate ambiguous unpacking
+            # if `dict.items()` is used so we won't flag them.
+            if any(isinstance(target, nodes.Starred) for target in targets):
+                return
+
+        if len(targets) != len(values):
+            details = _get_unpacking_extra_info(node, inferred)
+            self._report_unbalanced_unpacking(node, inferred, targets, values, details)
```
Pierre-Sassoulas, line 1164:
```suggestion
            self._report_unbalanced_unpacking("unbalanced-dict-unpacking", node, inferred, targets, values, details)
```

pylint/checkers/variables.py
```diff
@@ -2670,39 +2733,55 @@ def _check_unpacking(
...
+        if isinstance(node, (nodes.Tuple, nodes.List) + DICT_TYPES):
             return node.itered()  # type: ignore[no-any-return]
         if isinstance(node, astroid.Instance) and any(
             ancestor.qname() == "typing.NamedTuple" for ancestor in node.ancestors()
         ):
             return [i for i in node.values() if isinstance(i, nodes.AssignName)]
         return None
 
+    def _report_unbalanced_unpacking(
+        self,
```
Pierre-Sassoulas, line 2758:
```suggestion
        self,
        symbol: str,
```

pylint/checkers/variables.py
```diff
@@ -2670,39 +2733,55 @@ def _check_unpacking(
...
+            "" if len(targets) == 1 else "s",
+            len(values),
+            "" if len(values) == 1 else "s",
+        )
+
+        symbol = (
+            "unbalanced-dict-unpacking"
+            if isinstance(inferred, DICT_TYPES)
+            else "unbalanced-tuple-unpacking"
+        )
```
Pierre-Sassoulas, line 2777:
```suggestion
```
Giving the symbol hardcoded in the function call is probably better here (the perf gain is negligible, but it's easier to read/understand imo).

Pierre-Sassoulas (changes requested):
Thank you @clavedeluna, this is pretty great. Small clarity refactor and let's merge.

pylint/checkers/variables.py
clavedeluna: Your suggestion to pass the symbol as an arg would make sense except here we would still need to do
```
symbol = (
            "unbalanced-dict-unpacking"
            if isinstance(inferred, DICT_TYPES)
            else "unbalanced-tuple-unpacking"
        )
 self._report_unbalanced_unpacking(symbol....)
```

There is only one place we can call ` self._report_unbalanced_unpacking("unbalanced-dict-unpacking"...` since that code path only handles that msg, but in this code path we need to differentiate.

I would like to politely push back and leave the code as is. I think it's quite clean to have the symbol check in only one location.

Pierre-Sassoulas: Hmm right I missed the fact that we handled two cases here.

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Great work on this new checker @clavedeluna !

PR #7752, opened 2022-11-11
Flag `superfluous-parens` if parentheses are used during string concatenation.

### LINKED ISSUE #4792: Make ``superfluous-parens`` consider string combinations
### Current problem

Although #4784 fixes some problems with ``superfluous-parens`` it still does not consider the following cases:
```python
Z = "TestString"
X = ("Test " + "String") # Bad
Y = ("Test " + "String") in Z # Bad
assert "" + ("Version " + "String") in Z # Bad
```

This is because the current checker focuses on unnecessary parens after keywords (`if`, `else`, etc.) and only if the `(` immediately follows the keyword. See
```python
if tokens[start + 1].string != "(":
            return
``` 
on line `373` in `format.py` and the fact that we only run when Keywords are present.

Ideally, the checker would also consider string combinations and seeing if parens are necessary there.

### Desired solution

Raise ``superfluous-parens`` message on the test cases described above.

### Additional context

_No response_

clavedeluna:
## Type of Changes

## Description

Closes #4792

This PR ensures that `superfluous-parens` is emitted if str concatenation is happening but the parentheses are not necessary, that is without the parens the concat result is the same as with them.


tests/functional/i/invalid/invalid_all_format.py
```diff
@@ -2,6 +2,6 @@
 
 Tuples with one element MUST contain a comma! Otherwise it's a string.
 """
-__all__ = ("CONST")  # [invalid-all-format]
+__all__ = ("CONST")  # [invalid-all-format, superfluous-parens]
```
clavedeluna, line 5: I'm torn on what to do here, should we make sure the `superfluous-parens` check doesn't fire for the special case of __all__ ?

tests/functional/s/slots_checks.py
```diff
@@ -64,7 +64,7 @@ class SixthBad:  # [single-string-used-for-slots]
     __slots__ = "a"
 
 class SeventhBad:  # [single-string-used-for-slots]
-    __slots__ = ('foo')
+    __slots__ = ('foo')  # [superfluous-parens]
```
clavedeluna, line 67: Same as above, do we make a special case for __slots__ to not fire superlous-parens?

tests/functional/s/superfluous_parens.py
```diff
@@ -72,6 +65,19 @@ def __iter__(self):
...
+X = ("Test " + "String")  # [superfluous-parens]
+Y = ("Test " + "String") in Z  # [superfluous-parens]
+assert ("Test " + "String") in "hello"  # [superfluous-parens]
+assert ("Version " + "String") in ("Version " + "String")  # [superfluous-parens]
+
+hi = ("CONST")  # [superfluous-parens]
+hi = ("CONST",)
+
+#TODO: maybe get this line to report [superfluous-parens] without causing other false positives.
+assert "" + ("Version " + "String") in Z
```
clavedeluna, line 83: alright so this is a legitimate case for raising superlous-parens and was brought up in the issue, but I cannot fix it without causing other bigger issues. I think it's probably ok to leave this as a false negative?

tests/functional/s/slots_checks.py
mbyrnepr2: I think so because the existing message is more accurate in this scenario.

jacobtylerwalls: Sorry for not quite following the example. Can you help me understand the case for not emitting a message here?

clavedeluna: Both `__all__` and `__slots__` are special cases when if a user assigns them to `('foo')`, that evaluates to just string `foo`. 
Both examples currently raise messages `single-string-used-for-slots` or `invalid-all-format`, respectively, because instead of `('foo')`  we expect `('foo', )` (tuple, not str).

However, with this PR, both will now also raise `superfluous-parens`.

So the question to decide is:  are we OK with both examples raising BOTH messages? Or should we do some special casing for __all__ / __slots__ that if we see superfluous parens being raised along with `single-string-used-for-slots` or `invalid-all-format`, we prioritize the latter msg, which is more "accurate"?

If the latter, prioritizing a more specific/ accurate msg over the other, are there other locations where pylint does this already that I can take as an example? I can think of a few ways to do it, but I want to be consistent.

jacobtylerwalls: Ah, I see the question now. Yes, we should raise both messages. That allows users fine-grained control (especially if they have one message globally disabled, we don't want to rob them of the opportunity to hear about the other message if we just "pick" one.)

clavedeluna: cool, then code stays as is

tests/functional/i/invalid/invalid_all_format.py
clavedeluna: Given the conversation below on the slots case, this code stays as is

jacobtylerwalls:
Do you know what's up with the `alternative-union-syntax` test failure?

clavedeluna:
> Do you know what's up with the `alternative-union-syntax` test failure?

Fixed. Disabling superfulous-parens in alternative_union_syntax:52 is reasonable given that we aren't testing for that, but also the msg being raised is also correct there bc

`CustomTypedDict = TypedDict("CustomTypedDict", my_var=(int | str))`
==
`CustomTypedDict = TypedDict("CustomTypedDict", my_var=int | str)`

tests/functional/s/slots_checks.py
mbyrnepr2: We can check the slots checker is enabled first, right? The problem is that users will see two messages, one of which is made redundant by the other.

jacobtylerwalls: That could cause problems with `useless-suppression` etc. IMO it's better to not introduce closer coupling.

mbyrnepr2: OK @jacobtylerwalls. Having said all that, I would say 'superfluous-parens' should never be emitted for `__slots__` anyway since slots cannot be a string.

jacobtylerwalls: I think it [can](https://docs.python.org/3/reference/datamodel.html?highlight=__slots__#object.__slots__) be a string. I can also try to dig up some place we had a conversation about useless-suppression so that I'm sure I'm not misremembering :-D

mbyrnepr2: You're right :D I must have remembered it as something that was always a container. Perhaps confusing that we have the `single-string-used-for-slots` message.

But anyway enough talk from me; thanks for the discussion!

pylint/checkers/format.py
```diff
@@ -55,6 +55,7 @@
     "while",
     "yield",
     "with",
+    "=",
```
jacobtylerwalls, line 58: What do you think about adding `:=` and a test case for it?

clavedeluna:
I looked at some of the primer outputs and it actually seems like this is working exactly as we want!

jacobtylerwalls (changes requested):
I think it's not a big scope change to add `:=`, so marking request changes for now. I think you'll need to create a new test file with `min_pyver=3.8` in the corresponding `.rc` file, which you can find examples of.

tests/functional/s/superfluous_parens_walrus_py38.py
```diff
@@ -1,5 +1,5 @@
 """Test the superfluous-parens warning with python 3.8 functionality (walrus operator)"""
-# pylint: disable=missing-function-docstring, invalid-name, missing-class-docstring, import-error
+# pylint: disable=missing-function-docstring, invalid-name, missing-class-docstring, import-error, pointless-statement
```
jacobtylerwalls, line 2: `pointless-statement` for the walrus operator statments looks like a false positive. @clavedeluna would you be willing to open a new issue for that?


```python
(walrus := False)
print(walrus)
```
```
b.py:1:0: W0104: Statement seems to have no effect (pointless-statement)
```

jacobtylerwalls (dismissed):

tests/functional/s/superfluous_parens_walrus_py38.py
clavedeluna: oh great point! Will open

jacobtylerwalls:
Sorry, one more thing. You could change the news fragment to end with `false_negative`, which will ensure it gets organized in the release notes with the other false negatives automagically.

jacobtylerwalls (approved):

PR #7753, opened 2022-11-12
Add docs for unused / unsupported msgs

clavedeluna:
## Type of Changes

## Description

Refs #5953

### DIFF
--- a/doc/data/messages/u/unsupported-delete-operation/good.py
+++ b/doc/data/messages/u/unsupported-delete-operation/good.py
@@ -1,3 +1,3 @@
-FRUITS = ("apple", "orange", "berry")
+FRUITS = ["apple", "orange", "berry"]
 
-del list(FRUITS)[0]
+del FRUITS[0]
--- a/doc/data/messages/u/unsupported-delete-operation/good.py
+++ b/doc/data/messages/u/unsupported-delete-operation/good.py
@@ -1,1 +1,3 @@
-# This is a placeholder for correct code for this message.
+FRUITS = ("apple", "orange", "berry")
+
+del list(FRUITS)[0]
--- a/doc/data/messages/u/unsupported-membership-test/good.py
+++ b/doc/data/messages/u/unsupported-membership-test/good.py
@@ -1,7 +1,6 @@
 class Fruit:
     FRUITS = ["apple", "orange"]
-
-    def __getitem__(self, name):
+    def __contains__(self, name):
         return name in self.FRUITS
 
 
--- a/doc/data/messages/u/unsupported-membership-test/good.py
+++ b/doc/data/messages/u/unsupported-membership-test/good.py
@@ -1,1 +1,8 @@
-# This is a placeholder for correct code for this message.
+class Fruit:
+    FRUITS = ["apple", "orange"]
+
+    def __getitem__(self, name):
+        return name in self.FRUITS
+
+
+apple = "apple" in Fruit()
--- a/doc/data/messages/u/unused-private-member/good.py
+++ b/doc/data/messages/u/unused-private-member/good.py
@@ -1,9 +1,9 @@
 class Fruit:
-    FRUITS = ["apple", "orange"]
+    FRUITS = {"apple": "red", "orange": "orange"}
 
-    def __print_color(self, name):
-        print(f"{name}: red")
+    def __print_color(self, name, color):
+        print(f"{name}: {color}")
 
     def print(self):
-        for fruit in self.FRUITS:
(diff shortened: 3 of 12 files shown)


doc/data/messages/u/unsupported-delete-operation/good.py
```diff
@@ -1 +1,3 @@
-# This is a placeholder for correct code for this message.
+FRUITS = ("apple", "orange", "berry")
+
+del list(FRUITS)[0]
```
jacobtylerwalls, line 3: `list(FRUITS)` is immediately discarded, so I would consider constructing an example with a more realistic use of `del`.

doc/data/messages/u/unsupported-membership-test/good.py
```diff
@@ -1 +1,8 @@
-# This is a placeholder for correct code for this message.
+class Fruit:
+    FRUITS = ["apple", "orange"]
+
+    def __getitem__(self, name):
+        return name in self.FRUITS
+
+
+apple = "apple" in Fruit()
```
jacobtylerwalls, line 8: When I execute this file, the python interpreter just hangs.

clavedeluna: huh! you're right! even hangs in the REPL. I'll investigate but will change the example

clavedeluna: looks like an infinite loop 🤦‍♀️

jacobtylerwalls: Btw, the primer should finally be working now if you merge main. Sorry for bringing it down earlier this week!

clavedeluna: oh thank you!! that's excellent news as there's lots of PRs waiting for it

doc/data/messages/u/unused-private-member/good.py
```diff
@@ -1 +1,9 @@
-# This is a placeholder for correct code for this message.
+class Fruit:
+    FRUITS = ["apple", "orange"]
```
Pierre-Sassoulas, line 2:
```suggestion
    FRUITS = {"apple": "red", "orange": "orange"}
```

doc/data/messages/u/unused-private-member/good.py
```diff
@@ -1 +1,9 @@
...
+
+    def __print_color(self, name):
+        print(f"{name}: red")
+
+    def print(self):
+        for fruit in self.FRUITS:
+            self.__print_color(fruit)
```
Pierre-Sassoulas, line 9:
```suggestion
    def __print_color(self, name, color):
        print(f"{name}: {color}")

    def print(self):
        for fruit, color in self.FRUITS.items():
            self.__print_color(fruit, color)
```

doc/data/messages/u/unused-private-member/good.py
```diff
@@ -1 +1,9 @@
-# This is a placeholder for correct code for this message.
+class Fruit:
+    FRUITS = ["apple", "orange"]
+
+    def __print_color(self, name):
```
jacobtylerwalls, line 4: Sorry for the piecemeal review. One last Q. Is there a reason you used two leading underscores rather than one? AFAIK double underscores indicate "name mangling," and if that's not relevant here, then I just didn't want to encourage it in "good.py". Thoughts?

clavedeluna: Since we're demonstrating examples for unused-private-member, it wouldn't make a lot of sense to do an example of "good" by not having a private member at all, since "bad" has one. So here I"m using it.

Pierre-Sassoulas (commented):
Thank you ! Just a small suggestion.

Pierre-Sassoulas (approved):

doc/data/messages/u/unused-private-member/good.py
jacobtylerwalls: I meant a private member with one underscore rather than two. But it's alright.

clavedeluna: Isn't one underscore protected, not private?

Pierre-Sassoulas: Sorry I merged without waiting for the discussion to conclude. I think @clavedeluna is right, protected could be used in a child class, we do need a private member for this example.

jacobtylerwalls: You're right, I didn't look at the source code for the check and wrongly assumed it was dealing with both flavors private and protected. Never mind!

PR #7756, opened 2022-11-12
reimported should detect aliased imports

### LINKED ISSUE #4836: False negative ``reimported`` when using an import alias
### Current problem

I think it would be nice if Pylint could check if there is same import name from different packages.
For example, in the code below, import order is fine, but the 'Path' imports overlap, causing errors somewhere.
```python
from pathlib import Path
from FastAPI import Path
```

### Desired solution

overlapping imports found. `Path`

### Additional context

_No response_

clavedeluna:
## Type of Changes

## Description

Closes #4836

Issue brought up the fact that the `reimported` message was confusing and hinted at something else, while also giving us more test cases to improve the check for imports using aliases.

### DIFF
--- a/doc/whatsnew/fragments/4836.bugfix
+++ b/doc/whatsnew/fragments/4836.bugfix
@@ -1,0 +1,3 @@
+Update ``reimported`` help message for clarity and ensure aliased imports are also detected.
+
+Closes #4836
--- a/doc/whatsnew/fragments/4836.bugfix
+++ b/doc/whatsnew/fragments/4836.bugfix
@@ -1,3 +1,3 @@
-Update ``reimported`` help message for clarity and ensure aliased imports are also detected.
+Update ``reimported`` help message for clarity and add a ``shadowed-import`` message for aliased imports.
 
 Closes #4836
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -951,7 +951,9 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
                 )
                 if first is not None and msg is not None:
                     name = name if msg == "reimported" else alias
-                    self.add_message(msg, node=node, args=(name, first.fromlineno))
+                    self.add_message(
+                        msg, node=node, args=(name, first.fromlineno), confidence=HIGH
+                    )
 
     def _report_external_dependencies(
         self, sect: Section, _: LinterStats, _dummy: LinterStats | None
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -12,7 +12,7 @@ import os
 import sys
 from collections import defaultdict
 from collections.abc import ItemsView, Sequence
-from typing import TYPE_CHECKING, Any, Dict, List, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Tuple, Union
 
 import astroid
 from astroid import nodes
@@ -97,12 +97,14 @@ def _get_first_import(
     base: str | None,
     level: int | None,
     alias: str | None,
-) -> nodes.Import | nodes.ImportFrom | None:
+) -> Tuple[nodes.Import | nodes.ImportFrom | None, str | None]:
     """Return the node where [base.]<name> is imported or None if not found."""
     fullname = f"{base}.{name}" if base else name
 
     first = None
     found = False
+    msg = "reimported"
+
     for first in context.body:
         if first is node:
             continue
@@ -127,12 +129,13 @@ def _get_first_import(
                         break
                     if imported_name == alias:
                         found = True
+                        msg = "shadowed-import"
                         break
                 if found:
                     break
     if found and not astroid.are_exclusive(first, node):
-        return first
-    return None
+        return first, msg
+    return None, None
 
 
 def _ignore_import_failure(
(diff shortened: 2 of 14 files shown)


tests/functional/r/reimported.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=missing-docstring,unused-import,import-error, wildcard-import,unused-wildcard-import,redefined-builtin,no-name-in-module,ungrouped-imports,wrong-import-order
+# pylint: disable=missing-docstring,unused-import,import-error, wildcard-import,unused-wildcard-import,redefined-builtin,no-name-in-module,ungrouped-imports,wrong-import-order,wrong-import-position,consider-using-from-import
```
Pierre-Sassoulas, line 2:
```suggestion
# pylint: disable=missing-docstring,unused-import,import-error, wildcard-import,unused-wildcard-import,redefined-builtin,no-name-in-module
# pylint: disable=ungrouped-imports, wrong-import-order,wrong-import-position,consider-using-from-import
```

Pierre-Sassoulas (commented):
LGTM, I would go further and rename ``reimported`` to something like ``import-name-clash``or something clearer about what the problem is. "reimported" feel like what should be raised for:
```pyhon
from pathlib import Path
from pathlib import Path
```
Which is a lot less problematic than the problem we detect here.

Pierre-Sassoulas:
I realize by reading the issue that we already had this discussion. Now that I think about it again this feel like a new check, which should have a different message and probably also a different level (warning at least, while reimported is probably just a convention).

clavedeluna:
I followed the conclusion of the issue. If maintainers would like to change scope, lmk I will close this.

Pierre-Sassoulas:
Let's pursue the discussion in the original issue. I feel that your implementation would permit to separate the two use case relatively easily though.

Pierre-Sassoulas:
Let's separate the two checks between ``reimported`` and ``shadowed-import``. I also think that this is a false positive for ``shadowed-import`` as the import is not shadowed:
```python
from pandas._libs import (
    algos as libalgos,
    hashtable as ht,
)
import pandas.core.algorithms as algos
```

But this should raise shadowed-import:

```python
from pandas._libs import (
    algos,
    hashtable as ht,
)
import pandas.core.algorithms as algos  # [shadowed-import]
```

And this should raise reimported:

```python
from pandas._libs import (
    algos as libalgos,
    hashtable as ht,
)
import pandas._libs.algos as algos  # [reimported]
```

clavedeluna:
As a principle, I don't like changing the scope of a PR after work that _could_ be merged is complete. This PR could be merged as is, and through an iterative process, a new issue can be opened to then separate the two checks. Without this, you're asking a contributor to do more work than they initially signed up for in an issue. So not saying scope can't change -it always will. Just saying that we should manage it by cutting it with issues.

Pierre-Sassoulas:
Sorry about that but I don't think this is mergeable as is. This would make ``reimported`` confusing by adding a new use case that does not correspond to the current name and we'd have to make real breaking change to correct it later if we release it. (a user disable reimported somewhere and the message then change to import-shadowed and she has to disable it again)  So if we merge this it's going to make the main branch un-realeasable until someone fix it and that's not something I want to have to keep in mind.

pylint/checkers/imports.py
```diff
@@ -914,8 +932,10 @@ def _check_reimport(
         basename: str | None = None,
         level: int | None = None,
     ) -> None:
-        """Check if the import is necessary (i.e. not already done)."""
-        if not self.linter.is_message_enabled("reimported"):
+        """Check if a module with the same name is already imported or aliased."""
+        if not self.linter.is_message_enabled(
+            "reimported"
+        ) and not self.linter.is_message_enabled("shadowed-import"):
```
clavedeluna, line 938: I see a few similar is_message_enabled` checks in the codebase, should these actually be replaced with `@only_required_for_messages(...`?

Pierre-Sassoulas: I think ``only_required_for_messages`` only work for visitor function, so ``is_message_enabled`` is a suboptimization for non visitor function. Imo it's premature optimization to add them when implementing but also making the code worst to remove them so we should just leave it alone unless there's an actual problem with it.

doc/whatsnew/fragments/4836.bugfix
```diff
@@ -0,0 +1,3 @@
+Update ``reimported`` help message for clarity and ensure aliased imports are also detected.
```
Pierre-Sassoulas, line 1:
```suggestion
Update ``reimported`` help message for clarity and add a ``shadowed-import`` message for aliased imports.
```

pylint/checkers/imports.py
```diff
@@ -303,6 +316,11 @@ def _make_graph(
         "Used when an import statement is used anywhere other than the module "
         "toplevel. Move this import to the top of the file.",
     ),
+    "W0416": (
+        "Shadowed %r (imported line %s)",
+        "shadowed-import",
+        "Used when a module is aliased with a name that shadowed another import.",
```
Pierre-Sassoulas, line 322:
```suggestion
        "Used when a module is aliased with a name that shadows another import.",
```

Pierre-Sassoulas (changes requested):
Looks pretty great 👍 Some minor nitpicks and let's merge !

Pierre-Sassoulas (approved):

doc/whatsnew/fragments/4836.bugfix
```diff
@@ -0,0 +1,3 @@
+Update ``reimported`` help message for clarity and add a ``shadowed-import`` message for aliased imports.
```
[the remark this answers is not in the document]
Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/7903

PR #7762, opened 2022-11-13
Improve `invalid-slice-index` and add `invalid-slice-step`

cdce8p:
## Description
Followup to https://github.com/PyCQA/astroid/pull/1844#issuecomment-1312281340 by @nelfin.

This PR
* Improve the `invalid-slice-index` check to detect invalid indices on string and byte sequences.
```py
"Hallo"["a":"z"]  # invalid-slice-index (2x)
```
* Adds a new warning for `invalid-slice-step` when `0` is passed as step value for common builtin sequences.
```py
["a", "b", "c"][::0]  # invalid-slice-step
```

### DIFF
--- a/doc/data/messages/i/invalid-slice-step/bad.py
+++ b/doc/data/messages/i/invalid-slice-step/bad.py
@@ -1,0 +1,3 @@
+LETTERS = ["a", "b", "c", "d"]
+
+LETTERS[::0]  # [invalid-slice-step]
--- a/doc/data/messages/i/invalid-slice-step/good.py
+++ b/doc/data/messages/i/invalid-slice-step/good.py
@@ -1,0 +1,3 @@
+LETTERS = ["a", "b", "c", "d"]
+
+LETTERS[::2]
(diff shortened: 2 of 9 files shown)


Pierre-Sassoulas (approved):
LGTM, the error level check are always a pleasure to review :) (Line not covered is due to inference it's hard to test in the current state of the testutil)

cdce8p:
> (Line not covered is due to inference it's hard to test in the current state of the testutil)

It's actually a bit more complicated. `_check_invalid_slice_index` is only called from `_check_invalid_sequence_index` which includes a pretty strong filter of things too check. So all cases (with custom `__getitem__` methods) don't even get to the `slice_index` check.

I wouldn't want to remove the uncovered line in `_check_invalid_slice_index` though, as at some point we might want to loosen the strict guard clause a bit.

https://github.com/PyCQA/pylint/blob/435a5a1d8e542d1240b8a66985feb8950fb6d739/pylint/checkers/typecheck.py#L1695-L1701

Pierre-Sassoulas (approved):

PR #7766, opened 2022-11-14
Suppress `stop-iteration-return` on `itertools.cycle`

### LINKED ISSUE #7765: `stop-iteration-return` false positive when using `itertools.cycle`
### Bug description

Running pylint on this code gives a false positive:

```
from itertools import cycle


def test():
    vals = cycle('ABCD')
    while True:
        yield next(vals)
```

This is similar to #2158 which was fixed by adding a list of [known infinite iterators](https://github.com/PyCQA/pylint/blob/main/pylint/checkers/refactoring/refactoring_checker.py#L38).

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:7:14: R1708: Do not raise StopIteration in generator, use return statement instead (stop-iteration-return)
```

### Expected behavior

No warning, because `cycle` [repeats indefinitely](https://docs.python.org/3/library/itertools.html#itertools.cycle).

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.3
Python 3.10.7 (main, Nov  2 2022, 18:49:29) [GCC 12.2.0]
```

### OS / Environment

Ubuntu 22.10

### Additional dependencies

_No response_

tushar-deepsource:
## Type of Changes

## Description

Closes #7765


Pierre-Sassoulas (dismissed):
Clean fix, thank you 👌

doc/whatsnew/fragments/7766.false_positive
```diff
@@ -0,0 +1,3 @@
+Don't warn about ``stop-iteration-return`` when using ``next()`` over ``itertools.cycle``
```
Pierre-Sassoulas, line 1:
```suggestion
Don't warn about ``stop-iteration-return`` when using ``next()`` over ``itertools.cycle``.
```

Pierre-Sassoulas (approved):

Pierre-Sassoulas (approved):

PR #7775, opened 2022-11-16
Backporting branch 2.15.6

Pierre-Sassoulas:


DanielNoord:
@Pierre-Sassoulas Just to make sure, was it a conscious decision to allow merging without review again? Personally I think it is much better, as it gives a (bit of a) false sense of security and PRs like this should just be merged by the relevant maintainer, but I wanted to double check.

Pierre-Sassoulas:
Do you think being able to merge without review is a good thing or a bad thing, I'm not sure ? :)

I removed the review need because there's also the case of pushing a micro change and having to ask for a review all other again. (I trust all maintainer with write access to not abuse the power.)

I merged this without review because I wanted to test the backport workflow, and the workflow cherry-picked something on the maintenance branch in the wrong order it was going to ruin the chronological order of the history. Also simply cherry-picking when there is no conflict during the cherry-pick is consensual imo.

DanielNoord:
I think it is, although there are obviously security concerns...

Then again, why would anybody trust any review I give? I feel like it gives a false sense of security as there is no reason why a review from me is "better" than not having a review. I feel like being able to merge security hot fixes is more important than getting a review of an unknown internet stranger.

_But as I said, I certainly see the downsides.._

Pierre-Sassoulas:
>  I feel like being able to merge security hot fixes is more important than getting a review of an unknown internet stranger.
 
It was pretty fucking convenient when Ashley alone fixed a botched release when all the European maintainers where sleeping, that's for sure.

But there's definitely an element of trust here. I trust you to be the person who did 850+ commit in pylint already and have publish right in pypi when you review a merge request and not a "random internet stranger" 😄 (We might enforce two factor auth at some point to ensure that). I also trust everyone with write access to not merge non consensual changes or possibly buggy and un-reviewed changes on main.

DanielNoord:
Yeah, but I feel like the trust is for others who don't know us, through the OS package scores. Why does merging with review from people you don't know count for more than people merging with knowledge of the package (without a review). There is definitely a discussion to be had there.

I wouldn't be bother with keeping the setting turned off until somebody starts complaining about it.

Pierre-Sassoulas:
Yes, this lower a little our security metrics in openSSF, and I'm paid to make the metric look good. So it could be a problem. But on the other hand... no one is paid to review my merge request. And I'm not paid enough for a release to take 5 asynchronous days when I want to release a patch. Especially if it's a hot fix that needs to be released ASAP because every hour 10k broken pylint get downloaded. Flexibility is nice, I agree with you we should keep it that way for now.

PR #7776, opened 2022-11-16
[release process] Add a github actions to backport fix automatically

Pierre-Sassoulas:
## Type of Changes

## Description

Based on https://github.com/tibdex/backport which seems awfully conveniant.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,25 @@
+name: Backport
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    name: Backport
+    runs-on: ubuntu-latest
+    # Only react to merged PRs for security reasons.
+    # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
+    if: >
+      github.event.pull_request.merged && (
+        github.event.action == 'closed'
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'backport')
+        )
+      )
+    steps:
+      - uses: tibdex/backport@v2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/doc/development_guide/contributor_guide/release.md
+++ b/doc/development_guide/contributor_guide/release.md
@@ -49,17 +49,12 @@ branch
 
 ## Back-porting a fix from `main` to the maintenance branch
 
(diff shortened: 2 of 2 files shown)


.github/workflows/backport.yml
```diff
@@ -0,0 +1,25 @@
+name: Backport
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
```
DanielNoord, line 6: I guess this should only be closed then? If we only want to run on merged PRs?

Pierre-Sassoulas: It's taken from step 1 in https://github.com/tibdex/backport#usage directly

DanielNoord: I mean, I haven't tested this, but it feels a bit weird. It should only run whenever a pull request gets closed, see:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

Pierre-Sassoulas: Now that I think about it, it's probably so you can backport after merging by adding labels. (Which would have avoided the work in #7775 )

DanielNoord: Makes sense!

.github/workflows/backport.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: Backport
+    runs-on: ubuntu-latest
+    # Only react to merged PRs for security reasons.
+    # See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
+    if: >
+      github.event.pull_request.merged && (
+        github.event.action == 'closed'
+        || (
+          github.event.action == 'labeled'
+          && contains(github.event.label.name, 'backport')
```
DanielNoord, line 19:
```suggestion
          && contains(github.event.label.name, 'needs backport')
```

Pierre-Sassoulas: The github action's code force us to use this keyword. Also as it's done automatically we don't "need" to do anything 🎉

DanielNoord (approved):

Pierre-Sassoulas:
Seems like I still need to rune up the config because of ``tibdex/backport@v2 is not allowed to be used in PyCQA/pylint. Actions in this workflow must be: within a repository owned by PyCQA, created by GitHub, or matching the following: kanga333/comment-hider@*``. I reviewed the code of the action itself and I did not see anything nefarious looking. But it's not an official github action.

PR #7778, opened 2022-11-16
Better error message for unpacking-non-sequence and unbalanced-tuple-unpacking

Pierre-Sassoulas:
## Type of Changes

## Description

Necessary to be able to do https://github.com/PyCQA/pylint/pull/7750 without thinking about it.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2681,9 +2681,9 @@ class VariablesChecker(BaseChecker):
                     args=(
                         details,
                         len(targets),
-                        "s" if len(targets) > 1 else "",
+                        "" if len(targets) == 1 else "s",
                         len(values),
-                        "s" if len(values) > 1 else "",
+                        "" if len(values) == 1 else "s",
                     ),
                 )
         # attempt to check unpacking may be possible (i.e. RHS is iterable)
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -167,11 +167,11 @@ def _get_unpacking_extra_info(node: nodes.Assign, inferred: InferenceResult) -> 
     inferred_module = inferred.root().name
     if node.root().name == inferred_module:
         if node.lineno == inferred.lineno:
-            more = f" {inferred.as_string()}"
+            more = f"'{inferred.as_string()}'"
         elif inferred.lineno:
-            more = f" defined at line {inferred.lineno}"
+            more = f"defined at line {inferred.lineno}"
     elif inferred.lineno:
-        more = f" defined at line {inferred.lineno} of {inferred_module}"
+        more = f"defined at line {inferred.lineno} of {inferred_module}"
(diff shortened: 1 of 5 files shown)


clavedeluna (approved):

pylint/checkers/variables.py
```diff
@@ -2673,18 +2672,18 @@ def _check_unpacking(
                     "unbalanced-tuple-unpacking",
                     node=node,
                     args=(
-                        _get_unpacking_extra_info(node, inferred),
+                        details,
                         len(targets),
+                        "s" if len(targets) > 1 else "",
```
mbyrnepr2, line 2677:
```suggestion
                        "" if len(targets) == 1 else "s",
```

pylint/checkers/variables.py
```diff
@@ -2673,18 +2672,18 @@ def _check_unpacking(
...
                         len(values),
+                        "s" if len(values) > 1 else "",
```
mbyrnepr2, line 2679:
```suggestion
                        "" if len(values) == 1 else "s",
```

mbyrnepr2 (commented):
Looks great. I like the care for usability. Small change required I think!
The primer gives: `right side has 0 value`

pylint/checkers/variables.py
Pierre-Sassoulas: Damn TIL ! a search for "0 values or 0 value" bring nothing but it seems that apple is the universal example. https://www.quora.com/Is-zero-plural-You-have-one-apple-I-have-zero-appleS

Pierre-Sassoulas (commented):
Thank you for the grammar lesson @mbyrnepr2 😉

mbyrnepr2:
> Thank you for the grammar lesson @mbyrnepr2 😉

Zero problems 😅

pylint/checkers/variables.py
mbyrnepr2: 😁🍎

mbyrnepr2 (approved):

clavedeluna (commented):
nice!

pylint/checkers/variables.py
```diff
@@ -2671,6 +2669,7 @@ def _check_unpacking(
 
         # Attempt to check unpacking is properly balanced
         values = self._nodes_to_unpack(inferred)
+        details = _get_unpacking_extra_info(node, inferred)
```
clavedeluna, line 2672: this prob should've been below the if statement in ln 2675 but no big deal, I'll move it in the next PR

clavedeluna: nvm, I see it's used below!

Pierre-Sassoulas: No I thought of it too (and then forgot). Performance wise this is pretty bad, we'd better duplicate the code so we only create the details when something is going to be raised.

PR #7783, opened 2022-11-17
Deprecation following the separation of emacs related file to https://github.com/emacsorphanage/pylint

Pierre-Sassoulas:
## Type of Changes

## Description

Deprecation of the emacs support in favor of an external repository : https://github.com/emacsorphanage/pylint. epylint needs to be kept for now, but we'll remove it in 3.0.

Closes https://github.com/PyCQA/pylint/issues/7737


nickdrozd (approved):

pylint/epylint.py
```diff
@@ -168,6 +169,10 @@ def py_run(
     To silently run Pylint on a module, and get its standard output and error:
         >>> (pylint_stdout, pylint_stderr) = py_run( 'module_name.py', True)
     """
+    warnings.warn(
+        "'epylint' will be removed in pylint 3.0, use https://github.com/emacsorphanage/pylint instead.",
+        DeprecationWarning,
```
jacobtylerwalls, line 174:
```suggestion
        DeprecationWarning,
        stacklevel=2,
```

pylint/epylint.py
```diff
@@ -198,6 +203,10 @@ def py_run(
 
 
 def Run(argv: Sequence[str] | None = None) -> NoReturn:
+    warnings.warn(
+        "'epylint' will be removed in pylint 3.0, use https://github.com/emacsorphanage/pylint instead.",
+        DeprecationWarning,
```
jacobtylerwalls, line 208:
```suggestion
        DeprecationWarning,
        stacklevel=2,
```

DanielNoord (commented):
I don't really understand how we are deprecating stuff while also removing stuff? Shouldn't we remove the flymake files at `3.0` then?

I'd be fine with removing everything and dealing with the fallout in `2.x` if there is any, but deprecating and removing at then same time feels weird..

jacobtylerwalls (approved):
Great, thank you!

jacobtylerwalls:
> I don't really understand how we are deprecating stuff while also removing stuff? Shouldn't we remove the flymake files at 3.0 then?

`epylint` hasn't been removed yet AFIACT, just the `elisp` related startup files, whatever those are. The tests for epylint still pass, and they now assert that deprecation warnings are emitted.

I suppose the release note could go into more detail about the `elisp` files being removed, but if it requires figuring out what they are in order to describe them correctly, I wouldn't insist on it.

DanielNoord:
Okay, let's merge then. Removing stuff people don't know anyhow seems like a good change!

Pierre-Sassoulas:
For anyone reading this, there was a discussion with a pylint contributor using emacs before taking the decision. The short version is that:

> I just run Pylint in M-x compile and it works well enough.

https://github.com/PyCQA/pylint/issues/7737#issuecomment-1317890751

The slightly longer version is that epylint is necessary for flymake only. It seems there's a lot of ways to integrate pylint to emacs without flymake. We also have no contributor who use flymake and we didn't have any contributor knowledgeable with emacs until very recently.

jacobtylerwalls:
Thanks @Pierre-Sassoulas I was aware of that conversation as well, so I should have chosen words that reflected that when corresponding with Daniël! Thanks for the follow-up.

sandrotosi:
This PR was merged 20 days ago, but 2.15.8 (released 5 days ago) does not contain it, which surprised me. is it because `main` is only for the new 2.16 while 2.15.x comes from a diffrent branch?

Pierre-Sassoulas:
@sandrotosi yes 2.15.x is in maintenance mode we're only fixing bugs / false positives.

PR #7789, opened 2022-11-17
``unnecessary-list-index-lookup`` false negative

### LINKED ISSUE #7770: False negative `unnecessary-list-index-lookup` with `iterable` kwarg
### Bug description

While there are plenty of tests for unnecessary_list_index_lookup, looks like missing is the case for using `iterable` as a kwarg

```
# test.py
series = [1, 2, 3]
for idx, val in enumerate(iterable=series, start=0):
    print(series[idx])  # should report [unnecessary-list-index-lookup] but currently isn't
```

Same code with iterable as an arg correctly reports and is tested.

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
Your code has been rated at 10.00/10 (previous run: 0.00/10, +10.00)
```

### Expected behavior

`test.py:4:10: R1736: Unnecessary list index lookup, use 'val' instead (unnecessary-list-index-lookup)`

### Pylint version

```shell
pylint 2.16.0-dev
astroid 2.12.12
Python 3.11.0
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description

`unnecessary-list-index-lookup` code wasn't taking into account the fact that `enumerate` could be called with a kwarg.

Closes #7770

### DIFF
--- a/doc/whatsnew/fragments/7770.false_negative
+++ b/doc/whatsnew/fragments/7770.false_negative
@@ -1,0 +1,3 @@
+Fixes ``unnecessary-list-index-lookup`` false negative when ``enumerate`` is called with ``iterable`` as a kwarg.
+
+Closes #7770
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (approved):
👌

PR #7793, opened 2022-11-18
Fix false positive `used-before-assignment` in pattern matching with a guard

### LINKED ISSUE #5327: False positive `used-before-assignment` in pattern matching with a guard.
### Bug description

When using pattern matching with a guard, `pylint` claims that the variable the guard uses is used before assignment. 

### Example:

```python
match ("example", "one"):
    case (x, y) if x == "example":  # <-- pylint complains about 'x' here
        print("woo!")
    case _:
        print("boo...")
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:2:19: E0601: Using variable 'x' before assignment (used-before-assignment)

-------------------------------------
Your code has been rated at -10.00/10
```

### Expected behavior

Pattern matching with guards is correctly parsed and supported.

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.4
```

### OS / Environment

Mac OSX Big Sur 11.6 (20G165)

### Additional dependencies

_No response_

i-likecheese:
## Type of Changes

## Description

Fix false positive `used-before-assignment` in pattern matching with a guard #5327

Closes #5327


i-likecheese:
Any guidance on what would be a proper way of testing this?

Pierre-Sassoulas:
Hey @i-likecheese , thank you for contributing. You can create a functional test using a configuration file like [this](https://github.com/PyCQA/pylint/blob/main/tests/functional/g/generic_alias/generic_alias_related_py39.rc.) so it's executed only for python 3.10.  See [the functional tests documentation](https://pylint.pycqa.org/en/latest/development_guide/contributor_guide/tests/writing_test.html#functional-tests)

pylint/checkers/variables.py
```diff
@@ -1641,6 +1642,10 @@ def _check_consumer(
                 if not (
                     self._postponed_evaluation_enabled
                     and isinstance(stmt, (nodes.AnnAssign, nodes.FunctionDef))
+                ) and not (
+                    PY310_PLUS
+                    and isinstance(stmt, nodes.Match)
+                    and isinstance(node.parent, nodes.Compare)
```
jacobtylerwalls, line 1648: This doesn't solve this false positive, because the `.parent` is `nodes.Name`:
```python
match ("example", "one"):
    case (x, y) if x:  # <-- pylint complains about 'x' here
        print("woo!")
    case _:
        print("boo...")
```

jacobtylerwalls (changes requested):
Thanks for tackling this false positive!

I'd prefer to solve it closer to the source instead of at the very last minute, so that the variables checker, which is already pretty hard to reason about, stays somewhat workable.

I suggest handling this case in `_maybe_used_and_assigned_at_once`, perhaps like:
```diff
diff --git a/pylint/checkers/variables.py b/pylint/checkers/variables.py
index 627e82ef0..7ad0c0913 100644
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1963,6 +1963,7 @@ class VariablesChecker(BaseChecker):
                             nodes.AugAssign,
                             nodes.Expr,
                             nodes.Return,
+                            nodes.Match,
                         ),
                     )
                     and VariablesChecker._maybe_used_and_assigned_at_once(defstmt)
@@ -2053,6 +2054,8 @@ class VariablesChecker(BaseChecker):
         """Check if `defstmt` has the potential to use and assign a name in the
         same statement.
         """
+        if isinstance(defstmt, nodes.Match):
+            return any(case.guard for case in defstmt.cases)
         if isinstance(defstmt.value, nodes.BaseContainer) and defstmt.value.elts:
             # The assignment must happen as part of the first element
             # e.g. "assert (x:= True), x"
```

jacobtylerwalls:
Superseded by #7922.

PR #5765, opened 2022-02-02
Add check against initializing and then mutating a dictionary

nickdrozd:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :sparkles: New feature |

## Description

Add a new optional check against initializing a dictionary and then immediately mutating it.

Closes #2876

PR #7794, opened 2022-11-18
New check `dict-init-mutate`

### LINKED ISSUE #2876: New check: Use single-statement dict initialization
Here's an ugly pattern, commonly found in eg config files:

```python
config = {}
config['dir'] = 'bin'
config['user'] = 'me'
config['workers'] = 5
```

An empty dict is initialized and then mutated several times. It would
be better to use declarative literal syntax:

```python
config = {
    'dir': 'bin',
    'user': 'me',
    'workers': 5,
}
```

No mutation! No repetition!

I'm imagining a check that would look for an empty dict initialization
and then see if any following lines are bare dict assignments. It
could get tricky though, as this pattern horrible pattern is sometimes
recursive:

```python
config = {}
config['dir'] = 'bin'
config['user'] = 'me'
config['workers'] = 5
config['options'] = {}
config['options']['debug'] = False
config['options']['verbose'] = True
```

Yuck!

And I know, a check like this gets released, CIs start showing failing
builds, and everyone thinks the sky is falling. Maybe make it
optional? Or a warning or something? Anything to discourage this
pattern.

clavedeluna:
## Type of Changes

## Description

Taking over https://github.com/PyCQA/pylint/pull/5765

Add new check to detect mutating dict after having immediately created it.

Closes #2876


clavedeluna:
Primer output looks really good, in fact [this suggestion](https://github.com/getsentry/sentry/blob/9708ac751a6c27695ce5007fcc6661c3f5a9bcfc/src/sentry/testutils/cases.py#L1239) is just what we would've wanted to suggest.

doc/user_guide/checkers/extensions.rst
```diff
@@ -264,6 +265,21 @@ Design checker Messages
...
+Dict-Init-Mutate checker
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+This checker is provided by ``pylint.extensions.dict_init_mutate``.
+Verbatim name of the checker is ``dict-init-mutate``.
+
+Dict-Init-Mutate checker Messages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+:dict-init-mutate (W3301): *Dictionary mutated immediately after initialization*
+  Dictionaries can be initialized with a single statementusing dictionary
```
nickdrozd, line 279:
```suggestion
  Dictionaries can be initialized with a single statement using dictionary
```

doc/data/messages/d/dict-init-mutate/bad.py
```diff
@@ -0,0 +1,3 @@
+fruits = {}  # [dict-init-mutate]
```
nickdrozd, line 1: I suggest sticking with the `config` example here instead of fruits, so as to emphasize the real-world existence of this pattern.

clavedeluna: HA! @Pierre-Sassoulas  loves fruits ( I think?) so I just stick to them since he approves :)

pylint/extensions/dict_init_mutate.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        Detect dictionary mutation immediately after initialization.
+
+        At this time, detecting nested mutation is not supported.
+        """
+        if not isinstance(node.value, nodes.Dict):
+            return
+
+        dict_name = node.targets[0]
+        if len(node.targets) != 1 or not isinstance(dict_name, nodes.AssignName):
```
nickdrozd, line 37: It's annoying that we can't use the walrus operator here, because then these early returns could be fully streamlined.

clavedeluna: Indeed. https://endoflife.date/python 7 months

doc/data/messages/d/dict-init-mutate/bad.py
Pierre-Sassoulas: I'm using fruit and animals because I'm unimaginative but any example that a school children would understand works 😄 just no ``foo``, ``bar``, or ``baz``.

pylint/extensions/dict_init_mutate.py
```diff
@@ -0,0 +1,61 @@
...
+from pylint.checkers.utils import only_required_for_messages
+from pylint.interfaces import HIGH
+from pylint.lint.pylinter import PyLinter
+
+
+class DictInitMutateChecker(BaseChecker):
+    name = "dict-init-mutate"
+    msgs = {
+        "W3301": (
+            "Dictionary mutated immediately after initialization",
```
Pierre-Sassoulas, line 19: I think this message could be more explicit about the solution and the problem, but I'm not sure. We added a function to print big suggestions of dict recently, so maybe it's possible to do something like
```suggestion
            "Declare all known values directly in the constructor with '%s'",
```
Ie. result would be ``Declare all known values directly in the constructor with ' {"apple": 1, "banana": 10}'``  ``Declare all known values directly in the constructor with ' {"apple": 1, "melon" : 4, ... "banana": 10}'``

Pierre-Sassoulas (commented):
The change look pretty refined already, let's discuss how we can make the name / message clearer/better.

pylint/extensions/dict_init_mutate.py
```diff
@@ -0,0 +1,61 @@
...
+from pylint.checkers import BaseChecker
+from pylint.checkers.utils import only_required_for_messages
+from pylint.interfaces import HIGH
+from pylint.lint.pylinter import PyLinter
+
+
+class DictInitMutateChecker(BaseChecker):
+    name = "dict-init-mutate"
+    msgs = {
+        "W3301": (
```
jacobtylerwalls, line 18: Throughout pylint we have some warning inflation where things that should be `Convention`(C) are `Warning`(W). This seems like something that could be demoted to convention. What do you think?

clavedeluna: yeah I agree!

pylint/extensions/dict_init_mutate.py
clavedeluna: I like the updated language, but the suggestion with %s may be tricky because it may be multiple lines after. I'll see what I can do...

tests/functional/ext/dict_init_mutate.py
```diff
@@ -0,0 +1,33 @@
...
+config = {}  # [dict-init-mutate]
+config['dir'] = 'bin'
+config['user'] = 'me'
+config['workers'] = 5
+
+config = {}  # Not flagging calls to update for now
+config.update({"dir": "bin"})
+
+config = {}  # [dict-init-mutate]
+config['options'] = {}  # Identifying nested assignment not supporting this yet.
```
nickdrozd, line 23: I didn't give it much thought initially, but how difficult do you think it would be to catch this one?

tests/functional/ext/dict_init_mutate.py
```diff
@@ -0,0 +1,33 @@
+"""Example cases for dict-init-mutate"""
+# pylint: disable=use-dict-literal, invalid-name
+
+base = {}
+
+fruits = {}
+for fruit in ["apple", "orange"]:
+    fruits[fruit] += 1
```
nickdrozd, line 8: Outside the scope of this PR obviously, but shouldn't some kind of warning be raised here? After all, the code doesn't work.

clavedeluna: oof awful example, I'll change it

tests/functional/ext/dict_init_mutate.py
clavedeluna: For another PR...

pylint/extensions/dict_init_mutate.py
```diff
@@ -0,0 +1,118 @@
...
+
+        sibling_target = first_sibling.targets[0]
+        if not isinstance(sibling_target, nodes.Subscript):
+            return
+
+        sibling_name = sibling_target.value
+        if not isinstance(sibling_name, nodes.Name):
+            return
+
+        if sibling_name.name == dict_name.name:
```
clavedeluna, line 74: So all the complexity starting at this line is to be able to get the suggested dict str in the msg. I don't love how it turned out, it seems non-refactored, but refactoring was actually much harder than it looks. happy to take suggestions but I'd also like to improve this iteratively in future PRs.

pylint/extensions/dict_init_mutate.py
```diff
@@ -0,0 +1,118 @@
...
+    try:
+        key = node.targets[0].slice.value
+        value = node.value.value
+    except AttributeError:
+        # More complex cases will be handled later on
+        return ""
+
+    if isinstance(value, str):
+        value = f"'{value}'"
+    return f"'{key}': {value}"
```
Pierre-Sassoulas, line 30:
```suggestion
```
We can use ``node.as_string()``. Check ``RefactoringChecker._dict_literal_suggestion`` and the result in ``use-dict-literal`` functional tests.  I think it would be better to move this function to ``pylint.util`` and reuse it, it's pretty polished already. The only thing it could do better is to take care of really long terminating value when the last element is close to 60 chars.

pylint/extensions/dict_init_mutate.py
```diff
@@ -0,0 +1,118 @@
...
+
+        sibling_target = sibling.targets[0]
+        if not isinstance(sibling_target, nodes.Subscript):
+            return [node] + all_siblings
+
+        sibling_name = sibling_target.value
+        if not isinstance(sibling_name, nodes.Name):
+            return [node] + all_siblings
+
+        return [node] + self._get_dict_mutating_siblings(sibling, all_siblings)
```
nickdrozd, line 104: Much as I love recursion, this will crash Pylint if the dictionary is big enough:

```python
config = {}
config[0] = None
config[1] = None
...  # 2 ... 981
config[982] = None
config[983] = None
```

Hopefully nobody would write something like that by hand! But it could be an issue with generated files.

Maybe it would be best to limit the suggested replacement to checking only a few items (say, three). Then if there are more than that, add an elipsis: `Declare all known key/values when initializing the dictionary with 'config = {'dir': 'bin', 'user': 'me', 'workers': 5, ...}'`

This could be done in a later PR.

pylint/extensions/dict_init_mutate.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if sibling_name.name == dict_name.name:
+            siblings = self._get_dict_mutating_siblings(first_sibling)
+            suggestion = self._make_dict_suggestion(siblings, dict_name.name)
+            self.add_message(
+                "dict-init-mutate", node=node, args=(suggestion,), confidence=HIGH
+            )
+
+    def _get_dict_mutating_siblings(
+        self, node: nodes.Assign, all_siblings: List[nodes.Assign] | None = None
```
nickdrozd, line 82: Can't use optional-bar syntax in older Python versions :cry:

Pierre-Sassoulas: We have python > 3.7.2 we can do it as long as we use ``from __future__ import annotations``

nickdrozd:
Running this checker against Pylint itself yields a funny edge case:

```python
BUILTIN_HINTS = {"map": "Using a list comprehension can be clearer."}
BUILTIN_HINTS["filter"] = BUILTIN_HINTS["map"]
```

This gives a confusing error message:

```
pylint/extensions/bad_builtin.py:21:0: C3301: Declare all known key/values when initializing the dictionary with 'BUILTIN_HINTS = {'filter': Name.BUILTIN_HINTS(name='BUILTIN_HINTS')}' (dict-init-mutate)
```

Considering the difficulty in catching these edge cases and how much complexity it adds to the checker, it might be worth keeping the error message simple for now and thinking it over further. Perhaps that could be done in conjunction with figuring out how to catch nested dictionaries. The big challenge is determining what exactly goes in the dictionary.

pylint/extensions/dict_init_mutate.py
clavedeluna: AHA that's the magic line!!! I couldn't figure it out, thanks Pierre!

pylint/extensions/dict_init_mutate.py
clavedeluna: very good point, I actually I'm really not loving the additional suggestion. I would vote to simply `Declare all known key/values when initializing the dictionary` and not include the constructing dictionary help. It adds difficult to read and maintain code that doesn't add THAT much value. @Pierre-Sassoulas what do you think?

clavedeluna: That would be awesome. I'll update the PR to leave the general suggestion and open a new issue to add the improvement suggestion.

Pierre-Sassoulas (changes requested):
This looks pretty good at this point, the continuous integration needs to be fixed but we can merge otherwise 👍

doc/data/messages/d/dict-init-mutate/bad.py
Pierre-Sassoulas, line 1:
```suggestion
fruit_prices = {}  # [dict-init-mutate]
```

doc/data/messages/d/dict-init-mutate/good.py
```diff
@@ -0,0 +1 @@
+fruits = {"apple": 1, "banana": 10}
```
Pierre-Sassoulas, line 1:
```suggestion
fruit_prices = {"apple": 1, "banana": 10}
```

Pierre-Sassoulas (changes requested):
It's possible to rebase now that #7833 is merged.

Pierre-Sassoulas (approved):

uzzell:
I'm concerned about the warning for https://github.com/pandas-dev/pandas/blob/d1ecf63e2040daecad6e0a8485a35e8a23393795/pandas/tests/reshape/test_get_dummies.py#L110 (item 3 in the Pandas list in [this github-actions comment](https://github.com/PyCQA/pylint/pull/7794#issuecomment-1325382772)):
```
expected_counts = {"int64": 1, "object": 1}
expected_counts[dtype_name] = 3 + expected_counts.get(dtype_name, 0)
```
Here, `dtype_name` depends on an argument of the method that contains this snippet.  So, I don't see how to rewrite the snippet.  Doing `expected_counts = {"int64": 1, "object": 1, dtype_name: 3 + expected_counts.get(dtype_name, 0)}` would throw an error.

clavedeluna:
https://github.com/PyCQA/pylint/issues/7997

PR #7796, opened 2022-11-18
Enable broad-try-clause check

nickdrozd:
## Type of Changes

## Description

This enables the `broad-try-clause` check and sets the `max-try-statements` config option to not flag anything. I think it would be a good idea to try and lower that number, but that can be done later.


Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas:
It seems like the primer are using our pylintrc by default and not the default values for pylint which is probably not what we want..

PR #7797, opened 2022-11-19
Reduce `max-locals` from 25 to 18

nickdrozd:
## Type of Changes

## Description

This PR lowers the `max-locals` config value from 25 to 18. A few outlier functions with abnormally many local variables required some trimming. My approach was to cut any variable used only once and simply inline the value. After a little rearranging, there might even be some tiny performance improvements.

Why 18? Because `ImportsChecker._check_imports_order` stubbornly resisted my attempts to get rid of any of its variables :smile:

### DIFF
--- a/pylint/checkers/spelling.py
+++ b/pylint/checkers/spelling.py
@@ -417,10 +417,17 @@ class SpellingChecker(BaseTokenChecker):
                 col += word_start_at
                 if starts_with_comment:
                     col += 1
-                indicator = (" " * col) + ("^" * len(word))
                 all_suggestion = "' or '".join(suggestions)
-                args = (word, original_line, indicator, f"'{all_suggestion}'")
-                self.add_message(msgid, line=line_num, args=args)
+                self.add_message(
+                    msgid,
+                    line=line_num,
+                    args=(
+                        word,
+                        original_line,
+                        (" " * col) + ("^" * len(word)),
+                        f"'{all_suggestion}'",
+                    ),
+                )
 
     def process_tokens(self, tokens: list[tokenize.TokenInfo]) -> None:
         if not self.initialized:
--- a/pylint/lint/expand_modules.py
+++ b/pylint/lint/expand_modules.py
@@ -75,13 +75,11 @@ def expand_modules(
     path = sys.path.copy()
 
     for something in files_or_modules:
-        basename = os.path.basename(something)
         if _is_ignored_file(
             something, ignore_list, ignore_list_re, ignore_list_paths_re
         ):
             continue
-        module_path = get_python_path(something)
-        additional_search_path = [".", module_path] + path
+        additional_search_path = [".", get_python_path(something)] + path
         if os.path.exists(something):
             # this is a file or a directory
             try:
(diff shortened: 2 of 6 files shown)


pylint/lint/expand_modules.py
```diff
@@ -131,11 +128,14 @@ def expand_modules(
...
-        )
-        if has_init or is_namespace or is_directory:
+        if (
+            is_namespace
+            or is_directory
+            or (
+                not (modname.endswith(".__init__") or modname == "__init__")
+                and os.path.basename(filepath) == "__init__.py"
+            )
+        ):
```
jacobtylerwalls, line 138: This strikes me as an instance of something becoming worse.

jacobtylerwalls (commented):
Hi Nick, thanks for raising the idea. I think I'm not very positive on this, as I worry it addresses a symptom and not a root cause, which is that some of these methods should be decomposed. What do you think?

nickdrozd:
> Hi Nick, thanks for raising the idea. I think I'm not very positive on this, as I worry it addresses a symptom and not a root cause, which is that some of these methods should be decomposed. What do you think?

I agree that some of these could use a more substantial rewrite. But maybe the path forward can't reveal itself until we hack through the brush :herb:

pylint/lint/expand_modules.py
nickdrozd: If I'm reading it right, this is actually a micro performance improvement. `is_namespace` and `is_directory` are already sitting around, so calculating `has_init` before checking them is potentially unnecessary.

I guess the name `has_init` gives a little clue. I could add a comment to that effect if you like.

jacobtylerwalls: Nah, no need to touch this specific example, was just saying that some of these locals seemed to be here for self-documenting. (Although you have a point that there's a micro-optimization to be had.) I suppose my thought about doing this PR was why remove nicely self-documenting locals just to get the locals count down? Won't stand in the way though if my view is not seconded 🗣️

nickdrozd: Yes, there is a risk of sacrificing readability just for the sake of chasing stats. But my feeling from looking at these functions is that they have reached a point where the marginal value of additional variables is negative. It can be bewildering to keep track of what all these variables mean, particularly when so many are used only once.

I'm happy to add some back though, or add comments.

pylint/checkers/spelling.py
```diff
@@ -417,10 +417,17 @@ def _check_spelling(self, msgid: str, line: str, line_num: int) -> None:
...
+                self.add_message(
+                    msgid,
+                    line=line_num,
+                    args=(
+                        word,
+                        original_line,
+                        (" " * col) + ("^" * len(word)),
+                        f"'{all_suggestion}'",
+                    ),
+                )
```
Pierre-Sassoulas, line 430: Here we go from 2 lines to 7 just for the sake of following a metric. Imo the code is worse. There's also the very real possibility of introducing bug while doing it. Let's not refactor for refactoring sake.

Pierre-Sassoulas (commented):
Thank you for opening the MR @nickdrozd. There might be improvement here but I'm not sure the risk and triaging necessary is worth the possible benefit.

There are huge area of improvement in astroid/pylint though. Imo the things that would make the code better the most without being a feature or bug fix are:
- typing in astroid / pylint
- Fixing cyclic import in astroid
- Performance analysis on start time in astroid
- Multiprocessing groundwork to be able to parallelize pylint
- Removing all legacy functional tests in https://github.com/PyCQA/pylint/blob/main/tests/test_func.py#L127 (13 left), which means making the functional tests framework evolve a little.

All of those are massive undertaking (except legacy functional tests ..maybe) and can be improved upon progressively.

nickdrozd:
Okay, I'll close this one...for now :smile:

PR #7799, opened 2022-11-19
Merge maintenance/2.15.x in main following 2.15.6 release

Pierre-Sassoulas:


doc/whatsnew/fragments/7214.bugfix
```diff
@@ -1,3 +1,3 @@
-Message send to reporter are now copied so a reporter cannot modify the message sent to other reporters.
+Messages sent to reporter are now copied so a reporter cannot modify the message sent to other reporters.
 
 Closes #7214
```
cdce8p, line 3: This fragment should probably be removed as it's included in the changelog for `2.15.6`.

cdce8p (changes requested):

doc/whatsnew/fragments/7214.bugfix
Pierre-Sassoulas: I think it's voluntary to include the released notes from patches in minor's whatsnew. It's  also convenient as when you merge back the removal need to be done manually again.

Pierre-Sassoulas: ~~I can't find the discussion about it :(~~ Here we go : https://github.com/PyCQA/pylint/pull/7426#issuecomment-1238692051

jacobtylerwalls (approved):

doc/whatsnew/fragments/7214.bugfix
cdce8p: > The only reason we have a diff on this fragment versus the others is because I suggested a grammar fix on this fragment that Pierre committed at release time.

👍🏻

> I think it's voluntary to include the released notes from patches in minor's whatsnew. It's also convenient as when you merge back the removal need to be done manually again.

I was thinking about it the wrong way. We still "need" them for `main` as they technically aren't released yet, just backported.

cdce8p (approved):

PR #7802, opened 2022-11-19
Add flag `--clear-cache-post-run` to support server-like usage

jacobtylerwalls:
Use this flag if you expect the linted files to be altered between runs, for instance, if using pylint in a server-like mode. The flag clears astroid's in-memory caches.

## Type of Changes

## Description

Refs https://github.com/microsoft/vscode-pylint/issues/129

### DIFF
--- a/pylint/utils/utils.py
+++ b/pylint/utils/utils.py
@@ -57,7 +57,6 @@ GLOBAL_OPTION_BOOL = Literal[
     "suggestion-mode",
     "analyse-fallback-blocks",
     "allow-global-unused-variables",
-    "clear-cache-post-run",
 ]
 GLOBAL_OPTION_INT = Literal["max-line-length", "docstring-min-length"]
 GLOBAL_OPTION_LIST = Literal["ignored-modules"]
--- a/pylint/utils/utils.py
+++ b/pylint/utils/utils.py
@@ -57,6 +57,7 @@ GLOBAL_OPTION_BOOL = Literal[
     "suggestion-mode",
     "analyse-fallback-blocks",
     "allow-global-unused-variables",
+    "clear-cache-post-run",
 ]
 GLOBAL_OPTION_INT = Literal["max-line-length", "docstring-min-length"]
 GLOBAL_OPTION_LIST = Literal["ignored-modules"]
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -6,12 +6,14 @@ from __future__ import annotations
 
 import os
 from pathlib import Path
+from tempfile import TemporaryDirectory
 
 import pytest
 from pytest import CaptureFixture
 
 from pylint.interfaces import CONFIDENCE_LEVEL_NAMES
 from pylint.lint import Run as LintRun
+from pylint.testutils import create_files
 from pylint.testutils._run import _Run as Run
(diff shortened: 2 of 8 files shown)


Pierre-Sassoulas (commented):
I'm not sure if I understand the use case here. If you're launching pylint, through the entrypoint the cache is going to be cleared after a run in any case ? Shouldn't this be an option specific to the API ?

jacobtylerwalls:
> I'm not sure if I understand the use case here. If you're launching pylint, through the entrypoint the cache is going to be cleared after a run in any case ? Shouldn't this be an option specific to the API ?

Before, no caches were inetntionally cleared, it was just that when pylint exits with sys.exit, that takes care of killing the astroid module. If you're not exiting, i.e. with `exit=False` in the test I just pushed, which fails on main, or whatever Microsoft is doing (`runpy`?), then you still have live astroid modules with stale data in their module-level caches.

This was the solution discussed at https://github.com/microsoft/vscode-pylint/issues/129.

pylint/utils/utils.py
```diff
@@ -57,6 +57,7 @@
     "suggestion-mode",
     "analyse-fallback-blocks",
     "allow-global-unused-variables",
+    "clear-cache-post-run",
```
DanielNoord, line 60: This shouldn't be used anymore as far as I know, so if it is necessary for tests to pass we have broken something 😅

DanielNoord (changes requested):

pylint/utils/utils.py
jacobtylerwalls: It wasn't, so I can take it out. I wondered if it was here for a public API of some sort. Will do.

jacobtylerwalls, line 60:
```suggestion
```

DanielNoord: The `argparse` refactor should make all of this unnecessary!

tests/config/test_config.py
```diff
@@ -155,3 +157,19 @@ def test_argument_separator() -> None:
...
+
+
+def test_clear_cache_post_run() -> None:
+    modname = "changing.py"
+    with TemporaryDirectory() as tmp_dir:
+        create_files([modname], tmp_dir)
+        module = tmp_dir + os.sep + modname
+        # Run class does not produce the wanted failure
+        # must use LintRun to get pylint.lint.Run
+        runBeforeEdit = LintRun([module, "--clear-cache-post-run=y"], exit=False)
```
DanielNoord, line 169: Why is the suddenly camel case?

DanielNoord (changes requested):
Perhaps we should discuss with MS/VS Code team if this indeed solves their issue?

_Primer flakiness is once again interesting..._

Pierre-Sassoulas:
> Primer flakiness is once again interesting...

I think the commit with which we compare the PR is the latest commit on main main when it should be the most recent commit in common with the feature branch and main. I want to make the primer better but I have too much on my plate right now.

jacobtylerwalls:
> Perhaps we should discuss with MS/VS Code team if this indeed solves their issue?

Yes, I can wait to merge until we hear from them; I've already posted on their issue.

DanielNoord (approved):

PR #7805, opened 2022-11-20
Permit to launch tests when backporting on the maintenance branch

Pierre-Sassoulas:
## Type of Changes

## Description

Also remove the changelog job for everything except main so we don't have to laboriously label everything "skip news" all the time.

Refs #7804

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-      - 2.*
+      - "maintenance/**"
   pull_request: ~
 
 env:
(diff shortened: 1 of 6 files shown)


.github/workflows/checks.yaml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-      - 2.*
```
Pierre-Sassoulas, line 7: @cdce8p was this a remnant of an old configuration we used originally ?

Pierre-Sassoulas:
Closing because it's easier to test directly on #7804

Issue #7609, opened 2022-10-11
False positive used-before-assignment for variable annotations when using TYPE_CHECKING


mew1033:
### Bug description

I know this has been looked at before, but I just ran into it again with pylint 2.15.4. If an import is guarded behind TYPE_CHECKING, I'm getting an `used-before-assignment` error when using that import in a type definition.

### Command used

```shell
pylint file.py
```

### Pylint output

```shell
file.py:7:19: E0601: Using variable 'PublishBatchRequestEntryTypeDef' before assignment (used-before-assignment)
```

### Expected behavior

It shouldn't error

### Pylint version

```shell
pylint 2.15.4
astroid 2.12.11
Python 3.9.14 (main, Sep  7 2022, 23:43:29) 
[GCC 9.4.0]
```

mbyrnepr2:
Thanks for reporting this @mew1033. Can you please provide a minimal reproducible example to help us dig into this one?

mew1033:
@mbyrnepr2 So sorry, I got the example ready and then completely forgot to add it to the issue. Here's a snippet that shows the issue. If you remove the `if TYPE_CHECKING:` line, the issues goes away.

```python
from typing import TYPE_CHECKING
import boto3

if TYPE_CHECKING:
    from mypy_boto3_sns.type_defs import PublishBatchRequestEntryTypeDef


def publish_to_sns():
    """
    Create SNS messages and publish them
    """
    sns_client = boto3.client("sns")
    messages: list[PublishBatchRequestEntryTypeDef] = []
```

mbyrnepr2:
No worries - thank you!

jacobtylerwalls:
Thanks! Per [PEP 526](https://peps.python.org/pep-0526/#runtime-effects-of-type-annotations), there is a difference in the runtime effect of variable annotations versus function annotations. Given that, the requirement in 3159b1744caee14068b4d875bb5adaa5ffbebd44 that both branches of an `if TYPE_CHECKING:` case define the name is too strict for variable annotations.

jacobtylerwalls:
Some helpful test cases given in #7773

PR #7806, opened 2022-11-20
Emit `used-before-assignment` for further imports guarded by TYPE_CHECKING

### LINKED ISSUE #7539: Audit remaining uses of `TYPING_TYPE_CHECKS_GUARDS`
Along the lines of #7524 and #6787, there is probably a false positive for `used-before-assignment`/`unused-variable` lurking in the use of `defstmt_parent.test.as_string() in TYPING_TYPE_CHECKS_GUARDS`, which is not robust against importing `typing` under an alias.

I advised us against doing a drop-in replacement by `in_type_checking_block` without writing more tests. For discussion on that see https://github.com/PyCQA/pylint/pull/7525#discussion_r982354594.

To close this issue, we need a contributor to write a regression test that fails on `main` and ideally a test that also would have failed if we had taken the simple approach of just doing the drop-in replacement in https://github.com/PyCQA/pylint/pull/7525#discussion_r982024065.

jacobtylerwalls:
## Type of Changes

## Description
Previously, this message was only emitted for imports guarded directly under TYPE_CHECKING, not guarded two if-branches deep, nor when TYPE_CHECKING was imported from typing under an alias.

Closes #7539


Pierre-Sassoulas (dismissed):
``in_type_checking_block`` is indeed a lot more accurate 😄

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -74,3 +78,12 @@ def function(self, var: int) -> None:
 
     def other_function(self) -> None:
         _x: MyThirdClass = self
+
+
+class MyFourthClass:  # pylint: disable=too-few-public-methods
+    """Class to test conditional imports guarded by TYPE_CHECKING two levels
+    up then used in function annotation. See https://github.com/PyCQA/pylint/issues/7539"""
+
+    def comparator(self, comparator: math.isclose, first, second):  # [used-before-assignment]
```
jacobtylerwalls, line 87:
```suggestion
    def is_close(self, comparator: math.isclose, first, second):  # [used-before-assignment]
```

jacobtylerwalls:
The false positive in flask is already documented at #7609. It's true that this PR will ever so slightly drive more traffic into that false positive, but I don't think that's a reason not to merge it.

jacobtylerwalls:
@Pierre-Sassoulas This one's ready for review, if you get a chance.

Pierre-Sassoulas (approved):
LGTM 👌 Thanks for the reminder !

Issue #7305, opened 2022-08-13
proposal: a simplified and generalized invalid-name config


bukzor:
### Current problem

The current configuration system for naming style in pylint can be seen as simultaneously overly rigid and overly flexible. There are two distinct ways to configure naming schemes. Primarily there is a named-scheme selection where users choose among lower_case UPPER_CASE PascalCase, etc., and there is a regex-based (`-rgx`) configuration where the user meant to write a regular expression that matches all (and only) permissible variable names.

As a concrete example:

```ini
[BASIC]
const-naming-style=UPPER_CASE
# from https://github.com/openqasm/openqasm/blob/main/.pylintrc
variable-rgx=[a-z_][a-z0-9_]{2,30}$
```

The named-style is convenient for users, but gives no clarity as to what exactly will match, and admits no ability to make small tweaks to its definition. The regex configuration is its near opposite, being completely unambiguous and easy to modify but painful-to-impossible to use for the average human.

### Desired solution

This is my (tentative!) proposal: (keep in mind that all details will/would be modified to reflect maintainers' requirements)

Let's move this configuration out of "basic" to a new, separate "naming" namespace.
```ini
[NAMING]
variable = lower_case

[NAMING.lower_case]
allowed-characters = [a-zA-Z0-9_]
allowed-first-character = [a-z]
min-length = 3
max-length  = 0  # no limit
```

Or, alternatively: (I prefer the above)

```ini
[NAMING]
upper-case.allowed-characters = [A-Z0-9_]
upper-case.allowed-first-character = [A-Z]
upper-case.min-length = 0  # no limit
upper-case.max-length  = 30
```

Each naming style is be defined by just four characteristics: allowed characters, allowed first-letter character, minimum and maximum length. If we can expose those quadruplets to the configuration, then the configuration is easy-to-use (users can largely ignore these fine details), unambiguous, easy to modify, and there's no (obvious) need to expose users to the full footgun of regular expressions.

Note: The "character" configs are required to be a single character-class (i.e. a single-char regex beginning and ending with square brackets). This limitation allows excellent error messaging in the case of a invalid-name error, but also enables programmatic use.

This obviously enables the potential (but not requirement) to allow users to define custom named naming-schemes.

All naming schemes would then be defined as (approximately) `f'_{{0,2}}{allowed-first-character}{allowed-characters}{{{min-length},{max-length}}}'`

[...]

PR #7808, opened 2022-11-20
Update disallowed-name to flag module-level variables

### LINKED ISSUE #3701: ``disallowed-name`` check no longer checks non-constants
Commit https://github.com/PyCQA/pylint/commit/3422e4adc718c60a467d66ab0dedf4a7032c4fc5  changed the behavior of the `invalid-name` check, but also changed the behavior of `blacklisted-name`, which appears to be unintentional (and undocumented).

### Steps to reproduce

demo.py:

```python
foo = {}.keys()
foo = None
```

`pylint demo.py -d all -e blacklisted-name`

### Current behavior

From pylint 2.5.0 and later:

```
demo.py:2:0: C0102: Black listed name "foo" (blacklisted-name)
```

### Expected behavior

From pylint 2.4.4:

```
demo.py:1:0: C0102: Black listed name "foo" (blacklisted-name)
demo.py:2:0: C0102: Black listed name "foo" (blacklisted-name)
```

### pylint --version output

pylint 2.5.0
astroid 2.4.2
Python 3.7.4 (default, Oct 16 2019, 10:35:50) 
[Clang 11.0.0 (clang-1100.0.33.8)]

clavedeluna:
## Type of Changes

## Description

Closes #3701

The original issue brought up the fact that `foo = {}.keys()` was not getting flagged for disallowed-name. I discovered that in some past refactoring, one `else:` case wasn't included, which is checking all module-level var names.


clavedeluna:
I'm going to also point out a couple interesting examples from functional tests that don't seem to be flagged correctly.

In this first example, why would only Alias5 and Alias6 be flagged, and not all?
```
import collections.abc
from collections.abc import Callable
from typing import TYPE_CHECKING, Optional, Union

Alias1 = Optional[Callable[[int], None]]  # [broken-collections-callable]
Alias2 = Union[Callable[[int], None], None]  # [broken-collections-callable]
Alias3 = Optional[Callable[..., None]]
Alias4 = Union[Callable[..., None], None]
Alias5 = list[Callable[..., None]] # this is getting flagged as invalid-name
Alias6 = Callable[[int], None] # this is getting flagged as invalid-name
```

in this example, the first is a constant, so it's good, but the second `SOME_CLASS_VAR` isn't a constant. Shouldn't this be flagged as invalid-name, and an author should instead use lower_case?
```
class Something:
    SOME_CLASS_CONST = 1
    SOME_CLASS_VAR = [1, 2]  # shouldn't this be  [invalid-name]?
```

tests/functional/n/names_in__all__.py
```diff
@@ -27,7 +27,7 @@ class Dummy:
     """A class defined in this module."""
     pass
 
-DUMMY = Dummy()
+dummy = Dummy()
```
DanielNoord, line 30: I don't really understand what you're changing in this PR, but this change is incorrect.

`DUMMY` is a "constant" in `pylint`/`Python` terms and should be uppercased.

cdce8p: > `DUMMY` is a "constant" in `pylint`/`Python` terms and should be uppercased.

I agree with Daniel here. PEP 8 never differentiates between a constants and (true) constants. The pylint interpretation has been "consider every variable at module level a constant" for a long time. Changing it now would result in a massive breaking change that's just unreasonable.

Just an example. The change would result in over 7500+ `invalid-name` errors for Home-Assistant.

I'm strong -1.

cdce8p (dismissed):

Pierre-Sassoulas:
Counter argument to not doing anything would be that ``invalid-name`` is unusable (the first thing I always do in a new project is to disable it).  It's probably one of the most annoying thing when trying to use pylint. It's in pylint so it must be important, so you try to follow it at first... but then it does not make a lot of sense. Well we're not talking about #7305 there but of the more minor problem of module level "constant", so let's focus on that.

What we can't do do is make it even more annoying by increasing the constraints (7500+ new messages in a well maintained code base is just nuts). There's two way forward on this particular problem imo:
- keeping it like it is so those that actually use it do not have a breaking change (it's not very satisfying considering this check is one of the major pain point of pylint)
- Making the constraints laxer. Allowing both constant and variable style at the module level. Instead of trying to treat immutable and mutable differently: just treat them all with leniency. python does not have constants, we can't know, trust the user on this one.

I think it's clear what I would favor here 😄

cdce8p:
> Counter argument to not doing anything would be that `invalid-name` is unusable (the first thing I always do in a new project is to disable it). It's probably one of the most annoying thing when trying to use pylint.

I would disagree. In my personal experience it works fine as it is.

Clearly not everybody feels that way 😉 but that might rather be a thing that could be solved by config profiles. There was a small discussion about it in #3512.

> * keeping it like it is so those that actually use it do not have a breaking change (it's not very satisfying considering this check is one of the major pain point of pylint)
> * Making the constraints laxer. Allowing both constant and variable style at the module level. Instead of trying to treat immutable and mutable differently: just treat them all with leniency. python does not have constants, we can't know, trust the user on this one.

I would prefer the first option. For me it isn't so much weather or not it's a true constant but rather to know if a variable is defined at module level. Having the distinction between upper case and lower case does help with that.

Additionally, it adds a common style that, even if not perfect, everybody (who uses it regularly) understands. Relaxing the requirement will just lead to mixed cases which will require additional attention from reviewers. Especially for beginners, without experience in other languages, it won't be obvious when to use what and I suspect they will frequently use the wrong one.

Pierre-Sassoulas:
The issue is that there is no consensus on what the style should be for module level variable, pep8 is lax about it:
- https://peps.python.org/pep-0008/#constants
- https://peps.python.org/pep-0008/#variable-annotations
So I don't think pylint should take a stance about this, especially based on what we think is mutable / immutable which is another thing entirely.

The solution I would propose is to permit only **consistent** uppercase or **consistent** snake case but not anything else. We could add an option deactivated by default for that. The discussion about this is shattered in a dozen issues but I thought this was consensual.

cdce8p:
> So I don't think pylint should take a stance about this, especially based on what we think is mutable / immutable which is another thing entirely.

We don't do that. The difference is based on the scope a variable is defined in, e.g. `module` or `function`. IMO much easier to understand.

> The solution I would propose is to permit only **consistent** uppercase or **consistent** snake case, (or dunder casing, which is consistent snake case) but not anything else. We could add an option deactivated by default for that. The discussion about this is shattered in a dozen issues but I thought this was consensual.

That's already possible with naming styles. https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html#predefined-naming-styles For example, this should work
```ini
# pylintrc
const-naming-style=snake_case
variable-naming-style=snake_case
```

Pierre-Sassoulas:
> We don't do that. The difference is based on the scope a variable is defined in, e.g. module or function. IMO much easier to understand.

See https://github.com/PyCQA/pylint/pull/7808#issuecomment-1321234945

> That's already possible with naming styles. 

Well if you do ``const-naming-style=snake_case`` then you can't have const in UPPERCASE, right ? You could however create a complicated regex that match both. Maybe that's the easy fix.

clavedeluna:
I've now updated this PR to only modify disallowed-name, not invalid-name.

pylint/checkers/base/name_checker/checker.py
```diff
@@ -543,7 +549,11 @@ def _should_exempt_from_invalid_name(node: nodes.NodeNG) -> bool:
             warnings = bad_name_group.setdefault(match.lastgroup, [])  # type: ignore[union-attr, arg-type]
             warnings.append((node, node_type, name, confidence))
 
-        if match is None and not _should_exempt_from_invalid_name(node):
+        if (
+            match is None
+            and not disallowed_check_only
```
clavedeluna, line 554: This could've been done in other ways, too, including integrating with `_should_exempt_from_invalid_name` above but I chose this for simplicity. Lmk if another implementation makes more sense.

DanielNoord (dismissed):
Makes sense to me, but I'd like the opinion of others as well.

tests/functional/b/blacklisted_name.py
```diff
@@ -1,4 +0,0 @@
-# pylint: disable=missing-docstring
```
Pierre-Sassoulas, line 1: 👍 good work merging those two

Pierre-Sassoulas (approved):
LGTM ! I don't think we have to resolve the previous discussion to fix this issue and merge. We're going to have to do it regardless of the result of the discussion.

clavedeluna:
This should be good to merge

DanielNoord:
I'll let @Pierre-Sassoulas do this as I'm not sure if this should be back ported.

Pierre-Sassoulas:
This was a bug, but it's going to make new messages appear so let's delay it to 2.16.0 as we advertise the patch version to only remove message in the doc: https://pylint.pycqa.org/en/latest/user_guide/installation/upgrading_pylint.html#upgrading-pylint

clavedeluna:
any reason this can't go out with 2.16.0?

Pierre-Sassoulas:
No but, I'm trying to wrap up 2.16.0 and focus on what is absolutely necessary (known unreleased issues).

doc/data/messages/d/disallowed-name/good.py
```diff
@@ -1,2 +1,2 @@
-def something_meaningful():
-    pass
+def print_fruit():
+    print("apples")
```
Pierre-Sassoulas, line 2: We should add the print in Bad.py too if we change this.

tests/functional/d/disallowed_name.txt
```diff
@@ -0,0 +1,5 @@
+disallowed-name:3:0:3:7:baz:"Disallowed name ""baz""":UNDEFINED
```
Pierre-Sassoulas, line 1: I think it's high confidence, right ?

Pierre-Sassoulas (changes requested):
Two nits and let's merge, good job!

doc/data/messages/d/disallowed-name/good.py
clavedeluna: it's already there :)
```
def foo():  # [disallowed-name]
    print("apples")

```

Pierre-Sassoulas (approved):
Great !

PR #7795, opened 2022-11-18
Fix ``deprecated-method`` false positive

clavedeluna:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Original PR that added this code https://github.com/PyCQA/pylint/pull/4619/files
it seems that the unit test that was added led the code, which is a good practice, except I believe the unit test itself was incorrect and confused the naming of `deprecated_method`. This wasn't a necessary change at all, and in fact led to a false positive. 

Closes #5886

PR #7809, opened 2022-11-20
Fail pytest runs on warnings

jacobtylerwalls:
## Type of Changes

## Description
I've opened a few PRs over time to clean up warnings in the test suite when I see them. Figured we might as well fail CI so that we don't merge with known issues.

Failure expected due to #7795, mentioned here: https://github.com/PyCQA/pylint/pull/7795#issuecomment-1321240480. Edit: resolved.


tests/pyreverse/test_inspector.py
```diff
@@ -30,7 +30,14 @@ def project(get_project: GetProjectCallable) -> Generator[Project, None, None]:
     with _test_cwd(TESTS):
         project = get_project("data", "data")
         linker = inspector.Linker(project)
-        linker.visit(project)
+        if "clientmodule_test" in str(project):
+            with pytest.warns(DeprecationWarning):
+                # tests.data.clientmodule_test.py emits DeprecationWarning
+                # due to use of __implements__
+                # TODO after deprecation: remove __implements__
```
Pierre-Sassoulas, line 37:
```suggestion
                # TODO: 3.0 remove 'pytest.warns' once __implements__ is removed
```

clavedeluna (dismissed):
Really good catch here and sorry one of my PRs introduced a warning. Looks like now it's failing the test run with it! Let me know if you want to fix the warnings in this PR or you want me to intervene via another issue/ PR.

DudeNr33:
I have to look at this more closely this evening. Currently I don't see why this change should work as intended.

pyproject.toml
```diff
@@ -89,6 +89,7 @@ test = "pytest"
 testpaths = ["tests"]
 python_files = ["*test_*.py"]
 addopts = "--strict-markers"
+filterwarnings = "error"
```
clavedeluna, line 92: will this make pytest fail at any len(warnings) > 0?

jacobtylerwalls:
> Currently I don't see why this change should work as intended.

There's one pyreverse test that still makes use of `__implements__`, so it raises a DeprecationWarning when it's loaded. Or were you asking about something else?

pyproject.toml
jacobtylerwalls: Yep!

jacobtylerwalls:
> Let me know if you want to fix the warnings in this PR or you want me to intervene via another issue/ PR.

If you have time to help, that would be great. You can either push to my branch (GH's own editor makes that easier than it used to) or open a new PR, whatever works 👍🏻

tests/pyreverse/test_inspector.py
```diff
@@ -30,7 +30,14 @@ def project(get_project: GetProjectCallable) -> Generator[Project, None, None]:
     with _test_cwd(TESTS):
         project = get_project("data", "data")
         linker = inspector.Linker(project)
-        linker.visit(project)
+        if "clientmodule_test" in str(project):
```
DudeNr33, line 33: This is never true, as the `project` is always the same here (the `data` directory underneath `/tests`). The string representation of `project` does not contain the substring `clientmodule_test`.

I would suggest just get rid of the `if` condition here.

DudeNr33 (changes requested):
Managed to squeeze the review in. As suspected the change does not help in this case, but I can understand the confusion (the `Generator` return type together with `TESTS` sounding like plural it seems like we are iterating over different directories here).

Let me know if you need more help!

tests/pyreverse/test_inspector.py
jacobtylerwalls: I tried that, but I received many failures. I'll try again tonight. 🤔

DudeNr33: Hm, but with the current change we are always hitting the `else`, and nothing has changed effectively compared to before. 
I haven't looked at how to actually properly catch the warning, my guess would be that there are some other tests that don't catch this Warning yet (the majority of the legacy tests for `pyreverse` work with the same test data). I will also take a closer look at it this or tomorrow evening!

DudeNr33: Neither on your branch nor on main could I reproduce the deprecation warnings on the `pyreverse` tests. 
On a [recent CI run](https://github.com/PyCQA/pylint/actions/runs/3514404778/jobs/5888375701) there is only one `DeprecationWarning` for some other test case, but not for `pyreverse`.
Am I missing something here like an option I have to turn on first?

jacobtylerwalls: Sorry @DudeNr33 you're quite right, the `else` branch always executes. I've just pushed a change to just ignore warnings in this fixture.

Pierre-Sassoulas (approved):
Seems like no contributers used python 3.7 anymore :D

PR #7815, opened 2022-11-21
[github actions] Permit to launch tests when backporting on maintenance branch

Pierre-Sassoulas:
## Type of Changes

## Description

Remove changelog so we don't have to laboriously label everything "skip news" all the time. Cherry-picked to be tested in #7804 where it's going to be reviewed.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -25,6 +25,6 @@ jobs:
         )
       )
     steps:
-      - uses: tibdex/backport@v2
+      - uses: tibdex/backport@2e217641d82d02ba0603f46b1aeedefb258890ac
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -8,10 +8,7 @@ on:
       - labeled
 
 permissions:
-  checks: read
-  contents: write
   pull-requests: write
-  statuses: read
 
 jobs:
   backport:
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,7 +1,5 @@
 name: Backport
 on:
-  branches:
-    - main
   pull_request_target:
     types:
       - closed
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,5 +1,7 @@
 name: Backport
 on:
+  branches:
+    - main
   pull_request_target:
     types:
       - closed
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,13 +1,10 @@
 name: changelog
 
 on:
-  push:
-    branches:
-      - main
-    branches-ignore:
-      - "maintenance/**"
   pull_request:
     types: [opened, synchronize, labeled, unlabeled, reopened]
+    branches:
+      - main
     branches-ignore:
       - "maintenance/**"
 env:
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,9 +1,15 @@
 name: changelog
 
 on:
+  push:
+    branches:
+      - main
+    branches-ignore:
+      - "maintenance/**"
   pull_request:
     types: [opened, synchronize, labeled, unlabeled, reopened]
-
+    branches-ignore:
+      - "maintenance/**"
 env:
   CACHE_VERSION: 1
   KEY_PREFIX: base-venv
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -3,8 +3,6 @@ name: changelog
 on:
   pull_request:
     types: [opened, synchronize, labeled, unlabeled, reopened]
-    branches:
-      - main
     branches-ignore:
       - "maintenance/**"
 env:
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -9,7 +9,7 @@ on:
   push:
     branches:
       - main
-
+      - "maintenance/**"
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -9,7 +9,6 @@ on:
   push:
     branches:
       - main
-      - "maintenance/**"
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/primer_run_pr.yaml
+++ b/.github/workflows/primer_run_pr.yaml
@@ -5,9 +5,6 @@
 name: Primer / Run
 
 on:
-  branches:
-    - main
-    - "maintenance/**"
   pull_request:
     paths:
       - "pylint/**"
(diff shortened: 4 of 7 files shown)


.github/workflows/primer_run_pr.yaml
```diff
@@ -5,6 +5,9 @@
 name: Primer / Run
 
 on:
+  branches:
+    - main
+    - "maintenance/**"
```
DanielNoord, line 10: Like I said, I don't think we should ignore here.

Also doesn't this actually run this on `main` now instead of not running it?

cdce8p:
A bit late, but I don't really feel comfortable giving an action that much permissions.
https://github.com/PyCQA/pylint/blob/331e79389e10b9720d25dcdb6120ccf5cc33701c/.github/workflows/backport.yml#L8-L12

I do understand what you're trying to do here. However, automatically pushing commits to a release branch is a bit much. IMO there are (at least) two better options
1. Do what CPython does: After a label is applied a bot opens a backport PR which needs to be approved / checks run before it is merged.
2. Or have a (simple) script which automates the local cherry-picking. The changes are then combined in a PR which targets the maintenance branch. Home-Assistant does this as part of their https://github.com/home-assistant/hass-release package. (It also automates the release-notes generation and other things.)

.github/workflows/primer_run_pr.yaml
Pierre-Sassoulas: I'm not trying to ignore this is ``branches`` not ``branches-ignore``. If I understand well this is something that should be run only on the PR branch and we should ``branches-ignore`` both main and maintenance/** ?

Pierre-Sassoulas:
> I don't really feel comfortable giving an action that much permissions.

Would it be okay if we own the actions (i.e. we use my fork of it) ?

What CPython does look like exactly what this action is doing:
- #7795 is labelled as backport maintenance/2.15.x
- Action open #7804 automatically
```diff
- - I modify the github actions so tests are launched on #7804 because it's still a POC
```
- Tests execute including primers, maintainers approve and merge #7804

cdce8p:
> What CPython does look like exactly what this action is doing:

My understanding was that the action only opens PRs if the auto cherry-pick fails. Normally it would just push the changes to the maintenance branch.
https://github.com/PyCQA/pylint/blob/7fd56d9e30f55e3b222fe46b3ba927ca02670123/doc/development_guide/contributor_guide/release.md#L55-L65

I think it would be fine if the action `always` opens a PR for the backport.

> > I don't really feel comfortable giving an action that much permissions.
> 
> Would it be okay if we own the actions (i.e. we use my fork of it) ?

That could be an improvement. Although even then I'm not sure an action should have `contents: write` permissions.

Pierre-Sassoulas:
- When the cherry-pick fail it create a comment on the original PR with some (very bad and convoluted imo) instructions on how to cherry-pick:  https://github.com/PyCQA/pylint/pull/7800#issuecomment-1320979317. Then we need to cherry-pick manually.
- The backporting PR is opened all the time otherwise (https://github.com/PyCQA/pylint/actions/workflows/backport.yml)

> That could be an improvement. Although even then I'm not sure an action should have contents: write permissions.

I think it needs those write rights to be able to open a PR with the cherry-picked commit.

cdce8p:
> * When the cherry-pick fail it create a comment on the original PR with some (very bad and convoluted imo) instructions on how to cherry-pick:  [[github actions] Fix the permissions in backporting job #7800 (comment)](https://github.com/PyCQA/pylint/pull/7800#issuecomment-1320979317). Then we need to cherry-pick manually.
> * The backporting PR is opened all the time otherwise (https://github.com/PyCQA/pylint/actions/workflows/backport.yml)

So it will always try to open a PR? In that case I misunderstood it.

> > That could be an improvement. Although even then I'm not sure an action should have `contents: write` permissions.
> 
> I think it needs those write rights to be able to open a PR with the cherry-picked commit.

🤔 Just opening a PR should only require `pull-requests: write`. Maybe it's pushing the new branch to Github which could require `contents: write`? https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#git-1

If that's the case, does it make sense to fork the action?

Pierre-Sassoulas:
> Just opening a PR should only require pull-requests: write. Maybe it's pushing the new branch to Github which could require contents: write?

I added the rights myself thinking that in order to create the cherry-picked commit we need content rights. We could try that with only ``pull-requests: write`` if I misunderstood something about the required github actions permissions. 

I would prefer to use the action directly if possible so as not to have to maintain it myself :)

cdce8p:
> > Just opening a PR should only require pull-requests: write. Maybe it's pushing the new branch to Github which could require contents: write?
> 
> I added the rights myself thinking that in order to create the cherry-picked commit we need content rights. We could try that with only `pull-requests: write` if I misunderstood something about the required github actions permissions.

Worth a shot. Would be great if it works without it.

> I would prefer to use the action directly if possible so as not to have to maintain it myself :)

Let's at least pin to full tag then and not just `v2`.
https://github.com/PyCQA/pylint/blob/73dfd8218eef72b3c37754df6c55d520a6a226de/.github/workflows/backport.yml#L29

Pierre-Sassoulas:
I made the change, let me know if it work for you @cdce8p / @DanielNoord . The cherry-pick on maintenance should fail when we merge because backport.yml does not exists on the maintenance branch yet. So i could do it manually, then we can fix any right issues by testing at the same time that the backporting work without ``content: write``.

.github/workflows/changelog.yml
```diff
@@ -1,9 +1,15 @@
 name: changelog
 
 on:
+  push:
```
DanielNoord, line 4: Does the push even still have the label data? That seems weird, just double checking!

.github/workflows/primer_run_main.yaml
```diff
@@ -9,7 +9,7 @@ on:
   push:
     branches:
       - main
-
+      - "maintenance/**"
```
DanielNoord, line 12: The primer will probably still get the data from `main`. I don't think this is necessary.

Pierre-Sassoulas:
```suggestion
```

.github/workflows/changelog.yml
Pierre-Sassoulas: Hmm I did not understand that ? Do you mean we should put this under "pull_request" ? Not an expert in github actions :)

.github/workflows/primer_run_main.yaml
```diff
@@ -9,7 +9,6 @@ on:
   push:
     branches:
       - main
```
Pierre-Sassoulas, line 11:
```suggestion
      - main

```

.github/workflows/changelog.yml
DanielNoord: I think so? (It's already there).

It feels weird to check the labels of a `push` event, I don't know if that even works..

Pierre-Sassoulas: Make sense, thank you !

.github/workflows/changelog.yml
```diff
@@ -3,7 +3,10 @@ name: changelog
 on:
   pull_request:
     types: [opened, synchronize, labeled, unlabeled, reopened]
-
+    branches:
+      - main
```
Pierre-Sassoulas, line 7:
```suggestion
```

.github/workflows/backport.yml
```diff
@@ -1,15 +1,14 @@
 name: Backport
 on:
+  branches:
+    - main
```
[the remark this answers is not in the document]
DanielNoord: I don't know how this action works..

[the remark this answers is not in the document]
Pierre-Sassoulas: When opening a PR I think github action consider that we're not on "main" but on the feature branch. We want to execute the backport when something is merged from a feature branch to main. (The changelog was'nt working with this branch option I had to do [696fbbe](https://github.com/PyCQA/pylint/pull/7815/commits/696fbbeb7f364778b3feea84bb8a24978edb5226)). Anyway, this backport is canceled as soon as backporting is not present so it's not going to be a big deal,

PR #7810, opened 2022-11-20
Fix `used-before-assignment` for variable annotations guarded by `TYPE_CHECKING`

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description
Closes #7609

This part of the (terribly long) method is starting to get ugly, so if we merge this first and backport it, I could use #7806 to factor out a staticmethod.

PR #7826, opened 2022-11-23
Add ``content: write`` rights for backporting job

Pierre-Sassoulas:
## Type of Changes

## Description

We need it to push to the backporting branch after cherry-picking. See https://github.com/PyCQA/pylint/pull/7810\#issuecomment-1324742960

Also include change we did on the maintenance branch to not run primers.

Refs #7804, #7815 #7810

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -7,6 +7,7 @@ on:
 
 permissions:
   pull-requests: write
+  content: write
 
 jobs:
   backport:
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -24,6 +24,6 @@ jobs:
         )
       )
     steps:
-      - uses: tibdex/backport@2e217641d82d02ba0603f46b1aeedefb258890ac
+      - uses: tibdex/backport@2e217641d82d02ba0603f46b1aeedefb258890ac # v2.0.3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/primer-test.yaml
+++ b/.github/workflows/primer-test.yaml
@@ -4,14 +4,14 @@ on:
   push:
     branches:
       - main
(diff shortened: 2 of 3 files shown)


DanielNoord (dismissed):
Primer changes look good!

clavedeluna (approved):

.github/workflows/backport.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 permissions:
   pull-requests: write
+  content: write
```
cdce8p, line 10: It was worth a shot at least.

cdce8p (commented):
Could you add a version comment here?
https://github.com/PyCQA/pylint/blob/f7d681b5a79e5781ab8072fe64459b199955a1f6/.github/workflows/backport.yml#L26

Similar to the one here. Dependabot will auto-update that too.
https://github.com/PyCQA/pylint/blob/f7d681b5a79e5781ab8072fe64459b199955a1f6/.github/workflows/primer_comment.yaml#L134

.github/workflows/backport.yml
```diff
@@ -23,6 +24,6 @@ jobs:
         )
       )
     steps:
-      - uses: tibdex/backport@2e217641d82d02ba0603f46b1aeedefb258890ac
+      - uses: tibdex/backport@2e217641d82d02ba0603f46b1aeedefb258890ac # v0.4.0
```
cdce8p, line 27:
```suggestion
      - uses: tibdex/backport@2e217641d82d02ba0603f46b1aeedefb258890ac # v2.0.3
```

Pierre-Sassoulas: Duh, I trust you too much 😄

cdce8p (approved):
Permissions look fine. Haven't followed the discussion around primer runs though.
If it should only run on `main`, I guess it's also good.

PR #7829, opened 2022-11-23
Drop dependency on py module

frenzymadness:
pytest 7.2 no longer depends on py and py is kinda deprecated so it's better to drop it. `tmp_path` fixture is newer and uses `pathlib.Path` from stdlib instead of `LocalPath` from `py._path`.

An alternative for this is to add `py` into test dependencies.

## Type of Changes

### DIFF
--- a/tests/lint/test_pylinter.py
+++ b/tests/lint/test_pylinter.py
@@ -2,13 +2,14 @@
 # For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
 # Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
 
+import os
+from pathlib import Path
 from typing import Any, NoReturn
 from unittest import mock
 from unittest.mock import patch
 
 import pytest
 from _pytest.recwarn import WarningsRecorder
-from py._path.local import LocalPath
 from pytest import CaptureFixture
 
 from pylint.lint.pylinter import PyLinter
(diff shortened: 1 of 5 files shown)


frenzymadness:
If there is no issue for that, is it possible to use PR number for the changelog entry?

DudeNr33:
> If there is no issue for that, is it possible to use PR number for the changelog entry?

Yes, you can do that!

tests/pyreverse/test_pyreverse_functional.py
```diff
@@ -23,17 +22,15 @@
...
-            args = ["-o", f"{output_format}", "-d", str(tmpdir)]
+            args = ["-o", f"{output_format}", "-d", str(tmp_path)]
             args.extend(testfile.options["command_line_args"])
             args += [str(input_file)]
             Run(args)
         assert sys_exit.value.code == 0
         assert testfile.source.with_suffix(f".{output_format}").read_text(
             encoding="utf8"
-        ) == Path(tmpdir / f"classes.{output_format}").read_text(encoding="utf8")
+        ) == Path(tmp_path / f"classes.{output_format}").read_text(encoding="utf8")
```
DudeNr33, line 36:
```suggestion
        ) == (tmp_path / f"classes.{output_format}").read_text(encoding="utf8")
```
As `tmp_path` is already a `Path` instance we can skip this here.

DudeNr33 (dismissed):
Thank you for this cleanup!
I only left a minor remark, otherwise LGTM!

Pierre-Sassoulas (approved):
Thank you for the cleanup !

Pierre-Sassoulas (dismissed):

frenzymadness:
I squashed the proposed fix and added a short whatsnew entry.

Pierre-Sassoulas (changes requested):
I'd rather not have this in the changelog, user don't care that much about our internal :)  Could you revert the latest changes please ?

frenzymadness:
> I'd rather not have this in the changelog, user don't care that much about our internal :) Could you revert the latest changes please ?

Sure thing, but, from my point of view, the changelog is also very interesting for maintainers who maintain the software in Linux distros, and for those, the info about dropped dependency is useful. But if you don't want it there, I can drop it.

Pierre-Sassoulas (approved):
Congratulation on becoming a pylint contributor too 😄

tests/lint/test_pylinter.py
```diff
@@ -21,16 +22,16 @@ def raise_exception(*args: Any, **kwargs: Any) -> NoReturn:
...
     with pytest.warns(DeprecationWarning):
         args = linter.load_command_line_configuration([__file__])
-    linter.crash_file_path = str(tmpdir / "pylint-crash-%Y")
+    linter.crash_file_path = str(tmp_path / "pylint-crash-%Y")
     linter.check(args)
     out, err = capsys.readouterr()
     assert not out
     assert not err
-    files = tmpdir.listdir()  # type: ignore[no-untyped-call]
+    files = os.listdir(tmp_path)
```
DanielNoord, line 34: For future reference, this could also be achieved with `tmp_path.iterdir`. Seems like something a linter could warn about :smile:

Pierre-Sassoulas: This can even be generalized : #7835 :) And in python 3.12 Path.walk() will also exists so there's going to be a lot of message to raise.

Issue #3899, opened 2020-10-13
Hangs forever after a parallel job gets killed


martinpitt:
This is similar to issue #1495 -- there is apparently some giant memory leak in pylint which makes it eat ginormous amounts of RAM. This issue is about mishandling parallel workers with `-j` -- once one worker dies, pylint runs into a deadlock.

Steps to reproduce
1. `podman run -it --rm --memory=2G fedora:33`

This issues is not specific to podman -- you can use docker, or a VM, or anything where you can control the amount of RAM.

2. Inside the container, prepare:
```
dnf install -y python3-pip git
pip install pylint
git clone --depth=1 https://github.com/rhinstaller/anaconda
cd anaconda
```
3. Run pylint in parallel; my `nproc` is 4, thus `-j0` selects 4, but let's select `-j4` explicitly:
```
pylint -j4 pyanaconda/
```

### Current behavior

As per issue #1495, the 5 pylint processes pile up more and more RAM usage, until one gets killed because of OOM:

    Memory cgroup out of memory: Killed process 89155 (pylint) total-vm:573588kB, anon-rss:564804kB, file-rss:0kB, shmem-rss:8kB, UID:1000 pgtables:1164kB oom_score_adj:0

I still get some remaining output, then pylint hangs forever:
```
[...]
pyanaconda/payload/dnf/payload.py:41:0: C0411: standard import "from fnmatch import fnmatch" should be placed before "import dnf" (wrong-import-order)
pyanaconda/payload/dnf/payload.py:42:0: C0411: standard import "from glob import glob" should be placed before "import dnf" (wrong-import-order)
pyanaconda/payload/dnf/payload.py:46:0: C0411: third party import "from pykickstart.constants import GROUP_ALL, GROUP_DEFAULT, KS_MISSING_IGNORE, KS_BROKEN_IGNORE, GROUP_REQUIRED" should be placed before "import pyanaconda.localization" (wrong-import-order)
pyanaconda/payload/dnf/payload.py:48:0: C0411: third party import "from pykickstart.parser import Group" should be placed before "import pyanaconda.localization" (wrong-import-order)
pyanaconda/payload/dnf/payload.py:50:0: C0412: Imports from package pyanaconda are not grouped (ungrouped-imports)
```

The other processes are still around:
```
martin     89152  5.1  0.2 266584 40680 pts/0    Sl+  17:23   0:24 /usr/bin/python3 /usr/local/bin/pylint -j0 pyanaconda/
[...]

PR #7834, opened 2022-11-23
Avoid hanging forever after a parallel job was killed

daniel-wer:
## Type of Changes

## Description

Replace `multiprocessing.pool` with [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) to avoid deadlocks.

In a `multiprocessing.pool`, if a process terminates in a non-clean fashion (for example, due to OOM or a segmentation fault), the pool will silently replace said process, but the work that the process was supposed to do will never be done, causing pylint to hang indefinitely. The `concurrent.futures.ProcessPoolExecutor` will raise a [BrokenProcessPool exception](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.process.BrokenProcessPool) in that case, avoiding the hang.

See https://github.com/PyCQA/pylint/issues/3899#issuecomment-732712793 for reproduction instructions.

The hangs caused by this issue ate up quite a few CI minutes for us over time. Although this doesn't fix the underlying sporadic issue (which is likely due to OOM) at least pylint will fail early instead of hanging forever.

A followup could be to recover from this type of error by catching the `BrokenProcessPool` exception and then instantiating a new `ProcessPoolExecutor` to continue the linting.

Closes https://github.com/PyCQA/pylint/issues/3899

### DIFF
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -8,6 +8,7 @@ import functools
 import warnings
 from collections import defaultdict
 from collections.abc import Iterable, Sequence
+from concurrent.futures import ProcessPoolExecutor
 from typing import TYPE_CHECKING, Any
 
 import dill
@@ -137,9 +138,9 @@ def check_parallel(
     # is identical to the linter object here. This is required so that
     # a custom PyLinter object can be used.
     initializer = functools.partial(_worker_initialize, arguments=arguments)
-    with multiprocessing.Pool(
-        jobs, initializer=initializer, initargs=[dill.dumps(linter)]
-    ) as pool:
+    with ProcessPoolExecutor(
+        max_workers=jobs, initializer=initializer, initargs=(dill.dumps(linter))
+    ) as executor:
         linter.open()
         all_stats = []
         all_mapreduce_data: defaultdict[
@@ -158,7 +159,7 @@ def check_parallel(
             stats,
             msg_status,
             mapreduce_data,
-        ) in pool.imap_unordered(_worker_check_single_file, files):
+        ) in executor.map(_worker_check_single_file, files):
             linter.file_state.base_name = base_name
             linter.file_state._is_base_filestate = False
             linter.set_current_module(module, file_path)
@@ -167,9 +168,6 @@ def check_parallel(
             all_stats.append(stats)
             all_mapreduce_data[worker_idx].append(mapreduce_data)
             linter.msg_status |= msg_status
-
-        pool.close()
-        pool.join()
 
     _merge_mapreduce_data(linter, all_mapreduce_data)
     linter.stats = merge_stats([linter.stats] + all_stats)
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -31,7 +31,7 @@ except ImportError:
 if TYPE_CHECKING:
     from pylint.lint import PyLinter
 
-# PyLinter object used by worker processes when checking files using multiprocessing
+# PyLinter object used by worker processes when checking files using parallel mode
 # should only be used by the worker processes
 _worker_linter: PyLinter | None = None
 
@@ -39,8 +39,7 @@ _worker_linter: PyLinter | None = None
 def _worker_initialize(
     linter: bytes, arguments: None | str | Sequence[str] = None
 ) -> None:
-    """Function called to initialize a worker for a Process within a multiprocessing
-    Pool.
+    """Function called to initialize a worker for a Process within a concurrent Pool.
 
     :param linter: A linter-class (PyLinter) instance pickled with dill
     :param arguments: File or module name(s) to lint and to be added to sys.path
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -8,7 +8,6 @@ import functools
 import warnings
 from collections import defaultdict
 from collections.abc import Iterable, Sequence
-from concurrent.futures import ProcessPoolExecutor
 from typing import TYPE_CHECKING, Any
 
 import dill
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas:
This look promising, thank you for opening the PR !

daniel-wer:
Thank _you_ for your work maintaining this great tool :pray: The tests were failing due to a silly mistake, should be fixed.

Pierre-Sassoulas (changes requested):
Thank you for fixing the tests :) Would you mind adding a changelog's fragment please ? Also I don't know if it's feasible but could we imagine an automated test where we kill a job to test the original issue?

Pierre-Sassoulas:
> However, I'm not yet sure how to design the test where we kill a job to test the original issue. 

I don't know either. We're going to hit timing /concurrency issues and flakiness if we do it with an external command to kill a process launched inside our test. I never did that with pytest I was hoping a magical pytest fixture would exists to be honest 😄Maybe a manual test is enough ?

pylint/lint/parallel.py
```diff
@@ -168,8 +169,5 @@ def check_parallel(
             all_mapreduce_data[worker_idx].append(mapreduce_data)
             linter.msg_status |= msg_status
```
DanielNoord, line 171: Why did you remove the close and join here? These were added explicitly at some point, even though the context manager should handle these normally. Is there a specific reason for removing them? Doesn't `ProccessPoolExecutor` support them?

DanielNoord (changes requested):

pylint/lint/parallel.py
daniel-wer: > Is there a specific reason for removing them? Doesn't ProccessPoolExecutor support them?

Exactly, these are not supported by `ProccessPoolExecutor`. The context manager calls [shutdown](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown) upon exiting which is explicitly mentioned in the docs.

DanielNoord:
@daniel-wer Do you understand the failing test?

daniel-wer:
@DanielNoord I'm not entirely sure. My first suspicion was that it is timing-related. Since I configured the test to timeout after 1 second, it could happen that the time until the first pool process is initialized (and the `TypeError: 'int' object is not iterable` is thrown which leads to a `BrokenProcessPool` exception) is more than 1 second.

However, the captured stderr indicates that the `TypeError` was already raised. In all other configurations this led to the expected `BrokenProcessPool` exception. Might be that the timing was very unlucky in this case :thinking: 

We could try to configure a more generous timeout of 5 seconds since the timeout will only be triggered if the test fails and won't affect normal test run times. This way we could rule out timing-related issues. What do you think?

DanielNoord:
Let's try the higher timeout indeed! Sorry for the last response!

DanielNoord (commented):
I just did a check for references to `multiprocessing` and saw that we have a couple of checks that test whether it is available. Is this also a concern for `concurrency`?

Edit: It seems like it is, so we should probably change all those references and tests as well.

daniel-wer:
@DanielNoord Thank you, I was not aware of that. Fixed.

daniel-wer:
Could you give me a pointer to the relevant files? I had a look and from what I can see the `multiprocessing` module is used in some of the functional checker tests or to call `.current_process()` or `.cpu_count()`. I think neither of these usages can be replaced with the `concurrent` module (which doesn't have the `.current_process()` or `.cpu_count()` functionality).

The only thing I can see is that in `pylint/lint/run.py` in addition to `multiprocessing.cpu_count()` the existence of the `concurrent` module should be checked to fallback to a single process if it is not available, but I don't see where `multiprocessing` imports could be removed.

daniel-wer:
@DanielNoord 

> The last thing was indeed the only thing I thinking of.

Ok great, I adapted the check.

> Just to double check: does the determining of cpus with multiprocessing also work for the amount of workers that concurrent can use?

The [docs](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) state that the "ProcessPoolExecutor uses the multiprocessing module", so yes that should be the case. Also, this kind of functionality was not duplicated for the `concurrent.futures` module which is rather an alternative interface for interacting with a process pool but not changing the underlying process fundamentals like `current_process()` or `cpu_count()`.

@Pierre-Sassoulas

> Maybe a manual test is enough ?

Using the latest version of this PR, I again confirmed that killing a random pylint worker process during a parallel linting no longer leads to a deadlock, but immediately exits with a `BrokenProcessPool` exception, using the reproduction instructions from https://github.com/PyCQA/pylint/issues/3899#issuecomment-732712793.

DanielNoord:
I hope to find time to review this somewhere this weekend!

tests/test_check_parallel.py
```diff
@@ -28,6 +27,13 @@
 from pylint.typing import FileItem
 from pylint.utils import LinterStats, ModuleStats
 
+try:
+    from concurrent.futures import ProcessPoolExecutor
+    from concurrent.futures.process import BrokenProcessPool
+except ImportError:
+    ProcessPoolExecutor = None  # type: ignore[assignment,misc]
+    BrokenProcessPool = None  # type: ignore[assignment,misc]
```
DanielNoord, line 35:
```suggestion
from concurrent.futures import ProcessPoolExecutor
from concurrent.futures.process import BrokenProcessPool
```

Let's make this necessary for tests still.

pylint/lint/parallel.py
```diff
@@ -23,19 +23,23 @@
 except ImportError:
     multiprocessing = None  # type: ignore[assignment]
```
DanielNoord, line 24: This is one use of this still, it is used to get the id of the process. Can we do that with `concurrent`?

DanielNoord (changes requested):
I pushed two changes to docstrings.

Two last (for real this time) questions.

Sorry for taking so long. Parallelism is hard and I don't want to leave any stuff from the old implementation behind after merging this.

Pierre-Sassoulas (dismissed):
LGTM. (I was worried that the backport would conflict with #7814 but we did not merge it)

Pierre-Sassoulas (approved):

DanielNoord (approved):

DanielNoord:
Thanks for all the work here @daniel-wer

PR #7836, opened 2022-11-23
Feature: distinct Composition and Aggregation arrows

qequ:
## Type of Changes

## Description
Update pyreverse to differentiate between compositions and aggregations in classes. Only covers the basic cases of composition and aggregation(@DudeNr33 ).

e.g.  having a code like this
```python

""" docstring for file clientmodule.py """
from data.suppliermodule_test import Interface, DoNothing, DoNothing2

class Ancestor:
    """ Ancestor method """
    __implements__ = (Interface,)
    cls_member = DoNothing()

    def __init__(self, value):
        local_variable = 0
        self.attr = 'this method shouldn\'t have a docstring'
        self.__value = value

    def get_value(self):
        """ nice docstring ;-) """
        return self.__value

    def set_value(self, value):
        self.__value = value
        return 'this method shouldn\'t have a docstring'

class Specialization(Ancestor):
    TYPE = 'final class'
    top = 'class'

    def __init__(self, value, _id, relation2: DoNothing2):
        Ancestor.__init__(self, value)
        self._id = _id
        self.relation = DoNothing()
        self.relation2 = relation2

    @classmethod
    def from_value(cls, value: int):
        return cls(value, 0, DoNothing2())

    @staticmethod
    def transform_value(value: int) -> int:
        return value * 2

    def increment_value(self) -> None:
        self.set_value(self.get_value() + 1)


```

without the feature
classes_test

with the feature
classes_test

Refs #6543

CC: @dgutson

### DIFF
--- a/doc/whatsnew/fragments/6543.feature
+++ b/doc/whatsnew/fragments/6543.feature
@@ -1,3 +1,3 @@
-Update pyreverse to differentiate between aggregations and compositions.
+Update ``pyreverse`` to differentiate between aggregations and compositions.
 
 Refs #6543
--- a/doc/whatsnew/fragments/6543.feature
+++ b/doc/whatsnew/fragments/6543.feature
@@ -1,0 +1,1 @@
+Update pyreverse to differentiate between aggregations and compositions
--- a/doc/whatsnew/fragments/6543.feature
+++ b/doc/whatsnew/fragments/6543.feature
@@ -1,3 +1,4 @@
 Update ``pyreverse`` to differentiate between aggregations and compositions.
-
+``pyreverse`` checks if it's an Instance or a Call of an object via method parameters (via type hints)
+to decide if it's a composition or an aggregation.
 Refs #6543
--- a/doc/whatsnew/fragments/6543.feature
+++ b/doc/whatsnew/fragments/6543.feature
@@ -1,1 +1,3 @@
 Update pyreverse to differentiate between aggregations and compositions
+
+Refs #6543
--- a/doc/whatsnew/fragments/6543.feature
+++ b/doc/whatsnew/fragments/6543.feature
@@ -1,4 +1,5 @@
 Update ``pyreverse`` to differentiate between aggregations and compositions.
 ``pyreverse`` checks if it's an Instance or a Call of an object via method parameters (via type hints)
 to decide if it's a composition or an aggregation.
+
 Refs #6543
--- a/doc/whatsnew/fragments/6543.feature
+++ b/doc/whatsnew/fragments/6543.feature
@@ -1,3 +1,3 @@
-Update pyreverse to differentiate between aggregations and compositions
+Update pyreverse to differentiate between aggregations and compositions.
 
 Refs #6543
--- a/pylint/pyreverse/diagrams.py
+++ b/pylint/pyreverse/diagrams.py
@@ -219,7 +219,8 @@ class ClassDiagram(Figure, FilterMixIn):
             for name, values in list(node.aggregations_type.items()):
                 for value in values:
                     self.assign_association_relationship(
-                        value, obj, name, "aggregation")
+                        value, obj, name, "aggregation"
+                    )
 
             for name, values in list(node.associations_type.items()) + list(
                 node.locals_type.items()
@@ -227,9 +228,12 @@ class ClassDiagram(Figure, FilterMixIn):
 
                 for value in values:
                     self.assign_association_relationship(
-                        value, obj, name, "association")
+                        value, obj, name, "association"
+                    )
 
-    def assign_association_relationship(self, value: Any, obj: ClassEntity, name: str, type_relationship: str) -> None:
+    def assign_association_relationship(
+        self, value: Any, obj: ClassEntity, name: str, type_relationship: str
+    ) -> None:
         if value is astroid.Uninferable:
             return
         if isinstance(value, astroid.Instance):
--- a/pylint/pyreverse/diagrams.py
+++ b/pylint/pyreverse/diagrams.py
@@ -232,7 +232,7 @@ class ClassDiagram(Figure, FilterMixIn):
                     )
 
     def assign_association_relationship(
-        self, value: Any, obj: ClassEntity, name: str, type_relationship: str
+        self, value: astroid.NodeNG, obj: ClassEntity, name: str, type_relationship: str
     ) -> None:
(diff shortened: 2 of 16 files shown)


doc/whatsnew/fragments/6543.feature
```diff
@@ -0,0 +1,3 @@
+Update pyreverse to differentiate between aggregations and compositions.
```
DudeNr33, line 1:
```suggestion
Update ``pyreverse`` to differentiate between aggregations and compositions.
```

pylint/pyreverse/inspector.py
```diff
@@ -313,6 +325,55 @@ def _imported_module(
...
+
+    @abstractmethod
+    def handle(self, node: nodes.AssignAttr, parent: nodes.ClassDef) -> None:
+        pass
+
+
+class AbstractAssociationHandler(AssociationHandlerInterface):
+    """
+    The default chaining behavior can be implemented inside a base handler
+    class.
```
DudeNr33, line 343: I think it would be great to add some more explanation here why we need this chaining behaviour. It took me a while to see why you implemented it like this and don't just loop over all possible handlers.
But I think it is a good solution for the problem and allows for extension later on if we want to handle other association types better as well.

DudeNr33 (dismissed):
Thank you very much for this contribution. This was not a trivial task, and I like the solution as it can be extended later on. 👍 

@Pierre-Sassoulas Milestone 2.16.0?

pylint/pyreverse/inspector.py
qequ: Updated docs :+1:

pylint/pyreverse/diagrams.py
```diff
@@ -214,20 +214,35 @@ def extract_relationships(self) -> None:
                     self.add_relationship(obj, impl_obj, "implements")
                 except KeyError:
                     continue
-            # associations link
-            for name, values in list(node.instance_attrs_type.items()) + list(
+            # associations & aggregations links
+
```
DanielNoord, line 218:
```suggestion

            # associations & aggregations links
```

pylint/pyreverse/diagrams.py
```diff
@@ -214,20 +214,35 @@ def extract_relationships(self) -> None:
...
-                        associated_obj = self.object_from_node(value)
-                        self.add_relationship(associated_obj, obj, "association", name)
-                    except KeyError:
-                        continue
+                    self.assign_association_relationship(
+                        value, obj, name, "association"
+                    )
+
+    def assign_association_relationship(
+        self, value: Any, obj: ClassEntity, name: str, type_relationship: str
```
DanielNoord, line 235: Can we specify this? `object_from_node` only accepts `NodeNG`. Since `astroid` doesn't have a `py.typed` `mypy` doesn't understand this, but it would cause problems when it (finally) does.

pylint/pyreverse/inspector.py
```diff
@@ -178,9 +187,12 @@ def visit_classdef(self, node: nodes.ClassDef) -> None:
             baseobj.specializations = specializations
         # resolve instance attributes
         node.instance_attrs_type = collections.defaultdict(list)
+        node.aggregations_type = collections.defaultdict(list)
```
DanielNoord, line 190: Could we add typing here? Same goes for the line below?
Something along the lines of `defaultdict[str, list[...]]`

DanielNoord (changes requested):
Just reviewed some of the typing. Thanks for your contribution!

pylint/pyreverse/diagrams.py
qequ: Added type hint for obj :+1:

pylint/pyreverse/inspector.py
qequ: I'm having a mypy issue with this; it's complaining that I can't set type for non-self atributes ( `Type cannot be declared in assignment to non-self attribute  [misc]`).  The nodes come from astroid but the dicts are set by the Linker. I don't know if there is a workaround for this (@DudeNr33 ?)

DudeNr33: If `mypy` complains here I fear that it won't be easy to solve this. I also dislike that some random attributes are added to the `Node` instances here at runtime. The cleanest solution would probably be to create a new (data)class that wraps `nodes.ClassDef` and declares the additional attributes needed by `pyreverse` inside its `__init__`, but that's out of scope in this PR.

qequ: yes, I followed the functionality of the Linker to set attributes to nodes with these new attributes so I think it's a Linker problem. I'm gonna mark this as resolved and open a separate issue :+1:

DudeNr33 (approved):

pylint/pyreverse/inspector.py
qequ: issue; https://github.com/PyCQA/pylint/issues/7850

doc/whatsnew/fragments/6543.feature
```diff
@@ -0,0 +1,3 @@
+Update ``pyreverse`` to differentiate between aggregations and compositions.
```
Pierre-Sassoulas, line 1: I think it would be nice to add a short summary of how we distinguish between the two here.

Pierre-Sassoulas (approved):
Thank you for this new feature ! I think it can go in 2.16.0. I have a small suggestion, feel free to disregard.

doc/whatsnew/fragments/6543.feature
qequ: Updated fragment :+1:

Pierre-Sassoulas (approved):

DudeNr33 (approved):

DudeNr33:
@DanielNoord anything left to change from your side?

DanielNoord (approved):

DudeNr33:
Thank you @qequ for this enhancement!

dgutson:
Go @qequ go!

khashashin:
Thank you guys, great job 🎉

jacobtylerwalls:
@DudeNr33 Gentle reminder to wipe/reduce GitHub's suggested commit message when squashing. Thanks!

DudeNr33:
> @DudeNr33 Gentle reminder to wipe/reduce GitHub's suggested commit message when squashing. Thanks!

Ah, I did not know that this is the preferred way. I'll try to remember.

PR #7838, opened 2022-11-23
Upgrade the versions of astroid and dill

Pierre-Sassoulas:
## Type of Changes

## Description

Make sure that we use the real version of dill on all interpreter now that dill released version 0.3.6. also explicitly use the latest astroid.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,9 +33,7 @@ classifiers = [
 ]
 requires-python = ">=3.7.2"
 dependencies    = [
-    # TODO: Remove dill-pylint once dill 0.3.6 is released
-    "dill>=0.2;python_version<'3.11'",
-    "dill-pylint>=0.3.6.dev0;python_version>='3.11'",
+    "dill>=0.2",
     "platformdirs>=2.2.0",
     # Also upgrade requirements_test_min.txt.
     # Pinned to dev of second minor update to allow editable installs and fix primer issues,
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas:
Wow, I'm really sorry about the ping of every code owner, isort and pre-commit suddenly started to modify all the files for no apparent reason.

DanielNoord (dismissed):

pyproject.toml
```diff
@@ -33,14 +33,12 @@ classifiers = [
 ]
 requires-python = ">=3.7.2"
 dependencies    = [
-    # TODO: Remove dill-pylint once dill 0.3.6 is released
-    "dill>=0.2;python_version<'3.11'",
-    "dill-pylint>=0.3.6.dev0;python_version>='3.11'",
+    "dill>=0.2",
```
DudeNr33, line 36: Shouldn't we keep the `>=0.3.6` constraint at least for Python >= 3.11 if lower versions cause problems?

Pierre-Sassoulas: This should be enforced by pip and dill's metadata automatically.

DanielNoord: I was thinking the same but Pierre is right, that shouldn't be our responsibility.

Pierre-Sassoulas:
```suggestion
    "dill>=0.2,<1",
```
Following [pyflake's 3.0 release all old releases of autoflake now have broken 2.X typing support](https://github.com/PyCQA/autoflake/pull/186). Let's do better for pylint :)

PR #7841, opened 2022-11-24
Add an upper limit to all runtime dependencie requirements

Pierre-Sassoulas:
## Type of Changes

## Description

In order to avoid old releases from becoming unusable if one of our dependencie release a major with breaking changes. Inspired by the release of pyflakes 3 that broke all version of autoflake at once: https://github.com/PyCQA/autoflake/pull/186


pyproject.toml
```diff
@@ -33,18 +33,18 @@ classifiers = [
...
     # Pinned to dev of second minor update to allow editable installs and fix primer issues,
     # see https://github.com/PyCQA/astroid/issues/1341
     "astroid>=2.12.13,<=2.14.0-dev0",
     "isort>=4.2.5,<6",
     "mccabe>=0.6,<0.8",
-    "tomli>=1.1.0;python_version<'3.11'",
-    "tomlkit>=0.10.1",
-    "colorama>=0.4.5;sys_platform=='win32'",
-    "typing-extensions>=3.10.0;python_version<'3.10'",
+    "tomli>=1.1.0,<2;python_version<'3.11'",
```
Pierre-Sassoulas, line 44:
```suggestion
    "tomli>=1.1.0,<3;python_version<'3.11'",
```

pyproject.toml
```diff
@@ -33,18 +33,18 @@ classifiers = [
...
+    "tomlkit>=0.10.1,<1",
+    "colorama>=0.4.5,<1;sys_platform=='win32'",
+    "typing-extensions>=3.10.0,<4;python_version<'3.10'",
```
Pierre-Sassoulas, line 47:
```suggestion
    "typing-extensions>=3.10.0,<5;python_version<'3.10'",
```

pyproject.toml
```diff
@@ -33,18 +33,18 @@ classifiers = [
...
-    "dill>=0.2",
-    "platformdirs>=2.2.0",
+    "dill>=0.2,<1",
+    "platformdirs>=2.2.0,<3",
     # Also upgrade requirements_test_min.txt.
     # Pinned to dev of second minor update to allow editable installs and fix primer issues,
     # see https://github.com/PyCQA/astroid/issues/1341
     "astroid>=2.12.13,<=2.14.0-dev0",
     "isort>=4.2.5,<6",
     "mccabe>=0.6,<0.8",
```
Pierre-Sassoulas, line 43:
```suggestion
    "mccabe>=0.6,<1",
```

DanielNoord (commented):
I really don't like this but it is a bit of a chicken and egg. We don't know that we don't support this versions so we are now unnecessarily blocking people from upgrading their dependencies if they are on an older version of `pylint`. I really don't like that.

I prefer fixing dependencies issues as they arise instead of prematurely blocking everything, potentially unnecessarily. But that might be my own opinion...

Pierre-Sassoulas:
There's two side of the argument for sure. I think the downside of each needs to be examined in details here.

-  No upper limit downside:
    - A release of a new dependency can break all available versions of pylint
    - If you're using an unmaintained pylint you have to fix the issue yourself because a maintainer fixing all old versions is not realistic
    - Or you if you're using the latest pylint, wait for a maintainer to release a new version and remove pylint from your pipeline in the meantime
    - The only way to fix it yourself otherwise is to understand the problem and fix the upper limit yourself (you have to debug for a long time to understand that a new dependency came out.  It's hard to reproduce because you have old installation of pylint that work, and new installation of pylint that break)

- Upper limit downside:
    - You can't upgrade your dependencies instantly /silently without pylint's maintainers intervention
    - If you really need the new version of the dependency because you have another dependency that is incompatible you have to **force** pip to install it. (Is it even possible ?)
  
I think not being able to upgrade immediately is a small price to pay for the stability this brings. In one case you do not have the latest feature of a dependency you might not be even using directly. For the other case everything break in strange way and a build that was working one day do not work the next day. It feel like pylint is not reliable. In all case you will have to wait for maintainers intervention. But would you rather have to wait to upgrade or to wait for your pipeline to be green again ?

cdce8p:
I don't like this either.

Pylint, and other linting tools for that matter, are in a difficult spot. Usually the recommendation is for applications to pin their dependencies to one specific release. However, that only works if they are installed inside a clean environment. Thus this doesn't work for linting tools which often require to be installed inside the working venv (especially pylint and mypy).

> A release of a new dependency can break all available versions of pylint

Technically this isn't correct. Only `new installs` will be broken. If you upgrade pylint alone, pip won't update the dependencies. (Might be different for poetry though.)

> If you're using an unmaintained pylint you have to fix the issue yourself because a maintainer fixing all old versions is not realistic

I don't see it that way. In case we don't pin an upper bound the issue will likely also be present with the current pylint version. What's likely to happen is that someone will copy the error and create a new issue. We'll investigate it and discover the dependency issue. The fix then will just be to pin the dependency which caused the issue in the first place. Something you can do / try even for old pylint versions.

_Btw. this is something we have to do quite regularly for Home Assistant since we only pin primary dependencies and build new venvs from scratch each time a dependencies changes. Thus also fetching new secondary deps. We even have a dedicated constraints file just for that._

> You can't upgrade your dependencies instantly /silently without pylint's maintainers intervention

The reverse is also true however and I believe this is the most important point. If you're using an old pylint version, at some point the upper limits might be outdated. Even if a new major version would just work fine, you can't install it because we have pinned a upper bound (unnecessarily).

You can fix a missing upper limit, but not an unnecessarily low one. (At least one easily.)

PR #7846, opened 2022-11-25
Fix ``logging-fstring-interpolation`` FP

### LINKED ISSUE #4984: False positive for `logging-fstring-interpolation`
### Bug description

When using f-string without arguments in a logging message, Pylint would warn me about not using lazy % formatting, even though I do use it.

a.py:
```python
import logging
logging.error(f'Hello %s', 'World!')
```

output:
`ERROR:root:Hello World!`

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:2:0: W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)
a.py:2:14: W1309: Using an f-string that does not have any interpolated variables (f-string-without-interpolation)

```

### Expected behavior

Report only `f-string-without-interpolation`

### Pylint version

```shell
pylint 2.10.2
astroid 2.7.3
Python 3.9.4 (default, Apr 30 2021, 18:51:54) 
[GCC 10.2.0]
```

### OS / Environment

Manjaro Linux, Kernel 5.10

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description

``logging-fstring-interpolation`` should not be raised if logging with an f-str that also has %s formatting.

Closes #4984

### DIFF
--- a/pylint/checkers/logging.py
+++ b/pylint/checkers/logging.py
@@ -105,12 +105,7 @@ CHECKED_CONVENIENCE_FUNCTIONS = {
     "warning",
 }
 
-MOST_COMMON_FORMATTING = {
-    "%s",
-    "%d",
-    "%f",
-    "%r",
-}
+MOST_COMMON_FORMATTING = frozenset("%s", "%d", "%f", "%r")
 
 
 def is_method_call(
--- a/pylint/checkers/logging.py
+++ b/pylint/checkers/logging.py
@@ -403,7 +403,7 @@ def _count_supplied_tokens(args: list[nodes.NodeNG]) -> int:
     return sum(1 for arg in args if not isinstance(arg, nodes.Keyword))
 
 
-def str_formatting_in_f_string(node: nodes.NodeNG) -> bool:
+def str_formatting_in_f_string(node: nodes.JoinedStr) -> bool:
     """Determine whether the node represents an f-string with string formatting.
 
     For example: `f'Hello %s'`
@@ -416,6 +416,5 @@ def str_formatting_in_f_string(node: nodes.NodeNG) -> bool:
     )
 
 
-# Detect % formatting inside an f-string, such as `f'Hello %s'`
 def register(linter: PyLinter) -> None:
     linter.register_checker(LoggingChecker(linter))
--- a/pylint/checkers/logging.py
+++ b/pylint/checkers/logging.py
@@ -105,6 +105,13 @@ CHECKED_CONVENIENCE_FUNCTIONS = {
     "warning",
 }
 
+MOST_COMMON_FORMATTING = {
+    "%s",
+    "%d",
+    "%f",
+    "%r",
+}
+
 
 def is_method_call(
     func: bases.BoundMethod, types: tuple[str, ...] = (), methods: tuple[str, ...] = ()
@@ -262,12 +269,7 @@ class LoggingChecker(checkers.BaseChecker):
         elif isinstance(format_arg, nodes.Const):
             self._check_format_string(node, format_pos)
         elif isinstance(format_arg, nodes.JoinedStr):
-            if any(
-                "%s" in val.value
-                for val in format_arg.values
-                if isinstance(val, nodes.Const)
-            ):
-                # Detect % formatting inside an f-string, such as `f'Hello %s'`
+            if str_formatting_in_f_string(format_arg):
                 return
             self.add_message(
                 "logging-fstring-interpolation",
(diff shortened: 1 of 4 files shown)


pylint/checkers/logging.py
```diff
@@ -256,11 +257,18 @@ def _check_log_method(self, node: nodes.Call, name: str) -> None:
...
-            self._check_call_func(node.args[format_pos])
-        elif isinstance(node.args[format_pos], nodes.Const):
+        elif isinstance(format_arg, nodes.Call):
+            self._check_call_func(format_arg)
+        elif isinstance(format_arg, nodes.Const):
             self._check_format_string(node, format_pos)
-        elif isinstance(node.args[format_pos], nodes.JoinedStr):
+        elif isinstance(format_arg, nodes.JoinedStr):
+            if any(
+                "%s" in val.value
```
Pierre-Sassoulas, line 266: There's a lot of potential other formatting here %r %d, %0.3f, %10s... Maybe this isn't the right approach.

clavedeluna: Yeah that's what I assumed I'd hear back. I figured '%s' is the most common, maybe d, r, and we could at least get rid of that FP. Any better way to detect %x formatting?

Pierre-Sassoulas: I'm not sure but I think it's not easy to solve. @DanielNoord you worked a lot on f-string could you confirm ?

DanielNoord: Yeah this seems to open a can of worms, although the same could be said for the original issue.

I don't know, we might just try this and see if other requests pop up.

clavedeluna: Unless we think the perf hit is egregious (which I don't since this is in a very specific code section), having this code or maybe I could refactor to a well named method `_check_str_formatting` or something, and can add a few more formatting types like %d or %r which should cover probably 95% of what we'd ever see in this pattern.

Pierre-Sassoulas: +1 for checking only the most common one. The primer already show that this is useful. Let's do %s, %d, %f and %r. Probably something like is ``'%' in val.value and any(x in val.value for x in [%s, %d, %f, %r])``. Not sure if it's the best performance wise but probably better than ``any(x in val.value for x in [%s, %d, %f, %r])``

pylint/checkers/logging.py
```diff
@@ -393,5 +403,19 @@ def _count_supplied_tokens(args: list[nodes.NodeNG]) -> int:
...
+    """
+    # Check "%" presence first for performance.
+    return any(
+        "%" in val.value and any(x in val.value for x in MOST_COMMON_FORMATTING)
+        for val in node.values
+        if isinstance(val, nodes.Const)
+    )
+
+
+# Detect % formatting inside an f-string, such as `f'Hello %s'`
```
mbyrnepr2, line 419:
```suggestion
```

pylint/checkers/logging.py
```diff
@@ -393,5 +403,19 @@ def _count_supplied_tokens(args: list[nodes.NodeNG]) -> int:
     return sum(1 for arg in args if not isinstance(arg, nodes.Keyword))
 
 
+def str_formatting_in_f_string(node: nodes.NodeNG) -> bool:
```
mbyrnepr2, line 406:
```suggestion
def str_formatting_in_f_string(node: nodes.JoinedStr) -> bool:
```

Basing this suggestion on these lines:
```python
        elif isinstance(format_arg, nodes.JoinedStr):
            if str_formatting_in_f_string(format_arg):
```

tests/functional/l/logging/logging_fstring_interpolation_py37.py
```diff
@@ -3,3 +3,9 @@
 
 VAR = "string"
 logging.error(f"{VAR}")  # [logging-fstring-interpolation]
+
+WORLD = "world"
+logging.error(f'Hello {WORLD}')  # [logging-fstring-interpolation]
```
mbyrnepr2, line 8: For consistency we could use the `"` instead of `'` like `VAR` above. Perhaps we should run black on the test files 🤔

Pierre-Sassoulas: pylint need to be able to handle simple quote as well as double quote so the tests needs to stay diverse. If everything is perfectly blacked then we don't know if we handle non blacked code correctly.

clavedeluna: yeah most of the functional test modules are not black-ened

mbyrnepr2: 👍

mbyrnepr2 (dismissed):

pylint/checkers/logging.py
```diff
@@ -105,6 +105,13 @@
     "warning",
 }
 
+MOST_COMMON_FORMATTING = {
+    "%s",
+    "%d",
+    "%f",
+    "%r",
+}
```
Pierre-Sassoulas, line 113:
```suggestion
MOST_COMMON_FORMATTING = frozenset("%s", "%d", "%f", "%r")
```

Pierre-Sassoulas (dismissed):

pylint/checkers/logging.py
```diff
@@ -105,6 +105,8 @@
     "warning",
 }
 
+MOST_COMMON_FORMATTING = frozenset("%s", "%d", "%f", "%r")
```
Pierre-Sassoulas, line 108:
```suggestion
MOST_COMMON_FORMATTING = frozenset(["%s", "%d", "%f", "%r"])
```

Pierre-Sassoulas (approved):

Issue #7779, opened 2022-11-16
False positive used-before-assignment with inline if


Zeckie:
### Bug description

```python
str(foo if (foo := "BAR") else "").lower()
```

Pylint gives error: `foobar.py:1:4: E0601: Using variable 'foo' before assignment (used-before-assignment)`

When run in python:
```python
>>> str(foo if (foo := "BAR") else "").lower()
'bar'
```

### Command used

```shell
pylint foobar.py
```

### Pylint output

```shell
************* Module foobar
foobar.py:1:4: E0601: Using variable 'foo' before assignment (used-before-assignment)
```

### Expected behavior

No errors

### Pylint version

```shell
pylint 2.15.5
astroid 2.12.12
Python 3.10.6 (tags/v3.10.6:9c7b4bd, Aug  1 2022, 21:53:49) [MSC v.1932 64 bit (AMD64)]
```

### OS / Environment

Windows

nickdrozd:
Variations that don't raise the error:

```python
# reverse if / else values
str("" if (foo := "BAR") else foo).lower()

# drop .lower()
str(foo if (foo := "BAR") else "")

# simple assignment
var = foo if (foo := "BAR") else ""
```

PR #7847, opened 2022-11-25
Prevent `used-before-assignment` for variables defined in assignment expressions

jacobtylerwalls:
## Type of Changes

## Description
Fix an unreleased false positive noticed by Marc in https://github.com/PyCQA/pylint/pull/6677#discussion_r1031664559.

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -738,13 +738,13 @@ scope_type : {self._atomic.scope_type}
         if any(node.nodes_of_class(nodes.Break)):
             return True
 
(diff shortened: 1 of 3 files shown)


clavedeluna:
I think this _may_ address https://github.com/PyCQA/pylint/issues/7779 . If so, could you add the test cases in that issue?

jacobtylerwalls:
Good question, but no, it doesn't address that issue. This change is inside the new static methods added in #6677.

clavedeluna (approved):

Pierre-Sassoulas (approved):
👌

PR #7852, opened 2022-11-25
New check `consider-using-keyword-argument` raised for positional litterals in function call

### LINKED ISSUE #385: Warning to suggest keyword argument when using non-meaningfull litterals as positional argument
Originally reported by: **Claudiu Popa (BitBucket: [PCManticore](http://bitbucket.org/PCManticore), GitHub: @PCManticore)**

---

I see the following idiom a lot in some codebases:

```python
s.splitlines(True)
shutil.rmtree("something", True)
my_particular_function(obj, False)
```

The following is clearer and improves the code intent:

```python
s.splitlines(keepends=True)
shutil.rmtreee(something, ignore_errors=True)
my_particular_function(obj, remove=False)
```

We should emit a warning when detecting constants in function calls. Probably True, False, None should be enough, in order to avoid too many false positives. Also, we should ignore builtins. filter(None, something) is okay.

---
- Bitbucket: https://bitbucket.org/logilab/pylint/issue/385

clavedeluna:
## Type of Changes

## Description

Refs #XXXX

Closes #385


DanielNoord (commented):
I haven't reviewed the code, but I don't really like the name of the message as to me it implies something is actually missing while it is just a style preference. What about `consider-using-keyword-argument`?

clavedeluna:
> I haven't reviewed the code, but I don't really like the name of the message as to me it implies something is actually missing while it is just a style preference. What about `consider-using-keyword-argument`?

Yeah seems reasonable. I did this quickly without worrying about that stuff so we could see primer results and decide if we wanted this impl. Once we decide I'll go back and change that stuff, thanks!

Pierre-Sassoulas:
My first thought, is that most function don't need keyword args, it's only function with a large number of  arguments that are problematic. So we should raise only if there are more than 1 positional arguments excluding default value, maybe 2 to 5 depending on how much message we would get. I.e. remembering the order of arguments of the function is only a problem for 2 arguments or more excluding the one that have default value.

clavedeluna:
> My first thought, is that most function don't need keyword args, it's only function with a large number of arguments that are problematic. So we should raise only if there are more than 1 positional arguments excluding default value, maybe 2 to 5 depending on how much message we would get. I.e. remembering the order of arguments of the function is only a problem for 2 arguments or more excluding the one that have default value.

Seems like a reasonable approach. Another one could be only raising on types such as bool, int, etc, like we mentioned initially. But I can go with your suggestion first and see where we get to.

clavedeluna:
> more than 1 positional arguments excluding default value,

Does pylint currently have a way to detect when a param has a default kwarg? Asking in case  there's code to reuse

Pierre-Sassoulas:
> Does pylint currently have a way to detect when a param has a default kwarg?

Sorry no idea. I suppose it's possible to retrieve the explicitly defined argument from the node but I'm not sure.

DanielNoord:
`astroid` has this information in the `Arguments` node, but its methods are not perfect. Should be able to retrieve it though.

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,66 @@
...
+        ),
+    }
+
+    def visit_call(self, node: nodes.Call) -> None:
+        """Check that called functions/methods are provided keyword arguments."""
+        called = safe_infer(node.func)
+
+        if isinstance(called, nodes.ClassDef):
+            needed_keywords = called.instance_attrs.keys()
+            default_kwargs = []  # todo
```
clavedeluna, line 37: @DanielNoord would you happen to recall how to get the default kwargs for methods like `__init__`? I got it working for normal functions and instance methods but couldn't find it for these. If not, I could make the checker not check stuff like this `car = Car("a car", "name", "title"`.

DanielNoord: Doesn't `ClassDef.args` have a method for that?`default_value()`.

clavedeluna:
```
(Pdb) called.args
*** AttributeError: 'ClassDef' object has no attribute 'args'
```

?

DanielNoord: ``node.locals["__init__"].args``. Sorry!

clavedeluna: Awesome, super helpful!

The issue I'm having now is that I don't have a way to connect `defaults` to their name.  For this example

```
def make_car(name, owner, title, color="red", year=2022):
    return "car"

second_car = make_car("a car", "name", "title") # [missing-keyword-arg, missing-keyword-arg, missing-keyword-arg]
```

We only want to raise the new message 3x, not 5x, because color and year are defaults and the user didn't provide them and that's ok. 

So I need to figure out a way to tell my new checker to remove these defaults, but I don't have their names... all I have is

```
(Pdb) called.args.defaults
[<Const.str l.82 at 0x11a4da4f0>, <Const.int l.82 at 0x11a4da610>]
```

```
(Pdb) called.args.args
[<AssignName.name l.82 at 0x11a4da400>, <AssignName.owner l.82 at 0x11a4da520>, <AssignName.title l.82 at 0x11a4da550>, <AssignName.color l.82 at 0x11a4da580>, <AssignName.year l.82 at 0x11a4da5b0>]
```

but there isn't a way to connect them to each other. Does that make sense? If so, do you have any suggestions?

altendky: `dict(zip(called.args.args[-len(called.args.defaults):], called.args.defaults))` assuming they are properly parallel lists, I think.

clavedeluna: Good thinking! Just trying out one example it does seem we can assume order

doc/data/messages/c/consider-using-keyword-argument/bad.py
```diff
@@ -0,0 +1,4 @@
+def print_car(name, owner, title, color="red", year=2022):
+    print(name)
+
+print_car("a car", "name", "title") # [consider-using-keyword-argument, consider-using-keyword-argument, consider-using-keyword-argument]
```
Pierre-Sassoulas, line 4:
```suggestion
def print_car(name, owner, title, color="red", year=2022):
    print(name)

car = "a car"
print_car(car, "name", "title", "green") # [consider-using-keyword-argument, consider-using-keyword-argument]
```

doc/data/messages/c/consider-using-keyword-argument/good.py
```diff
@@ -0,0 +1,4 @@
+def print_car(name, owner, title, color="red", year=2022):
+    print(name)
+
+print_car(name="a car",owner= "your name", title="title")
```
Pierre-Sassoulas, line 4:
```suggestion
def print_car(name, owner, title, color="red", year=2022):
    print(name)

car = "a car"
print_car(car, owner="your name", title="title", color="green")
```

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,89 @@
...
+
+class KeywordChecker(BaseChecker):
+    """Checks related to keyword arguments."""
+
+    name = "kwargs"
+    msgs = {
+        "W3501": (
+            "Call to `%s` missing keyword argument `%s`.",
+            "consider-using-keyword-argument",
+            "Used when method is called without specifying a keyword argument.",
```
Pierre-Sassoulas, line 27:
```suggestion
            "When using a literal directly in a function call, it's very hard to know what the argument is"
            " if a positional argument is used because there's no variable name or attribute name to "
            "rely on when reading. Using a keyword argument there's at least the attribute name to "
            "help the reader understanding the call."
```

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,93 @@
+# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
+# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from astroid import BoundMethod, UnboundMethod, nodes
```
DanielNoord, line 9:
```suggestion
from astroid import bases, nodes
```

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,93 @@
...
+    from pylint.lint import PyLinter
+
+
+class KeywordChecker(BaseChecker):
+    """Checks related to keyword arguments."""
+
+    name = "kwargs"
+    msgs = {
+        "W3501": (
+            "Call to `%s` missing keyword argument `%s`.",
```
DanielNoord, line 25:
```suggestion
            "Call to `%s` misses keyword argument `%s`.",
```

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,93 @@
...
+            "consider-using-keyword-argument",
+            "When using a literal directly in a function call, it's very hard to know what the argument is"
+            " if a positional argument is used because there's no variable name or attribute name to "
+            "rely on when reading. Using a keyword argument there's at least the attribute name to "
+            "help the reader understanding the call.",
```
DanielNoord, line 30:
```suggestion
            "When using a literal directly in a function call, it can be very hard to know which argument it is "
            "if a positional argument is used. In that case there's no variable name or attribute name to "
            "rely on when reading. By using a keyword argument there's at least the attribute name to "
            "help the reader understand the call.",
```

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,93 @@
...
+        """Check that called functions/methods are provided keyword arguments."""
+        called = safe_infer(node.func)
+
+        if not called:
+            return
+
+        needed_keywords, default_kwarg_names = self._get_args(called)
+
+        if len(needed_keywords) - len(default_kwarg_names) < 2:
+            # This checker will not apply if there are less than 2 positional args.
```
DanielNoord, line 44: This feels quite arbitrary. Why was this limit chosen?

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,93 @@
...
+            return
+
+        provided_kwarg_names = [
+            kwarg.arg for kwarg in node.keywords
+        ] + default_kwarg_names
+
+        for arg_name in needed_keywords:
+            if arg_name not in provided_kwarg_names:
+                self.add_message(
```
DanielNoord, line 53: It would be nice to have one message instead of multiple.

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    def visit_call(self, node: nodes.Call) -> None:
+        """Check that called functions/methods are provided keyword arguments."""
+        called = safe_infer(node.func)
+
+        if not called:
+            return
+
+        needed_keywords, default_kwarg_names = self._get_args(called)
```
DanielNoord, line 41: This is not type safe. You should do an `isinstance` check before the call to `_get_args`.

pylint/extensions/kwargs.py
```diff
@@ -0,0 +1,93 @@
...
+        self, node: nodes.ClassDef | nodes.FunctionDef | BoundMethod | UnboundMethod
+    ) -> tuple[list[str], list[str]]:
+        default_kwargs = []
+        needed_keywords = []
+
+        if isinstance(node, nodes.ClassDef):
+            needed_keywords = node.instance_attrs.keys()
+            if "__init__" in node.locals:
+                init_func = node.locals["__init__"][0]
+                if isinstance(init_func, nodes.FunctionDef):
```
DanielNoord, line 73: Can't you just use `self._get_args` again here?

tests/functional/ext/keyword/missing_keyword_arg.py
```diff
@@ -0,0 +1,96 @@
+"""Tests for consider-using-keyword-argument"""
+# pylint: disable=unused-argument, missing-function-docstring, missing-class-docstring,
+# pylint: disable=invalid-name, too-few-public-methods, line-too-long, too-many-arguments
```
DanielNoord, line 3:
```suggestion
# pylint: disable=too-few-public-methods, line-too-long, too-many-arguments
```

It's often nice to use "Pythonic" names in the tests as well as they can be used by non-maintainers are reference to what the checker does. Making the tests "pythonic" helps there I think.

DanielNoord (changes requested):

pylint/extensions/kwargs.py
clavedeluna: @Pierre-Sassoulas 's suggestion in this PR and in issue.

DanielNoord: Shall we make it configurable?

Pierre-Sassoulas: For example if we take ``join`` or ``enumerate`` or ``range`` everyone knows about the first positional argument, it's the most used one and it's the obvious one. If we're going to raise for those kind of call, then the false positive rate will be enormous. As soon as there is more than one argument I think the call can benefit from keyword args for litterals. Right now it's hard to be sure of what we need to do because the primers are spammed by call without any args/kwargs with litteral in the default value. I think we should rethink this once we can see the actual primer result.

clavedeluna: The primer result is out and it looks pretty reasonable to me, what do you think?

Pierre-Sassoulas: Not sure if I'm looking at the wrong primer but we should not raise here https://github.com/PyCQA/astroid/blob/b717e99964bf9e601b0acb939865990e64da19ee/astroid/bases.py#L160. There's no positional literal argument used in this call. If we can't differentiate genuine explicit positional value from implicit default values the check will not be usable.

pylint/extensions/kwargs.py
clavedeluna: isinstance check on what exactly? Since _get_args does that already, maybe you can expand on this?

pylint/extensions/kwargs.py
clavedeluna: I don't understand your comment since this is within self._get_args

pylint/extensions/kwargs.py
DanielNoord: `called` is not of the type that you have as annotation in `self._get_args`. Namely `SuccesfulInferenceResult` vs. a number of node types. You either need to expand the annotation in `_get_args` to be broader or you need to check the type of `called` before calling `_get_args`.

pylint/extensions/kwargs.py
DanielNoord: Yes but you are duplicating logic of a few lines below. Why don't you call the same function recursively?

pylint/extensions/kwargs.py
DanielNoord: I don't think you adressed this comment? Did you look at the other comments as well?

pylint/extensions/kwargs.py
clavedeluna: I personally disagree here, I think each missing kwarg should be its own msg. Maybe another maintainer can break the tie? @Pierre-Sassoulas

pylint/extensions/kwargs.py
clavedeluna: I'm sorry but I don't think your comment applies. The code is slightly different for each if statement for a reason.

pylint/extensions/kwargs.py
Pierre-Sassoulas: I also prefer one message, it's better not to spam and we should also handle the case where 20 kwargs are warning and would make the message too long imo.

clavedeluna:
note to self, look at https://github.com/PyCQA/astroid/pull/2006#event-8441769697

pylint/extensions/kwargs.py
clavedeluna: I believe this has now been addressed

DanielNoord:
I'm closing this PR as I hasn't seen any activity in a long time. It should be rebased on `main` and use the `astroid` feature that is now released.

PR #7824, opened 2022-11-22
Fix crash when using ``enumerate`` with ``start`` and a class attribute

clavedeluna:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

fixes
```
    start_val = node.value
AttributeError: 'Attribute' object has no attribute 'value'
```

Closes #7821

PR #7855, opened 2022-11-26
Revert "Fix crash when using ``enumerate`` with ``start`` and a class attribute (#7824)"

Pierre-Sassoulas:
## Type of Changes

## Description

This reverts commit 86b8c649061f49c7d121843675ceefe6ff3a8113 which broke the test on the maintenance branch. Do you think this is fixable by cherry-picking something else first @clavedeluna ? Otherwise we can just wait for 2.16 to release the fix.

### DIFF
--- a/doc/whatsnew/fragments/7821.bugfix
+++ b/doc/whatsnew/fragments/7821.bugfix
@@ -1,3 +1,0 @@
-Fixes a crash in the ``unnecessary_list_index_lookup`` check when using ``enumerate`` with ``start`` and a class attribute.
-
-Closes #7821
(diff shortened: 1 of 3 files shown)


clavedeluna:
How did this happen? Is it because we merge without rebasing on to main first?

If so, I've personally always used the rebase strategy to prevent these situations. But since this repo doesn't have that practice then we get these situations.

I tried to reproduce this at no avail, with or without tox locally. Are you sure this failure wasnt a total fluke? The test runs on main here https://github.com/PyCQA/pylint/actions/workflows/tests.yaml?query=branch%3Amain do not indicate to me that there is any problem??

Pierre-Sassoulas:
> How did this happen? Is it because we merge without rebasing on to main first?

The branch we're trying to merge on is the ``maintenance/2.15.x`` branch where we only cherry-pick the false positives and bugs. We might miss some refactor from main. If it isn't easy to know which commit also need to cherry picked we could wait for 2.16.0 to release this. If it would also bring new feature or warning and is impossible to separate then we must wait for 2.16.0.

mbyrnepr2 (approved):
👍  Ok I can see the tests pass on the maintenance branch. Not 100% sure why they failed on the maintenance branch but we can always handle it a little bit later & get it merged.

PR #7877, opened 2022-11-30
Cut oldstyle checks

nickdrozd:
## Type of Changes

## Description

Looking at the coverage report, I noticed these `except NotImplementedError` blocks weren't being hit. I think it's because they are holdovers from Python 2 days. Oldstyle classes aren't possible anymore, so I don't think it's possible to reach this code. So, delete it.


DanielNoord (approved):
Wonderful!

PR #7878, opened 2022-11-30
Fix ``missing-param-doc`` FP

### LINKED ISSUE #7827: False positive ``missing-param-doc`` when escaping underscore in docstring
### Bug description

In this code I need to escape the underscore, because else it's not valid RST.

```py
def get_value(something: int, raise_: bool = False) -> bool:
    """Return the maximum length of a column.

    Args:
        something: the something
        raise\\_: the other

    Returns:
        something
    """
    return something and raise_
```

### Configuration

```ini
[MAIN]
load-plugins = ["pylint.extensions.docparams"]
```

### Command used

```shell
pylint foo.py
```

### Pylint output

```shell
************* Module foo
foo.py:4:0: W9015: "raise_" missing in parameter documentation (missing-param-doc)
```

### Expected behavior

Escaped underscore is understood

### Pylint version

```shell
pylint 2.15.6
astroid 2.12.12
Python 3.7.2 (default, Jun  2 2022, 07:23:31)
```

clavedeluna:
## Type of Changes

## Description

As mentioned in linked issue, if this is too a hacky solution, we can reject the PR and attempt to update the `re_param_line` regex.

Closes #7827

### DIFF
--- a/pylint/extensions/_check_docs_utils.py
+++ b/pylint/extensions/_check_docs_utils.py
@@ -471,7 +471,7 @@ class GoogleDocstring(Docstring):
 
     re_param_line = re.compile(
         rf"""
-        \s*  ((?:\\?\*{{0,2}})?\w+)     # identifier potentially with asterisks
+        \s*  ((?:\\?\*{{0,2}})?[\w\\]+) # identifier potentially with asterisks or escaped `\`
         \s*  ( [(]
             {re_multiple_type}
             (?:,\s+optional)?
@@ -645,12 +645,16 @@ class GoogleDocstring(Docstring):
         entries.extend(self._parse_section(self.re_keyword_param_section))
         for entry in entries:
             # Remove escape characters necessary for asterisks and trailing underscores
-            entry = entry.replace("\\", "")
             match = self.re_param_line.match(entry)
             if not match:
                 continue
 
-            param_name, param_type, param_desc = match.groups()
+            param_name = match.group(1)
+            # Remove escape characters necessary for asterisks
+            param_name = param_name.replace("\\", "")
+
+            param_type = match.group(2)
+            param_desc = match.group(3)
 
             if param_type:
                 params_with_type.add(param_name)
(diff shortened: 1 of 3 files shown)


Pierre-Sassoulas (approved):
It does feel hacky, but it's simple and I don't see what could go wrong.

Pierre-Sassoulas:
Blocked by tests not being launched on backported branch that needs fixing so I don't have to rebase this later. Also another review would be welcome.

pylint/extensions/_check_docs_utils.py
```diff
@@ -644,7 +644,7 @@ def match_param_docs(self) -> tuple[set[str], set[str]]:
         entries = self._parse_section(self.re_param_section)
         entries.extend(self._parse_section(self.re_keyword_param_section))
         for entry in entries:
-            match = self.re_param_line.match(entry)
+            match = self.re_param_line.match(entry.replace("\\", ""))
```
DanielNoord, line 647: Why don't we change `self._re_param_line`? We can include `\\` there.

In any case, we are also replacing `\\`a bit further down here so that should be fixed as well then if we continue down the current road.

DanielNoord (changes requested):

pylint/extensions/_check_docs_utils.py
clavedeluna: I have no idea how to change `_re_param_line`, it's really difficult to read. If you can, plz do paste your regex here!

DanielNoord: True, I'd rather not touch it as well, but just replacing all `\\` seems error prone as well.
I think it should be as easy as allowing both `\w` and `\\` in `_re_param_line`. You can check this with regex101.com

clavedeluna: > should be as easy

Oh I tried, I really tried 🤣 , but nothing quite "works". As soon as you try to add some variation of (\\)* or something (Tried lots of ways), it breaks other tests!

clavedeluna: > In any case, we are also replacing \\a bit further down here so that should be fixed as well then if we continue down the current road.

reasonable and doesn't seem to break tests

DanielNoord (changes requested):
I fixed the regex and made some suggestions! 😄

DanielNoord (approved):

PR #7880, opened 2022-11-30
Bugfix - extend ``magic-value-comparison`` support for parsing numerical types and `\\` in rcfile configuration

### LINKED ISSUE #7858: How to specify numbers, newlines (`'\n'`) and an empty string in the rcfile?
### Question

I want to add `'\n'` to [the `valid-magic-values` configuration option](https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#valid-magic-values)
I'm using `.pylintrc`, so the example there doesn't help. I got `__main__` to work by not quoting it, but `0`, `1`, `''`, and `'\n'` all don't work, with double quotes, no quotes, or single quotes.
`pylint --valid-magic-values=a,b --generate-rcfile` doesn't output anything related to valid-magic-values.

I expect details about how the rcfile works to be located in [the configuration documentation](https://pylint.pycqa.org/en/latest/user_guide/configuration/index.html)

### Additional context

Downloaded from `pylint@ git+https://github.com/PyCQA/pylint@a6a544613fd6cd28c3213751173f1b0e4205d1c0`
`pylint --version`:
```
pylint 2.16.0-dev
astroid 2.12.13
Python 3.9.15 (main, Nov 11 2022, 13:33:56) 
[GCC 9.4.0]
```

orSolocate:
## Type of Changes

## Description

The issue was that the rcfile parses the `valid-magic-values` config argument as a list of strings, which does not allow for `int`/`float` magic numbers, empty strings and backslashes like `\n`.

This is a hotfix that provides an all-around solution for potentially used cases, plus a documentation clarification.

For integers, floats, negative numbers, etc: just write them as is e.g. `valid-magic-values=1, -2, -3.54..`
For string, strings that represent numbers and empty strings just add `'` e.g.  `valid-magic-values='',  'magic', '15'`
For your `\n` just write it as is: `valid-magic-values='\n',  '\t'`

The functional test was modified to be tested on `rcfile` parsing instead of on a default value.

Closes #7858


pylint/extensions/magic_value.py
```diff
@@ -37,11 +42,30 @@ class MagicValueChecker(BaseChecker):
                 "default": (0, -1, 1, "", "__main__"),
                 "type": "csv",
                 "metavar": "<argument names>",
-                "help": " List of valid magic values that `magic-value-compare` will not detect.",
+                "help": " List of valid magic values that `magic-value-compare` will not detect."
+                "supports integers, floats, negative numbers, for empty string enter ``''``,"
```
DanielNoord, line 46:
```suggestion
                "help": " List of valid magic values that `magic-value-compare` will not detect. "
                "Supports integers, floats, negative numbers, for empty string enter ``''``,"
```

pylint/extensions/magic_value.py
```diff
@@ -74,10 +98,23 @@ def _check_constants_comparison(self, node: nodes.Compare) -> None:
...
 
-    def _is_magic_value(self, node: nodes.NodeNG) -> bool:
+    def _is_magic_value(self, node: nodes.Const) -> bool:
         return (not utils.is_singleton_const(node)) and (
-            node.value not in self.linter.config.valid_magic_values
+            node.value not in (self.valid_magic_vals)
+        )
+
+    @staticmethod
+    def _parse_rcfile_magic_numbers(parsed_val: str) -> float | str:
```
DanielNoord, line 107: This feels like it should in the configuration parsing instead of individual checkers. Have you tried if that breaks stuff?

orSolocate: Good idea it'd be much better, I thought this is a behavior that is not useful for other checkers initially.. I would have a try though.
Are all checkers with `pylintrc` config vars have functional tests for regression? I wouldn't be surprised if not...

orSolocate:
Any help with the `mypy` pre-commit check? I struggled with that a bit..

GideonBear:
It's probably better to parse all backslashes instead of just `\n` and `\t`. If someone needs to specify an escaped backslash, they should just use `\\`.

doc/whatsnew/fragments/7858.bugfix
```diff
@@ -0,0 +1,5 @@
+Extend support for ``magic-number`` plugin rcfile parsing. Now supports newline and tab characters,
+empty string and floating point numbers.
+You can use ``magic-number`` checker with ``--load-plugins=pylint.extensions.magic_value``.
+
+Closes #7858
```
Pierre-Sassoulas, line 5: No need to add this in the changelog as the issue was unreleased afaik.

Pierre-Sassoulas (changes requested):
Regarding the error with mypy it seems a Type is given when a Tuple was expected, what is the issue you're facing exactly ?

pylint/extensions/magic_value.py
DanielNoord: I think it makes sense to do this for all `pylintrc` files. We'll get a report about it if we break anything :smile:

Pierre-Sassoulas:
@orSolocate this is a blocker for releasing 2.16.0, do you think you'll have time for this in the near future or should we take over ?

orSolocate:
@Pierre-Sassoulas  - I will finish this upload in the next couple of days

orSolocate:
> It's probably better to parse all backslashes instead of just `\n` and `\t`. If someone needs to specify an escaped backslash, they should just use `\\`.

@GideonBear  - You are right. I am adding a support for all backslashes used. a single `\` when you want to escape and a double `\\` when you want a backslash. e.g. if you want the parser to ignore newline character you should write in the rcfile `\n`, if you want to ignore `\\b` in your code, you should write in the rcfile `\\b`. I will add a regression cases that cover those 2 :)

pylint/extensions/magic_value.py
orSolocate: Added this to `argument_manager.py` in `_parse_configuration_file` method in the begining (before the parser to `self.config`):

```
            for idx,arg in enumerate(arguments):
                if "--" not in arg:
                    arguments[idx] = ','.join([self._post_parse_arg(val) for val in arg.strip(',').replace(" ", "").split(',')])
```

running on `magic_value` test you get, as input to this method:

```
>>>arguments
['--load-plugins', 'pylint.extensions.magic_value,', '--valid-magic-values', "0, 1, -1, __main__,'', \\n, \\\\b"]
```

and after this snippet:

```
>>>arguments
['--load-plugins', 'pylint.extensions.magic_value', '--valid-magic-values', '0,1,-1,__main__,,\n,\\b']
```

which is kind of the wanted behavior (Except for maybe the empty string). but after:

```
            self.config, parsed_args = self._arg_parser.parse_known_args(
                arguments, self.config
            )
```

you get:

```
>>>self.config.valid_magic_values
['0', '1', '-1', '__main__', '\\b']
```

which kills the empty string and newline character. 

So I am not sure how to tackle this, I can't control the argparse package..
If it should be done, I would do it in another commit, right now uploading in `magic_value.py`

orSolocate: this is the extra method for comletion:

  ```
  @staticmethod
    def _post_parse_arg(parsed_val: str) -> float | str:
        parsed_val = parsed_val.encode().decode('unicode_escape')

        if parsed_val.startswith("'") and parsed_val.endswith("'"):
            return parsed_val[1:-1]

        return parsed_val
```

File: tests/functional/ext/magic_value_comparison/magic_value_comparison.py [hunk 4]
```diff
@@ -1,33 +1,35 @@
...
-shouldnt_raise = var == '__main__'
-shouldnt_raise = var == 1
-shouldnt_raise = var == 0
-shouldnt_raise = var == -1
-shouldnt_raise = var == True  # [singleton-comparison]
-shouldnt_raise = var == False  # [singleton-comparison]
-shouldnt_raise = var == None  # [singleton-comparison]
-celebration_started = Christmas.EVE.value == Christmas.MONTH.value
-shouldnt_raise = var == ""
+"""
```
Pierre-Sassoulas, line 1: I suppose the line ending was modified here, could you revert it ?

Pierre-Sassoulas (changes requested):
Thank you for tackling this @orSolocate !

tests/functional/ext/magic_value_comparison/magic_value_comparison.py
orSolocate: I think I did now

Pierre-Sassoulas: Yes thank you 👍

Pierre-Sassoulas (approved):
LGTM, i'm going to wait for Daniel's review before merging. We still have some small issue before releasing 2.16.0 alpha anyway :)

pylint/extensions/magic_value.py
```diff
@@ -37,11 +38,30 @@ class MagicValueChecker(BaseChecker):
                 "default": (0, -1, 1, "", "__main__"),
                 "type": "csv",
                 "metavar": "<argument names>",
-                "help": " List of valid magic values that `magic-value-compare` will not detect.",
+                "help": " List of valid magic values that `magic-value-compare` will not detect,"
+                "Supports integers, floats, negative numbers, for empty string enter ``''``,"
```
DanielNoord, line 42:
```suggestion
                "help": " List of valid magic values that `magic-value-compare` will not detect. "
                "Supports integers, floats, negative numbers, for empty string enter ``''``,"
```

pylint/extensions/magic_value.py
```diff
@@ -37,11 +38,30 @@ class MagicValueChecker(BaseChecker):
...
+                " for backslash values just use one backslash e.g \\n.",
             },
         ),
     )
 
+    def __init__(self, linter: PyLinter) -> None:
+        """Initialize checker instance."""
+        super().__init__(linter=linter)
+        self.valid_magic_vals: tuple[(float | str), ...] = ()
```
DanielNoord, line 51: I have never seen this syntax.. Shouldn't it be:
```suggestion
        self.valid_magic_vals: tuple[float | str, ...] = ()
```

pylint/extensions/magic_value.py
```diff
@@ -37,11 +38,30 @@ class MagicValueChecker(BaseChecker):
...
+
+    def open(self) -> None:
+        if self._is_default_magic_vals():
```
DanielNoord, line 54: Let's rename this to something like `_is_split_magic_vals_option` cause that is what we are essentially checking.

pylint/extensions/magic_value.py
```diff
@@ -74,11 +94,21 @@ def _check_constants_comparison(self, node: nodes.Compare) -> None:
                 confidence=HIGH,
             )
 
-    def _is_magic_value(self, node: nodes.NodeNG) -> bool:
+    def _is_magic_value(self, node: nodes.Const) -> bool:
```
DanielNoord, line 97: `mypy` probably won't like this when `astroid` finally has `py.typed` but that is something for another day.

pylint/extensions/magic_value.py
```diff
@@ -74,11 +94,21 @@ def _check_constants_comparison(self, node: nodes.Compare) -> None:
...
 
+    @staticmethod
+    def _parse_rcfile_magic_numbers(parsed_val: str) -> float | str:
+        parsed_val = parsed_val.encode().decode("unicode_escape")
+
+        if parsed_val.startswith("'") and parsed_val.endswith("'"):
+            return parsed_val[1:-1]
+
+        is_number = regex_match(r"[-+]?\d+(\.0*)?$", parsed_val) is not None
+        return float(parsed_val) if is_number else parsed_val
```
DanielNoord, line 110:
```suggestion
        is_number = regex_match(r"[-+]?\d+(\.0*)?$", parsed_val)
        return float(parsed_val) if is_number else parsed_val
```

DanielNoord (changes requested):
I guess this works as the tests pass, but I still don't really understand how a config option that is said to be of type `csv` ends up as a string.

I'd also like a test without spaces between the values in the config option to see how that behaves.

pylint/extensions/magic_value.py
orSolocate: we are checking if the valid-magic-values config is taken from default (being a tuple) or from the rcfile (not being a tuple)

orSolocate: maybe we can rename it to: `_is_magic_vals_externally_parsed`

pylint/extensions/magic_value.py
orSolocate: was already Const typing (maybe you compared previous commit?)

pylint/extensions/magic_value.py
DanielNoord: `_is_magic_vals_set_by_user`?

orSolocate:
> I guess this works as the tests pass, but I still don't really understand how a config option that is said to be of type `csv` ends up as a string.

why `csv`? when I followed the code that runs through `test_functional.py` it flowed through `ini` parsing.
It is weird but this config-value string seems to be split into sub arguments *only* in the `argparse` function...

DanielNoord:
> > I guess this works as the tests pass, but I still don't really understand how a config option that is said to be of type `csv` ends up as a string.
> 
> why `csv`? when I followed the code that runs through `test_functional.py` it flowed through `ini` parsing. It is weird but this config-value string seems to be split into sub arguments _only_ in the `argparse` function...

See L38. That says `"type": "csv",`. If that isn't parsed correctly in `ini` then we might have even more troubles 😅 Don't we call the `type` for the functional test?

orSolocate:
> I'd also like a test without spaces between the values in the config option to see how that behaves.

This is the rcfile in the experiment above:
```
[MAIN]
load-plugins=pylint.extensions.magic_value,

valid-magic-values=0, 1, -1, __main__,'', \n, \\b

```

so there was no space between the config option and value.

As for the spaces between values themselves, you can see in the way I parsed it that I got rid of the whitespaces:

` arguments[idx] = ','.join([self._post_parse_arg(val) for val in arg.strip(',').replace(" ", "").split(',')])`

DanielNoord:
> > I'd also like a test without spaces between the values in the config option to see how that behaves.
> 
> This is the rcfile in the experiment above:
> 
> ```
> [MAIN]
> load-plugins=pylint.extensions.magic_value,
> 
> valid-magic-values=0, 1, -1, __main__,'', \n, \\b
> ```
> 
> so there was no space between the config option and value.
> 
> As for the spaces between values themselves, you can see in the way I parsed it that I got rid of the whitespaces:
> 
> ` arguments[idx] = ','.join([self._post_parse_arg(val) for val in arg.strip(',').replace(" ", "").split(',')])`

Yeah, but to make sure your joining and replacing is working as intended I think it would be good to also test a situation where we have no spaces or even mixed usage (both spaces and not).

Pierre-Sassoulas:
There are configuration functional tests if required but imo refactor of the configuration parser is out of the scope of this MR :)

DanielNoord:
> There are configuration functional tests if required but imo refactor of the configuration parser is out of the scope of this MR :)

I think if there is a real issue it is that the functional test don't use the config parsing as a normal run does. But we would need to investigate that!

Pierre-Sassoulas:
> I think if there is a real issue it is that the functional test don't use the config parsing as a normal run does. But we would need to investigate that!

The testutil's coverage is very low. I've been thinking about moving the testutil in their own repository and making them better. (I went as far as creating the repo by filtering commits from pylint https://github.com/PyCQA/pylint-testutils). But there is a kind of circular dependencies if we do that, so I did not follow through. Doing this before 3.0 would be nice so we can deprecate / remove in 3.0. What do you think ?

orSolocate:
> See L38. That says `"type": "csv",`. If that isn't parsed correctly in `ini` then we might have even more troubles 😅 Don't we call the `type` for the functional test?

Oh now I understand what you were saying. I followed the call stack reaching the `_parse_configuration_file` of `arguments_manager,py`. the arguments parsed from the rcfile come from `parse_config_file` in `config_file_parser.py`, if you follow the function you see that it parses it with ` return self._parse_ini_file(file_path)`

This is why i said that I think it parses it as `ini`. I didn't see anything regarding the type or regarding `csv` parsing.. but maybe I missed it? I am not familiar with the code and it is not straight forward to track..

Aren't we verifying all the external configuration types are correctly configured with special testing? if not we could open another task for it, that seems like an important matter.

DanielNoord:
Yeah, would you mind making an issue for this? We can put it on the`2.17` milestone.

pylint/extensions/magic_value.py
```diff
@@ -37,11 +38,30 @@ class MagicValueChecker(BaseChecker):
                 "default": (0, -1, 1, "", "__main__"),
                 "type": "csv",
                 "metavar": "<argument names>",
-                "help": " List of valid magic values that `magic-value-compare` will not detect.",
+                "help": " List of valid magic values that `magic-value-compare` will not detect."
```
DanielNoord, line 41:
```suggestion
                "help": "List of valid magic values that `magic-value-compare` will not detect. "
```

pylint/extensions/magic_value.py
```diff
@@ -37,11 +38,30 @@ class MagicValueChecker(BaseChecker):
...
+        if self._magic_vals_ext_configured():
+            self.valid_magic_vals = tuple(
+                self._parse_rcfile_magic_numbers(value)
+                for value in self.linter.config.valid_magic_values
+            )
+        else:
+            self.valid_magic_vals = self.linter.config.valid_magic_values
+
+    def _magic_vals_ext_configured(self) -> bool:
+        return not isinstance(self.linter.config.valid_magic_values, tuple)
```
DanielNoord, line 63: One comment here would help I think.

DanielNoord (changes requested):
2 final nits. Rest LGTM!

Thanks @orSolocate, this proved to be more work than you might have expected!

DanielNoord (approved):

PR #7886, opened 2022-12-01
bugfix  - ``assignment-from-none`` False Positive #7853

### LINKED ISSUE #7853: False Positive - ``assignment-from-none`` raised when function not necessarily returns `None`
### Bug description

`assignment-from-none` is emitted where `func`  does not necessarily returns `None`

consider this `example.py` file to reproduce the error:

```
def get_func(param):

    if param is None:
        def func():
            return None
    else:
        def func():
            return param

    return func


def process_val(param):

    func = get_func(param)
    val = func()

    # Do stuff with 'val'; *None* is an useful case.
    return val
```

### Configuration

_No response_

### Command used

```shell
pylint example.py
```

### Pylint output

```shell
`:16:4: E1128: Assigning result of a function call, where the function returns None (assignment-from-none)`
```

### Expected behavior

No Error

### Pylint version

```shell
pylint 2.16.0-dev
astroid 2.12.13
Python 3.8.10
```

### OS / Environment

Windows 10

### Additional dependencies

_No response_

orSolocate:
## Type of Changes

|   | :sparkles: New feature |

## Description

See comment about this PR in [this comment](https://github.com/PyCQA/pylint/issues/7853#issuecomment-1329855791).

Closes #7853

### DIFF
--- a/doc/whatsnew/fragments/7853.bugfix
+++ b/doc/whatsnew/fragments/7853.bugfix
@@ -1,0 +1,3 @@
+Fix False Positive issue regarding a complex ``assignment-from-none``.
+
+Closes #7853.
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1371,13 +1371,29 @@ def safe_infer(
         raise AstroidError from e
 
     if value is not astroid.Uninferable:
-        inferred_types.add(_get_python_type_of_node(value))
+        if isinstance(value, nodes.FunctionDef):
+            try:
+                inferred_results = value.infer_call_result(context=context)
+                for inferred_result in inferred_results:
+                    if inferred_result is astroid.Uninferable:
+                        return None
+                    inferred_types.add(inferred_result.pytype())
+            except astroid.InferenceError:
+                return None  # There is some kind of ambiguity
+        else:
+            inferred_types.add(_get_python_type_of_node(value))
 
     try:
         for inferred in infer_gen:
-            inferred_type = _get_python_type_of_node(inferred)
-            if inferred_type not in inferred_types:
-                return None  # If there is ambiguity on the inferred node.
+            if isinstance(inferred, nodes.FunctionDef):
+                inferred_type = inferred.infer_call_result(context=context)
+                for inf_val_type in inferred_type:
+                    if inf_val_type not in inferred_types:
+                        return None
+            else:
+                inferred_type = _get_python_type_of_node(inferred)
+                if inferred_type not in inferred_types:
+                    return None  # If there is ambiguity on the inferred node.
             if (
                 compare_constants
                 and isinstance(inferred, nodes.Const)
--- a/tests/functional/a/assignment/assignment_from_none_7853.py
+++ b/tests/functional/a/assignment/assignment_from_none_7853.py
@@ -1,0 +1,22 @@
+"""
+Check False positive issue #7853
+"""
+# pylint: disable=missing-function-docstring
+
+
+def get_func(param):
+    if param is None:
+        def func():
+            return None
+    else:
+        def func():
+            return param
+    return func
+
+
+def process_val(param):
+
+    func_a = get_func(param)
+    val = func_a()
+    # Do stuff with 'val'; *None* is an useful case.
+    return val


orSolocate:
That is interesting:

```
Expected in testdata:
  35: possibly-unused-variable

Unexpected in testdata:
  35: unused-variable
FAILED tests/test_functional.py::test_functional[useless_with_lock] - AssertionError: Wrong results for file "useless_with_lock":
```

Pierre-Sassoulas (commented):
We should definitely not touch ``safe_infer`` for something happenning in a particular checker, this is going to affect a lot of code. What we can do is make astroid better or make the checker handle more specific use case.

orSolocate:
> We should definitely not touch `safe_infer` for something happening in a particular checker, this is going to affect a lot of code.

I uploaded a fix that refactors the `safe_infer` method and using its parts to build a `safe_infer_multiple` function. It took a lot of effort to make all regression tests pass locally.. since there are many logical flows :( 
The `safe_infer_multiple` returns all return nodes and checks all of them, no matter the order whatsoever. The `typecheck` checker use this fucntion to check for the `assignment-from-none` and the `assignment-no-return` messages.

There were some unnecessary things in the original function, like the use of a `set` that can have a maximum of  1 element, and then verifying that in a condition in the `return` statement..

orSolocate:
Not sure where it fails, when I run ` test_runner_with_arguments[run_pylint]` test locally it passes..

The `/home/runner/work/pylint/pylint/tests/test_pylint_runners.py:38:4: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)` error does not pop out on my machine...

Could anybody help solve this mystery? Maybe it has something to do with the tox environment?

pylint/checkers/utils.py
```diff
@@ -1361,45 +1361,95 @@ def safe_infer(
...
-    if value is not astroid.Uninferable:
-        inferred_types.add(_get_python_type_of_node(value))
 
+@lru_cache(maxsize=1024)
+def safe_infer_multiple(
+    node: nodes.NodeNG,
+    context: InferenceContext | None = None,
+    *,
+    compare_constants: bool = False,
+) -> Any | None:
```
DanielNoord, line 1388: Why is this `Any`?

DanielNoord (commented):
I'm still not sure whether I like this approach. It feels like we're complicating the `safe_infer` util for a very specific use case.
We could also consider doing the double infer call in the checker itself. We would need to duplicate a little bit of `safe_infer` inside the checker, but at least the `safe_infer` function remains somewhat legible.

Perhaps other maintainers have an opinion here?

Pierre-Sassoulas (changes requested):
We might need to extend ``safe_infer``, but I consider it to be an API. It's used too much in pylint as well as in lib depending on it, to be open to modification.

doc/whatsnew/fragments/7853.bugfix
```diff
@@ -0,0 +1,3 @@
+Fix False Positive issue regarding a complex ``assignment-from-none``.
```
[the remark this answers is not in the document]
orSolocate: Agree. yeah we just don't know what is the final solution yet.

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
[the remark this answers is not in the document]
orSolocate: yeah my bad, will remove it in my next commit

pylint/checkers/utils.py
orSolocate: I asked myself the same question. But this is what `mypy`  accepted. :S

orSolocate:
I agree that this function is super fragile. I had a really hard time making all the functional tests pass..

The purpose of the change was just to refactor `safe_infer` to be able to use its parts in a different function to be used by the problematic checker. The only parts that changed is that the `inferred_types` `set` is removed, since it can contain a maximum of 1 element, and the ending `return` `else` case removed, since it can never be reached, for the same reason.

pylint/checkers/utils.py
DanielNoord: Yes, that's because `mypy` doesn't do any checking on `astroid`. It should be `InferenceResult`.

orSolocate:
> We could also consider doing the double infer call in the checker itself.

@DanielNoord  - I think I didn't understand the suggestion here, could you explain more on that please?

DanielNoord:
> > We could also consider doing the double infer call in the checker itself.
> 
> @DanielNoord - I think I didn't understand the suggestion here, could you explain more on that please?

The issue is that for some nodes we want to do a double inference right? So not call `safe_infer` once, but actually twice? Can't we hide that logic inside the checker?

orSolocate:
> The issue is that for some nodes we want to do a double inference right? So not call `safe_infer` once, but actually twice? Can't we hide that logic inside the checker?

Not necessary twice, for our case which is a function infer, you want `safe_infer` to retrieve all function declarations, it can theoretically be any number, see this example:

"""
```
def get_func(param):
    if param is None:
        def func():
            return None
    elif isinstance(param, int):
        def func():
            return str(param)
    else:
        def func():
            return param
    return func
```
"""

In this case you want to check `return` nodes for all 3 `functionDefs`.
The `safe_infer` returns only the **first** `functionDef`, so you can't get around it the way I see it without modifying what it returns..

DanielNoord:
Can't we get away with calling `.infer()` then? We'd need to catch `StopIteration` and `InferenceError`, but we would get access to all inferred nodes.

jacobtylerwalls:
`safe_infer()`'s docstring should really be updated to say return the _first_ inferred value to avoid confusion.

jacobtylerwalls:
We already have a cached util for getting all inferred values, it's `utils.infer_all()`.

jacobtylerwalls:
Thanks for helping think through these issues. I think changing the assignment-from-none checker to use `infer_all()` instead of `safe_infer()` is the right approach. Feel free to reopen or start with a fresh PR if you'd like to continue with the contribution.

orSolocate:
Hey @jacobtylerwalls, it is nice to see you kept contributing to Pylint. A pleasure to work together again. I can look into it.

PR #7888, opened 2022-12-02
7788 no else return False Negative fix

### LINKED ISSUE #7788: False negative for `no-else-return` after `except`
### Bug description

This snippet returns in the `except` block, and therefore the `else` is unnecessary:

```python
def f() -> bool:
    try:
        print('asdf')
    except:  # pylint: disable = bare-except
        return False
    else:
        return True
```

It should raise `no-else-return`.

### Configuration

_No response_

### Command used

```shell
`pylint asdf.py`
```

### Pylint output

```shell
No warnings raised
```

### Expected behavior

Should raise `no-else-return`

### Pylint version

```shell
pylint 2.16.0-dev
astroid 2.13.0-dev0
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
```

clavedeluna:
## Type of Changes

## Description

`no else return` and `no else raise` now account for try/except/else pattern.

Closes #7788

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -726,6 +726,12 @@ class RefactoringChecker(checkers.BaseTokenChecker):
         msg_id: str,
         returning_node_class: nodes.NodeNG,
     ) -> None:
+        if isinstance(node, nodes.TryExcept) and isinstance(
+            node.parent, nodes.TryFinally
+        ):
+            # Not interested in try/except/else/finally statements.
+            return
+
         if not node.orelse:
             # Not interested in if /try statements without else.
             return
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -73,6 +73,16 @@ def _if_statement_is_always_returning(
     return any(isinstance(node, returning_node_class) for node in if_node.body)
 
 
+def _except_statement_is_always_returning(
+    node: nodes.TryExcept, returning_node_class: nodes.NodeNG
+) -> bool:
+    """Detect if all except statements return."""
+    return all(
+        any(isinstance(child, returning_node_class) for child in handler.body)
+        for handler in node.handlers
+    )
+
+
 def _is_trailing_comma(tokens: list[tokenize.TokenInfo], index: int) -> bool:
     """Check if the given token is a trailing comma.
 
@@ -531,7 +541,7 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             node.value.value, bool
         )
 
-    def _is_actual_elif(self, node: nodes.If) -> bool:
+    def _is_actual_elif(self, node: nodes.If | nodes.TryExcept) -> bool:
         """Check if the given node is an actual elif.
 
         This is a problem we're having with the builtin ast module,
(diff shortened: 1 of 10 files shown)


nickdrozd:
Tested locally; works!

clavedeluna:
Primer output looks spot on! 🚀

mbyrnepr2:
It may be good to add a test which uses `finally` - would it be clear from the emitted message what the user needs to change in the code? https://github.com/getsentry/sentry/blob/916fe38f45a23d27f2bde851e6521b66c1404299/src/sentry/utils/metrics.py#L262

clavedeluna:
@mbyrnepr2 good catch. I'm thinking for this type of example

```
    try:
        yield current_tags
    except Exception:
        current_tags["result"] = "failure"
        raise
    else:
        current_tags["result"] = "success"
    finally:
        timing(key, time.monotonic() - start, instance, current_tags, sample_rate)
```

or perhaps any with `finally`, we should actually never raise the message. What do you think @nickdrozd ?

nickdrozd:
I agree. The purpose of the check is to show code that can be simplified by removing an `else`. But with `finally` this cannot be done, so no warning should be raised.

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -531,7 +541,7 @@ def _is_bool_const(node: nodes.Return | nodes.Assign) -> bool:
             node.value.value, bool
         )
 
-    def _is_actual_elif(self, node: nodes.If) -> bool:
+    def _is_actual_elif(self, node: nodes.If | nodes.TryExcept) -> bool:
```
mbyrnepr2, line 544: Is this change necessary?

mbyrnepr2 (commented):
Looks solid overall. One question.

pylint/checkers/refactoring/refactoring_checker.py
clavedeluna: I believe so, because in this line https://github.com/PyCQA/pylint/pull/7888/files/341e7098da296c12fffefbfe43eb4c133426132b#diff-af1d215b95c8b3ac59218ba588186da4aefd936eef4732518196ea309c3cc788R739  a node of either type can land in there

mbyrnepr2 (approved):

pylint/checkers/refactoring/refactoring_checker.py
mbyrnepr2: 👍

DanielNoord:
@Pierre-Sassoulas Can this be merged with a backport? I have lot track of whether merging with the backport label works or doesn't :sweat_smile:

Pierre-Sassoulas:
Everything in 2.15.8 was backported automatically. The maintenance branch and the main branch were separated a long time ago so sometime we have cherry-picking issue but apart from that the backporting label is working well. (This should not be an issue anymore when we release 2.16.0). However we're only backporting false positives or crash/bug fixes so we don't add new messages in patch version so I don't think this one should be backported. (See https://pylint.pycqa.org/en/latest/user_guide/installation/upgrading_pylint.html)

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -707,23 +721,38 @@ def visit_with(self, node: nodes.With) -> None:
...
     ) -> None:
+        if isinstance(node, nodes.TryExcept) and isinstance(
+            node.parent, nodes.TryFinally
+        ):
+            # Not interested in try/except/else/finally statements.
+            return
+
         if not node.orelse:
-            # Not interested in if statements without else.
+            # Not interested in if /try statements without else.
```
Pierre-Sassoulas, line 736:
```suggestion
            # Not interested in if/try statements without else.
```

Pierre-Sassoulas (approved):
👍

Pierre-Sassoulas (dismissed):

Pierre-Sassoulas (approved):

Issue #6085, opened 2022-04-01
``bad-format-character`` false negative on ``.format()`` and f-strings


DudeNr33:
### Bug description

# pylint: disable=consider-using-f-string, missing-module-docstring
PI = 3.1415
print("Pi is %.2v" % PI)
print("Pi is {:.2v}".format(PI))
print(f"Pi is {PI:.2v}")

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:3:6: E1300: Unsupported format character 'v' (0x76) at index 9 (bad-format-character)
```

### Expected behavior

``bad-format-character`` is currently only emitted for the old-style string formatting.

It should also cover new style formatting and f-strings.

### Pylint version

```shell
pylint 2.14.0-dev0
astroid 2.11.2
Python 3.9.6 (default, Jul 28 2021, 21:15:06) 
[Clang 12.0.5 (clang-1205.0.22.9)]
```

### OS / Environment

macOS Big Sur Version 11.6

DanielNoord:
For anybody taking this on, after #6086 has landed we will need to update the documentation for this as well.

PR #7895, opened 2022-12-03
Update format parsing to support f-strings and the str.format function

clavedeluna:
## Type of Changes

## Description

This PR takes over https://github.com/PyCQA/pylint/pull/6555 in hopes of getting the work merged.

Closes https://github.com/PyCQA/pylint/issues/6163
Closes https://github.com/PyCQA/pylint/issues/6085


DanielNoord:
@clavedeluna Will you also work on fixing the tests? Otherwise it doens't make a lot of sense to keep this PR open.

clavedeluna:
> @clavedeluna Will you also work on fixing the tests? Otherwise it doens't make a lot of sense to keep this PR open.

Of course. What would be the point of this PR otherwise? The following work needs to happen:

- [ ] add typing
- [ ] fix tests

Typing is the first thing I think but I haven't had the energy to work on it much yet. If anyone is interested in contributing, please feel free to push commits.

DanielNoord:
I'd focus on the tests first. The original PR never got the test suite to pass so there might be crucial incompatible changes in here. Typing of those change might become redundant after the incompatibilities have been fixed.

clavedeluna:
I can now see why this was so difficult to get right in the first place. I did quite a bit but there are two tests that fail, and in trying to fix them so many other things break.

@DanielNoord maybe I won't be able to complete this, but I did find a couple functional tests that I'm not sure are right. In string_formatting.py:

```
"{0.missing.length}".format(ReturnYes())
 "{1.missing.length}".format(ReturnYes())
```

shouldn't those two lines emit some message, given that `missing` is not a real module?

DanielNoord:
@clavedeluna Which messages were you thinking of?

clavedeluna:
> @clavedeluna Which messages were you thinking of?

Given this
```
"{0.missing}".format(2) # [missing-format-attribute]
```

I would expect

```
from missing import Missing
class ReturnYes:
    """ can't be properly inferred """
    missing = Missing()


"{0.missing.length}".format(ReturnYes())
"{1.missing.length}".format(ReturnYes())
```

to also raise `missing-format-attribute`. In the first example, `2` doesn't have a `.missing` attr. In the second group of cases, `ReturnYes().missing.length` also doesn't exit.

DanielNoord:
Sorry for the late response, but I agree. That seems to make sense to me.

pylint/checkers/strings.py
```diff
@@ -404,10 +424,75 @@ def visit_binop(self, node: nodes.BinOp) -> None:
...
+                    continue
+                try:
+                    arg_type = utils.safe_infer(field)
+                except astroid.InferenceError:
+                    continue
+
+                if len(value.format_spec.values) == 0:
+                    continue
+
+                def convert_node_to_string(x):
```
clavedeluna, line 469: move this func out for performance

pylint/checkers/strings.py
```diff
@@ -449,6 +534,106 @@ def _detect_vacuous_formatting(
...
+                format_type = conversion
+
+            if (
+                format_type is not None
+                and arg_type
+                and arg_type != astroid.Uninferable
+                and not arg_matches_format_type(arg_type, format_type)
+            ):
+                self.add_message(
+                    "bad-string-format-type",
```
clavedeluna, line 632: try to dedupe with above code

pylint/checkers/strings.py
```diff
@@ -634,6 +846,51 @@ def _check_new_format_specifiers(
                     # can't check further if we can't infer it
                     break
 
+        for position, specifiers in pos_args.items():
```
clavedeluna, line 849: try to dedupe with above code

clavedeluna:
I'm thinking that the two primer reports are false positives but the one in music21 seems that it may be really hard to fix because `:d` is referring to the return from a function

Pierre-Sassoulas:
I think for music21 bools can be considered as ``int`` too so it should not raise even without considering the function return, right ?

clavedeluna:
I'm going to open this PR up for taking over since I just can't find the time to come back to it. Here is what I believe remains to be done:
- fix primer results as per Pierre's comment above
- fix pre-commit. this will require adding typing

DanielNoord:
Closing due to inactivity. Sorry this seems to have died due to me not responding actively enough. Life caught up to me.

PR #7912, opened 2022-12-08
Add more docs for messages

clavedeluna:
## Type of Changes

## Description

Refs #5953


Pierre-Sassoulas (changes requested):
Not sure about the lock example. Making a "real" example would mean a lot of boilerplate and creating ``Thread``... but I think it's better to show that for mutual exclusion you need the assignment.

Pierre-Sassoulas (approved):

PR #7931, opened 2022-12-12
Fix inconsistent argument exit code when argparse exit with its own error code

dmrlawson:
## Type of Changes

## Description

We were bitten by this causing us to think that pylint was completing successfully when in fact a typo in a configuration file was causing argparse to exit with exit code 2. This didn't happen in older versions of pylint for whatever reason. The [documentation for pylint](https://pylint.pycqa.org/en/latest/user_guide/usage/run.html#exit-codes) suggests that this means "error message issued" but that pylint executed properly when this is not the case.

It feels a bit clumsy to just add in `try..except`, but on the other hand it's perhaps unexpected that argparse will raise a `SystemExit`. I'm open to any suggestions to improve this.

### DIFF
--- a/doc/whatsnew/fragments/7931.bugfix
+++ b/doc/whatsnew/fragments/7931.bugfix
@@ -1,3 +1,3 @@
-Fix exit code when argparse raises SystemExit due to bad arguments.
+When pylint exit due to bad arguments being provided the exit code will now be the expected ``32``.
 
 Refs #7931
--- a/doc/whatsnew/fragments/7931.bugfix
+++ b/doc/whatsnew/fragments/7931.bugfix
@@ -1,0 +1,3 @@
+Fix exit code when argparse raises SystemExit due to bad arguments.
+
+Refs #7931
--- a/pylint/config/arguments_manager.py
+++ b/pylint/config/arguments_manager.py
@@ -256,9 +256,12 @@ class _ArgumentsManager:
 
     def _parse_configuration_file(self, arguments: list[str]) -> None:
         """Parse the arguments found in a configuration file into the namespace."""
-        self.config, parsed_args = self._arg_parser.parse_known_args(
-            arguments, self.config
-        )
+        try:
+            self.config, parsed_args = self._arg_parser.parse_known_args(
+                arguments, self.config
+            )
+        except SystemExit:
+            sys.exit(32)
         unrecognized_options: list[str] = []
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas (commented):
Nice catch, we indeed missed that argparse could exit with its own exit code and the inconsistency with pylint's error codes. I think we should use -1 because it's the only way our bit encoded error code can be extended to 64 if required. Thoughts @DanielNoord ?

dmrlawson:
> Nice catch, we indeed missed that argparse could exit with its own exit code and the inconsistency with pylint's error codes. I think we should use -1 because it's the only way our bit encoded error code can be extended to 64 if required. Thoughts @DanielNoord ?

`sys.exit(32)` appears throughout the code. The "usage error" exit code could be factored out to be a constant and then changed but it's a bit more than I was intending here.

```
$ grep -r 'sys.exit(32)'
pylint/pyreverse/utils.py:        sys.exit(32)
pylint/pyreverse/utils.py:        sys.exit(32)
pylint/lint/run.py:            sys.exit(32)
pylint/lint/run.py:            sys.exit(32)
pylint/lint/run.py:            sys.exit(32)
pylint/lint/run.py:                sys.exit(32)
pylint/config/arguments_manager.py:            sys.exit(32)
pylint/config/config_initialization.py:        sys.exit(32)
pylint/config/config_initialization.py:            sys.exit(32)
```

Pierre-Sassoulas:
You're right let's keep it as 32, I wasn't up to date on this aspect of the code sorry.

DanielNoord (dismissed):

Pierre-Sassoulas (changes requested):
Could you fix the tests (there's some other test that is somewhat fragile you'll need to add the new file to it) and add a changelog entry please ?

doc/whatsnew/fragments/7931.bugfix
```diff
@@ -0,0 +1,3 @@
+Fix exit code when argparse raises SystemExit due to bad arguments.
```
Pierre-Sassoulas, line 1:
```suggestion
When pylint exit due to bad arguments being provided the exit code will now be the expected ``32``.
```

Pierre-Sassoulas (dismissed):
Great job 👌

Pierre-Sassoulas (approved):

Issue #7897, opened 2022-12-04
Better message documentation: add various templates to be able to handle all messages


Pierre-Sassoulas:
### Current problem

In #5953, some messages do not follow the generic template with good code and bad code or can't be triggered easily. Among them:
#### internal errors
- [x] astroid-error 

#### configuration errors
- [x] b/bad-plugin-value 

#### Code that require multiple examples 

There already are multiple example mixed in the existing files.
- [ ] missing-final-newline
- [ ] A <= symbol < D #9006 
- [ ] D < symbol < Z TODO
 
#### Messages that depend on multiple file or dependency to trigger
- [x] import-error
- [x] cyclic-import #8322 
- [x] import-self
- [x] d/duplicate-code
- [x] n/non-ascii-file-name #8340 
#### other
- [x] too-many-lines (Hard and meaningless to do parity behavior between bad/good, creating a new file is often the solution)
- [x] b/bad-file-encoding (Need to be handled by the generating part)
- [ ] c/c-extension-no-member (This one is going to be hard to trigger, not for beginner)
- [x] r/relative-beyond-top-level

### Desired solution

We need to create templates for those message so we can generate the proper skeleton for them.

- For internal errors, probably only a message stating "You should never encounter this one, please open an issue if you do"
- For configuration errors a bad.toml and good.toml with some tests close to what we do for good.py / bad.py (maybe running on a dummy empty file)
- For code that require multiple examples, handle multiple example more gracefully than right now (we just add a longer example with multiple part)

### Additional context

The Sphinx extension that generates these pages needs to be modified see https://github.com/PyCQA/pylint/issues/5953#issuecomment-1336389098

ollie-iterators:
I think it would be beneficial if the errors that need to have a different template with good.py and bad.py were moved into this or a separate PR so that people could focus on the issues that can currently have documentation added.

There could also be an explanation of what needs to be added if different things need to be added for different errors.

ollie-iterators:
An alternative suggestion is to add notes to each issue that needs to have a different template on #5953.

ollie-iterators:
Do you still think that unexpected-line-ending-format should be included in this list? Because it has not had a commentn for that

Pierre-Sassoulas:
It's done in #8255

Pierre-Sassoulas:
It's possible to do some messages without a new template by removing good.py and bad.py and simply adding a details.rst and related.rst.

ollie-iterators:
[...]

PR #7936, opened 2022-12-13
Add documentation for emitted messages

zenlyj:
## Type of Changes

## Description

Refs #5953

Added documentation for the following messages:

1. import-error
2. import-self
3. disallowed-name
4. else-if-used

References:

1. https://vald-phoenix.github.io/pylint-errors/plerr/errors/imports/E0401
2. https://vald-phoenix.github.io/pylint-errors/plerr/errors/imports/W0406
3. https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/C0102
4. https://vald-phoenix.github.io/pylint-errors/plerr/errors/else-if-used/R5501


doc/data/messages/i/import-error/bad.py
```diff
@@ -0,0 +1,2 @@
+# you didn't install it via pip, setup.py, wheel
```
Pierre-Sassoulas, line 1:
```suggestion
```
You can add comment in details.rst. sorry I can't suggest it as the file was deleted.

doc/data/messages/i/import-error/good.py
```diff
@@ -1 +1,2 @@
-# This is a placeholder for correct code for this message.
+# package installed in your venv, globally, user space
+import third_party
```
Pierre-Sassoulas, line 2:
```suggestion
import sphinx
```
Trying to be clever by using a package we have in our env, but realistically this is not going to be something we can demonstrate by using our doc framework. If this doesn't work could you remove import-error from this merge request ?

doc/data/messages/i/import-self/bad.py
```diff
@@ -0,0 +1,4 @@
+cat > foo.py <<EOF
```
Pierre-Sassoulas, line 1: This one is also tricky, maybe it would work by importing ``bad`` ?

Pierre-Sassoulas (changes requested):
Thank you for your contribution ! Some of the message here are not the easiest to document, they are probably going to end up in #7897

doc/data/messages/i/import-self/bad.py
zenlyj: Thanks for the suggestion! It seems that the test error is resolved by importing `bad`, since the file is named `bad.py`.

doc/data/messages/i/import-error/bad.py
zenlyj: Shifted the comments to details.rst.

doc/data/messages/d/disallowed-name/good.py
```diff
@@ -1 +1,2 @@
-# This is a placeholder for correct code for this message.
+def get_something():
```
Pierre-Sassoulas, line 1:
```suggestion
def something_meaningful():
```

Pierre-Sassoulas (changes requested):
Some small suggestions.

zenlyj:
Thanks for the suggestions @Pierre-Sassoulas, I have updated the PR to address the requested changes.
May I also check, for disallowed-names, would it be clearer if we included the list of blacklisted/disallowed names in the additional details section?

doc/data/messages/e/else-if-used/bad.py
```diff
@@ -0,0 +1,7 @@
+if input():
+    pass
+else:
+    if len(input()) >= 10:  # [else-if-used]
+        pass
+    else:
+        pass
```
Pierre-Sassoulas, line 7: You're going to need a rc file to be able to load the extension here.

doc/data/messages/i/import-error/good.py
```diff
@@ -1 +1,3 @@
-# This is a placeholder for correct code for this message.
+from pathlib import Path
+
+import sphinx
```
Pierre-Sassoulas, line 3:
```suggestion
# import third_party
```
I don't think we can demonstrate this one easily. Might be better to just keep the typo only. If a beginner encounter ``# import third_party `` it will be confusing for them.

Pierre-Sassoulas:
> May I also check, for disallowed-names, would it be clearer if we included the list of blacklisted/disallowed names in the additional details section?

There's an option to change this, using a configuration file would make this clearer imo.

doc/data/messages/e/else-if-used/bad.py
zenlyj: Apologies for my ignorance, how should I go about using an .rc file in the workflow?

doc/data/messages/i/import-error/good.py
zenlyj: Makes sense to me, updated to keep the typo only.

zenlyj:
> > May I also check, for disallowed-names, would it be clearer if we included the list of blacklisted/disallowed names in the additional details section?
> 
> There's an option to change this, using a configuration file would make this clearer imo.

That is to say, show a rc file with the following content?
```
[BASIC]
# Bad variable names which should always be refused, separated by a comma
bad-names=foo,bar,baz,toto,tutu,tata
# Bad variable names regexes, separated by a comma. If names match any regex, they will always be refused
bad-names-rgxs=
```

Pierre-Sassoulas:
> That is to say, show a rc file with the following content?

Yeah we could do that. But now that I thougth about it this is something that we already have with : https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#bad-names. 

If we want to make this better then we have to make it better for all the messages generically. If we do it one by one manually it's going to be a pain to upgrade.

Pierre-Sassoulas (approved):
👌

Pierre-Sassoulas:
Great first contribution @zenlyj, the message that still had to be documented weren't easy. Congratulation on becoming a pylint contributor :)

zenlyj:
> Great first contribution @zenlyj, the message that still had to be documented weren't easy. Congratulation on becoming a pylint contributor :)

Thank you @Pierre-Sassoulas for guiding a newbie like myself. This is my first contribution to an open-source project, glad to help out!

Issue #3110, opened 2019-09-19
Unassigned instantiated classes should trigger a pointless-statement or a new message to be determined


vapier:
### Current status

pylint will issue `pointless-statement` when a class (e.g. Exception) is used.  this is great:
```sh
$ cat test.py
Exception
$ pylint3 test.py
test.py:1:0: W0104: Statement seems to have no effect (pointless-statement)
```

but if the exception is instantiated, it's not shown:
```sh
$ cat test.py
Exception()
$ pylint3 test.py
<nothing>
```

### Proposal

while instantiating random classes can have side-effects, exceptions rarely (if ever) have such behavior.  it's so uncommon that, imo, it should trigger a warning by default.  we ran into issues where people forgot to `raise` the exception (because the naming made it sound like it would throw an error), so code that was supposed to fail ended up not.

so how about:
```sh
$ cat test.py
Exception()
$ pylint3 test.py
test.py:1:0: W0104: Instantiated object seems to have no effect (pointless-exception-object)
```

feel free to bikeshed the exact naming/phrasing.

### Version info

```sh
$ pylint3 --version
pylint3 2.2.2
astroid 2.1.0
Python 3.6.8 (default, Jan  3 2019, 03:42:36) 
[GCC 8.2.0]
```

vapier:
maybe something like:
```python
    def visit_expr(self, node):
        if not isinstance(node.value, astroid.Call):
            return
        func = node.value.func
        try:
            cls = next(func.infer())
        except astroid.InferenceError:
            return
        if not isinstance(cls, astroid.ClassDef):
            return
        if any(x for x in cls.ancestors() if x.name == 'BaseException'):
            self.add_message('...', node=node, line=node.fromlineno)
```

PCManticore:
This sounds good to me, but I think we should do it for the general case not just for exceptions. Instantiating a class and not doing anything with it sounds like a smelling to me (note that might have a purpose, e.g. testing that the instantiation raises an exception)

PCManticore:
@vapier From my understanding, this new check will handle the following case:
```
Class()
```

while ```Class().run()``` and `c = Class(); c.run()` will not be caught.

The former looks more of an application bug than a code smell, as that instantiation might miss an additional method call or a variable assignment.

Pierre-Sassoulas:
This is still relevant as 
```python
from pathlib import Path

Exception()
Path()
```
does not raise any message.

jayaddison:
[...]

PR #7939, opened 2022-12-14
Add new check "pointless-exception-statement"

jayaddison:
It's possible to write Python statements that instantiate exception objects without assigning, raising or returning the exception object.

As described in #3110, this can happen when a `raise` statement is accidentally omitted.

For example:

```py
def pattern_number(pattern):
    if pattern.startswith("A"):
        return 1
    elif pattern.startswith("B"):
        return 2
    else:
        # it looks like the intent here was to raise this exception; instead the method will return None
        ValueError(f"expected '{pattern}' to begin with 'A' or 'B'")
```

This changeset introduces a new check, named `pointless-exception-statement`, that can warn about unused-instantiated-exceptions like this.

## Type of Changes

## Description

Refs #3110


doc/data/messages/p/pointless-exception-statement/bad.py
```diff
@@ -0,0 +1,2 @@
+if next_token is None:
+    Exception("this exception is a statement")  # [pointless-exception-statement]
```
Pierre-Sassoulas, line 2:
```suggestion
Exception("This exception is a statement.")  # [pointless-exception-statement]
```

doc/data/messages/p/pointless-exception-statement/good.py
```diff
@@ -0,0 +1,2 @@
+if 2 + 2 == 5:
+    raise Exception("this shouldn't happen")
```
Pierre-Sassoulas, line 2:
```suggestion
raise Exception("This will raise.")
```

Pierre-Sassoulas (commented):
Look pretty good already, I'm surprised we do not have this already.

pylint/checkers/base/basic_checker.py
```diff
@@ -444,6 +451,12 @@ def visit_expr(self, node: nodes.Expr) -> None:
             self.add_message("pointless-string-statement", node=node)
             return
 
+        # Warn W0133 for exceptions that are used as statements
+        if isinstance(expr, nodes.Call):
+            inferred = utils.safe_infer(expr)
+            if isinstance(inferred, objects.ExceptionInstance):
```
jayaddison, line 457: One question I had is whether checking against the `objects.ExceptionInstance` type is safe, in terms of API stability - I didn't find it listed in the documentation (and could raise a documentation update as a separate issue and/or pull request, if relevant).

jayaddison: Sorry: I should clarify that I didn't find it mentioned in the `astroid` documentation - and that's where I'd consider opening a corresponding issue/pull request.

Pierre-Sassoulas: We're using the ``qname`` attribute generally, you can check how it's one in checker for exceptions too broad raised/caught.

pylint/checkers/base/basic_checker.py
```diff
@@ -444,6 +451,12 @@ def visit_expr(self, node: nodes.Expr) -> None:
...
+                self.add_message("pointless-exception-statement", node=node)
```
Pierre-Sassoulas, line 458:
```suggestion
                self.add_message("pointless-exception-statement", node=node, confidence=INFERENCE)
```
This will require to import it and to regenerate the functional output.

pylint/checkers/base/basic_checker.py
```diff
@@ -444,6 +451,14 @@ def visit_expr(self, node: nodes.Expr) -> None:
             self.add_message("pointless-string-statement", node=node)
             return
 
+        # Warn W0133 for exceptions that are used as statements
+        if isinstance(expr, nodes.Call):
+            inferred = utils.safe_infer(expr)
```
jayaddison, line 456: This `utils.safe_infer` invocation seems to retrieve the first item from a generator.

During the `string_formatting` functional test, [these inference tests](https://github.com/PyCQA/pylint/blob/c61bc76806f7b389db4b7b484174dbc286fa2c11/tests/functional/s/string/string_formatting.py#L47-L48) appear to fail as a result of that.  Instead of `0.missing` accessing the attribute with the name 'missing' from the [`ReturnYes` class](https://github.com/PyCQA/pylint/blob/c61bc76806f7b389db4b7b484174dbc286fa2c11/tests/functional/s/string/string_formatting.py#L29-L31), there is an to access the same attribute from the integer with value `2` (from the previous test fixture perhaps?)

The underlying exception is a result of a [`StopIteration` exception handled in `astroid` here](https://github.com/PyCQA/astroid/blob/8bdec591f228e7db6a0be66b6ca814227ff50001/astroid/decorators.py#L146-L148).

jayaddison: Hm, I could be mistaken about the source of the `StopIteration`.  I forget exactly what brought me to that part of the code.

jayaddison: Inference tips and/or caching are something to do with the problem here, I think.

* Removing the `str.format` inference tip from https://github.com/PyCQA/astroid/blob/8bdec591f228e7db6a0be66b6ca814227ff50001/astroid/brain/brain_builtin_inference.py#L1002-L1004 allows the `string_formatting` test to succeed -- but doing so causes other functional tests (such as `alternative_union_syntax`) begin to fail.

* Alternatively, adding `AttributeError` to the [list of exceptions handled in `astroid` here](https://github.com/PyCQA/astroid/blob/8bdec591f228e7db6a0be66b6ca814227ff50001/astroid/brain/brain_builtin_inference.py#L957) also works.

Either way: I don't understand why performing an additional inference on `Call` objects affects this.

jayaddison:
Note to self:

The remaining test failures may be handled by PyCQA/astroid#1903 (allocated to upcoming `astroid` version 2.12.14).

If/when those changes are merged, then:

- [ ] Update the [minimum required `astroid` version in `pyproject.toml`](https://github.com/jayaddison/pylint/blob/1cf068a3579c8d4f451b484c38e92e339921a8a9/pyproject.toml#L41)
- [ ] Pin the `astroid` version in [`requirements_test_min.txt`](https://github.com/jayaddison/pylint/blob/1cf068a3579c8d4f451b484c38e92e339921a8a9/requirements_test_min.txt#L3)

pylint/checkers/base/basic_checker.py
DanielNoord: This issues has been resolved, right?

pylint/checkers/base/basic_checker.py
DanielNoord: I think `ExceptionInstance` is fine here. We're the maintainers of `astroid` as well. So if we ever break the API we are also in charge of fixing it. I don't think this is an issue here.

DanielNoord (approved):
Love it! Simple check and good tests. Thanks for this!

Pierre-Sassoulas:
Great check, let's fast track it in 2.16 :)

pylint/checkers/base/basic_checker.py
jayaddison: Yep, this was resolved by PyCQA/astroid#1903.

jayaddison:
Thanks @Pierre-Sassoulas @DanielNoord!

jayaddison:
Confirmed using [v2.16.0b0](https://github.com/PyCQA/pylint/releases/tag/v2.16.0b0) of `pylint` that a `pointless-exception-statement` is emitted as expected on the statement that originally caused a warning.

jayaddison:
d'oh - `s/caused a warning/should have caused a warning`

pylint/checkers/base/basic_checker.py
cdce8p, line 456: Inference can be expensive, especially if the lru cache has reached is maxsize and old entries are overwritten.

We should not call for `safe_infer` for every `Call` node. Instead I would recommend to use a heuristic and filter likely candidates first.
* Almost all builtin error names contain either `Exception`, `Error`, or `Warning`. That likely applies to custom exceptions as well.
* This could  be further improved with a check against other predefined names. For the stdlib, those would probably be `KeyboardInterrupt`, `StopAsyncIteration`, `StopIteration`.
* Lastly a new config option could be added to allow the user to specify additional names (maybe even regex).

The goal IMO should be to capture 95% of the likely errors without the huge performance impact.

cdce8p (commented):

```
before: 7:59
```

pylint/checkers/base/basic_checker.py
Pierre-Sassoulas: Outch thank you for the check @cdce8p this seems like a show stopper for 2.16.0 release and a great advertisement for better benchmarking in CI.

jayaddison: Thank you, @cdce8p!

I can take a look at this within the next 24h or so.  The name-based heuristics sound straightforward to implement - I have less confidence about the config option though, since I don't know much about how those work in `pylint` yet.

Is it fairly straightforward to re-run the benchmarking when a changeset is ready?

None of this should prevent either reverting the change, or someone else picking up the performance improvements if I don't get around to them soon.

cdce8p: > I can take a look at this within the next 24h or so. The name-based heuristics sound straightforward to implement - I have less confidence about the config option though, since I don't know much about how those work in `pylint` yet.

I would suggest to start with the name check first and implement that. It should get you most of the way.
Adding a new config option isn't too difficult though. As example, you can take a look at this
https://github.com/PyCQA/pylint/blob/b87cf1631f9fcef278895382481cc0169cb2301e/pylint/checkers/design_analysis.py#L389-L398
https://github.com/PyCQA/pylint/blob/b87cf1631f9fcef278895382481cc0169cb2301e/pylint/checkers/design_analysis.py#L413-L415

Feel free to ask if you have questions.

> Is it fairly straightforward to re-run the benchmarking when a changeset is ready?

I've set up a semi-automatic pipeline for that in my fork. Probably easiest if you ping me for it.
If you really want to test it on your own, you could fork my home-assistant fork and take a look at https://github.com/cdce8p/ha-core/pull/11. Although I'm fairly confident that any changes here should be enough to remedy the issue and it's probably much easier for me to run the test again.

jayaddison:
Confirming that this is working as expected with the v2.16.0 release (it catches a `ValueError` instantiation that is not raised).  Thanks for all the help @DanielNoord @Pierre-Sassoulas @cdce8p.

jayaddison:
I'm seeing some false positives with this rule, often similar to [this one](https://github.com/scrapy/scrapy/blob/88eec7a417c55668ebd0b8242e59bc99721d2816/tests/test_pipeline_images.py#L625), where instantiation of a class without assignment is being flagged by this check.

Given that detection by this check should be rare (especially after the filtering applied subsequently in #8075) I wonder whether it'd make sense to add an additional inference-based 'appears to be subclass of `Exception`' condition.

jayaddison:
Todo: confirm whether `is_subtype_of("Exception")` could help for that.

jayaddison:
> I'm seeing some false positives with this rule, often similar to [this one](https://github.com/scrapy/scrapy/blob/88eec7a417c55668ebd0b8242e59bc99721d2816/tests/test_pipeline_images.py#L625), where instantiation of a class without assignment is being flagged by this check.

Ok, after further inspection: that _is_ in fact instantiation of a class derived from `Exception`, so arguably it _is not_ a false positive.

> Given that detection by this check should be rare (especially after the filtering applied subsequently in https://github.com/PyCQA/pylint/pull/8075) I wonder whether it'd make sense to add an additional inference-based 'appears to be subclass of Exception' condition.

The [`isinstance(inferred, objects.ExceptionInstance)`](https://github.com/PyCQA/pylint/blob/08bac36eeb5006c7eb2e574049566e232bdbf5ed/pylint/checkers/base/basic_checker.py#L467) condition should already achieve that intended effect, more-or-less.

PR #7945, opened 2022-12-15
Fix the multiple-statements issue on pypy 3.8

Pierre-Sassoulas:
## Type of Changes

## Description

Problem appeared recently hard to know exactly why because we were using a cached environment for some time before.


tests/functional/a/alternative/alternative_union_syntax_error.py
```diff
@@ -1,10 +1,18 @@
-"""Test PEP 604 - Alternative Union syntax
-without postponed evaluation of annotations.
+"""Test PEP 604 - Alternative Union syntax with postponed evaluation of
+annotations enabled.
 
-For Python 3.7 - 3.9: Everything should fail.
+For Python 3.7 - 3.9: Most things should work.
```
Pierre-Sassoulas, line 4:
```suggestion
For Python 3.7 - 3.9: Everything should fail.
```

DanielNoord (commented):
I don't really understand this PR? Are all these formatting changes necessary for PyPy to pass?

Pierre-Sassoulas:
I formatted the disable because as I added 3 new lines I had to regenerate the output anyway, might as well clean up a little.

DanielNoord:
But there are also multiple file moves right? What's is the actual fix the unblock CI?

Pierre-Sassoulas:
[814ae24](https://github.com/PyCQA/pylint/pull/7945/commits/814ae244a991f7c735e775ed9edc462eeb2c0166) (There might be little bit of fix in the moved files from the two previous commit which is why I did not separate them).

DanielNoord (approved):

PR #7955, opened 2022-12-17
Warn on incomplete concrete classes that inherit from abstract classes

### LINKED ISSUE #7950: Subclasses of abstract class that do not inherit abc.ABC are considered abstract
### Bug description

Related to https://github.com/PyCQA/pylint/issues/3098. Subclasses of an abstract parent class which do not inherit directly from `abc.ABC` are considered abstract, and only throw an error during run time when trying to instantiate it.

For example, take this code below. PyCharm correctly determines through a IDE warning that Sub is not abstract, and says `Class Sub must implement all abstract methods`, but pylint will ignore this class, as its ancestors inherit from `abc.ABC`. Do we not get the best of both worlds (allowing abstract subclasses of abstract parent classes, ie. `AbsSub`, while catching non-abstract subclasses of abstract parent classes, ie. `Sub`) if we check the bases, instead of the ancestors where that PR to fix the issue above does it?

```python
import abc
import unittest

class Abstract(unittest.TestCase, abc.ABC):
  @abc.abstractmethod
  def test_something(self):
    pass

  @abc.abstractmethod
  def test_another_thing(self):
    pass


class AbsSub(Abstract, abc.ABC):
  def test_another_thing(self):
    return 1


class Sub(AbsSub):
  pass
```

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
`NOTHING`
```

### Expected behavior

`test.py:19:0: W0223: Method 'test_something' is abstract in class 'Abstract' but is not overridden (abstract-method)` (referring to `Sub`)

### Pylint version

```shell
2.5.0
```

### OS / Environment

Ubuntu 20.04

### Additional dependencies

_No response_

sshane:
Fixes: https://github.com/PyCQA/pylint/issues/7950

TODO:
- [x] Update test refs
- [x] Make it detect the root abc module correctly


pylint/checkers/utils.py
```diff
@@ -1172,8 +1172,12 @@ def class_is_abstract(node: nodes.ClassDef) -> bool:
         if meta.name == "ABCMeta" and meta.root().name in ABC_MODULES:
             return True
 
-    for ancestor in node.ancestors():
-        if ancestor.name == "ABC" and ancestor.root().name in ABC_MODULES:
+    # As well as directly inherited ABC class
+    for base in node.bases:
+        base_class_name = (
+            base.attrname if isinstance(base, astroid.Attribute) else base.as_string()
+        )
```
sshane, line 1179: I could not find any way to detect the module directly if you subclass ABC like this:

```
from abc import ABC
class MyTestCase(ABC):
  pass
````

Let me know if anyone has any ideas

Comment (DanielNoord) [reply pylint/checkers/utils.py:1179]: This is what we use `inference` for. You can look for calls to `safe_infer` which should hopefully set you on the right track!

**Comment by sshane** [comment]
Updated the test refs, but I'm not sure they're right. I'm getting lots of diffs for a lot of tests. Running pylint built from main on Windows 11.

**Comment by sshane** [comment]
Should be good now, let me know if there's any issues @Pierre-Sassoulas! I had to add a print into the tester to get the results :/

File: tests/functional/a/abstract/abstract_method.py [hunk 2]
```diff
@@ -44,7 +44,16 @@ class AbstractD(AbstractB, metaclass=abc.ABCMeta):
     """
 
 
-class Concrete(Abstract): # [abstract-method]
+class ConcreteA(AbstractC):
+    """
+    Concrete class.
+
+    Should trigger a warning since this class does
```
DanielNoord, line 51: I don't really understand this docstring. Shouldn't this then raise `abstract-method`?

pylint/checkers/utils.py
```diff
@@ -1172,9 +1172,14 @@ def class_is_abstract(node: nodes.ClassDef) -> bool:
...
-    for ancestor in node.ancestors():
-        if ancestor.name == "ABC" and ancestor.root().name in ABC_MODULES:
-            # abc.ABC inheritance
+    # As well as direct abc.ABC inheritance
+    for base in node.bases:
+        inferred_base = safe_infer(base)
+        if (
+            isinstance(inferred_base, nodes.ClassDef)
+            and inferred_base.root().name in ABC_MODULES
+            and inferred_base.name == "ABC"
```
DanielNoord, line 1181:
```suggestion
            and inferred_base.root().name in ABC_MODULES
            and inferred_base.name == "ABC"
```

I think this works fine. Note that you could also call `.qname()` for a similar result.

DanielNoord (changes requested):

tests/functional/a/abstract/abstract_method.py
```diff
@@ -44,7 +44,16 @@ class AbstractD(AbstractB, metaclass=abc.ABCMeta):
     """
 
 
-class Concrete(Abstract): # [abstract-method]
+class ConcreteA(AbstractC):
```
sshane, line 47:
```suggestion
class ConcreteA(AbstractC): # [abstract-method] 
```

pylint/checkers/utils.py
sshane: Yes, but do we not care about checking all the modules in `ABC_MODULES`?

DanielNoord:
> @DanielNoord is there anything else needing to be done?

The CI fails so we would first need to fix that before this can be merged.

sshane:
For whatever reason, both Python 3.8 and 3.10 do not work to update the abstract method file. I'm running `python test_functional.py --update-functional-tests`, but it just fails and doesn't update any files.

```
-e git+https://github.com/sshane/pylint@c8564bb99a393621341851acaa3ed042b3c53c3d#egg=pylint
astroid==2.12.14
Python 3.10.0
Ubuntu 20.04
```

```python
self = <pylint.testutils.functional.lint_module_output_update.LintModuleOutputUpdate object at 0x7fa1e3c6d690>

    def runTest(self) -> None:
>       self._runTest()
E       AssertionError: Wrong results for file "abstract_method":
E       
E         47: abstract-method

../pylint/testutils/lint_module_test.py:145: AssertionError
=========================================================================================================================================== short test summary info ===========================================================================================================================================
FAILED test_functional.py::test_functional[abstract_method] - AssertionError: Wrong results for file "abstract_method":
============================================================================================================================ 1 failed, 779 passed, 25 skipped in 62.76s (0:01:02) =============================================================================================================================
```

DanielNoord:
Could it be that the test simply fails? If haven't checked that myself locally

sshane:
Do you have any docs for what the comments do? For example, I'm not sure if `# [abstract-method]` means we're asserting an abstract method error from pylint, or we're catching an error to not fail the tests.

DanielNoord:
> Do you have any docs for what the comments do? For example, I'm not sure if `# [abstract-method]` means we're asserting an abstract method error from pylint, or we're catching an error to not fail the tests.
> 
> Are the functional tests a sort of blend of replay with the file outputs and unit tests?

https://pylint.readthedocs.io/en/latest/development_guide/contributor_guide/tests/writing_test.html#functional-tests

This (attempts to) describe the functional test framework. Basically it tries to lint the file and sees where we expect messages to be raised and compares that against the actual output of the linting.

DanielNoord:
@sshane This is quite close to being done. Do you have intention of working on it? Or shall we close it?

sshane:
I'm having the same problem I did when I last worked on it, I verified my ` # [abstract-method]` markers are accurate according to running pylint on the file, but still get some test failures:

```
    def runTest(self) -> None:
>       self._runTest()
E       AssertionError: Wrong message(s) raised for "abstract_method.py":
E       
E         47: abstract-method
```

Running `python tests/test_functional.py --update-functional-output -k "test_functional[abstract_method]"` does not update `abstract_method.txt`

sshane:
Narrowed it down to https://github.com/pylint-dev/pylint/issues/9688. I don't have any more time to spend on this for the time being, feel free to get it across the line if it's simple for you, else you can close.

sshane:
I was expecting each line's warnings to be compared to the annotations, not one by one! A more clear error message would have been nice :slightly_smiling_face:

sshane:
@DanielNoord I added the count which would have surfaced this for me much quicker, let me know if it's something you want and I can split it out

DanielNoord (commented):
Only needs a news entry using `towncrier`. I'm fine with merging without full coverage.

Pierre-Sassoulas:
As incredible as it seems, every new primer warning I checked seem to be an actual issue.

DanielNoord (approved):

PR #7974, opened 2022-12-23
update while-used doc

### LINKED ISSUE #7748: `while_used` checker shouldn't report `while True`
### Current problem

Currently, the `while_used` checker reports `while-used` for a `while True` loop:
```py
from time import sleep

while True:
    sleep(1)
```
```
main.py:3:0: W0149: Used `while` loop (while-used)
```
The documentation says:
> Unbounded while loops can often be rewritten as bounded for loops.

Which is untrue for `while True` loops.
It is possible with an infinite generator, but `while True` makes much more sense.

### Desired solution

The `while_used` checker should not report an error for a `while True` loop.

### Additional context

`pylint --version` output:
```
pytpylint 2.15.5
astroid 2.12.12
Python 3.8.12
[GCC 10.3.0]
```

clavedeluna:
## Type of Changes

## Description

Closes #7748

### DIFF
--- a/pylint/extensions/while_used.py
+++ b/pylint/extensions/while_used.py
@@ -24,7 +24,8 @@ class WhileChecker(BaseChecker):
         "W0149": (
             "Used `while` loop",
             "while-used",
-            "Unbounded `while` loops can often be rewritten as bounded `for` loops.",
+            "Unbounded `while` loops can often be rewritten as bounded `for` loops. "
+            "Exceptions can be made for cases such as event loops, listeners, etc.",
         )
     }
 


clavedeluna:
I know I need to run `tox -e docs` but I'm getting errors for it 🤔

pylint/extensions/while_used.py
```diff
@@ -24,7 +24,8 @@ class WhileChecker(BaseChecker):
         "W0149": (
             "Used `while` loop",
             "while-used",
-            "Unbounded `while` loops can often be rewritten as bounded `for` loops.",
+            "Unbounded `while` loops can often be rewritten as bounded `for` loops. ",
+            "Exceptions can be made for cases such as event loops, listeners, etc.",
```
mbyrnepr2, line 28: I think the added tuple item is an issue here. Mypy is highlighting it in the precommit output.

mbyrnepr2:
You've run into the rate limit issue @clavedeluna. Can you try force pushing in order to trigger the CI?

mbyrnepr2 (approved):

PR #7975, opened 2022-12-23
Fix `use-sequence-for-iteration` when unpacking a set with `*`

### LINKED ISSUE #5788: Possible false positive with `use-sequence-for-iteration` on in-place sets with dynamic content?
### Bug description

I'm getting a warning for `use-sequence-for-iteration` in my code, but I believe I have a valid reason to use it that should possibly be considered as an exception for the rule.

In my code I iterate over in-place sets with some in-place content, and some dynamic content to ensure that I iterate over the in-place content if it doesn't already exist in the dynamic content, while also ensuring I never iterate over the same thing twice.

Example:

```python
# test.py

def print_strings_and_hello(may_contain_hello):
    """
    Prints a set of strings without duplicates, while also ensuring 'hello' always gets
    printed
    """
    for string in {*may_contain_hello, "hello"}:
        print(string)


print_strings_and_hello(["hello", "goodbye"])  # prints both hello and goodbye
print_strings_and_hello(["goodbye"])  # prints both hello and goodbye

```

Do let me know if you think this is a bad practice!

I can't find much discussion about this warning online (which makes sense considering it was added recently) so I would love to engage and leave a paper trail for the next user who gets this warning for a similar usage.

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:8: [C0208(use-sequence-for-iteration), print_strings_and_hello] Use a sequence type when iterating over values

Your code has been rated at 8.00/10 (previous run: 8.00/10, +0.00)
```

### Expected behavior

No warning for the line

### Pylint version

```shell
pylint 2.12.2
astroid 2.9.0
Python 3.9.2 (default, Apr 15 2021, 10:01:45) 
```

clavedeluna:
## Type of Changes

## Description

Closes #5788

### DIFF
--- a/doc/whatsnew/fragments/5788.false_positive
+++ b/doc/whatsnew/fragments/5788.false_positive
@@ -1,0 +1,3 @@
+Fix ``use-sequence-for-iteration`` when unpacking a set with ``*``.
+
+Closes #5788
--- a/pylint/checkers/refactoring/recommendation_checker.py
+++ b/pylint/checkers/refactoring/recommendation_checker.py
@@ -338,7 +338,7 @@ class RecommendationChecker(checkers.BaseChecker):
         Sets using `*` are not considered in-place.
         """
         if isinstance(node.iter, nodes.Set) and not any(
-            isinstance(x, nodes.Starred) for x in node.iter.elts
+            utils.has_starred_node_recursive(node)
         ):
             self.add_message(
                 "use-sequence-for-iteration", node=node.iter, confidence=HIGH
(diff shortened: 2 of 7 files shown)


Pierre-Sassoulas (approved):
LGTM !

jacobtylerwalls (changes requested):
Thanks! Would you mind doing a recursive search? We should have some utils for this.

This fails ⬇️ 

```python
def deduplicate_nested_sets(input1, input2, input3, input4):
    for thing in {{*input1, *input2}, {*input3, *input4}}:
        print(thing)
```

jacobtylerwalls (approved):
👍🏻

PR #7987, opened 2022-12-26
class attrs should not emit assigning-non-slot msg

### LINKED ISSUE #6001: false positive "assigning-non-slot" error for inherited descriptor when slots are used
### Bug description

Minimized reproducer:
```python
"""Repro false positive bug in pylint."""


class MyDescriptor:
    """Basic descriptor."""

    def __get__(self, instance, owner):
        return 42

    def __set__(self, instance, value):
        pass


class Base:
    """Base class with slots."""

    __slots__ = ()

    attr2 = MyDescriptor()


class Repro(Base):
    """Reproduce false positive error using this class."""

    __slots__ = ()


repro = Repro()
repro.attr2 = 'false positive "assigning-non-slot" error'  # false positive error here
```

A real-world (non-minimized) example where pylint produces this false positive error can be found on this line: https://github.com/jab/bidict/blob/caf703e959ed4471bc391a7794411864c1d6ab9d/bidict/_orderedbase.py#L101

### Configuration

_No response_

### Command used

```shell
pylint repro.py
```

### Pylint output

```shell
************* Module repro
repro.py:32:0: E0237: Assigning to attribute 'attr2' not defined in class slots (assigning-non-slot)
```

### Expected behavior

No error should be flagged here.

### Pylint version

```shell
pylint 2.13.2
astroid 2.11.2
Python 3.10.2 (main, Feb  2 2022, 07:36:01) [Clang 12.0.0 (clang-1200.0.32.29)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description

Pylint was raising  ``assigning-non-slot`` when a class attribute was re-assigned, which I believe is incorrect given that `__slots__` is meant for instance attrs.

Closes #6001

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1629,7 +1629,9 @@ a metaclass class method.",
                     # Properties circumvent the slots mechanism,
                     # so we should not emit a warning for them.
                     return
-                if self._is_class_attribute(node.attrname, klass):
+                if node.attrname != "__class__" and utils.is_class_attr(
+                    node.attrname, klass
+                ):
                     return
                 if node.attrname in klass.locals:
                     for local_name in klass.locals.get(node.attrname):
@@ -1786,7 +1788,7 @@ a metaclass class method.",
                 if (
                     self._is_classmethod(node.frame(future=True))
                     and self._is_inferred_instance(node.expr, klass)
-                    and self._is_class_attribute(attrname, klass)
+                    and self._is_class_or_instance_attribute(attrname, klass)
                 ):
                     return
 
@@ -1833,7 +1835,7 @@ a metaclass class method.",
         return inferred._proxied is klass
 
     @staticmethod
-    def _is_class_attribute(name: str, klass: nodes.ClassDef) -> bool:
+    def _is_class_or_instance_attribute(name: str, klass: nodes.ClassDef) -> bool:
         """Check if the given attribute *name* is a class or instance member of the
         given *klass*.
 
@@ -1841,11 +1843,8 @@ a metaclass class method.",
         ``False`` otherwise.
         """
 
-        try:
-            klass.getattr(name)
+        if utils.is_class_attr(name, klass):
             return True
-        except astroid.NotFoundError:
-            pass
 
         try:
             klass.instance_attr(name)
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1646,7 +1646,12 @@ a metaclass class method.",
                     slots, node.parent.value
                 ):
                     return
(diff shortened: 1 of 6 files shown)


DanielNoord (changes requested):
Let's add `INFERENCE` confidence while we're here.

Also, could you add the example from the issue just to be sure?

Rest LGTM!

Pierre-Sassoulas (dismissed):

tests/functional/a/assigning/assigning_non_slot.py
```diff
@@ -147,7 +147,7 @@ class ClassReassingingInvalidLayoutClass:
     __slots__ = []
 
     def release(self):
-        self.__class__ = ClassWithSlots # [assigning-non-slot]
+        self.__class__ = ClassWithSlots
```
DanielNoord, line 150: This feels incorrect upon second review. This is an instance attribute and should still flag right?

DanielNoord (changes requested):

Pierre-Sassoulas:
I think we need an actual rebase on main to see the real codecov result @DanielNoord

DanielNoord:
> I think we need an actual rebase on main to see the real codecov result @DanielNoord

Yeah, sadly. Oh well, that's a small price to pay.

Sorry for the churn @clavedeluna, you'll need to rebase to get CI to pass :smile:

DanielNoord:
Is the test failure genuine?

Pierre-Sassoulas:
I swear the latest pypy 3.8 release is nothing but trouble :(

pylint/checkers/classes/class_checker.py
```diff
@@ -1621,7 +1621,6 @@ def _check_in_slots(self, node: nodes.AssignAttr) -> None:
             for ancestor in klass.ancestors()
         ):
             return
```
DanielNoord, line 1623:
```suggestion
            return

```

tests/functional/a/assigning/assigning_non_slot.py
```diff
@@ -147,7 +147,7 @@ class ClassReassingingInvalidLayoutClass:
     __slots__ = []
 
     def release(self):
-        self.__class__ = ClassWithSlots # [assigning-non-slot]
+        self.__class__ = ClassWithSlots  # [assigning-non-slot]
```
DanielNoord, line 150:
```suggestion
        self.__class__ = ClassWithSlots  # [assigning-non-slot]
        self.test  # [assigning-non-slot]
```

DanielNoord (changes requested):
Small nits 😄

DanielNoord (approved):

DanielNoord:
@clavedeluna Do you want to try this back port yourself?

Issue #4146, opened 2021-02-24
Checker to replace 'any/all(not condition)' by 'not any/all(condition)' when 'not condition' is slower than 'condition'


Pierre-Sassoulas:
The following function are equivalent but one is faster because it cut the execution tree at the first not instead of checking all the values.

```
import time
from timeit import timeit


def not_all(n, step=1):
    if not all(range(-n, n, step)):
        ...


def any_not(n, step=1):
    if any(x for x in range(-n, n, step)):
        ...


assert not_all(3, 1) == any_not(3, 1)
assert not_all(3, 2) == any_not(3, 2)
assert not_all(5, 1) == any_not(5, 1)

start = time.time()
for i in range(20):
    any_not(1000000)
end = time.time()
print(end - start)
start = time.time()
for i in range(20):
    not_all(1000000)
end = time.time()
print(end - start)

print(
    timeit(
        "import numpy as np; arr = np.ones(int(1e8)); arr[3]=0; np.any(arr<=0)",
        number=10,
    )
)
print(
    timeit(
        "import numpy as np; arr = np.ones(int(1e8)); arr[3]=0; np.all(arr>0)",
        number=10,
    )
)

```

Results:
```
1.52587890625e-05
0.28366875648498535
2.1408786789979786
1.6885079700005008
```

Notice that changing from one generator to two generators with a casting to bool is slower but just changing the condition is faster.

### Describe the solution you'd like

New checker for detecting `not all(values)` and `any(not v for v in values)` to advise to convert it to the most efficient version of the check.

Pierre-Sassoulas:
Actually having two generators instead of one makes the function slower (cf times I did not read properly).

Pierre-Sassoulas:
After thinking about it it seems the example I gave work because of the dataset used, `not all(...)` is not inherently faster than `any(not ...)`.

cdce8p:
> After thinking about it it seems the example I gave work because of the dataset used, `not all(...)` is not inherently faster than `any(not ...)` (unless you know something about the likeliness of something being False or True).

Not exactly, there is a performance difference although you need a dataset with a lot of `True` values before the first `False` one. The reason is that `not all(...)` calls `UNARY_NOT` only once whereas `any(not ...)` calls it on every value.

This example illustrates that
```py
lst = [True] * 99999 + [False]

def not_all_2():
    it = (v for v in lst)
    return not all(it)

def any_not_2():
    it = (not v for v in lst)
    return any(it), it

[...]

PR #7990, opened 2022-12-27
Add new checker `bad-chained-comparison`

### LINKED ISSUE #6559: Warn about ``if x is None == y is None``
### Current problem

We sometimes see code like this:

```python
def foo(*, left=None, right=None):
  """ ... """
  if (left is None) != (right is None):
    raise ValueError('Either both left= and right= need to be provided or none should.')
```

It is very easy to make the mistake of writing the check as:

```python
def foo(*, left=None, right=None):
  """ ... """
  if left is None != right is None:
    raise ValueError('Either both left= and right= need to be provided or none should.')
```
This actually is a chained comparison: ``(left is None != right is None)`` is equivalent to:

```python
(left is None) and (None != right) and (right is None)
```

...which is unsatisfiable, since `right is None` implies `not(None != right)`.

### Desired solution

According to the `comparison` rule in the Python grammar (https://docs.python.org/3/reference/grammar.html),
these comparison operators have the same precedence, and would lead to chained comparisons:

``in, not in, is, is not, <, <=, >, >=, !=, ==``

There are three groups: ``{'in', 'not in'}``, ``{'is', 'is not'}``, ``{'<', '<=', '>', '>=', '!=', '=='}``.

If the linter warned about chained comparisons where one expression is part of two comparisons that belong to different groups, this would flag up checks such as "left is None != right is None".

The price to pay is that this would also trigger on code such as...:

```python
if None is not f(x) > 0:
  ...
```

...but overall, it might be justified in those rare cases where a conditional just like that is appropriate to expect the author to silence the linter on that line.

zenlyj:
## Type of Changes

## Description

As suggested by the issue, a default checker is added to warn users about potential logical mistakes from a chained comparison consisting of comparisons that belong to different semantic categories/groups. Please take a look at the referenced issue as this implementation is pretty much a 1-1 of the solution suggested there.

Any advice is greatly appreciated!

Closes #6559


Pierre-Sassoulas:
The new message in primer seems genuine ! Amazing !

pylint/extensions/bad_chained_comparison.py
```diff
@@ -0,0 +1,47 @@
...
+    def _has_diff_comparison_groups(self, node: nodes.Compare) -> bool:
+        operators = [op[0] for op in node.ops]
+        # Check if comparison operators are in the same group
+        for comparison_group in [COMPARISON_OP, IDENTITY_OP, MEMBERSHIP_OP]:
+            if operators[0] in comparison_group:
+                group = comparison_group
+        for operator in operators:
+            if operator not in group:
+                return True
+        return False
```
Pierre-Sassoulas, line 38:
```suggestion
        return any(o not in group for o in operators)
```
Look like a false negative of the for-any-all checker.

pylint/extensions/bad_chained_comparison.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def visit_compare(self, node: nodes.Compare) -> None:
+        if self._has_diff_comparison_groups(node):
+            num_pieces = f"{len(node.ops)}"
+            self.add_message("bad-chained-comparison", node=node, args=(num_pieces))
```
Pierre-Sassoulas, line 43:
```suggestion
            self.add_message("bad-chained-comparison", node=node, args=(num_pieces), confidence=HIGH)

```
As we're not using inference during the check. Applying this change will require and import and regenerating the functional output.

Pierre-Sassoulas (changes requested):
This looks pretty good already, thank you !

nickdrozd:
I think it would be reasonable to enable this by default and not make it an extension. The one primer result appears to be a full-on error and my guess is that most hits would be also be errors.

I know we try to avoid false positives, but on the other hand, surfacing latent errors helps to highlight the value of the tool.

pylint/extensions/bad_chained_comparison.py
zenlyj: I think both options should have the same performance?

Let's say `o not in group for o in operators` evaluates to the following iterable: `[False, True, True]`.
Then, `o in group for o in operators` evaluates to the inverse: `[True, False, False]`.

`any([False, True, True])` short circuits at the first True, and conversely, `all([True, False, False])`  short circuits at the first False.

My point is, I don't think the frequency of operators that are not in the group matters, since the short circuit effect should always occur on the same index in both iterables.

Please correct me if I am missing something here.

zenlyj:
> I think it would be reasonable to enable this by default and not make it an extension. The one primer result appears to be a full-on error and my guess is that most hits would be also be errors.
> 
> I know we try to avoid false positives, but on the other hand, surfacing latent errors helps to highlight the value of the tool.

I agree, I feel that most hits would most likely be errors. I don't think there are many scenarios that warrant such confusing chained comparisons. Any thoughts on this? @Pierre-Sassoulas

pylint/extensions/bad_chained_comparison.py
zenlyj: Done!

pylint/extensions/bad_chained_comparison.py
zenlyj: Refactored according to pylint's suggestions.

Pierre-Sassoulas (changes requested):
I agree with @nickdrozd, as evidenced by the primer result this is a great check. It's not that often that we discover a real world bug in the primers with a 0% false positive rate from the start. In fact it never happened before. If we want to make this a default check we just have to move it to ``pylint/checker``, please keep the code independent :)

I suggested some wording change and a shorter message with more details, let me know what you think. (I did not review the logic yet)

pylint/extensions/bad_chained_comparison.py
Pierre-Sassoulas: You're right it's identical (Actually I initially made the same mistake in https://github.com/PyCQA/pylint/issues/4146). Resolving, sorry about the noise 😄

doc/whatsnew/fragments/6559.new_check
```diff
@@ -0,0 +1,4 @@
+Add a ``bad-chained-comparison`` extension that emits a warning when
+there is a chained comparison where one expression is part of two comparisons that belong to different groups.
```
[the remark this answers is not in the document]
zenlyj: `semantically incompatible with the other.` sounds good to me, updated!

pylint/extensions/bad_chained_comparison.py
```diff
@@ -0,0 +1,47 @@
...
+MEMBERSHIP_OP = frozenset(("in", "not in"))
+
+
+class BadChainedComparisonChecker(BaseChecker):
+    """Checks for unintentional usage of chained comparison."""
+
+    name = "bad-chained-comparison"
+    msgs = {
+        "W3501": (
+            "Expression gets interpreted as a %s-part chained comparison which straddles comparison groups. If this is not the intent, please parenthesize.",
```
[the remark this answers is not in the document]
zenlyj: This message should be easier to understand, updated!

pylint/extensions/bad_chained_comparison.py
```diff
@@ -0,0 +1,47 @@
...
+            "bad-chained-comparison",
+            "Used when there is a chained comparison where one expression is part of two comparisons that belong to different groups.",
```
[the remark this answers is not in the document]
zenlyj: Updated according to suggestion, made some minor grammar corrections at:
`mean the same thing as "is"` and `chaining them in`

tests/functional/ext/bad_chained_comparison/bad_chained_comparison.txt
```diff
@@ -0,0 +1,11 @@
...
+bad-chained-comparison:23:7:23:44:id_comparison_invalid:Expression gets interpreted as a 3-part chained comparison which straddles comparison groups. If this is not the intent, please parenthesize.:HIGH
+bad-chained-comparison:28:7:28:33:member_comparison_invalid:Expression gets interpreted as a 3-part chained comparison which straddles comparison groups. If this is not the intent, please parenthesize.:HIGH
+bad-chained-comparison:30:9:30:39:member_comparison_invalid:Expression gets interpreted as a 3-part chained comparison which straddles comparison groups. If this is not the intent, please parenthesize.:HIGH
+bad-chained-comparison:32:9:32:58:member_comparison_invalid:Expression gets interpreted as a 5-part chained comparison which straddles comparison groups. If this is not the intent, please parenthesize.:HIGH
+bad-chained-comparison:34:9:34:57:member_comparison_invalid:Expression gets interpreted as a 5-part chained comparison which straddles comparison groups. If this is not the intent, please parenthesize.:HIGH
```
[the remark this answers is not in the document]
zenlyj: Done, message should emit the unique operators (ordered by left to right in expression) used in the chained comparison now.

pylint/checkers/bad_chained_comparison.py
```diff
@@ -0,0 +1,58 @@
...
+class BadChainedComparisonChecker(BaseChecker):
+    """Checks for unintentional usage of chained comparison."""
+
+    name = "bad-chained-comparison"
+    msgs = {
+        "W3501": (
+            "Suspicious %s-part chained comparison using semantically incompatible operators (%s)",
+            "bad-chained-comparison",
+            "Used when there is a chained comparison where one expression is part of two comparisons that belong to different semantic groups "
+            ' ("<" does not mean the same thing as "is", chaining them in "0 < x is None" is probably a mistake).',
```
Pierre-Sassoulas, line 27:
```suggestion
            "Used when there is a chained comparison where one expression is part "
            "of two comparisons that belong to different semantic groups "
            ' ("<" does not mean the same thing as "is", chaining them in '
            '"0 < x is None" is probably a mistake).',

```

pylint/checkers/bad_chained_comparison.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def _has_diff_semantic_groups(self, operators: list) -> bool:
+        # Check if comparison operators are in the same semantic group
+        for semantic_group in (COMPARISON_OP, IDENTITY_OP, MEMBERSHIP_OP):
+            if operators[0] in semantic_group:
+                group = semantic_group
+        return not all(o in group for o in operators)
+
+    def visit_compare(self, node: nodes.Compare) -> None:
+        operators = self._get_distinct_operators(node)
```
Pierre-Sassoulas, line 43:
```suggestion
        operators = set(op[0] for op in node.ops)
```

pylint/checkers/bad_chained_comparison.py
```diff
@@ -0,0 +1,58 @@
...
+            "Suspicious %s-part chained comparison using semantically incompatible operators (%s)",
+            "bad-chained-comparison",
+            "Used when there is a chained comparison where one expression is part of two comparisons that belong to different semantic groups "
+            ' ("<" does not mean the same thing as "is", chaining them in "0 < x is None" is probably a mistake).',
+        )
+    }
+
+    def _get_distinct_operators(self, node: nodes.Compare) -> list:
+        operators = [op[0] for op in node.ops]
+        return list(dict.fromkeys(operators))
```
Pierre-Sassoulas, line 33:
```suggestion
```
We're trying to get a list of uniq operator right ? Shouldn't this be a set as in ``set(op[0] for op in node.ops)`` ?

Pierre-Sassoulas (changes requested):

pylint/checkers/bad_chained_comparison.py
zenlyj: Using `set(...)` gives us a list of unique operators, however, the operators will be unordered (from my understanding, python sets are unordered). Therefore, the message output will be non-deterministic.

The `list(dict.fromkeys(...))` gives a list of unique operators as well, and will order the operators according to their position in the expression https://stackoverflow.com/questions/39755464/python-unique-items-of-list-in-the-order-that-it-appears. 
For example: given the expression `(a < b <= c in d)`, the list will always be `['<', '<=', 'in']`. Whereas the set may create `{'<=', 'in', '<'}` in an instance, and output `{'in', '<', '<='}` in another instance.

In my opinion, it would be better if the message output is deterministic. Also, so that we can perform functional testing 🙂 Let me know what you think.

pylint/checkers/bad_chained_comparison.py
zenlyj: updated!

pylint/checkers/bad_chained_comparison.py
Pierre-Sassoulas: Maybe we can sort explicitly then ?

pylint/checkers/bad_chained_comparison.py
Pierre-Sassoulas:
```suggestion
        operators = sorted(set(op[0] for op in node.ops))

```

pylint/checkers/bad_chained_comparison.py
zenlyj: that works too, since the operator set is quite small, sorting shouldn't be too costly. let's go with sort so that the code will be easier to understand.

Pierre-Sassoulas (dismissed):

Pierre-Sassoulas:
I think it's a pretty nice checker, great job @zenlyj. Let's wait for another maintainer review, this is a default check after all, mistakes would affect a lot of users.

pylint/checkers/bad_chained_comparison.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import annotations
```
Pierre-Sassoulas, line 1:
```suggestion
# Licensed under the GPL: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html
# For details: https://github.com/PyCQA/pylint/blob/main/LICENSE
# Copyright (c) https://github.com/PyCQA/pylint/blob/main/CONTRIBUTORS.txt

from __future__ import annotations
```

pylint/checkers/bad_chained_comparison.py
```diff
@@ -0,0 +1,56 @@
...
+            "Suspicious %s-part chained comparison using semantically incompatible operators (%s)",
+            "bad-chained-comparison",
+            "Used when there is a chained comparison where one expression is part "
+            "of two comparisons that belong to different semantic groups "
+            '("<" does not mean the same thing as "is", chaining them in '
+            '"0 < x is None" is probably a mistake).',
+        )
+    }
+
+    def _has_diff_semantic_groups(self, operators: list) -> bool:
```
Pierre-Sassoulas, line 33:
```suggestion
    def _has_diff_semantic_groups(self, operators: list[str]) -> bool:

```

Pierre-Sassoulas:
I have a really hard time to understand the pre-commit error. I can reproduce locally but I can't seem to fix it.

zenlyj:
@Pierre-Sassoulas it seems that the pre-commit error is due to the files having Windows styles line breaks. It should be fixed now, sorry for the trouble!

Regarding the documentation check fail, is this expected for new documentation pages added?

Pierre-Sassoulas:
Strange, I fixed the line ending locally and it did not  work.

Pierre-Sassoulas:
#8007 has been merged, could you rebase on it please ?

Pierre-Sassoulas:
The documentation check is expected to fail as long as it's not merged in main and in the official doc.

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
First check that not only did not have a single primer false positives but even detected a real bug to fix in a very popular well scrutinized project.  It already detected and [permitted to fix a bug in sentry](https://github.com/getsentry/sentry/pull/42669), which is pretty impressive ! Great checker, amazing job 👌

Issue #6935, opened 2022-06-12
Should we switch to ``codecov`` instead of ``coveralls``


DanielNoord:
Lately I have been getting increasingly annoyed with ``coveralls``. This is for a number of reasons:
* It can take 15/20 minutes for their comment to actually be posted after the CI step passed
* They don't show the commit for which the comment was posted (see https://github.com/lemurheavy/coveralls-public/issues/1638), which is annoying as you can't be sure that the comment is updated due to 1.
* You can't see the annotated code without an account
* They handled their more than a week long outage very poorly this winter

That's why for my own project I have switched to [codecov](https://about.codecov.io). I think it is the main other service and it is used (for example) by `home-assistant` and `pytest`. Both of which are projects that I value highly in terms of CI and project maintenance.

The switch to `codecov` is pretty easy, see: https://github.com/DanielNoord/pydocstringformatter/pull/112
For public projects you don't need to set up any environment variables and can just start using the service.

It would be good to give some administration rights to some members though to perhaps tweak the PR comments.

I'd be willing to create a PR to do this, but I want some discussion/confirmation first.

_Tagging for input @Pierre-Sassoulas @jacobtylerwalls_

Pierre-Sassoulas:
>It would be good to give some administration rights to some members though to perhaps tweak the PR comments.

I need someone to be a member of PyCQA to be able to add him to the pylint's internal maintainer/dev/admin group. PCManticore want sigmavirus24 confirmation to grant me the right to add someone to PyCQA so right now it's not possible (but I'm pestering Ian about it with some regularity 😄). 

I don't have strong opinion about the subject, I think I got accustomed to coveralls more or less working over the years. But I trust you if you think codecov is better. Being able to merge multiple pipelines result would be nice as it's impossible to do with coveralls (but it's probably out of scope for the migration.)

jacobtylerwalls:
+0

> I don't have strong opinion about the subject, I think I got accustomed to coveralls more or less working over the years.

Same.

clavedeluna:
I'd like to help improve code coverage but as mentioned above, I can't see annotated code right now.

Pierre-Sassoulas:
You need to login with your GitHub account @clavedeluna coverall need to use your api key to make the call to get the code or they would be rate limited by GitHub

Pierre-Sassoulas:
I'm now admin in PyCQA, and Jacob, Marc and Daniel should be admin in pylint (let me know if that's not the case).

[...]

PR #7992, opened 2022-12-27
Use codecov instead of coveralls for coverage reporting

DanielNoord:
  included in the changelog. `<type>` can be one of: breaking, user_action, feature,
  new_check, removed_check, extension, false_positive, false_negative, bugfix, other, internal.
  If necessary you can write details or offer examples on how the new change is supposed to work.

## Type of Changes

## Description

Closes https://github.com/PyCQA/pylint/issues/6935

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -81,19 +81,14 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     needs: tests-linux
-    strategy:
-      matrix:
-        python-version: ["3.11"]
-    env:
-      COVERAGERC_FILE: .coveragerc
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v3.2.0
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python 3.11
         id: python
         uses: actions/setup-python@v4.4.0
         with:
-          python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


Pierre-Sassoulas (commented):
Very nice 👍 ! I don't know anything about codecov so this might be a lot of work or even impossible, but maybe it's possible to merge test's and primer's coverage ? It was impossible to do with coveralls.

DanielNoord:
> Very nice +1 ! I don't know anything about codecov so this might be a lot of work or even impossible, but maybe it's possible to merge test's and primer's coverage ? It was impossible to do with coveralls.

That is certainly possible, although we don't record coverage currently when running the primer. That would probably slow it down significantly, so I'm not sure how useful that would be.
We could start collecting coverage for PyPY and Windows though. That would be nice I guess as that would be the only way to ever reach 100% :smile:

Pierre-Sassoulas:
Yeah you're right, forget about primers, windows pypy  and lower/upper python interpreter makes a lot more sense

Pierre-Sassoulas:
This looks amazing by the way, I'm already like "What is this red little square there 😠 ??!"

Pierre-Sassoulas (approved):

DanielNoord:
Yeah I much prefer both the comment and the site of `codecov`. I'm going to (auto-)merge and then let `codecov` do one of the open PRs to make sure the settings are correct.

PR #7994, opened 2022-12-27
Cover missing line in dict-init-mutate test

nickdrozd:

## Description

This covers a missing line from the dict-init-mutate extension test.

### DIFF
--- a/tests/functional/ext/dict_init_mutate.py
+++ b/tests/functional/ext/dict_init_mutate.py
@@ -36,3 +36,6 @@ def update_dict(di):
     di["one"] = 1
 
 update_dict(config)
+
+config = {}
+globals()["config"]["dir"] = "bin"


Pierre-Sassoulas (approved):
👌

DanielNoord:
Bit weird that the `codecov/patch` doesn't show an increase and says "not affected". Anyway, awesome that this switch has already led to a coverage improving PR :smile:

tests/functional/ext/dict_init_mutate.py
```diff
@@ -36,3 +36,6 @@ def update_dict(di):
     di["one"] = 1
 
 update_dict(config)
+
+config = {}
+globals()["config"]["dir"] = "bin"
```
DanielNoord, line 41: Should we also test `globals()["config"]`?

nickdrozd: I'm sure there a thousand variations that could be added. In this case this was just the simplest piece of technically valid code I could think of to trigger the missing line.

Pierre-Sassoulas:
It was already visible with coverall but we missed it during review because you have to follow the link and login, and if the coveralls comment was modified a thousand time I forget to do that. That's pretty nice.

PR #8006, opened 2022-12-31
Fix ``abstract-class-instantiated`` false positive for ``__new__`` method.

### LINKED ISSUE #3060: False abstract-class-instantiated with pandas.ExcelWriter
Hello,

### Steps to reproduce

```python
"""Lorem ipsum"""

import pandas as pd


def demo():
    """Demo for the false positive"""
    with pd.ExcelWriter("demo.xlsx") as writer:
        print(writer)
```

### Current behavior
Pylint gives: ``abstract-class-instantiated(E0110): test.py:8:9: demo: Abstract class 'ExcelWriter' with abstract methods instantiated``

### Expected behavior

No issue

### pylint --version output
- Astroid 2.2.5
- Pylint 2.3.1
- Python 3.7.4
- GCC 7.4.0

### Discussion

This ticket has originally been opened on pandas github: https://github.com/pandas-dev/pandas/issues/27634

According to the discussions there, it seems that pylint has issues with introspection of objects defining __new__

clavedeluna:
## Type of Changes

## Description

Closes #3060

### DIFF
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -19,7 +19,6 @@ from pylint.checkers.base.basic_checker import _BasicChecker
 from pylint.checkers.utils import infer_all
 from pylint.interfaces import HIGH, INFERENCE
 
-ABC_METACLASSES = {"_py_abc.ABCMeta", "abc.ABCMeta"}  # Python 3.7+,
 # List of methods which can be redefined
 REDEFINABLE_METHODS = frozenset(("__module__",))
 TYPING_FORWARD_REF_QNAME = "typing.ForwardRef"
@@ -496,9 +495,6 @@ class BasicErrorChecker(_BasicChecker):
                     )
                     break
 
-            return
-
-        if metaclass.qname() not in ABC_METACLASSES:
             return
 
         if _new_without_super(inferred):
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -17,7 +17,7 @@ from astroid.typing import InferenceResult
 from pylint.checkers import utils
 from pylint.checkers.base.basic_checker import _BasicChecker
 from pylint.checkers.utils import infer_all
-from pylint.interfaces import HIGH
+from pylint.interfaces import HIGH, INFERENCE
 
 ABC_METACLASSES = {"_py_abc.ABCMeta", "abc.ABCMeta"}  # Python 3.7+,
 # List of methods which can be redefined
@@ -76,6 +76,38 @@ def _has_abstract_methods(node: nodes.ClassDef) -> bool:
     with `abc` decorators.
     """
     return len(utils.unimplemented_abstract_methods(node)) > 0
+
+
+def _new_correctly_implemented(node: nodes.ClassDef) -> bool:
+    """Check if node implements `__new__`.
+
+    If `__new__` is implemented, check if it calls `super().__new__(cls)`.
+    """
+    if "__new__" not in node.locals:
+        return False
+
+    new = next(node.igetattr("__new__"))
+    if not isinstance(new, astroid.UnboundMethod):
+        return False
+
+    calls = new.nodes_of_class(
+        nodes.Call, skip_klass=(nodes.FunctionDef, nodes.ClassDef)
+    )
+    for call in calls:
+        if not isinstance(call.func, nodes.Attribute):
+            continue
+        if call.func.attrname != "__new__":
+            continue
+
+        super_call = utils.safe_infer(call.func.expr)
+        if not isinstance(super_call, astroid.objects.Super):
+            continue
+
+        # `__new__` calls `super().__new__(cls)`.
+        return False
+
+    # If we get to this point, it means `__new__` has no calls to `super().__new__(cls)`.
+    return True
 
 
 def redefined_by_decorator(node: nodes.FunctionDef) -> bool:
@@ -446,7 +478,6 @@ class BasicErrorChecker(_BasicChecker):
             # body, we're expecting that it knows what it is doing.
             return
 
-        # __init__ was called
         abstract_methods = _has_abstract_methods(inferred)
 
         if not abstract_methods:
@@ -467,8 +498,16 @@ class BasicErrorChecker(_BasicChecker):
             return
 
         if metaclass.qname() in ABC_METACLASSES:
+            if _new_correctly_implemented(inferred):
+                # A class that implements `__new__` without calling `super().__new__(cls)`
+                # should not emit the message.
+                return
+
             self.add_message(
-                "abstract-class-instantiated", args=(inferred.name,), node=node
+                "abstract-class-instantiated",
+                args=(inferred.name,),
+                node=node,
+                confidence=INFERENCE,
             )
 
     def _check_yield_outside_func(self, node: nodes.Yield) -> None:
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -78,15 +78,18 @@ def _has_abstract_methods(node: nodes.ClassDef) -> bool:
     return len(utils.unimplemented_abstract_methods(node)) > 0
 
 
-def _new_instantiates_super(node: nodes.ClassDef) -> bool:
-    """Check if node implements `__new__`.
+def _new_without_super(node: nodes.ClassDef) -> bool:
+    """Check if node implements `__new__` but does not call `super()` on `__new__`.
 
     If `__new__` is implemented, check if it calls `super().__new__(cls)`.
     """
     if "__new__" not in node.locals:
         return False
 
-    new = next(node.igetattr("__new__"))
+    try:
+        new = next(node.igetattr("__new__"))
+    except astroid.InferenceError:
+        return False
 
     calls = new.nodes_of_class(
         nodes.Call, skip_klass=(nodes.FunctionDef, nodes.ClassDef)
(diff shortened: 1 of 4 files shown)


pylint/checkers/base/basic_error_checker.py
```diff
@@ -78,6 +78,38 @@ def _has_abstract_methods(node: nodes.ClassDef) -> bool:
...
+        return False
+
+    calls = new.nodes_of_class(
+        nodes.Call, skip_klass=(nodes.FunctionDef, nodes.ClassDef)
+    )
+    for call in calls:
+        if not isinstance(call.func, nodes.Attribute):
+            continue
+        if call.func.attrname != "__new__":
+            continue
```
DanielNoord, line 100: @Pierre-Sassoulas This is very fancy and a nice side effect of `codecov`.

pylint/checkers/base/basic_error_checker.py
```diff
@@ -78,6 +78,38 @@ def _has_abstract_methods(node: nodes.ClassDef) -> bool:
     return len(utils.unimplemented_abstract_methods(node)) > 0
 
 
+def _new_correctly_implemented(node: nodes.ClassDef) -> bool:
```
DanielNoord, line 81: I don't think we should call this "correctly". What about: `new_instantiates_super`?

DanielNoord (changes requested):

pylint/checkers/base/basic_error_checker.py
clavedeluna: lovely, naming is hard :)

pylint/checkers/base/basic_error_checker.py
clavedeluna: Is codecov not taking into account functional tests? Because if I add a breakpoint right before the `continue` in ln 100 my functional tests locally do stop, so I think it is covered.

clavedeluna: same with ln 104, it is covered by functional tests

Pierre-Sassoulas: Nice tooltip ! Is this a bug and really is covered though ? Maybe it depends on the interpreter being used ? What do you have locally @clavedeluna ?

clavedeluna: Ah good call, testing with python3.8 it's covered, but with 3.11 it isn't, so I assume codecov is running with the latter? Not sure how to verify that in codecov settings, it's not specified in the yaml

DanielNoord: It feels weird that `coverage` doesn't see this. I would need to investigate this as I have never seen such things in other projects that use `coverage`.

clavedeluna: Locally running pytest with coverage I can see the following

```
pylint/checkers/base/basic_error_checker.py                    242     10    96%   284, 434-435, 489-496, 603-604
```

which tells me the lines codecov is complaining about (L98, L102) are indeed covered by tests.

Happy to continue looking into this but I would need to find this data in the action run, which I don't. So I'll block this PR for now.

clavedeluna:
Blocked by codecov issue

pylint/checkers/base/basic_error_checker.py
DanielNoord: I never saw the edited comment. We don't install `pandas` so this indeed uncovered in the tests. `codecov` is right here.
The tests should work even without `pandas` in the environment as we don't have that as a test requirement.

clavedeluna: Aha! You found the issue, that's indeed the case. The only functional test that passes thru that code is the one using pandas. I'll add one to fix this. Thanks!

pylint/checkers/base/basic_error_checker.py
```diff
@@ -78,6 +78,36 @@ def _has_abstract_methods(node: nodes.ClassDef) -> bool:
...
 
+def _new_instantiates_super(node: nodes.ClassDef) -> bool:
+    """Check if node implements `__new__`.
+
+    If `__new__` is implemented, check if it calls `super().__new__(cls)`.
+    """
+    if "__new__" not in node.locals:
+        return False
+
+    new = next(node.igetattr("__new__"))
```
DanielNoord, line 89: This should probably be done in a `try...except`. Or with `safe_infer`.

pylint/checkers/base/basic_error_checker.py
```diff
@@ -78,6 +78,36 @@ def _has_abstract_methods(node: nodes.ClassDef) -> bool:
...
+
+        super_call = utils.safe_infer(call.func.expr)
+        if not isinstance(super_call, astroid.objects.Super):
+            continue
+
+        # `__new__` calls `super().__new__(cls)`.
+        return False
+
+    # If we get to this point, it means `__new__` has no calls to `super().__new__(cls)`.
+    return True
```
DanielNoord, line 108: This is the other way around from the docstring right? The function should probably be renamed.

pylint/checkers/base/basic_error_checker.py
```diff
@@ -467,8 +496,16 @@ def _check_inferred_class_is_abstract(
             return
 
         if metaclass.qname() in ABC_METACLASSES:
```
DanielNoord, line 498: Let's make L498 a `if not` and then return. There is no need for this indentation here.

tests/functional/a/abstract/abstract_class_instantiated.py
```diff
@@ -7,8 +7,10 @@
 # pylint: disable=abstract-method, import-error
 
 import abc
+import warnings
 import weakref
 from lala import Bala
+import pandas as pd
```
DanielNoord, line 13: Let's remove this test as we don't install `pandas`.

tests/functional/a/abstract/abstract_class_instantiated.py
```diff
@@ -141,3 +143,61 @@ def test(self):
...
+    def sheets(self):
+        """sheets."""
+
+    @property
+    @abc.abstractmethod
+    def book(self):
+        """book."""
+
+
+bad = GoodithComplexNew()
```
DanielNoord, line 203: What about:

```python
def __new__(self):
    cls = super().__new__()
    return cls

def __new__(self):
    cls = super().__new__()
    return "1"
```

DanielNoord (changes requested):
Thanks for fixing the coverage and test!

tests/functional/a/abstract/abstract_class_instantiated.py
clavedeluna: as long as the class has abstract methods, yes msg will be emitted. will add them as test cases

pylint/checkers/base/basic_error_checker.py
clavedeluna:
```
    try:
        new = next(node.igetattr("__new__"))
        return False
```

is the closest I see to other repo code using igetattr. However, I can't think of a single example where we'd actually raise `InferenceError` since __new__ is a dunder for pretty much any class. Do you have a suggestion? I can remove the try/except or ignore coverage for this line just for precaution.

DanielNoord: Hmm.. @Pierre-Sassoulas what do you want? Non-covered code? Or code that is known to raise a certain exception that we just don't know the edge case for yet?

pylint/checkers/base/basic_error_checker.py
```diff
@@ -466,10 +498,20 @@ def _check_inferred_class_is_abstract(
 
             return
 
-        if metaclass.qname() in ABC_METACLASSES:
-            self.add_message(
-                "abstract-class-instantiated", args=(inferred.name,), node=node
-            )
+        if metaclass.qname() not in ABC_METACLASSES:
```
DanielNoord, line 501: Can you add a test for this?

DanielNoord (changes requested):

pylint/checkers/base/basic_error_checker.py
clavedeluna: Turns out I can't come up with a single class example where we would land in this code, so going to remove this check. The previous if-checks above this handle the edge cases.

clavedeluna:
hm need to see what's up with that new primer output

clavedeluna:
I'm still stumped by 
```
abstract-class-instantiated:
Abstract class 'DontReadFromInput' with abstract methods instantiated
https://github.com/pytest-dev/pytest/blob/ca40380e99c2cdaab1d0c041f9f28cff37ef8ff9/src/_pytest/capture.py#L344
```

I tried to reproduce it locally bit that msg doesn't get raised. Maybe a maintainer has a better idea?

DanielNoord:
Have you checked wether `TextIO` has any "abstract" methods?

clavedeluna:
> Have you checked wether `TextIO` has any "abstract" methods?

It [does](https://github.com/python/cpython/blob/main/Lib/typing.py#L3273) ... while I can see `DontReadFromInput` getting some `abstract-method` msg raised, I'm not sure why we'd raise `abstract-class-instantiated`? Maybe I'm missing something subtle and the primer is right.

DanielNoord:
> > Have you checked wether `TextIO` has any "abstract" methods?
> 
> It [does](https://github.com/python/cpython/blob/main/Lib/typing.py#L3273) ... while I can see `DontReadFromInput` getting some `abstract-method` msg raised, I'm not sure why we'd raise `abstract-class-instantiated`? Maybe I'm missing something subtle and the primer is right.

I don't understand this comment. If `DontReadFromInput` has abstract methods that don't get implemented than it is correct that we raise this here as well right?

clavedeluna:
Maybe I'm misunderstanding the purpose of the msg `abstract-class-instantiated` versus msg `abstract-method`? What I understand from the docs is that the former is telling a user "don't directly instantiate a class with `abc.ABCMeta` and with abstract-methods", but if the class instantiated is inheriting from one of these, but the instantiated class DOES implement the abstract methods, as in the case of [DontReadFromInput](https://github.com/pytest-dev/pytest/blob/ca40380e99c2cdaab1d0c041f9f28cff37ef8ff9/src/_pytest/capture.py#L196) then it's fine.

Anyway, if I'm wrong, then this PR should be able to be merged, I think everything asked is complete?

DanielNoord:
> Maybe I'm misunderstanding the purpose of the msg `abstract-class-instantiated` versus msg `abstract-method`? What I understand from the docs is that the former is telling a user "don't directly instantiate a class with `abc.ABCMeta` and with abstract-methods", but if the class instantiated is inheriting from one of these, but the instantiated class DOES implement the abstract methods, as in the case of [DontReadFromInput](https://github.com/pytest-dev/pytest/blob/ca40380e99c2cdaab1d0c041f9f28cff37ef8ff9/src/_pytest/capture.py#L196) then it's fine.
> 
> Anyway, if I'm wrong, then this PR should be able to be merged, I think everything asked is complete?

But isn't one of the abstract methods of `DontReadFromInput` missing from the final implementation?

PR #8007, opened 2023-01-01
Correction for doc files line ending

zenlyj:
## Type of Changes

## Description

Just noticed that my previous 2 PRs #7970 and #7936 committed files with Windows style line endings instead of UNIX style line endings. This PR aims to modify the line endings to LF to keep things consistent across the repo, nothing functional is changed.


Pierre-Sassoulas (commented):
Thank you for this. I wanted to do that repo wide with a .gitattributes but some functional tests should use windows line ending so we don't regress on Windows. I delayed actually doing it. Do you think it's possible to normalize only part of the repo this way ?

zenlyj:
> Thank you for this. I wanted to do that repo wide with a .gitattributes but some functional tests should use windows line ending so we don't regress on Windows. I delayed actually doing it. Do you think it's possible to normalize only part of the repo this way ?

I have not done it myself, but here seems to say that it is possible? https://stackoverflow.com/questions/6670140/how-do-i-set-all-files-in-a-subdirectory-to-use-a-specific-eol-using-gitattribut

Pierre-Sassoulas:
Thank you for doing the research, do you mind adding the gitattributes and excluding the tests data ? Could be the same regex than 'fixtures' in the pre-commit config. I can also take over the PR if you want

zenlyj:
sure, I can give it a shot. though I need some time to read up on it and test things out.

Pierre-Sassoulas:
👍 Here's the regex for the paths that should stay as they are: ``tests(/\w*)*/functional/|tests/input|tests(/\w*)*data/``.

zenlyj:
Added git attributes to normalize the line endings to LF on file check in. Regular expressions cannot be used in git attributes, so I excluded the paths explicitly. Please help to do some sanity test to ensure that the behavior is as expected, thanks!

zenlyj:
Added the git attributes to normalize file endings to LF on file check ins. Regular expressions cannot be used in git attributes, so the paths are specified explicitly. Please help to do some sanity test to ensure that the behavior is as expected, thanks!

Pierre-Sassoulas (approved):
Thanks a lot @zenlyj, I've been wanting to do that for a long time, and it's finally done 🙏  🎉

PR #8021, opened 2023-01-04
feat(8015): `consider-refactoring-into-while-condition` checker

### LINKED ISSUE #8015: New check: `break` at the top of `while True` loop
### Current problem

A not-uncommon pattern is to check some condition at the top of a `while True` loop and break if the condition is met:

```python
while True:
    if condition:
        break

    do_stuff()
```

There is an instance of this in the Pylint codebase.

### Desired solution

It would be more straightforward and less verbose to negate the condition and move it into the `while` expression:

```python
while not condition:
    do_stuff()
```

Add a new check for the following circumstance:

1. There is a `while` loop with `True` condition (or more generally, constant condition);
2. the first statement of the loop body is an `if` statement; and
3. the first statement of the `if` body is a `break` statement.

In this case, emit a suggestion to negate the condition and move it to the `while`.

I don't know what the name of the warning should be. Suggestions welcome!

### Additional context

_No response_

yushao2:
## Type of Changes

### Background

This PR aims to implement a new checker `consider-refactoring-into-while-condition` as requested.

### Scope
- [x] Implemented new *conventions* checker, `consider-refactoring-into-while-condition` under `extensions` with id `R3501`
- [x] Added basic functional tests -- appreciate more examples
- [x] Added docs (news fragment, user_guide) 

Closes #8015


nickdrozd:
Nice! As expected, this flags one use in the Pylint code, so that will have to be fixed. It's also raised in an existing test (`membership_protocol.py`), and it can be disabled there. Other than that, LGTM.

yushao2:
> Nice! As expected, this flags one use in the Pylint code, so that will have to be fixed. It's also raised in an existing test (`membership_protocol.py`), and it can be disabled there. Other than that, LGTM.

hey, i updated the checker with better error messages with a variety of other possible use-cases -- let me know if this is better

tests/functional/c/consider/consider_refactoring_into_while_condition.py
```diff
@@ -0,0 +1,78 @@
...
+        while True:   # [consider-refactoring-into-while-condition]
+            if a_list:
+                break
+            a_list.append(1)
+
+    def test_error_message_3(self):
+        a_list = []
+        # Should recommend `while not a_list`
+        while True:   # [consider-refactoring-into-while-condition]
+            if a_list is not None:
```
nickdrozd, line 46: `not a_list` and `not a_list is None` are not equivalent. In this case, the former will fail immediately and the latter will never fail. I think it would be safer (and easier) to simply negate the condition, even if that would lead to something like a double negative.

nickdrozd: If the suggested change would be unidiomatic, that could be caught with a separate checker on a subsequent Pylint pass. Maybe requiring multiple passes could be considered undesirable. I don't know the best way to handle that.

nickdrozd (changes requested):
Maybe it would be better to do it the simpler way?

tests/functional/c/consider/consider_refactoring_into_while_condition.py
yushao2: yeah i'm stumped on this dilemma as well. ideally I want to generate an idiomatic suggestion which would be more useful

yushao2: @nickdrozd ,

wouldn't 

` if a_list is not None` be compacted to 

```py
while not (a_list is not None)
```

since `a_list is not None` can be expressed as simply `a_list` i would assume this recommendation of `while not a_list` to be correct

nickdrozd: The empty list is falsy, but it still isn't `None`. Try this:

```python
a_list = []

while not (a_list is not None):  # fails immediately
    print('asdf')  # unreachable

while not a_list:  # never fails, infinite loop
    print('zxcv')
```

pylint/checkers/base/basic_error_checker.py
```diff
@@ -396,9 +396,7 @@ def visit_unaryop(self, node: nodes.UnaryOp) -> None:
 
     def _check_nonlocal_without_binding(self, node: nodes.Nonlocal, name: str) -> None:
         current_scope = node.scope()
-        while True:
-            if current_scope.parent is None:
-                break
+        while current_scope.parent:
```
Pierre-Sassoulas, line 399:
```suggestion
        while current_scope.parent is not None:
```

Pierre-Sassoulas (changes requested):
I did not check in detail but could you add an example of good/bad code in the documentation please ? :)

nickdrozd:
Some of the suggested changes from the primer results look a little wonky. But the good news is that the check is triggering properly, and lots of hits show up!

DanielNoord (commented):
This feels really opinionated doesn't it? Shouldn't this be an extension?

Pierre-Sassoulas:
I think it should be an extension but I also don't think it's *that* opinionated. It clarify the code by transforming 3 lines of code with two concepts and 2 indentation  into a one liner using only one concept and one indentation. What I'm wondering is how well it handle case where one could genuinely want to use this (a very long list of condition with multiple ``if break`` to avoid a while with too much condition in it ?)

nickdrozd:
I don't think it's any more opinionated than existing checks, for example `no-else-return`.

tests/functional/c/consider/consider_refactoring_into_while_condition.py
yushao2: Ah okay, I see where you're coming from.

In this case, however I wanted to provide an example where sometimes a variable/attribute is checked to be None and if so breaks out of the loop.

In this case we can refactor it to while (var/attr)

I'll update the test cases to make it more clear. thanks

yushao2:
> I did not check in detail but could you add an example of good/bad code in the documentation please ? :)

where should i add them?

Pierre-Sassoulas:
When you upgrade the doc with ' tox -e docs' a directory should be created automatically Doc/data/messages/

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -60,6 +60,14 @@ class RecommendationChecker(checkers.BaseChecker):
             "which could potentially be a f-string. The use of f-strings is preferred. "
             "Requires Python 3.6 and ``py-version >= 3.6``.",
         ),
+        "C0210": (
+            "Consider using `while %s`",
```
Pierre-Sassoulas, line 64:
```suggestion
            "Consider using 'while %s' instead of 'while True:' an 'if', and a 'break'",
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -60,6 +60,14 @@ class RecommendationChecker(checkers.BaseChecker):
...
+            "consider-refactoring-into-while-condition",
+            "Emitted when `while <constant>:` loop is used with the first statement being a if statement "
+            "with a conditional check to break out of the loop. "
+            "The `if`` statement can be removed with the conditional check refactored to be "
+            "within the `while` statement",
```
Pierre-Sassoulas, line 69:
```suggestion
            "Emitted when `while True:` loop is used and the first statement is a break condition."
            "The `if / break`` construct can be removed if the check is inverted and moved to "
            "the `while` statement",
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -430,3 +438,44 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
...
+            elif ops == "==":
+                msg = f"{lhs.as_string()} != {rhs.as_string()}"
+            elif ops == "!=":
+                msg = f"{lhs.as_string()} == {rhs.as_string()}"
+
+        self.add_message(
+            "consider-refactoring-into-while-condition",
+            node=node,
+            line=node.lineno,
+            args=(msg,),
```
Pierre-Sassoulas, line 480:
```suggestion
            args=(msg,),
            confidence=HIGH,
```

doc/data/messages/c/consider-refactoring-into-while-condition/bad.py
```diff
@@ -0,0 +1,6 @@
+a = [1,2,3,4,5]
+
+while True:  # [consider-refactoring-into-while-condition]
+  if len(a) == 0:
+    break
+  a.pop()
```
Pierre-Sassoulas, line 6: I like example that are less abstract and understandable by school children, it doesn't have to be fruit, it's just that I don't have a lot of imagination :)
```suggestion
fruit_basket = ["apple", "orange", "banana", "cherry", "guava"]

while True:  # [consider-refactoring-into-while-condition]
  if len(fruit_basket ) == 0:
    break
    
  fruit = fruit_basket.pop()
  print(f"We removed {fruit} from the basket")
```

doc/data/messages/c/consider-refactoring-into-while-condition/good.py
```diff
@@ -0,0 +1,4 @@
+a = [1,2,3,4,5]
+
+while len(a) != 0:
+  a.pop()
```
Pierre-Sassoulas, line 4:
```suggestion
fruit_basket = ["apple", "orange", "banana", "cherry", "guava"]

while len(fruit_basket ) != 0:
  fruit = fruit_basket.pop()
  print(f"We removed {fruit} from the basket")
```

Pierre-Sassoulas (changes requested):
Thank you for working on this, this is looking pretty good already. I agree with Daniel that it should be an extension. Or at least it should be independent from the default refactoring checker even if we  choose to make it a default check later.

pylint/checkers/refactoring/recommendation_checker.py
yushao2: just wondering, what should I specify for confidence moving forward in future contributions?

uzzell:
I have one suggestion and one question about the replacements that the checker suggests.  First, it would be good to suggest `while foo not in bar`, rather than `while not foo in bar`.  Second, if the "break" condition is `foo >= 0`, the checker suggests `while not foo >= 0`.  Should we suggest `while foo < 0` instead?

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -430,3 +438,44 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
...
+
+            msg = f"not {node.body[0].test.as_string()}"
+            if ops == "is":
+                msg = f"{lhs.as_string()} is not {rhs.as_string()}"
+            elif ops == "is not":
+                msg = f"{lhs.as_string()} is {rhs.as_string()}"
+            elif ops == "==":
+                msg = f"{lhs.as_string()} != {rhs.as_string()}"
+            elif ops == "!=":
+                msg = f"{lhs.as_string()} == {rhs.as_string()}"
```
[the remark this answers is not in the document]
yushao2: refactoring done, please check to see if what i did makes sense or if it could be improved

yushao2:
> I have one suggestion and one question about the replacements that the checker suggests. First, it would be good to suggest `while foo not in bar`, rather than `while not foo in bar`. Second, if the "break" condition is `foo >= 0`, the checker suggests `while not foo >= 0`. Should we suggest `while foo < 0` instead?

yeah, i added those cases in, thanks for that

yushao2:
> Thank you for working on this, this is looking pretty good already. I agree with Daniel that it should be an extension. Or at least it should be independent from the default refactoring checker even if we choose to make it a default check later.

also, how should i make this an extension?

tests/functional/c/consider/consider_refactoring_into_while_condition.py
```diff
@@ -0,0 +1,147 @@
...
+            if k < 1:
+                break
+            k -= 1
+
+    def test_error_message_15(self):
+        k = 1
+        while True:  # [consider-refactoring-into-while-condition]
+            if k <= 1:
+                break
+            k -= 1
```
Pierre-Sassoulas, line 147:
```suggestion
            k -= 1

def test_error_message_multiple_break(self, k: int) -> None:
        while True:  # [consider-refactoring-into-while-condition]
            if k <= 1:
                break
            if k > 10:
                break
            k -= 1
```

Pierre-Sassoulas (changes requested):
Yes, this should be an extension, maybe be in the coding style checker :)

doc/data/messages/c/consider-refactoring-into-while-condition/bad.py
```diff
@@ -0,0 +1,8 @@
+fruit_basket = ["apple", "orange", "banana", "cherry", "guava"]
+
+while True:  # [consider-refactoring-into-while-condition]
+  if len(fruit_basket ) == 0:
+    break
```
mbyrnepr2, line 5:
```suggestion
while True:  # [consider-refactoring-into-while-condition]
    if len(fruit_basket) == 0:
        break
```

doc/data/messages/c/consider-refactoring-into-while-condition/good.py
```diff
@@ -0,0 +1,5 @@
+fruit_basket = ["apple", "orange", "banana", "cherry", "guava"]
+
+while len(fruit_basket ) != 0:
+  fruit = fruit_basket.pop()
+  print(f"We removed {fruit} from the basket")
```
mbyrnepr2, line 5:
```suggestion
while len(fruit_basket) != 0:
    fruit = fruit_basket.pop()
    print(f"We removed {fruit} from the basket")
```

doc/whatsnew/fragments/8015.new_check
```diff
@@ -0,0 +1,5 @@
+Add ``consider-refactoring-into-while-condition`` check to recommend refactoring when
+a while loop is defined with a constant condition with an immediate if
+statement to check for break condition as a first statement.
```
mbyrnepr2, line 3:
```suggestion
Add ``consider-refactoring-into-while-condition`` check to recommend refactoring when
a while loop is defined with a constant condition with an immediate ``if``
statement to check for ``break`` condition as a first statement.
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -430,3 +437,36 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
                 line=node.lineno,
                 col_offset=node.col_offset,
             )
+
+    @utils.only_required_for_messages("consider-refactoring-into-while-condition")
+    def visit_while(self, node: nodes.While) -> None:
+        self._check_breaking_after_while_true(node)
+
+    def _check_breaking_after_while_true(self, node: nodes.While) -> None:
+        """Check that any loop with an else clause has a break statement."""
```
mbyrnepr2, line 446:
```suggestion
        """Check that any loop with an `if` clause has a break statement."""
```
I think it is meant to be `if`?

nickdrozd:
This checks pairs nicely with the existing `consider-using-assignment-expr`. Suppose you doing something like reading lines from stdin:

```python
while True:
    line = getline()

    if line is None:
        break

    process(line)
```

Pylint will suggest using an assignment expression:

```python
while True:
    if (line := getline()) is None:
        break

    process(line)
```

Then the new check will in turn suggest inlining the condition:

```python
while (line := getline()) is not None:
    process(line)
```

I think that's a great improvement, and this is a fairly common pattern.

nickdrozd:
Actually, my previous comment glossed over a minor error. On this snippet:

```python
while True:
    if (line := getline()) is None:
        break

    process(line)
```

The suggested change is `while line := getline() is not None`. But this won't work. Parentheses around the assignment expression are required, or else the `line` variable will get an unintended value. So the suggestion should be `while (line := getline()) is not None`.

pylint/checkers/refactoring/recommendation_checker.py
yushao2:
```suggestion
        """Check that any loop with an ``if`` clause has a break statement."""
```

yushao2:
> Actually, my previous comment glossed over a minor error. On this snippet:
> 
> ```python
> while True:
>     if (line := getline()) is None:
>         break
> 
>     process(line)
> ```
> 
> The suggested change is `while line := getline() is not None`. But this won't work. Parentheses around the assignment expression are required, or else the `line` variable will get an unintended value. So the suggestion should be `while (line := getline()) is not None`.

i improved the error messages by adding brackets for expressions with lower precedence, thanks

doc/data/messages/c/consider-refactoring-into-while-condition/bad.py
```diff
@@ -0,0 +1,8 @@
+fruit_basket = ["apple", "orange", "banana", "cherry", "guava"]
+
+while True:  # [consider-refactoring-into-while-condition]
+    if len(fruit_basket) == 0:
+        break
+
+  fruit = fruit_basket.pop()
+  print(f"We removed {fruit} from the basket")
```
yushao2, line 8:
```suggestion
while True:  # [consider-refactoring-into-while-condition]
    if len(fruit_basket) == 0:
        break

    fruit = fruit_basket.pop()
    print(f"We removed {fruit} from the basket")
```

tests/functional/c/consider/consider_refactoring_into_while_condition_py.txt
```diff
@@ -0,0 +1,22 @@
+consider-refactoring-into-while-condition:7:8:10:18:Issue8015.bad:"Consider using 'while k != 10' instead of 'while True:' an 'if', and a 'break'":HIGH
+consider-refactoring-into-while-condition:14:8:17:32:Issue8015.another_bad:"Consider using 'while current_scope is not None' instead of 'while True:' an 'if', and a 'break'":HIGH
+consider-refactoring-into-while-condition:27:8:32:18:Issue8015.test_error_message_multiple_break:"Consider using 'while k > 1' instead of 'while True:' an 'if', and a 'break'":HIGH
```
Pierre-Sassoulas, line 3: Should we detect multiple breaks here and suggest this instead ?

```suggestion
consider-refactoring-into-while-condition:27:8:32:18:Issue8015.test_error_message_multiple_break:"Consider using 'while k > 1 and k <= 10' instead of 'while True:' an 'if', and a 'break'":HIGH
```

nickdrozd:
@jacobtylerwalls What do you think about the suggested changed to the music library?

jacobtylerwalls:
I see how the check would be useful to some, but yes, in that library, I expect that check will just be globally disabled.

> I don't think it's any more opinionated than existing checks, for example no-else-return.

Right, and in that case, that library already disables that warning also. It's good the new message is at the "C" level.

jacobtylerwalls:
Speaking of which, this should probably be an "R" message, but since we can't change the others in this checker, which is already called "Refactoring" for some reason, I get why "C" was chosen. 🤷🏻‍♂️

doc/user_guide/checkers/features.rst
```diff
@@ -879,6 +879,10 @@ Refactoring checker Messages
   Emitted when iterating over the keys of a dictionary and accessing the value
   by index lookup. Both the key and value can be accessed by iterating using
   the .items() method of the dictionary instead.
+:consider-refactoring-into-while-condition (C0210): *Consider using 'while %s' instead of 'while True:' an 'if', and a 'break'*
+  Emitted when `while True:` loop is used and the first statement is a break
+  condition.The `if / break`` construct can be removed if the check is inverted
+  and moved to the `while` statement
```
jacobtylerwalls, line 885:
```suggestion
  condition. The `if / break`` construct can be removed if the check is inverted
  and moved to the `while` statement.
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -60,6 +60,13 @@ class RecommendationChecker(checkers.BaseChecker):
             "which could potentially be a f-string. The use of f-strings is preferred. "
             "Requires Python 3.6 and ``py-version >= 3.6``.",
         ),
+        "C0210": (
+            "Consider using 'while %s' instead of 'while True:' an 'if', and a 'break'",
+            "consider-refactoring-into-while-condition",
+            "Emitted when `while True:` loop is used and the first statement is a break condition."
+            "The ``if / break`` construct can be removed if the check is inverted and moved to "
+            "the ``while`` statement",
```
jacobtylerwalls, line 68:
```suggestion
            "Emitted when `while True:` loop is used and the first statement is a break condition. "
            "The ``if / break`` construct can be removed if the check is inverted and moved to "
            "the ``while`` statement.",
```

jacobtylerwalls (changes requested):
Linting this:
```python
while True:
  if x ** 2:
    break
```
Gives:
```
    ops, rhs = test_node.ops[0]
               ^^^^^^^^^^^^^
AttributeError: 'BinOp' object has no attribute 'ops'
```

Pierre-Sassoulas (changes requested):
It should be an R message once we move it to extension. Also we can tests the message generation on all ifs tests instead of on the while with if/break only to catch more possible crashes with the primers. (Create the helper function and use it in 'visit_if')

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -430,3 +437,53 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
...
+                if isinstance(rhs, lower_priority_expressions)
+                else rhs.as_string()
+            )
+            msg = f"{lhs} {utils.get_inverse_comparator(ops)} {rhs}"
+
+        self.add_message(
+            "consider-refactoring-into-while-condition",
+            node=node,
+            line=node.lineno,
+            args=(msg,),
```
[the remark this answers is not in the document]
yushao2: implemented additional utils function

tests/functional/c/consider/consider_refactoring_into_while_condition_py.txt
yushao2: implemented this

tests/functional/c/consider/consider_refactoring_into_while_condition_py.py
```diff
@@ -0,0 +1,186 @@
...
+            k += 1
+            if k == 10:
+                break
+
+    def test_error_message_multiple_break(self, k: int) -> None:
+        while True:  # [consider-refactoring-into-while-condition]
+            if k <= 1:
+                break
+            if k > 10:
+                break
```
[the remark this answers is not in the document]
yushao2: not handling

jacobtylerwalls (changes requested):
A small thing. Mentioning `while True` is too specific, as it's catching all constants, see the first primer result for music21 with `while 1`.

Also, should we exclude or negate False constants?

```python
while False:
  if x ** 2:
    break
```
```
a.py:1:0: C0210: Consider using 'while not x**2' instead of 'while True:' an 'if', and a 'break' (consider-refactoring-into-while-condition)
```

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -60,6 +60,13 @@ class RecommendationChecker(checkers.BaseChecker):
             "which could potentially be a f-string. The use of f-strings is preferred. "
             "Requires Python 3.6 and ``py-version >= 3.6``.",
         ),
+        "C0210": (
+            "Consider using 'while %s' instead of 'while True:' an 'if', and a 'break'",
+            "consider-refactoring-into-while-condition",
+            "Emitted when `while True:` loop is used and the first statement is a break condition. "
+            "The ``if / break`` construct can be removed if the check is inverted and moved to "
+            "the ``while`` statement.",
+        ),
```
Pierre-Sassoulas, line 69: To make this an extension this part must be moved in [pylint/extensions](https://github.com/PyCQA/pylint/tree/main/pylint/extensions) instead of ``pylint/checkers`` along with the ``visit_while`` function.

Pierre-Sassoulas (changes requested):

nickdrozd:
> Also, should we exclude or negate False constants?

This check should only apply to loops with truthy constants, and not falsy ones.

However, loops with falsy constants should be flagged for a different reason: their bodies can never be executed, and therefore they are pointless. That should be a separate check.

yushao2:
Hi, I've updated the PR and all tests are now passing -- would like a re-review. Sorry I've been on vacation the past 2 weeks

jacobtylerwalls:
Thanks for the updates. My comments in https://github.com/PyCQA/pylint/pull/8021#pullrequestreview-1249284188 are still not reflected. "While True" is too specific if we're handling any truthy constant, and we should exclude falsy constants.

Pierre-Sassoulas (changes requested):
Regarding the django case I don't think our suggestion to replace this:

```python
        while True:
            if escaped:
                result.append(ch)
            elif ch == ".":
                # Replace "any character" with an arbitrary representative.
                result.append(".")
            elif ch == "|":
                # FIXME: One day we'll should do this, but not in 1.0.
                raise NotImplementedError("Awaiting Implementation")
            elif ch == "^":
                pass
            elif ch == "$":
                break
```
By this:
```python
        while ch != "$":
            if escaped:
                result.append(ch)
            elif ch == ".":
                # Replace "any character" with an arbitrary representative.
                result.append(".")
            elif ch == "|":
                # FIXME: One day we'll should do this, but not in 1.0.
                raise NotImplementedError("Awaiting Implementation")
            elif ch == "^":
                pass
```

is equivalent. If ``escaped`` is truethey then the break would not happen at the same time and we would not append ch to result. Concretely there would be a break before the end of the regex if we encounter an escaped terminator (``\$``) in the pattern.

I also agree with the suggestion from @uzzell  (https://github.com/PyCQA/pylint/pull/8021#issuecomment-1372273298) to make the ``not_condition_as_string`` function better. Creating unit tests for multiple use cases would be nice, I think this can be done in a preliminary merge requests because this is not a simple problem to solve. Also comments are getting hidden by github for UX reason in this one which is a problem you should click on load more to see them all:

load_more

yushao2:
> Regarding the django case I don't think our suggestion to replace this:
> 
> ```python
>         while True:
>             if escaped:
>                 result.append(ch)
>             elif ch == ".":
>                 # Replace "any character" with an arbitrary representative.
>                 result.append(".")
>             elif ch == "|":
>                 # FIXME: One day we'll should do this, but not in 1.0.
>                 raise NotImplementedError("Awaiting Implementation")
>             elif ch == "^":
>                 pass
>             elif ch == "$":
>                 break
> ```
> 
> By this:
> 
> ```python
>         while ch != "$":
>             if escaped:
>                 result.append(ch)
>             elif ch == ".":
>                 # Replace "any character" with an arbitrary representative.
>                 result.append(".")
>             elif ch == "|":
>                 # FIXME: One day we'll should do this, but not in 1.0.
>                 raise NotImplementedError("Awaiting Implementation")
>             elif ch == "^":
>                 pass
> ```
> 
> is equivalent. If `escaped` is truethey then the break would not happen at the same time and we would not append ch to result. Concretely there would be a break before the end of the regex if we encounter an escaped terminator (`\$`) in the pattern.
> 
> I also agree with the suggestion from @uzzell ([#8021 (comment)](https://github.com/PyCQA/pylint/pull/8021#issuecomment-1372273298)) to make the `not_condition_as_string` function better. Creating unit tests for multiple use cases would be nice, I think this can be done in a preliminary merge requests because this is not a simple problem to solve. Also comments are getting hidden by github for UX reason in this one which is a problem you should click on load more to see them all:
> 
> load_more

I believe the latest change has dealt with the `not_condition_as_string` and is suggesting the while condition from the abovementioned comment https://github.com/PyCQA/pylint/pull/8021#issuecomment-1372273298

For the issue with django, I believe it's a logic error on my part so I'm working on it

yushao2:
> A small thing. Mentioning `while True` is too specific, as it's catching all constants, see the first primer result for music21 with `while 1`.
> 
> Also, should we exclude or negate False constants?
> 
> ```python
> while False:
>   if x ** 2:
>     break
> ```
> 
> ```
> a.py:1:0: C0210: Consider using 'while not x**2' instead of 'while True:' an 'if', and a 'break' (consider-refactoring-into-while-condition)
> ```

what are some examples of falsey constants? Sorry for missing your earlier comment

jacobtylerwalls:
`False`, `0`, and `[]`, or anything that doesn't evaluate true is a falsy constant. It's unlikely to appear in user code, but pylint is for catching mistakes in code, and if a user has done `while False:`, it would be confusing to them for the check to report "don't write while True"`, since they haven't written that. We should only emit the warning if the bool value of the constant is true, and ideally be more flexible in the message so that it matches the user code (refer to "while 1" like in the music21 example, rather than "while True").

yushao2:
> `False`, `0`, and `[]`, or anything that doesn't evaluate true is a falsy constant. It's unlikely to appear in user code, but pylint is for catching mistakes in code, and if a user has done `while False:`, it would be confusing to them for the check to report "don't write while True"`, since they haven't written that. We should only emit the warning if the bool value of the constant is true, and ideally be more flexible in the message so that it matches the user code (refer to "while 1" like in the music21 example, rather than "while True").

thanks, just updated to handle this by ignoring all falsy values

yushao2:
> `False`, `0`, and `[]`, or anything that doesn't evaluate true is a falsy constant. It's unlikely to appear in user code, but pylint is for catching mistakes in code, and if a user has done `while False:`, it would be confusing to them for the check to report "don't write while True"`, since they haven't written that. We should only emit the warning if the bool value of the constant is true, and ideally be more flexible in the message so that it matches the user code (refer to "while 1" like in the music21 example, rather than "while True").

Also, to the second part of your comment: https://github.com/PyCQA/pylint/pull/8021/commits/ed7412146ed36918c71d21a557841177bdc06733

jacobtylerwalls (approved):

mbyrnepr2 (approved):

pylint/extensions/consider_refactoring_into_while_condition.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @utils.only_required_for_messages("consider-refactoring-into-while-condition")
+    def visit_while(self, node: nodes.While) -> None:
+        self._check_breaking_after_while_true(node)
+
+    def _check_breaking_after_while_true(self, node: nodes.While) -> None:
+        """Check that any loop with an ``if`` clause has a break statement."""
+        if not isinstance(node.test, nodes.Const) or not node.test.bool_value():
+            return
+        pri_candidates = []
```
nickdrozd, line 52:
```suggestion
        pri_candidates: list[nodes.If] = []
```

pylint/extensions/consider_refactoring_into_while_condition.py
```diff
@@ -0,0 +1,93 @@
...
+                orelse = orelse_node.orelse
+
+        candidates = [n for n in candidates if isinstance(n.body[0], nodes.Break)]
+        if len(candidates) == 0:
+            return
+        msg = " and ".join(
+            [f"({utils.not_condition_as_string(c.test)})" for c in candidates]
+        )
+        if len(candidates) == 1:
+            msg = utils.not_condition_as_string(candidates[0].test)
```
nickdrozd, line 81: Please double check, but I'm pretty sure this will be equivalent. But it won't re-assign and re-allocate `candidates`, and there is only one check made. (The formatting is probably wrong.)

```suggestion
        msg = " and ".join(
            [
                f"({utils.not_condition_as_string(c.test)})"
                for c in candidates
                if isinstance(n.body[0], nodes.Break)
            ]
        )

        if not msg:
            return
```

nickdrozd (changes requested):
This is great @yushao2! Much more comprehensive than I initially imagined. Just a few code tweaks.

Also, could you enable the check in the `pylintrc`? When the PR started it was enabled by default, but now as an extension it needs to be enabled explicitly.

pylint/extensions/consider_refactoring_into_while_condition.py
yushao2: this isn't equivalent as it doesn't take c are of the case when there's only a single candidate -- in that case it would not be wrapped with brackets

the error messages would look weird like:
`consider-refactoring-into-while-condition:7:8:10:18:Issue8015.bad:"Consider using 'while (k != 10)' instead of 'while 1:' an 'if', and a 'break'":HIGH`

nickdrozd (dismissed):
Thanks @yushao2!

doc/user_guide/checkers/extensions.rst
```diff
@@ -715,3 +715,19 @@ While Used checker Messages
...
+
+Consider refactoring into while condition checker
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This checker is provided by ``pylint.extensions.consider_refactoring_into_while_condition``.
+Verbatim name of the checker is ``consider_refactoring_into_while_condition``.
+
+Consider refactoring into while checker Messages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+:consider-refactoring-into-while-condition (R3501): *Consider using 'while %s' instead of 'while True:' an 'if', and a 'break'*
```
Pierre-Sassoulas, line 730: This need to be regenerated with latest code using the doc generation command.

pylint/extensions/consider_refactoring_into_while_condition.py
```diff
@@ -0,0 +1,93 @@
...
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class ConsiderRefactorIntoWhileConditionChecker(checkers.BaseChecker):
+    """Checks for instances where while loops are implemented with a constant condition
+    which.
+
+    always evaluates to true and the first statement(s) is/are if statements which, when
```
Pierre-Sassoulas, line 25:
```suggestion
    always evaluates to truethey and the first statement(s) is/are if statements which, when
```

Pierre-Sassoulas (changes requested):
That's a great new check, let's include it in 2.16.0 !

pylint/extensions/consider_refactoring_into_while_condition.py
nickdrozd: truthy

doc/user_guide/checkers/extensions.rst
yushao2: done

pylint/extensions/consider_refactoring_into_while_condition.py
yushao2: done

doc/user_guide/configuration/all-options.rst
```diff
@@ -1258,7 +1258,7 @@ Standard Checkers
 
 --spelling-dict
 """""""""""""""
-*Spelling dictionary name. Available dictionaries: en (aspell), en_AU (aspell), en_CA (aspell), en_GB (aspell), en_US (aspell).*
+*Spelling dictionary name. Available dictionaries: none. To make it work, install the 'python-enchant' package.*
```
yushao2, line 1261:
```suggestion
*Spelling dictionary name. Available dictionaries: en (aspell), en_AU (aspell), en_CA (aspell), en_GB (aspell), en_US (aspell).*```

pylint/extensions/consider_refactoring_into_while_condition.py
yushao2: it's causing an error in CI checks: https://github.com/PyCQA/pylint/actions/runs/4026257164/jobs/6920556064#step:9:50

should i rollback?

Pierre-Sassoulas (approved):

pylint/extensions/consider_refactoring_into_while_condition.py
Pierre-Sassoulas: We should add truthy in the pyenchant dict :)

yushao2: done, sorry i didn't know about pyenchant dict

Pierre-Sassoulas (approved):

Issue #8024, opened 2023-01-05
Support for cython pure python syntax: cython.declare


jtoledo1974:
### Bug description

"""Test for cython pure python cython.declare"""
import cython

MY_VAR = cython.declare(cython.int, 0)

### Configuration

```ini
[testoptions]
requires: cython
```

### Command used

```shell
pylint regression_cython_declare.py
```

### Pylint output

```shell
************* Module regression_cython_declare
tests/functional/r/regression_02/regression_cython_declare.py:4:24: E1101: Module 'cython' has no 'int' member; maybe 'bint'? (no-member)

------------------------------------------------------------------
Your code has been rated at 0.00/10 (previous run: 0.00/10, +0.00)
```

### Expected behavior

------------------------------------
Your code has been rated at 10.00/10

### Pylint version

```shell
pylint 2.16.0-dev
astroid 2.12.13
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
```

### OS / Environment

Needs cython. The default cython package as of today is
Cython==0.29.32

jtoledo1974:
Cython has a "pure python" mode with which you can write your python code using valid python syntax.

I believe pylint should be able to recognize it and avoid reporting correct syntax when using cython. In this case, it is expected that the cython module doesn't provide the members. 

I'll do a pull request with the test case, and will try to produce a patch.

jtoledo1974:
For reference, and if I fail to produce a patch, a workaround is to add this to the pyproject.toml file:

```toml
[tool.pylint.BASIC]
generated-members = "cython.*"
```

Pierre-Sassoulas:
Isn't a problem since [cython/cython@6e31b41f6](https://github.com/cython/cython/commit/6e31b41f6353bd623d98aa4de40d3645965e47ba), but ``cython.declare(cython.int, 0)`` still raises E1101 on main with Cython 0.29.33 through 3.2.9 on latest pylint main.

PR #8025, opened 2023-01-05
Cython pure python declare

jtoledo1974:
## Type of Changes

## Description

This Pull Requests provides a test case for #8024.
A pure python cython.declare(cython.int... is reported as no-member. These should be ignored.

Refs #8024

Closes #XXXX


jtoledo1974:
@Pierre-Sassoulas , I took a look at it. Both pure python cython problems get resolved by ignoring the cython module. I took a look to see whether this would be a problem for the cython developers, but they themselves don't seem to do cimports.

Problem is that just setting a default for ignored-imports doesn't cut it because it would be overridden by any user configuration.

The option reading code is complex and I don't know where in the codebase it is that you would like to have the forced ignored-import added. I'll do it if you give me a hint.

doc/whatsnew/fragments/8024.false_positive
```diff
@@ -0,0 +1,3 @@
+Added test for pure python cython.declare usage, which should be valid no matter what is imported from the cython module.
```
mbyrnepr2, line 1:
```suggestion
Added test for pure Python ``cython.declare`` usage, which should be valid no matter what is imported from the cython module.
```

tests/functional/r/regression_02/regression_cython_declare_8024.py
```diff
@@ -0,0 +1,13 @@
+"""Test for cython pure python false positives"""
```
mbyrnepr2, line 1:
```suggestion
"""Test for cython pure Python false positives"""
```

tests/functional/r/regression_02/regression_cython_declare_8024.py
```diff
@@ -0,0 +1,13 @@
...
+import cython
+
+# cython allows any import in pure python mode. It can only
+# be determined correct at compile time
+# Check that it doesn't raise neither an import-error nor a
```
mbyrnepr2, line 6:
```suggestion
# Check that it raises neither an import-error nor a
```

tests/functional/r/regression_02/regression_cython_declare_8024.py
```diff
@@ -0,0 +1,13 @@
+"""Test for cython pure python false positives"""
+import cython
+
+# cython allows any import in pure python mode. It can only
```
mbyrnepr2, line 4:
```suggestion
# cython allows any import in pure Python mode. It can only
```

DanielNoord (changes requested):
We don't test with `cython` right? So this PR doesn't make a lot of sense in its current state imo.

jtoledo1974:
> We don't test with `cython` right? So this PR doesn't make a lot of sense in its current state imo.

The point of the matter is that this is not just cython code. It is valid python interpreted code. Cython allows to compile it as well, but it shouldn't be flagged as invalid.

DanielNoord:
> > We don't test with `cython` right? So this PR doesn't make a lot of sense in its current state imo.
> 
> The point of the matter is that this is not just cython code. It is valid python interpreted code. Cython allows to compile it as well, but it shouldn't be flagged as invalid.

I don't really understand your point. I was just saying that this test will never run in CI which is why this PR has a successful CI even though there is no fix for the added test.

jtoledo1974:
> > > We don't test with `cython` right? So this PR doesn't make a lot of sense in its current state imo.
> > 
> > 
> > The point of the matter is that this is not just cython code. It is valid python interpreted code. Cython allows to compile it as well, but it shouldn't be flagged as invalid.
> 
> I don't really understand your point. I was just saying that this test will never run in CI which is why this PR has a successful CI even though there is no fix for the added test.

I apologize. I misunderstood and my choice of words was poor.

I was forcing the test to run to make sure that it failed first and then it passed. I would say it was run together with the other tests after a full run, but I can't say for sure. But if CI passes I guess it didn't.

PR #8029, opened 2023-01-07
Fix `used-before-assignment` false positive

### LINKED ISSUE #7779: False positive used-before-assignment with inline if

zenlyj:
## Type of Changes

## Description

After investigation, there seems to be two main cases that trigger used-before-assignment false positives:

1. Ternary expression with walrus operator within attribute calls. Eg. `(foo if (foo:='foo') else 'bar').upper()`
2. Ternary expression with walrus operator nested in function calls. Eg. `str(str(foo if (foo:='foo') else 'bar'))`

This fix enforces a stricter check on whether a name may be assigned and used in the same statement.

Closes #7779

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -2198,10 +2198,3 @@ def is_class_attr(name: str, klass: nodes.ClassDef) -> bool:
         return True
     except astroid.NotFoundError:
         return False
-
-def get_all_calls(node: nodes.NodeNG) -> list[nodes.Call]:
-    """Depth first search to get all call nodes under a node"""
-    calls = [node] if isinstance(node, nodes.Call) else []
-    for child in node.get_children():
-        calls.extend(get_all_calls(child))
-    return calls
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -2198,3 +2198,10 @@ def is_class_attr(name: str, klass: nodes.ClassDef) -> bool:
         return True
     except astroid.NotFoundError:
         return False
+
+def get_all_calls(node: nodes.NodeNG) -> list[nodes.Call]:
+    """Depth first search to get all call nodes under a node"""
+    calls = [node] if isinstance(node, nodes.Call) else []
+    for child in node.get_children():
+        calls.extend(get_all_calls(child))
+    return calls
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2246,14 +2246,17 @@ class VariablesChecker(BaseChecker):
             return True
         if isinstance(value, nodes.Lambda) and isinstance(value.body, nodes.IfExp):
             return True
-        if isinstance(value, nodes.Call):
-            for call in value.nodes_of_class(klass=nodes.Call):
-                if (any(isinstance(kwarg.value, nodes.IfExp) for kwarg in call.keywords)
-                    or any(isinstance(arg, nodes.IfExp) for arg in call.args)
-                    or (isinstance(call.func, nodes.Attribute) and isinstance(call.func.expr, nodes.IfExp))
-                ):
-                    return True
-        return False
+        if not isinstance(value, nodes.Call):
+            return False
+        return any(
+            any(isinstance(kwarg.value, nodes.IfExp) for kwarg in call.keywords)
+            or any(isinstance(arg, nodes.IfExp) for arg in call.args)
+            or (
+                isinstance(call.func, nodes.Attribute)
+                and isinstance(call.func.expr, nodes.IfExp)
+            )
+            for call in value.nodes_of_class(klass=nodes.Call)
+        )
 
     def _is_only_type_assignment(
         self, node: nodes.Name, defstmt: nodes.Statement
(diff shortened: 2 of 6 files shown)


pylint/checkers/variables.py
```diff
@@ -2246,10 +2246,18 @@ def _maybe_used_and_assigned_at_once(defstmt: nodes.Statement) -> bool:
             return True
         if isinstance(value, nodes.Lambda) and isinstance(value.body, nodes.IfExp):
             return True
-        return isinstance(value, nodes.Call) and (
-            any(isinstance(kwarg.value, nodes.IfExp) for kwarg in value.keywords)
-            or any(isinstance(arg, nodes.IfExp) for arg in value.args)
-        )
+        if isinstance(value, nodes.Call):
+            for call in utils.get_all_calls(value):
```
Pierre-Sassoulas, line 2250: We should use ``NodeNG.nodes_of_class(nodes.Call)`` from astroid directly.
```suggestion
            for call in value.nodes_of_class(klass=nodes.Call):
```

pylint/checkers/utils.py
```diff
@@ -2198,3 +2198,11 @@ def is_class_attr(name: str, klass: nodes.ClassDef) -> bool:
...
     except astroid.NotFoundError:
         return False
+
+
+def get_all_calls(node: nodes.NodeNG) -> list[nodes.Call]:
+    """Depth first search to get all call nodes under a node."""
+    calls = [node] if isinstance(node, nodes.Call) else []
+    for child in node.get_children():
+        calls.extend(get_all_calls(child))
+    return calls
```
Pierre-Sassoulas, line 2208:
```suggestion
```

Pierre-Sassoulas (changes requested):

pylint/checkers/variables.py
zenlyj: Thanks for pointing this out!

pylint/checkers/variables.py
```diff
@@ -2246,10 +2246,18 @@ def _maybe_used_and_assigned_at_once(defstmt: nodes.Statement) -> bool:
             return True
         if isinstance(value, nodes.Lambda) and isinstance(value.body, nodes.IfExp):
             return True
```
Pierre-Sassoulas, line 2248:
```suggestion
            return True
        if not isinstance(value, nodes.Call):            
            return False
```

pylint/checkers/variables.py
```diff
@@ -2246,10 +2246,18 @@ def _maybe_used_and_assigned_at_once(defstmt: nodes.Statement) -> bool:
...
+                if (
+                    any(isinstance(kwarg.value, nodes.IfExp) for kwarg in call.keywords)
+                    or any(isinstance(arg, nodes.IfExp) for arg in call.args)
+                    or (
+                        isinstance(call.func, nodes.Attribute)
+                        and isinstance(call.func.expr, nodes.IfExp)
+                    )
+                ):
+                    return True
+        return False
```
Pierre-Sassoulas, line 2260:
```suggestion
        return any(
            any(isinstance(kwarg.value, nodes.IfExp) for kwarg in call.keywords)
            or any(isinstance(arg, nodes.IfExp) for arg in call.args)
            or (
               isinstance(call.func, nodes.Attribute)
               and isinstance(call.func.expr, nodes.IfExp)
           ) 
           for call in value.nodes_of_class(klass=nodes.Call)
       )
```

Pierre-Sassoulas (commented):
Maybe a simplification with early return ?

mbyrnepr2:
@zenlyj the changelog file is required. [There are a number here](https://github.com/PyCQA/pylint/tree/main/doc/whatsnew/fragments) if you need to check how that is done

pylint/checkers/variables.py
zenlyj: Thanks, this looks cleaner 🙂

mbyrnepr2:
@zenlyj The _Python 3.7_ test in the CI is failing because Walrus is a syntax-error. It's possible to run this test for Python versions greater than 3.7. There are many examples in the existing functional tests where a `.rc` file, [similar to this one](https://github.com/PyCQA/pylint/blob/main/tests/functional/o/overridden_final_method_py38.rc), is added with the same file-name as the functional test module, for the tests to be run for the specification in the .rc file.

zenlyj:
@mbyrnepr2 Thank you for helping with the CI fails, very helpful advice.

Pierre-Sassoulas (approved):
Thank you for the fix, the test cases are pretty nice too. This will be available in 2.15.10 :)

PR #8030, opened 2023-01-07
Upgrade astroid to 2.13.2

Pierre-Sassoulas:


Pierre-Sassoulas:
@nelfin look at the primer result, there's a bunch of false positives that were fixed by your implementation support for is None :)

DanielNoord (changes requested):
See https://github.com/PyCQA/astroid/issues/1942

jacobtylerwalls:
I tried to reproduce the psycopg error outside the primer and couldn't (but I was in a rush). I'm not certain the way the primer packages are packaged all together under a single directory is realistic--I remember opening an astroid issue about primer packages polluting each other. I'm still interested to try for a reproducer, but I don't know that I'll have time this weekend.

DanielNoord:
@Pierre-Sassoulas Let's merge then and get some real world examples (if there are any) from the beta.

Pierre-Sassoulas:
Let's release the beta in the wild and see how it fare then :D

DanielNoord (approved):

PR #8065, opened 2023-01-16
[code owner] Remove areveny from her checkers temporarily

Pierre-Sassoulas:
## Type of Changes

## Description
No availability for review right now, so the automated review request is a bit much. It's possible to stay a pylint member without being spammed for it. (@areveny fyi)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -26,15 +26,6 @@ tests/pyreverse/*    @DudeNr33
 
 ### Extensions
 
-# For any all
-pylint/extensions/for_any_all.py @areveny
-tests/functional/ext/for_any_all/* @areveny
-
-# Private import
-pylint/extensions/private_import.py @areveny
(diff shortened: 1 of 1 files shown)


DanielNoord (changes requested):
I think it is better to just remove these then and then revert this commit when necessary.

Pierre-Sassoulas:
I think reverting the commit is harder than removing the ``#`` and I feel like this file is a documentation about who master what in the repositories regardless of who actually get pinged for review but feel free to suggest/change/merge :)

DanielNoord:
> I think reverting the commit is harder than removing the `#` and I feel like this file is a documentation about who master what in the repositories regardless of who actually get pinged for review but feel free to suggest/change/merge :)

Yeah completely agree, but I don't like keeping stuff on `main` that is dependent on people's live. I completely trust that @areveny hopes to have more time in the future, but we never know if that will actually happen. Feels better to handle that when it actually does.

PR #8070, opened 2023-01-17
Fixed a crash when inferring a value and using its qname on a slice

### LINKED ISSUE #8067: Inference crash: `Slice has no attribute 'qname'`
### Bug description

```python
x = slice(42)
x()
```

### Configuration

_No response_

### Command used

```shell
pylint asd.py
```

### Pylint output

```shell
************* Module asd
asd.py:1:0: C0114: Missing module docstring (missing-module-docstring)
Exception on node <Call l.2 at 0x1040c31f0> in file '/private/tmp/asd.py'
  File "/private/tmp/venv3/lib/python3.10/site-packages/pylint/utils/ast_walker.py", line 90, in walk
    callback(astroid)
  File "/private/tmp/venv3/lib/python3.10/site-packages/pylint/checkers/method_args.py", line 72, in visit_call
    and inferred.qname() in self.linter.config.timeout_methods
AttributeError: 'Slice' object has no attribute 'qname'
asd.py:1:0: F0002: asd.py: Fatal error while checking 'asd.py'. Please open an issue in our bug tracker so we address this. There is a pre-filled template that you can use in '/Users/tusharsadhwani/Library/Caches/pylint/pylint-crash-2023-01-17-14-57-43.txt'. (astroid-error)
```

### Expected behavior

No crash.

### Pylint version

```shell
pylint 2.15.10
astroid 2.13.2
Python 3.10.6 (main, Aug 30 2022, 04:58:14) [Clang 13.1.6 (clang-1316.0.21.2.5)]
```

### OS / Environment

MacOS Monterey

### Additional dependencies

_No response_

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #8067

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1611,18 +1611,17 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             # the result of this call was already assigned to a variable and will be checked when leaving the scope.
             return
         inferred = utils.safe_infer(node.func)
-        if not inferred:
+        if not inferred or not isinstance(
+            inferred, (nodes.FunctionDef, nodes.ClassDef, bases.UnboundMethod)
+        ):
             return
         could_be_used_in_with = (
             # things like ``lock.acquire()``
-            hasattr(inferred, "qname")
-            and (
-                inferred.qname() in CALLS_THAT_COULD_BE_REPLACED_BY_WITH
-                or (
-                    # things like ``open("foo")`` which are not already inside a ``with`` statement
-                    inferred.qname() in CALLS_RETURNING_CONTEXT_MANAGERS
-                    and not _is_part_of_with_items(node)
-                )
+            inferred.qname() in CALLS_THAT_COULD_BE_REPLACED_BY_WITH
+            or (
+                # things like ``open("foo")`` which are not already inside a ``with`` statement
+                inferred.qname() in CALLS_RETURNING_CONTEXT_MANAGERS
+                and not _is_part_of_with_items(node)
             )
         )
         if could_be_used_in_with and not _will_be_released_automatically(node):
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1615,11 +1615,14 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             return
         could_be_used_in_with = (
             # things like ``lock.acquire()``
-            inferred.qname() in CALLS_THAT_COULD_BE_REPLACED_BY_WITH
-            or (
-                # things like ``open("foo")`` which are not already inside a ``with`` statement
-                inferred.qname() in CALLS_RETURNING_CONTEXT_MANAGERS
-                and not _is_part_of_with_items(node)
+            hasattr(inferred, "qname")
+            and (
+                inferred.qname() in CALLS_THAT_COULD_BE_REPLACED_BY_WITH
+                or (
+                    # things like ``open("foo")`` which are not already inside a ``with`` statement
+                    inferred.qname() in CALLS_RETURNING_CONTEXT_MANAGERS
+                    and not _is_part_of_with_items(node)
+                )
             )
         )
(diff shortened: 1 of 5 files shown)


DanielNoord (commented):
I don't really like using `hasattr` as it is not exactly type safe. Can't we use `isinstance`?

DanielNoord:
> Maybe we could check that it's a base.Instance ? I've seen that somewhere. Did you have something in mind ?

Purely based on the next line (`self.linter.config.timeout_methods`) I guess we're looking for `FunctionDefs` here?

Pierre-Sassoulas:
Given our current test, it can be 
```
<class 'astroid.nodes.scoped_nodes.scoped_nodes.FunctionDef'>
<class 'astroid.nodes.scoped_nodes.scoped_nodes.ClassDef'>
<class 'astroid.bases.BoundMethod'>
```
(and create false negative according to our tests if we don't handle those cases). What's wrong with hasattr exactly ?

DanielNoord:
> Given our current test, it can be
> 
> ```
> <class 'astroid.nodes.scoped_nodes.scoped_nodes.FunctionDef'>
> <class 'astroid.nodes.scoped_nodes.scoped_nodes.ClassDef'>
> <class 'astroid.bases.BoundMethod'>
> ```
> 
> (and create false negative according to our tests if we don't handle those cases). What's wrong with hasattr exactly ?

```python
class A:
    def qname(self):
        return 1
```

:smile: 

There is nothing that says that `qname` returns a string like we expect it to. I'm also not opposed to adding it to all `nodes` which seems much simpler than tackling these one by one.

jacobtylerwalls:
> I'm also not opposed to adding it to all nodes which seems much simpler than tackling these one by one.

Marc was against that [here](https://github.com/PyCQA/astroid/pull/1284#pullrequestreview-823449216), but it also comes up regularly, see https://github.com/PyCQA/astroid/pull/1373.

Pierre-Sassoulas:
Hmm yeah, right thank you I understand. The problematic version would be this:
```python
class A:
    def qname(self):
        return "threading.lock.acquire"
```
Because then it's a false positive for ``consider-using-with``. Are you proposing to add qname to all node in astroid directly ? I was going to propose that because clearly ``Slice`` has in fact a qname, and it's ``builtin.Slice``, right ?

Pierre-Sassoulas:
Being able to use qname on all nodes would obviously be great in order to not crash when we get unexpected nodes. (Could assignAttr qname be the name of the variable with full namespace ?) But it's a big decisions with a required astroid upgrade, maybe hasattr is a decent temporary fix.

DanielNoord:
Yeah I'm not sure. `hasattr` is just asking for trouble further down the line. When we add `py.typed` to `astroid` all of our incorrect calls to `qname()` for nodes that don't have it will be flagged anyway.
Here it does feel like we could get away with doing a check for `ClassDef`, `FunctionDef` and the "lowest type" of either `UnboundMethod` or `BoundMethod` (I always forget which inherits which).

Pierre-Sassoulas:
Let's the how the primers react then

Pierre-Sassoulas:
There's a whole lot of other call to hasattr accross the codebase (qname, attrname, name). Probably something to do in another refactor merge request or after the astroid qname discussion.

tushar-deepsource:
Hey! This does fix the first crash in the issue, but it doesn't fix the second crash in `RefactoringChecker`, mentioned [here](https://github.com/PyCQA/pylint/issues/8067#issuecomment-1385095725).

Dropping the code snippet here as well:

```python
class Foo:
    def __init__(self, foo=slice(42)):
        self.foo = foo


def bar():
    i = Foo()
    i.foo()
    return 100
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1998,7 +2000,7 @@ def _is_function_def_never_returning(self, node: nodes.FunctionDef) -> bool:
             )
         try:
             return node.qname() in self._never_returning_functions
```
DanielNoord, line 2002: Clearly `node` is typed incorrectly, but oh well :smile:

DanielNoord (approved):

PR #8071, opened 2023-01-17
Fix `used-before-assignment` false positive for TYPE_CHECKING if/elif/else usage

### LINKED ISSUE #7574: False positive `used-before-assignment` when TYPE_CHECKING is used with if/elif/else blocks
### Bug description

Example

```python
# pylint: disable=import-error,missing-function-docstring,invalid-name
"""."""

from typing import TYPE_CHECKING, Optional
from package import runtime


if TYPE_CHECKING:
    from package import module
elif runtime.is_local:
    from package import module
else:
    module = None


def diagnose() -> Optional[module.Result]:
    if module is not None:
        return module.diagnose_result()
    return None
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:16:27: E0601: Using variable 'module' before assignment (used-before-assignment)

Your code has been rated at 5.45/10 (previous run: 5.45/10, +0.00)
```

### Expected behavior

No `used-before-assignment` should be raised.

### Pylint version

```shell
pylint 2.15.3
astroid 2.12.10
Python 3.8.10 (default, Jul  6 2022, 10:42:32)
[Clang 13.0.0 (clang-1300.0.29.30)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

zenlyj:
## Type of Changes

## Description

Extend existing logic to prevent false positive when TYPE_CHECKING is used with if/elif/else blocks.

Closes #7574


jacobtylerwalls (changes requested):
Thanks for the PR!

This still misses one false positive that I think is in scope. Perhaps you would find `_exhaustively_define_name_raise_or_return` or `_defines_name_raises_or_returns_recursive` useful to handle this? ~If so, we can (in a separate commit) promote it out of the NamesConsumer into utils.py.~ (EDIT: no, I wouldn't move it, it's too many related methods and should be discussed separately.)

```diff
diff --git a/tests/functional/u/used/used_before_assignment_typing.py b/tests/functional/u/used/used_before_assignment_typing.py
index 6e1f87131..94d28830b 100644
--- a/tests/functional/u/used/used_before_assignment_typing.py
+++ b/tests/functional/u/used/used_before_assignment_typing.py
@@ -11,7 +11,10 @@ if TYPE_CHECKING:
     import calendar
     from urllib.request import urlopen
 elif input():
-    import calendar
+    if input() + 1:
+        import calendar
+    else:
+        import calendar
 else:
     from urllib.request import urlopen
```

Gives:
```
E       AssertionError: Wrong results for file "used_before_assignment_typing":
E       
E        124: used-before-assignment
```

zenlyj:
@jacobtylerwalls Thanks for the review! I looked into the usage `_exhaustively_define_name_raise_or_return` and `_defines_name_raises_or_returns_recursive`, but I don't think they can be used without making significant changes to them.

I took reference from those 2 and implemented a new function in `utils.py`. Currently, the solution should handle the mentioned nested if/else case, alongside with a few other cases (reflected in the updated functional tests) including try-except-finally blocks and for/while blocks.

pylint/checkers/utils.py
```diff
@@ -2193,6 +2193,60 @@ def is_class_attr(name: str, klass: nodes.ClassDef) -> bool:
...
+    if isinstance(node, nodes.NamedExpr) and node.target.name == name:
+        return True
+
+    if isinstance(node, (nodes.Import, nodes.ImportFrom)) and any(
+        node_name[0] == name for node_name in node.names
+    ):
+        return True
+
+    if isinstance(node, nodes.With):
+        is_defined_so_far = any(
```
jacobtylerwalls, line 2209: For my own understanding, is there a reason to switch from the `return True` pattern?

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -8,7 +8,62 @@
     if True:  # pylint: disable=using-constant-test
         import math
     import datetime
+    import calendar
```
jacobtylerwalls, line 11: [cosmetic thing:] do you mind alphabetizing these?

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -8,7 +8,62 @@
...
+    import mailbox
+    import mimetypes
+    import base64
+    import binascii
     from urllib.request import urlopen
+elif input():
+    import calendar, bisect  # pylint: disable=multiple-imports
+    if input() + 1:
+        import heapq
+    elif (enum:=None):
```
jacobtylerwalls, line 34: I would put spaces around the `:=` here and below.

tests/functional/u/used/used_before_assignment_typing.rc
```diff
@@ -0,0 +1,2 @@
+[testoptions]
+min_pyver=3.8
```
jacobtylerwalls, line 2: This will skip a lot of tests on Py 3.7. An alternative would be to move the walrus operator test into `undefined_variable_py38.py`, which already tests for both `undefined-variable` and `used-before-assignment`, given that they're analyzed with extremely similar code paths.

jacobtylerwalls (dismissed):
This is amazing! Such a useful generalized solution (vastly outstripping the underlying issue in importance!) 🚀 

I left some cosmetic feedback, but this looks good to go.

pylint/checkers/variables.py
```diff
@@ -2036,7 +2036,7 @@ def _in_lambda_or_comprehension_body(
             parent = parent.parent
         return False
 
-    # pylint: disable = too-many-statements, too-many-branches
```
nickdrozd, line 2039: Nice! As the warnings indicate, `_is_variable_violation` is one of the nastiest functions in the codebase, and this is a measurable improvement.

pylint/checkers/variables.py
```diff
@@ -2220,24 +2220,17 @@ def _is_variable_violation(
...
+                            node.name, defstmt_parent.orelse[0]
+                        )
+                    else:
+                        defined_in_or_else = any(
+                            utils.is_defined(node.name, content)
+                            for content in defstmt_parent.orelse
+                        )
 
                     if not used_in_branch and not defined_in_or_else:
                         maybe_before_assign = True
```
nickdrozd, line 2236: It takes some effort to define `defined_in_or_else`, but it looks like the value won't get used if `used_in_branch` is true. Would something like this work?

```suggestion
                if not used_in_branch:
                    if defstmt_parent.has_elif_block():
                        defined_in_or_else = utils.is_defined(
                            node.name, defstmt_parent.orelse[0]
                        )
                    else:
                        defined_in_or_else = any(
                            utils.is_defined(node.name, content)
                            for content in defstmt_parent.orelse
                        )

                    if not defined_in_or_else:
                        maybe_before_assign = True
```

I think that should be equivalent, but the existing logic is very unpleasant and hard to understand, so please double check.

With a little further further shuffling, it might even be possible to get rid of a few more extra branches and eliminate the `too-many-branches` warning.

jacobtylerwalls (approved):

pylint/checkers/utils.py
zenlyj: For checks that do not `return True` immediately, there may be child nodes that define the given name. That is why `is_defined_so_far` is declared, whereby if it is False, recursively search for a child node that defines the given name.

Taking `nodes.With` as an example:
1. is_defined_so_far is True if an opened file is assigned the given name. `with open(...) as ...`
2. If the opened file is not assigned the given name, search the statements within the `with` block.

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: Done!

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: Done!

tests/functional/u/used/used_before_assignment_typing.rc
zenlyj: Thanks for pointing this out!

pylint/checkers/utils.py
jacobtylerwalls: Ah, I didn't look closely at the end of the function. Thanks, got it.

pylint/checkers/variables.py
zenlyj: Great suggestion! Refactored accordingly.

Regarding code readability, `is_variable_violation` is overloaded with logic and probably needs more function abstractions. But that is non-trivial and should probably be addressed in a separate PR.

zenlyj:
@Pierre-Sassoulas this PR should be ready for final review, whenever you are free. Thank you!

Pierre-Sassoulas (approved):
Amazing, thank you !

PR #8075, opened 2023-01-18
pointless-exception-statement: performance: filter to call nodes that reference named functions

### LINKED ISSUE #8073: Performance regression: `pointless-exception-statement`

```
before: 7:59
```

jayaddison:
Adds an additional preconditional check to the `pointless-exception-statement` message logic so that `astroid`'s inference functionality is called only when the identified `Call` node includes a [named](https://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.Name) function reference.

Some comparative (local, single-machine, limited sample-size) benchmark performance testing against the Home Assistant `core.git` repository (at commit https://github.com/home-assistant/core/commit/a44e44b7d01c2a10e0b84e22fbd2b20012024159):

```
pylint.git $ git checkout 89a20e2ccdb2927ac370f7c6663fc83318988e10
HEAD is now at 89a20e2cc Add new check "pointless-exception-statement" (#7939)

pylint.git $ time find home-assistant/core.git/ -name '*.py' -exec python3 -m pylint -j2 --output /dev/null {} +

real    12m33.726s
user    23m25.588s
sys     0m47.887s

pylint.git $ rm -rf ~/.cache/pylint/
pylint.git $ git checkout 56121344f6c9c223d8e4477c0cdb58c46c840b4b
HEAD is now at 56121344f [pre-commit.ci] pre-commit autoupdate (#8016)

pylint.git $ time find home-assistant/core.git/ -name '*.py' -exec python3 -m pylint -j2 --output /dev/null {} +

real    11m59.741s
user    22m58.198s
sys     0m46.425s

pylint.git $ rm -rf ~/.cache/pylint/
pylint.git $ git checkout 29d39b66d7b7c17b076638e46772472e9a5c73fc
HEAD is now at 29d39b66d pointless-exception-statement: performance: filter to call nodes that reference named functions

pylint.git $ time find .venv/lib/core/ -name '*.py' -exec python3 -m pylint -j2 --output /dev/null {} +

real	12m9.042s
user	23m3.343s
sys	0m46.171s
```

## Type of Changes

## Description

Refs #7939.

May resolve #8073.

### DIFF
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -452,7 +452,7 @@ class BasicChecker(_BasicChecker):
             return
 
         # Warn W0133 for exceptions that are used as statements
-        if isinstance(expr, nodes.Call):
+        if isinstance(expr, nodes.Call) and isinstance(expr.func, nodes.Name):
             inferred = utils.safe_infer(expr)
             if isinstance(inferred, objects.ExceptionInstance):
                 self.add_message(


pylint/checkers/base/basic_checker.py
```diff
@@ -452,7 +452,7 @@ def visit_expr(self, node: nodes.Expr) -> None:
             return
 
         # Warn W0133 for exceptions that are used as statements
-        if isinstance(expr, nodes.Call):
+        if isinstance(expr, nodes.Call) and isinstance(expr.func, nodes.Name):
```
jayaddison, line 455: I'm having second thoughts about this additional filter.  Perhaps it does improve performance.. but perhaps there are legitimate situations where a `Call` node is _not_ a named function can be inferred to be an `Exception` type here.  That situation could be worth identifying.  Could test coverage be added for it?  What would an example be?

jayaddison:
This isn't my preferred approach; more investigation is required to identify the cause of the performance issues reported in #8073.  Closing.

pylint/checkers/base/basic_checker.py
jayaddison: From discussion in the linked issue: `Attribute` is another valid case here, in addition to `Name`.

Those two classes correspond to `ast.Name` and `ast.Attribute` in Python, and, according to the [current upstream Py3 docs for `ast.Call`](https://docs.python.org/3/library/ast.html#ast.Call):

> `func` is the function, which will often be a [Name](https://docs.python.org/3/library/ast.html#ast.Name) or [Attribute](https://docs.python.org/3/library/ast.html#ast.Attribute) object.

So it's not a strong guarantee, exactly, but handling both `Name` and `Attribute` would catch most cases.

My guess at the moment, however, is that this change _would not_ fix any performance problems -- because it's adding additional boolean checks (a performance penalty) and in most cases those will evaluate to `True` (again, referencing the Py3 docs) - so we won't be able to skip the inference step -- where we think the cost is -- in many cases.

PR #8077, opened 2023-01-18
[doc] Add details and related links for bare/broad-except

Pierre-Sassoulas:
## Type of Changes

## Description

All messages do not have a good/bad example yet but the next step is giving the rationale. This one was easy to do. I feel like the rationale are lacking right now we have a lot of "emitted when ..." type message, but **when** it's emitted is / should be shown by the example, what we really need is a rationale for **why** the message exists in pylint. There's a lot of work if we want to change this part of the message other the whole code base.

### DIFF
--- a/doc/data/messages/b/bare-except/bad.py
+++ b/doc/data/messages/b/bare-except/bad.py
@@ -2,9 +2,3 @@ try:
     import platform_specific_module
 except:  # [bare-except]
     platform_specific_module = None
-
-
-try:
-    1 / 0
-except:  # [bare-except]
-    pass
--- a/doc/data/messages/b/bare-except/bad.py
+++ b/doc/data/messages/b/bare-except/bad.py
@@ -1,4 +1,10 @@
 try:
+    import platform_specific_module
+except:  # [bare-except]
+    platform_specific_module = None
+
+
+try:
     1 / 0
 except:  # [bare-except]
     pass
--- a/doc/data/messages/b/broad-exception-caught/bad.py
+++ b/doc/data/messages/b/broad-exception-caught/bad.py
@@ -1,4 +1,10 @@
 try:
+    import platform_specific_module
+except Exception:  # [broad-exception-caught]
+    platform_specific_module = None
+
+
+try:
     1 / 0
 except Exception:  # [broad-exception-caught]
     pass
--- a/doc/data/messages/b/broad-exception-caught/bad.py
+++ b/doc/data/messages/b/broad-exception-caught/bad.py
@@ -2,9 +2,3 @@ try:
     import platform_specific_module
 except Exception:  # [broad-exception-caught]
     platform_specific_module = None
-
-
-try:
-    1 / 0
-except Exception:  # [broad-exception-caught]
-    pass
--- a/doc/data/messages/b/broad-exception-caught/details.rst
+++ b/doc/data/messages/b/broad-exception-caught/details.rst
@@ -1,0 +1,6 @@
+If you use a naked ``except Exception:`` clause, you might end up catching exceptions other than the ones you expect to catch
+This can hide bugs or make it harder to debug programs when they aren't doing what you expect.
+
+For example, you're trying to import a library with required system dependencies and you catch
+everything instead of only import errors, you will miss the error message telling you, that
+your code could work if you had installed the system dependencies.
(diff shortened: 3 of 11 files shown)


doc/data/messages/b/broad-exception-caught/details.rst
```diff
@@ -0,0 +1,6 @@
+If you use a naked ``except:`` clause, you might end up catching exceptions other than the ones you expect to catch
```
Pierre-Sassoulas, line 1:
```suggestion
If you use a naked ``except Exception:`` clause, you might end up catching exceptions other than the ones you expect to catch
```

doc/data/messages/b/broad-exception-caught/bad.py
```diff
@@ -1,3 +1,9 @@
+try:
+    import platform_specific_module
+except Exception:  # [broad-except]
```
Pierre-Sassoulas, line 3:
```suggestion
except Exception:  # [broad-exception-caught]
```

doc/data/messages/b/bare-except/bad.py
```diff
@@ -1,3 +1,9 @@
+try:
```
DanielNoord, line 1: Why the added example?

pylint/checkers/exceptions.py
```diff
@@ -112,13 +112,18 @@ def _is_raising(body: list[nodes.NodeNG]) -> bool:
     "W0702": (
         "No exception type(s) specified",
         "bare-except",
-        "Used when an except clause doesn't specify exceptions type to catch.",
+        "A bare ``except:`` clause will catch ``SystemExit`` and ``KeyboardInterrupt``"
+        " exceptions, making it harder to interrupt a program with ``Control-C``, and "
```
DanielNoord, line 116: I don't really like the mixing of spaces at the end of the string and at the beginning of the string. Could you make that consistent?

pylint/checkers/exceptions.py
```diff
@@ -168,7 +173,11 @@ def _is_raising(body: list[nodes.NodeNG]) -> bool:
     "W0719": (
         "Raising too general exception: %s",
         "broad-exception-raised",
-        "Used when an except raises a too general exception.",
+        "Raising exceptions that are too generic force you to catch exception "
```
DanielNoord, line 176:
```suggestion
        "Raising exceptions that are too generic force you to catch exceptions "
```

DanielNoord (commented):
Couple of nits! This is indeed much better though!

Pierre-Sassoulas:
I thougth about this some more, this field is called description . Might be a better Idea to add a rational field and not change everything.

DanielNoord:
> I thougth about this some more, this field is called description . Might be a better Idea to add a rational field and not change everything.

I like the added detail in the description but I think we can make do with just one example and an accompanying test.

clavedeluna (dismissed):
Great clarification

pylint/checkers/exceptions.py
```diff
@@ -168,7 +173,11 @@ def _is_raising(body: list[nodes.NodeNG]) -> bool:
     "W0719": (
         "Raising too general exception: %s",
         "broad-exception-raised",
-        "Used when an except raises a too general exception.",
+        "Raising exceptions that are too generic force you to catch exceptions "
+        "generically too. It will force you to use a naked ``except Exception:``"
+        " clause. You might then end up catching exceptions other than the ones "
+        "you expect to catch. This can hide bugs or make it harder to debug programs"
+        " when unrelated errors are hidden.",
```
DanielNoord, line 180:
```suggestion
        "Raising exceptions that are too generic force you to catch exceptions "
        "generically too. It will force you to use a naked ``except Exception:`` "
        "clause. You might then end up catching exceptions other than the ones "
        "you expect to catch. This can hide bugs or make it harder to debug programs "
        "when unrelated errors are hidden. =",
```

DanielNoord (approved):

pylint/checkers/exceptions.py
Pierre-Sassoulas: Can't apply suggestion because there's a '=' at the end. Let's merge, we have bigger issues to fix :)

DanielNoord: I'll edit

DanielNoord (approved):

PR #8078, opened 2023-01-18
pointless-exception-statement: filter call inference to names that begin with an uppercase character

### LINKED ISSUE #8073: Performance regression: `pointless-exception-statement`

jayaddison:
## Type of Changes

## Description
This changeset addresses a performance regression identified in #8073, and does so by limiting the use of `astroid` inference added in #7939 to situations where the call-expression's name begins with an uppercase character.

Refs #7939.

May resolve #8073.

### DIFF
--- a/doc/whatsnew/fragments/8073.internal
+++ b/doc/whatsnew/fragments/8073.internal
@@ -1,0 +1,3 @@
+Use string-pattern-matching instead of ``astroid`` inference to implement ``pointless-exception-statement``.
+
+Refs #8073
--- a/doc/whatsnew/fragments/8073.internal
+++ b/doc/whatsnew/fragments/8073.internal
@@ -1,3 +1,3 @@
-Use string-pattern-matching instead of ``astroid`` inference to implement ``pointless-exception-statement``.
+Use string-pattern-matching as a precondition for ``astroid`` inference in ``pointless-exception-statement``.
 
 Refs #8073
--- a/doc/whatsnew/fragments/8073.internal
+++ b/doc/whatsnew/fragments/8073.internal
@@ -1,3 +1,0 @@
-Use string-pattern-matching as a precondition for ``astroid`` inference in ``pointless-exception-statement``.
-
-Refs #8073
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -453,11 +453,10 @@ class BasicChecker(_BasicChecker):
 
         # Warn W0133 for exceptions that are used as statements
         if isinstance(expr, nodes.Call):
-            inferred = utils.safe_infer(expr)
-            if isinstance(inferred, objects.ExceptionInstance):
-                self.add_message(
-                    "pointless-exception-statement", node=node, confidence=INFERENCE
-                )
+            name, _, _ = expr.as_string().partition("(")
+            exc_suffixes = ("Error", "Exc", "Exception", "Warning")
+            if any(name.endswith(suffix) for suffix in exc_suffixes):
+                self.add_message("pointless-exception-statement", node=node)
 
         # Ignore if this is :
         # * a direct function call
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -13,7 +13,7 @@ from collections.abc import Iterator
 from typing import TYPE_CHECKING, cast
 
 import astroid
-from astroid import nodes
+from astroid import nodes, objects
 
 from pylint import utils as lint_utils
 from pylint.checkers import BaseChecker, utils
@@ -456,7 +456,11 @@ class BasicChecker(_BasicChecker):
             name, _, _ = expr.as_string().partition("(")
             exc_suffixes = ("Error", "Exc", "Exception", "Warning")
             if any(name.endswith(suffix) for suffix in exc_suffixes):
-                self.add_message("pointless-exception-statement", node=node)
+                inferred = utils.safe_infer(expr)
+                if isinstance(inferred, objects.ExceptionInstance):
+                    self.add_message(
+                        "pointless-exception-statement", node=node, confidence=INFERENCE
+                    )
 
         # Ignore if this is :
         # * a direct function call
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -11,6 +11,7 @@ import itertools
 import sys
 from collections.abc import Iterator
 from typing import TYPE_CHECKING, cast
+import warnings
 
 import astroid
 from astroid import nodes, objects
@@ -23,6 +24,11 @@ from pylint.utils import LinterStats
 
 if TYPE_CHECKING:
     from pylint.lint.pylinter import PyLinter
+
+if sys.version_info >= (3, 8):
+    from functools import cached_property
+else:
+    from astroid.decorators import cachedproperty as cached_property
 
 if sys.version_info >= (3, 8):
     from typing import Literal
@@ -421,6 +427,10 @@ class BasicChecker(_BasicChecker):
         """
         self.linter.stats.node_count["klass"] += 1
 
+    @cached_property
+    def _heuristic_exception_detection(self) -> bool:
+        return bool(self.linter.config.heuristic_exception_detection)
+
     @utils.only_required_for_messages(
         "pointless-statement",
         "pointless-exception-statement",
@@ -453,14 +463,24 @@ class BasicChecker(_BasicChecker):
 
         # Warn W0133 for exceptions that are used as statements
         if isinstance(expr, nodes.Call):
-            name, _, _ = expr.as_string().partition("(")
-            exc_suffixes = ("Error", "Exc", "Exception", "Warning")
-            if any(name.endswith(suffix) for suffix in exc_suffixes):
-                inferred = utils.safe_infer(expr)
-                if isinstance(inferred, objects.ExceptionInstance):
-                    self.add_message(
-                        "pointless-exception-statement", node=node, confidence=INFERENCE
+            inferred = None
+            if self._heuristic_exception_detection is True:
+                name, _, _ = expr.as_string().partition("(")
+                exc_suffixes = ("Error", "Exc", "Exception", "Warning")
+                if any(name.endswith(suffix) for suffix in exc_suffixes):
+                    inferred = utils.safe_infer(expr)
+                else:
+                    skipped = expr.as_string()
+                    warnings.warn(
+                        f"Skipped W0133 check for {skipped!r} because heuristic exception detection is enabled",
+                        RuntimeWarning,
                     )
+            else:
+                inferred = utils.safe_infer(expr)
+            if isinstance(inferred, objects.ExceptionInstance):
+                self.add_message(
+                    "pointless-exception-statement", node=node, confidence=INFERENCE
+                )
 
         # Ignore if this is :
         # * a direct function call
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 import collections
 import itertools
 import sys
-import warnings
 from collections.abc import Iterator
 from typing import TYPE_CHECKING, cast
 
@@ -469,12 +468,6 @@ class BasicChecker(_BasicChecker):
                 exc_suffixes = ("Error", "Exc", "Exception", "Warning")
                 if any(name.endswith(suffix) for suffix in exc_suffixes):
                     inferred = utils.safe_infer(expr)
-                else:
-                    skipped = expr.as_string()
-                    warnings.warn(
-                        f"Skipped W0133 check for {skipped!r} because heuristic exception detection is enabled",
-                        RuntimeWarning,
-                    )
             else:
                 inferred = utils.safe_infer(expr)
             if isinstance(inferred, objects.ExceptionInstance):
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -460,6 +460,8 @@ class BasicChecker(_BasicChecker):
                 name = expr.func.attrname
 
             # Heuristic: only run inference for names that begin with an uppercase char
+            # This reduces W0133's coverage, but retains acceptable runtime performance
+            # For more details, see: https://github.com/PyCQA/pylint/issues/8073
             inferred = utils.safe_infer(expr) if name[:1].isupper() else None
             if isinstance(inferred, objects.ExceptionInstance):
                 self.add_message(
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -467,16 +467,16 @@ class BasicChecker(_BasicChecker):
                 self.add_message(
                     "pointless-exception-statement", node=node, confidence=INFERENCE
                 )
+            return
 
         # Ignore if this is :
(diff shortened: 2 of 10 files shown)


jayaddison:
The `Primer / Run / Run / 3.7` job seemed to be near completion and is/was beginning to write report results to the filesystem, but that has taken a long time and I wonder if it is stuck somehow.

jayaddison:
Hmm.  The false positives in Django aren't great.  I suppose that's one example of where inference can be useful, to provide a more-precise check in cases like those.

Supporting data would be useful around the performance tradeoff: my sense is that the suffix-checks should achieve a decently large improvement - so it may be OK to tentatively restore the use of inference (in _combination_ with string pattern matching, instead of replacing it).  But the truth of the matter is that I don't have a good way to evaluate that tradeoff at the moment.

Pierre-Sassoulas:
I think inferring only at the end after most condition were checked is nice. I'd rather pylint be a little slower than have the django false positive. Also it seem to me that it's really slower only for very big codebase where maybe increasing the cache size in astroid would be the solution.

doc/whatsnew/fragments/8073.internal
```diff
@@ -0,0 +1,3 @@
+Use string-pattern-matching as a precondition for ``astroid`` inference in ``pointless-exception-statement``.
+
+Refs #8073
```
cdce8p, line 3: The check isn't released yet, so I don't think we necessarily need a changelog entry for this PR. @Pierre-Sassoulas What do you think?

pylint/checkers/base/basic_checker.py
```diff
@@ -453,11 +453,14 @@ def visit_expr(self, node: nodes.Expr) -> None:
 
         # Warn W0133 for exceptions that are used as statements
         if isinstance(expr, nodes.Call):
-            inferred = utils.safe_infer(expr)
-            if isinstance(inferred, objects.ExceptionInstance):
-                self.add_message(
-                    "pointless-exception-statement", node=node, confidence=INFERENCE
-                )
+            name, _, _ = expr.as_string().partition("(")
```
cdce8p, line 456:
```suggestion
            name = expr.call.as_string()
```
That might work too.

pylint/checkers/base/basic_checker.py
```diff
@@ -453,11 +453,14 @@ def visit_expr(self, node: nodes.Expr) -> None:
...
+            exc_suffixes = ("Error", "Exc", "Exception", "Warning")
+            if any(name.endswith(suffix) for suffix in exc_suffixes):
```
cdce8p, line 458: I would like a regex filter here. It's common that these are suffixes but not a guarantee. E.g. `ExceptionGroup`

Something like `.*(Error|Exc|Exception|Warning).*`.

The regex should also be precompiled at module level to improve performance.

--
_If added, user defined regexes could be checked here too then._

cdce8p (changes requested):
We should also find a solution for the remaining stdlib exceptions at least which aren't covered yet: `KeyboardInterrupt`, `StopAsyncIteration`, `StopIteration`, `GeneratorExit`, `SystemExit`.

A solution here could be to add an `or` clause which checks if the name is part of a predefined (at module level) set of exception names.

doc/whatsnew/fragments/8073.internal
Pierre-Sassoulas: Yeap, a changelog would actually be confusing.

pylint/checkers/base/basic_checker.py
jayaddison: Do you have a use case in mind that requires regexes?  Most of this could be achieved by changing from a `name.endswith(suffix)` pattern to `substring in name`.

cdce8p: > Most of this could be achieved by changing from a `name.endswith(suffix)` pattern to `substring in name`.

Good point! That will work too. Just thought about combining it with the potential user input.

pylint/lint/base_options.py
```diff
@@ -391,6 +391,18 @@ def _make_linter_options(linter: PyLinter) -> Options:
...
+            {
+                "default": False,
+                "type": "yn",
+                "metavar": "<y or n>",
+                "help": "Use faster (but less accurate) heuristic checks as a "
+                "precondition for analysis of code that may contain Python "
+                "exceptions.  Disabling this may be useful to improve linting "
+                "performance on large codebases.",
+            },
+        ),
```
cdce8p, line 405: This should be moved to `basic_checker.py`.

pylint/checkers/base/basic_checker.py
```diff
@@ -421,6 +427,10 @@ def visit_classdef(self, _: nodes.ClassDef) -> None:
         """
         self.linter.stats.node_count["klass"] += 1
 
+    @cached_property
+    def _heuristic_exception_detection(self) -> bool:
+        return bool(self.linter.config.heuristic_exception_detection)
```
cdce8p, line 432: This can be defined as instance variable in `def open`.

pylint/checkers/base/basic_checker.py
```diff
@@ -453,7 +463,20 @@ def visit_expr(self, node: nodes.Expr) -> None:
...
+                name, _, _ = expr.as_string().partition("(")
+                exc_suffixes = ("Error", "Exc", "Exception", "Warning")
+                if any(name.endswith(suffix) for suffix in exc_suffixes):
+                    inferred = utils.safe_infer(expr)
+                else:
+                    skipped = expr.as_string()
+                    warnings.warn(
+                        f"Skipped W0133 check for {skipped!r} because heuristic exception detection is enabled",
+                        RuntimeWarning,
+                    )
```
cdce8p, line 477: That will be emitted for every instance and every run. It'll probably just spam the user. I'd recommend we remove it.

pylint/checkers/base/basic_checker.py
```diff
@@ -453,7 +463,20 @@ def visit_expr(self, node: nodes.Expr) -> None:
 
         # Warn W0133 for exceptions that are used as statements
         if isinstance(expr, nodes.Call):
-            inferred = utils.safe_infer(expr)
+            inferred = None
+            if self._heuristic_exception_detection is True:
+                name, _, _ = expr.as_string().partition("(")
+                exc_suffixes = ("Error", "Exc", "Exception", "Warning")
+                if any(name.endswith(suffix) for suffix in exc_suffixes):
+                    inferred = utils.safe_infer(expr)
```
cdce8p, line 471: Could you try the regex filter for PascalCase here?
```
[^\W\da-z][^\W_]+$
```

cdce8p (changes requested):
With https://github.com/PyCQA/pylint/issues/8073#issuecomment-1397664994, it's probably best to move forward with this one.

Added a few more comments. You you also take a look at the unresolved comments from my last review?

pylint/checkers/base/basic_checker.py
jayaddison: Possibly needs a leading caret (`^`) character?

(otherwise matches `laFoobar`, for example)

jayaddison: Attribute-accesses would also need handling, I think (`Foo.Bar`).

cdce8p: > Possibly needs a leading caret (`^`) character?
> 
> (otherwise matches `laFoobar`, for example)

You can use `re.Match`. That always starts at the beginning of the string.
https://docs.python.org/3/library/re.html#re.match

cdce8p: > Attribute-accesses would also need handling, I think (`Foo.Bar`).

This should work
```py
CLASS_NAME_RGX.match(
    expr.func.as_string().rsplit(".", maxsplit=1)[-1]
)
```

jayaddison: That's helpful, thank you - I'm reading up on how `CLASS_NAME_RGX` (and `PascalCaseStyle`) are used throughout the `pylint` codebase.

jayaddison: To be honest I feel like 'call name starts with an uppercase character' would achieve good-enough heuristic filtering to allow inference to proceed - but until I'm able to replicate the performance degradation, I don't have much confidence that I'll be able to confirm that the issue is resolved.

jayaddison:
> until I'm able to replicate the performance degradation, I don't have much confidence that I'll be able to confirm that the issue is resolved

I should probably eat my hat after making this statement :/

It's possible to look at the history of the successful Primer Run workflow jobs in GitHub actions, and I think that that provides a fairly good initial indication of improvement/degradation.

The latest changes here seem to have brought that back down to a more reasonable timescale, at least for the Python 3.11 version of the job.  I'll leave these changes open for consideration (although it's still perfectly ok by me to revert #7939 if doubt remains).

pylint/checkers/base/basic_checker.py
```diff
@@ -453,7 +453,14 @@ def visit_expr(self, node: nodes.Expr) -> None:
...
         if isinstance(expr, nodes.Call):
-            inferred = utils.safe_infer(expr)
+            name = ""
+            if isinstance(expr.func, nodes.Name):
+                name = expr.func.name
+            elif isinstance(expr.func, nodes.Attribute):
+                name = expr.func.attrname
+
+            # Heuristic: only run inference for names that begin with an uppercase char
+            inferred = utils.safe_infer(expr) if name[:1].isupper() else None
```
cdce8p, line 463: Could you add a small note regarding the performance issue here.
Just so we know later why we added it.

tests/functional/s/statement_without_effect.py
```diff
@@ -38,6 +38,10 @@ def to_be():
 class ClassLevelAttributeTest:
     """ test attribute docstrings. """
 
+    class ClassLevelException(Exception):
+        """ exception defined for access as a class attribute """
+        ...
```
cdce8p, line 43:
```suggestion
        """Exception defined for access as a class attribute."""
```

tests/functional/s/statement_without_effect.py
```diff
@@ -90,3 +94,7 @@ def unraised_exception():
     """Test that instantiating but not raising an exception is flagged as a pointless statement"""
     ValueError("pointless-statement")  # [pointless-exception-statement]
     ValueError(to_be())  # [pointless-exception-statement]
+    ClassLevelAttributeTest.ClassLevelException(to_be())  # [pointless-exception-statement]
+    ValueError ( "another-pointless-statement" ) # [pointless-exception-statement]
```
cdce8p, line 98:
```suggestion
    ValueError("another-pointless-statement")  # [pointless-exception-statement]
```

cdce8p (commented):
> > until I'm able to replicate the performance degradation, I don't have much confidence that I'll be able to confirm that the issue is resolved
> 
> I should probably eat my hat after making this statement :/
> 
> It's possible to look at the history of the [successful Primer Run workflow jobs in GitHub Actions](https://github.com/PyCQA/pylint/actions/workflows/primer_run_pr.yaml?query=is%3Asuccess), and I think that that provides a fairly good initial indication of improvement/degradation.
> 
> The latest changes here seem to have brought that back down to a more reasonable timescale, at least for the Python 3.11 version of the job. I'll leave these changes open for consideration (although it's still perfectly ok by me to revert #7939 if doubt remains).

Interesting 🤔

Personally, I like your current approach of checking for an upper case character. That should resolve it. Running the benchmarks now. Will post an update later or tomorrow.

Left a few styling comments, but aside from that I would be ok with merging it.
Reverting the original PR doesn't seem necessary if the heuristic is good enough. Am I missing something?

cdce8p:
> Running the benchmarks now. Will post an update later or tomorrow.

Checking Home Assistant, the time is back down to 8min. Roughly the same time it takes without the initial change which added the check. So all good!

jayaddison:
Great news, thank you @cdce8p!  Especially for having patience while I experimented with the changes and for providing benchmarking tests.

I'll apply your cleanup suggestions in a few moments (apart from one, where the unusual-but-valid code style is intentional).

tests/functional/s/statement_without_effect.py
jayaddison: The code here does look strange, but this was somewhat intentional test coverage to ensure that any string-contents checks used now or in future (like `expr.as_string().endswith(...)`) consider the case that the call name may not be directly adjacent to the parentheses.

cdce8p:
> Great news, thank you @cdce8p! Especially for having patience while I experimented with the changes and for providing benchmarking tests.

The result is much better for everyone. As we now know, my solution wouldn't have worked.
Thank you @jayaddison for working through it with us!

tests/functional/s/statement_without_effect.py
cdce8p: Had to check, but the spaces are automatically stripped by the Python parser when the code is converted to the AST.

pylint/checkers/base/basic_checker.py
```diff
@@ -453,7 +453,16 @@ def visit_expr(self, node: nodes.Expr) -> None:
...
+            name = ""
+            if isinstance(expr.func, nodes.Name):
+                name = expr.func.name
+            elif isinstance(expr.func, nodes.Attribute):
+                name = expr.func.attrname
+
+            # Heuristic: only run inference for names that begin with an uppercase char
+            # This reduces W0133's coverage, but retains acceptable runtime performance
+            # For more details, see: https://github.com/PyCQA/pylint/issues/8073
+            inferred = utils.safe_infer(expr) if name[:1].isupper() else None
```
Pierre-Sassoulas, line 465: This is pretty clever in term of filtering, I would not have bet on this being good enough to improve performance, but I guess it makes sense as it remove all the function call when pep8 is followed.

Pierre-Sassoulas (dismissed):
Amazing work on really going in depth about this ! And this was the last thing we needed to release 2.16.0 👌

cdce8p (dismissed):
Pushed a last small formatting change as it doesn't increase test coverage.

Looks good from my view, although I'm getting some mixed results with Github Actions lately. _Sometimes I still get +10min runs, just not sure they are a result of pylint and not Github Actions being slow. A usual run now is 8:20 (compared to 7:55 before)._

Anyway, it's a clear improvement over the current implementation and probably good enough.

jayaddison:
Thanks @cdce8p @Pierre-Sassoulas.

One further small performance optimization that we could attempt would be to `return` early after the `W0133` checks here (regardless of whether a pointless exception is found) -- and then we could remove the corresponding `Call` check here: https://github.com/PyCQA/pylint/blob/361a1f7533ccfcec58eb10fb517cc5b278fec411/pylint/checkers/base/basic_checker.py#L470

I don't think that that would make much difference, performance-wise.  But it's another possibility to add.

cdce8p:
> One further small performance optimization that we could attempt would be to `return` early after the `W0133` checks here (regardless of whether a pointless exception is found) -- and then we could remove the corresponding `Call` check here:
> 
> https://github.com/PyCQA/pylint/blob/361a1f7533ccfcec58eb10fb517cc5b278fec411/pylint/checkers/base/basic_checker.py#L470
> 
> I don't think that that would make much difference, performance-wise. But it's another possibility to add.

I agree, this likely won't have a performance impact. However it would be nice code style wise. Makes the code flow it a bit easier to follow. Would you like to change it?

jayaddison:
Sure, let's give it a try.  Pushing a commit with those changes in a few moments.

Pierre-Sassoulas (approved):

cdce8p (approved):

cdce8p:
Thanks @jayaddison 👍🏻

PR #8084, opened 2023-01-19
Improve performance for `pointless-exception-statement`

### LINKED ISSUE #8073: Performance regression: `pointless-exception-statement`

cdce8p:
## Description
Improve performance for `pointless-exception-statement` especially for large code bases.
Use 3 step approach outline in https://github.com/PyCQA/pylint/pull/7939#discussion_r1072865819 to catch almost all issues without the negative performance impact while still providing an option for users to check more / everything if they use a custom exception name pattern.

1. Check common names. Almost all exceptions contain one the these words: `Exception`, `Error`, or `Warning`.
2. Check not yet covered stdlib exceptions. https://docs.python.org/3/library/exceptions.html#exception-hierarchy
_If a 3rd party library has some other names, those could be added here as well. Although I would limit it to fairly popular ones, something like `requests` for example._
3. Allow users to define additional regex names to check.

Alternative to #8078
Closes #8073

### DIFF
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import collections
 import itertools
+import re
 import sys
 from collections.abc import Iterator
 from typing import TYPE_CHECKING, cast
@@ -60,6 +61,16 @@ DEFAULT_ARGUMENT_SYMBOLS = dict(
             "collections.UserList",
         )
     },
+)
+
+ADDITIONAL_EXCEPTION_NAMES: frozenset[str] = frozenset(
+    (
+        "GeneratorExit",
+        "KeyboardInterrupt",
+        "SystemExit",
+        "StopIteration",
+        "StopAsyncIteration",
+    )
 )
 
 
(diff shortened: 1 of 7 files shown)


Pierre-Sassoulas (approved):
Looks pretty good, what do you think @jayaddison ?  (We need to add you as co-author if we merge this imo)

pylint/checkers/base/basic_checker.py
```diff
@@ -265,6 +276,18 @@ class BasicChecker(_BasicChecker):
...
     }
+    options = (
+        (
+            "additional-exception-classes",
+            {
+                "default": [],
+                "type": "regexp_csv",
+                "metavar": "<pattern>[,<pattern>...]",
+                "help": "List of regular expressions of additional "
+                "exception names to check for 'pointless-exception-statement",
```
cdce8p, line 287: I'm open to suggestions if someone has a better name for it

jayaddison:
@cdce8p why is the default to perform non-thorough exception matching, with thorough matching only enabled by adding `.*` or similar as a pattern string?

I think that smaller codebases are likely to be the more common case (and it's more important for them to build good foundations as they grow), so we should provide for that - but allowing large codebases to reduce the thoroughness of checks (if they choose to) for performance reasons seems fine, too.

cdce8p:
Closing this one as #8078 is the better approach.
For details see: https://github.com/PyCQA/pylint/issues/8073#issuecomment-1397664994

PR #8088, opened 2023-01-20
Fix `consider-using-augmented-assign` nested attribute access

### LINKED ISSUE #8086: consider-augmented-assign false positive
### Bug description

Currently this checker doesn't appear to pay attention to the context of a field lookup and assumes that as long as the field name matches it can proceed, example:

```python
class A:
    def __init__(self) -> None:
        self.a = 1
        self.b = A()

    def test(self) -> None:
        self.a = self.a + 1 # should flag
        self.b.a = self.a + 1 # should NOT flag, the error is here
```

### Configuration

```ini
Appears to be not relevant as long as the offending rule is enabled
```

### Command used

Please don't scream, we need to move things around (suggestions welcome either way).

```shell
`find . -type f -name "*.py" -not -path "./.env/*" -not -path "*/src/common/*" | xargs pylint --jobs 4 --init-hook "import sys; sys.path.append(\"$PWD\")"`
```

### Pylint output

```shell
`Use '+=' to do an augmented assign directlypylint(consider-using-augmented-assign)`
```

### Expected behavior

The false positive should not be reported

### Pylint version

```shell
pylint 2.16.0b0
astroid 2.13.2
Python 3.10.8 (main, Oct 13 2022, 09:48:40) [Clang 14.0.0 (clang-1400.0.29.102)]
```

### OS / Environment

MacOS

### Additional dependencies

_No response_

cdce8p:
## Description

Closes #8086

### DIFF
--- a/doc/whatsnew/fragments/8086.false_positive
+++ b/doc/whatsnew/fragments/8086.false_positive
@@ -1,3 +1,0 @@
-Fix ``consider-using-augmented-assign`` checker for nested attribute access.
-
-Closes #8086
--- a/doc/whatsnew/fragments/8086.false_positive
+++ b/doc/whatsnew/fragments/8086.false_positive
@@ -1,0 +1,3 @@
+Fix ``consider-using-augmented-assign`` checker for nested attribute access.
+
+Closes #8086
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -2066,22 +2066,6 @@ def is_hashable(node: nodes.NodeNG) -> bool:
         return True
 
 
-def get_full_name_of_attribute(node: nodes.Attribute | nodes.AssignAttr) -> str:
-    """Return the full name of an attribute and the classes it belongs to.
-
-    For example: "Class1.Class2.attr"
-    """
-    parent = node.parent
-    ret = node.attrname or ""
-    while isinstance(parent, (nodes.Attribute, nodes.Name)):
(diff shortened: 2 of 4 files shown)


pylint/checkers/utils.py
```diff
@@ -2091,7 +2075,7 @@ def _is_target_name_in_binop_side(
             return target.name == side.name  # type: ignore[no-any-return]
         return False
     if isinstance(side, nodes.Attribute) and isinstance(target, nodes.AssignAttr):
-        return get_full_name_of_attribute(target) == get_full_name_of_attribute(side)
+        return target.as_string() == side.as_string()  # type: ignore[no-any-return]
```
Pierre-Sassoulas, line 2078: Well that's clever ! And probably faster too.

Pierre-Sassoulas (approved):
👌

doc/whatsnew/fragments/8086.false_positive
```diff
@@ -0,0 +1,3 @@
+Fix ``consider-using-augmented-assign`` checker for nested attribute access.
+
+Closes #8086
```
Pierre-Sassoulas, line 3: Nvm, could you remove this fragment as the checker was not released yet ?

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas:
We messed up when reading the primer in previous merge request 😅

Pierre-Sassoulas (approved):

PR #8089, opened 2023-01-20
Fix `invalid-name` for `typing_extensions.TypeVar`

cdce8p:
## Description
`typing_extensions` recently added it's own version of `TypeVar` and no longer aliases `typing.TypeVar`. Adjust name check for that.
https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-440-october-6-2022


Pierre-Sassoulas (approved):
👌

Issue #7957, opened 2022-12-17
Enable the use of preferred-modules with submodules


d33bs:
### Current problem

When using the [`prefered-modules` configuration extension](https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#preferred-modules) there doesn't appear to be a way to specify submodules. Adding this capability would allow for greater specificity with this functionality and potentially avoid custom checker implementations.

### Desired solution

For example, if we wanted to encourage the use of `pathlib` instead of `os.path` for filesystem operations without triggering warnings for `os` more broadly, one could use something like the following configuration line within their `.pylintrc` file:

```ini
preferred-modules=(os.path,pathlib)
```

e-gebes:
I also wanted to use the feature for "deprecating" os.path in favor of pathlib.

I found out that there *is* a warning, but only when the import is `import os.path`, and not only `import os`. So, when a package has an *implicit* package/submodule import, it will not work reliably because in the code base there are often just the bare `import os` statements. There is apparently no checking of `os.path.<func>` usage, only checking of the import statement.

The same is with `deprecated-modules=os.path`, which works in my case only when the import statement imports that module *explicitly*.

An improvement for the feature would need to check usages, I don't know if this is feasible.

PR #8093, opened 2023-01-21
Add test for existing preferred-modules functionality

d33bs:
## Type of Changes

## Description

The work in this PR adds a simple test for existing preferred-modules functionality.

Background: While exploring solutions for #7957, I noticed there was not yet a specific test for this configuration option. As a result, I opted to add this simple test and decouple the change from new functionality I hope to contribute later towards more fully addressing #7957.

Apologies in advance if I somehow missed the existing work in this area (please don't hesitate to let me know!).

Refs #7957

Thank you for any thoughts, suggestions, or edits you may have as part of the review!


tests/checkers/unittest_imports.py
```diff
@@ -114,3 +114,26 @@ def test_wildcard_import_non_init(self) -> None:
         )
         with self.assertAddsMessages(msg):
             self.checker.visit_importfrom(import_from)
+
+    @staticmethod
+    def test_preferred_module(capsys: CaptureFixture[str]) -> None:
+        """
+        Tests preferred-module configuration option
+        """
+
```
DanielNoord, line 123:
```suggestion
```

tests/checkers/unittest_imports.py
```diff
@@ -114,3 +114,26 @@ def test_wildcard_import_non_init(self) -> None:
...
+                "-e preferred-module",
+                # prefer sys instead of os (for triggering test)
+                "--preferred-modules=os:sys",
+            ],
+            exit=False,
+        )
+        output, errors = capsys.readouterr()
+
+        # assert that we saw preferred-modules triggered
+        assert len(output.split("\n")) == 7
```
DanielNoord, line 138: Could you assert for the message here? If we ever change the output format this will unnecessarily fail.

DanielNoord (changes requested):
Thanks! PRs like these are much appreciated!

tests/checkers/unittest_imports.py
d33bs: Thank you for the feedback here! I've made a change to this effect to address this in [4781b0b](https://github.com/PyCQA/pylint/pull/8093/commits/4781b0b96c26865c5a18d6281a01be37c78c95e4).

d33bs:
Thank you @DanielNoord for the review and suggestions! I've made some changes based on what you mentioned. Please don't hesitate to let me know if there is anything else we may address.

DanielNoord (approved):
Thanks! Love it!

PR #8095, opened 2023-01-22
Fix FP for `used-before-assignment` when reimporting name used in type annotation

### LINKED ISSUE #7882: False positive from used-before-assignment when variable is reimported
### Bug description

```python

if TYPE_CHECKING:
    from foo import Bar

def myfunc():
    result: Bar

    from foo import Bar # if I remove this, then no error is reported

```

Note how if I remove the `from foo import Bar` line, then the error goes away

### Command used

```shell
pylint --disable=all --enable=used-before-assignment t.py
```

### Pylint output

```shell
************* Module t
t.py:7:12: E0601: Using variable 'Bar' before assignment (used-before-assignment)

Your code has been rated at 1.67/10 (previous run: 1.67/10, +0.00)
```

### Expected behavior

No error reported

### Pylint version

```shell
pylint 2.15.7
astroid 2.12.13
Python 3.8.15 (default, Oct 12 2022, 19:15:16) 
[GCC 11.2.0]
```

jacobtylerwalls:
## Type of Changes

## Description
Closes #7882

### DIFF
--- a/doc/whatsnew/fragments/7882.false_positive
+++ b/doc/whatsnew/fragments/7882.false_positive
@@ -1,0 +1,4 @@
+Fix a false positive for ``used-before-assignment`` after a variable
+annotated with a name guarded by ``if TYPE_CHECKING:``.
+
+Closes #7882
--- a/doc/whatsnew/fragments/7882.false_positive
+++ b/doc/whatsnew/fragments/7882.false_positive
@@ -1,4 +1,3 @@
-Fix a false positive for ``used-before-assignment`` after a variable
-annotated with a name guarded by ``if TYPE_CHECKING:``.
+Fix a false positive for ``used-before-assignment`` when a name guarded by ``if TYPE_CHECKING:`` is used as a type annotation in a function body and later re-imported in the same scope.
 
 Closes #7882
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1798,10 +1798,14 @@ class VariablesChecker(BaseChecker):
             elif base_scope_type != "lambda":
                 # E0601 may *not* occurs in lambda scope.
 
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas (approved):
👌

doc/whatsnew/fragments/7882.false_positive
```diff
@@ -0,0 +1,4 @@
+Fix a false positive for ``used-before-assignment`` after a variable
+annotated with a name guarded by ``if TYPE_CHECKING:``.
```
mbyrnepr2, line 2:
```suggestion
Fix a false positive for ``used-before-assignment`` when a variable
is annotated with a name guarded by ``if TYPE_CHECKING:``.
```

jacobtylerwalls: Unfortunately, that version sounds more like #7609, which was already solved. This bug doesn't happen until _after_ the variable has already been annotated. I wasn't sure how best to describe it succinctly. Open to suggestions.

mbyrnepr2: Aha I hear you. How about this or something similar?
```suggestion
Fix a false positive for ``used-before-assignment`` when a name guarded by ``if TYPE_CHECKING:`` is used as a type annotation in a function body and later re-imported in the same scope.
```

clavedeluna (commented):
Would it make sense to add the issue's exact example as a test case?

jacobtylerwalls:
> Would it make sense to add the issue's exact example as a test case?

Open to different views here. I generally think that if we can fold it into the existing cases, it reduces bloat. I verified that this change fails before the fix. But I also understand the viewpoint that original cases should be added (as long as sufficiently small).

mbyrnepr2 (approved):

PR #8096, opened 2023-01-22
Add tests to improve coverage

clavedeluna:
## Type of Changes

## Description

This is PR with small code coverage increase. Each commit is a separate issue handling of either adding a test to increase coverage or remove code that's clearly unused.

### DIFF
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -234,9 +234,7 @@ class BaseChecker(_ArgumentsProvider):
             for msgid, msg_tuple in sorted(self.msgs.items())
         ]
 
-    def get_message_definition(
-        self, msgid: str
-    ) -> MessageDefinition:
+    def get_message_definition(self, msgid: str) -> MessageDefinition:
         # TODO: 3.0: Remove deprecated method
         warnings.warn(
             "'get_message_definition' is deprecated and will be removed in 3.0.",
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -234,7 +234,15 @@ class BaseChecker(_ArgumentsProvider):
             for msgid, msg_tuple in sorted(self.msgs.items())
         ]
 
-    def get_message_definition(self, msgid: str) -> MessageDefinition:
+    def get_message_definition(
+        self, msgid: str
+    ) -> MessageDefinition:  # pragma: no cover
+        # TODO: 3.0: Remove deprecated method
+        warnings.warn(
+            "'get_message_definition' is deprecated and will be removed in 3.0.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         for message_definition in self.messages:
             if message_definition.msgid == msgid:
                 return message_definition
--- a/pylint/checkers/base_checker.py
+++ b/pylint/checkers/base_checker.py
@@ -236,7 +236,7 @@ class BaseChecker(_ArgumentsProvider):
 
     def get_message_definition(
         self, msgid: str
-    ) -> MessageDefinition:  # pragma: no cover
+    ) -> MessageDefinition:
         # TODO: 3.0: Remove deprecated method
         warnings.warn(
             "'get_message_definition' is deprecated and will be removed in 3.0.",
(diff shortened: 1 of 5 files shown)


pylint/checkers/classes/special_methods_checker.py
```diff
@@ -296,9 +296,6 @@ def _is_dict(node: InferenceResult) -> bool:
 
     @staticmethod
     def _is_iterator(node: InferenceResult) -> bool:
-        if node is astroid.Uninferable:
-            # Just ignore Uninferable objects.
-            return True
         if isinstance(node, bases.Generator):
```
Pierre-Sassoulas, line 299: This kind of safety should not be removed just because we can't cover it. It can be something we added after a crash where someone opened an issue and someone made a hotfix (so lots of work came into it). It's hard to know when exactly astroid is going to not be able to infer something and it's very hard to prove we can remove this kind of thing safely, let's play it safe if we're not sure.

Pierre-Sassoulas (changes requested):

pylint/checkers/base_checker.py
```diff
@@ -234,14 +234,6 @@ def messages(self) -> list[MessageDefinition]:
             for msgid, msg_tuple in sorted(self.msgs.items())
         ]
 
-    def get_message_definition(self, msgid: str) -> MessageDefinition:
```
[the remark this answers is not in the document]
clavedeluna: makes perfect sense, I can happily do either, which do you prefer?

pylint/checkers/classes/special_methods_checker.py
clavedeluna: This PR is still in draft so I didn't get a chance to explain why I did this. If you inspect the code closely, we're never going to hit this. This is because when `_check_iter` is called within `visit_functiondef`, we already do a truthy check for

```
if (
            inferred
...
```

I checked and `astroid. Uninferable` is falsy. So I think we either keep the removal as I have it or remove the `if inferred` check. This is why it wasn't tested.

pylint/checkers/base_checker.py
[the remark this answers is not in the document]
Pierre-Sassoulas: Let's deprecate if another maintainer agree with me :)

pylint/checkers/typecheck.py
```diff
@@ -499,16 +499,7 @@ def _emit_no_member(
...
-    if (
-        owner.parent
-        and isinstance(owner.parent, nodes.ClassDef)
-        and owner.parent.name == "EnumMeta"
-        and owner_name == "__members__"
-        and node.attrname in {"items", "values", "keys"}
-    ):
-        # Avoid false positive on Enum.__members__.{items(), values, keys}
-        # See https://github.com/PyCQA/pylint/issues/4123
-        return False
```
clavedeluna, line 511: I saw ln 511 was uncovered and I investigated further. Turns out new code additions after this was added make it that this case is no longer necessary as the code short-circuits early. This is proven by the fact that functional tests under member_checks such as `for itm in Animal.__members__.items():` still pass.

pylint/checkers/base_checker.py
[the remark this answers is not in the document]
DanielNoord: Yeah we can't remove this sadly. Let's deprecate!

pylint/checkers/base_checker.py
```diff
@@ -234,7 +234,15 @@ def messages(self) -> list[MessageDefinition]:
...
+    def get_message_definition(
+        self, msgid: str
+    ) -> MessageDefinition:  # pragma: no cover
```
DanielNoord, line 239: I'd prefer if we actually wrote a test for this or just keep it uncovered. There is no good reason this isn't covered other than we don't have a test for it.

We have similar test for deprecated stuff as well.

Pierre-Sassoulas, line 239:
```suggestion
    ) -> MessageDefinition:
```

Pierre-Sassoulas (commented):
Agree with Daniel's last comment and let's merge :)

Pierre-Sassoulas (approved):

PR #8103, opened 2023-01-23
Fix ``invalid-sequence-index`` for call to ``scipy.fft.rfft``

### LINKED ISSUE #8018: False positive `invalid-sequence-index` with `scipy.fft.rfft`
### Bug description

# pylint: disable=missing-module-docstring, missing-function-docstring
# Using scipy.fft.rfft causes the error, removing this line also removes the error
# Seems related to issue #7702

import numpy as np
from scipy.fft import rfft

def some_func():
    arr = np.array([[1, 2, 3], [4, 5, 6]])
    arr_fft = rfft(arr)

    print(arr_fft[:, 0:2])

### Configuration

_No response_

### Command used

```shell
pylint false_positive_pylint.py
```

### Pylint output

```shell
************* Module false_positive_pylint
false_positive_pylint.py:11:10: E1126: Sequence index is not an int, slice, or instance with __index__ (invalid-sequence-index)

Your code has been rated at 3.75/10 (previous run: 3.75/10, +0.00)
```

### Expected behavior

No error reported

### Pylint version

```shell
pylint 2.15.9
astroid 2.12.13
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)]
```

### OS / Environment

Windows 10 using Powershell

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description

Fix ``invalid-sequence-index`` for call to ``scipy.fft.rfft``.

Closes #8018


clavedeluna:
I'd like feedback from maintainers on where I should add `scipy` as a test dependency

Pierre-Sassoulas (commented):
I think it's better to pinpoint the exact code construct that create the problem in scipy and then use that directly in the test. We won't have to install scipy, and the test will still be meaningful if  the scipy code is refactored and the problematic code construct disappear.

clavedeluna:
> I think it's better to pinpoint the exact code construct that create the problem in scipy and then use that directly in the test. We won't have to install scipy, and the test will still be meaningful if the scipy code is refactored and the problematic code construct disappear.

That makes sense, but I think I'll still need to at least test with `numpy` since the behavior happens when the objects are numpy arrays. I see some functional tests that do use numpy, but I don't see where numpy is installed??

Pierre-Sassoulas:
There are import from numpy but it's tested without installing numpy (it's simply an example using numpy). We would need a numpy astroid plugin if something is specific to numpy. We might need to discuss how we test c extension generically by providing the C code ourselves to be able to follow the same logic than m'y previous comment for c extensions.

clavedeluna:
I've pushed the latest code I have but will block this until maintainers let me know about the testing strategy for this type of thing.

DanielNoord:
I agree with Pierre that this is not how we would want to solve this.

PR #8105, opened 2023-01-24
Add new check "implicit-flag-alias"

### LINKED ISSUE #8102: Feature request: implicit-flag-overlap
### Current problem

The following behaviour seemed confusing:

```py
>>> from enum import IntFlag
>>> class TestFlag(IntFlag):
...     A = 1
...     B = 2
...     C = 3
... 
>>> print([i for i in TestFlag])
[<TestFlag.A: 1>, <TestFlag.B: 2>]
```

(note that this is an `IntFlag`, not an `IntEnum`)

### Desired solution

The reason this happens is that `Flag` types are designed for setwise operations (bitwise operations, in the case of `IntFlag`), and, because `C` can be composed of `A` and `B`...

```py
>>> int(TestFlag.A | TestFlag.B)
3
>>> int(TestFlag.C)
3
```

...internally, `IntFlag` considers `C` to be an alias -- and doesn't include it when iterating over the members of the flag.

A `pylint` message something along the lines of `implicit-flag-overlap` could be emitted by a relevant checker when scanning the class declaration.

### Additional context

_No response_

jayaddison:
## Type of Changes

## Description

This changeset implements an `implicit-flag-alias` check that inspects classes derived from `enum.Flag` and looks for member values that share common bit positions.

The commented sample code below may help explain:

```py
from enum import IntFlag

class Test(IntFlag):

    # valid cases:

    A = 1
    B = 2
    C = 2          # C is an alias (alternative name) for B, and is not highlighted by the check
    D = A | B      # D shares bit positions with A and B, but uses bitwise-operator syntax and is not highlighted

    # implicit-flag-alias cases:

    E = 3          # The value of E shares bit positions with A and B, so the check will highlight it
```

Closes #8102.

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -895,21 +895,19 @@ a metaclass class method.",
             )
 
         if ancestor.is_subtype_of("enum.IntFlag"):
-            value_nodes, bit_values = {}, defaultdict(set)
-            for assign_name in node.nodes_of_class(nodes.AssignName):
-                if not isinstance(assign_name.parent, nodes.Assign):
-                    continue  # Ignore non-assignment expressions
+            value_nodes = {
+                assign_name.parent.value.value: assign_name
+                for assign_name in node.nodes_of_class(nodes.AssignName)
+                if isinstance(assign_name.parent, nodes.Assign)
+                and isinstance(getattr(assign_name.parent.value, "value", None), int)
+            }
 
-                assigned = assign_name.parent.value
-                if not isinstance(getattr(assigned, "value", None), int):
-                    continue  # Ignore non-integer assignments
-
-                value_nodes[assigned.value] = assign_name
-
+            bit_values = defaultdict(set)
+            for value in value_nodes.keys():
                 # Create a mapping from bit-index to value(s)
-                for bit, char in enumerate(reversed(f"{assigned.value:b}")):
+                for bit, char in enumerate(reversed(f"{value:b}")):
                     if char == "1":
-                        bit_values[bit].add(assigned.value)
+                        bit_values[bit].add(value)
 
             # For each set of overlapping flags, pair min-and-max overlaps
             overlaps = {
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -910,10 +910,9 @@ a metaclass class method.",
             # Collect the minimum, unique values that each flag overlaps with
             overlap_sources = defaultdict(list)
             for flag_values in bit_flags.values():
-                if len(flag_values) > 1:
-                    source_value, *overlap_values = sorted(flag_values)
-                    for overlap_value in overlap_values:
-                        overlap_sources[overlap_value].append(source_value)
+                source_value, *overlap_values = sorted(flag_values)
+                for overlap_value in overlap_values:
+                    overlap_sources[overlap_value].append(source_value)
 
             def _flag_value_repr(value):
                 return f"<{node.name}.{assignments[value].name}: {value}>"
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -525,6 +525,12 @@ MSGS: dict[str, MessageDefinitionTuple] = {
         "beginning with an underscore) is access outside the class or a "
         "descendant of the class where it's defined.",
     ),
+    "W0213": (
+        "Value of flag member overlaps with another",
+        "implicit-flag-overlap",
+        "Used when a value declared on a class derived from enum.Flag partially "
+        "overlaps with another member value.",
+    ),
     "E0211": (
         "Method %r has no argument",
         "no-method-argument",
@@ -856,6 +862,7 @@ a metaclass class method.",
         "redefined-slots-in-subclass",
         "invalid-enum-extension",
         "subclassed-final-class",
+        "implicit-flag-overlap",
     )
     def visit_classdef(self, node: nodes.ClassDef) -> None:
         """Init visit variable _accessed."""
@@ -903,6 +910,33 @@ a metaclass class method.",
                         node=node,
                         confidence=INFERENCE,
                     )
+
+            if isinstance(ancestor, nodes.ClassDef) and ancestor.is_subtype_of(
+                "enum.Flag"
+            ):
+                previous_values, total, union, overlaps = set(), None, None, []
+                for assign_name in node.nodes_of_class(nodes.AssignName):
+                    if isinstance(assign_name.parent, nodes.Arguments):
+                        continue  # Ignore function arguments
+
+                    assigned = assign_name.parent.value
+                    if not isinstance(assigned, nodes.Const):
+                        continue  # Ignore non-literal assignments
+
+                    assert assigned.value is not None
+                    if assigned.value in previous_values:
+                        continue  # Ignore aliases
+                    previous_values.add(assigned.value)
+
+                    # Detect divergence between the sum-of-values and union-of-values
+                    total = assigned.value if total is None else total + assigned.value
+                    union = assigned.value if union is None else union | assigned.value
+                    if total != union:
+                        overlaps.append(assign_name)
+                        total = union  # reset divergence detection after each iteration
+
+                for overlap in overlaps:
+                    self.add_message("implicit-flag-overlap", node=overlap)
 
             if ancestor.name == object.__name__:
                 self.add_message(
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -910,29 +910,29 @@ a metaclass class method.",
                     if bit == "1":
                         bit_flags[position].append(flag_value)
 
-            # When multiple flags overlap on a bit position, pair the min-and-max
-            overlap_groups = (
-                sorted(flag_values)
-                for flag_values in bit_flags.values()
-                if len(flag_values) > 1
-            )
+            # When multiple flags overlap on a bit position, store the minimum as source
+            overlap_sources = {}
+            for flag_values in bit_flags.values():
+                if len(flag_values) > 1:
+                    source_value, *overlap_values = sorted(flag_values)
+                    for overlap_value in overlap_values:
+                        overlap_sources[overlap_value] = source_value
 
             # Report the overlapping value pairs
-            for source_value, *overlap_values in overlap_groups:
-                for overlap_value in overlap_values:
-                    self.add_message(
-                        "implicit-flag-alias",
-                        node=assignments[overlap_value],
-                        args={
-                            "class": node.name,
-                            "overlap": assignments[overlap_value].name,
-                            "overlap_value": overlap_value,
-                            "source": assignments[source_value].name,
-                            "source_value": source_value,
-                            "intersection_value": overlap_value & source_value,
-                        },
-                        confidence=INFERENCE,
-                    )
+            for overlap_value, source_value in overlap_sources.items():
+                self.add_message(
+                    "implicit-flag-alias",
+                    node=assignments[overlap_value],
+                    args={
+                        "class": node.name,
+                        "overlap": assignments[overlap_value].name,
+                        "overlap_value": overlap_value,
+                        "source": assignments[source_value].name,
+                        "source_value": source_value,
+                        "intersection_value": overlap_value & source_value,
+                    },
+                    confidence=INFERENCE,
+                )
 
     def _check_proper_bases(self, node: nodes.ClassDef) -> None:
         """Detect that a class inherits something which is not
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -910,7 +910,7 @@ a metaclass class method.",
             # Collect the minimum, unique values that each flag overlaps with
             overlap_sources = defaultdict(list)
             for flag_values in bit_flags.values():
-                source_value, *overlap_values = sorted(flag_values)
+                source_value, *overlap_values = flag_values
                 for overlap_value in overlap_values:
                     overlap_sources[overlap_value].append(source_value)
 
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -526,7 +526,7 @@ MSGS: dict[str, MessageDefinitionTuple] = {
         "descendant of the class where it's defined.",
     ),
     "W0213": (
-        "Flag member %s shares bit positions with %s",
+        "Flag member %(overlap)s shares bit positions with %(sources)s",
         "implicit-flag-alias",
         "Used when multiple integer values declared within an enum.IntFlag "
         "class share a common bit position.",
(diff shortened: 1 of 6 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -904,6 +911,33 @@ def _check_proper_bases(self, node: nodes.ClassDef) -> None:
...
+                    if not isinstance(assigned, nodes.Const):
+                        continue  # Ignore non-literal assignments
+
+                    assert assigned.value is not None
+                    if assigned.value in previous_values:
+                        continue  # Ignore aliases
+                    previous_values.add(assigned.value)
+
+                    # Detect divergence between the sum-of-values and union-of-values
+                    total = assigned.value if total is None else total + assigned.value
```
jayaddison, line 932: Note: there's an assumption here that subclasses of `enum.Flag` will implement addition/sum operations - I don't know whether that's reasonable (the check could be adjusted to only inspect subclasses of `enum.IntFlag`).

pylint/checkers/classes/class_checker.py
```diff
@@ -904,6 +911,33 @@ def _check_proper_bases(self, node: nodes.ClassDef) -> None:
                         confidence=INFERENCE,
                     )
 
+            if isinstance(ancestor, nodes.ClassDef) and ancestor.is_subtype_of(
```
clavedeluna, line 914: `visit_classdef` is quite bloated already, would you mind moving the new code to a `check_flag` (better name maybe) method?

pylint/checkers/classes/class_checker.py
```diff
@@ -904,6 +911,33 @@ def _check_proper_bases(self, node: nodes.ClassDef) -> None:
...
+
+                    # Detect divergence between the sum-of-values and union-of-values
+                    total = assigned.value if total is None else total + assigned.value
+                    union = assigned.value if union is None else union | assigned.value
+                    if total != union:
+                        overlaps.append(assign_name)
+                        total = union  # reset divergence detection after each iteration
+
+                for overlap in overlaps:
+                    self.add_message("implicit-flag-overlap", node=overlap)
```
clavedeluna, line 939: add_message should also have `confidence=INFERENCE` (and update functional test .txt output accordingly)

pylint/checkers/classes/class_checker.py
jayaddison: Would it be OK to relocate the `enum.Enum` code at the same time (`_check_enum_bases`, or similar)?  Reasoning: it might be possible to de-duplicate some of the conditional logic after doing that.

clavedeluna: yeah for sure!

mbyrnepr2:
I feel the naming, especially the word “overlap” isn’t intuitive. I believe we are talking about warning when a flag member alias is defined, so what about flag-member-alias-defined, or  something similar, as a more direct message?

jayaddison:
Thanks for the feedback @mbyrnepr2 - this check _excludes_ aliases (if `A = 1` and `B = 1` are two members defined in an enum, no warning will be emitted for either of them).

mbyrnepr2:
> Thanks for the feedback @mbyrnepr2 - this check _excludes_ aliases (if `A = 1` and `B = 1` are two members defined in an enum, no warning will be emitted for either of them).

I was referring to the initial code example in the issue and in the functional test where an IntFlag member is excluded from the iteration over the Enum when it can be the result of a bitwise or of two other members.

pylint/checkers/classes/class_checker.py
jayaddison: I'm experimenting with this a bit locally; it's making me wonder whether it'd be possible (and more efficient) to use the existing tree `visit_` handlers to achieve the same check.

jayaddison:
> where an IntFlag member is excluded from the iteration over the Enum when it can be the result of a bitwise or of two other members

Ok, my apologies - now I understand, I think - my use of the term _alias_ has been out-of-alignment with the documentation (in `A = 1, B = 2, C = 3`, `C` is considered an alias).

Previously I thought that _alias_ referred purely to secondary attribute names for a single value (as in my `A = 1, B = 1` example).

Spending some time to think about renaming the check.. initially `implicit-flag-alias` springs to mind (trying to keep the name concise and combining your correct use of the terminology).

pylint/checkers/classes/class_checker.py
```diff
@@ -874,6 +881,43 @@ def _check_consistent_mro(self, node: nodes.ClassDef) -> None:
...
+                # Detect divergence between the sum-of-values and union-of-values
+                total = assigned.value if total is None else total + assigned.value
+                union = assigned.value if union is None else union | assigned.value
+                if total != union:
+                    overlaps.append(assign_name)
+                    total = union  # reset divergence detection after each iteration
+
+            for overlap in overlaps:
+                self.add_message(
+                    "implicit-flag-overlap", node=overlap, confidence=INFERENCE
```
DanielNoord, line 918: I think it would be really nice to use the `args` argument of `add_message` to tell the user a little bit about which flag is overlapping (its name) and potentially also with which other flag.

DanielNoord (commented):
Sorry for another comment instead of a well structured review but the code already looks really good and this is the only major thing I see now.

Really nice check btw! Never would have thought of this myself and I can see this preventing some people from making mistakes!

jayaddison:
(and I'm also reading about some of the [`EnumCheck` decorator functionality](https://docs.python.org/3/library/enum.html#enum.EnumCheck))

DanielNoord:
I think the primer is also showing that this causes a `Fatal` error somewhere. Our documentation has instructions about running the primer locally, which should help debugging this.

pylint/checkers/classes/class_checker.py
```diff
@@ -887,23 +930,16 @@ def _check_proper_bases(self, node: nodes.ClassDef) -> None:
...
-                members = ancestor.getattr("__members__")
-                if members and isinstance(members[0], nodes.Dict) and members[0].items:
-                    self.add_message(
-                        "invalid-enum-extension",
-                        args=ancestor.name,
-                        node=node,
-                        confidence=INFERENCE,
-                    )
-
             if ancestor.name == object.__name__:
```
jayaddison, line 943: It's not directly relevant to this pull request, but I wonder whether moving this `baseclass-is-object` check before the inference steps, with a `continue` statement, might save some time:

```python
        for base in node.bases:
            if getattr(base, 'name', None) == object.__name__:
                self.add_message(
                    "useless-object-inheritance", args=node.name, node=node
                )
                continue
```

pylint/checkers/classes/class_checker.py
jayaddison: (maybe: for now, it seems simpler and likely more maintainable to keep the logic grouped textually)

pylint/checkers/classes/class_checker.py
jayaddison: That sounds good - yep, I'll give this a try :+1:

jayaddison: Partially-implemented: the message includes information about the flag member where an implicit alias was detected, but doesn't match this up with the corresponding flag(s) that conflict.  I couldn't think of a way to achieve that while also maintaining linear time and space bounding.

tests/functional/i/implicit/implicit_flag_alias.txt
```diff
@@ -0,0 +1,3 @@
+invalid-enum-extension:19:0:19:24:SubclassUnionFlags:"Extending inherited Enum class ""ExplicitUnionFlags""":INFERENCE
+implicit-flag-alias:27:4:27:5:ImplicitUnionFlags:"Flag member <ImplicitUnionFlags.C: 3> overlaps with another":INFERENCE
+implicit-flag-alias:28:4:28:5:ImplicitUnionFlags:"Flag member <ImplicitUnionFlags.D: 5> overlaps with another":INFERENCE
```
jayaddison, line 3: Suggested improvements for the message phrasing here would be great, I think it reads a bit awkwardly.

pylint/checkers/classes/class_checker.py
```diff
@@ -525,6 +525,12 @@ def _has_same_layout_slots(
         "beginning with an underscore) is access outside the class or a "
         "descendant of the class where it's defined.",
     ),
+    "W0213": (
+        "Flag member <%s.%s: %d> overlaps with another",
```
Pierre-Sassoulas, line 529:
```suggestion
        "Flag member <%s.%s: %d> implicitely hide <%s.%s: %d> as %s|%s=%s.",
```

Pierre-Sassoulas (commented):
Looks pretty good already !

mbyrnepr2:
> > where an IntFlag member is excluded from the iteration over the Enum when it can be the result of a bitwise or of two other members
> 
> Ok, my apologies - now I understand, I think - my use of the term _alias_ has been out-of-alignment with the documentation (in `A = 1, B = 2, C = 3`, `C` is considered an alias).
> 
> Previously I thought that _alias_ referred purely to secondary attribute names for a single value (as in my `A = 1, B = 1` example).
> 
> Spending some time to think about renaming the check.. initially `implicit-flag-alias` springs to mind (trying to keep the name concise and combining your correct use of the terminology).

My initial impression was that `alias` would be good to have in there somewhere because that is the [Python documentation's naming](https://docs.python.org/3/library/enum.html#enum.Flag) for it (see section on `__iter__`.

I would say don't change the naming unless someone else also finds that the suggestion makes sense - let's not be led by my impressions but validate with someone else first :D

pylint/checkers/classes/class_checker.py
mbyrnepr2:
```suggestion
        "Flag member <%s.%s: %d> implicitly hides <%s.%s: %d> as %s|%s=%s.",
```

(This is my new specialty :D)

pylint/checkers/classes/class_checker.py
jayaddison: This is now fully-[implemented](https://github.com/PyCQA/pylint/pull/8105/commits/ac21d3149c2b67365249c16ac5853309de93cea3), hopefully in a reasonably-efficient way.

pylint/checkers/classes/class_checker.py
jayaddison: Adapted these suggested changes slightly - I found it easier to understand the cause when shown the intersection (`A & B`) rather than the union (`A | B`) of bitmasks where this issue can occur.

jayaddison:
>  My understanding is that this checker is saying: `Color.BLUE` is an alias for `Color.GREEN & Color.BLUE` and that it may not be necessary to define `Color.BLUE` at all for that reason.

Mm, not quite - it's currently saying that the symbol for `Color.BLUE` and the symbol for `Color.GREEN` share some common elements -- in fact, in the example, `Color.BLUE` contains all of the symbols from `Color.GREEN` -- which may not be what the author intends.

(if they do, they can declare `BLUE = GREEN | RED` and the warning goes away)

>  I think the wording could be changed to something like `Flag member Color.BLUE: 3 is an alias of Color.RED | Color.GREEN` as a simpler wording

That's a nice idea, although there are two problems I can think of:

- I'm not sure it's possible to implement efficiently (can you suggest an algorithm?)
- I don't know what output you'd expect for a class containing `A = 1, B = 3, C = 5` - can you clarify?

jayaddison:
No problem, thanks @mbyrnepr2 - I'm still finding it slightly difficult to describe concisely what the check does (even after writing it), so that's probably a sign that the documentation and messaging could be improved further.

pylint/checkers/classes/class_checker.py
```diff
@@ -874,6 +884,55 @@ def _check_consistent_mro(self, node: nodes.ClassDef) -> None:
...
+            # For each bit position, collect all the flags that set the bit
+            bit_flags = defaultdict(set)
+            for flag_value in assignments.keys():
+                for position, bit in enumerate(reversed(f"{flag_value:b}")):
+                    if bit == "1":
+                        bit_flags[position].add(flag_value)
+
+            # When multiple flags overlap on a bit position, pair the min-and-max
+            overlaps = (
+                (min(*flag_values), max(*flag_values))
```
jayaddison, line 915: In hindsight, this is slightly odd: why only report the maximum overlapping value when multiple items share a common bit position?

(todo: more test coverage to examine this)

tests/functional/i/implicit/implicit_flag_alias.py
```diff
@@ -0,0 +1,34 @@
...
+    RWX = 7
+
+class ImplicitUnionFlags(IntFlag):
+    """Class with flags that overlap implicitly"""
+    A = 1
+    B = 2
+    C = 3  # [implicit-flag-alias]
+    D = 5  # [implicit-flag-alias]
+    E = 8
+    F = 17  # [implicit-flag-alias]
```
jayaddison, line 30: TODO: Add some (intentional) aliases - there are some edge cases to handle around those.

For example:

```
G = 17
```

pylint/checkers/classes/class_checker.py
jayaddison: There was indeed a problem here, and it's now resolved (with test coverage).

Basically, if `1, 5, 9` (for example) were declared as flag values, then the min/max logic that existed here previously would have resulting in the `9` as having a reported overlap, but would not have reported about the `5`.

In practice anyone who noticed that by using `pylint` would probably have run the check again after fixing the `9`, and _then_ been told about the `5`, but that'd be kinda an annoying workflow.

I think the pull request is ready for review again.

pylint/checkers/classes/class_checker.py
jayaddison: (in practice: no, this doesn't seem worthwhile to move or adjust at the moment; removing the check entirely doesn't result in a significant difference in the timings produced by `tests/benchmark/test_baseline_benchmarks.py` locally, so I don't think there's a huge opportunity for improvement here)

pylint/checkers/classes/class_checker.py
```diff
@@ -525,6 +525,12 @@ def _has_same_layout_slots(
         "beginning with an underscore) is access outside the class or a "
         "descendant of the class where it's defined.",
     ),
+    "W0213": (
+        "Flag member %(overlap)s shares bit positions with %(sources)s",
```
Pierre-Sassoulas, line 529: Great way to make the code more maintanable, I wonder if we should doit everywhere.

Pierre-Sassoulas (approved):
LGTM, great checker ! Could go in 2.16.0 maybe.

jayaddison:
Thanks again @Pierre-Sassoulas!  There's no hurry as far as I'm concerned, so I'll leave it up to you as maintainers.

It might be slightly easier for me to respond to any feedback for one change at a time, so I have a mild preference to wait until after 2.16 for this one.. but either way I'll follow the changes and watch for relevant tickets after release.

pylint/checkers/classes/class_checker.py
jayaddison: Unfortunately it's not possible to express in a `pylint` rule (because it also involves `diff` context), but I think my approach would be:

If a string is being modified by a changeset _and_ contains `%`-replacement characters, then consider also changing the template to use named variables (or f-strings) during the same change -- although with some caution about larger potential refactoring if doing so would result in line-reformatting due to `black` / `flake8` rules.

DanielNoord:
@jayaddison You removed your repository and therefore this branch. Does that imply that you no longer want to work on this PR?

jayaddison:
@DanielNoord argh, thanks - my mistake.  I forgot that this was still pending when deleting the repo after my contributions-alias edit (#8343) was merged.  I'm good to continue working on this PR and will re-open it.

Pierre-Sassoulas (changes requested):
The documentation fail is genuine, an example need to be added :)

jayaddison:
Okie doke - I think I'll open a second pull request (re-using the same commits from here) with an example added, since I'm not sure I'll be able to push further commits to this branch now that the repo has been deleted.

jayaddison:
Continues in #8345.

PR #8107, opened 2023-01-24
rename global-variable-not-assigned

### LINKED ISSUE #5073: False positive for global-variable-not-assigned when using with
### Bug description

```python
# File a.py
"""
The "global-variable-not-assigned" check does not take into account that a "with"
statement can very well change the global variable.
For example, when using a global condition variable by multiple threads:
"""

import threading

COND = threading.Condition()


def func1():
    "Function that is called in one thread"
    # pylint: disable=global-statement
    global COND
    with COND:
        COND.notify()
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:16:4: W0602: Using global for 'COND' but no assignment is done (global-variable-not-assigned)
```

### Expected behavior

Pylint realizes that "with" may change the global variable and thus counts it like it was an assignment, and thus does not raise the issue in this case.

### Pylint version

```shell
pylint 2.11.1
astroid 2.8.0
Python 3.9.7 (default, Sep  3 2021, 12:45:31) 
```

### OS / Environment

macOS

### Additional dependencies

Output of pipdeptree:
pylint==1.6.4
  - astroid [required: >=1.4.5,<1.5.0, installed: 1.4.9]
    - lazy-object-proxy [required: Any, installed: 1.4.2]
    - six [required: Any, installed: 1.15.0]
    - wrapt [required: Any, installed: 1.11.2]
  - backports.functools-lru-cache [required: Any, installed: 1.6.4]
  - configparser [required: Any, installed: 4.0.2]
  - isort [required: >=4.2.5, installed: 4.2.15]
  - mccabe [required: Any, installed: 0.6.1]
  - six [required: Any, installed: 1.15.0]

clavedeluna:
## Type of Changes

## Description

Closes #5073


DanielNoord:
Can't we do this with just changing the description but keeping the name?

clavedeluna:
> Can't we do this with just changing the description but keeping the name?

We can, but doesn't the name change read much better?

mbyrnepr2:
If we do change the checker name like this could we simply use `unnecessary-global`?

clavedeluna:
> If we do change the checker name like this could we simply use `unnecessary-global`?

Good suggestion, although that (to me) hints that the author should not use a global variable at all, and what I _think_ this msg is supposed to say is "you're dealing with a global variable, but you aren't re-assigning it (only printing or something) so you don't have to use the global keyword again". so maybe "unnecessary-global-keyword"?

Pierre-Sassoulas:
global-without-modification ? global-used-without-modification ?

mbyrnepr2:
I agree @Pierre-Sassoulas.
Now that you mention it I think there is another checker to highlight any use of global. But that’s a different story.

superfluous-global? Although that could also be confused with “never use”😄

clavedeluna:
I think @Pierre-Sassoulas's suggestion of `global-without-modification` is a great balance of name length and description, so going with that.

timmartin:
It looks good to me.

doc/user_guide/checkers/features.rst
```diff
@@ -1349,6 +1349,9 @@ Variables checker Messages
 :redefined-outer-name (W0621): *Redefining name %r from outer scope (line %s)*
   Used when a variable's name hides a name defined in an outer scope or except
   handler.
+:global-without-modification (W0645): *Unnecessary use of global for %r*
+  Used when a variable defined as a global is then invoked with the `global`
+  keyword but no assignment to this variable is done.
```
mbyrnepr2, line 1354:
```suggestion
:global-without-modification (W0645): *Unnecessary use of global for %r*
  Used when a name listed in a global statement has nothing assigned to it in the current code block.
```

mbyrnepr2 (commented):
It looks like we have a consensus on the current naming in the MR.
I have one suggestion for simplifying the description but other than that looks great!

tests/functional/g/globals.py
```diff
@@ -44,12 +44,12 @@ def global_with_import_from():
 
 def global_no_assign():
     """Not assigning anything to the global makes 'global' superfluous"""
-    global CONSTANT  # [global-variable-not-assigned]
+    global CONSTANT  # [global-without-modification]
     print(CONSTANT)
```
Pierre-Sassoulas, line 50:
```suggestion

RAN_DB_SET = set()
RAN_DB_DICT= dict()

def init_connection_state(alias):
   """Initialize the database connection settings."""
    global RAN_DB_SET 
    global RAN_DB_DICT
    RAN_DB_SET.add(alias)
    return RAN_DB_DICT.set_default("color", "Palomino")
```

Pierre-Sassoulas (changes requested):
I feel like the renaming this make the warning we see on primers become false positives. They were not false positives before as there were no assignments... but a global dict **is** modified if we call ``set_default`` on it (or add on a  global set) thus we should handle the case, probably not raise anything if a function is called on the variable ?

clavedeluna:
Really good point @Pierre-Sassoulas . I'm having some trouble updating the code to handle this case. The location would be updating `is_reassigned_after_current` (changing the name too). Trying to add `nodes.Name` to the list of nodes to look at, but I'm not quite there yet. Do you have any suggestion? Anything I come up with could either easily lead to FP or be too specific to objects like dicts.

tests/functional/g/globals.py
mbyrnepr2: Just so I understand - do you think the Pylint _code_ needs to change or the message _text_?

In this example the set and dict don't need global in order to be updated, right?
```python
(venv38) Marks-MacBook-Air-2:programming markbyrne$ python x.py 
{'color': 'Palomino'}
{'xxx'}
(venv38) Marks-MacBook-Air-2:programming markbyrne$ cat x.py 
RAN_DB_SET = set()
RAN_DB_DICT= dict()
def init_connection_state(alias):
    RAN_DB_SET.add(alias)
    return RAN_DB_DICT.update(color = "Palomino")


init_connection_state("xxx")
print(RAN_DB_DICT)
print(RAN_DB_SET)
```

Pierre-Sassoulas: Nice point, actually the message name was better before, and there's nothing to handle: the global keyword is only really required for assignment only. I think the documentation around it could be clearer though.

mbyrnepr2: I am inclined to agree. We can’t convey precisely an unambiguous meaning with message name alone. Users will usually need to double check the doc for further clarity.

mbyrnepr2: I think Daniël probably had this in mind also.

clavedeluna:
Alright so we've come full circle here and I think we've met @DanielNoord! It sounds like the action item here is just to update the doc to more clearly indicate this msg is indeed for assignment only, so the linked issue is not a FP, it is indeed correct because no assignment was being done.

clavedeluna:
Will open a new PR for the new requirement, just for keeping record.

PR #8108, opened 2023-01-24
docs: fix grammatical error

rnwst:
## Type of Changes

## Description

Fix grammatical error in explanation of `global-statement (W0603)`.

### DIFF
--- a/doc/user_guide/checkers/features.rst
+++ b/doc/user_guide/checkers/features.rst
@@ -1366,7 +1366,7 @@ Variables checker Messages
   or a generator expression) is used outside the loop.
 :global-statement (W0603): *Using the global statement*
   Used when you use the "global" statement to update a global variable. Pylint
-  just try to discourage this usage. That doesn't mean you cannot use it !
+  discourages its usage. That doesn't mean you cannot use it!
 :global-at-module-level (W0604): *Using the global statement at the module level*
   Used when you use the "global" statement at the module level since it has no
-  effect
+  effect.
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -441,14 +441,14 @@ MSGS: dict[str, MessageDefinitionTuple] = {
         "Using the global statement",  # W0121
         "global-statement",
         'Used when you use the "global" statement to update a global '
-        "variable. Pylint just try to discourage this "
-        "usage. That doesn't mean you cannot use it !",
+        "variable. Pylint discourages its usage. That doesn't mean you cannot "
+        "use it!",
     ),
     "W0604": (
         "Using the global statement at the module level",  # W0103
         "global-at-module-level",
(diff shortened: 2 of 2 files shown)


Pierre-Sassoulas (changes requested):
This is generated and would need to be modified in the checker directly.

clavedeluna:
@rnwst the steps you need to take are:
1. update the language in variables.py:444
2. run tox -e docs. Sometimes locally it also edits non-relevant stuff so only keep what's relevant here from this output
3. add a changelog with `towncrier create`

Pierre-Sassoulas (approved):
Thanks !

rnwst:
Thank you, @clavedeluna, for the step-by-step instructions. I've applied the changes in `variables.py` and ran `tox -e docs`. This resulted in some further changes unrelated to my change, which I haven't committed. These are as follows:
``` console
diff --git a/doc/user_guide/checkers/features.rst b/doc/user_guide/checkers/features.rst
index 465dc0e39..dc442c30f 100644
--- a/doc/user_guide/checkers/features.rst
+++ b/doc/user_guide/checkers/features.rst
@@ -474,7 +474,7 @@ Exceptions checker Messages
   program errors, use ``except Exception:`` (bare except is equivalent to
   ``except BaseException:``).
 :broad-exception-raised (W0719): *Raising too general exception: %s*
-  Raising exceptions that are too generic force you to catch exception
+  Raising exceptions that are too generic force you to catch exceptions
   generically too. It will force you to use a naked ``except Exception:``
   clause. You might then end up catching exceptions other than the ones you
   expect to catch. This can hide bugs or make it harder to debug programs when
diff --git a/doc/user_guide/configuration/all-options.rst b/doc/user_guide/configuration/all-options.rst
index d284d1194..21c9e6e70 100644
--- a/doc/user_guide/configuration/all-options.rst
+++ b/doc/user_guide/configuration/all-options.rst
@@ -1258,7 +1258,7 @@ Standard Checkers
 
 --spelling-dict
 """""""""""""""
-*Spelling dictionary name. Available dictionaries: en (aspell), en_AU (aspell), en_CA (aspell), en_GB (aspell), en_US (aspell).*
+*Spelling dictionary name. Available dictionaries: none. To make it work, install the 'python-enchant' package.*
``` 

To fix the second diff, I tried to install the python package `enchant`, but this doesn't seem to exist. The package closest to that name is called [`pyenchant`](https://pypi.org/project/pyenchant/). What's going on here?

@Pierre-Sassoulas, since `features.rst` is generated, why is it version-controlled?

Pierre-Sassoulas:
> @Pierre-Sassoulas, since features.rst is generated, why is it version-controlled?

When we modify the code actually doing the generation (that we're also maintaining in this repo), it's easier to review if the changes in the generated rst are visibles.

PR #8114, opened 2023-01-26
fix(4857): fix false positive for `use-maxsplit-arg` with custom spli…

### LINKED ISSUE #4857: `use-maxsplit-arg` false positive with custom split method
### Bug description
I get a false positive when implementing a `split` method on a custom class:

```python
class Obj:

    def split(self, point):
        return 0, 1


obj = Obj()
obj.split((0, 0))[0]
```

Problem seems to be in the `[0]` part. When I change this to `[1]`, no warning is shown.

### Configuration

_No response_

### Command used

```shell
pylint --rcfile=.pylintrc module
```

### Pylint output

```shell
C0207: Use obj.split((0, 0), maxsplit=1)[0] instead (use-maxsplit-arg)
```

### Expected behavior

No warning

### Pylint version

```shell
pylint 2.9.6
astroid 2.6.6
Python 3.6.14 (default, Aug  9 2021, 06:21:09)
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

yushao2:
…t method

## Type of Changes

## Description
Refer to linked issue. This PR resolves the false positive arising from custom defined `split` methods. This is accomplished by adding an additional check to ensure that the expression used before the `.split()` is inferred as a `Const` node with a `String` type value

Closes #4857

### DIFF
--- a/pylint/checkers/refactoring/recommendation_checker.py
+++ b/pylint/checkers/refactoring/recommendation_checker.py
@@ -121,6 +121,12 @@ class RecommendationChecker(checkers.BaseChecker):
             and isinstance(utils.safe_infer(node.func), astroid.BoundMethod)
         ):
             return
+        inferred_expr = utils.safe_infer(node.func.expr)
+        if not (
+            isinstance(inferred_expr, nodes.Const)
+            and not set(inferred_expr.nodes_of_class(nodes.ClassDef))
+        ):
+            return
 
         try:
             sep = utils.get_argument_from_call(node, 0, "sep")
--- a/pylint/checkers/refactoring/recommendation_checker.py
+++ b/pylint/checkers/refactoring/recommendation_checker.py
@@ -123,7 +123,7 @@ class RecommendationChecker(checkers.BaseChecker):
             return
         inferred_expr = utils.safe_infer(node.func.expr)
         if isinstance(inferred_expr, astroid.Instance) and any(
-            i for i in inferred_expr.nodes_of_class(nodes.ClassDef)
+            inferred_expr.nodes_of_class(nodes.ClassDef)
         ):
             return
 
--- a/pylint/checkers/refactoring/recommendation_checker.py
+++ b/pylint/checkers/refactoring/recommendation_checker.py
@@ -122,9 +122,8 @@ class RecommendationChecker(checkers.BaseChecker):
         ):
             return
         inferred_expr = utils.safe_infer(node.func.expr)
-        if not isinstance(inferred_expr, nodes.Const) and (
-            isinstance(inferred_expr, astroid.Instance)
-            and set(inferred_expr.nodes_of_class(nodes.ClassDef))
+        if isinstance(inferred_expr, astroid.Instance) and set(
+            inferred_expr.nodes_of_class(nodes.ClassDef)
         ):
             return
 
--- a/pylint/checkers/refactoring/recommendation_checker.py
+++ b/pylint/checkers/refactoring/recommendation_checker.py
@@ -122,7 +122,7 @@ class RecommendationChecker(checkers.BaseChecker):
         ):
             return
         inferred_expr = utils.safe_infer(node.func.expr)
-        if isinstance(inferred_expr, astroid.Instance) and set(
+        if isinstance(inferred_expr, astroid.Instance) and any(i for i in
             inferred_expr.nodes_of_class(nodes.ClassDef)
         ):
             return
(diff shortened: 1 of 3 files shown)


yushao2:
based on the primer test results, it feels like this might result in false negatives.

maybe i'll tighten up the checks to look for custom class methods instead

Pierre-Sassoulas (approved):
Very nice !

yushao2:
@Pierre-Sassoulas seems like I'm unable to merge the PR -- feel free to loop whoever you think should take a look at this as well or merge it if you think it's good to go 🎉

Pierre-Sassoulas:
I agree with your comment about the false negative though, it would be better to fix it.

yushao2:
I'll check again later in the day

yushao2 (commented):
fix: minor fix logic

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -121,6 +121,12 @@ def _check_use_maxsplit_arg(self, node: nodes.Call) -> None:
             and isinstance(utils.safe_infer(node.func), astroid.BoundMethod)
         ):
             return
+        inferred_expr = utils.safe_infer(node.func.expr)
+        if not (
+            isinstance(inferred_expr, nodes.Const)
+            and not set(inferred_expr.nodes_of_class(nodes.ClassDef))
```
yushao2, line 127:
```suggestion
            or not set(inferred_expr.nodes_of_class(nodes.ClassDef))
```

this should fix it

yushao2:
@Pierre-Sassoulas good to go 😀

yushao2:
@Pierre-Sassoulas is this good to go?

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -121,6 +121,12 @@ def _check_use_maxsplit_arg(self, node: nodes.Call) -> None:
             and isinstance(utils.safe_infer(node.func), astroid.BoundMethod)
         ):
             return
+        inferred_expr = utils.safe_infer(node.func.expr)
+        if not isinstance(inferred_expr, nodes.Const) and (
+            isinstance(inferred_expr, astroid.Instance)
```
DanielNoord, line 126: Can't the first line be removed? The `and` makes the first check redundant.

DanielNoord (changes requested):

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -121,6 +121,12 @@ def _check_use_maxsplit_arg(self, node: nodes.Call) -> None:
...
+            and set(inferred_expr.nodes_of_class(nodes.ClassDef))
+        ):
+            return
```
yushao2, line 129:
```suggestion
        if isinstance(inferred_expr, astroid.Instance)
           and set(inferred_expr.nodes_of_class(nodes.ClassDef)):
            return
```

pylint/checkers/refactoring/recommendation_checker.py
yushao2: updated :) thanks again

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -121,6 +121,11 @@ def _check_use_maxsplit_arg(self, node: nodes.Call) -> None:
             and isinstance(utils.safe_infer(node.func), astroid.BoundMethod)
         ):
             return
+        inferred_expr = utils.safe_infer(node.func.expr)
+        if isinstance(inferred_expr, astroid.Instance) and set(
+            inferred_expr.nodes_of_class(nodes.ClassDef)
+        ):
```
DanielNoord, line 127:
```suggestion
        if isinstance(inferred_expr, astroid.Instance) and any(i for i in
            inferred_expr.nodes_of_class(nodes.ClassDef)
        ):
```

This is your intent right?

yushao2: that's right :)

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -121,6 +121,11 @@ def _check_use_maxsplit_arg(self, node: nodes.Call) -> None:
             and isinstance(utils.safe_infer(node.func), astroid.BoundMethod)
         ):
             return
+        inferred_expr = utils.safe_infer(node.func.expr)
+        if isinstance(inferred_expr, astroid.Instance) and any(
+            i for i in inferred_expr.nodes_of_class(nodes.ClassDef)
+        ):
```
DanielNoord, line 127:
```suggestion
        if isinstance(inferred_expr, astroid.Instance) and any(
            inferred_expr.nodes_of_class(nodes.ClassDef)
        ):
```

Checked, and it can just be this. Even shorter 😄

DanielNoord: Could you als import `Instance` from `bases` instead of `astroid` directly?

Pierre-Sassoulas (approved):
👌

DanielNoord (approved):

PR #8115, opened 2023-01-26
Tighten design constraints

nickdrozd:
## Type of Changes

## Description

This PR tightens a few of the design constraints in the config file. No code is changed, although a few disable comments are added.

The rationale for the numbers is that they should be as low as possible without raising too many warnings. This way only the worst offenders are flagged. For example, there are only a few functions with more than 50 statements, and more with between 40 and 50 statements. So 50 is the cutoff. I'm not wed to any of the numbers in particular, but I think they're about where they should be.

Observe that several functions raise multiple warnings. Such functions are prime candidates for refactoring.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -425,30 +425,11 @@ max-branches = 20
 # Maximum number of statements in function / method body
 max-statements = 50
 
-# Maximum number of parents for a class (see R0901).
-max-parents=7
-
-# List of qualified class names to ignore when counting class parents (see R0901).
-ignored-parents=
-
 # Maximum number of attributes for a class (see R0902).
 max-attributes=11
 
-# Minimum number of public methods for a class (see R0903).
-min-public-methods=2
-
-# Maximum number of public methods for a class (see R0904).
-max-public-methods=25
-
-# Maximum number of boolean expressions in an if statement (see R0916).
-max-bool-expr=5
-
 # Maximum number of statements in a try-block
 max-try-statements = 7
-
-# List of regular expressions of class ancestor names to
-# ignore when counting public methods (see R0903).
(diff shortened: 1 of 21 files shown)


pylintrc
```diff
@@ -413,16 +413,16 @@ max-spelling-suggestions=2
 max-args=10
 
 # Maximum number of locals for function / method body
```
Pierre-Sassoulas, line 415: Could we use the default value and remove the description and the value ?

Pierre-Sassoulas (changes requested):
Thank you ! Let's cleanup the pylintrc and let's merge.

pylintrc
nickdrozd: You mean for all of them or for max locals in particular? For all of the settings in this PR at least, the configured value is higher than the default, so using the defaults would require a lot more disabling comments.

As for the rest of the config file, personally I would be happy to rip through and cut out a bunch of the text. Currently there's an awful lot of text, and I find it hard to read. How much cleanup were you thinking?

Pierre-Sassoulas: I was thinking of removing the default values and their description. It should be possible to do that automatically with the conf génération command iirc

nickdrozd: I'm not quite sure what you mean.

nickdrozd: Okay, I cut a few more of them.

nickdrozd:
Unrelated to this PR, it looks like the full set of primer tests runs even if linting fails. Is that something that should be changed? It seems wasteful to rerun everything after lint fixups.

Pierre-Sassoulas (approved):
Thank you nice clean up !

PR #8122, opened 2023-01-27
Fix issues with new typing Union syntax (Py310)

### LINKED ISSUE #8119: Issues with new Union Syntax
### Bug description

Home Assistant dropped Python 3.9 this week. While upgrading all runtime annotations to the new union syntax, I discovered some issues with pylint. It seems astroid doesn't know how to handle these.

```py
# pylint: disable=missing-docstring,too-few-public-methods,line-too-long,useless-suppression

from typing import Generic, TypeVar

T = TypeVar("T")


class Coordinator(Generic[T]):
    def __init__(self, update_interval=None) -> None:
        self.update_interval = update_interval


class Child(Coordinator[int | str]):  # <-- 'int | str' here seems to be the issue
    def __init__(self) -> None:
        Coordinator.__init__(self, update_interval=2)  # non-parent-init-called

    def _async_update_data(self):
        assert self.update_interval  # access-member-before-definition
        self.update_interval = 1  # attribute-defined-outside-init
```

### Configuration

```ini
[MAIN]
load-plugins=
    pylint.extensions.code_style,
    pylint.extensions.typing,
py-version=3.10

[TYPING]
runtime-typing=no
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
test.py:16:8: W0233: __init__ method from a non direct base class 'Coordinator' is called (non-parent-init-called)
test.py:19:15: E0203: Access to member 'update_interval' before its definition line 20 (access-member-before-definition)
test.py:20:8: W0201: Attribute 'update_interval' defined outside __init__ (attribute-defined-outside-init)
```

### Expected behavior

_No errors_

### Pylint version

```shell
pylint 2.16.0b1
astroid 2.14.0-dev0
Python 3.11.1
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

cdce8p:
## Description
Add regression tests for new typing Union syntax.
Bump astroid to `2.13.4`.

Requires https://github.com/PyCQA/astroid/pull/1977
Closes #8119

### DIFF
--- a/doc/whatsnew/fragments/8119.bugfix
+++ b/doc/whatsnew/fragments/8119.bugfix
@@ -1,0 +1,3 @@
+Fix issue with new typing Union syntax in runtime context for Python 3.10+.
+
+Closes #8119
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies    = [
     # Also upgrade requirements_test_min.txt.
     # Pinned to dev of second minor update to allow editable installs and fix primer issues,
     # see https://github.com/PyCQA/astroid/issues/1341
-    "astroid>=2.13.3,<=2.15.0-dev0",
+    "astroid>=2.14.1,<=2.16.0-dev0",
     "isort>=4.2.5,<6",
     "mccabe>=0.6,<0.8",
     "tomli>=1.1.0;python_version<'3.11'",
--- a/requirements_test_min.txt
+++ b/requirements_test_min.txt
@@ -1,6 +1,6 @@
 -e .[testutils,spelling]
 # astroid dependency is also defined in pyproject.toml
-astroid==2.13.3  # Pinned to a specific version for tests
+astroid==2.14.1  # Pinned to a specific version for tests
 typing-extensions~=4.4
 py~=1.11.0
 pytest~=7.2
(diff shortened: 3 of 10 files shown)


Pierre-Sassoulas:
It seems there's been a core dump two times in a row in the 3.7 tests and once in the 3.8 tests ?  I wonder if this what happen in case of timeout now or if this is a genuine problem with astroid 2.14 ?

cdce8p:
> It seems there's been a core dump two times in a row in the 3.7 tests and once in the 3.8 tests ? I wonder if this what happen in case of timeout now or if this is a genuine problem with astroid 2.14 ?

It's the recursion issue from https://github.com/PyCQA/astroid/pull/1982 again, sadly. I wouldn't consider astroid broken though so we don't need to yank the releases. However, it might make sense to revert the change. What do you think?

It's a bit of work, but I could do it. With that I would also consider releasing `2.13.5` and `2.14.1`.

Pierre-Sassoulas:
Hmm yeah, I have no idea how to fix this. Maybe a partial revert keeping the new added tests ? Also we don't need to maintain 2.13 anymore, we could only release 2.14.1.

cdce8p:
> Also we don't need to maintain 2.13 anymore, we could only release 2.14.1.

Yeah, also I kind of broke the last release of `2.13`. Would be nice to fix it even if we don't "need" to.
I can try to do the backports manually, it should probably work then.

cdce8p:
Did the revert, released both `2.13.5` + `2.14.1`, and merged everything back into `main`. This should work now 🤞🏻

Pierre-Sassoulas (approved):
Great work with all the release ! I suppose it prove it's easier to do now than before 😄

Issue #5159, opened 2021-10-15
Add dataclass tests


cdce8p:
### Current problem

```py
from dataclasses import dataclass, field

@dataclass
class Expandable:
    _api: field()
    var = []
```

### Desired solution

Detect invalid or problematic usage in dataclasses.

**Some ideas**
- [ ] Invalid `field` call. It should be the value of the assignment
- [ ] Missing type annotation
- [ ] Missing use of `field` -> `default_factory` for mutable types. Careful of `ClassVars` as it might be intended for them.
- [ ] Unsupported-version check as dataclasses require Python 3.7

--
It might be best to create a new checker class for it.

PR #8126, opened 2023-01-28
feat(5159): add new checker `invalid-field-call`

yushao2:
## Type of Changes

## Description

Added new checker `invalid-field-call` to check for invalid `dataclasses.field()` usage.

Correct usage should be only within dataclasses as well as within `make_dataclass()` function (see: https://docs.python.org/3/library/dataclasses.html#dataclasses.make_dataclass)

Would appreciate better exampels for good/bad documentation.

Refs #5159


yushao2:
documentation check is failing, but i'm not too sure how to fix it -- help appreciated

Pierre-Sassoulas:
It can't be fixed but it won't fail once it's merged

yushao2:
primer looks good now. ready for review

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,137 @@
...
+from pylint.checkers import BaseChecker, utils
+from pylint.interfaces import INFERENCE
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+def _is_dataclasses_module(node: nodes.NodeNG) -> bool:
+    """Utility function to check if node is from dataclasses_module."""
+    return isinstance(node, nodes.Module) and node.name == "dataclasses"
```
DanielNoord, line 22: Sadly there are other types of `dataclasses` as well. There is actually a list that we maintain in `astroid`. You could try importing it:
https://github.com/PyCQA/astroid/blob/main/astroid/brain/brain_dataclasses.py

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,137 @@
...
+) -> Any:
+    """Utility function to check either a Name/Attribute node's name/attrname with a
+    given string.
+    """
+    if isinstance(node, nodes.Name):
+        return node.name == check_with
+    return node.attrname == check_with
+
+
+def _is_annotated_with_dataclass(d_nodes: list[nodes.NodeNG]) -> bool:
```
DanielNoord, line 36: You should be able to do most of this with `ClassDef.is_dataclass`. You just need to infer the parents of the node.

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,137 @@
...
+        if (isinstance(n, nodes.Name) and n.name == "dataclass")
+        or (
+            isinstance(n, nodes.Attribute)
+            and n.attrname == "dataclass"
+            and _is_dataclasses_module(utils.safe_infer(n.expr))
+        )
+    )
+
+
+class DataclassChecker(BaseChecker):
```
DanielNoord, line 57: Don't we have other `dataclass` specific checks that should go here?

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,137 @@
...
+    """Checker that detects invalid or problematic usage in dataclasses.
+
+    Checks for
+    * invalid-field-call
+    """
+
+    name = "dataclass"
+    msgs = {
+        "E3601": (
+            "Invalid usage of field(), %s",
```
DanielNoord, line 67:
```suggestion
            "Invalid usage of field(), '%s'",
```

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,137 @@
...
+            "invalid-field-call",
+            "The dataclasses.field() specifier should only be used as the value of "
+            "an assignment within a dataclass, or within the make_dataclass() function.",
+            {"minversion": (3, 7)},
```
DanielNoord, line 71: Is this necessary now that we only support 3.7+ runtimes?

DanielNoord (commented):
I love the check but have some questions about the implementation. I think those should be resolved before a full review can be done!

Well done on implementing this though, this seems useful!

pylint/checkers/dataclass_checker.py
yushao2: i realised I can make use of this: https://github.com/PyCQA/astroid/blob/main/astroid/brain/brain_dataclasses.py#L49

do you think it's a good idea?

pylint/checkers/dataclass_checker.py
yushao2: I'm not too sure what there is tbh, but I'm following the suggestion of the issue linked to create a new checker for dataclass checks

pylint/checkers/dataclass_checker.py
DanielNoord: I think it would be better to use `is_dataclass`. We can leave the determining of dataclasses up to `astroid` and keep the logic in one package then.

Pierre-Sassoulas:
I moved the milestone to 2.17.0 to be able to focus on releasing astroid 2.13.4 / 2.14 and pylint 2.16 :)

yushao2:
> I moved the milestone to 2.17.0 to be able to focus on releasing astroid 2.13.4 / 2.14 and pylint 2.16 :)

sure! hopefully I can implement more checks relating to #5159 for 2.17 release

yushao2:
hey @DanielNoord , how should I proceed with this PR?

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def _is_dataclasses_module(node: nodes.NodeNG) -> bool:
+    """Utility function to check if node is from dataclasses_module."""
+    return isinstance(node, nodes.Module) and node.name in DATACLASS_MODULES
+
+
+def _check_name_or_attrname_eq_to(
+    node: nodes.Name | nodes.Attribute, check_with: str
+) -> Any:
```
DanielNoord, line 28:
```suggestion
) -> bool:
```

pylint/checkers/dataclass_checker.py
DanielNoord: I feel like there are other `dataclass` specific checks in our codebase already that should be moved to this class, but that is something for a later follow up PR.

pylint/checkers/dataclass_checker.py
DanielNoord: You are right!

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,114 @@
...
+        is used improperly within a dataclass.
+        """
+        if not isinstance(node.func, (nodes.Name, nodes.Attribute)):
+            return
+        if not _check_name_or_attrname_eq_to(node.func, "field"):
+            return
+        inferred_func = utils.safe_infer(node.func)
+        if not (
+            isinstance(inferred_func, nodes.FunctionDef)
+            and _is_dataclasses_module(inferred_func.parent)
```
DanielNoord, line 74:
```suggestion
            and _is_dataclasses_module(inferred_func.root)
```

Let's also support nested definitions. Using the root module is better here.

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,114 @@
...
+        ):
+            return
+        parent_node = node.parent
+        while parent_node and not isinstance(parent_node, (nodes.ClassDef, nodes.Call)):
+            parent_node = parent_node.parent
```
DanielNoord, line 79: Can we rename this? Something like `scope_node` or something.

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if isinstance(parent_node, nodes.Call):
+            inferred_func = utils.safe_infer(parent_node.func)
+        if not (
```
DanielNoord, line 83:
```suggestion
        
        if not (
```

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,114 @@
...
+            self.add_message(
+                "invalid-field-call",
+                node=node,
+                args=(
+                    "it should be used within a dataclass or the make_dataclass() function.",
+                ),
+                confidence=INFERENCE,
+            )
+            return
+        if isinstance(parent_node, nodes.Call):
```
DanielNoord, line 102: This feels really weird. Can we separate the check for `Call` and `ClassDef`. We would duplicate one `add_message` but that feels more logical than these `if` statements.

tests/functional/i/invalid/invalid_field_call.py
```diff
@@ -0,0 +1,57 @@
+# pylint: disable=unused-import, invalid-name, missing-module-docstring, missing-class-docstring, missing-function-docstring, too-few-public-methods, unsubscriptable-object, multiple-statements
```
DanielNoord, line 1:
```suggestion
"""Tests for the invalid-field-call message."""

# pylint: disable=unused-import, invalid-name, missing-class-docstring, missing-function-docstring, too-few-public-methods, unsubscriptable-object, multiple-statements
```

DanielNoord (changes requested):
What was left was me reviewing this 😄 

Thanks! This does look very nice, especially with the use of the `astroid` brain.

pylint/checkers/dataclass_checker.py
yushao2: I can't remember why but this doesn't pass type checking

pylint/checkers/dataclass_checker.py
yushao2: Hmm I think this doesn't work. using root breaks things

pylint/checkers/dataclass_checker.py
yushao2: Ah okay, I can cast the .name / .attrname attributes to string

pylint/checkers/dataclass_checker.py
yushao2: I slightly refactored it, let me know what you think

pylint/checkers/dataclass_checker.py
DanielNoord: That sounds like a bug. Do you have the errors?

pylint/checkers/dataclass_checker.py
DanielNoord: We should add an ignore then. This is due to `astroid` missing annotations

pylint/checkers/dataclass_checker.py
yushao2: sorry, i meant the test results would break and would not produce the same effect as intended

yushao2: i applied the suggestion. let's see

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        is used improperly within a dataclass.
+        """
+        if not isinstance(node.func, (nodes.Name, nodes.Attribute)):
+            return
+        if not _check_name_or_attrname_eq_to(node.func, "field"):
+            return
+        inferred_func = utils.safe_infer(node.func)
+        if not (
+            isinstance(inferred_func, nodes.FunctionDef)
+            and _is_dataclasses_module(inferred_func.root)
```
DanielNoord, line 74: My bad, I should have been much more explicit.
```suggestion
            and _is_dataclasses_module(inferred_func.root())
```

This will fix the tests :smile: Sorry about this and for wasting your time!

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        self, node: nodes.Call, scope_node: nodes.Call
+    ) -> None:
+        """Checks for special case where calling field is valid as an argument of the
+        make_dataclass() function.
+        """
+        inferred_func = utils.safe_infer(scope_node.func)
+        if (
+            scope_node.func.name == "make_dataclass"
+            and isinstance(inferred_func, nodes.FunctionDef)
+            and _is_dataclasses_module(inferred_func.parent)
```
yushao2, line 114: @DanielNoord should this be changed too? 
```suggestion
            and _is_dataclasses_module(inferred_func.root())
```

pylint/checkers/dataclass_checker.py
yushao2: no worries about it -- could you check my other comment? just a small question from me

pylint/checkers/dataclass_checker.py
DanielNoord: Yeah!

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        if isinstance(scope_node, nodes.Call):
+            self._check_invalid_field_call_within_call(node, scope_node)
+            return
+
+        if not scope_node or not scope_node.is_dataclass:
+            self.add_message(
+                "invalid-field-call",
+                node=node,
+                args=(
+                    "it should be used within a dataclass or the make_dataclass() function.",
```
Pierre-Sassoulas, line 90: I think we could make two messages without requiring to pass the args here. Maybe field-call-not-assigned and invalid-field-call ?

Pierre-Sassoulas (commented):
Looks pretty  good, clearly a lot of work went into this 👍

tests/functional/i/invalid/invalid_field_call.py
```diff
@@ -0,0 +1,59 @@
+"""Tests for the invalid-field-call message."""
+
+# pylint: disable=unused-import, invalid-name, missing-module-docstring, missing-class-docstring, missing-function-docstring, too-few-public-methods, unsubscriptable-object, multiple-statements
```
DanielNoord, line 3:
```suggestion
# pylint: disable=missing-class-docstring, missing-function-docstring, too-few-public-methods
```

I think it is better to disable all others in line so we see what is wrong there.

pylint/checkers/dataclass_checker.py
```diff
@@ -0,0 +1,128 @@
...
+from pylint.checkers import BaseChecker, utils
+from pylint.interfaces import INFERENCE
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+def _is_dataclasses_module(node: nodes.NodeNG) -> bool:
+    """Utility function to check if node is from dataclasses_module."""
+    return isinstance(node, nodes.Module) and node.name in DATACLASS_MODULES
```
DanielNoord, line 23:
```suggestion
def _is_dataclasses_module(node: nodes.Modules) -> bool:
    """Utility function to check if node is from dataclasses_module."""
    return node.name in DATACLASS_MODULES
```

DanielNoord (commented):
Sorry for all the poor previous reviews. I keep seeing stuff that I should have seen earlier. I think these are my last nits.

jacobtylerwalls (changes requested):
Sorry for the pest of a test case, but we've learned the hard way to not rely on `name` being present:

```diff
diff --git a/tests/functional/d/dataclass/dataclass_with_default_factory.py b/tests/functional/d/dataclass/dataclass_with_default_factory.py
index 2a3634652..1861cf5be 100644
--- a/tests/functional/d/dataclass/dataclass_with_default_factory.py
+++ b/tests/functional/d/dataclass/dataclass_with_default_factory.py
@@ -49,3 +49,10 @@ Test2.int_prop["key"] = "value"  # [unsupported-assignment-operation]
 class TEST3:
     """Test dataclass that puts call to field() in another function call"""
     attribute: int = cast(int, field(default_factory=dict))
+
+
+@dc.dataclass
+class TEST4:
+    def call_lambda(y):
+        lambda x: cast(int, x)(field(default_factory=y))
+    attribute: int = call_lambda(dict)
```

Gives:

```python
Exception on node <Call l.57 at 0x105f46c40> in file '/Users/.../pylint/tests/functional/d/dataclass/dataclass_with_default_factory.py'
  File "/Users/.../pylint/pylint/utils/ast_walker.py", line 91, in walk
    callback(astroid)
  File "/Users/.../pylint/pylint/checkers/dataclass_checker.py", line 56, in visit_call
    self._check_invalid_field_call(node)
  File "/Users/.../pylint/pylint/checkers/dataclass_checker.py", line 82, in _check_invalid_field_call
    self._check_invalid_field_call_within_call(node, scope_node)
  File "/Users/.../pylint/pylint/checkers/dataclass_checker.py", line 112, in _check_invalid_field_call_within_call
    scope_node.func.name == "make_dataclass"
AttributeError: 'Call' object has no attribute 'name'
```

tests/functional/i/invalid/invalid_field_call.py
jacobtylerwalls: Took your suggestion (but for invalid-name) in 
3f2a279.

pylint/checkers/dataclass_checker.py
jacobtylerwalls: Interesting idea. I think we can merge as is, though. Do you think there's an argument for one message in order to keep all the good/bad in the same place (and not proliferating the messages too much)?

jacobtylerwalls (dismissed):

jacobtylerwalls (dismissed):

Pierre-Sassoulas (approved):
Current fail is expected when we add a checker, thank you @jacobtylerwalls for taking over 👌

PR #8127, opened 2023-01-28
ci: parallelise primer run pr on modules

yushao2:
## Type of Changes

| ✓   | :sparkles: New feature (?)|

## Description

CI improvements to speed up primer runs by dynamically populating a matrix and adding mechanism to run primer tests per package.

### Test results
Initial tests on own repo yielded 22m for entire primer run -- which usually takes 45m or more
image

- `Primer/comment` job works with merging all test artifacts with jq but output seems wonky: https://github.com/yushao2/pylint/pull/4#issuecomment-1407426487

### DIFF
--- a/.github/workflows/primer_comment.yaml
+++ b/.github/workflows/primer_comment.yaml
@@ -98,7 +98,6 @@ jobs:
                archive_format: "zip",
             });
             fs.writeFileSync("primer_pr_number.zip", Buffer.from(downloadThree.data));
-      - run: sudo apt install -y jq
       - run: unzip primer_main_output.zip
       - run:
           jq -n 'reduce inputs as $item ({}; . *= $item)' output_${{
--- a/.github/workflows/primer_comment.yaml
+++ b/.github/workflows/primer_comment.yaml
@@ -98,9 +98,17 @@ jobs:
                archive_format: "zip",
             });
             fs.writeFileSync("primer_pr_number.zip", Buffer.from(downloadThree.data));
+      - run: sudo apt install -y jq
       - run: unzip primer_main_output.zip
-      - run: unzip primer_pr_output.zip
+      - run:
+          jq -n 'reduce inputs as $item ({}; . *= $item)' output_${{
+          steps.python.outputs.python-version }}_main*.txt > output_${{
+          steps.python.outputs.python-version }}_main.txt
       - run: unzip primer_pr_number.zip
+      - run:
+          jq -n 'reduce inputs as $item ({}; . *= $item)' output_${{
+          steps.python.outputs.python-version }}_pr*.txt > output_${{
+          steps.python.outputs.python-version }}_pr.txt
       - name: Compare outputs
         run: |
           . venv/bin/activate
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -31,7 +31,6 @@ jobs:
         uses: actions/checkout@v3.3.0
         with:
           fetch-depth: 0
-      - run: sudo apt install -y jq
       - name: Generate matrix
         id: set-matrix
         run:
--- a/.github/workflows/primer_run_main.yaml
+++ b/.github/workflows/primer_run_main.yaml
@@ -23,13 +23,32 @@ permissions:
   contents: read
 
 jobs:
-  run-primer:
-    name: Run / ${{ matrix.python-version }}
+  matrix-prep-packages:
+    name: Generate matrix of packages to prime
     runs-on: ubuntu-latest
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.3.0
+        with:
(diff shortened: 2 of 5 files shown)


.github/workflows/primer_comment.yaml
```diff
@@ -98,9 +98,17 @@ jobs:
                archive_format: "zip",
             });
             fs.writeFileSync("primer_pr_number.zip", Buffer.from(downloadThree.data));
+      - run: sudo apt install -y jq
```
cdce8p, line 101: `jq` is already installed on Github action runners.

cdce8p (changes requested):
Not sure this is a good idea. True, the primer tests are quite slow. However, we do only have a limited amount of concurrent jobs available. Could be 20 or 40 for the whole org.[^1] Blocking all with a single PR is unacceptable IMO.

[^1]: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

.github/workflows/primer_comment.yaml
yushao2: my bad. will remove

yushao2:
> Not sure this is a good idea. True, the primer tests are quite slow. However, we do only have a limited amount of concurrent jobs available. Could be 20 or 40 for the whole org.[1](#user-content-fn-1-767346f9f5e184629b128f55dff70f47) Blocking all with a single PR is unacceptable IMO.
> 
> ## Footnotes
> 1. https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits [↩](#user-content-fnref-1-767346f9f5e184629b128f55dff70f47)

Thanks for pointing that up! I didn't know about that. I'd also like to highlight that most of the primer tests could complete within a minute which might free up runners.

However as you also mentioned -- since there's a max cap of concurrency, if there are many jobs that are queued we might still end up with little to no benefit despite parallelising it

image

cdce8p:
I would have to look for it, but didn't we use to split the primer run into multiple jobs already some time ago? Seems like this was removed.

.github/workflows/primer_comment.yaml
yushao2, line 101:
```suggestion
```

.github/workflows/primer_run_main.yaml
```diff
@@ -23,13 +23,32 @@ permissions:
...
 jobs:
+  matrix-prep-packages:
+    name: Generate matrix of packages to prime
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.3.0
+        with:
+          fetch-depth: 0
+      - run: sudo apt install -y jq
```
yushao2, line 34:
```suggestion
```

.github/workflows/primer_run_pr.yaml
```diff
@@ -32,13 +32,32 @@ permissions:
...
 jobs:
+  matrix-prep-packages:
+    name: Generate matrix of packages to prime
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v3.3.0
+        with:
+          fetch-depth: 0
+      - run: sudo apt install -y jq
```
yushao2, line 43:
```suggestion
```

yushao2:
> I would have to look for it, but didn't we use to split the primer run into multiple jobs already some time ago? Seems like this was removed.

https://github.com/PyCQA/pylint/blob/57014d6a45672d76f6a58a9533ff998268525165/.github/workflows/primer-test.yaml#L137

It has gone through quite a major rework, but I doubt the splitting of runs was removed because it was bad, but perhaps happened naturally as a side-effect of the rework of primer tests. cmiiw

cdce8p (changes requested):
I'm -1 on this. It just generates too many jobs. We do already have a lot just for tests, no need to increase that even further.

If at all, the only way forward would be to split each primer job maybe into just 2 or 3 jobs. Still an improvement while keeping the number of jobs lower. Although, not sure that is even worth it at this point.

Maybe others want to add something here. I would especially be interested to know why it was removed previously.

Pierre-Sassoulas:
I suppose the primer results are easier to handle if you don't have to merge it from multiple jobs. But Daniel or Jacob could confirm. Another way to make this less painful would be to make the primer manual and non optional. That way we launch them only when we really need them. It's taking a lot of runner time and we don't need them on every commits. Also as noted by Niels right now it's not aborted when another check fail (only if the other primer fail).

DanielNoord:
I also don't like splitting this up. It just gives more stuff to think about.

I think there are other improvements as noted above that would be more worthwhile.

yushao2:
> I also don't like splitting this up. It just gives more stuff to think about.
> 
> I think there are other improvements as noted above that would be more worthwhile.

okay, I'm happy to shelf this idea and look at other improvements. thanks:)

PR #8135, opened 2023-01-30
[doc] Make sure that doc examples are of a reasonable length

Pierre-Sassoulas:
## Type of Changes

## Description

And upgrade existing unreasonable length. Reading ease is drastically reduced on read the doc after 103 chars (Because of horizontal scrolling)


mbyrnepr2 (approved):
Looks reasonable to me!

PR #8140, opened 2023-01-30
Fix dev requirements

cdce8p:
## Description
Similar to https://github.com/PyCQA/astroid/pull/1993

* Include `requirements_test_pre_commit.txt` in the cache key whenever `requirements_test.txt` is installed.
* Limit `flake8` and `isort` to  Python `>=3.8`.


Pierre-Sassoulas (approved):
Look like very few maintainers are bothering with "old" python interpreters nowadays. 😄

cdce8p:
> Look like very few maintainers are bothering with "old" python interpreters nowadays. 😄

Yeah. Although the `flake8` change was recently reverted to support `3.7` again. Probably only because the marker for `>=3.8` was missing. https://github.com/PyCQA/pylint/commit/5a93635a4a220b16b83b859c2c7f4421ce8d05cd

Issue #6597, opened 2022-05-12
Plain text outputs to stdout invalidate the non-text outputs formatting


yilei:
### Bug description

https://github.com/PyCQA/pylint/blob/d668f641ef5d5794f912cd5a23292bf493290905/pylint/utils/ast_walker.py#L98 prints the exception error message to stdout, making the output invalid.

### Command used

```shell
$ cat myfile.py
def myfunc():
  try:
    @property
    def myfunc():
      return None
  except TypeError:
    pass

  @myfunc.setter
  def myfunc():
    pass

  return myfunc()
$ pylint --output-format=json myfile.py 2>/dev/null
```

### Pylint output

```shell
Exception on node <Call l.13 at 0x7ff8da87f3a0> in file 'myfile.py'
[
    {
        "type": "warning",
        "module": "myfile",
        "obj": "",
        "line": 2,
        "column": 0,
        "endLine": null,
        "endColumn": null,
        "path": "myfile.py",
        "symbol": "bad-indentation",
        "message": "Bad indentation. Found 2 spaces, expected 4",
        "message-id": "W0311"
    },
    {
        "type": "warning",
        "module": "myfile",
        "obj": "",
        "line": 3,
        "column": 0,
        "endLine": null,
        "endColumn": null,
        "path": "myfile.py",
        "symbol": "bad-indentation",
        "message": "Bad indentation. Found 4 spaces, expected 8",
        "message-id": "W0311"
    },
    {
        "type": "warning",
        "module": "myfile",
        "obj": "",
        "line": 4,
        "column": 0,
        "endLine": null,
        "endColumn": null,
        "path": "myfile.py",
        "symbol": "bad-indentation",
        "message": "Bad indentation. Found 4 spaces, expected 8",
        "message-id": "W0311"
    },
    {
        "type": "warning",
        "module": "myfile",
        "obj": "",
        "line": 5,
        "column": 0,
        "endLine": null,
        "endColumn": null,
        "path": "myfile.py",
        "symbol": "bad-indentation",
        "message": "Bad indentation. Found 6 spaces, expected 12",
        "message-id": "W0311"
    },
    {
        "type": "warning",
        "module": "myfile",
        "obj": "",
        "line": 6,
        "column": 0,
        "endLine": null,
        "endColumn": null,
        "path": "myfile.py",
        "symbol": "bad-indentation",
        "message": "Bad indentation. Found 2 spaces, expected 4",
        "message-id": "W0311"
    },
    {
[...]

PR #8142, opened 2023-01-31
Print this error to stderr instead of stdout.

yilei:
## Type of Changes

## Description

Print this error to stderr instead of stdout, otherwise it produces invalid formatted outputs.

This is similar to https://github.com/PyCQA/pylint/issues/6597, please let me know if you want me to file a separate issue.


Pierre-Sassoulas (approved):
Thanks ! How did you encounter the issue ? (I.e. there's a crash in the wild ?)

yilei:
Same as #6597. We have our internal tool that's built on top of pylint, and it uses the stdout for structured lint error outputs (equivalent of --output-format=json).

PR #8151, opened 2023-02-01
Update docs for `global-variable-not-assigned`

### LINKED ISSUE #5073: False positive for global-variable-not-assigned when using with

clavedeluna:
## Type of Changes

## Description

Update explanation for ``global-variable-not-assigned`` and add inference.

Closes #5073

### DIFF
--- a/doc/whatsnew/fragments/5073.other
+++ b/doc/whatsnew/fragments/5073.other
@@ -1,0 +1,3 @@
+Update explanation for ``global-variable-not-assigned`` and add inference.
+
+Closes #5073
--- a/doc/whatsnew/fragments/5073.other
+++ b/doc/whatsnew/fragments/5073.other
@@ -1,3 +1,3 @@
-Update explanation for ``global-variable-not-assigned`` and add inference.
+Update explanation for ``global-variable-not-assigned`` and add confidence.
 
 Closes #5073
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -434,8 +434,9 @@ MSGS: dict[str, MessageDefinitionTuple] = {
     "W0602": (
         "Using global for %r but no assignment is done",
         "global-variable-not-assigned",
-        'Used when a variable is defined through the "global" statement '
-        "but no assignment to this variable is done.",
+        'Used when a name that is previously defined with a "global" statement '
+        "is being invoked with the `global` keyword but is not reassigned in"
+        "the current code block. For example, modifying a global dict.",
     ),
     "W0603": (
         "Using the global statement",  # W0121
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -434,9 +434,9 @@ MSGS: dict[str, MessageDefinitionTuple] = {
     "W0602": (
         "Using global for %r but no assignment is done",
         "global-variable-not-assigned",
-        'Used when a name that is previously defined with a "global" statement '
-        "is being invoked with the `global` keyword but is not reassigned in"
-        "the current code block. For example, modifying a global dict.",
+        "When a variable defined in the global scope is modified in an inner scope, "
+        "the 'global' keyword is required in the inner scope only if there is an "
+        "assignment operation done in the inner scope.",
     ),
     "W0603": (
         "Using the global statement",  # W0121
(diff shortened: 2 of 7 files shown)


pylint/checkers/variables.py
```diff
@@ -434,8 +434,9 @@ def _has_locals_call_after_node(stmt: nodes.NodeNG, scope: nodes.FunctionDef) ->
     "W0602": (
         "Using global for %r but no assignment is done",
         "global-variable-not-assigned",
-        'Used when a variable is defined through the "global" statement '
-        "but no assignment to this variable is done.",
+        'Used when a name that is previously defined with a "global" statement '
+        "is being invoked with the `global` keyword but is not reassigned in"
+        "the current code block. For example, modifying a global dict.",
```
Pierre-Sassoulas, line 439:
```suggestion
        'When a variable defined in the global scope is modified in an inner scope' "
        "the 'global' keyword is required in the inner scope only if there is an "
        "assignment operation done in the inner scope.",
```

Pierre-Sassoulas (changes requested):
I have a suggestion for the new explanation, let's wait for other opinion before applying and regenerating the doc :)

pylint/checkers/variables.py
mbyrnepr2: I like it 👌

pylint/checkers/variables.py
```diff
@@ -434,8 +434,9 @@ def _has_locals_call_after_node(stmt: nodes.NodeNG, scope: nodes.FunctionDef) ->
     "W0602": (
         "Using global for %r but no assignment is done",
         "global-variable-not-assigned",
-        'Used when a variable is defined through the "global" statement '
-        "but no assignment to this variable is done.",
+        "When a variable defined in the global scope is modified in an inner scope"
```
mbyrnepr2, line 437:
```suggestion
        "When a variable defined in the global scope is modified in an inner scope "
```

mbyrnepr2 (changes requested):
Looks good. Just need to add a space to prevent the two words sticking together 😁

pylint/checkers/variables.py
mbyrnepr2: Suggesting a comma too because the sentence is a mouthful

doc/whatsnew/fragments/5073.other
```diff
@@ -0,0 +1,3 @@
+Update explanation for ``global-variable-not-assigned`` and add inference.
```
Pierre-Sassoulas, line 1:
```suggestion
Update explanation for ``global-variable-not-assigned`` and add confidence.
```

Pierre-Sassoulas (approved):
Thank you @clavedeluna !

PR #8157, opened 2023-02-01
Fix some doc about messages

Pierre-Sassoulas:
## Type of Changes

## Description

Saw some issue while reading the doc.

Before:
image

After:
image

### DIFF
--- a/doc/data/messages/u/using-f-string-in-unsupported-version/details.rst
+++ b/doc/data/messages/u/using-f-string-in-unsupported-version/details.rst
@@ -1,1 +1,1 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
+You should upgrade to python 3.6 to use f-strings. Even python 3.6 is past end of life at this point.
--- a/doc/data/messages/b/bad-configuration-section/details.rst
+++ b/doc/data/messages/b/bad-configuration-section/details.rst
@@ -1,1 +1,2 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
+This error was raised when we encountered an unexpected value type in a toml
+configuration between pylint 2.12 and pylint 2.14 (before the argparse refactor).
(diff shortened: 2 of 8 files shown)


doc/data/messages/u/using-f-string-in-unsupported-version/details.rst
```diff
@@ -1 +1 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
+You should upgrade to python 3.6 to use f-strings. Even python 3.6 is past end of life at this point.
```
mbyrnepr2, line 1:
```suggestion
f-strings were introduced in Python version 3.6; to use them, please use a more recent version of Python.
```

Perhaps I'm being pedantic but it sounds a bit like saying you must use Python 3.6 in order to use f-strings. I'm suggesting something more general but feel free to disregard.

mbyrnepr2 (approved):
Looks good to me. Just one small suggestion but you have the deciding vote to use it or not :D

doc/data/messages/u/using-f-string-in-unsupported-version/details.rst
mbyrnepr2: Oh timing! I see its merged just this minute - well no worries :D

Pierre-Sassoulas: Huh, my bad it was auto-merge enabled. I shouldn't do that as it prevent approving with an optional suggestion (something I often do myself).

Pierre-Sassoulas: Do you want to open a MR for your suggestion ? I think it's a lot better too.

mbyrnepr2: I'm happy either way; what you have already isn't wrong at all. Sure I'll create a quick follow-up 👨‍💻

mbyrnepr2: https://github.com/PyCQA/pylint/pull/8180

PR #8185, opened 2023-02-04
Fix a false positive for 'missing-parentheses-for-call-in-test' and add doc example

Pierre-Sassoulas:
## Type of Changes

## Description

Part of #5953 Actual conversation with ChatGPT that led to this MR:

Pierre Sassoulas: Could you give me an example of bad code raising a 'missing-parentheses-for-call-in-test' message in pylint

Sure, here's an example of code that would raise the "missing-parentheses-for-call-in-test" message in Pylint:

```python
def example_func():
    return "example return value"

if example_func:
    print("example_func is defined")
```
In this code, the line if example_func: raises the "missing-parentheses-for-call-in-test" message, because Pylint is expecting to see a function call in the test expression of an if statement, but there are no parentheses after example_func. To correct this, you can add parentheses to make it a function call:

```
if example_func():
    print("example_func is defined")
```

Pierre Sassoulas
Could you make it more lively, the name is very generic ?

Sure, here's a more lively example that also includes a more descriptive function name:

```python
def say_hello():
    return "Hello, world!"

if say_hello:
    print("say_hello is defined")
```
In this code, the line if say_hello: raises the "missing-parentheses-for-test-call-in-if" message, because Pylint is expecting to see a function call in the test expression of an if statement, but there are no parentheses after say_hello. To correct this, you can add parentheses to make it a function call:

```
if say_hello():
    print("say_hello is defined")
```

Pierre Sassoulas
That's kinda expected and boring, could you be more imaginative ? Also I don't think it should return a string, it's supposed to return a boolean.

Sure, here's a more imaginative example with a function that returns a boolean:

```python

def is_it_a_good_day():
    import random
    return random.choice([True, False])

if is_it_a_good_day:
    print("Today is a good day!")
```

In this code, the line if is_it_a_good_day: raises the "missing-parentheses-for-call-in-test" message, because Pylint is expecting to see a function call in the test expression of an if statement, but there are no parentheses after is_it_a_good_day. To correct this, you can add parentheses to make it a function call:

```
if is_it_a_good_day():
    print("Today is a good day!")
```

Pierre Sassoulas
You need to add a ``# [missing-parentheses-for-call-in-test]`` on the line the actual message would be raised, could you update the bad example, please ?

Of course! Here's the updated bad example with a comment indicating the location of the "missing-parentheses-for-call-in-test" message:

```python
def is_it_a_good_day():
    import random
    return random.choice([True, False])

if is_it_a_good_day:  # [missing-parentheses-for-call-in-test]
    print("Today is a good day!")
```
In this code, the line if is_it_a_good_day: would raise the "missing-parentheses-for-call-in-test" message, because Pylint is expecting to see a function call in the test expression of an if statement, but there are no parentheses after is_it_a_good_day. To correct this, you can add parentheses to make it a function call, as shown in the corrected code in my previous answer.


Pierre-Sassoulas:
This started as a fun chat GPT fueled adventure and then I did not manage to make it raise like it should and went on another more painful adventure.

pylint/checkers/base/basic_checker.py
```diff
@@ -370,14 +370,15 @@ def _check_using_constant_test(
...
-                            self.add_message(
-                                "missing-parentheses-for-call-in-test",
-                                node=test,
-                                confidence=INFERENCE,
-                            )
-                            break
+                    # Just forcing the generator to infer all elements.
+                    # astroid.exceptions.InferenceError are false positives
+                    # see https://github.com/PyCQA/pylint/pull/8185
+                    list(call_inferred)
```
DanielNoord, line 376: This is a nit, but can't we do the unpacking above where we are already catching `InferenceError`?

tests/functional/m/missing/missing_parentheses_for_call_in_test.py
```diff
@@ -1,19 +1,28 @@
 """Verify if call to function or method inside tests are missing parentheses."""
 # pylint: disable=using-constant-test, missing-docstring
 # pylint: disable=invalid-name, expression-not-assigned, unnecessary-lambda-assignment
+import os
```
DanielNoord, line 4:
```suggestion

import os
```

DanielNoord (approved):
Two nits! Love this adventure 😄

pylint/checkers/base/basic_checker.py
Pierre-Sassoulas: Sure, that not a nit that's an optimization 😄

tests/functional/m/missing/missing_parentheses_for_call_in_test.py
DanielNoord: This was a nit though.. 😅

pylint/checkers/base/basic_checker.py
```diff
@@ -366,20 +366,19 @@ def _check_using_constant_test(
                     call_inferred = inferred.infer_call_result()
                 elif isinstance(inferred, nodes.Lambda):
                     call_inferred = inferred.infer_call_result(node)
```
DanielNoord, line 368:
```suggestion
                    call_inferred = list(inferred.infer_call_result(node))
```

Why not just do it in place?

Pierre-Sassoulas: Can't it be None ? Or is this just for lambdas ?

DanielNoord: As far as I know `infer_call_result` should always return a `generator`. If not, that is something we should fix.

I think we can just immediately unpack on L366 and L368.

Pierre-Sassoulas: We would get a (false positive I think) message on:
```
    if sched_getaffinity:
        cpu_count = len(sched_getaffinity(0))
```

DanielNoord: We also need the `list` on line 366

Pierre-Sassoulas: I reverted

pylint/checkers/base/basic_checker.py
```diff
@@ -366,20 +366,19 @@ def _check_using_constant_test(
...
+                # Just forcing the generator to infer all elements.
+                # astroid.exceptions.InferenceError are false positives
+                # see https://github.com/PyCQA/pylint/pull/8185
+                if call_inferred is not None:
+                    call_inferred = list(call_inferred)
```
DanielNoord, line 373:
```suggestion
                    call_inferred = list(inferred.infer_call_result())
                elif isinstance(inferred, nodes.Lambda):
                    call_inferred = list(inferred.infer_call_result(node))
```

I can't suggest to place the comment above here within the Github UI, but this should be functionally the same right?

pylint/checkers/base/basic_checker.py
Pierre-Sassoulas: Duh, of course, this work.

DanielNoord (approved):

PR #8186, opened 2023-02-04
Add import-based submodule detection for preferred-modules

d33bs:
## Type of Changes

## Description

This PR adds new functionality with preferred-modules configuration to detect submodules (along with relevant documentation and tests). The focus was intended to enhance the existing design via the `imports.py` without losing the original functionality.

I very much appreciate any comments or suggestions towards improving the work in this PR!

Note: I feel that #7957 may still need more work to close beyond this change to possibly detect submodule implementation when a "parent" module has been imported (without a direct submodule import statement to reference). It felt that non-import-specific submodule detection might need to be referenced or implemented outside of `imports.py` but I wasn't sure exactly where this would fit. Thank you in advance for any thoughts or feedback on this topic specifically!

Refs #7957

### DIFF
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -914,11 +914,20 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
 
     def _check_preferred_module(self, node: ImportNode, mod_path: str) -> None:
         """Check if the module has a preferred replacement."""
-        if mod_path in self.preferred_modules:
+
+        if isinstance(node, astroid.nodes.node_classes.ImportFrom):
+            mod_compare = [f"{node.modname}.{name[0]}" for name in node.names]
+        else:
+            mod_compare = [mod_path]
+
+        if len(set(mod_compare).intersection(self.preferred_modules)) > 0:
+            mod_path_intersect = list(
+                set(mod_compare).intersection(self.preferred_modules)
+            )[0]
             self.add_message(
                 "preferred-module",
                 node=node,
-                args=(self.preferred_modules[mod_path], mod_path),
+                args=(self.preferred_modules[mod_path_intersect], mod_path_intersect),
             )
 
     def _check_import_as_rename(self, node: ImportNode) -> None:
--- a/pylint/checkers/imports.py
+++ b/pylint/checkers/imports.py
@@ -915,19 +915,28 @@ class ImportsChecker(DeprecatedMixin, BaseChecker):
     def _check_preferred_module(self, node: ImportNode, mod_path: str) -> None:
         """Check if the module has a preferred replacement."""
 
+        mod_compare = [mod_path]
+        # build a comparison list of possible names using importfrom's
         if isinstance(node, astroid.nodes.node_classes.ImportFrom):
             mod_compare = [f"{node.modname}.{name[0]}" for name in node.names]
-        else:
-            mod_compare = [mod_path]
 
-        if len(set(mod_compare).intersection(self.preferred_modules)) > 0:
-            mod_path_intersect = list(
-                set(mod_compare).intersection(self.preferred_modules)
-            )[0]
+        # find whether there are matches with the import vs preferred_modules keys
+        matches = [
+            key
+            for key in list(self.preferred_modules.keys())
+            for compare_val in mod_compare
+            # determine if key is in compare_val only if key is smaller
+            if (len(compare_val) < len(key) and key in compare_val)
+            # if key and compare_val are exact matches
+            or key == compare_val
+        ]
+
+        # if we have matches, add message
+        if len(matches) > 0:
             self.add_message(
                 "preferred-module",
                 node=node,
-                args=(self.preferred_modules[mod_path_intersect], mod_path_intersect),
+                args=(self.preferred_modules[matches[0]], matches[0]),
             )
 
     def _check_import_as_rename(self, node: ImportNode) -> None:
--- a/tests/checkers/unittest_imports.py
+++ b/tests/checkers/unittest_imports.py
@@ -123,7 +123,7 @@ class TestImportsChecker(CheckerTestCase):
         # test preferred-modules case with base module import
         Run(
             [
-                f"{os.path.join(REGR_DATA, 'preferred_module')}",
+                f"{os.path.join(REGR_DATA, 'preferred_module/unpreferred_module.py')}",
                 "-d all",
                 "-e preferred-module",
                 # prefer sys instead of os (for triggering test)
@@ -135,6 +135,24 @@ class TestImportsChecker(CheckerTestCase):
 
         # assert that we saw preferred-modules triggered
         assert "Prefer importing 'sys' instead of 'os'" in output
+        # assert there were no errors
+        assert len(errors) == 0
+
+        # test preferred-modules case with submodules
+        Run(
+            [
+                f"{os.path.join(REGR_DATA, 'preferred_module/unpreferred_submodule.py')}",
+                "-d all",
+                "-e preferred-module",
+                # prefer sys instead of os (for triggering test)
+                "--preferred-modules=os.path:pathlib",
+            ],
+            exit=False,
+        )
+        output, errors = capsys.readouterr()
+
+        # assert that we saw preferred-modules triggered
+        assert "Prefer importing 'pathlib' instead of 'os.path'" in output
         # assert there were no errors
         assert len(errors) == 0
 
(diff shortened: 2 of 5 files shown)


DanielNoord:
@d33bs Do you need help investigating the failing test case?

d33bs:
Thank you @DanielNoord - yes, I may need help here. I am able to see there's something expected by perhaps the message in testdata but I'm not sure how to fix. I'm wondering but not sure if the change for items which could include `.` in the module name could have triggered this.

DanielNoord:
> Thank you @DanielNoord - yes, I may need help here. I am able to see there's something expected by perhaps the message in testdata but I'm not sure how to fix. I'm wondering but not sure if the change for items which could include `.` in the module name could have triggered this.

The issue is in https://github.com/PyCQA/pylint/blob/8d13dbe9d82da64f21c5a62b7c16d2332839a346/tests/functional/p/preferred_module.py#L6

The file indicates that we expect to raise a message on line 5, but with this PR we no longer do. You could try and run only that test and debug why there is a change in behaviour.

d33bs:
Thank you @DanielNoord, this was helpful! I changed the functionality to acknowledge the existing functional test for preferred-modules.

tests/checkers/unittest_imports.py
```diff
@@ -123,7 +123,7 @@ def test_preferred_module(capsys: CaptureFixture[str]) -> None:
         # test preferred-modules case with base module import
         Run(
             [
-                f"{os.path.join(REGR_DATA, 'preferred_module')}",
+                f"{os.path.join(REGR_DATA, 'preferred_module/unpreferred_module.py')}",
```
DanielNoord, line 126: Why was this changed?

tests/checkers/unittest_imports.py
```diff
@@ -138,6 +138,42 @@ def test_preferred_module(capsys: CaptureFixture[str]) -> None:
...
         assert len(errors) == 0
 
+        # test preferred-modules case with submodules
+        Run(
+            [
+                f"{os.path.join(REGR_DATA, 'preferred_module/unpreferred_submodule.py')}",
+                "-d all",
+                "-e preferred-module",
+                # prefer sys instead of os (for triggering test)
+                "--preferred-modules=os.path:pathlib",
```
DanielNoord, line 148: Could we also do a test with `os.something_that_is_not_path` and see if we don't still suggest it?

DanielNoord (changes requested):
Glad to see you figured it out! The `pylint` failure can be solved by adding some wordt to our internal dictionary.

Really like this change itself!

tests/checkers/unittest_imports.py
d33bs: I added a specific `.py` file for submodule tests under that directory. The added specification here is for the file with the "base" module test so as to not read in the submodule specific test file.

tests/checkers/unittest_imports.py
d33bs: Great idea, thank you! I'll add this into a further commit.

d33bs: Added in [9f42701](https://github.com/PyCQA/pylint/pull/8186/commits/9f4270151fde697ed8913818750435e9879cba3a)

pylint/checkers/imports.py
```diff
@@ -914,11 +914,27 @@ def _add_imported_module(self, node: ImportNode, importedmodname: str) -> None:
 
     def _check_preferred_module(self, node: ImportNode, mod_path: str) -> None:
         """Check if the module has a preferred replacement."""
-        if mod_path in self.preferred_modules:
+
+        mod_compare = [mod_path]
+        # build a comparison list of possible names using importfrom
+        if isinstance(node, astroid.nodes.node_classes.ImportFrom):
```
DanielNoord, line 920: This should always be true based on the typing. Is the typing of the signature incorrect or can we remove this check?

d33bs: Astroid provides an [`ImportFrom`](https://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.ImportFrom) as well as an [`Import`](https://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.Import) node. This if case was added in for `from ... import ...` cases where a comparison string was necessary to match from the preferred-modules pairs involving submodules with "dot" notation (for ex. `"module.submodule"` str construction). Otherwise with `Import` cases we only need the module name.

DanielNoord: Oh my, I misread `ImportNode` for `ImportFrom`. I'm stupid.. Thanks for this!

DanielNoord (approved):
Thanks @d33bs! 

@Pierre-Sassoulas Do you want to review as well?

pylint/checkers/imports.py
```diff
@@ -914,11 +914,27 @@ def _add_imported_module(self, node: ImportNode, importedmodname: str) -> None:
...
+        matches = [
+            key
+            for key in list(self.preferred_modules.keys())
+            for compare_val in mod_compare
+            # determine if key is in compare_val
+            if key in compare_val
+        ]
+
+        # if we have matches, add message
+        if len(matches) > 0:
```
Pierre-Sassoulas, line 933:
```suggestion
        if matches:
```
We have a check for that, it must be a false negative.

pylint/checkers/imports.py
```diff
@@ -914,11 +914,27 @@ def _add_imported_module(self, node: ImportNode, importedmodname: str) -> None:
...
+            mod_compare = [f"{node.modname}.{name[0]}" for name in node.names]
+
+        # find whether there are matches with the import vs preferred_modules keys
+        matches = [
+            key
+            for key in list(self.preferred_modules.keys())
+            for compare_val in mod_compare
+            # determine if key is in compare_val
+            if key in compare_val
+        ]
```
Pierre-Sassoulas, line 930:
```suggestion
        matches = [
            k
            for k in self.preferred_modules
            for mod in mod_compare
            if k in mod
        ]
```
Shouldn't this work ? when iterating on a dict the default is ``keys()`

tests/checkers/unittest_imports.py
```diff
@@ -138,6 +138,55 @@ def test_preferred_module(capsys: CaptureFixture[str]) -> None:
...
+            ],
+            exit=False,
+        )
+        output, errors = capsys.readouterr()
+
+        # assert that we saw preferred-modules triggered with base module
+        assert "Prefer importing 'pathlib' instead of 'os'" in output
+        # assert there were no errors
+        assert len(errors) == 0
+
```
Pierre-Sassoulas, line 189: This could be a small functional tests implemented in 10 lines with exact match and automated upgrade if something was to change. @DanielNoord are those unit test really faster or should we move everything to functional ?

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas:
Thank you for working on this @d33bs, this bug has been opened for a loooong time !

tests/checkers/unittest_imports.py
DanielNoord: In this case the unit tests aren't much faster, in general they probably are because there is less overhead.

Anyway, I don't really mind these as they built upon the previously existing test but I understand your point.

pylint/checkers/imports.py
d33bs: No worries at all! Thank you for calling it forward just the same.

pylint/checkers/imports.py
d33bs: Thank you, yes I believe you're correct, and this looks much cleaner. I'll add this into a forthcoming commit.

pylint/checkers/imports.py
d33bs: Thank you! I'll get this updated.

d33bs:
Thank you @Pierre-Sassoulas for the review and suggestions! I've applied the changes you suggested and also added two missing asserts for one of the test cases. Please let me know if I may add or update anything.

Pierre-Sassoulas (approved):
Great ! The strange comment I made previously about the issue being opened "for a long time" is because I mistaken that with https://github.com/PyCQA/pylint/issues/5 (Looks similar, let us know if this is the same kind of issue :)

ssbarnea:
I have reasons to believe that this change introduced a major regression, filed at https://github.com/PyCQA/pylint/issues/8453

PR #8188, opened 2023-02-05
Minor improvement to unittest_utils.py

andrehora:
Function `decoding_stream` may receive an empty encoding (see line 173):

https://github.com/PyCQA/pylint/blob/a57edea63fa7a1edfcb63074f9e2a373ebabb7cd/pylint/utils/utils.py#L167-L176

... but this case is untested `unittest_utils.py`.

Actually, the test `unittest_utils.py` checks two cases:
- test_decoding_stream_unknown_encoding
- test_decoding_stream_known_encoding 

Thus, this PR tests the untested case in `test_decoding_stream_empty_encoding`.

### DIFF
--- a/tests/utils/unittest_utils.py
+++ b/tests/utils/unittest_utils.py
@@ -18,6 +18,17 @@ def test_decoding_stream_unknown_encoding() -> None:
     assert ret == ["foo\n", "bar"]
 
 
+def test_decoding_stream_empty_encoding() -> None:
+    """Decoding_stream should fall back to *some* decoding when given an
+    empty encoding.
+    """
+    binary_io = io.BytesIO(b"foo\nbar")
+    stream = utils.decoding_stream(binary_io, "")
+    # should still act like a StreamReader
+    ret = stream.readlines()
+    assert ret == ["foo\n", "bar"]
+
+
 def test_decoding_stream_known_encoding() -> None:
     binary_io = io.BytesIO("€".encode("cp1252"))
     stream = utils.decoding_stream(binary_io, "cp1252")


Pierre-Sassoulas (commented):
Thank you for opening a pull request. I don't think having tests covering all tests cases on util function is super valuable. In fact there's very little chance that we change this part of the code and it's going to make our test suite slower all the time forever. It's also testing ``codecs.getreader`` and ``sys.getdefaultencoding())`` two functions that we did not develop ourselves.

I appreciate the willingness to work on pylint though, you can find good first issues [here](https://github.com/PyCQA/pylint/issues?q=is%3Aopen+is%3Aissue+label%3A%22Good+first+issue%22)

andrehora:
Thanks @Pierre-Sassoulas for the feedback.

Just would like to provide a few comments on the reasons I created this *minor* test:

1) `decoding_stream` is public, thus, it should be well-tested.
2) `decoding_stream` accepts three types of inputs: (1) known encoding, (2) unknown encoding, and (3) empty encoding. The first two are tested. The last one is not tested. This PR fixes it.
3)  although `decoding_stream` has 100% line coverage, it misses the input (3). Indeed, high coverage does not necessarily mean that the test is perfect or cover all relevant cases.
4) this PR is not trying to cover all possible cases (like you mentioned). It is only covering the three relevant cases accepted by `decoding_stream`.
5) the fact that a function is unlikely to change does not mean (IMHO) that it should not be properly tested.

Pierre-Sassoulas:
I hear your points. There are a lot of functions that could be considered public api but really aren't (because they were not initially marked as private, it doesn't mean waiting a full deprecation cycle to change them is reasonable). Also we do not have full coverage on essential functions (check the crash in 2.16.1 due to uncovered code) we do not  have typing in astroid and we have 650 opened issue in pylint + 125 in astroid, some with tons of impact. I'd rather focus mainteners' limited review time on impactful changes.

PR #8191, opened 2023-02-05
[return-arg-in-generator] Add a details.rst only as the message is not raised anymore

Pierre-Sassoulas:
## Type of Changes

## Description

Blocked by the inclusion of deleted messages in the doc (#6670) or the inclusion of a new message template #7897 

Refs #5953

### DIFF
--- a/doc/data/messages/r/return-arg-in-generator/details.rst
+++ b/doc/data/messages/r/return-arg-in-generator/details.rst
@@ -1,0 +1,16 @@
+This is a message that isn't going to be raised for python > 3.3. It was raised
+for code like::
+
+    def interrogate_until_you_find_jack(pirates):
+        for pirate in pirates:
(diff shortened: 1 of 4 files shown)


ollie-iterators:
If you are doing things that only need details.rst, I believe that the invalid-unicode-codec one only needs deatils.rst because it is about the encoding of the file.

Pierre-Sassoulas:
See https://pylint--8191.org.readthedocs.build/en/8191/user_guide/messages/error/return-arg-in-generator.html

mbyrnepr2:
Savvy! Although could probably do with removing the word “outside” from the description (not part of the current MR).

Pierre-Sassoulas:
Nice catch, I fixed it :)

mbyrnepr2 (approved):

PR #8182, opened 2023-02-03
docs: Add documentation for six pylint errors

ollie-iterators:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |
| ✓   | :sparkles: New feature |
| ✓   | :hammer: Refactoring   |
| ✓   | :scroll: Docs          |

## Description

This adds documentation for the following errors in pylint from pylint-errors:

- bad-mcs-classmethod-argument
- bad-mcs-method-argument
- inconsistent-mro
- docstring-first-line-empty
- differing-param-doc
- differing-type-doc

Refs [#5953](https://github.com/PyCQA/pylint/issues/5953)

Closes #XXXX

PR #8202, opened 2023-02-06
Add documentation examples

DanielNoord:
## Type of Changes

## Description

Supersedes https://github.com/PyCQA/pylint/pull/8182

Let's not let this work go to waste, but also let's not keep reviewing a PR with 140 commits for 6 doc examples.

### DIFF
--- a/doc/data/messages/b/bad-mcs-classmethod-argument/bad.py
+++ b/doc/data/messages/b/bad-mcs-classmethod-argument/bad.py
@@ -1,0 +1,4 @@
+class Meta(type):
+    @classmethod
+    def foo(some):  # [bad-mcs-classmethod-argument]
+        pass
--- a/doc/data/messages/b/bad-mcs-classmethod-argument/details.rst
+++ b/doc/data/messages/b/bad-mcs-classmethod-argument/details.rst
@@ -1,1 +1,0 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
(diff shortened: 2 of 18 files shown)


Pierre-Sassoulas (approved):
Thank you for the cleanup @DanielNoord.

DanielNoord:
@Pierre-Sassoulas Are the examples good enough for you? It feels a bit `foo ... bar`

ollie-iterators:
Why does the message about the error happening for the docstring errors like "docstring-first-line-empty" not go on the docstring, and instead go on the line above it?

DanielNoord:
> Why does the message about the error happening for the docstring errors like "docstring-first-line-empty" not go on the docstring, and instead go on the line above it?

The message gets raised on the function definition, as that is the object with the incorrect docstring. This is a bit of a side effect of `pylint` relying heavily on the Abstract Syntax Tree in which a `FunctionDef` node is much easier to access than a docstring which doesn't really have its own node.

Pierre-Sassoulas:
Yeah, well in a "normal" merge request I would have asked about it, but this one was already exhausting as is. I think we can deal with this later on (it's easy to grep foo and bar in the doc). Or we could change it but this would be almost redoing everything from zero...

ollie-iterators:
> > Why does the message about the error happening for the docstring errors like "docstring-first-line-empty" not go on the docstring, and instead go on the line above it?
> 
> The message gets raised on the function definition, as that is the object with the incorrect docstring. This is a bit of a side effect of `pylint` relying heavily on the Abstract Syntax Tree in which a `FunctionDef` node is much easier to access than a docstring which doesn't really have its own node.

Ok, thanks for the explanation

PR #8206, opened 2023-02-07
Fix invalid type false positive

### LINKED ISSUE #8205: `W1116 (isinstance-second-argument-not-valid-type)`  false positive for union types
### Bug description

```python
assert isinstance(0, int | str)  # ✘ raises W1116
```

### Configuration

_No response_

### Command used

```shell
pylint
```

### Pylint output

```shell
************* Module isinstance_second_argument_not_valid_type
bugs/pylint/isinstance_second_argument_not_valid_type.py:1:0: C0114: Missing module docstring (missing-module-docstring)
bugs/pylint/isinstance_second_argument_not_valid_type.py:5:7: W1116: Second argument of isinstance is not a type (isinstance-second-argument-not-valid-type)

```

### Expected behavior

Union types are supported by `isinstance` and `issubclass` since 3.10 https://docs.python.org/3/library/functions.html#isinstance

### Pylint version

```shell
pylint 2.16.1
astroid 2.14.1
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

nickdrozd:
## Type of Changes

## Description

Fixes false positive for `isinstance-second-argument-not-valid-type` with union types.

Closes #8205

### DIFF
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -24,6 +24,7 @@ import astroid.helpers
 from astroid import arguments, bases, nodes
 from astroid.typing import InferenceResult, SuccessfulInferenceResult
 
+from pylint.constants import PY310_PLUS
 from pylint.checkers import BaseChecker, utils
 from pylint.checkers.utils import (
     decorated_with,
@@ -86,9 +87,6 @@ class VERSION_COMPATIBLE_OVERLOAD:
 
 
 VERSION_COMPATIBLE_OVERLOAD_SENTINEL = VERSION_COMPATIBLE_OVERLOAD()
-
-
-PY310_PLUS = sys.version_info >= (3, 10)
 
 
 def _unflatten(iterable: Iterable[_T]) -> Iterator[_T]:
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -796,6 +796,10 @@ def _is_c_extension(module_node: InferenceResult) -> bool:
 
 def _is_invalid_isinstance_type(arg: nodes.NodeNG) -> bool:
     # Return True if we are sure that arg is not a type
+    if isinstance(arg, nodes.BinOp) and arg.op == "|":
+        return _is_invalid_isinstance_type(arg.left) or _is_invalid_isinstance_type(
+            arg.right
+        )
     inferred = utils.safe_infer(arg)
     if not inferred:
         # Cannot infer it so skip it.
@@ -806,10 +810,6 @@ def _is_invalid_isinstance_type(arg: nodes.NodeNG) -> bool:
         return False
     if isinstance(inferred, astroid.Instance) and inferred.qname() == BUILTIN_TUPLE:
         return False
-    if isinstance(arg, nodes.BinOp) and arg.op == "|":
-        return _is_invalid_isinstance_type(arg.left) or _is_invalid_isinstance_type(
-            arg.right
-        )
     return True
 
 
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -1407,7 +1407,11 @@ accessed. Python regular expressions are accepted.",
 
         second_arg = node.args[1]
         if _is_invalid_isinstance_type(second_arg):
-            self.add_message("isinstance-second-argument-not-valid-type", node=node)
+            self.add_message(
+                "isinstance-second-argument-not-valid-type",
+                node=node,
+                confidence=INFERENCE,
+            )
 
     # pylint: disable = too-many-branches, too-many-locals, too-many-statements
     def visit_call(self, node: nodes.Call) -> None:
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -24,7 +24,6 @@ import astroid.helpers
 from astroid import arguments, bases, nodes
 from astroid.typing import InferenceResult, SuccessfulInferenceResult
 
-from pylint.constants import PY310_PLUS
 from pylint.checkers import BaseChecker, utils
 from pylint.checkers.utils import (
     decorated_with,
@@ -49,6 +48,7 @@ from pylint.checkers.utils import (
     supports_membership_test,
     supports_setitem,
 )
+from pylint.constants import PY310_PLUS
 from pylint.interfaces import HIGH, INFERENCE
 from pylint.typing import MessageDefinitionTuple
 
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -810,6 +810,10 @@ def _is_invalid_isinstance_type(arg: nodes.NodeNG) -> bool:
         return False
     if isinstance(inferred, astroid.Instance) and inferred.qname() == BUILTIN_TUPLE:
         return False
+    if isinstance(inferred, bases.UnionType):
+        return _is_invalid_isinstance_type(
+            inferred.left
+        ) or _is_invalid_isinstance_type(inferred.right)
     return True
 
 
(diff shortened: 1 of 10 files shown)


tests/functional/i/isinstance_second_argument_py310.py
```diff
@@ -0,0 +1,5 @@
+# pylint: disable = missing-docstring
```
DanielNoord, line 1: Could you please add a docstring?

pylint/checkers/typecheck.py
```diff
@@ -806,6 +806,10 @@ def _is_invalid_isinstance_type(arg: nodes.NodeNG) -> bool:
         return False
     if isinstance(inferred, astroid.Instance) and inferred.qname() == BUILTIN_TUPLE:
         return False
+    if isinstance(arg, nodes.BinOp) and arg.op == "|":
```
DanielNoord, line 809: Can't we use the new `UnionType` here?

DanielNoord (changes requested):
This will need a changelog entry

pylint/checkers/typecheck.py
orSolocate, line 809: I reckon this check should only be done for python version that support PEP604 and the `|` operand for type annotations.  I think this condition should only be checked when supported, similar code in `pylint/extensions/typing.py`

```suggestion
    if _should_check_alternative_union_syntax and isinstance(arg, nodes.BinOp) and arg.op == "|":
```

where:
```
        py_version = self.linter.config.py_version
        self._py37_plus = py_version >= (3, 7)
        self._py310_plus = py_version >= (3, 10)
        self._should_check_alternative_union_syntax = self._py310_plus or (
            self._py37_plus and self.linter.config.runtime_typing is False

```

tests/functional/i/isinstance_second_argument_py310.rc
```diff
@@ -0,0 +1,2 @@
+[testoptions]
+min_pyver=3.10
```
orSolocate, line 2: maybe you could allow the test to run also for earlier versions like was done in `redundant-typehint-argument`:
```suggestion
[testoptions]
min_pyver=3.7

[TYPING]
runtime-typing=no
```

pylint/checkers/typecheck.py
nickdrozd: Because of what I assume are inference failures, I couldn't cover all the test cases without checking both the inferred value and the superficial value. Is there a better way to do this?

nickdrozd: Agreed, I'll add that once the rest is figured out. And maybe a shorter name would be nice :laughing:

nickdrozd:
I included test cases for false negatives from https://github.com/PyCQA/pylint/issues/8213.

pylint/checkers/typecheck.py
```diff
@@ -806,6 +810,10 @@ def _is_invalid_isinstance_type(arg: nodes.NodeNG) -> bool:
         return False
     if isinstance(inferred, astroid.Instance) and inferred.qname() == BUILTIN_TUPLE:
         return False
+    if isinstance(inferred, bases.UnionType):
```
Pierre-Sassoulas, line 813: Wondering if we should put this line 806 before the Tuple check for performance reason. Wet finger estimation but I suppose union happens more often (?)

tests/functional/i/isinstance_second_argument_py310.py
```diff
@@ -0,0 +1,27 @@
...
+isinstance(0, (float | ListOrDict) | IntOrStr)
+
+# True positives
+isinstance(0, int | 5)  # [isinstance-second-argument-not-valid-type]
+isinstance(0, str | 5 | int)  # [isinstance-second-argument-not-valid-type]
+INT = 5
+isinstance(0, INT | int)  # [isinstance-second-argument-not-valid-type]
+
+
+# FALSE NEGATIVES
```
Pierre-Sassoulas, line 21: 👍 I like that tests are already where pylint is not, yet

pylint/checkers/typecheck.py
nickdrozd: The union syntax is pretty new, and I wouldn't think adoption is all that widespread. For example, it's not used anywhere in Pylint code because of compatibility with old versions.

pylint/checkers/typecheck.py
orSolocate: maybe `_alt_union_syntax_enabled` ?

You can also rename it in `pylint/extensions/typing.py`

pylint/checkers/typecheck.py
Pierre-Sassoulas: Make sense !

Pierre-Sassoulas (dismissed):

tests/functional/i/isinstance_second_argument_py310.rc
nickdrozd: What does the `runtime-typing` option do? Also, is there a need to specify 3.7 as the min version, since that's already the min version being used?

nickdrozd: Now that I think about it, this check is different because it doesn't deal with type annotations. In this case the types involved will actually be called, so I think this option won't apply.

pylint/checkers/typecheck.py
```diff
@@ -88,6 +88,9 @@ class VERSION_COMPATIBLE_OVERLOAD:
 VERSION_COMPATIBLE_OVERLOAD_SENTINEL = VERSION_COMPATIBLE_OVERLOAD()
 
 
+PY310_PLUS = sys.version_info >= (3, 10)
```
orSolocate, line 91: Found this cool def already:

`from astroid.const import PY310_PLUS`

tests/functional/i/isinstance_second_argument_py310.rc
orSolocate: you don't have to explicitly write `min_pyver=3.7` but it would be just clearer for the test reader to understand why `runtime-typing` was used. There is some documentation `runtime-typing` at `pylint/extensions/typing.py`. It will identify type annotations if it can for python versions 3.7-3.9 with the `|` operation.

> Now that I think about it, this check is different because it doesn't deal with type annotations. In this case the types involved will actually be called, so I think this option won't apply.

I think you are right. These are not annotations it the actual code. so if someone uses the `|` operation they can only do that in py3.10+.

pylint/checkers/typecheck.py
```diff
@@ -806,6 +813,10 @@ def _is_invalid_isinstance_type(arg: nodes.NodeNG) -> bool:
         return False
     if isinstance(inferred, astroid.Instance) and inferred.qname() == BUILTIN_TUPLE:
         return False
+    if PY310_PLUS and isinstance(inferred, bases.UnionType):
```
orSolocate, line 816:
```suggestion
    if isinstance(inferred, bases.UnionType):
```
Is the `bases.UnionType` the legacy Union/Optional type? if so it is supported from 3.7+ I think.

pylint/checkers/typecheck.py
Pierre-Sassoulas: Let's use the one in pylint :)

pylint/checkers/typecheck.py
nickdrozd: No, `Union[int, str]` is an instance of `ClassDef` and not `UnionType`. I'm not sure what exactly `UnionType` corresponds to.

pylint/checkers/typecheck.py
nickdrozd: py310 wasn't defined in constants, so I added it there.

Pierre-Sassoulas: Let's define it here then: https://github.com/PyCQA/pylint/blob/main/pylint/constants.py#L21

Pierre-Sassoulas (approved):
👍

nickdrozd:
@DanielNoord Anything else for this one?

DanielNoord (commented):
Could you add `INFERENCE` as confidence to the message?>

Pierre-Sassoulas (approved):

DanielNoord (approved):
Thanks for this! 😄

PR #8219, opened 2023-02-07
Add too-many-lines

ollie-iterators:
## Type of Changes

## Description

Adding documentation for too-many-lines

Refs #5953 

Closes #XXXX


doc/data/messages/t/too-many-lines/details.rst
```diff
@@ -1 +1 @@
-You can help us make the doc better `by contributing <https://github.com/PyCQA/pylint/issues/5953>`_ !
+Used when a module has too many lines, reducing its readability.
```
Pierre-Sassoulas, line 1: You really need to **actually** check the result before pushing. I told you at least twice already, the description from the checker is automatically added to the doc. What does the checker contain ? "Used when a module has too many lines, reducing its readability." Now what is going to be in the result ? You should generate locally but here goes:

```
Description:

Used when a module has too many lines, reducing its readability.

....

Additional details:

Used when a module has too many lines, reducing its readability.

Created by the [format](https://github.com/PyCQA/pylint/blob/main/pylint/checkers/format.py) checker.
```
So what's wrong here ? Yes, the text is duplicated. Do I have to point this out each time ?

Also we have 40 concurrent runners for the whole organization, wasting runner time by pushing new commits all the time before event checking the result locally is impacting **dozen of persons**, please be more considerate

Pierre-Sassoulas (commented):
The code in good.py and bad.py should have an equivalent behavior, it hard to have this constraint for this message so it should be a special case where we don't have a good/bad example done in https://github.com/PyCQA/pylint/issues/7897

PR #8223, opened 2023-02-07
Add `--color-palette` option to `pyreverse`

### LINKED ISSUE #6738: Custom colour palettes for `pyreverse`
### Current problem

See [this SO question](https://stackoverflow.com/questions/72294217/pyreverse-uml-diagram-with-customized-color): 
The `--colorized` option uses a fixed colour palette. 
If one wants to use other colours, (s)he has to use a text-based output format and do a search-and-replace.

### Desired solution

As the colour palette is just a simple list of strings inside `pylint.pyreverse.writer.DiagramWriter` it should be possible to make this configurable.

### Additional context

_No response_

DudeNr33:
## Type of Changes

## Description

* Add support for user-defined color palettes
* Use a new default color palette, that is (hopefully) more visually pleasing and more accessible for color blind people. This palette is pretty close to what `seaborn` uses in their colorblind palette (see [here](https://seaborn.pydata.org/tutorial/color_palettes.html#qualitative-color-palettes))

The new default colors look like this:
image

Closes #6738

### DIFF
--- a/.pyenchant_pylint_custom_dict.txt
+++ b/.pyenchant_pylint_custom_dict.txt
@@ -281,6 +281,7 @@ rstrip
 rtype
 runtime
 se
+seaborn
 sep
 setcomp
 shortstrings
--- a/doc/whatsnew/fragments/6738.feature
+++ b/doc/whatsnew/fragments/6738.feature
@@ -1,0 +1,4 @@
+`pyreverse` now supports custom color palettes with the `--color-palette` option.
+The default color palette has been changed to be more accessible for color blind people.
+
+Closes #6738
--- a/doc/whatsnew/fragments/6738.feature
+++ b/doc/whatsnew/fragments/6738.feature
@@ -1,4 +1,3 @@
 `pyreverse` now supports custom color palettes with the `--color-palette` option.
-The default color palette has been changed to be more accessible for color blind people.
 
 Closes #6738
--- a/pylint/pyreverse/main.py
+++ b/pylint/pyreverse/main.py
@@ -51,18 +51,6 @@ DEFAULT_COLOR_PALETTE = (
     "gold",
     "hotpink",
     "mediumspringgreen",
-    # TODO for 3.0: replace the colors above with the ones below
-    # Those are more colorblind friendly
-    # "dodgerblue",
-    # "orange",
-    # "teal",
-    # "sienna",
-    # "orchid",
-    # "lightsalmon",
-    # "plum",
-    # "gray",
-    # "yellow",
-    # "steelblue",
 )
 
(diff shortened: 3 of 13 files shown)


nickdrozd (dismissed):
Defaults looks great. Tested input color flag; works. :art:

Pierre-Sassoulas (commented):
Maybe we should keep what we had because it's a kind of breaking change. I'm so not the person to tell you if the default color palette looks good or not... But now that it can be user defined then it's fine, user can change it *to taste* :)

nickdrozd:
> Maybe we should keep what we had because it's a kind of breaking change. I'm so not the person to tell you if the default color palette looks good or not... But now that it can be user defined then it's fine, user can change it _to taste_ :)

By that same logic, anyone who just has to have _alice blue_ and _antique white_ is also free to specify them! But somehow I have the feeling that few will actually choose those colors :laughing:

Pierre-Sassoulas:
You'd be surprised, people get accustomed to thing and if it changes there's always someone complaining.  https://imgs.xkcd.com/comics/workflow.png

DudeNr33:
I don't have a strong opinion on changing the default colors.
We can leave them as they are, and change them in 3.0.

Pierre-Sassoulas (dismissed):
I'd prefer to minimize churn but I'm also ok if @nickdrozd feels strongly about changing the default :)

DudeNr33:
I will leave it as it is currently (i.e. stick with the previous default colors and leave the TODO to switch for 3.0) and will take care of the spelling errors this evening.
Trying to install `enchant` yesterday resulted in a downwards spiral that made me update macOS. Sometimes you gotta love being a developer... 😅

Pierre-Sassoulas:
> Trying to install enchant yesterday resulted in a downwards spiral that made me update macOS. Sometimes you gotta love being a developer... 

https://www.youtube.com/watch?v=AbSehcT19u0

nickdrozd:
Sure, the default colors could be changed later, as part of the wider push for sane defaults. But they really should be changed. It's not just my opinion that default colors are bad; the [SO post](https://stackoverflow.com/questions/72294217/pyreverse-uml-diagram-with-customized-color) that prompted the color issue specifically mentions wanting to change the "faint blue color" to a darker one.

And it's not just about opinions. The new proposed colors are objectively better in terms of accessibility than the current defaults. Actually, the ugliness and the inaccessibility have the same root cause, which is that *alice blue* and *antique white* are both difficult to distinguish from each other and difficult to distinguish from a white background.

As an example, here is a package diagram generated by Pyreverse with the current colors:

packages

Pierre-Sassoulas (approved):

PR #8234, opened 2023-02-08
Fix `nested-min-max` output msg

### LINKED ISSUE #8168: False positive `nested-min-max`
### Bug description

The `nested-min-max` suggests `max(3, max(1, 2))` to be `max(3, 1, 2)`. However, if the `min` / `max` function takes an iterable `max(iterable)` instead of multiple inputs `max(a, b, c)`. The nested min/max cannot be rewritten as a single statement `max(x, max(iterable)) -> max(x, iterable)`.

```python
"""test.py"""

lst = [1, 2]
max(3, max(lst))
```

If rewrite it to `max(3, lst)`, it will raise:

```pytb
TypeError: '>' not supported between instances of 'list' and 'int'
```

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:4:0: W3301: Do not use nested call of 'max'; it's possible to do 'max(3, lst)' instead (nested-min-max)

```

### Expected behavior

No warnings.

### Pylint version

```shell
pylint 2.16.0
astroid 2.14.1
Python 3.8.13 (default, Oct 21 2022, 23:50:54)
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

clavedeluna:
## Type of Changes

## Description
Fix ``nested-min-max`` suggestion message to indicate it's possible to splat iterable objects.

Closes #8168


tests/test_self.py
```diff
@@ -1302,6 +1302,21 @@ def test_no_name_in_module(self) -> None:
             [module, "-E"], expected_output="", unexpected_output=unexpected
         )
 
+    def test_nested_min_max_splats(self) -> None:
```
Pierre-Sassoulas, line 1305: I think we can remove this if we also have functional tests(I like functional tests a lot)

pylint/checkers/utils.py
```diff
@@ -238,6 +238,13 @@
     {"_sitebuiltins.Quitter", "sys.exit", "posix._exit", "nt._exit"}
 )
 
+DICT_TYPES = (
```
Pierre-Sassoulas, line 241: Let's keep this in the variable checkers, we don't want to have checkers be coupled because of this constant and the module name 'utils' is horrible.

Pierre-Sassoulas (commented):
Looks amazing already !

pylint/checkers/utils.py
clavedeluna: fair point! the repo uses "utils" a ton. So just have both "DICT_TYPES" in each module regardless of duplication?

tests/test_self.py
Pierre-Sassoulas: If I understand correctly there is an issue with `*` in functional tests ?

clavedeluna: No no, no issue. You can see the functional test .txt file looks great. 

What I mean is that as far as I understand, the functional test will ensure that the message 
nested-min-max appears in a specific line, indicated by adding `# [nested-min-max]` to that line

but what I want to test is not just that the msg is emitted, but that the msg suggestion specifically says "it's possible to do 'max(3, *nums, *lst2)' i". I can only assert this in the unit test, not the functional test.

tests/functional/n/nested_min_max.txt
```diff
@@ -6,3 +6,9 @@ nested-min-max:8:0:8:25::Do not use nested call of 'min'; it's possible to do 'm
 nested-min-max:11:0:11:25::Do not use nested call of 'min'; it's possible to do 'min(1, 2, 3, 4)' instead:INFERENCE
 nested-min-max:12:0:12:40::Do not use nested call of 'min'; it's possible to do 'min(len([]), len([1]), len([1, 2]))' instead:INFERENCE
 nested-min-max:17:0:17:27::Do not use nested call of 'orig_min'; it's possible to do 'orig_min(1, 2, 3)' instead:INFERENCE
+nested-min-max:25:0:25:16::Do not use nested call of 'max'; it's possible to do 'max(3, *lst)' instead:INFERENCE
+nested-min-max:29:0:29:17::Do not use nested call of 'max'; it's possible to do 'max(3, *nums)' instead:INFERENCE
+nested-min-max:33:0:33:17::Do not use nested call of 'max'; it's possible to do 'max(3, *nums)' instead:INFERENCE
+nested-min-max:37:0:37:17::Do not use nested call of 'max'; it's possible to do 'max(3, *nums)' instead:INFERENCE
+nested-min-max:40:0:40:26::Do not use nested call of 'max'; it's possible to do 'max(3, *nums.values())' instead:INFERENCE
+nested-min-max:44:0:44:28::Do not use nested call of 'max'; it's possible to do 'max(3, *nums, *lst2)' instead:INFERENCE
```
Pierre-Sassoulas, line 14: The output of the functional tests is checked here @clavedeluna (And it can be upgraded automatically if the message change)

tests/test_self.py
Pierre-Sassoulas: https://github.com/PyCQA/pylint/pull/8234/files#r1100266806 (Modify the expected output in this file and the tests should fail provided you have python 3.8 ore more recent)

tests/functional/n/nested_min_max.txt
clavedeluna: ah hm good point... good point!

Pierre-Sassoulas (approved):
👍

PR #8247, opened 2023-02-08
Add example for deprecated-module

ollie-iterators:
## Type of Changes

## Description

Add example for deprecated-module

Refs #5953 

Closes #XXXX


DanielNoord (dismissed):
Please use one of the modules that is deprecated by default, I think that makes more sense in this case. Perhaps `distutils.copy` > `shutil.copy`? That's a nice transition to show in the `good` and `bad examples.

ollie-iterators:
Ok, I'll work on changing the module that is deprecated later.

doc/data/messages/d/deprecated-module/pylintrc
```diff
@@ -0,0 +1,2 @@
+[FORMAT]
+deprecated-module = copy
```
Pierre-Sassoulas, line 2: This need to have a pyversion set or it will start to break when we upgrade the interpreter for the documentation tests. See https://github.com/PyCQA/pylint/blob/main/doc/data/messages/c/continue-in-finally/pylintrc

doc/data/messages/d/deprecated-module/bad.py
```diff
@@ -0,0 +1,16 @@
+from copy import deepcopy  # [deprecated-module]
```
Pierre-Sassoulas, line 1: ``deepcopy`` is not the best example as there's no drop in replacement that will make the code in good an equivalent. Better use ``os.path`` => ``Path`` or ``distutils`` => ``setuptools`` or ``optparse`` => ``argparse``.

Pierre-Sassoulas (changes requested):
Thank you for opening the merge request :)

doc/data/messages/d/deprecated-module/bad.py
ollie-iterators: I understand, I just chose that tool because I noticed the deprecated-module setting in the pylintrc file which made me realize how to cause the error to happen.

ollie-iterators: So, to confirm before committing, I should change bad.py to "import distutils # [deprecated-module]", and good.py to "import setuptools" and pylintrc to "[main]
py-version=3.7" to make the example better?

ollie-iterators: Ok, I am using distutils and setuptools

doc/data/messages/d/deprecated-module/pylintrc
ollie-iterators: I added a pyversion to the code.

doc/data/messages/d/deprecated-module/pylintrc
```diff
@@ -0,0 +1,3 @@
+[main]
+py-version=3.7
+deprecated-modules=regsub,TERMIOS,Bastion,rexec
```
Pierre-Sassoulas, line 3:
```suggestion
deprecated-modules=whatever_you_want
```

doc/data/messages/d/deprecated-module/bad.py
```diff
@@ -0,0 +1 @@
+import distutils  # [deprecated-module]
```
Pierre-Sassoulas, line 1:
```suggestion
import distutils  # [deprecated-module]
import whatever_you_want  # [deprecated-module]
```

doc/data/messages/d/deprecated-module/good.py
```diff
@@ -1 +1 @@
-# This is a placeholder for correct code for this message.
+import setuptools
```
Pierre-Sassoulas, line 1:
```suggestion
import setuptools
import whatever_replacement_you_want
```

Pierre-Sassoulas (commented):
LGTM, but there might be an opportunity to document the other use case.

doc/data/messages/d/deprecated-module/bad.py
ollie-iterators: Added changes

doc/data/messages/d/deprecated-module/good.py
ollie-iterators: Added changes

Pierre-Sassoulas (approved):

PR #8253, opened 2023-02-10
Fix FP for used-before-assignment with assignment expressions in containers

### LINKED ISSUE #8252: `used-before-assignment` false positive for named expressions in a not-first position in container
This raises `used-before-assignment`:
```python
def expression_in_ternary_operator_inside_container_later_position():
    """Named expression follows unrelated item in container"""
    return [23, val3 if (val3 := 'something') else 'anything']
```

***

The check for `BaseContainer` should also check all elements. I can handle that in another PR.

_Originally posted by @jacobtylerwalls in https://github.com/PyCQA/pylint/pull/8176#pullrequestreview-1292221144_

jacobtylerwalls:
## Type of Changes

## Description
Closes #8252

### DIFF
--- a/tests/functional/u/undefined/undefined_variable_py38.py
+++ b/tests/functional/u/undefined/undefined_variable_py38.py
@@ -174,7 +174,11 @@ def expression_in_ternary_operator_inside_container_tuple():
 
 
 def expression_in_ternary_operator_inside_container_later_position():
-    """Named expression follows unrelated item in container"""
+    """
+    Named expression follows unrelated item in container.
+
+    If 23 is replaced with `val3`, there is currently a false negative,
+    but the false positive here is more important and likely to occur."""
     return [23, val3 if (val3 := 'something') else 'anything']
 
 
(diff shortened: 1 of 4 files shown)


tests/functional/u/undefined/undefined_variable_py38.py
```diff
@@ -173,9 +173,9 @@ def expression_in_ternary_operator_inside_container_tuple():
     return [(val3, val3) if (val3 := 'something') else 'anything']
 
 
-def expression_in_ternary_operator_inside_container_wrong_position():
-    """2-element list where named expression comes too late"""
-    return [val3, val3 if (val3 := 'something') else 'anything']  # [used-before-assignment]
+def expression_in_ternary_operator_inside_container_later_position():
+    """Named expression follows unrelated item in container"""
+    return [23, val3 if (val3 := 'something') else 'anything']
```
jacobtylerwalls, line 178: This PR trades fixing this false positive for not addressing the false negative previously tested here. I think this is both more important and more common.

Pierre-Sassoulas: We should put that comment directly in the code imo

jacobtylerwalls: 3aa45afd2e46455269d4af77606c088b8298a15f

Pierre-Sassoulas (approved):

PR #8255, opened 2023-02-10
doc unexpected-line-ending-format

hofrob:
## Type of Changes

## Description

Add doc for unexpected-line-ending-format / C0328

Refs #5953

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@ tests/**/functional/** -text
 tests/input/** -text
 tests/**/data/** -text
 tests/regrtest_data/** -text
+doc/data/messages/u/unexpected-line-ending-format/bad.py -text
--- a/doc/data/messages/u/unexpected-line-ending-format/bad.py
+++ b/doc/data/messages/u/unexpected-line-ending-format/bad.py
@@ -1,0 +1,2 @@
+print("Sometimes I drink soda.")  # CRLF # [unexpected-line-ending-format]
+print("Sometimes I drink sulfuric acid.")  # CRLF # [unexpected-line-ending-format]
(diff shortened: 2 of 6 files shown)


hofrob:
It's not that easy to commit an actual file with CRLF if all your tools are trying to smart and help you "do the right thing" :grimacing:

Pierre-Sassoulas:
Yes you'd need to modify our .gitattributes for that. But it's impossible to see that you did the right thing while reading the doc though :) Maybe something that don't need a bad/good that we could handle in #7897 with a nice story about the meaning and history of CRLF and type printer.

hofrob:
Right now it's correct, but you're right, this is not visible in the docs. I added inline comments like in the `missing-final-newline` docs.

I'm thinking about adding an explanation in a `details.rst`.

Pierre-Sassoulas:
Might want to link this https://stackoverflow.com/a/6521730/2519059 and this https://stackoverflow.com/a/10855862/2519059 in related too.

hofrob:
Ok, those links are awesome and have all the info needed (I think). I added those and a link to the awesome gitattributes template repo to `related.rst` now.

hofrob:
I don't get it. It worked yesterday. Now it doesn't anymore. The `bad.py` file has `CRLF` when I check it out locally. But something along the way replaces them with `LF` again.

:shrug:

.gitattributes
```diff
@@ -3,3 +3,4 @@ tests/**/functional/** -text
 tests/input/** -text
 tests/**/data/** -text
 tests/regrtest_data/** -text
+doc/data/messages/u/unexpected-line-ending-format/bad.py text eol=crlf
```
Pierre-Sassoulas, line 6:
```suggestion
doc/data/messages/u/unexpected-line-ending-format/bad.py -text
```
Isn't ``eol=crlf`` is normalizing ?

hofrob: Yes, that was it! Thanks for your help!

Pierre-Sassoulas (approved):

PR #8267, opened 2023-02-12
Add documentation for invalid-character-carriage-return

ollie-iterators:
## Type of Changes

## Description

Add documentation for invalid-character-carriage-return. NOTE: I have no idea why syntax-error is not being disabled.

Refs #5953 

Closes #XXXX


doc/data/messages/i/invalid-character-carriage-return/bad.py
```diff
@@ -0,0 +1,7 @@
+# pylint: disable=trailing-whitespace, missing-module-docstring, trailing-newlines, missing-final-newline, syntax-error
```
Pierre-Sassoulas, line 1: You don't need that, all message except the one we document and fatal one are disabled :)

ollie-iterators: Changed it

Pierre-Sassoulas:
@ollie-iterators there's a genuine fail here, it's not only ``self-import``

doc/data/messages/i/invalid-character-carriage-return/bad.py
```diff
@@ -0,0 +1,7 @@
+# pylint: disable = syntax-error
```
mbyrnepr2, line 1:
```suggestion
```

mbyrnepr2:
There's an [example](https://github.com/PyCQA/pylint/blob/aaa3770d325223ee5c13a637f09ff16e5bbc5851/tests/functional/b/bad_char/bad_char_carriage_return.py) in the functional tests which could be re-used.

ollie-iterators:
> There's an [example](https://github.com/PyCQA/pylint/blob/aaa3770d325223ee5c13a637f09ff16e5bbc5851/tests/functional/b/bad_char/bad_char_carriage_return.py) in the functional tests which could be re-used.
> 
> I think the `disable`s can be omitted from the good/bad examples because these other messages are ignored in the examples.

I have tried to copy the example, but it does not seem to get copied.

Issue #8109, opened 2023-01-25
Crash `AttributeError: 'NoneType' object has no attribute 'value'`


siner308:
### Bug description

pylint failed with Fatal error but code is working

```python
# number.py

from dataclasses import dataclass


@dataclass
class Number:
    amount: int | float
    round: int = 2

    def __str__(self):
        number_format = "{:,.%sf}" % self.round
        return number_format.format(self.amount).rstrip("0").rstrip(".")
```

### Command used

```shell
pylint number.py
```

### Pylint output

```shell
pylint crashed with a ``AstroidError`` and with the following stacktrace:

Traceback (most recent call last):
  File "/Users/siner308/Library/Caches/pypoetry/virtualenvs/myproject-6nZamvkJ-py3.11/lib/python3.11/site-packages/astroid/inference_tip.py", line 38, in _inference_tip_cached
    result = _cache[func, node]
             ~~~~~~^^^^^^^^^^^^
KeyError: (<function _infer_str_format_call at 0x1056fb6a0>, <Call l.11 at 0x107c0e250>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/siner308/Library/Caches/pypoetry/virtualenvs/myproject-6nZamvkJ-py3.11/lib/python3.11/site-packages/pylint/lint/pylinter.py", line 790, in _lint_file
    check_astroid_module(module)
  File "/Users/siner308/Library/Caches/pypoetry/virtualenvs/myproject-6nZamvkJ-py3.11/lib/python3.11/site-packages/pylint/lint/pylinter.py", line 1060, in check_astroid_module
    retval = self._check_astroid_module(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/siner308/Library/Caches/pypoetry/virtualenvs/myproject-6nZamvkJ-py3.11/lib/python3.11/site-packages/pylint/lint/pylinter.py", line 1110, in _check_astroid_module
    walker.walk(node)
  File "/Users/siner308/Library/Caches/pypoetry/virtualenvs/myproject-6nZamvkJ-py3.11/lib/python3.11/site-packages/pylint/utils/ast_walker.py", line 93, in walk
    self.walk(child)
  File "/Users/siner308/Library/Caches/pypoetry/virtualenvs/myproject-6nZamvkJ-py3.11/lib/python3.11/site-packages/pylint/utils/ast_walker.py", line 93, in walk
    self.walk(child)
[...]

PR #8268, opened 2023-02-12
Upgrade astroid to 2.14.2

Pierre-Sassoulas:
Also add regression test for #8109

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies    = [
     # Also upgrade requirements_test_min.txt.
     # Pinned to dev of second minor update to allow editable installs and fix primer issues,
     # see https://github.com/PyCQA/astroid/issues/1341
-    "astroid>=2.14.1,<=2.16.0-dev0",
+    "astroid>=2.14.2,<=2.16.0-dev0",
     "isort>=4.2.5,<6",
     "mccabe>=0.6,<0.8",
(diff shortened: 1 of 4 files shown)


DanielNoord (approved):
Not sure we need the regression test as it completely duplicates the test in `astroid`, but I'm fine either way

Pierre-Sassoulas:
The merge queue seems to not permit to choose between squash/rebase/merge, this is going to be a problem for messy MR we're going to be forced to ask for cleanup. Did you activate it ?

DanielNoord:
@Pierre-Sassoulas That's annoying. Disabling it again.

Pierre-Sassoulas:
It seem it's possible to bypass with admin rights, the intention is good though but we're not forcing to be on the latest main and to run the test on the latest main or we would not be merging anything. So it's also going to increase massively the runner used if we put it back because right now we don't. Like for this MR the test were running again when they weren't before. Or maybe it's the equivalent of running the test on main and then merging but not running the test on main ? Right now we're dropping the current the primer on main when there is a new commit merged so it's also going to take longer with merge queue (the primer is going to run fully for each commit on main?).

Pierre-Sassoulas:
OK, probably just a case of github needing to permit rebase/squash/merge in their interface too, then it's a very good thing. It's still in beta it seems, I don't know if we can provide feedback somewhere.

PR #8286, opened 2023-02-13
[doc] Do not use real code for 'self-import'

### LINKED ISSUE #8273: ``test_code_examples[import-self-bad]`` is flaky and the example "work" but is not very good
### Current problem

The test is very flaky, I noticed it in #8264, I don't know what could make this particular test flaky though. And I can't reproduce locally.

### Desired solution

Maybe use a new template from #7897, we exposes our internals to users because we have to import bad.py in the example and it work (tests were green) but it's not super clear if you don't know how it's implemented.

### Additional context

_No response_

Pierre-Sassoulas:
## Type of Changes

## Description

Closes #8273

### DIFF
--- a/doc/data/messages/i/import-self/bad.py
+++ b/doc/data/messages/i/import-self/bad.py
@@ -1,5 +1,0 @@
-from bad import a_function  # [import-self]
-
-
-def a_function():
-    pass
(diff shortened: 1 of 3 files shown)


mbyrnepr2 (approved):

Pierre-Sassoulas:
Thank you @mbyrnepr2 :)

PR #8287, opened 2023-02-13
Add support for more complicated documentation examples

### LINKED ISSUE #7143: Add a configuration file in documentation example for message
### Current problem

Right now we do not show the configuration used for example. We could and it would:

- show readers what need to be done to activate the check if it's an extension, or a standard but optional check (soon ™️)
- permits to create simpler examples by lowering the values to something sane (i.e. max line length = 15 instead of having 120+ character in the example, max-return-values = 2 instead of 11...)
- demonstrate some options that can affect the message (!)

### Desired solution

Show the configuration used in bad/good and use it as an example of configuration related to a message. There's some issue with this:

- How to identify related options?

It's not easy to determine that without reading the code, we struggle as much as user here.

- How to show the difference between the default value and the value being used in a test?

If the default value is 120 we need a way to show what it is if we change it to 15.

- How to handle duplication of the load-plugins warning

We already have a message for extensions right now.

### Additional context

Initially discussed in https://github.com/PyCQA/pylint/pull/7142#pullrequestreview-1031209077

DanielNoord:
## Type of Changes

## Description

Refs https://github.com/PyCQA/pylint/issues/7897
Closes #7143

### DIFF
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import os
 from collections import defaultdict
+from enum import Enum
 from inspect import getmodule
 from itertools import chain, groupby
 from pathlib import Path
@@ -39,14 +40,16 @@ class MessageData(NamedTuple):
     id: str
     name: str
     definition: MessageDefinition
-    good_code: str
-    bad_code: str
-    details: str
-    related_links: str
+    example_code: str
     checker_module_name: str
     checker_module_path: str
     shared: bool = False
     default_enabled: bool = True
+
+
+class ExampleType(str, Enum):
+    GOOD = "good"
+    BAD = "bad"
 
 
 MessagesDict = Dict[str, List[MessageData]]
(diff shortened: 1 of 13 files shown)


doc/data/messages/c/cyclic-import/bad/bad.py
```diff
@@ -0,0 +1,9 @@
+from .bad2 import count_to_two  # [cyclic-import]
```
Pierre-Sassoulas, line 1: ~~Can we take into account the fact that if we want to do multiple examples properly displayed we're going to want to have bad.py, bad2.py, bad3.py (...) too ? Maybe the display can change if there's no good2.py, (...)  matching the bad2.py (...) ?~~ Edit: Did not notice it's in a directory, I think it's compatible.

pylint/checkers/imports.py
```diff
@@ -555,13 +555,18 @@ def visit_importfrom(self, node: nodes.ImportFrom) -> None:
...
+        if imported_module_name is None:
             return
+
         for name, _ in node.names:
             if name != "*":
-                self._add_imported_module(node, f"{imported_module.name}.{name}")
+                self._add_imported_module(node, f"{imported_module_name}.{name}")
             else:
-                self._add_imported_module(node, imported_module.name)
+                self._add_imported_module(node, imported_module_name)
```
Pierre-Sassoulas, line 569: Seems like something that could be it's own PR.

Pierre-Sassoulas (commented):
Looks pretty good, thank you for tackling this !

doc/test_messages_documentation.py
```diff
@@ -52,10 +52,20 @@ def get_functional_test_files_from_directory(input_dir: Path) -> list[tuple[str,
                 suite.append(
                     (message_dir.stem, message_dir / "good.py"),
                 )
+
+            # Bad files can either consist of a single file or a directory.
             if (message_dir / "bad.py").exists():
                 suite.append(
                     (message_dir.stem, message_dir / "bad.py"),
                 )
+            elif (message_dir / "bad").is_dir():
```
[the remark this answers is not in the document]
DanielNoord: I think in that case it makes more sense to just use one file

pylint/checkers/imports.py
DanielNoord: Yeah and it is breaking tests... Without it this test is very flaky so I'm not sure what to do

DanielNoord:
@Pierre-Sassoulas I kept in the `cyclic-import` example to show this works. I'll remove it afterwards as the behaviour is flaky and it seems to be because of `astroid`.

I think this PR also closes https://github.com/PyCQA/pylint/issues/7143 right?

doc/test_messages_documentation.py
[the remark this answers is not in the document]
ollie-iterators: You could also do bad.py, neutral.py, lawful.py and good.py

Pierre-Sassoulas:
> I think this PR also closes https://github.com/PyCQA/pylint/issues/7143 right?

Yes, amazing !

doc/exts/pylint_messages.py
```diff
@@ -62,28 +59,67 @@ def _register_all_checkers_and_extensions(linter: PyLinter) -> None:
...
-        title="Problematic code", text=_get_python_code_as_rst(bad_py_path)
-    )
+    # Handle multi-file bad code templates
+    if bad_py_path.exists():
+        bad_code = _get_titled_rst(
+            title="Problematic code", text=_get_python_code_as_rst(bad_py_path)
+        )
+    elif (data_path / "bad").exists():
+        bad_one = f"""\
+File one:
```
Pierre-Sassoulas, line 85: Instead of ``File one`` / ``File two``, we should have the actual name of the file (``bad.py::``. This would also simplify the handling code that can become generic and handle as many file as we want. We can also name it however we want (it's the directory prefix that indicate badness / goodness).

DanielNoord: I like this suggestion! Will (hopefully) work on this tomorrow!

doc/exts/pylint_messages.py
```diff
@@ -62,28 +59,62 @@ def _register_all_checkers_and_extensions(linter: PyLinter) -> None:
...
+``{file.name}``:
+
+.. literalinclude:: /{file.relative_to(Path.cwd())}
+    :language: python
+
+"""
+                )
+        bad_code = _get_titled_rst(title="Problematic code", text="".join(bad_files))
+    else:
+        bad_code = ""
```
Pierre-Sassoulas, line 98: We should create a generic function for this as we also need to handle ``good/`` directories with multiple file imo. I can do it, it's a lot easier with pycharm than vscode.

doc/exts/pylint_messages.py
```diff
@@ -62,28 +59,62 @@ def _register_all_checkers_and_extensions(linter: PyLinter) -> None:
...
-        title="Problematic code", text=_get_python_code_as_rst(bad_py_path)
-    )
+    # Handle multi-file bad code templates
+    if bad_py_path.exists():
+        bad_code = _get_titled_rst(
+            title="Problematic code", text=_get_python_code_as_rst(bad_py_path)
+        )
+    elif (data_path / "bad").exists():
+        bad_files: list[str] = []
+        for file in (data_path / "bad").iterdir():
```
Pierre-Sassoulas, line 85: We need to sort here, so bad.py is before bad2.py
```suggestion
        for file in sorted((data_path / "bad").iterdir()):
```

Pierre-Sassoulas (changes requested):
Let me know if you agree with the good/ multiple fle and I'll do it :)

doc/exts/pylint_messages.py
DanielNoord: Yeah I didn't want to add code that we don't actively use right now, but go ahead and add it! Whatever makes it easier to complete this process the quickest!

doc/exts/pylint_messages.py
Pierre-Sassoulas: Turn out it's not that easy to sort Path 😄 took me embarrassingly long to understand

DanielNoord:
I have commits almost ready that clean up this PR. Hopefully can finish this tomorrow.

DanielNoord:
Not clearing that reporter after every file shows some underlying issues that we never seen. Nice to fix these as well I think :smile:

pylint/testutils/reporter_for_tests.py
```diff
@@ -72,9 +72,6 @@ def _display(self, layout: Section) -> None:
 
 
 class FunctionalTestReporter(BaseReporter):
-    def on_set_current_module(self, module: str, filepath: str | None) -> None:
-        self.messages = []
-
```
Pierre-Sassoulas, line 77: Nice !

Pierre-Sassoulas (commented):
LGTM but it's impossible to check in read the doc now that the cyclic-import is removed.

DanielNoord:
> LGTM but it's impossible to check in read the doc now that the cyclic-import is removed.

Yeah, like I said; the `cyclic-import` message is 100% flakey 😅 

If you know of another message where you want multiple files and that isn't flakey I can add it as well. Otherwise I think we should just merge this. I did test this locally with multiple files and checked that the tests still worked (which is how I found the issues with the reporter).

Pierre-Sassoulas:
Maybe too many lines ? It's a little insulting to readers though.  Otherwise, duplicated code ? I can add this and test locally myself :)

DanielNoord:
> Maybe too many lines ? It's a little insulting to readers though. Otherwise, duplicated code ? I can add this and test locally myself :)

Duplicated code would be a good one! Perhaps make a PR that has this branch as target branch?

Pierre-Sassoulas (approved):
I'll let you check the example, and then let's merge. The result is pretty good, excited to make this live ! 😄

DanielNoord:
Both https://pylint--8287.org.readthedocs.build/en/8287/user_guide/messages/warning/while-used.html and https://pylint--8287.org.readthedocs.build/en/8287/user_guide/messages/refactor/duplicate-code.html are definite improvements!

Indeed happy we have this now!

stdedos:
> Both [pylint--8287.org.readthedocs.build/en/8287/user_guide/messages/warning/while-used.html](https://pylint--8287.org.readthedocs.build/en/8287/user_guide/messages/warning/while-used.html) and [pylint--8287.org.readthedocs.build/en/8287/user_guide/messages/refactor/duplicate-code.html](https://pylint--8287.org.readthedocs.build/en/8287/user_guide/messages/refactor/duplicate-code.html) are definite improvements!
> 
> Indeed happy we have this now!

btw - is the `Similar lines in %s files %s` bug or feature?
Since it's "technically code", should the messages have the `backtick` formatting?

DanielNoord:
I think it might make sense indeed!

PR #8292, opened 2023-02-14
Only count obviously non-terminating while-loops as return-ended

### LINKED ISSUE #8280: inconsistent-return-statements not detected for while loop with flag
### Bug description

```python
    def get_the_answer(self):  # [inconsistent-return-statements]
        while self.is_running:
            read_message()
            if comunication_finished():
                return "done"
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
Your code has been rated at 10.00/10 (previous run: 9.97/10, +0.03)
```

### Expected behavior

I expect to get message:
`R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)` because when the `'self.is_running` flag changes to false, I'll receive `None` instead of str which isn't desired behaviour and should be detected by pylint.

### Pylint version

```shell
pylint 2.16.1
astroid 2.14.2
Python 3.10.9 (main, Dec 15 2022, 17:11:09) [Clang 14.0.0 (clang-1400.0.29.202)]
```

### OS / Environment

Macos

### Additional dependencies

_No response_

nickdrozd:
## Type of Changes

## Description

Fixes false negative for `inconsistent-returns` by only counting `while` loops as return-ended when they are obviously non-terminating.

Closes #8280

### DIFF
--- a/doc/whatsnew/fragments/8280.false_negative
+++ b/doc/whatsnew/fragments/8280.false_negative
@@ -1,3 +1,3 @@
 Fix false negative for inconsistent-returns with while-loops.
 
-Closes 8280
+Closes #8280
--- a/doc/whatsnew/fragments/8280.false_negative
+++ b/doc/whatsnew/fragments/8280.false_negative
@@ -1,0 +1,3 @@
+Fix false negative for inconsistent-returns with while-loops.
+
+Closes 8280
--- a/pylint/checkers/base/basic_error_checker.py
+++ b/pylint/checkers/base/basic_error_checker.py
@@ -46,7 +46,7 @@ def _get_break_loop_node(break_node: nodes.Break) -> nodes.For | nodes.While | N
     return parent
 
 
-def _loop_exits_early(loop: nodes.For | nodes.While) -> bool:
+def loop_exits_early(loop: nodes.For | nodes.While) -> bool:
     """Returns true if a loop may end with a break statement.
 
     Args:
@@ -474,7 +474,7 @@ class BasicErrorChecker(_BasicChecker):
 
     def _check_else_on_loop(self, node: nodes.For | nodes.While) -> None:
         """Check that any loop with an else clause has a break statement."""
-        if node.orelse and not _loop_exits_early(node):
+        if node.orelse and not loop_exits_early(node):
             self.add_message(
                 "useless-else-on-loop",
                 node=node,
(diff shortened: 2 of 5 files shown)


Pierre-Sassoulas (approved):
LGTM, nice 👍

doc/whatsnew/fragments/8280.false_negative
```diff
@@ -0,0 +1,3 @@
+Fix false negative for inconsistent-returns with while-loops.
+
+Closes 8280
```
nickdrozd, line 3: I forgot the `#` for the issue number. Is there a way to add that without rerunning all the tests?

Pierre-Sassoulas: No :D

nickdrozd:
I added a few more test cases based on initial primer results. Primer results look good to me now.

pylint/checkers/base/basic_error_checker.py
```diff
@@ -46,7 +46,7 @@ def _get_break_loop_node(break_node: nodes.Break) -> nodes.For | nodes.While | N
     return parent
 
 
-def _loop_exits_early(loop: nodes.For | nodes.While) -> bool:
+def loop_exits_early(loop: nodes.For | nodes.While) -> bool:
```
Pierre-Sassoulas, line 49: Let's keep our private functions private, we don't want to have to deprecate things if people assume we voluntarily made this public. The old functions that are not explicitly private are not public either but at least the current maintaining team did not make them private herself.

Pierre-Sassoulas (changes requested):

Pierre-Sassoulas (dismissed):

nickdrozd:
@Pierre-Sassoulas Anything further on this one?

Pierre-Sassoulas:
Nothing from me, but I'd like another review before merging if possible.

tests/functional/i/inconsistent/inconsistent_returns.py
```diff
@@ -1,5 +1,5 @@
 #pylint: disable=missing-docstring, no-else-return, no-else-break, invalid-name, unused-variable, superfluous-parens, try-except-raise
-#pylint: disable=disallowed-name
+#pylint: disable=disallowed-name,too-few-public-methods,no-member,undefined-variable,useless-else-on-loop
```
DanielNoord, line 2:
```suggestion
#pylint: disable=disallowed-name,too-few-public-methods,useless-else-on-loop
```
This is a bit of a nit, but I'd prefer if the tests were just readable and functioning code where possible. I think these can be refactored to not give a `SyntaxError`

DanielNoord (commented):
One nit, rest LGTM! Thanks!

Pierre-Sassoulas (approved):

DanielNoord (approved):

Issue #8224, opened 2023-02-07
consider-using-f-string is triggered in a functional context


sam-s:
### Bug description

Pylint suggest f-string in a functional context where it cannot be used:
```
map("{:,d}".format, [1,10,100,1000,10000])
```
The alternative:
```
(f"{x:,d}" for x in [...]) 
```
while, admittedly, shorter by a few characters, is not more clear.

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
[C0209(consider-using-f-string), ] Formatting a regular string which could be a f-string
```

### Expected behavior

no diagnostics

### Pylint version

```shell
pylint 2.15.10
astroid 2.13.5
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
```

### OS / Environment

ubuntu

Pierre-Sassoulas:
This check is not about readability or terseness, but after a quick check, I'm surprised that the map version using format is actually faster:
```
python -m timeit '(f"{x:,d}" for x in [1,10,100,1000,10000])'
2000000 loops, best of 5: 167 nsec per loop
python -m timeit '(map("{:,d}".format, [1,10,100,1000,10000]))'
2000000 loops, best of 5: 119 nsec per loop
```

clavedeluna:
I also consistently get that using map is faster than f-string equivalent. I'm not sure there's much more investigation we can do. Sounds like deciding if we want pylint to be consistent and always suggest using f-string or to prioritize map's performance and tell pylint to not emit `consider-using-fstring` in cases when `map` is used

sam-s:
> `(f"{x:,d}" for x in [1,10,100,1000,10000])` is insignificantly shorter than `(map("{:,d}".format, [1,10,100,1000,10000]))` and subjectively easier to read.

How would you rewrite
```
pandas_series.map("{:.2%}".format)
```

sam-s:
what about [`display.float_format`](https://pandas.pydata.org/docs/user_guide/options.html)?
```
with pd.option_context("display.float_format", '{:.4g}'.format):
    ...
```

PR #8299, opened 2023-02-15
Don't emit f-string warning for tuple variables

### LINKED ISSUE #8296: consider-using-f-string is triggered when the formatting arguments are implicit
### Bug description

the following code should produce no diagnostics:
```python
"https://github.com/PyCQA/pylint/issues/8296"

date = (2020,1,2,3,4,5)
print("%4d-%02d-%02d %02d:%02d:%02d" % date)

def test_fn(pa_l):
    "my doc"
    return ["Alpha: {alpha:.0%}, Power: {power:.0%}, Nobs={nobs:.1f}".format(**pa)
            for pa in pa_l]
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:4: [C0209(consider-using-f-string), ] Formatting a regular string which could be a f-string
a.py:8: [C0209(consider-using-f-string), test_fn] Formatting a regular string which could be a f-string
```

### Expected behavior

no diagnostics: it is impossible to replace these string formatting ops with f-strings.

### Pylint version

```shell
pylint 2.15.10
astroid 2.13.5
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
```

### OS / Environment

ubuntu

### Additional dependencies

_No response_

nickdrozd:
## Type of Changes

## Description

Don't emit f-string warning for tuple variables.

Closes #8296


pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -427,6 +424,11 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
             elif isinstance(inferred_right, nodes.List):
                 if len(inferred_right.elts) > 1:
                     return
+            # If the inferred value is a tuple and the value comes
```
nickdrozd, line 427: This part is the actual logic change. Everything is just using variables so we don't have to call `\.parent` over and over and have ugly line breaks.

doc/whatsnew/fragments/8292.false_positive
```diff
@@ -0,0 +1,3 @@
+Don't emit `consider-using-f-string` for `%` with tuple variables
```
mbyrnepr2, line 1:
```suggestion
Don't emit ``consider-using-f-string`` for ``%`` with tuple variables.
```

Double: code. Single: italic.

doc/whatsnew/fragments/8292.false_positive
```diff
@@ -0,0 +1,3 @@
...
+
+Fixes #8296
```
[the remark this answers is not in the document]
mbyrnepr2: The regex allows “Fixes part of” but I think the intention is “Closes”.

Pierre-Sassoulas:
It would be nice to also handle #8224 at the same time.

nickdrozd:
@Pierre-Sassoulas Whether these suggestions are appropriate or not is really a matter of opinion. In my opinion, the suggestion in that ticket is good, and the flagged code is strange. So I don't think I will make any attempts to "fix" it. But it someone else wanted to, I wouldn't object. Either way, I added the examples as test cases.

nickdrozd:
The primer results all look like cases where it would be annoying to get that warning, so the change seems good.

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -357,37 +357,34 @@ def visit_const(self, node: nodes.Const) -> None:
             ):
                 self._detect_replacable_format_call(node)
 
+    # pylint: disable = too-many-branches
```
Pierre-Sassoulas, line 360: We need to refactor first I think. This function became too big to understand easily. (It's also better to do the ``parent = node.parent`` refactor as a standalone so we can revert and understand the change easily).

pylint/checkers/refactoring/recommendation_checker.py
```diff
@@ -427,6 +424,11 @@ def _detect_replacable_format_call(self, node: nodes.Const) -> None:
...
                 if len(inferred_right.elts) > 1:
                     return
+            # If the inferred value is a tuple and the value comes
+            # from a variable, don't emit the warning
+            elif isinstance(inferred_right, nodes.Tuple):
+                if isinstance(parent.right, nodes.Name):
+                    return
 
             # If all tests pass, then raise message
             self.add_message(
```
Pierre-Sassoulas, line 434: We can add the confidence while we're at it in the previous refactor.

Pierre-Sassoulas (commented):
I think we should add an option to disable this (like ``--fstrings-no-exception`` with a default to False ?). It's a convention message after all.

nickdrozd:
Upon reflection, I don't care enough about this issue to work on it any further. Anyone who wants to take it over, feel free.

PR #8324, opened 2023-02-21
Work on documentation for non-ascii-file-name

ollie-iterators:
## Type of Changes

## Description

Working on documentation for non-ascii-file-name

This is because while the issue (https://bugs.python.org/issue20485) that caused this issue has been fixed since Python 3.5.   It is still not recommended to use non-ascii characters in a filename. 

This is the PR that fixed the issue: https://github.com/python/cpython/issues/64684#issuecomment-1093644763.

Refs #7897


Pierre-Sassoulas (commented):
This need a changelog message, and a cleanup of the existing code in the checkers, this is not going to be easy, a maintainer or someone experienced should probably do it. Also the current MR need to justify the removal if someone read it later.

ollie-iterators:
Ok, I updated the MR

Pierre-Sassoulas (changes requested):
Thanks for adding the description and the changelog :) Next step would be to remove all the code related to this (functional tests, all the part of the pylint code specific to this check), do you want to do it or should we take over ?

ollie-iterators:
I'll try to do it

ollie-iterators:
> Thanks for adding the description and the changelog :) Next step would be to remove all the code related to this (functional tests, all the part of the pylint code specific to this check), do you want to do it or should we take over ?

Thanks, but I think I'll try to do it

ollie-iterators:
Ok, I removed non-ascii-file-name from all the files except for full.rst, summary.rst and the files that are necessary to show that it was removed.

pylint/checkers/non_ascii_names.py
```diff
@@ -41,26 +41,6 @@ class NonAsciiNameChecker(base_checker.BaseChecker):
...
-                # Some = PyCharm at the time of writing didn't display the non_ascii_name_loł
-                # files and had big troubles with git.
-                # Probably only a bug shows the problem quite good.
-                # That's also why this is a warning and not only a convention!
-                "Some editors don't support non-ASCII file names properly. "
-                "Even though Python supports UTF-8 files since Python 3.5 this isn't "
-                "recommended for interoperability. Further reading:\n"
-                "- https://peps.python.org/pep-0489/#export-hook-name\n"
-                "- https://peps.python.org/pep-0672/#confusing-features\n"
-                "- https://bugs.python.org/issue20485"
```
Pierre-Sassoulas, line 61: Hmm, I did not realize this was well thought out. What happen if we name a file ``non_ascii_name_loł`` in the doc example ?

ollie-iterators: The documentation could be changed so that there is a link to non-ascii characters so that a bad.py or good.py file is not needed.

ollie-iterators: I also reverted the change that removed non-ascii-file-name (for now)

Pierre-Sassoulas:
All right thank you for reverting. Let's simply document the message here.

doc/user_guide/checkers/features.rst
```diff
@@ -734,7 +734,9 @@ Nonascii-Checker checker Messages
   interoperability. Further reading: -
   https://peps.python.org/pep-0489/#export-hook-name -
   https://peps.python.org/pep-0672/#confusing-features -
-  https://bugs.python.org/issue20485
+  https://bugs.python.org/issue20485 -
+  https://github.com/python/cpython/issues/64684#issuecomment-1093644763 -
+  https://terpconnect.umd.edu/~zben/Web/CharSet/htmlchars.html
```
Pierre-Sassoulas, line 739: Everything starting from "further reading" should go in ``related.rst`` imo (and be removed from the checker's code).

doc/data/messages/n/non-ascii-file-name/pylintrc
```diff
@@ -0,0 +1,2 @@
+[main]
+py-version = 3.4
```
Pierre-Sassoulas, line 2: Do we still need that if there's no example ? Also I think the intent of the check is to be activated even in python 3.5+

Pierre-Sassoulas (changes requested):

doc/user_guide/checkers/features.rst
ollie-iterators: Ok, reverting

doc/data/messages/n/non-ascii-file-name/pylintrc
ollie-iterators: The thing that was agreed on was that either it would be removed or add a py-version checker. Since the alert is not going to be removed, I am adding a py-version checker.

Pierre-Sassoulas: Yeah sorry about the confusion, I did not check the reasoning behind the check in detail at first. The comment in the code made me reevaluate what we should do (check even in python 3.5+ for interoperability reasons).

ollie-iterators: Ok, so should I change the py-version to 3.5 or something above 3.5?

Pierre-Sassoulas: No py-version / pylintrc at all if possible.

ollie-iterators: Ok, I've removed the pylintrc

Pierre-Sassoulas:
Thank you for working on this. Sorry I'm going to close because the first link is just something that point to PEP489 that we already linked, and the second link point to a list of non ascii character that is not relevant to the issue. Also we can actually document the problem properly see https://github.com/PyCQA/pylint/pull/8340

PR #8328, opened 2023-02-22
[doc] Add an example for suppressed-message

Pierre-Sassoulas:
## Type of Changes

## Description

The message would not be triggered if implicit-str-concat is not activated too. I initially wanted to activate invalid-name so we follow pep8 in our examples but there's more than 50 violations right now. So 'implicit-str-concat' it is.

Refs #7897 #5953

### DIFF
--- a/doc/data/messages/s/suppressed-message/bad.py
+++ b/doc/data/messages/s/suppressed-message/bad.py
@@ -1,1 +1,9 @@
-test = ["a" "b"]  # pylint: disable=implicit-str-concat # [suppressed-message]
+### This is a contrived example, to show how suppressed-message works.
+### First we enable all messages
+# pylint: enable=all
+
+### Here we disable two messages so we get two warnings
+# pylint: disable=locally-disabled, useless-suppression # [suppressed-message, suppressed-message]
+
+### Here we disable a message, so we get a warning for suppressed-message again.
+"A"  # pylint: disable=pointless-statement # [suppressed-message, suppressed-message]
--- a/doc/data/messages/s/suppressed-message/bad.py
+++ b/doc/data/messages/s/suppressed-message/bad.py
@@ -1,0 +1,1 @@
+test = ["a" "b"]  # pylint: disable=implicit-str-concat # [suppressed-message]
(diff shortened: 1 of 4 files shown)


doc/data/messages/s/suppressed-message/bad.py
```diff
@@ -0,0 +1 @@
+test = ["a" "b"]  # pylint: disable=implicit-str-concat # [suppressed-message]
```
DanielNoord, line 1: I'd prefer to do this with a `pylintrc` that enables the message and then disable it here.

The enable in the running of the tests is just additional "headspace" that gets occupied when dealing with these tests.
The fact that we had so many issues with just disabling the cleaning of messages after each file shows that neither of us (or other maintainers) fully understand all the implications of the functional test runners.

Pierre-Sassoulas: I tried that too (and also enable in file) but I think pylintrc is the baseline that is modified by the command line argument. Not sure if it's possible.

ollie-iterators: I agree. I tried to use disable-next to enable the suppressed message and that did not work for me

Pierre-Sassoulas: Yeah, not sure why, I think the right order for message control should be "configuration < commande line < file directive according to scope":
```bash
$ cat a.py
# pylint: enable=invalid-name
tEsT = "apple"  # pylint: disable=invalid-name
$ pylint a.py --disable=all --enable=invalid-name,suppressed-message,astroid-error,syntax-error
************* Module a
a.py:2:0: I0020: Suppressed 'invalid-name' (from line 2) (suppressed-message)
$ pylint a.py --disable=all --enable=suppressed-message,astroid-error,syntax-error 
```
Clearly fixing the message control is a task that is not going to be tackled like a young ballerina by Sébastien Chabal.

DanielNoord:
@Pierre-Sassoulas I know this is awful, but at least it shows what the behaviour is..

Pierre-Sassoulas:
Hmm maybe it's due to the way "all" behave and take precedence. Definitely something to consider in https://github.com/PyCQA/pylint/issues/3696

Pierre-Sassoulas:
Let's merge this ! You'll have to approve your own change @DanielNoord 😄

DanielNoord:
Something fishy is going on for sure as the second `suppressed-message` on the last line is for `useless-suppression`..

Anyway, I much prefer to keep all of these weird issues contained to one documentation example rather than affecting all other tests for it.

DanielNoord (approved):

Pierre-Sassoulas:
Removing errors and pep8 smells from documentation examples is valuable in itself but this is something from another time. There's 50+ invalid-name violations currently. But we clearly have better things to do.

Issue #5784, opened 2022-02-09
false positive


martimlobao:
### Bug description

Just tried this out with the latest version of the main branch (any version after bfeca4c) and there seems to be a case where #5547 is still raising a false positive unexpected-keyword-arg error.

```python
def example_decorator(*new_func_args, **new_func_kwargs):
    def decorator(func):
        def wrapper(*args, **kwargs):
            print_me = kwargs.pop("print_me", None)
            def inner(*inner_args, **inner_kwargs):
                print(print_me)
                return func(*inner_args, **inner_kwargs)

            return inner(*args, **kwargs)

        return wrapper

    if len(new_func_args) == 1 and len(new_func_kwargs) == 0 and callable(new_func_args[0]):
        return decorator(new_func_args[0])
    return decorator

@example_decorator
def example_func(num):
    return num ** 2

example_func(6, print_me="hello")
```

(Note that no error is raised by pylint if we remove the top-level function (`example_decorator`). The reason for someone to define the decorator this way is so that it's possible to use the bare `@example_decorator`, as well as with arguments `@example_decorator(foo=1)`.)

Using pylint on commit hash e75e37a2908ef1c003ea5816c1e234523bdd38be, this is the error that's printed out:

```
E1123: Unexpected keyword argument 'print_me' in function call (unexpected-keyword-arg)
```

Calling `example_func` works fine:

```
>>> example_task(6, print_me="hello")
hello
36
```

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
E1123: Unexpected keyword argument 'print_me' in function call (unexpected-keyword-arg)
```

### Expected behavior

No `unexpected-keyword-arg` error raised

### Pylint version

```shell
current pylint version in the `main` branch (set for 2.13.0)
```

rmorshea:
Unsure if this is a true solution, but perhaps related: https://github.com/PyCQA/pylint/pull/8332

Archmonger:
All potentially related issues:
- #7643
- #7437
- #8303
- #5735
- #3957
- #3954
- #3662
- #3641
- #5784
- #2578
- #1694

mamiu:
Have the same problem. Here's a very simplified version of the code that displays this error:

```python
def clean_html(html):
    elements_to_delete = [
        {"search_term": "footer", "delete_after": True}
    ]

    for etd in elements_to_delete:
        search_term = etd.pop("search_term", None)
        elements = html.find_all(search_term)

[...]

PR #2926, opened 2019-05-19
Can choose decorators that mutate a function's signature

AWhetter:
## Description
This adds a new option `function_mutators` that allows `too-many-function-args`, `unexpected-keyword-arg`, and `no-value-for-parameter` to not get raised for functions that are decorated with a decorator mentioned in `function_mutators`.

I figured it would be too difficult to try and infer what a decorator was doing so this seemed like the best solution, even though it means that there will be cases where a function is called with incorrect arguments will be missed by pylint when the function is decorated with a mutator decorator.

## Type of Changes
|   | Type |
| ------------- | ------------- |
| ✓  | :bug: Bug fix  |
| ✓  | :sparkles: New feature |
|    | :hammer: Refactoring  |
|    | :scroll: Docs |

## Related Issue

Close #259

Issue #8303, opened 2023-02-16
decorators not processed successfully with alpha_vantage


AnthonyZ71:
### Bug description

Using `pip install --upgrade alpha_vantage` and the following application (av_pylint_test.py):
```
#!/usr/bin/env python
"""Example application to demonstrate pylint false positives"""
import os

from alpha_vantage.timeseries import TimeSeries

api_key = os.getenv("AV_API_KEY")
time_series = TimeSeries(key=api_key)
av_data, _ = time_series.get_daily_adjusted("IBM", "compact")
av_quote_dates = av_data.keys()
for quote_date in av_quote_dates:
    print(f"{quote_date} : {av_data[quote_date]['4. close']}")
```

Unfortunately, to get an [ALPHA VANTAGE API key](https://www.alphavantage.co/support/#api-key) they do require to register with their website.  Using the `demo` key doesn't work.

This *may* be a duplicate of #7437 but I'm not sure if that is the case or not.

Following the code a little bit:
```
    @av._output_format
    @av._call_api_on_func
    def get_daily_adjusted(self, symbol, outputsize='compact'):
        _FUNCTION_KEY = "TIME_SERIES_DAILY_ADJUSTED"
        return _FUNCTION_KEY, 'Time Series (Daily)', 'Meta Data'
```
The _call_api_on_func looks like it creates and returns a wrapper function that invokes the API which is then passed into _output_format which creates another wrapper around the function to convert the output of the API into the requested format.

### Command used

```shell
pylint av_pylint_test.py
```

### Pylint output

```shell
************* Module av_pylint_test
av_pylint_test.py:9:0: W0632: Possible unbalanced tuple unpacking with sequence defined at line 76 of alpha_vantage.timeseries: left side has 2 labels, right side has 3 values (unbalanced-tuple-unpacking)
av_pylint_test.py:10:17: E1101: Instance of 'str' has no 'keys' member (no-member)

------------------------------------------------------------------
Your code has been rated at 2.50/10 (previous run: 2.50/10, +0.00)
```

### Expected behavior

pylint would follow the decorators and understand the actual output.

### Pylint version

```shell
pylint 2.16.2
astroid 2.14.2
Python 3.11.2 (main, Feb 11 2023, 02:24:27) [GCC 10.2.1 20210110]
```

### OS / Environment

vscode devcontainer based upon mcr.microsoft.com/devcontainers/python:3

### Additional dependencies

alpha-vantage==2.3.1
pandas==1.5.3

[...]

Issue #7643, opened 2022-10-17
Support for Overload Decorators


WilliamMiteff:
### Current problem

When using `@overload` only the first version in the code is pulled into the classDiagram. 

### Desired solution

I would like to see all overload options in the class diagram.

jacobtylerwalls:
Likely duplicate of PyCQA/astroid#1015

WilliamMiteff:
Any update on this?

PR #8332, opened 2023-02-23
Declare function mutators with inline comment

rmorshea:
_Preface: There may be a better way to do this, but this is my first contribution and I'm unfamiliar with the code base. I'm mostly hoping this can be a starting point for discussion on how this should actually be achieved. I'll update docs if/when this approach is given the go-ahead._

This addresses [this comment](https://github.com/PyCQA/pylint/pull/2926#issuecomment-747463194) requesting a way to mark decorators that mutate function signatures via a comment. This is vital for package authors since, if they create a mutative decorator, all their users would have to manually populate their `function_mutators` config.

I was able to allow mutative decorators to be marked with inline comments by creating an "error" called `decorator-preserves-signature` which, when disabled via a comment on a decorator definition, will disable function argument checks on calls to functions that have that decorator. I used an error to do this in order to consume the inline comment state via the `linter.file_state._supression_mapping`.

So for example, all functions decorated with the following decorator would not have their arguments checked

```python
def mutative_decorator(fun):  # pylint: disable=signature-mutator
    """Yet another decorator that changes a function's signature"""
    def wrapper(*args, do_something=None, **kwargs):
        if do_something:
            return fun(*args, **kwargs)

        return None

    return wrapper

@mutative_decorator
def my_func(): pass

my_func(do_something='a-thing')  # OK
```

## Type of Changes

|   maybe    | :bug: Bug fix          |
|       | :hammer: Refactoring   |
|       | :scroll: Docs          |

## Description

Refs https://github.com/PyCQA/pylint/pull/2926#issuecomment-747463194
Maybe closes this issue https://github.com/PyCQA/pylint/issues/5784

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,3 @@ build-stamp
 .pytest_cache/
 .mypy_cache/
 .benchmarks/
-venv
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ build-stamp
 .pytest_cache/
 .mypy_cache/
 .benchmarks/
+venv
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -381,6 +381,11 @@ MSGS: dict[str, MessageDefinitionTuple] = {
         "Used when a slice step is 0 and the object doesn't implement "
         "a custom __getitem__ method.",
     ),
+    "E1145": (
+        "",
+        "decorator-preserves-signature",
+        "Ignore invalid argument errors on calls to this function",
+    ),
     "W1113": (
         "Keyword argument before variable positional arguments list "
         "in the definition of %s function",
@@ -1454,10 +1459,18 @@ accessed. Python regular expressions are accepted.",
             return
 
         # Has the function signature changed in ways we cannot reliably detect?
-        if hasattr(called, "decorators") and decorated_with(
-            called, self.linter.config.signature_mutators
-        ):
-            return
+        if getattr(called, "decorators", None):
+            if decorated_with(called, self.linter.config.signature_mutators):
+                return
+
+            called_decorator: astroid.NodeNG
+            for called_decorator in filter(
+                None, map(safe_infer, called.decorators.nodes)
+            ):
+                if not self.linter.file_state._module_msgs_state.get("E1145", {}).get(
+                    called_decorator.lineno, True
+                ):
+                    return
 
         num_positional_args = len(call_site.positional_arguments)
         keyword_args = list(call_site.keyword_arguments.keys())
--- a/pylint/checkers/typecheck.py
+++ b/pylint/checkers/typecheck.py
@@ -382,9 +382,10 @@ MSGS: dict[str, MessageDefinitionTuple] = {
         "a custom __getitem__ method.",
     ),
     "E1145": (
-        "",
-        "decorator-preserves-signature",
-        "Ignore invalid argument errors on calls to this function",
+        "Decorator does not preserve function signature",
+        "signature-mutator",
+        "Emitted when a decorator does not preserve the signature "
+        "of the functions is takes as inputs.",
     ),
     "W1113": (
         "Keyword argument before variable positional arguments list "
--- a/tests/functional/a/arguments.py
+++ b/tests/functional/a/arguments.py
@@ -240,6 +240,17 @@ def other_mutation_decorator(fun):
     return wrapper
 
 
+def yet_another_mutation_decorator(fun):  # pylint: disable=decorator-preserves-signature
+    """Yet another decorator that changes a function's signature"""
+    def wrapper(*args, do_something=True, **kwargs):
+        if do_something:
+            return fun(*args, **kwargs)
+
+        return None
+
+    return wrapper
+
+
 @mutation_decorator
 def mutated_function(arg):
     return arg
@@ -250,11 +261,17 @@ def mutated(arg):
     return arg
 
 
+@yet_another_mutation_decorator
+def another_mutated_function(arg):
+    return arg
+
+
 mutated_function(do_something=False)
 mutated_function()
 
 mutated(do_something=True)
 
+another_mutated_function(do_something=False)
 
 def func(one, two, three):
     return one + two + three
--- a/tests/functional/a/arguments.py
+++ b/tests/functional/a/arguments.py
@@ -240,7 +240,7 @@ def other_mutation_decorator(fun):
     return wrapper
 
 
-def yet_another_mutation_decorator(fun):  # pylint: disable=decorator-preserves-signature
+def yet_another_mutation_decorator(fun):  # pylint: disable=signature-mutator
     """Yet another decorator that changes a function's signature"""
     def wrapper(*args, do_something=True, **kwargs):
         if do_something:


AntonyMilneQB:
I am not at all familiar with pylint so can't comment on the implementation but this feature would be extremely helpful I think, so thank you very much @rmorshea for working on it!

Just two questions:
1. Are there other examples in pylint where you can disable warnings at the point of function definition rather than point of call?
2. If this effectively adds a function to `signature-mutators`, should the name of the "error" `decorator-preserves-signature` match that somehow? e.g. `signature-mutated` or `signature-unmutated` depending on which way round you want to call it.

rmorshea:
@AntonyMilneQB

1. As far as I'm aware, the indirect effects of disabling this error are unique in PyLint. If it's not I'd definitely want to copy whatever approach was taken previously.
2. I renamed the error. The comment is now `pylint: disable=signature-mutator`.

DanielNoord:
@Pierre-Sassoulas What do you think about this? I don't know if I prefer this approach but also don't have an immediate better solution yet..

rmorshea:
@DanielNoord, given that the source of the problem arises from the decorator definition, rather than from the call-site for functions that use that decorator, I think some level of indirection is required. The only other option would be to have PyLint inspect and analyze the annotated return type of the offending decorator, but at that point, you require the decorator author to use type hints and you have to implement some complex static analysis.

Archmonger:
All potentially related issues:
- #7643
- #7437
- #8303
- #5735
- #3957
- #3954
- #3662
- #3641
- #5784
- #2578
- #1694

rmorshea:
Wow that's quite a list! Thanks for collecting that.

rmorshea:
- #7643 - seems like this could be useful patch until a true solution is found, but would require adding the ability to add an ignore comment at the decorator usage rather then decorator definition since overload is built-in.
-  #7437 - seems unrelated
- #8303 - can't tell - too little information in the issue
- #5735 - seems unrelated
- #3957 - duplicate of #2578
- #3954 - duplicate of #2578
- #3662 - unrelated
- #3641 - unrelated
- #5784 - :heavy_check_mark: related
- #2578 - related, but I don't think this would fix it unless we could tell PyLint to ignore type checks on the return type of decorators marked as signature mutators.
- #1694 duplicate of #2578

.gitignore
```diff
@@ -27,3 +27,4 @@ build-stamp
 .pytest_cache/
 .mypy_cache/
 .benchmarks/
+venv
```
Pierre-Sassoulas, line 30:
```suggestion
```
Don't put env specific things here, you can create a global gitignore to not have to add 'venv' or `.idea` in all your projects see https://stackoverflow.com/a/7335487/2519059.

pylint/checkers/typecheck.py
```diff
@@ -381,6 +381,12 @@ def _missing_member_hint(
         "Used when a slice step is 0 and the object doesn't implement "
         "a custom __getitem__ method.",
     ),
+    "E1145": (
```
Pierre-Sassoulas, line 384:
```suggestion
    "E1145": (
```
Should not be an error imo, but I also think it should not be a message either. Or at least some prior design is in order. Isn't the goal to auto-populate the function-mutator option based on comment by the lib authors directly in their code ?  I like this the end goal, but this would be something new in pylint, with the expected associated maintenance cost. Maybe a hack like this is using the existing message control framework is good, maybe it would be simpler to suggest to add a value to ``function-mutator`` manually in the message ? This is not something that is done often, even less so if we do configuration template, this way it's done once per library and become very reasonable.

Pierre-Sassoulas (commented):
Thank you for the design proposal, this is thought provoking.

pylint/checkers/typecheck.py
rmorshea: > Should not be an error imo, but I also think it should not be a message either.

Agreed. I just needed a simple way to hook into the inline comment system. It feels like this comment is meant to "mark" decorator definitions as being signature-mutators. Thus, a more ideal comment might read:

```python
def my_decorator(func): # pylint: mark=signature-mutator
    ...
```

I'm not really sure how to do this though.

> Maybe it would be simpler to suggest to add a value to function-mutator manually in the message?

If I understand this correctly you're suggesting that, if PyLint discovers invalid arguments, and the function in question has a decorator, that PyLint report the error, but also recommend adding the decorator to the `signature-mutators` list manually. If so, I think this could be problematic. Signature mutating decorators are a fairly advanced concept and users might add decorators to that list without fully understanding what they're doing. I foresee bug reports to PyLint from people who have done this and are getting false negatives for invalid function calls. Further, as a lib author, even with this addition to the error message, I'm still probably going to get bug reports from people complaining about this PyLint false positive. Ultimately, as a lib author myself, I didn't want to deal with that and [chose to just not use the decorator syntax](https://github.com/idom-team/idom/blob/main/src/idom/html.py#L400-L446).

AntonyMilneQB: I don't have any useful suggestions on implementation (sorry), but just to add a bit of weight to the cause... I'm coming from a completely unrelated project from @rmorshea and was independently looking for a solution to the exact same problem. I agree with everything he's said here. I understand there's no other patterns like this in pylint at the moment, but it would be much, much more convenient if this `disable` could be somehow specified at point of function definition rather than at point of call. I'm developing a package that defines decorated functions that will unfortunately raise `unexpected-keyword-arg` when called. At the moment there's a burden on every user that uses pylint and calls my function to update their config; it would be much more maintainable if the burden fell on the package author instead.

Archmonger: Perhaps rather than creating a new pattern, should pylint consider disabling this error on all decorated functions?

In theory, this change should only be a temporary until a method is determined on how to properly parse and analyze these functions.

Pierre-Sassoulas: We have ``function-mutator`` to handle this. Some functions are just too dynamic and cannot reasonably be parsed but that's not all functions and pylint would be crippled if we just refuse to analyses all decorators. Maybe we could suggest to populate ``function_mutator`` only on decorator we're not sure about. Maybe we can add more default values in function-mutator, for well known libs.

Archmonger: Brutal opinion here: If I need to start populating a list of decorated functions within pylintrc, I'd likely remove pylint from my workflow instead.

DanielNoord:
I don't think this proposal ever went anywhere. Personally I still don't really like using comments as it quickly allows for more complexity to seep into `pylint`. I'd suggest closing this as a nice thought provoking experiment but not something we will act on now.

rmorshea:
Happy to close if maintainers don't see it going anywhere.

It might be better for users to just disable `too-many-function-args`, `unexpected-keyword-arg`, and `no-value-for-parameter` - I'd argue that other tools do a better job of performing those checks and maintaining a `function_mutators` list is quite burdensome. Give all that, the complexity this adds to PyLint's internals doesn't seem worth it.

Pierre-Sassoulas:
I think we can close, but we should create a related issue because a lot of info are here only and would be lost otherwise. It feel like pylint need to understand function mutators better (easier said than done) and prefer false negative over false positive as long as it does not understand them.

Pierre-Sassoulas:
They were two existing "generic" issues about decorators, #3954 and #2578, closing after I added a comment there.

PR #8337, opened 2023-02-24
Prevent emitting ``invalid-name``

### LINKED ISSUE #8307: 'Constant name doesn't conform' warning on `global` statement instead of declaration
### Bug description
```python
import typing as _tp

_foo: _tp.Optional["Foo"] = None


class Foo:
    ...


def setup_shared_foo():
    global _foo

    _foo = Foo()
```

### Configuration

_No response_

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
test.py:11:4: C0103: Constant name "_foo" doesn't conform to UPPER_CASE naming style ('([^\\W\\da-z][^\\Wa-z]*|__.*__)$' pattern) (invalid-name)
```

### Expected behavior

Pylint should warn about the 'incorrect' variable name on line 3, where the variable is defined, not line 11, where it is used in a `global` statement.

### Pylint version

```shell
pylint 2.16.2
astroid 2.14.2
```

### OS / Environment

Windows 10

### Additional dependencies

_No response_

mbyrnepr2:
Prevent emitting ``invalid-name`` for the line on which a ``global`` statement is declared.

## Type of Changes

## Description
The `visit_global` function seems to be redundant because the `visit_assignname` method is already running the [self._check_name("const", node.name, node)](https://github.com/PyCQA/pylint/blob/main/pylint/checkers/base/name_checker/checker.py#L446) method on the node representing the module-level assignment.

Closes #8307

### DIFF
--- a/tests/functional/g/globals.py
+++ b/tests/functional/g/globals.py
@@ -102,3 +102,12 @@ def init_connection_state(alias):
     global RAN_DB_DICT  # [global-variable-not-assigned]
     RAN_DB_SET.add(alias)
     return RAN_DB_DICT.setdefault("color", "Palomino")
+
+
+# Prevent emitting `invalid-name` for the line on which `global` is declared
+# https://github.com/PyCQA/pylint/issues/8307
+
+_foo: str = "tomato"
+def setup_shared_foo():
+    global _foo  # [global-statement]
+    _foo = "potato"
--- a/doc/whatsnew/fragments/8307.bugfix
+++ b/doc/whatsnew/fragments/8307.bugfix
@@ -1,0 +1,3 @@
+Prevent emitting ``invalid-name`` for the line on which a ``global`` statement is declared.
+
+Closes #8307
--- a/pylint/checkers/base/name_checker/checker.py
+++ b/pylint/checkers/base/name_checker/checker.py
@@ -388,11 +388,6 @@ class NameChecker(_BasicChecker):
 
     visit_asyncfunctiondef = visit_functiondef
 
-    @utils.only_required_for_messages("disallowed-name", "invalid-name")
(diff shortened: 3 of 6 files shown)


tests/functional/g/globals.py
```diff
@@ -102,3 +102,12 @@ def init_connection_state(alias):
     global RAN_DB_DICT  # [global-variable-not-assigned]
     RAN_DB_SET.add(alias)
     return RAN_DB_DICT.setdefault("color", "Palomino")
+
+
+# Prevent emitting `invalid-name` for the line on which `global` is declared
+# https://github.com/PyCQA/pylint/issues/8307
+
+_foo: str = "tomato"
```
Pierre-Sassoulas, line 110: But it should be raised here, right ?
```suggestion
_foo: str = "tomato"  # [disallowed-name]
```

Pierre-Sassoulas: Yeah sorry I'm confused about it, there's a lot of discussion around invalid-name, hard to know what's done or not (https://github.com/PyCQA/pylint/issues/7305). There's also almost too much to digest in https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/invalid-name.html and yet no explanation about Final.

mbyrnepr2: We have a [global-variable-undefined](https://github.com/PyCQA/pylint/blob/a0b28f9019fabddd8ef428be75659082377abb4c/pylint/checkers/variables.py#L440) message which will be emitted when the global name is not defined in the module scope. So, when it is defined in the module scope the `invalid-name` rule-checks are triggered by `visit_assignname`.

Pierre-Sassoulas (approved):

PR #8353, opened 2023-02-27
security update, bump dockerfile to python:3.10.10-alpine3.17

nashcontrol:
## Type of Changes

## Description

reduce potential security vulnerabilities by upgrading container base image to python:3.10.10-alpine3.17


Pierre-Sassoulas (commented):
👍. We're not using this dockerfile it's not going up to date. Might be a good idea to move this to the doc as an 'example of dockerfile' ? What do you think ?

Pierre-Sassoulas (approved):

PR #8363, opened 2023-02-28
Adding more types to the code

ollie-iterators:
## Type of Changes

## Description

Working on adding more types to the code so that it can run faster.

Refs #XXXX

Closes #XXXX

### DIFF
--- a/pylint/testutils/lint_module_test.py
+++ b/pylint/testutils/lint_module_test.py
@@ -141,7 +141,7 @@ class LintModuleTest:
         ):
             pytest.skip("Test excluded from --minimal-messages-config")
 
-    def runTest(self) -> None:
+    def runTest(self):
         self._runTest()
 
     def _should_be_skipped_due_to_version(self) -> bool:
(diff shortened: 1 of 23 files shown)


pylint/testutils/lint_module_test.py
```diff
@@ -141,7 +141,7 @@ def setUp(self) -> None:
         ):
             pytest.skip("Test excluded from --minimal-messages-config")
 
-    def runTest(self) -> None:
```
Pierre-Sassoulas: Look like you actually removed typing ?

tests/functional/r/regression_02/regression_5244.py
```diff
@@ -8,5 +8,5 @@ class MyClass:
     def some_func(self):
         return lambda: 42
 
-    def __len__(self):
+    def __len__(self) -> int:
```
Pierre-Sassoulas, line 11: Code in functional tests or test data should not be modified to add typing (there's no value and it might actually be harmful).

Pierre-Sassoulas (changes requested):
Thank you for the effort but don't think there's a single change that we can keep in this merge request.

Pierre-Sassoulas:
Also adding typing does not make the code run faster, it permits to detect error when using a type checker like mypy :)

Issue #6473, opened 2022-04-29
os._exit is incorrectly flagged as a "protected-access" issue


ngie-eign:
`os._exit` is a supported function, matching the Unix libcall, `_exit`. It should not be flagged, but is incorrectly flagged by pylint as a W0212 (protected-access) issue.

jacobtylerwalls:
I think there's value in leaving the warning. We don't want to get involved deciding which underscored methods are nevertheless allowable to call. The docs indicate this is [usually not to be called](https://docs.python.org/3/library/os.html?highlight=_exit#os._exit), other than cleaning up after a `fork()`. That means on balance we should warn about its usage, no?

ngie-eign:
> I think there's value in leaving the warning. We don't want to get involved deciding which underscored functions are nevertheless allowable to call. The docs indicate this is [usually not to be called](https://docs.python.org/3/library/os.html?highlight=_exit#os._exit), other than cleaning up after a `fork()`. That means on balance we should warn about its usage, no?

Generally speaking, post-fork(), but there are other instances where os._exit should be used to avoid calling atexit handlers, flushing I/O streams, etc.

Either way, os._exit is a false positive in this case that should be handled differently than it currently is. I realize there needs to be a minimum bar of understanding when using APIs like fork/_exit, but (selfishly) I would like to decrease the barrier of entry so folks who are more junior or less aware of systems programming can trust tools like pylint to help inform them of items which should be handled in code.

jacobtylerwalls:
> Either way, os._exit is a false positive in this case that should be handled differently than it currently is.

I think this is a classic use case for `exclude-protected` in your `pylintrc`. Regarding adding it to the default `pylintrc` generated by `--generate-rcfile`, do you think the false positives outweigh the false negatives from people incorrectly calling `_exit` such that we should add it to the list?

Pierre-Sassoulas:
Adding this issue in the 2.15 milestone so we take a decision regarding ``os._exit`` being in the default value of ``exclude-protected`` by then.

DanielNoord:
@jacobtylerwalls  @Pierre-Sassoulas Any opinion about this?

Pierre-Sassoulas:
I say let's put this in default value, one user want it and took the time to ask for it. Also the documentation clearly explain the context in which you need to call it, I don't think there's an alternative for "impossible to catch sys exit" : https://docs.python.org/3/library/os.html#os._exit

DanielNoord:
[...]

PR #8364, opened 2023-02-28
Accept values of the form ``<class name>.<attribute name>`` for the `exclude-protected` list

### LINKED ISSUE #7343: ``exclude-protected`` using the class name and attribute name (astroid's ``qname``) instead of only the attribute name
### Current problem

Right now only the name is taken into account. For example if we want to make ``os._exit`` permissible we'd make all ``x._exit`` permissible. 

### Desired solution

We use the qname instead so we can allow one off class/attr combination.

### Additional context

See https://github.com/PyCQA/pylint/issues/6473

mbyrnepr2:
## Type of Changes

## Description

Closes #7343

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -1784,6 +1784,12 @@ a metaclass class method.",
         ):
             return
 
+        # Typing annotations in function definitions can include protected members
+        if utils.is_node_in_type_annotation_context(node):
+            return
+
+        # Return if `attrname` is defined at the module-level or as a class attribute
+        # and is listed in `exclude-protected`.
         inferred = safe_infer(node.expr)
         if (
             inferred
@@ -1793,19 +1799,14 @@ a metaclass class method.",
             return
 
         klass = node_frame_class(node)
+        if klass is None:
+            # We are not in a class, no remaining valid case
+            self.add_message("protected-access", node=node, args=attrname)
+            return
 
         # In classes, check we are not getting a parent method
         # through the class object or through super
         callee = node.expr.as_string()
-
-        # Typing annotations in function definitions can include protected members
-        if utils.is_node_in_type_annotation_context(node):
-            return
-
-        # We are not in a class, no remaining valid case
-        if klass is None:
-            self.add_message("protected-access", node=node, args=attrname)
-            return
 
         # If the expression begins with a call to super, that's ok.
         if (
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -815,6 +815,7 @@ a metaclass class method.",
                     "_replace",
                     "_source",
                     "_make",
+                    "os._exit",
                 ),
                 "type": "csv",
                 "metavar": "<protected access exclusions>",
@@ -1776,10 +1777,13 @@ a metaclass class method.",
             Klass.
         """
         attrname = node.attrname
-        if (
-            is_attr_protected(attrname)
-            and attrname not in self.linter.config.exclude_protected
-        ):
+        if is_attr_protected(attrname):
+            if attrname in self.linter.config.exclude_protected:
+                return
+            inferred = safe_infer(node.expr)
+            if inferred and isinstance(inferred, (nodes.ClassDef, nodes.Module)):
+                if f"{inferred.name}.{attrname}" in self.linter.config.exclude_protected:
+                    return
             klass = node_frame_class(node)
 
             # In classes, check we are not getting a parent method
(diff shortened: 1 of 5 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -1776,10 +1777,16 @@ def _check_protected_attribute_access(
             Klass.
         """
         attrname = node.attrname
```
Pierre-Sassoulas, line 1779:
```suggestion
        attrname = node.attrname
        if not is_attr_protected(attrname):
            return
```
Then we can de-indent everything after that.

pylint/checkers/classes/class_checker.py
```diff
@@ -1776,10 +1777,16 @@ def _check_protected_attribute_access(
...
-        if (
-            is_attr_protected(attrname)
-            and attrname not in self.linter.config.exclude_protected
-        ):
+        if is_attr_protected(attrname):
+            if attrname in self.linter.config.exclude_protected:
+                return
+            inferred = safe_infer(node.expr)
```
Pierre-Sassoulas, line 1783: The inference is costly, might want to do that at the end, after other checks like "it's a typing it's ok".

Pierre-Sassoulas (changes requested):
Looks pretty good already, some nit and possible perf improvements.

pylint/checkers/classes/class_checker.py
```diff
@@ -1776,85 +1777,94 @@ def _check_protected_attribute_access(
...
-            # In classes, check we are not getting a parent method
-            # through the class object or through super
-            callee = node.expr.as_string()
+        # If the expression begins with a call to type(self), that's ok.
+        if self._is_type_self_call(node.expr):
+            return
 
-            # Typing annotations in function definitions can include protected members
-            if utils.is_node_in_type_annotation_context(node):
+        inferred = safe_infer(node.expr)
```
Pierre-Sassoulas, line 1815: Didn't we already infer that line 1784 ?

Pierre-Sassoulas (commented):
Sorry my refactor suggestion made a mess of the diff, how about we do it in another prior merge  request ?

pylint/checkers/classes/class_checker.py
mbyrnepr2: Good one - I must have copied the code-added-by-me to this location but not removed it from where it was in the first commit. This commit is fully reverted now until the refactor MR is merged.

pylint/checkers/classes/class_checker.py
mbyrnepr2: Right good point. However we may need to re-think this method. I realise now that currently it is geared towards class-attributes. If we move this new inference logic below [these lines](https://github.com/PyCQA/pylint/pull/8364/files#diff-28b08eca9b507acd5447a6f7107823b8d5e627963cccc780601da4f88f842813R1805-R1807) then the new logic is never called as that `klass` object tends to be `None`.

Pierre-Sassoulas (commented):
Some suggestion about order of execution regarding performance.

pylint/checkers/classes/class_checker.py
```diff
@@ -1777,11 +1778,20 @@ def _check_protected_attribute_access(
         """
         attrname = node.attrname
 
-        if (
+         if (
```
Pierre-Sassoulas, line 1781:
```suggestion
        if (
```

pylint/checkers/classes/class_checker.py
```diff
@@ -1776,12 +1777,20 @@ def _check_protected_attribute_access(
...
+
+        inferred = safe_infer(node.expr)
+        if (
+            inferred
+            and isinstance(inferred, (nodes.ClassDef, nodes.Module))
+            and f"{inferred.name}.{attrname}" in self.linter.config.exclude_protected
+        ):
+            return
+
         klass = node_frame_class(node)
```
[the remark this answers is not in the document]
mbyrnepr2: Sounds good! If you meant to do the refactor in a separate MR let me know. I've made the update in the current MR but that can be undone.

pylint/checkers/classes/class_checker.py
```diff
@@ -1782,21 +1783,31 @@ def _check_protected_attribute_access(
...
+        klass = node_frame_class(node)
         if klass is None:
+            # We are not in a class, no remaining valid case
             self.add_message("protected-access", node=node, args=attrname)
             return
 
+        # In classes, check we are not getting a parent method
+        # through the class object or through super
+        callee = node.expr.as_string()
+
```
Pierre-Sassoulas, line 1810: Could we move that to line 1825  ?

Pierre-Sassoulas (changes requested):
Last perf nit and let's merge.

Pierre-Sassoulas (approved):

mbyrnepr2:
Thanks @Pierre-Sassoulas. Can we also close https://github.com/PyCQA/pylint/issues/6473 now also?

PR #8362, opened 2023-02-28
Changing flake8 to have max-line-length of 115

ollie-iterators:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
|    | :bug: Bug fix          |
|    | :sparkles: New feature |
| ✓   | :hammer: Refactoring   |
|    | :scroll: Docs          |

## Description

Changing flake8 to have a max-line-length of 115 because there are issues with changing it to 103. This makes the code easier to understand.

Refs #XXXX

Closes #XXXX

PR #8372, opened 2023-03-02
[pre-commit] Transition from flake8/autoflake to ruff

Pierre-Sassoulas:
## Type of Changes

## Description

ruff is faster and support pyproject.toml and python 3.7. flake8-typing-import is not implemented but it checks for things in python version < 3.7.2 that we do not support. Fix main following #8362 that broke it.


Pierre-Sassoulas:
The continuous integration takes a long time and it's very rare that a problem happen. When it happens we fix main and it does not impact contributors too much. So it's a balance we try to achieve between performance and not breaking main.

DanielNoord (approved):

PR #8380, opened 2023-03-04
Document ignoring pragmas for calculating line length

### LINKED ISSUE #4802: Make pylint consider `disable params` when checking for line length
### Current problem

As discussed in #1682 `pylint` currently does not consider the length of the disable params when checking for line length.
For example, when max-length is set to 80 the following code would not raise a `line-too-long` warning.
```python
Line with length of 80 columns  # pylint: disable=not-callable
```
According to the issue, this is due to:
https://github.com/PyCQA/pylint/blob/14da5ce258d3818037304a6e0f61aba1462e251d/pylint/checkers/format.py#L1004
Which seems fair.

However, this is not what users might intuitively expect from this warning, as a comment in that issue shows. Some users might want the max-length to also include comments.
Although #4797 fixes some of the issues in the original issue, it does not solve this particular issue. This issue serves to track it.

### Desired solution

Make `pylint` consider comments when checking for max-length of lines.

**Open question:**
What should be default behaviour? Including comments or excluding comments? Current behaviour is excluding, but I would argue that it would be more intuitive to include them. Especially now that users can use `disable-next` to split comments and code over multiple lines.

jacobtylerwalls:
## Type of Changes

## Description
Closes #4802 with a doc update as described there.

### DIFF
--- a/doc/data/messages/l/line-too-long/details.rst
+++ b/doc/data/messages/l/line-too-long/details.rst
@@ -1,3 +1,5 @@
+Pragma controls such as ``# pylint: disable=all`` are not counted toward line length for the purposes of this message.
+
 If you attempt to disable this message via ``# pylint: disable=line-too-long`` in a module with no code, you may receive a message for ``useless-suppression``. This is a false positive of ``useless-suppression`` we can't easily fix.
 
 See https://github.com/PyCQA/pylint/issues/3368 for more information.


Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Thank you !

PR #8383, opened 2023-03-05
Change line-length to 103

ollie-iterators:
## Type of Changes

## Description

Change line-length to 103. This makes the code easier to read, which makes it easier to improve the code.

Refs #XXXX

Closes #XXXX

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - "maintenance/**"
+      - "flake8100" # TODO: Remove flake8100 when it is time to make a PR
   pull_request:
     branches:
       - main
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - "maintenance/**"
-      - "flake8100" # TODO: Remove flake8100 before submitting a PR
   pull_request:
     branches:
       - main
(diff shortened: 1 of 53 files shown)


Pierre-Sassoulas (commented):
Some comments on the first files I checked.

ollie-iterators:
It looks like there are two lines in the code that are relevant for tests that need to be altered for pre-commit to be fixed. I made pre-commit not work because I thought one test failing was easier to fix than having all the 'run / python version' tests be failing

DudeNr33 (commented):
I appreciate your motivation to contribute! 😊 

The discussion about line length certainly comes down to personal taste, but looking at some of the changes I personally feel like we are moving in the wrong direction here. 

While I agree that for pure *code* the line length of 88 characters (automatically enforced by `black`!) is a good thing, I would tend to keep the 115 for comments and strings. 
Reading text is not as mentally demanding as reading code. 
There are now several changes in this PR where a sentence that easily fit into one line before is now split over two lines instead.
A good example here would be the change in `pylint/message/message.py`.

In my opinion this is more offsetting than having an extra 12 characters in one line.

I won't object if the other maintainers prefer the new limit, this is just my personal opinion.

Anyways, thank you for taking the time and effort contributing to Pylint!

Pierre-Sassoulas (commented):
Reviewing this is a lot of work because of hard to spot white-spaces errors (2 got past review in the latest changes and were only caught in #8395) and as @DudeNr33 said it's not a certainty that this make the code better. Sorry for the work you already done on this but trying to merge it would means a lot of work on top of it so I'm going to close.

PR #8385, opened 2023-03-05
Add regression test for #5288

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #5288

PR #8387, opened 2023-03-06
Upgrade to astroid 2.15.0

Pierre-Sassoulas:
## Type of Changes

## Description

Refs #8385

### DIFF
--- a/doc/whatsnew/fragments/8387.feature
+++ b/doc/whatsnew/fragments/8387.feature
@@ -1,3 +1,3 @@
-* pylint now supports ``TryStar`` nodes from python 3.11 and should be fully compatible with python 3.11.
+pylint now supports ``TryStar`` nodes from Python 3.11 and should be fully compatible with Python 3.11.
 
 Closes #8387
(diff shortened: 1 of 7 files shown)


doc/whatsnew/fragments/8387.feature
```diff
@@ -0,0 +1,3 @@
+* pylint now supports ``TryStar`` nodes from python 3.11 and should be fully compatible with python 3.11.
```
DanielNoord, line 1:
```suggestion
pylint now supports ``TryStar`` nodes from Python 3.11 and should be fully compatible with Python 3.11.
```

pylint/pyreverse/inspector.py
```diff
@@ -300,14 +300,14 @@ def visit_importfrom(self, node: nodes.ImportFrom) -> None:
...
-        """Return true if the module should be added to dependencies."""
+    def compute_module(self, context_name: str, mod_path: str) -> bool:
+        """Should the module be added to dependencies ?"""
         package_dir = os.path.dirname(self.project.path)
         if context_name == mod_path:
-            return 0
-        if astroid.modutils.is_standard_module(mod_path, (package_dir,)):
-            return 1
-        return 0
+            return True
```
DanielNoord, line 307: This is a change in behaviour? `bool(0) == False`?

Pierre-Sassoulas: Huh, nice catch. It seems no test caught that 😱 
```suggestion
            return False
```

DanielNoord (approved):

Issue #7607, opened 2022-10-11
What's left for 3.0?


DanielNoord:
Just because I want to gather some thoughts before we can decide when to release `3.0`.

For me the main blocker was always turning on `--strict` in `mypy` as that would lead to many deprecations. If we (ever) fix the typing in `astroid` we will likely see many more things we need to fix in `pylint` as well but at least it somewhat works now. That means I personally have only four main "stretch goals" for `3.0` that I can think of:

G1) Take one last look at `multiprocessing` and see if it can be fixed
G2) Create a new JSON reporter
G3) Overhaul the enable/disable vs. extend configuration
G4) Make a start with offering pylint configuration templates

I'm not sure if they should all be part of `3.0` but I think we should at least start a discussion about what our expectations are and align some of our goals. For me personally only G2 is a real blocker.
It might even make more sense to postpone G3 and G4 to `4.0` but start working on it immediately. We probably have some fallout from the `3.0` release as we deprecate a lot and can then make `4.0` the `pylint-config` and "Update-your-configuration-Update".
G1 I think is also very important as it refers to our current users. Although configuration is a pain for a new user, I think a much larger "threat" is people turning off `pylint` because it is too slow. It's just that we clearly lack the expertise/time to fix this right now.

So, what do others think? Are there any other features that should definitely go in `3.0`? And what do you think is the right timeframe to think about?

Pierre-Sassoulas:
In the same vein there's:

- #5701 This one is a blocker but also trivial to implements.
- https://github.com/PyCQA/pylint/pull/7163 It needs to be discussed / reviewed merged asap, because the only reason blocking it is that we do not have a 3.0 branch yet. 
- #3512 it make more sense to apply it in 3.0 even if we do change this kind of thing in minors and it can be done when we decide to. 
- https://github.com/PyCQA/pylint/issues/5275 is related but I think it should be postponed indefinitely.
- We need to apply all the deprecation and breaking changes that are already planned and advertised. There's a lot, (a search for warning.warn with DeprecationWarning would already show a lot of thing.) It's going to be some work and as soon as we start this we're going to have to maintain the 2.x maintenance branch at the same time as we try to do the 3.0 release which could prove difficult.

> G1 I think is also very important

Performance is important but it can be done in any releases imo.

> For me personally only G2 is a real blocker.

[...]

PR #8401, opened 2023-03-08
[deprecation] Remove ``get_message_definitions`` from ``BaseChecker``

Pierre-Sassoulas:
## Type of Changes


Pierre-Sassoulas:
We're maintaining the last minor version in general but even when we release a major. So we're going to maintain 2.x until we release 3.0, possibly for another minor, see #7607

PR #8403, opened 2023-03-08
Add Python 3.8+ `asyncSetUp` to "defining-attr-methods" list

HorlogeSkynet:
## Type of Changes

|  | :sparkles: New feature |
|   | :hammer: Refactoring   |

## Description

This patch adds `unittest.IsolatedAsyncioTestCase` asynchronous "setUp" coroutine to the list of "defining-attr-methods", in order to silence "attribute-defined-outside-init" (W0201) false positive.

---

Feel free to edit this PR as needed; I'm not sure how we should handle Python 3.8+ constraint here.

Thanks, bye :wave:

### DIFF
--- a/doc/user_guide/configuration/all-options.rst
+++ b/doc/user_guide/configuration/all-options.rst
@@ -626,7 +626,7 @@ Standard Checkers
 """""""""""""""""""""""
 *List of method names used to declare (i.e. assign) instance attributes.*
 
-**Default:**  ``('__init__', '__new__', 'setUp', '__post_init__')``
+**Default:**  ``('__init__', '__new__', 'setUp', 'asyncSetUp', '__post_init__')``
 
 
 --exclude-protected
@@ -663,7 +663,7 @@ Standard Checkers
    [tool.pylint.classes]
    check-protected-access-in-special-methods = false
 
-   defining-attr-methods = ["__init__", "__new__", "setUp", "__post_init__"]
+   defining-attr-methods = ["__init__", "__new__", "setUp", "asyncSetUp", "__post_init__"]
 
    exclude-protected = ["_asdict", "_fields", "_replace", "_source", "_make", "os._exit"]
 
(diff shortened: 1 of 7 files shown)


tests/functional/a/attribute_defined_outside_init.py
```diff
@@ -82,3 +86,10 @@ def prop(self, value):
 class DataClass:
     def __post_init__(self):
         self.a = 42
+
+
+if sys.version_info >= (3, 8):
+    class G(unittest.IsolatedAsyncioTestCase):
```
HorlogeSkynet, line 92: Hey @Pierre-Sassoulas !
Any idea about how we can make this line _pass_ Pylint against 3.7 **and** 3.8+ (3.9 currently in CI) ? 😇 
(It raises `no-member` against 3.7 and when we suppress this warning it then raises `useless-suppression` against 3.9 ☹️)

Pierre-Sassoulas: You should probably set the min and max version of the functional test using a pylintrc and multiple files, here's an example with py-version too : https://github.com/PyCQA/pylint/blob/main/tests/functional/ext/typing/redundant_typehint_argument_py310.rc

Pierre-Sassoulas: min_pyver=3.10 means the test won't be executed at all under python 3.10, the py-version means that when the interpreter is python 3.11 pylint will still raise python 3.10 related messages.

HorlogeSkynet: Thanks for the tip ! PR has been updated (CI seems to pass and a changelog fragment has been added).

:wave:

Pierre-Sassoulas (dismissed):
LGTM, thank you !

HorlogeSkynet:
(One check is still failing against Python 3.11, I guess this is unrelated to these changes as you've encountered the same error elsewhere lately)

Pierre-Sassoulas:
I don't know why the primer repeatedly fail. Could you try rebasing on upstream main so that the base commit is newer ? It could be that the cache from the common commit between your branch and the main branch is not available anymore.

Pierre-Sassoulas (approved):

PR #8407, opened 2023-03-08
[FileState] Deprecation enactment for 3.0.0

Pierre-Sassoulas:
## Type of Changes

## Description

``modname`` and ``msg_store`` are now required to be given in FileState, ``collect_block_lines`` has also been removed.

### DIFF
--- a/pylint/lint/parallel.py
+++ b/pylint/lint/parallel.py
@@ -5,7 +5,6 @@
 from __future__ import annotations
 
 import functools
-import warnings
 from collections import defaultdict
 from collections.abc import Iterable, Sequence
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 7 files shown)


pylint/lint/parallel.py
```diff
@@ -165,6 +165,7 @@ def check_parallel(
             msg_status,
             mapreduce_data,
         ) in executor.map(_worker_check_single_file, files):
+            assert isinstance(base_name, str)
```
DanielNoord, line 168: This should now always be `str` without an assertion

Pierre-Sassoulas: mypy does not like it:
```
pylint/lint/parallel.py:168: error: Incompatible types in assignment (expression has type "Optional[str]", variable has type "str")  [assignment]

Found 1 error in 1 file (checked 3 source files)
```

DanielNoord: The typing on `PyLinter` should be updated!

Pierre-Sassoulas: I rebased so we can rebase and merge, there was quite a lot more to change than I thought :D

DanielNoord (approved):

PR #8409, opened 2023-03-08
Remove deprecated functions and classes

DanielNoord:
## Type of Changes

## Description

We're getting there.


doc/whatsnew/fragments/8409.breaking
```diff
@@ -0,0 +1,7 @@
+A number of old utility functions and classes have been removed.
+These are ``MapReduceMixin``, ``is_inside_lambda``, ``check_messages``,
+``is_class_subscriptable_pep585_with_postponed_evaluation_enabled``,
+``get_python_paht``, ``fix_import_path`` and ``get_global_option``.
```
Pierre-Sassoulas, line 4:
```suggestion
``get_python_path``, ``fix_import_path`` and ``get_global_option``.
```

Pierre-Sassoulas (changes requested):
We need to explain what's the alternative, for example "To make a checker reduce map data simply implement get_map_data and reduce_map_data." etc. etc. (We can copy paste from the deprecation message and adapt).

DanielNoord:
> We need to explain what's the alternative, for example "To make a checker reduce map data simply implement get_map_data and reduce_map_data." etc. etc. (We can copy paste from the deprecation message and adapt).

I was wondering about that. It feels as if the changelog can become very long. Shouldn't we just refer to the code and let the old `DeprecationWarnings` show what needs to be changed?

Pierre-Sassoulas:
Yeah it's going to be long but it's the place where people are going to look first when something break, so it needs to explain what to do imo (also if someone is taking care of the deprecation warnings they won't need it).

Pierre-Sassoulas (approved):

DanielNoord:
But docs fail :(

Pierre-Sassoulas (approved):
I don't think we can use enumeration in the fragments, because it's already inside an enumeration. It's probably fixable but also not worth imo.

PR #8412, opened 2023-03-09
[colorize_ansi] Remove the possibility to use anything else than a MessageStyle

Pierre-Sassoulas:
## Type of Changes

## Description

I think we could be a lot more liberal with the breaking change for small internal things like that.

### DIFF
--- a/pylint/reporters/text.py
+++ b/pylint/reporters/text.py
@@ -37,7 +37,7 @@ class MessageStyle(NamedTuple):
     style: tuple[str, ...] = ()
     """Tuple of style strings (see `ANSI_COLORS` for available values)."""
 
-    def _get_ansi_code(self) -> str:
+    def __get_ansi_code(self) -> str:
         """Return ANSI escape code corresponding to color and style.
 
         :raise KeyError: if a nonexistent color or style identifier is given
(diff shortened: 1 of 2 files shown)


pylint/reporters/text.py
```diff
@@ -37,6 +37,24 @@ class MessageStyle(NamedTuple):
     style: tuple[str, ...] = ()
     """Tuple of style strings (see `ANSI_COLORS` for available values)."""
 
+    def get_ansi_code(self) -> str:
```
DanielNoord, line 40: Let's make this private

Pierre-Sassoulas: It's a little complicated to make a public class function private (then we need to call it with ``.__MessageStyle_`` ... I'm wondering if we could use ``colorama`` entirely instead. Or make this a public API, it's unlikely it's ever going away.

DanielNoord: I don't understand your comment. We can just make the function private and keep `MessageStyle` public right? We have similar patterns for all of our other classes: the class itself is public but not all method are public.

Pierre-Sassoulas:
```python
>>> class MessageStyle:
...     def __ansi_color(self):
...         print("ansi")
... 
>>> m = MessageStyle()
>>> m.__ansi_color()
AttributeError: 'MessageStyle' object has no attribute '__ansi_color'
>>> m._MessageStyle__ansi_color()
ansi
```

DanielNoord: I meant `_get_ansi_code`. That signals that it is private and not part of `MessageStyle`'s API

Pierre-Sassoulas: Well it could be made private with a small refactor [64e5b95](https://github.com/PyCQA/pylint/pull/8412/commits/64e5b95986604ae3265ad839b0b137bae1799669).

DanielNoord (approved):

PR #8415, opened 2023-03-09
`pyreverse`: use colorblind friendly default colors

### LINKED ISSUE #8251: Use better default colors for `pyreverse`
### Current problem

The current default colors used by `pyreverse` when specifying `--colorized` are not visually pleasing and probably also not very colorblind friendly.

### Desired solution

Replace the current default colors with:

```python
[
    "dodgerblue",
    "orange",
    "teal",
    "sienna",
    "orchid",
    "lightsalmon",
    "plum",
    "gray",
    "yellow",
    "steelblue",
]
```

### Additional context

See discussion in https://github.com/PyCQA/pylint/pull/8223

DudeNr33:
## Type of Changes

## Description

As discussed in the issue, with 3.0 we can now change the default color palette.

Closes #8251


Pierre-Sassoulas (approved):
I did not review the graphical result... Code's LGTM :smile:

DudeNr33:
For reference, this is how the new colors would look like:

image

Pierre-Sassoulas:
It looks fine with protanopia (no red) but it's a little hard with other form of colorblindness. 

No blue:

tritanopia (no blue)

No coolor:
achromatopia (no color at all)

No green:
deuteranopia (no green)

orange 2 and goldenrod 9 are not great in particular. This is not something that can be improvised. What were the color before that (maybe it's already taken into account by plantuml) ? Otherwise some research are in order I think :D

DudeNr33:
Oh that's cool, what tool did you use to make the different versions? Or did you just desaturate it yourself?

I tried to adapt the [colorblind palette from seaborn](https://seaborn.pydata.org/tutorial/color_palettes.html#qualitative-color-palettes) with the closest matching HTML colors. 
But I will try to find something better. 👍

Pierre-Sassoulas:
It's firefox dev tools for accessibility (ctrl + shift + I), but there's probably the same thing in chrome. If it's already a colorblind palette then I don't think we're going to do much better than that, this is a hard problem to solve. They probably made the better result for red color blind because it's the most prevalent, it can't be perfect.

DanielNoord:
I am actually colorblind and these colours are indeed sub-optimal. Funilly enough the "no green" output Pierre gives works really well for my type of colour blindness but I don't know if it would work for others.

Pierre-Sassoulas:
I've found three possible color palettes:
IBM's : https://lospec.com/palette-list/ibm-color-blind-safe (It's IBM but there's only six of them)
Paul Tol's : https://personal.sron.nl/~pault/  (7, up to 10)
Ban wong: https://www.nature.com/articles/nmeth.1618 (8)

What's the best one in your opinion @DanielNoord ? Also do you know what type of color blindness you have ? It seems we should target red/green rather than blue/yellow, because red/green is a lot more common.

DanielNoord:
I have red/green, the second link works quite well for me!

DanielNoord:
> All figures in the second one, or a particular one ? :)

All are quite good, but figure 1 is definitely the best!

Pierre-Sassoulas:
I think we should use figure 7 then, there's 9 different values and the contrast with the black text is still high enough (compared to say the dark blue in figure 4 where there is more values but I can't read the value in it personally).

fig7

Let me know if you want me to translate that in plantuml color @DudeNr33 :)

Pierre-Sassoulas (changes requested):
(Colorblindness to take into account)

DudeNr33:
I used the hex codes directly and made the `PlantUMLPrinter` work with both hex codes and CSS names.

Here is an example:
image

Updated the functional test file to make sure that both CSS names and HEX values can be given through the command line.

pylint/pyreverse/main.py
```diff
@@ -35,23 +35,15 @@
...
-    "mediumspringgreen",
+    "#77AADD",  # light blue
+    "#99DDFF",  # light cyan
+    "#44BB99",  # mint
+    "#BBCC33",  # pear
+    "#AAAA00",  # olive
+    "#EEDD88",  # light yellow
+    "#EE8866",  # orange
+    "#FFAABB",  # pink
+    "#DDDDDD",  # pale greay
```
Pierre-Sassoulas, line 46:
```suggestion
    "#DDDDDD",  # pale grey
```

pylint/pyreverse/main.py
```diff
@@ -35,23 +35,15 @@
...
-    "cornflowerblue",
-    "cyan",
-    "darkgoldenrod",
-    "darkseagreen",
-    "dodgerblue",
-    "forestgreen",
-    "gold",
-    "hotpink",
-    "mediumspringgreen",
+    "#77AADD",  # light blue
```
Pierre-Sassoulas, line 38:
```suggestion
    # colorblind friendly palette taken from... (sorry I'm on mobile)
    "#77AADD",  # light blue
```

Pierre-Sassoulas (changes requested):
Amazing we even got an enhancement for hexcode !

pylint/pyreverse/main.py
DudeNr33: Oopsie

Pierre-Sassoulas (dismissed):
👌

Pierre-Sassoulas (approved):

nickdrozd:
I just saw the new colors -- they look great! :art:

PR #8418, opened 2023-03-09
Removal of ini support in configuration

Pierre-Sassoulas:
## Type of Changes

## Description

Reading comments it seems this was planned, but are we really doing it ? It seem the rst generated rely on the functions. What do you think @DanielNoord ?


DanielNoord:
I think this is blocked on `pylint-config` actually providing an updated and upgraded `toml` config file. It can generate a file based on current settings already but we want it to be a bit fancier.

After that I think most of this can be removed as there is no need to `generate-rcfile` anymore. I would still be in favour of deprecating the old command for `3.0` already as I think even `pylint.toml` is so much easier to parse as a format than the old `ini` formats.

Pierre-Sassoulas:
Removing from 3.0 then, we should deprecate and remove later.

DanielNoord:
Coming back to this: now that a lot of CIs will switch over to `ruff` I think we actually provide a lot of value for "older" project that don't want to change much and keep using old formats. Shall we keep the deprecation and point to `.toml` but just not act on it (and close this PR)?

Pierre-Sassoulas:
I think it's the same thing than for EOL python interpreters: Old projects that don't change much can still use the old version of pylint.  That being said this is a small cleanup that can wait and we can always revisit this once we have an automigration tool for toml. (And touching the conf and the args parsing is making me really nervous tbf,)

PR #8426, opened 2023-03-10
`pyreverse`: drop support for `.vcg` format

### LINKED ISSUE #8416: Deprecate / remove support for `.vcg` files in `pyreverse`
[Visual Compiler Graphs](https://www.rw.cdl.uni-saarland.de/people/sander/private/html/gsvcg1.html) is currently one of the supported output formats for `pyreverse`.
However this is a very old format (the linked article talks about MS Windows 3.1 😀), and it is really hard to even find tools that can render this format. 
The [link](https://rw4.cs.uni-sb.de/people/sander/private/html/gshome.html) included in the source code does not contain a download link as promised, and the mentioned debian package [cannot be found anymore](https://packages.debian.org/search?keywords=vcg&searchon=names&suite=stable&section=all).

I talked to various colleagues that I thought might have heard about this format, but no one ever did.

Especially because verifying changes to the `.vcg` testdata is hard and might even scare potential contributors, I propose to deprecate - or even better, as it is seemingly very outdated and probably not used anymore - remove `.vcg` support entirely.

Maybe we could put the deprecation warning in 2.17 and remove it for 3.0?

DudeNr33:
## Type of Changes

... "Refactoring" 😅 

## Description

As discussed, let's drop the support for this outdated format.

Closes #8416


Pierre-Sassoulas (approved):
The best kind of refactoring !

PR #8428, opened 2023-03-10
[pre-commit] Configure ruff to sort import in the doc examples

Pierre-Sassoulas:
## Type of Changes

## Description

Faster, but I'm not sure if ruff handle first party / third parties correctly. It's good enough for the doc example anyway, and the per file configuration is better than a regex...

### DIFF
--- a/requirements_test_pre_commit.txt
+++ b/requirements_test_pre_commit.txt
@@ -4,7 +4,7 @@ pre-commit~=3.1;python_version>='3.8'
 bandit==1.7.4
 black==23.1.0
 flake8==6.0.0;python_version>='3.8'
-ruff
+ruff==0.0.254
 flake8-typing-imports==1.14.0;python_version>='3.8'
 isort==5.12.0;python_version>='3.8'
 mypy==1.0.1
--- a/requirements_test_pre_commit.txt
+++ b/requirements_test_pre_commit.txt
@@ -1,10 +1,10 @@
-# Everything in this file should reflect the pre-commit configuration
-# in .pre-commit-config.yaml
-pre-commit~=3.1;python_version>='3.8'
-bandit==1.7.4
-black==23.1.0
-flake8==6.0.0;python_version>='3.8'
-ruff==0.0.254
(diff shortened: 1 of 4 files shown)


requirements_test_pre_commit.txt
```diff
@@ -4,7 +4,7 @@ pre-commit~=3.1;python_version>='3.8'
 bandit==1.7.4
 black==23.1.0
 flake8==6.0.0;python_version>='3.8'
-ruff
+ruff==0.0.254
```
DanielNoord, line 7: I'd rather just have this install latest right? Pinning feels unnecessary as `pre-commit` will get the version as specified anyway.

Pierre-Sassoulas: Yeah, I'm fed up with approving dependabot's dozen of MR too.

Pierre-Sassoulas: Should I do another commit for that, or change this in another MR?

DanielNoord: Can we just remove this file perhaps?

DanielNoord: Let's do a separate commit in this PR

Pierre-Sassoulas: We had this argument with Marc, he wanted to have the possibility to use tools without pre-commit, I did not like the duplication personally. I'm not sure about pinging him for that though.

DanielNoord: With dependabot we are on the latest version anyway in `pre-commit`. Specifying the versions in a file does not improve the workflow significantly as you would still need to `pip install` yourself manually.

Specifying the tools we use seems fine, pinning them to versions and having to update seems a timewaste

Pierre-Sassoulas: So let's keep the file and not pin then. Do you think we need another MR of should I just fix this here ?

DanielNoord: Let's fix here

DanielNoord (approved):

Issue #8167, opened 2023-02-02
False positive  `used-before-assignment` when usage and assignment guarded by the same test in different statements


tvandeveenzepp:
### Bug description

The `used-before-assignment` check reports false positive for using the same variable inside multiple `if __name__ == '__main__'` statements.

```
# pylint: disable=missing-docstring

if __name__ == '__main__':
    X = 1

if __name__ == '__main__':
    print(X)
```

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
************* Module test
test.py:7:10: E0601: Using variable 'X' before assignment (used-before-assignment)

-----------------------------------
Your code has been rated at 0.00/10
```

### Expected behavior

No error.

### Pylint version

```shell
pylint 2.16.0
astroid 2.14.1
Python 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)]
```

clavedeluna:
It would appear `if __name__ == '__main__':` is causing different behavior because if you do:
```
# pylint: disable=missing-docstring

if True:
    X = 1

if True:
    print(X)
```

the msg is not emitted. My hunch is this is happening because pylint isn't loading it as a script which is what `if __name__ == '__main__':` requires.

jacobtylerwalls:
Likely met the "always false" test in #6677.

jacobtylerwalls:
We may need to check if the variable is guarded again by the same(ish) test.

saranti:
This issue wasn't present in v2.15.10. I'll take this up if no one objects

jacobtylerwalls:
This also affects `if TYPE_CHECKING:` e.g.:

```python
if TYPE_CHECKING:
    import math

if TYPE_CHECKING:
    assert math
```

This is because `TYPE_CHECKING`, like `__name__ == '__main__'`, is false at runtime except under special circumstances.

PR #8431, opened 2023-03-11
Fix `used-before-assignment` TYPE_CHECKING false negatives

### LINKED ISSUE #8198: ``used-before-assignment`` false negative with ``TYPE_CHECKING``
### Bug description

Within functions ``used-before-assignment`` doesn't get raised for some edge cases.

```python

if TYPE_CHECKING:
    from datetime import datetime

# pylint: disable=missing-function-docstring, missing-module-docstring


def func_two():
    second = datetime.now()
    return second


def func():
    first: datetime
    first = datetime.now()
    second = datetime.now()
    return first, second

```

### Configuration

```ini
Default config.
```

### Command used

```shell
pylint
```

### Pylint output

```shell
************* Module test
test.py:10: [E0601(used-before-assignment), func_two] Using variable 'datetime' before assignment


Your code has been rated at 4.55/10 (previous run: 10.00/10, -5.45)
```

### Expected behavior

```shell
************* Module test
test.py:10: [E0601(used-before-assignment), func_two] Using variable 'datetime' before assignment
test.py:16: [E0601(used-before-assignment), func_two] Using variable 'datetime' before assignment


Your code has been rated at 4.55/10 (previous run: 10.00/10, -5.45)
```

### Pylint version

```shell
2.16.1
```

### OS / Environment

/

### Additional dependencies

_No response_

zenlyj:
## Type of Changes

## Description

This PR addresses false negative cases for `used-before-assignment`, when imports are made within `TYPE_CHECKING` guard clauses.

The root cause of these false negatives is the consumption of `Import`, `ImportFrom` nodes within `TYPE_CHECKING` clauses. Upon consumption, subsequent usages are not evaluated by the checker.

My approach is to not consume these nodes and keep track of usage scopes, so that errors can be emitted in multiple scopes.

Feel free to discuss alternative solutions! 🙂 

Closes #8198

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1257,6 +1257,7 @@ class VariablesChecker(BaseChecker):
             tuple[nodes.ExceptHandler, nodes.AssignName]
         ] = []
         """This is a queue, last in first out."""
+        self._evaluated_type_checking_scopes: dict[str, list[nodes.NodeNG]] = {}
         self._postponed_evaluation_enabled = False
 
     @utils.only_required_for_messages(
@@ -1717,30 +1718,16 @@ class VariablesChecker(BaseChecker):
         if found_nodes is None:
             return (VariableVisitConsumerAction.CONTINUE, None)
         if not found_nodes:
-            if (
-                not (
-                    self._postponed_evaluation_enabled
-                    and utils.is_node_in_type_annotation_context(node)
-                )
-                and not self._is_builtin(node.name)
-                and not self._is_variable_annotation_in_function(node)
-            ):
-                confidence = (
-                    CONTROL_FLOW
-                    if node.name in current_consumer.consumed_uncertain
-                    else HIGH
-                )
-                self.add_message(
-                    "used-before-assignment",
-                    args=node.name,
-                    node=node,
-                    confidence=confidence,
-                )
+            self._report_unfound_name_definition(node, current_consumer)
             # Mark for consumption any nodes added to consumed_uncertain by
             # get_next_to_consume() because they might not have executed.
+            nodes_to_consume = current_consumer.consumed_uncertain[node.name]
+            nodes_to_consume = self._filter_type_checking_import_from_consumption(
+                node, nodes_to_consume
+            )
             return (
                 VariableVisitConsumerAction.RETURN,
-                current_consumer.consumed_uncertain[node.name],
+                nodes_to_consume,
             )
 
         self._check_late_binding_closure(node)
@@ -1905,6 +1892,61 @@ class VariablesChecker(BaseChecker):
                     return (VariableVisitConsumerAction.RETURN, found_nodes)
 
         return (VariableVisitConsumerAction.RETURN, found_nodes)
+
+    def _report_unfound_name_definition(
+        self, node: nodes.NodeNG, current_consumer: NamesConsumer
+    ) -> None:
+        """Reports used-before-assignment when all name definition nodes
+        gets filtered out by NamesConsumer.
+        """
+        if (
+            self._postponed_evaluation_enabled
+            and utils.is_node_in_type_annotation_context(node)
+        ):
+            return
+        if self._is_builtin(node.name):
+            return
+        if self._is_variable_annotation_in_function(node):
+            return
+        if (
+            node.name in self._evaluated_type_checking_scopes
+            and node.scope() in self._evaluated_type_checking_scopes[node.name]
+        ):
+            return
+
+        confidence = (
+            CONTROL_FLOW if node.name in current_consumer.consumed_uncertain else HIGH
+        )
+        self.add_message(
+            "used-before-assignment",
+            args=node.name,
+            node=node,
+            confidence=confidence,
+        )
+
+    def _filter_type_checking_import_from_consumption(
+        self, node: nodes.NodeNG, nodes_to_consume: list[nodes.NodeNG]
+    ) -> list[nodes.NodeNG]:
+        """Do not consume type checking import node as used-before-assignment
+        may invoke in different scopes.
+        """
+        type_checking_import = next(
+            (
+                n
+                for n in nodes_to_consume
+                if isinstance(n, (nodes.Import, nodes.ImportFrom))
+                and in_type_checking_block(n)
+            ),
+            None,
+        )
+        # If used-before-assignment reported for usage of type checking import
+        # keep track of its scope
+        if type_checking_import and not self._is_variable_annotation_in_function(node):
+            self._evaluated_type_checking_scopes.setdefault(node.name, []).append(
+                node.scope()
+            )
+        nodes_to_consume = [n for n in nodes_to_consume if n != type_checking_import]
+        return nodes_to_consume
 
     @utils.only_required_for_messages("no-name-in-module")
     def visit_import(self, node: nodes.Import) -> None:
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -1897,7 +1897,7 @@ class VariablesChecker(BaseChecker):
         self, node: nodes.NodeNG, current_consumer: NamesConsumer
     ) -> None:
         """Reports used-before-assignment when all name definition nodes
-        gets filtered out by NamesConsumer.
+        get filtered out by NamesConsumer.
         """
         if (
             self._postponed_evaluation_enabled
@@ -1927,7 +1927,7 @@ class VariablesChecker(BaseChecker):
     def _filter_type_checking_import_from_consumption(
         self, node: nodes.NodeNG, nodes_to_consume: list[nodes.NodeNG]
     ) -> list[nodes.NodeNG]:
-        """Do not consume type checking import node as used-before-assignment
+        """Do not consume type-checking import node as used-before-assignment
         may invoke in different scopes.
         """
         type_checking_import = next(
--- a/doc/whatsnew/fragments/8198.bugfix
+++ b/doc/whatsnew/fragments/8198.bugfix
@@ -1,0 +1,4 @@
+Fix ``used-before-assignment`` false negative when TYPE_CHECKING imports
+are used in multiple scopes.
+
+Closes #8198
--- a/tests/functional/u/used/used_before_assignment_scoping.py
+++ b/tests/functional/u/used/used_before_assignment_scoping.py
@@ -1,0 +1,18 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from datetime import datetime
+
+
+def func_two():
+    second = datetime.now()  # [used-before-assignment]
+    return second
+
+
+def func():
+    first: datetime
+    first = datetime.now()  # [used-before-assignment]
+    second = datetime.now()
+    return first, second
--- a/tests/functional/u/used/used_before_assignment_scoping.txt
+++ b/tests/functional/u/used/used_before_assignment_scoping.txt
@@ -1,0 +1,2 @@
+used-before-assignment:10:13:10:21:func_two:Using variable 'datetime' before assignment:CONTROL_FLOW
+used-before-assignment:16:12:16:20:func:Using variable 'datetime' before assignment:CONTROL_FLOW


tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -7,6 +7,15 @@
...
         import math
+    elif input():
+        import pickle
+    else:
+        with input() as temp:
+            import posix
+    if TYPE_CHECKING:
+        pass
+    else:
+        import math
```
jacobtylerwalls, line 18: Nothing fails if I comment this out.

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -26,11 +35,13 @@
...
     import zoneinfo
 elif input():
-    import calendar, bisect  # pylint: disable=multiple-imports
+    import calendar, bisect, pathlib  # pylint: disable=multiple-imports
     if input() + 1:
         import heapq
     else:
         import heapq
+    if TYPE_CHECKING:
+        import math
```
jacobtylerwalls, line 44: Same, this doesn't seem to be targeted by any tests.

pylint/checkers/utils.py
```diff
@@ -1805,7 +1805,14 @@ def is_typing_guard(node: nodes.If) -> bool:
...
     """Return True if node is part for guarded `typing.TYPE_CHECKING` if block."""
-    return isinstance(node.parent, nodes.If) and is_typing_guard(node.parent)
+    ancestor = get_node_first_ancestor_of_type(node, nodes.If)
+    if ancestor is None:
+        return False
+    if not is_typing_guard(ancestor) or any(
+        node in n.nodes_of_class(nodes.NodeNG) for n in ancestor.orelse
+    ):
+        return is_node_in_typing_guarded_import_block(ancestor)
+    return True
```
jacobtylerwalls, line 1815: I'd like to(*) remove this method in favor of `in_type_checking_block`, which is more robust. Is the implementation of this PR any easier if it's changed to just use `in_type_checking_block`?

* PR coming shortly

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -180,6 +192,10 @@ def defined_in_nested_if_else(self) -> heapq:
         print(heapq)
         return heapq
 
+    def defined_in_type_checking(self) -> posix:  # [used-before-assignment]
+        print(pickle)  # [used-before-assignment]
+        return posix
+
```
jacobtylerwalls, line 198: This test passes for me on main (and the title of the method doesn't tell me what it's testing).

jacobtylerwalls (commented):
Hey, thanks for jumping on this! Mainly I'm wondering if what happens if we use `in_type_checking_block`, which has seen much more development. Is the implementation any simpler?

pylint/checkers/utils.py
zenlyj: `in_type_checking_block` returns True when a name is defined any TYPE_CHECKING conditional branch. Whereas this implementation of `is_node_in_typing_guarded_import_block` is a stricter check that returns True only if a name is defined in the TYPE_CHECKING block.

In the following snippet, `is_node_in_typing_guarded_import_block` returns False while `in_type_checking_block`  returns True.
```
if TYPE_CHECKING:
    pass
else:
    import math
```

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: This tests for false negative, both names are defined in the TYPE_CHECKING block and should raise an error.

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: My intention of this was to ensure that `math` is not imported by line 18, and `used-before-assignment` will be raised.

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: Similar to above, the intention was to ensure that `math` is not defined by line 44 and `used-before-assignment` is raised.

pylint/checkers/utils.py
jacobtylerwalls: Ah, I see. That seems like a useful improvement, but it also feels different enough from the reported issue and the variables checker changes that it should be its own PR.

My original concern is that we shouldn't have a mixture of utilities doing very similar things. (For instance, only `in_type_checking_block` is robust against importing `typing` under an alias, like `if t.TYPE_CHECKING`, which at least one of the packages we regularly test in the primer against does.)

Would something like this work?

```diff
diff --git a/pylint/checkers/utils.py b/pylint/checkers/utils.py
index b7691f88e..1d3f13d20 100644
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1931,6 +1931,8 @@ def in_type_checking_block(node: nodes.NodeNG) -> bool:
     for ancestor in node.node_ancestors():
         if not isinstance(ancestor, nodes.If):
             continue
+        if node in ancestor.orelse:
+            return in_type_checking_block(ancestor)
         if isinstance(ancestor.test, nodes.Name):
             if ancestor.test.name != "TYPE_CHECKING":
                 continue
```

***

Finally, if we go down this path, it's worth noting that we're not distinguishing negations like this:

```python
if not TYPE_CHECKING:
    pass
else:
    import math
```

tests/functional/u/used/used_before_assignment_typing.py
jacobtylerwalls: But they do raise an error already without this PR, is what I'm saying. And the total number of errors goes from 3 down to 2, which isn't an improvement.

Apply this diff to main, and it passes:
```diff
diff --git a/tests/functional/u/used/used_before_assignment_ternary.txt b/tests/functional/u/used/used_before_assignment_ternary.txt
index d991970e4..bcb21df1c 100644
--- a/tests/functional/u/used/used_before_assignment_ternary.txt
+++ b/tests/functional/u/used/used_before_assignment_ternary.txt
@@ -1,3 +1,3 @@
-used-before-assignment:6:14:6:15:invalid:Using variable 'a' before assignment:HIGH
-used-before-assignment:8:14:8:15:invalid:Using variable 'b' before assignment:HIGH
-used-before-assignment:9:10:9:11:invalid:Using variable 'c' before assignment:HIGH
+used-before-assignment:6:14:6:15:invalid:Using variable 'a' before assignment:HIGH
+used-before-assignment:8:14:8:15:invalid:Using variable 'b' before assignment:HIGH
+used-before-assignment:9:10:9:11:invalid:Using variable 'c' before assignment:HIGH
diff --git a/tests/functional/u/used/used_before_assignment_typing.py b/tests/functional/u/used/used_before_assignment_typing.py
index a685bdabc..7f464f061 100644
--- a/tests/functional/u/used/used_before_assignment_typing.py
+++ b/tests/functional/u/used/used_before_assignment_typing.py
@@ -7,6 +7,8 @@ from typing import List, Optional, TYPE_CHECKING
 if TYPE_CHECKING:
     if True:  # pylint: disable=using-constant-test
         import math
+    elif input():
+        import pickle
     from urllib.request import urlopen
     import array
     import base64
@@ -195,3 +197,7 @@ class TypeCheckingMultiBranch:  # pylint: disable=too-few-public-methods,unused-
     def defined_in_with(self) -> base64:
         print(binascii)
         return base64
+
+    def defined_in_type_checking(self) -> posix:  # [undefined-variable]
+        print(pickle)  # [used-before-assignment]
+        return posix  # [undefined-variable]
diff --git a/tests/functional/u/used/used_before_assignment_typing.txt b/tests/functional/u/used/used_before_assignment_typing.txt
index c0a31fae0..54e15e0cb 100644
--- a/tests/functional/u/used/used_before_assignment_typing.txt
+++ b/tests/functional/u/used/used_before_assignment_typing.txt
@@ -1,5 +1,8 @@
-undefined-variable:68:21:68:28:MyClass.incorrect_typing_method:Undefined variable 'MyClass':UNDEFINED
-undefined-variable:73:26:73:33:MyClass.incorrect_nested_typing_method:Undefined variable 'MyClass':UNDEFINED
-undefined-variable:78:20:78:27:MyClass.incorrect_default_method:Undefined variable 'MyClass':UNDEFINED
-used-before-assignment:139:35:139:39:MyFourthClass.is_close:Using variable 'math' before assignment:HIGH
-used-before-assignment:152:20:152:28:VariableAnnotationsGuardedByTypeChecking:Using variable 'datetime' before assignment:HIGH
+undefined-variable:70:21:70:28:MyClass.incorrect_typing_method:Undefined variable 'MyClass':UNDEFINED
+undefined-variable:75:26:75:33:MyClass.incorrect_nested_typing_method:Undefined variable 'MyClass':UNDEFINED
+undefined-variable:80:20:80:27:MyClass.incorrect_default_method:Undefined variable 'MyClass':UNDEFINED
+used-before-assignment:141:35:141:39:MyFourthClass.is_close:Using variable 'math' before assignment:HIGH
+used-before-assignment:154:20:154:28:VariableAnnotationsGuardedByTypeChecking:Using variable 'datetime' before assignment:HIGH
+undefined-variable:201:42:201:47:TypeCheckingMultiBranch.defined_in_type_checking:Undefined variable 'posix':UNDEFINED
+used-before-assignment:202:14:202:20:TypeCheckingMultiBranch.defined_in_type_checking:Using variable 'pickle' before assignment:HIGH
+undefined-variable:203:15:203:20:TypeCheckingMultiBranch.defined_in_type_checking:Undefined variable 'posix':UNDEFINED
```

jacobtylerwalls: Oh, I have an outdated version, this doesn't use `posix` anymore.

zenlyj: My bad, I was changing the names to try and resolve some CI failures. Not very sure why the tests are failing on pypy, could be because of the unsafe implementation of `is_node_in_typing_guarded_import_block`.

jacobtylerwalls: I understand this one better now. I think this is a good to leave on this PR, since I can't get the current version of this test to pass on main 👍

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -180,6 +192,10 @@ def defined_in_nested_if_else(self) -> heapq:
         print(heapq)
         return heapq
 
+    def defined_in_type_checking(self) -> fractions:  # [used-before-assignment]
```
jacobtylerwalls, line 195:
```suggestion
    def defined_in_always_false_branch(self) -> fractions:  # [used-before-assignment]
```

tests/functional/u/used/used_before_assignment_typing.py
jacobtylerwalls: Shouldn't `used-before-assignment` _not_ be raised if `math` is imported at runtime? I'm still having trouble connecting this highlighted portion to the assertions, but I'm hoping you can help me out.

zenlyj: Correct me if I am wrong, but my understanding is that `math` is not imported at runtime as lines 15-18 are within a TYPE_CHECKING block (refer to line 7). Therefore, assertions will be raised for any usages of `math`.

pylint/checkers/utils.py
zenlyj: That was my concern as well, I think this PR is overloaded. 😅 Shall I remove the changes that is meant to address the other false negative case and focus on the VariableChecker changes first?

tests/functional/u/used/used_before_assignment_typing.py
jacobtylerwalls: Thanks, that helps. I see now this is here to prevent a bad implementation of the proposed change that treats `import math` under an else as not guarded by TYPE_CHECKING because here it is guarded two levels up, and this case would have already passed because we were not excluding anything under else branches at all 👍

pylint/checkers/utils.py
jacobtylerwalls: Separating would be very helpful, thank you! You could open a separate issue for your Case 2 in the body of your PR.

zenlyj: Gotcha, apologies for the confusion!

jacobtylerwalls:
I'm also sort of wondering if the variables checker shouldn't be using the typing guard at all. We have a concept in the variables checker of names that are "always false". This seems to raise the warning we want, although I don't know what effect it has on unused-variable:

```diff
diff --git a/pylint/checkers/variables.py b/pylint/checkers/variables.py
index 734231103..a0016606e 100644
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -780,10 +780,11 @@ scope_type : {self._atomic.scope_type}
         """
         uncertain_nodes = []
         for other_node in found_nodes:
-            if in_type_checking_block(other_node):
-                continue
-
-            if not isinstance(other_node, nodes.AssignName):
+            if isinstance(other_node, nodes.AssignName):
+                name = other_node.name
+            elif isinstance(other_node, (nodes.Import, nodes.ImportFrom)):
+                name = node.name
+            else:
                 continue
 
             closest_if = utils.get_node_first_ancestor_of_type(other_node, nodes.If)
@@ -796,7 +797,7 @@ scope_type : {self._atomic.scope_type}
 
             # Name defined in every if/else branch
             if NamesConsumer._exhaustively_define_name_raise_or_return(
-                other_node.name, closest_if
+                name, closest_if
             ):
                 continue
 ```

**Review by Pierre-Sassoulas — CHANGES_REQUESTED** [review]
Could you rebase on the latest main, please ?

**Comment by zenlyj** [comment]
I'm on it, made a mistake while trying to rebase and squash to tidy up the log

**Comment by zenlyj** [comment]
Let me know if there is anything that needs to be updated according to the latest primer results.

**Review by Pierre-Sassoulas — DISMISSED** [review]
Neat change ! Sorry for the delay in reviewing, I was conferencing and teaching the youngling.

File: pylint/checkers/variables.py [hunk 6]
```diff
@@ -1906,6 +1893,61 @@ def _check_consumer(
 
         return (VariableVisitConsumerAction.RETURN, found_nodes)
 
+    def _report_unfound_name_definition(
+        self, node: nodes.NodeNG, current_consumer: NamesConsumer
+    ) -> None:
+        """Reports used-before-assignment when all name definition nodes
+        gets filtered out by NamesConsumer.
```
jacobtylerwalls, line 1900:
```suggestion
        get filtered out by NamesConsumer.
```

pylint/checkers/variables.py
```diff
@@ -1906,6 +1893,61 @@ def _check_consumer(
...
+            "used-before-assignment",
+            args=node.name,
+            node=node,
+            confidence=confidence,
+        )
+
+    def _filter_type_checking_import_from_consumption(
+        self, node: nodes.NodeNG, nodes_to_consume: list[nodes.NodeNG]
+    ) -> list[nodes.NodeNG]:
+        """Do not consume type checking import node as used-before-assignment
```
jacobtylerwalls, line 1930:
```suggestion
        """Do not consume type-checking import node as used-before-assignment
```

jacobtylerwalls (dismissed):
Thanks for fixing this difficult false negative! Nice refactors also.

> Didn't look closely at the primer results and no time to do it atm.

The last one is a real bug in sentry. The others are already captured in #8167, which we have some time to fix before 3.0.

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):

PR #8433, opened 2023-03-11
Deprecate redundant type checking guard utils

jacobtylerwalls:
## Type of Changes

## Description
Suggesting we backport this to 2.17.x so that we can get the notice out before 3.0.


DanielNoord (approved):

Pierre-Sassoulas (approved):
Kinda wondering if this should be part of the astroid's node API. It would be conveniant and discoverable.

jacobtylerwalls:
> Kinda wondering if this should be part of the astroid's node API.

😅 
https://github.com/PyCQA/astroid/blob/395bfbd670a6bf3fa043f5e7b509b9df5b83a714/ChangeLog#L43-L46

Pierre-Sassoulas:
Yeah I remember that, but how are we going to handle control flow ? It seems like this is similar to control flow understanding. And imo it will need to be object oriented instead of functional if we want them to be easy to use (if there was a single function on the node for that we would not have created 3/4 identical functions to solve a single problem?).

jacobtylerwalls:
Gotcha. I see how control flow ideas can be similar to `astroid.are_exclusive`. But I think with something all mixed up like this, where we're also inferring names in the middle, is not really a pure-ast-analysis thing like `are_exclusive`. I don't have a strong opinion though.

Pierre-Sassoulas:
Anyway, it's going to be a design discussion in astroid later, and we need to deprecate the duplicated functions in any cases.

doc/whatsnew/fragments/8433.user_action
```diff
@@ -0,0 +1,10 @@
+The following utilities are deprecated in favor of the more robust ``in_type_checking_block``
```
Pierre-Sassoulas, line 1: I think it should be an internal change not user action. I think user action should be for user and not plugin maintainers.

Pierre-Sassoulas (changes requested):

doc/whatsnew/fragments/8433.user_action
jacobtylerwalls: [f066909](https://github.com/PyCQA/pylint/pull/8433/commits/f0669098ea5055c328d9b5a31440d93efdea7404)

Pierre-Sassoulas (approved):

PR #8440, opened 2023-03-12
add checker for return in finally

### LINKED ISSUE #8260: Prevent usage of `return`, `yield` in `try/finally`
### Current problem

Code in `finally` blocks will always be executed. If there's a `return` inside the `try`, it's possible to overwrite this return value with a `return` in the `finally`.

```py
def add(a: int, b: int) -> int:
    try:
        return a + b
    finally:
        return 0


if __name__ == "__main__":
    print(add(1, 2))  # those args will be ignored; "0" is printed
```

### Desired solution

There should not be any `return` or `yield` statements in a finally block.

### Additional context

https://docs.python.org/3/reference/compound_stmts.html#finally-clause

> The return value of a function is determined by the last return statement executed.

hofrob:
## Type of Changes

## Description

Closes #8260

### DIFF
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -260,6 +260,12 @@ class BasicChecker(_BasicChecker):
             "Used when an exception is created without being assigned, raised or returned "
             "for subsequent use elsewhere.",
         ),
+        "W0134": (
+            "'return' shadowed by the 'finally' clause.",
+            "return-in-finally",
+            "Emitted when a 'return' statement is found in a 'finally' block. This will overwrite "
+            "the return value of a function and should be avoided.",
+        ),
     }
 
     reports = (("RP0101", "Statistics by type", report_by_type_stats),)
@@ -766,6 +772,10 @@ class BasicChecker(_BasicChecker):
         """Update try...finally flag."""
         assert self._tryfinallys is not None
         self._tryfinallys.append(node)
+
+        for final_node in node.finalbody:
+            for return_node in final_node.nodes_of_class(nodes.Return):
+                self.add_message("return-in-finally", node=return_node, confidence=HIGH)
 
     def leave_tryfinally(self, _: nodes.TryFinally) -> None:
         """Update try...finally flag."""
--- a/tests/functional/r/return_in_finally.py
+++ b/tests/functional/r/return_in_finally.py
@@ -1,0 +1,41 @@
+# pylint: disable=missing-docstring,too-few-public-methods,useless-return,lost-exception,inconsistent-return-statements
+
+def second_favorite():
+    fruits = ["apple"]
+    try:
+        return fruits[1]
+    finally:
+        return fruits[0]  # [return-in-finally]
+
+
+def all_favorites():
+    fruits = ["orange", "kiwi", "pineapple"]
+
+    for fruit in fruits:
+        try:
+            return fruit
+        finally:
+            return  # [return-in-finally]
+
+
+def more_favorites():
+    fruits = ["orange", "kiwi", "pineapple"]
+
+    for fruit in fruits:
+        try:
+            return fruit
+        finally:
+            if len(fruit) > 7:
+                return "too many fruits"  # [return-in-finally]
+
+
+def even_more_favorites():
+    fruits = ["orange", "kiwi", "pineapple"]
+
+    for fruit in fruits:
+        try:
+            return fruit
+        finally:
+            if len(fruit) > 7:
+                for fruit_name in fruits:
+                    return f"please remove {fruit_name}"  # [return-in-finally]
--- a/tests/functional/r/return_in_finally.txt
+++ b/tests/functional/r/return_in_finally.txt
@@ -1,0 +1,4 @@
+return-in-finally:8:8:8:24:second_favorite:'return' shadowed by the 'finally' clause.:HIGH
+return-in-finally:18:12:18:18:all_favorites:'return' shadowed by the 'finally' clause.:HIGH
+return-in-finally:29:16:29:40:more_favorites:'return' shadowed by the 'finally' clause.:HIGH
+return-in-finally:41:20:41:56:even_more_favorites:'return' shadowed by the 'finally' clause.:HIGH
--- a/doc/data/messages/r/return-in-finally/bad.py
+++ b/doc/data/messages/r/return-in-finally/bad.py
@@ -1,0 +1,7 @@
+def second_favorite():
+    fruits = ["kiwi", "pineapple"]
+    try:
+        return fruits[1]
+    finally:
+        # because of this `return` statement, this function will always return "kiwi"
+        return fruits[0]  # [return-in-finally]
--- a/doc/data/messages/r/return-in-finally/good.py
+++ b/doc/data/messages/r/return-in-finally/good.py
@@ -1,0 +1,8 @@
+def second_favorite():
+    fruits = ["kiwi", "pineapple"]
+    try:
+        return fruits[1]
+    except KeyError:
+        ...
+
+    return fruits[0]
--- a/doc/data/messages/r/return-in-finally/related.rst
+++ b/doc/data/messages/r/return-in-finally/related.rst
@@ -1,0 +1,1 @@
+- `Python 3 docs 'finally' clause <https://docs.python.org/3/reference/compound_stmts.html#finally-clause>`_
--- a/doc/whatsnew/fragments/8260.new_check
+++ b/doc/whatsnew/fragments/8260.new_check
@@ -1,0 +1,3 @@
+Add ``return-in-finally`` to emit a message if a return statement was found in a finally clause.
+
+Closes #8260
--- a/tests/functional/l/lost_exception.py
+++ b/tests/functional/l/lost_exception.py
@@ -1,4 +1,4 @@
-# pylint: disable=missing-docstring, using-constant-test,cell-var-from-loop
+# pylint: disable=missing-docstring, using-constant-test,cell-var-from-loop,return-in-finally
 
 def insidious_break_and_return():
     for i in range(0, -5, -1):


pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -656,17 +662,32 @@ def leave_module(self, _: nodes.Module) -> None:
...
     visit_while = visit_tryexcept
 
+    def _check_yield_return_in_finally(self, node: nodes.NodeNG) -> None:
+        if isinstance(node, nodes.Return):
+            self.add_message("return-or-yield-in-finally", node=node, args=("return",))
+        if isinstance(node, nodes.Yield):
+            self.add_message("return-or-yield-in-finally", node=node, args=("yield",))
+
+        for child in node.get_children():
+            self._check_yield_return_in_finally(child)
```
Pierre-Sassoulas, line 689: It's possible to use ``for child_node in node.nodes_of_class(nodes.Yield, nodes.Return)`` to recover only the yield/return nodes in a generator.

Pierre-Sassoulas (dismissed):
Thank you for working on this, it looks pretty good already !

hofrob:
I'm not too happy with those examples. The problem is, that pylint should probably not make recommendations on how to fix it. The only real suggestion this checker could offer, would be "the `finally` should only contain shutdown/close commands". If there's a good source where it says what to do in more detail, I'd add it.

I added this checker to the `pylint/checkers/refactoring/refactoring_checker.py` module, but if you have another suggestion, I'll gladly move it. I'm also walking through all statements in a `finalbody`. This might impact performance, though I hope that projects do not have a lot of logic inside of their `finally` clauses :grimacing:.

Other than that, let me know if I should improve the wording or anything else really.

The fact that this new checker only produces one more message in all the checked repos is a good sign in my opinion. I'm not sure if the usage of yield in finally in pycopg2 is a legitimate one?

Pierre-Sassoulas:
It's a very good sign when there's only one new message raised and it's hard to tell if it's a false positive :)

pylint/checkers/refactoring/refactoring_checker.py
hofrob: Oh my, that is good to know. Thanks!

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -490,6 +490,12 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             "value by index lookup. "
             "The value can be accessed directly instead.",
         ),
+        "R1737": (
```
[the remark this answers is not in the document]
hofrob: I moved it to `pylint/checkers/base/basic_checker.py`

hofrob:
After looking at the generated docs, I thought adding code to the bad example might be a bit more helpful.

doc/data/messages/r/return-or-yield-in-finally/bad.py
```diff
@@ -0,0 +1,20 @@
+def second_favorite():
```
Pierre-Sassoulas, line 1: It's possible to create multiple files in a directory too, for multiple examples (a bad directory and a good directory) I think this would be better. We've never done it before so we'll have to deal with the surprises if they are any :)

hofrob: Ok, let's see what happens :grinning:

hofrob: https://pylint--8440.org.readthedocs.build/en/8440/user_guide/messages/warning/return-or-yield-in-finally.html

:champagne:

doc/data/messages/r/return-or-yield-in-finally/good/return.py
```diff
@@ -0,0 +1,8 @@
+def second_favorite():
+    fruits = ["apple"]
```
Pierre-Sassoulas, line 2:
```suggestion
    fruits = ["kiwi", "pineapple"]
```

Pierre-Sassoulas (commented):
I like the embedded comment in the doc

nickdrozd:
The primer is showing one hit that looks like a false positive:

```python
    def _exit_gen(self) -> PQGen[None]:
        """
        Exit current pipeline by sending a Sync and fetch back all remaining results.
        """
        try:
            self._enqueue_sync()
            yield from self._communicate_gen()
        finally:
            # No need to force flush since we emitted a sync just before.
            yield from self._fetch_gen(flush=False)
```

That function yields at least one item in any case, and possibly more. But I don't see any bug there.

I think the warning should only trigger if there is a `return` somewhere.

hofrob:
> ```python
>     def _exit_gen(self) -> PQGen[None]:
>         """
>         Exit current pipeline by sending a Sync and fetch back all remaining results.
>         """
>         try:
>             self._enqueue_sync()
>             yield from self._communicate_gen()
>         finally:
>             # No need to force flush since we emitted a sync just before.
>             yield from self._fetch_gen(flush=False)
> ```

I'm trying to understand why this `yield` is in a `finally` clause. It more or less looks like they're trying to skip "bare-except" lint warnings this way 😀. Because that's what this actually does, right?

So I'm still not convinced that this is actually a false positive. `yield from self._fetch_gen(flush=False)` should come after the `try`/`except` with actual exceptions they want to catch.

😕

nickdrozd:
> I'm trying to understand why this `yield` is in a `finally` clause. It more or less looks like they're trying to skip "bare-except" lint warnings this way grinning. Because that's what this actually does, right?

I'm not sure why it's like that. It's not skipping bare-except because it doesn't actually catch anything. So if there's an exception in `try`, it will interfere with the `yield` in `finally`? But `self._fetch_gen` needs to get called in any case, which might result in side effects.

Barring side effects, I can't think of a good reason to do it this way. So maybe it isn't a false positive. But I could be wrong.

pylint/checkers/base/basic_checker.py
```diff
@@ -266,6 +266,12 @@ class BasicChecker(_BasicChecker):
             "Used when an exception is created without being assigned, raised or returned "
             "for subsequent use elsewhere.",
         ),
+        "W0134": (
+            "'%s' shadowed by the 'finally' clause.",
+            "return-or-yield-in-finally",
+            "Emitted when a 'return' or 'yield' statement is found in a 'finally' block. "
+            "This will overwrite the return value of a function and should be avoided.",
```
jacobtylerwalls, line 273: Speaking a bit out of my depth here, but I think if `return`ing from a generator, the return value won't be overwritten strictly speaking, it will just be passed to a coroutine. So this is slightly incorrectly described for `yield`s, leaving me to suggest to back out all the yield-related stuff.

doc/data/messages/r/return-or-yield-in-finally/bad/yield.py
```diff
@@ -0,0 +1,11 @@
...
+    fruits = ["orange", "kiwi", "pineapple"]
+
+    for fruit in fruits:
+        try:
+            yield fruit
+        finally:
+            # this `yield` statement will add `None` after every fruit
+            # list(all_favorites()) will result in:
+            # ["orange", None, "kiwi", None, "pineapple", None]
+            yield  # [return-or-yield-in-finally]
```
jacobtylerwalls, line 11: Strikes me as a little contrived, who's to say that interleaving None between those fruits is not intended/useful? It's not a crazy use case to need to interleave a constant into a every other position in a sequence.

doc/data/messages/r/return-or-yield-in-finally/bad/return.py
```diff
@@ -0,0 +1,7 @@
+def second_favorite():
+    fruits = ["kiwi", "pineapple"]
+    try:
+        return fruits[1]
+    finally:
+        # because of this `return` statement, this function will always return "kiwi"
+        return fruits[0]  # [return-or-yield-in-finally]
```
jacobtylerwalls, line 7: Great!

jacobtylerwalls (changes requested):
Thanks for this!

Agree with @nickdrozd we should just deal with `return`. I don't think `yield` is really a problem in `finally`. I'd be happy to merge this if just affecting `return`.

hofrob:
Ok, I see that including `yield` is a step too far. I removed references to `yield` now but will do a _final_ (:wink:) check and wait for the CI to finish.

Thanks for chiming in! This PR has bothered me for weeks. But I totally agree now, that this should only look at `return` statements.

tests/functional/r/return_in_finally.py
```diff
@@ -0,0 +1,41 @@
...
+
+    for fruit in fruits:
+        try:
+            return fruit
+        finally:
+            if len(fruit) > 7:
+                return "too many fruits"  # [return-in-finally]
+
+
+def even_more_favorites():
```
jacobtylerwalls, line 32: What about an example with `return` in the finally block but lacks a `return` in the try block. (And would you consider that something that should still emit a message?)

jacobtylerwalls: I'm imagining a legit use case like:

```python
def divide(numerator, denominator):
    try:
        result = numerator / denominator
    except ZeroDivisionError:
        try:
            send_error()
        finally:
            return
    return result
```

hofrob: A `return` in a `finally` will _always_ overwrite the return value. It doesn't matter if there's a `return` inside the `try` itself.

So even if your function catches an error and continues without raising any other exceptions, it will execute the `finally` and if there was a return statement, it will return this value and discard whatever was returned afterwards.

```py
def foo():
    try:
        raise Exception
    finally:
        return 1

    return 2

print(foo())
```

This prints `1`.

hofrob: Wait no. Scrap that (ignore my example).

Let me think about it for a bit.

jacobtylerwalls: But in my example, only if the inner `try` executes, right?
```python
>>> divide(1, 4)
0.25
```

I think I'm asking what you think of checking whether or not the try/finally in question is guarded under `except`.

hofrob: Do you have any real world examples? If I see something like this in a MR/PR, I'd question if there's something wrong in general here. For example, the `send_error()` method should probably take care of any errors. So it would look like this:

```py
def divide(numerator, denominator):
    result = None
    try:
        result = numerator / denominator
    except ZeroDivisionError:
        send_error()  # send_error should handle its own exceptions
    return result
```

So I can't really think of a situation where this would make sense to have a `return` in a `finally`, let alone a `try` in an `except` clause.

Pierre-Sassoulas:
The doc fail is due to a bug that was fixed on main :)

tests/functional/r/return_in_finally.py
jacobtylerwalls: I think the most realistic case to consider would be when you're not in control of `send_error()`,  or if it's more idiomatic for send_error to raise documented, idiomatic exceptions like NotAuthorized or Http500Exception or whatnot. But I don't think we need to block a merge over this; a user in this scenario can just disable `return-in-finally` since they are doing something a little brittle.

pylint/checkers/base/basic_checker.py
```diff
@@ -767,6 +773,10 @@ def visit_tryfinally(self, node: nodes.TryFinally) -> None:
         assert self._tryfinallys is not None
         self._tryfinallys.append(node)
 
+        for final_node in node.finalbody:
+            for return_node in final_node.nodes_of_class(nodes.Return):
+                self.add_message("return-in-finally", node=return_node)
```
jacobtylerwalls, line 778: We add confidence to new messages
```suggestion
                self.add_message("return-in-finally", node=return_node, confidence=HIGH)
```

tests/functional/r/return_in_finally.txt
```diff
@@ -0,0 +1,4 @@
+return-in-finally:8:8:8:24:second_favorite:'return' shadowed by the 'finally' clause.:UNDEFINED
+return-in-finally:18:12:18:18:all_favorites:'return' shadowed by the 'finally' clause.:UNDEFINED
+return-in-finally:29:16:29:40:more_favorites:'return' shadowed by the 'finally' clause.:UNDEFINED
+return-in-finally:41:20:41:56:even_more_favorites:'return' shadowed by the 'finally' clause.:UNDEFINED
```
jacobtylerwalls, line 4:
```suggestion
return-in-finally:8:8:8:24:second_favorite:'return' shadowed by the 'finally' clause.:HIGH
return-in-finally:18:12:18:18:all_favorites:'return' shadowed by the 'finally' clause.:HIGH
return-in-finally:29:16:29:40:more_favorites:'return' shadowed by the 'finally' clause.:HIGH
return-in-finally:41:20:41:56:even_more_favorites:'return' shadowed by the 'finally' clause.:HIGH
```

jacobtylerwalls (approved):

tests/functional/r/return_in_finally.py
hofrob: I know we're drifting off-topic now, but in such a case I would just have a module that takes care of external stuff I can't handle and call my "proxy"-methods instead.

I know it's not a thorough explanation of my thought process though and there still might be edge cases where one has to live with such a situation.

pylint/checkers/base/basic_checker.py
hofrob: Thanks! Done.

jacobtylerwalls (approved):
Thanks for the contribution!

Pierre-Sassoulas:
Great new check, thank you !

PR #8441, opened 2023-03-12
Fix `used-before-assignment` false positive for `TYPE_CHECKING` elif branch imports 

### LINKED ISSUE #8437: `used-before-assignment` false positive for import nodes under `elif` after `TYPE_CHECKING`
### Bug description

```python

if TYPE_CHECKING:
    pass
elif 42:
    import math


print(math)
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:5:5: W0125: Using a conditional statement with a constant value (using-constant-test)
a.py:9:6: E0601: Using variable 'math' before assignment (used-before-assignment)

Your code has been rated at 0.00/10 (previous run: 10.00/10, -10.00)
```

### Expected behavior

************* Module a
a.py:5:5: W0125: Using a conditional statement with a constant value (using-constant-test)

Your code has been rated at 0.00/10 (previous run: 10.00/10, -10.00)

### Pylint version

```shell
pylint 3.0.0a6
astroid 2.16.0dev0
Python 3.11.0
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

zenlyj:
## Type of Changes

## Description

This PR resolves false positives when imports are made in the `TYPE_CHECKING` else if branch.

The issue lies in `is_variable_violation()`. The logic is supposed to rely on the usage of the `TYPE_CHECKING` `if` statement node, but it is relying on the `elif` statement node for such cases.

This solution searches the import node ancestors for the `TYPE_CHECKING` base branch.

Additionally, a new static function `_is_variable_annotation_in_function()` is introduced to refactor the related logic and reduce nesting levels.

Closes #8437

### DIFF
--- a/pylint/checkers/variables.py
+++ b/pylint/checkers/variables.py
@@ -2210,34 +2210,25 @@ class VariablesChecker(BaseChecker):
                 and isinstance(defstmt.parent, nodes.If)
                 and in_type_checking_block(defstmt)
                 and not in_type_checking_block(node)
+                and not VariablesChecker._is_variable_annotation_in_function(node)
+                and not defstmt.parent.parent_of(node)
             ):
-                defstmt_parent = defstmt.parent
-
-                maybe_annotation = utils.get_node_first_ancestor_of_type(
-                    node, nodes.AnnAssign
-                )
-                if not (
-                    maybe_annotation
-                    and utils.get_node_first_ancestor_of_type(
-                        maybe_annotation, nodes.FunctionDef
+                # Exempt those definitions that are used inside the type checking
+                # guard or that are defined in any elif/else type checking guard branches.
+                outermost_type_checking_guard = defstmt.parent
+                defstmt_ancestors_in_type_checking = [n for n in defstmt.node_ancestors() if in_type_checking_block(n)]
+                if defstmt_ancestors_in_type_checking:
+                    guard = utils.get_node_first_ancestor_of_type(defstmt_ancestors_in_type_checking[-1], nodes.If)
+                    outermost_type_checking_guard = guard if guard else outermost_type_checking_guard
+                if outermost_type_checking_guard.has_elif_block():
+                    maybe_before_assign = not utils.is_defined(
+                        node.name, outermost_type_checking_guard.orelse[0]
                     )
-                ):
-                    # Exempt those definitions that are used inside the type checking
-                    # guard or that are defined in any elif/else type checking guard branches.
-                    used_in_branch = defstmt_parent.parent_of(node)
-                    if not used_in_branch:
-                        if defstmt_parent.has_elif_block():
-                            defined_in_or_else = utils.is_defined(
-                                node.name, defstmt_parent.orelse[0]
-                            )
-                        else:
-                            defined_in_or_else = any(
-                                utils.is_defined(node.name, content)
-                                for content in defstmt_parent.orelse
-                            )
-
-                        if not defined_in_or_else:
-                            maybe_before_assign = True
+                else:
+                    maybe_before_assign = not any(
+                        utils.is_defined(node.name, content)
+                        for content in outermost_type_checking_guard.orelse
+                    )
 
         return maybe_before_assign, annotation_return, use_outer_definition
 
@@ -2377,6 +2368,13 @@ class VariablesChecker(BaseChecker):
             if inferred_test.value is False and defnode == defnode_parent.body:
                 return True
         return False
+
+    @staticmethod
+    def _is_variable_annotation_in_function(node: nodes.NodeNG) -> bool:
+        is_annotation = utils.get_node_first_ancestor_of_type(node, nodes.AnnAssign)
+        return is_annotation and utils.get_node_first_ancestor_of_type(
+            is_annotation, nodes.FunctionDef
+        )
 
     def _ignore_class_scope(self, node: nodes.NodeNG) -> bool:
         """Return True if the node is in a local class scope, as an assignment.
--- a/tests/functional/u/used/used_before_assignment_typing.py
+++ b/tests/functional/u/used/used_before_assignment_typing.py
@@ -26,7 +26,7 @@ if TYPE_CHECKING:
     import types
     import zoneinfo
 elif input():
-    import calendar, bisect  # pylint: disable=multiple-imports
+    import calendar, bisect, dbm  # pylint: disable=multiple-imports
     if input() + 1:
         import heapq
     else:
@@ -168,6 +168,7 @@ class TypeCheckingMultiBranch:  # pylint: disable=too-few-public-methods,unused-
     """Test for defines in TYPE_CHECKING if/elif/else branching"""
     def defined_in_elif_branch(self) -> calendar.Calendar:
         print(bisect)
+        print(dbm)
         return calendar.Calendar()
 
     def defined_in_else_branch(self) -> urlopen:
--- a/doc/whatsnew/fragments/8437.bugfix
+++ b/doc/whatsnew/fragments/8437.bugfix
@@ -1,0 +1,4 @@
+Fix a ``used-before-assignment`` false positive when imports 
+are made under the ``TYPE_CHECKING`` else if branch.
+
+Closes #8437


pylint/checkers/variables.py
```diff
@@ -2378,6 +2375,13 @@ def _is_never_evaluated(
                 return True
         return False
 
+    @staticmethod
+    def _is_variable_annotation_in_function(node: nodes.NodeNG) -> bool:
+        is_annotation = utils.get_node_first_ancestor_of_type(node, nodes.AnnAssign)
+        return is_annotation and utils.get_node_first_ancestor_of_type(
```
Pierre-Sassoulas, line 2381:
```suggestion
        return is_annotation and utils.get_node_first_ancestor_of_type( # type: ignore[return-value]
```
astroid does not declare its types, but it's going to be a boolean not Any.

pylint/checkers/variables.py
```diff
@@ -2210,34 +2210,31 @@ def _is_variable_violation(
                 and isinstance(defstmt.parent, nodes.If)
                 and in_type_checking_block(defstmt)
                 and not in_type_checking_block(node)
+                and not VariablesChecker._is_variable_annotation_in_function(node)
+                and not defstmt.parent.parent_of(node)
```
zenlyj, line 2214: pylint raises `too-many-boolean-expressions` here, but at the same time, I don't think it is a good idea to add another `if` branch just to bypass this warning.

Any suggestions? One workaround would be to create another method to perform a few of these checks.

pylint/checkers/variables.py
zenlyj: Thanks for the suggestion!

pylint/checkers/variables.py
jacobtylerwalls: We may need to disable the message here. But first, do you know why `and isinstance(defstmt.parent, nodes.If)` is needed? That seems fishy, since `in_type_checking_block(defstmt)` is supposed to take care of that (and allow for defstmt.parent being something else, like a `with`).

jacobtylerwalls: I guess it's because it's used as `outermost_type_checking_guard` below.

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -26,7 +26,7 @@
     import types
     import zoneinfo
 elif input():
-    import calendar, bisect  # pylint: disable=multiple-imports
+    import calendar, bisect, dbm  # pylint: disable=multiple-imports
```
jacobtylerwalls, line 29: I feat this is starting to get too complex to understand. Are `calendar` and `bisect` here in order to test the same imports being under both branches of TYPE_CHECKING? Or are they here to test multiple imports on the same line? Either way, why would we add `dbm` to the same line, since we want `dbm` to be something not imported above. I wonder if we can just remove `bisect` from the TYPE_CHECKING branch and let that be the regression test--what do you think?

pylint/checkers/variables.py
jacobtylerwalls, line 2214: Although it was here before, nothing fails when I comment out `and not defstmt.parent.parent_of(node)`. So that's a good candidate for a test case.

jacobtylerwalls (changes requested):
Hey, thanks for this. I'm starting to worry we're putting more effort into a flawed implementation. For instance, with these changes, we now have a false negative for this case:

```python

if TYPE_CHECKING:
    pass
elif TYPE_CHECKING:
    import dbm

print(dbm)
```

No pressure, but I'm wondering if you explored whether https://github.com/PyCQA/pylint/pull/8431#issuecomment-1465011806 was helpful in any way?

pylint/checkers/variables.py
Pierre-Sassoulas: If only... :D it would be nice if mypy knew and pre-commit warned us of that kind of thing.

zenlyj:
@jacobtylerwalls thanks, just saw your comment https://github.com/PyCQA/pylint/pull/8431#issuecomment-1465011806. That does seem like a more fitting way for these checks, by filtering out the always false nodes. Let me look into that and in the meantime, I will mark this as draft.

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: makes sense to me, removing `bisect` should achieve the same effect as well. thanks for the suggestion!

ollie-iterators:
Could this be used to fix the false positive for multiple-imports? Because an example of that can be seen in the file: "tests/functional/u/used/used_before_assignment_typing.py" on line 29

zenlyj:
@ollie-iterators line 29 is supposed to raise `multiple-imports`, since it has multiple imports on a single line.

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: Removing `bisect` should achieve the same effect. Thanks for the suggestion!

ollie-iterators:
Oops, I thought multiple imports was because calendar was imported in the if statement and the elif statement.

Pierre-Sassoulas:
The primer's result is surprising because this number of message usually means that we have false positives. But I checked two of them and it seems those are real problems. For example here: https://github.com/psf/black/blob/ef6e079901d53a42dfae4ab10b081ce7a73a47b5/src/blib2to3/pytree.py#L848 if `` if hasattr(sys, "getrefcount"):`` is False then save_stderr is not defined.

zenlyj:
The PR definitely needs more work. You are right most are false positives that point out flaws in the solution.

zenlyj:
@Pierre-Sassoulas What is your opinion on the latest primer results?

It seems that there are many new positives in code like:

```
if ...
    var1 = ...

print(var1)
```

and

```
if ...
    var1 = ...
elif ...
    var1 = ...

print(var1)
```

Would you say that it is reasonable to warn users for these cases, since `var1` may potentially be undefined? Or would it be too opinionated?

Pierre-Sassoulas:
For ``unused-variable`` we have a friend message called ``possibly-unused-variable``, maybe we could do the same thing in this case, idk. But this would be something that is going to make the scope of this PR go *3 and require a prior refactor.

In general we're preferring false negatives vs false positives. I think as long as pylint do not understand basic control flow we're going to make pylint too noisy to be useful if we raise (because this much issue in the primer means that a condition that is always true or false makes this a false positive that is hard to analyse).

tests/functional/u/used/used_before_assignment_typing.py
```diff
@@ -7,11 +7,12 @@
 if TYPE_CHECKING:
     if True:  # pylint: disable=using-constant-test
         import math
+    import dbm
+    print(dbm)
```
jacobtylerwalls, line 11: May I ask what this line tests?

pylint/checkers/variables.py
```diff
@@ -2029,7 +2066,6 @@ def _in_lambda_or_comprehension_body(
             parent = parent.parent
         return False
 
-    # pylint: disable = too-many-branches
```
jacobtylerwalls, line 2032: 🚀

pylint/checkers/variables.py
```diff
@@ -1687,16 +1716,22 @@ def _check_consumer(
...
-            else:
-                confidence = HIGH
-            self.add_message(
-                "used-before-assignment",
-                args=node.name,
-                node=node,
-                confidence=confidence,
-            )
+            if (
+                not self._postponed_evaluation_enabled
```
jacobtylerwalls, line 1720: This seems to be silencing checks for used-before-assignment having nothing to do with annotations, like this case, which used to raise a warning, and which I simplified from the latest primer results for sentry:

```python

if not annotations:
   asfd = 1

print(asfd)
```

pylint/checkers/variables.py
```diff
@@ -728,23 +733,43 @@ def _exhaustively_define_name_raise_or_return(
...
+        ):
+            return NamesConsumer._branch_handles_name(name, node.orelse)
+        # Only search if branch when test condition is inferred to be true
+        if all_inferred and any(
+            isinstance(inferred, nodes.Const)
+            and inferred.value != NotImplemented
+            and inferred.value
+            for inferred in all_inferred
+        ):
+            return NamesConsumer._branch_handles_name(name, node.body)
```
jacobtylerwalls, line 751: This is a neat solution, but can we iterate `all_inferred` only once to determine the correct call to make?

pylint/checkers/variables.py
```diff
@@ -695,26 +694,32 @@ def get_next_to_consume(self, node: nodes.Name) -> list[nodes.NodeNG] | None:
...
-                        name, handler
-                    )
-                    for handler in node.handlers
+            try_except_node = node
+            if isinstance(node, nodes.TryFinally):
+                try_except_node = next(
+                    (
+                        child
+                        for child in node.get_children()
+                        if isinstance(child, nodes.TryExcept)
```
jacobtylerwalls, line 710: I think this is equivalent.
```suggestion
                        for child in node.nodes_of_class(nodes.TryExcept)
```

jacobtylerwalls (changes requested):
Thanks, this is a huge help. I think we're pretty close. One substantive thing to discuss.

Pierre-Sassoulas (dismissed):
The false positives being resolved (useless-suppression now emitted in pandas !) are very encouraging, I think it show the new approach is the right approach. It also read like a cleanup, which is nice ! I'll let Jacob merge :)

pylint/checkers/variables.py
zenlyj: sure, i updated it to iterate over `all_inferred` once

tests/functional/u/used/used_before_assignment_typing.py
zenlyj: I believe this was discussed in previous review: https://github.com/pylint-dev/pylint/pull/8441/files/82fdb4e39cd4716a5b5a080f3fa3f56a3ffb1f03#diff-1c142afc02e8b8c4cf872f61ff8693a0b25646d6db50f16f8027dc130e7f719f.

But the test is to ensure that no warning is emitted when a variable is defined and used in an always false branch.

Pierre-Sassoulas:
Seeing music 21 the following test case should be added:

```python
import typing as t

if t.TYPE_CHECKING:
    from music 21 import Stream

class StreamExample:
    def my_func():
        if t.TYPE_CHECKING:
           assert isinstance(self, Stream)

```

The panda one also seem genuine as we return in the try then we raise in the final except, so fixturedef is always going to be defined:

```python
    def _get_active_fixturedef(
        self, argname: str
    ) -> Union["FixtureDef[object]", PseudoFixtureDef[object]]:
        try:
            return self._fixture_defs[argname]
        except KeyError:
            try:
                fixturedef = self._getnextfixturedef(argname)
            except FixtureLookupError:
                if argname == "request":
                    cached_result = (self, [0], None)
                    return PseudoFixtureDef(cached_result, Scope.Function)
                raise
        # Remove indent to prevent the python3 exception
        # from leaking into the call.
        self._compute_fixture_value(fixturedef)
```

zenlyj:
@Pierre-Sassoulas The music21 case is interesting, because `Stream` is used under a `TYPE_CHECKING` branch. Since `TYPE_CHECKING` is False at runtime, code under it should not execute at all. So, I am not sure if we should raise an error for it. What do you think?

jacobtylerwalls:
The music21 assert is just there to give a hint to mypy.

Pierre-Sassoulas:
I agree we should not raise if both are guarded in TYPE_CHECKING, but what about the following:
```python
import typing as t

if t.TYPE_CHECKING:
    from music 21 import Stream


class StreamExample(MusicStream): # Definitely used-before-assignment
    def my_func(self):
        if t.TYPE_CHECKING:
           assert isinstance(self, Stream) # No used-before-assignment
           assert isinstance(self, MusicStream) # used-before-assignment imo

```

tests/functional/u/used/used_before_assignment_typing.py
jacobtylerwalls: Thanks, sorry for the repeat question. At your option you could add a brief comment to this effect, but no worries.

jacobtylerwalls:
The mypy-hint false positive we're discussing is essentially the same as #8167, where I [mentioned](https://github.com/pylint-dev/pylint/issues/8167#issuecomment-1413784493) we'll eventually want to be smart enough to know it was _the very same test_ that was guarding both the declaration and the use of the variable.

If this PR is only driving a small amount of traffic into that existing false positive, we may not need to solve it right now on this PR.

jacobtylerwalls:
> The panda one also seem genuine as we return in the try then we raise in the final except, so fixturedef is always going to be defined:

Actually, I think the pandas example should (probably) warn. If `self._getnextfixturedef(argname)` divides by zero, none of the excepts will catch it, and fixturedef will blow up. It's reasonable for pandas not to code around it if unlikely, but pylint should be warning.

That said, we did choose in #5402 to assume that try blocks succeed, and only warn when definitions are made in an except. Someone could make the argument that's that's what's happening here, just with a nested try. Not that I would be all that convinced :-)

pylint/checkers/variables.py
jacobtylerwalls: Looks great! See below for reasons I think we can postpone discussing those edge cases.

I think the last test we should add here would be one for this very issue with `from __future__ import annotations`, then I'd be glad to merge!

jacobtylerwalls (dismissed):
I think it [would be good](https://github.com/pylint-dev/pylint/pull/8441#discussion_r1148447782) to add a test (in a separate file) for the failed case from the last review, but pending that, adding a green check!

jacobtylerwalls (approved):

tests/functional/u/used/used_before_assignment_postponed_evaluation.py
```diff
@@ -0,0 +1,13 @@
+"""Tests for used-before-assignment when postponed evaluation of annotations is enabled"""
+# pylint: disable=missing-function-docstring, invalid-name
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    var = 1
+    import math
```
Pierre-Sassoulas, line 8:
```suggestion
    import math
    from music21 import Stream
```

tests/functional/u/used/used_before_assignment_postponed_evaluation.py
```diff
@@ -0,0 +1,13 @@
...
+
+print(var)  # [used-before-assignment]
+
+def function_one(m: math):  # no error for annotations
+    return m
```
Pierre-Sassoulas, line 13:
```suggestion
    return m


class StreamExample(MusicStream):   # [used-before-assignment]
    def my_func(self):
        if TYPE_CHECKING:
           assert isinstance(self, Stream)
           assert isinstance(self, MusicStream)  # [used-before-assignment]
```

Pierre-Sassoulas (changes requested):
LGTM, let's just add the test case the new code is handling, it's easy to forget about it otherwise.

tests/functional/u/used/used_before_assignment_postponed_evaluation.py
jacobtylerwalls: I thought we were deferring this to #8167 per https://github.com/pylint-dev/pylint/pull/8441#issuecomment-1489614574?

Pierre-Sassoulas: Ho my bad did not read in detail :)

PR #8442, opened 2023-03-12
Increase new primer timeout to 120m

jacobtylerwalls:
## Type of Changes

## Description
The 3.7 primer job on main regularly takes about an hour and fails most of the time now, especially since we added `poetry-core`. Eventually, a better multiprocessing design will help. For now, bump the limit to the same as the workflow on PRs:

https://github.com/PyCQA/pylint/blob/b968fa062f801626426ef399401c4530e13fcc49/.github/workflows/primer_run_pr.yaml#L38


Pierre-Sassoulas (approved):
👍 Next step would be to parallelize each repo I guess.

PR #8420, opened 2023-03-10
Clear LRU caches on pylint utilities

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #8361 

Created a new backport label (should I have?)

PR #8448, opened 2023-03-13
[cyclic-import] Break cycle between pylint.checkers.util / variables

Pierre-Sassoulas:
## Type of Changes

## Description

Seen when adding ruff specific checks, it suddenly appeared following unrelated changes, I don't know why.

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -2277,13 +2277,13 @@ def overridden_method(
         return None
     try:
         meth_node = parent[name]
-    except KeyError:
+    except KeyError:  # pragma: no-cover
         # We have found an ancestor defining <name> but it's not in the local
         # dictionary. This may happen with astroid built from living objects.
         return None
     if isinstance(meth_node, nodes.FunctionDef):
         return meth_node
-    return None
+    return None  # pragma: no-cover
 
 
 def clear_lru_caches() -> None:
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -2266,11 +2266,28 @@ def not_condition_as_string(
     return msg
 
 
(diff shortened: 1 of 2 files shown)


jacobtylerwalls (approved):
Nice, I almost did this myself when doing #8420 but wanted it to be backportable. 👍

Pierre-Sassoulas (commented):
Sorry to reping you @jacobtylerwalls, I added pragmas for the coverage. Let me know if you have a simple method to cover this :D ((feel like something that should be in astroid now that I have to think how I would test it)

pylint/checkers/utils.py
```diff
@@ -2266,11 +2266,28 @@ def not_condition_as_string(
...
+    klass: nodes.LocalsDictNodeNG, name: str | None
+) -> nodes.FunctionDef | None:
+    """Get overridden method if any."""
+    try:
+        parent = next(klass.local_attr_ancestors(name))
+    except (StopIteration, KeyError):
+        return None
+    try:
+        meth_node = parent[name]
+    except KeyError:  # pragma: no-cover
```
Pierre-Sassoulas, line 2280:
```suggestion
    except KeyError:  #  pragma: no cover
```

pylint/checkers/utils.py
```diff
@@ -2266,11 +2266,28 @@ def not_condition_as_string(
...
+        # We have found an ancestor defining <name> but it's not in the local
+        # dictionary. This may happen with astroid built from living objects.
+        return None
+    if isinstance(meth_node, nodes.FunctionDef):
+        return meth_node
+    return None  # pragma: no-cover
```
Pierre-Sassoulas, line 2286:
```suggestion
    return None  #  pragma: no cover
```

PR #8469, opened 2023-03-20
Work in progress remove ``_format-option-value``

Pierre-Sassoulas:
## Type of Changes

## Description

Planned deprecation but I don't know what I wanted to do. I suppose it's because the parameter is a deprecated option dict from optparse ? But it seems there's no replacement right now.


pylint/utils/__init__.py
```diff
@@ -14,7 +14,6 @@
     HAS_ISORT_5,
     IsortDriver,
     _check_csv,
```
DanielNoord, line 16: I think we can remove it from this file and updated the docstring to be "Refactor the code to not use this old function".

PR #8470, opened 2023-03-20
[test multiprocessing] Add regression test against removing dill for unpickable plugins

Pierre-Sassoulas:
## Type of Changes

## Description

It was fixed previously I don't know when.


DanielNoord (changes requested):
This was fixed because of removing deprecated code.

Ideally we would keep this test as it is a good regression test against removing `dill` (or checking if we actually can).

DanielNoord (commented):
You didn't need to change anything and this worked again?

Pierre-Sassoulas:
Pretty much !

DanielNoord (approved):

PR #8475, opened 2023-03-21
[deprecation] Remove all duplicated typing guard check functions

Pierre-Sassoulas:
## Type of Changes

## Description

Enacting planned deprecations.

### DIFF
--- a/doc/whatsnew/fragments/8475.internal
+++ b/doc/whatsnew/fragments/8475.internal
@@ -1,0 +1,5 @@
+Following a deprecation period, ``is_typing_guard``, ``is_node_in_typing_guarded_import_block`` and
+``is_node_in_guarded_import_block``: from ``pylint.utils`` were removed: use a combination of
+``is_sys_guard`` and ``in_type_checking_block`` instead.
+
+Refs #8475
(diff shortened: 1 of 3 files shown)


doc/whatsnew/fragments/8475.internal
```diff
@@ -0,0 +1,5 @@
+Following a deprecation period, ``is_typing_guard``, ``is_node_in_typing_guarded_import_block`` and
+``is_node_in_guarded_import_block``: from ``pylint.utils`` were removed: use a combination of
+``is_sys_guard`` and ``in_type_checking_block`` instead.
```
Pierre-Sassoulas, line 3: We should probably create ``in_sys_guarded_block`` here, what do you think ?

DanielNoord: Do we have a need for this ourselves? If not I don't think so.

Pierre-Sassoulas: We're probably doing something complicated to get the englobing if, and then check that it's a sys guard, but the design of ``in_type_checking_block`` is superior imo.  Also it's breaking change time and it would make those two functions consistent together, but ultimately you're right.

Pierre-Sassoulas: >   ``is_sys_guard`` is still available, which was part of  ``is_node_in_guarded_import_block``.

So we do have the consistent equivalent.

PR #8474, opened 2023-03-21
[deprecation] Make 'OutputLine' require a fixed number of args

Pierre-Sassoulas:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :hammer: Refactoring   |

## Description

Enacting planned deprecation.

PR #8484, opened 2023-03-22
[testutil] Add proper test for the functional tests helpers

Pierre-Sassoulas:
## Type of Changes

## Description

Required for #8474, the output update was been broken a lot previously. We don't need the user warning because either we're going to fail with a message telling to update the output or we're already going to update the output.


pylint/testutils/output_line.py
```diff
@@ -139,13 +138,7 @@ def from_csv(
...
             raise IndexError
         except Exception:  # pylint: disable=broad-except
-            warnings.warn(
-                "Expected 'msg-symbolic-name:42:27:MyClass.my_function:The message:"
-                f"CONFIDENCE' but we got '{':'.join(row)}'. Try updating the expected"
-                f" output with:\npython tests/test_functional.py {UPDATE_OPTION}",
-                UserWarning,
-                stacklevel=2,
-            )
+            # We need this to not fail for the update script to work.
```
DanielNoord, line 141: Why don't we raise the warning anymore?

tests/testutils/test_lint_module_output_update.py
```diff
@@ -6,15 +6,20 @@
...
 import pytest
 
-from pylint.constants import PY38_PLUS
-from pylint.testutils import FunctionalTestFile
+from pylint.constants import IS_PYPY, PY38_PLUS, PY39_PLUS
+from pylint.testutils import FunctionalTestFile, LintModuleTest
 from pylint.testutils.functional import LintModuleOutputUpdate
 
+FIXTURE_DIRECTORY = Path(__file__).parent / "data/functional"
+DIRECTORIES = list(FIXTURE_DIRECTORY.iterdir())
```
DanielNoord, line 20: Let's not make this a constant. It increases the start up time for every test invocation.

tests/testutils/test_functional_testutils.py
```diff
@@ -41,8 +41,33 @@ def test_parsing_of_pylintrc_init_hook() -> None:
 
 def test_get_functional_test_files_from_directory() -> None:
     """Test that we correctly check the functional test directory structures."""
-    with pytest.raises(AssertionError, match="using_dir.py should not go in"):
-        get_functional_test_files_from_directory(DATA_DIRECTORY)
+    with pytest.raises(AssertionError) as exc_info:
```
DanielNoord, line 44: Why not use `match` in the `raises` already? Seems much easier.

DanielNoord (commented):
I left some early comments. Note that I find it quite hard to review, feels like a lot of refactoring and actual code changes are included here. Perhaps split the refactoring + fixes and the adding of tests in two separate PRs?

pylint/testutils/output_line.py
Pierre-Sassoulas: We don't need the user warning because either we're going to fail with a message telling to update the output or we're already going to update the output. See the 8 added test cases. The UserWarning actually prevented the update in case of broken output.

tests/testutils/test_lint_module_output_update.py
Pierre-Sassoulas: We then need to compute it twice to have the proper ids ?

tests/testutils/test_functional_testutils.py
```diff
@@ -41,8 +41,33 @@ def test_parsing_of_pylintrc_init_hook() -> None:
...
+        get_functional_test_files_from_directory(DATA_DIRECTORY / "u")
+    assert exc_info.match("'use_dir.py' should go in 'use'")
+    assert exc_info.match(
+        "using_dir.py should go in a directory that starts with the "
+        "first letters of 'using_dir'"
+    )
+    with pytest.raises(AssertionError):
+        exc_info.match("incredibly_bold_mischief.py")
```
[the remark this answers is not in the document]
Pierre-Sassoulas: It's still the ``exc_info`` from l. 44

tests/testutils/test_lint_module_output_update.py
```diff
@@ -6,15 +6,20 @@
...
 
 import pytest
 
-from pylint.constants import PY38_PLUS
-from pylint.testutils import FunctionalTestFile
+from pylint.constants import IS_PYPY, PY38_PLUS, PY39_PLUS
+from pylint.testutils import FunctionalTestFile, LintModuleTest
 from pylint.testutils.functional import LintModuleOutputUpdate
 
+FIXTURE_DIRECTORY = Path(__file__).parent / "data/functional"
```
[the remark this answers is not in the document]
Pierre-Sassoulas: How about that ?
```suggestion
FUNCTIONAL_TEST_HELPERS_FIXTURES_PATH = Path(__file__).parent / "data/functional"
```

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -50,27 +51,38 @@ def get_functional_test_files_from_directory(
...
+                yield _file_or_dir.resolve()
+
     # Collect all sub-directories and files in directory
-    for file_or_dir in directory.iterdir():
-        if file_or_dir.is_file():
-            if file_or_dir.suffix == ".py" and not file_or_dir.stem.startswith("_"):
-                files.add(file_or_dir)
-        elif file_or_dir.is_dir():
+    for file_or_dir in walk(directory):
+        if file_or_dir.stem.startswith("_"):
```
DanielNoord, line 98: We should probably do the check for directory length after excluding ignored files

tests/testutils/test_functional_testutils.py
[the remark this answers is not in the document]
DanielNoord: Then I don't understand what this is testing.. Could you explain? 😄

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -62,17 +62,21 @@ def _check_functional_tests_structure(directory: Path) -> None:
...
-                    f"{_file_or_dir} contains too many functional tests files "
-                    + f"({len(_files)} > {REASONABLY_DISPLAYABLE_VERTICALLY})."
-                )
+                if len(_files) > REASONABLY_DISPLAYABLE_VERTICALLY:
+                    violations.append((_file_or_dir, len(_files)))
                 yield _file_or_dir
                 yield from walk(_file_or_dir)
             else:
                 yield _file_or_dir.resolve()
+        if violations:
```
DanielNoord, line 88: I said this before, why not just assert and do this at the end? This makes the code a lot more complex and harder to read.

tests/testutils/test_functional_testutils.py
```diff
@@ -54,6 +54,22 @@ def test_get_functional_test_files_from_directory() -> None:
...
 
+def test_get_functional_test_files_from_crowded_directory() -> None:
+    """Test that we correctly check the functional test directory structures."""
+    with pytest.raises(AssertionError) as exc_info:
+        get_functional_test_files_from_directory(
+            DATA_DIRECTORY / "m", max_file_per_directory=1
+        )
+    assert exc_info.match("m: 4 when the max is 1")
+    assert exc_info.match("max_overflow: 3 when the max is 1")
+    with pytest.raises(AssertionError) as exc_info:
```
DanielNoord, line 65: Shouldn't this test be with 3 so that the output is different?

DanielNoord (commented):
Thanks @Pierre-Sassoulas! Wonderful that all of this code is finally getting tested.

I reviewed this commit by commit as just final diff is still quite large. I think it is probably best to go through them commit by commit as well as the comments make more sense then!

Again, love this! Makes it much easier to use this for plugins as well!

tests/testutils/test_lint_module_output_update.py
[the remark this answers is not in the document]
DanielNoord: See other comment about whether this is actually necessary

tests/testutils/test_lint_module_output_update.py
Pierre-Sassoulas: It would use the full path that is very long though.

tests/testutils/test_functional_testutils.py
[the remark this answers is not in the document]
Pierre-Sassoulas: It tests that we do **not** match ``"incredibly_bold_mischief.py`` in exc_info

tests/testutils/test_functional_testutils.py
```diff
@@ -41,7 +41,11 @@ def test_parsing_of_pylintrc_init_hook() -> None:
 
 def test_get_functional_test_files_from_directory() -> None:
     """Test that we correctly check the functional test directory structures."""
-    with pytest.raises(AssertionError, match="using_dir.py should not go in"):
+    match = (
+        "using_dir.py should go in a directory that starts with the "
+        "first letters of 'using_dir'"
+    )
+    with pytest.raises(AssertionError, match=match):
```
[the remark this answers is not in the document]
Pierre-Sassoulas: What do you mean by "full name message" ?

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -71,17 +71,27 @@ def _check_functional_tests_structure(directory: Path) -> None:
         f"{directory} contains too many functional tests files "
         + f"({len(files)} > {REASONABLY_DISPLAYABLE_VERTICALLY})."
     )
-
+    directory_does_not_exists: list[tuple[Path, Path]] = []
```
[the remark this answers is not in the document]
Pierre-Sassoulas: Raising at the first violation make the message incomplete (you have to relaunch n times to see all the violations, it's annoying) and it also make the message being raised platform dependent depending on the order of iterdir (thus hard to test).

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: It makes the error friendlier for users and easier to tests. We have to do it at the end if we want to cover everything, or the first violations prevent from testing the other use case.

tests/testutils/test_functional_testutils.py
Pierre-Sassoulas: It tests with ``2`` so that the output is different 😄

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: [5bee0d6](https://github.com/PyCQA/pylint/pull/8484/commits/5bee0d67db163a463d861181820a3759fb32e070)

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -50,27 +51,38 @@ def get_functional_test_files_from_directory(
...
 
+    def walk(path: Path) -> Iterator[Path]:
+        for _file_or_dir in path.iterdir():
+            if _file_or_dir.is_dir():
+                _files = list(_file_or_dir.iterdir())
+                assert len(_files) <= REASONABLY_DISPLAYABLE_VERTICALLY, (
+                    f"{_file_or_dir} contains too many functional tests files "
+                    + f"({len(_files)} > {REASONABLY_DISPLAYABLE_VERTICALLY})."
+                )
+                yield _file_or_dir
```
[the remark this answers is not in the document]
Pierre-Sassoulas: [00779f1](https://github.com/PyCQA/pylint/pull/8484/commits/00779f1c5e3c378fe98e4236ff3d40a9754717e4)

tests/testutils/test_lint_module_output_update.py
```diff
@@ -76,4 +81,65 @@ def test_lint_module_output_update_remove_useless_txt(
...
+    for filename in new_path.iterdir():
+        if filename.suffix != ".py":
+            continue
+        new_path_str = str(new_path)
+        ftf = FunctionalTestFile(directory=new_path_str, filename=filename.name)
+        # Standard functional test helper
+        lint_module = LintModuleTest(ftf)
+        # Functional test helper that automatically update the output
+        lint_module_output_update = LintModuleOutputUpdate(ftf)
+        if "wrong_test" in new_path_str:
```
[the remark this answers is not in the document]
Pierre-Sassoulas: [3e4ea4a](https://github.com/PyCQA/pylint/pull/8484/commits/3e4ea4ad5a1c495a6aa684ea2a7733b12cfd7ecc)

tests/testutils/test_lint_module_output_update.py
DanielNoord: The ids are only necessary for `pytest -k ID` right? Then a long name isn't that much of an issue.

tests/testutils/test_functional_testutils.py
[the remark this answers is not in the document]
DanielNoord: `"using_dir.py should go in a directory that starts with the "
        "first letters of 'using_dir'"`

We could just do `match="should go in a directory"`. We don't really care about the specifics here I think.

tests/testutils/test_lint_module_output_update.py
```diff
@@ -76,4 +81,65 @@ def test_lint_module_output_update_remove_useless_txt(
...
-    assert not (expected_output_file).exists()
+    assert not expected_output_file.exists()
+
+
+@pytest.mark.skipif(
+    not PY38_PLUS or (IS_PYPY and not PY39_PLUS),
+    reason="Requires accurate 'end_col' value to update output",
+)
+@pytest.mark.parametrize("directory_path", DIRECTORIES, ids=DIRECTORIES_NAMES)
+def test_update_of_functional_output(directory_path: Path, tmp_path: Path) -> None:
```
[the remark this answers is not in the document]
DanielNoord: Yeah I understand that but I would have preferred a solution where we made the case by hand in the test itself something like:
```python
@dataclass
class Case:
    test_is_correct: bool
    output_is_correct: bool
```

Then the `Case` could get the correct file from the right template. Right now the definition of the test case and the test runner is very decoupled which makes it harder to read imo. Anyway, let's leave this for now. Perhaps something to consider in the future.

pylint/testutils/functional/find_functional_tests.py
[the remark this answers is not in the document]
DanielNoord: Hmm okay. I don't really mind running multiple times (bad input > bad output, as with `pydocstringformatter`) but testability is nice I guess

tests/testutils/test_functional_testutils.py
DanielNoord: Let's test with 3 instead so we are sure that we are not actually raising a message each time but actually check the number as well.

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -49,39 +51,89 @@ def get_functional_test_files_from_directory(
...
+        )
+        if len(files_without_leading_underscore) > max_file_per_directory:
+            violations.append((path, len(files_without_leading_underscore)))
+        return files_without_leading_underscore
+
+    def walk(path: Path) -> Iterator[Path]:
+        violations: list[tuple[Path, int]] = []
+        violations_msgs: set[str] = set()
+        parent_dir_files = _get_files_from_dir(path, violations)
+        error_msg = (
```
DanielNoord, line 82: Let's just assign this on L98. I don't think there is a need to do so here?

tests/testutils/test_functional_testutils.py
Pierre-Sassoulas: I think we're doing what you suggest only with 2 instead of 3:
```python
    with pytest.raises(AssertionError) as exc_info:
        get_functional_test_files_from_directory(
            DATA_DIRECTORY / "m", max_file_per_directory=1 #1
        )
    assert exc_info.match("m: 4 when the max is 1")  #1
    assert exc_info.match("max_overflow: 3 when the max is 1")  #1
        get_functional_test_files_from_directory(
            DATA_DIRECTORY / "m", max_file_per_directory=2   #2
        )
    assert exc_info.match("m: 4 when the max is 2")    #2
    assert exc_info.match("max_overflow: 3 when the max is 2")    #2

File: pylint/testutils/functional/find_functional_tests.py [hunk 15]
```diff
@@ -49,39 +51,89 @@ def get_functional_test_files_from_directory(
...
+            "The following directory contains too many functional tests files:\n"
+        )
+        for _file_or_dir in parent_dir_files:
+            if _file_or_dir.is_dir():
+                _files = _get_files_from_dir(_file_or_dir, violations)
+                yield _file_or_dir.resolve()
+                try:
+                    yield from walk(_file_or_dir)
+                except AssertionError as e:
+                    violations_msgs.add(str(e).replace(error_msg, ""))
```
DanielNoord, line 92: If I am reading this correctly `violations` can be empty while `violations_msgs` isn't, which breaks the logic on L95

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -49,39 +51,89 @@ def get_functional_test_files_from_directory(
...
+            raise AssertionError(_msg)
+
     # Collect all sub-directories and files in directory
-    for file_or_dir in directory.iterdir():
-        if file_or_dir.is_file():
-            if file_or_dir.suffix == ".py" and not file_or_dir.stem.startswith("_"):
-                files.add(file_or_dir)
-        elif file_or_dir.is_dir():
+    for file_or_dir in walk(directory):
+        if file_or_dir.stem.startswith("_"):
```
DanielNoord, line 105: Is this still necessary?

pylint/testutils/functional/find_functional_tests.py
```diff
@@ -49,39 +51,89 @@ def get_functional_test_files_from_directory(
...
             not len(file.parent.stem) == 1  # First letter sub-directories
             and file.parent.stem not in IGNORED_PARENT_DIRS
             and file.parent.parent.stem not in IGNORED_PARENT_PARENT_DIRS
         ):
-            assert file.stem.startswith(
-                file.parent.stem
-            ), f"{file} should not go in {file.parent}"
+            if not file.stem.startswith(file.parent.stem):
+                misplaced_file.append(file)
+    if directory_does_not_exists or misplaced_file:
```
DanielNoord, line 126: Please add some new lines in between 103-139. These blocks seem to served different purposes (like the comments indicate) so imo they should be separated

tests/testutils/test_functional_testutils.py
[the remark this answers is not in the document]
DanielNoord: Yeah but why are we doing that on the old `exc_info`? This feels like an anti-pattern: why not write one `regex` that includes the word we want to match "should go in" and does not match whenever `incredibly_bold_mischief.py` is included?

tests/testutils/test_functional_testutils.py
DanielNoord: Yeah but are we sure we don't raise `"max_overflow: 3 when the max is 3"`? We are only testing the emitting and `f-stringing` here, not the actual comparison which is the main use of the function we are testing.

DanielNoord (commented):
Sorry for taking so long to review this! Hard to find the time to get back into this. Think we are getting there though! Love the idea of all of this being actually tested 😄

tests/testutils/test_functional_testutils.py
Pierre-Sassoulas: I finally understood.

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: Feel like a nitpick, we have a lot to do in pylint, and this is blocking 3.0.0a6, moving assignments around in functional test's tests is not an efficient use of our time.

tests/testutils/test_functional_testutils.py
Pierre-Sassoulas: a955ed8ef

tests/testutils/test_functional_testutils.py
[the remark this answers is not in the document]
Pierre-Sassoulas: a955ed8ef

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: 2430c6a5e

tests/testutils/test_lint_module_output_update.py
Pierre-Sassoulas: 0ac141cd8

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: Nice catch fixed in aa87ebb14

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: We need it line 93 in ``violations_msgs.add(str(e).replace(error_msg, ""))``

DanielNoord: That I completely missed! Resolving!

DanielNoord: I understand that you feel this way but I must say that only through that scrutiny we found the logic flaw in `violations` that would have actually added a regression in this PR that is supposed to only add coverage.

There is a fine balance to strike between scrutiny and just merging because we want to continue and perhaps in this case I was a bit too strict, but I do think we should be somewhat strict about our code. Stuff like the adding of default values to `__init__` and `postinit` in `astroid` and some of the non-understandable tests we have there could all have been avoided by carefully considering the PRs and the code in them.

As you know I do really appreciate the effort that went into this PR and I know this was hard to refactor and add tests for. I have only one open comment and then we can go ahead and merge this 🚀

DanielNoord (dismissed):

DanielNoord:
Thanks again @Pierre-Sassoulas for all the work that went into this!

Pierre-Sassoulas:
I'm going to cleanup and rebase :)

DanielNoord:
> I'm going to cleanup and rebase :)

I'll be right here to approve and let auto-merge do its magic 😄

DanielNoord (approved):

pylint/testutils/functional/find_functional_tests.py
Pierre-Sassoulas: It's true that reviewing made us realize a problem but even if we merged with the violations msg issue, it's an error in a helper that enforce that there's not too much files in our functional test directory. What I mean is that we could live with that kind of issues 😄 Especially when we're living just fine with the multiprocessing that give very wrong results when activated (without even being that much faster), or #3512 not implemented 3 years after being opened, the new json reporter either, astroid not typed, or even typable, etc. etc. you know as well as me that some issues are pretty important and pretty much only doable by a maintainer because they are so hard to design or start on. I think that sometime when the cost of a mistake is low we should *just* merge because the changes makes the code better even if it's not perfect and because we're going to loose contributors if they dread our reviews.

Issue #8509, opened 2023-03-29
aiohttp.ClientSession() crashes pylint


ppickfor:
### Bug description

import aiohttp
s = aiohttp.ClientSession() 
[pylint-crash-2023-03-28-20-44-10.txt](https://github.com/PyCQA/pylint/files/11096417/pylint-crash-2023-03-28-20-44-10.txt)

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
see attached
```

### Expected behavior

not to crash

### Pylint version

```shell
pylint 2.17.1
astroid 2.15.1
Python 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)]
```

### OS / Environment

Windows

### Additional dependencies

aiodns==3.0.0
aiohttp==3.8.4
aiosignal==1.3.1
astroid==2.15.1
async-timeout==4.0.2
attrs==22.2.0
autopep8==2.0.2
certifi==2022.12.7
cffi==1.15.1
charset-normalizer==3.1.0
colorama==0.4.6
dill==0.3.6
frozenlist==1.3.3
idna==3.4
isort==5.12.0
lazy-object-proxy==1.9.0
mccabe==0.7.0
multidict==6.0.4
mypy==1.1.1
mypy-extensions==1.0.0
platformdirs==3.2.0
pycares==4.3.0
pycodestyle==2.10.0
pycparser==2.21
pylint==2.17.1
requests==2.28.2
tomlkit==0.11.7
types-requests==2.28.11.17
types-urllib3==1.26.25.10
typing_extensions==4.5.0
urllib3==1.26.15
wrapt==1.15.0
yarl==1.8.2

DanielNoord:
Please copy the content of the stack trace into the issue description. Maintainers are not likely to download random files from the internet because of the apparent security risks :smile:

Pierre-Sassoulas:
Thank you for opening the issue :) Don't worry, I included the stacktrace in the issue.

DanielNoord:
@jacobtylerwalls Are you able to reproduce this?

I can't:

```python
❯ cat test.py
import aiohttp

s = aiohttp.ClientSession()
❯ pylint --version
pylint 3.0.0b1
astroid 2.16.0dev0
Python 3.11.0b3 (main, Jun 12 2022, 16:13:11) [Clang 13.0.0 (clang-1300.0.29.30)]
❯ pylint test.py

[...]

PR #8510, opened 2023-03-29
[crash template] Upgraded crash template for easier issue creation

Pierre-Sassoulas:
## Type of Changes

## Description

Refs #8509 where I had to do quite a lot to make the crash txt displayable properly.


pylint/lint/utils.py
```diff
@@ -46,11 +47,27 @@ def prepare_crash_report(ex: Exception, filepath: str, crash_file_path: str) ->
...
 ```
+
+### Pylint output
+
+<details open>
+    <summary>
+        pylint crashed with a ``{ex.__class__.__name__}`` and with the following stacktrace:
+    </summary>
+
+```python
```
DanielNoord, line 63: A code block can't be in ``details``. I'd prefer the codeblock over the minimizing.

Pierre-Sassoulas: Are you sure ? It works in #8509 (with the proper amount of line break)

Pierre-Sassoulas:
Drafting because we can link to a blank issue and provide most of the other infos

pylint/lint/utils.py
DanielNoord: TIL! This was broken for years I think

Pierre-Sassoulas: Ho it was broken 5mn ago too, I edited the thing 5 times 😄 Adding line breaks helps it seems.

DanielNoord (approved):

PR #8520, opened 2023-03-31
Add Pyreverse option to exclude standalone nodes

### LINKED ISSUE #8476: Pyreverse: Add option to ignore classes without relationships
### Current problem

Class diagrams are great for understanding the relationships between classes. But the class diagrams produced by Pyreverse include all classes, including those that have no relationships. Those objects clutter the diagrams.

In this example, there are seven unconnected nodes:

classes

### Desired solution

It would be nice to have an option to exclude unconnected classes from the class diagram. With the option enabled, Pyreverse would only display classes that have some kind of relationship to other classes.

### Additional context

Maybe the package diagram too?

nickdrozd:
## Type of Changes

## Description

Closes #8476

### DIFF
--- a/tests/pyreverse/test_writer.py
+++ b/tests/pyreverse/test_writer.py
@@ -149,6 +149,7 @@ def _setup(
     for fname in (
         DOT_FILES
         + COLORIZED_DOT_FILES
+        + NO_STANDALONE_FILES
         + PUML_FILES
         + COLORIZED_PUML_FILES
         + MMD_FILES
--- a/tests/pyreverse/test_writer.py
+++ b/tests/pyreverse/test_writer.py
@@ -35,6 +35,7 @@ _DEFAULTS = {
     "show_stdlib": False,
     "only_classnames": False,
     "output_directory": "",
+    "no_standalone": False,
 }
 
 TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "..", "data")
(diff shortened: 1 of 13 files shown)


nickdrozd:
From manual test, this seems to work fine. I have no idea how the Pyreverse tests work.

DudeNr33:
> From manual test, this seems to work fine. I have no idea how the Pyreverse tests work.

For class diagrams you can use the functional tests similar to how the regular `test_functional.py` works for messages:

* choose or create a directory under `tests/pyreverse/functional/class_diagrams`
* Create a Python file with sample code, e. g. `no_standalone.py`
* In your case which requires special command line options: create a `no_standalone.rc`, and add `command_line_args=<your args>`
* Create a corresponding expected output file  (default format is MermaidJS) with the same name, i.e. `no_standalone.mmd`

Unfortunately we don't have a similar concept yet for package diagrams. You can take a look at `test_writer.py`, which creates class and package diagrams from the files located under `tests/data`, and add a suitable fixture for a config including the new option there. As far as I see there should already be suitable classes and "packages" (modules also count as packages here) present that should be omitted with `--no-standalone`.

nickdrozd:
@DudeNr33 I keep getting a `FileNotFoundError` in the package test. Any ideas? There must be some obvious mistake somewhere.

tests/pyreverse/test_writer.py
```diff
@@ -87,6 +89,15 @@ def setup_colorized_dot(
     yield from _setup(project, colorized_dot_config, writer)
 
 
+@pytest.fixture()
+def setup_no_standalone_dot(
+    no_standalone_dot_config: PyreverseConfig, get_project: GetProjectCallable
+) -> Iterator[None]:
+    writer = DiagramWriter(no_standalone_dot_config)
+    project = get_project(TEST_DATA_DIR)
```
DudeNr33, line 97: You have to specify `name` here like above in `setup_colorized_dot`. Also you have to rename the files from `no_standalone_classes.dot` to `classes_no_standalone.dot` (same for package diagram).

DudeNr33 (approved):
Looks good to me, thank you for the addition! 👍

PR #8534, opened 2023-04-03
Merge maintenance 2.17.x in main following 2.17.2 release

Pierre-Sassoulas:


DanielNoord (commented):
Shouldn't this also remove some news fragments?

Pierre-Sassoulas:
We're voluntarily keeping them for the next minor release.

DanielNoord:
So the entry shows up in `2.18.0` and `2.17.2`?

Pierre-Sassoulas:
Yeah I had the same question than you for 2.16 actually but I don't remember exactly where it was discussed. (The fact that it make merge easier is convenient).

DanielNoord:
Hmm this does feel weird, but okay I guess?

DanielNoord (approved):

PR #8553, opened 2023-04-08
Enhance checker "Dangerous default value" detecting callables

### LINKED ISSUE #4659: Dangerous default argument should include "Stateful" default arguments as well

A clear and concise description of what the problem is.

When a function is defined, a function will maintain the state of the default argument. This can be dangerous if the default argument is something like `datetime.date.today()`. In this [article](https://docs.python-guide.org/writing/gotchas/), it states that:

> Python’s default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, Ruby).

Thus, if a production environment has a pod that evaluates a function once and the pod doesn't rotate for a couple days, it will continue to define that default argument incorrectly. 

It would be nice if `pylint` warns us about these dangerous default "stateful" arguments, similar to the warning about dangerous "mutable" default arguments, e.g. http://pylint-messages.wikidot.com/messages:w0102

### Additional context

Add any other context about the feature request here.

alblasco:
## Type of Changes

## Description

Refs #4659 
Closes #4659

Notes: 
* This is my first contribution to pylint, and not familiar with AST so please any feedback is welcome
* My approach is something KISS. 
  - Detect any callable 
  - excluding a small list of builtins immutables as frozenset

### DIFF
--- a/doc/whatsnew/fragments/4659.feature
+++ b/doc/whatsnew/fragments/4659.feature
@@ -1,0 +1,3 @@
+Dangerous-default-value warns also about callables: functions, methods, lambdas, ...
+
+Closes #4659
--- a/pylint/checkers/base/basic_checker.py
+++ b/pylint/checkers/base/basic_checker.py
@@ -61,6 +61,16 @@ DEFAULT_ARGUMENT_SYMBOLS = dict(
         )
     },
 )
+DEFAULT_ARGUMENT_CALLABLE_IMMUTABLE_BUILTINS = [
+    "frozenset",
+    "bytes",
+    "tuple",
+    "str",
+    "int",
+    "float",
+    "complex",
+    "range",
+]
 
 
 def report_by_type_stats(
@@ -605,6 +615,16 @@ class BasicChecker(_BasicChecker):
         def is_iterable(internal_node: nodes.NodeNG) -> bool:
             return isinstance(internal_node, (nodes.List, nodes.Set, nodes.Dict))
 
+        def is_callable_immutable_builtins(
+            default_as_string: str, value: nodes.NodeNG
+        ) -> bool:
+            a = (
+                default_as_string in DEFAULT_ARGUMENT_CALLABLE_IMMUTABLE_BUILTINS
+                and isinstance(value, astroid.Instance)
+                and value.qname() == f"builtins.{default_as_string}"
+            )
+            return a
+
         defaults = (node.args.defaults or []) + (node.args.kw_defaults or [])
         for default in defaults:
             if not default:
@@ -638,6 +658,13 @@ class BasicChecker(_BasicChecker):
                     # this argument is a name
                     msg = f"{default.as_string()} ({DEFAULT_ARGUMENT_SYMBOLS[value.qname()]})"
                 self.add_message("dangerous-default-value", node=node, args=(msg,))
+
+            elif isinstance(default, astroid.nodes.node_classes.Call):
+                if hasattr(default, "func") and not is_callable_immutable_builtins(
+                    default.func.as_string(), value
+                ):
+                    msg = f"{default.as_string()} Callable"
+                    self.add_message("dangerous-default-value", node=node, args=(msg,))
 
     @utils.only_required_for_messages("unreachable", "lost-exception")
     def visit_return(self, node: nodes.Return) -> None:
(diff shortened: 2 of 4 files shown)


jacobtylerwalls:
Thanks for jumping on this. Just a quick note to say I'm not sure about adding this into `dangerous-default-value`. I personally wouldn't want to use this check, but I'd want to keep `dangerous-default-value`. There's nothing dangerous about creating a constant with a callable. It's dangerous if you're creating an instance and manipulating that instance, or if the constant is time-sensitive, but the check is going to flag all kinds of safe cases, like the primer results show.

Pierre-Sassoulas:
I agree with Jacob here, I think the check needs a list of problematic calls (optimally a configurable list) instead of a list of non problematic calls. The primer's result shows that too imo.

alblasco:
Thanks for the feedback:
> Just a quick note to say I'm not sure about adding this into dangerous-default-value. I personally wouldn't want to use this check, but I'd want to keep dangerous-default-value. There's nothing dangerous about creating a constant with a callable. It's dangerous if you're creating an instance and manipulating that instance, or if the constant is time-sensitive, but the check is going to flag all kinds of safe cases, like the primer results show.

Agreed. Using a new different check isolates both results (dangerous-default-value is a basic check working since a long time).
And we can also disable the new check by default. 

> I agree with Jacob here, I think the check needs a list of problematic calls (optimally a configurable list) instead of a list of non problematic calls. The primer's result shows that too imo.

Not so sure. In theory any callable is potentially dangerous, so a configurable list seems not enough imo. Even if it's a callable returning a constant, you have to assure that this callable is not calling to any external method to retrieve this constant or that the cyclomatic complexity of that callable is not more than one (i.e. not ifs, for, ... or any loop or conditional programming is used). 

I just implemented a kiss approach (every callable is potentially unsafe, except some immutable cases). With previous conditions most callables will be identified as potentially unsafe. So probably too much work for so little benefit.

Any way I am willing to keep contributing. But for any other approach than kiss, I would need additional help about ast, to learn how to dig inside the callable to inspect it. E.g: a commit example in my fork would be very much appreciated.

Pierre-Sassoulas (commented):
Reading the primer in more details, I'm not entirely sure all of those numerous new messages are false positives... but I think there are indeed false positives between them, because if the [left padding in django](https://github.com/django/django/blob/2eb1f37260f0e0b71ef3a77eb5522d2bb68d6489/django/db/models/functions/text.py#L193) had a real problem with consequences, it would have been fixed a long time ago. Let me know what you think :) 

The pylint team favor false negative over false positives (see [our fundamental tenets](https://pylint.readthedocs.io/en/latest/development_guide/contributor_guide/governance.html#what-are-the-fundamental-tenets-of-pylint-development)). I think the proposed check would make sense as an external plugin, but even if we create a new message (Maybe ``call-as-default-value`` or  ``dangerous-call-as-default-value`` ?), we should avoid raising when what is returned from the call is an immutable that still make sense after 10k calls or a year of uptime. Checking that a call return something safe is a hard problem to solve generically and in a reasonable time (both dev and runtime) which is why I was proposing a set of known dangerous calls where a problem can be expected (like the example given in #4659 with "today").

jacobtylerwalls:
> which is why I was proposing a set of known dangerous calls where a problem can be expected (like the example given in https://github.com/pylint-dev/pylint/issues/4659 with "today").

Agree that to move this forward we should only check for known dangerous calls that are time-sensitive or expensive to compute or misleading (e.g. random.random(), which will be anything but random).

jacobtylerwalls:
Re: the case from #8826, we could detect if the callable is a class (and not a frozen dataclass). That's easy to distinguish from a method that returns a constant.

jacobtylerwalls:
Don't forget to update the `dangerous-default-value` message definition to avoid referring to "dicts and lists" only.

alblasco:
FYI I'm done
What I was looking for is already available at https://github.com/PyCQA/flake8-bugbear
> B008: Do not perform function calls in argument defaults. The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call. If this is intended, assign the function call to a module-level variable and use that variable as a default value.

DanielNoord:
Closing as the original author indicated they won't work on this.

PR #8556, opened 2023-04-08
Fix false positive for ``positional-only-arguments-expected``

### LINKED ISSUE #8555: False positive for `positional-only-arguments-expected`
### Bug description

a.py:
```python
def function_with_kwargs(apple, banana="Yellow banana", /, **kwargs):
    """
    Calling this function with the `banana` keyword should not emit
    `positional-only-arguments-expected` since it is added to `**kwargs`.

    >>> function_with_kwargs("Red apple", banana="Green banana")
    >>> "Red apple"
    >>> "Yellow banana"
    >>> {"banana": "Green banana"}
    """
    print(apple)
    print(banana)
    print(kwargs)
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
a.py:15:0: E3102: `function_with_kwargs()` got some positional-only arguments passed as keyword arguments: 'banana' (positional-only-arguments-expected)
```

### Expected behavior

No error emitted in this case as the argument passed to the function is part of `**kwargs`.
The purpose of this checker is to indicate when an error would occur at runtime, for example:
`TypeError: function_with_kwargs() got some positional-only arguments passed as keyword arguments: 'banana'`.

However, in this case, there is no error running this Python code.

### Pylint version

```shell
pylint 3.0.0b1
astroid 2.16.0dev0
Python 3.10.4
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

mbyrnepr2:
## Type of Changes

## Description

Fix false positive for ``positional-only-arguments-expected`` when a
function contains both a positional-only parameter that has a default value, and ``**kwargs``.

Closes #8555

### DIFF
--- a/doc/whatsnew/fragments/8555.false_positive
+++ b/doc/whatsnew/fragments/8555.false_positive
@@ -1,0 +1,3 @@
+Fix false positive for ``positional-only-arguments-expected`` when a function contains both a positional-only parameter that has a default value, and ``**kwargs``.
+
+Closes #8555
--- a/pylint/checkers/method_args.py
+++ b/pylint/checkers/method_args.py
@@ -111,6 +111,8 @@ class MethodArgsChecker(BaseChecker):
             and inferred_func.args.posonlyargs
         ):
             return
+        if inferred_func.args.kwarg:
+            return
         pos_args = [a.name for a in inferred_func.args.posonlyargs]
         kws = [k.arg for k in node.keywords if k.arg in pos_args]
         if not kws:
--- a/tests/functional/p/positional_only_arguments_expected.py
+++ b/tests/functional/p/positional_only_arguments_expected.py
@@ -16,3 +16,21 @@ cake.nihon(1, 2, i=3)  # [positional-only-arguments-expected]
 cake.nihon(1, r=2, i=3)  # [positional-only-arguments-expected]
 cake.nihon(a=1, r=2, i=3)  # [positional-only-arguments-expected]
 cake.nihon(1, r=2, i=3, cheese=True)  # [positional-only-arguments-expected]
+
+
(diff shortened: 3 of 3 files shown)


jacobtylerwalls (changes requested):
Nice edge case! I find this profoundly weird, and honestly, after reading PEP 570 a couple times, I feel like this could even be reported to CPython as a bug. But even if it's not a bug, it should be a pylint warning on the function def, not a pylint error on the call.

However, I did find this regression, based on the example in PEP 570 ("Semantic Corner Case"). Would you be able to fix and add a test case?

```python
def foo(name, /, **kwds):
    return 'name' in kwds

foo(name=3)
```
Gives:
```python
TypeError: foo() missing 1 required positional argument: 'name'
```

`main` emits:
```
corner.py:4:0: E3102: `foo()` got some positional-only arguments passed as keyword arguments: 'name' (positional-only-arguments-expected)
```

PR emits nothing.

jacobtylerwalls:
Opened #8558 to track emitting a different message for the original reported case.

mbyrnepr2:
> However, I did find this regression, based on the example in PEP 570 ("Semantic Corner Case"). Would you be able to fix and add a test case?

I don't think this is a regression. The TypeError you've identified is `missing 1 required positional argument` whereas this checker deals with the TypeError `got some positional-only arguments passed as keyword arguments`.
This is similar behaviour to the current fix in the sense that `name=3` is collected by `**kwds` and hence the `name` argument is indeed missing. Here would be a working example:
```python
def foo(name, /, **kwds):
    print(kwds)
    return 'name' in kwds

>>> foo(1, name=3)
>>> {'name': 3}
```

jacobtylerwalls (approved):
Ah, I see, makes sense.

PR #8566, opened 2023-04-12
Fix `unused-import` to check`dummy-variables-rgx`

### LINKED ISSUE #8500: `unused-import` should observe `dummy-variables-rgx` setting
### Bug description

Python convention is that any variable name consisting entirely of underscores is to be considered a placeholder, used when a value needs to be assigned somewhere but you don't actually care about it after that. Thus no warnings should be emitted about such a variable being unused, overwritten, or changing type.

However, another common convention is to import the `gettext()` translation function as `_` for convenience. Thus current best-practice is to use names of two or more underscores for placeholders.

`test2.py`:
```python
"""Test"""

import prettyprinter as __
```

### Configuration

_No response_

### Command used

```shell
pylint test2.py
```

### Pylint output

```shell
************* Module test2
test2.py:3:0: W0611: Unused prettyprinter imported as __ (unused-import)

```

### Expected behavior

No `unused-import` should be emitted if the module is imported with a name consisting entirely of underscores.

### Pylint version

```shell
pylint 2.16.2
astroid 2.14.2
```

### OS / Environment

Win10

### Additional dependencies

_No response_

RSTdefg:
## Type of Changes

## Description

Fixed `unused-import` so that it observes the `dummy-variables-rgx` option. Also added one testcase to test this behavior.

Closes #8500


Pierre-Sassoulas (changes requested):
Pretty good first contribution, thank you ! This also fixed a false positive in pylint which is why the pylint check is failing :)

RSTdefg:
> Pretty good first contribution, thank you ! This also fixed a false positive in pylint which is why the pylint check is failing :)

That's what I figured as well!

RSTdefg (commented):
Reviewed

Pierre-Sassoulas (approved):
👌

PR #8571, opened 2023-04-12
Fix false positive for ``keyword-arg-before-vararg``

### LINKED ISSUE #8570: False positive for `keyword-arg-before-vararg` with `positional-only` parameter
### Bug description

`keyword-arg-before-vararg` is emitted by Pylint for the following `a.py` where we have a positional-only parameter with a default value followed by `*args`:
```python
def name(param1=True, /, *args): ...
```

This is a false positive because the `keyword-arg-before-vararg` checker exists to warn us about function signatures that can lead to `TypeError: name() got multiple values for argument "param1"` when this function would be called with the intention of passing an arbitrary number of *args together with a keyword-argument for `param1`.
However, in this case, we have the following, since it isn't possible to pass a positional-only parameter as a keyword anyway:
```python
  File "/Users/markbyrne/programming/a.py", line 3, in <module>
    name1(1, 2, 3, 4, 5, param1=False)
TypeError: name1() got some positional-only arguments passed as keyword arguments: 'param1'
```

I propose suppressing the warning in this situation.

For further context on `keyword-arg-before-vararg` see [comment](https://github.com/pylint-dev/pylint/issues/2481#issuecomment-419700017).

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
a.py:1:0: W1113: Keyword argument before variable positional arguments list in the definition of name1 function (keyword-arg-before-vararg)
```

### Expected behavior

No warning is emitted.

### Pylint version

```shell
pylint 3.0.0b1
astroid 2.16.0dev0
Python 3.10.4
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

mbyrnepr2:
## Type of Changes

## Description
Fix false positive for ``keyword-arg-before-vararg`` when a positional-only parameter with a default value precedes ``*args``.

Closes #8570


pylint/checkers/typecheck.py
```diff
@@ -1001,8 +1001,17 @@ def _compiled_generated_members(self) -> tuple[Pattern[str], ...]:
 
     @only_required_for_messages("keyword-arg-before-vararg")
     def visit_functiondef(self, node: nodes.FunctionDef) -> None:
-        # check for keyword arg before varargs
+        """Check for keyword arg before varargs.
```
DanielNoord, line 1004: Bit of a nit but let's keep this as a comment. That makes it easier to later add another check to this `visit` method.

DanielNoord (approved):
One nit, rest LGTM!

pylint/checkers/typecheck.py
mbyrnepr2: Sure! Let me know if it still doesn't look right.

DanielNoord (approved):

Issue #746, opened 2015-12-17
Getting sane in 2.0


PCManticore:
I want to start a discussion regarding what we want to have in pylint 2.0, from the point of view of the usability of the tool, rather than the static analysis capabilities. There are a lot of folks out there who could use pylint, but they don't use it currently because of various reasons, most of them having
real arguments that I'd like to address with this issue.

When talking about this 2.0, I'm not referring to the previous 2.0 that arised in discussions we had over IRC, with advanced static analysis capabilities, such as control flow graphs, points-to analysis and abstract interpretation. We'll have to call the latter one 3.0 though, since it's going to be another major API breaker. 2.0, in my opinion, should have a main focus of reducing the annoyances that users have with the tool from a UX perspective.

Without further ado, here are the issues I'd like to address with 2.0:

1. Sane defaults

Pylint's output is too verbose by default. While I agree that most of the messages should
be left as is, we could try to activate the -E flag by default. It should offer the most useful set of errors that users shouldn't have in their project, leaving style checking and other non-critical checks to be activated explicitly by those interested in them.

The reports are not so useful, we could activate "--reports=n" by default. They tend to be useful in some situations, but impedes the normal workflow when working with pylint. The only capability that I'd like to have activated by default is the score, though, since it tends to make the code quality a gamification for beginners.

Remove or downgrade to extensions some checkers, such as too-few-public-methods, too-many-public-methods etc. I'll have to compile a full list for these, but I'm interested in what messages don't seem so useful for  the community.

Reevaluate constants for checker's options, such as the maximum number of statements, maximum number of arguments etc. For me, some of them are set too low and I always change them.

   
2. Prefer false negatives over false positives

Currently we're doing the opposite thing, we emit a message even if there is a high risk of a false positive to occur. Things changed quite a bit in the last year though regarding this aspect.
For instance, we don't emit no-member if we don't detect all the base classes of the class which owns the attribute we're trying to access. 

But we're not there yet, we still have tons of false positives, but this is a topic that's not going to go away so soon, requiring a higher level understanding of Python than we currently have
and more static analysis concepts brought into our need (call graphs, abstract interpretation, flow control understanding, points-to and so much more). What we can do is to try not to emit a message if we're not 100% confident that it's actually a problem. The first example that comes into my
mind is no-member, which we could try not to emit if we can infer multiple values for the same object. 
[...]

PR #8572, opened 2023-04-12
[doc] Mention that info messages are disabled by default

matusvalo:
## Type of Changes

## Description

Document that information catagory is disabled by default - see https://github.com/pylint-dev/pylint/commit/fe8f7dce2eb5fe9bc55ca269482ed7dd81258409.

Refs #746

### DIFF
--- a/doc/exts/pylint_messages.py
+++ b/doc/exts/pylint_messages.py
@@ -313,7 +313,7 @@ def _generate_single_message_body(message: MessageData) -> str:
 
 *{message.definition.description}*
 """
-    if not message.default_enabled or MSG_TYPES[message.id[0]] == 'info':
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (changes requested):
Makes sense but this needs a changelog with the associated user actions. Probably 'if you made a plugin with info level message you now need to explicitely enable them using...'

matusvalo:
#8576 

> Probably 'if you made a plugin with info level message you now need to explicitely enable them using...'

Yes and moreover I found out that at least `c-extension-no-member` Info Message is not disabled by default:

https://github.com/pylint-dev/pylint/blob/56fa5dce747a46f1dcba6eca003bb22fcc347247/pylint/lint/run.py#L158

PR #8576, opened 2023-04-13
Disable Info linters in definition instead of Run class

matusvalo:
## Type of Changes

## Description

This PR refactors code in a way that it removes explicit disable of Info Messages during creation of `Run` class but instead define it in message definition. This fixes issue when Message documentation is not showing Caution Bar that message is disabled by default.

### DIFF
--- a/pylint/testutils/testing_pylintrc
+++ b/pylint/testutils/testing_pylintrc
@@ -7,3 +7,7 @@ disable=
     suppressed-message,
     locally-disabled,
     useless-suppression,
+
+enable=
+    deprecated-pragma,
+    use-symbolic-message-instead,
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,13 +16,13 @@ repos:
             doc/data/messages/m/missing-final-newline/bad.py|
           )$
   - repo: https://github.com/charliermarsh/ruff-pre-commit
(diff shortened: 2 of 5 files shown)


matusvalo:
Not sure whether we need a changlog entry or not.

matusvalo:
I was too confident. Tests are failing. I need to investigate why...

Pierre-Sassoulas:
Some messages will need to be explicitely enabled in some tests imo.

matusvalo:
For now I am marking as skip news. If news entry is needed I will gladly create one.

pylint/testutils/testing_pylintrc
```diff
@@ -7,3 +7,7 @@ disable=
     suppressed-message,
     locally-disabled,
     useless-suppression,
+
+enable=
```
Pierre-Sassoulas, line 11: We might want to enable them for their functional tests directly. Although it's also a good thing if we have up to date functional tests with only symbolic messages. It's also an internal change for plugin dev but I don't think we advertised that file so we don't need to add a changelog entry (thinking out loud before approving to document that this was considered during review)

Pierre-Sassoulas (approved):
Great !

matusvalo:
Thank you @Pierre-Sassoulas for code review!

PR #8590, opened 2023-04-18
Allow parenthesized implicitly concatenated strings inside calls, to be more compatible with Black.

### LINKED ISSUE #8552: `implicit-str-concat`: allow wrapping implicitly concatenated strings in parenthesis.
### Current problem

Internally, we have enabled the `--check-str-concat-over-line-jumps` option for `implicit-str-concat`. But we also allow wrapping implicitly concatenated strings in parenthesis. This is done by our internal patch to Pylint ~3 years ago.

Example code:

```python
a = [
    "aa bb"
    "cc dd",
]
```

It raises `implicit-str-concat` on the `"aa bb"` line, that's good to catch bugs of missing commas.

But for long strings that must be put on multiple lines, currently you need to use explicit str concatenations:

```python
a = [
    "aaaaaaaa aaaaaaaa "
    + "aaaaaaaa aaaaaaaa ",
]
```

We think it should also be fine to use implicit str concatenations BUT wrapped inside parenthesis:

```python
a = [
    ("aaaaaaaa aaaaaaaa "
     "aaaaaaaa aaaaaaaa "),
]
```

### Desired solution

Introduce an opt-in option `allow-parenthesized-str-concat`. When enabled, allow the use of parenthesis.

### Additional context

WDYT? If you agree, I'll send out our patch.

yilei:
## Type of Changes

## Description

For discussion in #8552.

Closes #8552 

For reference, here is _Black_'s preview style results with relevant usages: [playground](https://black.vercel.app/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AMkASVdAD2IimZxl1N_WkJh9laRHh-GWxqULQiCW8Yn2RV6NgpGv7VyvjzJVYfBofZMzxDYXYeo72ye9kvUTowtSMtbr3LWSIsNPKcVenZgo7OFhylLa75jeZXrhowgSR7bmxcuzhTnkA6-eFyJG3BUIb63hatN-Zj75WvgPW57YlhAf33N_uex55RSrTYHCD9CP4c0QgYh85x1P9zvF4roHeK9qJFjweKNtnF7FzHivgGbVoYlrIvkDgNBgsGz_eM_H0lzVwA-z5FfXUNRu76aGjqdxXtLLexxKSoGocc0tQc_oGA7v5Fk_3OJMxhM3RB8dXzEBRmbMplqwc9PcRsPzA61FwVP6FE8IV48bXz50_XOjFjnVhHPWxz0g2HkC5FD9qrjk506-vLqAAAAALFDMj-nMjuqAAHBAqUGAAAp3xkdscRn-wIAAAAABFla)

Alternatively, we could add another option (default to keep the existing behavior) to silence parenthesized `implicit-str-concat`.


DanielNoord:
@yilei I guess nobody has time to review this yet, but as a start could you look at the coverage issue? You seem to be missing a line.

Pierre-Sassoulas (commented):
Sorry to react only after I read the functional tests I was not understanding things properly before. I think I thought and started to create test case about this earlier but I lost the change when my hard drive crashed because I can't find the branch anymore. What I wanted to do at some point was to check problem on a single line even inside multiline:

```python

print(
    "Lorem ipsum dolor sit amet, ""consectetur adipiscing elit,"  # [implicit-str-concat] (notice the "" before 'consectetur')
    "Lorem ipsum dolor sit amet, consectetur adipiscing elit,"
    " sed do eiusmod tempor incididunt ut labore et dolore "
    "magna aliqua. Ut enim ad minim veniam, quis nostrud "
    "exercitation ullamco laboris nisi ut aliquip ex ea "
)
```
I.e. if the string concat happens on a single line then raise ``implicit-str-concat`` else do not raise. But the approach from the checker was to be changed completely for that to work (use token I think?). Let me know what you think :)

yilei:
@DanielNoord The coverage issue should be addressed. The missed line was actually dead code that can't be reached.

@Pierre-Sassoulas Thanks for taking a look. If they are on the same line, `implicit-str-concat` is always raised even in a multi line case. I added your example as a test case. The checker is using tokens and checking token's line numbers. Or did I misunderstand something?

Pierre-Sassoulas (dismissed):
LGTM, thank you !

pylint/checkers/strings.py
```diff
@@ -821,10 +879,18 @@ def check_for_concatenated_strings(
...
-                    next_token.start[0] == elt.lineno
-                    or self.linter.config.check_str_concat_over_line_jumps
+            if (
+                matching_token != elt.value
+                and next_token is not None
+                and next_token.type == tokenize.STRING
+            ):
+                if next_token.start[0] == elt.lineno or (
+                    self.linter.config.check_str_concat_over_line_jumps
+                    # Allow implicitly concatenated strings in in parens.
```
jacobtylerwalls, line 889:
```suggestion
                    # Allow implicitly concatenated strings in parens.
```

pylint/checkers/strings.py
```diff
@@ -744,10 +750,62 @@ def process_tokens(self, tokens: list[tokenize.TokenInfo]) -> None:
...
+    ) -> bool:
+        # Must NOT be preceded by a string literal
+        prev_token = self._find_prev_token(index, tokens)
+        if prev_token and prev_token.type == tokenize.STRING:
+            return False
+        # Must be followed by a string literal token.
+        next_token = self._find_next_token(index, tokens)
+        return bool(next_token and next_token.type == tokenize.STRING)
+
+    def _is_parenthesised(self, index: int, tokens: list[tokenize.TokenInfo]) -> bool:
```
jacobtylerwalls, line 772:
```suggestion
    def _is_parenthesized(self, index: int, tokens: list[tokenize.TokenInfo]) -> bool:
```

pylint/checkers/strings.py
```diff
@@ -744,10 +750,62 @@ def process_tokens(self, tokens: list[tokenize.TokenInfo]) -> None:
                     # to match with astroid `.col_offset`
                     start = (start[0], len(line[: start[1]].encode(encoding)))
                 self.string_tokens[start] = (str_eval(token), next_token)
+                is_parenthesized = self._is_initial_string_token(
+                    i, tokens
+                ) and self._is_parenthesised(i, tokens)
```
jacobtylerwalls, line 755:
```suggestion
                ) and self._is_parenthesized(i, tokens)
```

jacobtylerwalls (approved):
Thanks! I think this is ready to merge. I found some minor typos.

Finally, it would be worth removing the reference to this PR added in #8649 and just show beneath the example there how to rewrite it correctly (as black will do it, soon).

jacobtylerwalls (approved):

jacobtylerwalls (approved):

Issue #8460, opened 2023-03-20
Pylint does not respect false values for boolean config options in toml files


GrimzEcho:
### Bug description

When pylint loads its configuration from a .toml file, it treats the mere presence of a key as enabling that flag or option, without regard to its actual value.

For instance, setting `from-stdin = false` in a toml config will cause pylint to attempt to read from stdin. Setting `exit-zero = false` will cause pylint to always exit with code 0 even if there are errors or warnings.

The only way to disable the flag is to comment out the line. This is directly at odds with the example configuration provided in the documentation (https://pylint.readthedocs.io/en/latest/user_guide/configuration/all-options.html) which has many examples of `option = false`.

This same behavior seems to happen with _pylintrc_ configuration files as well, but I only did some basic testing.

### Configuration

```toml
# example 1:
[tool.pylint.main]
from-stdin = false
exit-zero = false

# example 2:
[tool.pylint.main]
# from-stdin = false
exit-zero = false

# example 3:
[tool.pylint.main]
# from-stdin = false
# exit-zero = false
```

### Command used

```shell
pylint test.py; echo $?
```
```python
# test.py
import datetime
```

### Pylint output

```shell
# example 1
# hangs because it is expecting input from stdin ... must be killed with ctrl-c


# example 2
************* Module test
test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
test.py:1:0: W0611: Unused import datetime (unused-import)
------------------------------------------------------------------
Your code has been rated at 0.00/10 (previous run: 0.00/10, +0.00)
0


# example 3
************* Module test
test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
test.py:1:0: W0611: Unused import datetime (unused-import)
------------------------------------------------------------------
Your code has been rated at 0.00/10 (previous run: 0.00/10, +0.00)
20
```

### Expected behavior

Examples 1 and 2 should produce the same output as example 3

### Pylint version

```shell
pylint 2.17.0
astroid 2.15.0
Python 3.7.3 (default, Oct 31 2022, 14:04:00) 
[GCC 8.3.0]
```

### OS / Environment

Debian 10

### Additional dependencies

None

DanielNoord:
This is an issue with `store_true` options. Basically, these options don't work with `false` or `true` but act as flags.

We should indeed fix this and also update the documentation. Thanks for spotting it!

ianrose1:
hello, can I try and work on this issue?

[...]

PR #8591, opened 2023-04-19
Fix Pylint Configuration Parsing Bug Causing Incorrect Option Settings #8460

ianrose1:
## Type of Changes

|   | :sparkles: New feature |
|   | :hammer: Refactoring   |
|   | :scroll: Docs          |

## Description

Refs #8460

Closes #XXXX


pylint/config/config_file_parser.py
```diff
@@ -79,7 +79,8 @@ def _parse_toml_file(self, file_path: Path) -> tuple[dict[str, str], list[str]]:
                 for config, value in values.items():
                     value = _parse_rich_type_value(value)
                     config_content[config] = value
-                    options += [f"--{config}", value]
+                    if value.lower() != "false":
```
DanielNoord, line 82: I think we should also check the type of the config option. If somebody put in `false` for a non-`store_true` option we don't want to this this. A test for this would also be good!

ianrose1: Sounds good! Do you know if there's an efficient way to check the type of config? I was looking at the function "_convert_option_to_argument" in pylint/config/utils.py, but not sure if that would work.

DanielNoord: No, although you could perhaps wait with parsing them until you reach:

```python
https://github.com/pylint-dev/pylint/blob/4cdf9222ae7ed5a3122f633d04b8d8c144e60c1c/pylint/config/arguments_manager.py#L211-L214
```

This should have knowledge of all registered arguments.

DanielNoord:
Closing due to inactivity.

PR #8605, opened 2023-04-23
Cache `class_is_abstract()`

jacobtylerwalls:
## Type of Changes

## Description
For one project in the pylint primer (music21), provides 15% speedup _linting the entire project_. (171.6s -> 145.6s on an M1 mac running python3.11.)

I think we should batch the performance improvements in 3.0 and describe them all with a set of benchmarks at the end, so adding a skip news label.

### DIFF
--- a/towncrier.toml
+++ b/towncrier.toml
@@ -65,3 +65,8 @@ showcontent = true
 directory = "internal"
 name = "Internal Changes"
 showcontent = true
+
+[[tool.towncrier.type]]
+directory = "performance"
+name = "Performance Improvements"
+showcontent = true
--- a/doc/whatsnew/fragments/1954.performance
+++ b/doc/whatsnew/fragments/1954.performance
@@ -1,0 +1,3 @@
+Cache `class_is_abstract()`.
+
+Refs #1954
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas (commented):
How about we create the entry right now and add an entry each time we do a performance fix ? This way we're not going to have to do it all at once and each MR can verify that a new detail is added.

towncrier.toml
```diff
@@ -65,3 +65,8 @@ showcontent = true
 directory = "internal"
 name = "Internal Changes"
 showcontent = true
+
+[[tool.towncrier.type]]
+directory = "performance"
+name = "Performance Improvements"
+showcontent = true
```
Pierre-Sassoulas, line 72: 👍 but the doc and issue template needs to be updated for this.

jacobtylerwalls: No problem, but not sure where to look. Were you thinking of the PR template?

Pierre-Sassoulas: Yes but nvm we changed it to this recently: https://github.com/pylint-dev/pylint/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L9

Pierre-Sassoulas (approved):

PR #5534, opened 2021-12-16
Fixed extremely long processing of long lines with comma's

DanielNoord:
- [x] Add yourself to CONTRIBUTORS if you are a new contributor.
- [x] Add a ChangeLog entry describing what your PR does.
- [x] If it's a new feature, or an important bug fix, add a What's New entry in
      `doc/whatsnew/<current release.rst>`.
- [x] Write a good description on what the PR does.

## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #5483 

The problem was not actually in a regex pattern but in how we checked lines for trailing comma's.

The test is a bit arbitrary and the timing was difficult to get right: I don't want to have the tests fail because the CI runner has an off-day. But I checked with the original reported code and this is much faster.

PR #8606, opened 2023-04-23
Optimize `is_trailing_comma()`

jacobtylerwalls:
## Type of Changes

## Description

Refs #5534

When profiling a lint over `pygame`, I see a savings of over 2s on my M1 Mac/Python3.11:

**Before**
   493189    3.580    0.000    3.640    0.000 refactoring_checker.py:88(_is_trailing_comma)

**After**
   493189    1.346    0.000    1.400    0.000 refactoring_checker.py:88(_is_trailing_comma)

### DIFF
--- a/doc/whatsnew/fragments/1954.performance
+++ b/doc/whatsnew/fragments/1954.performance
@@ -1,0 +1,3 @@
+Optimize `is_trailing_comma()`.
+
+Refs #1954
--- a/doc/whatsnew/fragments/1954.performance
+++ b/doc/whatsnew/fragments/1954.performance
@@ -1,3 +1,4 @@
-Optimize `is_trailing_comma()`.
+- Optimize `is_trailing_comma()`.
+- Cache `class_is_abstract()`.
 
 Refs #1954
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas (commented):
Great ! Reading the MR you linked was entertaining: we increased the timeout on it and did no deep analysis of the performance. I suppose we were reaching the timeout regularly before and this was the straw that made us reach the timeout too much. Motivation to add benchmark regression primers (discussed in the issue) reactivated :D

Pierre-Sassoulas (approved):

jacobtylerwalls:
Thanks for the review! I have an additional commit to push to this branch, but wanted to wait for the primer 3.7 job to finish first (hopefully it's almost done!)--just to make sure it finishes reasonably.

Pierre-Sassoulas (approved):
Great ! The second optimization with ``self.linter.is_message_enabled(`` would be huge if used everywhere, but this would require design to be done right imo (``@utils.only_required_for_messages`` working for token based checker and a check to verify that it's used when required and not used when not required ?)

PR #8607, opened 2023-04-23
Fix `FunctionDef` isinstance checks

cdce8p:
## Description
Fallout from https://github.com/pylint-dev/astroid/pull/2115 - required for the next astroid update.
Fix `isinstance` checks now that `nodes.FunctionDef` doesn't inherit from `nodes.Lambda`.

Changes are compatible with the current astroid version, so don't need to wait to fix these.

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -360,7 +360,6 @@ def is_defined_before(var_node: nodes.Name) -> bool:
             defnode_scope, (*COMP_NODE_TYPES, nodes.Lambda, nodes.FunctionDef)
         ):
             # Avoid the case where var_node_scope is a nested function
-            # FunctionDef is a Lambda until https://github.com/pylint-dev/astroid/issues/291
             if isinstance(defnode_scope, nodes.FunctionDef):
                 var_node_scope = var_node.scope()
                 if var_node_scope is not defnode_scope and isinstance(
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -356,7 +356,9 @@ def is_defined_before(var_node: nodes.Name) -> bool:
         if defnode is None:
             continue
         defnode_scope = defnode.scope()
-        if isinstance(defnode_scope, (*COMP_NODE_TYPES, nodes.Lambda)):
+        if isinstance(
+            defnode_scope, (*COMP_NODE_TYPES, nodes.Lambda, nodes.FunctionDef)
+        ):
             # Avoid the case where var_node_scope is a nested function
             # FunctionDef is a Lambda until https://github.com/pylint-dev/astroid/issues/291
             if isinstance(defnode_scope, nodes.FunctionDef):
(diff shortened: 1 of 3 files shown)


pylint/checkers/variables.py
```diff
@@ -2500,7 +2500,7 @@ def _loopvar_name(self, node: astroid.Name) -> None:
         # FunctionDef subclasses Lambda due to a curious ontology. Check both.
         # See https://github.com/pylint-dev/astroid/issues/291
         # TODO: Revisit when astroid 3.0 includes the change
```
jacobtylerwalls, line 2502:
```suggestion
```

pylint/checkers/utils.py
```diff
@@ -356,7 +356,9 @@ def is_defined_before(var_node: nodes.Name) -> bool:
         if defnode is None:
             continue
         defnode_scope = defnode.scope()
-        if isinstance(defnode_scope, (*COMP_NODE_TYPES, nodes.Lambda)):
+        if isinstance(
+            defnode_scope, (*COMP_NODE_TYPES, nodes.Lambda, nodes.FunctionDef)
+        ):
             # Avoid the case where var_node_scope is a nested function
             # FunctionDef is a Lambda until https://github.com/pylint-dev/astroid/issues/291
```
jacobtylerwalls, line 363:
```suggestion
```

jacobtylerwalls (commented):
Thanks! Just so all reviewers are aware, we will still need another round of cleanup after bumping to astroid 2.16, when we could remove things like this last line:

https://github.com/pylint-dev/pylint/blob/3d036b78fca6e80222f9ca2ffd0f25950d261aed/pylint/checkers/typecheck.py#L676-L678



Either then or now, we could tidy up some comments like this:

There's also a comment to clean up in `_determine_callable(), but we can do that later:
https://github.com/pylint-dev/pylint/blob/3d036b78fca6e80222f9ca2ffd0f25950d261aed/pylint/checkers/typecheck.py#L614-L615

jacobtylerwalls (approved):

DanielNoord:
Yeah, I think this change is very likely to cause a need to do a `3.0` and perhaps `4.0` quickly after it. Then again, still think it is a good change.

Thanks @cdce8p and @jacobtylerwalls for taking these first steps towards compatibility!

jacobtylerwalls:
> Yeah, I think this change is very likely to cause a need to do a 3.0 and perhaps 4.0 quickly after it. Then again, still think it is a good change.

Agree about 3.0. Let's slow the release cadence down for a little bit, and release both 3.0 in pylint and astroid when we're ready for python 3.12. We may not need a 4.0 then.

DanielNoord:
> > Yeah, I think this change is very likely to cause a need to do a 3.0 and perhaps 4.0 quickly after it. Then again, still think it is a good change.
> 
> Agree about 3.0. Let's slow the release cadence down for a little bit, and release both 3.0 in pylint and astroid when we're ready for python 3.12. We may not need a 4.0 then.

Also fine with me, although we should definitely release `3.0a1` soon. I'd like to see behaviour in `pylint` as I'm not all that sure about the changes to some of the constructors. I think we're making definitive improvements but might have been too strict in some places.

Pierre-Sassoulas:
+1, we can keep the release cadence by releasing alpha/betas and still wait to support python 3.12 for pylint 3.0.  There's a lot to do in #3512 still, we'll probably release pylint 3.0 after python 3.12 is out in any case.

Issue #6306, opened 2022-04-14
Cleanup after we drop support for Python 3.7


Pierre-Sassoulas:
When we do drop support because 3.7 has reached its end of life after 2023-06-27, we'll need to do some cleanup:

- [x] Remove the code for ``_ExtendAction`` (#8793)
- [x] Import ``Final``, ``Literal``, and ``TypedDict`` from typing directly
- [x] Remove ``functools.cached_property`` import guards

cdce8p:
EOL for 3.7 is scheduled for `2023-06-27`.

Once 3.7 is dropped, the most notable changes for us are probably
* We can start using assignment expressions
* `Final`, `Literal`, `TypedDict` can be imported from `typing` directly
* We no longer need import guards for `functools.cached_property`

Pierre-Sassoulas:
We also need to upgrade functional tests that are exclusive to python 3.7 so they use py-version instead.

jacobtylerwalls:
> We also need to upgrade functional tests that are exclusive to python 3.7 so they use py-version instead.

For future reference, this was done when adjusting the `max_pyver` test setting, not the `min_pyver` setting. We discussed it [here](https://github.com/pylint-dev/pylint/pull/8811#discussion_r1248981435).

PR #8609, opened 2023-04-23
Drop support for Python 3.7

cdce8p:
## Description
Python 3.7 will reach its EOL in two months (2023-06-27). I think it would make sense to drop support with the next release `3.0`. That will allow us to cleanup even more old code.

I'm **not** proposing to remove the ability to lint `3.7` compatible code. If users want that, they can still set `py-version=3.7`.

Refs #6306


DanielNoord:
I would defer this choice to https://github.com/pylint-dev/astroid/pull/2137#issuecomment-1519106161, as I think this is where we will make the first/final decision about this.

cdce8p:
> I would expect some change in the functional tests too (py 3.7 exclusive tests that now need a py-version to run).

This is just a first step. Besides the test cases, there are other areas which could be improved. Had to start somewhere.

> Assignment expressions

Yeah, we could start using them with `3.8`. Although I'm not sure we should enable the code-style check for it. Sometimes using `:=` is an opinionated decision which the check just can't do.

Pierre-Sassoulas:
Regarding the functional tests we could have a regression if it never run in the CI, right ? (Added https://github.com/pylint-dev/pylint/issues/6306#issuecomment-1519118923 so we do not forget later)

cdce8p:
> Regarding the functional tests we could have a regression if it never run in the CI, right ? (Added [#6306 (comment)](https://github.com/pylint-dev/pylint/issues/6306#issuecomment-1519118923) so we do not forget later)

`py-version` isn't a full replacement. For example for `bad-reversed-sequence` we check the argument `locals` to see if it contains the `__reversed__` protocol method. This will always be true when pylint is run with `>= 3.8` so it can't detect an issue for `3.7`. That's why I would usually recommend to run pylint with the lowest supported Python version.

However, as this is the only small regression I still think dropping 3.7 now is worth it. If someone really depended on it, he can continue to use pylint `2.17.x`.

Pierre-Sassoulas (commented):
I think the primer fail is expected in this PR. I'm a little worried that all MR will need to be rebased on main after we merge but I don't think we have a way to avoid that.

cdce8p:
> I think the primer fail is expected in this PR. I'm a little worried that all MR will need to be rebased on main after we merge but I don't think we have a way to avoid that.

Yeah, that's just the nature of these cleanups.
However, I would rebase on a need bases only, not proactive. It's likely some of the breaking changes in astroid will cause additional merge conflicts down the line.

Pierre-Sassoulas (approved):

PR #8610, opened 2023-04-23
Added escaping of vertical bar character in annotation labels

### LINKED ISSUE #8603: Type hinting a return type as nullable renders `None`
### Bug description

Hi there,

the following code

```
class Foo:
    def bar() -> int | None:
        pass
```

produces

image

meaning `None` is not considered as part of the type hint. My current workaround is to revert back to `Optional[T]` like

```
class Foo:
    def bar() -> Optional[int]:
        pass
```

which correctly produces

image

I think the handling for the *new* syntax for nullable is missing, which would explain why `Optional[T]` is working but `T | None` is not.

### Configuration

```ini
`pyreverse --version`


pyreverse is included in pylint:
pylint 2.17.2
astroid 2.15.3
```
```


### Command used

```shell
`pyreverse -o png -m y src/mt5_proxies`
```

### Pylint output

```shell
See bug description
```

### Expected behavior

Type hint `T | None` is correctly handled like `Option[T]`

### Pylint version

```shell
Python 3.11.2
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

ViRuSTriNiTy:
…  to ensure it is not treated as field separator of record-based nodes, see discussion in #8603.

## Type of Changes

## Description

Currently the DOT printer produces a label that matches exactly the type hint. In case of nullable return types like `int | None` the `|` character (vertical bar) has the special meaning of a fiel separator for record-based nodes in DOT language. Hence it must be escaped as mentioned in the [docs](https://graphviz.org/doc/info/shapes.html#record):

> Braces, vertical bars and angle brackets must be escaped with a backslash character if you wish them to appear as a literal character.

Refs #8603

Closes #8603


ViRuSTriNiTy:
How to proceed with the failing tasks?

Pierre-Sassoulas (commented):
The expected result changed in a lot of tests that would need to be updated. But I think a manual check in each printers (dot, plantuml, mermaidjs) is required first. I'm not sure it can be escaped generically, maybe it's a solution specific to dot.

ViRuSTriNiTy:
@Pierre-Sassoulas Ok, at the time I posted my comment on how to proceed with the failing tasks the status of the checks was looking different. I then committed another change to get rid of the failing tasks. Now, as you already mentioned, many tests are failing because of the escaping implemented in a method that seems to be used n other printers as well. Didn't know that and the reference checks in VSCode did not reveal this. Hopefully I can find another solution to fix the missing escaping.

tests/pyreverse/data/classes_No_Name.html
```diff
@@ -2,7 +2,7 @@
   <body>
     <script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>
       <div class="mermaid">
-
+
```
DudeNr33, line 5: Just a minor nitpick: can you remove the whitespace here and in line 50 please?

tests/pyreverse/data/packages_No_Name.html
```diff
@@ -2,18 +2,20 @@
   <body>
     <script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>
       <div class="mermaid">
-
+
```
DudeNr33, line 5: Same nitpick as above, please remove the whitespace, and again on line 18.

DudeNr33 (commented):
Thank you for providing a fix for this issue!

Looks good from my side, there is only some whitespace in the HTML output that should be removed again.
You also have to provide a towncrier fragment which we use to generate our changelog. Simply run:

```console
towncrier create 8603.bugfix
```

and then add a short description of the issue that was solved, and a reference to the issue:

```rst
<your description>

Closes #8603
```


As a side note, if you want to contribute to `pyreverse` again:
At least for class diagrams, it is most of the times easier to create a *functional test* file instead of expanding the existing test data under `tests/data`. The latter has the drawback (as you probably noticed 😁) that a lot of files - possibly for output formats unaffected by this change - have to be adapted, and the tests for `test_diadefs` and `test_inspector` have to be updated as well. 
For this PR you can leave the tests as they are, but this could save you some time in the future!

tests/pyreverse/data/classes_No_Name.html
ViRuSTriNiTy: Good catch, removed in commit c6cea8088cc38ee6fbb1eef6077b2bad6f3503cd

tests/pyreverse/data/packages_No_Name.html
ViRuSTriNiTy: Good catch, removed in commit c6cea8088cc38ee6fbb1eef6077b2bad6f3503cd

ViRuSTriNiTy:
> Thank you for providing a fix for this issue!
> 
> Looks good from my side, there is only some whitespace in the HTML output that should be removed again.

The whitespaces are produced by the file generators. I just copied the generated files from my temp directory to the data folder. Anyway, I removed it as requested, looks better this way for sure.

> You also have to provide a towncrier fragment which we use to generate our changelog. Simply run:
> 
> ```
> towncrier create 8603.bugfix
> ```
> 
> and then add a short description of the issue that was solved, and a reference to the issue:
> 
> ```
> <your description>
> 
> Closes #8603
> ```

Yep, done in commit 28d168a5960c166ef5565ea687f06a958b1b945d 

> As a side note, if you want to contribute to `pyreverse` again: At least for class diagrams, it is most of the times easier to create a _functional test_ file instead of expanding the existing test data under `tests/data`. The latter has the drawback (as you probably noticed 😁) that a lot of files - possibly for output formats unaffected by this change - have to be adapted, and the tests for `test_diadefs` and `test_inspector` have to be updated as well. For this PR you can leave the tests as they are, but this could save you some time in the future!

I tried out various approaches to this and to be honest, the functional folder scared me of a bit as I have never used such a approach ^^. My thinking here was to provide an entry point for testing other parts of the code where nullable syntax can be used like method arguments. I initially got "creative" with extending `clientmodule_test.py` but I couldn't extend it in a way where I could say for my self: "Ok, this code makes sense".  Then I found the `nullable_pattern.py` ...

Anyway, great to see the PR being accepted.

Pierre-Sassoulas (approved):

doc/whatsnew/fragments/8603.bugfix
```diff
@@ -0,0 +1,3 @@
+Added escaping of vertical bar character in annotation labels produced by DOT printer to ensure it is not treated as field separator of record-based nodes.
```
DudeNr33, line 1:
```suggestion
``pyreverse``: added escaping of vertical bar character in annotation labels produced by DOT printer to ensure it is not treated as field separator of record-based nodes.
```

DudeNr33 (changes requested):
Thanks for incorporating the feedback. We should make it clear in the news fragment that this change deals with `pyreverse`, otherwise it is fine.

doc/whatsnew/fragments/8603.bugfix
ViRuSTriNiTy: Done in commit 72bb5af492c484e7d6d9a2924992dd6ce969ff08

ViRuSTriNiTy:
@DudeNr33 Do you normally resolve the open conversations resulting from the review process or does the PR creator do that? I would assume the former.

DudeNr33 (approved):
We do not have a strict rule that all conversations have to be resolved before a PR can be merged. I normally also don't resolve conversations as - at least for small PRs - I find it easier to see what was discussed and what the result was. But that is personal preference.

ViRuSTriNiTy:
Who merges the PR?

DudeNr33:
One of the maintainers that have the required rights to merge. Typically one of those that also reviewed the changes. 

Congratulations on becoming a contributor! 😊

PR #8612, opened 2023-04-24
Add home-assistant to primer and remove old external primer

### LINKED ISSUE #5359: Better primer tests (faster, less duplication, more diverse)

jacobtylerwalls:
## Type of Changes

## Description
Restore `home-assistant` to the new primer, and finally remove the old (external) primer. The old (check for fatal only) primer still runs on the stdlib.

We were waiting for the 3.7 primer job to take less time before doing this, but we fixed that in #8608, and we're dropping 3.7 soon.

The commit string was too large for the cache action (512 chars ceiling), so I shortened the SHAs it uses.

Removed `keras`, as it wasn't linting successfully in this context, unfortunately.

[Successful run on my fork](https://github.com/jacobtylerwalls/pylint/actions/runs/4781412105/jobs/8499847553)

Closes #5359


Pierre-Sassoulas (commented):
Isn't keras failing in the "new" primer a valid reason to have it in the primer ? (If we only add projects that work from the get go we're selecting projects that are somewhat more compatible with pylint than the general projects pool)

jacobtylerwalls:
> Isn't keras failing in the "new" primer a valid reason to have it in the primer ? (If we only add projects that work from the get go we're selecting projects that are somewhat more compatible with pylint than the general projects pool)

Keras works locally when fewer projects are involved. We're blowing out available memory on Github Actions, I assume. We probably have too many projects in the primer anyway. If we ever parallelized the primer job, we could consider adding it back.

Pierre-Sassoulas (approved):
👌

Pierre-Sassoulas:
I was going to squash but it occurred to me that keeping the 4 commit separated in the history could be preferable, but there's a conflict so it's impossible to rebase. I'll let you choose what to do :)

jacobtylerwalls:
We can squash it; we don't need a super-fine resolution on a git blame for this testing tool...

DanielNoord:
I think this regressed? See https://github.com/pylint-dev/pylint/pull/8638, primer can't run anymore.

jacobtylerwalls:
Taking a look

PR #8624, opened 2023-04-24
Use assignment expressions

nickdrozd:
## Type of Changes

## Description

Enable the `consider-using-assignment-expr` check and address all warnings.

### DIFF
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -467,7 +467,7 @@ def _is_attribute_property(name: str, klass: nodes.ClassDef) -> bool:
         if inferred.pytype() != property_name:
             continue
 
-        if (_cls := node_frame_class(inferred)) == klass.declared_metaclass():
+        if node_frame_class(inferred) == klass.declared_metaclass():
             continue
         return True
     return False
--- a/pylint/checkers/classes/class_checker.py
+++ b/pylint/checkers/classes/class_checker.py
@@ -467,8 +467,7 @@ def _is_attribute_property(name: str, klass: nodes.ClassDef) -> bool:
         if inferred.pytype() != property_name:
             continue
 
-        cls = node_frame_class(inferred)
-        if cls == klass.declared_metaclass():
+        if (_cls := node_frame_class(inferred)) == klass.declared_metaclass():
             continue
         return True
     return False
@@ -742,8 +741,7 @@ class ScopeAccessMap:
     def set_accessed(self, node: _AccessNodes) -> None:
         """Set the given node as accessed."""
 
-        frame = node_frame_class(node)
-        if frame is None:
+        if (frame := node_frame_class(node)) is None:
             # The node does not live in a class.
             return
         self._scopes[frame][node.attrname].append(node)
@@ -938,8 +936,7 @@ a metaclass class method.",
         a class or a type.
         """
         for base in node.bases:
-            ancestor = safe_infer(base)
-            if not ancestor:
+            if not (ancestor := safe_infer(base)):
                 continue
             if isinstance(ancestor, astroid.Instance) and ancestor.is_subtype_of(
                 "builtins.type"
@@ -968,8 +965,7 @@ a metaclass class method.",
         if not self._py38_plus:
             return
         for base in node.bases:
-            ancestor = safe_infer(base)
-            if not ancestor:
+            if not (ancestor := safe_infer(base)):
                 continue
 
             if isinstance(ancestor, nodes.ClassDef) and (
@@ -1252,8 +1248,7 @@ a metaclass class method.",
                         return
 
                 # Infer the decorator and see if it returns something useful
-                inferred = safe_infer(decorator)
-                if not inferred:
+                if not (inferred := safe_infer(decorator)):
                     return
                 if isinstance(inferred, nodes.FunctionDef):
                     # Okay, it's a decorator, let's see what it can infer.
@@ -1506,9 +1501,8 @@ a metaclass class method.",
         }
 
         # Slots which are common to `node` and its parent classes
-        redefined_slots = ancestors_slots_names.intersection(slots_names)
 
-        if redefined_slots:
+        if redefined_slots := ancestors_slots_names.intersection(slots_names):
             self.add_message(
                 "redefined-slots-in-subclass",
                 args=([name for name in slots_names if name in redefined_slots],),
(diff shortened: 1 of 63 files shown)


pylint/checkers/classes/class_checker.py
```diff
@@ -467,8 +467,7 @@ def _is_attribute_property(name: str, klass: nodes.ClassDef) -> bool:
         if inferred.pytype() != property_name:
             continue
 
-        cls = node_frame_class(inferred)
-        if cls == klass.declared_metaclass():
+        if (_cls := node_frame_class(inferred)) == klass.declared_metaclass():
```
cdce8p, line 470: Keeping it separate is better here IMO

pylint/checkers/classes/class_checker.py
```diff
@@ -1506,9 +1501,8 @@ def _check_redefined_slots(
         }
 
         # Slots which are common to `node` and its parent classes
-        redefined_slots = ancestors_slots_names.intersection(slots_names)
 
-        if redefined_slots:
+        if redefined_slots := ancestors_slots_names.intersection(slots_names):
```
cdce8p, line 1505: This is equally hard too read

cdce8p (changes requested):
Thanks for the PR. However I don't think we should move forward with it in its current form.
When I originally added this check to promote the usage of `:=` in Home Assistant, I had a similar goal: "Replace every possible instance. Fewer lines are always better." That was quite a while ago.

In the meantime, I've realized that it isn't always the case. One of the tradeoffs is complexity and while it's still possible to understand what the code is doing, it sometimes takes a moment longer. Thus leading me to give up on the initial goal for Home Assistant.

I do still use them from time to time when I think it makes sense but mostly for fairly simple assignments. The check isn't sophisticated enough to know which these are unfortunately.

An additional argument against it is that these kind of refactorings (sometimes) destroy the visual code flow. There are cases where an author specifically wanted to group certain operations together, maybe even with a nice comment, that will now be broken up. Personally I don't think that is worth if.

--
**What to do now?**
My suggestion would be not to move forward with in in this form.
- If there truly is a desire for refactoring, I would suggest to break it up into smaller PRs to be able to review it better. Furthermore, I would suggest to limit the changes to the most obvious examples where a clear benefit is visible.
- I think it would make sense to disable the `consider-using-assignment-expression` check be default. As explained earlier, it just can't detect these nuances that are required here.

pylint/checkers/classes/class_checker.py
nickdrozd: Why? That variable isn't used anywhere else.

pylint/checkers/classes/class_checker.py
nickdrozd: Sounds like you don't like the language feature then? This is straightforward.

pylint/checkers/strings.py
```diff
@@ -873,19 +871,15 @@ def process_non_raw_string_token(
         # sequence <https://docs.python.org/reference/lexical_analysis.html>
         #
         index = 0
-        while True:
-            index = string_body.find("\\", index)
-            if index == -1:
-                break
+        while (index := string_body.find("\\", index)) != -1:
```
nickdrozd, line 874: Here's a nice case where using an assignment expression makes the logic dramatically simpler.

pylint/checkers/classes/class_checker.py
Pierre-Sassoulas: There's seven concepts on the line (variable name, assignment expression, function call, function args, equal operator, namespace, function call) and a human can handle 5 to 7 at once on a good day. It's too much info packed on the same line.

nickdrozd: Okay, I cut out the unused variable. Now the line looks like this:

```python
        if node_frame_class(inferred) == klass.declared_metaclass():
            continue
```
It is now logically simpler than a clause just above it:
```python
        if isinstance(inferred, nodes.FunctionDef) and decorated_with_property(
            inferred
        ):
            return True
```
It seems like there's a general perception that all this code is already written in an optimally readable style. But in my experience that's far from the case.

Pierre-Sassoulas: > It seems like there's a general perception that all this code is already written in an optimally readable style. 

There is above all a very high risk of introducing bug with this kind of stylistic changes. The code as it is might not be "optimally readable" but it is **good enough** and has the benefit of being battle tested. 

> If there truly is a desire for refactoring, I would suggest to break it up into smaller PRs to be able to review it better. Furthermore, I would suggest to limit the changes to the most obvious examples where a clear benefit is visible.

Can't agree more with that statement.

> I think it would make sense to disable the consider-using-assignment-expression check be default. As explained earlier, it just can't detect these nuances that are required here.

We can also learn a lot about what is wrong with the check atm, in particular I saw a false positive on a list comprehension that we had to disable, there are line that would be too complex or too long if we apply the suggestion blindly. This is an opportunity to make this check better.

cdce8p: > > I think it would make sense to disable the consider-using-assignment-expression check be default. As explained earlier, it just can't detect these nuances that are required here.
> 
> We can also learn a lot about what is wrong with the check atm, in particular I saw a false positive on a list comprehension that we had to disable, there are line that would be too complex or too long if we apply the suggestion blindly. This is an opportunity to make this check better.

I did attempt to do this at one point but it never felt good enough, so I didn't create a PR for it. One of the keys seems to be to make it more customizable but this undoubtedly adds (a lot) of complexity. This is my working branch: https://github.com/pylint-dev/pylint/compare/main...cdce8p:customize_assignment-expr

Just for reference the config I tested for it with Home Assistant.
<details>
<summary>pylint config</summary>

```
[tool.pylint.CODE_STYLE]
max-line-length-suggestions = 72
assignment-expr-not = true
assignment-expr-compare-none = true
assignment-expr-compare-const = true
assignment-expr-compare-in = false
assignment-expr-compare-call = false
assignment-expr-assign-not = true
assignment-expr-assign-getitem = true
assignment-expr-assign-annotation = false
assignment-expr-assign-await = true
assignment-expr-assign-ifexp = false
assignment-expr-assign-comprehension = false
assignment-expr-assign-lambda = false
assignment-expr-assign-compare = false
assignment-expr-assign-boolop = false
assignment-expr-assign-binop = false
assignment-expr-assign-call = true
assignment-expr-assign-call-func-names = [
#    "get",
#    "async_get",
#    "async_get_last_state",
    "search",
    "match",
    "parse_datetime",
    "parse_time",
    "parse_date",
    "async_load",
#    "get_entity",
    "async_prepare_reload",
    "pop",
    "getattr",
]
assignment-expr-assign-call-func-names-rgxs = [
    "^_?(async_)?get(_.*)?$",
    "^_?convert(_.*)?$",
]
```
</details>

PR #8629, opened 2023-04-26
Add applicability notes for `compare-to-empty-string/zero` (#8592)

### LINKED ISSUE #8592: `compare-to-empty-string` false positive when the type is a union
### Bug description

```python
a: str | int = "" if some_condition else 0
if a == "":
    print("some_condition")
```

### Configuration

```ini
load-plugins = "pylint.extensions.emptystring"
```

### Command used

```shell
pylint
```

### Pylint output

```shell
C1901: "a == ''" can be simplified to "not a" as an empty string is falsey (compare-to-empty-string)
```

### Expected behavior

No output as `a == ''` can not be simplified to `not a` in this case.  Note that `a` is `int | str`.

I know that using a datatype of `int | str` is a bad practice but apparently someone in our team decided to do this anyhow.  In making the code compliant to pylint, this causes the behavior of the code to change.

I would suggest pylint to take type analysis into account when suggesting the falsey replacement, as `can be simplified` is misleading somehow.  If this is not feasible, I feel that adding `can be simplified if the type of "a" is "str"` to the message, or adding a warning to the `pylint.extensions.emptystring` plugin would be helpful.

### Pylint version

```shell
pylint 2.17.2
astroid 2.15.3
Python 3.10.10 (main, Feb  8 2023, 14:49:42) [GCC 7.5.0]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

dotkrnl:
The extension `compare-to-empty-string` is only applicable when the expression being compared is strictly a `str`.  The extension `compare-to-zero` is only applicable when the expression being compared is strictly an `int`.

This PR adds the applicability notes of those two extensions in the docs and also in the messages.

## Type of Changes

## Description

Closes #8592.

### DIFF
--- a/doc/user_guide/checkers/extensions.rst
+++ b/doc/user_guide/checkers/extensions.rst
@@ -96,9 +96,20 @@ Compare-To-Empty-String checker
 This checker is provided by ``pylint.extensions.emptystring``.
 Verbatim name of the checker is ``compare-to-empty-string``.
 
+Compare-To-Empty-String checker Documentation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can use this plugin for finding patterns such as ``a == ''`` and ``a != ''`` in your code base.
+It is usually possible to replace ``a == ''`` with a Pythonic expression of ``not a``
+and replace ``a != ''`` with ``a``, if the variable ``a`` is strictly a ``str``.
+
+Note that this replacement is not always safe to perform when the expression
+being compared to an empty string is not an ``str``, or a union of ``str`` with other types.
+For example, ``False != ''`` is ``True``, while the expression ``False`` is falsey.
+
 Compare-To-Empty-String checker Messages
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


Pierre-Sassoulas (commented):
Thank you for opening this PR ! I did some work to uniformise those kind of checks (use-implicit-booleaness-x) in https://github.com/pylint-dev/pylint/pull/6870 that is not merged yet. I think it would be neat to build on it for this check (so the message is generic, for example "if %s is strictly typed" or "if %s is strictly a %s". What do you think ?

dotkrnl:
@Pierre-Sassoulas sounds great, I think "if %s is strictly a %s" is better as "int | str" is also typed.  Also maybe we would consider "a(n)" instead of "a" or check the spelling of the type in the code.  I do not have bandwidth recently so I would appreciate it if you could take over this case if we want to build it upon the uniformization :-).

Pierre-Sassoulas:
Sure thing, I'll cherry-pick your commit, thank you for the suggestion and the better/clearer message.

Pierre-Sassoulas:
Closing after cherry-picking this onto #6870, thank you for the suggestion, it's a tricky one, the warning was necessary. Congratulation on becoming a pylint contributor :)

PR #8639, opened 2023-04-30
[primer] Fix commit string generation

jacobtylerwalls:
Follow-up to 87a4d672a8661aa61d5717fd902315d9563e33b6.

The shortening of the commit SHA to 8 chars was done in only 1 of 3 places, making it so the primer would fail the subsequent time after regenerating the list of projects.

### DIFF
--- a/pylint/testutils/_primer/primer_prepare_command.py
+++ b/pylint/testutils/_primer/primer_prepare_command.py
@@ -15,7 +15,7 @@ class PrepareCommand(PrimerCommand):
     def run(self) -> None:
         commit_string = ""
         version_string = ".".join(str(x) for x in sys.version_info[:2])
-        # Shorten the SHA to avoid exceeeding GitHub's 512 char ceiling
+        # Shorten the SHA to avoid exceeding GitHub's 512 char ceiling
         if self.config.clone:
             for package, data in self.packages.items():
                 local_commit = data.lazy_clone()
--- a/pylint/testutils/_primer/primer_prepare_command.py
+++ b/pylint/testutils/_primer/primer_prepare_command.py
@@ -15,16 +15,17 @@ class PrepareCommand(PrimerCommand):
     def run(self) -> None:
         commit_string = ""
         version_string = ".".join(str(x) for x in sys.version_info[:2])
+        # Shorten the SHA to avoid exceeeding GitHub's 512 char ceiling
         if self.config.clone:
             for package, data in self.packages.items():
                 local_commit = data.lazy_clone()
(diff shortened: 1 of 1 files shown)


pylint/testutils/_primer/primer_prepare_command.py
```diff
@@ -15,16 +15,17 @@ class PrepareCommand(PrimerCommand):
...
+                commit_string += local_commit[:8] + "_"
         elif self.config.check:
             for package, data in self.packages.items():
                 local_commit = Repo(data.clone_directory).head.object.hexsha
                 print(f"Found '{package}' at commit '{local_commit}'.")
-                commit_string += local_commit + "_"
+                commit_string += local_commit[:8] + "_"
         elif self.config.make_commit_string:
             for package, data in self.packages.items():
                 remote_sha1_commit = (
```
jacobtylerwalls, line 31: Just below is where the prior commit shortened the SHA to 8 chars.

pylint/testutils/_primer/primer_prepare_command.py
```diff
@@ -15,16 +15,17 @@ class PrepareCommand(PrimerCommand):
     def run(self) -> None:
         commit_string = ""
         version_string = ".".join(str(x) for x in sys.version_info[:2])
+        # Shorten the SHA to avoid exceeeding GitHub's 512 char ceiling
```
jacobtylerwalls, line 18:
```suggestion
        # Shorten the SHA to avoid exceeding GitHub's 512 char ceiling
```

DanielNoord (commented):
Can we just apply some sort of hash to the commit string in the workflow? That seems more future proof and keeps giving us the full string in the debug file

jacobtylerwalls:
Hm. You'd find the longer strings/SHAs easier to work with? I don't know that I would.

DanielNoord:
I haven't been looking at the primer for some time, but while initially working on it just being able to copy the full sha made it quite easy to find the commit we were priming against.

jacobtylerwalls:
Gotcha. I'm in the habit of copying the short SHAs. TBH, I'm just trying to unblock main at this point. Re: hashing the commit string, that just adds a level of indirection when debugging the workflows in Actions, it seems to me. I was able to debug this very issue more quickly because I could see the commit string being used as a key and quickly identify that it was too long. With more indirection, I can't say I'd be as effective. WDYT?

DanielNoord:
Fine with me! If we ever have too much projects so that even this fix doesn't work will cross that bridge!

PR #8644, opened 2023-05-01
Add new checker `kwarg-superseded-by-positional-arg` and fix a false positive

### LINKED ISSUE #8558: Warn when positional-only arguments defeated by variadic keyword arguments
### Current problem

When writing a function, you can defeat the point of a PEP 570 positional-only argument (to the left of `/`) if:
- it has a default
- later, there is a variadic positional-or-keyword argument e.g. `**kwargs`

**GOOD** example (adapted from PEP 570, "[Semantic Corner Case](https://peps.python.org/pep-0570/#semantic-corner-case)"):
```python
def foo(name, /, **kwds):
    return name
```
See that name is actually positional-only:
```python
>>> foo(name="Jacob")
TypeError: foo() missing 1 required positional argument: 'name'
```

**BAD** example:
```python
def foo(name="Sarah", /, **kwds):
    return name
```

See that name is completely ignored:
```python
>>> foo(name="Jacob")
'Sarah'
```

### Desired solution

I suggest a new warning e.g. `defeated-positional-only-argument`

### Additional context

Inspired by #8555 

[PEP 570](https://peps.python.org/pep-0570)

mbyrnepr2:
## Type of Changes

|  ✓  | :scroll: Docs          |

## Description

Add a new checker ``kwarg-superseded-by-positional-arg`` to warn when a function is called with a keyword argument which shares a name with a positional-only parameter and the function contains a keyword variadic parameter dictionary. It may be surprising behaviour when the keyword argument is added to the keyword variadic parameter dictionary.

This prevents a false positive, in the above scenario, for ``redundant-keyword-arg``.

Closes #8558


tests/functional/a/arguments.py
```diff
@@ -278,3 +278,29 @@ def _print_selection(self):
...
+
+
+def name2(apple, /, orange, **kwargs):
+    """
+    Positional-only parameter with positional-or-keyword parameter and `**kwargs`.
+    """
+
+
+# +1:[redundant-keyword-arg]
+name2("Red apple", "Green pear", apple="Green apple", orange="Yellow pear")
```
jacobtylerwalls, line 306: Nice. Wondering why the oranges are actually pears?

pylint/checkers/typecheck.py
```diff
@@ -1548,6 +1548,12 @@ def visit_call(self, node: nodes.Call) -> None:
 
         # 2. Match the keyword arguments.
         for keyword in keyword_args:
+            # Skip if `keyword` is the same name as a positional-only-parameter
+            # and a `**kwargs` parameter exists.
+            if called.args.kwarg and keyword in [
+                arg.name for arg in called.args.posonlyargs
+            ]:
+                continue
```
jacobtylerwalls, line 1556: What do you think about a followup PR for 3.0 that emits a new message for this scenario? The code works, but it's pretty suspect (it's more likely a user made a mistake than a user is trying to call a signature that intentionally uses this weird semantics).

jacobtylerwalls (approved):

Pierre-Sassoulas:
Moving to 2.17.5 because 2.17.4 is already generated in #8664

pylint/checkers/typecheck.py
jacobtylerwalls: In that case, we could leave the linked issue open until we create the "lower severity message" it asks for.

mbyrnepr2: I'm open to that. So for now I can use #Refs instead of #Closes in the fragment and we can separately address adding the lower-severity message in a follow-up.

mbyrnepr2: I think your last comment on the issue made me think you were leaning against the idea of adding a new message and were considering this primarily as a false positive @jacobtylerwalls but happy to leave the issue open anyway until we reach a conclusion.

jacobtylerwalls: > I think your last comment on the issue made me think you were leaning against the idea of adding a new message and were considering this primarily as a false positive @jacobtylerwalls but happy to leave the issue open anyway until we reach a conclusion.

I was extraordinarily indecisive!

But on second thought, yeah, I think a new message would be good. This is such an edge case, though, it doesn't need to take a high priority.

jacobtylerwalls (approved):

pylint/checkers/typecheck.py
Pierre-Sassoulas: What would the new message be ? It feels like this should be a warning, the value in kwargs is misleading (is it going to be superseded by the one in args ? Frankly without reading the example and comments I would not know). Also what if the code does ``apple = kwargs.get("apple", Default apple")`` (like you probably expect if you're giving kwargs), isn't it a redundant keyword args anyway ?

jacobtylerwalls: > What would the new message be ? It feels like this should be a warning, the value in kwargs is misleading (is it going to be superseded by the one in args ?

My idea was that it needs to be a lower severity message than redundant-keyword-arg because it's a valid language feature, as opposed to something that will raise TypeError when tried, like the other use cases for redundant-keyword-arg.

However, it's a very, very unlikely language feature, and more likely that someone just wrote a bad call to a function that _didn't_ have the kind of signature shown in the test cases. We could in that case emit a lower-severity info message that's like "hey, are you sure you're want to be doing this strange thing?"

jacobtylerwalls: Maybe `hidden-kwarg`?

Pierre-Sassoulas: ``redundant-keyword-arg`` would describes both the error and the warning well imo. In fact ``redundant-keyword-arg`` describes the warning better, while the error would be better described by ``multiple-value-for-argument`` (Like the text for the type error raised by python in this case: `` got multiple values for argument 'x'``). Our message name need to be unique and retro-compatible so a possible solution would be to rename ``redundant-keyword-arg`` => ``multiple-value-for-argument``, and create ``redundant-keyword-argument``. That's convoluted and confusing in particular for ``redundant-keyword-arg`` / ``redundant-keyword-argument`` though. Maybe we can live with ``redundant-keyword-arg`` for both ?

mbyrnepr2: There's good reasons for either of the two approaches now mentioned.
How about we: Revert the code-change in this merge-request (i.e. we emit `redundant-keyword-arg` for this case) and merge the rest (the test-cases and the updated test output) which can be useful if we re-visit this to add a new low-severity message?

Pierre-Sassoulas: We could also name the warning ``kwargs-hidden-by-positional-args`` or ``hidden-kwarg`` or something else (?) like @jacobtylerwalls suggested so there's no confusion between the two. (And we can rename the error anyway, or not)

Pierre-Sassoulas (approved):
Let's discuss the name in the next MR then.

mbyrnepr2:
I was just about to update with the new checker :D.

mbyrnepr2:
Oh ok; well lets split it then. I can open a new MR for the checker.

tests/functional/a/arguments_positional_only.py
mbyrnepr2, line 1: Moved the contents of this module into `tests/functional/a/arguments.py` since Python 3.8 is the minimum version on the main branch now; the original reason for having this module was to let the positional-only argument syntax be used which required python 3.8 minimum as configured in the `arguments_positional_only.rc` file.

doc/data/messages/h/hidden-kwarg/bad.py
mbyrnepr2, line 1: Shamelessly stealing Jacob's examples :D

tests/functional/p/positional_only_arguments_expected.py
```diff
@@ -1,5 +1,5 @@
 # pylint: disable=missing-docstring,unused-argument,pointless-statement
-# pylint: disable=too-few-public-methods
+# pylint: disable=too-few-public-methods, hidden-kwarg
```
jacobtylerwalls, line 2: Interesting, which line raises hidden-kwarg in this file?

tests/functional/h/hidden_kwarg.py
```diff
@@ -0,0 +1,31 @@
...
+    {"apple": "Green apple"}
+    """
+
+    print(apple)
+    print(banana)
+    print(kwargs)
+
+
+# `apple` is hidden:
+name1("Red apple", apple="Green apple", banana="Green banana")  # [hidden-kwarg]
```
jacobtylerwalls, line 29: Can you add a test case for a signature having a default for `apple=` and only kwargs in the call? I feel like that's the most likely mistake. Such an example would also be good for the `good.py` file.

mbyrnepr2: You got it. I don't think I understand the second part of your comment; I think the `good.py` example already has something like you describe; can you clarify please?

jacobtylerwalls: My fault, I was viewing just a commit and thought I was viewing the PR. Apologies for the rushed comment!

tests/functional/a/arguments_positional_only.py
jacobtylerwalls: Either way is fine with me, no need to change on my account.

jacobtylerwalls (dismissed):

tests/functional/h/hidden_kwarg.py
mbyrnepr2: No worries! Thanks for your help reviewing 👍

tests/functional/p/positional_only_arguments_expected.py
jacobtylerwalls: No need. Thanks.

pylint/checkers/typecheck.py
```diff
@@ -399,6 +399,14 @@ def _missing_member_hint(
         "isinstance-second-argument-not-valid-type",
         "Emitted when the second argument of an isinstance call is not a type.",
     ),
+    "W1117": (
+        "The keyword argument %r will be added to the keyword variadic "
+        "parameter dictionary %r since it has the same name as a positional-only parameter",
+        "hidden-kwarg",
```
Pierre-Sassoulas, line 405:
```suggestion
        "duplicated-arg-hidden-in-kwarg",
```

```suggestion
        "duplicate-keyword-argument",
```

```suggestion
        "kwarg-hidden-by-posarg",
```

I'm trying to come up with something more descriptive. We have https://pylint.pycqa.org/en/latest/user_guide/messages/error/duplicate-argument-name.html that is close.

mbyrnepr2: Thanks for thinking along on this @Pierre-Sassoulas. I found it difficult to think of an accurate name :D.

- I think `hidden` may be a bit confusing as a term.
- Regarding `duplicate-keyword-argument`; it suggests to me 2 or more keywords were passed to the function call explicitly which isn't the case - also the message is specifically emitted when positional-only parameters are present, so I thought it would be good to make that explicit.

What do you think of this one:
`same-keyword-as-positional-with-kwargs` (inspired by the name of a [unittest in cpython](https://github.com/python/cpython/blob/main/Lib/test/test_positional_only_arg.py#LL326C21-L326C21))
Bit of a mouthful but I like that it uses the words _positional_ and _kwargs_ which are required for the message to be emitted.

Pierre-Sassoulas: Maybe "kwarg-overridden-by-positional-arg", "kwarg-superseded-by-positional-arg" ?

mbyrnepr2: I'd say the latter suggestion, `kwarg-superseded-by-positional-arg`, is more accurate; let's try it out.

pylint/checkers/typecheck.py
```diff
@@ -1548,6 +1555,18 @@ def visit_call(self, node: nodes.Call) -> None:
 
         # 2. Match the keyword arguments.
         for keyword in keyword_args:
+            # Skip if `keyword` is the same name as a positional-only-parameter
```
mbyrnepr2, line 1558:
```suggestion
            # Skip if `keyword` is the same name as a positional-only parameter
```

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Great new check, the doc's bad/good example is especially good.

PR #8647, opened 2023-05-01
spellchecking: add multi-locale support

### LINKED ISSUE #8083: Feature request: support multiple locales in spelling-dict option
### Current problem

Recently, spellchecking the word `onwards` using an `en_US` [failed during a `pylint.git` continuous integration workflow](https://github.com/PyCQA/pylint/actions/runs/3958657181/jobs/6780499845#step:9:49); it would likely succeed with most `en_GB` dictionaries.

The problem is more general, though: codebases may have contributors who read and write using a variety of locales.

### Desired solution

The [`spelling-dict` option](https://github.com/PyCQA/pylint/blob/3ba77586cde33de15480869b6872e920eb94c580/doc/user_guide/configuration/all-options.rst#--spelling-dict) could be extended to accept multiple locale options to spellcheck against.

### Additional context

[`pyenchant`](https://pypi.org/project/pyenchant/) supports [multi-language spellchecking](https://pyenchant.github.io/pyenchant/tutorial.html#provider-ordering).

jayaddison:
This change allows multiple locale codes to be specified using the existing 'spelling-dict' configuration key.

In practice, multiple locales can be provided by using TOML-based configuration, or by repeating the '--spelling-dict' option in the pylint commandline arguments.

The change maintains backwards compatibility for ini-format pylintrc files.

## Type of Changes

## Description

Closes #8083


jayaddison:
Ah, hrm.  The non-benchmark tests added here require `en_US` and `en_CA` dictionaries.  Re-using the [dictionary install steps from the self-checks workflow](https://github.com/pylint-dev/pylint/blob/f8288842e58a8b6007ff15ab380c84b00eb9b7a3/.github/workflows/checks.yaml#L114-L117) should be straightforward for the `ubuntu`-based unit test jobs.. maybe another approach would be required for Windows and MacOS, though.

pylint/checkers/spelling.py
```diff
@@ -229,11 +231,12 @@ class SpellingChecker(BaseTokenChecker):
         (
             "spelling-dict",
             {
-                "default": "",
+                "default": [],
                 "type": "choice",
                 "metavar": "<dict name>",
                 "choices": _get_enchant_dict_choices(enchant_dicts),
```
Pierre-Sassoulas, line 237:
```suggestion
```
We can also remove ``_get_enchant_dict_choices`` if it's not used elsewhere. Disregard this suggestion if the expected use is: ``--spelling-dict=en_gb --spelling-dict=en_us``, but then how would it work in the configuration file ? (It needs to be compatible with command line interface and configuration file at the same time.)

Pierre-Sassoulas (commented):
Looks nice, we would need functional configuration tests to have a better vision of what's going to happen. See https://github.com/pylint-dev/pylint/tree/main/tests/config/functional for examples.

pylint/checkers/spelling.py
jayaddison: I think CSV-representation is reasonable (and backwards-compatible) in the rc-configuration file, array configuration would be better in the TOML file, and repeated parameters (`--spelling-dict=en_gb --spelling-dict=en_us`) (with choice prompts available) are good on the command-line.

That could require refactoring work - I'd be OK looking at developing that as a dependency for this change.

It's possible that the complexity of the results wouldn't be worthwhile, though.  Sometimes simple (even with known limitations) is good.

Pierre-Sassoulas: I think it's best to keep it simple and stupid and have a single field, the same one that pyenchant expect to get.

jayaddison:
I'll continue this at some point in future; I need a bit more time to reflect and to figure out the way ahead with it (I haven't forgotten about it).  That said, I don't mind if someone else wants to run with it.

jayaddison:
Todo:

- [ ] Functional tests
- [ ] Changelog entry

jayaddison:
Ah, I see that [`enchant` is an optional Python package dependency](https://github.com/pylint-dev/pylint/blob/65959d3c310f9da6e1d89ace3c1abb3bed4c8757/pyproject.toml#L56), and not installed during most of the continuous integration tests; that makes sense.  To allow them to pass (reducing review noise, initially), I'll add it as a dependency, although that might not be ideal - it's only likely to be used for a very small number of tests (only one, currently).

jayaddison:
> Ah, I see that [`enchant` is an optional Python package dependency](https://github.com/pylint-dev/pylint/blob/65959d3c310f9da6e1d89ace3c1abb3bed4c8757/pyproject.toml#L56), and not installed during most of the continuous integration tests; that makes sense. To allow them to pass (reducing review noise, initially), I'll add it as a dependency, although that might not be ideal - it's only likely to be used for a very small number of tests (only one, currently).

Hmm.  I'm beginning to learn some of the continuous integration dependencies here, and that's going to take a little while.  Also: if `aspell-en` dictionaries become a required test-time dependency on Linux, then equivalent installation steps might be required on MacOS and Windows (very few tests are skipped, and generally only for `platform == win32` at the moment).

I think I'll roll back the continuous integration changes and the unit test, and then figure out a more thoughtful way to proceed (if I can :)).

jayaddison:
Hold on: shouldn't the `@skip_on_missing_package_or_dict` decorator handle the presence/absence of those dictionary/library dependencies?  Time to check whether that's working..

jayaddison:
> Hold on: shouldn't the `@skip_on_missing_package_or_dict` decorator handle the presence/absence of those dictionary/library dependencies? Time to check whether that's working..

Nope, never mind; it skips if the `en_US` dictionary (loaded at module-level) is missing, but we are using two other dictionaries in the added test case for this branch.  Makes (some) sense.  Perhaps the easiest solution is to re-use the `en_US` dictionary for the test, so that the decorator will skip the test when spellchecking appears unavailable using the existing logic at test-time.

jayaddison:
> > Hold on: shouldn't the `@skip_on_missing_package_or_dict` decorator handle the presence/absence of those dictionary/library dependencies? Time to check whether that's working..
> 
> Nope, never mind; it skips if the `en_US` dictionary (loaded at module-level) is missing, but we are using two other dictionaries in the added test case for this branch. Makes (some) sense. Perhaps the easiest solution is to re-use the `en_US` dictionary for the test, so that the decorator will skip the test when spellchecking appears unavailable using the existing logic at test-time.

I should stop talking.  I've still misinterpreted the situation - we _are_ using `en_US` in that test; for some reason I thought I'd chosen `en_CA` and `en_GB` previously.  However: we're setting the config value twice, and I don't think that that pattern is likely to work any more.

tl;dr - ignore my stream of consciousness debugging, I'll figure it out eventually.

jayaddison:
Failures continue, while most of the tests in `unittest_spelling` are _not_ being skipped (see [this build log from 39bb55f7a0462ed7c9d4513e77f2ef96c1d0f63b on py311/Linux](https://github.com/pylint-dev/pylint/actions/runs/6291497582/job/17079919187?pr=8647#step:7:60), for example).

So: the decorator doesn't appear to be skipping most of the tests, meaning that `enchant` and an `en_US` dictionary are [found on the system](https://github.com/pylint-dev/pylint/blob/39bb55f7a0462ed7c9d4513e77f2ef96c1d0f63b/tests/checkers/unittest_spelling.py#L14-L26).  However the `en_CA` dictionary could be absent, since the [word `colour` is what fails the spellchecking test](https://github.com/pylint-dev/pylint/actions/runs/6291497582/job/17079919187?pr=8647#step:7:192).

Maybe I'm a bit tired and have overlooked something else, but I feel like that's a possibility.  The test passes for me locally, but I do have `aspell-en` installed.

jayaddison:
> Maybe I'm a bit tired and have overlooked something else, but I feel like that's a possibility. The test passes for me locally, but I do have `aspell-en` installed.

Ok, yep: uninstalling `aspell-en` allows me to replicate the failure; so that seems to support the idea that some other fallback EN-US-locale dictionary is found on the system that means that these tests are not being skipped.  The decorator could be adjusted to accept an optional list of dictionaries to check for.

jayaddison:
> Failures continue, while most of the tests in unittest_spelling are _not_ being skipped (see [this build log from 39bb55f7a0462ed7c9d4513e77f2ef96c1d0f63b on py311/Linux](https://github.com/pylint-dev/pylint/actions/runs/6291497582/job/17079919187?pr=8647#step:7:60), for example).

My best guess why this happens is that the `actions/cache` that's being used for many of the test runs here is shared between the spelling tests and non-spelling tests, meaning that `pyenchant` remains installed in the `venv` across commits, and so the spelling-related tests will run during most builds, except for the first one.

There may also be a default `en_US`-language dictionary installed on most of the Ubuntu runners; checking the contents of the `spell_dict.provider` property would be one way to confirm that.

pylint/checkers/spelling.py
```diff
@@ -310,17 +313,23 @@ def open(self) -> None:
...
+            self.spelling_dicts = [
+                enchant.Dict(dict_name) for dict_name in dict_names if dict_name
+            ]
 
         if self.linter.config.spelling_store_unknown_words:
             self.unknown_words: set[str] = set()
 
         self.tokenizer = get_tokenizer(
-            dict_name,
+            tokenization_language,
```
jayaddison, line 332: Arbitrarily selecting the first configured locale to use as a tokenization locale is probably the part of this pull request that I'm most dubious about.

In defense of it, though: we only allow choice of _locales_, not choice of entirely separate _languages_ -- so `en_US, en_GB` is OK, but `en_GB, de_DE` is not valid here.

Do different locales within a language always share the same tokenization rules?  I'm not entirely qualified to say, but my guess from experience of these things is likely "no, there's some edge case locales" -- but hopefully it's relatively rare.

jayaddison: See also [previous discussion](https://github.com/pylint-dev/pylint/issues/8083#issuecomment-1529446843) about this.

jayaddison:
This pull request is a bit of a mess; I'm going to close this down, perhaps do some refactoring of the spelling tests, and then re-approach this again once I have a better idea of what's going on (rather than hacking away at this).

PR #8668, opened 2023-05-09
Fix bug with unclear function signature #3613

### LINKED ISSUE #3613: Pylint gets TensorFlow's tf.split all wrong
### Steps to reproduce
```
import tensorflow as tf
tf.split([0, 0], num_or_size_splits=[1, 1], axis=-1)
print("Done.")
```

### Current behavior
Code runs fine, but
```
bug.py:3:0: E1123: Unexpected keyword argument 'num_or_size_splits' in function call (unexpected-keyword-arg)
bug.py:3:0: E1124: Argument 'axis' passed by position and keyword in function call (redundant-keyword-arg)
bug.py:3:0: E1120: No value for argument 'value' in function call (no-value-for-parameter)
bug.py:3:0: E1120: No value for argument 'num_split' in function call (no-value-for-parameter)
``` 

### Expected behavior
Perfect score

### pylint --version output
```
tensorflow 2.2.0
pylint 2.5.2
astroid 2.4.1
Python 3.8.2 (default, Feb 26 2020, 09:52:54) 
[GCC 7.5.0]
```

hmc-cs-mdrissi:
## Type of Changes

## Description
Fixes long open issue of getting tensorflow function signatures incorrect and producing false positives. Due to import/function definition of tensorflow functions sometimes infer finds multiple function definitions and picks wrong one. safe_infer does check that each function definition has same number of arguments, but it's possible the names/defaults vary. 

Also added a few simple test cases.

Closes #3613

### DIFF
--- a/pylint/checkers/utils.py
+++ b/pylint/checkers/utils.py
@@ -1362,12 +1362,10 @@ def safe_infer(
                 return None
             if (
                 isinstance(inferred, nodes.FunctionDef)
-                and inferred.args.args is not None
                 and isinstance(value, nodes.FunctionDef)
-                and value.args.args is not None
-                and len(inferred.args.args) != len(value.args.args)
+                and inferred.args.format_args() != value.args.format_args()
             ):
-                return None  # Different number of arguments indicates ambiguity
+                return None  # Different arguments indicates ambiguity
     except astroid.InferenceError:
         return None  # There is some kind of ambiguity
     except StopIteration:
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (changes requested):
Looks great ! I think formatting the args make sense (it seems like it could be a performance hit a lot is done in this function). It does simplify the code and fix the issue though... Could you add a fragment for the changelog, please ?

jacobtylerwalls:
> I think formatting the args make sense (it seems like it could be a performance hit a lot is done in this function).

Agree. Let's try just comparing argnames() instead.

jacobtylerwalls:
We're touching `safe_infer()` so let's not backport this.

Pierre-Sassoulas (approved):
The primer looks great !

PR #8674, opened 2023-05-10
Fix a false negative for ``too-many-arguments`` and positional-only and keyword-only arguments

### LINKED ISSUE #8667: R0913 excludes keyword- and positional-only arguments
### Bug description

Pylint's `too-many-arguments` rule seems to exclude keyword- and positional-only arguments from the "total argument" count:

```py
def f(a, b, c, d, e, *, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t):
    ...


def g(a, b, c, d, e, f):
    ...
```

### Configuration

_No response_

### Command used

```shell
pylint a.py
```

### Pylint output

```shell
************* Module a
a.py:5:0: R0913: Too many arguments (6/5) (too-many-arguments)
```

### Expected behavior

My instinct is that it should include keyword- and positional-only arguments, but I'm wondering if this is intentional.

### Pylint version

```shell
pylint 2.17.2
astroid 2.15.3
Python 3.11.1 (main, Dec 23 2022, 09:25:23) [Clang 14.0.0 (clang-1400.0.29.202)]
```

### OS / Environment

_No response_

### Additional dependencies

_No response_

mbyrnepr2:
## Type of Changes

## Description
Currently the logic behind the `too-many-arguments` checker does not consider positional-only and keyword-only arguments. This fix factors those arguments into the equation.

Closes #8667


mbyrnepr2:
The CI shows that the Pylint codebase has had this false negative hidden from sight in the `arguments.py` module:
```python
************* Module pylint.config.argument
[43](https://github.com/pylint-dev/pylint/actions/runs/4940248807/jobs/8831716489?pr=8674#step:7:44)
pylint/config/argument.py:232:4: R0913: Too many arguments (10/9) (too-many-arguments)
[44](https://github.com/pylint-dev/pylint/actions/runs/4940248807/jobs/8831716489?pr=8674#step:7:45)
pylint/config/argument.py:309:4: R0913: Too many arguments (10/9) (too-many-arguments)
[45](https://github.com/pylint-dev/pylint/actions/runs/4940248807/jobs/8831716489?pr=8674#step:7:46)
pylint/config/argument.py:357:4: R0913: Too many arguments (11/9) (too-many-arguments)
[46](https://github.com/pylint-dev/pylint/actions/runs/4940248807/jobs/8831716489?pr=8674#step:7:47)
pylint/config/argument.py:401:4: R0913: Too many arguments (10/9) (too-many-arguments)
[47](https://github.com/pylint-dev/pylint/actions/runs/4940248807/jobs/8831716489?pr=8674#step:7:48)
pylint/config/argument.py:438:4: R0913: Too many arguments (10/9) (too-many-arguments)
```

This code would need a refactor in order to proceed with this merge-request.

Looking at the above I notice an unrelated issue to the change in the current merge-request where `self` is counted as one arguments in the `too-many-arguments` calculation in methods. That seems a bit inconsistent since:

too-many-arguments not emitted:
```python
def name1(param1, param2, param3, param4, param5):
    ...
```

too-many-arguments is emitted:
```python
class A:
    def name2(self, param1, param2, param3, param4, param5):  # R0913: Too many arguments (6/5) (too-many-arguments)
        ...
```

Pierre-Sassoulas:
Let's fix the inconsistencies, but we can disable the existing violations in pylint.

mbyrnepr2:
> Let's fix the inconsistencies, but we can disable the existing violations in pylint.

Ok @Pierre-Sassoulas! The refactoring may take some time to get right so I think that's a pragmatic approach.
Do we agree to fix the inconsistencies in a separate issue? I can create that now.

Pierre-Sassoulas (commented):
Another solution would be to increase the default value and our own value by 2 to take *args / **kwargs into account.

mbyrnepr2:
> Another solution would be to increase the default value and our own value by 2 to take *args / **kwargs into account.

It also crossed my mind to do this but thought against it as we probably want our functions to trend towards the magic number (5 - the current default). Or do you think this default value should now also increase for all us?

mbyrnepr2:
Created [this merge-request](https://github.com/pylint-dev/pylint/pull/8676) since the primer comment is not showing up here.

Pierre-Sassoulas:
Thank you for fixing this, I was wondering why the primer was'nt working in some other MR !

Pierre-Sassoulas:
Regarding the default we could use 7 to reduce disturbance following the false negative fix (we had to add a lot of disable ourselves, the primer could also what to expect.). Also because 7 is the number of concept a human can keep in mind at the same time so it makes sense.

mbyrnepr2:
We are missing 8 packages or so from the primer output because the comment size has reached its limit.
For the ones that are showing I see nothing unusual, the trend is a reduction of messages emitted due to changing the default maximum-parameter number from 5 to 7; Django, for example goes from 196 down to 67 messages emitted for `too-many-arguments`.

Pierre-Sassoulas:
Maybe we can revert 5a6e5bb8f2047a58200765325b07737f0217e2fe in order to be able to see the newly raised messages in the primer ?

Pierre-Sassoulas:
It seems we're fixing much more than what was signaled in the original issue. In Black this was not detected before: https://github.com/psf/black/blob/64887aab032c0fd64f9238cdab6684f2fc0c7f33/src/black/__init__.py#L620 but there was more than 5 values already. We might want to explicitly add the test case so we do not regress later and upgrade the changelog, what do you think ?

jacobtylerwalls:
I'm wondering if we should get more input on the original request. It's not a violation of LSP to accept additional keyword-only arguments in subclasses, right? We already have `arguments-differ` for purists, so maybe we should distinguish `too-many-arguments` in some way like "differs, but preserves LSP". That's a huge number of new messages in well-maintained projects, suggesting a low-value message.

mbyrnepr2:
@Pierre-Sassoulas what the `black` primer you've shared shows is the false negative which this merge-request is fixing. The reason `too-many-arguments` is not emitted on main is because the keyword-parameters (all the parameters after the `*`) were not being used by Pylint in the calculation for total number of parameters.

Here is the relevant debugging on the `main` branch showing why this is so:
```python
************* Module black
> /Users/markbyrne/programming/pylint/pylint/checkers/design_analysis.py(511)visit_functiondef()
-> ignored_argument_names = self.linter.config.ignored_argument_names
(Pdb) node.args.args
[]
(Pdb) node.args
<Arguments l.622 at 0x1121b0160>
(Pdb) node.args.args
[]
(Pdb) node.args.kwonlyargs
[<AssignName.ctx l.622 at 0x1121b01f0>, <AssignName.src l.623 at 0x1121b0220>, <AssignName.quiet l.624 at 0x1121b0250>, <AssignName.verbose l.625 at 0x1121b0280>, <AssignName.include l.626 at 0x1121b02b0>, <AssignName.exclude l.627 at 0x1121b02e0>, <AssignName.extend_exclude l.628 at 0x1121b0310>, <AssignName.force_exclude l.629 at 0x1121b0340>, <AssignName.report l.630 at 0x1121b0370>, <AssignName.stdin_filename l.631 at 0x1121b03a0>]
(Pdb) node.args.posonlyargs
[]
```

Here is the function for reference:
```python
def get_sources(
    *,
    ctx: click.Context,
    src: Tuple[str, ...],
    quiet: bool,
    verbose: bool,
    include: Pattern[str],
    exclude: Optional[Pattern[str]],
    extend_exclude: Optional[Pattern[str]],
    force_exclude: Optional[Pattern[str]],
    report: "Report",
    stdin_filename: Optional[str],
)
```

So, Pylint sees nothing before the `*` and decides that there are no parameters at all in this function.

mbyrnepr2:
@jacobtylerwalls I don't think we are on the same page about the rationale for `too-many-arguments`. I think of it in terms of the function definition itself and the potential for complexity in its body due to handling a large(this is subjective of course) number of parameters.

I think we can always rely on teams configuring the value for `max-args` in their pyproject.toml if it doesn't suit their project.
_pylint isn't smarter than you_ etc.

Some notes on too-many-arguments configuration of our current primer packages which do use Pylint:
```
coverage: 
too-many-arguments
max-args=15

homeassistant:
too-many-arguments disabled

pandas:
too-many-arguments disabled

music21:
too-many-arguments disabled

pygame:
too-many-arguments disabled
```

jacobtylerwalls:
@mbyrnepr2 got it now, thank you! I had forgotten that we don't use the primer project's own pylintrc with respect to enables/disables.

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):

PR #8681, opened 2023-05-13
Add optional `prefer-typing-namedtuple` message

### LINKED ISSUE #8660: Discourage collections.namedtuple in favor of typing.NamedTuple
### Current problem

when using Python 3.8+, collections.namedtuple is archaic, and should be flagged for upgrading

### Desired solution

when collections.namedtuple is seen, recommend using typing.NamedTuple instead

### Additional context

_No response_

jacobtylerwalls:
## Type of Changes

## Description
Closes #8660

### DIFF
--- a/doc/data/messages/p/prefer-typing-namedtuple/good.py
+++ b/doc/data/messages/p/prefer-typing-namedtuple/good.py
@@ -1,0 +1,6 @@
+from typing import NamedTuple
+
+class Philosophy(NamedTuple):
+    goodness: str
+    truth: bool
+    beauty: float
--- a/doc/whatsnew/fragments/8660.extension
+++ b/doc/whatsnew/fragments/8660.extension
@@ -1,0 +1,5 @@
+Add new ``prefer-typing-namedtuple`` message to the ``CodeStyleChecker`` to suggest
+rewriting calls to ``collections.namedtuple`` as classes inheriting from ``typing.NamedTuple``
+on Python 3.8+.
+
+Closes #8660
--- a/pylint/extensions/code_style.py
+++ b/pylint/extensions/code_style.py
@@ -69,6 +69,13 @@ class CodeStyleChecker(BaseChecker):
                 "default_enabled": False,
             },
         ),
+        "R6105": (
+            "Prefer typing.NamedTuple over collections.namedtuple",
+            "prefer-typing-namedtuple",
+            "Emitted when a call to collections.namedtuple() is detected. "
+            "This can be rewritten more elegantly as a class with type annotations. "
+            "Requires Python 3.8 and ``py-version >= 3.8``.",
+        ),
     }
     options = (
         (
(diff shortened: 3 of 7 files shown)


DanielNoord:
Looking at the primer, are we sure we want to add this while we can't disable messages by default? This feels like it should be `opt-in` even if you have already enabled the plug-in.

pylint/extensions/code_style.py
```diff
@@ -69,6 +69,17 @@ class CodeStyleChecker(BaseChecker):
                 "default_enabled": False,
             },
         ),
+        "R6105": (
+            "Prefer typing.NamedTuple over collections.namedtuple",
+            "prefer-typing-namedtuple",
+            "Emitted when a call to collections.namedtuple() is detected. "
+            "This can be rewritten more elegantly as a class with type annotations. "
```
Pierre-Sassoulas, line 76:
```suggestion
            "'typing.NamedTuple' permits to add type hints, and use the "
            "well known 'class' keyword (it's also faster as it does not "
            "require an internal exec call)."
```
See [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple) for a way to add type hints for named tuples. It also provides an elegant notation using the [class](https://docs.python.org/3/reference/compound_stmts.html#class) keyword:

pylint/extensions/code_style.py
```diff
@@ -69,6 +69,17 @@ class CodeStyleChecker(BaseChecker):
                 "default_enabled": False,
             },
         ),
+        "R6105": (
+            "Prefer typing.NamedTuple over collections.namedtuple",
```
Pierre-Sassoulas, line 73:
```suggestion
            "Prefer 'typing.NamedTuple' over 'collections.namedtuple'",
```

doc/whatsnew/fragments/8660.extension
```diff
@@ -0,0 +1,5 @@
+Add new ``prefer-typing-namedtuple`` message to the ``CodeStyleChecker`` to suggest
+rewriting calls to ``collections.namedtuple`` as classes inheriting from ``typing.NamedTuple``
+on Python 3.8+.
```
Pierre-Sassoulas, line 3:
```suggestion
on Python 3.8+.

Requires ``load-plugins=pylint.extensions.code_style`` and ``enable=prefer-typing-namedtuple`` to be raised.
```

Pierre-Sassoulas (changes requested):
👌, some nit about the documentation. I know historically the message often say "emitted when blablabla", but the rational is a lot more important imo. (And I don't feel like the description of when we emit the message is very helpful in the doc as we have the good/bad example already.)

doc/data/messages/p/prefer-typing-namedtuple/good.py
```diff
@@ -0,0 +1,7 @@
+from typing import NamedTuple
+
+
+class Philosophy(NamedTuple):
+    goodness: str
+    truth: bool
+    beauty: float
```
Pierre-Sassoulas, line 7: A link for related.rst: https://docs.python.org/3/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields (in the "see also")

jacobtylerwalls: What about linking to `typing.NamedTuple` instead?

jacobtylerwalls:
I think the version needs to be 3.6+, not 3.8+. I was just going off the report.

doc/data/messages/p/prefer-typing-namedtuple/good.py
Pierre-Sassoulas: Yeah, that make sense. Or both, I'll let you choose :)

Pierre-Sassoulas (approved):

PR #8682, opened 2023-05-13
Add regression test for #8554

jacobtylerwalls:
## Type of Changes

|     | Type                   |
| --- | ---------------------- |
| ✓   | :bug: Bug fix          |

## Description

Closes #8554

Depends on https://github.com/pylint-dev/astroid/pull/2177

PR #8685, opened 2023-05-14
Upgrade astroid to 3.0.0a3

Pierre-Sassoulas:
## Type of Changes

## Description

Refs #8682

### DIFF
--- a/pylint/checkers/modified_iterating_checker.py
+++ b/pylint/checkers/modified_iterating_checker.py
@@ -93,7 +93,7 @@ class ModifiedIterationChecker(checkers.BaseChecker):
             self.add_message(
                 msg_id,
                 node=node,
-                args=(iter_obj._repr_name(),),
+                args=(iter_obj.repr_name(),),
                 confidence=interfaces.INFERENCE,
             )
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies    = [
     # Also upgrade requirements_test_min.txt.
     # Pinned to dev of second minor update to allow editable installs and fix primer issues,
     # see https://github.com/pylint-dev/astroid/issues/1341
-    "astroid>=3.0.0a2,<=3.1.0-dev0",
+    "astroid>=3.0.0a3,<=3.1.0-dev0",
     "isort>=4.2.5,<6",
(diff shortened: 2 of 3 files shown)


jacobtylerwalls:
Thanks, I'll push some commits to this branch. 👍

jacobtylerwalls:
We should bump all the way to astroid 3.0.0a4, because the primer will be busted without pylint-dev/astroid#2181.

jacobtylerwalls:
> Re: the remaining failure: I remember looking at this a couple weeks ago. The test passes in isolation, so there's a pollution issue. I haven't made it much farther.

See pylint-dev/astroid#2185 for a fix.

jacobtylerwalls:
Just waiting on astroid 3.0.0a4

jacobtylerwalls (approved):

jacobtylerwalls:
I'm looking into the PyPy 3.8 failure, but it doesn't look serious. Thinking about writing a follow-up issue (and xfailing it with a link)

jacobtylerwalls:
So we have a test `regression_5408` extremely similar to `test_recursion_on_inference_tip` in astroid.

The failure of it on PyPy 3.8 is transient; rerunning it passed eventually. This is not stable enough for CI, so we should add the same skip we added in astroid, see https://github.com/pylint-dev/astroid/pull/2181#issuecomment-1547116406 and pylint-dev@1f9ba55a1bbe03763095a66bc59450d680338ab1.

Where exactly the failure is happening is interesting. We're correctly catching the RecursionError and reraising as InferenceError, but chaining it to RecursionError as the cause is becoming problematic, maybe because InferenceError is bloated with all kinds of references to nodes.

I'm going to experiment by only chaining the exceptions on versions 3.9+ (and push the commit to this branch).

jacobtylerwalls:
> I'm going to experiment by only chaining the exceptions on versions 3.9+ (and push the commit to this branch).

Oof, I'd have to do this in astroid. Will come back with a better proposal.

jacobtylerwalls (approved):

jacobtylerwalls:
I skipped the test on Python 3.8. I mentioned in the commit message for the skip that the file lints fine on python 3.8, rather it's pytest that's choking. Not necessary to look into at this juncture IMO.

PR #8697, opened 2023-05-17
[primer] Parallelize primer workflow into package batches

jacobtylerwalls:
## Type of Changes

## Description
The primer run now completes in [half the time](https://github.com/jacobtylerwalls/pylint/actions/runs/5003019925) (53m -> 28m) by dividing primer packages into batches.

The upside mainly consists of separating `home-assistant` from `pandas`, the two largest projects.

## Motivations
- Better development experience, shorter feedback loops
- Primer tool less likely to crash with out-of-memory issues trying to write huge .txt files (see https://github.com/pylint-dev/pylint/pull/8663#pullrequestreview-1427118576)
- Makes it more pleasant to debug multiprocessing issues (getting the primer to work with --jobs=0 would then provide more benefit)

## Notes
- We could shave a few more seconds off if `home-assistant` and `pandas` are given their own exclusive runs, but the few seconds it takes to lint another package in a batch after `pandas` is not substantial. My hope is that this is good enough and more maintainable.
-  Since the new arguments are optional, this is backwards-compatible with the existing way of running the primer locally without batching.


.github/workflows/primer_run_pr.yaml
```diff
@@ -33,12 +33,14 @@ permissions:
...
   run-primer:
-    name: Run / ${{ matrix.python-version }}
+    name: Run / ${{ matrix.python-version }} / batch index ${{ matrix.batchIdx }}
     runs-on: ubuntu-latest
-    timeout-minutes: 120
+    timeout-minutes: 45
     strategy:
       matrix:
         python-version: ["3.8", "3.11"]
+        batches: [4]
```
DanielNoord, line 42: Guess we can't do `len(batchIdx)` or something like that?

.github/workflows/primer_comment.yaml
```diff
@@ -103,8 +103,9 @@ jobs:
           . venv/bin/activate
           python tests/primer/__main__.py compare \
           --commit=${{ github.event.workflow_run.head_sha }} \
-          --base-file=output_${{ steps.python.outputs.python-version }}_main.txt \
-          --new-file=output_${{ steps.python.outputs.python-version }}_pr.txt
+          --base-file=output_${{ steps.python.outputs.python-version }}_main_BATCHIDX.txt \
```
DanielNoord, line 106: Have you tested that the downloading of artifacts and unzipping works? As in, do the different batch jobs get zipped into one single file?

pylint/testutils/_primer/primer_run_command.py
```diff
@@ -30,14 +30,22 @@ class RunCommand(PrimerCommand):
     def run(self) -> None:
         packages: PackageMessages = {}
         fatal_msgs: list[Message] = []
-        for package, data in self.packages.items():
+        package_data_iter = (
+            self.packages.items()
+            if self.config.batches is None
+            else list(self.packages.items())[
+                self.config.batchIdx :: self.config.batches
```
DanielNoord, line 37: Smart way to iterate this!

pylint/testutils/_primer/primer_run_command.py
```diff
@@ -30,14 +30,22 @@ class RunCommand(PrimerCommand):
...
+        for package, data in package_data_iter:
             messages, p_fatal_msgs = self._lint_package(package, data)
             fatal_msgs += p_fatal_msgs
             local_commit = Repo(data.clone_directory).head.object.hexsha
             packages[package] = PackageData(commit=local_commit, messages=messages)
         path = (
             self.primer_directory
-            / f"output_{'.'.join(str(i) for i in sys.version_info[:3])}_{self.config.type}.txt"
+            / f"output_{'.'.join(str(i) for i in sys.version_info[:3])}_{self.config.type}"
+            f"_batch{self.config.batchIdx}.txt"
```
DanielNoord, line 48: Should we add a check here whether there are batches defined? Otherwise this name doesn't make a lot of sense.

DanielNoord (commented):
Thanks for doing this! Very nice!

.github/workflows/primer_run_pr.yaml
jacobtylerwalls: Yeah, unless we add a step to use bash or python to do math, I don't think I can do it inside a string formatting expression.

.github/workflows/primer_comment.yaml
jacobtylerwalls: The 4 files get [zipped up](https://github.com/jacobtylerwalls/pylint/actions/runs/5003019925) into one .zip, and when you download it you can examine the four files.

pylint/testutils/_primer/primer_run_command.py
jacobtylerwalls: Right, it will just say `_batchNone` as of now. I can clean this up a little.

jacobtylerwalls: b94d0131

Pierre-Sassoulas (commented):
Looks pretty good ! I think the python parts could be tested in ``tests/testutils/``. (This would be reused in a pylint-testutil package accessible to all pylint plugins maintainer).

Pierre-Sassoulas:
I think we had issues with primers in the past that would not have happened if the python parts were covered from the get go (even if it's impossible to test the github actions code). That being said, the goal of this MR is not to add tests for everything that is not tested at the moment (only the new batch processing mechanism). I'd approve anyway if you don't want to invest the time this is clearly an improvement compared to what we have right now.

jacobtylerwalls:
@Pierre-Sassoulas @DanielNoord Does anyone know why the `prettier` pre-commit hook is failing?

Pierre-Sassoulas:
It seems as if prettier was outputting the path of every file in the repository without separators ?

Pierre-Sassoulas (approved):
LGTM. I fixed pre-commit but the tests now fails (?)

jacobtylerwalls:
> I fixed pre-commit but the tests now fails (?)

Just looks like a transient failure from the Action we're using.

Thanks for the reviews!

PR #8707, opened 2023-05-21
[config] Make the linter optional during config parsing

Pierre-Sassoulas:
## Type of Changes

## Description

This will permit to parse config file without having to instantiate a linter for tool working on pylint configuration like #5462. another way to do this would be to extract the logic and create another class that we could use in ``_ConfigurationFileParser`` like here: https://github.com/Pierre-Sassoulas/pylint/commit/4ccbee1d1e73900f2d34dd67c00d5eb665cf2a04

### DIFF
--- a/pylint/config/config_file_parser.py
+++ b/pylint/config/config_file_parser.py
@@ -28,9 +28,9 @@ class _RawConfParser:
 
     @staticmethod
     def parse_ini_file(file_path: Path) -> tuple[dict[str, str], list[str]]:
-        """Parse and handle errors of a ini configuration file.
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas:
I think I like the alternate way more. Cleaner final result it's harder to badly instantiate the classes.

DanielNoord:
Yes, let's go for the alternative! Nice solution.

One early review comment, let's make the new class private (or perhaps even just make the whole file private by underscoring it). I never want to deal with `DeprecationWarnings` ever again :smile:

DanielNoord (approved):

PR #8708, opened 2023-05-21
Docs: optparse not removed in Python 3.12

hugovk:
## Type of Changes

## Description

The optparse module is deprecated, and it's better to use argparse, but there's no current plans to remove it.

https://docs.python.org/3.12/library/optparse.html shows it's still present in 3.12 and says:

> _Deprecated since version 3.2:_ The [optparse](https://docs.python.org/3.12/library/optparse.html#module-optparse) module is deprecated and will not be developed further; development will continue with the [argparse](https://docs.python.org/3.12/library/argparse.html#module-argparse) module.

### DIFF
--- a/doc/whatsnew/2/2.14/summary.rst
+++ b/doc/whatsnew/2/2.14/summary.rst
@@ -15,8 +15,7 @@ We migrated to ``argparse`` from ``optparse`` and refactored the configuration h
 thanks to Daniël van Noord. On the user side it should change the output of the
 ``--help`` command, and some inconsistencies and bugs should disappear. The behavior
 between options set in a config file versus on the command line will be more consistent. For us,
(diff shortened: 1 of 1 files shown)


Pierre-Sassoulas (approved):
I wonder where I saw that it was going to be removed. Thank you for the correction.

hugovk:
Maybe in the [PEP 594](https://peps.python.org/pep-0594/) ("Removing dead batteries from the standard library") discussions, but they decided to keep it:

* https://peps.python.org/pep-0594/#optparse

DanielNoord:
They just announced the intent to keep it after we had moved all of our own code to `argparse` :sweat_smile:

Pierre-Sassoulas:
I don't regret the move though, argparse is better and the intertwining of PyLinter and OptParser was pretty nasty 😄

PR #8717, opened 2023-05-26
CI: Only run stale cron on upstream

hugovk:
## Type of Changes

## Description

There's no need to run the "close stale issues and pr" or CodeQL cron for forks.

More info: https://dev.to/hugovk/til-how-to-disable-cron-for-github-forks-2d0l

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -29,6 +29,8 @@ concurrency:
 
 jobs:
   analyze:
+    if:
+      ${{ github.repository_owner == 'pylint-dev' || github.event_name != 'schedule' }}
     name: Analyze
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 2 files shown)


Pierre-Sassoulas (commented):
Thank you ! I think we can also remove the ``debug-only: true`` flag and actually run the job for real seeing that only two PR would be staled : https://github.com/pylint-dev/pylint/actions/runs/5086370028/jobs/9140773077 (I'm not sure how to see which PR would be stalled though.) What do you think @mbyrnepr2 ?

mbyrnepr2:
Sure @Pierre-Sassoulas let’s try it for real👍

mbyrnepr2:
By the way I haven’t looked into the primer failure, although that needn’t be a blocker for this merge request.

mbyrnepr2 (approved):

PR #8720, opened 2023-05-27
Avoid `consider-using-f-string` on modulos with brackets in template

progval:
## Type of Changes

## Description

Brackets can be inconvenient in f-string expressions.

For example, `"{%s}%s" % (namespace, tag)` (idiomatic when using ElementTree) would be `f"{{{namespace}}}{tag}"` as a f-string


Pierre-Sassoulas (commented):
I think it make sense, I'll wait for another opinion before merging.

progval:
hmm, what should I do about this failing test?

https://github.com/pylint-dev/pylint/blob/c5f3ff4c37a2b5f50e6bca2529f9290f8a58f592/tests/functional/r/regression_02/regression_8109.py#L14

Pierre-Sassoulas:
Hmm it seems #8109 is a regression tests about a crash and not about actually raising ``consider-using-f-string`` in this case. So I would be okay not raising it anymore. Something to consider during review though.

Pierre-Sassoulas:
@progval do you think you'll be able to upgrade the tests to make the CI succeed ? :)

progval:
Sorry, thanks for the reminder. Done now

jacobtylerwalls (approved):

jacobtylerwalls:
Thanks for the contribution!

Pierre-Sassoulas:
Congratulation on becoming a pylint contributor, the primer result looks pretty good on this one, that's a lot of false positives avoided ! 💪

PR #8721, opened 2023-05-29
Extend test cases for 'condition-evals-to-constant'

### LINKED ISSUE #7933: Extend "using-constant-test" to checks involving boolean operators
### Current problem

We have recently found a case in which Pylint could help.
Specifically, we had this check:

```python
if get_text() == "string A":
    pass
```

That we wanted to extend for "string B".
A bug was then introduced, in this  form:
```python
if get_text() == "string A" or "string B":
    pass
```
which always evaluate to True.

I tried to see whether "using-constant-text" could be extended to cater for this type of bugs, but I think it might not be that easy - or at all feasible.
In case it is, I'd be happy to contribute, but I guess I will need some pointers.

### Desired solution

Ideally a warning should appear to highlight the fact that the comparison is not what the author likely wants.

### Additional context

_No response_

crazybolillo:
Closes #7933.

The original issue reported pylint as missing bugs in comparisons like:

`if get_text() == "string A" or "string B":`

In reality this is already covered by the message
'condition-evals-to-constant', however test cases with strings were missing, they have been added to self-document other cases this checker applies to and extend coverage.

### DIFF
--- a/tests/functional/c/condition_evals_to_constant.py
+++ b/tests/functional/c/condition_evals_to_constant.py
@@ -44,3 +44,10 @@ if True:  # pylint: disable=using-constant-test
 CONSTANT or True
 bool(CONSTANT or OTHER)
 bool(func(CONSTANT or True))
+
+# Strings also evaluate as True (they are constants)
+if func("a") == "b" or "c":  # [condition-evals-to-constant]
+    pass
+
+if 1 == func(2) or "fermi":  # [condition-evals-to-constant]
+    pass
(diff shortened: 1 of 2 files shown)


crazybolillo:
Hello @Pierre-Sassoulas 

Apparently the message already catches the cases mentioned by @cfax, I noticed this as soon as soon as I started work by extending the test cases so there is not much to do.

Anyways I am making the PR, at the very least it gets me acquainted with pylint's CI for future contributions of mine (hopefully).

Thanks for your time.

Pierre-Sassoulas (approved):
Thank you ! Glad to see you plan to contribute :) Should we close #7933 then ?

crazybolillo:
Yes, seems fine to me, I can't think of any other edge cases we should add. Thanks :)

Pierre-Sassoulas:
Congratulations on becoming a pylint contributor :)

PR #8728, opened 2023-05-31
Fix false positive `unspecified-encoding` when using kwargs

### LINKED ISSUE #8719: False positive `unspecified-encoding` when passing encoding as kwarg
### Bug description

When passing an `encoding` as a kwarg to `open`, I get `unspecified-encoding` thrown.

```python
CSV_KWARGS = {"newline": "", "encoding": "utf-8"}
with open("foo.csv", **CSV_KWARGS):
    pass
```

### Configuration

_No response_

### Command used

```shell
pylint --score=false a.py
```

### Pylint output

```shell
************* Module quick_play
a.py:2:5: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
```

### Expected behavior

I think it would be cool if `pylint` wouldn't throw `unspecified-encoding` in this case.

### Pylint version

```shell
pylint 2.17.4
astroid 2.15.4
Python 3.10.10 (main, Feb 19 2023, 17:57:18) [Clang 14.0.0 (clang-1400.0.29.102)]
```

### OS / Environment

macOS Monterey v12.6

### Additional dependencies

_No response_

crazybolillo:
False positives were being generated when passing arguments as kwargs to open() calls. This has been resolved by using inference inside `get_argument_from_call`.

This is an opt-in function, for inference to be used, `infer=True` must be specified when calling `get_argument_from_call`. This ensures that the rest of the codebase does not suffer a performance hit unnecessarily.

Closes #8719

### DIFF
--- a/tests/functional/u/unspecified_encoding_py38.py
+++ b/tests/functional/u/unspecified_encoding_py38.py
@@ -162,3 +162,35 @@ open(FILENAME, mode=None)  # [bad-open-mode, unspecified-encoding]
 
 # Test for crash reported in https://github.com/pylint-dev/pylint/issues/6414
 open('foo', mode=2)  # [bad-open-mode, unspecified-encoding]
+
+# Infer kwargs
+KWARGS = {"mode": "rb"}
+open(FILENAME, **KWARGS)
+
+KWARGS = {"mode": "w", "encoding": "utf-8"}
+open(FILENAME, **KWARGS)
+io.open(FILENAME, **KWARGS)
+Path(FILENAME).open(**KWARGS)
+
+KWARGS = {"mode": 5}
+open(FILENAME, **KWARGS)  # [bad-open-mode, unspecified-encoding]
+io.open(FILENAME, **KWARGS)  # [bad-open-mode, unspecified-encoding]
+
+KWARGS = {"mode": "wt", "encoding": None}
+with open(FILENAME, **KWARGS) as fd:  # [unspecified-encoding]
+    pass
+
+Path(FILENAME).open(**KWARGS)  # [unspecified-encoding]
+
+KWARGS = {"encoding": None}
+Path(FILENAME).write_text("hello", **KWARGS)  # [unspecified-encoding]
+
+KWARGS = {"encoding": "utf-8"}
+Path(FILENAME).write_text("goodbye", **KWARGS)
+
+# No one does it this way, but it is a possibility
+KWARGS = {"encoding": None}
+Path(FILENAME).read_text(**KWARGS)  # [unspecified-encoding]
+
+KWARGS = {"encoding": "utf-8"}
+Path(FILENAME).read_text(**KWARGS)
--- a/tests/functional/u/unspecified_encoding_py38.txt
+++ b/tests/functional/u/unspecified_encoding_py38.txt
@@ -1,33 +1,41 @@
-unspecified-encoding:13:0:13:14::Using open without explicitly specifying an encoding:UNDEFINED
-unspecified-encoding:14:0:14:20::Using open without explicitly specifying an encoding:UNDEFINED
-unspecified-encoding:15:0:15:20::Using open without explicitly specifying an encoding:UNDEFINED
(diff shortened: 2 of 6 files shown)


tests/functional/u/unspecified_encoding_py38.py
```diff
@@ -162,3 +162,10 @@ class IOArgs:
 
 # Test for crash reported in https://github.com/pylint-dev/pylint/issues/6414
 open('foo', mode=2)  # [bad-open-mode, unspecified-encoding]
+
+# Infer kwargs
+KWARGS = {"mode": "rb"}
+open(FILENAME, **KWARGS)
```
Pierre-Sassoulas, line 168:
```suggestion
open(FILENAME, **KWARGS)  # [unspecified-encoding]
```
This should raise, right ?

Pierre-Sassoulas: Ha, right. Could we add the same use cases than for the base check then ?

crazybolillo: Do you mean adding test cases similar to existing ones but for inferred kwargs (e.g. with no encoding or "encoding": None)

Pierre-Sassoulas: Yes

tests/functional/u/unspecified_encoding_py38.txt
```diff
@@ -31,3 +31,11 @@ bad-open-mode:161:0:161:25::"""None"" is not a valid mode for open.":UNDEFINED
 unspecified-encoding:161:0:161:25::Using open without explicitly specifying an encoding:UNDEFINED
 bad-open-mode:164:0:164:19::"""2"" is not a valid mode for open.":UNDEFINED
 unspecified-encoding:164:0:164:19::Using open without explicitly specifying an encoding:UNDEFINED
+bad-open-mode:176:0:176:24::"""5"" is not a valid mode for open.":UNDEFINED
```
Pierre-Sassoulas, line 34: Here the confidence level should be INFERENCE for all the new case, and be HIGH for the existing cases. It means we're going to have to get that information from ``get_argument_from_call``. Maybe it's an opportunity to not modify ``get_argument_from_call`` signature and do things differently because changing the return would complexify the design too much ? We might want to add the confidence HIGH in another MR first if it's harder than simply adding the parameter to ``add_message`` (I did not check).

tests/functional/u/unspecified_encoding_py38.py
Pierre-Sassoulas: Yes let's add a changelog entry. It's a false positive fix and we try to release them early in the maintenance branch.

Pierre-Sassoulas (changes requested):
Thank you for adding the test cases, pretty thorough !

tests/functional/u/unspecified_encoding_py38.txt
crazybolillo: Interesting, I will research into adding CONFIDENCE levels and get back

Pierre-Sassoulas: See this snippet:

https://github.com/pylint-dev/pylint/blob/6bfa0402fe1aaf3db8b40b30afa979eb0156aa84/pylint/checkers/bad_chained_comparison.py#L51-L56

crazybolillo: Thanks, I have added the confidence levels, I had to change the bad_mode.txt as well since now it outputs HIGH confidence as well.

jacobtylerwalls (approved):

Pierre-Sassoulas (approved):
Sorry for the delay, just realized it was ready when seeing Jacob's review :)

Pierre-Sassoulas (approved):

Pierre-Sassoulas:
Thank you, great contribution !

jamesbraza:
Thank you all!! 🥳 🎉

crazybolillo:
Thanks to you guys as well for the help :handshake:

Issue #2493, opened 2018-09-15
Add support for noqa: ERROR MESSAGE


PCManticore:
### Is your feature request related to a problem? Please describe
Yes. The problem is that every linter out there has its own pragma control, leading to a mess if you want to support bandit, pylint, pyflakes, flake8, mypy:
`… # type: … # noqa: E501 # pylint: disable=line-too-long # nosec`

I think it would help if we'd also support `noqa` with an error message (not bare `noqa` which I'm still against using in `pylint`). It could also help with the adoption as well with keeping your codebase linter agnostic if you plan to use just `pylint` and `flake8`.

### Describe the solution you'd like

Support `noqa: error message` so `noqa: no-member` or `noqa: E1101` in `pylint`.

tuukkamustonen:
Would it need some extra identifier in `noqa: E1101` to indicate that we are about to ignore a *pylint* error and not error from some other tool? Is there really no overlap between pylint, pycodestyle, pydocstyle, flake8, others?

The-Compiler:
FWIW I love this idea simply because `# noqa: no-member` is much shorter than `# pylint: disable=no-member`, so there are higher chances of easily fitting it after a line of code with a conservative line width limit.

edran:
Could you not add a namespace to differentiate between `flake8` and `pylint` errors? 

For instance instead of parsing the error codes directly, you could match codes that look like `pylint-errorcode`, so that you could then mix the errors at will. 

I.e.: 

`# noqa: flake8error, pylint-error`

PCManticore:
I don't understand what you mean. Are you referring to add a `pylint-` or `flake8-` prefix to the error codes? That unfortunately doesn't seem like a big of change than the current situation where you'd use `# pylint` instead and it would probably be a larger discussion for the whole PyCQA org to find a common ground on the `noqa` support. What this current issue tries to accomplish is to allow folks that already use `noqa` to switch to `pylint` without too much hassle.

lokesh1729:
👍+1

AWhetter:
Upon further discussion, is does not seem possible for pylint to introduce `#noqa` support without breaking things in flake8, and making maintenance more difficult in the future for both projects. Therefore I'm closing this as something that we are unable to implement.

socketpair:
So, just how to use noqa and pylint: disable=xxxxx on the same line ?!

rachmadaniHaryono:
[...]

PR #8731, opened 2023-06-01
Enable ``deprecated-pragma`` by default.

### LINKED ISSUE #8725: `pylint: disable=all` comment is applied to the whole file
### Bug description

Consider the following example:

```python
"""Module level docstring"""

def enable_here():
    """This function should have a warning."""
    x = "bad name"
    return x


def disable_here():
    """No pylint warnings in this function, please."""
    # pylint: disable=all
    x = "bad name"
    return x
```

For some reason, the `disable=all` comment disables the warnings for the **whole file**, instead of only for the current block/current line/next line/etc. Compare this with the exact same file, but with `disable=invalid-name` instead of `disable=all`.

### Configuration

None

### Command used

```shell
pylint test.py
```

### Pylint output

```shell
(no output)
```

### Expected output

```
************* Module test
test.py:5:4: C0103: Variable name "x" doesn't conform to snake_case naming style (invalid-name)
```

### Pylint version

```shell
pylint 2.17.4
astroid 2.15.5
Python 3.10.10 (main, Mar  5 2023, 22:26:53) [GCC 12.2.1 20230201]
```

### OS / Environment

Manjaro (Arch Linux)

### Additional dependencies

```
astroid==2.15.5
dill==0.3.6
isort==5.12.0
lazy-object-proxy==1.9.0
mccabe==0.7.0
platformdirs==3.5.1
pylint==2.17.4
tomli==2.0.1
tomlkit==0.11.8
typing_extensions==4.6.2
wrapt==1.15.0
```

mbyrnepr2:
## Type of Changes

## Description
Enable ``deprecated-pragma`` by default.

Closes #8725

### DIFF
--- a/doc/whatsnew/fragments/8725.bugfix
+++ b/doc/whatsnew/fragments/8725.bugfix
@@ -1,0 +1,3 @@
+Enable ``deprecated-pragma`` by default.
+
+Closes #8725
--- a/pylint/lint/message_state_handler.py
+++ b/pylint/lint/message_state_handler.py
@@ -388,6 +388,7 @@ class _MessageStateHandler:
                                 "deprecated-pragma",
                                 line=start[0],
                                 args=("disable=all", "skip-file"),
+                                confidence=HIGH,
                             )
                             self.linter.add_message("file-ignored", line=start[0])
                             self._ignore_file = True
--- a/pylint/lint/pylinter.py
+++ b/pylint/lint/pylinter.py
@@ -196,7 +196,6 @@ MSGS: dict[str, MessageDefinitionTuple] = {
         {
             "old_names": [("I0014", "deprecated-disable-all")],
             "scope": WarningScope.LINE,
-            "default_enabled": False,
         },
     ),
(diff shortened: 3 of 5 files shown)


Pierre-Sassoulas:
I think the removal can be automated as part of #5462 and cleanup and removal of deprecated disable see https://github.com/pylint-dev/pylint/issues/2493#issuecomment-1510491501.

Making it a default check before we're able to automate the migration would create a lot of unnecessary work for user imo.

mbyrnepr2:
> I think the removal can be automated as part of #5462 and cleanup and removal of deprecated disable see [#2493 (comment)](https://github.com/pylint-dev/pylint/issues/2493#issuecomment-1510491501).
> 
> Making it a default check before we're able to automate the migration would create a lot of unnecessary work for user imo.

Right thanks for that note @Pierre-Sassoulas. I didn't consider the wider context of that ongoing work you've mentioned. Let's close the issue with the rationale that this would be superseded by that design change.

PR #8732, opened 2023-06-01
Unused function call extension

### LINKED ISSUE #7935: FR: report unused function return values
### Current problem

This is a request to implement a check similar to the `[[nodiscard]]` function attribute in C++. A typical mistake is when people assume that the method modifies the current object while it actually creates and returns a copy of self. Calling this method without using the return value doesn't make sense and could be easily caught.

An example with sqlalchemy. Instead of:

```
if (id):
    query.filter(row.id == id)

```
You should write:

```
if (id):
    query = query.filter(row.id == id)

```

The problem is the first version looks legit: query.filter reads as "filter the query", so it's hard to spot this mistake. I've seen even experienced sqlalchemy users fall into this trap once in a while.

### Desired solution

I'd suggest something like `# pylint: enable=unused-return-value` pragma for functions with a subsequent analysis of the call site.

There's already a very similar warning `expression-not-assigned / W0106`, but it specifically ignores the function calls. Which makes sense because in many cases functions have side effects besides just the return value. In some cases though they're limited to the return value and it would be nice to catch these.

### Additional context

https://stackoverflow.com/questions/51858215/how-to-make-pylint-report-unused-return-values
https://stackoverflow.com/questions/74785720/python-prevent-discarding-of-the-function-return-value

LukasKrocek:
## Type of Changes

## Description

Adding extension to check for unassigned non-nullable return values from function/method calls. This update encourages proper handling of returned values, which is especially handy when working with immutable classes and collections.

Closes #7935


Pierre-Sassoulas (changes requested):
Thank you for opening the pull request :)

Could you add a changelog entry (https://github.com/pylint-dev/pylint/blob/main/doc/whatsnew/fragments/5159.new_check), an example of bad/good code in the documentation and cover the missing line please (https://github.com/pylint-dev/pylint/tree/main/doc/data/messages/c/comparison-of-constants) ? There's an easier way to cover a lot of use cases with functional tests, see https://github.com/pylint-dev/pylint/blob/main/tests/functional/ext/check_elif/check_elif.py for example.

pylint/extensions/unassigned_function_call.py
```diff
@@ -0,0 +1,91 @@
...
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class UnusedReturnValueChecker(BaseChecker):
+    name = "unused_return_value"
+    msgs = {
+        "W5486": (
+            "Function returned value which is never used",
+            "unused-return-value",
```
Pierre-Sassoulas, line 27: I just realized that we have a message for expression that aren't assigned : https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/expression-not-assigned.html. Maybe grouping those two in the same checker and calling this one the same way would make sense:
```suggestion
            "function-return-not-assigned",
```

Pierre-Sassoulas: Relevant issues:
- https://github.com/pylint-dev/pylint/issues/8129
- https://github.com/pylint-dev/pylint/issues/2380 
 
We might want to check some ternary construct.

LukasKrocek: Thanks for the review. Docs and examples added, switched to functional tests, message was renamed to `function-return-not-assigned`.

tests/functional/ext/function_return_not_assigned/function_return_not_assigned.py
```diff
@@ -0,0 +1,56 @@
...
+def func_with_no_explicit_return():
+    print("I am doing something")
+
+
+func_that_returns_none()
+func_with_no_explicit_return()
+
+some_var = ""
+# next line should probably raise?
+func_that_returns_something() if some_var else func_that_returns_none()
```
LukasKrocek, line 30: parent of all three `func_that_returns_something` `some_var` and `func_that_returns_none` is IfExpr. If `some_var` would be function call, I would consider it to be used while `func_that_returns_something` `func_that_returns_none` not. `func_that_returns_something` should raise `function-return-not-assigned`

Pierre-Sassoulas (changes requested):
Thank you for your work on this. I think it need to take https://github.com/pylint-dev/pylint/issues/7935#issuecomment-1350733948 into account, seeing the truncated primer result, there's indeed too much false positive otherwise.

DanielNoord:
Considering the review here was inconclusive this is also not really ready for take over.

Issue #4813, opened 2021-08-08
Take type information into account instead of inferring


Pierre-Sassoulas:
### Current problem

Right now pylint relies on its own inference engine, which ultimately is somewhat basic. This leads to quite a lot of false positives for deep inspections such as no-member, since what the entire inference is doing is just solving the variables into what values they might represent. In some cases it can properly infer the true value, in other cases we just infer a None or raise an inference error. This leads to quite a few false positives in which we cannot properly deduce the actual types due to the limited inference capabilities.

### Desired solution

While mypy is a fully fledged type checker, in our case we only need some support of type inference for function signatures and assignment signatures as well, in which case we could assume the type passed from the type signature instead of relying on our inference. This could also mean better performance, as we won't have to walk the tree calling the inference to figure out the type. A similar project that works in the same vein is jedi which doesn't need to do full type analysis as mypy is doing.

### Additional context

Taken from the comment from @PCManticore [here](https://github.com/PyCQA/pylint/issues/647#issuecomment-529359236)

DanielNoord:
I'm not sure we should actually do this. How would we handle incorrectly typed functions/variables? We would need to run `mypy` ourselves to make sure the typing is correct.

The biggest benefit of this would be that you can explicitly type stuff that `pylint` incorrectly infers right now, often due to control flow inference. I think we should focus our efforts on improving the inference, rather than also become a type checker that can sport incorrect typing.

Pierre-Sassoulas:
(If we use typing I would do it only when inference failed and assume it's correct instead of re-implementing a type checker entirely) But I think control flow capability is indeed more important. The strength of pylint lie in it's inference, so it make sense to make the inference better before trying to use typing as crutches when inference fails.

DanielNoord:
Indeed! Let's close this: we will likely never close this as it isn't within the current scope of the project and there are far too many things that are within that scope and haven't been fixed yet.

Pierre-Sassoulas:
I see that as a really long term goal. But I think keeping the discussion open permit to point to it when an issue is opened and solvable only by using annotations so we can close the other issue without writting a novel.

ollie-iterators:
Why does the milestones skip version 4.0.0 for pylint and go from version 3 to version 5?

Pierre-Sassoulas:
It's a long term goal, putting it in the farther milestone is supposed to convey that, but we could also remove the milestone entirely.

ollie-iterators:
[...]

PR #8750, opened 2023-06-07
fix(inconsistent-return-statements): Fix false positive due to pylint not seeing that a method returns `NoReturn`

### LINKED ISSUE #8747: False positive for `inconsistent-return-statements` with method annotated with `NoReturn` return type
### Bug description

`repro.py`
```python
from typing import NoReturn

class Foo:
    def _never_returns() -> NoReturn:
        raise Exception

    def bar(self) -> int:
        try:
            return 1
        except Exception:
            self._never_returns()
```

### Configuration

```ini
Default Config
```

### Command used

```shell
pylint --disable=all --enable=inconsistent-return-statements repro.py
```

### Pylint output

```shell
************* Module repro
repro.py:7:4: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)

Your code has been rated at 8.89/10 (previous run: 0.00/10, +8.89)
```

### Expected behavior

```console
```

### Pylint version

```shell
My pylint configuration is pinned to 


pylint 2.15.8
astroid 2.13.5
Python 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]


But I was able to repro this on the latest released version of pylint (2.17.4 at time of writing).
```

### OS / Environment

```console
$ lsb_release -a
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.2 LTS
Release:	22.04
Codename:	jammy
$ uname -r
5.15.0-72-generic
```

### Additional dependencies

_No response_

---

I did an initial investigation, and _might_ have an idea of what's going on:

* During the `inconsistent-return-statements` lint, there's an attempt made infer whether the return type of a function call is `NoReturn`.

  https://github.com/pylint-dev/pylint/blob/a67ffbdd261dffa01045162506614bd8c1e21be5/pylint/checkers/refactoring/refactoring_checker.py#L2001-L2016

* But `_is_function_def_never_returning` only inspects the return type if the node passed in is a `nodes.FunctionDef`. But calling `node.func.inferred()[0]` on the method call (e.g. `self._never_returns()`) returns the corresponding method definition node as type `astroid.BoundMethod`. Since that isn't a `nodes.FunctionDef`, `_is_function_def_never_returning` never even looks to see if the return type is `NoReturn`.

  https://github.com/pylint-dev/pylint/blob/a67ffbdd261dffa01045162506614bd8c1e21be5/pylint/checkers/refactoring/refactoring_checker.py#L1956-L1957

kdestin:
## Type of Changes

## Description

Closes #8747

### DIFF
--- a/pylint/checkers/refactoring/refactoring_checker.py
+++ b/pylint/checkers/refactoring/refactoring_checker.py
@@ -1998,16 +1998,18 @@ class RefactoringChecker(checkers.BaseTokenChecker):
             next_sibling = next_sibling.next_sibling()
         return False
 
-    def _is_function_def_never_returning(self, node: nodes.FunctionDef) -> bool:
+    def _is_function_def_never_returning(
+        self, node: nodes.FunctionDef | astroid.BoundMethod
+    ) -> bool:
         """Return True if the function never returns, False otherwise.
 
         Args:
-            node (nodes.FunctionDef): function definition node to be analyzed.
+            node (nodes.FunctionDef or astroid.BoundMethod): function definition node to be analyzed.
 
         Returns:
             bool: True if the function never returns, False otherwise.
         """
-        if isinstance(node, nodes.FunctionDef) and node.returns:
+        if isinstance(node, (nodes.FunctionDef, astroid.BoundMethod)) and node.returns:
             return (
                 isinstance(node.returns, nodes.Attribute)
                 and node.returns.attrname == "NoReturn"
--- a/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
+++ b/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
@@ -84,9 +84,9 @@ class ClassUnderTest:
         except ValueError:
             self._does_return_method()
 
-    def bug_pylint_8747_incorrect_annotation_trumps_inference(self, s: str) -> int:
-        """Every return is consistent since pylint does not currently use inference
-        to detect that the NoReturn annotation is incorrect and the function actually returns
+    def bug_pylint_8747_incorrect_annotation(self, s: str) -> int:
+        """Every return is consistent since pylint does not attempt to detect that the
+        NoReturn annotation is incorrect and the function actually returns
         """
         try:
             n = int(s)
--- a/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
+++ b/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
@@ -58,6 +58,9 @@ class ClassUnderTest:
     def _no_return_method(self) -> typing.NoReturn:
         sys.exit(1)
 
+    def _falsely_no_return_method(self) -> typing.NoReturn:
+        return 1
+
     def _does_return_method(self) -> int:
         return 1
 
@@ -80,3 +83,16 @@ class ClassUnderTest:
             return n
         except ValueError:
             self._does_return_method()
+
+    def bug_pylint_8747_underhanded(self, s: str) -> int:
+        """Every return is not consistent because self._falsely_no_return_method() returns a value.
+
+        Known false negative because we take the typing into account and not inference
+        contrary to what we usually do."""
+        try:
+            n = int(s)
+            if n < 1:
+                raise ValueError
+            return n
+        except ValueError:
+            self._falsely_no_return_method()
--- a/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
+++ b/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
@@ -84,11 +84,10 @@ class ClassUnderTest:
         except ValueError:
             self._does_return_method()
 
-    def bug_pylint_8747_underhanded(self, s: str) -> int:
-        """Every return is not consistent because self._falsely_no_return_method() returns a value.
-
-        Known false negative because we take the typing into account and not inference
-        contrary to what we usually do."""
+    def bug_pylint_8747_incorrect_annotation_trumps_inference(self, s: str) -> int:
+        """Every return is consistent since pylint does not currently use inference
+        to detect that the NoReturn annotation is incorrect and the function actually returns
+        """
         try:
             n = int(s)
             if n < 1:
--- a/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
+++ b/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
@@ -53,3 +53,30 @@ def bug_pylint_4122_bis(s):
         return n
     except ValueError:
         parser_error_name('parser error')
+
+class ClassUnderTest:
+    def _no_return_method(self) -> typing.NoReturn:
+        sys.exit(1)
+
+    def _does_return_method(self) -> int:
+        return 1
+
+    def bug_pylint_8747(self, s: str) -> int:
+        """Every return is consistent because self._no_return_method hints NoReturn"""
+        try:
+            n = int(s)
+            if n < 1:
+                raise ValueError
+            return n
+        except ValueError:
+            self._no_return_method()
+
+    def bug_pylint_8747_wrong(self, s: str) -> int:  # [inconsistent-return-statements]
+        """Every return is not consistent because self._does_return_method() returns a value"""
+        try:
+            n = int(s)
+            if n < 1:
+                raise ValueError
+            return n
+        except ValueError:
+            self._does_return_method()
--- a/doc/whatsnew/fragments/8747.false_positive
+++ b/doc/whatsnew/fragments/8747.false_positive
@@ -1,0 +1,3 @@
+Fix a false positive where pylint was ignoring method calls annotated as ``NoReturn`` during the ``inconsistent-return-statements`` check.
+
+Closes #8747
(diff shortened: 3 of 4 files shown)


pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1998,16 +1998,18 @@ def _has_return_in_siblings(node: nodes.NodeNG) -> bool:
             next_sibling = next_sibling.next_sibling()
         return False
 
-    def _is_function_def_never_returning(self, node: nodes.FunctionDef) -> bool:
+    def _is_function_def_never_returning(
+        self, node: nodes.FunctionDef | astroid.BoundMethod
```
Pierre-Sassoulas, line 2002:
```suggestion
        self, node: nodes.FunctionDef | nodes.BoundMethod
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1998,16 +1998,18 @@ def _has_return_in_siblings(node: nodes.NodeNG) -> bool:
...
+    ) -> bool:
         """Return True if the function never returns, False otherwise.
 
         Args:
-            node (nodes.FunctionDef): function definition node to be analyzed.
+            node (nodes.FunctionDef or astroid.BoundMethod): function definition node to be analyzed.
```
Pierre-Sassoulas, line 2007:
```suggestion
            node (nodes.FunctionDef or nodes.BoundMethod): function definition node to be analyzed.
```

pylint/checkers/refactoring/refactoring_checker.py
```diff
@@ -1998,16 +1998,18 @@ def _has_return_in_siblings(node: nodes.NodeNG) -> bool:
...
 
         Returns:
             bool: True if the function never returns, False otherwise.
         """
-        if isinstance(node, nodes.FunctionDef) and node.returns:
+        if isinstance(node, (nodes.FunctionDef, astroid.BoundMethod)) and node.returns:
```
Pierre-Sassoulas, line 2012:
```suggestion
        if isinstance(node, (nodes.FunctionDef, nodes.BoundMethod)) and node.returns:
```

tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
```diff
@@ -53,3 +53,30 @@ def bug_pylint_4122_bis(s):
         return n
     except ValueError:
         parser_error_name('parser error')
+
+class ClassUnderTest:
+    def _no_return_method(self) -> typing.NoReturn:
+        sys.exit(1)
```
Pierre-Sassoulas, line 59:
```suggestion
    def _no_return_method(self) -> typing.NoReturn:
        sys.exit(1)
        
    def _falsely_no_return_method(self) -> typing.NoReturn:
        return 1
```

tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
```diff
@@ -53,3 +53,30 @@ def bug_pylint_4122_bis(s):
...
+    def bug_pylint_8747(self, s: str) -> int:
+        """Every return is consistent because self._no_return_method hints NoReturn"""
+        try:
+            n = int(s)
+            if n < 1:
+                raise ValueError
+            return n
+        except ValueError:
+            self._no_return_method()
+
```
Pierre-Sassoulas, line 73:
```suggestion

        def bug_pylint_8747_underhanded(self, s: str) -> int:  # [inconsistent-return-statements]
        """Every return is not consistent because self._falsely_no_return_method() returns a value"""
        try:
            n = int(s)
            if n < 1:
                raise ValueError
            return n
        except ValueError:
            self._falsely_no_return_method()
```

Pierre-Sassoulas (commented):
Looks great, thank you for contributing !

pylint/checkers/refactoring/refactoring_checker.py
kdestin: Applying this suggestion raises `AttributeError: module 'astroid.nodes' has no attribute 'BoundMethod'`, with: 

```console
$ pylint --version
pylint 3.0.0b1
astroid 3.0.0a4
Python 3.8.16 (default, Jun  7 2023, 02:00:15) 
[GCC 11.3.0]
```

tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
kdestin: This suggested test may be revealing a separate false negative?

Pylint _doesn't_ raise `inconsistent-return-statements` in this case.

pylint/checkers/refactoring/refactoring_checker.py
Pierre-Sassoulas: My bad, not sure if we import with ``bases.B...`` elsewhere in the code though. nvm

tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
Pierre-Sassoulas: It's surprising that's it's the existing behavior because usually we do not take typing into account and use inference instead (#4813) -- it was the comment I was going to make by reading the title.  I upgraded the suggestion, I guess we can simply add this test case to document the false negative then.

kdestin: Resolved by https://github.com/pylint-dev/pylint/pull/8750/commits/ab7679406c63111bcda01bcdeb92b7a08064e33d

tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
kdestin: Resolved by https://github.com/pylint-dev/pylint/pull/8750/commits/ab7679406c63111bcda01bcdeb92b7a08064e33d


See https://github.com/pylint-dev/pylint/pull/8750#discussion_r1221019911 for full context

Pierre-Sassoulas (approved):
Great first contribution, I think it's fine to merge and backport considering the check is not using inference like we expect most of our checks to do, but ti's a pre-existing caveat. Let's see what others thinks

pylint/checkers/refactoring/refactoring_checker.py
jacobtylerwalls: We do, but for some reason the `astroid...` form is more prevalent. IDK.

tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
```diff
@@ -53,3 +53,46 @@ def bug_pylint_4122_bis(s):
...
+                raise ValueError
+            return n
+        except ValueError:
+            self._does_return_method()
+
+    def bug_pylint_8747_underhanded(self, s: str) -> int:
+        """Every return is not consistent because self._falsely_no_return_method() returns a value.
+
+        Known false negative because we take the typing into account and not inference
+        contrary to what we usually do."""
```
jacobtylerwalls, line 91:
```suggestion
    def bug_pylint_8747_inference_trumps_incorrect_annotation(self, s: str) -> int:
```
I think the behavior as is is correct: why would we want to emit a message here? The slight asymmetry between whether to trust typing is fine with me: we use the typing only for preventing false positives, not when building the confidence to emit a message.

kdestin: I have no strong feelings about this. From my perspective, the behavior captured by that test falls outside the original scope of this PR.

Given that:

* Should I keep that test in the PR?
  * If so, @jacobtylerwalls would you prefer I commit your suggested rename?
* Otherwise, should I remove it? If this is something that requires further discussion/action, an issue could be made to track it.

jacobtylerwalls: Let's keep the test. My suggestion is just to remove all the talk about the false negative and let the test stand for itself as expected behavior 👍

kdestin: Resolved by https://github.com/pylint-dev/pylint/pull/8750/commits/3c9a77c45689563f51f631ced8f3cfde0ad8c88b

tests/functional/i/inconsistent/inconsistent_returns_noreturn.py
```diff
@@ -84,11 +84,10 @@ def bug_pylint_8747_wrong(self, s: str) -> int:  # [inconsistent-return-statemen
         except ValueError:
             self._does_return_method()
 
-    def bug_pylint_8747_underhanded(self, s: str) -> int:
-        """Every return is not consistent because self._falsely_no_return_method() returns a value.
-
-        Known false negative because we take the typing into account and not inference
-        contrary to what we usually do."""
+    def bug_pylint_8747_incorrect_annotation_trumps_inference(self, s: str) -> int:
+        """Every return is consistent since pylint does not currently use inference
```
jacobtylerwalls, line 88: This is good enough for now. I would have just removed the whole docstring, as it's kind of confusing (why would need to use inference? we just tell from the AST node that it returns something.)

jacobtylerwalls: My fault for suggesting the word inference in the test name. Do you mind if I push a trivial change to the test name? We're not paying attention to the incorrect annotation, so we shouldn't say that it "trumps" anything.

kdestin: Potentially resolved by https://github.com/pylint-dev/pylint/pull/8750/commits/d9738b74d3846ab616c3b8b93fdb4df72a5ec003 ?

jacobtylerwalls (approved):
Thank you!

jacobtylerwalls:
(The documentation check failure is unrelated; I'll look into it.)

